PR #234, opened 2020-08-11
Add Double and Int Convenience Properties

vermont42:
Building on the [work](https://github.com/apple/swift-syntax/pull/148/files) of @ahoppen on [SR-11580](https://bugs.swift.org/browse/SR-11580), this PR adds convenience properties to get `Double` and `Int` values from `FloatLiteralExprSyntax` and `IntegerLiteralExprSyntax`, respectively.

The two new properties, `floatingValue` and `intValue`, use failable initializers of `Double` and `Int`: `init?<S>(_ text: S) where S : StringProtocol` and `init?<S>(_ text: S, radix: Int) where S : StringProtocol`. These initializers fail if the `String`s passed in do not represent valid `Double`s or `Int`s. The question arises: what should `floatingValue` and `intValue` return, if anything, if initialization fails? I considered the following possibilities, ultimately choosing the fourth.

1. `floatingValue` could return `Double.nan`. This would make sense because, for example, ``"🍕.🌍"`` is truly not a number, floating-point or otherwise. The problem with this approach is that there is no corresponding `Int.nan`, which would prevent `floatingValue` and `intValue` from returning the same sort of value given invalid input.  `Decimal` _does_ have a [`nan`](https://developer.apple.com/documentation/foundation/decimal/1780019-nan) value, so if `intValue` returned a `Decimal` rather than an `Int`, that would solve the problem, but I, as a consumer of the API, would find this counter-intuitive because `"42"` in a source file represents an `Int`, not a `Decimal`.

2. The return types of `floatingValue` and `intValue` could be optional. If `Double` or `Int` initialization fails, `floatingValue` and `intValue` could therefore return `nil`. I preferred not to impose on API consumers the burden of unwrapping.

3. The return types of `floatingValue` and `intValue` could be `Result<Double, LiteralError>` and `Result<Int, LiteralError>`, respectively. If `Double` or `Int` initialization fails, `floatingValue` and `intValue` could return `.failure(.invalid)`. I preferred not to impose on API consumers the burden of `switch`ing on the returned value.

[...]

PR #148, opened 2019-10-03
Add convenience methods to read the value of float and int literals

ahoppen:
I was trying to get the integer/float value of integer/float literals and was surprised to find there is no method for it and you need to parse the value from the characters in the digits token yourself. I feel like there should be a convenience method for that. 

If you disagree and don't think we should increase the API surface for these kinds of methods, I‘m also happy with the decision.

PR #239, opened 2020-09-18
Add Double and Int Convenience Properties

vermont42:
Building on the [work](https://github.com/apple/swift-syntax/pull/148/files) of @ahoppen on [SR-11580](https://bugs.swift.org/browse/SR-11580), this PR adds convenience properties to get `Double` and `Int` values from `FloatLiteralExprSyntax` and `IntegerLiteralExprSyntax`, respectively.

The two new properties, `floatingValue` and `intValue`, use failable initializers of `Double` and `Int`: `init?<S>(_ text: S) where S : StringProtocol` and `init?<S>(_ text: S, radix: Int) where S : StringProtocol`. These initializers fail if the `String`s passed in do not represent valid `Double`s or `Int`s. The question arises: what should `floatingValue` and `intValue` return, if anything, if initialization fails? I considered the following possibilities, ultimately choosing the fourth.

1. `floatingValue` could return `Double.nan`. This would make sense because, for example, ``"🍕.🌍"`` is truly not a number, floating-point or otherwise. The problem with this approach is that there is no corresponding `Int.nan`, which would prevent `floatingValue` and `intValue` from returning the same sort of value given invalid input.  `Decimal` _does_ have a [`nan`](https://developer.apple.com/documentation/foundation/decimal/1780019-nan) value, so if `intValue` returned a `Decimal` rather than an `Int`, that would solve the problem, but I, as a consumer of the API, would find this counter-intuitive because `"42"` in a source file represents an `Int`, not a `Decimal`.

2. The return types of `floatingValue` and `intValue` could be optional. If `Double` or `Int` initialization fails, `floatingValue` and `intValue` could therefore return `nil`. I preferred not to impose on API consumers the burden of unwrapping.

3. The return types of `floatingValue` and `intValue` could be `Result<Double, LiteralError>` and `Result<Int, LiteralError>`, respectively. If `Double` or `Int` initialization fails, `floatingValue` and `intValue` could return `.failure(.invalid)`. I preferred not to impose on API consumers the burden of `switch`ing on the returned value.

4. The result of the `Double` or `Int` initialization can be force-unwrapped. This causes a crash if initialization fails. Crashing is [considered harmful](https://homepages.cwi.nl/~storm/teaching/reader/Dijkstra68.pdf), though [not](https://t.co/30VaX2WWHg?amp=1) by everyone. Force-unwrapping is the most convenient option for API consumers, which is why I chose it. I observe that force-unwrapping does appear elsewhere in SwiftSyntax, for example [here](https://github.com/apple/swift-syntax/blob/27bc2a11a7723dfe9bc5bda3caee49d5f5e2df59/Sources/SwiftSyntax/Diagnostic.swift#L173), [here](https://github.com/apple/swift-syntax/blob/27bc2a11a7723dfe9bc5bda3caee49d5f5e2df59/Sources/SwiftSyntax/SyntaxOtherNodes.swift#L97), [here](https://github.com/apple/swift-syntax/blob/27bc2a11a7723dfe9bc5bda3caee49d5f5e2df59/Sources/SwiftSyntax/SyntaxData.swift#L308), and [here](https://github.com/apple/swift-syntax/blob/27bc2a11a7723dfe9bc5bda3caee49d5f5e2df59/Sources/SwiftSyntax/Syntax.swift#L326), and I suspect that the authors of this code made similar judgments about the convenience of force-unwrapping versus the crashing risk it imposes.

Feedback on this code occurred in [this PR](https://github.com/apple/swift-syntax/pull/234), which I closed due to a `rebase` error described [here](https://github.com/apple/swift-syntax/pull/234#issuecomment-694642590).

### DIFF
--- a/Sources/SwiftSyntax/SyntaxConvenienceMethods.swift
+++ b/Sources/SwiftSyntax/SyntaxConvenienceMethods.swift
@@ -1,0 +1,80 @@
+//===- SyntaxConvenienceMethods.swift - Convenience funcs for syntax nodes ===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2020 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+public extension FloatLiteralExprSyntax {
+  var floatingValue: Double {
+    return potentialFloatingValue!
+  }
+
+  fileprivate var potentialFloatingValue: Double? {
+    let floatingDigitsWithoutUnderscores = floatingDigits.text.filter {
+      $0 != "_"
+    }
+    return Double(floatingDigitsWithoutUnderscores)
+  }
+}
+
+public extension IntegerLiteralExprSyntax {
+  var integerValue: Int {
+    return potentialIntegerValue!
+  }
+
+  fileprivate var potentialIntegerValue: Int? {
+    let text = digits.text
+    let (prefixLength, radix) = IntegerLiteralExprSyntax.prefixLengthAndRadix(text: text)
+    let digitsStartIndex = text.index(text.startIndex, offsetBy: prefixLength)
+    let textWithoutPrefix = text.suffix(from: digitsStartIndex)
+
+    let textWithoutPrefixOrUnderscores = textWithoutPrefix.filter {
+      $0 != "_"
+    }
+
+    return Int(textWithoutPrefixOrUnderscores, radix: radix)
+  }
+
+  private static func prefixLengthAndRadix(text: String) -> (Int, Int) {
+    let nonDecimalPrefixLength = 2
+
+    let binaryPrefix = "0b"
+    let octalPrefix = "0o"
+    let decimalPrefix = ""
+    let hexadecimalPrefix = "0x"
+
+    let binaryRadix = 2
+    let octalRadix = 8
+    let decimalRadix = 10
+    let hexadecimalRadix = 16
+
+    switch String(text.prefix(nonDecimalPrefixLength)) {
+    case binaryPrefix:
+      return (binaryPrefix.count, binaryRadix)
+    case octalPrefix:
+      return (octalPrefix.count, octalRadix)
+    case hexadecimalPrefix:
+      return (hexadecimalPrefix.count, hexadecimalRadix)
+    default:
+      return (decimalPrefix.count, decimalRadix)
+    }
+  }
+}
+
+public extension IntegerLiteralExprSyntax {
+  var isValid: Bool {
+    potentialIntegerValue != nil
+  }
+}
+
+public extension FloatLiteralExprSyntax {
+  var isValid: Bool {
+    potentialFloatingValue != nil
+  }
+}
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,6 +4,8 @@ Note: This is in reverse chronological order, so newer entries are added to the 
 
 ## Swift 5.3
 
+* Introduced `integerValue` and `floatingValue` properties to `IntegerLiteralExprSyntax` and `FloatLiteralExprSyntax`, respectively. Converted their `digits` and `floatingDigits` setters, respectively, into throwing functions.
+
 * Introduced `FunctionCallExprSyntax.additionalTrailingClosures` property with type `MultipleTrailingClosureElementListSyntax?` for supporting [SE-0279 Multiple Trailing Closures](https://github.com/apple/swift-evolution/blob/master/proposals/0279-multiple-trailing-closures.md).
 
 * Introduced `syntaxNodeType` property for all types conforming to `SyntaxProtocol`, which returns the underlying syntax node type. It is primarily intended as a debugging aid during development.
--- a/README.md
+++ b/README.md
@@ -181,7 +181,7 @@ swift/utils/build-script --swiftsyntax --swiftpm --llbuild -t --skip-test-cmark 
 ```
 This command will build SwiftSyntax and all its dependencies, tell the build script to run tests, but skip all tests but the SwiftSyntax tests.
 
-Note that it is not currently supported to SwiftSyntax while building the Swift compiler using Xcode.
+Note that it is not currently supported to build SwiftSyntax while building the Swift compiler using Xcode.
 
 ### CI Testing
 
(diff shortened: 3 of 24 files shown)


ahoppen:
https://github.com/apple/swift/pull/33948

@swift-ci Please test

akyrtzi:
The failures are because the generated files differ:
```
***************
*** 1080 ****
!   public init?(_ build: (inout FloatLiteralExprSyntaxBuilder) -> Void) {
--- 1080 ----
!   public init(_ build: (inout FloatLiteralExprSyntaxBuilder) -> Void) {
```
Does the swift repo PR need update?

ahoppen:
https://github.com/apple/swift/pull/33948

@swift-ci Please test

@vermont42 Could you squash the commits again?

ahoppen:
https://github.com/apple/swift/pull/33948

@swift-ci Please test

vermont42:
@ahoppen I will do so, but I just realized that because I was using a pre-concurrency version of `ExprNodes.py`, my generated files were incorrect. I am regenerating.

ahoppen:
Ah, I see. Let me know when you have regenerated the files and I will trigger CI again.

vermont42:
@akyrtzi That is my conclusion as well. I modified those Swift-repo GYB files repo more than a month ago and didn't notice until this morning that they had changed in upstream. That said, I have verified that my current Swift [PR](https://github.com/apple/swift/pull/33948) is correct in that it includes the recent concurrency changes.

ahoppen:
@vermont42 What you need to do, is to rebase your swift PR so that the version in your branch contains both the concurrency changes as well as your changes and then re-generate the files for this PR. If you look at the diff for the generated files in this PR they should contain no concurrency-related changes.

vermont42:
https://github.com/apple/swift/pull/33948

@swift-ci Please test

I fixed the generated files to restore the recent concurrency enhancements, and tests passed locally. In the beginning of this comment, I attempt to invoke CI, not knowing whether t have permission to do so. (No prob if I don't.) When CI passes as a result of this invocation or someone else's, I will squash commits.

akyrtzi:
apple/swift#33948

@swift-ci Please test

rintaro (approved):
Looks good to me

Sources/SwiftSyntax/SyntaxConvenienceMethods.swift
```diff
@@ -0,0 +1,80 @@
...
+    let text = digits.text
+    let (prefixLength, radix) = IntegerLiteralExprSyntax.prefixLengthAndRadix(text: text)
+    let digitsStartIndex = text.index(text.startIndex, offsetBy: prefixLength)
+    let textWithoutPrefix = text.suffix(from: digitsStartIndex)
+
+    let textWithoutPrefixOrUnderscores = textWithoutPrefix.filter {
+      $0 != "_"
+    }
+
+    return Int(textWithoutPrefixOrUnderscores, radix: radix)
```
xwu, line 41: I'm sorry to have missed the initial review, but this would crash for valid integer literals that are too large to be represented in `Int`. Is this intentional? If so, shouldn't this API be `intValue` and the floating-point API `doubleValue` and not `integerValue`/`floatingValue` (the latter implying that any valid integer or floating-point value would be returned)?

vermont42: According to the `Int` initializer's [documentation](https://github.com/apple/swift/blob/103961a7d5f8853ac518f4d8151de0f5b64d8ccf/stdlib/public/core/IntegerParsing.swift#L136), "if the value [that the `String` parameter] denotes in the given `radix` is not representable, the result is `nil`." I confirmed this by running the following code:
```
  let goodInt = Int("42", radix: 10)
  let badInt = Int("42424242424242424242424242424242424242424242424242424242424242424242424242", radix: 10)
  print("@@@ \(goodInt) @@@ \(badInt) @@@")
```
The output was:
```
@@@ Optional(42) @@@ nil @@@
```

vermont42:
Many thanks to Xiaodi, Rintaro, Harlan, and Αργύριος for their thoughtful review and input on the PR. _Extra_ thanks to Alex for opening the starter bug, creating the initial implementation, providing me the GYB strategy, and answering my questions.

Sources/SwiftSyntax/SyntaxConvenienceMethods.swift
xwu: Right, so force unwrapping that value will crash. Is this the intended behavior here?

vermont42: A valid `IntegerLiteralExprSyntax` would not cause a crash.

The only way a crash would occur, via force unwrap, is if a client called `integerValue` on an invalid `IntegerLiteralExprSyntax`. However, there are guardrails in place to prevent clients from creating invalid `IntegerLiteralExprSyntax`es. Those guardrails, not initially present in the PR, were the primary focus of feedback.

xwu: I might be misunderstanding then. My assumption was that 'IntegerLiteralExprSyntax' can represent any valid integer literal? If so, then calling 'integerValue' on some valid 'IntegerLiteralExprSyntax' values would crash, because not all valid integer literals can be represented as an 'Int'.

vermont42: `IntegerLiteralExprSyntax`es represent `Int` literals in source files. A non-representable `Int` won't compile.

Screen Shot 2020-09-21 at 8 33 20 AM

allevato: This isn't quite accurate; `IntegerLiteralExprSyntax` is not limited to what can fit into the `Int` type; it's any sequence of characters that is syntactically a valid integer literal. The error you're showing occurs later on at the semantic analysis phase, but the compiler and SwiftSyntax will still parse it without error. If we try the following command:

```shell
echo 'let foo = 42424242424242424242424242424242424242424242424242424242424242424242424242' | \
  swiftc -frontend -emit-syntax -
```

we'll see this in the output:

```json
...
{
  "id":6,
  "kind":"IntegerLiteralExpr",
    "layout":[{
      "id":5,
      "tokenKind": {
        "kind":"integer_literal",
        "text":"42424242424242424242424242424242424242424242424242424242424242424242424242"
      },
```

So `potentialIntegerValue` called on this would return `nil`, and `integerValue` would crash.

akyrtzi: I agree with @allevato, this "overflows" error is not even coming from typechecking, it's coming from SIL code generation.
Many thanks to @xwu for catching this, I believe we need to make `integerValue` return an optional.

xwu: Right, integer literals are untyped; they are _not_ “Int literals.”

vermont42: I [now](https://media.giphy.com/media/AuwBPJztsEWkw/giphy.gif) agree with @xwu's concern. Accordingly, I intend to create a story in the bug tracker and raise a follow-up PR.

PR #240, opened 2020-09-22
Implement Additional Feedback on PR for SR-11580

vermont42:
After the [PR](https://github.com/apple/swift-syntax/pull/239) for [SR-11580](https://bugs.swift.org/browse/SR-11580) merged, there was consensus, based on feedback by @xwu, that the new convenience methods `floatingValue` and `integerValue`, should be optional in order to prevent potential crashes. This PR implements that change.

This PR also implements new names, suggested by @xwu, for `floatingValue` and `integerValue`: `floatLiteralValue` and `integerLiteralValue`. These names LGTM.

@xwu also [stated](https://github.com/apple/swift-syntax/pull/239#discussion_r492173603), "Ideally, [the convenience methods] would be generic over `T: ExpressibleByIntegerLiteral` and `T: ExpressibleByFloatLiteral` so users can get the value represented in any type they want." I am not a generics [power](https://giphy.com/gifs/zCv1NuGumldXa/html5) user, but I am familiar with their use in the context of, for example, making a `Node` type generic over an `Element` type. I was unable to translate this understanding of the purpose and use of generics into an implementation of @xwu's suggestion. I would welcome a suggested new signature for `floatLiteralValue` or `integerLiteralValue`.

### DIFF
--- a/Sources/SwiftSyntax/SyntaxConvenienceMethods.swift
+++ b/Sources/SwiftSyntax/SyntaxConvenienceMethods.swift
@@ -11,24 +11,20 @@
 //===----------------------------------------------------------------------===//
 
 public extension FloatLiteralExprSyntax {
-  var floatingValue: Double {
-    return potentialFloatingValue!
-  }
-
-  fileprivate var potentialFloatingValue: Double? {
+  var floatLiteralValue: Double? {
     let floatingDigitsWithoutUnderscores = floatingDigits.text.filter {
       $0 != "_"
     }
     return Double(floatingDigitsWithoutUnderscores)
   }
+
+  var isValid: Bool {
+    floatLiteralValue != nil
+  }
 }
 
 public extension IntegerLiteralExprSyntax {
-  var integerValue: Int {
-    return potentialIntegerValue!
-  }
-
-  fileprivate var potentialIntegerValue: Int? {
+  var integerLiteralValue: Int? {
     let text = digits.text
     let (prefixLength, radix) = IntegerLiteralExprSyntax.prefixLengthAndRadix(text: text)
     let digitsStartIndex = text.index(text.startIndex, offsetBy: prefixLength)
@@ -65,16 +61,8 @@ public extension IntegerLiteralExprSyntax {
       return (decimalPrefix.count, decimalRadix)
     }
   }
-}
 
-public extension IntegerLiteralExprSyntax {
   var isValid: Bool {
-    potentialIntegerValue != nil
-  }
-}
-
-public extension FloatLiteralExprSyntax {
-  var isValid: Bool {
-    potentialFloatingValue != nil
+    integerLiteralValue != nil
   }
 }
--- a/Tests/SwiftSyntaxTest/SyntaxConvenienceMethodsTests.swift
+++ b/Tests/SwiftSyntaxTest/SyntaxConvenienceMethodsTests.swift
@@ -2,14 +2,16 @@ import XCTest
 import SwiftSyntax
 
 public class SyntaxConvenienceMethodsTests: XCTestCase {
-
-  public func testFloatingValues() {
-    testFloatingValue(text: "5.3_8", expectedValue: 5.38)
-    testFloatingValue(text: "12e3", expectedValue: 12000.0)
-    testFloatingValue(text: "32E1", expectedValue: 320.0)
-    testFloatingValue(text: "0xdEFACE.C0FFEEp+1", expectedValue: 0xdEFACE.C0FFEEp+1)
-    testFloatingValue(text: "0xaffab1e.e1fP-2", expectedValue: 0xaffab1e.e1fP-2)
-    testFloatingValue(text: "🥥", expectedValue: nil)
+  public func testFloatValues() {
+    testFloatValue(text: "5.3_8", expectedValue: 5.38)
+    testFloatValue(text: "12e3", expectedValue: 12000.0)
+    testFloatValue(text: "32E1", expectedValue: 320.0)
+    testFloatValue(text: "0xdEFACE.C0FFEEp+1", expectedValue: 0xdEFACE.C0FFEEp+1)
+    testFloatValue(text: "0xaffab1e.e1fP-2", expectedValue: 0xaffab1e.e1fP-2)
+    testFloatValue(text: "🥥", expectedValue: nil)
+    // The following results in a valid FloatLiteralExprSyntax, but because Double.greatestFiniteMagnitude
+    // is 1.7976931348623157e+308, floatLiteralValue should be nil.
+    testFloatValue(text: "1.7976931348623157e+309", expectedValue: nil)
   }
 
   public func testIntegerValues() {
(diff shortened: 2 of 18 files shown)


Sources/SwiftSyntax/SyntaxConvenienceMethods.swift
```diff
@@ -11,24 +11,20 @@
 //===----------------------------------------------------------------------===//
 
 public extension FloatLiteralExprSyntax {
-  var floatingValue: Double {
-    return potentialFloatingValue!
-  }
-
-  fileprivate var potentialFloatingValue: Double? {
```
vermont42, line 18: One benefit of this PR is that this property is no longer needed.

Tests/SwiftSyntaxTest/SyntaxConvenienceMethodsTests.swift
```diff
@@ -18,25 +20,36 @@ public class SyntaxConvenienceMethodsTests: XCTestCase {
...
+    if let expectedValue = expectedValue {
+      XCTAssertEqual(literalExpr.integerLiteralValue!, expectedValue)
+    } else {
+      XCTAssertNil(literalExpr.integerLiteralValue)
+    }
   } else {
-    XCTAssertNil(literalExpr)
+    XCTAssertNil(expectedValue)
   }
-}
```
vermont42, line 42: `git diff` called out this absence of a terminating newline, so I added one.

ahoppen (commented):
After a quick scan of the PR, I’ve got two directions in which this could go:

### 1. The properties don’t need to be optional at all

AFAICT, the `ExpressibleByFloatLiteral` protocol has a requirement for an initialiser that takes a `ExpressibleByFloatLiteral.FloatLiteralType`. That type needs to conform to `_ExpressibleByBuiltinFloatLiteral` (see [here](https://github.com/apple/swift/blob/main/stdlib/public/core/CompilerProtocols.swift#L344)). However, the only types that conform to `_ExpressibleByBuiltinFloatLiteral` according to [here](https://github.com/apple/swift/blob/main/utils/SwiftFloatingPointTypes.py) and [here](https://github.com/apple/swift/blob/main/stdlib/public/core/FloatingPointTypes.swift.gyb) are `Float16`, `Float`, `Double` and `Float80`. So we cannot have a floating point literal that doesn’t fit in a `Float80`. 

I know that this is a semantic property, but I would say that for the sake of usability, it would be alright to harness this semantic property. What do you think @xwu  and @akyrtzi?

For Int the argument is similar with the only difference that we need to take into account signed and unsigned integers (`Int` and `UInt` being the biggest types). But I think we can tackle that problem by returning a type like the following from the `intLiteralValue` property and using either the `signed` or `unsigned` case depending on whether the literal starts with a `-` or not:

```swift
enum IntLiteralValue {
  case signed(Int)
  case unsigned(UInt)

  var signedValue: Int? { /* … */ }
  var unsignedValue: Int? { /* … */ }
}
```

### 2. We don’t need the `must_uphold_invariant` thing anymore

If we decide that we want to make the `floatLiteralValue` and `intLiteralValue` properties optional, we don’t need the entire `must_uphold_invariant` thing anymore. 

In fact, as far as it’s implemented right now, there are currently two issues with the invariant:
a) It is no longer true that every syntactically valid literal upholds its invariant because it might not fit into the property’s type
b) We cannot set the contents of a float literal to anything that’s larger than the property’s type.

akyrtzi:
> I know that this is a semantic property, but I would say that for the sake of usability, it would be alright to harness this semantic property. What do you think @xwu and @akyrtzi?

That would require that we change the parser to make overflowing literals syntactically invalid. But I don't think that is appropriate, I'd recommend that we adhere to what is valid syntax and avoid bringing in semantic/code-generation concepts.

xwu:
It'll take me until the weekend to write up my feedback for this PR--sorry for the delay. Overall, it's going to be a hair more complicated than what's presented here.

@ahoppen Both `_ExpressibleByBuiltinFloatLiteral` and which types conform to it are internal implementation details (and unfortunate ones at that) of the standard library.

We can very much have a valid float literal that doesn't fit in a `Float80`; it leads to double-rounding error in today's Swift (see bugs [SR-920](https://bugs.swift.org/browse/SR-920), [SR-3317](https://bugs.swift.org/browse/SR-3317), [SR-3970](https://bugs.swift.org/browse/SR-3970), [SR-7124](https://bugs.swift.org/browse/SR-7124), among others) and is why using float literals with `Foundation.Decimal` is a footgun. If necessary, using this _implementation_ detail (i.e., bug) for an _implementation_ here that could be kept in sync with the standard library should be fine, but using this implementation detail for the _interface_ here is doubleplus ungood.

For integer literals, `Int` is not the widest built-in integer literal type (on 32-bit platforms, it is `Int64`); additionally, `DoubleWidth` could be brought back in the future, at which time there will be no limit to the widest built-in integer literal type. Meanwhile, the underlying representation of integer literals was long ago increased from 64 bits to 2048 bits, and then even that was superseded by an arbitrary-precision type in time for ABI stability; in other words, there truly is no limit to how big an integer literal can be.

The purpose of all of this work (including any future overhaul of `ExpressibleByFloatLiteral` to represent binary and decimal floating-point values losslessly with arbitrary precision) is precisely to make it so that float and integer literals are suitable for use with any floating-point or integer type, even third-party arbitrary precision ones. That is to say, in order to fulfill their semantic purpose, they must be able to represent values that are unrepresentable in _any particular_ conforming type, or for that matter unrepresentable even in _all available_ conforming types.

ahoppen:
On second thought, I quite agree with @xwu  and @akyrtzi that we shouldn’t rely on semantic functionality. AFAICT it should be sufficient if we make the properties optional and remove the `must_uphold_invariant` thing as outlined above. 

I’m sorry for all the work you have put into this, @vermont42, only so it needs to be scrapped now…

akyrtzi:
@xwu would you be fine with making these simple changes for now (make the properties optional and remove `must_uphold_invariant`) so that at least we avoid potential crashes in the interface, and if you are able to outline in a forum post what do you recommend as an API I'd really appreciate it 🙏

xwu (changes requested):
Sorry for the delay. Yes, I do have some comments on what an ideal API might be, but that shouldn't hold up this PR because it could be addressed further down the road.

However, I also have some comments on the correctness of the implementation, and I think those _are_ pressing because they affect the raison d'être of this API: Users want to know, and this API purports to tell us, what value is represented by a particular literal expression. If it gives false positives and/or false negatives, then it does not serve that purpose, and it would not be a usable API.

As it happens, I think my comments on the API design are actually fairly straightforward to address, though not pressing, but my comments on the correctness of the implementation are quite hard to address but very much pressing. In fact, the potential crash is actually the _least_ pressing in my view, as in that circumstance at least execution does not proceed with a misleading result; it's where the API returns an inaccurate result _without_ crashing that really makes me worry. I really don't mind if we don't address the entirety of my comment on the API design, but if addressing the other issues prove to be unworkable in the short term, perhaps in the interim we could revert the previous PR?

Sources/SwiftSyntax/SyntaxConvenienceMethods.swift
```diff
@@ -11,24 +11,20 @@
...
-  var integerValue: Int {
-    return potentialIntegerValue!
+  var isValid: Bool {
+    floatLiteralValue != nil
   }
+}
 
-  fileprivate var potentialIntegerValue: Int? {
+public extension IntegerLiteralExprSyntax {
+  var integerLiteralValue: Int? {
```
xwu, line 27: OK, so here's my one comment on API design. I think users will want to know what value is represented by their integer literal expression for types other than the default `Int` and `Double`. It would be very reasonable to know, what's the `UInt32` value represented by a certain literal, for instance.

Therefore, we can use a (generic) method instead of a property to allow the user to pass a variable indicating the desired type:

```swift
func integerLiteralValue<T>(ofType _: T.Type) -> T
where T: ExpressibleByIntegerLiteral { ... }
```

Users could then write: `literalExpr.integerLiteralValue(ofType: Int32.self)`. Now, it's likely that users will most often want a value of the _default_ literal types, so it's reasonable to add that:

```swift
func integerLiteralValue<T>(ofType _: T.Type = IntegerLiteralType) -> T
where T: ExpressibleByIntegerLiteral { ... }
```

Now, the use site simply differs from what you have currently in this PR by a pair of parentheses. Pretty good. Now, if you were to try to implement this API, you'd find that you actually need to constrain `T` a bit further in order to use the String conversion APIs.

It's tempting to write `where T: ExpressibleByIntegerLiteral, T: LosslessStringConvertible`, but in fact this raises a problem: there's no _semantic_ guarantee that the _string format_ from which a type can be losslessly converted is the same as the _literal format_. But in practice, we might be able to get away with it. However, you'd run into another problem when it comes to integer literals, because if you want to pass the `radix` argument, for which you'd need to constrain to `T: FixedWidthInteger` instead. This is fine, but floating-point types are expressible by _integer_ literals but _not_ fixed-width integer types; it'd be difficult to make it possible to convert from binary or octal integer literals to floating-point types without implementing your own conversions from scratch here. This is certainly not necessary for this PR and likely out of scope for this project entirely, but if we ignore floating-point types and integer literals, you can get 80% of the way just by making a few adjustments to the API design.

Tests/SwiftSyntaxTest/SyntaxConvenienceMethodsTests.swift
```diff
@@ -2,14 +2,16 @@ import XCTest
...
-    testFloatingValue(text: "🥥", expectedValue: nil)
+  public func testFloatValues() {
+    testFloatValue(text: "5.3_8", expectedValue: 5.38)
+    testFloatValue(text: "12e3", expectedValue: 12000.0)
+    testFloatValue(text: "32E1", expectedValue: 320.0)
+    testFloatValue(text: "0xdEFACE.C0FFEEp+1", expectedValue: 0xdEFACE.C0FFEEp+1)
+    testFloatValue(text: "0xaffab1e.e1fP-2", expectedValue: 0xaffab1e.e1fP-2)
+    testFloatValue(text: "🥥", expectedValue: nil)
+    // The following results in a valid FloatLiteralExprSyntax, but because Double.greatestFiniteMagnitude
+    // is 1.7976931348623157e+308, floatLiteralValue should be nil.
```
xwu, line 13: It should _not_ be `nil`. If you write `let x = 1.7976931348623157e+309`, you get an actual `Double` which is `infinity`. This is one of those divergences I discuss above between the string conversion APIs and the float literal APIs.

vermont42: The last commit was about removing `must_uphold_invariant` and verifying that unit tests still pass. I have not yet explored additional changes.

shahmishal:
The Swift project moved the default branch to `main` and deleted `master` branch, so GitHub automatically closed the PR. Please re-create the pull request with `main` branch.

More detail about the branch update - https://forums.swift.org/t/updating-branch-names/40412

PR #242, opened 2020-10-13
Revert "Add Double and Int Convenience Properties (#239)"

rintaro:
This reverts commit #239
This couldn't handle integer literal syntax more than `Int.max` e.g. `0x9e3779b97f4a0000`


rintaro:
https://github.com/apple/swift/pull/34292
@swift-ci Please test

rintaro:
@vermont42 Sorry, I have to revert it for now. This causes an error for valid `UInt64` integer literals. like: https://github.com/apple/swift-argument-parser/blob/d424bca0ec66ec0e3888ebdc118379d348a67da3/Examples/roll/SplitMix64.swift#L19-L25

akyrtzi:
> This causes an error for valid UInt64 integer literals

To be more clear, this was causing a `fatalError` crash.

PR #243, opened 2020-10-19
Make DiagnosticEngine thread safe

keith:
This allows consumers to emit diagnostics from multiple threads.

Primarily motivated by https://github.com/apple/swift-format/pull/117


akyrtzi (approved):

keith:
Thanks!

PR #246, opened 2020-11-04
[Test] Disable a test case  while investigating non-deterministic failure

rintaro:
rdar://problem/71046813


rintaro:
@swift-ci Please test

PR #253, opened 2021-01-13
Add assertions verifying that trailing objects of RawSyntax fit inside RawSyntaxDataElement

ahoppen:
Assert that `RawSyntaxDataElement` holds enough space for each trailing object. This should be satisfied now but might allow us to catch regressions later more easily.


ahoppen:
@swift-ci Please test

akyrtzi (approved):

PR #254, opened 2021-01-15
Make SourceLocation and related types Hashable

MaxDesiatov:
It seems natural to be able to compare source locations for equality, or to add locations to a `Set` or use them as keys in dictionaries. Currently this requires declaring manual `Hashable` conformance in libraries that use SwiftSyntax, which is error-prone.

### DIFF
--- a/Sources/SwiftSyntax/SourceLocation.swift
+++ b/Sources/SwiftSyntax/SourceLocation.swift
@@ -12,7 +12,7 @@
 
 /// Represent the user-facing part of SourceLocation that can be calculated
 /// on demand.
-struct ComputedLocation: Equatable, Codable, CustomDebugStringConvertible {
+struct ComputedLocation: Hashable, Codable, CustomDebugStringConvertible {
   /// The line in the file where this location resides. 1-based.
   let line: Int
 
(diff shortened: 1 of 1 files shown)


MaxDesiatov:
@swift-ci please test

MaxDesiatov:
@swift-ci please test

MaxDesiatov:
@swift-ci please test

akyrtzi (approved):

MaxDesiatov:
Please feel free to merge when you see fit, I don't have commit access here.

Sources/SwiftSyntax/SourceLocation.swift
```diff
@@ -43,7 +43,7 @@ struct ComputedLocation: Codable, CustomDebugStringConvertible {
 }
 
 /// Represents a source location in a Swift file.
-public struct SourceLocation: Codable, CustomDebugStringConvertible {
+public struct SourceLocation: Hashable, Codable, CustomDebugStringConvertible {
```
ahoppen, line 46: Just to be clear: Should two `SourceLocation`s be different if one has a `ComputedLocation` and the other does not? After all, they still refer to the same location in the source code.

I’m not 100% sure on this, but I’m leaning towards making the identity of `SourceLocation` be purely based on the value of `offset`. What do you think, @MaxDesiatov? 

CC: @akyrtzi

MaxDesiatov: In my opinion they shouldn't be equal since `line`, `column`, and `file` will return different results for them. The one without a `ComputedLocation` could refer to a different file altogether, as `file` is inferred from `compLoc`.

ahoppen: That’s a good point. I didn’t consider `file`. With that in mind, I

PR #255, opened 2021-01-20
Minor internal API improvements

ahoppen:
Two minor internal API improvements:

1. Remove `advancedToEndOfChildren`
These methods are not used anywhere and are potentially incorrect (`indexInParent` should probably be initialized with `raw.numberOfChildren - 1`). Remove the until we really need them
2. Rename `advancedBySibling`/`reversedBySibling` to `advancedBy`/`reversedBy`
We are not actually advancing by a sibling but just by a node, simplify the method name to reflect this.


ahoppen:
@swift-ci Please test

akyrtzi (approved):

PR #257, opened 2021-02-02
Add a parsing performance test

ahoppen:
This is a small performance test, testing the time it takes to parse a source file into SwiftSyntax. I intend to use this to make sure we don't regress in performance during the libSyntax migration.


ahoppen:
@swift-ci Please test

akyrtzi (approved):

PR #260, opened 2021-03-03
Adjust for modified C API that only returns a range for token nodes

ahoppen:
Adjustments for the changes to the C API in https://github.com/apple/swift/pull/36249


ahoppen:
https://github.com/apple/swift/pull/36249

@swift-ci Please test

PR #261, opened 2021-03-04
Pass length of source file to C parsing actions

ahoppen:
Previously, the string was implicitly terminated by the first null character. But a source file might contain a null character which shouldn't terminate it.

### DIFF
--- a/Sources/SwiftSyntax/SyntaxParser.swift
+++ b/Sources/SwiftSyntax/SyntaxParser.swift
@@ -192,9 +192,11 @@ public enum SyntaxParser {
       swiftparse_parser_set_diagnostic_handler(c_parser, diagHandler)
     }
 
-    let c_top = swiftparse_parse_string(c_parser, source)
+    let c_top = source.withCString { buf in
+      swiftparse_parse_string(c_parser, buf, source.utf8.count)
+    }
 
-  // Get ownership back from the C parser.
+    // Get ownership back from the C parser.
     return RawSyntax.moveFromOpaque(c_top)!
   }
 }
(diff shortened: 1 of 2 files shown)


ahoppen:
https://github.com/apple/swift/pull/36283

@swift-ci Please test

Sources/SwiftSyntax/SyntaxParser.swift
```diff
@@ -192,7 +192,7 @@ public enum SyntaxParser {
       swiftparse_parser_set_diagnostic_handler(c_parser, diagHandler)
     }
 
-    let c_top = swiftparse_parse_string(c_parser, source)
+    let c_top = swiftparse_parse_string(c_parser, source, source.utf8.count)
```
rintaro, line 195: I believe passing `String` to `const char *` [calls `String.utf8CString`](https://github.com/apple/swift/blob/b6f0ff7a033fc81691d4e9ce24c4e42d724125de/stdlib/public/core/Pointer.swift#L361-L369) which [always copies the buffer](https://github.com/apple/swift/blob/b6f0ff7a033fc81691d4e9ce24c4e42d724125de/stdlib/public/core/StringUTF8View.swift#L253-L272). Probably
```swift
let c_top = source.withUTF8 {
  $0.withMemoryRebound(to: CChar.self) { buf in
    swiftparse_parse_string(c_parser, buf.baseAddress, buf.count)
  }
}
```
is faster.

ahoppen: It needs to be null-terminated, otherwise `lvm::MemoryBuffer::getMemBuffer` is not happy (also we need the terminating null char to lex it as the EOF token).

rintaro: OK, in that case, I suggest
```swift
let c_top = source.withCString { buf in
  swiftparse_parse_string(c_parser, buf.baseAddress, buf.count)
}
```
`withCString` doesn't copy the buffer if the underlying string buffer is null terminated.

ahoppen: Sounds good 👍

akyrtzi (approved):

ahoppen:
Updated to use `withCString` as @rintaro suggested. Nice catch on performance.

apple/swift#36283

@swift-ci Please test

PR #262, opened 2021-03-05
Add extension builder

kimdv:
Hi! 👋

I have been looking a bit in to the SwiftSyntaxBuilder to generate some code. 

I made some things so I thought more people could benefit from it, so I will now try to open a PR! 

I hope it fits the project? 

Thanks ! 🚀

### DIFF
--- a/Sources/SwiftSyntaxBuilder/DeclBuildables.swift
+++ b/Sources/SwiftSyntaxBuilder/DeclBuildables.swift
@@ -185,3 +185,45 @@ public struct Struct: DeclBuildable {
     return DeclSyntax(structDecl)
   }
 }
+
+// MARK: - Extension
+
+public struct Extension: DeclBuildable {
+  let name: String
+  let memberList: DeclListBuildable
+
+  public init(
+    _ name: String,
+    @DeclListBuilder buildMemberList: () -> DeclListBuildable = { DeclList.empty }
+  ) {
+    self.name = name
+    self.memberList = buildMemberList()
+  }
+
+  public func buildDecl(format: Format, leadingTrivia: Trivia) -> DeclSyntax {
+    let extensionKeyword = SyntaxFactory.makeExtensionKeyword().withTrailingTrivia(.spaces(1))
+
+    let declList = memberList.buildDeclList(
+      format: format._indented(),
+      leadingTrivia: .zero
+    )
+
+    let extensionDecl = ExtensionDeclSyntax {
+      $0.useExtensionKeyword(extensionKeyword)
+      $0.useExtendedType(SyntaxFactory.makeTypeIdentifier(name))
+      $0.useMembers(MemberDeclBlockSyntax {
+        $0.useLeftBrace(SyntaxFactory.makeLeftBraceToken().withLeadingTrivia(.spaces(1)))
+        $0.useRightBrace(SyntaxFactory.makeRightBraceToken().withLeadingTrivia(.newlines(1) + format._makeIndent()))
+
+        let format = format._indented()
+        for decl in declList {
+          let member = SyntaxFactory
+            .makeMemberDeclListItem(decl: decl, semicolon: nil)
+            .withLeadingTrivia(.newlines(1) + format._makeIndent())
+          $0.addMember(member)
+        }
+      })
+    }
+    return DeclSyntax(extensionDecl)
+  }
+}
--- a/Tests/SwiftSyntaxBuilderTest/ExtensionTests.swift
+++ b/Tests/SwiftSyntaxBuilderTest/ExtensionTests.swift
@@ -1,0 +1,59 @@
+import XCTest
+import SwiftSyntax
+
+import SwiftSyntaxBuilder
+
+final class ExtensionTests: XCTestCase {
+  func testEmptyExtension() {
+    let leadingTrivia = Trivia.garbageText("␣")
+
+    let buildable = Extension("TestExtension")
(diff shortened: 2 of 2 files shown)


ktoso:
This is awesome! I've long wanted to poke a bit more at this and complete it but sadly unable to make the time...

This looks good but I don't maintain swift syntax so let's ask @akyrtzi or @rintaro for review :-) 

Do you think you'd be interested in filling in the various other builders and perhaps even get this pretty complete, so even functions etc can be declared using it?

ktoso:
@swift-ci please smoke test

kimdv:
@ktoso yes I have some time, and will take a small piece at the same! 
Was thinking about also adding things like `public`, `private`, `static`, `final` and more. 

Maybe I need some guidance and help, but I can give it a try! 😅

ahoppen:
I can only agree with @ktoso: Thanks for working on this!

If you would like to add SyntaxBuilders for all types (which I would absolutely love), I would suggest that we gyb-generate them, instead of hand-writing the changes in the same way that we also generate the `Syntax` nodes [e.g. here](https://github.com/apple/swift-syntax/blob/main/Sources/SwiftSyntax/SyntaxNodes.swift.gyb.template). There are simply too many nodes to write everything by hand and it’ll be fragile to maintain.

If you would like to go that road, I can mentor you through it. Just let me know.

ahoppen (approved):

kimdv:
@ahoppen I'm up to for it! 
It will take some time and I'll ask you A LOT of questions and some of them will also be stupid. 
But if you will mentor I'm ready for it 🎉

First question, I have no clue of where to start looking and figure out how everything works?

ahoppen:
Nice. Feel free to ask as many questions as you want.

So, I’ve got a first task for you 😉:
You already mentioned that you wanted to support `public`, `private`, `static`, `final` etc. While you could add them to `Tokens.swift` by hand, I’d say we automatically generate *all* tokens. Here are a couple of steps you might want to take for this.

### Set up your local development environment to allow gyb-generation
Since you need to generate `gyb` files (those in `Sources/SwiftSyntax/gyb_generated`), you need to checkout the main [swift compiler repo](http://github.com/apple/swift/) next to `swift-syntax`. Your directory tree should now look as follows:
```
swift-src (or whatever you want to call the parent folder)
  - swift
  - swift-syntax
```
(Note: there’s no need to actually build the Swift compiler, you just need to check out the repo)

The way that `gyb` works, is that it loads the syntax definition files from the main Swift repo ([here](https://github.com/apple/swift/tree/main/utils/gyb_syntax_support)) and then generates Swift files from them. Any line in a `.gyb` file that starts with a `%` is a python instruction line and any line that does not start with a `%` gets printed into the generated file. For example [here](https://github.com/apple/swift-syntax/blob/main/Sources/SwiftSyntax/TokenKind.swift.gyb#L23) we are looping over all `SYNTAX_TOKENS` (defined [here](https://github.com/apple/swift/blob/main/utils/gyb_syntax_support/__init__.py#L24)) and are printing a case statement for it. The result looks like [this](https://github.com/apple/swift-syntax/blob/main/Sources/SwiftSyntax/gyb_generated/TokenKind.swift#L16).

With that setup at hand, you should be able to gyb-generate the files yourself. Go to `swift-syntax` and run
```bash
./build-script.py --toolchain /Library/Developer/Toolchains/swift-DEVELOPMENT-SNAPSHOT-XXXXXXXXXX.xctoolchain/usr/ --degyb-only
```
Replace the toolchain by a `main` development snapshot you have installed from [swift.org](https://swift.org/download/#trunk-development-main). To verify that things work, you could try adding a new test keyword to [Token.py](https://github.com/apple/swift/blob/main/utils/gyb_syntax_support/Token.py), re-generate your Swift files and verify that it shows up.

### Enable `gyb`-generation of files inside `Sources/SwiftSyntaxBuilder`
The actual `gyb`-generation in `build-script.py` happens [here](https://github.com/apple/swift-syntax/blob/main/build-script.py#L212). Notice that we currently only loop over `.gyb` files in `Sources/SwiftSyntax`. Since we want to also use `.gyb` files in `SwiftSyntaxBuilder`, you’ll need to duplicate the loop to also iterate over the files in `SwiftSyntaxBuilder`.

### Actually generate `Tokens.swift`
Now we’ve finally reached the stage where we can make meaningful changes 🎉. Rename `Tokens.swift` to `Tokens.swift.gyb` and re-run `gyb`-generation. If everything works, you should have a `Tokens.swift` file in `Sources/SwiftSyntaxBuilder/gyb_genreated` now and SwiftSyntax should still compile.

### Generate all tokens
You should now be able to generate all keyword tokens similar to what is done in the loop in `TokenKinds.swift.gyb`. To figure out the method names in `SyntaxFactory`, you can look up how they are generated [here](https://github.com/apple/swift-syntax/blob/main/Sources/SwiftSyntax/SyntaxFactory.swift.gyb#L60).

The tricky part that we still need to figure out, is how we want to handle formatting. For now, I would suggest, you only generate `Tokens` cases for keywords, since for these at least, we know we want to append a single trailing space. Maybe you’ve got ideas if there is a nice general rule to whitespaces around tokens, depending on whether they are punctuators, keywords etc. Otherwise we could think about adding that information to the `Token.py` or delegate all the formatting to [swift-format](https://github.com/apple/swift-format) (although we’d need to figure out how to resolve the dependency cycle there (swift-format requires swift-syntax, so swift-syntax can’t require swift-format). Let’s maybe postpone that discussion though.

ahoppen:
I’m closing this because it has been superseded by #263.

PR #263, opened 2021-03-05
Add gyb for swift syntax builder keyword tokens

kimdv:
From #262 

Okay! I tried and got something to work here, thanks a lot @ahoppen ! 🎉

### DIFF
--- a/Sources/SwiftSyntaxBuilder/Tokens.swift.gyb
+++ b/Sources/SwiftSyntaxBuilder/Tokens.swift.gyb
@@ -1,8 +1,16 @@
+%{
+  from gyb_syntax_support import *
+  from gyb_syntax_support.kinds import lowercase_first_word
+  # -*- mode: Swift -*-
+  # Ignore the following admonition it applies to the resulting .swift file only
+}%
+//// Automatically Generated From Tokens.swift.gyb.
+//// Do Not Edit Directly!
 //===----------------------------------------------------------------------===//
 //
 // This source file is part of the Swift.org open source project
 //
-// Copyright (c) 2014 - 2019 Apple Inc. and the Swift project authors
+// Copyright (c) 2014 - 2021 Apple Inc. and the Swift project authors
 // Licensed under Apache License v2.0 with Runtime Library Exception
 //
 // See https://swift.org/LICENSE.txt for license information
@@ -16,17 +24,12 @@ import SwiftSyntax
 enum Tokens {
   // MARK: Keywords
 
-  /// `"let "`
-  static let `let` = SyntaxFactory.makeLetKeyword().withTrailingTrivia(.spaces(1))
-
-  /// `"var "`
-  static let `var` = SyntaxFactory.makeVarKeyword().withTrailingTrivia(.spaces(1))
-
-  /// `"import "`
-  static let `import` = SyntaxFactory.makeImportKeyword().withTrailingTrivia(.spaces(1))
-
-  /// `"struct "`
-  static let `struct` = SyntaxFactory.makeStructKeyword().withTrailingTrivia(.spaces(1))
+% for token in SYNTAX_TOKENS:
+%   if token.is_keyword:
+  /// The `${token.text}` keyword
+  static let `${lowercase_first_word(token.name)}` = SyntaxFactory.make${token.name}Keyword().withTrailingTrivia(.spaces(1))
+%   end
+% end
 
   // MARK: Punctuations and Signs
 
--- a/build-script.py
+++ b/build-script.py
@@ -16,6 +16,9 @@ import tempfile
 
 PACKAGE_DIR = os.path.dirname(os.path.realpath(__file__))
 WORKSPACE_DIR = os.path.dirname(PACKAGE_DIR)
+SOURCES_DIR = os.path.join(PACKAGE_DIR, "Sources")
+SWIFTSYNTAX_DIR = os.path.join(SOURCES_DIR, "SwiftSyntax")
+SWIFTSYNTAXBUILDER_DIR = os.path.join(SOURCES_DIR, "SwiftSyntaxBuilder")
 
 LLVM_DIR = os.path.join(WORKSPACE_DIR, "llvm-project", "llvm")
 SWIFT_DIR = os.path.join(WORKSPACE_DIR, "swift")
@@ -164,37 +167,38 @@ def generate_single_gyb_file(
     check_call(rsync_command, verbose=verbose)
 
 
-def generate_gyb_files(gyb_exec, verbose, add_source_locations, destination=None):
+def generate_gyb_files(
+    gyb_exec, verbose, add_source_locations, 
+    swiftsyntax_destination=None, swiftsyntaxbuilder_destination=None
+):
     print("** Generating gyb Files **")
 
     check_gyb_exec(gyb_exec)
     check_rsync()
 
-    swiftsyntax_sources_dir = os.path.join(PACKAGE_DIR, "Sources", "SwiftSyntax")
     temp_files_dir = tempfile.gettempdir()
 
-    if destination is None:
-        destination = os.path.join(swiftsyntax_sources_dir, "gyb_generated")
+    if swiftsyntax_destination is None:
+        swiftsyntax_destination = os.path.join(
+            SWIFTSYNTAX_DIR, "gyb_generated")
+    
+    if swiftsyntaxbuilder_destination is None:
+        swiftsyntaxbuilder_destination = os.path.join(
+            SWIFTSYNTAXBUILDER_DIR, "gyb_generated")
 
-    template_destination = os.path.join(destination, "syntax_nodes")
+    template_destination = os.path.join(swiftsyntax_destination, "syntax_nodes")
 
-    if not os.path.exists(temp_files_dir):
-        os.makedirs(temp_files_dir)
-    if not os.path.exists(destination):
-        os.makedirs(destination)
-    if not os.path.exists(template_destination):
-        os.makedirs(template_destination)
+    make_dir_if_needed(temp_files_dir)
+    make_dir_if_needed(swiftsyntax_destination)
+    make_dir_if_needed(swiftsyntaxbuilder_destination)
+    make_dir_if_needed(template_destination)
 
     # Clear any *.swift files that are relics from the previous run.
-    for previous_gyb_gen_file in os.listdir(destination):
-        if previous_gyb_gen_file.endswith(".swift"):
-            gyb_file = os.path.join(
-                swiftsyntax_sources_dir, previous_gyb_gen_file + ".gyb"
-            )
-            if not os.path.exists(gyb_file):
-                check_call(
-                    ["rm", previous_gyb_gen_file], cwd=destination, verbose=verbose
-                )
+    clear_gyb_files_from_previous_run(
+        SWIFTSYNTAX_DIR, swiftsyntax_destination)
+
+    clear_gyb_files_from_previous_run(
+        SWIFTSYNTAXBUILDER_DIR, swiftsyntaxbuilder_destination)
 
     for previous_gyb_gen_file in os.listdir(template_destination):
         if previous_gyb_gen_file.endswith(".swift"):
@@ -205,35 +209,18 @@ def generate_gyb_files(gyb_exec, verbose, add_source_locations, destination=None
                     verbose=verbose,
                 )
 
-    # Generate the new .swift files in a temporary directory and only copy them
-    # to Sources/SwiftSyntax/gyb_generated if they are different than the files
-    # already residing there. This way we don't touch the generated .swift
-    # files if they haven't changed and don't trigger a rebuild.
-    for gyb_file in os.listdir(swiftsyntax_sources_dir):
-        if not gyb_file.endswith(".gyb"):
-            continue
+    generate_gyb_files_helper(gyb_exec, SWIFTSYNTAX_DIR, swiftsyntax_destination,
+        temp_files_dir, add_source_locations, verbose)
 
-        gyb_file_path = os.path.join(swiftsyntax_sources_dir, gyb_file)
-
-        # Slice off the '.gyb' to get the name for the output file
-        output_file_name = gyb_file[:-4]
-
-        generate_single_gyb_file(
-            gyb_exec,
-            gyb_file_path,
-            output_file_name,
-            destination,
-            temp_files_dir,
-            add_source_locations,
-            additional_gyb_flags=[],
-            verbose=verbose,
-        )
+    generate_gyb_files_helper(gyb_exec, SWIFTSYNTAXBUILDER_DIR, 
+        swiftsyntaxbuilder_destination, temp_files_dir, add_source_locations, 
+        verbose)
 
     for base_kind in BASE_KIND_FILES:
         output_file_name = BASE_KIND_FILES[base_kind]
 
         gyb_file = os.path.join(
-            swiftsyntax_sources_dir, "SyntaxNodes.swift.gyb.template"
+            SWIFTSYNTAX_DIR, "SyntaxNodes.swift.gyb.template"
         )
 
         generate_single_gyb_file(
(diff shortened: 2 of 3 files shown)


build-script.py
```diff
@@ -170,30 +170,46 @@ def generate_gyb_files(gyb_exec, verbose, add_source_locations, destination=None
     check_gyb_exec(gyb_exec)
     check_rsync()
 
-    swiftsyntax_sources_dir = os.path.join(PACKAGE_DIR, "Sources", "SwiftSyntax")
```
kimdv, line 173: There is some duplicated code here. 
Was thinking of making a new method to avoid this? 

Should I do that?

Sources/SwiftSyntaxBuilder/Tokens.swift.gyb
```diff
@@ -1,8 +1,15 @@
...
+  # Ignore the following admonition it applies to the resulting .swift file only
+}%
+//// Automatically Generated From Tokens.swift.gyb.
+//// Do Not Edit Directly!
 //===----------------------------------------------------------------------===//
 //
 // This source file is part of the Swift.org open source project
 //
-// Copyright (c) 2014 - 2019 Apple Inc. and the Swift project authors
+// Copyright (c) 2014 - 2021 Apple Inc. and the Swift project authors
```
kimdv, line 12: Is this okay?

Sources/SwiftSyntaxBuilder/DeclBuildables.swift
```diff
@@ -159,7 +159,7 @@ public struct Struct: DeclBuildable {
   }
 
   public func buildDecl(format: Format, leadingTrivia: Trivia) -> DeclSyntax {
-    let structKeyword = Tokens.struct.withLeadingTrivia(leadingTrivia)
+    let structKeyword = Tokens.structKeyword.withLeadingTrivia(leadingTrivia)
```
kimdv, line 162: Should the new generated one have the `keyword` removed or is this good?

build-script.py
ahoppen: Would be great. I didn’t realize there would be this much code duplication.

Also: Could you make sure that the lines are at most 80 columns. It’s not a hard requirement here, but would be nice if they didn’t get too long.

ahoppen (commented):
Looking really good 👍. 

I’ve left some comments inline, mostly responding to issues you already raised. Other than that I don’t have much to add.

ahoppen:
Triggering CI already to see if it’s got something to complain about already.

@swift-ci Please test

Sources/SwiftSyntaxBuilder/DeclBuildables.swift
kimdv: I also think it is some verbose, but I didn't know what you preferred. 
I tried to se if there was a way to lower case first letter in `${token.name}`

kimdv:
I just saw there was a `verify_generated_files` function: https://github.com/apple/swift-syntax/pull/263/files#diff-cfa908195f6a017af0a6da67bab31be3a9b97bbc55b3e4d939944786cff075fdR357

Should I also make one for the `SwiftSyntaxBuilder` gyps?

ahoppen:
> Should I also make one for the SwiftSyntaxBuilder gyps?

Oh yeah, definitely.

build-script.py
```diff
@@ -170,30 +170,46 @@ def generate_gyb_files(gyb_exec, verbose, add_source_locations, destination=None
...
+    sources_dir = os.path.join(PACKAGE_DIR, "Sources")
+    swiftsyntax_sources_dir = os.path.join(sources_dir, "SwiftSyntax")
+    swiftsyntaxbuilder_sources_dir = os.path.join(sources_dir, "SwiftSyntaxBuilder")
     temp_files_dir = tempfile.gettempdir()
 
     if destination is None:
-        destination = os.path.join(swiftsyntax_sources_dir, "gyb_generated")
+        swiftsyntax_destination = os.path.join(swiftsyntax_sources_dir, "gyb_generated")
+        swiftsyntaxbuilder_destination = os.path.join(swiftsyntaxbuilder_sources_dir, "gyb_generated")
```
ahoppen, line 180: This should probably be 

```python
if swiftsyntax_destination is None:
  swiftsyntax_destination = os.path.join(swiftsyntax_sources_dir, "gyb_generated")
if swiftsyntaxbuilder_destination is None:
  swiftsyntaxbuilder_destination = os.path.join(swiftsyntaxbuilder_sources_dir, "gyb_generated")
```

That’s what CI is failing over.

build-script.py
```diff
@@ -16,6 +16,9 @@
 
 PACKAGE_DIR = os.path.dirname(os.path.realpath(__file__))
 WORKSPACE_DIR = os.path.dirname(PACKAGE_DIR)
+SOURCES_DIR = os.path.join(PACKAGE_DIR, "Sources")
```
kimdv, line 19: Add those new constants as they was used multiple places

build-script.py
kimdv: I have tried to refactor, but the `BASE_KIND_FILES` depend on the destination from SwiftSyntax, so I couldn't find a generic way. But I rested some duplicated code

build-script.py
```diff
@@ -315,28 +349,36 @@ def build(self, product_name, module_group_path=""):
 
 
 def verify_generated_files(gyb_exec, verbose):
```
kimdv, line 351: I tried to run this locally to see if it worked, but this or `main` worked for me 🤷‍♂️

Sources/SwiftSyntaxBuilder/DeclBuildables.swift
ahoppen: You can do it similar to how `swift_kind` is computed in [Token.py](https://github.com/apple/swift/blob/main/utils/gyb_syntax_support/Token.py#L24):

```swift
from gyb_syntax_support/kinds import lowercase_first_word

// ...

static let `${lowercase_first_word(token.name)}` = SyntaxFactory.make${token.name}Keyword().withTrailingTrivia(.spaces(1))
```

(or something like this, I haven’t tested the code)

build-script.py
```diff
@@ -250,6 +236,54 @@ def generate_gyb_files(gyb_exec, verbose, add_source_locations, destination=None
...
+
+
+# Generate the new .swift files in a temporary directory and only copy them
+# to Sources/SwiftSyntaxBuilder/gyb_generated if they are different than the
+# files already residing there. This way we don't touch the generated .swift
+# files if they haven't changed and don't trigger a rebuild.
+def generate_gyb_files_helper(
+    gyb_exec, verbose, add_source_locations, 
+    temp_files_dir, sources_dir, destination_dir
+):
```
ahoppen, line 265: Some minor documentation improvements.

`Sources/SwiftSyntaxBuilder/gyb_generated` should probably be `destination_dir`.
`temporary directory` could be further specified as `temp_files_dir`.

Also, super nitpicky: Could we re-arrange the arguments here a little to make `verbose` the last argument and the boolean flag `add_source_locations` the second to last one? It better matches the other functions.

build-script.py
```diff
@@ -250,6 +236,54 @@ def generate_gyb_files(gyb_exec, verbose, add_source_locations, destination=None
     print("Done Generating gyb Files")
 
 
+def make_dir_if_needed(path):
+    if not os.path.exists(path):
+        os.makedirs(path)
+
+
+def clear_gyb_files_from_previous_run(sources_dir, destination_dir):
```
ahoppen, line 244: Could you add a comment here like. It took me a while to understand what this is actually doing even though I was probably who wrote the code snippet.

```
# Remove any files in the `gyb_generated` directory that no longer have a corresponding `.gyb` file in the `Sources` directory.
```

ahoppen (commented):
I’ve added some more nitpicky comments inline, mostly about better documentation. The refactoring of `build-script` looks good to me.

All that remains to be done before we can merge this, would be to drop the `Keyword` part from the `Tokens` cases.

ahoppen:
I’ve also though about what the next steps are for a bit and would like to share them with you. This is not a a definite guide. Read this more as “these are the steps I would be planning to do next”. If you’ve got different ideas on how to approach them or would like to proceed in a different order, that’s very good.

### 1. Generate cases for the remaining tokens

After a quick scan through `Token.py`, all that we have left still are [Punctuators](https://github.com/apple/swift/blob/main/utils/gyb_syntax_support/Token.py#L210) and the [Misc](https://github.com/apple/swift/blob/main/utils/gyb_syntax_support/Token.py#L320) tokens. I don’t see a general pattern which of these need a trailing space, so we’d need to annotate them manually. For that, I’d add another property to the Python [Token](https://github.com/apple/swift/blob/main/utils/gyb_syntax_support/Token.py#L6) class, like `requires_trailing_space`. For keywords this is always set to `true` and for all other tokens, we need to specify it manually in the definition (e.g. `{` requires a trailing space, but `.` does not). On the Swift side, we can check for this property and call `withTrailingTrivia` or not.

### 2. Actually create the `Buildable` types

I think with that we’re already to the main part. Here is a random collection on thoughts, how I would approach it. Again a disclaimer: I haven’t tested any of this and it might be whofully wrong.

First off, I would try and get a raw API up and running. At this stage, I wouldn’t worry about a nice API or result builders yet, I would just want to get an initialiser-based API up and running with which all syntax nodes can be created.
- As a baseline, I would start with [SyntaxBuilders.swift](https://github.com/apple/swift-syntax/blob/main/Sources/SwiftSyntax/SyntaxBuilders.swift.gyb) and modify it according to our needs.
- I wouldn’t be too focused on the current design of the buildables since we probably can’t automatically re-create a generic modelling like it’s currently done for variables [here](https://github.com/apple/swift-syntax/blob/main/Sources/SwiftSyntaxBuilder/DeclBuildables.swift#L104). I’m fine with any design that’s easy to use and we are not concerned about source breakage here.
- As a first step, I would try modifying the `SyntaxBuilders` so they are purely initialiser based, that is all children are passed in the initialiser and we don’t need the `use*` methods anymore.

After this, I am expecting the API to be *much worse* than it is today, but it should be exhaustive now. This means, we now need to make it easy to use again. I’ve got a couple of (more or less) necessary improvements in my head already, I’m sure there are more once you get to it.
- Of course the entire result builder thing. Any syntax collection should be buildable using a corresponding result builder.
- Any non-optional token without custom text (i.e. a keyword or punctuator) ccan be implicitly synthesized and does not have a corresponding argument in the initialiser
	- Example: We don't need to specify the open `{` for a `struct` decl.
- And as far as I can tell we’re now already at a stage, where custom specialisations are needed. These can be made in a separate file through `extensions`.
	- Example: According to the specification of a [VariableDecl](https://github.com/apple/swift/blob/main/utils/gyb_syntax_support/DeclNodes.py#L566), it is (simplified) a `var`/`let` keyword with a following `PatternBindingList`. But that’s not how anybody thinks about a `Var` decl. We should create an initialiser that’s closer to what we have [today](https://github.com/apple/swift-syntax/blob/main/Sources/SwiftSyntaxBuilder/DeclBuildables.swift#L109), which correctly creates the pattern binding etc.
	- Example: According to the definition of a [StructDecl](https://github.com/apple/swift/blob/main/utils/gyb_syntax_support/DeclNodes.py#L232), it takes its members as a [MemberDeclBlock](https://github.com/apple/swift/blob/main/utils/gyb_syntax_support/DeclNodes.py#L287), consisting of a pair of braces and decls which might be followed by a semicolon. The braces can already be implicitly synthesized as described above and nobody cares about semicolons. So we should have a specialised initialiser taking a list of decls (through a result builder) that correctly creates the `MemberDeclBlock` etc.

I think the advantage of this approach (auto-generate a exhaustive, but verbose API and add convenience extensions on top) apposed to the previous approach (hand write everything), is that there is always a fall-back to create every syntax element, even cutting-edge ones. I expect the convenience initialisers to concern the most commonly used syntax elements, which should also be the most stable, so I expect they should be fairly easy to maintain.

What do you think @kimdv? Does this sound good to you?

kimdv:
@ahoppen I should have resolved all your comments now. Please reopen if I'm wrong. 😁

I think it sounds like a good plan! 
I'm not that familiar with the whole generating thing, and underlying stuff in SwiftSyntax and Swift. But I'll learn it 💪

I agree in that we should make the last tokens first, and then we can take it from there? 
So I will start looking into it when this is merged? 
Then we can take the discussion over there?

build-script.py
```diff
@@ -205,35 +209,17 @@ def generate_gyb_files(gyb_exec, verbose, add_source_locations, destination=None
...
-            gyb_file_path,
-            output_file_name,
-            destination,
-            temp_files_dir,
-            add_source_locations,
-            additional_gyb_flags=[],
-            verbose=verbose,
-        )
+    generate_gyb_files_helper(gyb_exec, SWIFTSYNTAXBUILDER_DIR, 
+    swiftsyntaxbuilder_destination, temp_files_dir, add_source_locations, verbose)
```
ahoppen, line 216: I think this line should be indented.
```suggestion
        swiftsyntaxbuilder_destination, temp_files_dir, add_source_locations, 
        verbose)
```

build-script.py
```diff
@@ -315,28 +355,36 @@ def build(self, product_name, module_group_path=""):
...
     )
 
     self_generated_dir = tempfile.mkdtemp()
     generate_gyb_files(
         gyb_exec,
         verbose=verbose,
         add_source_locations=False,
-        destination=self_generated_dir,
+        swiftsyntax_destination=self_generated_dir,
+        swiftsyntaxbuilder_destination=self_generated_dir,
```
ahoppen, line 372: I just checked and `self_generated_dir` needs to be different for `SwiftSyntax` and `SwiftSyntaxBuilder`. As a little background what `verify_generated_files` is for: You already noticed that we don’t only check in the `.swift.gyb` files but also the generated `.swift` files. This is because SwiftSyntax should be a standalone Swift package and users of SwiftSyntax wouldn't be able to generate the files themselves (because SwiftPM doesn’t allow source generation, because the user probably doesn't have `gyb` installed on their system and because they are missing the Python files in the main Swift repo). Now, we do of course want to make sure that the generated files actually match the `.gyb` templates and that you don’t forget to check them in. This is why, in CI we are (more or less) running 

```bash
./build-script.py --toolchain /path/to/toolchain/ --degyb-only --verify-generated-files
```

which will fail if `gyb`-generation produces different files than those checked into the repo. If you want, you can try it yourself: Modify a file in `gyb_generated` and run the above command.

Back to the code, I was fiddling around locally and would propose the following change if you like it.

```python
def verify_generated_files(gyb_exec, verbose):
    user_swiftsyntax_generated_dir = os.path.join(
        SWIFTSYNTAX_DIR, "gyb_generated"
    )

    user_swiftsyntaxbuilder_generated_dir = os.path.join(
        SWIFTSYNTAXBUILDER_DIR, "gyb_generated"
    )

    self_swiftsyntax_generated_dir = tempfile.mkdtemp()
    self_swiftsyntaxbuilder_generated_dir = tempfile.mkdtemp()
    generate_gyb_files(
        gyb_exec,
        verbose=verbose,
        add_source_locations=False,
        swiftsyntax_destination=self_swiftsyntax_generated_dir,
        swiftsyntaxbuilder_destination=self_swiftsyntaxbuilder_generated_dir,
    )

    check_generated_files_match(self_swiftsyntax_generated_dir, user_swiftsyntax_generated_dir)
    check_generated_files_match(self_swiftsyntaxbuilder_generated_dir, user_swiftsyntaxbuilder_generated_dir)

def check_generated_files_match(self_generated_dir, user_generated_dir):
    command = [
        "diff",
        "-r",
        "-x",
        ".*",  # Exclude dot files like .DS_Store
        "--context=0",
        self_generated_dir,
        user_generated_dir,
    ]
    check_call(command)
```

ahoppen (commented):
We’re almost there. I’ve run the tests locally and they all pass. Only `--verify-generated-files` isn’t quite working yet. I’ve added comments inline. After that we’re good to merge.

ahoppen:
> I'm not that familiar with the whole generating thing, and underlying stuff in SwiftSyntax and Swift. But I'll learn it 💪

I think you’ve already gotten the core by generating the tokens. Everything else is just generating more complex code. 😉

> I agree in that we should make the last tokens first, and then we can take it from there?
> So I will start looking into it when this is merged?
> Then we can take the discussion over there?

Sounds good. Just open another PR and we can discuss further on there. I just wanted to give you a high-level overview of what needs to be done and didn’t know of a better place to post it than here.

kimdv:
So! I added your verify code. I understand it now and it makes sense, thanks!

ahoppen (approved):
Looking good now. 🚢 it.

ahoppen:
@swift-ci Please test

ahoppen:
Nevermind those first test failures, it’s an oddity of our CI. Real CI is running now.

ahoppen:
Congrats on your first contribution to Swift @kimdv 🎉

ktoso:
Wohoo, awesome -- so looking forward to this, thanks a lot for your work @kimdv @ahoppen

PR #264, opened 2021-03-10
Add gyb for Punctuators and Misc tokens

kimdv:
From #263 
> ### 1. Generate cases for the remaining tokens
> After a quick scan through `Token.py`, all that we have left still are [Punctuators](https://github.com/apple/swift/blob/main/utils/gyb_syntax_support/Token.py#L210) and the [Misc](https://github.com/apple/swift/blob/main/utils/gyb_syntax_support/Token.py#L320) tokens. I don’t see a general pattern which of these need a trailing space, so we’d need to annotate them manually. For that, I’d add another property to the Python [Token](https://github.com/apple/swift/blob/main/utils/gyb_syntax_support/Token.py#L6) class, like `requires_trailing_space`. For keywords this is always set to `true` and for all other tokens, we need to specify it manually in the definition (e.g. `{` requires a trailing space, but `.` does not). On the Swift side, we can check for this property and call `withTrailingTrivia` or not.
> 

I have added tokens for Punctuators and Misc, but as you mention we should add `requires_trailing_space `. 
I have seen that the `=` also require a leading space, so we should also we also add an `requires_leading_space ` for those few cases? 

You mention that we should add to [Token](https://github.com/apple/swift/blob/main/utils/gyb_syntax_support/Token.py#L6). 
Is that an extension or should we add this to the main swift repo? 😁

### DIFF
--- a/Sources/SwiftSyntaxBuilder/Tokens.swift.gyb
+++ b/Sources/SwiftSyntaxBuilder/Tokens.swift.gyb
@@ -22,27 +22,36 @@ import SwiftSyntax
 
 /// Namespace for commonly used tokens with default trivia.
 enum Tokens {
-  // MARK: Keywords
-
 % for token in SYNTAX_TOKENS:
 %   if token.is_keyword:
-  /// The `${token.text}` keyword
-  static let `${lowercase_first_word(token.name)}` = SyntaxFactory.make${token.name}Keyword().withTrailingTrivia(.spaces(1))
+  /// The `${token.text.encode('utf-8').decode('unicode_escape')}` keyword
+  static let `${lowercase_first_word(token.name)}` = SyntaxFactory.make${token.name}Keyword()
+%   if token.requires_leading_space:
+    .withLeadingTrivia(.spaces(1))
 %   end
+%   if token.requires_trailing_space:
+    .withTrailingTrivia(.spaces(1))
+%   end
+%   elif token.text:
+  /// The `${token.text.encode('utf-8').decode('unicode_escape')}` token
+  static let `${lowercase_first_word(token.name)}` = SyntaxFactory.make${token.name}Token()
+%   if token.requires_leading_space:
+    .withLeadingTrivia(.spaces(1))
+%   end
+%   if token.requires_trailing_space:
+    .withTrailingTrivia(.spaces(1))
+%   end
+%   else:
+  static func `${lowercase_first_word(token.name)}`(_ text: String) -> TokenSyntax {
+    SyntaxFactory.make${token.name}(text)
+%   if token.requires_leading_space:
+    .withLeadingTrivia(.spaces(1))
+%   end
+%   if token.requires_trailing_space:
+    .withTrailingTrivia(.spaces(1))
+%   end
+  }
+%   end
+
 % end
-
-  // MARK: Punctuations and Signs
-
-  /// `":"`
-  static let colon = SyntaxFactory.makeColonToken().withTrailingTrivia(.spaces(1))
-
-  /// `" = "`
-  static let equal = SyntaxFactory.makeEqualToken().withLeadingTrivia(.spaces(1))
-                                                   .withTrailingTrivia(.spaces(1))
-
-  /// `"{"`
-  static let leftBrace = SyntaxFactory.makeLeftBraceToken()
-
-  /// `"}"`
-  static let rightBrace = SyntaxFactory.makeRightBraceToken()
 }
--- a/Sources/SwiftSyntaxBuilder/gyb_generated/Tokens.swift
+++ b/Sources/SwiftSyntaxBuilder/gyb_generated/Tokens.swift
@@ -16,181 +16,474 @@ import SwiftSyntax
 
 /// Namespace for commonly used tokens with default trivia.
 enum Tokens {
-  // MARK: Keywords
-
   /// The `associatedtype` keyword
-  static let `associatedtype` = SyntaxFactory.makeAssociatedtypeKeyword().withTrailingTrivia(.spaces(1))
+  static let `associatedtype` = SyntaxFactory.makeAssociatedtypeKeyword()
+    .withTrailingTrivia(.spaces(1))
+
   /// The `class` keyword
-  static let `class` = SyntaxFactory.makeClassKeyword().withTrailingTrivia(.spaces(1))
+  static let `class` = SyntaxFactory.makeClassKeyword()
+    .withTrailingTrivia(.spaces(1))
+
   /// The `deinit` keyword
-  static let `deinit` = SyntaxFactory.makeDeinitKeyword().withTrailingTrivia(.spaces(1))
+  static let `deinit` = SyntaxFactory.makeDeinitKeyword()
+    .withTrailingTrivia(.spaces(1))
+
   /// The `enum` keyword
-  static let `enum` = SyntaxFactory.makeEnumKeyword().withTrailingTrivia(.spaces(1))
+  static let `enum` = SyntaxFactory.makeEnumKeyword()
+    .withTrailingTrivia(.spaces(1))
+
   /// The `extension` keyword
-  static let `extension` = SyntaxFactory.makeExtensionKeyword().withTrailingTrivia(.spaces(1))
+  static let `extension` = SyntaxFactory.makeExtensionKeyword()
+    .withTrailingTrivia(.spaces(1))
+
   /// The `func` keyword
-  static let `func` = SyntaxFactory.makeFuncKeyword().withTrailingTrivia(.spaces(1))
+  static let `func` = SyntaxFactory.makeFuncKeyword()
+    .withTrailingTrivia(.spaces(1))
+
   /// The `import` keyword
-  static let `import` = SyntaxFactory.makeImportKeyword().withTrailingTrivia(.spaces(1))
+  static let `import` = SyntaxFactory.makeImportKeyword()
+    .withTrailingTrivia(.spaces(1))
+
   /// The `init` keyword
-  static let `init` = SyntaxFactory.makeInitKeyword().withTrailingTrivia(.spaces(1))
+  static let `init` = SyntaxFactory.makeInitKeyword()
+    .withTrailingTrivia(.spaces(1))
+
   /// The `inout` keyword
-  static let `inout` = SyntaxFactory.makeInoutKeyword().withTrailingTrivia(.spaces(1))
+  static let `inout` = SyntaxFactory.makeInoutKeyword()
+    .withTrailingTrivia(.spaces(1))
+
   /// The `let` keyword
-  static let `let` = SyntaxFactory.makeLetKeyword().withTrailingTrivia(.spaces(1))
+  static let `let` = SyntaxFactory.makeLetKeyword()
+    .withTrailingTrivia(.spaces(1))
+
   /// The `operator` keyword
-  static let `operator` = SyntaxFactory.makeOperatorKeyword().withTrailingTrivia(.spaces(1))
+  static let `operator` = SyntaxFactory.makeOperatorKeyword()
+    .withTrailingTrivia(.spaces(1))
+
   /// The `precedencegroup` keyword
-  static let `precedencegroup` = SyntaxFactory.makePrecedencegroupKeyword().withTrailingTrivia(.spaces(1))
+  static let `precedencegroup` = SyntaxFactory.makePrecedencegroupKeyword()
+    .withTrailingTrivia(.spaces(1))
+
   /// The `protocol` keyword
-  static let `protocol` = SyntaxFactory.makeProtocolKeyword().withTrailingTrivia(.spaces(1))
+  static let `protocol` = SyntaxFactory.makeProtocolKeyword()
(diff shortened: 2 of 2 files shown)


Sources/SwiftSyntaxBuilder/Tokens.swift.gyb
```diff
@@ -33,16 +33,19 @@ enum Tokens {
...
+% end
 
-  /// `"{"`
-  static let leftBrace = SyntaxFactory.makeLeftBraceToken()
+  // MARK: Punctuations and Signs
 
-  /// `"}"`
-  static let rightBrace = SyntaxFactory.makeRightBraceToken()
+% for token in SYNTAX_TOKENS:
+%   if token.macro_name() == "MISC":
```
kimdv, line 46: I'm not sure if this is the way to do it? 
Can't find a thing similar to `is_keyword`? Should we add one or should we generate for all tokens and not only `keyword`, `mist` and `punctuator` ?

ktoso (commented):
I noticed the `actor` keyword is not in the generated list here, does the gyb generation perhaps have to run with -enable-experimental-concurrency or is there some other way to get that?

ahoppen:
> I have added tokens for Punctuators and Misc, but as you mention we should add requires_trailing_space .
I have seen that the = also require a leading space, so we should also we also add an requires_leading_space for those few cases?

Ah, yes, forgot about `=` (or spaced binary operators in general which should also have a leading space). Also adding a `requires_leading_space` property sounds good 👍 

> You mention that we should add to Token.
Is that an extension or should we add this to the main swift repo? 😁

Add it to the main swift repo, set up a PR there for your changes and just link to this PR in the description.

ahoppen:
> I noticed the actor keyword is not in the generated list here, does the gyb generation perhaps have to run with -enable-experimental-concurrency or is there some other way to get that?

That’s because `actor` is not in the list of tokens in [Token.py](https://github.com/apple/swift/blob/main/utils/gyb_syntax_support/Token.py). I think that should be resolved in the main Swift repo and is outside the scope of this PR.

Sources/SwiftSyntaxBuilder/Tokens.swift.gyb
ahoppen: I think I would just differentiate between keywords and non-keywords. I think there’s little benefit in distinguishing punctuators and miscellaneous tokens unless we would have been able to find a general rule regarding whitespace (which we didn’t).

kimdv:
@ahoppen I opened a PR in the Swift repo. 
I'm not completely sure how and where we should add all those cases. 

Should there be some default in the `__init__` and if there is some, that need another way we can add it in the `SYNTAX_TOKENS` list for that special case?

Etc: `Punctuator` have default leading = false and trailing = true.  


`Punctuator('Equal', 'equal', text='=', requires_leading_space=True, serialization_code=86),` and `Punctuator('Period', 'period', text='.', requires_trailing_space=False, serialization_code=85),`

ahoppen:
At a first glance the Swift PR looks very good. I think you went a very good way with the defaults already.

I think for all the other tokens we can reasonably assume a default of `False` in the initialiser and only add `requires_leading_space=True` (or `requires_trailing_space=True`) for those tokens that require it, just like you suggested (although I think that `=` should also have a trailing space 😉).

If you don’t know what some of the tokens are or whether they need trivia, you can also just leave them unspaced for now. It’s probably easier to go and revisit the decisions later once we have the builder up and running and get a feeling for how it behaves.

kimdv:
I have now pushed a new gyb file for Tokens. 
I think we can make it to one loop as the [Token.py](https://github.com/apple/swift/pull/36383/files#diff-d3bd1752ba5f7bfa3cedcdb1f753244e79f9a21c7c3689fb12f8697d72ed9247R6) contains all information about leading/trailing. 

As you say we can leave them unspaced, and then change later if needed.
I have tried to find the ones that need a space.

ktoso:
> That’s because actor is not in the list of tokens in Token.py. I think that should be resolved in the main Swift repo and is outside the scope of this PR.

Okey so I confirmed that piece. We indeed have it as a contextual keyword, same with await and async.

For the builders once we get there it'll be nice to construct an `actor` the same way as just a `class`, but that's later I suppose - the actual builder DSL  :)

Sources/SwiftSyntaxBuilder/gyb_generated/Tokens.swift
```diff
@@ -19,178 +19,340 @@ enum Tokens {
...
+  /// The `?` keyword
+  static let `postfixQuestionMark` = SyntaxFactory.makePostfixQuestionMarkKeyword()
+  /// The `?` keyword
+  static let `infixQuestionMark` = SyntaxFactory.makeInfixQuestionMarkKeyword()
+  /// The `\"` keyword
+  static let `stringQuote` = SyntaxFactory.makeStringQuoteKeyword()
+  /// The `\'` keyword
+  static let `singleQuote` = SyntaxFactory.makeSingleQuoteKeyword()
+  /// The `\"\"\"` keyword
+  static let `multilineStringQuote` = SyntaxFactory.makeMultilineStringQuoteKeyword()
```
ahoppen, line 330: It’s a bit unfortunate that we over-escape the text in `Token.py` so it ends up `\`-escaped here.

The simplest solution I found on the internet would be print it as
```python
token.text.encode('utf-8').decode('unicode_escape')
```

Not sure if that breaks any other comments though. If it does and you can’t find an easy solution to remove the `\` here, just leave it as is for now.

Sources/SwiftSyntaxBuilder/gyb_generated/Tokens.swift
```diff
@@ -19,178 +19,340 @@ enum Tokens {
...
-                                                   .withTrailingTrivia(.spaces(1))
-
-  /// `"{"`
-  static let leftBrace = SyntaxFactory.makeLeftBraceToken()
+  // MARK: Punctuations and Signs
 
-  /// `"}"`
-  static let rightBrace = SyntaxFactory.makeRightBraceToken()
+  /// The `` keyword
+  static let `unknown` = SyntaxFactory.makeUnknownKeyword()
```
ahoppen, line 335: I think we shouldn’t print a doc-comment if the token's text is empty.

ahoppen (commented):
The PR looks very good overall (also the Swift side, I’ve left a comment there as well).

> I think we can make it to one loop as the Token.py contains all information about leading/trailing.

I agree that one loop should be sufficient. As a bonus you could adjust the comment to say "the … keyword" or "the … token" depending on whether the token is a keyword or not, but I’m also fine with just having the comment say "the … token".
Also, I don’t care about the `MARK` comments. Feel free to remove them if they complicate things.

Other than that, I’ve got a few minor comments about the doc-comments, although they are not critical. Once we generate all tokens in one loop, I think we’re good to merge.

ahoppen:
> Okey so I confirmed that piece. We indeed have it as a contextual keyword, same with await and async.
>
> For the builders once we get there it'll be nice to construct an actor the same way as just a class, but that's later I suppose - the actual builder DSL :)

Sounds like a case where we need to hand write a constructor for those contextual keywords. We can do that once we’re approaching a close with the auto-generated stuff.

Sources/SwiftSyntaxBuilder/gyb_generated/Tokens.swift
```diff
@@ -19,178 +19,340 @@ enum Tokens {
...
+  static let `period` = SyntaxFactory.makePeriodKeyword()
+  /// The `.` keyword
+  static let `prefixPeriod` = SyntaxFactory.makePrefixPeriodKeyword()
+  /// The `,` keyword
+  static let `comma` = SyntaxFactory.makeCommaKeyword()
+    .withLeadingTrivia(.spaces(1))
+  /// The `...` keyword
+  static let `ellipsis` = SyntaxFactory.makeEllipsisKeyword()
+  /// The `:` keyword
+  static let `colon` = SyntaxFactory.makeColonKeyword()
```
kimdv, line 296: Should we also change `SyntaxFactory.makeColonKeyword()` -> `SyntaxFactory.makeColonToken()` @ahoppen?

ahoppen: I guess so because it doesn’t compile otherwise or does it?

I suggest you look at [SyntaxFactory.swift.gyb](https://github.com/apple/swift-syntax/blob/main/Sources/SwiftSyntax/SyntaxFactory.swift.gyb#L104) to figure out how the call to `SyntaxFactory` should look like so it compiles. 

I also just noticed that there are tokens that don't have a fixed text (for which `token.text` is not set). For those, we might need to generate functions instead of constants. E.g.

```swift
static func identifier(_ text: String) {
  return SyntaxFactory.makeIdentifier(text)
}
```

kimdv: Ah yes! Good one. 
In that way we can also keep the `keyword` in the docs. 

Pushed new updated code

Sources/SwiftSyntaxBuilder/gyb_generated/Tokens.swift
```diff
@@ -16,181 +16,450 @@ import SwiftSyntax
...
-  static let `poundImageLiteral` = SyntaxFactory.makePoundImageLiteralKeyword().withTrailingTrivia(.spaces(1))
+  static let `poundImageLiteral` = SyntaxFactory.makePoundImageLiteralKeyword()
+    .withLeadingTrivia(.spaces(1))
+
   /// The `#colorLiteral` keyword
-  static let `poundColorLiteral` = SyntaxFactory.makePoundColorLiteralKeyword().withTrailingTrivia(.spaces(1))
+  static let `poundColorLiteral` = SyntaxFactory.makePoundColorLiteralKeyword()
+    .withLeadingTrivia(.spaces(1))
+
+  static let `integerLiteral` = SyntaxFactory.makeIntegerLiteral()
```
ahoppen, line 433: I don’t think this compiles. `makeIntegerLiteral` has a required `text` parameter. 

Stupid question but are you checking if your code compiles before pushing the changes?

kimdv: Note to my self: Remember to delete `DerivedData` when chugging toolchain, sorry about that!  😞
`SwiftSyntax` started not to compile.. Strange.. But it should be fixed now and build and tests pass locally. 

Pushed to https://github.com/apple/swift/pull/36383 as tests did fail with `{` and `}` because we added trailing space. We can not generalise that.

Sources/SwiftSyntaxBuilder/Tokens.swift.gyb
```diff
@@ -22,27 +22,24 @@ import SwiftSyntax
...
+%   else:
+  static func `${lowercase_first_word(token.name)}`(_ text: String) -> TokenSyntax {
+    SyntaxFactory.make${token.name}(text)
+  }
+%   end
+%   if token.requires_leading_space:
+    .withLeadingTrivia(.spaces(1))
+%   end
+%   if token.requires_trailing_space:
+    .withTrailingTrivia(.spaces(1))
```
ahoppen, line 41: This will break once we have a token with text that requires a leading or trailing space, because then we’d get

```swift
static func `${lowercase_first_word(token.name)}`(_ text: String) -> TokenSyntax {
  SyntaxFactory.make${token.name}(text)
}
.withLeadingTrivia(.spaces(1))
```

which doesn’t compile.

kimdv: I'm not sure that I understand?

So we have to move

```
%   if token.requires_leading_space:
     .withLeadingTrivia(.spaces(1))
 %   end
```
and
```
%   if token.requires_leading_space:
     .withLeadingTrivia(.spaces(1))
 %   end
```

inside so it look like

```
  /// The `${token.text.encode('utf-8').decode('unicode_escape')}` keyword
   static let `${lowercase_first_word(token.name)}` = SyntaxFactory.make${token.name}Keyword()
%   if token.requires_leading_space:
     .withLeadingTrivia(.spaces(1))
 %   end
 %   elif token.text:
   /// The `${token.text.encode('utf-8').decode('unicode_escape')}` token
   static let `${lowercase_first_word(token.name)}` = SyntaxFactory.make${token.name}Token()
 %   if token.requires_leading_space:
     .withLeadingTrivia(.spaces(1))
 %   end
 %   else:
   static func `${lowercase_first_word(token.name)}`(_ text: String) -> TokenSyntax {
     SyntaxFactory.make${token.name}(text)
%   if token.requires_leading_space:
     .withLeadingTrivia(.spaces(1))
 %   end
   }
```

ahoppen: Yes, exactly.

kimdv: Got it. Should be pushed. 
I've run the `gyb` generation, to ensure nothing have changed.

ahoppen (approved):
Looking good now!

ahoppen:
https://github.com/apple/swift/pull/36383

@swift-ci Please test

PR #268, opened 2021-03-18
Add Buildable gyb

kimdv:
From #263 

> ### 2. Actually create the `Buildable` types
> I think with that we’re already to the main part. Here is a random collection on thoughts, how I would approach it. Again a disclaimer: I haven’t tested any of this and it might be whofully wrong.
> 
> First off, I would try and get a raw API up and running. At this stage, I wouldn’t worry about a nice API or result builders yet, I would just want to get an initialiser-based API up and running with which all syntax nodes can be created.
> 
> * As a baseline, I would start with [SyntaxBuilders.swift](https://github.com/apple/swift-syntax/blob/main/Sources/SwiftSyntax/SyntaxBuilders.swift.gyb) and modify it according to our needs.
> * I wouldn’t be too focused on the current design of the buildables since we probably can’t automatically re-create a generic modelling like it’s currently done for variables [here](https://github.com/apple/swift-syntax/blob/main/Sources/SwiftSyntaxBuilder/DeclBuildables.swift#L104). I’m fine with any design that’s easy to use and we are not concerned about source breakage here.
> * As a first step, I would try modifying the `SyntaxBuilders` so they are purely initialiser based, that is all children are passed in the initialiser and we don’t need the `use*` methods anymore.
> 
> After this, I am expecting the API to be _much worse_ than it is today, but it should be exhaustive now. This means, we now need to make it easy to use again. I’ve got a couple of (more or less) necessary improvements in my head already, I’m sure there are more once you get to it.
> 
> * Of course the entire result builder thing. Any syntax collection should be buildable using a corresponding result builder.
> * Any non-optional token without custom text (i.e. a keyword or punctuator) ccan be implicitly synthesized and does not have a corresponding argument in the initialiser
>   
>   * Example: We don't need to specify the open `{` for a `struct` decl.
> * And as far as I can tell we’re now already at a stage, where custom specialisations are needed. These can be made in a separate file through `extensions`.
>   
>   * Example: According to the specification of a [VariableDecl](https://github.com/apple/swift/blob/main/utils/gyb_syntax_support/DeclNodes.py#L566), it is (simplified) a `var`/`let` keyword with a following `PatternBindingList`. But that’s not how anybody thinks about a `Var` decl. We should create an initialiser that’s closer to what we have [today](https://github.com/apple/swift-syntax/blob/main/Sources/SwiftSyntaxBuilder/DeclBuildables.swift#L109), which correctly creates the pattern binding etc.
>   * Example: According to the definition of a [StructDecl](https://github.com/apple/swift/blob/main/utils/gyb_syntax_support/DeclNodes.py#L232), it takes its members as a [MemberDeclBlock](https://github.com/apple/swift/blob/main/utils/gyb_syntax_support/DeclNodes.py#L287), consisting of a pair of braces and decls which might be followed by a semicolon. The braces can already be implicitly synthesized as described above and nobody cares about semicolons. So we should have a specialised initialiser taking a list of decls (through a result builder) that correctly creates the `MemberDeclBlock` etc.
> 
> I think the advantage of this approach (auto-generate a exhaustive, but verbose API and add convenience extensions on top) apposed to the previous approach (hand write everything), is that there is always a fall-back to create every syntax element, even cutting-edge ones. I expect the convenience initialisers to concern the most commonly used syntax elements, which should also be the most stable, so I expect they should be fairly easy to maintain.
> 
> What do you think @kimdv? Does this sound good to you?

Okay, now we are ready for the next step. 🚀
(I have modified this file, not sure if it is the right, but then we can start a conversation)

I'm a little confused on how I should approach this. 
> * As a baseline, I would start with [SyntaxBuilders.swift](https://github.com/apple/swift-syntax/blob/main/Sources/SwiftSyntax/SyntaxBuilders.swift.gyb) and modify it according to our needs.
> * I wouldn’t be too focused on the current design of the buildables since we probably can’t automatically re-create a generic modelling like it’s currently done for variables [here](https://github.com/apple/swift-syntax/blob/main/Sources/SwiftSyntaxBuilder/DeclBuildables.swift#L104). I’m fine with any design that’s easy to use and we are not concerned about source breakage here.
> * As a first step, I would try modifying the `SyntaxBuilders` so they are purely initialiser based, that is all children are passed in the initialiser and we don’t need the `use*` methods anymore.

So the idea is that inside etc [StructDeclSyntaxBuilder](https://github.com/apple/swift-syntax/blob/main/Sources/SwiftSyntax/gyb_generated/SyntaxBuilders.swift#L4025) should not have `add*` but all possibilities should be basses within the `init` method?


ahoppen:
Disclaimer up front: I had a fairly good idea how the tokens should look like, my ideas further down become a lot more vague – there might be an altogether better design. Don’t feel like you have to stick to what I tell you as you dive into the implementation.

With that said, the way I would approach it, would be to start of by generating something like the following

```swift
// Simple syntax element
public struct IntegerLiteral: ExprBuildable {
  let digits: Token

  public init(digits: Token) {
    self.digits = digits
  }

  public func buildExpr(format: Format, leadingTrivia: Trivia) -> ExprSyntax {
    let integerLiteral = SyntaxFactory.makeIntegerLiteralExpr(
      digits: SyntaxFactory.makeIntegerLiteral(String(value))
    ).withLeadingTrivia(leadingTrivia)
    return ExprSyntax(integerLiteral)
  }
}

// More complex element
public struct ImportDecl: DeclBuildable {
  let attributes: AttributeListSyntax
  let modifiers: ModifeirListSyntax
  let importTok: Token
  let importKind: Token
  let path: AccessPathSyntax

  public init(
    attributes: AttributeListSyntax,
    modifiers: ModifeirListSyntax,
    importTok: Token,
    importKind: Token?,
    path: AccessPathSyntax,
  ) {
    self.attributes = attributes
    self.modifiers = modifiers
    self.importTok = importTok
    self.importKind = importKind
    self.path = path
  }

  public func buildDecl(format: Format, leadingTrivia: Trivia) -> DeclSyntax {
    let importDecl = SyntaxFactory.makeImportDecl(
      attributes: attributes, 
      modifiers: modifiers, 
      importTok: importTok, 
      importKind: importKind, 
      path: path
    ).withLeadingTrivia(leadingTrivia)
    return DeclSyntax(importDecl)
  }
}

// Syntax collection
public struct CodeBlockItemList: SyntaxBuildable {
  let elements: [CodeBlockItem]

  public init(elements: [CodeBlockItem]) {
    elements = elements
  }

  public func buildSyntax(format: Format, leadingTrivia: Trivia) -> DeclSyntax {
    let codeBlockItemList = makeCodeBlockItemList(elements)
      .withLeadingTrivia(leadingTrivia)
    return Syntax(codeBlockItemList)
}
```

Notice how each buildable takes exactly the arguments that are specified, so it’s essentially a wrapper around `SyntaxFactory`.

Once we have that, I imagine we can improve it as I described in the previous long post.

build-script.py
```diff
@@ -195,10 +195,10 @@ def generate_gyb_files(
 
     # Clear any *.swift files that are relics from the previous run.
     clear_gyb_files_from_previous_run(
-        SWIFTSYNTAX_DIR, swiftsyntax_destination)
```
kimdv, line 198: Found that if an gyb file was removed, the clean missed the verbose.
Should I add this in another PR so it faster get in the main branch?

Sources/SwiftSyntaxBuilder/DeclBuildables.swift.gyb
```diff
@@ -185,3 +192,55 @@ public struct Struct: DeclBuildable {
     return DeclSyntax(structDecl)
   }
 }
+
+// MARK: Generated
+
+% for node in SYNTAX_NODES:
```
kimdv, line 198: Okay, this ended up pretty messy. :( 
Should we add some methods in `Node.py` or is this okay?
Maybe I've missed some methods that already is in the swift repository that could make it more clean?

build-script.py
ahoppen: Oh, good catch. 
Yes that would be great if you could open another PR of that!

kimdv: @ahoppen opened in #269

ahoppen (commented):
Looks quite good already. I see it really starting to take shape. I’ve posted a few comments inline.

I think it would also be great to have one or two test cases for the buildables. Not only to test that they actually work, but also to give us a feeling for how the API currently behaves (and how we can improve it in the future 😉). Doesn’t have to be anything fancy. I think just an `enum` case, a struct decl or something like this should be enough, just pick whatever is easiest to construct.

Sources/SwiftSyntaxBuilder/DeclBuildables.swift.gyb
kimdv: I was seeing we have a list here of [kinds here](https://github.com/apple/swift/blob/main/utils/gyb_syntax_support/kinds.py#L5).

We could loop over them and generate the code protocols

ahoppen: Sounds good 👍

kimdv:
Okay @ahoppen I have pushed my progress now, but tests don't pass yet!

I have tried to make what you mentioned in (this comment)[https://github.com/apple/swift-syntax/pull/268#discussion_r598647295]. 

I found out that the API with `leadingTrivia` and `format` is a little hard (as I see it) to control, like have some custom indent.

If we take the SourceFileTests
We expect: 
```swift
    ␣import SwiftSyntax
    struct ExampleStruct {
        let syntax: Syntax
    }
```

But get 
```swift
␣import ␣SwiftSyntax␣struct ExampleStruct ␣{
  ␣let ␣syntax␣: ␣Syntax}
```

I have also tried to add some of the convince extensions for `ImportDecl`, `SourceFile`, `StructDecl`.
I have also tried to make some generic `Var` and `Let`. 


The Python code is still some messy for me. 
Etc there is some special handling for two types [here](https://github.com/apple/swift-syntax/pull/268/files#diff-1692562863ad5b0153cfede0d8b3fa518b3792f644fa2425ad41b1ac2f9ed33fR123).
I found there is an `IdentifierToken` and `Token`. They didn't fit in the other cases. 
Maybe you can guide me to some place, so it can be more clean and easier to read?

Thanks!

ahoppen:
## Trivia and indentation

> I found out that the API with leadingTrivia and format is a little hard (as I see it) to control, like have some custom indent.

Really good indentation will probably be hard to get to work in any case, so for now I’d settle with indentation that compiles, is not too horrible to look at and which e.g. [swift-format](https://github.com/apple/swift-format) could improve. But what we have right now is definitely not what we want.

What do you think about adding a newline after every line in a code block and after every member declaration? I.e. add a boolean flag to [MemberDeclList](https://github.com/apple/swift/blob/main/utils/gyb_syntax_support/DeclNodes.py#L296) and [CodeBlock](https://github.com/apple/swift/blob/main/utils/gyb_syntax_support/CommonNodes.py#L41) like `elements_separated_by_newline` and if that flag is set, in the builder you then add a newline between the items like so:

```swift
SyntaxFactory.make${node.syntax_kind}(elements.map { 
  $0.build${node.collection_element}(format: format, leadingTrivia: .newlines(1) + format._makeIndent())
})
```
(side node: you are currently passing `leadingTrivia` down to every child. I think that’s not correct since the leading trivia parameter is being handled by the `withLeadingTrivia` [here](https://github.com/apple/swift-syntax/pull/268/files#diff-1692562863ad5b0153cfede0d8b3fa518b3792f644fa2425ad41b1ac2f9ed33fR145). The `leadingTrivia` should only add that trivia to the currently constructed node, not to *every* nested node. `format` is responsible for doing that)

With that, I think, we would get the following result:

```swift
import␣SwiftSyntax␣
struct ExampleStruct ␣{
let␣syntax␣:␣Syntax
let␣syntax2␣:␣Syntax}
```

which is already a lot better (and compiles 🎉). Now two things remain:
1. Indent the members inside the struct
2. Add a newline before the closing brace

For 1.) I would suggest to add a property `is_indented` to the `Statements` child of `CodeBlock` [here](https://github.com/apple/swift/blob/main/utils/gyb_syntax_support/CommonNodes.py#L49) (and similarly for the member decls). If that property is set, you don’t pass down `format` [here](https://github.com/apple/swift-syntax/pull/268/files#diff-1692562863ad5b0153cfede0d8b3fa518b3792f644fa2425ad41b1ac2f9ed33fR103) but `format._indented()`, just like it is currently done in the [old Struct Syntax buildable](https://github.com/apple/swift-syntax/blob/cb7298fb1e9e0fb559a69cd7d54302e151671794/Sources/SwiftSyntaxBuilder/DeclBuildables.swift#L176), e.g.
```python
if child.is_indented:
  let format = format._indented()
```

For 2.) I would suggest adding a property `requires_leading_newline` to the `RightBrace` child of `CodeBlock` [here](https://github.com/apple/swift/blob/main/utils/gyb_syntax_support/CommonNodes.py#L51) (again similarly for member decls). If that property is set, you again use `.newline(1) + format._makeIndent()` as the child’s leading trivia [here](https://github.com/apple/swift-syntax/pull/268/files#diff-1692562863ad5b0153cfede0d8b3fa518b3792f644fa2425ad41b1ac2f9ed33fR103). Again, as noted above, passing down `leadingTrivia` like you do today, isn’t correct.

With 1.) and 2.) in, I think we should get what we want

```swift
import␣SwiftSyntax␣
struct ExampleStruct ␣{
␣␣let␣syntax␣:␣Syntax
␣␣let␣syntax2␣:␣Syntax
}
```

A background note on my suggestion: I would implement the entire indentation behavior on the level of the `Child` classes in Python, not the `Node` classes because the indentation is IMHO the property of a syntax node occurring at a specific location inside a parent, not a property of the syntax node itself. For `SyntaxCollection`s the `Child` nodes are implicit. Here the `elements_separated_by_newline` is the equivalent of having a `requires_leading_newline` on every of its implicit child nodes.

## Convenience extensions

> I have also tried to add some of the convince extensions for ImportDecl, SourceFile, StructDecl.

I would like to keep any convenience extensions out of this PR for now to (a) make it smaller and easier to review and (b) to see how far we can get with the purely generated code. If this means you need to delete some test cases, that’s totally fine with me. We can always add them again later.

## Python code

> The Python code is still some messy for me.

Actually, I think it’s fine. I just skimmed over it and was able to understand what’s going on. Perhaps we can move the code that determines what type a `SyntaxBuildable` child should have to a function to avoid duplication and make it clearer. E.g.

```python
def syntax_buildable_child_type(syntax_kind, is_optional=False):
  if 'Token' in syntax_kind: # Alternatively, add a `is_token` method to `Node`
    buildable_type = 'TokenSyntax'
  elif node.syntax_kind in SYNTAX_BASE_KINDS:
    buildable_type = syntax_kind + 'Buildable'
  else:
    buildable_type = syntax_kind

  if is_optional:
    buildable_type += '?'

  return buildable_type
```

and then use it in [line 69](https://github.com/apple/swift-syntax/pull/268/files#diff-1692562863ad5b0153cfede0d8b3fa518b3792f644fa2425ad41b1ac2f9ed33fR69) and [line 123](https://github.com/apple/swift-syntax/pull/268/files#diff-1692562863ad5b0153cfede0d8b3fa518b3792f644fa2425ad41b1ac2f9ed33fR123).

> I found there is an IdentifierToken and Token. They didn't fit in the other cases.

`IdentifierToken`s are just plain `Tokens` as far as the `SwiftSyntax` API is concerned. If you use [is_token()](https://github.com/apple/swift/blob/63d5cf1e5efff46698680a81b3137648a265dbc7/utils/gyb_syntax_support/Child.py#L69) on `Child` the two types are treated exactly the same.

kimdv:
> ## Trivia and indentation
> > I found out that the API with leadingTrivia and format is a little hard (as I see it) to control, like have some custom indent.
> 
> Really good indentation will probably be hard to get to work in any case, so for now I’d settle with indentation that compiles, is not too horrible to look at and which e.g. [swift-format](https://github.com/apple/swift-format) could improve. But what we have right now is definitely not what we want.
> 
> What do you think about adding a newline after every line in a code block and after every member declaration? I.e. add a boolean flag to [MemberDeclList](https://github.com/apple/swift/blob/main/utils/gyb_syntax_support/DeclNodes.py#L296) and [CodeBlock](https://github.com/apple/swift/blob/main/utils/gyb_syntax_support/CommonNodes.py#L41) like `elements_separated_by_newline` and if that flag is set, in the builder you then add a newline between the items like so:
> 
> ```swift
> SyntaxFactory.make${node.syntax_kind}(elements.map { 
>   $0.build${node.collection_element}(format: format, leadingTrivia: .newlines(1) + format._makeIndent())
> })
> ```
> 
> (side node: you are currently passing `leadingTrivia` down to every child. I think that’s not correct since the leading trivia parameter is being handled by the `withLeadingTrivia` [here](https://github.com/apple/swift-syntax/pull/268/files#diff-1692562863ad5b0153cfede0d8b3fa518b3792f644fa2425ad41b1ac2f9ed33fR145). The `leadingTrivia` should only add that trivia to the currently constructed node, not to _every_ nested node. `format` is responsible for doing that)
> 
> With that, I think, we would get the following result:
> 
> ```swift
> import␣SwiftSyntax␣
> struct ExampleStruct ␣{
> let␣syntax␣:␣Syntax
> let␣syntax2␣:␣Syntax}
> ```
> 
> which is already a lot better (and compiles 🎉). Now two things remain:
> 
> 1. Indent the members inside the struct
> 2. Add a newline before the closing brace
> 
> For 1.) I would suggest to add a property `is_indented` to the `Statements` child of `CodeBlock` [here](https://github.com/apple/swift/blob/main/utils/gyb_syntax_support/CommonNodes.py#L49) (and similarly for the member decls). If that property is set, you don’t pass down `format` [here](https://github.com/apple/swift-syntax/pull/268/files#diff-1692562863ad5b0153cfede0d8b3fa518b3792f644fa2425ad41b1ac2f9ed33fR103) but `format._indented()`, just like it is currently done in the [old Struct Syntax buildable](https://github.com/apple/swift-syntax/blob/cb7298fb1e9e0fb559a69cd7d54302e151671794/Sources/SwiftSyntaxBuilder/DeclBuildables.swift#L176), e.g.
> 
> ```python
> if child.is_indented:
>   let format = format._indented()
> ```
> 
> For 2.) I would suggest adding a property `requires_leading_newline` to the `RightBrace` child of `CodeBlock` [here](https://github.com/apple/swift/blob/main/utils/gyb_syntax_support/CommonNodes.py#L51) (again similarly for member decls). If that property is set, you again use `.newline(1) + format._makeIndent()` as the child’s leading trivia [here](https://github.com/apple/swift-syntax/pull/268/files#diff-1692562863ad5b0153cfede0d8b3fa518b3792f644fa2425ad41b1ac2f9ed33fR103). Again, as noted above, passing down `leadingTrivia` like you do today, isn’t correct.
> 
> With 1.) and 2.) in, I think we should get what we want
> 
> ```swift
> import␣SwiftSyntax␣
> struct ExampleStruct ␣{
> ␣␣let␣syntax␣:␣Syntax
> ␣␣let␣syntax2␣:␣Syntax
> }
> ```
> 
> A background note on my suggestion: I would implement the entire indentation behavior on the level of the `Child` classes in Python, not the `Node` classes because the indentation is IMHO the property of a syntax node occurring at a specific location inside a parent, not a property of the syntax node itself. For `SyntaxCollection`s the `Child` nodes are implicit. Here the `elements_separated_by_newline` is the equivalent of having a `requires_leading_newline` on every of its implicit child nodes.

I have added your suggestions and almost something that generates some code (But will not compile yet).
I have some thoughts that I wanted some feedback on before adding it.

Should we add something like `requires_leading_space` for places like (here)[https://github.com/apple/swift/pull/36726/files#diff-de4680aa2bbb7151cb6595294ce91fb29fec2ecd719ae270ae1cdbd4c37a3c87R48] (and all other related places, where there is a left brace or Identifier etc).
And if `requires_leading_space ` is set we will add a `withLeadingTrivia(.spaces(1))`. 
The reason is, that now we are generating 
```␣structTestStruct{
}```

Or can we make some general rule somewhere for applying all places so we don't need to add the something multiple places?

> ## Python code
> > The Python code is still some messy for me.
> 
> Actually, I think it’s fine. I just skimmed over it and was able to understand what’s going on. Perhaps we can move the code that determines what type a `SyntaxBuildable` child should have to a function to avoid duplication and make it clearer. E.g.
> 
> ```python
> def syntax_buildable_child_type(syntax_kind, is_optional=False):
>   if 'Token' in syntax_kind: # Alternatively, add a `is_token` method to `Node`
>     buildable_type = 'TokenSyntax'
>   elif node.syntax_kind in SYNTAX_BASE_KINDS:
>     buildable_type = syntax_kind + 'Buildable'
>   else:
>     buildable_type = syntax_kind
> 
>   if is_optional:
>     buildable_type += '?'
> 
>   return buildable_type
> ```
> 
> and then use it in [line 69](https://github.com/apple/swift-syntax/pull/268/files#diff-1692562863ad5b0153cfede0d8b3fa518b3792f644fa2425ad41b1ac2f9ed33fR69) and [line 123](https://github.com/apple/swift-syntax/pull/268/files#diff-1692562863ad5b0153cfede0d8b3fa518b3792f644fa2425ad41b1ac2f9ed33fR123).
> 
> > I found there is an IdentifierToken and Token. They didn't fit in the other cases.
> 
> `IdentifierToken`s are just plain `Tokens` as far as the `SwiftSyntax` API is concerned. If you use [is_token()](https://github.com/apple/swift/blob/63d5cf1e5efff46698680a81b3137648a265dbc7/utils/gyb_syntax_support/Child.py#L69) on `Child` the two types are treated exactly the same.

I have tried to add it to the gyb file? 
Maybe I have misunderstood something, but should it be added in `Buildables.swift.gyb` or as an method on node/child?

**Comment by ahoppen** [comment]
I haven’t looked at the code yet. Let me know once you think that is ready for review.

> Should we add something like requires_leading_space for places like [here](https://github.com/apple/swift/pull/36726/files#diff-de4680aa2bbb7151cb6595294ce91fb29fec2ecd719ae270ae1cdbd4c37a3c87R48) (and all other related places, where there is a left brace or Identifier etc).
And if requires_leading_space is set we will add a withLeadingTrivia(.spaces(1)).

I don’t think that’s necessary. IIUC when we generate the struct-Keyword with `Tokens.struct`, we already receive `struct␣`. For now I think we can put it in the user’s responsibility to create an identifier token with a trailing space so that we would get
```swift
struct␣TestStruct␣{
}
```

Please correct me if I’m wrong.

Later, when we create convenience constructors that take Strings instead of `IdentifierToken`s, these could add the trailing space.

> I have tried to add it to the gyb file?
Maybe I have misunderstood something, but should it be added in Buildables.swift.gyb or as an method on node/child?

Sorry, I don’t quite understand the question. What is “it” referring to?

Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
```diff
@@ -0,0 +1,172 @@
...
+  from gyb_syntax_support.kinds import SYNTAX_BASE_KINDS
+  # -*- mode: Swift -*-
+  # Ignore the following admonition it applies to the resulting .swift file only
+
+  def syntax_buildable_child_type(syntax_kind, is_token=False, is_optional=False):
+    if syntax_kind in SYNTAX_BASE_KINDS:
+      buildable_type = syntax_kind + 'Buildable'
+    elif not is_token:
+      buildable_type = syntax_kind
+    elif 'List' in syntax_kind:
```
kimdv, line 13: This is needed to add correct type (herre)[https://github.com/apple/swift-syntax/pull/268/files#diff-685068f90f6719a81556c55a8d5cd4e0b2d4f3fe395a24a21af3dbcbb1d46165R1539] etc.

Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
```diff
@@ -0,0 +1,172 @@
...
+
+import SwiftSyntax
+
+// MARK: Protocols
+
+% for kind in SYNTAX_BASE_KINDS:
+%   if kind != 'SyntaxCollection':
+%     build_kind = 'Syntax' if kind == 'Syntax' else kind + 'Syntax'
+%     if kind == 'Syntax':
+public protocol ${kind}ListBuildable {
```
kimdv, line 44: Should we add a convenience extension without `leadingTrivia` so we don't need to pass `nil`?

Sources/SwiftSyntaxBuilder/gyb_generated/Buildables.swift
```diff
@@ -0,0 +1,8325 @@
...
+
+public struct ClosureCaptureItem: SyntaxBuildable {
+  let specifier: TokenListSyntax?
+  let name: TokenSyntax?
+  let assignToken: TokenSyntax?
+  let expression: ExprBuildable
+  let trailingComma: TokenSyntax?
+
+  public init(
+    specifier: TokenListSyntax?,
```
kimdv, line 1546: Should we add `specifier: TokenListSyntax? = nil` here so it is possible to skip the parameters?

kimdv:
Okay @ahoppen, after some evenings working on it, I have something that works now. 
I have commented below. 

> I haven’t looked at the code yet. Let me know once you think that is ready for review.
> 
> > Should we add something like requires_leading_space for places like [here](https://github.com/apple/swift/pull/36726/files#diff-de4680aa2bbb7151cb6595294ce91fb29fec2ecd719ae270ae1cdbd4c37a3c87R48) (and all other related places, where there is a left brace or Identifier etc).
> > And if requires_leading_space is set we will add a withLeadingTrivia(.spaces(1)).
> 
> I don’t think that’s necessary. IIUC when we generate the struct-Keyword with `Tokens.struct`, we already receive `struct␣`. For now I think we can put it in the user’s responsibility to create an identifier token with a trailing space so that we would get
> 
> ```swift
> struct␣TestStruct␣{
> }
> ```
> 
> Please correct me if I’m wrong.
> 
> Later, when we create convenience constructors that take Strings instead of `IdentifierToken`s, these could add the trailing space.

If you have time, it would be nice if you would look a little on the tests I have added.
I think it will give you an idea on how the API works now and if it matches your vision.
I have added some comments in the code, with some thoughts/ideas

> > I have tried to add it to the gyb file?
> > Maybe I have misunderstood something, but should it be added in Buildables.swift.gyb or as an method on node/child?
> 
> Sorry, I don’t quite understand the question. What is “it” referring to?

I figured it out, had some problems where to add the code, that is added [here](https://github.com/apple/swift-syntax/pull/268/files#diff-1692562863ad5b0153cfede0d8b3fa518b3792f644fa2425ad41b1ac2f9ed33fR8).

I first tried to add it add bottom of the gyb file, like this
```
%  def syntax_buildable_child_type(syntax_kind, is_token=False, is_optional=False):
%    if syntax_kind in SYNTAX_BASE_KINDS:
%      buildable_type = syntax_kind + 'Buildable'
%    elif not is_token:
%      buildable_type = syntax_kind
%
%    else:
%      buildable_type = 'TokenSyntax'
%
%    if is_optional:
%      buildable_type += '?'
%
%    return buildable_type
```

Again, thanks for your patience 😁

Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
```diff
@@ -0,0 +1,172 @@
...
+%         end            
+%         param_name = param_name + ".build" + child.syntax_kind + "(format: " + format + ", leadingTrivia: nil)"
+%       end
+%       child_params.append("%s: %s" % (child.swift_name, param_name))
+%     end
+    let ${node.swift_syntax_kind} = SyntaxFactory.make${node.syntax_kind}(
+      ${',\n      '.join(child_params)}
+    )
+    
+    if let leadingTrivia = leadingTrivia {
```
kimdv, line 120: We need to do this, because we otherwise need to pass `.zero`.

Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
```diff
@@ -0,0 +1,172 @@
...
+%       if node.elements_separated_by_newline:
+%         leading_trivia = '.newlines(1) + format._makeIndent()'
+%       end
+    let collection = SyntaxFactory.make${node.syntax_kind}(elements.map {
+      $0.build${node.collection_element}(format: format, leadingTrivia: ${leading_trivia})
+    })
+%     else:
+    let collection = SyntaxFactory.make${node.syntax_kind}(elements)
+%     end
+      return collection
```
ahoppen, line 157: Any reason why you are creating `collection` and are not directly returning it?

Tests/SwiftSyntaxBuilderTest/EnumCaseElementTests.swift
```diff
@@ -0,0 +1,32 @@
...
+
+final class EnumCaseElementTests: XCTestCase {
+  func test_enumCaseElementInit() {
+    let leadingTrivia = Trivia.garbageText("␣")
+    let string = SyntaxFactory.makeStringSegment("Hello World")
+    let segment = StringSegment(content: string)
+    let segments = StringLiteralSegments([segment])
+
+    let stringLiteralExpr: ExprBuildable = StringLiteralExpr(openDelimiter: nil,
+                                                             openQuote: SyntaxFactory.makeStringQuoteToken(),
```
ahoppen, line 13: Can’t you just use `.stringQuote` which you defined in your last PR [here](https://github.com/apple/swift-syntax/blob/main/Sources/SwiftSyntaxBuilder/gyb_generated/Tokens.swift#L339)? Also further down for every use of `SyntaxFactory.make*Token`.

Since the token convenience constructors from you last PR also already have the trivia set, I think you don’t need any more calls to `with*Trivia` on them.

ahoppen (commented):
I’ve added some comments inline but they are all mostly stylistic now, nothing major. 

Once you’ve addressed them, I will give the PR another review, run the test suite and get this merged. I’ve lost track but are there any related. If there are any, could you also push any pending changes to your apple/swift PR for that?

The tests look really good to me and I think we are on the right track to some really nice and exhaustive API if you imagine the following changes (which I believe are minor/trivial, but still would like to keep out of this PR)! 🎉 
1. Allow omitting arguments that are `nil`
2. Allow omitting arguments that take a specific token and default them to that token
3. Allow passing string literals for identifiers
4. Create a result builder API for `SyntaxCollections`

Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
```diff
@@ -0,0 +1,172 @@
...
+    
+    if let leadingTrivia = leadingTrivia {
+      return ${node.swift_syntax_kind}
+        .withLeadingTrivia(leadingTrivia)
+    }
+
+    return ${node.swift_syntax_kind}
+  }
+
+  public func build${node.base_kind}(format: Format, leadingTrivia: Trivia?) -> ${return_type} {
```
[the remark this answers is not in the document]
kimdv: Stupid question, what is "IIUC"?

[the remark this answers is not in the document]
ahoppen: It’s an abbreviation for “If I understand correctly”. There’s a couple of these that get frequently used, Urban Dictionary usually has explanations for them: https://www.urbandictionary.com/define.php?term=IIUC

Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
kimdv: I have just tried this, and ended up with something that we wanted.

Look at the example below:
```swift
public struct AvailabilityVersionRestriction: SyntaxBuildable {
  let platform: TokenSyntax
  let version: VersionTupleSyntax?

  public init(
    platform: TokenSyntax,
    version: VersionTupleSyntax?
  ) {
    self.platform = platform
    self.version = version
  }

  ...
}
```

But we want 

```swift
public struct AvailabilityVersionRestriction: SyntaxBuildable {
  let platform: TokenSyntax
  let version: VersionTuple?

  public init(
    platform: TokenSyntax,
    version: VersionTuple?
  ) {
    self.platform = platform
    self.version = version
  }

  ...
}
```

Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
```diff
@@ -0,0 +1,172 @@
...
+
+% for kind in SYNTAX_BASE_KINDS:
+%   if kind != 'SyntaxCollection':
+%     build_kind = 'Syntax' if kind == 'Syntax' else kind + 'Syntax'
+%     if kind == 'Syntax':
+public protocol ${kind}ListBuildable {
+%     else:
+public protocol ${kind}ListBuildable: SyntaxListBuildable {
+%     end
+  func build${kind}List(format: Format, leadingTrivia: Trivia?) -> [${build_kind}]
```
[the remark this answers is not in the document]
kimdv: The reason is that if we pass `.zero` is replaces the leading `Trivia`.

We expect 
`␣import SwiftSyntax`
or 
```swift
␣struct TestStruct {
}
```
But get 

but get `␣importSwiftSyntax`

```swift
␣struct TestStruct{
}
```

Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
```diff
@@ -0,0 +1,172 @@
...
+% end
+
+// MARK: - Buildables
+
+% for node in SYNTAX_NODES:
+%   if node.is_buildable():
+%     return_type = 'Syntax'
+%     if node.base_kind != 'Syntax':
+%       return_type = node.base_kind + 'Syntax'
+%     end
```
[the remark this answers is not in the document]
kimdv: Moved it down. 
But I think we should still call it `return_type` then, because it is only used for that now.

Tests/SwiftSyntaxBuilderTest/EnumCaseElementTests.swift
kimdv: It is not public. 
We could make them public and I think actually we should do that

Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
```diff
@@ -0,0 +1,179 @@
...
+}
+
+%   end
+% end
+
+// MARK: - Buildables
+
+% for node in SYNTAX_NODES:
+%   if node.is_buildable():
+public struct ${node.syntax_kind}: ${node.base_kind}Buildable {
```
kimdv, line 87: @ahoppen would it also make sense to add documentation here`
```
%     for line in dedented_lines(node.description):
/// ${line}
%     end
```

Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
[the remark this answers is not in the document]
kimdv: I have added a comment. Looked a bit on how it was done in the SwiftSyntax. 
Is it good enough or should there be more? 😁

kimdv:
I have address more less all off them, I have some small questions and added some comments.

> I’ve added some comments inline but they are all mostly stylistic now, nothing major.
> 
> Once you’ve addressed them, I will give the PR another review, run the test suite and get this merged. I’ve lost track but are there any related. If there are any, could you also push any pending changes to your apple/swift PR for that?

Yes, there is one here: https://github.com/apple/swift/pull/36726

Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
kimdv: And maybe also for the `init` and all the children?

Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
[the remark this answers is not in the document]
ahoppen: Ah, I see. Makes total sense. Could you add a comment describing the behaviour?

Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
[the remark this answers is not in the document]
ahoppen: Yeah, looks good now that the variable definition is close to its use.

Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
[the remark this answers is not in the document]
ahoppen: 👍 

I think it’s OK. I wouldn't object to more (I think almost nobody every objects to more good comments) but there’s 
nothing pressing I can think of.

Tests/SwiftSyntaxBuilderTest/EnumCaseElementTests.swift
ahoppen: I didn’t realize that. I think they should definitely be public, otherwise I can’t think of what they are good for.

Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
ahoppen: Yeah, would be good to use the documentation we already have. It doesn’t have to be in this PR though if you’d rather want to get this merged as soon as possible and tackle it in a follow-up PR. However you like it.

Tests/SwiftSyntaxBuilderTest/EnumCaseElementTests.swift
kimdv: Should it be in this PR or another? 
What do you prefer?

ahoppen: Let’s put it in a follow-up PR before any more discussions about it come up, complicating this PR even further.

ahoppen:
Let’s see what CI has to say about this.

https://github.com/apple/swift/pull/36726

@swift-ci Please test

Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
kimdv: Will do it another.

Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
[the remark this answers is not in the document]
kimdv: Would some documentation like this help?

```
  /// Builds list of `${kind}`s
  /// - Parameter format: The `Format` to use.
  /// - Parameter leadingTrivia: Replaces the the last leading trivia if not nil.
```

ahoppen:
🎉 CI passed. I’m proposing that we merge this as is and you address the following outstanding things in follow-up PRs:

- Add comment describing behavior of `nil` `leadingTrivia` (https://github.com/apple/swift-syntax/pull/268#discussion_r616595072)
- Use `Token` static properties instead of constructing tokens manually in tests (https://github.com/apple/swift-syntax/pull/268#discussion_r616522144)
- Add syntax node description comments to buildables (https://github.com/apple/swift-syntax/pull/268#discussion_r616537232)

ahoppen (approved):

kimdv:
Yes!
Let's get it merged and I will fix the 3 things you mentioned after. 

I just rebased on `main`.

ahoppen:
apple/swift#36726

@swift-ci Please test

ahoppen:
If it didn’t come through when you put up the PR and during review, I wanted to give you a huge thank you for working on this and having the perseverance to see this patch through to the end (I only now realized that it was started more than a month ago). 🙏🏼

ktoso:
I love this work so much, thank you @kimdv! 🥳

kimdv:
You're welcome! 
And again thanks for being patient for my, some times stupid, questions 🎉

PR #271, opened 2021-03-26
Add precondition that a token’s range returned by the Parser is in the source file

ahoppen:
We have been hitting crashes in
```swift
let begin = utf8.index(utf8.startIndex, offsetBy: startOffset)
```
presumably because `startOffset` is outside of the `UTF8View`’s bounds. I still can’t explain how this happens.

To further narrow down the issue, I propose we add preconditions that validate the start and end location of the range. This way, we will be crashing in our own code, hopefully giving us more insight into why the crash is happening.

Whenever these preconditions are violated, we should already be crashing today.


akyrtzi (approved):

ahoppen:
@swift-ci Please test

PR #272, opened 2021-03-26
[build] Invoke lit.py with python3

edymtt:
Also provide an environment variable to allow to launch `lit.py`
directly.

LLVM is transitioning its script infrastructure to python3 and we may
run into issues if we continue using the system-default python2 on some
platforms.

Addresses rdar://75886813


edymtt:
@swift-ci please test

edymtt:
@swift-ci please test

ahoppen (approved):
Approving if this is transitional and we want to remove the environment variable again in the future.

edymtt:
Thanks -- in the long term we would like to move to `llvm-lit` instead (which references the correct Python), the environment variable was meant as an "escape hatch" in case my testing did not account for some scenarios.

PR #278, opened 2021-04-24
Remove parameter that is not used

kimdv:


Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
```diff
@@ -150,13 +149,13 @@ public struct ${node.syntax_kind}: SyntaxBuildable {
     self.elements = elements
   }
 
-  public func build${node.syntax_kind}(format: Format, leadingTrivia: Trivia? = nil) -> ${node.syntax_kind}Syntax {
+  public func build${node.syntax_kind}(format: Format) -> ${node.syntax_kind}Syntax {
```
kimdv, line 152: `leadingTrivia` was never used 🤦‍♂️

Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
```diff
@@ -150,13 +145,13 @@ public struct ${node.syntax_kind}: SyntaxBuildable {
...
 %     if not node.is_token():
 %       leading_trivia = 'nil'
 %       if node.elements_separated_by_newline:
 %         leading_trivia = '.newlines(1) + format._makeIndent()'
 %       end
-    return SyntaxFactory.make${node.syntax_kind}(elements.map {
+    SyntaxFactory.make${node.syntax_kind}(elements.map {
```
ahoppen, line 154: Small nitpick: I would prefer to keep the `return` to make it more explicit what the function is doing (especially since the statement in the `else` branch has a `return` keyword. gyb is hard enough to read even without these syntactic shortcuts.

ahoppen (commented):
One small nitpick, otherwise looks good to me. Thanks for cleaning it up.

ahoppen (approved):

ahoppen:
@swift-ci Please test

PR #280, opened 2021-04-29
Omit parameters in init that is optional

kimdv:


Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
```diff
@@ -81,17 +81,23 @@ extension ${kind}Buildable {
...
+%     let_variables = []
 %     for child in node.children:
 %       param_type = syntax_buildable_child_type(child.type_name, child.syntax_kind, child.is_token(), child.is_optional)
-%       child_params.append("%s: %s" % (child.swift_name, param_type))
+%       let_variables.append("%s: %s" % (child.swift_name, param_type))
 %     end
-%     for child_param in child_params:
-  let ${child_param}
+%     for let_variable in let_variables:
+  let ${let_variable}
```
ahoppen, line 90: Since we are no longer using `let_variables` for the initializer, I think we can just print the straight away instead of storing them in an array first.

ahoppen (commented):
One small comment inline, otherwise looks good to me.

Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
kimdv: Done! That was a little brain fart! 🤦‍♂️

ahoppen (approved):
LGTM

ahoppen:
@swift-ci Please test

PR #283, opened 2021-05-01
build: use Foundation instead of C for `getenv`

compnerd:
`getenv` is different across different platforms.  Use Foundation to
abstract over the platform differences.

This cherry-picks the build changes from main to the 5.4 release to enable building on Windows.


compnerd:
@swift-ci please test

harlanhaskins (approved):

PR #284, opened 2021-05-09
Add convenience inits

kimdv:
Hi @ahoppen ! 

I have been working on some of the convenience inits. Not sure if I should split more up or just make some more in this PR? 

What do you think? 
Have you any ideas for more inits that I should add?

### DIFF
--- a/Sources/SwiftSyntaxBuilder/ResultBuilders.swift.gyb
+++ b/Sources/SwiftSyntaxBuilder/ResultBuilders.swift.gyb
@@ -1,0 +1,39 @@
+%{
+  from gyb_syntax_support import *
+  from gyb_syntax_support.kinds import SYNTAX_BASE_KINDS, lowercase_first_word, syntax_buildable_child_type
+  NODE_MAP = create_node_map()
+  # -*- mode: Swift -*-
+  # Ignore the following admonition it applies to the resulting .swift file only
+}%
+//// Automatically Generated From ResultBuilders.swift.gyb.
+//// Do Not Edit Directly!
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2021 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
+% for node in SYNTAX_NODES:
+%   if node.is_syntax_collection():
+@resultBuilder
+public struct ${node.syntax_kind}Builder {
+%     element_type = syntax_buildable_child_type(node.collection_element_type, node.collection_element, node.is_token())
+  public static func buildBlock(_ elements: ${element_type}...) -> ${node.syntax_kind} {
+    ${node.syntax_kind}(elements)
+  }
+}
+
+extension ${node.syntax_kind} {
+  public static let empty: ${node.syntax_kind} = ${node.syntax_kind}([])
+}
+
+%   end
+% end
--- a/Tests/SwiftSyntaxBuilderTest/StructTests.swift
+++ b/Tests/SwiftSyntaxBuilderTest/StructTests.swift
@@ -9,13 +9,8 @@ final class StructTests: XCTestCase {
     let members = MemberDeclBlock(leftBrace: Tokens.leftBrace.withLeadingTrivia(.spaces(1)),
                                   members: MemberDeclList([]),
                                   rightBrace: Tokens.rightBrace)
-    let buildable = StructDecl(attributes: nil,
-                               modifiers: nil,
-                               structKeyword: Tokens.struct,
+    let buildable = StructDecl(structKeyword: Tokens.struct,
                                identifier: SyntaxFactory.makeIdentifier("TestStruct"),
-                               genericParameterClause: nil,
-                               inheritanceClause: nil,
-                               genericWhereClause: nil,
                                members: members)
     let syntax = buildable.buildSyntax(format: Format(), leadingTrivia: leadingTrivia)
 
@@ -30,19 +25,18 @@ final class StructTests: XCTestCase {
 
   func testNestedStruct() {
     let leadingTrivia = Trivia.garbageText("␣")
-    let structKeyword = SyntaxFactory.makeStructKeyword()
     let emptyMembers = MemberDeclBlock(leftBrace: Tokens.leftBrace.withLeadingTrivia(.spaces(1)),
                                   members: MemberDeclList([]),
                                   rightBrace: Tokens.rightBrace)
-    let nestedStruct = StructDecl(structKeyword: structKeyword,
-                               identifier: SyntaxFactory.makeIdentifier("NestedStruct").withLeadingTrivia(.spaces(1)),
+    let nestedStruct = StructDecl(structKeyword: Tokens.struct,
+                               identifier: SyntaxFactory.makeIdentifier("NestedStruct"),
                                members: emptyMembers)
     let members = MemberDeclBlock(leftBrace: Tokens.leftBrace.withLeadingTrivia(.spaces(1)),
                                   members: MemberDeclList([MemberDeclListItem(decl: nestedStruct)]),
                                   rightBrace: Tokens.rightBrace)
-    let testStruct = StructDecl(structKeyword: structKeyword,
-                               identifier: SyntaxFactory.makeIdentifier("TestStruct").withLeadingTrivia(.spaces(1)),
-                               members: members)
+    let testStruct = StructDecl(structKeyword: Tokens.struct,
+                                identifier: SyntaxFactory.makeIdentifier("TestStruct"),
+                                members: members)
     let syntax = testStruct.buildSyntax(format: Format(), leadingTrivia: leadingTrivia)
 
     var text = ""
@@ -51,6 +45,40 @@ final class StructTests: XCTestCase {
     XCTAssertEqual(text, """
     ␣struct TestStruct {
         struct NestedStruct {
+        }
+    }
+    """)
+  }
+
+  func testEmptyStructWithBuilder() {
+    let leadingTrivia = Trivia.garbageText("␣")
+    let test = StructDecl(identifier: "TestStruct", members: MemberDeclBlock())
+
+    let syntax = test.buildSyntax(format: Format(), leadingTrivia: leadingTrivia)
+
+    var text = ""
+    syntax.write(to: &text)
+
+    XCTAssertEqual(text, """
+    ␣struct TestStruct{
+    }
+    """)
+  }
+
+  func testNestedStructWithBuilder() {
+    let leadingTrivia = Trivia.garbageText("␣")
+    let memberDeclBlock = MemberDeclBlock {
+      MemberDeclListItem(decl: StructDecl(identifier: "NestedStruct", members: MemberDeclBlock()))
+    }
+    let test = StructDecl(identifier: "TestStruct", members: memberDeclBlock)
+    let syntax = test.buildSyntax(format: Format(), leadingTrivia: leadingTrivia)
+
+    var text = ""
+    syntax.write(to: &text)
+
+    XCTAssertEqual(text, """
+    ␣struct TestStruct{
+        struct NestedStruct{
         }
     }
     """)
--- a/Tests/SwiftSyntaxBuilderTest/VariableTests.swift
+++ b/Tests/SwiftSyntaxBuilderTest/VariableTests.swift
@@ -1,0 +1,43 @@
+import XCTest
+import SwiftSyntax
+
+import SwiftSyntaxBuilder
+
+final class VariableTests: XCTestCase {
+  func testVariableDecl() {
+    let leadingTrivia = Trivia.garbageText("␣")
+
+    let buildable = VariableDecl(letOrVarKeyword: Tokens.let) {
+        PatternBinding(pattern: IdentifierPattern(identifier: SyntaxFactory.makeIdentifier("color")),
+                       typeAnnotation: TypeAnnotation(colon: Tokens.colon, type: SimpleTypeIdentifier("UIColor")))
+    }
+
+    let syntax = buildable.buildSyntax(format: Format(), leadingTrivia: leadingTrivia)
+
+    var text = ""
+    syntax.write(to: &text)
+
+    XCTAssertEqual(text, """
+    ␣let color: UIColor
+    """)
+  }
+
+  func testVariableDeclWithValue() {
+    let leadingTrivia = Trivia.garbageText("␣")
+
+    let buildable = VariableDecl(letOrVarKeyword: Tokens.var) {
+        PatternBinding(pattern: IdentifierPattern(identifier: SyntaxFactory.makeIdentifier("number")),
+                       typeAnnotation: TypeAnnotation(colon: Tokens.colon, type: SimpleTypeIdentifier("Int")),
+                       initializer: InitializerClause(equal: Tokens.equal, value: IntegerLiteralExpr(digits: Tokens.integerLiteral("123"))))
+    }
+
+    let syntax = buildable.buildSyntax(format: Format(), leadingTrivia: leadingTrivia)
+
+    var text = ""
+    syntax.write(to: &text)
+
+    XCTAssertEqual(text, """
+    ␣var number: Int = 123
+    """)
+  }
+}
--- a/Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
+++ b/Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
@@ -2,23 +2,12 @@
   from gyb_syntax_support import *
   from gyb_syntax_support.kinds import lowercase_first_word
   from gyb_syntax_support.kinds import SYNTAX_BASE_KINDS
+  from gyb_syntax_support.kinds import syntax_buildable_child_type
   # -*- mode: Swift -*-
   # Ignore the following admonition it applies to the resulting .swift file only
-
-  def syntax_buildable_child_type(type_name, syntax_kind, is_token, is_optional=False):
-    if syntax_kind in SYNTAX_BASE_KINDS:
-      buildable_type = syntax_kind + 'Buildable'
-    elif not is_token:
-      buildable_type = syntax_kind
-    else:
-      buildable_type = type_name
-
-    if is_optional:
-      buildable_type += '?'
-
-    return buildable_type
 }%
 //// Automatically Generated From DeclBuildables.swift.gyb.
+//// Do Not Edit Directly!
 //===----------------------------------------------------------------------===//
 //
 // This source file is part of the Swift.org open source project
--- a/Sources/SwiftSyntaxBuilder/BuildablesConvenienceInitializers.swift.gyb
+++ b/Sources/SwiftSyntaxBuilder/BuildablesConvenienceInitializers.swift.gyb
@@ -1,0 +1,91 @@
+%{
+  from gyb_syntax_support import *
+  from gyb_syntax_support.kinds import SYNTAX_BASE_KINDS, lowercase_first_word, syntax_buildable_child_type
+  NODE_MAP = create_node_map()
+  # -*- mode: Swift -*-
+  # Ignore the following admonition it applies to the resulting .swift file only
+}%
+//// Automatically Generated From DeclBuildables.swift.gyb.
+//// Do Not Edit Directly!
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2021 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
+% for node in SYNTAX_NODES:
+%   if node.is_buildable():
+%     has_syntax_collection_child = False
+%     for child in node.children:
+%       child_node = NODE_MAP.get(child.syntax_kind)
+%       if child_node and child_node.is_syntax_collection():
+%         has_syntax_collection_child = True
+%       end
+%     end
+%     if has_syntax_collection_child:
+extension ${node.syntax_kind} {
+  public init(
+%       init_parameters = []
+%       for child in node.children:
+%         child_node = NODE_MAP.get(child.syntax_kind)
+%         token = SYNTAX_TOKEN_MAP.get(child.syntax_kind)
+%         if child_node and child_node.is_syntax_collection():
+%           # Allow initializing syntax collections with result builders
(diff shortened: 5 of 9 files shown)


Sources/SwiftSyntaxBuilder/DeclBuildables.swift
```diff
@@ -0,0 +1,34 @@
...
+    self.init(structKeyword: Tokens.struct,
+              identifier: SyntaxFactory.makeIdentifier(identifier),
+              members: MemberDeclBlock(leftBrace: Tokens.leftBrace.withLeadingTrivia(.spaces(1)),
+                                       members: membersBuilder(),
+                                       rightBrace: Tokens.rightBrace))
+  }
+}
+
+extension VariableDecl {
+  public init(_ letOrVarKeyword: TokenSyntax,
```
kimdv, line 27: Should we also have one for a more simple variable? That takes a identifier, type and expr as parameter?

ahoppen (commented):
I would prefer if we tried some generic approaches first before looking at hand-written convenience initializers. Based on the convenience initializers you wrote, I think we should first look into:
- Trying to generate result builder initializers for any syntax type that has a child which is a syntax collection
- Accepting string literals for all children that are of type `IdentifierToken`
- Default an initializer argument if it has a concrete keyword type (like the `struct` keyword for a struct declaration)

I think that way we wouldn’t need a hand-written convenience initializer for `StructDecl`, `VariableDecl` and `SimpleTypeIdentifier` at all.

I think the other convenience initializers (integer literal, string literal, pattern binding) make sense but would like to hold them back until we’ve tried as much as possible in terms of generic code.

Sources/SwiftSyntaxBuilder/DeclBuildables.swift.gyb
```diff
@@ -0,0 +1,83 @@
...
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
+% for node in SYNTAX_NODES:
+%   if node.is_buildable():
+%     should_generate_init = False
```
ahoppen, line 39: I think a better name for `should_generate_init` would be `has_syntax_collection_child`.

Sources/SwiftSyntaxBuilder/DeclBuildables.swift.gyb
```diff
@@ -0,0 +1,83 @@
...
+%         end
+%       end
+    ${',\n    '.join(init_parameters)}
+  ) {
+    self.init(
+%     init_parameters = []
+%     for child in node.children:
+%       child_node = NODE_MAP.get(child.syntax_kind)
+%       if child_node and child_node.is_syntax_collection():
+%         if child_node.collection_element is "Syntax":
```
ahoppen, line 67: I think the `if` and `else` branch are doing the same thing, so you don’t need the condition.

Sources/SwiftSyntaxBuilder/DeclBuildables.swift.gyb
```diff
@@ -0,0 +1,83 @@
+%{
+  from gyb_syntax_support import *
+  from gyb_syntax_support.kinds import lowercase_first_word
+  from gyb_syntax_support.kinds import SYNTAX_BASE_KINDS
+  NODE_MAP = create_node_map()
+  # -*- mode: Swift -*-
+  # Ignore the following admonition it applies to the resulting .swift file only
+  
+  def syntax_buildable_child_type(type_name, syntax_kind, is_token, is_optional=False):
```
ahoppen, line 9: Could you move this function to some common place in e.g. `gyb_syntax_support` that we can import instead of duplicating it?

ahoppen (commented):
> I have tried to generate result builders for each collection. Can you verify that I'm on the right path?

I have only looked at it from a high level so far, but I think it’s looking good so far. I’ve added a few comments inline that immediately jumped to my mind. Let me know when you’re ready and I’ll look at it in more detail.

> How would we do that? 
Should we see if a child name contains `Keyword` like `Child('StructKeyword', kind='StructToken')` and then have a default parameter `Tokens.struct` ?

That was my idea exactly. 

Could you do that in another PR though? And could you also create a separate PR for the other convenience initializers (like the one fro `SimpleTypeIdentifier`). It just makes it easier to review this one if anything that’s not result-builder specific is in a different PR.

Sources/SwiftSyntaxBuilder/DeclBuildables.swift.gyb
kimdv: That was actually a question. 
We have one place [here](https://github.com/apple/swift-syntax/pull/284/files/4c2ea03308554b7c35071e78d4d0d2e62571da77#diff-30af98704bd9407c6b3ee049cc324ffe08d525bb80f1760107f81497aafafb20R84) etc where we have a TokenListBuilder. 
It returns a `TokenList`.
But it expects a `TokenListSynx` [here](https://github.com/apple/swift-syntax/blob/main/Sources/SwiftSyntaxBuilder/gyb_generated/Buildables.swift#L1720)

So in this special case (I only found two places where it is used) we should maybe do something line 
`tokenListBuilder().buildTokenListSyntax(format: ...)`

So what do we need to do here? 
Should we ignore this case? 

It affects `ClosureCaptureItem` and `Attribute`

kimdv: `Cannot convert value of type 'TokenList' to expected argument type 'TokenListSyntax?'`

ahoppen: I think your “here“ comments don’t contain line numbers. Could you add them to your comment. That way I don’t have to search 😉

kimdv: GitHub won't add line numbers for the first "here", so line 84 in DeclBuildables.swift 😑🙃
The `ClosureCaptureItem`

The second "here" should direct to the line

Sources/SwiftSyntaxBuilder/DeclBuildables.swift.gyb
kimdv: Should it be placed in the swift or swift-syntax repo?

Sources/SwiftSyntaxBuilder/DeclBuildables.swift.gyb
ahoppen: I think [Buildables.swift:1720](https://github.com/apple/swift-syntax/blob/main/Sources/SwiftSyntaxBuilder/gyb_generated/Buildables.swift#L1720) should take a `TokenList`, not a `TokenListSyntax`. Looking into this a bit I *believe* this is because `TokenListSyntax` is classified as a token in [Child.py:39](https://github.com/apple/swift/blob/main/utils/gyb_syntax_support/Child.py#L39), which is *probably* a bug and should be checking for a `Token` suffix instead of a `Token` substring. Could you maybe look into it?

Sources/SwiftSyntaxBuilder/gyb_generated/DeclBuildables.swift
```diff
@@ -0,0 +1,1082 @@
...
+extension StringLiteralExpr {
+  public init(
+    openDelimiter: String?,
+    openQuote: TokenSyntax,
+    @StringLiteralSegmentsBuilder segmentsBuilder: () -> StringLiteralSegments = { .empty },
+    closeQuote: TokenSyntax,
+    closeDelimiter: String?
+  ) {
+    self.init(
+      openDelimiter: openDelimiter == nil ? nil : Tokens.rawStringDelimiter(openDelimiter!),
```
kimdv, line 198: I have moved default value down here? 
What do you think?

Sources/SwiftSyntaxBuilder/gyb_generated/DeclBuildables.swift
```diff
@@ -0,0 +1,1082 @@
...
+    )
+  }
+}
+
+extension FunctionParameter {
+  public init(
+    @AttributeListBuilder attributesBuilder: () -> AttributeList = { .empty },
+    firstName: TokenSyntax? = nil,
+    secondName: TokenSyntax? = nil,
+    colon: TokenSyntax? = Tokens.`colon`,
```
kimdv, line 444: What should we do with optional parameters?

Sources/SwiftSyntaxBuilder/gyb_generated/DeclBuildables.swift
```diff
@@ -0,0 +1,1082 @@
...
+    type: TypeBuildable? = nil,
+    ellipsis: TokenSyntax? = Tokens.`ellipsis`,
+    defaultArgument: InitializerClause? = nil,
+    trailingComma: TokenSyntax? = Tokens.`comma`
```
kimdv, line 448: What should we do with optional parameters? 
As I see, this could be a problem. Because here we are adding a comma. But the last parameter should not have one. 
I don't even think the generated code would compile?

Sources/SwiftSyntaxBuilder/gyb_generated/DeclBuildables.swift
```diff
@@ -0,0 +1,1082 @@
...
+
+extension InitializerDecl {
+  public init(
+    @AttributeListBuilder attributesBuilder: () -> AttributeList = { .empty },
+    @ModifierListBuilder modifiersBuilder: () -> ModifierList = { .empty },
+    initKeyword: TokenSyntax = Tokens.`init`,
+    optionalMark: TokenSyntax? = nil,
+    genericParameterClause: GenericParameterClause? = nil,
+    parameters: ParameterClause,
+    throwsOrRethrowsKeyword: TokenSyntax? = nil,
```
kimdv, line 495: Keywords with an `or` in do we ignore

Sources/SwiftSyntaxBuilder/gyb_generated/DeclBuildables.swift
```diff
@@ -0,0 +1,1082 @@
...
+  }
+}
+
+extension AccessorDecl {
+  public init(
+    @AttributeListBuilder attributesBuilder: () -> AttributeList = { .empty },
+    modifier: DeclModifier? = nil,
+    accessorKind: TokenSyntax,
+    parameter: AccessorParameter? = nil,
+    asyncKeyword: String?,
```
kimdv, line 577: I have seen multiple placeses where the `async` keyword is a string.
Is this meant to be this way?

Sources/SwiftSyntaxBuilder/gyb_generated/DeclBuildables.swift
```diff
@@ -0,0 +1,1082 @@
...
+      rightBrace: rightBrace
+    )
+  }
+}
+
+extension VariableDecl {
+  public init(
+    @AttributeListBuilder attributesBuilder: () -> AttributeList = { .empty },
+    @ModifierListBuilder modifiersBuilder: () -> ModifierList = { .empty },
+    letOrVarKeyword: TokenSyntax,
```
kimdv, line 611: Again, we can't make a default token here.

Sources/SwiftSyntaxBuilder/gyb_generated/DeclBuildables.swift
```diff
@@ -0,0 +1,1082 @@
...
+  }
+}
+
+extension ExpressionSegment {
+  public init(
+    backslash: TokenSyntax = Tokens.`backslash`,
+    delimiter: String?,
+    leftParen: TokenSyntax = Tokens.`leftParen`,
+    @TupleExprElementListBuilder expressionsBuilder: () -> TupleExprElementList = { .empty },
+    rightParen: TokenSyntax = Tokens.`stringInterpolationAnchor`
```
kimdv, line 177: There is something wrong here. 
I think it relates to this [ExpressionSegment](https://github.com/apple/swift/blob/main/utils/gyb_syntax_support/ExprNodes.py#L504)

On line 510 left paren is `  Child('LeftParen', kind='LeftParenToken', classification='StringInterpolationAnchor', force_classification=True),`

But the right paren `Child('RightParen', kind='StringInterpolationAnchorToken')`

I think it is a bug?

Sources/SwiftSyntaxBuilder/DeclBuildables.swift.gyb
```diff
@@ -0,0 +1,89 @@
+%{
```
kimdv, line 1: And then the naming of this file? 
Should it be something like `BuildableExtensions.swift.gyb` ?

Tests/SwiftSyntaxBuilderTest/StructTests.swift
```diff
@@ -55,4 +49,56 @@ final class StructTests: XCTestCase {
...
+  func testEmptyStructWithBuilder() {
+    let leadingTrivia = Trivia.garbageText("␣")
+    let test = StructDecl(identifier: "TestStruct", members: MemberDeclBlock())
+
+    let syntax = test.buildSyntax(format: Format(), leadingTrivia: leadingTrivia)
+
+    var text = ""
+    syntax.write(to: &text)
+
+    XCTAssertEqual(text, """
```
kimdv, line 62: This test fails and generates
```
␣struct TestStruct{
}
```

We need a leading space before the `{`.
We could add `requires_leading_space `on [LeftBrace](https://github.com/apple/swift/blob/main/utils/gyb_syntax_support/Token.py#L217).

kimdv:
@ahoppen I have pushed some more, and we are having something that looks good I think. 

I have left some comments in the code with some thought and findings 😁

Sources/SwiftSyntaxBuilder/DeclBuildables.swift.gyb
```diff
@@ -0,0 +1,90 @@
...
+extension ${node.syntax_kind} {
+  public init(
+%       init_parameters = []
+%       for child in node.children:
+%         child_node = NODE_MAP.get(child.syntax_kind)
+%         token = SYNTAX_TOKEN_MAP.get(child.syntax_kind)
+%         if child_node and child_node.is_syntax_collection():
+%             default_value = "? = { nil }" if child.is_optional else " = { .empty }"
+%             init_parameters.append("@%sBuilder %sBuilder: () -> %s%s" % (child.syntax_kind, child.swift_name, child.syntax_kind, default_value))
+%         elif child.syntax_kind is "IdentifierToken" or token and not token.text:
```
ahoppen, line 45: Nitpick: I think parenthesis around `token and not token.text` would improve readability.

------

Also, I think it would be good to have some comments on the different branches to describe what kind of transformation you are doing. Like for the first `if`: Allow initializing syntax collections with result builders, etc.

You should be able to put comments in `gyb` by prefixing the line with `% #` (`%` tells `gyb` that it’s Python line `#` makes it a Python comment).

The same (or shorter abbreviated) comments would probably also help below when you’re creating the call to `self.init`.

Sources/SwiftSyntaxBuilder/DeclBuildables.swift.gyb
```diff
@@ -0,0 +1,90 @@
...
+    self.init(
+%     init_parameters = []
+%     for child in node.children:
+%       child_node = NODE_MAP.get(child.syntax_kind)
+%       token = SYNTAX_TOKEN_MAP.get(child.syntax_kind)
+%       if child_node and child_node.is_syntax_collection():
+%         init_parameters.append("%s: %sBuilder()" % (child.swift_name, child.swift_name))
+%       elif child.syntax_kind is "IdentifierToken":
+%         if child.is_optional:
+%           init_parameters.append("%s: %s == nil ? nil : SyntaxFactory.makeIdentifier(%s!)" % (child.swift_name, child.swift_name, child.swift_name))
```
ahoppen, line 69: Just a suggestion if you like it better:
```diff
- %s == nil ? nil : SyntaxFactory.makeIdentifier(%s!)
+ %s.map(SyntaxFactory.makeIdentifier)
```

Sources/SwiftSyntaxBuilder/DeclBuildables.swift.gyb
```diff
@@ -0,0 +1,90 @@
...
+%         else:
+%           init_parameters.append("%s: SyntaxFactory.makeIdentifier(%s)" % (child.swift_name, child.swift_name))
+%         end
+%       elif token and not token.text:
+%         if child.is_optional:
+%           init_parameters.append("%s: %s == nil ? nil : Tokens.%s(%s!)" % (child.swift_name, child.swift_name, lowercase_first_word(token.name), child.swift_name))
```
ahoppen, line 75: Same here. If you like it better, you could replace the ternary by `map`

```diff
- %s == nil ? nil : Tokens.%s(%s!)
+ %s.map(Tokens.%s)
```

Sources/SwiftSyntaxBuilder/DeclBuildables.swift.gyb
```diff
@@ -0,0 +1,90 @@
...
+%         else:
+%         init_parameters.append("%s: Tokens.%s(%s)" % (child.swift_name, token.name, child.swift_name))
```
ahoppen, line 77: Missing indentation?

Sources/SwiftSyntaxBuilder/DeclBuildables.swift.gyb
```diff
@@ -0,0 +1,90 @@
+%{
+  from gyb_syntax_support import *
+  from gyb_syntax_support.kinds import lowercase_first_word
+  from gyb_syntax_support.kinds import SYNTAX_BASE_KINDS
+  from gyb_syntax_support.kinds import syntax_buildable_child_type
```
ahoppen, line 5: I think you can unify those two imports

```suggestion
  from gyb_syntax_support.kinds import SYNTAX_BASE_KINDS, syntax_buildable_child_type
```

Sources/SwiftSyntaxBuilder/ResultBuilders.swift.gyb
```diff
@@ -0,0 +1,53 @@
+%{
+  from gyb_syntax_support import *
+  from gyb_syntax_support.kinds import lowercase_first_word
+  from gyb_syntax_support.kinds import SYNTAX_BASE_KINDS
+  NODE_MAP = create_node_map()
+  # -*- mode: Swift -*-
+  # Ignore the following admonition it applies to the resulting .swift file only
+
+  def syntax_buildable_child_type(type_name, syntax_kind, is_token, is_optional=False):
```
ahoppen, line 9: Can you import `syntax_buildable_child_type` like you do in `Buildables.swift.gyb`?

Tests/SwiftSyntaxBuilderTest/StructTests.swift
ahoppen: Requiring a leading space for left brace sounds reasonable to me, but I’m not too upset about the current behavior either – it compiles, everything else just just best effort and we’ll never match a proper formatter in that.

Could you do that change in a follow-up PR and just leave it as is for now?

Tests/SwiftSyntaxBuilderTest/VariableTests.swift
```diff
@@ -0,0 +1,42 @@
...
+import SwiftSyntax
+
+import SwiftSyntaxBuilder
+
+final class VariableTests: XCTestCase {
+  func testVariableDecl() {
+    let leadingTrivia = Trivia.garbageText("␣")
+
+
+
```
ahoppen, line 11: Are these three empty lines intentional or just a leftover?

ahoppen (commented):
Wohoo. It’s starting to take real shape now with the result builders. Once this is merged I need to really start playing around with it myself.

I’ve added some comments inline. 

Could you also put the convenience initializers in `ExprBuildables.swift` and `SyntaxBuildables.swift` in a follow-up PR? Then this one would be all about result builders (and some other minor improvements in the generated convenience initializers).

Sources/SwiftSyntaxBuilder/DeclBuildables.swift.gyb
kimdv: When I use `asyncKeyword.map(SyntaxFactory.makeIdentifier)` I get the error `Cannot convert value of type '(String, Trivia, Trivia) -> TokenSyntax' to expected argument type '(String) throws -> TokenSyntax'`

So I have used `asyncKeyword.map({ SyntaxFactory.makeIdentifier($0) })`
Little more clean and easier to read.

Sources/SwiftSyntaxBuilder/DeclBuildables.swift.gyb
kimdv: I have adde comments now. 
Please resolve this if you thinks it's fine

Sources/SwiftSyntaxBuilder/DeclBuildables.swift.gyb
ahoppen: Ah, I forgot that `SyntaxFactory.makeIdentifier` has defaulted arguments and we thus can’t use it as a function of type `(String) -> TokenSyntax` (because it’s not, it’s of type `(String, Trivia, Trivia) -> TokenSyntax` like the error says. 

Your solution looks good to me 👍

Sources/SwiftSyntaxBuilder/gyb_generated/DeclBuildables.swift
ahoppen: Sorry, only saw these comments now. I was looking through your changed files and didn’t open the generated file. Sorry for that.

I don’t quite understand what you’re trying to say but the code looks good to me.

Sources/SwiftSyntaxBuilder/gyb_generated/DeclBuildables.swift
ahoppen: Hmm. Maybe it’d be better to not default the token in that case. That way the user needs to be explicit whether he wants it or not.

Sources/SwiftSyntaxBuilder/gyb_generated/DeclBuildables.swift
kimdv: It's okay! 

Not sure what I have thought. Forgot it. Sorry 😆
Will close

Sources/SwiftSyntaxBuilder/gyb_generated/DeclBuildables.swift
ahoppen: I’m not quite sure what you’re trying to say, but I don’t think there’s much we can do here. The user needs to choose between three options (no keyword, `throws` or `rethrows`).

Sources/SwiftSyntaxBuilder/gyb_generated/DeclBuildables.swift
ahoppen: Yes, that’s correct. `async` is a contextual keyword (which means that you can use it in identifiers etc.). I think that’s mostly for historical reasons. If I remember correctly, someone wanted to change it but it’s not a top priority.

Sources/SwiftSyntaxBuilder/gyb_generated/DeclBuildables.swift
ahoppen: That’s fine with me.

ahoppen (approved):
Looking good to me. Again, thanks so much for working on this!

ahoppen:
@swift-ci Please test

kimdv:
@ahoppen you forgot the PR in Swift repo: apple/swift#37426 


> Looking good to me. Again, thanks so much for working on this!

No problem! I like to work on it, and I hope to see some developers use it 🎉

ahoppen:
https://github.com/apple/swift/pull/37426

@swift-ci Please test

PR #286, opened 2021-05-14
[build] invoke lit with python3 unconditionally

edymtt:
The mechanism to revert to the old mechanism added in #272 was only
introduced for an abudance of caution.


edymtt:
@swift-ci please test

ahoppen (approved):
LGTM if my assumption that we never used `SWIFT_SYNTAX_LIT_TESTS_USE_PYTHON_DEFAULT` is correct.

edymtt:
I personally did not resort to use that, and cannot find it used in Swift codebase as well.

PR #288, opened 2021-05-24
Add convenience initializers for simple expr values

kimdv:
I have tried to make some initializers for the types I thought would be helpfull. 

I could also try to generate all Expr that have a Keyword in the init parameter like [NilLiteralExpr](https://github.com/apple/swift/blob/main/utils/gyb_syntax_support/ExprNodes.py#L99)

Something like

```swift
extension NilLiteralExpr {
    public init(
        nilKeyword: TokenSyntax = Tokens.nil
    ) {
        self.init(nilKeyword: nilKeyword)
    }
}
```

### DIFF
--- a/Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
+++ b/Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
@@ -2,7 +2,7 @@
   from gyb_syntax_support import *
   from gyb_syntax_support.kinds import lowercase_first_word
   from gyb_syntax_support.kinds import SYNTAX_BASE_KINDS
-  from gyb_syntax_support.kinds import syntax_buildable_child_type
+  from gyb_syntax_support.kinds import syntax_buildable_child_type, syntax_buildable_default_init_value
   # -*- mode: Swift -*-
   # Ignore the following admonition it applies to the resulting .swift file only
 }%
@@ -103,8 +103,9 @@ public struct ${node.syntax_kind}: ${node.base_kind}Buildable {
   public init(
 %     init_parameters = []
 %     for child in node.children:
+%       child_token = SYNTAX_TOKEN_MAP.get(child.syntax_kind)
 %       param_type = syntax_buildable_child_type(child.type_name, child.syntax_kind, child.is_token(), child.is_optional)
-%       default_value = " = nil" if child.is_optional else ""
+%       default_value = syntax_buildable_default_init_value(child, child_token)
 %       init_parameters.append("%s: %s%s" % (child.swift_name, param_type, default_value))
 %     end
     ${',\n    '.join(init_parameters)}
--- a/Sources/SwiftSyntaxBuilder/BuildablesConvenienceInitializers.swift.gyb
+++ b/Sources/SwiftSyntaxBuilder/BuildablesConvenienceInitializers.swift.gyb
@@ -1,11 +1,11 @@
 %{
   from gyb_syntax_support import *
-  from gyb_syntax_support.kinds import SYNTAX_BASE_KINDS, lowercase_first_word, syntax_buildable_child_type
+  from gyb_syntax_support.kinds import SYNTAX_BASE_KINDS, lowercase_first_word, syntax_buildable_child_type, syntax_buildable_default_init_value
   NODE_MAP = create_node_map()
   # -*- mode: Swift -*-
   # Ignore the following admonition it applies to the resulting .swift file only
 }%
-//// Automatically Generated From DeclBuildables.swift.gyb.
+//// Automatically Generated From BuildablesConvenienceInitializers.swift.gyb.
 //// Do Not Edit Directly!
 //===----------------------------------------------------------------------===//
 //
@@ -23,36 +23,35 @@ import SwiftSyntax
 
 % for node in SYNTAX_NODES:
 %   if node.is_buildable():
-%     has_syntax_collection_child = False
+%     should_create_convenience_initializer = False
 %     for child in node.children:
 %       child_node = NODE_MAP.get(child.syntax_kind)
-%       if child_node and child_node.is_syntax_collection():
-%         has_syntax_collection_child = True
+%       child_token = SYNTAX_TOKEN_MAP.get(child.syntax_kind)
+%       if (child_node and child_node.is_syntax_collection()) or (child_token and not child_token.text):
+%         # If the child token doesn’t have a text, we can create a convenience initializer that takes a string.
+%         # If the child token doesn’t have a text, it takes a String as parameter, we can create a convenient initializer.
+%         should_create_convenience_initializer = True
 %       end
 %     end
-%     if has_syntax_collection_child:
+%     if should_create_convenience_initializer:
 extension ${node.syntax_kind} {
   public init(
 %       init_parameters = []
 %       for child in node.children:
 %         child_node = NODE_MAP.get(child.syntax_kind)
-%         token = SYNTAX_TOKEN_MAP.get(child.syntax_kind)
+%         child_token = SYNTAX_TOKEN_MAP.get(child.syntax_kind)
 %         if child_node and child_node.is_syntax_collection():
 %           # Allow initializing syntax collections with result builders
 %           default_value = "? = { nil }" if child.is_optional else " = { .empty }"
 %           init_parameters.append("@%sBuilder %sBuilder: () -> %s%s" % (child.syntax_kind, child.swift_name, child.syntax_kind, default_value))
-%         elif child.syntax_kind is "IdentifierToken" or (token and not token.text):
+%         elif child_token and not child_token.text:
 %           # Allow initializing identifier or a token without a text with String value
 %           param_type = "String?" if child.is_optional else "String"
 %           init_parameters.append("%s: %s" % (child.swift_name, param_type))
 %         else:
 %           # When type is not handled above, use default value
 %           param_type = syntax_buildable_child_type(child.type_name, child.syntax_kind, child.is_token(), child.is_optional)
-%           default_value = ""
-%           if token and token.text and not child.is_optional:
-%             default_value = " = Tokens.`%s`" % lowercase_first_word(token.name)
-%           elif child.is_optional:
-%             default_value = " = nil"
+%           default_value = syntax_buildable_default_init_value(child, child_token)
 %           init_parameters.append("%s: %s%s" % (child.swift_name, param_type, default_value))
 %         end
 %       end
@@ -75,7 +74,7 @@ extension ${node.syntax_kind} {
 %         if child.is_optional:
 %           init_parameters.append("%s: %s.map(Tokens.%s)" % (child.swift_name, child.swift_name, lowercase_first_word(token.name)))
 %         else:
-%           init_parameters.append("%s: Tokens.%s(%s)" % (child.swift_name, token.name, child.swift_name))
+%           init_parameters.append("%s: Tokens.%s(%s)" % (child.swift_name, lowercase_first_word(token.name), child.swift_name))
 %         end
 %       else:
 %         init_parameters.append("%s: %s" % (child.swift_name, child.swift_name))
--- a/Tests/SwiftSyntaxBuilderTest/StructTests.swift
+++ b/Tests/SwiftSyntaxBuilderTest/StructTests.swift
@@ -6,55 +6,10 @@ import SwiftSyntaxBuilder
 final class StructTests: XCTestCase {
   func testEmptyStruct() {
     let leadingTrivia = Trivia.garbageText("␣")
-    let members = MemberDeclBlock(leftBrace: Tokens.leftBrace.withLeadingTrivia(.spaces(1)),
-                                  members: MemberDeclList([]),
-                                  rightBrace: Tokens.rightBrace)
-    let buildable = StructDecl(structKeyword: Tokens.struct,
-                               identifier: SyntaxFactory.makeIdentifier("TestStruct"),
+    let members = MemberDeclBlock(members: MemberDeclList([]))
+    let buildable = StructDecl(identifier: "TestStruct",
                                members: members)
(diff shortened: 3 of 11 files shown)


Sources/SwiftSyntaxBuilder/IntegerLiteralExprConvenienceInitializers.swift
```diff
@@ -0,0 +1,25 @@
...
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
+extension IntegerLiteralExpr {
+  public init(_ value: Int) {
```
ahoppen, line 16: One thing, I think might be confusing is that 

`0x10` converts to the integer literal `16`. Or `1_000` represents the integer literal `1000` (without the underscore). Similarly for the float literals. I think the value of the concise syntax outweighs the cost. @akyrtzi Do you have a strong opinion on this?

kimdv: @ahoppen I'm not sure that I understand the "problem" and confusing part here?
Can you try to explain what you mean with `I think the value of the concise syntax outweighs the cost.`

ahoppen: I just think that the following behavior might be confusing
```swift
func printIntegerLiteral(_ literal: IntegerLiteralExpr) {
  var text = ""
  literal.write(to: &text)
  print(text)
}

printIntegerLiteral(0x10) // prints 16, not 0x10
printIntegerLiteral(1_000) // prints 1000, not 1_000
```

But then, I think, in >95% of all cases you wouldn’t write code with hex literals or underscores and the convenience initializer does exactly what you would expect.

kimdv: I see and understand now. 

I'm currently looking into generate convenience initializers for all Expr.
So there will also be a

```swift
extension IntegerLiteralExpr {
  public init(
    digits: String
  ) {
    self.init(
      digits: Tokens.integerLiteral(digits)
    )
  }
}
```

Sources/SwiftSyntaxBuilder/ExprConvenienceInitializers.swift.gyb
```diff
@@ -0,0 +1,91 @@
...
+%     for child in node.children:
+%       child_node = NODE_MAP.get(child.syntax_kind)
+%       if child_node and child_node.is_syntax_collection():
+%         has_syntax_collection_child = True
+%       end
+%       if SYNTAX_TOKEN_MAP.get(child.syntax_kind) or child.syntax_kind is "IdentifierToken":
+%         has_token_as_child = True
+%       end
+%     end
+%     if not has_syntax_collection_child and has_token_as_child:
```
ahoppen, line 37: I haven’t looked at this in detail but wouldn’t we get the same (or something similar) if we dropped the `has_syntax_collection_child` requirement from [`BuildablesConvenienceInitializers.swift.gyb`](https://github.com/apple/swift-syntax/blob/18880228fa4de8c3368d66ead6f4e0dfd4c7f1f9/Sources/SwiftSyntaxBuilder/BuildablesConvenienceInitializers.swift.gyb#L33)?

Sources/SwiftSyntaxBuilder/IntegerLiteralExprConvenienceInitializers.swift
ahoppen: A string initializer also sounds good 👍 Since @akyrtzi is also on board, I think we should also continue to offer the literal initializers. Feel free to do the changes in one PR or split them up, whatever you like best.

kimdv: I will split the PRs up!

Sources/SwiftSyntaxBuilder/ExprConvenienceInitializers.swift.gyb
kimdv: I have removed the file and made `BuildablesConvenienceInitializers.swift.gyb` generate more.

ahoppen:
It’s not directly related to this PR, but I felt a bit uneasy about the `BuildablesConvenienceInitializers` and I understand why now: There are currently three areas through which we optimize the ergonomics of the initializers:
1. Providing default values of `nil` and `Tokens.something` for optional children and token children without text
2. Taking a `String` parameter for tokens with text and constructing a `TokenSyntax` from the string
3. Supporting result builders for children that are syntax collections

I think 1. should already be implemented in `Buildables.swift.gyb`. Currently, we only default optional children to `nil` there. Also defaulting tokens would be nice, I think. Ideally, we could factor out the code that computes the default value and reuse it in `BuildablesConvenienceInitializers.swift.gyb`.

Regarding 2. and 3., I think it’s reasonable that we only provide an initializer that applies all optimizations and ignore use cases where a user might want to initialize a token by a string but not use the result builder. This is what we are doing already.

And just so recap, we need to be careful not to create a convenience initializer if neither 2. nor 3. apply because then we’re redeclaring the initializer defined in `Buildables.swift.gyb`. This, again, is what you’re already doing.

Long story short, I think the following changes would make the implementation more streamlined and easier to understand:
- Factor out the code that determines default values for parameters and use them in `Buildables.swift.gyb` as well as `BuildablesConvenienceInitializers.swift.gyb`. Default values should be `nil` and `Token.something` where applicable.
- In `BuildablesConvenienceInitializers.swift.gyb`, when doing the pre-scan whether a convenience initializer should be generated ([link](https://github.com/apple/swift-syntax/blob/18880228fa4de8c3368d66ead6f4e0dfd4c7f1f9/Sources/SwiftSyntaxBuilder/BuildablesConvenienceInitializers.swift.gyb#L25-L32)) just keep track of a single boolean variable `should_create_convenience_initializer`. Set it to `True` if either of the conditions in 2. or 3. are met
- Always expose all parameters in the convenience initializer. If I understand correctly, with this PR you are sometimes omitting defaulted parameters in the convenience initializer to avoid redeclaring the one in `Buildables.swift.gyb`. If we discovered that the conditions of 2. or 3. are satisfied, we know that we aren’t redeclaring the initializer in `Buildables.swift.gyb` so it’s safe to include them.

Does this make sense to you @kimdv?

kimdv:
I thinks it makes sense yes! 
I think I will start with the first thing. Adding `nil` or `Tokens.something` in the `Buildables.swift.gyb`. 😁

kimdv:
@ahoppen I think I made what you wrote.

I can split change in `Buildables.swift.gyb` in another PR if you want? 

The helper method is put in this PR: https://github.com/apple/swift/pull/37382

Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
```diff
@@ -103,8 +103,9 @@ public struct ${node.syntax_kind}: ${node.base_kind}Buildable {
   public init(
 %     init_parameters = []
 %     for child in node.children:
+%       child_token = SYNTAX_TOKEN_MAP.get(child.syntax_kind)
 %       param_type = syntax_buildable_child_type(child.type_name, child.syntax_kind, child.is_token(), child.is_optional)
-%       default_value = " = nil" if child.is_optional else ""
+%       default_value = syntax_buildable_default_init_value(child, child_token) # " = nil" if child.is_optional else ""
```
ahoppen, line 108: Is the `# " = nil" if child.is_optional else ""` still a comment left over from development?

Sources/SwiftSyntaxBuilder/BuildablesConvenienceInitializers.swift.gyb
```diff
@@ -23,36 +23,36 @@ import SwiftSyntax
...
-%     has_syntax_collection_child = False
+%     should_create_convenience_initializer = False
 %     for child in node.children:
 %       child_node = NODE_MAP.get(child.syntax_kind)
-%       if child_node and child_node.is_syntax_collection():
-%         has_syntax_collection_child = True
+%       child_token = SYNTAX_TOKEN_MAP.get(child.syntax_kind)
+%       if (child_node and child_node.is_syntax_collection()) or (child_token and not child_token.text) or (child.syntax_kind is "IdentifierToken"):
+%         # If node have syntax collection as child, we can create a convenient initializer
+%         # If the child token have dont have a text, it takes a String as parameter, we can create a convenient initializer
```
ahoppen, line 32:
```suggestion
%         # If the child token doesn’t have a text, it takes a String as parameter, we can create a convenient initializer
```

Sources/SwiftSyntaxBuilder/BuildablesConvenienceInitializers.swift.gyb
```diff
@@ -23,36 +23,36 @@ import SwiftSyntax
...
 %   if node.is_buildable():
-%     has_syntax_collection_child = False
+%     should_create_convenience_initializer = False
 %     for child in node.children:
 %       child_node = NODE_MAP.get(child.syntax_kind)
-%       if child_node and child_node.is_syntax_collection():
-%         has_syntax_collection_child = True
+%       child_token = SYNTAX_TOKEN_MAP.get(child.syntax_kind)
+%       if (child_node and child_node.is_syntax_collection()) or (child_token and not child_token.text) or (child.syntax_kind is "IdentifierToken"):
+%         # If node have syntax collection as child, we can create a convenient initializer
```
ahoppen, line 31: I’d add `we can create a convenience initializer using a result builder`

Sources/SwiftSyntaxBuilder/BuildablesConvenienceInitializers.swift.gyb
```diff
@@ -23,36 +23,36 @@ import SwiftSyntax
...
 % for node in SYNTAX_NODES:
 %   if node.is_buildable():
-%     has_syntax_collection_child = False
+%     should_create_convenience_initializer = False
 %     for child in node.children:
 %       child_node = NODE_MAP.get(child.syntax_kind)
-%       if child_node and child_node.is_syntax_collection():
-%         has_syntax_collection_child = True
+%       child_token = SYNTAX_TOKEN_MAP.get(child.syntax_kind)
+%       if (child_node and child_node.is_syntax_collection()) or (child_token and not child_token.text) or (child.syntax_kind is "IdentifierToken"):
```
ahoppen, line 30: Shouldn’t `child.syntax_kind is "IdentifierToken"` already be covered by `child_token and not child_token.text`. `IdentifierToken`s should always produce a `child_token` and don’t have  `text` set. Or am I missing something here? (same applies a bit further down)

ahoppen (commented):
This is looking very good overall. I’ve added a few remarks/questions inline.

kimdv:
@ahoppen I have resolved your comment and pushed. 
Also rebased https://github.com/apple/swift/pull/37712 with latest main

Sources/SwiftSyntaxBuilder/BuildablesConvenienceInitializers.swift.gyb
```diff
@@ -23,36 +23,36 @@ import SwiftSyntax
...
-%     has_syntax_collection_child = False
+%     should_create_convenience_initializer = False
 %     for child in node.children:
 %       child_node = NODE_MAP.get(child.syntax_kind)
-%       if child_node and child_node.is_syntax_collection():
-%         has_syntax_collection_child = True
+%       child_token = SYNTAX_TOKEN_MAP.get(child.syntax_kind)
+%       if (child_node and child_node.is_syntax_collection()) or (child_token and not child_token.text):
+%         # If node have syntax collection as child, we can create a convenience initializer using a result builder
+%         # If the child token doesn’t have a text, it takes a String as parameter, we can create a convenient initializer
```
ahoppen, line 32: I think this would be more understandable as 

```suggestion
%         # If the child token doesn’t have a text, we can create a convenience initializer that takes a string.
```

Sources/SwiftSyntaxBuilder/BuildablesConvenienceInitializers.swift.gyb
```diff
@@ -23,36 +23,36 @@ import SwiftSyntax
...
+%         # If child token is an IdentifierToken it can be made as a String, we can create a convenient initializer
```
ahoppen, line 33: I think this comment is a leftover from the `IdentifierToken` case.

Tests/SwiftSyntaxBuilderTest/StructTests.swift
```diff
@@ -6,76 +6,33 @@ import SwiftSyntaxBuilder
...
   func testNestedStruct() {
     let leadingTrivia = Trivia.garbageText("␣")
-    let emptyMembers = MemberDeclBlock(leftBrace: Tokens.leftBrace.withLeadingTrivia(.spaces(1)),
-                                  members: MemberDeclList([]),
-                                  rightBrace: Tokens.rightBrace)
-    let nestedStruct = StructDecl(structKeyword: Tokens.struct,
-                               identifier: SyntaxFactory.makeIdentifier("NestedStruct"),
+    let emptyMembers = MemberDeclBlock(members: MemberDeclList([]))
+    let nestedStruct = StructDecl(identifier: "NestedStruct",
                                members: emptyMembers)
```
ahoppen, line 27: Nitpick: `members` should be indented by 3 more spaces.

ahoppen (approved):
I’ve got a few suggestions on the wording for comments inline, otherwise looks good to me!

kimdv:
@ahoppen I have pushed the suggested changes. Also rebased apple/swift#37712 with latest main

Happy WWDC days! 🎉

ahoppen:
> Happy WWDC days! 🎉

Same to you!

https://github.com/apple/swift/pull/37712
@swift-ci Please test

PR #294, opened 2021-06-23
Move result builder as last parameters

kimdv:
Moved the result builders to end end of init parameters. This will give a more clean code.

Idea from #291


ahoppen (approved):
LGTM

ahoppen:
@swift-ci Please test

PR #293, opened 2021-06-22
Move Token helpers to TokenSyntax extension

kimdv:
Move `Token` helpers to `TokenSyntax` extension, so we can use dot notation

Idea from #291 

Swift PR https://github.com/apple/swift/pull/38026

PR #295, opened 2021-06-23
Make SyntaxToken extend ExpressibleByStringLiteral

kimdv:
@ahoppen I have open a PR, so we can discuss it here. 

From #291 

> 4a. TokenSyntax should be ExpressibleByStringLiteral. That way we wouldn’t need SyntaxFactory.makeIdentifier("myFirstVar"). It might also remove the need for convenience of passing in a String for a TokenSyntax in the convinience initializers because TokenSyntax can now just be expressed by a string literal.

I'm not sure how implement this.

Because if we add `ExpressibleByStringLiteral ` to `SyntaxToken`, how can it now to create a `identifier` token, just like when we use `SyntaxFactory.makeIdentifier("myOtherVar")`

### DIFF
--- a/Sources/SwiftSyntaxBuilder/TokenSyntaxConvenienceInitializers.swift
+++ b/Sources/SwiftSyntaxBuilder/TokenSyntaxConvenienceInitializers.swift
@@ -1,0 +1,19 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2021 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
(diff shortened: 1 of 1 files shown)


ahoppen:
Argh, I temporarily forgot that `TokenSyntax` nodes carry a kind. My initial idea was that `let myToken: TokenSyntax = "hello"` should create an identifier with the text `hello`. But since `TokenSyntax` carries a type, we need to decide whether `myToken ` should indeed be an identifier or maybe some other token like a contextual keyword. So I take back my idea.

I also just realized that since #293 we can write 
```swift
IdentifierPattern(identifier: .identifier("myOtherVar")),
```
instead of
```swift
IdentifierPattern(identifier: SyntaxFactory.makeIdentifier("myOtherVar")),
```
which I think is a very precise and short syntax for what we are trying to achieve.

kimdv:
Good. I was a bit confused. Maybe I didn't know something. 

But will close then.

PR #299, opened 2021-07-21
Copy file contents to a new buffer before passing it to `swiftparse_parse_string`

ahoppen:
It appears that under certain circumstances the Swift parser is modifying its source buffer (rdar://72848263).
To prevent it from modifying the contents of `source`, create a new memory buffer with the contents of `source` and pass that one to the Swift parser.

This is a tentative fix for rdar://72848263. If it doesn’t fix the issue, we can revert this change.


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
Figured out that this is not the problem. Closing.

PR #301, opened 2021-08-09
Clarify that edits passed to `IncrementalParseTransition` are concurrent and add function to transform sequential edits to concurrent

ahoppen:
To clarify that the edits passed to `IncrementalParseTransition` are applied concurrently, introduce a new `ConcurrentEdit` type that provides the guarantee and allows translation of sequentially applied edits to the expected concurrent form.

Fixes rdar://72848263

### DIFF
--- a/Sources/SwiftSyntax/IncrementalParseTransition.swift
+++ b/Sources/SwiftSyntax/IncrementalParseTransition.swift
@@ -45,7 +45,7 @@ public final class IncrementalParseReusedNodeCollector:
 /// occurred since it was created.
 public final class IncrementalParseTransition {
   fileprivate let previousTree: SourceFileSyntax
-  fileprivate let edits: [SourceEdit]
+  fileprivate let edits: ConcurrentEdits
   fileprivate let reusedDelegate: IncrementalParseReusedNodeDelegate?
 
   /// - Parameters:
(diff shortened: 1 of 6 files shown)


ahoppen:
@swift-ci Please test

akyrtzi:
`isEditArrayValid()` can be made internal now.
Also `normalizeEdits()` is only useful to be made public for testing purposes, should we have it underscored?

ahoppen:
OK, I refactored the PR pretty significantly (although the translation algorithm stayed mostly the same).

`IncrementalParseTransition` now takes `ConcurrentEdits` instead of `[SourceEdit]`. `ConcurrentEdits` can either be initialized from `[SourceEdit]` that are already guaranteed to be in concurrent form or from an array of sequentially applied edits, which are then translated to concurrent edits in the expected form. This way, the user of the API always needs to make a deliberate choice which constructor of `ConcurrentEdits` to invoke and thus shouldn’t make any false assumptions about how the edits are interpreted. 

The downside is that the change changes the API in a breaking way. We could consider whether we want to keep providing a deprecated initializer to `IncrementalParseTransition` that still takes `[SourceEdit]` instead of `ConcurrentEdits` and just converts it internally using `ConcurrentEdits.init(concurrent:)`.

ahoppen:
@swift-ci Please test

akyrtzi:
This API isn't used much, a search on GitHub only showed the stress-tester, but having a transitional step is goodness.

Sources/SwiftSyntax/IncrementalParseTransition.swift
```diff
@@ -57,19 +57,122 @@ public final class IncrementalParseTransition {
...
+///  - insert 'x' at offset 0
+///  - insert 'y' at offset 1
+///  - insert 'z' at offset 2
+///  to '012345' results in 'x0y1z2345'.
+///
+/// The raw `edits` of this struct are guaranteed to
+///   1. not be overlapping.
+///   2. be in increasing source offset order.
+public struct ConcurrentEdits {
+  /// The raw concurrent edits. Are guranteed to satisfy the requirements
```
akyrtzi, line 104: Typo "guranteed"

akyrtzi (approved):

ahoppen:
@swift-ci Please test

PR #304, opened 2021-08-25
[SwiftSyntaxBuilder] Add support for control flow statements in result builders

ahoppen:
Add the `buildOptional` etc. result builder constructs to add support for control-flow constructs inside the result builders.

CC: @czechboy0


ahoppen:
@swift-ci Please test

PR #305, opened 2021-08-27
Add result builder to decl blocks

kimdv:
Hi @ahoppen 👋

I have not forgotten you, just had a busy at work and then summer vacation!
And European Championship in football where Denmark did it very good ⚽️
But I'll get some time to look at somethings again. 

From #291
This PR aims to A `MemberDeclBlock` should be directly creatable from a `MemberDeclListBuilder` result builder. This would probably take some hardcoding of `MemberDeclBlock` in the convenience initializers where the initialization of a `MemberDeclBlock` with a `MemberDeclListBuilder` is another convenience added (besides the current ones of: syntax collection through result builder, tokens by string literals and defaulting of tokens)

This will make it possible to make things like

```
StructDecl("Foo") {
    StructDecl("Bar")
}
```

I will next week start too look into it again, and catch up.
I will properly ask some questions

### DIFF
--- a/Sources/SwiftSyntaxBuilder/StringLiteralExprConvenienceInitializers.swift
+++ b/Sources/SwiftSyntaxBuilder/StringLiteralExprConvenienceInitializers.swift
@@ -13,14 +13,14 @@
 import SwiftSyntax
 
 extension StringLiteralExpr {
-  public init(_ value: String) {
+  public init(_ value: String, openQuote: TokenSyntax = .stringQuote, closeQuote: TokenSyntax = .stringQuote) {
     let content = SyntaxFactory.makeToken(TokenKind.stringSegment(value), presence: .present)
     let segment = StringSegment(content: content)
     let segments = StringLiteralSegments([segment])
 
-    self.init(openQuote: .stringQuote,
+    self.init(openQuote: openQuote,
               segments: segments,
-              closeQuote: .stringQuote)
+              closeQuote: closeQuote)
   }
 }
 
--- a/Tests/SwiftSyntaxBuilderTest/ExpressibleBuildablesTests.swift
+++ b/Tests/SwiftSyntaxBuilderTest/ExpressibleBuildablesTests.swift
@@ -1,0 +1,86 @@
+import XCTest
+import SwiftSyntax
+
+import SwiftSyntaxBuilder
+
+final class ExpressibleBuildablesTests: XCTestCase {
+    func testExpressibleAsMemberDeclListItem() {
+        let myStruct = StructDecl(identifier: "MyStruct", members: MemberDeclBlock(membersBuilder: {
+            VariableDecl(.var, name: "myFirstVar", type: "Int")
+
+            // We use `MemberDeclListItem` to ensure and show we can combine it with `ExpressibleAsMemberDeclListItem`
+            MemberDeclListItem(decl: VariableDecl(letOrVarKeyword: .let, bindingsBuilder: {
+                    PatternBinding(pattern: IdentifierPattern(identifier: .identifier("myOtherLet")),
+                                   typeAnnotation: TypeAnnotation(type: SimpleTypeIdentifier("String")))
+                })
+            )
+
+            StructDecl(identifier: "InnerStruct", members: MemberDeclBlock())
+        })
+        )
+
+        let syntax = myStruct.buildSyntax(format: Format())
+        XCTAssertEqual(syntax.description, """
+        struct MyStruct{
+            var myFirstVar: Int
+            let myOtherLet: String
+            struct InnerStruct{
+            }
+        }
+        """)
+    }
+
+    func testExpressibleAsCodeBlockItem() {
+        let myCodeBlock = SourceFile(eofToken: SyntaxFactory.makeToken(.eof, presence: .present)) {
+            StructDecl(identifier: "MyStruct1", members: MemberDeclBlock())
+
+            StructDecl(identifier: "MyStruct2", members: MemberDeclBlock())
+        }
+
+        let syntax = myCodeBlock.buildSyntax(format: Format())
+        XCTAssertEqual(syntax.description, """
+
+        struct MyStruct1{
+        }
+        struct MyStruct2{
+        }
+        """)
+    }
+
+    func testExpressibleAsSwitchStmt() {
+        let versions = [("version_1", "1.0.0"), ("version_2", "2.0.0"), ("version_3", "3.0.0"), ("version_3_1", "3.1.0")]
+        let expression = IdentifierExpr(identifier: SyntaxFactory.makeIdentifier("version"))
+
+        let switchStmt = SwitchStmt(labelName: nil,
+                                    expression: expression,
+                                    leftBrace: .leftBrace.withTrailingTrivia(.newlines(1)),
+                                    rightBrace: .rightBrace.withLeadingTrivia(.newlines(1)),
+                                    casesBuilder: {
+            for (version, semVer) in versions {
+                SwitchCase(label: SwitchCaseLabel(caseItemsBuilder: {
+                    CaseItem(pattern: EnumCasePattern(caseName: version))
+                }), statementsBuilder: {
+                    ReturnStmt(expression: StringLiteralExpr(semVer, closeQuote: .stringQuote.withTrailingTrivia(.newlines(1))))
+                })
+            }
+        })
+
+        let syntax = switchStmt.buildSyntax(format: Format())
+
+        // The generated code contains whitespace after `:`.
+        // So replacing whitespace with `␣`.
+        XCTAssertEqual(syntax.description.replacingOccurrences(of: " ", with: "␣"), """
+        switch␣version{
+        case␣.version_1:␣
+        return␣"1.0.0"
+        case␣.version_2:␣
+        return␣"2.0.0"
+        case␣.version_3:␣
+        return␣"3.0.0"
+        case␣.version_3_1:␣
+        return␣"3.1.0"
+
+        }
+        """)
+    }
+}
--- a/Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
+++ b/Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
@@ -39,10 +39,19 @@ public protocol ${kind}ListBuildable: SyntaxListBuildable {
   func build${kind}List(format: Format, leadingTrivia: Trivia?) -> [${build_kind}]
 }
 
+%     buildable_type = kind + 'Buildable'
+public protocol ExpressibleAs${buildable_type} {
+  func create${buildable_type}() -> ${buildable_type}
+}
+
 %     if kind == 'Syntax':
-public protocol ${kind}Buildable: ${kind}ListBuildable {
+public protocol ${buildable_type}: ExpressibleAs${buildable_type}, ${kind}ListBuildable {
+%     elif kind == 'Decl':
+public protocol ${buildable_type}: ExpressibleAs${buildable_type}, SyntaxBuildable, ${kind}ListBuildable, ExpressibleAsMemberDeclListItem, ExpressibleAsCodeBlockItem {
+%     elif kind == 'Stmt':
+public protocol ${buildable_type}: ExpressibleAs${buildable_type}, SyntaxBuildable, ${kind}ListBuildable, ExpressibleAsCodeBlockItem {
 %     else:
-public protocol ${kind}Buildable: SyntaxBuildable, ${kind}ListBuildable {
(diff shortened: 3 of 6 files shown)


ahoppen:
Cool. Thanks! And there’s no need to apologize, you are working on this in your free time after all. 😉 

Let me know when the code is ready for review or if you have any questions. I just took a look at the code and it still appears to be in an incomplete state to me.

ahoppen:
@kimdv What’s the state of this PR? Is it ready for review yet or still work in progress?

kimdv:
@ahoppen I think it is in a state where it is ready for review. 

I have rebased with latest master

ahoppen (commented):
I just looked through the changes and thought it was a little over-complicated to only generate the `ExpressibleAs` protocols for types that are children of syntax collections – why not have them for all syntax nodes as they might also come in handy later? I also thought it was a bit of a shame that `ExprBuildable` didn’t conform to `ExpressibleAsExprBuildable`. 

I couldn’t resist the temptation to play around with your design myself and think it’s easiest to just implement the `ExpressibleAs` protocols in `Buildables.swift.gyb`. That way we can also conform protocols like `ExprBuildable` to their corresponding `ExpressibleAs` protocols. I pushed my draft idea to https://github.com/ahoppen/swift-syntax/commit/d6d4f12596451fda73b19cca58ecb55c26ee8122. What do you think?

kimdv:
> I just looked through the changes and thought it was a little over-complicated to only generate the `ExpressibleAs` protocols for types that are children of syntax collections – why not have them for all syntax nodes as they might also come in handy later?

 Sure we can generate for them all. Did that initially, but went away because I wouldn't introduce a lot of protocols that wasn't used. Will add them again.

> I also thought it was a bit of a shame that `ExprBuildable` didn’t conform to `ExpressibleAsExprBuildable`.

My thought was just as we have a `ExprBuildable` protocol it would be a bit duplicated to have a `ExpressibleAsExprBuildable`?
But maybe there is some scenarios I haven't thought of.


> I couldn’t resist the temptation to play around with your design myself and think it’s easiest to just implement the `ExpressibleAs` protocols in `Buildables.swift.gyb`. That way we can also conform protocols like `ExprBuildable` to their corresponding `ExpressibleAs` protocols. I pushed my draft idea to [ahoppen@d6d4f12](https://github.com/ahoppen/swift-syntax/commit/d6d4f12596451fda73b19cca58ecb55c26ee8122). What do you think?

So we should move everything from `ExpressibleBuildables.swift.gyb` to `Buildables.swift.gyb`.

ahoppen:
> Sure we can generate for them all. Did that initially, but went away because I wouldn't introduce a lot of protocols that wasn't used. Will add them again.

My thought process was that we are exposing these protocols publicly and it’s hard to explain why certain syntax kinds have `ExpressibleAs` protocols and some don’t in case someone wants to use them in their own codebase. Generating all just eliminates that problem.

>  My thought was just as we have a `ExprBuildable` protocol it would be a bit duplicated to have a `ExpressibleAsExprBuildable`?
But maybe there is some scenarios I haven't thought of.

That’s a good point. I hadn’t thought of that. But I think that 
- if we have an `ExressibleAsExprBuildable` protocol, `ExprBuildable` should definitely conform to it (how is `ExprBuildable` not expressible as `ExprBuildable` 🤯) or
- we should not have the `ExpressibleAsExprBuildable` protocol at all. In that case, `ResultBuilders.swift.gyb` would need to check whether `element_type` has an `ExpressibleAs` protocol or
- `typealias ExpressibleAsExprBuildable = ExprBuildable` and add a comment that this type alias only exists to simplify code generation.

I don’t have a strong preference for either of these options.

> So we should move everything from `ExpressibleBuildables.swift.gyb` to `Buildables.swift.gyb`.

Not necessarily. I only did so to conform `ExprBuildable` to `ExpressibleAsExprBuildable` (and reuse some of the loops). If we decide that we don’t need `ExressibleAsExprBuildable` as a protocol, everything can live inside `ExpressibleBuildables.swift.gyb` again.

kimdv:
@ahoppen I have applied your suggestions 👍

Tests/SwiftSyntaxBuilderTest/ExpressibleBuildablesTests.swift
```diff
@@ -0,0 +1,95 @@
...
+                    CaseItem(pattern: EnumCasePattern(caseName: version))
+                }), statementsBuilder: {
+                    ReturnStmt(expression: StringLiteralExpr(semVer, closeQuote: .stringQuote.withTrailingTrivia(.newlines(1))))
+                })
+            }
+        })
+
+        let syntax = switchStmt.buildSyntax(format: Format())
+
+        XCTAssertEqual(syntax.description, """
```
kimdv, line 75: I have added this test case so we also have something more than `struct` examples.

BUT is keeps failing, and I don't know why. 
The expected result and the actual result seems equal.
Maybe you can help me @ahoppen ?

Screenshot 2021-09-23 at 22 11 10
?

ahoppen:
> After some thoughts I think it would be strange that `ExprBuildable` would be the only one with no `ExpressibleAs` as protocol.

It wouldn’t have been the only one. `DeclBuildable` also wouldn’t have had an `ExpressibleAs` protocol 😉 Still I agree it makes sense to add them for consistency.

Sources/SwiftSyntaxBuilder/StringLiteralExprConvenienceInitializers.swift
```diff
@@ -13,14 +13,14 @@
 import SwiftSyntax
 
 extension StringLiteralExpr {
-  public init(_ value: String) {
+    public init(_ value: String, openQuote: TokenSyntax = .stringQuote, closeQuote: TokenSyntax = .stringQuote) {
```
ahoppen, line 16: Incorrect indentation?

Tests/SwiftSyntaxBuilderTest/ExpressibleBuildablesTests.swift
```diff
@@ -0,0 +1,95 @@
...
+    func testExpressibleAsMemberDeclListItem() {
+        let myStruct = StructDecl(identifier: "MyStruct", members: MemberDeclBlock(membersBuilder: {
+            VariableDecl(.var, name: "myFirstVar", type: "Int")
+
+            VariableDecl(letOrVarKeyword: .let, bindingsBuilder: {
+                PatternBinding(pattern: IdentifierPattern(identifier: .identifier("mySecondLet")),
+                               typeAnnotation: TypeAnnotation(type: SimpleTypeIdentifier("CGFloat")))
+            })
+
+            MemberDeclListItem(decl: VariableDecl(letOrVarKeyword: .let, bindingsBuilder: {
```
ahoppen, line 16: Do you want to explicitly test we can put a `MemberDeclListItem` here? If so, I think it would be good to add a comment explaining your intention.

Tests/SwiftSyntaxBuilderTest/ExpressibleBuildablesTests.swift
```diff
@@ -0,0 +1,95 @@
+import XCTest
+import SwiftSyntax
+
+import SwiftSyntaxBuilder
+
+final class ExpressibleBuildablesTests: XCTestCase {
+    func testExpressibleAsMemberDeclListItem() {
+        let myStruct = StructDecl(identifier: "MyStruct", members: MemberDeclBlock(membersBuilder: {
+            VariableDecl(.var, name: "myFirstVar", type: "Int")
```
ahoppen, line 9: This is really starting to look like you can generate real-world code with `SwiftSyntaxBuilder` 😍

Tests/SwiftSyntaxBuilderTest/ExpressibleBuildablesTests.swift
```diff
@@ -0,0 +1,95 @@
...
+
+            VariableDecl(letOrVarKeyword: .let, bindingsBuilder: {
```
ahoppen, line 11: Do you want to test that both the `VariableDecl` convenience initializer and the `bindingsBuilder` initializer work here? If so, could you add a comment explaining your intention?

Tests/SwiftSyntaxBuilderTest/ExpressibleBuildablesTests.swift
```diff
@@ -0,0 +1,95 @@
...
+            let mySecondLet: CGFloat
+            let myOtherVar: String
+            struct InnerStruct{
+            }
+        }
+        """)
+    }
+
+    func testExpressibleAsCodeBlockItem() {
+        let myCodeBlock = SourceFile(eofToken: SyntaxFactory.makeToken(.eof, presence: .present)) {
```
ahoppen, line 39: Not part of this PR but would it make sense to add 
```swift
extension TokenSyntax {
  static let eof = SyntaxFactory.makeToken(.eof, presence: .present)
}
```
So we could improve this to 
```suggestion
        let myCodeBlock = SourceFile(eofToken: .eof) {
```
And possibly try to remove the need to specify the `eofToken` in the `SourceFile` initializer altogether.

Tests/SwiftSyntaxBuilderTest/ExpressibleBuildablesTests.swift
ahoppen: The generated code contains trailing whiteface after the `:`. It’s a bit annoying that you can’t have trailing whitespace in the string literal you’re matching against 😬 Maybe trim all trailing whitespace from the generated string before comparing or replace all whitespace with `␣`?

Tests/SwiftSyntaxBuilderTest/ExpressibleBuildablesTests.swift
```diff
@@ -0,0 +1,95 @@
...
+        return "3.0.0"
+        case .version_3_1:
+        return "3.1.0"
+
+        }
+        """)
+    }
+}
+
+extension ReturnStmt: ExpressibleAsCodeBlockItem {
```
ahoppen, line 91: This should probably live in `SwiftSyntaxBuilder` and not the test suite, right?

ahoppen (commented):
Looking very good to me. And the test cases really start to look like what I’ve always envisioned `SwiftSyntaxBuilder` to look like 🤩 

I’ve added some minor comments regarding the test cases inline.

Tests/SwiftSyntaxBuilderTest/ExpressibleBuildablesTests.swift
kimdv: My intention is to make sure they can be combined and showcasing it.
I will add a comment

ahoppen: Thats’s good! Just wanted to make sure we’re on the same page what should be tested here.

Tests/SwiftSyntaxBuilderTest/ExpressibleBuildablesTests.swift
kimdv: Oh yes. Sorry! 
Was playing around. I will also make Statements implement the `ExpressibleAsCodeBlockItem`. 
Then people can add switch case, return, if and so on.

Tests/SwiftSyntaxBuilderTest/ExpressibleBuildablesTests.swift
kimdv: Removing it. cannot remember why.

Tests/SwiftSyntaxBuilderTest/ExpressibleBuildablesTests.swift
kimdv: Replaced with `␣` it makes it more clean to read the test case I think. 😬

Tests/SwiftSyntaxBuilderTest/ExpressibleBuildablesTests.swift
kimdv: @ahoppen added `ExpressibleAsCodeBlockItem` protocol to `Stmt`

Tests/SwiftSyntaxBuilderTest/ExpressibleBuildablesTests.swift
kimdv: Yes! It look really good and it gets more and more clean to read! 🎉

ahoppen (approved):
Nice. Looks good to me. 🚀

ahoppen:
@swift-ci Please test

kimdv:
Rebased on latest master

Tests/SwiftSyntaxBuilderTest/ExpressibleBuildablesTests.swift
kimdv: Added in #314

ahoppen:
@swift-ci Please test

PR #311, opened 2021-09-23
Make `ConcurrentEdits.init(concurrent:)` throw instead of crash if edits are not concurrent

ahoppen:
The user of the `ConcurrentEdits.init(concurrent:)` should be able to control what happens if the edits passed to it do not satisfy the requirements. We shouldn’t presume that crashing is the right way to handle the failure.


ahoppen:
@swift-ci Please test

akyrtzi (approved):

PR #313, opened 2021-09-23
Add an environment variable that enables assertions in release builds

ahoppen:
SwiftSyntax’s assertions catch a lot of internal inconsistencies but are disabled in Release builds. Add the `SWIFT_SYNTAX_FORCE_ENABLE_ASSERTIONS` environment variable which dynamically enables these checks in Release builds to verify that no SwiftSyntax invariants are violated.

I couldn’t detect a performance difference between the old `assert` and the new `swiftSyntaxAssert` in release builds without the `SWIFT_SYNTAX_FORCE_ENABLE_ASSERTIONS` environment variable set.


akyrtzi:
For performance checking did you try both creating and walking a large tree?

ahoppen:
I checked all performance tests we have in the repo and none of them changed. Also I wouldn’t have expected that a simple new `if` statement would have made a significant performance impact.
- `ParsingPerformanceTests.testParsingPerformance`
- `VisitorPerformanceTests.testEmptyVisitorPerformance`
- `VisitorPerformanceTests.testEmptyRewriterPerformance`
- `VisitorPerformanceTests.testEmptyAnyVistorPerformance`

ahoppen:
@swift-ci Please test

ICodeMuch:
Is this swift?
sorry i dont yet

ahoppen:
@swift-ci Please test

ahoppen:
Hmm. Looking at the numbers again, it looks like `SyntaxClassifierPerformanceTests.swift` regresses by 50% (from 0.020s to 0.030s) due to this change. No idea why I missed it when looking at the numbers last time.

Sources/SwiftSyntax/SwiftSyntaxAssert.swift
```diff
@@ -0,0 +1,35 @@
...
+import Foundation
+
+fileprivate let forceEnableAssertions: Bool = {
+	ProcessInfo.processInfo.environment["SWIFT_SYNTAX_FORCE_ENABLE_ASSERTIONS"] != nil
+}()
+
+/// A wrapper around `Swift.assert` that not only traps in `-Onone` builds but
+/// also if the `SWIFT_SYNTAX_FORCE_ENABLE_ASSERTIONS` environment variable
+/// is set.
+func swiftSyntaxAssert(
```
tbkka, line 22: You almost certainly want this to be inlined.  Otherwise, you're potentially paying a function call in addition to the check for whether assertions should be enabled.

Sources/SwiftSyntax/SwiftSyntaxAssert.swift
```diff
@@ -0,0 +1,35 @@
...
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+
+fileprivate let forceEnableAssertions: Bool = {
+	ProcessInfo.processInfo.environment["SWIFT_SYNTAX_FORCE_ENABLE_ASSERTIONS"] != nil
+}()
```
tbkka, line 17: You could make this ever-so-slightly cheaper by making this a simple public Bool variable and doing the environment check once at the top of your main function.  That would avoid the additional lazy initialization check.  You should benchmark and see if such a change actually makes a measurable difference, of course.

PR #317, opened 2021-09-30
Add fibonacci function example/test

kimdv:
I have added a `FunctionDecl` test. 
Made it a bit more advanced, so I made the Fibonacci function in swift. 

I think it will give us a good idea of what we could do, so it is easier to use and a bit more convenient.
Here is some ideas

`ReturnStmt` should take a buildable as parameter, so we don't need to wrap it in something like `SequenceExpr`.
  Example api could be 
```Swift
ReturnStmt(expressionBuilder: {
  IntegerLiteralExpr(digits: "n")
})
```

Same for `IfStmt` we need, somehow, to skip `ExprList`.
Example api 
``` Swift
IfStmt(labelName: nil, body: ifCodeBlock, conditionsBuilder: {
  ConditionElement(conditionBuilder: {
    IntegerLiteralExpr(digits: "n"),

    BinaryOperatorExpr(operatorToken: .unspacedBinaryOperator("<=")),

    IntegerLiteralExpr(1)
  }
})
```

Maybe also a convenient api so we could do   

``` Swift
IfStmt(labelName: nil, body: ifCodeBlock, conditionsBuilder: {
  IntegerLiteralExpr(digits: "n"),

  BinaryOperatorExpr(operatorToken: .unspacedBinaryOperator("<=")),

  IntegerLiteralExpr(1)
})
```

`FunctionCallExpr` should be able to take a `String` as identifier `FunctionCallExpr(calledExpression: IdentifierExpr(identifier: .identifier("fibonacci"))` -> `FunctionCallExpr(calledExpression: "fibonacci")`

This was just some of my findings! 
Let me hear what you think @ahoppen

### DIFF
--- a/Tests/SwiftSyntaxBuilderTest/FunctionTests.swift
+++ b/Tests/SwiftSyntaxBuilderTest/FunctionTests.swift
@@ -1,0 +1,70 @@
+import XCTest
+import SwiftSyntax
+import SwiftSyntaxBuilder
+
+final class FunctionTests: XCTestCase {
+  func testEmptyStruct() {
+    let leadingTrivia = Trivia.garbageText("␣")
+
+    let input = ParameterClause(parameterListBuilder: {
+      FunctionParameter(firstName: .wildcard, secondName: .identifier("n"), colon: .colon, type: SimpleTypeIdentifier("Int"), attributesBuilder: {})
+    })
+
+    let ifCodeBlock = CodeBlock(statementsBuilder: {
+      ReturnStmt(expression: SequenceExpr(elementsBuilder: {
+        IntegerLiteralExpr(digits: "n")
+      }))
+    })
+
+    let signature = FunctionSignature(input: input, output: ReturnClause(returnType: SimpleTypeIdentifier("Int")))
+    
+
+    let codeBlock = CodeBlock(statementsBuilder: {
+      IfStmt(labelName: nil, body: ifCodeBlock, conditionsBuilder: {
+        ConditionElement(condition: ExprList([
+          IntegerLiteralExpr(digits: "n"),
+
+          BinaryOperatorExpr(operatorToken: .unspacedBinaryOperator("<=")),
+
+          IntegerLiteralExpr(1)
+        ]))
+      })
+
+      ReturnStmt(expression: SequenceExpr(elementsBuilder: {
+        FunctionCallExpr(calledExpression: IdentifierExpr(identifier: .identifier("fibonacci")), leftParen: .leftParen, rightParen: .rightParen, argumentListBuilder: {
+          TupleExprElement(expression: SequenceExpr(elementsBuilder: {
+            IntegerLiteralExpr(digits: "n")
+
+            BinaryOperatorExpr(operatorToken: .unspacedBinaryOperator("-"))
+
+            IntegerLiteralExpr(1)
+          }))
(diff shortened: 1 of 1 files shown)


ahoppen (approved):
Very good idea to use this as an example of how we can further improve the syntax builder 👍

ahoppen:
@swift-ci Please test

kimdv:
How should we approach the ResultBuilders in the `IfStmt`.
There are many more cases, should it would be best to generate the code for all cases.

Could you give me some hints @ahoppen so I can make it 😁

ahoppen:
OK, some ideas from me. We are starting with
```swift
IfStmt(labelName: nil, body: ifCodeBlock, conditionsBuilder: {
  ConditionElement(condition: ExprList([
    IntegerLiteralExpr(digits: "n"),

    BinaryOperatorExpr(operatorToken: .unspacedBinaryOperator("<=")),

    IntegerLiteralExpr(1)
  ]))
})
```

### Step 1:
If we make `ExpressibleAsExprBuildable` conform to `ExpressibleAsConditionElement`, we could write instead
```swift
IfStmt(labelName: nil, body: ifCodeBlock, conditionsBuilder: {
  ExprList([
    IntegerLiteralExpr(digits: "n"),

    BinaryOperatorExpr(operatorToken: .unspacedBinaryOperator("<=")),

    IntegerLiteralExpr(1)
  ])
})
```

### Step 2:
Change the `IfStmt` initializer (and probably all other initializers in `Buildables.swift`) to take `ExpressibleAsTypes`, i.e. change it to:
```swift
public init(
    labelName: ExpressibleAsTokenSyntax? = nil,
    labelColon: ExpressibleAsTokenSyntax? = nil,
    ifKeyword: ExpressibleAsTokenSyntax = TokenSyntax.`if`,
    conditions: ExpressibleAsConditionElementList,
    body: ExpressibleAsCodeBlock,
    elseKeyword: ExpressibleAsTokenSyntax? = nil,
    elseBody: ExpressibleAsSyntaxBuildable? = nil
  )
```
This doesn’t improve the syntax yet.

### Step 3:
Make `ExpressibleAsConditionElement` conform to `ExpressibleAsConditionElementList`. That we we receive:
```swift
IfStmt(labelName: nil, body: ifCodeBlock, conditions: ExprList([
  IntegerLiteralExpr(digits: "n"),

  BinaryOperatorExpr(operatorToken: .unspacedBinaryOperator("<=")),

  IntegerLiteralExpr(1)
])
```

Which is, I think, as clean as it’ll get. 

---

The problematic thing is that we can’t declare these protocol conformances retroactively through extensions, so I think it might make sense to model these conformances at the Python level, e.g. by having a map like the following
```python
conformances = {
  'ConditionElement': ['ConditionElementList'],
  'ExprBuildable': ['ConditionElement']
}
```
and then emit the conformances when generating the protocols in `Buildabeles.swift.gyb`. You could then probably also describe the current explicit conformances (`Syntax` to `SyntaxList`, `Decl` to `MemberDeclListItem` etc.) using that model. Note that we would no longer conform `Decl` to `ExpressibleAsMemberDeclListItem` but instead `Decl` to `ExpressibleAsDecl` (which we have already today) and `ExpressibleAsDecl` to `ExpressibleAsMemberDeclListItem`.

Does this make sense to you?

PR #318, opened 2021-10-01
Build and install SwiftSyntaxParser from `build-script.py`

ahoppen:
The stress tester is using SwiftSyntaxParser and is expecting to find the dylib in the toolchain, but it’s failing to find it because we’re not installing `SwiftSyntaxParser.dylib` into the toolchain. Install it to fix the issue.

rdar://83727870

### DIFF
--- a/build-script.py
+++ b/build-script.py
@@ -325,8 +325,8 @@ def clear_gyb_files_from_previous_run(sources_dir, destination_dir, verbose):
 # Building SwiftSyntax
 
 
-def get_installed_dylib_name():
-    return "libSwiftSyntax.dylib"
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
Verified that the toolchain generated with this PR applied contains both `libSwiftSyntax.dylib` and `libSwiftSyntaxParser.dylib`.

akyrtzi (approved):

ahoppen:
Also verified that the stress tester integration test disabled by https://github.com/apple/swift-integration-tests/pull/89 is passing again in the newly built toolchain.

PR #319, opened 2021-10-01
Get the realpath of the build directory when building SwiftSyntax

ahoppen:
We were currently canonicalizing the build directory when testing SwiftSyntax but not when building it. This caused errors if the build directory was a symlink because module caches were created for the non-canonicalized path during the build and then referenced with the canonicalized path during testing.


ahoppen:
@swift-ci Please test

akyrtzi (approved):

PR #320, opened 2021-10-04
Use protocols conformance map to add protocols

kimdv:
Okay @ahoppen I gave it my first try here!


Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
```diff
@@ -97,17 +101,25 @@ extension ${buildable_type} {
...
 }
-%     if kind == 'Decl':
+%     if buildable_protocols and 'ExpressibleAsConditionElement' in buildable_protocols:
+
+extension ${buildable_type} {
+  public func createConditionElement() -> ConditionElement {
+    ConditionElement(condition: self)
+  }
+}
+%     end
```
kimdv, line 111: Could we somehow share this kind of blocks?
If so, how?

ktoso:
@swift-ci please smoke test

Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
```diff
@@ -125,7 +137,12 @@ extension ${kind}Buildable {
 /// ${line}
 %       end
 %     end
+%     protocols = PROTOCOL_CONFORMANCES_MAP.get(node.syntax_kind)
+%     if protocols:
+public struct ${node.syntax_kind}: ${node.base_kind}Buildable, ${', '.join(protocols)} {
```
ahoppen, line 142: Again, I think we don’t need this if we express all conformances on the `ExpressibleAs` level.

Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
```diff
@@ -3,6 +3,7 @@
   from gyb_syntax_support.kinds import lowercase_first_word
   from gyb_syntax_support.kinds import SYNTAX_BASE_KINDS
   from gyb_syntax_support.kinds import syntax_buildable_child_type, syntax_buildable_default_init_value
+  from gyb_syntax_support.protocolsMap import PROTOCOL_CONFORMANCES_MAP
```
ahoppen, line 6: Since you asked in the other PR regarding naming: If we only model conformances using the `ExpressibleAs` protocols, I would suggest: `SYNTAX_BUILDABLE_EXPRESSIBLE_AS_CONFORMANCES`.

ahoppen (commented):
I’ve put some comments inline again.

kimdv:
@ahoppen I have pushed changes that you suggested in https://github.com/apple/swift-syntax/pull/320#discussion_r721384814

Sources/SwiftSyntaxBuilder/BooleanLiteralExprConvenienceInitializers.swift
```diff
@@ -10,9 +10,11 @@
 //
 //===----------------------------------------------------------------------===//
 
+import SwiftSyntax
+
 extension BooleanLiteralExpr {
   public init(_ value: Bool) {
-    self.init(booleanLiteral: value ? .true : .false)
+    self.init(booleanLiteral: value ? TokenSyntax.true : TokenSyntax.false)
```
kimdv, line 17: This is now needed as we use `ExpressibleAsTokenSyntax` and not  `TokenSyntax`

Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
```diff
@@ -229,15 +251,37 @@ public struct ${node.syntax_kind}: SyntaxBuildable {
...
   func create${node.syntax_kind}() -> ${node.syntax_kind}
 }
 
-extension ${node.syntax_kind}: ExpressibleAs${node.syntax_kind} {
+extension ${node.syntax_kind}: ${expressible_as_type} {
   public func create${node.syntax_kind}() -> ${node.syntax_kind} {
     self
   }
 }
+%     if expressible_as_protocols and 'ExpressibleAsConditionElementList' in expressible_as_protocols:
```
ahoppen, line 269: It’s a shame we need to split those conforming extensions to two places… What do you think about adding all of them toward the end of `Buildable.swift.gyb` by iterating over all keys in `SYNTAX_BUILDABLE_EXPRESSIBLE_AS_CONFORMANCES`?

Sources/SwiftSyntaxBuilder/BooleanLiteralExprConvenienceInitializers.swift
ahoppen: What do you think about changing `Tokens.swift` to extend `ExpressibleAsTokenSyntax` instead of `TokenSyntax`. IIUC we could still use leading dot syntax in that case.

Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
kimdv: Makes it more clean now.

Sources/SwiftSyntaxBuilder/BooleanLiteralExprConvenienceInitializers.swift
kimdv: We get and `Contextual member reference to static property 'true' requires 'Self' constraint in the protocol extension`.

I'm not that familiar with static properties and protocols. 
I have tried to make it return `Self` instead of `ExpressibleAsTokenSyntax `. I also tried to return `TokenSyntax,  but no luck. `

ahoppen: I honestly don’t fully understand *why* it works, but 
```swift
extension ExpressibleAsTokenSyntax where Self == TokenSyntax {
  static var `associatedtype`: TokenSyntax { SyntaxFactory.makeAssociatedtypeKeyword()
    .withTrailingTrivia(.spaces(1)) }
}
```
appears to work.

kimdv: No problem! 

Should we in the readme or another file make a Known issues list? 
Just so we remember it at some points and inform users of the library.

ahoppen (approved):
LGTM

PR #321, opened 2021-10-05
Move the Diagnostic-related types back to the SwiftSyntax module

ahoppen:
Moving `Diagnostic` etc. to the `SwiftSyntaxParser` module required `swift-format` to import `SwiftSyntaxParser` in a bunch of places where it shouldn’t be needed. Move the types back to the `SwiftSyntax` module for now until we completely refactor the parser to just return `Diagnostics` and not require a `DiagnsoticEngine` (tracked by SR-15280).


ahoppen:
@swift-ci Please test

PR #322, opened 2021-10-06
Remove group info for SwiftSyntax

ahoppen:
Depends on #321.

---

AFAICT the group info was only used for swift modules built by `build-script.py`. But since we don’t ship the swift modules as part of the toolchain anymore and the intended way to build SwiftSyntax for clients is through the SwiftPM package, I think the maintainance cost is higher than the value it provides (which I think has always been marginal).


ahoppen:
@swift-ci Please test

PR #323, opened 2021-10-06
Always build SwiftSyntax as a static library

ahoppen:
Instead of including `SwiftSyntax.dylib` and `SwiftSyntaxParser.dylib` in the toolchain, make all tools that depend on SwiftSyntax link against it statically. This simplifies the way SwiftSyntax is built and slightly reduces the toolchain size, because we only include on copy of SwiftSyntax (statically linked into `sk-stress-test`) instead of two copies (`SwiftSyntax.dylib` and `SwiftSyntaxParser.dylib`).

rdar://83757810


ahoppen:
https://github.com/apple/swift/pull/39682

@swift-ci Please test

PR #324, opened 2021-10-06
Remove support for installing the swift modules in a toolchain

ahoppen:
We haven’t been installing the modules into the toolchain since late 2019, there’s no need to have an option for it.


ahoppen:
@swift-ci Please test

PR #326, opened 2021-10-06
Add test case that tests the performance of the SyntaxClassifier

ahoppen:
Just found this branch in my local repo. Can’t remember what I created it for but seems useful.


ahoppen:
@swift-ci Please test

likeremix715 (approved):

PR #331, opened 2021-11-01
Add `ExtensionDecl` test

kimdv:
Hi @ahoppen ! 👋

I have played a but around with SwiftSyntaxBuilder. Started with making a test case that should make something like the [Tokens.swift](https://github.com/apple/swift-syntax/blob/main/Sources/SwiftSyntaxBuilder/gyb_generated/Tokens.swift).
Just to have something realistic. 

Made a bunch of convenient API's and other small improvements. 
This PR is a bit messy (a lot of changes and far from done). But I improve the API, and try different Idas when I think something can be done smarter or more clean. 

How should I progress this? 
Do you want PR's with small changes, and I keep rebasing this branch on main as they get merged or just one big PR here?

Depends on:
#335 
#334

### DIFF
--- a/Sources/SwiftSyntaxBuilder/IdentifierExprConvenienceInitializers.swift
+++ b/Sources/SwiftSyntaxBuilder/IdentifierExprConvenienceInitializers.swift
@@ -15,9 +15,3 @@ extension IdentifierExpr {
     self.init(identifier: .identifier(identifier))
   }
 }
-
-extension IdentifierExpr: ExpressibleByStringLiteral {
-  public init(stringLiteral value: String) {
-    self.init(value)
-  }
-}
--- a/Sources/SwiftSyntaxBuilder/IdentifierPatternConvenienceInitializers.swift
+++ b/Sources/SwiftSyntaxBuilder/IdentifierPatternConvenienceInitializers.swift
@@ -15,9 +15,3 @@ extension IdentifierPattern {
     self.init(identifier: .identifier(identifier))
   }
 }
-
-extension IdentifierPattern: ExpressibleByStringLiteral {
-  public init(stringLiteral value: String) {
-    self.init(value)
-  }
-}
--- a/Sources/SwiftSyntaxBuilder/SimpleTypeIdentifierConvenienceInitializers.swift
+++ b/Sources/SwiftSyntaxBuilder/SimpleTypeIdentifierConvenienceInitializers.swift
@@ -14,13 +14,7 @@ import SwiftSyntax
 
 extension SimpleTypeIdentifier {
   public init(_ name: String) {
-    self.init(name: SyntaxFactory.makeIdentifier(name),
+    self.init(name: .identifier(name),
               genericArgumentClause: nil)
-  }
-}
-
-extension SimpleTypeIdentifier: ExpressibleByStringLiteral {
-  public init(stringLiteral value: String) {
-    self.init(value)
   }
 }
--- a/Sources/SwiftSyntaxBuilder/StringLiteralExprConvenienceInitializers.swift
+++ b/Sources/SwiftSyntaxBuilder/StringLiteralExprConvenienceInitializers.swift
@@ -14,18 +14,12 @@ import SwiftSyntax
 
 extension StringLiteralExpr {
   public init(_ value: String, openQuote: TokenSyntax = .stringQuote, closeQuote: TokenSyntax = .stringQuote) {
-    let content = SyntaxFactory.makeToken(TokenKind.stringSegment(value), presence: .present)
+    let content = SyntaxFactory.makeToken(.stringSegment(value), presence: .present)
     let segment = StringSegment(content: content)
     let segments = StringLiteralSegments([segment])
 
     self.init(openQuote: openQuote,
               segments: segments,
               closeQuote: closeQuote)
-  }
-}
-
-extension StringLiteralExpr: ExpressibleByStringLiteral {
-  public init(stringLiteral value: String) {
-    self.init(value)
   }
 }
--- a/Sources/SwiftSyntaxBuilder/TypeAnnotationConvenienceInitializers.swift
+++ b/Sources/SwiftSyntaxBuilder/TypeAnnotationConvenienceInitializers.swift
@@ -15,9 +15,3 @@ extension TypeAnnotation {
     self.init(type: SimpleTypeIdentifier(type))
   }
 }
-
-extension TypeAnnotation: ExpressibleByStringLiteral {
-  public init(stringLiteral value: String) {
-    self.init(value)
-  }
-}
--- a/Tests/SwiftSyntaxBuilderTest/ExtensionDeclTests.swift
+++ b/Tests/SwiftSyntaxBuilderTest/ExtensionDeclTests.swift
@@ -1,0 +1,46 @@
+import XCTest
+import SwiftSyntax
+import SwiftSyntaxBuilder
+
+final class ExtensionDeclTests: XCTestCase {
+  func testExtensionDecl() {
+    let keywords = ["associatedtype", "class"].map { keyword -> VariableDecl in
+      // We need to use `CodeBlock` here to ensure there is braces around.
+      let body = CodeBlock {
+        FunctionCallExpr("SyntaxFactory.make\(keyword)Keyword",
+                         leftParen: .leftParen,
+                         rightParen: .rightParen)
+      }
+
+      return VariableDecl(letOrVarKeyword: .var,
+                          modifiersBuilder: { TokenSyntax.public },
+                          bindingsBuilder: {
+        PatternBinding(pattern: "`\(keyword)`",
+                       typeAnnotation: "TokenSyntax",
+                       initializer: nil,
+                       accessor: body)
+
+      })
+    }
+    let members = MemberDeclList(keywords)
+    let buildable = ExtensionDecl(modifiers: nil,
+                                  extendedType: "TokenSyntax",
+                                  members: members)
+
+    let syntax = buildable.buildSyntax(format: Format())
+
+    var text = ""
+    syntax.write(to: &text)
+
+    XCTAssertEqual(text, """
+    extension TokenSyntax{
+        public var `associatedtype`: TokenSyntax{
+            SyntaxFactory.makeassociatedtypeKeyword()
+        }
+        public var `class`: TokenSyntax{
+            SyntaxFactory.makeclassKeyword()
+        }
+    }
+    """)
+  }
+}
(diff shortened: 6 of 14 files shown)


Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
```diff
@@ -281,5 +281,61 @@ extension ${protocol} {
...
+    MemberDeclBlock(members: self)
+  }
+}
+
+%   end
+%   if 'ExpressibleAsPatternBindingList' in conformances:
+extension ${protocol} {
+  public func createPatternBindingList() -> PatternBindingList {
+    PatternBindingList([self])
+  }
```
kimdv, line 297: We have a lot of those. 
I think we should try to generate them all, so we don't have to add them in the [protocolsMap.py](https://github.com/apple/swift/blob/main/utils/gyb_syntax_support/protocolsMap.py)

ahoppen: If we could make these conformances more generic so we don’t have to spell all of them out, I think that would be a big win. I suppose it also shouldn’t be too hard but haven’t thought about it in detail yet.

Sources/SwiftSyntaxBuilder/IdentifierPatternConvenienceInitializers.swift
```diff
@@ -21,3 +21,13 @@ extension IdentifierPattern: ExpressibleByStringLiteral {
     self.init(value)
   }
 }
+
+extension String: ExpressibleAsIdentifierPattern {
+  public func createPatternBuildable() -> PatternBuildable {
```
ahoppen, line 26: Should this go into a separate extension? `ExpressibleAsIdentifierPattern` only requires `createIdentifierPattern`.

Sources/SwiftSyntaxBuilder/SimpleTypeIdentifierConvenienceInitializers.swift
```diff
@@ -24,3 +24,13 @@ extension SimpleTypeIdentifier: ExpressibleByStringLiteral {
     self.init(value)
   }
 }
+
+extension String: ExpressibleAsSimpleTypeIdentifier {
+  public func createSimpleTypeIdentifier() -> SimpleTypeIdentifier {
```
ahoppen, line 29: Same as for `ExpressibleAsIdentifierPattern`.

Tests/SwiftSyntaxBuilderTest/ExtensionDeclTests.swift
```diff
@@ -0,0 +1,60 @@
...
+    print(text)
+
+    XCTAssertEqual(text, """
+    ␣extension ExpressibleAsTokenSyntax where Self == TokenSyntax{
+        public let `associatedtype`: TokenSyntax{
+            return SyntaxFactory.makeAssociatedtypeKeyword()
+                   .withTrailingTrivia(.spaces(1))
+        }
+        public let `class`: TokenSyntax{
+            return n
```
ahoppen, line 55: Typo?

ahoppen (commented):
I added some preliminary comments inline. I like the idea of trying to generate SwiftSyntax sources using SwiftSyntax. 👍

Sources/SwiftSyntaxBuilder/IdentifierPatternConvenienceInitializers.swift
kimdv: > Should this go into a separate extension?

Do you mean file or do you mean that `public func createPatternBuildable() -> PatternBuildable` should be in another `extension String: ...` ? 

The reason I placed it there is that `ExpressibleAsIdentifierPattern` inherence from `ExpressibleAsPatternBuildable`. 

I could make something like 

```swift
extension {
  public func createPatternBuildable() -> PatternBuildable {
    IdentifierPattern(self)
  }
```

Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
kimdv: I haven't thought about how to make it yet. 
Some quick thoughts: 

1: We need to identify all Collections children, like `PatternBindingList` -> `ExpressibleAsPatternBinding`.
2: When looping through all nodes that is not a SyntaxCollection, we need to see if it's child to a SyntaxCollection.
2: a: If it's not a child ignore
2: b: If it's a child, then we need to add default conformance to the SyntaxCollection

Sources/SwiftSyntaxBuilder/IdentifierPatternConvenienceInitializers.swift
ahoppen: I think in that case we should have 

```
extension ExpressibleAsIdentifierPattern {
  public func createPatternBuildable() -> PatternBuildable {
    return self
  }
}
```

Or doesn’t that work? (I haven’t checked)

Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
ahoppen: Sound good to me. I suppose in that case we don’t need to explicitly specify the collection-element to collection conformances (e.g. `ExpressibleAsConditionElement` to `ExpressibleAsConditionElementList`) in `protocolsMap.py`. 

And we could probably also generate the remaining protocol extensions because they always follow the same pattern:
```swift
extension <ConformingType> {
   public func create<ConfomringToType>() -> <ConformingToType> {
     <ConformingToType>(<someLabel>: self)
   }
 }
```

kimdv: PR in #334

Sources/SwiftSyntaxBuilder/IdentifierPatternConvenienceInitializers.swift
kimdv: PR in #335

Sources/SwiftSyntaxBuilder/StringLiteralExprConvenienceInitializers.swift
```diff
@@ -23,9 +23,3 @@ extension StringLiteralExpr {
               closeQuote: closeQuote)
   }
 }
-
-extension StringLiteralExpr: ExpressibleByStringLiteral {
```
kimdv, line 27: This is not needed anymore as we extend `String`

Sources/SwiftSyntaxBuilder/TypeAnnotationConvenienceInitializers.swift
```diff
@@ -15,9 +15,3 @@ extension TypeAnnotation {
     self.init(type: SimpleTypeIdentifier(type))
   }
 }
-
-extension TypeAnnotation: ExpressibleByStringLiteral {
```
kimdv, line 19: This is not needed anymore as we extend `String`

Sources/SwiftSyntaxBuilder/IdentifierPatternConvenienceInitializers.swift
```diff
@@ -15,9 +15,3 @@ extension IdentifierPattern {
     self.init(identifier: .identifier(identifier))
   }
 }
-
-extension IdentifierPattern: ExpressibleByStringLiteral {
```
kimdv, line 19: This is not needed anymore as we extend `String`

Sources/SwiftSyntaxBuilder/IdentifierExprConvenienceInitializers.swift
```diff
@@ -15,9 +15,3 @@ extension IdentifierExpr {
     self.init(identifier: .identifier(identifier))
   }
 }
-
-extension IdentifierExpr: ExpressibleByStringLiteral {
```
kimdv, line 19: This is not needed anymore as we extend `String`

Sources/SwiftSyntaxBuilder/SimpleTypeIdentifierConvenienceInitializers.swift
```diff
@@ -14,13 +14,7 @@ import SwiftSyntax
 
 extension SimpleTypeIdentifier {
   public init(_ name: String) {
-    self.init(name: SyntaxFactory.makeIdentifier(name),
+    self.init(name: .identifier(name),
               genericArgumentClause: nil)
   }
 }
-
-extension SimpleTypeIdentifier: ExpressibleByStringLiteral {
```
kimdv, line 22: This is not needed anymore as we extend `String`

kimdv:
@ahoppen I think this is now at a point where it makes sense too review :)

Tests/SwiftSyntaxBuilderTest/ExtensionDeclTests.swift
```diff
@@ -0,0 +1,48 @@
+import XCTest
+import SwiftSyntax
+import SwiftSyntaxBuilder
+
+final class ExtensionDeclTests: XCTestCase {
+  func testExtensionDecl() {
+    let leadingTrivia = Trivia.garbageText("␣")
+
+    let keywords = ["associatedtype", "class"].map { keyword -> VariableDecl in
```
kimdv, line 9: I could add a `Token` struct with some of the properties that the [Token.py](https://github.com/apple/swift/blob/main/utils/gyb_syntax_support/Token.py) have. 
And then make an array with a few more tokens? 

Or is this good enough as an example?

ahoppen: I’m not sure I fully understand what you mean. For this test case, I think a list of strings is totally sufficient. Just generating more tokens doesn’t really give us any benefit, I’d say. Or did you think of something else?

Tests/SwiftSyntaxBuilderTest/ExtensionDeclTests.swift
```diff
@@ -0,0 +1,48 @@
+import XCTest
+import SwiftSyntax
+import SwiftSyntaxBuilder
+
+final class ExtensionDeclTests: XCTestCase {
+  func testExtensionDecl() {
+    let leadingTrivia = Trivia.garbageText("␣")
```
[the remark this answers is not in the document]
kimdv: There is no reason. Removed.

kimdv:
https://github.com/apple/swift/pull/40358

@swift-ci Please test

ahoppen (approved):

eeckstein:
@kimdv This breaks the build: https://ci.swift.org/job/oss-swift-incremental-RA-linux-ubuntu-18_04-long-test//6392/console

ahoppen:
CI just got very unlucky with its checkout. It was started at exactly the same time that the PRs were merged and picked up https://github.com/apple/swift-syntax/pull/331 but not https://github.com/apple/swift/pull/40358. It should be fixed in the next build.

PR #332, opened 2021-11-03
Return ExpressibleAs in buildables convenience initializers

kimdv:
Okay!

I thought that making smaller pull requests will make it easier to review than the big #331 

So I started with a simple one. 

I don't know why we missed this, but I think it makes sense also to use the protocol here.


ahoppen (approved):
LGTM

ahoppen:
@swift-ci Please test

PR #334, opened 2021-11-10
Add default conformance to syntax collection

kimdv:
https://github.com/apple/swift/pull/40120

Added default conformance to syntax collection as discussed in #331

### DIFF
--- a/Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
+++ b/Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
@@ -4,6 +4,7 @@
   from gyb_syntax_support.kinds import SYNTAX_BASE_KINDS
   from gyb_syntax_support.kinds import syntax_buildable_child_type, syntax_buildable_default_init_value
   from gyb_syntax_support.protocolsMap import SYNTAX_BUILDABLE_EXPRESSIBLE_AS_CONFORMANCES
+  SYNTAX_COLLECTION_EXPRESSIBLE_AS_CONFORMANCES = syntax_collection_element_to_collection_mapping()
   # -*- mode: Swift -*-
   # Ignore the following admonition it applies to the resulting .swift file only
 }%
@@ -42,7 +43,9 @@ public protocol ${kind}ListBuildable: SyntaxListBuildable {
 
 %     buildable_type = kind + 'Buildable'
 %     expressible_as_type = 'ExpressibleAs' + buildable_type
-%     expressible_as_protocols = SYNTAX_BUILDABLE_EXPRESSIBLE_AS_CONFORMANCES.get(expressible_as_type)
+%     expressible_as_protocols = SYNTAX_COLLECTION_EXPRESSIBLE_AS_CONFORMANCES.get(buildable_type) or []
+%     expressible_as_protocols += SYNTAX_BUILDABLE_EXPRESSIBLE_AS_CONFORMANCES.get(buildable_type) or []
+%     expressible_as_protocols = map(lambda x : 'ExpressibleAs' + x, expressible_as_protocols)
 %     if expressible_as_protocols:
 public protocol ${expressible_as_type}: ${', '.join(expressible_as_protocols)} {
 %     else:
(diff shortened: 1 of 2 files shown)


ahoppen (commented):
LGTM. Will probably need some adjustment based on which changes will be made to https://github.com/apple/swift/pull/40120.

Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
```diff
@@ -42,8 +43,10 @@ public protocol ${kind}ListBuildable: SyntaxListBuildable {
 
 %     buildable_type = kind + 'Buildable'
 %     expressible_by_type = 'ExpressibleBy' + buildable_type
-%     expressible_by_protocols = SYNTAX_BUILDABLE_EXPRESSIBLE_BY_CONFORMANCES.get(expressible_by_type)
-%     if expressible_by_protocols:
+%     expressible_by_protocols = SYNTAX_COLLECTION_EXPRESSIBLE_BY_CONFORMANCES.get(buildable_type) or []
+%     expressible_by_protocols += SYNTAX_BUILDABLE_EXPRESSIBLE_BY_CONFORMANCES.get(buildable_type) or []
+%     expressible_by_protocols = map(lambda x : 'ExpressibleBy' +x, expressible_by_protocols)
```
ahoppen, line 48:
```suggestion
%     expressible_by_protocols = map(lambda x : 'ExpressibleBy' + x, expressible_by_protocols)
```

Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
```diff
@@ -42,8 +43,10 @@ public protocol ${kind}ListBuildable: SyntaxListBuildable {
...
+%     if expressible_by_protocols and expressible_by_protocols is not []:
```
ahoppen, line 49: I think 
```suggestion
%     if expressible_by_protocols:
```
should be enough.

Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
```diff
@@ -220,8 +223,10 @@ public struct ${node.syntax_kind}: SyntaxBuildable {
 %   end
 %   if node.is_buildable() or node.is_syntax_collection():
 %     expressible_by_type = 'ExpressibleBy' + node.syntax_kind
-%     expressible_by_protocols = SYNTAX_BUILDABLE_EXPRESSIBLE_BY_CONFORMANCES.get(expressible_by_type)
-%     if expressible_by_protocols:
+%     expressible_by_protocols = SYNTAX_COLLECTION_EXPRESSIBLE_BY_CONFORMANCES.get(node.syntax_kind) or []
+%     expressible_by_protocols += SYNTAX_BUILDABLE_EXPRESSIBLE_BY_CONFORMANCES.get(node.syntax_kind) or []
+%     expressible_by_protocols = map(lambda x : 'ExpressibleBy' +x, expressible_by_protocols)
```
ahoppen, line 228:
```suggestion
%     expressible_by_protocols = map(lambda x : 'ExpressibleBy' + x, expressible_by_protocols)
```

Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
```diff
@@ -249,37 +254,40 @@ extension TokenSyntax: ExpressibleByTokenSyntax {
...
 }
 
 %   end
-%   if 'ExpressibleByConditionElement' in conformances:
-extension ${protocol} {
+% end
+
+% for protocol, conformances in SYNTAX_BUILDABLE_EXPRESSIBLE_BY_CONFORMANCES.items():
+%   if 'ConditionElement' in conformances:
+extension ExpressibleBy${protocol} {
```
ahoppen, line 270: Not part of this PR, but couldn’t we also generate these `extension` implementations now that syntax collection conformances live somewhere else?

kimdv: I would be nice! 
But the thing is, that the init method different for each type. 

We can enrich the `protocols.py`file with the needed information.

ahoppen: Why is that? Am I missing something obvious? Shouldn’t the following work?

```swift
extension ExpressibleBy${protocol} {
   public func create${expressible_as_protocol}() -> ${expressible_as_protocol} {
     ${expressible_as_protocol}(${retrieve only non-optional child name here}: self)
   }
 }

```

kimdv:
```swift
extension ExpressibleByDeclBuildable {
  public func createCodeBlockItem() -> CodeBlockItem {
    CodeBlockItem(item: self)
  }
}

extension ExpressibleByDeclBuildable {
  public func createMemberDeclListItem() -> MemberDeclListItem {
    MemberDeclListItem(decl: self)
  }
}
```

Yes. Sorry bad explanation 😅
The `${retrieve only non-optional child name here}` do we need to find somehow. 
Not sure how to do, but I will look into it, and then we can take it from there. 

It is possible, somehow 😬

ahoppen (approved):
Looking good. Could you rebase because I just reverted my change of `ExpressibleAs` -> `ExpressibleBy` because I had a brain fart yesterday. The protocols actually describe that these types are expressible **as** other types and not **by** other types. Also make sure to rename all of your `by`s by `as`s.

kimdv:
I saw that! 

I have rebased 🚀

kimdv:
apple/swift#40120

@swift-ci Please test

ahoppen:
apple/swift#40120

@swift-ci Please test

PR #336, opened 2021-11-14
Make syntax collections conform to `ExpressibleByArrayLiteral`

kimdv:


ahoppen:
Without trying it myself, have you tried 
```swift
extension ExpressibleAsExprList: ExpressibleByArrayLiteral {
  …
}
```

or, if that doesn’t work, add the protocol conformance to `ExpressibleByArrayLiteral` to the definition of `ExpressibleAsExprList`.

---

Unrelated to this PR: I just realized that the standard library protocols are called Expressible**By** and ours are called Expressible**As**. Maybe we should change our protocols to also be called Expressible**By**.

kimdv:
> Unrelated to this PR: I just realized that the standard library protocols are called Expressible**By** and ours are called Expressible**As**. Maybe we should change our protocols to also be called Expressible**By**.

Funny. I actually also noticed that yesterday when playing around with this 😆

kimdv:
The problem is that protocols like:

```swift
public protocol ExpressibleByConditionElement: ExpressibleByConditionElementList {
  func createConditionElement() -> ConditionElement
}
```
```swift
public protocol ExpressibleByConditionElementList: ExpressibleByArrayLiteral {
  func createConditionElementList() -> ConditionElementList
}
```

`ConditionElement` then need to implement `init(arrayLiteral elements: Self.ArrayLiteralElement...)` because of this. 

There will come more of those cases as we add more conformances in [protocols.py](https://github.com/apple/swift/blob/main/utils/gyb_syntax_support/protocolsMap.py)

ahoppen:
This shouldn’t be an issue though if we implement `it(arrayLiteral elements: Self.ArrayLiteralElement…)` as an extension on `ExpressibleByConditionElement` right?

kimdv:
> This shouldn’t be an issue though if we implement `it(arrayLiteral elements: Self.ArrayLiteralElement…)` as an extension on `ExpressibleByConditionElement` right?

I'm not sure I understand the question?
Do you think that `ExpressibleByConditionElement ` need to implement the init(arrayLiteral elements: Self.ArrayLiteralElement…) or should `ExpressibleByConditionElementList` implement it? 

 
Because I don't this this would be possible (I haven't tried it)
```swift
public protocol ExpressibleByConditionElement {
  init(arrayLiteral elements: Self.ArrayLiteralElement…) {
  self. createConditionElement(self)  
  }
}
```

Do we not need to call an `init` method inside an `init`?

ahoppen:
I was thinking of something like this.
```swift
// NEW: Add ExpressibleByArrayLiteral conformance here
protocol ExpressibleAsExprList : ExpressibleByArrayLiteral {}

protocol ExpressibleAsExprBuildable {}

// NEW: Generate this extension
extension ExpressibleAsExprList {
  init(arrayLiteral elements: ExpressibleAsExprBuildable...) {
    // self.init(elements) or something like this. `elements` has type `Array<ExpressibleAsExprBuildable>`
  }
}

struct ExprList : ExpressibleAsExprList {}

struct ExprBuildable : ExpressibleAsExprBuildable {}

```

kimdv:
> I was thinking of something like this.
> 
> ```swift
> // NEW: Add ExpressibleByArrayLiteral conformance here
> protocol ExpressibleAsExprList : ExpressibleByArrayLiteral {}
> 
> protocol ExpressibleAsExprBuildable {}
> 
> // NEW: Generate this extension
> extension ExpressibleAsExprList {
>   init(arrayLiteral elements: ExpressibleAsExprBuildable...) {
>     // self.init(elements) or something like this. `elements` has type `Array<ExpressibleAsExprBuildable>`
>   }
> }
> 
> struct ExprList : ExpressibleAsExprList {}
> 
> struct ExprBuildable : ExpressibleAsExprBuildable {}
> ```

We cannot access the `self.init` on `ExpressibleAsExprList` as it is a protocol?
Or am I missing something? 

When `ExpressibleByConditionElement` implements `ExpressibleByConditionElementList` be doing

https://github.com/apple/swift-syntax/blob/af253dc89d179b33bf8d0c4340678f42c18738ed/Sources/SwiftSyntaxBuilder/gyb_generated/Buildables.swift#L11205-L11209

So I'm not sure how we don't need to implement the `init(arrayLiteral elements: ExpressibleAsExprBuildable...)`

ahoppen:
@swift-ci Please test

kimdv:
@ahoppen when you have time, will you see my [comment](https://github.com/apple/swift-syntax/pull/336#issuecomment-970697065)?

ahoppen:
Sorry, I completely missed your comment. Thanks for pinging me.

Of course, you’re right. We don’t want to be able to create an arbitrary `ExpressibleAsExprList` from an array literal, but only an `ExprList` itself, so we shouldn’t be conforming `ExpressibleAsExprList: ExpressibleByArrayLiteral`, but `ExprList: ExpressibleByArrayLiteral`, so something like this:

```swift
extension ExprList: ExpressibleByArrayLiteral {
  init(arrayLiteral elements: ExpressibleAsExprBuildable...) {
    self.init(elements)
  }
}
```

This also matches your initial description of the PR where `createExprList` creates an `ExprList` and not an `ExpressibleAsExprList`.

kimdv:
@swift-ci Please test

ahoppen (approved):
Looks like a very nice addition to me.

kimdv:
@swift-ci Please test

kimdv:
@swift-ci Please test Linux

PR #337, opened 2021-11-15
Replace ExpressibleAs protocols by ExpressibleBy protocols

ahoppen:
The protocols in the standard library are called `ExpressibleBy*` and SwiftSyntax should follow the same naming convention.

CC: @kimdv

PR #338, opened 2021-11-16
Revert "Replace ExpressibleAs protocols by ExpressibleBy protocols"

ahoppen:
Reverts apple/swift-syntax#337

---

Stupid me, the protocols actually describe that these types are expressible **as** other types and not **by** other types.


ahoppen:
https://github.com/apple/swift/pull/40200

@swift-ci Please test

PR #339, opened 2021-11-20
Generate extensions for protocolsMap.py

kimdv:
Generate extensions for [protocolsMap.py](https://github.com/apple/swift/blob/main/utils/gyb_syntax_support/protocolsMap.py).

### DIFF
--- a/Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
+++ b/Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
@@ -5,6 +5,7 @@
   from gyb_syntax_support.kinds import syntax_buildable_child_type, syntax_buildable_default_init_value
   from gyb_syntax_support.protocolsMap import SYNTAX_BUILDABLE_EXPRESSIBLE_AS_CONFORMANCES
   SYNTAX_COLLECTION_EXPRESSIBLE_AS_CONFORMANCES = syntax_collection_element_to_collection_mapping()
+  NODE_MAP = create_node_map()
   # -*- mode: Swift -*-
   # Ignore the following admonition it applies to the resulting .swift file only
 }%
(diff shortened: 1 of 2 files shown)


Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
```diff
@@ -264,57 +265,17 @@ extension TokenSyntax: ExpressibleAsTokenSyntax {
...
-% for protocol, conformances in SYNTAX_BUILDABLE_EXPRESSIBLE_AS_CONFORMANCES.items():
-%   if 'ExpressibleAsConditionElementList' in conformances:
-extension ${protocol} {
-  public func createConditionElementList() -> ConditionElementList {
-    ConditionElementList([self])
-  }
-}
-
-%   end
-% end
```
kimdv, line 277: This one properly slipped through when added #334

Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
```diff
@@ -264,57 +265,17 @@ extension TokenSyntax: ExpressibleAsTokenSyntax {
...
-
 % for protocol, conformances in SYNTAX_BUILDABLE_EXPRESSIBLE_AS_CONFORMANCES.items():
-%   if 'ConditionElement' in conformances:
+%   for conformance in conformances:
+%     node = NODE_MAP.get(conformance)
+%     if node and node.children:
+%       param = filter(lambda x : x.type_name != "TokenSyntax", node.children)[0].swift_name
```
ahoppen, line 272: I don’t think this is correct. We want to the first child that doesn’t have a default value. Currently optional children also have a default value of `nil` and `TokenSyntax` children might not have a default value if they don’t have a default text (see https://github.com/apple/swift/blob/17f1cf92973d2c1b3d5fe6b70f4a33771a303e3a/utils/gyb_syntax_support/kinds.py#L58-L64)

I think a better implementation would be 

```python
non_defaulted_params = filter(lambda child : syntax_buildable_default_init_value(child, SYNTAX_TOKEN_MAP.get(child.syntax_kind)) == None, node.children)
assert len(non_defaulted_params) == 1, "ExpressibleAs conformances expects the conforming type to have an initializer with a single non-optional child"
param = non_defaulted_params[0].swift_name
```

kimdv: Ah, I didn't see that. 
Added. Tough changes the `== None` to `== ""` as we return an empty string.

kimdv:
@swift-ci Please test

ahoppen (approved):

ahoppen:
@swift-ci Please test

PR #123, opened 2019-05-12
SR-10241 Consolidate StringInterpolationExprSyntax to StringLiteralExprSyntax

kitasuke:
## Description

This PR is required by https://github.com/apple/swift/pull/24718 to apply same changes to SwiftSyntax

- [x] Fix test case for tree
- [x] Modify factory method in SyntaxFactory

PR #341, opened 2021-11-27
Add asserts in init methods

kimdv:
The idea is to add checks to help us and other developers to make valid swift code.

This only add for `text_choices` and `token_choices`. 
I have not figured out how to add for `node_choices`, but this is a start.

https://github.com/apple/swift/pull/40359

### DIFF
--- a/Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
+++ b/Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
@@ -159,6 +159,20 @@ public struct ${node.syntax_kind}: ${node.base_kind}Buildable {
 %       else:
     self.${child.swift_name} = ${child.swift_name}${create_method_dot}create${param_type}()
 %       end
+%       assert_choices = []
+%       if child.text_choices:
+%         assert_choices = child.text_choices
+%       elif child.token_choices:
+%         assert_choices = map(lambda x : SYNTAX_TOKEN_MAP.get(x.name + 'Token').text, child.token_choices)
+%         # If a Token has `None` as `text`, it can contain arbitrary text.
+%         # If any of the token choices may contain arbitrary text, set assert_choices to [] so we don’t assert on the token’s text below.
+%         assert_choices = [] if None in assert_choices else assert_choices
+%       end
+%       if assert_choices:
+%         optional_condition = child.swift_name + ' == nil || ' if child.is_optional else ""
+%         unwrap = "!" if child.is_optional else ""
+    assert(${optional_condition}[${', '.join(map(lambda x : '"' + x + '"', assert_choices))}].contains(self.${child.swift_name}${unwrap}.text))
+%       end
 %     end
   }
   
--- a/Sources/SwiftSyntaxBuilder/gyb_generated/Buildables.swift
+++ b/Sources/SwiftSyntaxBuilder/gyb_generated/Buildables.swift
@@ -369,6 +369,7 @@ public struct CodeBlockItem: SyntaxBuildable {
   ) {
     self.item = item.createSyntaxBuildable()
     self.semicolon = semicolon
+    assert(semicolon == nil || [";"].contains(self.semicolon!.text))
     self.errorTokens = errorTokens?.createSyntaxBuildable()
   }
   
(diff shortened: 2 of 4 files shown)


Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
```diff
@@ -136,6 +136,20 @@ public struct ${node.syntax_kind}: ${node.base_kind}Buildable {
 %       child_token = SYNTAX_TOKEN_MAP.get(child.syntax_kind)
 %       param_type = syntax_buildable_child_type(child.type_name, child.syntax_kind, child.is_token())
     self.${child.swift_name} = ${child.swift_name}${create_method_dot}create${param_type}()
+%       assert_choices = []
+%       if child.text_choices:
+%         assert_choices = map(lambda x : '"' + x + '"', child.text_choices)
+%       elif child.token_choices:
+%         assert_choices = map(lambda x : '"' + SYNTAX_TOKEN_MAP.get(x.name + 'Token').text + '"', child.token_choices)
+%         # some cases, like IdentifierToken, will add an empty string.
+%         # So if there is any empty string we just make it empty
```
ahoppen, line 145: If I understand your comment correctly, we currently set a `Token`’s `text` to `""` if it essentially is a wildcard token and doesn’t have a fixed text, correct?

Since we now start relying on that behavior, I think it would be good if we document it in `Token.py`. E.g.
```
If `text` is not empty, a token of this type must always contain the specified `text`. If `text` is empty, the token can contain arbitrary text.
```
And while we’re at it, I think it would be nicer if we could use `None` instead of an empty string to denote the behavior that a token can contain arbitrary text – not sure what other changes that entails though.

--- 

Also a suggestion to make this comment clearer:
```
If a Token has an empty text, it can contain arbitrary text. If any of the token choices may contain arbitrary text, set assert_choices to [] so we don’t assert on the token’s text below.
```

kimdv: Yes the `text` in a `Token` is set to `""` if it is `None`.
https://github.com/apple/swift/blob/f00f2841abf79b81a20b01936137eb095bd9b1df/utils/gyb_syntax_support/Token.py#L21

I agree that it would be more nice with an `None` instead of an `""` in Token.

Added the comment above text #123

kimdv: I have run degyb locally where `Token.text` is None. Thus removed `or ""`.
It didn't generate any other files? 

Are there any other repos depending on it? 
I could push and we could try to run CI?

ahoppen: I’m not entirely if I understood your questions correctly, but I’m trying to answer as best as I can.

> I have run degyb locally where `Token.text` is None. Thus removed `or ""`.
> It didn't generate any other files? 

It could very well be that changing `Token.text` from `""` to `None` doesn’t change any of the generated files in the SwiftSyntax repo, so this might be expected.

> Are there any other repos depending on it? 
`gyb_syntax_support` is only used by the compiler itself (in `include/swift/Syntax` and `lib/Syntax`) and SwiftSyntax. So any breakage will be found by CI.

> I could push and we could try to run CI? 
Sounds good to me.

kimdv:
https://github.com/apple/swift/pull/40359

@swift-ci Please test

Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
kimdv: CI did pass 🥳

Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
```diff
@@ -159,6 +159,20 @@ public struct ${node.syntax_kind}: ${node.base_kind}Buildable {
 %       else:
     self.${child.swift_name} = ${child.swift_name}${create_method_dot}create${param_type}()
 %       end
+%       assert_choices = []
+%       if child.text_choices:
+%         assert_choices = child.text_choices
+%       elif child.token_choices:
+%         assert_choices = map(lambda x : SYNTAX_TOKEN_MAP.get(x.name + 'Token').text, child.token_choices)
+%         # If a Token has an empty text, it can contain arbitrary text.
```
ahoppen, line 167:
```suggestion
%         # If a Token has `None` as `text`, it can contain arbitrary text. 
```

ahoppen (approved):
A small comment nitpick. Otherwise 👍

kimdv:
https://github.com/apple/swift/pull/40359

@swift-ci Please test

PR #343, opened 2021-11-27
Remove expressible as syntax token

kimdv:
Depends on #339

### DIFF
--- a/Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
+++ b/Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
@@ -8,6 +8,13 @@
   NODE_MAP = create_node_map()
   # -*- mode: Swift -*-
   # Ignore the following admonition it applies to the resulting .swift file only
+
+  # Use [:] to make sure we copy the conformances so that we don't modify
+  # `SYNTAX_COLLECTION_EXPRESSIBLE_AS_CONFORMANCES` when adding elements to `expressible_as_protocols` below.
+  def get_expressible_as_conformances(conforming_type):
+    expressible_as_protocols = (SYNTAX_COLLECTION_EXPRESSIBLE_AS_CONFORMANCES.get(conforming_type) or [])[:]
+    expressible_as_protocols += (SYNTAX_BUILDABLE_EXPRESSIBLE_AS_CONFORMANCES.get(conforming_type) or [])[:]
+    return map(lambda x : 'ExpressibleAs' + x, expressible_as_protocols)
 }%
 //// Automatically Generated From DeclBuildables.swift.gyb.
 //// Do Not Edit Directly!
@@ -44,9 +51,7 @@ public protocol ${kind}ListBuildable: SyntaxListBuildable {
 
 %     buildable_type = kind + 'Buildable'
 %     expressible_as_type = 'ExpressibleAs' + buildable_type
-%     expressible_as_protocols = SYNTAX_COLLECTION_EXPRESSIBLE_AS_CONFORMANCES.get(buildable_type) or []
-%     expressible_as_protocols += SYNTAX_BUILDABLE_EXPRESSIBLE_AS_CONFORMANCES.get(buildable_type) or []
-%     expressible_as_protocols = map(lambda x : 'ExpressibleAs' + x, expressible_as_protocols)
+%     expressible_as_protocols = get_expressible_as_conformances(buildable_type)
 %     if expressible_as_protocols:
 public protocol ${expressible_as_type}: ${', '.join(expressible_as_protocols)} {
 %     else:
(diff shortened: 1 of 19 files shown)


ahoppen (approved):

ahoppen:
@swift-ci Please test

kimdv:
@swift-ci Please test

Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
```diff
@@ -258,21 +268,21 @@ extension ${node.syntax_kind}: ${expressible_as_type} {
...
-  public func createTokenSyntax() -> TokenSyntax {
-    self
-  }
+% expressible_as_protocols = SYNTAX_COLLECTION_EXPRESSIBLE_AS_CONFORMANCES.get('TokenSyntax') or []
+% expressible_as_protocols += SYNTAX_BUILDABLE_EXPRESSIBLE_AS_CONFORMANCES.get('TokenSyntax') or []
+% expressible_as_protocols = map(lambda x : 'ExpressibleAs' + x, expressible_as_protocols)
+% if expressible_as_protocols:
+extension TokenSyntax: ${', '.join(expressible_as_protocols)} {
 }
+% end
```
kimdv, line 278: @ahoppen added this section after your approval

Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
```diff
@@ -258,36 +268,38 @@ extension ${node.syntax_kind}: ${expressible_as_type} {
...
 % end
-public protocol ExpressibleAsTokenSyntax {
-  func createTokenSyntax() -> TokenSyntax
+% expressible_as_protocols = (SYNTAX_COLLECTION_EXPRESSIBLE_AS_CONFORMANCES.get('TokenSyntax') or [])[:]
+% expressible_as_protocols += (SYNTAX_BUILDABLE_EXPRESSIBLE_AS_CONFORMANCES.get('TokenSyntax') or [])[:]
+% expressible_as_protocols = map(lambda x : 'ExpressibleAs' + x, expressible_as_protocols)
+% if expressible_as_protocols:
+extension TokenSyntax: ${', '.join(expressible_as_protocols)} {
 }
+% end
```
kimdv, line 277: @ahoppen added this section after your approval. 

I also discovered, in another PR, that if we don't add `[:]` ad the end, we are copy by reference. 
I'm not sure why it workers the other places. 

Correct me if I'm wrong

ahoppen: Oh, that is surprising. Very good catch.

To make sure we don’t hit this again, I think it would be good if we wrapped `SYNTAX_COLLECTION_EXPRESSIBLE_AS_CONFORMANCES.get` in a function. 
E.g.
```python
def get_expressible_as_conformances(conforming_type):
  expressible_as_protocols = (SYNTAX_COLLECTION_EXPRESSIBLE_AS_CONFORMANCES.get(conforming_type) or [])[:]
  expressible_as_protocols += (SYNTAX_BUILDABLE_EXPRESSIBLE_AS_CONFORMANCES.get(conforming_type) or [])[:]
  return map(lambda x : 'ExpressibleAs' + x, expressible_as_protocols)
```

and also use that in the other places.

kimdv: Something like this?

kimdv:
@swift-ci Please test

Sources/SwiftSyntaxBuilder/Buildables.swift.gyb
```diff
@@ -8,6 +8,11 @@
   NODE_MAP = create_node_map()
   # -*- mode: Swift -*-
   # Ignore the following admonition it applies to the resulting .swift file only
+
+  def get_expressible_as_conformances(conforming_type):
+    expressible_as_protocols = (SYNTAX_COLLECTION_EXPRESSIBLE_AS_CONFORMANCES.get(conforming_type) or [])[:]
```
ahoppen, line 13: Could you add a comment why `[:]` is necessary here? Something like
```
# Use [:] to make sure we copy the conformances so that we don't modify `SYNTAX_COLLECTION_EXPRESSIBLE_AS_CONFORMANCES` when adding elements to `expressible_as_protocols` below.
```

ahoppen (approved):
One request for a comment, otherwise looking good.

ahoppen:
🎉 CI testing is honoring your requests, @kimdv 🎉

kimdv:
@swift-ci Please test

kimdv:
@ahoppen I have added the comment and triggered the CI. 
I assume that I can merge, now that you have approved it? 

If not, sorry. Please tell then!

ahoppen:
Yes, if CI passes and you’ve got approval, you may merge.

PR #344, opened 2021-11-29
Add convenience initializer for `BinaryOperatorExpr`

kimdv:
https://github.com/apple/swift/pull/40355
~~Depends on #343~~

### DIFF
--- a/Sources/SwiftSyntaxBuilder/BinaryOperatorExprConvenienceInitializers.swift
+++ b/Sources/SwiftSyntaxBuilder/BinaryOperatorExprConvenienceInitializers.swift
@@ -1,0 +1,19 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2021 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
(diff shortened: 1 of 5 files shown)


Sources/SwiftSyntaxBuilder/BinaryOperatorExprConvenienceInitializers.swift
```diff
@@ -0,0 +1,25 @@
...
+
+import SwiftSyntax
+
+extension BinaryOperatorExpr {
+  public init(_ text: String) {
+    self.init(operatorToken: TokenSyntax.spacedBinaryOperator(text))
+  }
+}
+
+extension BinaryOperatorExpr: ExpressibleByStringLiteral {
```
ahoppen, line 21: Do we actually need the `ExpressibleByStringLiteral` conformance? IIUC we should be able to use `String`s for binary operators in all relevant places because we have `String: ExpressibleAsBinaryOperatorExpr`.

kimdv: It's possible yes.

kimdv:
https://github.com/apple/swift/pull/40355

@swift-ci Please test

ahoppen (approved):

PR #345, opened 2021-11-29
Add convenience initializer for `ReturnClause`

kimdv:


Sources/SwiftSyntaxBuilder/ReturnClauseConvenienceInitializers.swift
```diff
@@ -0,0 +1,19 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
+extension ReturnClause: ExpressibleByStringLiteral {
```
ahoppen, line 15: Do we actually need the `ExpressibleByStringLiteral` conformance? IIUC we should be able to use `String`s for binary operators in all relevant places because we have `String: ExpressibleAsReturnClause `.

kimdv: Fixed.

ahoppen (approved):

kimdv:
@swift-ci Please test

PR #351, opened 2022-01-13
Include Swift 5.5’s `_InternalSwiftSyntaxParser.dylib` in the repo

ahoppen:
Previously, clients of SwiftSyntax always needed to provide a `_InternalSwiftSyntaxParser.dylib` that was matching the SwiftSyntax version they are using. This was fragile and annoying.

Add the `_InternalSwiftSyntaxParser.dylib` that was released as part of Swift 5.5 to the repo as a binary framework dependency. This way, users can just depend on the SwiftSyntax package and it will bring along the `_InternalSwiftSyntaxParser.dylib` it needs.

If you want to try it, create a Swift package and add the following dependency

```
.package(url: "https://github.com/ahoppen/swift-syntax.git", .branch("pr-5.5/include-parserlib"))
```

This PR also contains a few changes to `Package.swift` to fix warnings that resulted from upgrading the package manifest version to 5.3.

Since we need `_InternalSwiftSyntaxParser.dylib` to be built before we can add it to the repo, we can’t include the framework in the `main` development branch. My plan is to add it to the release branches once the corresponding Swift version has been released. If we merge this, I would like to tag the resulting commit `0.50500.1`.

### To-Dos

- [ ] The parser library is currently vended as an `.xcframework` that is only compatible with macOS. AFAICT it is just ignored on Linux. @neonichu Could you verify that this is indeed the case or do we need special checks for macOS in `Package.swift` before we include the binary target? When we release Swift 5.6 and have [SE-0305](https://github.com/apple/swift-evolution/blob/main/proposals/0305-swiftpm-binary-target-improvements.md) we should also be able to include the parser library for Linux.
- [ ] Update the parser library’s URL to a GitHub release hosted on Apple’s repository instead of my personal one.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.1
+// swift-tools-version:5.3
 
 import PackageDescription
 import Foundation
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please smoke test

ahoppen:
I have just played around with hosting the parser library via GitHub’s releases by creating a release on my personal GitHub [Link](https://github.com/ahoppen/swift-syntax/releases/tag/ahoppen-0.50500.2).

ahoppen:
Created a snapshot prerelease of SwiftSyntax 0.50600.0 that vendors the parser library here: https://github.com/apple/swift-syntax/releases/tag/0.50600.0-SNAPSHOT-2022-01-24

PR #352, opened 2022-01-20
Update Package manifest to tools version 5.3

ahoppen:
Vending the parser library as a binary dependency requires Swift tools version 5.3. In prepration for that, update the tools version, without making any semantic changes.

Syntactically, the following has chagned:
- Some cleanup to add all targets and products in the `Package` initializer call. It looks like previous, dynamic addition is not needed anymore, possibly because SwiftSyntax now always links statically
- Explicitly add non-Swift files to the exclude list, as required by 5.3 tools version.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,24 +1,9 @@
-// swift-tools-version:5.1
+// swift-tools-version:5.3
 
 import PackageDescription
 import Foundation
 
-let package = Package(
-  name: "SwiftSyntax",
-  targets: [
-    .target(name: "_CSwiftSyntax"),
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

akyrtzi (approved):

ahoppen:
@swift-ci Please smoke test Linux

ahoppen:
@swift-ci Please test Linux

ahoppen:
Linux is failing because it’s running on Ubuntu 16.04 but Swift requires a more recent clang than installed on those machines. We need to wait until the CI job is updated to a more recent Ubuntu.

PR #356, opened 2022-02-08
Add `ProtocolDecl` test cases

kimdv:


kimdv:
@swift-ci please test

Tests/SwiftSyntaxBuilderTest/ProtocolDeclTests.swift
```diff
@@ -0,0 +1,29 @@
+import XCTest
+import SwiftSyntax
+import SwiftSyntaxBuilder
+
+final class ProtocolDeclTests: XCTestCase {
+  func testProtocolDecl() {
+    let returnClause = ArrayType(elementType: "DeclSyntax")
+    let input = ParameterClause(parameterListBuilder: {
+      FunctionParameter(firstName: .identifier("format"), colon: .colon, type: "Format", trailingComma: .comma, attributesBuilder: {})
+      FunctionParameter(firstName: .identifier("leadingTrivia"), colon: .colon, type: OptionalType(wrappedType: "Trivia"), attributesBuilder: {})
```
ahoppen, line 10: Is the `attributesBuilder` argument necessary? Shouldn’t it be defaulted?

Tests/SwiftSyntaxBuilderTest/ProtocolDeclTests.swift
```diff
@@ -0,0 +1,29 @@
...
+    })
+    let functionSignature = FunctionSignature(input: input, output: returnClause)
+
+    // FIXME: We need to add the `modifiersBuilder` with a none empty value, otherwise will the builder omit newline.
```
ahoppen, line 14: Is this comment still up-to-date? I don’t see a `modifiersBuilder` with an empty value.

ahoppen (approved):
Some minor comments, otherwise LGTM.

Tests/SwiftSyntaxBuilderTest/ProtocolDeclTests.swift
kimdv: Maybe my comment is not clear, it should not be empty.
`none empty value`

ahoppen: Ah, I see. In that case the comment the comment should say `non-empty` instead of `none empty`.

Tests/SwiftSyntaxBuilderTest/ProtocolDeclTests.swift
kimdv: It should yes. 
But the problem is that we then get `Ambiguous use of 'init'`.

I think we should, somehow, solve this as there is many `init` methods that look like each other so we need to do things like this to avoid the error. Now it works now, but I think it makes some noice when using it.

This is a long shot, and I'm not sure if it is technically possible, but I'll try to explain. 

It would be nice if we could remove this type of `init` methods, where the `@resultBuilder` is the only difference.
This will mean, in this case, that our `AttributeListBuilder` need to conform to `ExpressibleAsAttributeList`.
And then we could things like `FunctionParameter(modifiers: { TokenSyntax.public }, firstName: .identifier("format"), colon: .colon, type: "Format", trailingComma: .comma)` or 
`FunctionParameter(modifiers: TokenSyntax.public, firstName: .identifier("format"), colon: .colon, type: "Format", trailingComma: .comma)`. 

I think we before have talked about that the `@resultBuilders` should conform to some of the protocols.

kimdv:
@swift-ci please test

kimdv:
Another question @ahoppen.
Do you know any place where this is used? 
Because it would be cool to see how it is used and what we can do to make this package even better!

kimdv:
@swift-ci please test linux

kimdv:
Pinging @ahoppen 😊

Tests/SwiftSyntaxBuilderTest/ProtocolDeclTests.swift
ahoppen: I see, makes sense 👍 I’m good with merging this as-is.

As to your other question: 
If I understand you correctly, you are saying that you’d like to remove the second initializer listed in the code snippet and make `AttributeListBuilder` conform to `ExpressibleAsAttributeList`. In that case you expect to be able to use the `AttributeListBuilder` result builder wherever an `ExpressibleAsAttributeList` is expected, right? 
If this is what you’re suggesting, I’m pretty sure it won’t work. I think the result builder needs to be declared using `@AttributeListBuilder` on the parameter or otherwise the type checker won’t try to apply it.

ahoppen:
Thanks for the ping, I missed your other comments.

> Do you know any place where this is used?

Unfortunately, I don’t know if it’s used at all, but I agree that I think the syntax builder has matured enough that we could start using it for real and refine it based on real world usage. What do you think about the following
- Let’s update the [README.md](https://github.com/apple/swift-syntax/blob/main/Sources/SwiftSyntaxBuilder/README.md) to the new syntax and include instructions how to use it (that is, which dependencies to declare in SwiftPM etc.)
- When we’ve got a good readme with a few compelling examples, make a Swift Forums post in the [Source Tooling category](https://forums.swift.org/c/development/tooling/39), announcing SwiftSyntaxBuilder, asking people to try it and give feedback (we can use the SwiftSyntax component on JIRA for bugs). Since you are the reason that SwiftSyntaxBuilder exists in the current state at all, I think you should also have the honor of creating the blog post if you wish

What do you think?

kimdv:
That sounds like a good idea! 
I will update the README, and make a draft post, that we then can post in the forums!

PR #357, opened 2022-02-08
Added README.md to exclude list

kimdv:
The README.md was missing, so it generated a warning. 
Just excluded it as the other readme


ahoppen (approved):
Thanks

PR #358, opened 2022-02-11
rewrite swiftSettings to avoid causing SPM build failures

kelvin13:
fixes the issue [described here](https://forums.swift.org/t/cannot-edit-spm-package-with-a-swiftsyntax-dependency/55281) by kicking the unsafe flags logic up a level, so non-CI builds pass an empty array of `SwiftSetting`s instead of an empty array of unsafe flags.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -6,7 +6,7 @@ import Foundation
 /// If we are in a controlled CI environment, we can use internal compiler flags
 /// to speed up the build or improve it.
 let swiftSyntaxSwiftSettings: [SwiftSetting] 
-if case _? = ProcessInfo.processInfo.environment["SWIFT_BUILD_SCRIPT_ENVIRONMENT"] {
+if ProcessInfo.processInfo.environment["SWIFT_BUILD_SCRIPT_ENVIRONMENT"] != nil {
   let groupFile = URL(fileURLWithPath: #file)
     .deletingLastPathComponent()
     .appendingPathComponent("utils")
--- a/Package.swift
+++ b/Package.swift
@@ -3,22 +3,21 @@
 import PackageDescription
 import Foundation
 
-var swiftSyntaxUnsafeSwiftFlags: [String] = []
-
 /// If we are in a controlled CI environment, we can use internal compiler flags
 /// to speed up the build or improve it.
-if ProcessInfo.processInfo.environment["SWIFT_BUILD_SCRIPT_ENVIRONMENT"] != nil {
+let swiftSettings:[SwiftSetting] = ProcessInfo.processInfo.environment["SWIFT_BUILD_SCRIPT_ENVIRONMENT"].map { _ in 
   let groupFile = URL(fileURLWithPath: #file)
     .deletingLastPathComponent()
     .appendingPathComponent("utils")
     .appendingPathComponent("group.json")
-
-  swiftSyntaxUnsafeSwiftFlags += ["-Xfrontend", "-group-info-path",
-                                "-Xfrontend", groupFile.path]
-  // Enforcing exclusivity increases compile time of release builds by 2 minutes.
-  // Disable it when we're in a controlled CI environment.
-  swiftSyntaxUnsafeSwiftFlags += ["-enforce-exclusivity=unchecked"]
-}
+  return  [.unsafeFlags([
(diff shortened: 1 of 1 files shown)


ahoppen (commented):
Thanks for fixing this. For readability, however, I would prefer to use an `if` instead the `map`, and just change `swiftSyntaxUnsafeSwiftFlags 

```swift
var swiftSyntaxSwiftSettings: [SwiftSetting] = []

/// If we are in a controlled CI environment, we can use internal compiler flags
/// to speed up the build or improve it.
if ProcessInfo.processInfo.environment["SWIFT_BUILD_SCRIPT_ENVIRONMENT"] != nil {
  let groupFile = URL(fileURLWithPath: #file)
    .deletingLastPathComponent()
    .appendingPathComponent("utils")
    .appendingPathComponent("group.json")

  swiftSyntaxSwiftSettings.append(.unsafeFlags([
    "-Xfrontend", "-group-info-path",
    "-Xfrontend", groupFile.path,
    // Enforcing exclusivity increases compile time of release builds by 2 minutes.
    // Disable it when we're in a controlled CI environment.
    "-enforce-exclusivity=unchecked"
  ]))
}
```

kelvin13:
rewrote it without the `map`, is

```swift 
if case _? = ProcessInfo.processInfo.environment["SWIFT_BUILD_SCRIPT_ENVIRONMENT"] {
```

fine?

Package.swift
```diff
@@ -3,21 +3,23 @@
 import PackageDescription
 import Foundation
 
-var swiftSyntaxUnsafeSwiftFlags: [String] = []
-
 /// If we are in a controlled CI environment, we can use internal compiler flags
 /// to speed up the build or improve it.
-if ProcessInfo.processInfo.environment["SWIFT_BUILD_SCRIPT_ENVIRONMENT"] != nil {
+let swiftSyntaxSwiftSettings:[SwiftSetting]
```
ahoppen, line 8: Nitpick to match the existing style in SwiftSyntax
```suggestion
let swiftSyntaxSwiftSettings: [SwiftSetting] 
```

Package.swift
```diff
@@ -3,21 +3,23 @@
 import PackageDescription
 import Foundation
 
-var swiftSyntaxUnsafeSwiftFlags: [String] = []
-
 /// If we are in a controlled CI environment, we can use internal compiler flags
 /// to speed up the build or improve it.
-if ProcessInfo.processInfo.environment["SWIFT_BUILD_SCRIPT_ENVIRONMENT"] != nil {
+let swiftSyntaxSwiftSettings:[SwiftSetting] 
+if case _? = ProcessInfo.processInfo.environment["SWIFT_BUILD_SCRIPT_ENVIRONMENT"] {
```
ahoppen, line 9: Is there a reason why you don’t want to use `!= nil`? IMO it’s easier to read than the pattern matching syntax. If there is a reason I would be curious to know it.
```suggestion
if ProcessInfo.processInfo.environment["SWIFT_BUILD_SCRIPT_ENVIRONMENT"] != nil {
```

ahoppen (commented):
Thanks for continuing to drive this. I’ve got two more stylistic comments inline.

Package.swift
kelvin13: it’s clearer about the intent, since it doesn’t call into the `Equatable` implementation of `Optional<T>`. in this case, we are unwrapping the optional, but don’t care about the result

ahoppen: Hmm, I see where you’re coming from but still think that the benefits of `!= nil` (I can visually parse it in <1s whereas my brain takes a couple of seconds to understand what `case _? =` is doing) outweigh the benefits not calling into the `Equatable` implementation of `Optional` (which are very, very low; if that is broken I assume almost nothing will work in Swift). 
`!=` also matches how we check for optionality of `Optional` in the rest of the project. Could you change it?

kelvin13: done

ahoppen (approved):
Thanks. Looks good to me now.

ahoppen:
@swift-ci Please test

kelvin13:
is this ready to merge?

ahoppen:
Sorry, forgot to hit the merge button. Thanks for the ping.

PR #362, opened 2022-02-18
Add `SourceFile` tests

kimdv:
Also adding a simple convenient initialiser for `ExpressibleAsAccessPathComponent`

### DIFF
--- a/Sources/SwiftSyntaxBuilder/ClassDeclConvenienceInitializers.swift
+++ b/Sources/SwiftSyntaxBuilder/ClassDeclConvenienceInitializers.swift
@@ -1,0 +1,38 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2022 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
+extension ClassDecl {
+  /// A convenience initializer that allows passing in members using a result builder instead of having to wrap them in a `MemberDeclBlock`.
+  public init(
+    classOrActorKeyword: TokenSyntax,
+    identifier: String,
+    genericParameterClause: ExpressibleAsGenericParameterClause? = nil,
+    inheritanceClause: ExpressibleAsTypeInheritanceClause? = nil,
+    genericWhereClause: ExpressibleAsGenericWhereClause? = nil,
+    @AttributeListBuilder attributesBuilder: () -> ExpressibleAsAttributeList? = { nil },
+    @ModifierListBuilder modifiersBuilder: () -> ExpressibleAsModifierList? = { nil },
+    @MemberDeclListBuilder membersBuilder: () -> ExpressibleAsMemberDeclList = { MemberDeclList([]) }
+  ) {
+    self.init(
+      attributes: attributesBuilder(),
+      modifiers: modifiersBuilder(),
+      classOrActorKeyword: classOrActorKeyword,
+      identifier: TokenSyntax.identifier(identifier),
+      genericParameterClause: genericParameterClause,
+      inheritanceClause: inheritanceClause,
+      genericWhereClause: genericWhereClause,
+      members: membersBuilder()
+    )
+  }
+}
--- a/Sources/SwiftSyntaxBuilder/FunctionDeclConvenienceInitializers.swift
+++ b/Sources/SwiftSyntaxBuilder/FunctionDeclConvenienceInitializers.swift
@@ -1,0 +1,38 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2022 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
+extension FunctionDecl {
+  /// A convenience initializer that allows passing in members using a result builder instead of having to wrap them in a `MemberDeclBlock`.
+  public init(
+    funcKeyword: TokenSyntax = TokenSyntax.`func`,
+    identifier: TokenSyntax,
+    genericParameterClause: ExpressibleAsGenericParameterClause? = nil,
+    signature: ExpressibleAsFunctionSignature,
+    genericWhereClause: ExpressibleAsGenericWhereClause? = nil,
+    @AttributeListBuilder attributesBuilder: () -> ExpressibleAsAttributeList? = { nil },
+    @ModifierListBuilder modifiersBuilder: () -> ExpressibleAsModifierList? = { nil },
+    @CodeBlockItemListBuilder bodyBuilder: () -> CodeBlockItemList? = { nil }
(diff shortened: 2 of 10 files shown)


kimdv:
@swift-ci Please test

Tests/SwiftSyntaxBuilderTest/SourceFileTests.swift
```diff
@@ -0,0 +1,29 @@
...
+import SwiftSyntaxBuilder
+
+final class SourceFileTests: XCTestCase {
+  func testSourceFile() {
+    let source = SourceFile(eofToken: .eof, statementsBuilder: {
+      ImportDecl(path: "Foundation")
+      ImportDecl(path: "UIKit")
+      ClassDecl(classOrActorKeyword: .class, identifier: "SomeViewController", members: MemberDeclBlock(membersBuilder: {
+        VariableDecl(.let, name: "tableView", type: "UITableView")
+      }))
```
ahoppen, line 12: I think it would be great if this would look like
```suggestion
      ClassDecl(classOrActorKeyword: .class, identifier: "SomeViewController") {
        VariableDecl(.let, name: "tableView", type: "UITableView")
      }
```

What do you think of adding a convenience initializer to `ClassDecl` (probably also for `StructDecl`, `ProtocolDecl` etc.) that takes a `MemberDeclListBuilder`? I think we’d need to hand-craft that convenience initializer because I can’t think of a way to teach our code-generation to generate result builder parameters based on `ExpressibleAs` conformances.

Tests/SwiftSyntaxBuilderTest/SourceFileTests.swift
```diff
@@ -0,0 +1,29 @@
...
+    var text = ""
+    syntax.write(to: &text)
+
+    XCTAssertEqual(text, """
+
+      import UIKit
+      import Foundation
+      class SomeViewController{
+          let view: UIView
+      }
```
ahoppen, line 26: Should probably be
```suggestion

      import Foundation
      import UIKit
      class SomeViewController{
          let view: UITableView
      }
```

Tests/SwiftSyntaxBuilderTest/SourceFileTests.swift
```diff
@@ -0,0 +1,29 @@
+import XCTest
+import SwiftSyntax
+import SwiftSyntaxBuilder
+
+final class SourceFileTests: XCTestCase {
+  func testSourceFile() {
+    let source = SourceFile(eofToken: .eof, statementsBuilder: {
```
ahoppen, line 7: I think it would be good to hand-craft a 
```swift
public init(@CodeBlockItemListBuilder statements: () -> ExpressibleAsCodeBlockItemList)
```
initializer for `SourceFile`. That way this line could be
```suggestion
    let source = SourceFile {
```

Sources/SwiftSyntaxBuilder/StringConvenienceInitializers.swift
```diff
@@ -46,6 +46,12 @@ extension String: ExpressibleAsReturnClause {
   }
 }
 
+extension String: ExpressibleAsAccessPathComponent {
```
ahoppen, line 49: Do we need this extensions for the test case you add in the PR?

kimdv: Yes, otherwise we need to do `ImportDecl(path: AccessPathComponent(name: "Foundation"))` etc. 

Can add this extension in another PR to separate things.

Tests/SwiftSyntaxBuilderTest/ProtocolDeclTests.swift
```diff
@@ -11,9 +11,10 @@ final class ProtocolDeclTests: XCTestCase {
     })
     let functionSignature = FunctionSignature(input: input, output: returnType)
 
-    // FIXME: We need to add the `modifiersBuilder` with a non-empty value, otherwise will the builder omit newline.
-    let functionDecl = FunctionDecl(identifier: .identifier("buildDeclList"), signature: functionSignature, modifiersBuilder: { TokenSyntax.public })
-    let buildable = ProtocolDecl(modifiers: TokenSyntax.public, identifier: .identifier("DeclListBuildable"), members: functionDecl)
+    let buildable = ProtocolDecl(identifier: "DeclListBuildable", attributesBuilder: { TokenSyntax.public }, membersBuilder: {
```
kimdv, line 14: We cannot remove the parameter `membersBuilder`. Otherwise we get the warning `Backward matching of the unlabeled trailing closure is deprecated; label the argument with 'membersBuilder' to suppress this warning`

Tests/SwiftSyntaxBuilderTest/SourceFileTests.swift
kimdv: We need the last parameter `membersBuilder` otherwise we get `Backward matching of the unlabeled trailing closure is deprecated; label the argument with 'membersBuilder' to suppress this warning`

kimdv: I have added multiple extensions where there was no builder.

Sources/SwiftSyntaxBuilder/FunctionDeclConvenienceInitializers.swift
```diff
@@ -0,0 +1,41 @@
...
+  /// A convenience initializer that allows:
+  ///  - Initializing syntax collections using result builders
+  ///  - Initializing tokens without default text using strings
+  public init(
+    funcKeyword: TokenSyntax = TokenSyntax.`func`,
+    identifier: TokenSyntax,
+    genericParameterClause: ExpressibleAsGenericParameterClause? = nil,
+    signature: ExpressibleAsFunctionSignature,
+    genericWhereClause: ExpressibleAsGenericWhereClause? = nil,
+
```
ahoppen, line 25: Superfluous newline?

Sources/SwiftSyntaxBuilder/ClassDeclConvenienceInitializers.swift
```diff
@@ -0,0 +1,40 @@
...
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
+extension ClassDecl {
+  /// A convenience initializer that allows:
+  ///  - Initializing syntax collections using result builders
+  ///  - Initializing tokens without default text using strings
```
ahoppen, line 18: The comment is probably still from the gyb-generated version. I think a better comment would be 
```suggestion
  /// A convenience initializer that allows passing in members using a result builder instead of having to wrap them in a `MemberDeclBlock`.
```

Similar for the other convenience initializers.

Sources/SwiftSyntaxBuilder/SourceFileConvenienceInitializers.swift
```diff
@@ -0,0 +1,25 @@
...
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+extension SourceFile {
+    /// A convenience initializer that allows:
+    ///  - Initializing syntax collections using result builders
+    ///  - Initializing tokens without default text using strings
+    public init(
+      @CodeBlockItemListBuilder statementsBuilder: () -> ExpressibleAsCodeBlockItemList = { CodeBlockItemList([]) }
```
ahoppen, line 18: I don’t think having a default value here really makes sense. You very rarely want to generate an empty source file 😉

Tests/SwiftSyntaxBuilderTest/ProtocolDeclTests.swift
ahoppen: I think that’s reasonable 👍

ahoppen (approved):
Some minor suggestions around comments. Feel free to merge after addressing them.

kimdv:
@swift-ci Please test

PR #363, opened 2022-02-18
Update readme

kimdv:
I was think if we should  `SwiftSyntax` specific info from root readme into a new readme inside SwiftSyntax source, just like SwiftSyntaxBuilder.
The same for SwiftSyntaxParser.

Then keep the root readme to things that is common for all sources? 
Like how to get started developing, report bugs, license etc?


Sources/SwiftSyntaxBuilder/README.md
```diff
@@ -1,30 +1,64 @@
...
+  ],
+  targets: [
+    .target(name: "MyTool", dependencies: ["SwiftSyntaxBuilder"]),
+  ]
+)
+```
+
+Replace `<#Specify Release tag#>` by the version of SwiftSyntaxBuilder that you want to use. see the following [table](https://github.com/apple/swift-syntax#declare-swiftpm-dependency-with-release-tag) for mapping details.
+
+Then, import `SwiftSyntaxBuilder` in your Swift code.
```
ahoppen, line 56: I think it would be good to also say something about code formatting, like the following

Note that SwiftSyntaxBuilder has limited support for pretty-printing code. We generally suggest running a formatter like [swift-format](https://github.com/apple/swift-format) on the generated code.

README.md
```diff
@@ -13,10 +13,11 @@ Its API is designed for performance-critical applications. It uses value types a
 ## Usage
 
 ### Declare SwiftPM dependency with release tag
+
 Add this repository to the `Package.swift` manifest of your project:
 
 ```swift
-// swift-tools-version:5.2
+// swift-tools-version:5.3
```
ahoppen, line 20: Is there a particular reason to update the swift-tools-version? Doesn’t 5.2 work anymore?

ahoppen: Haha, of course. I was even the one who updated that package version. Thanks for catching it 👍

Sources/SwiftSyntaxBuilder/README.md
```diff
@@ -1,30 +1,70 @@
...
+
+Replace `<#Specify Release tag#>` by the version of SwiftSyntaxBuilder that you want to use. See this following [table](https://github.com/apple/swift-syntax#declare-swiftpm-dependency-with-release-tag) for mapping details.
+
+Then, import `SwiftSyntaxBuilder` in your Swift code.
+
+Note that SwiftSyntaxBuilder has limited support for pretty-printing code. We generally suggest running a formatter like [swift-format](https://github.com/apple/swift-format) on the generated code.
+
+### Reporting Issues
+
+If you should hit any issues while using SwiftSyntaxBuilder, we appreciate bug reports on [bugs.swift.org](https://bugs.swift.org) in the [SwiftSyntax component](https://bugs.swift.org/issues/?jql=component%20%3D%20SwiftSyntax).
```
kimdv, line 66: I think we should make a reference to the README in the root.

I have some other ideas in this comment https://github.com/apple/swift-syntax/pull/363#issue-1143085810

ahoppen:
> I was think if we should `SwiftSyntax` specific info from root readme into a new readme inside SwiftSyntax source, just like SwiftSyntaxBuilder. The same for SwiftSyntaxParser.
> 
> Then keep the root readme to things that is common for all sources? Like how to get started developing, report bugs, license etc?

I would actually prefer to go the other way because I think the readme in `Sources/SwiftSyntaxBuilder` is not very discoverable and I think the three modules should always be considered together since you don’t have to make a choice and use either one or the other. But then the top-level readme is already getting quite large. What I’m thinking might be beneficial would be to create top-level “Documentation” and “Examples” folder that contains all the documentation that’s not necessary to get started. 

The Documentation folder would include documents to 
- Declare SwiftPM dependency with nightly build
- Embedding SwiftSyntax in an Application (includes Embedding in an iOS Application)
- Contributing

The examples folder would contain 
- A markdown document with example users
- `.swift` files with for the examples we currently show (`AddOneToIntegerLiterals` and your example to use SwiftSyntaxBuilder)
  - In CI we could also make sure that these examples actually compile. I just noticed that the example in the main `README` doesn’t even compile because it doesn’t `import SwiftSyntaxParser` (as a follow-up change)

The main `README.md` would then include links to the documents in the Documentation and Examples folder. What do you think?

Sources/SwiftSyntaxBuilder/README.md
```diff
@@ -1,30 +1,70 @@
...
+**Swift Syntax Builder** is an open-source package for generating Swift code in a declarative way.
 
-## Example Usage
+> Note: SwiftSyntaxBuilder is still in development, and the API is not guaranteed to
+> be stable. It's subject to change without warning.
 
 ```swift
-let sourceFile = SourceFile {
-  Import("SwiftSyntax")
+import SwiftSyntaxParser
```
ahoppen, line 9: Wouldn’t this need to be 
```suggestion
import SwiftSyntaxBuilder
```

kimdv:
> I would actually prefer to go the other way because I think the readme in `Sources/SwiftSyntaxBuilder` is not very discoverable and I think the three modules should always be considered together since you don’t have to make a choice and use either one or the other. But then the top-level readme is already getting quite large. What I’m thinking might be beneficial would be to create top-level “Documentation” and “Examples” folder that contains all the documentation that’s not necessary to get started.

Agree on that it can be hard to find and a single readme will become to large and messy. 

What about using DocC?
I have seen the documentation from DocC self is quite good. We have the possibility to make guides, how to get started, examples etc.
They have recently added support for static hosted sites: https://github.com/apple/swift-docc/pull/56

> The Documentation folder would include documents to
> 
> * Declare SwiftPM dependency with nightly build
> * Embedding SwiftSyntax in an Application (includes Embedding in an iOS Application)
> * Contributing

I think that contributing should be in the main readme as for me it is more "repository" related documentation on how to contribute, where SwiftPM and SwiftSyntax embedding is more documentation for how to use it. 
Can you follow me?

> The examples folder would contain
> 
> * A markdown document with example users
> * `.swift` files with for the examples we currently show (`AddOneToIntegerLiterals` and your example to use SwiftSyntaxBuilder)
>   
>   * In CI we could also make sure that these examples actually compile. I just noticed that the example in the main `README` doesn’t even compile because it doesn’t `import SwiftSyntaxParser` (as a follow-up change)

It would be pretty cool if we could make the CI actually test/compile the examples. 💪

> The main `README.md` would then include links to the documents in the Documentation and Examples folder. What do you think?

And then I think the contribution, license, issues and so on is added in the main readme.

ahoppen:
> What about using DocC? I have seen the documentation from DocC itself. It looks really good. We have the possibility to make guides, how to get started, examples etc. They have recently added support for static hosted sites: [apple/swift-docc#56](https://github.com/apple/swift-docc/pull/56)

Exploring to use DocC would definitely be interesting but there’s a non-trivial deployment story we need to solve. We always want the latest documentation for the current main branch. But that means that we need to do one of the following
- You need to manually rebuild the documentation when you make any changes
  - I think this is not practical
- We need to have a CI job that generates and pushes the documentation to a different repository
  - In that case we wouldn’t have any documentation inside this repository itself, which I don’t really like but maybe having a basic README + reference to the DocC-generated documentation would be alright
  - Creating a new repository + the necessary CI job is not something we can set up in the short-term
- We require users who wish to view the documentation to download the repository and open the documentation themselves
  - I think this raises the entry barrier of figuring out what SwiftSyntax too much because cloning/downloading the repo just to view the basic documentation seems like a fairly big commitment to me

I might have missed an option here because I’m not familiar with DocC but I would suggest that for now we stick with markdown files. 

As a separate task, we could consider improving the doc comments of SwiftSyntax to generate better documentation which adopters can build themselves by invoking `docc` locally.

> I think that contributing should be in the main readme as for me it is more "repository" related documentation on how to contribute, where SwiftPM and SwiftSyntax embedding is more documentation for how to use it. Can you follow me?

I disagree here. This repo is the main entry point to whoever wants to use SwiftSyntax as a Swift package dependency and I expect those to be the vast majority compared to those who contribute to the repository. So I think the documentation should primarily target adopters of SwiftSyntax. Contributors are already slightly more committed and I think we can expect them to click a link to view the “Contributing” documentation.

> And then I think the contribution, license, issues and so on is added in the main readme.

I’m not entirely sure if I understand what you mean here but my idea is that the main readme is fairly lightweight. IMO “Contributing” and “License” should just be links. If the documentation to write bug reports it might fit into the main readme.

CONTRIBUTING.md
```diff
@@ -0,0 +1,55 @@
+# Contributing
```
kimdv, line 1: I have added this file by following this GitHub guide: https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors#adding-a-contributing-file

README.md
```diff
@@ -1,22 +1,13 @@
...
-SwiftSyntax is a set of Swift bindings for the
-[libSyntax](https://github.com/apple/swift/tree/main/lib/Syntax) library. It
-allows Swift tools to parse, inspect, generate, and transform Swift source
-code.
-
-Its API is designed for performance-critical applications. It uses value types almost exclusively and aims to avoid existential conversions where possible.
-
-> Note: SwiftSyntax is still in development, and the API is not guaranteed to
-> be stable. It's subject to change without warning.
-
```
ahoppen, line 12: I think we should continue to include this in the main README because it gives a high-level overview of what the library does (or at least the best that we currently have).

Documentation/SwiftSyntax/README.md
```diff
@@ -0,0 +1,41 @@
...
+[libSyntax](https://github.com/apple/swift/tree/main/lib/Syntax) library. It
+allows Swift tools to parse, inspect, generate, and transform Swift source
+code.
+
+Its API is designed for performance-critical applications. It uses value types almost exclusively and aims to avoid existential conversions where possible.
+
+> Note: SwiftSyntax is still in development, and the API is not guaranteed to
+> be stable. It's subject to change without warning.
+
+## Some Example Users
```
ahoppen, line 13: I think we could put those example users in the Examples folder.

CONTRIBUTING.md
ahoppen: Thanks. That’s great.

Documentation/SwiftSyntax/README.md
kimdv: I have moved those, and called I `Some Examples Usages` (I think the Users is a typo)

README.md
kimdv: Moved this part, and the part above back to the main readme.

README.md
```diff
@@ -32,217 +33,27 @@ let package = Package(
...
 |:-------------------:|:-------------------:|:-------------------------:|
 | Xcode 13.0   | swift-5.5-RELEASE   | 0.50500.0 |
 | Xcode 12.5   | swift-5.4-RELEASE   | 0.50400.0 |
 | Xcode 12.0   | swift-5.3-RELEASE   | 0.50300.0 |
 | Xcode 11.4   | swift-5.2-RELEASE   | 0.50200.0 |
 
-
 Then, import `SwiftSyntax` in your Swift code.
 
+### Documentation
```
ahoppen, line 45: This should probably be
```suggestion
## Documentation
```
to match the other headings.

Documentation/README.md
```diff
@@ -0,0 +1,58 @@
+# Documentation
+
+## Usage
```
ahoppen, line 3: I don’t really think we need this heading. We can just remove it and make all the following headings `##` instead of `###`.

Examples/SwiftSyntax.md
```diff
@@ -0,0 +1,52 @@
+# Swift Syntax Examples
```
ahoppen, line 1: This just duplicates the code that’s in `AddOneToIntegerLiteral.swift`.

I think we can just include the non-code parts as comments in `AddOneToIntegerLiteral.swift` and delete this file.

Examples/SwiftSyntaxBuilder.md
```diff
@@ -0,0 +1,30 @@
+# Swift Syntax Builder Examples
```
ahoppen, line 1: Same here as for SwiftSyntax.md. I’d include the non-code parts as comments in the .swift file.

Examples/README.md
```diff
@@ -0,0 +1,59 @@
+# Examples
+
```
ahoppen, line 2: To give some visibility to the examples in the repo, I think it would be good to include them in this README. Something like

```
## Examples

- Command line tool to add one to every integer literal in a source file (AddOneToIntegerLiterals.swift)
- Code-generate a simple source file using SwiftSyntaxBuilder (SourceFile.swift)
```

And adding links to the example source files would also be good IMO.

Examples/SourceFile.swift
```diff
@@ -0,0 +1,16 @@
+import SwiftSyntaxBuilder
```
ahoppen, line 1: A better name for this file would probably be CodeGenerationUsingSwiftSyntaxBuilder.swift.

Documentation/SwiftSyntaxBuilder.md
```diff
@@ -0,0 +1,5 @@
+# Swift Syntax Builder
```
ahoppen, line 1: I’m not sure if this is really providing any value or if we should just remove it.

README.md
```diff
@@ -32,217 +33,27 @@ let package = Package(
...
-[**Pecker**](https://github.com/woshiccm/Pecker): a tool to detect unused code based on [SwiftSyntax](https://github.com/apple/swift-syntax.git) and [IndexStoreDB](https://github.com/apple/indexstore-db.git).
-
-[**Piranha**](https://github.com/uber/piranha): a tool for refactoring code related to feature flags.
-
-[**STAR**](https://github.com/thumbtack/star): a tool to find how often specified Swift type(s) are used in a project.
-
-### Reporting Issues
-
-If you should hit any issues while using SwiftSyntax, we appreciate bug reports on [bugs.swift.org](https://bugs.swift.org) in the [SwiftSyntax component](https://bugs.swift.org/issues/?jql=component%20%3D%20SwiftSyntax).
+Documentation can be found [here](https://github.com/apple/swift-syntax/blob/main/Documentation/README.md).
```
ahoppen, line 47: Maybe also reference the examples:
```suggestion
Documentation can be found [here](https://github.com/apple/swift-syntax/blob/main/Documentation) and some examples of using SwiftSyntax can be found [here](https://github.com/apple/swift-syntax/blob/main/Examples).
```
I would also reference the documentation directory instead of the REAMDE. It still renders the README quite prominently but you also have access to the other files in the directory that way.

ahoppen (approved):

ahoppen:
@swift-ci Please test

PR #365, opened 2022-02-24
Fix leading trivia is not applied if raw syntax is empty

kimdv:
Not sure if we should create a test class for `RawSyntax` where we can add the test case, or it is enough with the current ones

### DIFF
--- a/Sources/SwiftSyntax/RawSyntax.swift
+++ b/Sources/SwiftSyntax/RawSyntax.swift
@@ -1113,7 +1113,11 @@ final class RawSyntax: ManagedBuffer<RawSyntaxBase, RawSyntaxDataElement> {
     }
   }
 
-  func withLeadingTrivia(_ leadingTrivia: Trivia) -> RawSyntax {
+  /// Replaces the leading trivia of the first token in this syntax tree by `leadingTrivia`.
+  /// If the syntax tree did not contain a token and thus no trivia could be attached to it, `nil` is returned.
+  /// - Parameters:
+  ///   - leadingTrivia: The trivia to attach.
+  func withLeadingTrivia(_ leadingTrivia: Trivia) -> RawSyntax? {
     if isToken {
       return RawSyntax.createAndCalcLength(
         kind: formTokenKind()!,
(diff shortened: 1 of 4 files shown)


kimdv:
@swift-ci Please test

Sources/SwiftSyntax/SyntaxData.swift
```diff
@@ -356,10 +356,18 @@ struct SyntaxData {
   }
 
   func withLeadingTrivia(_ leadingTrivia: Trivia) -> SyntaxData {
-    return replacingSelf(raw.withLeadingTrivia(leadingTrivia))
+    if let raw = raw.withLeadingTrivia(leadingTrivia) {
+      return replacingSelf(raw)
+    } else {
+      return replacingSelf(raw)
```
ahoppen, line 362: I think in this case calling `replacingSelf(raw)` is a no-op so we can just `return self` here. Same below.

Sources/SwiftSyntax/RawSyntax.swift
```diff
@@ -1113,7 +1113,7 @@ final class RawSyntax: ManagedBuffer<RawSyntaxBase, RawSyntaxDataElement> {
     }
   }
 
-  func withLeadingTrivia(_ leadingTrivia: Trivia) -> RawSyntax {
+  func withLeadingTrivia(_ leadingTrivia: Trivia) -> RawSyntax? {
```
ahoppen, line 1116: Could you add documentation here what returning `nil` means? I’d suggest
```swift
/// Replaces the leading trivia of the first token in this syntax tree by `leadingTrivia`.
/// If the syntax tree did not contain a token and thus no trivia could be attached to it, `nil` is returned.
```

Similar below.

Tests/SwiftSyntaxBuilderTest/FunctionTests.swift
```diff
@@ -52,7 +52,7 @@ final class FunctionTests: XCTestCase {
     let syntax = buildable.buildSyntax(format: Format(), leadingTrivia: leadingTrivia)
 
     XCTAssertEqual(syntax.description, """
-      func fibonacci(_ n: Int)-> Int{
+      ␣func fibonacci(_ n: Int)-> Int{
```
ahoppen, line 55: Haha, good catch 👍

ahoppen (commented):
Thanks. Looks good to me. I think the current test coverage is fine. AFAICT we don’t have any tests at the `RawSyntax` level at the moment.

Sources/SwiftSyntax/RawSyntax.swift
kimdv: Added! 
Also added parameter documentation

kimdv:
@swift-ci Please test

ahoppen (approved):

PR #371, opened 2022-03-15
Explicitly add rpath for lib_InternalSwiftSyntaxParser.dylib

ahoppen:
Previously, we were relying on SwiftPM adding the toolchain’s stdlib to the rpaths and we were finding `lib_InternalSwiftSyntaxParser.dylib` inside the stdlib directory. This behavior was removed in apple/swift-package-manager#4208.

Explicitly add the directory that contains `lib_InternalSwiftSyntaxParser.dylib` as an rpath to SwiftSyntaxParser when compiling SwiftSyntax using build-script.py.

The major downsides of this approach are
- You can no longer run SwiftSyntaxParser tests for SwiftSyntax in Xcode with an open source toolchain
  - Maybe this isn’t too bad because it only affects SwiftSyntaxParser tests and those can be run from the command line
- All adopters of SwiftSyntax also need to explicitly specify the `lib_InternalSwiftSyntaxParser.dylib` rpath
  - Maybe this isn’t too bad because on Linux `lib_InternalSwiftSyntaxParser.dylib` is in the standard search paths and on macOS we start shipping `lib_InternalSwiftSyntaxParser.dylib` as a binary dependency.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -5,7 +5,7 @@ import Foundation
 
 /// If we are in a controlled CI environment, we can use internal compiler flags
 /// to speed up the build or improve it.
-let swiftSyntaxSwiftSettings: [SwiftSetting] 
+let swiftSyntaxSwiftSettings: [SwiftSetting]
 if ProcessInfo.processInfo.environment["SWIFT_BUILD_SCRIPT_ENVIRONMENT"] != nil {
   let groupFile = URL(fileURLWithPath: #file)
     .deletingLastPathComponent()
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

ktoso (approved):
Thanks for the quick turnaround here 👍

ahoppen:
https://github.com/apple/swift-stress-tester/pull/175

@swift-ci Please test

dduan:
This change made the package un-buildable for users with the following errors

> the target 'SwiftSyntax' in product 'SwiftSyntax' contains unsafe build flags

Test manifest

```swift
// swift-tools-version:5.5

import PackageDescription

let package = Package(
    name: "SwiftSyntaxProblem",
    products: [
        .executable(
            name: "SwiftSyntaxProblem",
            targets: ["SwiftSyntaxProblem"]
        ),
    ],
    dependencies: [
        .package(
            name: "SwiftSyntax",
            url: "https://github.com/apple/swift-syntax.git",
            .exact("0.50600.0")
        ),
    ],
    targets: [
        .executableTarget(
            name: "SwiftSyntaxProblem",
            dependencies: ["SwiftSyntax"]),
    ]
)
```

```
swift-driver version: 1.26.21 Apple Swift version 5.5.2 (swiftlang-1300.0.47.5 clang-1300.0.29.30)
Target: arm64-apple-macosx12.0
```

dduan:
https://bugs.swift.org/browse/SR-15989

PR #374, opened 2022-03-17
Correct the Syntax of Primary Associated Types

CodaFi:
The prior syntax tree did not take into account that the clause itself should own the angle brackets.


CodaFi:
https://github.com/apple/swift/pull/41848
@swift-ci Please test

PR #378, opened 2022-03-30
Migrate to use Python 3

dduan:
Migrate Python code to Python 3 by:

1. update scripts' shebang
2. migrate syntax in Python modules
3. delete unnecessary imports

PR #379, opened 2022-04-06
Adding SwiftSyntaxBuilder to the build process

vrjuliao:
The build process of SwiftSyntax, should build three shared objects:
* SwiftSyntax
* SwiftSyntaxParser
* SwiftSyntaxBuilder

The SwiftSyntaxBuilder one has not been called to be built, then this pull request adds such build call to the `build-script.py` file. Also, there are two points in the build process that was producing erros during the SwiftSyntaxBuilder build.
1. Python3 strings does not support `string_escape` encode. However, python2 strings supports encoding for `utf-8` and decoding as `unicode_escape`.\
I'm not sure about differences between `string.encode('utf-8').decode('unicode_escape')` and `string.decode('string_escape')` in Python2, but looking for the data that will be used to fill `Tokens.swift.gyb` we only have UTF-8 characters, and for that case these approaches are equivalent. See the values in the file [swift/utils/gyb_syntax_support/Token.py](https://github.com/apple/swift/blob/main/utils/gyb_syntax_support/Token.py#L144).
2. The local module import in [SyntaxBuildableWrappers.py](Sources/SwiftSyntaxBuilder/gyb_helpers/SyntaxBuildableWrappers.py) and [\_\_init\_\_.py](Sources/SwiftSyntaxBuilder/gyb_helpers/__init__.py) is failing. The dot must to be added to tell to pyhton3 interpreter that we are dealing with the local module (actually, the same directory).

### DIFF
--- a/Sources/SwiftSyntax/TokenKind.swift.gyb
+++ b/Sources/SwiftSyntax/TokenKind.swift.gyb
@@ -76,7 +76,7 @@ public enum TokenKind {
     case .eof: return .zero
 % for token in SYNTAX_TOKENS:
 %   if token.text:
-    case .${token.swift_kind()}: return SourceLength(utf8Length: ${len(token.text.decode('string_escape'))})
+    case .${token.swift_kind()}: return SourceLength(utf8Length: ${len(token.text.encode('utf-8').decode('unicode_escape'))})
 %   else:
     case .${token.swift_kind()}(let text): return SourceLength(of: text)
 %   end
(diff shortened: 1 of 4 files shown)


vrjuliao:
HI @ahoppen!
I didn't see the pr #378 . I'm sorry about that. I will get this modifications back, and only keep the SwiftSyntaxBuilder one.

ahoppen:
No worries. Would you mind squashing your changes into one commit and force pushing. Having five commits do and undo some changes just looks noisy in the git history.

kimdv:
This have been fixed in #381 🎉

kimdv:
@ahoppen think we just can close this one. 
https://github.com/apple/swift-syntax/pull/379#issuecomment-1140064774

ahoppen:
Yes, you’re right 👍

PR #381, opened 2022-04-07
Use SwiftSyntaxBuilder to generate SwiftSyntaxBuilder Tokens

kimdv:


Sources/SwiftSyntaxBuilderGeneration/SwiftSyntaxBuilderGeneration.swift
```diff
@@ -0,0 +1,138 @@
...
+    let generatedPath = URL(fileURLWithPath: #file)
+      .deletingLastPathComponent()
+      .deletingLastPathComponent()
+      .appendingPathComponent("SwiftSyntaxBuilder")
+      .appendingPathComponent("gyb_generated")
+
+    try FileManager.default.createDirectory(atPath: generatedPath.path, withIntermediateDirectories: true, attributes: nil)
+
+    let filePath = generatedPath.appendingPathComponent("Tokens.swift")
+    try tokenSyntax.description.write(to: filePath, atomically: true, encoding: .utf8)
```
kimdv, line 122: Still WIP. 
We should abstract this out, so it can be used multiple places.

Sources/SwiftSyntaxBuilder/gyb_generated/Tokens.swift
```diff
@@ -1,723 +1,693 @@
...
+static var `associatedtype`: TokenSyntax{
     SyntaxFactory.makeAssociatedtypeKeyword()
     .withTrailingTrivia(.spaces(1))
   }
-
-  /// The `class` keyword
-  static var `class`: TokenSyntax {
+  
+/// The `class` keyword
+static var `class`: TokenSyntax{
```
kimdv, line 14: This is an bug, where it don't respect the inject from the formatter.

Sources/SwiftSyntaxBuilder/gyb_generated/Tokens.swift
```diff
@@ -1,723 +1,693 @@
...
-//
-// This source file is part of the Swift.org open source project
-//
-// Copyright (c) 2014 - 2022 Apple Inc. and the Swift project authors
-// Licensed under Apache License v2.0 with Runtime Library Exception
-//
-// See https://swift.org/LICENSE.txt for license information
-// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
-//
-//===----------------------------------------------------------------------===//
```
kimdv, line 13: Right now there is no nice way to add comments like this. 
We could add it the the `import` token

kimdv:
We also need to be able to run the executable from the CI, so we can ensure everything are up to date.

Sources/SwiftSyntaxBuilder/gyb_generated/Tokens.swift
```diff
@@ -1,723 +1,693 @@
-//// Automatically Generated From Tokens.swift.gyb.
```
kimdv, line 1: Right now it also adds the file to the `gyb_generated` folder. 
I think it's fine to continue with that.

Sources/SwiftSyntaxBuilder/gyb_generated/Tokens.swift
kimdv: The reason is properly because the `withLeadingTrivia()` is overriding and not adding it.

Sources/SwiftSyntaxBuilder/gyb_generated/Tokens.swift
ahoppen: Yeah, I think that’s fine for now.

Sources/SwiftSyntaxBuilder/gyb_generated/Tokens.swift
ahoppen: Fixing the indentation issue in a separate PR would be great. I think we can tackle this separately from this PR though.

Sources/SwiftSyntaxBuilderGeneration/String+Extensions.swift
```diff
@@ -0,0 +1,17 @@
...
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+
+extension StringProtocol {
+    var firstLowercased: String { prefix(1).lowercased() + dropFirst() }
```
ahoppen, line 16: How does this behave on an empty string? Also maybe `withFirstCharacterLowercased` would be clearer.

Sources/SwiftSyntaxBuilderGeneration/Tokens.swift.gyb
```diff
@@ -0,0 +1,136 @@
...
+public class SilKeyword: Keyword { }
+
+public class PoundKeyword: Token {
+  public init(name: String, kind: String, serializationCode: Int, text: String, classification: String = "Keyword") {
+    super.init(name: name, kind: "pound_\(kind)", serializationCode: serializationCode, unprefixedKind: kind, text: text, classification: classification, isKeyword: true, requiresTrailingSpace: true)
+  }
+}
+
+public class PoundObjectLiteral: PoundKeyword {
+  public let `description`: String
```
ahoppen, line 88: I don’t think `description` needs to be in backticks (also below).

Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version:5.3
+// swift-tools-version:5.4
```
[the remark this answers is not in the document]
kimdv: It was to use the `@main` feature. 
Reverted to 5.3 again, and just using a more simple main file.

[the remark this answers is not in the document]
ahoppen: Feel free to update to 5.4, I don’t have any objections whatsoever. I was just curios

Sources/SwiftSyntaxBuilder/gyb_generated/Tokens.swift
kimdv: It did just hit me, that the degyb will remove all files, from the folder if there is not related gyb file. 

So made a new folder `generated`.

Sources/SwiftSyntaxBuilderGeneration/SwiftSyntaxBuilderGeneration.swift
```diff
@@ -0,0 +1,153 @@
...
+  }
+}
+
+fileprivate extension FunctionCallExpr {
+  private static var spacing: FunctionCallExpr {
+    FunctionCallExpr(calledExpression: TokenSyntax.identifier(".spaces"), leftParen: .leftParen, argumentList: TupleExprElement(expression: IntegerLiteralExpr(1)), rightParen: .rightParen)
+  }
+
+  static var leadingTriviaFunction: FunctionCallExpr {
+    FunctionCallExpr.init(calledExpression: TokenSyntax.identifier(".withLeadingTrivia"), leftParen: .leftParen, argumentList: TupleExprElement.init(expression: spacing), rightParen: .rightParen)
```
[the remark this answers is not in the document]
kimdv: Question: 
https://github.com/apple/swift/blob/11dd84e4b00e47396cdbfa1d09f07cb6908b2cfe/utils/gyb_syntax_support/ExprNodes.py#L440

Why is the kind an `Expr` and not and `MemberAccessExpr`?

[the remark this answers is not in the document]
kimdv: And is this also wrong then: https://github.com/apple/swift-syntax/blob/0fdca233a03c8cf039824ad82df0693a3657cc97/Sources/SwiftSyntaxBuilder/FunctionCallExprConvenienceInitializers.swift#L26

build-script.py
```diff
@@ -717,6 +794,15 @@ def main():
                 verbose=args.verbose,
                 add_source_locations=args.add_source_locations,
             )
+
+            run_code_generation(
```
kimdv, line 798: Added code generation here, as I think it makes sense to run this if the gyb have changed.

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test

Sources/SwiftSyntaxBuilder/MemberAccessExprConvenienceInitializers.swift
```diff
@@ -0,0 +1,30 @@
...
+
+import SwiftSyntax
+
+extension MemberAccessExpr {
+    /// Creates a `MemberAccessExpr` using the provided parameters.
+    /// - Parameters:
+    ///   - base:
+    ///   - dot:
+    ///   - name:
+    ///   - declNameArguments:
```
ahoppen, line 21: To keep this PR focussed, do you think we could pull this change out into a separate PR? Also, I think you’re missing descriptions for the parameters. I think we should either remove the per-parameter documentation altogether or provide descriptions.

Sources/SwiftSyntaxBuilderGeneration/Tokens.swift.gyb
```diff
@@ -0,0 +1,135 @@
...
+let SYNTAX_TOKENS: [Token] = [
+% for token in SYNTAX_TOKENS:
+%   class_name = type(token).__name__
+%   text = 'text: "' + token.text + '", ' if token.text else ""
+%   if class_name in ['Keyword', 'SwiftKeyword', 'DeclKeyword', 'StmtKeyword', 'ExprKeyword', 'PatternKeyword', 'SilKeyword']:
+    ${class_name}(name: "${token.name}", serializationCode: ${token.serialization_code}, text: "${token.text}"),
+%   elif class_name in ['PoundKeyword', 'PoundConfig', 'PoundDirectiveKeyword', 'PoundConditionalDirectiveKeyword']:
+    ${class_name}(name: "${token.name}", kind: "${token.kind}", serializationCode: ${token.serialization_code}, text: "${token.text}"),
+%   elif class_name in ['Punctuator', 'Misc']:
+    ${class_name}(name: "${token.name}", kind: "${token.kind}", serializationCode: ${token.serialization_code}, ${text}requiresLeadingSpace: ${str(token.requires_leading_space).lower()}, requiresTrailingSpace: ${str(token.requires_trailing_space).lower()}),
```
ahoppen, line 126: Small nitpick: If we omitted `requiresLeadingSpace` and `requiresTrailingSpace` if their value is `False`, I think the generated code would pretty much 1:1 match what we have in Token.py, which I think would be kind of nice. It’s not really required but would be a nice touch, I think.

Sources/SwiftSyntaxBuilderGeneration/Tokens.swift.gyb
```diff
@@ -0,0 +1,135 @@
...
+%   elif class_name == 'Literal':
+    ${class_name}(name: "${token.name}", kind: "${token.kind}", serializationCode: ${token.serialization_code}),
+%   elif class_name == 'PoundObjectLiteral':
+    ${class_name}(name: "${token.name}", kind: "${token.kind}", serializationCode: ${token.serialization_code}, ${text}description: "${token.description}", protocol: "${token.protocol}"),
+%   else:
+    fatalError("Unknown token `${class_name}`")
```
ahoppen, line 132: I don’t think `fatalError` is the right thing to do here, because you would insert `fatalError("Unknown token…")` into the generated source code. What you want is probably closer to 
```suggestion
%     fatal_error("Unknown token `${class_name}`")
```

with 
```python
def printerr(message):
    print(message, file=sys.stderr)


def fatal_error(message):
    printerr(message)
    sys.exit(1)
```

taken from build-script.py. If this is the only place where we need a Python `fatal_error`, we can also just inline the function bodies here.

Sources/SwiftSyntaxBuilderGeneration/main.swift
```diff
@@ -0,0 +1,199 @@
...
+//===----------------------------------------------------------------------===//
+
+import Foundation
+import SwiftSyntax
+import SwiftSyntaxBuilder
+
+private let format = Format(indentWidth: 2)
+
+private let copyrightHeader = """
+    //// Automatically Generated
```
ahoppen, line 20:
```suggestion
    //// Automatically Generated by SwiftSyntaxBuilderGeneration
```

Just so we know it’s not generated by `.gyb` files and keep time spending looking for them.

Sources/SwiftSyntaxBuilderGeneration/main.swift
```diff
@@ -0,0 +1,199 @@
...
+                       bindingsBuilder: {
+            // We need to use `CodeBlock` here to ensure there is braces around.
+
+            let accessor = CodeBlock {
+              FunctionCallExpr(calledExpression: MemberAccessExpr(base: "SyntaxFactory", name: "make\(token.name)Keyword"),
+                               leftParen: .leftParen,
+                               rightParen: .rightParen)
+
+              if token.requiresLeadingSpace {
+                FunctionCallExpr.leadingTriviaFunction
```
ahoppen, line 61: I can’t really tell why but `FunctionCallExpr.leadingTriviaFunction` doesn’t sound like it generates a `withLeadingTrivia` call to me. I think it’s mostly because I wouldn't expect anything SwiftSyntaxBuilder code generation builder-specific to be declared on `FunctionCallExpr`. I think I’d prefer to have a free function like `createWithLeadingTriviaCall` to be declared at the top of the file.

Sources/SwiftSyntaxBuilderGeneration/main.swift
```diff
@@ -0,0 +1,199 @@
...
+        PatternBinding.createTokenSyntaxPatternBinding("open", accessor: body)
+      })
+    })
+}
+
+let tokenSyntax = tokensFile.buildSyntax(format: format)
+
+let arguments = CommandLine.arguments
+  .filter { $0.starts(with: "--") }
+  .map { $0.split(separator: "=").map { String($0) } }
```
ahoppen, line 165: What do you think about just using the first argument as the destination? That would remove any need for command line argument parsing here. I.e. `let destination = CommandLine.arguments[1]` (plus some checking that we actually have a first argument).

Also, in the case that `destination` is not passed, we should output a little help message with the arguments SwiftSyntaxBuilderGenerat/main.swift expects.

build-script.py
```diff
@@ -376,6 +397,35 @@ def build(self, product_name):
         check_call(command, env=env, verbose=self.verbose)
 
 
+# -----------------------------------------------------------------------------
+# Running
+
+
+def run_code(runner, arguments, toolchain, build_dir, multiroot_data_file, release, verbose):
```
ahoppen, line 404: I think we can just inline this to `run_code_generation` because `run_code_generation` is currently very simple and this is also pretty straight-forward straight-line code.

build-script.py
ahoppen: I don’t think generation of files in SwiftSyntaxGeneration should be an implicit step in `build-script.py` because otherwise you might run into bootstrapping issues. Imagine you are making a breaking change to `SwiftSyntaxBuilder`. What you need to do is:
1. Change `SwiftSyntaxBuilderGeneration` to generate the code with the breaking change
2. Run `SwiftSyntaxBuilderGeneration` to actually generate the `SwiftSyntaxBuilder` code that contains the breaking change.
3. Update `SwiftSyntaxBuilderGeneration` to adopt the breaking change
4. Run step 1 and 2 again to make sure things are compatible
5. Actually build all the real SwiftSyntax products

Step (3) requires manual user intervention so we can’t just perform the other steps automatically in build-script.py.

What I would imagine is that there is a mode for build-script.py that just build + runs SwiftSyntaxBuilderGeneration.

build-script.py
```diff
@@ -757,6 +853,7 @@ def main():
         # targets simultaneously. For now, just build one product after the other.
         builder.build("SwiftSyntax")
         builder.build("SwiftSyntaxParser")
+        builder.build("SwiftSyntaxBuilder")
```
ahoppen, line 856: I think we should also build SwiftSyntaxBuilderGeneration here to make sure it’s actually compatible with the current SwiftSyntaxBuilder, just to be sure.

Sources/SwiftSyntaxBuilderGeneration/main.swift
kimdv: Moved all extensions to the top as a free function

Sources/SwiftSyntaxBuilder/MemberAccessExprConvenienceInitializers.swift
kimdv: Fixed in https://github.com/apple/swift-syntax/pull/450

build-script.py
kimdv: So we should remove `run_code ` and move all the logic into `run_code_generation`?

ahoppen: That’s what I thought

Sources/SwiftSyntaxBuilderGeneration/main.swift
```diff
@@ -0,0 +1,193 @@
...
+        }
+
+        createTokenSyntaxPatternBinding("open", accessor: body)
+      })
+    })
+}
+
+let tokenSyntax = tokensFile.buildSyntax(format: format)
+
+guard !CommandLine.arguments.isEmpty else {
```
ahoppen, line 182: `CommandLine.arguments` is never empty, the first argument is always the executable that’s being called. I think this should be 
```suggestion
guard CommandLine.arguments.count > 1 else {
```

build-script.py
```diff
@@ -740,6 +817,18 @@ def main():
...
         sys.exit(0)
 
+    run_code_generation(
+        toolchain=args.toolchain,
+        build_dir=realpath(args.build_dir),
+        multiroot_data_file=args.multiroot_data_file,
+        release=args.release,
+        verbose=args.verbose,
+        swiftsyntaxbuilder_destination=os.path.join(SWIFTSYNTAXBUILDER_DIR, "generated")
+    )
```
ahoppen, line 827: Sorry if I wasn’t clear in https://github.com/apple/swift-syntax/pull/381#discussion_r876791560. I was saying that we shouldn’t run the `SwiftSyntaxBuilderGeneration` as an implicit step in `build-script.py` because `SwiftSyntaxBuilderGeneration` itself might no longer compile after `SwiftSyntaxBuilder` has been re-generated.

My idea was that we should have a flag `--run-swiftsyntaxbuilder-generation` (or something of the sort) and then have this be like
```suggestion
    if args.run_swiftsyntaxbuilder_generation:
      run_code_generation(
          toolchain=args.toolchain,
          build_dir=realpath(args.build_dir),
          multiroot_data_file=args.multiroot_data_file,
          release=args.release,
          verbose=args.verbose,
          swiftsyntaxbuilder_destination=os.path.join(SWIFTSYNTAXBUILDER_DIR, "generated")
      )
      sys.exit(0)
```

In which case we could also move it up to the top of `main`.

Also: Since `build-script-helper.py` now does so many things, we should probably migrate it to using argparse subcommands. But that’s something for a completely standalone PR.

ahoppen (approved):
Let’s 🚢 it

ahoppen:
@swift-ci Please test

Sources/SwiftSyntaxBuilderGeneration/Tokens.swift.gyb
kimdv: Will fix this in a follow up PR if this gets merged

ahoppen:
Looks like the generated files don’t match. Could you re-generate them?

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test

PR #382, opened 2022-04-11
Support `open` in swiftSyntaxBuilder

iMostfa:
This pull request adds support for the 'open' keyword as TokenSyntax in SwiftSyntaxBuilder
it's implemented as syntactic sugar based on `makeContextualKeyword ` 
and accessed in swift like this: TokenSyntax.public

I'm open to your feedback and suggestation to enhance my pull request 

greetings, Mostfa.

### DIFF
--- a/Sources/SwiftSyntaxBuilder/Tokens.swift.gyb
+++ b/Sources/SwiftSyntaxBuilder/Tokens.swift.gyb
@@ -62,4 +62,9 @@ public extension TokenSyntax {
   static var eof: TokenSyntax {
     SyntaxFactory.makeToken(.eof, presence: .present)
   }
+  /// The `open` contextual token
+  static var open: TokenSyntax {
+        SyntaxFactory.makeContextualKeyword("open")
+          .withTrailingTrivia(.spaces(1))
+    }
 }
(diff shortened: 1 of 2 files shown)


Sources/SwiftSyntaxBuilder/Tokens.swift.gyb
```diff
@@ -62,4 +62,9 @@ public extension TokenSyntax {
   static var eof: TokenSyntax {
     SyntaxFactory.makeToken(.eof, presence: .present)
   }
+  /// The `open` token
```
ahoppen, line 65: One minor nitpick: Could you rephrase this to 
```suggestion
  /// The `open` contextual token
```

It should make it clear why it is not generated using `gyb`.

ahoppen (commented):
Thanks. This looks good to me. Just one minor comment inline.

iMostfa:
Thank you @ahoppen 
i've added your comment in Token.swift.gyb and the generated file

ahoppen (approved):

ahoppen:
Thanks for the PR and welcome to SwiftSyntax development.

PR #384, opened 2022-04-13
Fixed comments in trivia piece

iMostfa:
Hi, 

While working on better API support for comments in SwiftSyntaxBuilder, 

I noticed that all variations of comments in `TriviaPiece` have the same implementation, which just adds regular text to the output of Syntax, which doesn't reflect the naming of a used trivia.

 because the clients using the trivia would need to explicitly add the `///` when using it
`.docLineComment("/// This's a docLine comment")`

```swift
// This PR will let the users just pass the comment needed, without /* or /// ...etc

.docLineComment("This's a docLine comment")    // -> generated output: /// This's a docLine comment
.lineComment("This's a reuglar line comment") // -> generated output:  // This's a reuglar line comment
.blockComment("This's a block comment")      // -> generated output:  /* This's a block comment */


```

and same applies to the rest of the comment types

Please let me what do you think about the changes, and if there's something better we should do.

Thanks,

### DIFF
--- a/Sources/SwiftSyntax/Trivia.swift.gyb
+++ b/Sources/SwiftSyntax/Trivia.swift.gyb
@@ -53,11 +53,18 @@ extension TriviaPiece: TextOutputStreamable {
 %   joined = ''.join(trivia.swift_characters)
     case let .${trivia.lower_name}s(count):
       printRepeated("${joined}", count: count)
-%   else:
-    case let .${trivia.lower_name}(text):
-      target.write(text)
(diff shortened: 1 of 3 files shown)


iMostfa:
ah! my bad, thank you @ahoppen for clarifying this to me.
I will use AST-Explorer before making changes in the future to be sure that behaviors are matched between
`SwiftSyntax`, and `SwiftSyntaxBuilder`.

PR #448, opened 2022-05-23
Add API to enable or disable bare slash regex parsing

ahoppen:
Depends on https://github.com/apple/swift/pull/59018

---

Add a parameter to the `SyntaxParser.parse` functions to enable or disable bare slash regex parsing.

rdar://93750821

### DIFF
--- a/Sources/SwiftSyntaxParser/SyntaxParser.swift
+++ b/Sources/SwiftSyntaxParser/SyntaxParser.swift
@@ -58,6 +58,12 @@ public enum SyntaxParser {
   ///   - source: The source string to parse.
   ///   - parseTransition: Optional mechanism for incremental re-parsing.
   ///   - filenameForDiagnostics: Optional file name used for SourceLocation.
+  ///   - languageVersion: Interpret input according to a specific Swift
+  ///       language version number. If `nil`, this inherits the default from
+  ///       the syntax parser library.
+  ///   - enableBareSlashRegexLiteral: Enable or disable the use of forward
+  ///       slash regular-expression literal syntax. If `nil`, this inherits the
+  ///       default from the syntax parser library.
   ///   - diagnosticHandler: Optional callback that will be called for all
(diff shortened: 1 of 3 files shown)


Sources/SwiftSyntaxParser/SyntaxParser.swift
```diff
@@ -92,33 +104,55 @@ public enum SyntaxParser {
...
   ///       diagnostics the parser emits
   /// - Returns: A top-level Syntax node representing the contents of the tree,
   ///            if the parse was successful.
   /// - Throws: `ParserError`
-  public static func parse(_ url: URL,
-      diagnosticHandler: ((Diagnostic) -> Void)? = nil) throws -> SourceFileSyntax {
+  public static func parse(
+    _ url: URL,
+    languageVersion: String? = nil,
+    enableBareSlashRegexLiteral: Bool? = nil,
```
rintaro, line 121: I *feel* like these should be a `Options` object. But I'm not sure how the API should look like, for example
```swift
// array of enums
SyntaxParse.parse(
    url,
    options: [.languageVersion("5.6"), .bareSlashRegexLiteral(true)],
    diagnosticHandler: diagnosticHandler)
// Just a "options" object
SyntaxParse.parse(
    url,
    options: SyntaxParseOptions(languageVersion: "5.1", enableBareSlashRegexLiteral: true),
    diagnosticHandler: diagnosticHandler)
```
The advantage of "options" object is that you can _easily_ pass around and store the object in the clients, when they want to create options at _not_  exact the call site.

WDYT?

rintaro: Alex and I chatted about this and we decided to accept this as is for now until the "language features" proposal is settled.

rintaro (approved):
Just resolve the conflict :)

ahoppen:
https://github.com/apple/swift/pull/59018

@swift-ci Please test

PR #449, opened 2022-05-23
Show the command that needs to be run when SwiftSyntax files don't match

ahoppen:
This error message will show up if someone modifies the syntax definition files in the main swift repo without modifying the generated SwiftSyntax files accordingly. Tell them what to do in order to fix the problem.

rdar://89777232


ahoppen:
@swift-ci Please test

build-script.py
```diff
@@ -730,7 +730,9 @@ def main():
             printerr(
                 "FAIL: Gyb-generated files committed to repository do "
                 "not match generated ones. Please re-generate the "
-                "gyb-files and recommit them."
+                "gyb-files using the following command, open a PR to the "
+                "SwiftSyntax project and merge it alongside the main PR."
+                "$ swift-syntax/build-script.py --degyb-only --toolchain whatever"
```
bnbarham, line 735: Not sure if there's a brief way to say this, but it could be worth noting that it doesn't need to be a toolchain. Ie. `/path/to/swift-build/` will generally work.

ahoppen: What do you think about the following?
```
swift-syntax/build-script.py --degyb-only --toolchain /dev/null
```

bnbarham (approved):

PR #450, opened 2022-05-24
Add convenience initialiser for `MemberAccessExpr`

kimdv:
From comment: https://github.com/apple/swift-syntax/pull/381#discussion_r876745511

> To keep this PR focussed, do you think we could pull this change out into a separate PR? Also, I think you’re missing descriptions for the parameters. I think we should either remove the per-parameter documentation altogether or provide descriptions.

@ahoppen the parameters are from the generated init method. 
I think the best would be to actually add them in the gyb files, to have documentation.

Removed them here.


kimdv:
@swift-ci please test

ahoppen (approved):

PR #451, opened 2022-05-26
Code Generation example is fixed for 0.50600.1

evnik:
I'm not sure if it is different for the latest `main`, but here are the changes I need to do to make the example code compilable with the `0.50600.1`.
BTW, `0.50500.0` requires other fixes, I think it worths to be mentioned somewhere in the docs

### DIFF
--- a/Examples/CodeGenerationUsingSwiftSyntaxBuilder.swift
+++ b/Examples/CodeGenerationUsingSwiftSyntaxBuilder.swift
@@ -10,12 +10,12 @@ import SwiftSyntaxBuilder
 /// }
 ///```
 
-let source = SourceFile {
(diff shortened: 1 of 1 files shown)


ahoppen:
Thanks for bringing this up, but this is expected. The SwiftSyntaxBuilder syntax has been improved quite a bit since the 0.50600.1 release, so it’s expected that the current examples only compile with `main` and not the 0.50600.1 release. 

If you want to see examples for the Swift 0.50600.1 release, I would suggest that you look at the [source state for that release](https://github.com/apple/swift-syntax/tree/0.50600.1/Sources/SwiftSyntaxBuilder), but I have to admit that we had a lot fewer examples back then.

PR #452, opened 2022-05-29
Improvements in `FunctionCallExpr` convenience initializers

evnik:
* Default values for left/right paren arguments
* Separate initializer accepting `ExpressibleAsMemberAccessExpr`
* Tests and initializer usages are updated/simplified


Sources/SwiftSyntaxBuilder/FunctionCallExprConvenienceInitializers.swift
```diff
@@ -31,4 +31,22 @@ extension FunctionCallExpr {
       additionalTrailingClosures: additionalTrailingClosuresBuilder()
     )
   }
+
+  public init(
```
ahoppen, line 35: I think it would be more general if we changed the `ExpressibleAsIdentifierExpr` above into `ExpressibleByExprBuildable` in the convenience initializer above instead of adding another overload.

ahoppen (commented):
Thanks. This really improves readability of `SwiftSyntaxBuilderGeneration`.

I’ve got one suggestion inline.

kimdv:
@swift-ci please test

PR #454, opened 2022-06-02
Override trailing comma for elements in ResultBuilders

evnik:
Test coverage is added for some of ResultBuilders
I think that should be enough for now and more coverage could be added later in subsequent PRs

### DIFF
--- a/Sources/SwiftSyntaxBuilder/BuildableNodes.swift.gyb
+++ b/Sources/SwiftSyntaxBuilder/BuildableNodes.swift.gyb
@@ -1,6 +1,6 @@
 %{
   from gyb_syntax_support import SYNTAX_NODES
-  from gyb_helpers import SyntaxBuildableNode
+  from gyb_helpers import SyntaxBuildableNode, conformance_clause
   # -*- mode: Swift -*-
   # Ignore the following admonition it applies to the resulting .swift file only
 }%
@@ -23,10 +23,15 @@ import SwiftSyntax
 % for node in [SyntaxBuildableNode(syntax_node) for syntax_node in SYNTAX_NODES if syntax_node.is_buildable()]:
 %   type = node.type()
 %   base_type = node.base_type()
+%   conformances = [base_type.buildable(), type.expressible_as()]
+%   has_with_trailing_comma_trait = node.node.traits and 'WithTrailingComma' in node.node.traits
+%   if has_with_trailing_comma_trait:
+%     conformances.append("HasTrailingComma")
+%   end
 %   if node.documentation():
 /// ${node.documentation()}
 %   end
-public struct ${type.buildable()}: ${base_type.buildable()}, ${type.expressible_as()} {
+public struct ${type.buildable()}${conformance_clause(conformances)} {
 %   children = node.children()
 %   for child in children:
   let ${child.name()}: ${child.type().buildable()}
@@ -114,6 +119,23 @@ public struct ${type.buildable()}: ${base_type.buildable()}, ${type.expressible_
   public func create${type.buildable_base_name()}() -> ${type.buildable()} {
     return self
   }
+%   if has_with_trailing_comma_trait:
+%     trailing_comma_init_args = []
+%     for child in children:
+%       if child.name() == 'trailingComma':
+%         trailing_comma_init_args.append('%s: withComma ? .comma : nil' % (child.name()))
+%       else:
+%         trailing_comma_init_args.append('%s: %s' % (child.name(), child.name()))
+%       end
+%     end
+
+  /// Conformance to `HasTrailingComma`.
+  public func withTrailingComma(_ withComma: Bool) -> Self {
+      return Self.init(
+        ${',\n        '.join(trailing_comma_init_args)}
+      )
+  }
+%   end
 
   /// `${type.buildable()}` might conform to `${base_type.expressible_as()}` via different `ExpressibleAs*` paths.
   /// Thus, there are multiple default implementations for `create${base_type.buildable_base_name()}`, some of which perform conversions through `ExpressibleAs*` protocols.
--- a/Sources/SwiftSyntaxBuilder/ResultBuilders.swift.gyb
+++ b/Sources/SwiftSyntaxBuilder/ResultBuilders.swift.gyb
@@ -85,6 +85,12 @@ public struct ${type.result_builder()} {
   public static func buildFinalResult(_ component: Component) -> FinalResult {
 %   if element_type.is_token():
     return .init(component)
+%   elif element_type.has_with_trailing_comma_trait():
+    let lastIndex = component.count - 1
+    return .init(component.enumerated().map({ index, source in
+      let element = source.create${element_type.buildable_base_name()}()
+      return index < lastIndex ? element.withTrailingComma(true) : element
+    }))
 %   else:
     return .init(component.map { $0.create${element_type.buildable()}() })
 %   end
--- a/Sources/SwiftSyntaxBuilder/gyb_helpers/SyntaxBuildableWrappers.py
+++ b/Sources/SwiftSyntaxBuilder/gyb_helpers/SyntaxBuildableWrappers.py
@@ -283,6 +283,15 @@ class SyntaxBuildableType:
         result.append(node.type())
     return result
 
+  def has_with_trailing_comma_trait(self):
+    """
+    Returns if this type has `WithTrailingComma` trait
+    """
+    for node in [SyntaxBuildableNode(node) for node in SYNTAX_NODES]:
+      if node.type() == self:
+        return node.node.traits and 'WithTrailingComma' in node.node.traits
+    return False
+
   def convertible_to_types(self):
     """
     Types that take a single non-optional parameter of this types and to which this type is thus convertible.
--- a/Sources/SwiftSyntaxBuilder/HasTrailingComma.swift
+++ b/Sources/SwiftSyntaxBuilder/HasTrailingComma.swift
@@ -1,0 +1,16 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2022 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+public protocol HasTrailingComma {
+  /// Returns this node overriding presence of the trailing comma
+  func withTrailingComma(_ withComma: Bool) -> Self
+}
--- a/Sources/SwiftSyntaxBuilder/gyb_generated/BuildableNodes.swift
+++ b/Sources/SwiftSyntaxBuilder/gyb_generated/BuildableNodes.swift
@@ -1629,7 +1629,7 @@ public struct DictionaryExpr: ExprBuildable, ExpressibleAsDictionaryExpr {
     return self
   }
 }
-public struct TupleExprElement: SyntaxBuildable, ExpressibleAsTupleExprElement {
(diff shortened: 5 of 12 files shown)


ahoppen (commented):
Thanks, this is a great improvement and exactly the kind of thing SwiftSyntaxBuilder needs 👍 

I think adding the trailing comma shouldn’t be done via the `create*` methods though. I put up a more detailed comment inline.

Sources/SwiftSyntaxBuilder/BuildableNodes.swift.gyb
```diff
@@ -23,10 +23,15 @@ import SwiftSyntax
...
+%   additionalConformance = ""
+%   has_with_trailing_comma_trait = node.node.traits and 'WithTrailingComma' in node.node.traits
+%   if has_with_trailing_comma_trait:
+%     additionalConformance = ", HasTrailingComma"
+%   end
 %   if node.documentation():
 /// ${node.documentation()}
 %   end
-public struct ${type.buildable()}: ${base_type.buildable()}, ${type.expressible_as()} {
+public struct ${type.buildable()}: ${base_type.buildable()}, ${type.expressible_as()}${additionalConformance} {
```
ahoppen, line 34: You could use the `conformance_clause` helper function to generate the conformance clause here. That way you don’t need to deal with adding the `,` between `ExpressibleAs*` and `HasTrailingComma`.

Sources/SwiftSyntaxBuilder/BuildableNodes.swift.gyb
```diff
@@ -114,6 +119,24 @@ public struct ${type.buildable()}: ${base_type.buildable()}, ${type.expressible_
...
+}%
+% if trailing_comma_init_args:
+
+  /// Conformance to `HasTrailingComma`.
+  public func withTrailingComma(_ withComma: Bool) -> Self {
+      return Self.init(
+        ${',\n        '.join(trailing_comma_init_args)}
+      )
+  }
+% end
```
ahoppen, line 139: Is there a reason why you can’t you move the `trailing_comma_init_args` initialization into the `if` statement to make it clear it’s not being used anywhere else?

```suggestion
% if has_with_trailing_comma_trait:
%   trailing_comma_init_args = []
%   for child in children:
%     if child.name() == 'trailingComma':
%       trailing_comma_init_args.append('%s: withComma ? .comma : nil' % (child.name()))
%     else:
%       trailing_comma_init_args.append('%s: %s' % (child.name(), child.name()))

  /// Conformance to `HasTrailingComma`.
  public func withTrailingComma(_ withComma: Bool) -> Self {
      return Self.init(
        ${',\n        '.join(trailing_comma_init_args)}
      )
  }
% end
```

Sources/SwiftSyntaxBuilder/gyb_helpers/SyntaxBuildableWrappers.py
```diff
@@ -281,6 +281,15 @@ def element_in_collections(self):
...
 
+  def has_with_trailing_comma_trait(self):
+    """
+    Returns if this type has `WithTrailingComma` trait
+    """
+    for node in [SyntaxBuildableNode(node) for node in SYNTAX_NODES]:
+      if node.type() == self:
+        return node.node.traits and 'WithTrailingComma' in node.node.traits
+    return False
+
```
ahoppen, line 292: Isn’t this equivalent to 

```suggestion
    return 'WithTrailingComma' in self.node.traits
```

Or am I missing something obvious?

evnik: I don't think the type has info about the node
> AttributeError: 'SyntaxBuildableType' object has no attribute 'node'

Sources/SwiftSyntaxBuilder/BuildableCollectionNodes.swift.gyb
```diff
@@ -35,6 +35,11 @@ public struct ${type.buildable()}: ExpressibleByArrayLiteral, SyntaxBuildable, $
   public init(_ elements: [${element_type.expressible_as()}]) {
 %   if element_type.is_token():
     self.elements = elements
+%   elif element_type.has_with_trailing_comma_trait():
+    let lastIndex = elements.count - 1
+    self.elements = elements.enumerated().map({ index, element in
+      element.create${element_type.buildable_base_name()}().withTrailingComma(index < lastIndex)
```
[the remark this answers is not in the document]
evnik: Originally I was expecting to unconditionally override trailing commas in the Result Builders. `create*(isLastElement: Bool?)` methods required to apply commas in both Result Builders and Element List types initializers. Now with the separate `withTrailingComma` method it's possible to limit that impact to Result Builders, which I missed in my recent commit. So I'm going to move that logic again, in that way trailing commas will be preserved in the Element List types initializers (so there is a way to define them manually), but always overridden in the Result Builders. Does it make sense?

Sources/SwiftSyntaxBuilder/gyb_helpers/SyntaxBuildableWrappers.py
ahoppen: Oh, sorry, I thought this was defined on `SyntaxBuildableNode` not `SyntaxBuildableType`. In that case, it makes total sense 👍

Sources/SwiftSyntaxBuilder/BuildableCollectionNodes.swift.gyb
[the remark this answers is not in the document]
ahoppen: Thanks for catching this. Yes, I think limiting the automatic trailing commas to result builders makes sense to me 👍 I think you should be able to implement that in the `buildFinalResult` of the result builder.

Sources/SwiftSyntaxBuilder/BuildableNodes.swift.gyb
```diff
@@ -114,6 +119,23 @@ public struct ${type.buildable()}: ${base_type.buildable()}, ${type.expressible_
...
   }
+% if has_with_trailing_comma_trait:
+%   trailing_comma_init_args = []
+%   for child in children:
+%     if child.name() == 'trailingComma':
+%       trailing_comma_init_args.append('%s: withComma ? .comma : nil' % (child.name()))
+%     else:
+%       trailing_comma_init_args.append('%s: %s' % (child.name(), child.name()))
+%     end
+%   end
```
ahoppen, line 130: I think your indentation is wrong here. This should be indented two more spaces because it’s inside the `for` loop started in line 23. This also applies to line 138.

Sources/SwiftSyntaxBuilder/BuildableNodes.swift.gyb
```diff
@@ -23,10 +23,15 @@ import SwiftSyntax
 % for node in [SyntaxBuildableNode(syntax_node) for syntax_node in SYNTAX_NODES if syntax_node.is_buildable()]:
 %   type = node.type()
 %   base_type = node.base_type()
+%   conformances = [base_type.buildable(), type.expressible_as()]
+%   has_with_trailing_comma_trait = node.node.traits and 'WithTrailingComma' in node.node.traits
```
ahoppen, line 27: I didn’t check in detail (and might be wrong) but could this be simplified to `type.has_with_trailing_comma_trait()`? In which case you could also eliminate the `has_with_trailing_comma_trait` variable.

Sources/SwiftSyntaxBuilder/ResultBuilders.swift.gyb
```diff
@@ -85,6 +85,11 @@ public struct ${type.result_builder()} {
   public static func buildFinalResult(_ component: Component) -> FinalResult {
 %   if element_type.is_token():
     return .init(component)
+%   elif element_type.has_with_trailing_comma_trait():
+    let lastIndex = component.count - 1
+    return .init(component.enumerated().map({ index, element in
+      element.create${element_type.buildable_base_name()}().withTrailingComma(index < lastIndex)
```
ahoppen, line 91: I think I mentioned this before but unconditionally calling `withTrailingComma` here effectively prevents you from adding a trailing comma using result builders (because the result builder will remove the comma on the last element).

I think we don’t have this problem if this would be something like

```suggestion
      var buildable = element.create${element_type.buildable_base_name()}()
      if index < lastIndex {
        buildable = buildable.withTrailingComma(index < lastIndex)
      }
      return buildable
```

I think it would also be worth adding a test case that explicitly adds a trailing comma to the last element and assert that the trailing comma is retained through the result builder.

Sources/SwiftSyntaxBuilder/BuildableNodes.swift.gyb
evnik: `has_with_trailing_comma_trait()` looks for a node related to the type in `SYNTAX_NODES`. Since we already have the `node` here, I think it is more optimal to use it right away. But if we don't care about those scripts performance, then yes, I can simplify it

Sources/SwiftSyntaxBuilder/ResultBuilders.swift.gyb
evnik: I thought [you've agreed](https://github.com/apple/swift-syntax/pull/454#discussion_r895417798) that it's fine if trailing commas are preserved in the Element List types initializers, but always overridden in the Result Builders (in both ways added and removed).

ahoppen: Sorry, I should have maybe been clearer (or I’m misunderstanding you). What I mean is that I think you should be able to generate e.g. an array literal with trailing comma like `[1,2,]` using the result builder syntax using something like the following (pseudo-code written in GitHub, probably doesn’t compile)
```swift
ArrayExpr {
  ArrayElement(IntegerLiteralExpr(1))
  ArrayElement(IntegerLiteralExpr(2), trailingComma: .comma)
}
```

So basically my thinking is that the result builder should only add commas but never remove any. My thinking is that some people prefer to have trailing commas in their array/dictionary literals and I don’t think we should deprive them of that ability.

evnik: I don't see much value in it because Result Builders don't allow explicit trivias on non-last trailing commas anyway. I think list type initializer should be used in the scenario you described. But since you insist, I've updated the PR with that and I've added a [test](https://github.com/apple/swift-syntax/pull/454/files#diff-168af55d0fe26fcb8e58da2a1e597fa45b6d9cbc7832c1a1d5a4273d50b7737eR31) for it as well.

ahoppen: I’m totally open to debate here. I would have just been surprised if the example I put above dropped the trailing comma. Do you disagree? Or do you have an example where the behavior you implemented before would be less surprising than what I’m suggesting?

ahoppen:
@swift-ci Please test

Sources/SwiftSyntaxBuilder/ResultBuilders.swift.gyb
evnik: Trailing comma in a single line array literal (like `[1, 2, 3, ]`) doesn't make much sense to me, however I see the point in case of multiline:
```
[
  1,
  2,
  3,
]
```
However current implementation won't allow you to generate code like the second example with result builder, because it'll overwrite trailing commas removing explicit new line trivias.
It might be surprising, but I'd say that it works as designed: you either use convenience of result builder with no way to customize the output, or you use more verbose Element List initializer and has full control on the output.
Finally, I'd like to merge this PR ASAP and then the logic could be improved as a separate change

ahoppen: OK, now I understand what you meant by “don't allow explicit trivias on non-last trailing commas”. With that in mind, I don’t have strong opinions in either direction anymore.

I’ve merged the PR as-is for now. If you want, I won’t object to reverting the behavior back to what you originally implemented. 😉

PR #455, opened 2022-06-04
Builder based initializer for `DictionaryExpr`

evnik:
Allow to use `ArrayType` and `DictionaryType` as `PatternBinding.typeAnnotation` argument

### DIFF
--- a/Sources/SwiftSyntaxBuilder/DictionaryExprConvenienceInitializers.swift
+++ b/Sources/SwiftSyntaxBuilder/DictionaryExprConvenienceInitializers.swift
@@ -1,0 +1,30 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2022 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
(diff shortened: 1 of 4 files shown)


ahoppen (commented):
This looks good but I think there’s a slight mismatch between PR title, description in code. AFAICT
- The description mentions “Allow to use ArrayType and DictionaryType as PatternBinding.typeAnnotation argument” covers all changes except DictionaryExprConvenienceInitializers.swift. These changes are looking great. ✅ 
- The title “Builder based initializer for DictionaryExpr” covers the changes in DictionaryExprConvenienceInitializers.swift. It would be good to have test coverage for those before merging. Maybe we can just cherry-pick the test case from https://github.com/apple/swift-syntax/pull/454.

evnik:
My intention was to avoid adding trailing commas, so I won't need to remove them in #454, but that's fine as well

ahoppen (approved):
Thanks. Looks good 👍 

I just prefer to keep the PRs isolated so we can’t forget to add the test case if for some reason we should decide to take a different direction on the other PR. It also makes for a nicer git history where you can see the test cases that were added for a particular code change.

ahoppen:
@swift-ci Please test

PR #458, opened 2022-06-04
Clean build-script by adding subcommands

kimdv:
Based on this comment: https://github.com/apple/swift-syntax/pull/381#discussion_r881655025
> Also: Since `build-script-helper.py` now does so many things, we should probably migrate it to using argparse subcommands. But that’s something for a completely standalone PR.

I tried to come with an example on how it could be structured. A command could look like: `python3 ./build-script.py degyb --degyb-only --verbose`
What do you think @ahoppen ?

Then we don't need to pass the `--toolchain` etc if it's not needed

### DIFF
--- a/build-script.py
+++ b/build-script.py
@@ -351,6 +351,7 @@ def clear_gyb_files_from_previous_run(sources_dir, destination_dir, verbose):
                     verbose=verbose
                 )
 
+
 # -----------------------------------------------------------------------------
 # Building SwiftSyntax
 
@@ -636,6 +637,115 @@ def run_xctests(toolchain, build_dir, multiroot_data_file, release, verbose):
     env["SWIFT_SYNTAX_PARSER_LIB_SEARCH_PATH"] = os.path.join(toolchain, "lib", "swift", "macosx")
     return call(swiftpm_call, env=env, verbose=verbose) == 0
 
+# -----------------------------------------------------------------------------
+# Arugment Parsing functions
+
+
+def generate_xcodeproj(args):
+    xcode_gen(config=args.xcconfig_path)
+
+
+def generate_source_code_command(args):
+    try:
+        generate_gyb_files(
+            args.gyb_exec,
+            verbose=args.verbose,
+            add_source_locations=args.add_source_locations,
+        )
+    except subprocess.CalledProcessError as e:
+        printerr("FAIL: Generating .gyb files failed")
+        printerr("Executing: %s" % " ".join(e.cmd))
+        printerr(e.output)
+
+    run_code_generation(
+        toolchain=args.toolchain,
+        build_dir=realpath(args.build_dir),
+        multiroot_data_file=args.multiroot_data_file,
+        release=args.release,
+        verbose=args.verbose,
+        swiftsyntaxbuilder_destination=os.path.join(SWIFTSYNTAXBUILDER_DIR, "generated")
+    )
+
+
+def verify_source_code_command(args):
+    try:
+        verify_generated_files(args.gyb_exec, verbose=args.verbose)
+
+        verify_code_generated_files(
+            toolchain=args.toolchain,
+            build_dir=realpath(args.build_dir),
+            multiroot_data_file=args.multiroot_data_file,
+            release=args.release,
+            verbose=args.verbose,
+        )
+
+        verify_c_syntax_nodes_match()
+    except subprocess.CalledProcessError as e:
+        printerr(
+            "FAIL: Gyb-generated files committed to repository do "
+            "not match generated ones. Please re-generate the "
+            "gyb-files using the following command, open a PR to the "
+            "SwiftSyntax project and merge it alongside the main PR."
+            "$ swift-syntax/build-script.py generate-source-code --toolchain /path/to/toolchain.xctoolchain/usr"
+        )
+
+
+def build_command(args):
+    try:
+        builder = Builder(
+            toolchain=args.toolchain,
+            build_dir=realpath(args.build_dir),
+            multiroot_data_file=args.multiroot_data_file,
+            release=args.release,
+            verbose=args.verbose,
+            disable_sandbox=args.disable_sandbox,
+        )
+        # Until rdar://53881101 is implemented, we cannot request a build of multiple
+        # targets simultaneously. For now, just build one product after the other.
+        builder.build("SwiftSyntax")
+        builder.build("SwiftSyntaxParser")
+        builder.build("SwiftSyntaxBuilder")
+        builder.build("SwiftSyntaxBuilderGeneration")
+    except subprocess.CalledProcessError as e:
+        printerr("FAIL: Building SwiftSyntax failed")
+        printerr("Executing: %s" % " ".join(e.cmd))
+        printerr(e.output)
+
+
+def test_command(args):
+    try:
+        builder = Builder(
+            toolchain=args.toolchain,
+            build_dir=realpath(args.build_dir),
+            multiroot_data_file=args.multiroot_data_file,
+            release=args.release,
+            verbose=args.verbose,
+            disable_sandbox=args.disable_sandbox,
+        )
+        
+        builder.build("lit-test-helper")
+        
+        success = run_tests(
+            toolchain=args.toolchain,
+            build_dir=realpath(args.build_dir),
+            multiroot_data_file=args.multiroot_data_file,
+            release=args.release,
+            filecheck_exec=realpath(args.filecheck_exec),
+            skip_lit_tests=args.skip_lit_tests,
+            verbose=args.verbose,
+        )
+        if not success:
+            # An error message has already been printed by the failing test
+            # suite
+            sys.exit(1)
+        else:
+            print("** All tests passed **")
+    except subprocess.CalledProcessError as e:
+        printerr("FAIL: Running tests failed")
+        printerr("Executing: %s" % " ".join(e.cmd))
+        printerr(e.output)
+        sys.exit(1)
+
 
 # -----------------------------------------------------------------------------
 # Arugment Parsing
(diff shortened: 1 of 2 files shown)


ahoppen:
Yes, that’s pretty much what I was thinking as well.

I think the possible subcommands could be
- `build`: Similar to `./build-script.py` without any arguments
- `test`: Similar to `./build-script.py --test`
- `generate-xcodeproj`: Similar to `./build-script.py --generate-xcodeproj`
- `degyb`: Similar to `./build-script.py --degyb-only`
- `run-swiftsyntaxbuilder-generation`: Similar to `./build-script.py --run-swiftsyntaxbuilder-generation`

We should be able to get rid of the options I mentioned in the “Similar to ” commands once we move to sub parsers.

build-script.py
```diff
@@ -671,216 +771,198 @@ def parse_args():
...
-                # suite
-                sys.exit(1)
-            else:
-                print("** All tests passed **")
-        except subprocess.CalledProcessError as e:
-            printerr("FAIL: Running tests failed")
-            printerr("Executing: %s" % " ".join(e.cmd))
-            printerr(e.output)
-            sys.exit(1)
+    # verify_c_syntax_nodes_match()
```
kimdv, line 965: Not sure where this should be places?
It this part of the `verify-generated-files` command?

kimdv:
Okay. 
I've tried to add the commands. 
There is some arguments that are duplicated. 

Should we add them to the `parser` so all sub-commands can use them?

ahoppen:
I just took a look at the actions that are currently supported by `build-script.py` and think it would make sense to provide the following sub-parser actions, with these options. Given that we now have gyb and SwiftSyntaxBuilderGeneration, I think it also makes sense to combine the two generation steps into one sub-command and decouple code-generation from building.

```
Actions:

generate-xcodeproj  Generate an Xcode project for SwiftSyntax.
  -v, --verbose         Enable verbose logging.
  --xcconfig-path XCCONFIG_PATH

generate-source-code    Generate source code using gyb and SwiftSyntaxBuilderGeneration
  -v, --verbose         Enable verbose logging.
  --add-source-locations
  --gyb-exec GYB_EXEC   Path to the gyb tool
  -v, --verbose         Enable verbose logging.
  -r, --release         Build in release mode.
  --build-dir BUILD_DIR
  --disable-sandbox     Disable sandboxes when building with SwiftPM
  --multiroot-data-file MULTIROOT_DATA_FILE
  --toolchain TOOLCHAIN
  --verify              Instead of writing the newly generated files to the source directory, assert that the generated files match the ones currently in the source directory

build                   Build all SwiftSyntax modules (does not re-generate source code - this is a change from the current behavior)
  -v, --verbose         Enable verbose logging.
  -r, --release         Build in release mode.
  --build-dir BUILD_DIR
  --disable-sandbox     Disable sandboxes when building with SwiftPM
  --multiroot-data-file MULTIROOT_DATA_FILE
  --toolchain TOOLCHAIN

test                    Run tests (does not re-generate source code - this is a change from the current behavior)
  -v, --verbose         Enable verbose logging.
  -r, --release         Build in release mode.
  --build-dir BUILD_DIR
  --disable-sandbox     Disable sandboxes when building with SwiftPM
  --multiroot-data-file MULTIROOT_DATA_FILE
  --toolchain TOOLCHAIN
  --skip-lit-tests      Don't run lit-based tests
  --filecheck-exec FILECHECK_EXEC
```

Regarding the common args, I think it makes sense to have two functions that add these common args (one for `--verbose` and one for all the build flags), similar to https://github.com/apple/sourcekit-lsp/blob/472a06c88af97135b6faa450c5854e8f1a0ab8de/Utilities/build-script-helper.py#L169-L181

kimdv:
@ahoppen I added your suggestions on groups. 
Made it a bit more clean.

build-script.py
```diff
@@ -661,115 +768,117 @@ def run_xctests(toolchain, build_dir, multiroot_data_file, release, verbose):
...
-        "line-directive.",
+    generate_source_code_parser.add_argument(
+        "--gyb-exec",
+        default=GYB_EXEC,
+        help="Path to the gyb tool (default: %(default)s).",
     )
 
-    build_group.add_argument(
+    generate_source_code_parser.add_argument(
         "--degyb-only",
```
kimdv, line 837: Added `--degyb-only` to `generate-source-code` because in #381 you mention that it should not be implicit.
https://github.com/apple/swift-syntax/pull/381#discussion_r881655025

ahoppen (commented):
Thanks. I think this looks a lot cleaner than our old build-script and I like the separation of the subcommands in different functions 👍

build-script.py
```diff
@@ -636,6 +637,112 @@ def run_xctests(toolchain, build_dir, multiroot_data_file, release, verbose):
...
+        # Until rdar://53881101 is implemented, we cannot request a build of multiple
+        # targets simultaneously. For now, just build one product after the other.
+        builder.build("SwiftSyntax")
+        builder.build("SwiftSyntaxParser")
+        builder.build("SwiftSyntaxBuilder")
+        builder.build("SwiftSyntaxBuilderGeneration")
+
+        # Only build lit-test-helper if we are planning to run tests
+        if args.test:
+            builder.build("lit-test-helper")
```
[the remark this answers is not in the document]
kimdv: Should this also not be moved to thee `test` command?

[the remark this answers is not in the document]
ahoppen: I initially thought that building `lit-test-helper` even as part of `./build-script.py build` would make sense just to make sure there are no syntax errors in it. But I now realized that we don’t build the tests as part of `./build-script.py build` either, so you’re right, we should probably only build it in the `test` subcommand.

kimdv:
@ahoppen I have fixed all the comments. 
How do we setup the CI? 

I'm not sure it knows about the new sub-commands?

kimdv:
https://github.com/apple/swift/pull/59651

@swift-ci please test

kimdv:
https://github.com/apple/swift/pull/59651

@swift-ci please test

build-script.py
```diff
@@ -636,6 +637,114 @@ def run_xctests(toolchain, build_dir, multiroot_data_file, release, verbose):
...
+        )
+
+        verify_c_syntax_nodes_match()
+    except subprocess.CalledProcessError as e:
+        printerr(
+            "FAIL: Gyb-generated files committed to repository do "
+            "not match generated ones. Please re-generate the "
+            "gyb-files using the following command, open a PR to the "
+            "SwiftSyntax project and merge it alongside the main PR."
+            "$ swift-syntax/build-script.py generate-source-code --toolchain <path-to-toolchain>"
```
ahoppen, line 745: Nitpick: To disambiguate how the toolchain should be passed, I like to include the final `/usr` in the path. Otherwise I’m often not sure whether `<path-to-toolchain>` should point to the `.xctoolchain` bundle or `.xctoolchain/usr`

```suggestion
            "$ swift-syntax/build-script.py generate-source-code --toolchain /path/to/toolchain.xctoolchain/usr"
```

build-script.py
```diff
@@ -661,114 +770,116 @@ def run_xctests(toolchain, build_dir, multiroot_data_file, release, verbose):
...
-        "--degyb-only",
-        action="store_true",
-        help="The script only generates swift files from gyb and skips the "
-        "rest of the build",
-    )
-
-    build_group.add_argument(
-        "--run-swiftsyntaxbuilder-generation",
+    generate_source_code_parser.add_argument(
+        "--verify",
```
ahoppen, line 846: I don’t think `--verify` is used anywhere because we now have the `verify-source-code` subcommand.

build-script.py
```diff
@@ -661,114 +770,116 @@ def run_xctests(toolchain, build_dir, multiroot_data_file, release, verbose):
...
     )
 
-    test_group.add_argument(
-        "--gyb-exec",
-        default=GYB_EXEC,
-        help="Path to the gyb tool (default: %(default)s).",
-    )
+    # -------------------------------------------------------------------------
+    verify_source_code_parser = sub_parsers.add_parser("verify-source-code")
+    verify_source_code_parser.set_defaults(func=verify_source_code_command)
```
ahoppen, line 879: AFAICT `verify-source-code` also needs a `gyb-exec` parameter because we’re referencing that from the `verify_source_code_command` function.

build-script.py
```diff
@@ -636,6 +637,114 @@ def run_xctests(toolchain, build_dir, multiroot_data_file, release, verbose):
...
+            sys.exit(1)
+        else:
+            print("** All tests passed **")
+    except subprocess.CalledProcessError as e:
+        printerr("FAIL: Running tests failed")
+        printerr("Executing: %s" % " ".join(e.cmd))
+        printerr(e.output)
+        sys.exit(1)
+
+def verify_source_code_command(args):
```
ahoppen, line 726: Nitpick: I think it would be nice if we defined `verify_source_code_command` right below `generate_source_code_command` because they perform very similar tasks. Similarly, I think defining the `verify-source-code` subcommand right after `generate-source-code` would make sense to me.

kimdv:
https://github.com/apple/swift/pull/59651

@swift-ci please test

kimdv:
Failed because it didn't recognise `--verbose`. 
It need to come before the `build` argument, as it is in the root parser.

This syntax seems a bit strange @ahoppen ?

kimdv:
https://github.com/apple/swift/pull/59651

@swift-ci please test

ahoppen:
> Failed because it didn't recognise `--verbose`. It need to come before the `build` argument, as it is in the root parser.
> 
> This syntax seems a bit strange @ahoppen ?

Ah, I remember. This is because `--verbose` is an argument to the top level parser. We probably want to add `--verbose` as an option to every sub parser.

kimdv:
https://github.com/apple/swift/pull/59651

@swift-ci please test

kimdv:
https://github.com/apple/swift/pull/59651

@swift-ci please test

ahoppen (approved):
One minor thing: `CONTRIBUTING.md` needs to be updated to include the `build` subcommand.

Afterwards, let’s 🚢 it

kimdv:
https://github.com/apple/swift/pull/59651

@swift-ci please test

kimdv:
@ahoppen should https://github.com/apple/swift/pull/59651 also be rebuild?

ahoppen:
No, I mixed up the PRs. I thought this was about making matching Swift files for the Python definitions. No need to re-build swift if you just changed CONTRIBUTING.md

PR #459, opened 2022-06-05
Omit parenthesis for calls with no arguments and a trailing closure

evnik:


Tests/SwiftSyntaxBuilderTest/FunctionTests.swift
```diff
@@ -60,4 +60,43 @@ final class FunctionTests: XCTestCase {
...
+    XCTAssertEqual(syntax.description, "test{}")
+  }
+
+  func testParensOmittedForMultipleTrailingClosures() {
+    let buildable = FunctionCallExpr("test", additionalTrailingClosuresBuilder: {
+      MultipleTrailingClosureElement(label: .identifier("a").withLeadingTrivia(.spaces(1)), closure: ClosureExpr())
+      MultipleTrailingClosureElement(label: .identifier("b").withLeadingTrivia(.spaces(1)), closure: ClosureExpr())
+    })
+    let syntax = buildable.buildSyntax(format: Format())
+    XCTAssertEqual(syntax.description, "test a: {} b: {}")
```
ahoppen, line 100: This will not compile because the first trailing closure needs to be unlabeled. Could you change the test case to pass the first trailing closure as the `trailingClosure` argument instead of in the `additionalTrailingClosuresBuilder`?

Sources/SwiftSyntaxBuilder/FunctionCallExprConvenienceInitializers.swift
```diff
@@ -16,19 +16,33 @@ import SwiftSyntax
 extension FunctionCallExpr {
   public init(
     _ calledExpression: ExpressibleAsExprBuildable,
-    leftParen: TokenSyntax? = .leftParen,
-    rightParen: TokenSyntax? = .rightParen,
+    leftParen: TokenSyntax? = nil,
+    rightParen: TokenSyntax? = nil,
```
ahoppen, line 20: With the current signature, I would expect that there are no parenthesis if I explicitly pass `nil` to `leftParen` and `rightParen`, even if it results in invalid code.

But since this is a convenience initializer, what do you think about dropping the `leftParen` and `rightParen` arguments altogether and adding a comment to the function that the presence of the parenthesis will be inferred by the presence of arguments?

Sources/SwiftSyntaxBuilder/FunctionCallExprConvenienceInitializers.swift
```diff
@@ -16,19 +16,33 @@ import SwiftSyntax
...
     trailingClosure: ExpressibleAsClosureExpr? = nil,
     @TupleExprElementListBuilder argumentListBuilder: () -> ExpressibleAsTupleExprElementList = { TupleExprElementList([]) },
     @MultipleTrailingClosureElementListBuilder additionalTrailingClosuresBuilder: () -> MultipleTrailingClosureElementList? = { nil }
   ) {
+    let additionalTrailingClosures = additionalTrailingClosuresBuilder()
+    let argumentList = argumentListBuilder().createTupleExprElementList()
+    let shouldSetParens = argumentList.elements.isEmpty == false
+      || (trailingClosure == nil && additionalTrailingClosures?.elements.isEmpty != false)
+    var leftParen = leftParen
+    if leftParen == nil && shouldSetParens {
```
ahoppen, line 30: Nitpick: The preferred spelling in Swift would be

```suggestion
    if leftParen == nil, shouldSetParens {
```

Tests/SwiftSyntaxBuilderTest/FunctionTests.swift
evnik: I think we shouldn't allow building invalid code, similarly to the case with no parenthesis.
Should we drop the `trailingClosure` argument and populate it from the `additionalTrailingClosuresBuilder`? Alternatively we might provide two initializers, one with just optional `trailingClosure`(no `additionalTrailingClosuresBuilder`) and another with both non-optional `trailingClosure` and `additionalTrailingClosuresBuilder`. What do you think?

ahoppen: I don’t think we should try too hard to prevent users from constructing SwiftSyntax trees that won’t compile. They will always be able to construct such trees using the `SyntaxFactory` API. My standpoint was that everything that matches the constraints in `gyb_syntax_support` should be considered a valid SwiftSyntax tree.

I think the cleanest solution here would be to restructure `FunctionCallExpr` in `gyb_syntax_support` in a way that enforces that there is an unlabeled trailing closure in front of labelled trailing closures. That way we can also make the same assumption when consuming a SwiftSyntax tree. Such a change would be slightly bigger but still manageable IMHO.

Sources/SwiftSyntaxBuilder/FunctionCallExprConvenienceInitializers.swift
```diff
@@ -14,21 +14,26 @@
...
-    leftParen: TokenSyntax? = .leftParen,
-    rightParen: TokenSyntax? = .rightParen,
     trailingClosure: ExpressibleAsClosureExpr? = nil,
     @TupleExprElementListBuilder argumentListBuilder: () -> ExpressibleAsTupleExprElementList = { TupleExprElementList([]) },
     @MultipleTrailingClosureElementListBuilder additionalTrailingClosuresBuilder: () -> MultipleTrailingClosureElementList? = { nil }
   ) {
+    let additionalTrailingClosures = additionalTrailingClosuresBuilder()
+    let argumentList = argumentListBuilder().createTupleExprElementList()
+    let shouldEmitParens = argumentList.elements.isEmpty == false
+      || (trailingClosure == nil && additionalTrailingClosures?.elements.isEmpty != false)
```
ahoppen, line 29: A couple of suggestions here that improve legibility IMHO.
- I found this easier to read if it‘s storing whether to omit parenthesis rather than when to emit them. I think this is because I’m mentally thinking about the edge case where they are not necessary rather than the normal cases where they are required
- Since for valid code having additional trailing closures implies having a “normal” trailing closure, I don’t think we need to check for `additionalTrailingClosures?.elements.isEmpty` here

```suggestion
    let shouldOmitParens = argumentList.elements.isEmpty && trailingClosure != nil
```

ahoppen (approved):

ahoppen:
@swift-ci Please test

PR #460, opened 2022-06-07
Take attribute and modifier lists directly in declaration convenience initializers

fwcd:
As suggested by @ahoppen, this branch explores taking attributes and modifiers directly in the convenience initializers for declarations in `SwiftSyntaxBuilder` rather than as result builders, primarily to simplify the DSL and to improve the usability of trailing closures for the 'body' of most constructs.

Although the branch passes the build, there are a few challenges that I have encountered with this approach. For this reason I will mark the PR as draft for now, so these points can be discussed and addressed first:

#### We cannot use `ExpressibleByArrayLiteral` conformances in parameters

Since we pass attributes and modifiers as existentials (`ExpressibleAsAttributeList` and `ExpressibleAsModifierList`, respectively), we cannot use the conformance of `AttributeList`/`ModifierList` to `ExpressibleByArrayLiteral` directly. More specifically, we have to explicitly wrap these parameters as

```swift
ExtensionDecl(
  modifiers: ModifierList([...]),
  ...
)
```

rather than being able to say

```swift
ExtensionDecl(
  modifiers: [...],
  ...
)
```

As a potential solution we could take `AttributeList`/`ModifierList` directly in these initializers, at the cost of making the initializers less polymorphic (and less consistent, considering that the other arguments are taken existentially too).

- **Update:** Fixed by #461 

#### Not using result builders increases ambiguity with generated initializers

For example, `FunctionDecl` has three initializers:

- Two generated initializers in `BuildableNodes` (a direct initializers and one that takes attributes/modifiers as builders)
- A convenience initializer in `FunctionDeclConvenienceInitializers`, which now only takes the body as builder

The compiler can now only disambiguate these initializers by the final `bodyBuilder` closure, even passing it as a trailing closure (which we want to encourage with this API design) can cause ambiguities with the modifier builder closure from the generated initializer.

A potential solution would be to not generate builder-based initializers automatically at all and only include them in convenience initializers, this could potentially be a larger change though.

Thoughts? @ahoppen @kimdv

### DIFF
--- a/Sources/SwiftSyntaxBuilderGeneration/Templates/TokensFile.swift
+++ b/Sources/SwiftSyntaxBuilderGeneration/Templates/TokensFile.swift
@@ -18,22 +18,14 @@ let tokensFile = SourceFile {
   ImportDecl(importTok: TokenSyntax.import.withLeadingTrivia(.docLineComment(copyrightHeader)), path: "SwiftSyntax")
 
   ExtensionDecl(
+    modifiers: ModifierList([TokenSyntax.public.withLeadingTrivia(.newlines(1) + .docLineComment("/// Namespace for commonly used tokens with default trivia.") + .newlines(1))]),
     extendedType: "TokenSyntax",
-    modifiersBuilder: {
-      TokenSyntax.public.withLeadingTrivia(.newlines(1) + .docLineComment("/// Namespace for commonly used tokens with default trivia.") + .newlines(1))
-    },
     membersBuilder: {
       for token in SYNTAX_TOKENS {
         if token.isKeyword {
           VariableDecl(
-            letOrVarKeyword: .var,
-            modifiersBuilder: {
-              if let text = token.text {
-                TokenSyntax.static.withLeadingTrivia(.newlines(1) + .docLineComment("/// The `\(text)` keyword") + .newlines(1))
-              } else {
-                TokenSyntax.static
-              }
-            },
+            modifiers: ModifierList([token.text.map { TokenSyntax.static.withLeadingTrivia(.newlines(1) + .docLineComment("/// The `\($0)` keyword") + .newlines(1)) } ?? TokenSyntax.static]),
+            .var,
             bindingsBuilder: {
               // We need to use `CodeBlock` here to ensure there is braces around.
 
@@ -52,16 +44,10 @@ let tokensFile = SourceFile {
               createTokenSyntaxPatternBinding("`\(token.name.withFirstCharacterLowercased)`", accessor: accessor)
             }
           )
-        } else if token.text != nil {
+        } else if let text = token.text {
           VariableDecl(
-            letOrVarKeyword: .var,
-            modifiersBuilder: {
-              if let text = token.text {
-                TokenSyntax.static.withLeadingTrivia(.newlines(1) + .docLineComment("/// The `\(text)` token") + .newlines(1))
-              } else {
-                TokenSyntax.static
-              }
-            },
+            modifiers: ModifierList([TokenSyntax.static.withLeadingTrivia(.newlines(1) + .docLineComment("/// The `\(text)` token") + .newlines(1))]),
+            .var,
             bindingsBuilder: {
               // We need to use `CodeBlock` here to ensure there is braces around.
               let accessor = CodeBlock {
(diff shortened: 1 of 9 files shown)


Sources/SwiftSyntaxBuilderGeneration/Templates/TokensFile.swift
```diff
@@ -18,22 +18,14 @@ let tokensFile = SourceFile {
   ImportDecl(importTok: TokenSyntax.import.withLeadingTrivia(.docLineComment(copyrightHeader)), path: "SwiftSyntax")
 
   ExtensionDecl(
+    modifiers: ModifierList([TokenSyntax.public.withLeadingTrivia(.newlines(1) + .docLineComment("/// Namespace for commonly used tokens with default trivia.") + .newlines(1))]),
```
fwcd, line 21: Also, we might want to extract these `.withLeadingTrivia`s, rather than 'attaching' them to the first modifier token. I haven't dug too deep into this yet, perhaps we could extend constructs like `ExtensionDecl` to have a `withLeadingTrivia` method?

```swift
SourceFile {
  ...
  ExtensionDecl(
    ...
  )
  .withLeadingTrivia(.newlines(1) + .docLineComment(...) + .newlines(1))
}
```

Or perhaps declare them separately, something very roughly along the lines of

```swift
SourceFile {
  ...
  Trivia([.newlines(1) + .docLineComment(...) + .newlines(1)])
  ExtensionDecl(
    ...
  )
}
```

(straw-man syntax, I haven't checked whether we could actually do something like this)

Anyway, this is probably something for another PR.

ahoppen: I think with the current design your first proposal is easier to implement for now, the second one looks promising but there are a few challenges like deciding whether to attach the trivia as leading trivia to `ExtensionDecl` or as trailing trivia to the previous node.

@iMostfa started working on your first proposal in https://github.com/apple/swift-syntax/pull/385. Maybe you could pick up from there, but I agree that this is something for a follow-up PR.

fwcd: Interesting, I didn't know about the other PR. For the second idea, perhaps something like

```swift
SourceFile {
  ...
  LeadingTrivia(...)
  ExtensionDecl(...)
  TrailingTrivia(...)
  ...
}
```

would be possible, having the builder attach the trivia to the adjacent node internally. This way the builder closure would closely resemble the generated code.

ahoppen: That sounds like an interesting idea. We need to figure out how to express this in the type system, but that should be possible with a couple more protocols.

fwcd:
I have looked at bit into your second suggestion. The gyb-generated builder initializers only apply to structures that are initialized directly by syntax collections, e.g. `MemberDeclBlock` being initialized by a `MemberDeclList`. Many structures, like `EnumDecl` for example, take a block which is not considered to be a syntax collection. Since declaring all of these blocks gets verbose in practice, the current solution to this is a wide range of explicitly declared convenience initializers (located in `XyzDeclConvenienceInitializers.swift`). I have a few ideas how we could simplify this:

- We replace the explicitly declared convenience initializers with more sophisticated gyb generation of initializers that e.g. automatically generate builders for blocks (which only indirectly are syntax collections). This would probably be the least verbose and most maintainable solution.
- We remove the builder-based initializers from the gyb generation and only use explicitly declared convenience initializers for builder closures. This make it easy to precisely control where builders can be used and potentially reduce ambiguities between initializers.
- We leave everything as is and only refine the gyb generation. The main drawback I see here is that the builder-based initializers generated by gyb aren't actually used much in practice, since the explicitly declared convenience initializers are almost always more convenient.

Thoughts?

ahoppen:
> * We replace the explicitly declared convenience initializers with more sophisticated gyb generation of initializers that e.g. automatically generates builders for blocks (which only indirectly are syntax collections). This would probably be the least verbose and most maintainable solution.

Just to clarify I understand you correctly: You are saying that if we limit the automatically generated result builders to only a few syntax collections (e.g. only `MemberDeclList` and `CodeBlockItemList`), we could teach the result builder generation to look through these types. That is, if a type takes a `MemberDeclBlock`, we allow initializing it using a `@MemberDeclListBuilder`, correct? If this is what you are suggesting, I think it’s a great idea.

fwcd:
Yes, that was my idea and in doing so we would no longer need all of the explicitly declared convenience initializers, since they essentially work around this.

fwcd:
Replaced with #465.

PR #461, opened 2022-06-09
Conform `Array` to various `ExpressibleAs` protocols in `BuildableCollectionNodes`

fwcd:
In a separate PR (from #460) for easier reviewability. This adds a bunch of `Array` conformances that let us initialize syntax collections more conveniently, for example:

```swift
ExtensionDecl(
  modifiers: [...],
  ...
)
```

instead of

```swift
ExtensionDecl(
  modifiers: ModifierList([...]),
  ...
)
```

(This applies whenever the parameter has an existential type a la `ExpressibleAsModifierList`, in which case the compiler cannot infer that we would like to use the `ExpressibleByArrayLiteral` conformance from `ModifierList`)

cc @ahoppen

### DIFF
--- a/Sources/SwiftSyntaxBuilder/BuildableCollectionNodes.swift.gyb
+++ b/Sources/SwiftSyntaxBuilder/BuildableCollectionNodes.swift.gyb
@@ -81,4 +81,12 @@ public struct ${type.buildable()}: ExpressibleByArrayLiteral, SyntaxBuildable, $
   }
 }
 
+% if not type.generated_expressible_as_conformances():
+extension Array: ${type.expressible_as()} where Element == ${element_type.expressible_as()} {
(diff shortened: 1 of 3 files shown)


kimdv:
@swift-ci please test

kimdv (approved):

ahoppen (commented):
Nice. This looks great. Could you add a couple test cases?

fwcd:
Sure, I've added a few tests. Additionally, I have updated the conformance clause to always test for equality against the element type, i.e. use existentials. This is slightly more flexible and more consistent with the existing initializers e.g. in `ModifierList`, which also take a list of existentials (`[ExpressibleAsDeclModifier]`).

kimdv:
@swift-ci please test

ahoppen (approved):
Thanks. Looks great!

PR #462, opened 2022-06-12
Allow combining code block item lists

anreitersimon:
Hi!

I was trying to write a SwiftPackage to allow generating some code.

As part of my setup i tried to introduce a abstraction of a `CodeGenStep`.

The idea is that i would define multiple `CodeGenStep`, each of them would generate some (top-level) code
And i wanted to then loop over all of them and combine their outputs.

I used `CodeBlockItemListBuilder` since that felt most appropriate to generate code in that scope.

```swift
protocol CodeGenStep {

    @CodeBlockItemListBuilder
    var code: CodeBlockItemList { get }
}
```

This worked great!
My problem then was combining these `CodeBlockItemList`s and generating the final result.

```swift
let steps: [CodeGenStep] = [...]

let sourceFile = SourceFile(
    eofToken: .eof,
    statementsBuilder: {
      for step in steps {
        step.code // this currently emits a error
      }
    }
)
```

Also manually merging (outside of the result builder) did also not work, since `CodeBlockItemList.elements` has `internal` accessLevel so i cannot manually merge them

I can imagine this being a common use case so i thought it open up a PR.

## My Solution

This basically adds a `buildExpression` to the `CodeBlockItemListBuilder` which makes the code snippet above legal.
Additionally it provides a initializer to allow combining multiple `CodeBlockItemList`s into one.


anreitersimon:
Thinking about this some more it might make sense to extend this approach to more constructs.

Basically everything ending in `List` (for instance `MemberDeclList`, `DeclNameArgumentList`)

The pattern would be really similar to the one in the PR, which makes me think it could potentially be added to the code-generation.

Following this template

```swift
extension {{Base}}ListBuilder {

     public static func buildExpression(_ expression: {{Base}}List) -> Component {
         return expression.elements
     }
 }

 extension{{Base}}List {

     public init(combining lists: [ExpressibleAs{{Base}}List]) {
         self.elements = lists.flatMap {
             $0.create{{Base}}().elements
         }
     }

 }
```

ahoppen:
Thanks for your PR @anreitersimon, this sounds like a wonderful idea. I also agree that this is not only useful for `CodeBlockItemListBuilder` but for all syntax collections that can be initialized by result builders.

As you might have already seen, we use `.swift.gyb` files to code-generate most of SwiftSyntaxBuilder, so you should be able to add exactly your pattern to these files and make them apply to all syntax collections. The result builder extension would need to go into [ResultBuilders.swift.gyb](https://github.com/apple/swift-syntax/blob/main/Sources/SwiftSyntaxBuilder/ResultBuilders.swift.gyb) and your combining initializer into [BuildableCollectionNodes.swift.gyb](https://github.com/apple/swift-syntax/blob/main/Sources/SwiftSyntaxBuilder/BuildableCollectionNodes.swift.gyb). If you need any help writing the gyb code, please let me know.

anreitersimon:
@ahoppen Thanks for the pointer, i think i figured it out.

Sources/SwiftSyntaxBuilder/BuildableCollectionNodes.swift.gyb
```diff
@@ -39,6 +39,13 @@ public struct ${type.buildable()}: ExpressibleByArrayLiteral, SyntaxBuildable, $
     self.elements = elements.map { $0.create${element_type.buildable_base_name()}() }
 %   end
   }
+  
+  public init(combining lists: [${type.expressible_as()}]) {
+    self.elements = lists.flatMap {
+        $0.create${type.buildable()}().elements
```
ahoppen, line 45: Nitpick: Could you indent with 2 spaces to match the surrounding style?

Sources/SwiftSyntaxBuilder/BuildableCollectionNodes.swift.gyb
```diff
@@ -39,6 +39,13 @@ public struct ${type.buildable()}: ExpressibleByArrayLiteral, SyntaxBuildable, $
     self.elements = elements.map { $0.create${element_type.buildable_base_name()}() }
 %   end
   }
+  
+  public init(combining lists: [${type.expressible_as()}]) {
```
ahoppen, line 43: Could you add a doc comment to this function? Something like:
```
/// Creates a new ${type.buildable()} by flattening the elements in `lists`
```

ahoppen (commented):
Thanks. That looks great. I’ve added a few comments inline regarding doc-comments inline. Could you also add a test case for this new behavior?

Sources/SwiftSyntaxBuilder/ResultBuilders.swift.gyb
```diff
@@ -49,7 +49,13 @@ public struct ${type.result_builder()} {
   public static func buildExpression(_ expression: Expression) -> Component {
     return [expression]
   }
-
+  
+  /// If declared, provides contextual type information for statement
+  /// expressions to translate them into partial results.
```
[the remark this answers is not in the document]
ahoppen: I don’t think the existing comment
```swift
```

provides much value. If you agree, you could delete it – or keep it, I don’t have strong opinions.

ahoppen (commented):
> Im specifically wondering if the tests should cover all of the builders, or if it's ok to just cover a few to test that it works in general.

Just testing one builder would be totally fine. I don’t think there’s much value in testing two of them since they are all generated using the same technique.

After finishing your changes, could you squash all of your commits to keep the Git history clean?

anreitersimon:
@ahoppen I think all the Feedback should be addressed.

* added `CollectionNodeFlatteningTests.swift` with two tests (one for  the result-builder and another one for the constructor)
* Removed the generic doc-comments on the resultbuilder method, keeping only the comment about flattening
* squashed the commits into a single commit

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

PR #464, opened 2022-06-16
Add default initialization for EOF token to auto-generate SourceFile convenience initializer

fwcd:
The EOF token currently does not yield a `.default_initialization()` since it is not part of the `SYNTAX_TOKEN_MAP` (`.token()` therefore evaluates to `None`). This makes it slightly inconvenient to initialize `SourceFile`, since the `eof` argument has to be passed explicitly:

```swift
public init(
  eofToken: TokenSyntax,
  @CodeBlockItemListBuilder statementsBuilder: () -> ExpressibleAsCodeBlockItemList = { CodeBlockItemList([]) }
)
```

While the convenience initializer from `SourceFileConvenienceInitializers.swift` currently works around this, this PR makes the explicit convenience initializer superfluous, since the generated initializer is now essentially equivalent:

```swift
public init(
  eofToken: TokenSyntax = TokenSyntax.eof,
  @CodeBlockItemListBuilder statementsBuilder: () -> ExpressibleAsCodeBlockItemList = { CodeBlockItemList([]) }
)
```

cc @ahoppen


kimdv (approved):

kimdv:
@swift-ci please test

ahoppen (approved):
LGTM 👍

PR #465, opened 2022-06-16
Clean up generation of builder-based convenience initializers

fwcd:
Based on #464 (but in a separate PR for easier reviewability).

This branch implements the proposed ideas from [this thread](https://github.com/apple/swift-syntax/pull/460#issuecomment-1149569900), replacing the various explicit `...DeclConvenienceInitializers` with improved gyb-generation of initializers that:

- automatically generates initializers for `...Block`s too (previously only direct syntax collections were possible)
- only generates builder parameters for a handful of block-like structures, in an effort to clean up and to e.g. make trailing closures more ergonomic in these APIs
  - currently this includes `CodeBlock`, `MemberDeclBlock`, `CodeBlockItemList`, `MemberDeclList` and `PatternBindingList`, but the list is easy to extend as needed
- aims to make as many explicitly declared convenience initializers as possible redundant

cc @ahoppen


kimdv:
@swift-ci please test

Sources/SwiftSyntaxBuilder/BuildableNodes.swift.gyb
```diff
@@ -54,17 +54,37 @@ public struct ${type.buildable()}: ${base_type.buildable()}, ${type.expressible_
   }
 
 %{
+    RESULT_BUILDER_TYPES = {
+      'CodeBlock': 'CodeBlockItemList',
+      'MemberDeclBlock': 'MemberDeclList',
+      'CodeBlockItemList': None,
+      'MemberDeclList': None,
+      'PatternBindingList': None,
+    }
```
ahoppen, line 63: Do you think it makes sense to move this definition to its own file similar to ExpressibleAsConformances.py?

Sources/SwiftSyntaxBuilder/BuildableNodes.swift.gyb
```diff
@@ -54,17 +54,37 @@ public struct ${type.buildable()}: ${base_type.buildable()}, ${type.expressible_
...
+
+    def is_initializable_by_result_builder(type):
+      return type.non_optional().buildable() in RESULT_BUILDER_TYPES
+    
+    def result_builder_type(type):
+      buildable = type.non_optional().buildable()
+      return SyntaxBuildableType(RESULT_BUILDER_TYPES.get(buildable, None) or buildable, is_optional=type.is_optional)
```
ahoppen, line 70: What do you think about making this a method on `SyntaxBuildableType`?

ahoppen (commented):
LGTM. I’ve got a couple minor suggestions inline. Also:
- I think we could use trailing closure syntax inside `Templates/TokensFile.swift`

fwcd:
I have rebased the PR on `main` and addressed your remarks, so the PR should be ready for a second round of review

evnik:
Could you please keep using Result Builders for collections of elements with trailing commas? Otherwise #454 won't make much sense

fwcd:
Sure! I think result builders absolutely make sense for things like collection literals where the trailing closure syntax looks nice too. The main goal here is to reduce the number of builders that prevent us from having nice trailing-closure APIs (like modifier/attribute lists etc.), but more use cases remain to be investigated.

Sources/SwiftSyntaxBuilderGeneration/Templates/TokensFile.swift
```diff
@@ -18,148 +18,123 @@ let tokensFile = SourceFile {
...
+            rightParen: .rightParen.withTrailingTrivia(.spaces(1))
+          ),
+          output: "TokenSyntax"
+        )
+
+        FunctionDecl(
+          modifiers: [TokenSyntax.static],
+          identifier: .identifier("`\(token.name.withFirstCharacterLowercased)`"),
+          signature: signature,
+          bodyBuilder: {
```
ahoppen, line 85: I think `bodyBuilder` could also be trailing closure syntax.

ahoppen (approved):
LGTM 👍 Just one minor inline comment.

Sources/SwiftSyntaxBuilder/gyb_helpers/BuilderInitializableTypes.py
```diff
@@ -0,0 +1,15 @@
...
+    'MemberDeclList': None,
+    'PatternBindingList': None,
+    'SwitchCaseList': None,
+    'ArrayElementList': None,
+    'TupleExprElementList': None,
+    'EnumCaseElementList': None,
+    'FunctionParameterList': None,
+    'GenericParameterList': None,
+    'GenericRequirementList': None,
+    'InheritedTypeList': None,
```
fwcd, line 14: I have added quite a few new types here, mostly since the array-based initializers do not support the new trailing comma logic that was recently merged. In general my suggestion would be that we try to use builders for every parameter that can be reasonably well expressed using a single trailing closure, since that usually produces very concise and nice-looking DSL code.

fwcd (commented):
I have rebased the PR onto `main` and squashed the commits. A note regarding the extended use of builders below...

Sources/SwiftSyntaxBuilder/gyb_helpers/BuilderInitializableTypes.py
fwcd: That also plays well e.g. with function parameter lists generated by for-loops (within a builder)

fwcd: Another thing I thought about was whether we would want to automatically create builders for every `HasTrailingComma` type. This would e.g. in `IfStmt`, however, lead to an initializer with multiple builder closures (`conditions` and `body`), which we'd like to avoid, therefore I went with the manual approach for now.

ahoppen: I agree that maintaining this list manually is the better option because, as you mentioned, it gives us the flexibility to try and avoid initializers that take multiple result builders.

ahoppen:
@swift-ci Please test

fwcd:
@swift-ci please test

ahoppen:
@swift-ci please test

Sources/SwiftSyntaxBuilder/gyb_generated/BuildableNodes.swift
```diff
@@ -11043,13 +10795,13 @@ public struct CatchClause: SyntaxBuildable, ExpressibleAsCatchClause {
   ///  - Initializing tokens without default text using strings
   public init(
     catchKeyword: TokenSyntax = TokenSyntax.`catch`,
-    body: ExpressibleAsCodeBlock,
-    @CatchItemListBuilder catchItemsBuilder: () -> ExpressibleAsCatchItemList? = { nil }
```
evnik, line 11047: Please keep this as `CatchItem` conforms to `HasTrailingComma`

Sources/SwiftSyntaxBuilder/gyb_generated/BuildableNodes.swift
```diff
@@ -11695,20 +11447,6 @@ public struct PrimaryAssociatedTypeClause: SyntaxBuildable, ExpressibleAsPrimary
     assert(rightAngleBracket.text == ">")
   }
 
-  /// A convenience initializer that allows:
-  ///  - Initializing syntax collections using result builders
-  ///  - Initializing tokens without default text using strings
-  public init(
-    leftAngleBracket: TokenSyntax = TokenSyntax.`leftAngle`,
-    rightAngleBracket: TokenSyntax = TokenSyntax.`rightAngle`,
-    @PrimaryAssociatedTypeListBuilder primaryAssociatedTypeListBuilder: () -> ExpressibleAsPrimaryAssociatedTypeList = { PrimaryAssociatedTypeList([]) }
```
evnik, line 11704: Please keep this as `PrimaryAssociatedType` conforms to `HasTrailingComma`

Sources/SwiftSyntaxBuilder/gyb_generated/BuildableNodes.swift
```diff
@@ -12522,20 +12250,6 @@ public struct TupleType: TypeBuildable, ExpressibleAsTupleType {
     assert(rightParen.text == ")")
   }
 
-  /// A convenience initializer that allows:
-  ///  - Initializing syntax collections using result builders
-  ///  - Initializing tokens without default text using strings
-  public init(
-    leftParen: TokenSyntax = TokenSyntax.`leftParen`,
-    rightParen: TokenSyntax = TokenSyntax.`rightParen`,
-    @TupleTypeElementListBuilder elementsBuilder: () -> ExpressibleAsTupleTypeElementList = { TupleTypeElementList([]) }
```
evnik, line 12531: Please keep this as `TupleTypeElement` conforms to `HasTrailingComma`
The same for the `FunctionType`

Sources/SwiftSyntaxBuilder/gyb_generated/BuildableNodes.swift
```diff
@@ -12834,20 +12512,6 @@ public struct GenericArgumentClause: SyntaxBuildable, ExpressibleAsGenericArgume
     assert(rightAngleBracket.text == ">")
   }
 
-  /// A convenience initializer that allows:
-  ///  - Initializing syntax collections using result builders
-  ///  - Initializing tokens without default text using strings
-  public init(
-    leftAngleBracket: TokenSyntax = TokenSyntax.`leftAngle`,
-    rightAngleBracket: TokenSyntax = TokenSyntax.`rightAngle`,
-    @GenericArgumentListBuilder argumentsBuilder: () -> ExpressibleAsGenericArgumentList = { GenericArgumentList([]) }
```
evnik, line 12843: Please keep this as `GenericArgument` conforms to `HasTrailingComma`

Sources/SwiftSyntaxBuilder/gyb_generated/BuildableNodes.swift
```diff
@@ -13252,20 +12916,6 @@ public struct TuplePattern: PatternBuildable, ExpressibleAsTuplePattern {
     assert(rightParen.text == ")")
   }
 
-  /// A convenience initializer that allows:
-  ///  - Initializing syntax collections using result builders
-  ///  - Initializing tokens without default text using strings
-  public init(
-    leftParen: TokenSyntax = TokenSyntax.`leftParen`,
-    rightParen: TokenSyntax = TokenSyntax.`rightParen`,
-    @TuplePatternElementListBuilder elementsBuilder: () -> ExpressibleAsTuplePatternElementList = { TuplePatternElementList([]) }
```
evnik, line 13261: Please keep this as `TuplePatternElement` conforms to `HasTrailingComma`

evnik (commented):
Some more cases to keep

Sources/SwiftSyntaxBuilder/gyb_generated/BuildableNodes.swift
```diff
@@ -9052,15 +8842,15 @@ public struct WhileStmt: StmtBuildable, ExpressibleAsWhileStmt {
     labelName: String?,
     labelColon: TokenSyntax? = nil,
     whileKeyword: TokenSyntax = TokenSyntax.`while`,
-    body: ExpressibleAsCodeBlock,
-    @ConditionElementListBuilder conditionsBuilder: () -> ExpressibleAsConditionElementList = { ConditionElementList([]) }
```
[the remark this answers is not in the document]
fwcd: Hm, `ConditionElement` was actually a node I intentionally excluded since it introduces a second builder closure parameter, see https://github.com/apple/swift-syntax/pull/465#discussion_r901814595

fwcd:
More of a general question: Could we put the trailing comma insertion logic into the `XyzList` initializers (in `BuildableCollectionNodes`) themselves instead of the builders? That would allow us to share the logic and remove this inconsistency between the use of builders and arrays, decoupling it from the decision which nodes are best expressed as builders. (I haven't dug too deep into this yet, so maybe this isn't really feasible after all)

evnik:
> Could we put the trailing comma insertion logic into the `XyzList` initializers (in `BuildableCollectionNodes`) themselves instead of the builders? 

We can, but this logic is explicitly placed into Result Builders to allow to define trailing commas manually through list initializers (some context could be found [here](https://github.com/apple/swift-syntax/pull/454#discussion_r894950466)). Otherwise it's going to be no way to define trivia (for example line break) for trailing comma.

ahoppen:
@evnik I disagree that syntax nodes should be initializable by result builders purely because of the fact that they have the trailing comma trait, but I agree that some of the types you mentioned should be initializable by result builders. My take on result builders is that we should reserve those for the main bodies of nodes. The collections that we allow to be initialized by result builders are usually the last nodes in the syntax node anyway (modulo some trailing closing token). Most importantly, we want to try and only have a single result builder in the initializers because the multiple result builder syntax didn’t improve legibility. 

I put in some case-by-case comments for these nodes. @fwcd If you agree, could you adjust the list accordingly before we merge this PR?


> > Could we put the trailing comma insertion logic into the `XyzList` initializers (in `BuildableCollectionNodes`) themselves instead of the builders?
> 
> We can, but this logic is explicitly placed into Result Builders to allow to define trailing commas manually through list initializers (some context could be found [here](https://github.com/apple/swift-syntax/pull/454#discussion_r894950466)). Otherwise it's going to be no way to define trivia (for example line break) for trailing comma.

What do you think about transparently inserting the trailing comma before the trailing trivia of the node’s last token (i.e. drop the trailing trivia of the last token in the node and make it the trailing comma’s trailing trivia)? I think that would cover 95% of all cases and if you wanted to add trivia before the comma, you can always specify the comma yourself.

Sources/SwiftSyntaxBuilder/gyb_generated/BuildableNodes.swift
ahoppen: ✅ Similar to `TupleTypeElement`.

Sources/SwiftSyntaxBuilder/gyb_generated/BuildableNodes.swift
ahoppen: ✅ I think result builder would make sense here because there’s no reason to assume that types only have a single generic argument.

Sources/SwiftSyntaxBuilder/gyb_generated/BuildableNodes.swift
ahoppen: ✅ I agree that using a result builder here makes sense because a `TupleType` contains multiple types by design.

Sources/SwiftSyntaxBuilder/gyb_generated/BuildableNodes.swift
ahoppen: ❌ Again, IMO the dominating use case is that there is a single primary associated type so there’s no need for a result builder.

Sources/SwiftSyntaxBuilder/gyb_generated/BuildableNodes.swift
ahoppen: ❌ Similar to `if`. The body is the primary body here.

Sources/SwiftSyntaxBuilder/gyb_generated/BuildableNodes.swift
```diff
@@ -10970,20 +10736,6 @@ public struct SwitchCaseLabel: SyntaxBuildable, ExpressibleAsSwitchCaseLabel {
     assert(colon.text == ":")
   }
 
-  /// A convenience initializer that allows:
-  ///  - Initializing syntax collections using result builders
-  ///  - Initializing tokens without default text using strings
-  public init(
-    caseKeyword: TokenSyntax = TokenSyntax.`case`,
-    colon: TokenSyntax = TokenSyntax.`colon`,
-    @CaseItemListBuilder caseItemsBuilder: () -> ExpressibleAsCaseItemList = { CaseItemList([]) }
```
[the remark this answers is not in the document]
ahoppen: 🤷🏽‍♂️/ ✅ Similar to `ClosureCaptureItemList`, having the result builder initializer doesn’t hurt.

Sources/SwiftSyntaxBuilder/gyb_generated/BuildableNodes.swift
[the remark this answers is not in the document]
ahoppen: ❌ I agree that this should not be a result builder because making the conditions because the body is without doubt the primary body of the node and making the conditions a result builder significantly reduces legibility IMHO. Also, IMO, the usual case is that `if` and `guard` statements have a single condition so it won’t benefit much from result builders.

Sources/SwiftSyntaxBuilder/gyb_generated/BuildableNodes.swift
```diff
@@ -2352,20 +2328,6 @@ public struct ClosureCaptureSignature: SyntaxBuildable, ExpressibleAsClosureCapt
     assert(rightSquare.text == "]")
   }
 
-  /// A convenience initializer that allows:
-  ///  - Initializing syntax collections using result builders
-  ///  - Initializing tokens without default text using strings
-  public init(
-    leftSquare: TokenSyntax = TokenSyntax.`leftSquareBracket`,
-    rightSquare: TokenSyntax = TokenSyntax.`rightSquareBracket`,
-    @ClosureCaptureItemListBuilder itemsBuilder: () -> ExpressibleAsClosureCaptureItemList? = { nil }
```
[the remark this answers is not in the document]
ahoppen: 🤷🏽‍♂️/✅   I don’t have strong opinions about this one. I’m fine with making it a result builder again because nobody is forced to use the result builder initializer.

Sources/SwiftSyntaxBuilder/gyb_generated/BuildableNodes.swift
ahoppen: I didn’t expect us to be using `TupleTypeElementList` for `FunctionType` as well. In that case I agree with you that it’s better to not have `TupleTypeElementList` be result builder type.

ahoppen:
@swift-ci Please test

ahoppen:
> Just raised #472, which fixes overwriting of existing trailing commas. Not sure if with that we are fine to move the logic into list initializers. 

Thanks. That PR looks good. What I thought was that the following code (basically the test case in your PR just with the newline attached to the integer literal and without the explicit comma) generates the same output as in your PR.
```swift
    let buildable = VariableDecl(letOrVarKeyword: .let, bindings: [
       PatternBinding(pattern: "a", initializer: InitializerClause(value: ArrayExpr(elementsBuilder: {
       PatternBinding(pattern: "a", initializer: InitializerClause(value: ArrayExpr(
         leftSquare: .`leftSquareBracket`.withTrailingTrivia(.newlines(1)),
         elementsBuilder: {
         for i in 1...3 {
           ArrayElement(expression: "\(i)", trailingComma: .comma)
           ArrayElement(expression: IntegerLiteralExpr(i).withTrailingTrivia(.newlines(1)))
         }
       })))
     ])
```
My idea would be that your code takes the trailing trivia of the last token in the element, removes it and attaches it to the newly inserted comma.

> Another thought I have is we might wrap result builder not related to main body into another type, so it's possible to use trailing closure for both of them.

This would require us to re-architect the way the syntax tree is layed out and I don’t think it’s worth doing that just for the SwiftSyntaxBuilder API.

ahoppen:
> > My idea would be that your code takes the trailing trivia of the last token in the element, removes it and attaches it to the newly inserted comma.
> 
> That looks interesting, but won't it be surprising? Is there a way to attach the trivia to the element itself?

I think it’s no less surprising than the fact that the result builder inserts the comma. And if you accept the fact that it inserts the comma, I think inserting it before the trailing trivia makes sense.

Trivia can only be attached to tokens, so no.

> > > Another thought I have is we might wrap result builder not related to main body into another type, so it's possible to use trailing closure for both of them.
> > 
> > 
> > This would require us to re-architect the way the syntax tree is layed out and I don’t think it’s worth doing that just for the SwiftSyntaxBuilder API.
> 
> My thought was to put those wrapper types out of syntax tree, using the fact the argument should conform to the `ExpressibleAsXyzElementList` protocol

Ah, makes sense. I’m still not sure if it solves that many real-world use cases though to warrant the additional complexity.

Out of curiosity, are you interested in using SwiftSyntaxBuilder for some application/library you are working on?

evnik:
> Out of curiosity, are you interested in using SwiftSyntaxBuilder for some application/library you are working on?

Yes, currently it's in stealth mode, I'll make sure to update `Examples/README.md` after making it public

PR #469, opened 2022-06-18
Generate `Node`, `Child`, `Classification`, `NodeSerializationCodes`, `Traits` and `Trivia`

kimdv:
https://github.com/apple/swift/pull/59707


Sources/SwiftSyntaxBuilderGeneration/gyb_generated/Trivia.swift
```diff
@@ -0,0 +1,108 @@
...
+
+let TRIVIAS: [Trivia] = [
+  Trivia(name: "Space",
+         comment: "A space ' ' character.",
+         serializationCode: 0),
+  Trivia(name: "Tab",
+         comment: "A tab '\t' character.",
+         serializationCode: 1),
+  Trivia(name: "VerticalTab",
+         comment: "A vertical tab '\v' character.",
```
kimdv, line 62: The `\v` produces a 'Unterminated string literal'.
How should we solve this? Is there a way to escape things?

ahoppen: We could use raw string literals
```suggestion
         comment: #"A vertical tab '\v' character."#,
```

Sources/SwiftSyntaxBuilderGeneration/Trivia.swift.gyb
```diff
@@ -0,0 +1,83 @@
...
+%   'name: "%s"' % trivia.name,
+%   'comment: "%s"'  % trivia.comment,
+%   'serializationCode: %d' % trivia.serialization_code
+%   ]
+%   if trivia.is_new_line:
+%     parameters += ['isNewLine: true']
+%   end
+%   if trivia.is_comment:
+%     parameters += ['isComment: true']
+%   end
```
ahoppen, line 70: You aren’t passing in `characters` or `swiftCharacters` here.

Sources/SwiftSyntaxBuilderGeneration/Trivia.swift.gyb
```diff
@@ -0,0 +1,83 @@
...
+  Trivia(${",\n         ".join(parameters)}),
+% end
+]
+
+func verifyNoDuplicateSerializationCodes(trivias: [Trivia]) {
```
ahoppen, line 75: Again, I don’t think we need this function because we are verifying that there are no duplicate serialization codes on the Python side.

ahoppen (commented):
Thanks. This is a big step to bootstrap the rest of SwiftSyntaxBuilder. I’ve got quite a few comments inline but they should all be minor.

Also, I think that all the types should be internal instead of public because they are only meant to be used in SwiftSyntaxBuilderGeneration for now.

Sources/SwiftSyntaxBuilderGeneration/Node.swift.gyb
```diff
@@ -0,0 +1,125 @@
...
+    return syntaxKind.contains("Token") || collectionElement.contains("Token")
+  }
+
+  public init(name: String,
+              description: String? = nil,
+              kind: String,
+              traits: [String]? = nil,
+              children: [Child]? = nil,
+              element: String? = nil,
+              elementName: String? = nil,
```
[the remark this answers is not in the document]
kimdv: if `elementName` is `nil` it default to `element`. 
So if we make it non-option, it will have a different behaviour. 

https://github.com/apple/swift/blob/528bd11db8a215255a3b50eca36bd72a37551f68/utils/gyb_syntax_support/Node.py#L47

Sources/SwiftSyntaxBuilderGeneration/gyb_helpers/utils.py
```diff
@@ -0,0 +1,92 @@
...
+  if child.description:
+    parameters += ['description: "%s"' % flat_documentation(child.description)]
+
+  if child.is_optional:
+    parameters += ['isOptional: true']
+
+  if child.token_choices:
+    mapped_token_choices = child.token_choices
+    if child.token:
+      mapped_token_choices.remove(child.token)
```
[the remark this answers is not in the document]
kimdv: I remove it here, and then add in the init, to get the exact same behaviour.

The order of the tokens matter and it will add the `syntaxKind` to the `tokenChoices` in the `Child.init` and if we not remove it here, we will have it twice in the list.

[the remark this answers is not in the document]
kimdv: And updated `Child.init` method

kimdv:
@ahoppen I've pushed the changes and suggestions you had. 
Left two comments open for discussion.

Sources/SwiftSyntaxBuilderGeneration/gyb_generated/AvailabilityNodes.swift
```diff
@@ -0,0 +1,103 @@
...
+let AVAILABILITY_NODES: [Node] = [
+  Node(name: "AvailabilitySpecList",
+       kind: "SyntaxCollection",
+       element: "AvailabilityArgument"),
+
+  Node(name: "AvailabilityArgument",
+       description: "A single argument to an `@available` argument like `*`, `iOS 10.1`,or `message: \"This has been deprecated\"`.",
+       kind: "Syntax",
+       children: [
+        Child(name: "Entry",
```
ahoppen, line 24: This is currently indented by one space, it should probably be two.

Sources/SwiftSyntaxBuilderGeneration/gyb_helpers/utils.py
```diff
@@ -0,0 +1,101 @@
...
+
+  if child.is_indented:
+    parameters += ['isIndented: true']
+
+  if child.requires_leading_newline:
+    parameters += ['requiresLeadingNewline: true']
+
+
+#  spacing = ' ' * spaces
+#  return 'Child(%s)' % (',\n%s' % spacing).join(parameters)
```
ahoppen, line 92: Stale comment?

Sources/SwiftSyntaxBuilderGeneration/gyb_generated/Trivia.swift
```diff
@@ -0,0 +1,141 @@
...
+          Character(#"\f"#)
+         ],
+         swiftCharacters: [
+          Character(#"\u{240C}"#)
+         ]),
+  Trivia(name: "Newline",
+         comment: #"A newline '\n' character."#,
+         serializationCode: 4,
+         characters: [
+          Character(#"\n"#)
```
ahoppen, line 94: I think this should be indented by 2 spaces instead of 1.

Sources/SwiftSyntaxBuilderGeneration/Node.swift.gyb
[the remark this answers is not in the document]
ahoppen: You’re right. I thought you always pass in a string literal and missed that 
```swift
parameters += ['elementName: "%s"' % node.element_name]
```
is guarded by an `if`

ahoppen (commented):
I think you missed one comment from the first review. I’ve got a couple more minor comments. I think this is almost ready to be merged.

Sources/SwiftSyntaxBuilderGeneration/gyb_generated/AvailabilityNodes.swift
kimdv: I used the indentation from Xcode, and afterwards adjusted the `gyb` to it.

It might be something in Xcode then 🤷‍♂️

Will change. 
Should the wrong indent bug be reported ?

ahoppen: If you have a reproducer in Xcode, it would be great if you could file a bug report on http://feedbackassistant.apple.com.

kimdv:
https://github.com/apple/swift/pull/59707

@swift-ci please test

Sources/SwiftSyntaxBuilderGeneration/gyb_generated/AvailabilityNodes.swift
kimdv: Openend: FB10505542

ahoppen: Thanks for filing the bug report 👍

PR #473, opened 2022-06-25
Automatically generate leading/trailing space trivia for tokens in `SyntaxFactory`

fwcd:
In an effort to clean up the `TokensFile` template in `SwiftSyntaxBuilderGeneration`, this branch moves the generation of leading/trailing space trivia around tokens into `SyntaxFactory`. In other words, calls like

```
SyntaxFactory.makeSwitchKeyword()
```

will now automatically generate a trailing space (although this is still overridable by explicitly passing `trailingTrivia: []` as a parameter).

cc @ahoppen

### DIFF
--- a/Sources/SwiftSyntaxBuilderGeneration/Tokens.swift.gyb
+++ b/Sources/SwiftSyntaxBuilderGeneration/Tokens.swift.gyb
@@ -28,8 +28,6 @@ public class Token {
   public let text: String?
   public let classification: String
   public let isKeyword: Bool
-  public let requiresLeadingSpace: Bool
-  public let requiresTrailingSpace: Bool
 
   public var swiftKind: String {
     let name = self.name
@@ -41,7 +39,7 @@ public class Token {
     }
   }
 
-  public init(name: String, kind: String, serializationCode: Int, unprefixedKind: String? = nil, text: String? = nil, classification: String = "None", isKeyword: Bool = false, requiresLeadingSpace: Bool = false, requiresTrailingSpace: Bool = false) {
+  public init(name: String, kind: String, serializationCode: Int, unprefixedKind: String? = nil, text: String? = nil, classification: String = "None", isKeyword: Bool = false) {
     self.name = name
     self.kind = kind
     self.serializationCode = serializationCode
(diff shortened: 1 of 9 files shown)


Sources/SwiftSyntaxBuilderGeneration/Tokens.swift.gyb
```diff
@@ -28,8 +28,6 @@ public class Token {
   public let text: String?
   public let classification: String
   public let isKeyword: Bool
-  public let requiresLeadingSpace: Bool
-  public let requiresTrailingSpace: Bool
```
fwcd, line 32: Since the `SyntaxFactory` gyb-template takes care of this, I have removed these (no longer needed) attributes from `Tokens`. Is this fine?

Tests/SwiftSyntaxBuilderTest/VariableTests.swift
```diff
@@ -25,7 +25,7 @@ final class VariableTests: XCTestCase {
     }
 
     let syntax = buildable.buildSyntax(format: Format(), leadingTrivia: leadingTrivia)
-    XCTAssertEqual(syntax.description, "␣var d: [String: Int] = [:]")
+    XCTAssertEqual(syntax.description, "␣var d: [String: Int] = [: ]")
```
fwcd, line 28: The empty dictionary literal is now generated as `[: ]` since the colon token seems to require a trailing space. I don't think this should be an issue, but perhaps worth pointing out.

kimdv (approved):

kimdv:
@swift-ci please test

fwcd (commented):
A few notes on the implementation below. The `open` keyword also still needs an explicit trailing trivia call (in [`TokensFile.swift:100`](https://github.com/fwcd/swift-syntax/blob/109949e40f2dc43367e6e10c158d81894fd4cb20/Sources/SwiftSyntaxBuilderGeneration/Templates/TokensFile.swift#L100)) since contextual keywords as generated in [`makeContextualKeyword`](https://github.com/fwcd/swift-syntax/blob/109949e40f2dc43367e6e10c158d81894fd4cb20/Sources/SwiftSyntax/gyb_generated/SyntaxFactory.swift#L5952-L5960) apparently do not require a trailing space.

fwcd:
@swift-ci please test

fwcd:
@swift-ci please test

Tests/SwiftSyntaxBuilderTest/VariableTests.swift
evnik: You should be able to fix it in the [`DictionaryExprConvenienceInitializers.swift`](https://github.com/apple/swift-syntax/blob/0ec7b5a7f96aa652aac910767b4e4371f6cda488/Sources/SwiftSyntaxBuilder/DictionaryExprConvenienceInitializers.swift#L26)

fwcd: Good catch, thanks!

fwcd:
@swift-ci please test

Sources/SwiftSyntaxBuilderGeneration/Tokens.swift.gyb
ahoppen: Yes 👍

ahoppen (approved):
Wow, this is so much simpler than what we had before 😍

PR #474, opened 2022-06-25
Add `Trivia.isEmpty`

fwcd:
A small addition in preparation of a larger PR regarding leading/trailing trivia generation.

PR #475, opened 2022-06-25
Add ability to attach leading trivia directly to buildable nodes

fwcd:
Based on #474.

This branch adds the ability to attach leading trivia directly to `BuildableNodes` as a parameter/attribute (currently this can only be done by passing it as a parameter directly to the build syntax method), thereby addressing the [points outlined here](https://github.com/apple/swift-syntax/pull/460#discussion_r891749401). This is particularly convenient in `TokensFile`, where we can now attach doc comments directly to the nodes:

```swift
ExtensionDecl(
  leadingTrivia: .newlines(1) + .docLineComment("/// Namespace for commonly used tokens with default trivia.") + .newlines(1),
  extendedType: "TokenSyntax"
)
```

rather than having to modify the first token (as done currently):

```swift
ExtensionDecl(
  modifiers: [TokenSyntax.public.withLeadingTrivia(.newlines(1) + .docLineComment("/// Namespace for commonly used tokens with default trivia.") + .newlines(1))],
  extendedType: "TokenSyntax"
)
```

thereby making the DSL more readable and more aligned with the actual generated code.

cc @ahoppen @kimdv


fwcd:
@swift-ci please test

kimdv:
I like the API. Clean and easy to understand 👏

Was wondering if it would make sense to add  a single test that tests if we combine the leading `leadingTrivia` with a leading `trivia` on a modifier etc?
Just to ensure that works and will not break 😁

``` swift
VariableDecl(
       leadingTrivia: .newlines(1) + .docLineComment("/// The `open` contextual token") + .newlines(1),
       modifiers: [TokenSyntax.static.withLeadingTrivia(.newlines(1) )],
       letOrVarKeyword: .var
     )
```

fwcd:
@swift-ci please test

Sources/SwiftSyntaxBuilder/BuildableNodes.swift.gyb
```diff
@@ -37,18 +37,23 @@ public struct ${type.buildable()}${conformance_clause(conformances)} {
   let ${child.name()}: ${child.type().buildable()}
 %   end
 
+  /// The leading trivia attached to this syntax once built.
```
ahoppen, line 40: I got a little confused what this `leadingTrivia` and the `additionalLeadingTrivia` were used for. What do you think about adding a comment like
```swift
/// This is typically used to add comments to syntax nodes
```

to this member and something like

```swift
/// `leadingTrivia` is typically used to add the correct amount of indentation to this node.
```
to the `build*(` method.

Sources/SwiftSyntaxBuilder/BuildableNodes.swift.gyb
```diff
@@ -95,28 +100,31 @@ public struct ${type.buildable()}${conformance_clause(conformances)} {
...
-  func build${type.base_name()}(format: Format, leadingTrivia: Trivia? = nil) -> ${type.syntax()} {
+  func build${type.base_name()}(format: Format, leadingTrivia additionalLeadingTrivia: Trivia? = nil) -> ${type.syntax()} {
     let result = SyntaxFactory.make${type.base_name()}(
       ${',\n      '.join(['%s: %s' % (child.name(), child.generate_expr_build_syntax_node(child.name(), 'format')) for child in children])}
     )
-    if let leadingTrivia = leadingTrivia {
-      return result.withLeadingTrivia(leadingTrivia + (result.leadingTrivia ?? []))
-    } else {
+    let combinedLeadingTrivia = leadingTrivia + (additionalLeadingTrivia ?? []) + (result.leadingTrivia ?? [])
+    if combinedLeadingTrivia.isEmpty {
```
ahoppen, line 118: Is this `if`-statement really necessary or can we just 
```swift
return result.withLeadingTrivia(combinedLeadingTrivia)
```

ahoppen (commented):
Yes, this really improves readability of TokensFile.swift 👍

Sources/SwiftSyntaxBuilder/BuildableNodes.swift.gyb
fwcd: Yeah, we could probably do that, I wasn't sure whether there was some internal overhead to attaching potentially empty trivia to every node.

ahoppen: I don’t think we should need to worry about micro-performance-optimization in SwiftSyntaxBuilder.

Sources/SwiftSyntaxBuilder/BuildableNodes.swift.gyb
```diff
@@ -95,28 +100,31 @@ public struct ${type.buildable()}${conformance_clause(conformances)} {
...
   ) {
     self.init(
+      leadingTrivia: leadingTrivia,
       ${',\n      '.join(delegated_init_args)}
     )
   }
 %   end
 
-  func build${type.base_name()}(format: Format, leadingTrivia: Trivia? = nil) -> ${type.syntax()} {
+  func build${type.base_name()}(format: Format, leadingTrivia additionalLeadingTrivia: Trivia? = nil) -> ${type.syntax()} {
```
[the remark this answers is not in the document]
ahoppen: I would suggest that we keep it as-is for now and could consider refactoring it to be non-optional in a follow-up PR.

fwcd:
@swift-ci please test

PR #476, opened 2022-06-25
Properties returning collection trivia with count equal to 1

evnik:


Sources/SwiftSyntax/Trivia.swift.gyb
```diff
@@ -99,19 +99,24 @@ public struct Trivia {
...
-    /// Return a piece of trivia for some number of '${joined}' characters.
-    public static func ${trivia.lower_name}s(_ count: Int) -> Trivia {
-      return [.${trivia.lower_name}s(count)]
-    }
+  /// Returns a piece of trivia for some number of '${joined}' characters.
+  public static func ${trivia.lower_name}s(_ count: Int) -> Trivia {
+    return [.${trivia.lower_name}s(count)]
+  }
+
+  /// Gets a piece of trivia for '${joined}' characters.
```
ahoppen, line 111: I think the current comment isn’t grammatically correct.

```suggestion
  /// Gets a piece of trivia with a single '${joined}' character.
```

Sources/SwiftSyntax/Trivia.swift.gyb
```diff
@@ -99,19 +99,24 @@ public struct Trivia {
...
+  public static var ${trivia.lower_name}: Trivia {
+    return [.${trivia.lower_name}s(1)]
```
ahoppen, line 113: I don’t even think we need the `[` and `]`. Without those we’re just calling into the function defined above.

Tests/SwiftSyntaxBuilderTest/VariableTests.swift
```diff
@@ -30,15 +30,13 @@ final class VariableTests: XCTestCase {
...
-        for i in 1...3 {
-          ArrayElement(
-            expression: IntegerLiteralExpr(i),
-            trailingComma: .comma.withoutTrailingTrivia().withTrailingTrivia(.newlines(1)))
-        }
-      })))
+      PatternBinding(pattern: "a", initializer: InitializerClause(
+        value: ArrayExpr(leftSquare: .`leftSquareBracket`.withTrailingTrivia(.newline)) {
+          for i in 1...3 {
+            ArrayElement(expression: "\(i)", trailingComma: .comma.withoutTrailingTrivia().withTrailingTrivia(.newline))
```
ahoppen, line 36: I don’t think this transformation is technically correct. The old example generates an `IntegerLiteralExpr` while this creates an `IdentifierExpr` (which conforms to `ExpressibleAsIdentifierExpr`). Identifiers are e.g. variable names, which shouldn’t consist purely of digits. 

While both generate the same source code, the SwiftSyntax tree with the `IdentifierExpr` shouldn’t technically be valid.

ahoppen (commented):
Thanks. This looks good. I’ve got a couple nitpicks inline.

kimdv:
@swift-ci please test

ahoppen:
@swift-ci Please test

PR #477, opened 2022-06-28
Migrate examples to Swift snippets

fwcd:
[This recently merged proposal](https://github.com/apple/swift-evolution/blob/main/proposals/0356-swift-snippets.md) will bring native support for running small scripts (snippets) in the context of the package directly via SPM (`swift run [snippet]`). Since the examples from this repo effectively serve the same purpose, this branch renames the `Examples` folder to `Snippets` to make them discoverable by SPM. Additionally, it adds a few minor tweaks (cleaning up the paths in the `README.md` and using trailing closures in the syntax builder example).


kimdv (approved):
Cool! 🎉

fwcd:
I have rebased and tested this with a recent toolchain, `swift build` now builds the snippets too. The only thing left to do would be to update the build script, so the snippets are built/checked in CI too. Is there a way to do so without having to explicitly list all snippets in `build_command`?

ahoppen:
My idea would be to change `Builder.build()` to build all products (i.e. remove the `--product` option from the `swift build` invocation). That could give us a performance improvement because all products are built in parallel and also builds the snippets.

fwcd:
Since this PR doesn't really regress anything regarding the snippets/examples, wdyt about moving the updates to the build script to a future PR once we figure out #531?

@swift-ci please test

ahoppen (approved):
Sounds good to me 👍

fwcd:
The CI failure looks unrelated, if I read this correctly? I could try again later.

ahoppen:
I think CI got confused because you mentioned another PR in the same comment that you triggered CI, so it tried to test this PR together with the mentioned PR.

ahoppen:
@swift-ci Please test

PR #478, opened 2022-06-29
Convenience initializer for TupleExprElement

evnik:


Sources/SwiftSyntaxBuilder/TupleExprElementConvenienceInitializers.swift
```diff
@@ -0,0 +1,22 @@
...
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
+public extension TupleExprElement {
+  /// A convenience initializer that allows passing in label as an optional string.
+  /// The presence of the colon will be inferred based on the presence of the label.
+  init(_ label: String? = nil, expression: ExpressibleAsExprBuildable) {
```
ahoppen, line 18: I don’t have super strong opinions but I think I would prefer `label` to be a labelled argument instead of an unlabelled one. What was your reasoning for making it unlabeled?

evnik: Just wanted to make the signature different to the existing one, but I'm fine to make it labelled

kimdv:
@swift-ci Please test

ahoppen (approved):
Thanks that you keep adding these little convenience utilities. 👍

PR #479, opened 2022-06-30
Generate `@resultbuilder` with SwiftSyntaxBuilder

kimdv:
Depends on
- [x]  #480 
- [x] #485 

This is still WIP, and there is still some way to go! 

**With gyb**
```swift
@resultBuilder
public struct CodeBlockItemListBuilder {

  /// The type of individual statement expressions in the transformed function,
  /// which defaults to Component if buildExpression() is not provided.
  public typealias Expression = ExpressibleAsCodeBlockItem

  /// The type of a partial result, which will be carried through all of the
  /// build methods.
  public typealias Component = [ExpressibleAsCodeBlockItem]

  /// The type of the final returned result, which defaults to Component if
  /// buildFinalResult() is not provided.
  public typealias FinalResult = CodeBlockItemList

  /// Required by every result builder to build combined results from
  /// statement blocks.
  public static func buildBlock(_ components: Component...) -> Component {
    return components.flatMap { $0 }
  }

  public static func buildExpression(_ expression: Expression) -> Component {
    return [expression]
  }
  
  /// Add all the elements of `expression` to this result builder, effectively flattening them.
  public static func buildExpression(_ expression: ExpressibleAsCodeBlockItemList) -> Component {
    return expression.createCodeBlockItemList().elements
  }
  
  /// Enables support for `if` statements that do not have an `else`.
  public static func buildOptional(_ component: Component?) -> Component {
    return component ?? []
  }

  /// With buildEither(second:), enables support for 'if-else' and 'switch'
  /// statements by folding conditional results into a single result.
  public static func buildEither(first component: Component) -> Component {
    return component
  }

  /// With buildEither(first:), enables support for 'if-else' and 'switch'
  /// statements by folding conditional results into a single result.
  public static func buildEither(second component: Component) -> Component {
    return component
  }

  /// Enables support for 'for..in' loops by combining the
  /// results of all iterations into a single result.
  public static func buildArray(_ components: [Component]) -> Component {
    return components.flatMap { $0 }
  }

  /// If declared, this will be called on the partial result of an 'if
  /// #available' block to allow the result builder to erase type
  /// information.
  public static func buildLimitedAvailability(_ component: Component) -> Component {
    return component
  }

  /// If declared, this will be called on the partial result from the outermost
  /// block statement to produce the final returned result.
  public static func buildFinalResult(_ component: Component) -> FinalResult {
    return .init(component.map { $0.createCodeBlockItem() })
  }
}
```

**With SwiftSyntaxBuilderGeneration**
```swift
@resultBuilder
public struct CodeBlockItemListBuilder{
  
/// Required by every result builder to build combined results from
/// statement blocks.
public static func buildBlock(_ components: Component...)-> Component{
    return components.flatMap()
  }
  
public static func buildExpression(_ expression: Expression)-> Component{
    return [expression]
  }
  
/// Add all the elements of `expression` to this result builder, effectively flattening them.
public static func buildExpression(_ expression: ExpressibleAsCodeBlockItemList)-> Component{
    return components.flatMap()
  }
  
/// Enables support for `if` statements that do not have an `else`.
public static func buildOptional(_ component: Component?)-> Component{
    return component??[]
  }
  
/// With buildEither(second:), enables support for 'if-else' and 'switch'
/// statements by folding conditional results into a single result.
public static func buildEither(firstcomponent: Component)-> Component{
    return component
  }
  
/// With buildEither(first:), enables support for 'if-else' and 'switch'
/// statements by folding conditional results into a single result.
public static func buildEither(secondcomponent: Component)-> Component{
    return component
  }
  
/// Enables support for 'for..in' loops by combining the
/// results of all iterations into a single result.
public static func buildArray(_ components: [Component])-> Component{
    return .component
  }
  
/// If declared, this will be called on the partial result of an 'if'
/// #available' block to allow the result builder to erase type
/// information.
public static func buildLimitedAvailability(_ component: Component)-> Component{
    return component
  }
  
/// If declared, this will be called on the partial result from the outermost
/// block statement to produce the final returned result.
public static func buildFinalResult(_ component: Component)-> Component{
  }
}
```


Sources/SwiftSyntaxBuilderGeneration/Templates/ResultBuildersFile.swift
```diff
@@ -0,0 +1,355 @@
...
+            .newlines(1),
+            .docLineComment("/// The type of individual statement expressions in the transformed function,"),
+            .newlines(1),
+            .docLineComment("/// which defaults to Component if buildExpression() is not provided."),
+            .newlines(1)
+        ],
+                      attributes: nil,
+                      modifiers: [TokenSyntax.public],
+                      identifier: TokenSyntax.identifier("Expression"),
+                      initializer: TypeInitializerClause(value: node.expressibleAs))
```
fwcd, line 48: What do you think about indenting these (and the other) decls as

```suggestion
        TypealiasDecl(
          leadingTrivia: [
            .newlines(1),
            .docLineComment("/// The type of individual statement expressions in the transformed function,"),
            .newlines(1),
            .docLineComment("/// which defaults to Component if buildExpression() is not provided."),
            .newlines(1)
          ],
          attributes: nil,
          modifiers: [TokenSyntax.public],
          identifier: TokenSyntax.identifier("Expression"),
          initializer: TypeInitializerClause(value: node.expressibleAs)
        )
```

so the indentation doesn't jump back and forth?

kimdv: The `attributes`, `modifiers` have actually the correct indentation they are under the `leadingTrivia`.
It is the array that is wrongly indented by Xcode. 

We had the same issue here: https://github.com/apple/swift-syntax/pull/469#discussion_r908564456

Properly a bug in Xcode

kimdv: We could use a syntax like 


```
        TypealiasDecl(leadingTrivia:
                        [
                          .newlines(1),
                          .docLineComment("/// The type of individual statement expressions in the transformed function,"),
                          .newlines(1),
                          .docLineComment("/// which defaults to Component if buildExpression() is not provided."),
                          .newlines(1)
                        ],
                      attributes: nil,
                      identifier: TokenSyntax.identifier("Expression"),
                      initializer: TypeInitializerClause(value: node.expressibleAs))
```

fwcd: Yeah, that makes sense. Though I think putting the parameters on a new indentation level would still benefit readability, especially with long lines (like the doc comments).

kimdv: I saw you did the same in `TokensFile`. 
I think we should do the same, so will change 💪

kimdv: Something like this @fwcd ?

fwcd: Yes, that looks great!

Sources/SwiftSyntaxBuilder/generated/ResultBuilders.swift
```diff
@@ -3389,4 +3389,3 @@ public extension AvailabilitySpecList {
     self = itemsBuilder()
   }
 }
-
```
evnik, line 3392: Now the only difference is this blank line, this is pretty impressive! 🚀

Sources/SwiftSyntaxBuilderGeneration/Templates/ResultBuildersFile.swift
```diff
@@ -0,0 +1,387 @@
...
+    let type = SyntaxBuildableType(syntaxKind: node.syntaxKind)
+    let elementType = node.collectionElementType
+    
+    StructDecl(
+      attributes: TokenSyntax.identifier("@resultBuilder").withLeadingTrivia(.newlines(1)).withTrailingTrivia(.newlines(1)),
+      modifiers: TokenSyntax.public,
+      identifier: type.syntaxKind + "Builder") {
+        TypealiasDecl(
+          leadingTrivia: [
+            .newlines(1),
```
fwcd, line 32: I think we can shorten these to `.newline`

Sources/SwiftSyntaxBuilderGeneration/Templates/ResultBuildersFile.swift
```diff
@@ -0,0 +1,387 @@
...
+            .docLineComment("/// Required by every result builder to build combined results from"),
+            .newlines(1),
+            .docLineComment("/// statement blocks."),
+            .newlines(1)],
+          modifiers: [TokenSyntax.public, TokenSyntax.static],
+          identifier: .identifier("buildBlock"),
+          signature: FunctionSignature(
+            input: ParameterClause {
+              FunctionParameter(
+                firstName: TokenSyntax.wildcard,
```
fwcd, line 82: `TokenSyntax` can be omitted in parameters that only expect `TokenSyntax`, i.e. we can use implicit member expressions, e.g. `.wildcard` here.

Sources/SwiftSyntaxBuilderGeneration/Templates/ResultBuildersFile.swift
```diff
@@ -0,0 +1,387 @@
...
+                secondName: TokenSyntax.identifier("expression"),
+                colon: TokenSyntax.colon,
+                type: "Expression")
+            },
+            output: "Component")) {
+              ReturnStmt(
+                expression: ArrayElement(
+                  expression: IdentifierExpr("expression")
```
fwcd, line 123: I believe we can omit `IdentifierExpr`, the string literal should suffice for the compiler to pick up on the right `ExpressibleAs` chain IIRC.

fwcd (commented):
A few style nitpicks, but otherwise this looks good! 👍

Package.swift
```diff
@@ -79,6 +79,7 @@ let package = Package(
       dependencies: ["SwiftSyntax"],
       exclude: [
         "gyb_helpers",
+        "BuildableNodes.swift.gyb",
```
fwcd, line 82: This can probably be removed since we already merged #555 

```suggestion
```

ahoppen:
rdar://99492975

Sources/SwiftSyntaxBuilder/generated/ResultBuildersFile.swift
```diff
@@ -16,73 +17,61 @@ import SwiftSyntax
 
 @resultBuilder
 public struct CodeBlockItemListBuilder {
-
```
kimdv, line 19: Is there a way with the new API to add new lines, without should adding a empty line at the beginning of each `Decl`

kimdv:
@swift-ci please test

CodeGeneration/Sources/generate-swiftsyntaxbuilder/ResultBuildersFile.swift
```diff
@@ -0,0 +1,194 @@
...
+          public typealias FinalResult = \(type.buildable)
+          """
+        )
+        
+        FunctionDecl(
+          """
+          /// Required by every result builder to build combined results from
+          /// statement blocks.
+          public static func buildBlock(_ components: Component...) -> Component {
+            return components.flatMap { $0 }
```
fwcd, line 62: Since all of these functions return single expressions, we could omit the `return` here (and in the other builder functions).

CodeGeneration/Sources/generate-swiftsyntaxbuilder/ResultBuildersFile.swift
```diff
@@ -0,0 +1,194 @@
...
+          /// information.
+          public static func buildLimitedAvailability(_ component: Component) -> Component {
+            return component
+          }
+          """
+        )
+        
+        FunctionDecl(
+          leadingTrivia: [
+            .newlines(1),
```
fwcd, line 138: `.newlines(1)` be shortened to `.newline` IIRC

CodeGeneration/Sources/generate-swiftsyntaxbuilder/ResultBuildersFile.swift
```diff
@@ -0,0 +1,194 @@
...
+                  let lastIndex = component.count - 1
+                  """
+                )
+                
+                ReturnStmt(
+                  """
+                  return .init(component.enumerated().map({ index, source in
+                    let element = source.create\(elementType.buildableBaseName)()
+                    return index < lastIndex ? element.ensuringTrailingComma() : element
+                  }))
```
fwcd, line 172: If I'm not mistaken, we should be able to use a trailing closure here

```suggestion
                  return .init(component.enumerated().map { index, source in
                    let element = source.create\(elementType.buildableBaseName)()
                    return index < lastIndex ? element.ensuringTrailingComma() : element
                  })
```

Sources/SwiftSyntaxBuilder/ClosureParamConvenienceInitializer.swift
```diff
@@ -0,0 +1,23 @@
...
+
+extension ClosureParam {
+  public init (leadingTrivia: Trivia = [], unexpectedBeforeName: ExpressibleAsUnexpectedNodes? = nil, name: String, unexpectedBetweenNameAndTrailingComma: ExpressibleAsUnexpectedNodes? = nil, trailingComma: Token? = nil) {
+    self.init(leadingTrivia: leadingTrivia,
+              unexpectedBeforeName: unexpectedBeforeName,
+              name: Token(tokenSyntax: .identifier(name)),
+              unexpectedBetweenNameAndTrailingComma: unexpectedBetweenNameAndTrailingComma,
+              trailingComma: trailingComma)
+  }
+}
```
fwcd, line 23: Just curious, do we use this anywhere? Also wdyt about removing the unexpected nodes here (since they add noise, may complicate overload resolution and are probably not used by anyone constructing nodes with the convenience initializer)?

kimdv: Left over, sorry about that

CodeGeneration/Sources/generate-swiftsyntaxbuilder/ResultBuildersFile.swift
ahoppen: I have a general preference for explicit `return`, but that might be personal. I gave up actively fighting against implicit returns 😆

ahoppen (approved):
Wohoo. This was so easy to review.

ahoppen:
@swift-ci Please test

CodeGeneration/Sources/generate-swiftsyntaxbuilder/ResultBuildersFile.swift
kimdv: Personally I also like returns in methods as it read better IMO 😬

kimdv:
@swift-ci please test

PR #481, opened 2022-07-02
Add `--gyb-only` flag to build script's `generate-source-code`

fwcd:
When working on `SwiftSyntaxBuilderGeneration`, it can occasionally be useful to only generate `gyb` templates (skipping `SwiftSyntaxBuilderGeneration`'s Swift templates). Previously this could be achieved with `--degyb-only`, with the refactoring of the build script, however, this option was removed in favor of `generate-source-code`, which performs all forms of code generation at once. While that is what most people want to do most of the time, having the option to only perform gyb generation can still be a useful tool sometimes, hence this new flag.

cc @kimdv @ahoppen

### DIFF
--- a/build-script.py
+++ b/build-script.py
@@ -656,15 +656,16 @@ def generate_source_code_command(args):
         printerr("FAIL: Generating .gyb files failed")
         printerr("Executing: %s" % " ".join(e.cmd))
         printerr(e.output)
-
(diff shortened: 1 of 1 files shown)


fwcd:
@swift-ci please test

ahoppen (approved):
SGTM 👍

PR #482, opened 2022-07-02
Generate `SYNTAX_NODES` in `SwiftSyntaxBuilderGeneration`

fwcd:
A small patch that generates the Swift declarations for `SYNTAX_NODES` (thereby also moving [this declaration](https://github.com/apple/swift-syntax/blob/674aa55f9386392428c74d42a54fb5fa07c56676/Sources/SwiftSyntaxBuilderGeneration/Templates/ResultBuildersFile.swift#L17-L26) into its own file for #479). These are required by many `gyb` templates in `SwiftSyntaxBuilder` and let us move forward with the migration to Swift templates.

The definition of `SYNTAX_NODES` is analogous to [the corresponding declaration](https://github.com/apple/swift/blob/198b9746221cbccd7c525b1e8de18338f8336e51/utils/gyb_syntax_support/__init__.py#L21-L23) in `gyb_syntax_support`.

cc @ahoppen @kimdv

### DIFF
--- a/Sources/SwiftSyntaxBuilderGeneration/SyntaxNodes.swift
+++ b/Sources/SwiftSyntaxBuilderGeneration/SyntaxNodes.swift
@@ -1,0 +1,21 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2022 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+let SYNTAX_NODES: [Node] = COMMON_NODES
+  + EXPR_NODES
+  + DECL_NODES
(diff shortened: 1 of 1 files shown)


fwcd:
@swift-ci please test

kimdv:
> A small note on the implementation: Instead of manually concatenating `EXPR_NODES`, `DECL_NODES` etc. we use gyb to generate all of these nodes again. While this causes us to generate new instances for all of these nodes, it eliminates the redundancy of having to manually enumerate all of the different node kinds (though feel free to comment on whether you would prefer the latter approach instead).

Personally I like merge of the arrays better, otherwise we have the same node twice. 
So I would prefer the 

```
let SYNTAX_NODES: [Node] = COMMON_NODES +
    EXPR_NODES +
    DECL_NODES +
    ATTRIBUTE_NODES +
    STMT_NODES +
    GENERIC_NODES +
    TYPE_NODES +
    PATTERN_NODES +
    AVAILABILITY_NODES
```

fwcd:
I don't have a strong opinion here, so if you prefer the concatenation, let's do that instead 👍

fwcd:
@swift-ci please test

kimdv (approved):

ahoppen (approved):
LGTM

PR #483, opened 2022-07-03
Add convenience initializers that conforms to `ExpressibleAsArrayExpr`

kimdv:
Part of #479

### DIFF
--- a/Sources/SwiftSyntaxBuilder/gyb_generated/BuildableCollectionNodes.swift
+++ b/Sources/SwiftSyntaxBuilder/gyb_generated/BuildableCollectionNodes.swift
@@ -171,11 +171,6 @@ public struct ArrayElementList: ExpressibleByArrayLiteral, SyntaxBuildable, Expr
   }
 }
 
-extension Array: ExpressibleAsArrayElementList where Element == ExpressibleAsArrayElement {
-  public func createArrayElementList() -> ArrayElementList {
-    return ArrayElementList(self)
-  }
-}
 
 /// `DictionaryElementList` represents a collection of `DictionaryElement`s.
 public struct DictionaryElementList: ExpressibleByArrayLiteral, SyntaxBuildable, ExpressibleAsDictionaryElementList {
--- a/Sources/SwiftSyntaxBuilder/generated/ExpressibleAsProtocols.swift
+++ b/Sources/SwiftSyntaxBuilder/generated/ExpressibleAsProtocols.swift
@@ -142,8 +142,14 @@ public extension ExpressibleAsPoundColumnExpr {
 public protocol ExpressibleAsTupleExprElementList {
   func createTupleExprElementList() -> TupleExprElementList
 }
-public protocol ExpressibleAsArrayElementList {
+public protocol ExpressibleAsArrayElementList: ExpressibleAsArrayExpr {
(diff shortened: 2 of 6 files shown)


kimdv:
@swift-ci please test

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test

ahoppen (dismissed):

Sources/SwiftSyntaxBuilder/ArrayElementListConvenienceInitializers.swift
```diff
@@ -0,0 +1,19 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
+extension ArrayElementList: ExpressibleAsArrayExpr {
```
ahoppen, line 15: Do we really need to hand-craft a convenience initializer here? Can’t we just add this to `ExpressibleAsConformances`?

Sources/SwiftSyntaxBuilder/ArrayElementConvenienceInitializers.swift
```diff
@@ -0,0 +1,19 @@
...
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
+extension ArrayElement: ExpressibleAsArrayExpr {
+  public func createArrayExpr() -> ArrayExpr {
+    ArrayExpr(elements: ArrayElementList([self]))
```
ahoppen, line 17: I’m not entirely sure whether I am too fond of this. I think we should strive to make `ArrayElement` as transparent as possible (i.e. make sure that we can implicitly convert `Expr` to `ArrayElement` but I would prefer to keep an explicit wrap to `ArrayExpr` because it is modifying the semantic meaning.

kimdv:
@swift-ci please test

kimdv:
@ahoppen, there was to much spacing so cleaned a bit too 🧹

ahoppen (approved):

Sources/SwiftSyntaxBuilder/gyb_generated/BuildableCollectionNodes.swift
```diff
@@ -171,11 +171,6 @@ public struct ArrayElementList: ExpressibleByArrayLiteral, SyntaxBuildable, Expr
   }
 }
 
-extension Array: ExpressibleAsArrayElementList where Element == ExpressibleAsArrayElement {
-  public func createArrayElementList() -> ArrayElementList {
-    return ArrayElementList(self)
-  }
-}
```
fwcd, line 178: Hm, this removes the `Array: ExpressibleAsArrayElementList` conformance since `ExpressibleAsArrayElementList` now has a superprotocol (see #528 for details why generating `Array` conformances for these hierarchies is hard).

But I guess we can just use `ArrayElementList`s instead of `ArrayExpr`s in those places now.

PR #485, opened 2022-07-03
Port `SyntaxBuildableWrappers` to Swift

fwcd:
Based on #482.

This branch ports the `SyntaxBuildableWrappers` from `gyb_helpers` to Swift for use in `SwiftSyntaxBuilderGeneration`. The API is kept as close as possible to the original (with slight modifications to make it more swifty), hoping that it will make the translation of gyb templates to Swift as seamless as possible.

Currently WIP since only a subset of the API surface has been ported.

### DIFF
--- a/Sources/SwiftSyntaxBuilderGeneration/Node.swift
+++ b/Sources/SwiftSyntaxBuilderGeneration/Node.swift
@@ -28,18 +28,6 @@ class Node {
   let elementsSeparatedByNewline: Bool
   let collectionElement: String
 
-  var baseType: String {
-    if self.baseKind == "SyntaxCollection" {
-      return "Syntax"
-    } else {
-      return kindToType(kind: baseKind)
-    }
-  }
-
-  var collectionElementType: String {
-    return kindToType(kind: collectionElement)
-  }
-
   /// Returns `True` if this node declares one of the base syntax kinds.
   var isBase: Bool {
     return SyntaxBaseKinds.contains(syntaxKind)
--- a/Sources/SwiftSyntaxBuilderGeneration/SyntaxBuildableNode.swift
+++ b/Sources/SwiftSyntaxBuilderGeneration/SyntaxBuildableNode.swift
@@ -1,0 +1,60 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2022 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// Extension to the `Node` type to provide functionality specific to
+/// SwiftSyntaxBuilder.
+extension Node {
+  /// The node's syntax kind as `SyntaxBuildableType`.
+  var type: SyntaxBuildableType {
+    SyntaxBuildableType(syntaxKind: syntaxKind)
+  }
+
+  /// The node's syntax kind as `SyntaxBuildableType`.
+  var baseType: SyntaxBuildableType {
+    SyntaxBuildableType(syntaxKind: baseKind)
+  }
+
+  /// If documentation exists for this node, return it as a single-line string.
+  /// Otherwise return an empty string.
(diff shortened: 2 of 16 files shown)


fwcd:
@kimdv @ahoppen I would be curious to hear your thoughts on this.

The alternative would be to not port this abstraction layer to Swift and integrate the convenience properties directly into the Swift versions of `Child`, `Node` etc. Since that would require us to diverge from the Python API, we would have to figure out if how we'd want to e.g. represent syntax kinds, which in the `SyntaxBuildableWrappers` currently use `SyntaxBuildableType`.

kimdv:
Hmm, personally I prefer to have extensions on the types or add computed propperties on the types. 
I think it make it more readable.

fwcd:
That's fair, do you think it would make sense to add some kind of wrapper abstraction around syntax kinds anyway for the functionality previously provided by `SyntaxBuildableType`?

kimdv:
[Here](https://github.com/apple/swift-syntax/pull/479/files#diff-e0192affe9ef12453d67e416b7f45bb451f0baa95c8ae62d7c96ea14053cf214R415) I've added an extension that get a collection element as a `Node`.
After that we have the same capabilities as a node. 
I think it's a bit more readable and easier to understand what happens in the code when generating. Specially for new comers to the project. 

But that's just my person opinion 😁

Note: It's still under development, so I will properly use what this PR brings 👏

ahoppen:
Just a thought: What do you think about making `SyntaxBuildableChild` and `SyntaxBuildableNode` extensions on `Child` and `Node` that are internal to `SwiftSyntaxBuilderGeneration`?

Sources/SwiftSyntaxBuilderGeneration/SyntaxBuildableNode.swift
```diff
@@ -0,0 +1,53 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// Extension to the `Node` type to provide functionality specific to
+/// SwiftSyntaxBuilder.
+extension Node {
```
fwcd, line 15: While we could also move these types into `Node` directly, I kept them in an extension for clarity (the same goes for `Child` and its extension). I don't have a strong opinion here though, so if you think it's cleaner to have them all in one place, we could also move them to their respective classes.

Sources/SwiftSyntaxBuilderGeneration/Node.swift
```diff
@@ -28,18 +28,6 @@ class Node {
...
-    if self.baseKind == "SyntaxCollection" {
-      return "Syntax"
-    } else {
-      return kindToType(kind: baseKind)
-    }
-  }
-
-  var collectionElementType: String {
-    return kindToType(kind: collectionElement)
-  }
```
fwcd, line 41: These are replaced with `baseType: SyntaxBuildableType` and `collectionElementType: SyntaxBuildableType`, i.e. an API break to be aware of.

fwcd (commented):
I have updated `SyntaxBuildableNode`/`SyntaxBuildableChild` to be extensions on `Node`/`Child` and implemented `SyntaxBuildableType` (which should be mostly API-complete, the `Node`/`Child` extensions are still WIP).

A few notes below.

Sources/SwiftSyntaxBuilderGeneration/SyntaxBuildableNode.swift
ahoppen: SGTM. Eventually we might want to extract `Node` etc. to a different target if we want to also generate other parts of SwiftSyntax using `SwiftSyntaxBuilder`. But these extensions are specific to the generation of `SwiftSyntaxBuilder` itself.

fwcd:
@swift-ci please test

fwcd:
@swift-ci please test

fwcd:
@swift-ci please test

ahoppen (approved):
I think it would be great if we could migrate the methods that generate Swift source code in SyntaxBuildableChild.swift and SyntaxBuildableType.swift to use SwiftSyntaxBuilder as well. But that’s something that can be done in a follow-up PR after we actually started using these methods and have test coverage for them.

PR #486, opened 2022-07-05
Add a method to SyntaxProtocol to dump a node recursively for debugging purposes

ahoppen:
Swift’s reflection mechanism is actually really good to print syntax trees, we just need a convenient way to access it.

This produces output as can be seen in [CustomReflectableTests.swift](https://github.com/apple/swift-syntax/blob/main/Tests/SwiftSyntaxTest/CustomReflectableTests.swift).


ahoppen:
@swift-ci Please test

kimdv (approved):

PR #489, opened 2022-07-06
Add Keyword Convenience Initializer To TokenKind

CodaFi:
Add an API that allows the construction of specifically keyword token
kinds from text.

### DIFF
--- a/Sources/SwiftSyntax/TokenKind.swift.gyb
+++ b/Sources/SwiftSyntax/TokenKind.swift.gyb
@@ -31,6 +31,21 @@ public enum TokenKind {
   case ${kind}
 % end
 
+  /// Initializes a keyword token kind from its string representation. If the
+  /// given string is not a keyword, this function returns `nil`.
+  public init?(keyword: String) {
+    switch keyword {
+% for token in SYNTAX_TOKENS:
+%   if token.is_keyword:
+    case "${token.text}":
+      self = .${token.swift_kind()}
+%   end
+% end
+    default:
+      return nil
+    }
+  }
+
   /// The textual representation of this token kind.
   var text: String {
     switch self {
--- a/Tests/SwiftSyntaxParserTest/TokenTest.swift
+++ b/Tests/SwiftSyntaxParserTest/TokenTest.swift
@@ -1,5 +1,5 @@
 import XCTest
-import SwiftSyntax
+@_spi(Testing) import SwiftSyntax
 import SwiftSyntaxParser
 
 public class TokenTests: XCTestCase {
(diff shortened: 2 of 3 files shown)


CodaFi:
@swift-ci test

Sources/SwiftSyntax/TokenKind.swift.gyb
```diff
@@ -31,6 +31,29 @@ public enum TokenKind {
   case ${kind}
 % end
 
+  /// Initializes a keyword token kind from its string representation. If the
+  /// given string is not a keyword, this function returns `nil`.
+  public init?(keyword: String) {
+    switch keyword.utf8.count {
+% from itertools import groupby
+% for length, items in groupby(sorted(list(filter(lambda x: x.is_keyword, SYNTAX_TOKENS)), key=lambda x: len(x.text)), key=lambda x: len(x.text)):
```
CodaFi, line 39: If there's a better way to express this, I'm all ears.

CodaFi: The idea is to take all the keywords, sort them into buckets by the number of characters they have, then take `(length, bucket)` pairs and render switch statements out of each bucket after testing the length.

Tests/SwiftSyntaxParserTest/TokenTest.swift
```diff
@@ -1,5 +1,5 @@
 import XCTest
-import SwiftSyntax
+@testable import SwiftSyntax
```
ahoppen, line 2: I would very much like to avoid using `@testable`. In CI we currently only build SwiftSyntax in Release mode with testability disabled. If we require `@testable` we need to build SwiftSyntax twice (once as part of the “build” phase without testability and once for the “test” phase with testability). 

The paradigm we are using in SourceKit-LSP is to add a computed underscored property. I would suggest we add a new TestSupport.swift file in the SwiftSyntax target with
```swift
extension TokenKind {
  /// *Public for testing only*
  public var _text: String {
    return text
  }
}
```

Sources/SwiftSyntax/TokenKind.swift.gyb
ahoppen: I assume there is there a measurable performance benefit from do the switch on the length first? If not, we could just keep it simple and switch on the string contents themselves.

CodaFi: There is a significant benefit in my testing. But I agree let's keep things simple.

CodaFi:
@swift-ci test

CodaFi:
⛵

PR #497, opened 2022-07-11
Add `SyntaxBuildableType.listBuildable`

fwcd:
For some reason I missed this method in #485, therefore a small follow-up PR here.


fwcd:
@swift-ci please test

PR #495, opened 2022-07-10
Generate `ExpressibleAsProtocols` with `SwiftSyntaxBuilderGeneration`

fwcd:
This PR ports the currently gyb-generated `ExpressibleAsProtocols` to `SwiftSyntaxBuilder`'s DSL as part of the ongoing effort to bootstrap `SwiftSyntaxBuilder` with `SwiftSyntaxBuilderGeneration`.

PR #480, opened 2022-07-02
Fix indent when there is a leading new line

kimdv:
The list of Tricia pieces look likes this

```
TriviaPiece with 1 Newlines
TriviaPiece with 4 Spaces
TriviaPiece with '/// A nested struct' DocLineComment
TriviaPiece with 1 Newlines
TriviaPiece with '/// with multi line comment' DocLineComment
TriviaPiece with 1 Newlines
```

That need to be replaces with

```
TriviaPiece with 1 Newlines
TriviaPiece with 4 Spaces
TriviaPiece with '/// A nested struct' DocLineComment
TriviaPiece with 1 Newlines
TriviaPiece with 4 Spaces
TriviaPiece with '/// with multi line comment' DocLineComment
TriviaPiece with 1 Newlines
TriviaPiece with 4 Spaces
```

PR #498, opened 2022-07-11
Generate `BuildableBaseProtocols` with `SwiftSyntaxBuilderGeneration`

fwcd:
Based on #497.

This PR ports the currently gyb-generated `BuildableBaseProtocols` to `SwiftSyntaxBuilder`'s DSL as part of the ongoing effort to bootstrap `SwiftSyntaxBuilder` with `SwiftSyntaxBuilderGeneration`.


Sources/SwiftSyntaxBuilderGeneration/Templates/BuildableBaseProtocolsFile.swift
```diff
@@ -0,0 +1,168 @@
...
+          InheritedType(typeName: conformance)
+        }
+      }
+    ) {
+      FunctionDecl(
+        leadingTrivia: [
+          "/// Builds list of `\(type.syntax)`s.",
+          "/// - Parameter format: The `Format` to use.",
+          "/// - Parameter leadingTrivia: Replaces the last leading trivia if not nil.",
+        ].map { .docLineComment($0) + .newline }.reduce([], +),
```
fwcd, line 50: Generating multi-line doc strings is probably a common pattern that we should move to a utility function. (The reason for not doing a single `.docLineComment` with a multi-line string is that #480 cannot handle indentation correctly in that case IIUC.)

Sources/SwiftSyntaxBuilderGeneration/Templates/BuildableBaseProtocolsFile.swift
```diff
@@ -0,0 +1,168 @@
...
+    var listConformances: [String] = isSyntax ? [] : [syntaxType.buildable]
+
+    ProtocolDecl(
+      modifiers: [TokenSyntax.public],
+      identifier: type.listBuildable,
+      inheritanceClause: listConformances.isEmpty ? nil : TypeInheritanceClause {
+        for conformance in listConformances {
+          InheritedType(typeName: conformance)
+        }
+      }
```
fwcd, line 43: Generating a (non-nil) conformance clause iff there are conformances seems to be a common pattern too (that I've also encountered in #495). Perhaps we should move this either into a utility function (similar to `conformance_clause` from the gyb helpers) or simplify the DSL in some form, e.g. by adding new convenience initializers?

fwcd:
@swift-ci please test

Sources/SwiftSyntaxBuilderGeneration/Templates/BuildableBaseProtocolsFile.swift
ahoppen: I agree that it would make sense to add this as a utility function in `SwiftSyntaxBuilderGeneration`.

Sources/SwiftSyntaxBuilderGeneration/Templates/BuildableBaseProtocolsFile.swift
```diff
@@ -0,0 +1,168 @@
...
+    leadingTrivia: .docLineComment(copyrightHeader),
+    path: "SwiftSyntax"
+  )
+
+  // `SyntaxCollectionBuildable` and `ExpressibleAsSyntaxCollectionBuildable` don't exist because of the following reason:
+  // Through `ExpressibleAs*` conformances, a syntax kind might conform to `ExpressibleAsSyntaxCollectionBuildable` via different paths, thus the implementation of `createSyntaxCollectionBuildable` is ambiguous.
+  // We have the same issue for `ExpressibleAsSyntaxBuildable`, but in that case the solution is simple: Create a custom implementaiton of `createSyntaxBuildable` that doesn't perform any of the conversions via the `ExpressibleAs` protocols.
+  // For `SyntaxCollection` we would need to perform at least one conversion to end up with a type that we can put inside a `SyntaxCollection`, so there is no single canonical implementation.
+  // Since the types don't provide any value, we don't generate them for now.
+  for kind in SYNTAX_BASE_KINDS where kind != "SyntaxCollection" {
```
fwcd, line 28: One thing I've stumbled upon is that we probably want to sort this set (`SYNTAX_BASE_KINDS`) to enforce a deterministic iteration order (otherwise regenerating may yield different results).

ahoppen: Oh, yes, absolutely, we need deterministic iteration here.

fwcd:
@swift-ci please test

Sources/SwiftSyntaxBuilder/generated/BuildableBaseProtocols.swift
```diff
@@ -0,0 +1,217 @@
...
+// Copyright (c) 2014 - 2022 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+public protocol DeclListBuildable: SyntaxBuildable{
```
evnik, line 17: Shouldn't it be `SyntaxListBuildable`?
```suggestion
public protocol DeclListBuildable: SyntaxListBuildable{
```

fwcd: Good catch, thanks! Syntax and list conformances were generated in the wrong order, should be fixed now.

Sources/SwiftSyntaxBuilder/generated/BuildableBaseProtocols.swift
```diff
@@ -0,0 +1,217 @@
...
+// Copyright (c) 2014 - 2022 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+public protocol DeclListBuildable: SyntaxListBuildable{
```
evnik, line 17: Please rebase on top of the latest `main` and regenerate this file. It should add a space before left braces
```suggestion
public protocol DeclListBuildable: SyntaxListBuildable {
```

fwcd: Done!

fwcd:
@swift-ci please test

PR #500, opened 2022-07-14
[NFC] Scale Back Foundation Dependency

CodaFi:
Remove SwiftSyntax's dependency on Foundation - there's nothing in this library that really needs that big a dependency.


CodaFi:
@swift-ci test

ahoppen (approved):

PR #504, opened 2022-07-19
Add an API to compare two separate syntax trees for equivalence

bnbarham:
Given two syntax trees, add an API to find the first difference between
them. Use this to provide an XCTest assertion that two trees are equal,
which dumps out both with a marker for the different node otherwise.

Modifies `Format` to allow removing the builder's auto-formatting so
that the trivia in the expected case is more obvious.

Adds a new `SwiftSyntaxSupport` target and moves the various test
methods into it.


Sources/SwiftSyntax/Syntax.swift
```diff
@@ -81,6 +81,76 @@ extension Syntax: CustomReflectable {
...
+      target.write(" [\(range.start)...\(range.end)]")
+    }
+
+    if self == mark {
+      target.write(" ***")
+    }
+
+    if includeChildren {
+      let childIndent = indent + 2
+      for (num, child) in children.enumerated() {
```
bnbarham, line 143: It's probably useful to print missing children, where as this is only the present children. Same when comparing them as well - at the moment if we were expecting to see a parameter list with missing paren tokens but instead got a missing parameter list, that would just show up as a .missingNode.

Sources/SwiftSyntax/SyntaxComparison.swift
```diff
@@ -0,0 +1,80 @@
...
+
+public extension Syntax {
+  func isEquivalent(to other: Syntax, includeTrivia: Bool = false) -> DifferenceReason? {
+    if syntaxNodeType != other.syntaxNodeType {
+      return .nodeType
+    }
+
+    if isPresent != other.isPresent {
+      return .presence
+    }
```
bnbarham, line 41: This will never be possible at the moment, since we only ever compare present children.

Sources/SwiftSyntax/Syntax.swift
```diff
@@ -81,6 +81,76 @@ extension Syntax: CustomReflectable {
...
+      topLevelConverter = SourceLocationConverter(file: "", tree: file)
+    }
+
+    if self == mark {
+      target.write("*** ")
+    }
+
+    if isMissing {
+      target.write("- ")
+    }
```
bnbarham, line 112: Not actually possible right now, but IMO walking missing (expected) children would make sense.

Sources/SwiftSyntax/Syntax.swift
```diff
@@ -81,6 +81,76 @@ extension Syntax: CustomReflectable {
...
+    debugWrite(to: &str, includeChildren: includeChildren, includeTrivia: includeTrivia, converter: converter, mark: mark, indent: indent)
+    return str
+  }
+
+  public func debugWrite<Target: TextOutputStream>(to target: inout Target,
+                                                   includeChildren: Bool = true, includeTrivia: Bool = false,
+                                                   converter: SourceLocationConverter? = nil, mark: Syntax? = nil,
+                                                   indent: Int = 0) {
+    var topLevelConverter = converter
+    if topLevelConverter == nil, let file = `as`(SourceFileSyntax.self) {
```
ahoppen, line 102: Maybe this would read better as
```suggestion
    if topLevelConverter == nil, let file = self.as(SourceFileSyntax.self) {
```

Similar below. I don’t really like the look if the escaped `as`

Sources/SwiftSyntax/SyntaxComparison.swift
```diff
@@ -0,0 +1,80 @@
...
+
+  public init(lhs: Syntax, rhs: Syntax, reason: DifferenceReason) {
+    self.node = lhs
+    self.comparison = rhs
+    self.reason = reason
+  }
+}
+
+public extension Syntax {
+  func isEquivalent(to other: Syntax, includeTrivia: Bool = false) -> DifferenceReason? {
```
ahoppen, line 34: Since this returns a `DifferenceReason`, I think `isDifferent` would be a better name. With the current name, I would expect it to return something `true`-ish (i.e. non-`nil`) if the trees are equivalent.

Also, I think it’s worth mentioning that this does not compare children in a doc comment and because of this reason make it private, if possible.

Sources/SwiftSyntax/SyntaxComparison.swift
```diff
@@ -0,0 +1,80 @@
...
+      }
+      if includeTrivia && (leadingTrivia != other.leadingTrivia ||
+                           trailingTrivia != other.trailingTrivia) {
+        return .trivia
+      }
+    }
+    return nil
+  }
+
+  func findFirstDifference(to other: Syntax, includeTrivia: Bool = false) -> TreeDifference? {
```
ahoppen, line 58: Worth renaming `other` to `baseline`? That way you would know which tree has an additional node if the difference reason is `additionalNode`.

Sources/SwiftSyntax/gyb_generated/SyntaxFactory.swift
```diff
@@ -5467,7 +5467,7 @@ public enum SyntaxFactory {
                      trailingTrivia: trailingTrivia)
   }
   public static func makeLeftBraceToken(
-    leadingTrivia: Trivia = .space,
+    leadingTrivia: Trivia = [],
```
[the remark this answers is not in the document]
ahoppen: This got changed in https://github.com/apple/swift/pull/59980. I think you need to update your swift checkout.

Package.swift
```diff
@@ -90,6 +90,10 @@ let package = Package(
       ],
       linkerSettings: swiftSyntaxParserLinkerSettings
     ),
+    .target(
+      name: "SwiftSyntaxSupport",
```
ahoppen, line 94: Shouldn’t this be named `SwiftSyntaxTestSupport` if it imports `XCTest`?

ahoppen (commented):
Looking good overall. I added some inline comments. 

I think to make reviews easier, it would be good if this could be split into multiple PRs:
- Introduce `SwiftSyntaxTestSupport` and move methods which already exist to that module (+ I think you did a few general improvements to the tests)
- Make the `_makeNewline` change in `SwiftSyntaxBuilder`
- Introduce SwiftSyntaxComparison

I think it would also be nice to have test cases for the output of SwiftSyntaxComparison.

Package.swift
bnbarham: Yeah, probably should 👍

Sources/SwiftSyntax/SyntaxComparison.swift
bnbarham: Sounds reasonable, will do!

Sources/SwiftSyntax/SyntaxComparison.swift
bnbarham: Good point 👍

PR #505, opened 2022-07-19
Use `let` bindings in `BuildableBaseProtocolsFile`

fwcd:
A minor stylistic oversight that originated from translating a variable mutation from Python, which we however ended up rewriting in an immutable style. This patch therefore updates these variables to be `let`-bindings.


fwcd:
@swift-ci please test

ahoppen (approved):

PR #506, opened 2022-07-20
Generate `BuildableCollectionNodes` with `SwiftSyntaxBuilderGeneration`

fwcd:
This PR ports the currently gyb-generated `BuildableCollectionNodes` to `SwiftSyntaxBuilder`'s DSL as part of the ongoing effort to bootstrap `SwiftSyntaxBuilder` with `SwiftSyntaxBuilderGeneration`.


fwcd:
@swift-ci please test

Sources/SwiftSyntaxBuilderGeneration/Templates/BuildableCollectionNodesFile.swift
```diff
@@ -0,0 +1,257 @@
...
+      ) {
+        SequenceExpr(elements: ExprList([
+          MemberAccessExpr(base: "self", name: "elements"),
+          AssignmentExpr(),
+          elementType.isToken
+            ? "elements"
+            : FunctionCallExpr(MemberAccessExpr(base: "elements", name: "map"), trailingClosure: ClosureExpr {
+                FunctionCallExpr(MemberAccessExpr(base: "$0", name: "create\(elementType.buildableBaseName)"))
+              }),
+        ]))
```
fwcd, line 64: The `Array: ExpressibleAsExprList` conformance (as proposed in https://github.com/apple/swift-syntax/pull/465#discussion_r899162231) would really come in handy since we use quite a few `SequenceExpr`s here and in the other generated files.

Perhaps noteworthy that `BuildableCollectionNodes` (this file) is also responsible for generating these conformances, though I would still consider this to be more of a candidate for a future PR.

Sources/SwiftSyntaxBuilderGeneration/Templates/BuildableCollectionNodesFile.swift
```diff
@@ -0,0 +1,257 @@
...
+        identifier: .identifier("build\(type.baseName)"),
+        signature: FunctionSignature(
+          input: createFormatLeadingTriviaParameters(withDefaultTrivia: true),
+          output: type.syntax
+        )
+      ) {
+        VariableDecl(letOrVarKeyword: .let) {
+          PatternBinding(
+            pattern: "result",
+            initializer: InitializerClause(
```
fwcd, line 123: Perhaps we could add an optional parameter or a new convenience initializer in `VariableDeclConvenienceInitializers` for variable declarations binding an initializer clause? This seems like a common use case and might be worth sugaring.

Sources/SwiftSyntaxBuilderGeneration/Templates/BuildableCollectionNodesFile.swift
```diff
@@ -0,0 +1,257 @@
...
+            ))
+          }),
+        ]))
+      }
+
+      InitializerDecl(
+        modifiers: [TokenSyntax.public],
+        parameters: ParameterClause(
+          parameterList: [
+            FunctionParameter(
```
fwcd, line 97: Having `Array: ExpressibleAsParameterClause` might be cool

Sources/SwiftSyntaxBuilderGeneration/Templates/BuildableCollectionNodesFile.swift
```diff
@@ -0,0 +1,257 @@
...
+                  ]))
+                }
+              ]))
+            })
+          },
+          elseKeyword: .else,
+          elseBody: CodeBlock {
+            ReturnStmt(expression: "result")
+          }
+        )
```
fwcd, line 178: `IfStmt` might actually look nice with a multiple-trailing-closure-style API:

```swift
IfStmt(conditions: ...) {
  // ...
} elseBody: {
  // ...
}
```

Perhaps we would want to add an extra initializer for this in an `IfStmtConvenienceInitializers` file? Not sure whether that would complicate overload resolution or make the existing initializers ambiguous though.

Sources/SwiftSyntaxBuilderGeneration/Templates/BuildableCollectionNodesFile.swift
```diff
@@ -0,0 +1,257 @@
...
+      }
+
+      FunctionDecl(
+        modifiers: [TokenSyntax.public],
+        identifier: .identifier("buildSyntax"),
```
fwcd, line 183: We are using `TokenSyntax.identifier(...)` in a lot of places. This got me thinking whether we might be able to conform `TokenSyntax` to `ExpressibleByStringLiteral`?

fwcd (commented):
A note on the implementation and a few ideas for (future) improvements to the DSL below

Sources/SwiftSyntaxBuilderGeneration/Templates/BuildableCollectionNodesFile.swift
fwcd: -> Moving the discussion to #507.

Sources/SwiftSyntaxBuilderGeneration/Templates/BuildableCollectionNodesFile.swift
```diff
@@ -0,0 +1,257 @@
...
+                      }
+                    })
+                  )
+                }
+              }
+            )
+          )
+        }
+        IfStmt(
+          conditions: ConditionElementList([
```
ahoppen, line 152: Would it be worth making adding a `Array: ConditionElementList` conformance? 

Also I think we should make `OptionalBindingCondition: ExpressibleAsConditionElement`. Combined, that should allow us to write `conditions: OptionalBindingCondition`.

Sources/SwiftSyntaxBuilderGeneration/Templates/BuildableCollectionNodesFile.swift
ahoppen: Yes, I agree. It’s one of the few places where I actually agree that multiple trailing closures might actually be nice.

Sources/SwiftSyntaxBuilderGeneration/Templates/BuildableCollectionNodesFile.swift
```diff
@@ -0,0 +1,257 @@
...
+                        TupleExprElement(
+                          label: "leadingTrivia",
+                          expression: node.elementsSeparatedByNewline
+                            ? SequenceExpr(elements: ExprList([
+                                MemberAccessExpr(base: "Trivia", name: "newline"),
+                                BinaryOperatorExpr("+"),
+                                FunctionCallExpr(MemberAccessExpr(base: "format", name: "_makeIndent")),
+                              ]))
+                            : "nil"
+                        )
```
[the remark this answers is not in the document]
fwcd: IIUC assigning variables isn't allowed in result builders, so we would have to use ternaries anyway (or factor out some of this into a separate function). I agree though, the builder is nested quite deeply, I think if we could make `VariableDecl` more convenient to initialize and ideally also make the `TupleExprElement` wrappers implicit, we could get ride of a quite a few levels of nesting.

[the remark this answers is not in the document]
ahoppen: AFAIK it’s totally fine to declare variables inside result builders.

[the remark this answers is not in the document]
fwcd: Yes, but reassigning (even if it's initializing a constant) yields errors in my testing.

Sources/SwiftSyntaxBuilderGeneration/Templates/BuildableCollectionNodesFile.swift
fwcd: -> #518

Sources/SwiftSyntaxBuilderGeneration/Templates/BuildableCollectionNodesFile.swift
fwcd: -> #523

Sources/SwiftSyntaxBuilderGeneration/Templates/BuildableCollectionNodesFile.swift
```diff
@@ -0,0 +1,257 @@
...
+      ) {
+        SequenceExpr(elements: ExprList([
+          "elements",
+          AssignmentExpr(),
+          FunctionCallExpr(MemberAccessExpr(base: "lists", name: "flatMap"), trailingClosure: ClosureExpr {
+            CodeBlockItem(item: MemberAccessExpr(
+              base: FunctionCallExpr(MemberAccessExpr(base: "$0", name: "create\(type.buildable)")),
+              name: "elements"
+            ))
+          }),
```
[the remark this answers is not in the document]
fwcd: -> #524

Sources/SwiftSyntaxBuilderGeneration/Templates/BuildableCollectionNodesFile.swift
fwcd: -> #525

This should get us the `OptionalBindingCondition` -> `ConditionElement` -> `ConditionElementList` conformance path. The `Array` conformance already exists AFAICT.

fwcd:
Whoops, didn't intend to auto-close this :D

Sources/SwiftSyntaxBuilderGeneration/Templates/BuildableCollectionNodesFile.swift
fwcd: Moving the discussion to #528.

Sources/SwiftSyntaxBuilderGeneration/Templates/BuildableCollectionNodesFile.swift
```diff
@@ -0,0 +1,248 @@
...
+            }
+          }
+        )
+        IfStmt(
+          conditions: OptionalBindingCondition(
+            letOrVarKeyword: .let,
+            pattern: "leadingTrivia",
+            initializer: "leadingTrivia"
+          )
+        ) {
```
fwcd, line 155: Something I thought about while working on this was whether we might want to remove this `if`-statement and replace e.g.

```swift
public func buildTupleExprElementList(format: Format, leadingTrivia: Trivia? = nil) -> TupleExprElementListSyntax {
  ...
  if let leadingTrivia = leadingTrivia {
    return result.withLeadingTrivia(leadingTrivia + (result.leadingTrivia ?? []))
  } else {
    return result
  }
}
```

with

```swift
public func buildTupleExprElementList(format: Format, leadingTrivia: Trivia? = nil) -> TupleExprElementListSyntax {
  ...
  let combinedLeadingTrivia = (leadingTrivia ?? []) + (result.leadingTrivia ?? [])
  return result.withLeadingTrivia(combinedLeadingTrivia)
}
```

Although we would deviate from the original translation there, the latter solution feels more consistent with `BuildableNodes` (which takes a similar approach) and would present a potential opportunity to unify these implementations by factoring the syntax generation into a utility function.

ahoppen: That sounds like a good idea to me 👍

fwcd:
I have cleaned up the template and factored out the generated functions along with some doc comments, so it should be a bit more readable now.

@swift-ci please test

Sources/SwiftSyntaxBuilderGeneration/Templates/BuildableCollectionNodesFile.swift
```diff
@@ -0,0 +1,283 @@
...
+      // Generate initializers
+      createArrayInitializer(type: type, elementType: elementType)
+      createCombiningInitializer(type: type)
+      createArrayLiteralInitializer(elementType: elementType)
+
+      // Generate function declarations
+      createBuildFunction(node: node, type: type, elementType: elementType)
+      createBuildSyntaxFunction(type: type)
+      createExpressibleAsCreateFunction(type: type)
+      createSyntaxBuildableCreateFunction(type: type)
```
fwcd, line 48: A question regarding style: I have passed `type`, `elementType` and `node` separately to these functions for now. Since `type` and `elementType` can easily be derived from just `node`, would it be better to just pass `node`?

ahoppen: I don’t think there’s a reason we need to pass `type` and `elementType` if they can be derived from `node` and there’s no reason you ever wanted to pass e.g. a different `type` than `node.type`.

ahoppen (approved):

Sources/SwiftSyntaxBuilderGeneration/Templates/BuildableCollectionNodesFile.swift
```diff
@@ -0,0 +1,296 @@
...
+                  MemberAccessExpr(base: "result", name: "leadingTrivia")
+                  BinaryOperatorExpr("??")
+                  ArrayExpr(elements: [])
+                }
+              }
+            }
+          },
+          name: "addingSpacingAfterNewlinesIfNeeded"
+        )))
+      })
```
fwcd, line 235: Contrary to what I proposed in https://github.com/apple/swift-syntax/pull/506#discussion_r933672865, we may have to stick with the `if-let` for now since the variant I proposed has slightly different semantics (the `if`-version will only consider `result.leadingTrivia` if the function parameter `leadingTrivia` is set). This becomes relevant as otherwise `.addingSpacingAfterNewlinesIfNeeded()` will generate duplicate indentation in some cases.

One alternative solution would be to use optional chaining and only apply the spacing-after-newlines to the function argument `leadingTrivia`:

```swift
let combinedLeadingTrivia = (leadingTrivia?.addingSpacingAfterNewlinesIfNeeded() ?? []) + (result.leadingTrivia ?? [])
return combinedLeadingTrivia
```

instead of the current

```swift
if let leadingTrivia = leadingTrivia {
  return result.withLeadingTrivia((leadingTrivia + (result.leadingTrivia ?? [])).addingSpacingAfterNewlinesIfNeeded())
} else {
  return result
}
```

While that seems to yield the correct result, I am not sure whether there are some edge cases where this logic breaks down (or whether we make too many assumptions about the kind of leading trivia here, e.g. whitespace trivia is 'nice' in the sense that it commutes and therefore may not exercise certain logic errors).

For this reason I decided not to make this change until we have figured that out. Any thoughts on this?

fwcd:
@swift-ci please test

Sources/SwiftSyntaxBuilderGeneration/Templates/BuildableCollectionNodesFile.swift
ahoppen: > One alternative solution would be to use optional chaining and only apply the spacing-after-newlines to the function argument leadingTrivia

I think that sounds reasonable. I would suggest we discuss it in a separate PR.

Sources/SwiftSyntaxBuilderGeneration/Templates/BuildableCollectionNodesFile.swift
```diff
@@ -0,0 +1,296 @@
...
+
+/// Generate an initializer taking an array of elements.
+private func createArrayInitializer(node: Node) -> InitializerDecl {
+  let type = node.type
+  let elementType = node.collectionElementType
+  return InitializerDecl(
+    leadingTrivia: [
+      "/// Creates a `\(type.buildable)` with the provided list of elements.",
+      "/// - Parameters:",
+      "/// - elements: A list of `\(elementType.expressibleAs)`",
```
ahoppen, line 91: The elements parameter needs to be indented so source editors will pick it up as a parameter documentation IIRC.

```suggestion
      "///   - elements: A list of `\(elementType.expressibleAs)`",
```

ahoppen (approved):
One minor comment regarding the generated documentation, otherwise LGTM.

fwcd:
@swift-ci please test

PR #511, opened 2022-07-23
Improvements for `do` and related statements

evnik:
* Leading/trailing spaces are updated to make statements well formatted
* `CatchClause` convenience initializer that calculates spacing around the `catch` keyword
* Test coverage for above changes

https://github.com/apple/swift/pull/60204

### DIFF
--- a/Tests/SwiftSyntaxBuilderTest/DoStmtTests.swift
+++ b/Tests/SwiftSyntaxBuilderTest/DoStmtTests.swift
@@ -1,0 +1,48 @@
+import XCTest
+import SwiftSyntaxBuilder
+
+final class DoStmtTests: XCTestCase {
+  func testDoStmt() {
+    let buildable = DoStmt(
+      labelName: nil,
+      body: CodeBlock(statementsBuilder: {
+        CodeBlockItem(item: TryExpr(expression: FunctionCallExpr(MemberAccessExpr(base: "a", name: "b"))))
+      }),
+      catchClauses: [
+        CatchClause(CatchItemList {
+          CatchItem(pattern: "Error1")
+          CatchItem(pattern: "Error2")
+        }) {
+          FunctionCallExpr("print") {
+            TupleExprElement(expression: StringLiteralExpr("Known error"))
+          }
+        },
+        CatchClause(CatchItemList {
+          CatchItem(
+            pattern: "Error3", whereClause: WhereClause(guardResult: MemberAccessExpr(base: "error", name: "isError4")))
+        }) {
+          ThrowStmt(expression: MemberAccessExpr(base: "Error4", name: "error3"))
+        },
+        CatchClause {
+          FunctionCallExpr("print") {
+            TupleExprElement(expression: "error")
+          }
+        }
+      ])
+
+    let syntax = buildable.buildSyntax(format: Format())
+    XCTAssertEqual(
+      syntax.description,
+      """
+      do {
+          try a.b()
+      } catch Error1, Error2 {
+          print("Known error")
+      } catch Error3 where error.isError4 {
+          throw Error4.error3
+      } catch {
+          print(error)
+      }
+      """)
+  }
+}
--- a/Tests/SwiftSyntaxBuilderTest/StructTests.swift
+++ b/Tests/SwiftSyntaxBuilderTest/StructTests.swift
@@ -28,7 +28,7 @@ final class StructTests: XCTestCase {
       ],
       structKeyword: .struct,
       identifier: "NestedStruct",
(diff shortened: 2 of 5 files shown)


Tests/SwiftSyntaxBuilderTest/StructTests.swift
```diff
@@ -22,7 +22,7 @@ final class StructTests: XCTestCase {
     let nestedStruct = StructDecl(
       structKeyword: .struct.withLeadingTrivia(.docLineComment("/// A nested struct\n")),
       identifier: "NestedStruct",
-      genericParameterClause: GenericParameterClause {
+      genericParameterClause: GenericParameterClause(rightAngleBracket: .rightAngle.withoutTrailingTrivia()) {
```
evnik, line 25: This is a temporary fix needed because of updates for the `where` keyword. I'm going to remove it with adjusting spaces around angle brackets in my next PR

Tests/SwiftSyntaxBuilderTest/DoStmtTests.swift
```diff
@@ -0,0 +1,48 @@
...
+    let buildable = DoStmt(
+      labelName: nil,
+      body: CodeBlock(statementsBuilder: {
+        CodeBlockItem(item: TryExpr(expression: FunctionCallExpr(MemberAccessExpr(base: "a", name: "b"))))
+      }),
+      catchClauses: [
+        CatchClause(CatchItems {
+          CatchItem(pattern: "Error1")
+          CatchItem(pattern: "Error2")
+        }) {
```
ahoppen, line 15: I’m not entirely convinced this is much better than the default implementation that’s already provided.

```suggestion
        CatchClause(catchItems: CatchItemList([
          CatchItem(pattern: "Error1"),
          CatchItem(pattern: "Error2")
        ])) {
```

Do you have some example where the result builder syntax is superior than using an array here?

Tests/SwiftSyntaxBuilderTest/StructTests.swift
ahoppen: I’m not sure how big the changes for the angle brackets are but feel free to include them in this PR.

ahoppen (commented):
The spacing changes look good to me. I’m not entirely convinced about `CatchItems`.

Tests/SwiftSyntaxBuilderTest/DoStmtTests.swift
evnik: 1. Result builder automatically inserts trailing commas
1. This `CatchClause` convenience initializer adds trailing space to the `catch` keyword

Tests/SwiftSyntaxBuilderTest/StructTests.swift
evnik: Not big, however I'd like to make a separate PR for them

Tests/SwiftSyntaxBuilderTest/DoStmtTests.swift
ahoppen: 1. I really don’t like the idea of introducing yet another type for catch items. We now have `CatchItemsListSyntax`, `CatchItemsList` and `CatchItems`. What do you think about adding an initializer like the following to `CatchItemList`. We could add the same for all syntax collections:
```swift
public init(@CatchItemListBuilder itemsBuilder: () -> ExpressibleAsCatchItemList) {
  self = itemsBuilder().createCatchItemList()
}
```
2. Can’t we just mark the `catch` keyword as `requires_trailing_space` in https://github.com/apple/swift/blob/5dfa894ea773c1cb82106d831cd4d2df4a837b75/utils/gyb_syntax_support/Token.py#L188?

evnik: 1. Great idea, I've raised #519 for it
2. I actually did the opposite to avoid double space [when there is no catch items](https://github.com/apple/swift-syntax/pull/511/files#diff-36fc9e19ca4760fd7343a8a7b18a4b2759cc0a821e1396ec5f804cde41476de0R43), so I'd like to keep this convenience initializer

evnik:
@ahoppen any comments for the updated implementation?

ahoppen (approved):
Sorry that I missed this. LGTM.

ahoppen:
@swift-ci Please test

ahoppen:
https://github.com/apple/swift/pull/60204

@swift-ci Please test

ahoppen:
https://github.com/apple/swift/pull/60204

@swift-ci Please test

ahoppen:
https://github.com/apple/swift/pull/60204

@swift-ci Please test

PR #512, opened 2022-07-23
[WIP][Experimental] Experiment new SwiftSyntax structure.

rintaro:
(Since this branch is in `apple/swift-syntax` I can't push changes to the branch. https://github.com/apple/swift-syntax/compare/main...rintaro:swift-syntax:bump-ptr-allocation-and-everything is more up-to-date)
This change is huge, and still doesn't build test cases. But I'm opening a PR anyway for RFC

* `SyntaxArena` holds the source buffer, and text data in token are now slices of the source buffer
* Use bump-pointer allocation for `RawSyntax` data memory management
* The root `SyntaxData` node owns the `SyntaxArena` so users don't need to manage the lifetime of it.
* Make `TokenKind` a trivial enum, so that we can `if token.kind == .identifier { ... }`
* Removing notion of `SourcePrecence`. All non-optional descendant structure are automatically created in `makeBlank()` APIs, including tokens with empty `text`. So we still can know the syntax. is "missing" by checking the token byte length is zero
* Simplified data structure for `RawSyntax`. `ManagedBuffer` was too complicated. New structure is easier to maintain.
* Moving `SyntaxParser` specific logic (`_InternalSwiftSyntaxParser` interaction) to `SwiftSyntaxParser` by making `RawSytnax` a SPI (`@_spi(RawSyntax)`).
* Add typed raw syntax `RawTokenSyntax`, `RawFunctionDeclSyntax` etc. To make raw tree creation more efficient

I will add more explanation, after I make this at least testable.

---
# Data Strucuture

## Green Tree (`RawSyntax` tree)

`RawSyntax` tree is a traditional tree structure. A node has children or a leaf node data, no reference to the parent node. This is immutable and persistent. A `RawSyntax` node can be reused in different parent.

### `RawSyntax`

The "surface" of the green tree. Just a pointer to `RawSyntaxData` which resides in a `SyntaxArena`.

### `Raw{...}Syntax`

Wrapper of `RawSyntax` for providing `SyntaxKind` specific APIs

### `RawSyntaxData`

The "green" tree node data. Since this is managed by a bump allocator (`SyntaxArena`), all substructure must be strictly trivial (i.e. no-refcounted objects inside (e.g. class instance, existential, indirect enum, etc.))

- Tagged union (`enum RawSyntaxData.Payload`)  of `RawSyntaxData.Layout`, `RawSyntaxData.ParsedToken`,  and `RawSyntaxData.MaterializedToken`.
- A pointer to the `SyntaxArena` where the data itself resides.

### `RawSyntaxData.Layout`

Data for "layout" node. This itself doesn't represent any source text. But it has children which might be tokens, or other layout nodes.

* Kind
* List of child `RawSyntax`. The buffer is allocated in the same `SyntaxArena`
* Total number of byte length (as a precomputed cache)

### `RawSyntaxData.ParsedToken`

Token data for parsed tokens. In this data, leading/trailing trivia are just string slices. trivia pieces are materialized on-demand, by using a function registered in `SyntaxArena`.

* Token kind
* Whole token text (slice from the source buffer) including the leading/trailing trivia
* Range of the token text in the whole text. Meaning, the text before this range is leading trivia, the after is trailing trivia.

### `RawSyntaxData.MaterializedToken`

Token data for constructed tokens. I.e. tokens created by `makeToken(kind:text:leadingTrivia:trailingTrivia:)` API. All text data for custom text pieces are copied to `SyntaxArena`, and this data only holds the references to them.

* Token kind
* Token text
* List of leading/trailing trivia pieces (`RawTriviaPiece`) whose optional text is a slice of the source buffer
* Total number of byte length (as a precomputed cache)

## Red Tree (`Syntax` tree)

`Syntax` tree is a tree with value semantics. You can safely edit the tree without worrying about breaking somewhere unrelated. For each edit, the tree structure is recreated, but reusing other `RaxSyntax`s that are not affected by the edit.

From the implementation POV, `Syntax` tree can be considered as a wrapper of `RawSyntax` tree, but providing a reference to the parent node. Using that, a node can know the absolute position (byte offset, index, etc) in the tree.

### `Syntax`

Just a wrapper of `SyntaxData`. Provides public APIs using its `SyntaxData`

### `{...}Syntax`

Wrapper of `Syntax` for providing `SyntaxKind` specific APIs

### `SyntaxData`

The "red" tree node data. Child nodes are created on-demand. 

* Unowned pointer to a `RawSyntaxData` (as a `RawSyntax`)
* Tagged union of:
  * `SyntaxData.Info.Root` for the root node
  * `SyntaxData.Info.Edge` for non-root nodes
 
### `SyntaxData.Info.Root`

Data for "root" node.

* Owned `SyntaxArena`. This `SyntaxArena` owns the all `RawSyntaxData` in this tree, directly or indirectly by `arena.children`

### `SyntaxData.Info.Edge`

This data is owned by `SyntaxData.Info` as an indirect enum case (`indirect case edge(SyntaxData.Info.Edge)`)

* Parent `SyntaxData`.
* Index in the parent node.
* Precomputed absolute position values
  * Byte offset
  * Index in the whole tree


akyrtzi:
> Use bump-pointer allocation for RawSyntax data memory management

Bump allocation is great for short-lived processes (like command-line tools) but it makes memory management rather complicated for long-lived processes.

When SwiftSyntax gets integrated in the compiler pipeline, what I expect that we do in `sourcekitd` is that we keep the parsed trees of the source files and re-use them instead of re-parsing the same files over and over again from scratch while creating new ASTs. And also we should take advantage of incremental parsing, where you have subsequent trees use parts of the previous ones while adding new nodes, and with a bump-allocation scheme you have a non-stop growing buffer that you can only get rid of wholesale and start completely from scratch again.

I would urge to also consider memory management for a long-lived process when evaluating bump vs heap. You can get better timings for parsing tests with bump-allocation but a) how much significant is it within the context of a fully typechecked compilation, and b) you have to weight these gains against the memory management complexity for a long-running process and the unpredictable performance losses when you are forced to dispose the bump-allocation buffer and start from scratch again.

rintaro:
@akyrtzi 
The allocator (i.e. `SyntaxArena`) is only owned by the root `SyntaxData`, and `SyntaxArena` can be owned by other `SyntaxArena`s as a child (`SyntaxArena.children`). For incremental parsing, For example, say all nodes are initially resides in `arena1` and constructed like:
```
[a-1] @ arena1
 +- [b-1] @ arena1
 +- [c-1] @ arena1
 `- [d-1] @ arena1
```
 When `[c-1]` is rewritten, we reuse `[b-1]` and `[d-1]` and recreate `[c-2]` and `[a-2]` in newly created `arena2`
```
[a-2] @ arena2
 +- [b-1] @ arena1
 +- [c-2] @ arena2
 `- [d-1] @ arena1
 ```
`arena1` is now owned by `arena2`, `arena2` is owned by `[a-2]`.
Then again, when `c-2` is rewritten, we reuse `[b-1]` and `[d-1]`, create `[c-3]` and `[a-3]` in newly created `arena3`.
```
[a-3] @ arena3
 +- [b-1] @ arena1
 +- [c-3] @ arena3
 `- [d-1] @ arena1
 ```
At this point, since nobody owns it, `arena2` is destructed,  then `arena1` is now owned by `arena3`, `arena3` is owned by `[a-3]`.
So we _can_ free memory at some point.
Of course there're concerns like, when every leaf node is owned by different arena, the total memory size might be huge. So we might want to make some mechanism to consolidate the whole tree to a single arena if the total number of arena became huge. But I believe it's manageable.

akyrtzi:
Thanks for providing more details, it'd be useful to quantify the performance benefits with this scheme.

ahoppen:
The overall design sounds great but I think this is too big to review at the moment. I’ve got the feeling that a lot of this can be split up into multiple, reviewable PRs though.

Sources/SwiftSyntax/StringRef.swift
```diff
@@ -0,0 +1,205 @@
...
+#else
+  dst.initialize(from: src, count: count)
+#endif
+}
+
+/// Represent a string.
+///
+/// This type does not own the string data. The data reside in some other buffer
+/// whose lifetime extends past that of the StringRef.
+public struct StringRef {
```
CodaFi, line 41: A nit, but I really dislike the name `StringRef` even in C++. I'd suggest something like `SyntaxString` or `SyntaxText` or `TextSpan`.

Sources/SwiftSyntax/StringRef.swift
```diff
@@ -0,0 +1,205 @@
...
+        copyMemory(strBuffer.baseAddress!, stringRef.baseAddress!, count)
+        return count
+      }
+    } else {
+      self.init(decoding: stringRef, as: UTF8.self)
+    }
+  }
+
+  public mutating func withStringRef<T>(_ body: (StringRef) -> T) -> T {
+    withUTF8 { buffer in
```
CodaFi, line 201: This can be `rethrows`.

Sources/SwiftSyntax/StringRef.swift
rintaro: Let's go with `SyntaxText`

rintaro:
Closing this PR. RawSyntax changes are landed. `Syntax`/`SyntaxData` changes will be in followups

PR #513, opened 2022-07-24
Return non-zero exit code in build-script.py if source generation fails

ahoppen:
Previously, build-script.py would return with a zero exit code, even if source generation failed.


ahoppen:
@swift-ci Please smoke test

kimdv (approved):

ahoppen:
@swift-ci Please test

PR #517, opened 2022-07-26
Introduce `SyntaxText`  - light weight string representation

rintaro:
(Part of #512)

'SyntaxText' represents a string, but without owning the data. In future, this will be used in Syntax tree as string representation.

Provides basic `String`-like functionalities including:
- `Hashable` and `Equatable`
- Collection of `UInt8`
- Slicing
- Prefix/Postfix check
- Interconvertible with `Swift.String`
- Expressible by string literal

### DIFF
--- a/Sources/SwiftSyntax/SyntaxText.swift
+++ b/Sources/SwiftSyntax/SyntaxText.swift
@@ -148,7 +148,7 @@ extension SyntaxText: Hashable {
     if lhs.buffer.count != rhs.buffer.count {
       return false
     }
-    if lhs.buffer.baseAddress == rhs.buffer.baseAddress {
+    if lhs.isEmpty || lhs.buffer.baseAddress == rhs.buffer.baseAddress {
       return true
     }
     return compareMemory(lhs.baseAddress!, rhs.baseAddress!, lhs.count)
--- a/Sources/SwiftSyntax/SyntaxText.swift
+++ b/Sources/SwiftSyntax/SyntaxText.swift
@@ -1,0 +1,226 @@
+//===---------- SyntaxText.swift - Unowned String Representation  ---------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2022 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+#if canImport(Darwin)
+import Darwin
+#elseif canImport(Glibc)
+import Glibc
+#endif
+
+/// Represent a string.
+///
+/// This type does not own the string data. The data reside in some other buffer
+/// whose lifetime extends past that of the SyntaxText.
+///
+/// `SyntaxText` conforms to `Collection` where the element is `UInt8`.
+@_spi(Testing) // SPI name is subject to change
+public struct SyntaxText {
+  var buffer: UnsafeBufferPointer<UInt8>
+
+  public init(baseAddress: UnsafePointer<UInt8>?, count: Int) {
+    assert(count == 0 || baseAddress != nil,
+           "If count is not zero, base address must be exist")
+    buffer = .init(start: baseAddress, count: count)
+  }
+
+  /// Creates an empty `SyntaxText`
+  public init() {
+    self.init(baseAddress: nil, count: 0)
+  }
+
+  /// Creates a `SyntaxText` from a `StaticString`
+  public init(_ string: StaticString) {
+    self.init(baseAddress: string.utf8Start, count: string.utf8CodeUnitCount)
+  }
+
+  /// Base address of the memory range this string refers to.
+  public var baseAddress: UnsafePointer<UInt8>? {
+    buffer.baseAddress
+  }
+
+  /// Byte length of this string.
+  public var count: Int {
+    buffer.count
+  }
+
+  /// A Boolean value indicating whether a string has no characters.
+  public var isEmpty: Bool {
+    buffer.isEmpty
+  }
+
+  /// Returns `true` if the memory range of this string is a part of `other`.
+  ///
+  /// `text[n ..< m].isSliceOf(text)` is always true as long as `n` and `m` are
+  /// valid indices.
+  public func isSliceOf(_ other: SyntaxText) -> Bool {
+    guard !isEmpty && !other.isEmpty else {
+      return isEmpty == other.isEmpty
+    }
+    return (other.baseAddress! <= baseAddress! &&
+            baseAddress! + count <= other.baseAddress! + other.count)
+  }
+
+  /// Returns `true` if `other` is a substring of this `SyntaxText`.
+  ///
+  /// This should not be used because `SyntaxText` is a `Collection<UInt8>` and
+  /// `contains(_:)` is usually for checking a collection contains the element.
+  /// Using `contains(_:)` for subsequence is confusing.
+  @available(*, deprecated, message: "Use 'firstRange(of:) != nil'")
+  public func contains(_ other: SyntaxText) -> Bool {
+    return firstRange(of: other) != nil
+  }
+
+  /// Finds and returns the range of the first occurrence of `other` within this
+  /// string. Returns `nil` if `other` is not found.
+  public func firstRange(of other: SyntaxText) -> Range<Index>? {
+    if other.isEmpty { return nil }
+    let needle = other.baseAddress!
+    let stop = self.count - other.count
+    var start = 0
+    // If 'other' is longer than 'self', stop < 0.
+    while start <= stop {
+      if compareMemory(self.baseAddress! + start, needle, other.count) {
+        return start ..< (start + other.count)
+      } else {
+        start += 1
+      }
+    }
+    return nil
+  }
+
+  /// Returns `true` if the string begins with the specified prefix.
+  public func hasPrefix(_ other: SyntaxText) -> Bool {
+    guard self.count >= other.count else { return false }
+    guard !other.isEmpty else { return true }
+    return self[0..<other.count] == other
+  }
+
+  /// Returns `true` if the string ends with the specified suffix.
+  public func hasSuffix(_ other: SyntaxText) -> Bool {
+    guard self.count >= other.count else { return false }
+    guard !other.isEmpty else { return true }
+    return self[(self.count - other.count)..<self.count] == other
+  }
+}
+
+/// `SyntaxText` is a collection of `UInt8`.
+extension SyntaxText: RandomAccessCollection {
+  public typealias Element = UInt8
+  public typealias Index = Int
+  public typealias SubSequence = SyntaxText
+
+  public var startIndex: Index { buffer.startIndex }
+  public var endIndex: Index { buffer.endIndex }
+
+  public subscript(bounds: Range<Index>) -> SyntaxText {
+    if isEmpty && bounds.isEmpty { return self }
+    return SyntaxText(
+      baseAddress: buffer.baseAddress! + bounds.lowerBound,
+      count: bounds.count)
+  }
+
+  public subscript(position: Index) -> Element {
+    unsafeAddress { buffer.baseAddress!.advanced(by: position) }
+  }
+}
+
+extension SyntaxText: Hashable {
+  public static func ==(lhs: SyntaxText, rhs: SyntaxText) -> Bool {
+    if lhs.buffer.count != rhs.buffer.count {
+      return false
+    }
+    if lhs.buffer.baseAddress == rhs.buffer.baseAddress {
+      return true
+    }
+    return compareMemory(lhs.baseAddress!, rhs.baseAddress!, lhs.count)
+  }
+
+  public func hash(into hasher: inout Hasher) {
+    hasher.combine(bytes: .init(buffer))
+  }
+}
+
+extension SyntaxText: ExpressibleByStringLiteral {
+  public init(stringLiteral value: StaticString) { self.init(value) }
+  public init(unicodeScalarLiteral value: StaticString) { self.init(value) }
+  public init(extendedGraphemeClusterLiteral value: StaticString) { self.init(value) }
+}
+
+extension SyntaxText: CustomStringConvertible {
+  public var description: String { String(syntaxText: self) }
+}
+
+extension SyntaxText: CustomDebugStringConvertible {
+  public var debugDescription: String { description.debugDescription }
+}
+
+extension String {
+  /// Creates a `String` from a `SyntaxText`.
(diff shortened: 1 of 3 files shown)


rintaro:
@swift-ci Please test

Sources/SwiftSyntax/SyntaxText.swift
```diff
@@ -0,0 +1,226 @@
...
+import Glibc
+#endif
+
+/// Represent a string.
+///
+/// This type does not own the string data. The data reside in some other buffer
+/// whose lifetime extends past that of the SyntaxText.
+///
+/// `SyntaxText` conforms to `Collection` where the element is `UInt8`.
+@_spi(Testing) // SPI name is subject to change
```
rintaro, line 25: `@_spi` is just to prevent clients to start relying on this for now.

rintaro:
@swift-ci Please test

Sources/SwiftSyntax/SyntaxText.swift
```diff
@@ -0,0 +1,226 @@
...
+  /// Returns `true` if the memory range of this string is a part of `other`.
+  ///
+  /// `text[n ..< m].isSliceOf(text)` is always true as long as `n` and `m` are
+  /// valid indices.
+  public func isSliceOf(_ other: SyntaxText) -> Bool {
+    guard !isEmpty && !other.isEmpty else {
+      return isEmpty == other.isEmpty
+    }
+    return (other.baseAddress! <= baseAddress! &&
+            baseAddress! + count <= other.baseAddress! + other.count)
```
ahoppen, line 69: Just wondering whether this would read easier as
```suggestion
    guard if isEmpty == other.isEmpty {
      return false
    }

    if isEmpty {
      return true
    } else {
      return (other.baseAddress! <= baseAddress! &&
           baseAddress! + count <= other.baseAddress! + other.count)
    }
```

Sources/SwiftSyntax/SyntaxText.swift
```diff
@@ -0,0 +1,226 @@
...
+  }
+
+  /// Returns `true` if `other` is a substring of this `SyntaxText`.
+  ///
+  /// This should not be used because `SyntaxText` is a `Collection<UInt8>` and
+  /// `contains(_:)` is usually for checking a collection contains the element.
+  /// Using `contains(_:)` for subsequence is confusing.
+  @available(*, deprecated, message: "Use 'firstRange(of:) != nil'")
```
ahoppen, line 77: Why do we need this function if we just deprecate it straight away?

Sources/SwiftSyntax/SyntaxText.swift
```diff
@@ -0,0 +1,226 @@
...
+  /// Finds and returns the range of the first occurrence of `other` within this
+  /// string. Returns `nil` if `other` is not found.
+  public func firstRange(of other: SyntaxText) -> Range<Index>? {
+    if other.isEmpty { return nil }
+    let needle = other.baseAddress!
+    let stop = self.count - other.count
+    var start = 0
+    // If 'other' is longer than 'self', stop < 0.
+    while start <= stop {
+      if compareMemory(self.baseAddress! + start, needle, other.count) {
```
ahoppen, line 91: Any reason why we need needle and can’t just do `other.baseAddress` here?

Sources/SwiftSyntax/SyntaxText.swift
```diff
@@ -0,0 +1,226 @@
...
+  /// Using `contains(_:)` for subsequence is confusing.
+  @available(*, deprecated, message: "Use 'firstRange(of:) != nil'")
+  public func contains(_ other: SyntaxText) -> Bool {
+    return firstRange(of: other) != nil
+  }
+
+  /// Finds and returns the range of the first occurrence of `other` within this
+  /// string. Returns `nil` if `other` is not found.
+  public func firstRange(of other: SyntaxText) -> Range<Index>? {
+    if other.isEmpty { return nil }
```
ahoppen, line 85: Don’t we also need to check that `self` is not empty?

```suggestion
    if isEmpty || other.isEmpty { return nil }
```

Sources/SwiftSyntax/SyntaxText.swift
```diff
@@ -0,0 +1,226 @@
...
+extension SyntaxText: RandomAccessCollection {
+  public typealias Element = UInt8
+  public typealias Index = Int
+  public typealias SubSequence = SyntaxText
+
+  public var startIndex: Index { buffer.startIndex }
+  public var endIndex: Index { buffer.endIndex }
+
+  public subscript(bounds: Range<Index>) -> SyntaxText {
+    if isEmpty && bounds.isEmpty { return self }
```
ahoppen, line 125: Do you think it would be worth adding an 
```swift
assert(bounds.upperBound < count) 
```
here to catch any memory issues?

Sources/SwiftSyntax/SyntaxText.swift
```diff
@@ -0,0 +1,226 @@
...
+    return SyntaxText(
+      baseAddress: buffer.baseAddress! + bounds.lowerBound,
+      count: bounds.count)
+  }
+
+  public subscript(position: Index) -> Element {
+    unsafeAddress { buffer.baseAddress!.advanced(by: position) }
```
ahoppen, line 132: Just out of curiosity: What does `unsafeAddress` do here?

Sources/SwiftSyntax/SyntaxText.swift
```diff
@@ -0,0 +1,226 @@
...
+  public init(_ string: StaticString) {
+    self.init(baseAddress: string.utf8Start, count: string.utf8CodeUnitCount)
+  }
+
+  /// Base address of the memory range this string refers to.
+  public var baseAddress: UnsafePointer<UInt8>? {
+    buffer.baseAddress
+  }
+
+  /// Byte length of this string.
```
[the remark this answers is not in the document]
rintaro: `SytnaxText` "expects" that the content is UTF8, but it's not guaranteed. It's just a collection of `UInt8` at this point. This must represents the exact source buffer which might contains ill-formed UTF8 sequence. Only when you convert them to `String`, the ill-formed sequences are converted to `\u{FFFD}`

Sources/SwiftSyntax/SyntaxText.swift
rintaro: I just wanted to avoid optional unwrapping in this hot-path, but it also applies to `self.baseAddress!`... I will update this.

ahoppen: Is there a measurable performance overhead associated with force unwrapping? Or is this just speculative optimization?

Sources/SwiftSyntax/SyntaxText.swift
rintaro: I will, because I will so `let selfBaseAddress = self.baseAddress!` outside the loop, but it was not needed because `self` was empty, `while` condition is never satisfied.

Sources/SwiftSyntax/SyntaxText.swift
rintaro: Ah, sure.

Sources/SwiftSyntax/SyntaxText.swift
rintaro: I added this because `llvm::StringRef` has `contains(StringRef Other)` method and I feel that is a common operation. When C++ compiler contributor starts to work on this project, they would expect `SyntaxText` has `contains(_: Self)` method.

Sources/SwiftSyntax/SyntaxText.swift
ahoppen: I see. I thought it was a function and couldn’t find its definition.

Sources/SwiftSyntax/SyntaxText.swift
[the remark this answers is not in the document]
ahoppen: Do you think you could add this explanation to the doc comment?

Sources/SwiftSyntax/SyntaxText.swift
ahoppen: I see. I’m not sure if it makes too much sense now anymore though, as the type is no longer named  `StringRef`. I think this should either be a properly supported method or removed.

Sources/SwiftSyntax/SyntaxText.swift
```diff
@@ -0,0 +1,226 @@
...
+private func copyMemory(
+  _ dst: UnsafeMutablePointer<UInt8>, _ src: UnsafePointer<UInt8>, _ count: Int
+) {
+  assert(count > 0)
+#if canImport(Darwin)
+  Darwin.memcpy(dst, src, count)
+#elseif canImport(Glibc)
+  Glibc.memcpy(dst, src, count)
+#else
+  dst.initialize(from: src, count: count)
```
CodaFi, line 224: UnsafeMutablePointer<UInt8>.initialize(from:count:) folds to a `memcpy` at -O.

Sources/SwiftSyntax/SyntaxText.swift
```diff
@@ -0,0 +1,226 @@
...
+    guard !other.isEmpty else { return true }
+    return self[(self.count - other.count)..<self.count] == other
+  }
+}
+
+/// `SyntaxText` is a collection of `UInt8`.
+extension SyntaxText: RandomAccessCollection {
+  public typealias Element = UInt8
+  public typealias Index = Int
+  public typealias SubSequence = SyntaxText
```
CodaFi, line 119: We may want to use `Slice<SyntaxText>` to satisfy the indices sharing requirement of `SubSequence`:

> The subsequence shares indices with the original collection.

`Data` suffers from this by being a self-slicing type with integer indices. So you can slice it and suddenly `data[0] != slice[0]` if the base of your slice is `> 0`.

Sources/SwiftSyntax/SyntaxText.swift
```diff
@@ -0,0 +1,226 @@
...
+#elseif canImport(Glibc)
+  return Glibc.memcmp(s1, s2, count) == 0
+#else
+  return UnsafeBufferPointer(start: s1, count: count)
+    .elementsEqual(UnsafeBufferPointer(start: s2, count: count))
+#endif
+}
+
+private func copyMemory(
+  _ dst: UnsafeMutablePointer<UInt8>, _ src: UnsafePointer<UInt8>, _ count: Int
```
CodaFi, line 216: A little QoL nit: Please add `from`, `to`, and `count` labels here. I always have to put a little extra thought into the order of arguments to memcpy/memcmp/memmove.

Sources/SwiftSyntax/SyntaxText.swift
```diff
@@ -0,0 +1,226 @@
...
+  /// A Boolean value indicating whether a string has no characters.
+  public var isEmpty: Bool {
+    buffer.isEmpty
+  }
+
+  /// Returns `true` if the memory range of this string is a part of `other`.
+  ///
+  /// `text[n ..< m].isSliceOf(text)` is always true as long as `n` and `m` are
+  /// valid indices.
+  public func isSliceOf(_ other: SyntaxText) -> Bool {
```
CodaFi, line 64: Nit: `isSlice(of other: SyntaxText)`

Sources/SwiftSyntax/SyntaxText.swift
rintaro: Oh, I didn't know the requirement. Thanks!
So I guess we need to add `init(rebasing: Sebsequence)`.

Sources/SwiftSyntax/SyntaxText.swift
rintaro: But I'm going stop using this for now. Let's measure the performance impact later.

Sources/SwiftSyntax/SyntaxText.swift
rintaro: Oh thanks!

Sources/SwiftSyntax/SyntaxText.swift
rintaro: Oh I was wrong. [Swift 5.7 is actually adding `contains(_: Collection<Self.Element>)`](https://github.com/apple/swift-experimental-string-processing/blob/e47cc63a0a99b76ced6279292fe30db3979f4ff0/Sources/_StringProcessing/Algorithms/Algorithms/Contains.swift#L24-L37). So I'm going to remove `@available` here and leave this as a specialized method.

Sources/SwiftSyntax/SyntaxText.swift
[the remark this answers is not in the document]
rintaro: Done.

Sources/SwiftSyntax/SyntaxText.swift
rintaro: I ended up with:
```swift
  public func isSlice(of other: SyntaxText) -> Bool {
    // If either of it is empty, returns 'true' only if both are empty.
    // Otherwise, returns 'false'.
    }
    return (other.baseAddress! <= self.baseAddress! &&
            self.baseAddress! + count <= other.baseAddress! + other.count)
  }
```

Sources/SwiftSyntax/SyntaxText.swift
rintaro: Was just speculative. Since there is not much difference in the generated code, I ended up with
`self.baseAddress! + start, other.baseAddress!` We will revisit this for performance optimization anyway.

Sources/SwiftSyntax/SyntaxText.swift
rintaro: Since I stoped hoisting optional unwrapping. I left this as-is. with some code comments.

Sources/SwiftSyntax/SyntaxText.swift
rintaro: I removed `subscript(bounds: Range<Index>)` because the default implementation is provided as I changed the `Subsequence` to `Slice<Self>`

Sources/SwiftSyntax/SyntaxText.swift
ahoppen: Makes sense 👍

Sources/SwiftSyntax/SyntaxText.swift
rintaro: I removed `copyMemory()` because I think `buffer.initialize(from:)` is enough.

Sources/SwiftSyntax/SyntaxText.swift
rintaro: Renamed!

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

CodaFi (approved):
❤️

ahoppen (approved):

rintaro:
Added a commit to fix an assertion failure case

rintaro:
@swift-ci Please test

PR #518, opened 2022-07-26
Add a convenience initializer for `IfStmt` (optionally taking a second trailing builder closure)

fwcd:
As proposed in https://github.com/apple/swift-syntax/pull/506#discussion_r925970287, this adds a new convenience initializer for `IfStmt` making both `if` and `if-else` statements nicely expressible using (multiple) trailing builder closures:

```swift
IfStmt(conditions: ...) {
  // if-body
}

IfStmt(conditions: ...) {
  // if-body
} elseBody: {
  // else-body
}
```

### DIFF
--- a/Sources/SwiftSyntaxBuilder/IfStmtConvenienceInitializers.swift
+++ b/Sources/SwiftSyntaxBuilder/IfStmtConvenienceInitializers.swift
@@ -1,0 +1,34 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2022 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
+public extension IfStmt {
+  /// A convenience initializer that uses builder closures to express an
+  /// if body, potentially with a second trailing builder closure for an else
+  /// body.
+  init(
+    leadingTrivia: Trivia = [],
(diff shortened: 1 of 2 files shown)


fwcd:
@swift-ci please test

Sources/SwiftSyntaxBuilder/IfStmtConvenienceInitializers.swift
```diff
@@ -0,0 +1,34 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
+public extension IfStmt {
+  /// A convenience initializer that uses builder closures to express an
+  /// if body, potentially with a second trailing builder closure for an else
+  /// body.
+  init(
```
ahoppen, line 19: I’m not sure if we have other types that take an optional parameter that’s expressible by a result builder but do you think it might be worth generalizing this case and putting it into the generated code?

fwcd: That's an interesting question, I am not sure whether this would generalize well though. AFAIK Swift doesn't have many if-else-ish constructs with multiple blocks. Aside from that, I don't think we're actively discriminating against multiple closures currently: The else-body in `IfStmt` doesn't seem to be a `CodeBlock`, but a `Syntax`, which also is the reason why an initializer of this form isn't autogenerated (I am not entirely sure why this is the case since I can't think of a case where an `else` wouldn't be followed by a `CodeBlock`, but that is also the reason why we need to perform this wrapping explicitly in this convenience initializer).

ahoppen: I see. Looking at `gyb_syntax_support`, the else body could also be an `IfStmt` to represent an `else if` clause. With that in mind, I think this PR makes perfect sense 👍

ahoppen (approved):

PR #521, opened 2022-07-27
Improvements to build-script.py

ahoppen:
- Add type annotations to functions
- Lint the build-script using flake8
- Make sure that we always call `check_call` for subprocesses to make sure build-script.py fails with a non-zero exit code if any subprocess fails.

### DIFF
--- a/build-script.py
+++ b/build-script.py
@@ -6,6 +6,7 @@ import platform
 import subprocess
 import sys
 import tempfile
+from concurrent.futures import ThreadPoolExecutor
 from typing import Dict, List, Optional
 
 
@@ -206,14 +207,7 @@ def generate_gyb_files_helper(
     clear_gyb_files_from_previous_run(
         sources_dir, destination_dir, verbose)
 
-    # Generate the new .swift files in `temp_files_dir` and only copy them
-    # to `destiantion_dir` if they are different than the
-    # files already residing there. This way we don't touch the generated .swift
-    # files if they haven't changed and don't trigger a rebuild.
-    for gyb_file in os.listdir(sources_dir):
-        if not gyb_file.endswith(".gyb"):
-            continue
-
+    def generate_gyb_file(gyb_file):
         gyb_file_path = os.path.join(sources_dir, gyb_file)
 
         # Slice off the '.gyb' to get the name for the output file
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please smoke test

ahoppen:
@swift-ci Please test

kimdv (approved):

Sources/SwiftSyntax/TraversalMode.swift
```diff
@@ -0,0 +1,8 @@
+//
+//  File.swift
+//  
+//
+//  Created by Alex Hoppen on 27.7.22.
```
kimdv, line 5: Sorry, but reviewing on my phone.

But is this file used?

kimdv (approved):
Single question about a file.
Otherwise does it look good 🙌

Sources/SwiftSyntax/TraversalMode.swift
ahoppen: Oh, that was committed accidentally. It belongs to a different PR. Thanks for catching it.

ahoppen:
I also added a new improvement now to run gyb-generation on multiple threads, speeding it up by 3x from ~12s to ~4s.

ahoppen:
@swift-ci Please test

build-script.py
```diff
@@ -166,36 +184,30 @@ def generate_single_gyb_file(
...
 
-    # Generate the new .swift files in `temp_files_dir` and only copy them
-    # to `destiantion_dir` if they are different than the
-    # files already residing there. This way we don't touch the generated .swift
-    # files if they haven't changed and don't trigger a rebuild.
-    for gyb_file in os.listdir(sources_dir):
-        if not gyb_file.endswith(".gyb"):
-            continue
-
+    def generate_gyb_file(gyb_file):
```
kimdv, line 210: You didn't add a type on `gyb_file`?

kimdv (approved):
Just a small nit-pick 🙌

build-script.py
```diff
@@ -505,36 +562,27 @@ def verify_c_syntax_nodes_match():
...
+    release: bool, filecheck_exec: Optional[str], skip_lit_tests: bool, verbose: bool
 ):
     print("** Running SwiftSyntax Tests **")
 
-    if skip_lit_tests:
-        lit_success = True
-    else:
-        lit_success = run_lit_tests(
+    if not skip_lit_tests:
+        run_lit_tests(
```
rintaro, line 571: I know this is the current behavior but why we don't run xctest cases when lit-tests fails? I feel we should run them both unconditionally, and raise a failure when either of them fails.

build-script.py
ahoppen: Thanks, good catch 👍

build-script.py
ahoppen: I think there are three reasons:
1. I like the way build-script fails early. That way the log output of lit-based tests are closer to the bottom (and thus easier to read) than when we continue to execute xctest commands
2. It follows the current paradigm of the build-script that if one command fails it shouldn’t execute any other commands
3. It’s slightly easier to implement this way

If you have strong opinions on this, we can change it but I like it the way it is.

ahoppen:
@swift-ci Please test

build-script.py
rintaro: I see. I agree that it's nice we can see the failures closer to the bottom. I still think it's also nice too all failures for both test suites, but it's not a strong opinion, and I can skip lit tests if I want. So let's keep it as it is.

rintaro (approved):

PR #526, opened 2022-07-27
Make `ExprBuildable` expressible as `InitializerClause`

fwcd:
This generalizes #523 slightly, letting the DSL accept expressions anywhere an `InitializerClause` is required (e.g. also in `OptionalBindingCondition` or `PatternBinding`):

```swift
// let x: String = "abc"
OptionalBindingCondition(
  letOrVarKeyword: .let,
  pattern: "x",
  typeAnnotation: "String",
  initializer: StringLiteralExpr("abc") // <- no longer needs an explicit InitializerClause wrap
)
```


fwcd:
@swift-ci please test

ahoppen (approved):

PR #527, opened 2022-07-27
Make `SyntaxBuildableType` hashable

fwcd:
This is a small change that implements `__hash__` for `SyntaxBuildableType` on the Python-side, thereby making it possible to use `SyntaxBuildableType` in sets and dictionaries as keys.

Since the type already implements a fieldwise `__eq__`, having a corresponding `__hash__` feels like a natural extension.

> Note that the Swift version of this type (in `SwiftSyntaxBuilderGeneration`) is also `Hashable`.


fwcd:
@swift-ci please test

ahoppen (approved):

PR #528, opened 2022-07-27
[Experimental] Make `Array` expressible as every collection node via dynamism

fwcd:
**This PR is intended to be a starting point for discussion, not a final solution for now.**

As mentioned in a few earlier PRs (see https://github.com/apple/swift-syntax/pull/506#discussion_r925962691, https://github.com/apple/swift-syntax/pull/506#discussion_r925968494, https://github.com/apple/swift-syntax/pull/465#discussion_r899162231), it would be cool if we could generalize #461 to generate `Array` conformances for every collection node. This has proven a bit tricky since the tree of `ExpressibleAs` superprotocols spans some types that are not collection nodes themselves. An example of this is the following conformance hierarchy (omitting the `ExpressibleAs` prefix for clarity):

```
                               +-> ConditionElementList
                               |
ExprList -> ConditionElement --+
                               |
                               +-> SyntaxBuildable
                                |
CodeBlockItemList -> CodeBlock -+
```

Ideally we would like to have

```swift
extension Array: ExpressibleAsExprList where Element == ExpressibleAsExprBuildable { ... }
extension Array: ExpressibleAsCodeBlockItemList where Element == ExpressibleAsCodeBlockItem { ... }
```

To make these retroactive conformances valid, however, we would need to conform `Array` to the entire tree of superprotocols, in particular we would need the following two conformances:

```swift
extension Array: ExpressibleAsSyntaxBuildable where Element == ExpressibleAsExprBuildable { ... }
extension Array: ExpressibleAsSyntaxBuildable where Element == ExpressibleAsCodeBlockItem { ... }
```

Unfortunately, Swift doesn't let us add separate protocol implementations that only differ in their generic constraints. For this reason, this PR explores a slightly different approach that uses dynamic casting to work around this particular limitation of the type system:

```swift
extension Array: ExpressibleAsSyntaxBuildable {
  public func createSyntaxBuildable() -> SyntaxBuildable {
    if let self = self as? [ExpressibleAsCodeBlockItem] {
      return CodeBlockItemList(self).createSyntaxBuildable()
    }
    if let self = self as? [ExpressibleAsExprBuildable] {
      return ExprList(self).createSyntaxBuildable()
    }
    ...
    fatalError("Cannot create SyntaxBuildable from this array")
  }
}
```

A particular drawback of this solution, besides performance would be that many semantically invalid arrays would now typecheck and first fail at runtime, e.g.

```swift
struct EntirelyUnrelatedType {}
let x: ExpressibleAsExprList = [EntirelyUnrelatedType()]
```

I would therefore be very interested in hearing your thoughts about this, whether such a tradeoff would be acceptable and perhaps more type-safe ideas on how to solve this.

cc @ahoppen

### DIFF
--- a/Sources/SwiftSyntaxBuilder/BuildableCollectionNodes.swift.gyb
+++ b/Sources/SwiftSyntaxBuilder/BuildableCollectionNodes.swift.gyb
@@ -1,6 +1,6 @@
 %{
   from gyb_syntax_support import SYNTAX_NODES
-  from gyb_helpers import flat_documentation, SyntaxBuildableChild, SyntaxBuildableNode
+  from gyb_helpers import flat_documentation, conformance_clause, SyntaxBuildableChild, SyntaxBuildableNode
   # -*- mode: Swift -*-
   # Ignore the following admonition it applies to the resulting .swift file only
 }%
@@ -20,7 +20,9 @@
 
 import SwiftSyntax
 
-% for node in [SyntaxBuildableNode(syntax_node) for syntax_node in SYNTAX_NODES if syntax_node.is_syntax_collection()]:
+% collection_nodes = [SyntaxBuildableNode(syntax_node) for syntax_node in SYNTAX_NODES if syntax_node.is_syntax_collection()]
+%
+% for node in collection_nodes:
 %   type = node.type()
 %   element_type = node.collection_element_type()
 %   if node.documentation():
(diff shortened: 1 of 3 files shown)


ahoppen:
I think the dynamic cast implementation that does runtime checks is an option because we lose type safety, which is a major feature of SwiftSyntaxBuilder. Unfortunately, I don’t have any better ideas of how to represent this in the type system either…

fwcd:
Yeah, that was my feeling as well, verbosity is probably preferable to losing type-safety.

PR #529, opened 2022-07-28
Add 'BumpPtrAllocator'

rintaro:
For future usage. The implementation is basically Swift copy of `llvm::BumpPtrAllocator`, but without customizable slab size etc.

### DIFF
--- a/Sources/SwiftSyntax/BumpPtrAllocator.swift
+++ b/Sources/SwiftSyntax/BumpPtrAllocator.swift
@@ -1,0 +1,114 @@
+//===---------- BumpPtrAllocator.swift - Bump-pointer Allocaiton ----------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2022 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// Bump-pointer allocation.
+struct BumpPtrAllocator {
+  typealias Slab = UnsafeMutableRawBufferPointer
+
+  static private var SLAB_SIZE: Int = 4096
+  static private var GLOWTH_DELAY: Int = 128
+  static private var SLAB_ALIGNMENT: Int = 8
+
+  private var slabs: [Slab]
+  private var currentPtr: UnsafeMutableRawPointer?
+  private var endPtr: UnsafeMutableRawPointer?
+  private var customSizeSlabs: [Slab]
+  private(set) var totalSizeAllocated: Int
+
+  init() {
+    slabs = []
+    currentPtr = nil
+    endPtr = nil
+    customSizeSlabs = []
+    totalSizeAllocated = 0
+  }
+
+  /// Calculate the size of the slab at the index.
+  private static func slabSize(at index: Int) -> Int {
+    // Double the slab size every 'GLOWTH_DELAY' slabs.
+    return SLAB_SIZE * (1 << min(30, index / GLOWTH_DELAY))
+  }
+
+  private mutating func startNewSlab() {
+    let newSlabSize = Self.slabSize(at: slabs.count)
+    let newSlab = Slab.allocate(
+      byteCount: newSlabSize, alignment: Self.SLAB_ALIGNMENT)
+    slabs.append(newSlab)
+    currentPtr = newSlab.baseAddress!
+    endPtr = currentPtr!.advanced(by: newSlabSize)
+  }
+
+  /// Allocate 'byteCount' of memory.
+  mutating func allocate(byteCount: Int, alignment: Int) -> UnsafeMutableRawBufferPointer {
+
+    precondition(alignment <= Self.SLAB_ALIGNMENT)
+    guard byteCount > 0 else {
+      return .init(start: nil, count: 0)
+    }
+
+    // Track the total size we allocated.
+    totalSizeAllocated += byteCount
+
+    // Check if the current slab have enough space.
+    if !slabs.isEmpty {
+      let aligned = currentPtr!.alignedUp(toMultipleOf: alignment)
+      if (aligned + byteCount <= endPtr!) {
+        currentPtr = aligned + byteCount
+        return .init(start: aligned, count: byteCount)
+      }
+    }
+
+    // If the size is too big, allocate a dedicated slab for it.
+    if (byteCount >= Self.SLAB_SIZE) {
+      let customSlab = Slab.allocate(
+        byteCount: byteCount, alignment: alignment)
+      customSizeSlabs.append(customSlab)
+      return customSlab
+    }
+
+    // Otherwise, start a new slab and try again.
+    startNewSlab()
+    let aligned = currentPtr!.alignedUp(toMultipleOf: alignment)
+    assert(aligned + byteCount <= endPtr!, "Unable to allocate memory!")
+    currentPtr = aligned + byteCount
+    return .init(start: aligned, count: byteCount)
+  }
+
+  /// Allocate a chunk of memory of `MemoryLayout<T>.stride * count'.
+  mutating func allocate<T>(_: T.Type, count: Int) -> UnsafeMutableBufferPointer<T> {
+
+    let allocated = allocate(byteCount: MemoryLayout<T>.stride * count,
+                             alignment: MemoryLayout<T>.alignment)
+    return allocated.bindMemory(to: T.self)
+  }
+
+  /// Check if the address is managed by this allocator.
+  func contains(address: UnsafeRawPointer) -> Bool {
+    func test(_ slab: Slab) -> Bool {
+      slab.baseAddress! <= address && address < slab.baseAddress! + slab.count
+    }
+    return slabs.contains(where: test(_:)) || customSizeSlabs.contains(where: test(_:))
+  }
+
+  /// Deallocate all memory.
+  mutating func reset() {
+    totalSizeAllocated = 0
+    currentPtr = nil
+    endPtr = nil
+    while !slabs.isEmpty {
+      slabs.removeLast().deallocate()
+    }
+    while !customSizeSlabs.isEmpty {
+      customSizeSlabs.removeLast().deallocate()
+    }
+  }
+}
--- a/Sources/SwiftSyntax/BumpPtrAllocator.swift
+++ b/Sources/SwiftSyntax/BumpPtrAllocator.swift
@@ -11,7 +11,8 @@
 //===----------------------------------------------------------------------===//
 
 /// Bump-pointer allocation.
-class BumpPtrAllocator {
+@_spi(Testing)
+public class BumpPtrAllocator {
   typealias Slab = UnsafeMutableRawBufferPointer
 
   static private var SLAB_SIZE: Int = 4096
@@ -19,29 +20,32 @@ class BumpPtrAllocator {
   static private var SLAB_ALIGNMENT: Int = 8
 
   private var slabs: [Slab]
+  /// Points to the next unused address in `slabs.last`.
   private var currentPtr: UnsafeMutableRawPointer?
+  /// Points to the end address of `slabs.last`. If `slabs` is not empty, this
+  /// is equivalent to `slabs.last!.baseAddress! + slabs.last!.count`
   private var endPtr: UnsafeMutableRawPointer?
   private var customSizeSlabs: [Slab]
-  private(set) var totalSizeAllocated: Int
+  private var _totalBytesAllocated: Int
 
-  init() {
+  public init() {
     slabs = []
     currentPtr = nil
     endPtr = nil
     customSizeSlabs = []
-    totalSizeAllocated = 0
+    _totalBytesAllocated = 0
   }
 
   deinit {
     /// Deallocate all memory.
-    totalSizeAllocated = 0
+    _totalBytesAllocated = 0
     currentPtr = nil
     endPtr = nil
-    while !slabs.isEmpty {
-      slabs.removeLast().deallocate()
+    while let slab = slabs.popLast() {
+      slab.deallocate()
     }
-    while !customSizeSlabs.isEmpty {
-      customSizeSlabs.removeLast().deallocate()
+    while let slab = customSizeSlabs.popLast() {
+      slab.deallocate()
     }
   }
 
@@ -61,7 +65,10 @@ class BumpPtrAllocator {
   }
 
   /// Allocate 'byteCount' of memory.
-  func allocate(byteCount: Int, alignment: Int) -> UnsafeMutableRawBufferPointer {
+  ///
+  /// The returned buffer is not bound to any types, nor initialized.
+  /// Clients should never call `deallocate()` on the returned buffer.
+  public func allocate(byteCount: Int, alignment: Int) -> UnsafeMutableRawBufferPointer {
 
     precondition(alignment <= Self.SLAB_ALIGNMENT)
     guard byteCount > 0 else {
@@ -69,9 +76,9 @@ class BumpPtrAllocator {
     }
 
     // Track the total size we allocated.
-    totalSizeAllocated += byteCount
+    _totalBytesAllocated += byteCount
 
-    // Check if the current slab have enough space.
+    // Check if the current slab has enough space.
     if !slabs.isEmpty {
       let aligned = currentPtr!.alignedUp(toMultipleOf: alignment)
       if aligned + byteCount <= endPtr! {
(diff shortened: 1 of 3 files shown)


Tests/SwiftSyntaxTest/BumpPtrAllocatorTests.swift
```diff
@@ -0,0 +1,39 @@
+#if DEBUG
+
+import XCTest
+@testable import SwiftSyntax
```
rintaro, line 4: Since `BumpPtrAllocator` should never be a public interface (even as a SPI), run tests only when built with `@testable`

Sources/SwiftSyntax/BumpPtrAllocator.swift
```diff
@@ -0,0 +1,114 @@
...
+// Copyright (c) 2014 - 2022 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// Bump-pointer allocation.
+struct BumpPtrAllocator {
```
CodaFi, line 14: I think this ought to be a class. Each allocator has a distinct identity, and we want to be able to deallocate slabs when the last reference to the allocator is gone.

Sources/SwiftSyntax/BumpPtrAllocator.swift
```diff
@@ -0,0 +1,114 @@
...
+      return .init(start: nil, count: 0)
+    }
+
+    // Track the total size we allocated.
+    totalSizeAllocated += byteCount
+
+    // Check if the current slab have enough space.
+    if !slabs.isEmpty {
+      let aligned = currentPtr!.alignedUp(toMultipleOf: alignment)
+      if (aligned + byteCount <= endPtr!) {
```
CodaFi, line 64: Nit: Extraneous parens

Sources/SwiftSyntax/BumpPtrAllocator.swift
```diff
@@ -0,0 +1,114 @@
...
+        currentPtr = aligned + byteCount
+        return .init(start: aligned, count: byteCount)
+      }
+    }
+
+    // If the size is too big, allocate a dedicated slab for it.
+    if (byteCount >= Self.SLAB_SIZE) {
```
CodaFi, line 71: Same here

CodaFi: This check needs to take into account that the requested size will differ from the allocation size.

Sources/SwiftSyntax/BumpPtrAllocator.swift
```diff
@@ -0,0 +1,114 @@
...
+  /// Check if the address is managed by this allocator.
+  func contains(address: UnsafeRawPointer) -> Bool {
+    func test(_ slab: Slab) -> Bool {
+      slab.baseAddress! <= address && address < slab.baseAddress! + slab.count
+    }
+    return slabs.contains(where: test(_:)) || customSizeSlabs.contains(where: test(_:))
+  }
+
+  /// Deallocate all memory.
+  mutating func reset() {
```
CodaFi, line 103: This is a profoundly dangerous operation. Do we need it?

Sources/SwiftSyntax/BumpPtrAllocator.swift
CodaFi: Same comment here: The requested byte count and the actual aligned allocation size are different and the aligned allocation size should be the subject of this check.

Sources/SwiftSyntax/BumpPtrAllocator.swift
CodaFi: Although maybe it doesn't. The allocation will always be at least as large as the requested size given alignment constraints...

CodaFi (commented):
Mostly some nits. Overall it looks pretty good

Sources/SwiftSyntax/BumpPtrAllocator.swift
rintaro: Eh, I was planning to use this as a member of `SyntaxArena` that is a class. in #512 . But yeah, I will change this to class.

Sources/SwiftSyntax/BumpPtrAllocator.swift
rintaro: I was planning to call this from SytnaxArena's `deinit`, but I will move this to this own `deinit` as I'm changing this to a class

Sources/SwiftSyntax/BumpPtrAllocator.swift
rintaro: Since I limited the max alignment to the slab alignment,
```
    precondition(alignment <= Self.SLAB_ALIGNMENT)
```
I thought we don't need to consider the alignment here. Am I missing anything?

CodaFi (approved):

rintaro:
@swift-ci Please test

Sources/SwiftSyntax/BumpPtrAllocator.swift
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  deinit {
+    /// Deallocate all memory.
+    totalSizeAllocated = 0
+    currentPtr = nil
+    endPtr = nil
+    while !slabs.isEmpty {
+      slabs.removeLast().deallocate()
+    }
```
ahoppen, line 42: What’s the reason that you need to remove the slabs from the array yourself and can’t just do the following?
```swift
for slab in slabs {
  slab.deallocate()
}
```

Sources/SwiftSyntax/BumpPtrAllocator.swift
```diff
@@ -0,0 +1,113 @@
...
+class BumpPtrAllocator {
+  typealias Slab = UnsafeMutableRawBufferPointer
+
+  static private var SLAB_SIZE: Int = 4096
+  static private var GLOWTH_DELAY: Int = 128
+  static private var SLAB_ALIGNMENT: Int = 8
+
+  private var slabs: [Slab]
+  private var currentPtr: UnsafeMutableRawPointer?
+  private var endPtr: UnsafeMutableRawPointer?
```
ahoppen, line 23: Could you add comments to these?

```suggestion
  /// Points to the next unused address in `slabs.last`
  private var currentPtr: UnsafeMutableRawPointer?
  /// Points to the end address of `slabs.last`. If `slabs` is not empty, this is equivalent to
  /// `slabs.last!.baseAddress! + slabs.last!.count`
  private var endPtr: UnsafeMutableRawPointer?
```

Also, I wonder if it makes sense to start an initial slab when the bump allocator is created so these can be non-optional. I think the use case of creating a bump allocator but never allocating anything in it are very rare.

Sources/SwiftSyntax/BumpPtrAllocator.swift
```diff
@@ -0,0 +1,113 @@
...
+  private var customSizeSlabs: [Slab]
+  private(set) var totalSizeAllocated: Int
```
ahoppen, line 25: Technically, this is only an underestimation of the bump allocator’s memory footprint because if there are bytes in a slab that were (not yet) used, they don’t contribute to this count, right? Maybe it’s worth pointing out that distinction in a comment.

Sources/SwiftSyntax/BumpPtrAllocator.swift
```diff
@@ -0,0 +1,113 @@
...
+
+    precondition(alignment <= Self.SLAB_ALIGNMENT)
+    guard byteCount > 0 else {
+      return .init(start: nil, count: 0)
+    }
+
+    // Track the total size we allocated.
+    totalSizeAllocated += byteCount
+
+    // Check if the current slab have enough space.
```
ahoppen, line 74: Nitpick 
```suggestion
    // Check if the current slab has enough space.
```

Tests/SwiftSyntaxTest/BumpPtrAllocatorTests.swift
```diff
@@ -0,0 +1,43 @@
+#if DEBUG
+// 'BumpPtrAllocator' is not a public API. Test it only when the module is built
+// with @testable (i.e. `DEBUG`)
```
ahoppen, line 3: I’m not sure if this is a great idea because it means that these tests won’t be run in CI. Why are you so afraid to mark `BumpPtrAllocator` as `@_spi(Testing)`

Sources/SwiftSyntax/BumpPtrAllocator.swift
```diff
@@ -0,0 +1,113 @@
...
+    // Otherwise, start a new slab and try again.
+    startNewSlab()
+    let aligned = currentPtr!.alignedUp(toMultipleOf: alignment)
+    assert(aligned + byteCount <= endPtr!, "Unable to allocate memory!")
+    currentPtr = aligned + byteCount
+    return .init(start: aligned, count: byteCount)
+  }
+
+  /// Allocate a chunk of memory of `MemoryLayout<T>.stride * count'.
+  func allocate<T>(_: T.Type, count: Int) -> UnsafeMutableBufferPointer<T> {
```
ahoppen, line 100: Maybe I’m wrong but my assumption would be that it’s very common to allocate memory for a single element. Would it make sense to default `count` to 1?

Tests/SwiftSyntaxTest/BumpPtrAllocatorTests.swift
```diff
@@ -0,0 +1,43 @@
...
+      let typedEmptyBuffer = allocator.allocate(String.self, count: 0)
+      XCTAssertNil(typedEmptyBuffer.baseAddress)
+      XCTAssertEqual(typedEmptyBuffer.count, 0)
+
+      XCTAssertTrue(allocator.contains(address: typedBuffer.baseAddress!))
+      XCTAssertTrue(allocator.contains(address: typedBuffer.baseAddress!.advanced(by: typedBuffer.count)))
+
+      XCTAssertEqual(allocator.totalSizeAllocated,
+                     42 + MemoryLayout<UInt64>.stride * 12)
+    }
```
ahoppen, line 40: Do you think it’s worth adding a test case that shows how to allocate a Swift class on the bump allocator and pass it around? I think it would be nice to have a reduced example how to use the bump allocator in Swift.

Sources/SwiftSyntax/BumpPtrAllocator.swift
rintaro: I just don't want to keep the deallocated slabs in the array even in `deinit`. Retting `startPtr` et al, is for the same reason. I don't want to keep any dangling pointers.

Sources/SwiftSyntax/BumpPtrAllocator.swift
rintaro: :+1: for the comments.
Two reasons I didn't make these non-optional:
* Need to allocate the first slab _before_ initializing these, that means I cannot use `startNewSlab` for the first slab
* As I commented above, I want to reset these before deallocating the first slab

Sources/SwiftSyntax/BumpPtrAllocator.swift
rintaro: Yeah, you are right. We probably want two metrics like:
- `totalMemorySizeAllocatedFromSystem`: `llvm::BumpPtrAllocator.getTotalMemory()`
- `totalMemorySizeClaimedByClients`: `llvm::BumpPtrAllocator.getBytesAllocated()`
Not sure about the naming though. Any suggestions?

Sources/SwiftSyntax/BumpPtrAllocator.swift
ahoppen: OK, fair enough. It seem a little too cautious to me but I’m fine with it.

CodaFi: I mean, another way to write this and maintain the style is 

```
while let slab = slabs.popLast() {
  slab.deallocate()
}
```

Tests/SwiftSyntaxTest/BumpPtrAllocatorTests.swift
rintaro: You should not use memory allocated by `BumpPtrAllocator` to place Swift `class` because it doesn't deinitialize them automatically. You should only place strictly trivial types i.e. no classes, closures, existentials, indirect enums, etc. (I will add doc-comments about it)

I'm not sure what you are suggesting here. The usage of the allocated memory is the same as Swift's `UnsafeXXXPointer`, the only difference is that you can't call `deallocate()` to them.

Sources/SwiftSyntax/BumpPtrAllocator.swift
rintaro: Updated to `popLast()`

Sources/SwiftSyntax/BumpPtrAllocator.swift
rintaro: Ended up with mildly following `llvm::BumpPtrAllocator`.
`totalMemorySize` / `totalByteSizeAllocated`

Tests/SwiftSyntaxTest/BumpPtrAllocatorTests.swift
rintaro: It was just a personal preference. Updated to `@_spi(Testing)`

Tests/SwiftSyntaxTest/BumpPtrAllocatorTests.swift
ahoppen: OK, I think I meant a `struct` instead. I think I’m just not very familiar with the `UnsafeXXXPointer` APIs. I’m fine to also keep this out of this PR and be surprised by the next PR in which you start using the bump allocator.

Sources/SwiftSyntax/BumpPtrAllocator.swift
ahoppen: Fair enough 👍 I’m also fine to keep it as-is.

Tests/SwiftSyntaxTest/BumpPtrAllocatorTests.swift
ahoppen: We disable testable imports in CI because it would cause us to rebuild the entire package when testing it after we have already built it in a first step.

Sources/SwiftSyntax/BumpPtrAllocator.swift
ahoppen: If you think it’s better this way, that’s fine. But I imagine that we’ll be passing `1` in 90% of all cases here. But we can always revisit this later if we actually see the call sites.

Sources/SwiftSyntax/BumpPtrAllocator.swift
```diff
@@ -0,0 +1,147 @@
...
+  static private var SLAB_ALIGNMENT: Int = 8
+
+  private var slabs: [Slab]
+  /// Points to the next unused address in `slabs.last`.
+  private var currentPtr: UnsafeMutableRawPointer?
+  /// Points to the end address of `slabs.last`. If `slabs` is not empty, this
+  /// is equivalent to `slabs.last!.baseAddress! + slabs.last!.count`
+  private var endPtr: UnsafeMutableRawPointer?
+  private var customSizeSlabs: [Slab]
+  private var _totalBytesAllocated: Int
```
ahoppen, line 29: Why can’t this be `private(set) var totalByteSizeAllocated: Int` so we could remove the computed property below?

Tests/SwiftSyntaxTest/BumpPtrAllocatorTests.swift
```diff
@@ -0,0 +1,39 @@
+import XCTest
```
ahoppen, line 1: This file is missing the copyright header

rintaro: So as other test files in this repo. If we want them, let's do it in a separate PR.

Sources/SwiftSyntax/BumpPtrAllocator.swift
rintaro: Because I want `totalMemorySize` / `totalByteSizeAllocated` at the bottom of the file (because it's not a main functionality), and side by size, but I don't want to write a stored properties after the initializer.

ahoppen: OK, sounds good 👍

Tests/SwiftSyntaxTest/BumpPtrAllocatorTests.swift
ahoppen: Sounds good. I think adding these headers would be a good thing to do. I don’t see a reason why some of these files shouldn’t contain them.

ahoppen (approved):

ahoppen:
@swift-ci Please test

PR #530, opened 2022-07-29
Make `ExprList` initializable by result builder

fwcd:
Since the `Array: ExpressibleAsExprList` conformance is [unlikely to be added](https://github.com/apple/swift-syntax/pull/528#issuecomment-1197399869) due to type system limitations, here is a small patch to make expression lists a bit more convenient to use by making them builder-initializable (currently only affects `SequenceExpr` and `YieldList`).

Using this we can shorten

```swift
SequenceExpr(elements: ExprList {
  ...
})
```

to

```swift
SequenceExpr {
  ...
}
```


fwcd:
@swift-ci please test

ahoppen (approved):

PR #531, opened 2022-07-29
Build all products (instead of specifying them individually) in the build script

fwcd:
As per the suggestion in https://github.com/apple/swift-syntax/pull/477#issuecomment-1197368041, this small patch updates the build script's build command to build all products. This obviates the need to specify them manually and also makes it easier to integrate snippets into the build process in the future (see #477).


fwcd:
@swift-ci please test

ahoppen (dismissed):

fwcd:
Oh okay 🤔 Is there a way to build all products (or targets) of a package even in a multi-project workspace setting?

ahoppen:
I’m not sure. I’ll come back to you once I know.

ahoppen:
I just checked and there’s no way to explicitly build all snippets from a snippet. I think what we need to do is build each snippet explicitly. The corresponding product name is the snippet’s filename without the extension. I.e. I think we need to include the following in the build script.

```python
# Build snippets
builder.build("AddOneToIntegerLiterals")
builder.build("CodeGenerationUsingSwiftSyntaxBuilder")
```

fwcd:
@swift-ci please test

ahoppen:
rdar://99492976

fwcd:
Moved to #1069

PR #534, opened 2022-08-01
When gyb-generation fails on a subthread, raise that exception in the parent thread

ahoppen:
An exception raised in a thread managed by `ThreadPoolExecutor` only gets raised if its value is retrieved. Iterate over all the 'None' results to cause the exceptions to be propagated.

Otherwise, gyb-generation can silently fail.


ahoppen:
@swift-ci Please test

rintaro (approved):

PR #535, opened 2022-08-01
Introduce a new syntax kind to represent missing base nodes

ahoppen:
In the future, we only want to attach a source presence to tokens, not to nodes. All concreate nodes which are missing can be represented by creating the node and marking all tokens as missing, but if a syntax node carries a child, that has a base kind (like `Decl`), we can’t decide which concrete node to instantiate. Introduce `MissingDecl` etc. node for this purpose.

While at it, also remove the base kinds from `SyntaxKind` because we should never instantiate nodes of these base kinds.

rdar://97908258
rdar://97775360

### DIFF
--- a/Sources/SwiftSyntax/Misc.swift.gyb
+++ b/Sources/SwiftSyntax/Misc.swift.gyb
@@ -54,7 +54,7 @@ extension Syntax {
       return node
     case .unknown(let node):
       return node
-% for node in SYNTAX_NODES:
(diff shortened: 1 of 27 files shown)


ahoppen:
@swift-ci Please smoke test

ahoppen:
https://github.com/apple/swift/pull/60334

@swift-ci Please test

CodaFi (approved):

PR #536, opened 2022-08-01
Make IfConfigClause Elements Optional

CodaFi:
A #if clause need not actually contain any elements. Consider

```
#if !FEATURE_ENABLED
#else
class Foo {

}
#endif
```


CodaFi:
apple/swift#60342

@swift-ci test

PR #539, opened 2022-08-02
Deprecate SyntaxFactory

ahoppen:
Instead of always writing `SyntaxFactory.make*` we can just use the initializers of the nodes that should be created, which leads to a lot nicer code IMO.

CC @akyrtzi @harlanhaskins in case you remember any compelling reasons why we decided to use `SyntaxFactory` instead of initializers when this was implemented.

rdar://97910890

### DIFF
--- a/Sources/SwiftSyntax/SyntaxFactory.swift.gyb
+++ b/Sources/SwiftSyntax/SyntaxFactory.swift.gyb
@@ -28,6 +28,7 @@
 //===----------------------------------------------------------------------===//
 
 public enum SyntaxFactory {
+  @available(*, deprecated, message: "Use initializer on TokenSyntax")
   public static func makeToken(_ kind: TokenKind, presence: SourcePresence,
                                leadingTrivia: Trivia = [],
                                trailingTrivia: Trivia = []) -> TokenSyntax {
@@ -37,6 +38,7 @@ public enum SyntaxFactory {
     return TokenSyntax(data)
   }
 
+  @available(*, deprecated, message: "Use initializer on UnknownSyntax")
   public static func makeUnknownSyntax(tokens: [TokenSyntax]) -> UnknownSyntax {
     let raw = RawSyntax.createAndCalcLength(kind: .unknown,
       layout: tokens.map { $0.raw }, presence: .present)
@@ -65,6 +67,7 @@ public enum SyntaxFactory {
 %       end
 %     end
 %     child_params = ', '.join(child_params)
+  @available(*, deprecated, message: "Use initializer on ${node.name}")
   public static func make${node.syntax_kind}(${child_params}) -> ${node.name} {
     let layout: [RawSyntax?] = [
 %     for child in node.children:
@@ -81,6 +84,7 @@ public enum SyntaxFactory {
     return ${node.name}(data)
   }
 %   elif node.is_syntax_collection():
+  @available(*, deprecated, message: "Use initializer on ${node.name}")
   public static func make${node.syntax_kind}(
     _ elements: [${node.collection_element_type}]) -> ${node.name} {
     let raw = RawSyntax.createAndCalcLength(kind: SyntaxKind.${node.swift_syntax_kind},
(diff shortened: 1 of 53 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

bnbarham (approved):
LGTM assuming there was no good reason for it (I assume just an easy place to see all the nodes that could be created?)

harlanhaskins:
From what I remember in conversations with @bitjammer, the goal of `SyntaxFactory` was to make autocompletion much easier -- rather than needing to know the kinds of nodes available, you'd type `SyntaxFactory.make` and autocomplete would show you what's available.

I personally don't know if that's necessarily worth it, especially as we're moving toward a Result-builder DSL future for Swift

ahoppen:
Yes, I agree that the verbosity of `SyntaxFactory` is not really worth the improved code completion behavior. Good documentation would probably serve a better job here.

bitjammer:
That’s right. The motivation was that there were just an incredible number of node types. If the tools and documentation could be improved, or there is another approach that could provide scoped hints to code completion, there’s no need.

rintaro:
I personally like a factory methods on the type instead of `init(...)` and `blank()`
```swift
static func make(...) -> Self
static func blank() -> Self
```

One drawback of `SyntaxFactory.makeXXX` was that we couldn't use implicit member expression like
```swift
func genFunctionDecl() -> FunctionDeclSyntax {
   ...
   return .make(....)
}
```
I feel `return .make(...)` is more clean than `return .init(...)`

Sources/SwiftSyntaxBuilder/TupleExprElementConvenienceInitializers.swift
```diff
@@ -17,6 +17,6 @@ public extension TupleExprElement {
   /// The presence of the colon will be inferred based on the presence of the label.
   init(label: String? = nil, expression: ExpressibleAsExprBuildable) {
     self.init(
-      label: label.map(TokenSyntax.identifier), colon: label == nil ? nil : .colon, expression: expression)
+      label: label.map { TokenSyntax.identifier($0) }, colon: label == nil ? nil : .colon, expression: expression)
```
evnik, line 20: Isn't it equivalent to what it was?

ahoppen: The old implementation doesn’t work anymore because `TokenSyntax.identifier` now takes three arguments, two of which are defaulted. But the argument default values are not used when referencing the function by its base name.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

Sources/SwiftSyntax/SyntaxFactory.swift.gyb
```diff
@@ -119,6 +124,7 @@ public enum SyntaxFactory {
 %   leading_trivia = token_trivia(token.requires_leading_space)
 %   trailing_trivia = token_trivia(token.requires_trailing_space)
 %   if token.is_keyword:
+  @available(*, deprecated, message: "Use TokenSyntax.${token.swift_kind()}Keyword instead")
```
harlanhaskins, line 127: Nit: update this deprecation and remove the Keyword suffix

Sources/SwiftSyntax/Tokens.swift.gyb
```diff
@@ -0,0 +1,101 @@
...
+    return []
+  }
+}
+
+extension TokenSyntax {
+% for token in SYNTAX_TOKENS:
+%   leading_trivia = '.space' if token.requires_leading_space else '[]'
+%   trailing_trivia = '.space' if token.requires_trailing_space else '[]'
+%   if token.is_keyword:
+  public static func `${lowercase_first_word(token.name)}`(
```
rintaro, line 35: What's wrong with `token.swift_name()`?  I feel, for example, `TokenSyntax.is(presence: .missing)` doesn't seem to be creating a new instance.

ahoppen:
@swift-ci Please test

rintaro (approved):
We decided to go with initializers, instead of factory methods.

PR #541, opened 2022-08-02
Allow source locations relative to any node

bnbarham:
Have `SourceLocationConverter` take a `Syntax` rather than a
`SourceFileSyntax` so that *any* node can be the root. Keep the
`SourceFileSyntax` initializer as a convenience init to avoid an API
change.


bnbarham:
@swift-ci please test

bnbarham:
@swift-ci please test

ahoppen (approved):

bnbarham:
Going to skip this change, Rintaro pointed out that the semantics of passing in a non-root node don't really make sense. Tests can just write out the text and then use the normal `String` initializer anyway.

PR #542, opened 2022-08-02
Migrate from strings to typed buildable nodes in `SyntaxBuildableWrappers`

fwcd:
As proposed in https://github.com/apple/swift-syntax/pull/485#pullrequestreview-1033905250, this PR migrates the generated snippets of Swift code in `SyntaxBuildableType` and `SyntaxBuildableChild` to `ExprBuildable`s and `TypeBuildable`s.

This will pave the way for a future PR porting `BuildableNodes.swift.gyb` to `SwiftSyntaxBuilderGeneration`.

### DIFF
--- a/Sources/SwiftSyntaxBuilderGeneration/SyntaxBuildableChild.swift
+++ b/Sources/SwiftSyntaxBuilderGeneration/SyntaxBuildableChild.swift
@@ -10,6 +10,9 @@
 //
 //===----------------------------------------------------------------------===//
 
+import SwiftSyntax
+import SwiftSyntaxBuilder
+
 /// Extension to the `Child` type to provide functionality specific to
 /// SwiftSyntaxBuilder.
 extension Child {
@@ -31,17 +34,37 @@ extension Child {
   /// Generate a Swift expression that creates a proper SwiftSyntax node of type
   /// `type.syntax` from a variable named `varName` of type `type.buildable` that
   /// represents this child node.
-  func generateExprBuildSyntaxNode(varName: String, formatName: String) -> String {
+  func generateExprBuildSyntaxNode(varName: String, formatName: String) -> ExpressibleAsExprBuildable {
     if type.isToken {
       if requiresLeadingNewline {
-        return "\(varName).withLeadingTrivia(.newline + \(formatName)._makeIndent() + (\(varName).leadingTrivia ?? []))"
+        return FunctionCallExpr(MemberAccessExpr(base: varName, name: "withLeadingTrivia")) {
+          SequenceExpr {
+            MemberAccessExpr(name: "newline")
+            BinaryOperatorExpr("+")
+            FunctionCallExpr(MemberAccessExpr(base: formatName, name: "_makeIndent"))
+            BinaryOperatorExpr("+")
+            TupleExpr {
+              SequenceExpr {
+                MemberAccessExpr(base: varName, name: "leadingTrivia")
+                BinaryOperatorExpr("??")
+                ArrayExpr(elements: [])
+              }
+            }
+          }
+        }
       } else {
         return varName
       }
     } else {
-      let format = formatName + (isIndented ? "._indented()" : "")
-      let expr = "\(varName)\(type.optionalQuestionMark)"
-      return "\(expr).build\(type.baseName)(format: \(format), leadingTrivia: nil)"
+      var format: ExpressibleAsExprBuildable = formatName
+      if isIndented {
+        format = FunctionCallExpr(MemberAccessExpr(base: format, name: "_indented"))
+      }
+      let expr = type.optionalChained(expr: varName)
+      return FunctionCallExpr(MemberAccessExpr(base: expr, name: "build\(type.baseName)")) {
+        TupleExprElement(label: "format", expression: format)
+        TupleExprElement(label: "leadingTrivia", expression: "nil")
+      }
     }
   }
 
--- a/Sources/SwiftSyntaxBuilderGeneration/SyntaxBuildableChild.swift
+++ b/Sources/SwiftSyntaxBuilderGeneration/SyntaxBuildableChild.swift
@@ -63,7 +63,7 @@ extension Child {
       let expr = type.optionalChained(expr: varName)
       return FunctionCallExpr(MemberAccessExpr(base: expr, name: "build\(type.baseName)")) {
         TupleExprElement(label: "format", expression: format)
-        TupleExprElement(label: "leadingTrivia", expression: "nil")
+        TupleExprElement(label: "leadingTrivia", expression: NilLiteralExpr())
       }
     }
   }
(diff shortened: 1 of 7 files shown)


Sources/SwiftSyntaxBuilderGeneration/SyntaxBuildableType.swift
```diff
@@ -19,11 +22,6 @@ struct SyntaxBuildableType: Hashable {
   let tokenKind: String?
   let isOptional: Bool
 
-  /// If optional `?`, otherwise the empty string.
-  var optionalQuestionMark: String {
-    isOptional ? "?" : ""
-  }
```
fwcd, line 25: This property is replaced with `optionalWrapped(type:)` and `optionalChained(expr:)`, depending on whether a type or an expression is needed.

Sources/SwiftSyntaxBuilderGeneration/SyntaxBuildableType.swift
```diff
@@ -68,81 +66,98 @@ struct SyntaxBuildableType: Hashable {
...
+    }
   }
 
-  /// The `ExpressibleAs*` Swift type for this syntax kind. Tokens don't
-  /// have an `ExpressibleAs*` type, so for those this method just returns
-  /// `TokenSyntax`. If the type is optional, this terminates with a `?`.
-  var expressibleAs: String {
+  /// The `ExpressibleAs*` Swift type for this syntax kind without any
+  /// question marks attached.
+  var expressibleAsBaseName: String {
```
fwcd, line 112: This PR factors out a few `...BaseName` properties (`expressibleAs`, `syntax`, `buildable`). These represent a non-optional type and will still be strings (as opposed to, e.g. `ExpressibleAsTypeBuildable`s) to make them easy to embed in string interpolations and across type and expression contexts.

Thus `.xyzBaseName` replaces the earlier `.nonOptional.xyz` idiom.

Sources/SwiftSyntaxBuilderGeneration/Templates/BuildableBaseProtocolsFile.swift
```diff
@@ -30,67 +30,67 @@ let buildableBaseProtocolsFile = SourceFile {
     let syntaxType = SyntaxBuildableType(syntaxKind: "Syntax")
     let isSyntax = type == syntaxType
     // Types that the `*Buildable` conforms to
-    let buildableConformances: [String] = [type.expressibleAs, type.listBuildable] + (isSyntax ? [] : [syntaxType.buildable])
-    let listConformances: [String] = isSyntax ? [] : [syntaxType.listBuildable]
+    let buildableConformances: [String] = [type.expressibleAsBaseName, type.listBuildableBaseName] + (isSyntax ? [] : [syntaxType.buildableBaseName])
+    let listConformances: [String] = isSyntax ? [] : [syntaxType.listBuildableBaseName]
```
fwcd, line 34: Previously we relied on these types being non-optional. Since optional types wouldn't make sense e.g. in conformance lists and dealing with strings is easier, we use the base names now in a few more places (this shouldn't change anything in the generated code).

Sources/SwiftSyntaxBuilderGeneration/Templates/BuildableBaseProtocolsFile.swift
```diff
@@ -30,67 +30,67 @@ let buildableBaseProtocolsFile = SourceFile {
...
     ProtocolDecl(
       modifiers: [TokenSyntax.public],
-      identifier: type.listBuildable,
+      identifier: type.listBuildableBaseName,
       inheritanceClause: createTypeInheritanceClause(conformances: listConformances)
     ) {
       FunctionDecl(
         leadingTrivia: [
-          "/// Builds list of `\(type.syntax)`s.",
+          "/// Builds list of `\(type.syntaxBaseName)`s.",
```
fwcd, line 43: Same goes here (and for the other places where base names are now used).

fwcd (commented):
A couple of notes on the implementation

fwcd:
@swift-ci please test

Sources/SwiftSyntaxBuilderGeneration/SyntaxBuildableChild.swift
```diff
@@ -31,17 +34,37 @@ extension Child {
   /// Generate a Swift expression that creates a proper SwiftSyntax node of type
   /// `type.syntax` from a variable named `varName` of type `type.buildable` that
   /// represents this child node.
-  func generateExprBuildSyntaxNode(varName: String, formatName: String) -> String {
+  func generateExprBuildSyntaxNode(varName: String, formatName: String) -> ExpressibleAsExprBuildable {
```
ahoppen, line 37: I consider `ExpressibleAsExprBuildable` to be mostly used as an input type to functions, not really something that’s produced. Unless there’s a reason against it, I would prefer to return `ExprBuildable` here.

Sources/SwiftSyntaxBuilderGeneration/SyntaxBuildableChild.swift
```diff
@@ -31,17 +34,37 @@ extension Child {
...
-      let expr = "\(varName)\(type.optionalQuestionMark)"
-      return "\(expr).build\(type.baseName)(format: \(format), leadingTrivia: nil)"
+      var format: ExpressibleAsExprBuildable = formatName
+      if isIndented {
+        format = FunctionCallExpr(MemberAccessExpr(base: format, name: "_indented"))
+      }
+      let expr = type.optionalChained(expr: varName)
+      return FunctionCallExpr(MemberAccessExpr(base: expr, name: "build\(type.baseName)")) {
+        TupleExprElement(label: "format", expression: format)
+        TupleExprElement(label: "leadingTrivia", expression: "nil")
```
ahoppen, line 66: This produces a `nil` identifier instead of a `nil` keyword. It doesn’t really matter in terms of generated source code but it would still be good to use the `nil` keyword here IMO.

ahoppen (approved):
Two minor comment, otherwise LGTM.

Also, I think that `optionalWrapped` and `optionalChained` are good examples of how SwiftSyntaxBuilder simplifies source generation. 👍

Sources/SwiftSyntaxBuilderGeneration/SyntaxBuildableChild.swift
fwcd: Should we do this for all output types, i.e. also all of the properties? (I have no strong preference either way)

Sources/SwiftSyntaxBuilderGeneration/SyntaxBuildableChild.swift
fwcd: Ah, good catch, thanks! For some reason I thought `nil` was parsed as an identifier, like `self`.

fwcd:
@swift-ci please test

PR #546, opened 2022-08-04
Fix documentation for syntax collections in `SwiftSyntaxBuilderGeneration`

fwcd:
The Swift translation of`Node.documentation` (from [the Python version](https://github.com/apple/swift-syntax/blob/8875897090059060c6ea7846261187380b1f81dc/Sources/SwiftSyntaxBuilder/gyb_helpers/SyntaxBuildableWrappers.py#L102-L110)) generates empty documentations for syntax collections without a `documentation`, which previously generated the description of its element type as a fallback. This small patch fixes this slight mistranslation.


fwcd:
@swift-ci please test

ahoppen (approved):

PR #548, opened 2022-08-04
Fix `ArrayExpr` in `SyntaxBuildableChild`

fwcd:
This is a small patch that fixes the (currently broken) `SwiftSyntaxBuilderGeneration` due to the now missing `Array: ExpressibleAsArrayElementList` conformance (see https://github.com/apple/swift-syntax/pull/483#discussion_r937747605).


ahoppen (approved):
Thanks. I just started looking into the same issue.

PR #550, opened 2022-08-04
Mark `TokenKind` as `Hashable`

ahoppen:
Useful so we can create a `Set<TokenKind>`, which clients can use to check if a token has one of a set of expected kinds.


ahoppen:
@swift-ci Please test

fwcd (approved):

PR #551, opened 2022-08-04
Use the builder-based `ParameterClause` initializer where possible

fwcd:
This looks both prettier and no longer requires us to worry about trailing commas thanks to #454.


fwcd:
@swift-ci please test

ahoppen (approved):

PR #552, opened 2022-08-04
Use `NilLiteralExpr` in `BuildableCollectionNodes`

fwcd:
A small patch that uses `NilLiteralExpr` in`BuildableCollectionNodes` instead of generating `nil` as an identifier.

One minor thing that we might want to figure out is why `nil` tokens generate a trailing space unlike identifier tokens.

### DIFF
--- a/Sources/SwiftSyntaxBuilderGeneration/Templates/BuildableCollectionNodesFile.swift
+++ b/Sources/SwiftSyntaxBuilderGeneration/Templates/BuildableCollectionNodesFile.swift
@@ -200,7 +200,7 @@ private func createBuildFunction(node: Node) -> FunctionDecl {
                         BinaryOperatorExpr("+")
                         FunctionCallExpr(MemberAccessExpr(base: "format", name: "_makeIndent"))
                       }
-                    : "nil"
+                    : NilLiteralExpr()
                 )
               }
             })


fwcd:
@swift-ci please test

ahoppen:
> One minor thing that we might want to figure out is why nil tokens generate a trailing space unlike identifier tokens.

AFAICT that’s because keywords require a trailing space by default. We might want to revisit that for literals.

https://github.com/apple/swift/blob/df2d938604ca19427aeed52d741674013403e645/utils/gyb_syntax_support/Token.py#L44

Sources/SwiftSyntaxBuilderGeneration/Templates/BuildableCollectionNodesFile.swift
```diff
@@ -200,7 +200,7 @@ private func createBuildFunction(node: Node) -> FunctionDecl {
                         BinaryOperatorExpr("+")
                         FunctionCallExpr(MemberAccessExpr(base: "format", name: "_makeIndent"))
                       }
-                    : "nil"
+                    : NilLiteralExpr()
```
evnik, line 203: Should it be just a keyword? 
```suggestion
                    : SyntaxFactory.makeNilKeyword(trailingTrivia: [])
```

I'd like to get rid of default trailing space after https://github.com/apple/swift/pull/60204 is merged

fwcd: Not sure, a token syntax can be converted [to identifiers and binary operators](https://github.com/apple/swift-syntax/blob/2380374d1ed52d89674adad21d98b8dd1d1be14c/Sources/SwiftSyntaxBuilder/gyb_helpers/ExpressibleAsConformances.py#L41), neither of which seem to be correct in this case.

Since the expression is a nil literal, `NilLiteralExpr` should be the right way to spell it in `SwiftSyntaxBuilder` too IMO

fwcd:
@swift-ci please test

ahoppen (approved):

Issue #549, opened 2022-08-04
Printed string literal expressions are not escaped correctly


fwcd:
Printed string literal expressions are not escaped correctly (with backslashes) when generated programmatically. I stumbled upon this while beginning to port `BuildableNodes.swift.gyb` to `SwiftSyntaxBuilderGeneration`, since the template generates such literals.

Consider the following example:

```swift
import SwiftSyntaxBuilder

let source = SourceFile {
  StringLiteralExpr("\\")
}

let syntax = source.buildSyntax(format: Format())
print(syntax.description)
```

### Checklist

- [x] If possible, I've reproduced the issue using the `main` branch of this package.
- [x] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/apple/swift-syntax/issues).

### Expected behavior

The example should print `"\\"`.

### Actual behavior

The example prints `"\"`, leaving the string literal unterminated.

### Steps to Reproduce

Run the program above.

ahoppen:
rdar://98129594

fwcd:
Closing as being the intended behavior as per our discussion.

PR #555, opened 2022-08-04
Generate `BuildableNodes` with `SwiftSyntaxBuilderGeneration`

fwcd:
This PR ports the currently gyb-generated `BuildableNodes` to `SwiftSyntaxBuilder`'s DSL as part of the ongoing effort to bootstrap `SwiftSyntaxBuilder` with `SwiftSyntaxBuilderGeneration`.

Currently blocked on

- #571 
- #574 
- #575 
- #576 

To do:

- [x] Generate `BuildableNodes` using `SwiftSyntaxBuilderGeneration`
- [x] Remove the current gyb-generation (and its associated exclude from the `Package.swift`)


Sources/SwiftSyntaxBuilderGeneration/SyntaxUtilities.swift
```diff
@@ -82,3 +82,30 @@ func createFormatLeadingTriviaParameters(withDefaultTrivia: Bool = false) -> Par
...
+
+/// Generate the `create...` function for an `ExpressibleAs...` conformance
+/// that includes an explanation as to how the function disambiguates a conformance.
+func createDisambiguatingExpressibleAsCreateFunction(type: SyntaxBuildableType, baseType: SyntaxBuildableType) -> FunctionDecl {
+  createExpressibleAsCreateFunction(type: baseType, additionalDocComments: [
+    "/// `\(type.buildableBaseName)` may conform to `\(baseType.expressibleAsBaseName)` via different `ExpressibleAs*` paths.",
+    "/// Thus, there are multiple default implementations of `create\(baseType.buildableBaseName)`, some of which perform conversions",
+    "/// through `ExpressibleAs*` protocols. To resolve the ambiguity, provie a fixed implementation that doesn't perform any conversions.",
+  ]) 
+}
```
fwcd, line 111: This PR factors out the `create...` functions generated for the `ExpressibleAs...` conformances. This is particularly nice since we can share a lot of logic between `BuildableNodes` and `BuildableCollectionNodes`, including this lengthy doc comment that explains how the function disambiguates the implementation.

Sources/SwiftSyntaxBuilderGeneration/Templates/BuildableNodesFile.swift
```diff
@@ -0,0 +1,371 @@
...
+    for child in children {
+      SequenceExpr {
+        MemberAccessExpr(base: "self", name: child.swiftName)
+        AssignmentExpr()
+        child.type.generateExprConvertParamTypeToStorageType(varName: child.swiftName)
+      }
+      // FIXME: This is currently broken due to https://github.com/apple/swift-syntax/issues/549
+      // if let assertStmt = child.generateAssertStmtTextChoices(varName: child.swiftName) {
+      //   assertStmt
+      // }
```
fwcd, line 121: As mentioned above, we need to uncomment this once #549 is fixed.

fwcd (commented):
Two notes on the implementation below. Another thing that we might want to look into is producing nicer formatting for very long parameter lists, for example by inserting `.newline` trivias:

```swift
func someFunction(
  first: First,
  second: Second,
  third: Third
)
```

instead of

```swift
func someFunction(first: First, second: Second, third: Third)
```

This would, if handled generically, be out-of-scope for this PR, but ideas or thoughts on this are of course welcome.

fwcd:
@swift-ci please test

Sources/generate-swift-syntax-builder/Templates/BuildableNodesFile.swift
```diff
@@ -0,0 +1,372 @@
...
+    // Generate node struct
+    StructDecl(
+      leadingTrivia: node.documentation.isEmpty
+        ? []
+        : .docLineComment("/// \(node.documentation)") + .newline,
+      modifiers: [TokenSyntax.public],
+      identifier: type.buildableBaseName,
+      inheritanceClause: createTypeInheritanceClause(conformances: conformances)
+    ) {
+      let children = node.children
```
ahoppen, line 38: Is there any specific reason why we need the `children` variable and can’t just use `node.children` below?

Sources/generate-swift-syntax-builder/Templates/BuildableNodesFile.swift
```diff
@@ -0,0 +1,372 @@
...
+
+      VariableDecl(
+        leadingTrivia: [
+          "/// The leading trivia attached to this syntax node once built.",
+          "/// This is typically used to add comments (e.g. for documentation).",
+        ].map { .docLineComment($0) + .newline }.reduce([], +),
+        .let,
+        name: "leadingTrivia",
+        type: "Trivia"
+      )
```
ahoppen, line 53: Nitpick: Since the leading trivia is displayed before the first node, what do you think about moving it in front of the children variables?

Sources/generate-swift-syntax-builder/Templates/BuildableNodesFile.swift
```diff
@@ -0,0 +1,372 @@
...
+      colon: .colon,
+      type: OptionalType(wrappedType: "Trivia"),
+      defaultArgument: NilLiteralExpr()
+    )
+  }
+}
+
+/// Generate the `withTrailingComma` function.
+private func createWithTrailingCommaFunction(node: Node) -> FunctionDecl {
+  let children = node.children
```
ahoppen, line 338: Again, why do we need the `children` variable?

ahoppen (approved):
Some minor stylistic comments inline. Otherwise LGTM.

fwcd:
@swift-ci please test

PR #557, opened 2022-08-04
Use SWIFTCI_USE_LOCAL_DEPS to check whether to clone swift-docc-plugin

ahoppen:
Only SwiftSyntax’s build-script sets `SWIFT_BUILD_SCRIPT_ENVIRONMENT`. Once we migrate other projects into the unified SwiftPM build, the environment variable is not set and SwiftPM tries to download swift-docc-plugin in CI.

Use `SWIFTCI_USE_LOCAL_DEPS` as an indicator whether the remote plugin should be loaded.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -143,6 +143,9 @@ let package = Package(
   ]
 )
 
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

kimdv (approved):

PR #558, opened 2022-08-04
Add tests for TypeInitializerClauseTests

kimdv:
Test I initially wrote for something in #479 

Thought we could use that


ahoppen (approved):

kimdv:
@swift-ci please test

PR #538, opened 2022-08-02
Add methods for structurally comparing syntax trees

bnbarham:
Adds a `findFirstDifference` API to `Syntax`, `XCTAssertSameStructure`
assertions, and a `SubtreeMatch` to compare subtrees.

PR #559, opened 2022-08-05
Get rid of shorthand `if let` syntax to fix compilation with Swift 5.6

evnik:
Similar issue as [here](https://github.com/apple/swift-syntax/pull/538#discussion_r935207895)
There is also usage of the new APIs (like [`UnsafeMutableRawPointer.alignedUp(toMultipleOf:)`](https://developer.apple.com/documentation/swift/unsafemutablerawpointer/alignedup(tomultipleof:)?changes=latest_minor)) in the [`BumpPtrAllocator.swift`](https://github.com/apple/swift-syntax/blob/2380374d1ed52d89674adad21d98b8dd1d1be14c/Sources/SwiftSyntax/BumpPtrAllocator.swift#L83), but I'm not sure how to fix that properly

### DIFF
--- a/Sources/SwiftSyntax/Syntax.swift
+++ b/Sources/SwiftSyntax/Syntax.swift
@@ -529,10 +529,10 @@ public extension SyntaxProtocol {
     if let token = Syntax(self).as(TokenSyntax.self) {
       target.write(String(describing: token.tokenKind))
       if includeTrivia {
-        if let leadingTrivia, !leadingTrivia.isEmpty {
+        if let leadingTrivia = leadingTrivia, !leadingTrivia.isEmpty {
           target.write(" leadingTrivia=\(leadingTrivia.debugDescription)")
         }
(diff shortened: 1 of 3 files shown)


CodaFi:
Thanks!

@swift-ci test

CodaFi (approved):

CodaFi:
For the alignment problem, we'll need to recreate this primitive.

```
#if swift(<5.7)
extension UnsafeMutableRawPointer {
  fileprivate func alignedUp(toMultipleOf alignment: Int) -> Self {
    let mask = UInt(alignment) &- 1
    assert(
      alignment > 0 && UInt(alignment) & mask == 0,
      "alignment must be a whole power of 2."
    )
    let bits = (UInt(bitPattern: self) &+ mask) & ~mask
    return .init(bitPattern: bits)!
  }
}
#endif
```

PR #560, opened 2022-08-05
Restore the Ability to Build with 5.6/5.5

CodaFi:
We still support building with tools from Xcode 13, so we cannot adopt
features after Swift 5.5. Recreate UMRP.alignedUp(toMultipleOf:) for
this environment and conditionalize docc support in the build script
on the newer language features being around.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.6
+// swift-tools-version:5.5
 
 import PackageDescription
 import Foundation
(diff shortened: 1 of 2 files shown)


CodaFi:
@swift-ci test

Package.swift
```diff
@@ -142,9 +142,11 @@ let package = Package(
   ]
 )
 
-// If `SWIFTCI_USE_LOCAL_DEPS` is set, we are in a CI enviornment that might disallow 
+#if swift(>=5.6)
```
CodaFi, line 145: I don't understand why we have this dependency. We do not need swift-docc-plugin to use docc, just the 5.5 package manifest or later. I would like to remove it.

ahoppen: I thought that we needed it to build/preview the documentation using the following command. But maybe I’m missing something.
```
swift package --disable-sandbox preview-documentation --target SwiftSyntax
```

CC @kimdv

ahoppen (approved):
Thanks

PR #565, opened 2022-08-09
Add an action to verify that round-tripping a source file works

ahoppen:
This is useful to reduce round-tripping failures.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #566, opened 2022-08-09
Allow starting a SourceLocationConverter on a node thats's not Source FileSyntax

ahoppen:
For test cases, it’s useful to work on syntax subtrees that don’t have a `SourceFileSyntax` as their root. In these cases it’s also useful to start a `SourceLocationConverter` on any syntax node.


ahoppen:
@swift-ci Please test

PR #567, opened 2022-08-09
Remove underscore from SyntaxViewingMode._all

ahoppen:
I believe this syntax viewing mode can be useful in general and there’s no need to make it semi-private with an underscore.

rdar://98368959

### DIFF
--- a/Sources/SwiftSyntax/SyntaxTreeViewMode.swift
+++ b/Sources/SwiftSyntax/SyntaxTreeViewMode.swift
@@ -25,9 +25,7 @@ public enum SyntaxTreeViewMode {
   case fixedUp
 
   /// Both missing and garbage nodes will be traversed.
-  /// Mostly useful for SwiftSyntax internal verifiers that verify the integrity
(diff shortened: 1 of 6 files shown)


ahoppen:
@swift-ci Please test

bnbarham (approved):

Sources/SwiftSyntax/SyntaxTreeViewMode.swift
```diff
@@ -27,7 +27,7 @@ public enum SyntaxTreeViewMode {
   /// Both missing and garbage nodes will be traversed.
   /// Mostly useful for SwiftSyntax internal verifiers that verify the integrity
   /// of the entire tree
```
bnbarham, line 29: Comment still true?

ahoppen: Good catch. I think we should get rid of it.

ahoppen:
@swift-ci Please test

PR #571, opened 2022-08-11
Port garbage node generation to `SwiftSyntaxBuilderGeneration`'s `Node`

fwcd:
This ports the [new garbage node generation](https://github.com/apple/swift/blob/b0fb7c54b2c5159be5f23daeec277f0deded9be3/utils/gyb_syntax_support/Node.py#L27-L47) from `gyb_syntax_support` to the Swift equivalent in `SwiftSyntaxBuilderGeneration` in preparation for #555.

### DIFF
--- a/Sources/SwiftSyntaxBuilderGeneration/Node.swift
+++ b/Sources/SwiftSyntaxBuilderGeneration/Node.swift
@@ -22,6 +22,7 @@ class Node {
   let baseKind: String
   let traits: [String]
   let children: [Child]
+  let nonGarbageChildren: [Child]
   let collectionElementName: String?
   let collectionElementChoices: [String]?
   let omitWhenEmpty: Bool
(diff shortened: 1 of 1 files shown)


fwcd:
@swift-ci please test

Sources/SwiftSyntaxBuilderGeneration/Node.swift
```diff
@@ -84,9 +85,30 @@ class Node {
...
+      // any two defined children
+      self.children = children.enumerated().flatMap { (i, child) in
+        [
+          Child(
+            name: i == 0
+              ? "GarbageBefore\(child.name)"
+              : "GarbageBetween\(children[i - 1].name)And\(children[i].name)",
+            kind: "GarbageNodes",
+            isOptional: true,
+            collectionElementName: name
```
ahoppen, line 103: `collectionElementName` should be the same as the parameter passed to `name` but right now it’s the node’s name.

fwcd: Good catch, thanks!

ahoppen (approved):

PR #572, opened 2022-08-11
Add typed raw syntax nodes

rintaro:
* `RawXXXSyntax` is a typed `RawSyntax` providing `SyntaxKind` specific API including a factory initializer.
* Expose `RawSyntax`, `RawXXXSyntax`, and other related types as SPI (`@_spi(RawSyntax)`)
* Move layout validation to `RawSyntax` and perform it whenever the layout is created. (debug build only). Previously, layout validation was only performed In `SyntaxRewriter`


rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

Sources/SwiftSyntax/TokenKind.swift.gyb
```diff
@@ -166,6 +166,28 @@ public enum RawTokenKind {
 %   if token.text:
     case .${token.swift_kind()}: return "${token.text}"
 %   end
+% end
+    default: return nil
+    }
+  }
+
+  @_spi(RawSyntax)
+  public init?(keyword text: SyntaxText) {
```
ahoppen, line 175: What is this used for?

CC @fwcd We could maybe use this to make `TokenSyntax` conform to `ExpressibleByStringLiteral`.

ahoppen (commented):
LGTM. Some minor comments inline.

Sources/SwiftSyntax/RawSyntaxNodeProtocol.swift
```diff
@@ -0,0 +1,124 @@
...
+  /// Creates a `TokenSyntax`. `text` and trivia must be managed by the same
+  /// `arena`.
+  public init(
+    kind: RawTokenKind,
+    text: SyntaxText,
+    leadingTriviaPieces: [RawTriviaPiece],
+    trailingTriviaPieces: [RawTriviaPiece],
+    arena: SyntaxArena
+  ) {
+    let totalTriviaCount = leadingTriviaPieces.count + trailingTriviaPieces.count
```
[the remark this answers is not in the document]
rintaro: We can. However, they can actually be a static text like:
```
RawTokenSyntax(
  tokenKind: .structKeyword, tokenText: "struct",
  leadingTriviaPiecies: [.docLineComment("/// Foo"), .newlines(1)], trailingTriviaPieces: [],
  arena: arena)
```
This is safe because static text are always alive. Do you think this feature is not needed?

Sources/SwiftSyntax/TokenKind.swift.gyb
rintaro: This is currently not used not necessary for this PR. Should I make a separate PR?

Sources/SwiftSyntax/RawSyntaxNodeProtocol.swift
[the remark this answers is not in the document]
ahoppen: This was basically referring to your comment in line 90: “`text` and trivia must be managed by the same `arena`.” So I thought whether we could enforce that requirement using assertions.

Sources/SwiftSyntax/TokenKind.swift.gyb
ahoppen: Yes, let’s move this to the PR that actually starts using it.

rintaro:
@swift-ci Please test

Sources/SwiftSyntax/RawSyntaxNodeProtocol.swift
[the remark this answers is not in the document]
rintaro: Added the assertions.

Sources/SwiftSyntax/TokenKind.swift.gyb
rintaro: Removed

PR #575, opened 2022-08-11
Improve assert statement generation and fix token choices

fwcd:
In preparation for #555, this patch migrates `generateAssertStmtTextChoices` to return an expr node instead of a string (as a small follow-up to #542). Additionally, it adds a utility function for escaping string literals (to address #549) and a function for generating force-unwraps based on a type's optionality.


Sources/SwiftSyntaxBuilderGeneration/Child.swift
```diff
@@ -95,14 +96,18 @@ class Child {
...
+    var mappedTokenChoices = [Token?]()
     
-    if syntaxKind.hasSuffix("Token"), let token = SYNTAX_TOKEN_MAP[syntaxKind] {
+    if isToken, let token = SYNTAX_TOKEN_MAP[syntaxKind] {
       mappedTokenChoices.append(token)
     }
 
-    mappedTokenChoices.append(contentsOf: tokenChoices.compactMap { SYNTAX_TOKEN_MAP[$0] })
-    self.tokenChoices = mappedTokenChoices
+    mappedTokenChoices.append(contentsOf: tokenChoices.map { SYNTAX_TOKEN_MAP["\($0)Token"] })
```
fwcd, line 106: This fixes a bug where `self.tokenChoices` was always empty since the `tokenChoices` do not have the `Token` suffix.

fwcd:
@swift-ci please test

ahoppen (approved):

fwcd:
@swift-ci please test

PR #576, opened 2022-08-11
Add utility property for enclosing a string in backticks

fwcd:
A small convenience property in preparation of #555.

### DIFF
--- a/Sources/SwiftSyntaxBuilderGeneration/String+Extensions.swift
+++ b/Sources/SwiftSyntaxBuilderGeneration/String+Extensions.swift
@@ -14,4 +14,5 @@ import Foundation
 
 extension StringProtocol {
     var withFirstCharacterLowercased: String { prefix(1).lowercased() + dropFirst() }
+    var backticked: String { "`\(self)`" }
 }
--- a/Sources/SwiftSyntaxBuilderGeneration/Templates/TokensFile.swift
+++ b/Sources/SwiftSyntaxBuilderGeneration/Templates/TokensFile.swift
@@ -35,10 +35,10 @@ let tokensFile = SourceFile {
           // We need to use `CodeBlock` here to ensure there is braces around.
 
           let accessor = CodeBlock {
-            FunctionCallExpr(MemberAccessExpr(base: "TokenSyntax", name: "\(token.swiftKind)"))
+            FunctionCallExpr(MemberAccessExpr(base: "TokenSyntax", name: token.swiftKind))
           }
 
-          createTokenSyntaxPatternBinding("`\(token.name.withFirstCharacterLowercased)`", accessor: accessor)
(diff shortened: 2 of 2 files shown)


fwcd:
@swift-ci please test

fwcd:
@swift-ci please test

Sources/SwiftSyntaxBuilderGeneration/String+Extensions.swift
```diff
@@ -14,4 +14,5 @@ import Foundation
 
 extension StringProtocol {
     var withFirstCharacterLowercased: String { prefix(1).lowercased() + dropFirst() }
+    var backticked: String { "`\(replacingOccurrences(of: "`", with: ""))`" }
```
ahoppen, line 17: Why do we need `replacingOccurrences(of: "`", with: "")`? Is `token.name` sometimes already backticked?

fwcd: I don't think so, my thinking was to make `.backticked` idempotent. But yeah, it would probably be simpler to just remove it.

ahoppen: I think just surrounding the text with `` ` `` is the most intuitive behavior, in particular if we don’t need idempotency.

fwcd:
@swift-ci please test

Sources/SwiftSyntaxBuilderGeneration/Templates/TokensFile.swift
```diff
@@ -51,7 +51,30 @@ let tokensFile = SourceFile {
...
+          },
+          output: "TokenSyntax"
+        )
+
+        FunctionDecl(
+          modifiers: [TokenSyntax.static],
+          identifier: .identifier(token.name.withFirstCharacterLowercased.backticked),
+          signature: signature
+        ) {
+          FunctionCallExpr(MemberAccessExpr(base: "SyntaxFactory", name: "make\(token.name)")) {
```
ahoppen, line 75: Could you migrate this to not use `SyntaxFactory`, which I recently deprecated. I think this should be something like
```suggestion
          FunctionCallExpr(MemberAccessExpr(base: "TokenSyntax", name: token.name)) {
```

ahoppen (commented):
Did you forget to commit the generated file?

Sources/SwiftSyntaxBuilderGeneration/Templates/TokensFile.swift
fwcd: Ah whoops, looks like I accidentally picked the wrong side when resolving the conflict. The generated `Tokens.swift` shouldn't have changed at all.

fwcd:
@swift-ci please test

ahoppen (approved):

PR #579, opened 2022-08-11
Add convenience initializer for raw string literals

fwcd:
This patch adds a small convenience initializer for raw string literals (in preparation of #575) and extends the existing convenience initializer to take custom delimiters.

```swift
StringLiteralExpr(raw: "abc")

// now generates

#"abc"#
```

### DIFF
--- a/Sources/SwiftSyntaxBuilder/StringLiteralExprConvenienceInitializers.swift
+++ b/Sources/SwiftSyntaxBuilder/StringLiteralExprConvenienceInitializers.swift
@@ -11,15 +11,58 @@
 //===----------------------------------------------------------------------===//
 
 import SwiftSyntax
+import Foundation
+
+/// A regular expression matching sequences of `#`s with an adjacent quote or
+/// interpolation. Used to determine the number of `#`s for a raw string literal.
+private let rawStringPotentialEscapesPattern = try! NSRegularExpression(
+  pattern: [
+    #""(#+)"#, // Match potential opening delimiters
+    #"(#+)""#, // Match potential closing delimiters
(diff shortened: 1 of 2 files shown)


fwcd:
@swift-ci please test

Sources/SwiftSyntaxBuilder/StringLiteralExprConvenienceInitializers.swift
```diff
@@ -13,13 +13,33 @@
...
+      openDelimiter: openDelimiter,
+      openQuote: openQuote,
+      segments: segments,
+      closeQuote: closeQuote,
+      closeDelimiter: closeDelimiter
+    )
+  }
+
+  /// Creates a raw string literal.
+  public init(raw value: String) {
```
ahoppen, line 38: What do you think about adding an additional parameter that describes how many `#` should be added and defaulting it to 1? That way you could create a string literal like `##" The raw string delimiter is #". "##`

fwcd: Sounds good. Perhaps we could take it even further and scan the string for the longest sequence of consecutive hashes + adjacent quote and do that automatically. Or would that be too much magic?

ahoppen: Sure, you might also want to check for the longest sequence of `####\(` which can also be escaped using raw strings.

fwcd:
@swift-ci please test

ahoppen (approved):

Issue #577, opened 2022-08-11
SwiftSyntaxBuilderGeneration Cannot Depend on SwiftSyntaxBuilder


CodaFi:
This creates a chicken-and-egg problem when any aspect of the syntax nodes changes that also affects the builders. Consider the addition of a hypothetical node `FooSyntax`. When a user goes to update their checkouts and reruns the code generation script, they are greeted with a build failure. At first blush, this is because `ExpressibleByFooSyntax` is not regenerated by `build-script.py generate-source-code`. So the next thought would then be - just `swift run SwiftSyntaxBuilderGeneration` and regenerate the builder. But! that module no longer builds thanks to the prior source code generation script invocation.

Okay, so start over. Regenerate the builders, then regenerate syntax. Also doesn't work - the builder depends on the updated syntax nodes being present. But in order to do that... you have to run `build-script.py generate-source-code`.

Going forward, there cannot be a dependency between these two modules. We must be able to bootstrap SwiftSyntax when things change here.

fwcd:
This is a valid point and a problem inherent to bootstrapping: If we want to use `SwiftSyntaxBuilder` to generate part of itself, it has to depend on itself in some form. However, I see that this could become an issue when trying to extend the syntax.

One solution I could think of would be to pin down a version of `SwiftSyntaxBuilder` (e.g. to the last Swift release?) That would prevent to from using any newer language features in the generated files, but I don't think that would be a big issue. Since I don't think a package can depend on an older version of itself, this might involve moving `SwiftSyntaxBuilderGeneration` to another repo, which might require some changes to the build infrastructure.

If everything else fails, we could of course also go back to the gyb-generation we had earlier, although it would be nice if we could find a solution that doesn't involve discarding all of the work done on porting the templates to Swift.

cc @ahoppen @kimdv

ahoppen:
rdar://98765046

ahoppen:
The underlying issue is solved by https://github.com/apple/swift-syntax/pull/583, which also has some more explanation.

CodaFi:
Reopening this issue because it has not been resolved. As long as this tool is defined in terms of itself, it's possible for changes that are made to mainline and immediately consumed to harm the builds of people that do not have their generated copy of the code match ToT. 

We ought to remove the `--gyb-only` option since that's just a way to skip the first stage of bootstrapping and will basically never do the right thing. I still think we ought to consider splitting this tool out of the main repository.

CodaFi:
[...]

PR #580, opened 2022-08-11
Move `SwiftSyntaxBuilderGeneration` to separate package that pins a version of `swift-syntax`

fwcd:
This PR addresses #577 by moving `SwiftSyntaxBuilderGeneration` to a separate package, i.e. replaces the current package structure

```
swift-syntax
└─ Sources
   ├─ SwiftSyntaxBuilder
   ├─ SwiftSyntaxBuilderGeneration
   └─ ...
```

with

```
swift-syntax
├─ Sources
│  ├─ SwiftSyntaxBuilder
│  └─ ...
└─ SwiftSyntaxGeneration
   └─ Sources
      └─ swift-syntax-builder-generation
```

The subpackage pins a specific commit of `swift-syntax` as a dependency and thereby is no longer susceptible to the bootstrapping problems outlined in #577 when adding new syntax nodes.

Things to figure out:
- [x] The build script, in particular how it has to be updated
- [ ] How this plays with CI and its workspaces

Therefore I will mark this PR as a draft for now, so we can discuss these things here (and whether it's a good approach to take at all).

cc @ahoppen @CodaFi


SwiftSyntaxGeneration/Package.swift
```diff
@@ -0,0 +1,41 @@
+// swift-tools-version:5.6
+
+import PackageDescription
+
+let package = Package(
+  name: "SwiftSyntaxGeneration",
+  products: [
+    .executable(name: "swift-syntax-builder-generation", targets: ["swift-syntax-builder-generation"]),
```
fwcd, line 8: Minor note: The reason for renaming `SwiftSyntaxBuilderGeneration` to `swift-syntax-builder-generation` (kebab-case) is twofold:

1. Until the pinned revision of `swift-syntax` no longer has a target named `SwiftSyntaxBuilderGeneration` the name will conflict with this target
2. Most Swift repos follow the convention of naming executable targets in kebab-case (e.g. this repo's `lit-test-helper`, SourceKit-LSP's `sourcekit-lsp`, SwiftFormat's `swift-format`, ...). Therefore this would be good opportunity to update the name.

ahoppen:
I don’t think this is necessary. We just need to make code generation a three-step instead of a two-step process, which I’m doing in https://github.com/apple/swift-syntax/pull/583.

ahoppen:
I think renaming `SwiftSyntaxBuilderGeneration` to `swift-syntax-builder-generation` would still be a change that’s worthwhile.

ahoppen:
The underlying build issue has been fixed by https://github.com/apple/swift-syntax/pull/583

PR #583, opened 2022-08-12
Implement code generation as a three-step process

### LINKED ISSUE #577: SwiftSyntaxBuilderGeneration Cannot Depend on SwiftSyntaxBuilder

ahoppen:
To make sure code generation works when adding new nodes, implement it as a three-stage process.

1. Re-generate the gyb files in SwiftSyntaxBuilderGeneration. This only re-generates the Swift files which essentially match the Python files in gyb_syntax_support. This does not generate any new syntax nodes etc.
2. Run code generation using SwiftSyntaxBuilder (i.e. run SwiftSyntaxBuilderGeneration). This still compiles because stage 1 did not modify any types in SwiftSyntax
3. Generate the remaining gyb files to match the files that were generated using SwiftSyntaxBuilderGeneration in step 2. Since this is Python, it will always pass.

Resolves #577

### DIFF
--- a/build-script.py
+++ b/build-script.py
@@ -284,20 +284,28 @@ def generate_syntax_node_template_gyb_files(
             verbose=verbose,
         )
 
-
+# Gyb-files that should be generated before running code generation using SwiftSyntaxBuilder.
 # Maps directories containing .gyb files to the directories the generated files should
 # live in.
-def gyb_dir_mapping(
+def first_stage_gyb_dir_mapping(
+    generateswiftsyntaxbuilder_destination: Optional[str] = None,
+) -> Dict[str, Optional[str]]:
+    return {
+        GENERATESWIFTSYNTAXBUILDER_DIR: generateswiftsyntaxbuilder_destination,
+    }
+
+# Gyb-files that should be generated after running code generation using SwiftSyntaxBuilder.
+# Maps directories containing .gyb files to the directories the generated files should
+# live in.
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

CodaFi (approved):
Approach LGTM. I was able to successfully use https://github.com/apple/swift/pull/60629 to test this.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

PR #586, opened 2022-08-12
Use explicit return type in Node initializer's flatMap

fwcd:
This fixes a build issue with the latest Swift release since the improved multi-statement closure type inference is only available on recent nightly toolchains.


fwcd:
@swift-ci please test

ahoppen (approved):

PR #587, opened 2022-08-12
Generate trailing newlines in `SwiftSyntaxBuilderGeneration`

fwcd:
This is a small patch that generates trailing newlines in files generated by `SwiftSyntaxBuilderGeneration`, both to adhere to the Unix convention and to fix an issue with editors that automatically insert trailing newlines and thus cause these files to change when opening them.


fwcd:
@swift-ci please test

kimdv (approved):

ahoppen (approved):

PR #589, opened 2022-08-13
Fix RawTokenSyntax initializer

rintaro:
Token text was not taken into account in the byteLength calcuration.


rintaro:
@swift-ci Please test

CodaFi (approved):

PR #595, opened 2022-08-16
SyntaxText: improve SyntaxText.== function

rintaro:
Since comparing the same memory region is rare in SwiftSyntax, drop the same-address shortcut to save the execution branching.

(For those who are interested [this](https://godbolt.org/z/5aqYo4G5r) is how `RawTokenKind.init(keyword: SyntaxString)` is compiled with `-O`)


ahoppen (approved):

rintaro:
@swift-ci Please test

PR #596, opened 2022-08-16
Add a recursive `hasError` flag to `RawSyntax`

ahoppen:
This allows us to check whether a subtree contains a syntax error in O(1).

rdar://98575249

### DIFF
--- a/Sources/SwiftSyntax/RawSyntax.swift
+++ b/Sources/SwiftSyntax/RawSyntax.swift
@@ -13,6 +13,20 @@
 typealias RawSyntaxBuffer = UnsafeBufferPointer<RawSyntax?>
 typealias RawTriviaPieceBuffer = UnsafeBufferPointer<RawTriviaPiece>
 
+fileprivate extension SyntaxKind {
+  /// Whether this node kind should be considered as `hasError` for purposes of `RecursiveRawSyntaxFlags`.
+  var hasError: Bool {
+    return self == .unexpectedNodes || self.isMissing
+  }
+}
+
+struct RecursiveRawSyntaxFlags: OptionSet {
+  let rawValue: UInt8
+
+  /// Whether the tree contained by this layout has any missing or unexpected nodes.
+  static let hasError = RecursiveRawSyntaxFlags(rawValue: 1 << 0)
+}
+
 /// Node data for RawSyntax tree. Tagged union plus common data.
 internal struct RawSyntaxData {
   internal enum Payload {
@@ -54,6 +68,7 @@ internal struct RawSyntaxData {
     var byteLength: Int
     /// Number of nodes in this subtree, excluding this node.
     var descendantCount: Int
+    var recursiveFlags: RecursiveRawSyntaxFlags
   }
 
   fileprivate var payload: Payload
@@ -213,6 +228,19 @@ extension RawSyntax {
   /// Whether or not this node is an unknown one.
   var isUnknown: Bool {
     kind.isUnknown
+  }
+
+  var recursiveFlags: RecursiveRawSyntaxFlags {
+    switch rawData.payload {
+    case .materializedToken, .parsedToken:
+      var recursiveFlags: RecursiveRawSyntaxFlags = []
+      if presence == .missing {
+        recursiveFlags.insert(.hasError)
+      }
+      return recursiveFlags
+    case .layout(let dat):
+      return dat.recursiveFlags
+    }
   }
 
   /// Child nodes.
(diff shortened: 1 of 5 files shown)


Sources/SwiftSyntax/RawSyntax.swift
```diff
@@ -172,6 +174,13 @@ extension RawSyntax {
     kind.isUnknown
   }
 
+  var hasError: Bool {
+    switch rawData.payload {
+    case .materializedToken(_): return self.presence == .missing
```
ahoppen, line 179: I would prefer to detect the presence of the node purely based on the `materializedToken` but I’m worried that will break if we change whether tokens without source text but with trivia are considered missing.

Sources/SwiftSyntax/RawSyntax.swift
```diff
@@ -36,6 +36,8 @@ internal struct RawSyntaxData {
     var byteLength: Int
     /// Number of nodes in this subtree, excluding this node.
     var descendantCount: Int
+    /// Whether the tree contained by this layout has any missing or unexpected nodes.
+    var hasError: Bool
```
rintaro, line 40: Could we model this a `OptionSet` backed by a e.g.`UInt8`? I feel we will need other cached flags in the future.

ahoppen: Good idea. I thought we could do this once we add more flags but it’s a good idea to do it now.

Sources/SwiftSyntax/RawSyntax.swift
```diff
@@ -736,9 +751,20 @@ extension RawSyntax {
     arena: SyntaxArena
   ) -> RawSyntax {
     validateLayout(layout: layout, as: kind)
+    var recursiveFlags = RecursiveRawSyntaxFlags()
```
rintaro, line 754: I consider this is a memberwise initializer.
Instead of calculating this in this function, please add `recursiveFlags: RecursiveRawSyntaxFlags` parameter to this, and edit  `makeLayout(arena:kind:uninitializedCount:initializingWith:)` and  `makeEmptyLayout(arena:kind:)`

ahoppen: 👍 Done

ahoppen:
@swift-ci Please test

Sources/SwiftSyntax/RawSyntax.swift
```diff
@@ -767,18 +790,36 @@ extension RawSyntax {
       descendantCount += node.totalNodes
       arena.addChild(node.arenaReference)
     }
+    var recursiveFlags = RecursiveRawSyntaxFlags()
+    if kind.hasError || layoutBuffer.contains(where: { $0?.hasError ?? false }) {
+      recursiveFlags.insert(.hasError)
+    }
```
rintaro, line 796: Please do this in the loop above
```swift
var recursiveFlags = RecursiveRawSyntaxFlags()
for case let node? in layoutBuffer {
  byteLength += node.byteLength
  descendantCount += node.totalNodes
  if node.hasError {
    recursiveFlags.insert(.hasError)
  }
  arena.adChild(node.arenaReference)
}
```

rintaro:
I realized that we might need/want this flag in `ParsedToken`(#597) too. Because for example the parser might want to lex e.g. unterminated regex literal as a `.regexLiteral`. For later diagnostics, without the flag in `ParsedToken`, we need to look into every `.regexLiteral` token if it has any error or not. Same for `stringLiteral`.

Sources/SwiftSyntax/RawSyntax.swift
ahoppen: Done. Also changed `RawSyntax` to have `recursiveFlags` property instead of `hasError` such that we can just do `recursiveFlags.insert(node.recursiveFlags)`, which will also work for all new recursive flags.

ahoppen:
@swift-ci Please test

Sources/SwiftSyntax/RawSyntax.swift
```diff
@@ -172,6 +187,19 @@ extension RawSyntax {
     kind.isUnknown
   }
 
+  var recursiveFlags: RecursiveRawSyntaxFlags {
+    switch rawData.payload {
```
rintaro, line 191: Since I merged `ParsedToken` PR, this `switch` needs to handle it. Sorry 🙇

ahoppen:
@swift-ci Please test

rintaro (approved):

PR #597, opened 2022-08-16
Introduce `RawSyntaxData.ParsedToken`

rintaro:
A new `RawSyntax` data variation. This is intended to be used for parsed token. Actual usage will be in follow-ups.

In this variation, unlike the existing `MaterializedToken`, leading/trailing trivia are stored in a `SyntaxText` form, and are lazily materialized by a function registered in `SyntaxArena`. Since the parser doesn't need to materialize them while parsing, it'd gain the "parsing" performance boost, in exchange for a performance reduction when accessing trivia.

### DIFF
--- a/Sources/SwiftSyntax/RawSyntax.swift
+++ b/Sources/SwiftSyntax/RawSyntax.swift
@@ -16,8 +16,26 @@ typealias RawTriviaPieceBuffer = UnsafeBufferPointer<RawTriviaPiece>
 /// Node data for RawSyntax tree. Tagged union plus common data.
 internal struct RawSyntaxData {
   internal enum Payload {
+    case parsedToken(ParsedToken)
     case materializedToken(MaterializedToken)
     case layout(Layout)
+  }
+
+  /// Token with lazy trivia parsing.
+  ///
+  /// The RawSyntax's `arena` must have a valid trivia parsing function to
+  /// lazily materialize the leading/trailing trivia pieces.
+  struct ParsedToken {
+    var tokenKind: RawTokenKind
+
+    /// Whole text of this token including leading/trailing trivia.
+    var wholeText: SyntaxText
+
+    /// Range of the actual token’s text.
+    ///
+    /// Text in `wholeText` before `textRange.lowerBound` is leading trivia and
+    /// after `textRange.upperBound` is trailing trivia.
+    var textRange: Range<SyntaxText.Index>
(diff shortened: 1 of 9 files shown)


rintaro:
@swift-ci Please test

Sources/SwiftSyntax/RawSyntax.swift
```diff
@@ -16,10 +16,21 @@ typealias RawTriviaPieceBuffer = UnsafeBufferPointer<RawTriviaPiece>
...
   }
 
+  /// Token with lazy trivia parsing.
+  ///
+  /// The RawSyntax's `arena` must have a valid trivia parsing function to
+  /// lazily materialize the leading/trailing trivia.
+  struct ParsedToken {
+    var tokenKind: RawTokenKind
+    var wholeText: SyntaxText
+    var textRange: Range<SyntaxText.Index>
```
ahoppen, line 31: Maybe worth adding a doc comment like
```swift
/// Range of the actual token’s text. Text in `wholeText` before `textRange.lowerBound` is leading trivia and after `textRange.upperBound` is trailing trivia.
```

rintaro:
@swift-ci Please test

Sources/SwiftSyntax/RawSyntax.swift
```diff
@@ -24,10 +24,17 @@ internal struct RawSyntaxData {
   /// Token with lazy trivia parsing.
   ///
   /// The RawSyntax's `arena` must have a valid trivia parsing function to
-  /// lazily materialize the leading/trailing trivia.
+  /// lazily materialize the leading/trailing trivia xpieces.
```
ahoppen, line 27: Typo?
```suggestion
  /// lazily materialize the leading/trailing trivia pieces.
```

Sources/SwiftSyntax/RawSyntax.swift
```diff
@@ -24,10 +24,17 @@ internal struct RawSyntaxData {
...
   struct ParsedToken {
     var tokenKind: RawTokenKind
+
+    /// Whole of this token including leading/trailing trivia.
```
ahoppen, line 31:
```suggestion
    /// Whole text of this token including leading/trailing trivia.
```

ahoppen (approved):

rintaro:
@swift-ci Please test

PR #599, opened 2022-08-17
Consdider empty string segments as present

ahoppen:
String segments can be empty if they occur in an empty string literal or in between two interpolation segments.


ahoppen:
@swift-ci Please test

CodaFi (approved):

rintaro (approved):

PR #601, opened 2022-08-17
Restore the Ability to Build with Xcode 13

CodaFi:


CodaFi:
@swift-ci test

Sources/_SwiftSyntaxTestSupport/Syntax+Assertions.swift
```diff
@@ -73,12 +75,14 @@ public func XCTAssertHasSubstructure(
 }
 
 /// See `SubtreeMatcher.assertSameStructure`.
-public func XCTAssertHasSubstructure(
-  _ actualTree: SyntaxProtocol,
-  _ expected: SyntaxProtocol,
+public func XCTAssertHasSubstructure<ActualTree, ExpectedTree>(
```
ahoppen, line 78: Can’t this just be `<ActualTree: SyntaxProtocl, ExpectedTree: SyntaxProtocol>` instead of the `where` clause?

bnbarham (approved):

Sources/_SwiftSyntaxTestSupport/Syntax+Assertions.swift
CodaFi: It can, but that's 57 characters of generic signature on top of the 38 here which is just really cluttered.

ahoppen: OK, I think I prefer declaring the requirement where the generic parameters are defined. But if you don't, that's fine.

PR #602, opened 2022-08-18
Make SyntaxArena.allocator immutable

rintaro:
`var` -> `let`. `SyntaxAnena` and `BumpPtrAllocator` lifetime should be 1:1.
Furthermore, if it's `var`, the arena needs to `retain`/`release` the allocator everytime calling it.


rintaro:
@swift-ci Please test

PR #603, opened 2022-08-18
Performance tweak for RawXXXSyntax initializers

rintaro:
* `RawXXXListSyntax.init` now only accepts `[Element]` instead of `Collection<Element>` to avoid potential cross module generics performance issue
* Avoid temporary `Array` creation in `RawXXXListSyntax.init`. Use a low-level initializer and directly initialize the elements buffer
* Add `__shared` to `arena: SyntaxArena` parameters for `RawXXXSyntax` initializers.

### DIFF
--- a/Sources/SwiftSyntax/RawSyntaxNodes.swift.gyb
+++ b/Sources/SwiftSyntax/RawSyntaxNodes.swift.gyb
@@ -59,7 +59,7 @@ public struct Raw${node.name}: Raw${node.name if node.is_base() else node.base_t
 %   end
 %
 %   if node.is_syntax_collection():
-  public init(elements: [Raw${node.collection_element_type}], arena: SyntaxArena) {
+  public init(elements: [Raw${node.collection_element_type}], arena: __shared SyntaxArena) {
     let raw = RawSyntax.makeLayout(
       kind: .${node.swift_syntax_kind}, uninitializedCount: elements.count, arena: arena) { layout in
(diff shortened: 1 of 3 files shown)


rintaro:
@swift-ci Please test

Sources/SwiftSyntax/RawSyntaxNodes.swift.gyb
```diff
@@ -59,8 +59,15 @@ public struct Raw${node.name}: Raw${node.name if node.is_base() else node.base_t
 %   end
 %
 %   if node.is_syntax_collection():
-  public init<C: Collection>(elements: C, arena: SyntaxArena)  where C.Element == Raw${node.collection_element_type} {
-    let raw = RawSyntax.makeLayout(kind: .${node.swift_syntax_kind}, from: elements.map { $0.raw }, arena: arena)
+  public init(elements: [Raw${node.collection_element_type}], arena: __shared SyntaxArena) {
```
ahoppen, line 62: Could we get the same performance improvement by adding
```swift
@_specialize(where C == Array<Raw${node.collection_element_type}>)
```
to the declaration while still having the convenience of being able to pass other collections in?

rintaro: Probably yes. But it would increase the code size. Since most clients should never use these `RawSyntax` APIs, I don't think it's worth the bother.

PR #608, opened 2022-08-19
Improve BumpPtrAllocator performance

rintaro:
Eliminate unnecessary slab array access when allocating from the current slab. Also factored out "allocate from the current slab" to a function.

### DIFF
--- a/Sources/SwiftSyntax/BumpPtrAllocator.swift
+++ b/Sources/SwiftSyntax/BumpPtrAllocator.swift
@@ -20,18 +20,19 @@ public class BumpPtrAllocator {
   static private var SLAB_ALIGNMENT: Int = 8
 
   private var slabs: [Slab]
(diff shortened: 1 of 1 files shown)


Sources/SwiftSyntax/BumpPtrAllocator.swift
```diff
@@ -79,12 +98,8 @@ public class BumpPtrAllocator {
     _totalBytesAllocated += byteCount
 
     // Check if the current slab has enough space.
-    if !slabs.isEmpty {
```
rintaro, line 82: This was the "unnecessary slab array access."

rintaro:
@swift-ci Please test

ahoppen (approved):

rintaro:
@swift-ci Please test Linux

PR #611, opened 2022-08-19
Generate `Format` using `SwiftSyntaxBuilder`

fwcd:
In preparation for some a small refactor to the formatting/leading trivia generation, this PR ports `Format.swift` to a file generated by SwiftSyntaxBuilder's DSL.


fwcd:
@swift-ci please test

fwcd:
Some observations I've made while tinkering on `FormatFile` that aren't directly related to this PR, but to the future plan of making `Format` responsible for generating leading trivia based on a node rather than the generated `build...` methods:

- We can't determine `requiresLeadingNewline` solely based on a passed node, since it is a property of `Child` rather than `Node`
- Even in collection nodes, the newline trivia attached to the elements crucially depends on the fact that the parent is a collection
- `_makeIndent()` has to be public or at least internal to be testable (so we cannot fully abstract these internals away, but that shouldn't be much of an issue AFAICT)

My current experiments can be found [here](https://github.com/fwcd/swift-syntax/blob/a159564ddc1077dafe90554ccabc6ea5e158dc70/Sources/SwiftSyntaxBuilder/generated/Format.swift#L33-L40) and [here](https://github.com/fwcd/swift-syntax/blob/6b093310cbfa609c5d859632683d0954c0b33933/Sources/SwiftSyntaxBuilder/generated/Format.swift#L33-L46).

Since the node-child relationship determines most the newline-related formatting anyway, I am not entirely sure whether this approach would really be feasible. Any thoughts?

cc @ahoppen

ahoppen (approved):

ahoppen:
> * We can't determine `requiresLeadingNewline` solely based on a passed node, since it is a property of `Child` rather than `Node`
> * Even in collection nodes, the newline trivia attached to the elements crucially depends on the fact that the parent is a collection

Since we `SwiftSyntaxBuilder` builds the tree top-down, I think we would pass the parent node to `_leadingTrivia`. That way you should be able to determine all the child properties, I believe.

> * `_makeIndent()` has to be public or at least internal to be testable (so we cannot fully abstract these internals away, but that shouldn't be much of an issue AFAICT)

You could use `@_spi(Testing)` like we do here: https://github.com/apple/swift-syntax/blob/6ce8c4871484da39be1e92b04ca885461c1995b0/Sources/SwiftSyntax/gyb_generated/TokenKind.swift#L315

PR #612, opened 2022-08-19
Add `+=` operator for `Trivia`

fwcd:
A small convenience that makes it more convenient to construct `Trivia` imperatively.

### DIFF
--- a/Sources/SwiftSyntax/Trivia.swift.gyb
+++ b/Sources/SwiftSyntax/Trivia.swift.gyb
@@ -210,9 +210,16 @@ extension Trivia: CustomStringConvertible {
   }
 }
 
-/// Concatenates two collections of `Trivia` into one collection.
-public func +(lhs: Trivia, rhs: Trivia) -> Trivia {
-  return Trivia(pieces: lhs.pieces + rhs.pieces)
+extension Trivia {
+  /// Concatenates two collections of `Trivia` into one collection.
+  public static func +(lhs: Trivia, rhs: Trivia) -> Trivia {
+    return Trivia(pieces: lhs.pieces + rhs.pieces)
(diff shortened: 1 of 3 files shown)


fwcd:
@swift-ci please test

fwcd:
@swift-ci please test

Sources/SwiftSyntax/Trivia.swift.gyb
```diff
@@ -215,6 +215,11 @@ public func +(lhs: Trivia, rhs: Trivia) -> Trivia {
   return Trivia(pieces: lhs.pieces + rhs.pieces)
 }
 
+/// Concatenates two collections of `Trivia` into the left-hand side.
+public func +=(lhs: inout Trivia, rhs: Trivia) {
+  lhs = lhs + rhs
+}
+
```
ahoppen, line 222: I have a vague memory that from a type checking performance perspective, it’s better to implement these as static methods inside the `Trivia` type itself.

```suggestion
extension Trivia {
  /// Concatenates two collections of `Trivia` into the left-hand side.
  public static func +=(lhs: inout Trivia, rhs: Trivia) {
    lhs = lhs + rhs
  }
}

```

fwcd: Does this apply to the `+` operator too?

CodaFi: It used to, but we removed the designated types optimization that supported it. Now it _should_ mostly a hygiene thing, and from that perspective I agree with Alex that this should be in extension context.

fwcd:
@swift-ci please test

CodaFi (approved):

PR #616, opened 2022-08-22
Add a Swift parser library that is written in Swift

ahoppen:
This starts an effort to implement the parser for the Swift compiler in Swift itself.

For more information, please see the announcement forum post at: https://forums.swift.org/t/a-new-swift-parser-for-swiftsyntax/59813

### DIFF
--- a/Sources/SwiftParser/README.md
+++ b/Sources/SwiftParser/README.md
@@ -1,0 +1,70 @@
+## Overview
+
+The `SwiftParser` framework implements a parser that accepts Swift source text
+as input and produces a SwiftSyntax syntax tree. This module is under active development and is not yet ready to completely replace `SwiftSyntaxParser`. For more information about the design of this module, please see [the module documentation](SwiftParser.docc/SwiftParser.md).
+
+## Quickstart
+
+The easiest way to parse Swift source code is to call the `Parser.parse` method, providing it with a string containing the source code:
+
+```swift
+import SwiftParser
+import SwiftSyntax
+
+let sourceText =
+"""
+func greeting(name: String) {
+  print("Hello, \(name)!")
+}
+"""
+
+// Parse the source code in sourceText into a syntax tree
+let sourceFile: SourceFileSyntax = Parser.parse(source: sourceText)
+
+// The "description" of the source tree is the source-accurate view of what was parsed.
+assert(sourceFile.description == sourceText)
+
+// Visualize the complete syntax tree.
+dump(sourceFile)
+```
+
+## Implementation Status
+
+There are a number of implementation tasks involved in realizing the parser, which include:
+
+* [ ] Parsing
+  * [x] Rough parse of full grammar
+  * [x] Recovery from unexpected syntax
+  * [x] Insert missing syntax where needed
+  * [ ] SIL syntax
+  * [ ] [Regular expression literals](https://github.com/apple/swift-experimental-string-processing)
+  * [x] Parse into raw syntax nodes
+* [ ] Diagnostics
+  * [ ] Render diagnostics to a terminal with color, highlights, etc.
+  * [ ] Fix-Its to add/remove/replace syntax
+  * [ ] Emit diagnostics in the same binary format that the Swift compiler/driver do
+  * [x] Syntax visitor to emit diagnostics for all missing and unexpected syntax
+* [ ] Tooling and automation
+  * [ ] Improve test harness to easily check the expected parse trees + diagnostics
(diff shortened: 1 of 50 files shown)


ahoppen:
@swift-ci Please test

CodaFi (approved):

rintaro (approved):

etcwilde:
@compnerd, if you have a cycle or two, it's probably worth making sure this works on Windows.

DougGregor (approved):

zoecarver (approved):

compnerd:
Doesn't seem to build on Windows (which is a regression).  At the very least we will need https://github.com/apple/swift-docc-plugin/pull/26.

CodaFi:
The docc plugin is not required for this repository in general, and is not a dependency for this PR specifically. Docc support is built into Xcode, I am not sure why we have a plugin dependency as well.

compnerd:
I couldn't answer why we have that.  We should invert the condition so that it is not used by default.

DougGregor:
I'm fine with removing the DocC plugin dependency or inverting the condition. We shouldn't land this in a form that breaks the Windows build.

CodaFi:
Let's strike it for now. We shouldn't have merged it without verifying that it worked for Windows

https://github.com/apple/swift-syntax/pull/617

Sources/swift-parser-test/swift-parser-test.swift
```diff
@@ -0,0 +1,312 @@
...
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+
+import SwiftSyntax
+import SwiftParser
+import Foundation
+import ArgumentParser
```
compnerd, line 17:
```suggestion
import ArgumentParser
#if os(Windows)
import WinSDK
#endif
```

Sources/swift-parser-test/swift-parser-test.swift
```diff
@@ -0,0 +1,312 @@
...
+      let sema = DispatchSemaphore(value: 0)
+      process.standardOutput = FileHandle.nullDevice
+      process.standardError = FileHandle.nullDevice
+      process.terminationHandler = { process in
+        sema.signal()
+      }
+
+      try process.run()
+      if sema.wait(timeout: DispatchTime.now() + .seconds(1)) == .timedOut {
+        kill(pid_t(process.processIdentifier), SIGKILL)
```
compnerd, line 232:
```suggestion
#if os(Windows)
        _ = TerminateProcess(process.processHandle, 0)
#else
        kill(pid_t(process.processIdentifier), SIGKILL)
#endif
```

compnerd (commented):
This at least allows this PR to be built on Windows with:
```
swift build -Xlinker -Xcc -IC:\Library\Developer\Toolchains\unknown-Asserts-development.xctoolchain\usr\include
```

Note that building this requires a locally built toolchain as there was a fix for SPM that is needed that has not had a new toolchain snapshot built due to other failures introduced by the rebranch.

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

Sources/SwiftParser/README.md
```diff
@@ -0,0 +1,70 @@
+## Overview
```
benlangmuir, line 1: (not a blocker) The existence of this file triggers a warning, at least with 5.7:

```
    Sources/SwiftParser/README.md
```

wu-s-john:
Nice. Swift would really benefit from code generating it's own code rather than C++.

Sources/SwiftParser/README.md
DougGregor: I'll merge and we'll fix it up afterward. Thanks!

compnerd:
> I can assure you that this does currently not work on Windows. I just tried using this framework in an SPM project on Windows.

Please use a build from main.  At the very least, a recent snapshot.

broken-bytes:
I used the most recent version.

```dependencies: [
        .package(
            url: "https://github.com/apple/swift-syntax.git", from: "0.50600.1"),
    ],
    ```
    As far as I am concerned, this is the most recent snapshot

compnerd:
5.6 should be buildable - https://github.com/compnerd/swift-build/releases/tag/swift-format-5.6-RELEASE was built and would pull in swift-syntax 5.6.  I wonder if the .1 has anything to do with it 🤔

DougGregor:
@broken-bytes If you're looking to try the new parser that's the subject of this pull request, you'll need `main`.

PR #617, opened 2022-08-22
Remove Docc Plugin Dependency

CodaFi:
Resolve a regression in the Windows build. We can restore this
dependency once we've verified the correct way for all clients
of this library to consume it.

NOTE: This does not affect existing docc documentation. You can build
that at your desk in Xcode with Product > Build Documentation.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -149,12 +149,3 @@ let package = Package(
     ),
   ]
 )
-
-#if swift(>=5.6)
-// If `SWIFTCI_USE_LOCAL_DEPS` is set, we are in a CI enviornment that might disallow
-// internet access, so we can't load swift-docc-plugin. Simply don't load it in these
-// environments because we don't build docc in CI at the moment.
-if ProcessInfo.processInfo.environment["SWIFTCI_USE_LOCAL_DEPS"] == nil {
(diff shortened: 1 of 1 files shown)


CodaFi:
@swift-ci test

Package.swift
```diff
@@ -149,12 +149,3 @@ let package = Package(
     ),
   ]
 )
-
-#if swift(>=5.6)
```
ethan-kusters, line 153: Should we consider just also adding a conditional check for Windows here? Swift-DocC doesn't ship in the Windows toolchain so the plugin won't be applicable there for some time.

It might be worth leaving it around for non-Xcode users in the meantime though.

CodaFi: It's still unclear to me why we need a dependency on this plugin to ship docc docs outside of Xcode. To my mind, this is something SourceKit-lsp should take care of.

ethan-kusters: It's for building and previewing documentation locally. Longer term, I agree that the plugin should probably be included in the toolchain by default (I'm not sure if sourcekit-lsp is the right place or if we should have a notion of default SwiftPM plugins) so that we have out-of-the-box support for docs. But we would need to have more discussion around that.

compnerd: The plugin can be built on Windows even if the rest of DocC cannot be supported due to the dependency on NIO.

ahoppen (approved):

PR #619, opened 2022-08-23
Introduce a library to handle user-defined operators

DougGregor:
Introduce a library that implements the semantics of Swift user-defined operators and precedence groups. The primary operation is a "fold", which takes a sequence expression (i.e., something like `a + b * c`) and turns it into a structured syntax tree that reflects the relative precedence of the operators involved (here, `*` has higher precedence than `+`). This is one step beyond simple parsing that provides the necessary structure for later stages of processing of a Swift syntax tree.

There are several pieces to this new `SwiftOperators` library, including:
* `PrecedenceGroup` and `Operator`, which are semantic representations of precedence groups and operator declarations.
* `PrecedenceGraph`, which keeps track of a set of precedence groups and their relationships to each other, and is able to answer questions like "how to the precedences of two different groups compare?".
* Syntactic -> semantic translation to go from `PrecedenceGroupDeclSyntax` and `OperatorDeclSyntax` nodes to `PrecedenceGroup` and `Operator`, by walking the child syntax.
* `OperatorTable`, which contains a `PrecedenceGraph` that can be created by processing a source file to pick out the various operator and precedence-group declarations. It provides the core "fold" operation that turns a `SequenceExprSyntax` node into one that is fully-structured to show the order of operations.
* Prefabricated `OperatorTable` instances for the logical operators (`&&` and `||`, as used in Swift `#if` conditions) and the Swift standard library operators (all the ones most Swift developers are used to), which can handle most Swift code.
* `OperatorError`, which describes the kinds of errors that can occur when handling operators, from missing precedence groups to ambiguous parses. There is a mapping from `OperatorError` to `Diagnostic` so that we can emit diagnostics for each error via the normal means.
* A new `--fold-sequences` option to `-swift-parser-test` to apply sequence folding to the input file as part of processing.

All that said, there are a few things that need to be finished before this library is ready:
* [x] Finish remaining syntactic expressions that use operator precedence:
  * [x] Explicit casts
  * [x] Ternary expressions (`? :`)
  * [x] Assignment expressions
  * [x] Arrow expression (`->`)
* [x] Integration in `swift-parser-test` via an option, so we can try this on lots more code
* [ ] Detect cycles in precedence groups
* [ ] Diagnostics for semantic errors in operator and precedence-group declarations
* [ ] Diagnostics for `try` that doesn't extend all the way to the right of ternary and assignment expressions
* [ ] Throwing lots and lots of code at it to make sure it works


DougGregor:
@swift-ci please test

Sources/SwiftOperatorPrecedence/OperatorPrecedence+Semantics.swift
```diff
@@ -0,0 +1,158 @@
...
+extension Operator {
+  /// Form the semantic definition of an operator given its syntax.
+  ///
+  /// TODO: This ignores all semantic errors.
+  init(from syntax: OperatorDeclSyntax) {
+    self.syntax = syntax
+
+    let kindModifier = syntax.modifiers?.first { modifier in
+      OperatorKind(rawValue: modifier.name.text) != nil
+    }
```
ahoppen, line 73: To avoid the force unwrap and `if` below, what do you think about
```suggestion
    kind = syntax.modifiers?.compactMap({ OperatorKind(rawValue: $0.name.text) }).first ?? .infix
```

Sources/SwiftOperatorPrecedence/OperatorPrecedenceError+Diagnostics.swift
```diff
@@ -0,0 +1,98 @@
...
+    case .missingGroup:
+      return "missing_group"
+
+    case .groupAlreadyExists:
+      return "group_already_exists"
+    }
+  }
+
+  public var diagnosticID: DiagnosticMessageID {
+      DiagnosticMessageID("SwiftOperatorPrecedence/\(diagnosticCaseID)")
```
ahoppen, line 61: Why not use reflection here?
```swift
DiagnosticMessageID("\(type(of: self)).\(self)")
```

Sources/SwiftOperatorPrecedence/OperatorPrecedenceError+Diagnostics.swift
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  /// Produce the syntax node at which a diagnostic should be displayed.
+  var diagnosticDisplayNode: Syntax {
+    switch self {
+    case .incomparableOperators(let leftOperator, _, _, _):
+      return Syntax(leftOperator)
+
+    case .missingOperator(_, let node):
+      return fixupDiagnosticDisplayNode(node)
```
ahoppen, line 81: A diagnostic should always be associated to a syntax node, otherwise it’s just not useful at all. Can’t we make the syntax nodes that we use here non-optional?

Sources/SwiftOperatorPrecedence/OperatorPrecedenceError.swift
```diff
@@ -0,0 +1,45 @@
...
+import SwiftSyntax
+
+/// Describes errors that can occur when working with operator precedence graphs.
+public enum OperatorPrecedenceError: Error {
+  /// Error produced when a given precedence group already exists in the
+  /// precedence graph.
+  case groupAlreadyExists(existing: PrecedenceGroup, new: PrecedenceGroup)
+
+  /// The named precedence group is missing from the precedence graph.
+  case missingGroup(PrecedenceGroupName, referencedFrom: SyntaxProtocol?)
```
ahoppen, line 21: The general paradigm is that we want to avoid `SyntaxProtocol` wherever possible to avoid the cost of existentials, so this should be `Syntax`. The only places where I’m usually using `SyntaxProtocol` as a type that’s not a generic requirement are test helpers that are not performance critical at all.

Sources/SwiftOperatorPrecedence/PrecedenceGroup.swift
```diff
@@ -0,0 +1,120 @@
...
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+import SwiftSyntax
+
+/// Names a precedence group.
+///
+/// TODO: For now, we'll use strings, but we likely want to move this to
+/// a general notion of an Identifier.
+public typealias PrecedenceGroupName = String
```
ahoppen, line 18: Just an idea: I’ve had good experience with declaring a new type
```swift
struct PrecedenceGroupName {
  let name: String
}
```

That way you can’t accidentally cast it to/from `String`, which would break if we switch the breaking storage later. 

That’s what I’m doing in `DiagnosticMessageID` at least.

Sources/SwiftOperatorPrecedence/SwiftOperatorPrecedence.docc/SwiftOperatorPrecedence.md
```diff
@@ -0,0 +1,89 @@
...
+```
+
+## Error handling
+
+By default, any of the operations that can produce an error, whether folding a sequence or parsing a source file's operators and precedence groups into a table, will throw an instance of <doc:OperatorPrecedenceError>. However, each entry point takes an optional error handler (of type <doc:OperatorPrecedenceErrorHandler>) that will be called with each error that occurs. For example, we can capture errors like this:
+
+```swift
+var errors: [OperatorPrecedenceError] = []
+let foldedExpr2e = opPrecedence.foldSingle(sequenceExpr2) { error in 
+    errors.append(error)
```
ahoppen, line 84:
```suggestion
  errors.append(error)
```

Sources/SwiftOperatorPrecedence/PrecedenceGraph.swift
```diff
@@ -0,0 +1,121 @@
...
+      for relation in currentGroup.relations {
+        if relation.kind == .lowerThan {
+          // If we hit our start group, we're done.
+          let otherGroupName = relation.groupName
+          if otherGroupName == startGroupName {
+            return .lowerThan
+          }
+
+          if groupsSeen.insert(otherGroupName).inserted {
+            stack.append((otherGroupName, relation.syntax))
```
ahoppen, line 85: Not really related to this line: Do we have some way to diagnose cycles in the precedence graph?

Sources/SwiftOperatorPrecedence/PrecedenceGraph.swift
```diff
@@ -0,0 +1,121 @@
...
+          if otherGroupName == endGroupName {
+            return .higherThan
+          }
+
+          if groupsSeen.insert(otherGroupName).inserted {
+            stack.append((otherGroupName, relation.syntax))
+          }
+        }
+      }
+    }
```
ahoppen, line 116: Would it be worth factoring this loop into a local function that can be used for both the `higherThan` and `lowerThan` scan?

Sources/SwiftOperatorPrecedence/OperatorPrecedence+Folding.swift
```diff
@@ -0,0 +1,364 @@
...
+
+  /// Determine the associativity between two precedence groups.
+  private func associativity(
+    firstGroup: PrecedenceGroupName?,
+    firstGroupSyntax: SyntaxProtocol?,
+    secondGroup: PrecedenceGroupName?,
+    secondGroupSyntax: SyntaxProtocol?,
+    errorHandler: OperatorPrecedenceErrorHandler = { throw $0 }
+  ) rethrows -> Associativity {
+    guard let firstGroup = firstGroup, let secondGroup = secondGroup else {
```
ahoppen, line 84: Why not make `firstGroup` and `secondGroup` non-optional if the first thing we do is return if they are `nil`?

Sources/SwiftOperatorPrecedence/OperatorPrecedence+Folding.swift
```diff
@@ -0,0 +1,364 @@
...
+      errorHandler: errorHandler
+    ) == .higherThan {
+      return .left
+    }
+
+    if try precedence(
+      relating: secondGroup, to: firstGroup,
+      startSyntax: secondGroupSyntax, endSyntax: firstGroupSyntax,
+      errorHandler: errorHandler
+    ) == .higherThan {
```
ahoppen, line 111: Isn’t `precedence` guaranteed to be commutative. I.e. should’t this check be the same as checking that the `precedence` call in the `if`-statement above returned `.lowerThan`?

Sources/SwiftOperatorPrecedence/OperatorPrecedence+Folding.swift
```diff
@@ -0,0 +1,364 @@
...
+    }
+
+    override func visit(_ node: SequenceExprSyntax) -> ExprSyntax {
+      // If the error handler threw in response to an error, don't
+      // continue folding.
+      if firstFatalError != nil {
+        return ExprSyntax(node)
+      }
+
+      let newNode = super.visit(node).as(SequenceExprSyntax.self)!
```
ahoppen, line 313: I don’t think this is necessary. The base `SyntaxRewriter` is a no-op.

Sources/SwiftOperatorPrecedence/OperatorPrecedence+Folding.swift
```diff
@@ -0,0 +1,364 @@
...
+      let folder = SequenceFolder(
+        opPrecedence: self, errorHandler: errorHandler
+      )
+      let result = folder.visit(Syntax(node))
+
+      // If the sequence folder encountered an error that caused the error
+      // handler to throw, invoke the error handler again with the original
+      // error.
+      if let origFatalError = folder.firstFatalError {
+        try errorHandler(origFatalError)
```
ahoppen, line 358: Doesn’t this call `errorHandler` twice? Once from `SequenceFolder` and once from here?

Sources/SwiftOperatorPrecedence/OperatorPrecedenceError+Diagnostics.swift
DougGregor: I agree that it would be much better for us always to have a syntax node. The main issue right now is that we don't have syntax nodes for the prefabricated operator tables... but there's no reason I can't fabricate them for diagnostic purposes! I'll give it a try.

Sources/SwiftOperatorPrecedence/OperatorPrecedenceError.swift
DougGregor: Okay!

Sources/SwiftOperatorPrecedence/OperatorPrecedence+Folding.swift
DougGregor: I needed it to visit the children, or else this transform wouldn't be recursive. Am I holding it wrong?

Sources/SwiftOperatorPrecedence/OperatorPrecedence+Folding.swift
DougGregor: Yeah, the problem here is that `rethrows` checking cannot account for errors thrown in the visitor. I'll document this as a limitation. The real fix is something like https://github.com/apple/swift-evolution/pull/1271.

DougGregor:
@ahoppen thanks for the great review! I answered a few things up front, and will go through and address the rest of your comments in the code before replying.

Sources/SwiftOperatorPrecedence/OperatorPrecedence+Folding.swift
```diff
@@ -0,0 +1,364 @@
...
+      )
+    }
+
+    // FIXME: Handle all of the language-defined precedence relationships.
+    return nil
+  }
+
+  /// Form a binary operation expression for, e.g., a + b.
+  private func makeBinaryOperationExpr(
+    lhs: ExprSyntax, op: ExprSyntax, rhs: ExprSyntax
```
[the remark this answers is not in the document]
ahoppen: OK, that makes sense 👍

Sources/SwiftOperatorPrecedence/OperatorPrecedence+Folding.swift
ahoppen: Nevermind, I was stupid.

Sources/SwiftOperatorPrecedence/OperatorPrecedence+Defaults.swift
```diff
@@ -0,0 +1,479 @@
...
+  /// example as it is used in `#if` processing.
+  public static var logicalOperators: OperatorPrecedence {
+    var opPrecedence = OperatorPrecedence()
+
+    try! opPrecedence.record(
+      Operator(kind: .infix, name: "&&",
+               precedenceGroup: "LogicalConjunctionPrecedence"))
+    try! opPrecedence.record(
+      Operator(kind: .infix, name: "||",
+               precedenceGroup: "LogicalDisjunctionPrecedence"))
```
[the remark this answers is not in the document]
DougGregor: We do. I haven't added all of the prefix/postfix operators to the table, because right now they're just getting dropped... but I can go do that.

Sources/SwiftOperatorPrecedence/OperatorPrecedence.swift
```diff
@@ -0,0 +1,105 @@
...
+/// providing the core operations for "folding" sequence expression syntax into
+/// a structured expression syntax tree.
+public struct OperatorPrecedence {
+  var precedenceGraph: PrecedenceGraph = .init()
+  var operators: [OperatorName : Operator] = [:]
+
+  public init() { }
+
+  /// Record the operator, if it matters.
+  /// FIXME: Terrible API used only for tests
```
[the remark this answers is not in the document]
DougGregor: I've made it internal, since I no longer need it for testing.

Sources/SwiftOperatorPrecedence/OperatorPrecedence.swift
```diff
@@ -0,0 +1,105 @@
...
+  public mutating func record(
+    _ op: Operator,
+    errorHandler: OperatorPrecedenceErrorHandler = { throw $0 }
+  ) rethrows {
+    // FIXME: Could do operator-already-exists checking for prefix/postfix
+    // operators as well, since we parse them.
+    if op.kind != .infix { return }
```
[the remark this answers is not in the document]
DougGregor: I was thinking that for prefix/postfix operators we could still use all of the same logic for (e.g.) diagnosing conflicting operators and such. Perhaps that means the whole thing is mis-named, and we should call it the "operator table" rather than "operator precedence".

Sources/SwiftOperatorPrecedence/OperatorPrecedence.swift
```diff
@@ -0,0 +1,105 @@
...
+      try errorHandler(
+        .missingOperator(operatorName, referencedFrom: syntax))
+      return nil
+    }
+
+    return op.precedenceGroup
+  }
+
+  /// Look for the precedence group corresponding to the given operator.
+  func lookupOperatorPrecedenceGroup(
```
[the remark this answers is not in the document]
DougGregor: Oh hey, you're right!

Sources/SwiftOperatorPrecedence/OperatorPrecedence+Folding.swift
```diff
@@ -0,0 +1,364 @@
...
+    let isStrict: Bool
+    let syntax: SyntaxProtocol?
+  }
+
+  /// Determine whether we should consider an operator in the given group
+  /// based on the specified bound.
+  private func shouldConsiderOperator(
+    fromGroup groupName: PrecedenceGroupName?,
+    in bound: PrecedenceBound,
+    fromGroupSyntax: SyntaxProtocol?,
```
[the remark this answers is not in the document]
DougGregor: Good point, this really should be "operatorSyntax", because it's the syntax node for the operator that we're evaluating.

Sources/SwiftOperatorPrecedence/OperatorPrecedence+Semantics.swift
```diff
@@ -0,0 +1,158 @@
...
+  /// Form the semantic definition of a precedence group given its syntax.
+  ///
+  /// TODO: This ignores all semantic errors.
+  init(from syntax: PrecedenceGroupDeclSyntax) {
+    self.name = syntax.identifier.text
+    self.syntax = syntax
+
+    for attr in syntax.groupAttributes {
+      // Relation (lowerThan, higherThan)
+      if let relation = attr.as(PrecedenceGroupRelationSyntax.self) {
```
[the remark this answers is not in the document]
DougGregor: Very cool, I like it!

Sources/SwiftOperatorPrecedence/PrecedenceGraph.swift
DougGregor: Yes, it's much cleaner this way.

Sources/SwiftOperatorPrecedence/OperatorPrecedence.swift
```diff
@@ -0,0 +1,105 @@
...
+  ) rethrows {
+    try precedenceGraph.add(group, errorHandler: errorHandler)
+  }
+}
+
+extension OperatorPrecedence {
+  /// Look for the precedence group corresponding to the given operator.
+  func lookupOperatorPrecedenceGroupName(
+    _ operatorName: OperatorName,
+    referencedFrom syntax: SyntaxProtocol?,
```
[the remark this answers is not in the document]
DougGregor: It's not just operators, but I'll eliminate the existential to use Syntax instead.

Sources/SwiftOperatorPrecedence/Operator.swift
```diff
@@ -0,0 +1,50 @@
...
+
+/// Describes an operator.
+public struct Operator {
+  public let kind: OperatorKind
+  public let name: OperatorName
+  public let precedenceGroup: PrecedenceGroupName?
+  public let syntax: OperatorDeclSyntax?
+
+  public init(
+    kind: OperatorKind, name: OperatorName,
```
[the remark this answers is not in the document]
ahoppen: OK, I think that makes sense. I’m fine to keep it.

Sources/SwiftOperatorPrecedence/OperatorPrecedence+Defaults.swift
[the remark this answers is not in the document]
ahoppen: Ah, I didn’t realize that. If that’s something you know still needs to be done, I’m fine with that.

Sources/SwiftOperatorPrecedence/OperatorPrecedence.swift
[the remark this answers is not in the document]
ahoppen: I agree, if this was named `OperatorTable`, using it to diagnose conflicting operators would make sense 👍

Sources/SwiftOperatorPrecedence/OperatorPrecedence+Folding.swift
```diff
@@ -0,0 +1,467 @@
...
+    }
+
+    // FIXME: Fallback that we should never need
+    fatalError("Unknown binary operator")
+  }
+
+  /// Determine the associativity between two precedence groups.
+  private func associativity(
+    firstGroup: PrecedenceGroupName?,
+    firstGroupSyntax: SyntaxProtocol?,
```
ahoppen, line 179: Similar to https://github.com/apple/swift-syntax/pull/619#discussion_r952216153, this should probably also be named differently.

Sources/SwiftOperatorPrecedence/PrecedenceGraph.swift
```diff
@@ -0,0 +1,149 @@
...
+      errorHandler: errorHandler
+    ).map {
+      $0.flipped
+    } ?? searchRelationships(
+      initialGroupName: endGroupName, initialSyntax: endSyntax,
+      targetGroupName: startGroupName, direction: .higherThan,
+      errorHandler: errorHandler
+    ).map {
+      $0.flipped
+    } ?? .unrelated
```
ahoppen, line 147: Reading this I initially thought that `$0.flipped` matched all previous checks (my brain somehow thought that `??`) has higher precedence than `.`. Just wondering whether this would read easier as follows (I’m also not a huge fan of the 20 line expression)
```suggestion
    if let relationship = try searchRelationships(
      initialGroupName: endGroupName, initialSyntax: endSyntax,
      targetGroupName: startGroupName, direction: .lowerThan,
      errorHandler: errorHandler
    ) {
      return relationship
    } else if let relationship = try searchRelationships(
      initialGroupName: startGroupName, initialSyntax: startSyntax,
      targetGroupName: endGroupName, direction: .higherThan,
      errorHandler: errorHandler
    ) {
      return relationship
    } else if let relationship = try searchRelationships(
      initialGroupName: startGroupName, initialSyntax: startSyntax,
      targetGroupName: endGroupName, direction: .lowerThan,
      errorHandler: errorHandler
    ) {
      return relationship.flipped
    } else if relationship = try searchRelationships(
      initialGroupName: endGroupName, initialSyntax: endSyntax,
      targetGroupName: startGroupName, direction: .higherThan,
      errorHandler: errorHandler
    ) {
      return relationship.flipped
    } else {
      return .unrelated
    }
```

Sources/SwiftOperatorPrecedence/OperatorPrecedence.swift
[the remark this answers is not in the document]
ahoppen: Mostly out of curiosity: What other types can `referencedFrom` be other than `TokenSyntax`?

Sources/SwiftOperatorPrecedence/PrecedenceGraph.swift
```diff
@@ -0,0 +1,149 @@
...
+
+  /// Determine the precedence relationship between two precedence groups.
+  ///
+  /// Follow the precedence relationships among the precedence groups to
+  /// determine the precedence of the start group relative to the end group.
+  ///
+  /// - Returns: Precedence.lowerThan if startGroupName has lower precedence
+  /// than endGroupName, Precedence.higherThan if startGroupName has higher
+  /// precedence than endGroup name, and Precedence.unrelated otherwise.
+  func precedence(
```
ahoppen, line 114: https://github.com/apple/swift-syntax/pull/619#discussion_r952214981 still applies here. I think a doc comment describing how `startSyntax` and `endSyntax` relate to `startGroupName` and `endGroupName` would be useful here.

Sources/SwiftOperatorPrecedence/OperatorPrecedenceError+Diagnostics.swift
```diff
@@ -0,0 +1,103 @@
...
+      if leftGroup == rightGroup {
+        return "adjacent operators are in non-associative precedence group '\(leftGroup)'"
+      }
+
+      return "adjacent operators are in unordered precedence groups '\(leftGroup)' and '\(rightGroup)'"
+    }
+  }
+
+  /// A string representation of each case.
+  private var diagnosticCaseID: String {
```
ahoppen, line 46: Is there any particular reason why you want to manually specify these names instead of deriving them using reflection like we do in the parser?

ahoppen:
rdar://99492978

Sources/SwiftOperatorPrecedence/OperatorPrecedence+Folding.swift
DougGregor: That would force callers to effectively do the same computation, so it's better to have it done once here.

Sources/SwiftOperatorPrecedence/PrecedenceGroup.swift
DougGregor: I rather hope we don't need a specific abstraction here in this library, but instead that we can introduce the notion of an `Identifier` into `SwiftSyntax` itself.

Sources/SwiftOperatorPrecedence/PrecedenceGraph.swift
DougGregor: No, we don't yet have a way to diagnose cycles, but it should!

Sources/SwiftOperatorPrecedence/OperatorPrecedence.swift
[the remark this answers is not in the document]
DougGregor: `UnresolvedTernaryExprSyntax`, `UnresolvedIsExprSyntax`, `UnresolvedAsExprSyntax`. We need the whole expression node here, although we'll take it apart when we resolve it.

Sources/SwiftOperatorPrecedence/OperatorPrecedenceError+Diagnostics.swift
DougGregor: No reason; I'll switch to reflection.

Sources/SwiftOperatorPrecedence/PrecedenceGraph.swift
DougGregor: To my eye, both look equally messy. I'd rather find a better answer, perhaps with a real data structure.

Sources/SwiftOperatorPrecedence/PrecedenceGroup.swift
ahoppen: OK, makes sense 👍

Sources/SwiftOperatorPrecedence/OperatorPrecedence+Folding.swift
ahoppen: I think I would have probably done it on the call side, but I’m fine to keep it as-is.

Sources/SwiftOperatorPrecedence/PrecedenceGraph.swift
DougGregor: (Added to my TODO list)

DougGregor:
@swift-ci please test

Sources/SwiftOperatorPrecedence/OperatorPrecedenceError+Diagnostics.swift
DougGregor: I went ahead and added the fabrication code to eliminate the optionals.

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

PR #623, opened 2022-08-23
Several improvements to diagnostics in the new parser

ahoppen:
- Move diagnostic types to their own modules (this will make it easier to use these diagnostics e.g. from the operator precedence library https://github.com/apple/swift-syntax/pull/619)
- Add severity, Fix-It and highlight range support
- Add a domain to `DiagnosticID`

The goal of this PR is to provide the mentioned functionality above. At this point, the goal is not to review ParseDiagnosticsGenerator.swift. I know that the generator has super rough edges at the moment but before optimizing it for better ergonomics, I would like to implement a few more diagnostics to see what the common patterns are.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -59,6 +59,10 @@ let package = Package(
       ]
     ),
     .target(
+      name: "SwiftDiagnostics",
+      dependencies: ["SwiftSyntax"]
+    ),
+    .target(
       name: "SwiftSyntax",
       dependencies: ["_CSwiftSyntax"],
       exclude: [
@@ -105,7 +109,7 @@ let package = Package(
     ),
     .target(
       name: "SwiftParser",
-      dependencies: ["SwiftSyntax"],
+      dependencies: ["SwiftDiagnostics", "SwiftSyntax"],
       exclude: [
         "README.md"
       ]
@@ -116,7 +120,7 @@ let package = Package(
     ),
     .executableTarget(
       name: "swift-parser-test",
-      dependencies: ["SwiftSyntax", "SwiftParser", .product(name: "ArgumentParser", package: "swift-argument-parser")]
+      dependencies: ["SwiftDiagnostics", "SwiftSyntax", "SwiftParser", .product(name: "ArgumentParser", package: "swift-argument-parser")]
     ),
     .executableTarget(
         name: "generate-swift-syntax-builder",
@@ -162,7 +166,7 @@ let package = Package(
     ),
     .testTarget(
       name: "SwiftParserTest",
-      dependencies: ["SwiftParser", "_SwiftSyntaxTestSupport"]
+      dependencies: ["SwiftDiagnostics", "SwiftParser", "_SwiftSyntaxTestSupport"]
     ),
   ]
 )
--- a/Sources/SwiftDiagnostics/FixIt.swift
+++ b/Sources/SwiftDiagnostics/FixIt.swift
@@ -1,0 +1,46 @@
+//===--- FixIt.swift ------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2022 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
+/// Types conforming to this protocol represent Fix-It messages that can be
+/// shown to the client.
+/// The messages should describe the change that the Fix-It will perform
+public protocol FixItMessage {
+  /// The diagnostic message that should be displayed in the client.
+  var message: String { get }
+
+  /// See ``MessageID``.
+  var fixItID: MessageID { get }
+}
+
+
+/// A Fix-It that can be applied to resolve a diagnostic.
+public struct FixIt {
+  public enum Change {
+    /// Replace `oldNode` by `newNode`.
+    case replace(oldNode: Syntax, newNode: Syntax)
+  }
+
+  /// A description of what this Fix-It performs.
+  public let message: FixItMessage
+
+  /// The changes that need to be performed when the Fix-It is applied.
+  public let changes: [Change]
+
+  public init(message: FixItMessage, changes: [Change]) {
+    assert(!changes.isEmpty, "A Fix-It must have at least one diagnostic associated with it")
+    self.message = message
+    self.changes = changes
+  }
+}
+
--- a/Sources/SwiftParser/Diagnostics/ParseDiagnosticsGenerator.swift
+++ b/Sources/SwiftParser/Diagnostics/ParseDiagnosticsGenerator.swift
@@ -96,13 +96,12 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
     if shouldSkip(node) {
       return .skipChildren
     }
+    // Detect C-style for loops based on two semicolons which could not be parsed between the 'for' keyword and the '{'
     // This is mostly a proof-of-concept implementation to produce more complex diagnostics.
-    if let unexpectedCondition = node.body.unexpectedBeforeLeftBrace {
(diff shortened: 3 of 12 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

DougGregor (approved):

Package.swift
```diff
@@ -58,6 +58,10 @@ let package = Package(
         "README.md"
       ]
     ),
+    .target(
+      name: "Diagnostics",
```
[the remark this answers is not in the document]
ahoppen: Good idea. I changed it.

Sources/SwiftParser/Diagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -90,13 +115,25 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
     // This is mostly a proof-of-concept implementation to produce more complex diagnostics.
-    if let unexpectedCondition = node.body.unexpectedBeforeLeftBrace {
-      // Detect C-style for loops based on two semicolons which could not be parsed between the 'for' keyword and the '{'
-      if unexpectedCondition.tokens(withKind: .semicolon).count == 2 {
-        addDiagnostic(node, CStyleForLoopDiagnostic())
-        markNodesAsHandled(node.inKeyword.id, unexpectedCondition.id)
-      }
+    if let unexpectedCondition = node.body.unexpectedBeforeLeftBrace,
+       unexpectedCondition.tokens(withKind: .semicolon).count == 2 {
+      // FIXME: This is aweful. We should have a way to either get all children between two cursors in a syntax node or highlight a range from one node to another.
```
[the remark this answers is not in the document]
ahoppen: I totally agree that this is awful. The reason why I didn’t want to use a range was that that might now become a runtime error if the end node is not after the start node (or even in a different syntax tree). 

As I mentioned in the PR description, I would like to implement a few more diagnostics first to find common patterns that we use in diagnostic generation before designing the entire diagnostic system. E.g. if it turns out that this is the only diagnostic that needs to highlight >3 syntax nodes, I would maybe be fine with it. Or it would certainly influence our design.

ahoppen:
@swift-ci Please test

Sources/SwiftParser/Diagnostics/ParseDiagnosticsGenerator.swift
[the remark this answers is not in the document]
ahoppen: Basically, I would like to get this merged so I can start implementing new diagnostics on top of it.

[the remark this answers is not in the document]
DougGregor: Oh yes, let's merge and we can iterate in the tree.

allevato:
Are there plans to make the `SwiftDiagnostics` module its own lightweight standalone package/repository? Fundamentally, it's not really related to parsing. As the compiler is factored out into different libraries, it would be great to have a single canonical implementation of diagnostics that has all the expected support for different kinds of consumers/printers, serialization, and so forth, and everyone (including tools built on top of these libraries, like swift-format) could get those capabilities for free.

There's also the one in swift-tools-support-core, but that package is something of a grab bag that brings in a lot of other things that aren't needed for just diagnostics.

DougGregor:
> Are there plans to make the `SwiftDiagnostics` module its own lightweight standalone package/repository? Fundamentally, it's not really related to parsing.

We do need to figure out the layering strategy here. This `SwiftDiagnostics` is a bit tied to `Syntax` nodes, so it cannot sink below `SwiftSyntax` unless we abstract that away. 

> There's also the one in swift-tools-support-core, but that package is something of a grab bag that brings in a lot of other things that aren't needed for just diagnostics.

Right, and I think we'd like to avoid a dependency on swift-tools-support-core here, because it also brings in a dependency on Foundation, which complicates the compiler build.

I suspect this does mean we need a common layer underlying these tools, which knows how to render diagnostics with varying levels of source information (file/line/column, syntax nodes, etc.), emission to the Clang diagnostic file format, etc. I'd like to understand what else might live at that layer, or whether diagnostics is enough to form its own package.

Sources/SwiftDiagnostics/FixIt.swift
```diff
@@ -0,0 +1,46 @@
...
+// Copyright (c) 2014 - 2022 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
```
bnbarham, line 14: IMO we want a `Note` (but not with that name because it's the same as severity). Notes can have messages and a possibly-empty list of changes. Diagnostics can have any numbers of notes. Ie. it's just additional information on a diagnostic.

Sources/SwiftDiagnostics/FixIt.swift
```diff
@@ -0,0 +1,46 @@
...
+/// Types conforming to this protocol represent Fix-It messages that can be
+/// shown to the client.
+/// The messages should describe the change that the Fix-It will perform
+public protocol FixItMessage {
+  /// The diagnostic message that should be displayed in the client.
+  var message: String { get }
+
+  /// See ``MessageID``.
+  var fixItID: MessageID { get }
```
bnbarham, line 23: Do these (or notes in my scheme) need IDs? Seems like clients would only ever care about the ID of the diagnostic itself.

Sources/SwiftDiagnostics/FixIt.swift
```diff
@@ -0,0 +1,46 @@
...
+}
+
+
+/// A Fix-It that can be applied to resolve a diagnostic.
+public struct FixIt {
+  public enum Change {
+    /// Replace `oldNode` by `newNode`.
+    case replace(oldNode: Syntax, newNode: Syntax)
```
bnbarham, line 31: I suspect we'll end up needing more than this but no harm in starting with this.

Sources/SwiftParser/Diagnostics/ParserDiagnosticMessages.swift
```diff
@@ -35,13 +38,68 @@ extension Syntax {
...
+
+  public var diagnosticID: MessageID {
+    MessageID(domain: diagnosticDomain, id: "\(type(of: self)).\(self)")
+  }
+
+  public var severity: DiagnosticSeverity { .error }
+}
+
+public enum StaticParserFixIt: String, FixItMessage {
+  case moveThrowBeforeArrow = "Move 'throws' in before of '->'"
```
bnbarham, line 91:
```suggestion
  case moveThrowBeforeArrow = "Move 'throws' before '->'"
```

Tests/SwiftParserTest/DiagnosticTests.swift
```diff
@@ -65,6 +70,11 @@ public class DiagnosticTests: XCTestCase {
       Syntax(raw: $0.parseFunctionSignature().raw).as(FunctionSignatureSyntax.self)!
     }
 
-    XCTAssertSingleDiagnostic(in: signature, line: 1, column: 7, message: "'throws' may only occur before '->'")
+    XCTAssertSingleDiagnostic(
+      in: signature,
+      line: 1, column: 7,
```
bnbarham, line 75: It'd be nice to add a marker equivalent to the structure tests. Maybe not as important here, I just highly dislike referencing line/col in tests in general 😅

Sources/SwiftDiagnostics/FixIt.swift
ahoppen: I agree. Let’s add them when the need comes up.

Sources/SwiftDiagnostics/FixIt.swift
ahoppen: I think it might be useful if clients want to hide certain fix-its or otherwise transform them. Adding IDs to things never hurts IMO.

Sources/SwiftDiagnostics/FixIt.swift
ahoppen: Absolutely, that’s why I made it an enum. I’ll add more cases as uses come up.

PR #625, opened 2022-08-23
Fix parser issues found by asserting that self-parsing should not produce diagnostics

ahoppen:
Flip the switch on `AssertParse` to check that no diagnostics are produced by default. Also check that no diagnostics are produced during the self-parse.

Making this change required two changes to:
- Fix an issue in attribute skipping, where we previously weren’t consuming `)` while skipping attributes
- Split a line in `swift-parser-test.swift` to disambiguate it as not being parsed as a regex literal (rdar://98736315)

In a follow-up PR, I will change the `AssertParse` statements that now `allowErrors` to check that the correct diagnostics are produced.

rdar://98575270


ahoppen:
@swift-ci Please test

CodaFi (approved):

PR #626, opened 2022-08-23
Improve bug filing workflow

ahoppen:
- Migrate issue template to issue forms: I think the interaction behavior of these forms is nicer than the pre-filled markdown that we currently provide in the template.
- Allow blank/freestyle issues: For frequent contributors (like myself), it’s sometimes useful to file bug reports without being tied to the issue template I created above
- Add an issue template for parser bugs: This is the main change.


ahoppen:
@swift-ci Please test

kimdv (approved):

DougGregor (approved):

PR #628, opened 2022-08-23
[Parser] Simplify sequence expression parsing 

rintaro:
(Depends on https://github.com/apple/swift/pull/60731)

`SequenceExprSyntax` now gurantees that it has odd number elements using `UnresolvedTernaryExpr`/`UnresolvedAsExpr`/`UnresolvedIsExpr`.

Factored out the operator parsing into `parseSequenceExpressionOperator()` which returns the operator expresssion, and optionally the RHS, so we can avoid nested loops.
Also, `parseSequenceExprssion()` now has a performance optimization that avoids heap `Array` construction unless the expression is actually sequenced.


Sources/SwiftParser/Expressions.swift
```diff
@@ -42,23 +42,6 @@ extension Parser {
 }
 
 extension Parser {
-  struct PendingTernary {
```
DougGregor, line 45: I'm so glad to see this struct go away. The parser is much simpler now!

Sources/SwiftParser/Expressions.swift
```diff
@@ -69,187 +52,190 @@ extension Parser {
...
+  ///
+  /// Grammar
+  /// =======
+  ///
+  ///     assignment-operator → '='
+  ///     conditional-operator → '?' expression ':'
+  ///     type-casting-operator → 'is' type
+  ///     type-casting-operator → 'as' type
+  ///     type-casting-operator → 'as' '?' type
+  ///     type-casting-operator → 'as' '!' type
```
DougGregor, line 138: Do we need to document the arrow operator here?

DougGregor (approved):
This is fantastic, and a huge help for the operator-precedence library in https://github.com/apple/swift-syntax/pull/619.

rintaro:
https://github.com/apple/swift/pull/60731
@swift-ci Please test

PR #630, opened 2022-08-23
Get rid of `XXXSyntax.Cursor` type

rintaro:
They are internal types which served as layout 'subscript' keys e.g. `layout[Cursor.propertyName]`. That looks nice, but in reality, the name is usually written very close, like:

```
  var propertyName: PropertyType {
    get { layout[Cursor.propertyName] }
  }
```

So it didn't give us much benefit. In the contrary, it hid the actual "index" information which might be useful when investigating issues. So just get rid of them for now.

Since they are just internal types, there's no external API changes.

### DIFF
--- a/Sources/SwiftSyntax/Raw/RawSyntaxLayoutView.swift
+++ b/Sources/SwiftSyntax/Raw/RawSyntaxLayoutView.swift
@@ -149,14 +149,6 @@ struct RawSyntaxLayoutView {
     }
   }
 
-  /// Returns the child at the provided cursor in the layout.
-  /// - Parameter index: The index of the child you're accessing.
-  /// - Returns: The child at the provided index.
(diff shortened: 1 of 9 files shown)


rintaro:
@swift-ci Please test

ahoppen (approved):

rintaro:
Maybe it's convenient for getting a property name from an index? Like `String(describing: FunctionDeclSyntax.Cursor(rawValue: 1)!)`.
I guess we'd just do 
```swift
extension FunctionDeclSyntax {
  static func propertyName(at i: Int) -> String {
    switch i {
    case 0: return "unexpectedBeforeAttributes"
    case 1: return "attributes"
    ...
    }
  }
}
```
or similar when we want.

PR #633, opened 2022-08-24
Improvements to test case reduction using `swift-parser-test`

ahoppen:
- Increase reduce timeout in swift-stress-tester to 2 seconds
  - I’ve had a few cases where parsing didn’t finish in 1 second but 2 were sufficient (in debug build of swift-syntax-test). Increase the timeout so that reduction works properly
- Stop reduce subcommand if it’s not making progress
  - If the chunk sizes are really tiny compared to the reduced source file, it looks like we aren't making any progress reducing. Just abort the reduction in that case.

### DIFF
--- a/Sources/swift-parser-test/swift-parser-test.swift
+++ b/Sources/swift-parser-test/swift-parser-test.swift
@@ -231,7 +231,7 @@ class Reduce: ParsableCommand {
       }
 
       try process.run()
-      if sema.wait(timeout: DispatchTime.now() + .seconds(1)) == .timedOut {
+      if sema.wait(timeout: DispatchTime.now() + .seconds(2)) == .timedOut {
 #if os(Windows)
         _ = TerminateProcess(process.processHandle, 0)
 #else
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

bnbarham (approved):
Nit on:
> Increase reduce timeout in swift-stress-tester to 2 seconds
Ie. both "increase" and "reduce" :)

What made you add the chunk size abort? Were you noticing the reduction taking a long time on large files but not actually getting anywhere?

ahoppen:
> Nit on:
> 
> > Increase reduce timeout in swift-stress-tester to 2 seconds
> > Ie. both "increase" and "reduce" :)

This was meant as Increase “reduce timeout”, i.e. increase the timeout for the reduce subcommand. I’ll just merge it as is though, not sure if updating the commit message is worth having to run CI again.

> What made you add the chunk size abort? Were you noticing the reduction taking a long time on large files but not actually getting anywhere?

This was because of files that failed to parse in the 2 second timeout. If we reduced them too much, they started parsing in under one second.

PR #639, opened 2022-08-25
Skip leading trivia generation if empty in BuildableNodes

fwcd:
Although we [discussed this earlier](https://github.com/apple/swift-syntax/pull/475#discussion_r907132139), I would like to revisit the optimization to skip leading trivia generation on buildable nodes in `SwiftSyntaxBuilder` when empty, since it actually does have a noticeable performance impact on `generate-swift-syntax-builder`: The optimization cuts down generation time from ~13 seconds to ~7 seconds on my machine, nearly a 50% improvement in performance! Profiling the executable with Instruments reveals that this indeed seems to be the issue, the heaviest stack trace points to `withLeadingTrivia`.

The improvement in performance with this patch becomes especially noticeable during development of generated files where shorter feedback cycles greatly improve the workflow.

(This change should be mostly orthogonal to the upcoming formatting-related changes)

### DIFF
--- a/Sources/generate-swift-syntax-builder/Templates/BuildableNodesFile.swift
+++ b/Sources/generate-swift-syntax-builder/Templates/BuildableNodesFile.swift
@@ -277,12 +277,18 @@ private func createBuildFunction(node: Node) -> FunctionDecl {
         }
       }
     )
-    ReturnStmt(expression: FunctionCallExpr(MemberAccessExpr(base: "result", name: "withLeadingTrivia")) {
-      TupleExprElement(expression: FunctionCallExpr(MemberAccessExpr(
-        base: "combinedLeadingTrivia",
-        name: "addingSpacingAfterNewlinesIfNeeded"
-      )))
-    })
+    IfStmt(
+      conditions: ExprList([MemberAccessExpr(base: "combinedLeadingTrivia", name: "isEmpty")])
+    ) {
+      ReturnStmt(expression: "result")
+    } elseBody: {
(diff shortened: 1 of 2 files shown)


Sources/generate-swift-syntax-builder/Templates/BuildableNodesFile.swift
```diff
@@ -277,12 +277,18 @@ private func createBuildFunction(node: Node) -> FunctionDecl {
...
+    ) {
+      ReturnStmt(expression: "result")
+    } elseBody: {
+      ReturnStmt(expression: FunctionCallExpr(MemberAccessExpr(base: "result", name: "withLeadingTrivia")) {
+        TupleExprElement(expression: FunctionCallExpr(MemberAccessExpr(
+          base: "combinedLeadingTrivia",
+          name: "addingSpacingAfterNewlinesIfNeeded"
+        )))
+      })
+    }
```
fwcd, line 291: One idea might be to move this logic into `withLeadingTrivia` itself (and similarly into `withTrailingTrivia`). However since we cannot assume that the syntax doesn't already have trivia (like in this case) the check would likely get more expensive, so I am not sure whether the more general solution would be worth it there.

fwcd:
@swift-ci please test

ahoppen (approved):
Wow, that performance difference is way bigger than I expected. Thanks for catching this 👍

fwcd:
@swift-ci please test macOS

PR #641, opened 2022-08-26
Unify all parser tests to use an enhanced version `AssertParse`

ahoppen:
This is an assortment of multiple test-only changes, migrating all tests to an enhanced version of `AssertParse`.

- Add an environment variable `SKIP_SELF_PARSE` to skip the self parsing test for local development
  - Not running `testSelfParse` reduces the test time from ~30 seconds to ~2 seconds and it’s really useful for local development to iterate faster.
- Add the following capabilities to `AssertParse`:
  - From `XCTAssertSingleDiagnostics`:
    - Check which diagnostics are produced
    - Check that applying the Fix-Its from the diagnostics produces an expected fixed source text
  - From `XCTAssertHasSubstructure`:
    - Check that the parsed syntax tree contains a substructure
- Remove `XCTAssert[Single]Diagnostics` and `XCTAssertHasSubstructure` and migrate those tests to the now enhanced `AssertParse`
- Allow adding markers to the source code of the form `#^MARKER^#` and assert that diagnostics are produced at the locations of these markers
- Move tests from RecoveryTests.swift and DiagnosticTests.swift to the other test files because the tests are now no longer fundamentally different than the parsing tests in these files
- Change `AssertParse` to take the source text as a proper parameter instead of a closure. I found the way of passing the source text via a closure rather confusing.

Asserting for the diagnostics showed that we weren’t producing diagnostics in all expected cases and that some of the produced diagnostics are bad. These cases should be fixed in follow-up PRs but it won’t be the job of this PR because it’s quite large already.


ahoppen:
@swift-ci Please test

CodaFi:
> Change AssertParse to take the source text as a proper parameter instead of a closure. I found the way of passing the source text via a closure rather confusing.

I designed it that way so that we could write computed text. But I suppose if we wanted to recover that, a local function would suffice.

CodaFi (approved):

Tests/SwiftParserTest/Attributes.swift
```diff
@@ -0,0 +1,21 @@
...
+      """
+      @_dynamicReplacement(#^DIAG_1^#
+      func #^DIAG_2^#test_dynamic_replacement_for2() {
+      }
+      """,
+      diagnostics: [
+        // FIXME: We should be complaining about the missing ')' for the attribute
+        DiagnosticSpec(locationMarker: "DIAG_1", message: "Expected 'for'"),
+        DiagnosticSpec(locationMarker: "DIAG_1", message: "Expected ':'"),
+        DiagnosticSpec(locationMarker: "DIAG_2", message: "Expected ')'"),
```
bnbarham, line 17: Obviously not for this PR and even somewhat unrelated to these diagnostics - but should we limit the expected diagnostic to one per location? Not sure it's entirely useful to have both `for` and `:` here.

Tests/SwiftParserTest/Declarations.swift
```diff
@@ -375,47 +309,370 @@ final class DeclarationTests: XCTestCase {
...
+
+  func testMissingSubscriptReturnClause() {
+    AssertParse(
+      """
+      struct Foo {
+        subscript(x: String) #^DIAG^#{}
+      }
+      """,
+      diagnostics: [
+        DiagnosticSpec(message: "Expected '->'")
```
bnbarham, line 453: Probably worth specializing this one 🤔

Tests/SwiftParserTest/Declarations.swift
```diff
@@ -375,47 +309,370 @@ final class DeclarationTests: XCTestCase {
...
+
+  func testClassWithLeadingNumber() {
+    AssertParse(
+      """
+      class #^DIAG^#23class {
+        // expected-error@-1 {{class name can only start with a letter or underscore, not a number}}
+        // expected-error@-2 {{'c' is not a valid digit in integer literal}}
+        func 24method() {}
+        // expected-error@-1 {{function name can only start with a letter or underscore, not a number}}
+        // expected-error@-2 {{'m' is not a valid digit in integer literal}}
```
bnbarham, line 466: Were these left in to help with the FIXME?

bnbarham (approved):

Sources/_SwiftSyntaxTestSupport/Syntax+Assertions.swift
```diff
@@ -174,6 +110,9 @@ public struct SubtreeMatcher {
 
   /// Same as `XCTAssertSameStructure`, but uses the subtree found from parsing
```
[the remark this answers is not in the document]
ahoppen: Oh, I already added the replacement comment below, I just forgot to remove the original.

Tests/SwiftParserTest/Assertions.swift
```diff
@@ -49,33 +51,165 @@ func AssertEqualTokens(_ actual: [Lexer.Lexeme], _ expected: [Lexer.Lexeme], fil
...
+  init(locationMarker: String = "DIAG", id: MessageID? = nil, message: String?, highlight: String? = nil, fixIts: [String]? = nil) {
+    self.locationMarker = locationMarker
+    self.id = id
+    self.message = message
+    self.highlight = highlight
+    self.fixIts = fixIts
+  }
+}
+
+class FixItApplier: SyntaxRewriter {
```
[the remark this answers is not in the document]
ahoppen: Yes, there are two reasons:
- It’s still incomplete. For example, there’s no way to specify which Fix-It should be applied if a diagnostic offers multiple. I’d like to flesh it out in tests first. If we’re happy with the result at some point, we can consider making it public. 
- The Fixed-Up view of the syntax tree should provide this functionality to clients.

Tests/SwiftParserTest/Assertions.swift
```diff
@@ -49,33 +51,165 @@ func AssertEqualTokens(_ actual: [Lexer.Lexeme], _ expected: [Lexer.Lexeme], fil
...
+
+/// Parse `markedSource` and perform the following assertions:
+///  - The parsed syntax tree should be printable back to the original source code (round-tripping)
+///  - Parsing produced the given `diagnostics` (`diagnostics = []` asserts that the parse was successful)
+///  - If `fixedSource` is not `nil`, assert that applying all fixes from the diagnostics produces `fixedSource`
+/// The source file can be marked with markers of the form `#^DIAG^#` to mark source locations that can be referred to by `diagnostics`.
+/// These markers are removed before parsing the source file.
+/// By default, `DiagnosticSpec` asserts that the diagnostics is produced at a location marked by `#^DIAG^#`.
+/// `parseSyntax` can be used to adjust the production that should be used as the entry point to parse the source code.
+/// If `substructure` is not `nil`, asserts that the parsed syntax tree contains this substructure.
```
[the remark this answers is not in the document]
ahoppen: No, there is not. I wanted to add it once we need it so that we actually have a test case that tests it.

Tests/SwiftParserTest/Declarations.swift
ahoppen: I’ll add a FIXME.

Tests/SwiftParserTest/Declarations.swift
ahoppen: Yes, exactly.

Tests/SwiftParserTest/Attributes.swift
ahoppen: I’m not sure it that’s easy to enforce. We only resolve source locations from syntax nodes when printing the diagnostics and before that, these diagnostics are anchored to different nodes (the missing `for` token and the missing `:` token). So the only place I think we could realistically enforce this would be `AssertParse`.

ahoppen:
@swift-ci Please test

PR #643, opened 2022-08-26
Add a new visitor: `SyntaxTransformVisitor`.

zoecarver:
This visitor returns a transformed Syntax node rather than exposing a pre-post visitor. This will be helpful for ASTGen.


Sources/SwiftSyntax/SyntaxVisitor.swift.gyb
```diff
@@ -141,3 +141,75 @@ open class SyntaxVisitor {
     }
   }
 }
+
+public protocol SyntaxTransformVisitor {
+  associatedtype ResultType = Void
```
zoecarver, line 146: @CodaFi we had discussed having one of these for each kind of thing we might want to transform (decls, exprs, ec.), but I think one is OK for now. If the need arises to have different ones, we can update the visitor.

Sources/SwiftSyntax/SyntaxVisitor.swift.gyb
```diff
@@ -141,3 +141,75 @@ open class SyntaxVisitor {
...
+% for node in SYNTAX_NODES:
+%   if is_visitable(node):
+  /// Visiting `${node.name}` specifically.
+  ///   - Parameter node: the node we are visiting.
+  ///   - Returns: nil by default.
+  public func visit(_ node: ${node.name}) -> ResultType? {
+%   if node.is_base():
+      // Avoid calling into visitChildren if possible.
+      if !node.raw.layoutView!.children.isEmpty {
+        return visitChildren(node).first
```
zoecarver, line 174: Oops, fixing this.

Sources/SwiftSyntax/SyntaxVisitor.swift.gyb
```diff
@@ -141,3 +141,75 @@ open class SyntaxVisitor {
...
+    case .${node.swift_syntax_kind}:
+      let node = ${node.name}(data)
+      return visit(node)
+  % end
+    }
+  }
+
+  public func visitChildren<SyntaxType: SyntaxProtocol>(_ node: SyntaxType) -> [ResultType] {
+    let syntaxNode = Syntax(node)
+        return NonNilRawSyntaxChildren(syntaxNode, viewMode: .sourceAccurate).compactMap { childRaw in
```
zoecarver, line 210: Should `viewMode` be an input?

Sources/SwiftSyntax/SyntaxData.swift
```diff
@@ -213,7 +213,7 @@ class SyntaxBox: CustomStringConvertible,
 ///
 /// SyntaxData is an implementation detail, and should not be exposed to clients
 /// of SwiftSyntax.
-struct SyntaxData {
+public struct SyntaxData {
```
CodaFi, line 216: Let's not expose this. Take is as `@_spi(RawSyntax)` if you need this.

Sources/SwiftSyntax/SyntaxVisitor.swift.gyb
```diff
@@ -141,3 +141,75 @@ open class SyntaxVisitor {
...
+      if !node.raw.layoutView!.children.isEmpty {
+        return visitChildren(node)
+      }
+      return []
+%   end
+  }
+%   end
+% end
+
+  public func visit(_ data: SyntaxData) -> [ResultType] {
```
CodaFi, line 188: We don't want clients visiting SyntaxData directly.

ahoppen: I agree, `SyntaxTransformVisitor` should be a `class` and this should be `private` like we do in all the other initializers.

Sources/SwiftSyntax/SyntaxVisitor.swift.gyb
```diff
@@ -141,3 +141,75 @@ open class SyntaxVisitor {
...
+    // ~50KB per call to this function. rdar://55929175
+  % for node in NON_BASE_SYNTAX_NODES:
+    case .${node.swift_syntax_kind}:
+      let node = ${node.name}(data)
+      return visit(node)
+  % end
+    }
+  }
+
+  public func visitChildren<SyntaxType: SyntaxProtocol>(_ node: SyntaxType) -> [ResultType] {
```
ahoppen, line 208: Similarly to `visit(_ data: SyntaxData)`, this should be `private`.

Sources/SwiftSyntax/SyntaxVisitor.swift.gyb
```diff
@@ -141,3 +141,75 @@ open class SyntaxVisitor {
     }
   }
 }
+
+public protocol SyntaxTransformVisitor {
```
ahoppen, line 145: Let’s put this in a different file. `SyntaxRewriter` also has its dedicated file.

ahoppen (commented):
I’m a little worried about this implementation because IIUC its default implementation discards any recursive transformations by default. My suggestion would be to have a `visitAny` method (similar to `SyntaxAnyVisitor`) that all the `visit` methods default to and have that `visitAny` method `fatalError` by default. This way you’ll trap at runtime rather than just dropping results.

Also, I’m not a huge fan of tying this to `Array` as the transformation result. What do you think about having two generic types: `NodeTransformResult` and `TransformResult` (feel free to come up with a better name) and then combining them using a `func reduce(initialResult: TransformResult?, nextPartialResult: NodeTransformResult)`. That way you could also return strings `testStringify` instead of arrays of strings.

ahoppen:
rdar://99492982

Sources/SwiftSyntax/SyntaxVisitor.swift.gyb
zoecarver: Why do we want this to be a class? I think this should just take a `Syntax`.

Sources/SwiftSyntax/SyntaxVisitor.swift.gyb
zoecarver: Why? It is useful to API consumers (see the FuncCounter test).

zoecarver:
> I’m a little worried about this implementation because IIUC its default implementation discards any recursive transformations by default. My suggestion would be to have a visitAny method (similar to SyntaxAnyVisitor) that all the visit methods default to and have that visitAny method fatalError by default. This way you’ll trap at runtime rather than just dropping results.

I'm not sure what you mean by this. The default implementation will visit all the children and flatMap them together. Nothing should be lost...

> Also, I’m not a huge fan of tying this to Array as the transformation result. What do you think about having two generic types: NodeTransformResult and TransformResult (feel free to come up with a better name) and then combining them using a func reduce(initialResult: TransformResult?, nextPartialResult: NodeTransformResult). That way you could also return strings testStringify instead of arrays of strings.

As long as there can be N children, I think this implementation will be hard. For a generic implementation, the reducer will either have to drop something on the floor or be non-trivial. I like that with the array approach we can handle any input and pass it up to the client for processing/validation. 

If performance is a concern, maybe we can introduce a SingleTransformVisitor that returns a `T?` instead of `[T]` and asserts that all nodes have one or no children. Otherwise, maybe we can work on your suggested implementation offline or commit it as a follow-up change?

zoecarver:
@swift-ci please test

zoecarver:
@swift-ci test

zoecarver:
@swift-ci test

zoecarver:
The `visitAny` method was a great idea. The stringify test did not change, but the FuncCounter got a lot simpler :) Thanks, @ahoppen!

Sources/SwiftSyntax/SyntaxTransform.swift.gyb
```diff
@@ -0,0 +1,122 @@
...
+  public func visit(_ token: TokenSyntax) -> ResultType { visitAny(Syntax(token)) }
+  public func visit(_ node: UnknownSyntax) -> ResultType { visitAny(Syntax(node)) }
+
+% for node in SYNTAX_NODES:
+%   if is_visitable(node):
+  /// Visiting `${node.name}` specifically.
+  ///   - Parameter node: the node we are visiting.
+  ///   - Returns: nil by default.
+  public func visit(_ node: ${node.name}) -> ResultType {
+      visitAny(Syntax(node))
```
ahoppen, line 48: Nitpick
```suggestion
    visitAny(Syntax(node))
```

zoecarver:
@swift-ci please test

Sources/SwiftSyntax/SyntaxTransform.swift.gyb
```diff
@@ -0,0 +1,122 @@
...
+  }
+%   end
+% end
+
+  public func visit(_ data: Syntax) -> ResultType {
```
ahoppen, line 53: Nit: Now that this is no longer `SyntaxData`, the parameter should be named `node`.

Sources/SwiftSyntax/SyntaxTransform.swift.gyb
```diff
@@ -0,0 +1,122 @@
...
+    switch data.raw.kind {
```
ahoppen, line 54: You should be able to use `as(SyntaxEnum.self)` here. That would eliminate the need for force unwraps.

Something like
```swift
    switch node.as(SyntaxEnum.self) {
    case .token(let node):
      return visit(node)
    case .unknown(let node):
      return visit(node)
% for node in NON_BASE_SYNTAX_NODES:
    case .${node.swift_syntax_kind}(let node):
      return visit(node)
% end
```

Sources/SwiftSyntax/SyntaxTransform.swift.gyb
```diff
@@ -0,0 +1,122 @@
...
+  %   if node.base_kind == "Expr":
+    case .${node.swift_syntax_kind}:
+      let node = data.as(${node.name}.self)!
+      return visit(node)
+  %   end
+  % end
+    default:
+      fatalError("Not expression?")
+    }
+  }
```
ahoppen, line 85: Why do we need this method? I don’t think we should ever end up calling `visit` with a base node type. Same for the methods below.

Tests/SwiftParserTest/SyntaxTransformVisitor.swift
```diff
@@ -0,0 +1,112 @@
...
+      
+      public func visitType(_ paramType: TypeSyntaxProtocol) -> String {
+        if let id = SimpleTypeIdentifierSyntax(Syntax(fromProtocol: paramType)) {
+          return visit(id)
+        }
+        if let arr = ArrayTypeSyntax(Syntax(fromProtocol: paramType)) {
+          return visit(arr)
+        }
+        return "<Unknown>"
+      }
```
ahoppen, line 55: Do we even need this. I think `visit(Syntax)` should directly dispatch to `visit(SimpleTypeIdentifierSyntax)` or `visit(ArrayTypeSyntax)`.

Tests/SwiftParserTest/SyntaxTransformVisitor.swift
```diff
@@ -0,0 +1,112 @@
...
+      
+      public func visit(_ node: SimpleTypeIdentifierSyntax) -> String {
+        node.name.text
+      }
+
+      public func visit(_ node: ArrayTypeSyntax) -> String {
+          "[" + visitType(node.elementType) + "]"
+      }
+    }
+    XCTAssertNoThrow(try {
```
ahoppen, line 87: You don’t need `XCTAssertNoThrow` if you make the test case throwing. IMHO it just reads better without XCTAssertNoThrow.

Tests/SwiftParserTest/SyntaxTransformVisitor.swift
```diff
@@ -0,0 +1,112 @@
...
+              }
+            }
+          }
+          """)
+      let count = FuncCounter().visit(parsed)
+      XCTAssertEqual(count, 4)
+    }())
+  }
+  
+  public func testStringify() {
```
ahoppen, line 41: `testStringify` is a somewhat misleading name for this test IMO. What do you think about `testPrintFunctionType`

Tests/SwiftParserTest/SyntaxTransformVisitor.swift
```diff
@@ -0,0 +1,112 @@
...
+    struct Stringify: SyntaxTransformVisitor {
```
ahoppen, line 42: Similarly to the test name, what do you think about renaming this:

```suggestion
    /// Assuming there is a single functioned declaration in the source file, print the type of that function.
    struct PrintFunctionType: SyntaxTransformVisitor {
```

zoecarver:
@swift-ci please test

zoecarver:
@swift-ci please test

PR #644, opened 2022-08-27
Add attributes to VariableDecl convenience initializers

fwcd:
A small patch to make attributed declarations such as `@Test var x: Int` easier to generate (this could also be interesting for users who want to generate uses of property wrappers).


fwcd:
@swift-ci please test macOS

kimdv (approved):

fwcd:
@swift-ci please test macOS

PR #645, opened 2022-08-27
Factor out all `leadingTrivia` logic into `Format`

fwcd:
This PR moves the entire `leadingTrivia` logic into `Format`, thereby allowing for a cleaner separation of concerns: The `build...` methods in `BuildableNodes`/`BuildableCollectionNodes` only generate the syntax, the `format._format` methods then deal with indentation.

In particular, the `leadingTrivia` parameter that was passed alongside `format: Format` is no longer needed. Since many test cases, however, rely on this ad-hoc leading trivia generation, I have migrated as many tests as possible to the `leadingTrivia`-as-field. For smaller nodes, such as literals, which are heavily sugared through their respective convenience initializers, this would become too verbose, therefore I have removed the unexpected trivia node in those cases. Feel free to discuss this though!

Some notes:

- The use of `@_spi(Testing)` for currently public indentation methods/properties on `Format` will come in a future PR since it depends on #644 and is mostly orthogonal to the changes in this PR.
- The rationale for not passing the leading trivia back to the `build...` methods is that this becomes complicated for children that may or may not require leading newlines and due to some other complications such as not being able to switch over the parent's `SyntaxEnum` before constructing its syntax (using `SyntaxKind` would have been an option, but would have required us to use the low-level SPI). In general, I think the overloaded `_format` methods are the more elegant solution here, also since we don't add a lot of unneeded empty trivias, but I would be curious to hear your thoughts on this!
- Performance is pretty much exactly as before, I measure 7-8 seconds of generation time both on this branch and `main` on debug builds of `generate-swift-syntax-builder`.
- The most interesting diff in this PR is probably [Sources/SwiftSyntaxBuilder/generated/Format.swift](https://github.com/apple/swift-syntax/pull/645/files#diff-3d3182fcf9c55cb4f979b4f719269afab620b92591544215460135c81b1f5d63).


fwcd:
@swift-ci please test

fwcd:
@swift-ci please test macOS

Tests/SwiftSyntaxBuilderTest/ImportTests.swift
```diff
@@ -7,9 +7,12 @@ final class ImportTests: XCTestCase {
     let leadingTrivia = Trivia.unexpectedText("␣")
     let identifier = TokenSyntax.identifier("SwiftSyntax")
 
-    let importDecl = ImportDecl(path: AccessPath([AccessPathComponent(name: identifier)]))
+    let importDecl = ImportDecl(
+      leadingTrivia: leadingTrivia,
```
ahoppen, line 11: Actually, I don’t know why we always add the `␣` as leading trivia here. I think it’s something that was done in one of the first test cases and then just carried over to all the other tests. I would be totally fine to remove `␣` from all test cases because it’s rather weird anyway. I’m fine to keep it as-is in this PR though.

ahoppen (approved):

PR #646, opened 2022-08-27
Introduce an intentionally-partial CMake-based build system

DougGregor:
Add a CMake-based build system that builds only the main modules
(SwiftSyntax, SwiftDiagnostics, SwiftParser) that would be used in
later stages of a host compiler. It does not run tests or build
auxiliar convenience modules like SwiftSyntaxBuilder that aren't
expected to be used in the compiler itself.


DougGregor:
@swift-ci please test

DougGregor:
That macOS failure makes no sense, because this PR should be a no-op for CI. Trying again...

DougGregor:
@swift-ci please test

PR #647, opened 2022-08-27
Use a dedicated `Token` type in SwiftSyntaxBuilder

ahoppen:
At the moment, tokens were a little bit of an oddity in SwiftSyntaxBuilder: All nodes were represented by dedicated types in SwiftSyntaxBuilder but tokens were represented by the `TokenSyntax` from the SwiftSyntax module. This has lead to a few problems:
- We needed to add extensions on `TokenSyntax` to add convienience properties that are only designed for SwiftSyntaxBuilder
- `TokenSyntax` was not `SyntaxBuildable` and thus couldn’t be used as elements in unexpected node collections

In general, it seems cleaner if we just add a custom `Token` type in SwiftSyntaxBuilder.

Also: Move the files that declare convenience initializers in SwiftSyntaxBuilder to their own directory.

rdar://99030016


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test macOS

Sources/SwiftSyntaxBuilder/gyb_helpers/ExpressibleAsConformances.py
```diff
@@ -38,7 +38,7 @@
     'StmtBuildable': [
         'CodeBlockItem'
     ],
-    'TokenSyntax': [
+    'Token': [
```
fwcd, line 41: Do we  actually use these? I think we specify them manually in `SwiftSyntaxBuilder/Token.swift`, do you think it would be worth generating `Token.swift` just to keep these conformances in sync?

Sources/generate-swift-syntax-builder/Templates/BuildableBaseProtocolsFile.swift
```diff
@@ -34,7 +34,7 @@ let buildableBaseProtocolsFile = SourceFile {
     let listConformances: [String] = isSyntax ? [] : [syntaxType.listBuildableBaseName]
 
     ProtocolDecl(
-      modifiers: [TokenSyntax.public],
+      modifiers: [Token.public],
```
fwcd, line 37: Just a random idea, what would it take to make it possible to write implicit member expressions here, i.e. `[.public]`? Could we declare tokens that are valid modifiers as static constants on `DeclModifier`? (Though that probably wouldn't work through the layer of `ExpressibleAs...` protocols)

Sources/SwiftSyntaxBuilder/Token.swift
```diff
@@ -14,34 +11,65 @@
...
+
   /// Conformance to ExpressibleAsDeclModifier
   public func createDeclModifier() -> DeclModifier {
     return DeclModifier(name: self)
   }
+
   /// Conformance to ExpressibleAsIdentifierExpr
   public func createIdentifierExpr() -> IdentifierExpr {
     return IdentifierExpr(identifier: self)
   }
```
fwcd, line 67: Since we no longer treat `TokenSyntax` specially, could we generate this file (along with these conformances)?

fwcd (commented):
Looks good, a few minor notes/questions below

kimdv (commented):
There is a space in the path name `Convenience Initializers`

Is that intended?

fwcd:
I stumbled on that too, I think in macOS-land it's not an uncommon convention to use spaces in folder names (see e.g. `~/Library/Application Support`), but in the Linux world it's rather uncommon, mostly since it requires an extra escape and introduces a risk of subtle bugs in shell scripts that do not properly quote strings.

I think `ConvenienceInitializers` would look just as good and avoid these problems, wdyt?

Sources/SwiftSyntaxBuilder/Token.swift
ahoppen: Good idea. I generated it.

Sources/SwiftSyntaxBuilder/gyb_helpers/ExpressibleAsConformances.py
ahoppen: Now that I generate Token.swift again, it’s being used now :wink

Sources/generate-swift-syntax-builder/Templates/BuildableBaseProtocolsFile.swift
ahoppen: Yes, I think that’s the problem, we can’t use implicit member lookup on protocols.

ahoppen:
> There is a space in the path name `Convenience Initializers`
> 
> Is that intended?

I didn’t put too much thought in it. But I agree that it’s easier for tooling in general if it doesn’t contain a space, so I changed it to `ConvenienceIntializers`. Thanks.

ahoppen:
@swift-ci Please test

Sources/generate-swift-syntax-builder/Templates/TokenFile.swift
```diff
@@ -0,0 +1,288 @@
...
+            )
+          },
+          output: "Token"
+        )
+      ) {
+        FunctionCallExpr("Token") {
+          TupleExprElement(
+            label: "tokenSyntax",
+            expression: FunctionCallExpr(MemberAccessExpr(base: "tokenSyntax", name: "with\(leadingTrailing)Trivia")) {
+              TupleExprElement(expression: IdentifierExpr("trivia"))
```
fwcd, line 93: I think we can shorten this to

```suggestion
              TupleExprElement(expression: "trivia")
```

Sources/generate-swift-syntax-builder/Templates/TokenFile.swift
```diff
@@ -0,0 +1,288 @@
...
+    FunctionDecl(
+      leadingTrivia: .newline,
+      modifiers: [Token.public],
+      identifier: .identifier("buildToken"),
+      signature: FunctionSignature(
+        input: ParameterClause(),
+        output: "TokenSyntax"
+      )
+    ) {
+      IdentifierExpr("tokenSyntax")
```
fwcd, line 109: Same here (unless you intended to make it explicit, but since we omit it in the other files too, being consistent is probably a good idea).

```suggestion
      "tokenSyntax"
```

Sources/generate-swift-syntax-builder/Templates/TokenFile.swift
```diff
@@ -0,0 +1,288 @@
...
+            firstName: .identifier("leadingTrivia"),
+            colon: .colon,
+            type: OptionalType(wrappedType: "Trivia")
+          )
+        },
+        output: "Syntax"
+      )
+    ) {
+      FunctionCallExpr("Syntax") {
+        TupleExprElement(expression: IdentifierExpr("tokenSyntax"))
```
fwcd, line 133:
```suggestion
        TupleExprElement(expression: "tokenSyntax")
```

Sources/generate-swift-syntax-builder/Templates/TokenFile.swift
```diff
@@ -0,0 +1,288 @@
...
+            colon: .colon,
+            type: "TokenSyntax"
+          )
+        }
+      )
+    ) {
+      SequenceExpr {
+        MemberAccessExpr(base: "self", name: "tokenSyntax")
+        AssignmentExpr()
+        IdentifierExpr("tokenSyntax")
```
fwcd, line 68:
```suggestion
        "tokenSyntax"
```

fwcd (approved):
Some minor stylistic comments below, otherwise this looks very good, thanks! I also like how we could unify `Tokens.swift`, `TokenSyntax.swift` and the new `Token` type into a single file 👍

ahoppen:
@swift-ci Please test

PR #652, opened 2022-08-29
Store presence of a token in an explicit field

ahoppen:
We recently changed a token’s presence to be implicitly determined by whether it contains any source code. It turns out that this has two downsides:
- We cannot synthesize missing tokens with explicitly known text, like a missing `for` identifier token for the `@_dynamicReplacable` attribute (this is the issue that I just stumbled over)
- We cannot synthesize missing tokens with trivia (this has been an issue in the past that we have been able to work around)

### DIFF
--- a/Sources/SwiftSyntax/Raw/RawSyntaxNodeProtocol.swift
+++ b/Sources/SwiftSyntax/Raw/RawSyntaxNodeProtocol.swift
@@ -116,10 +116,16 @@ public struct RawTokenSyntax: RawSyntaxNodeProtocol {
     kind: RawTokenKind,
     wholeText: SyntaxText,
     textRange: Range<SyntaxText.Index>,
+    presence: SourcePresence,
     arena: __shared SyntaxArena
   ) {
     let raw = RawSyntax.parsedToken(
-      kind: kind, wholeText: wholeText, textRange: textRange, arena: arena)
+      kind: kind,
+      wholeText: wholeText,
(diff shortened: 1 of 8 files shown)


ahoppen:
@swift-ci Please test

Sources/SwiftSyntax/Raw/RawSyntaxNodeProtocol.swift
```diff
@@ -148,8 +157,13 @@ public struct RawTokenSyntax: RawSyntaxNodeProtocol {
   /// Creates a missing `TokenSyntax` with the specified kind.
   public init(missing kind: RawTokenKind, arena: __shared SyntaxArena) {
     self.init(
-      kind: kind, text: "", leadingTriviaPieces: [], trailingTriviaPieces: [],
-      arena: arena)
+      kind: kind,
+      text: "",
```
rintaro, line 161: Inconsistent with `RawSyntax.makeMissingToken()`. Should we use `kind.defaultText` here?

ahoppen: I did actually change that in some follow-up PR but it looks like it didn’t make it into this PR. Updated it now.

rintaro (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

PR #653, opened 2022-08-30
Add progress conditions to all loops in the parser

ahoppen:
This makes infinite loop bugs easier to debug.

### DIFF
--- a/Sources/SwiftParser/Expressions.swift
+++ b/Sources/SwiftParser/Expressions.swift
@@ -592,12 +592,13 @@ extension Parser {
           //   #if CONDITION_1
           //     #if CONDITION_2
           //       .someMember
+          var loopProgress = LoopProgressCondition()
           repeat {
             backtrack.eat(.poundIfKeyword)
             while !backtrack.at(.eof) && !backtrack.currentToken.isAtStartOfLine {
               backtrack.skipSingle()
             }
-          } while backtrack.at(.poundIfKeyword)
+          } while loopProgress.evaluate(backtrack.currentToken) && backtrack.at(.poundIfKeyword)
 
           guard backtrack.isAtStartOfPostfixExprSuffix() else {
             break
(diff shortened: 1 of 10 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

CodaFi:
I believe the general pattern here ought to be `self.consume(if: <delimiter>) != nil && loopProgress.evaluate(currentToken)` in case we recover from a bogus element production by consuming the delimiter and continuing. That's good progress.

Sources/SwiftParser/Patterns.swift
```diff
@@ -194,11 +198,12 @@ extension Parser.Lookahead {
...
 
     if !self.at(.rightParen) {
+      var loopProgress = LoopProgressCondition()
       repeat {
         guard self.canParsePattern() else {
           return false
         }
-      } while self.consume(if: .comma) != nil
+      } while loopProgress.evaluate(currentToken) && self.consume(if: .comma) != nil
     }
```
CodaFi, line 207: In lookahead, we should consume the token, then check for progress.

Sources/SwiftParser/Expressions.swift
```diff
@@ -2012,7 +2029,8 @@ extension Parser.Lookahead {
     } else if lookahead.currentToken.isIdentifier || lookahead.at(.wildcardKeyword) {
       // Parse identifier (',' identifier)*
       lookahead.consumeAnyToken()
-      while lookahead.consume(if: .comma) != nil {
+      var parametersProgress = LoopProgressCondition()
+      while parametersProgress.evaluate(lookahead.currentToken) && lookahead.consume(if: .comma) != nil {
```
CodaFi, line 2033: Same here

Sources/SwiftParser/Patterns.swift
ahoppen: I think we would hit the loop progress assertion in either case but if we move the check after `consume` we might be able to abort the loop one iteration earlier. 

For consistency I moved `loopProgress.evaluate` to be the last condition of all loops. On second thought that also reads nicer because it makes the actual condition more prominent.

ahoppen:
@swift-ci Please test

ahoppen:
rdar://99492985

ahoppen:
@swift-ci Please test macOS

CodaFi (approved):

ahoppen:
@swift-ci Please test

PR #655, opened 2022-08-30
Remove `recover` methods from the parser

ahoppen:
Instead of sticking unexpected tokens into `errorTokens` of a `CodeBlock`, we should be putting them into the tree as unexpected nodes, where they can be diagnosed. These unexpected nodes are typically created by `Parser.expect` calls.


ahoppen:
@swift-ci Please test

ahoppen:
https://github.com/apple/swift/pull/60847

@swift-ci Please test

ahoppen:
@swift-ci Please test

CodaFi (approved):
We'll take it for now. There's some tricky cases especially with extraneous right braces that this won't handle correctly.

PR #659, opened 2022-08-30
Move AbsoluteSyntaxInfo to "edge" SyntaxData information

rintaro:
The root node doesn't need to hold its `AbsoluteSyntaxInfo` because it's always zero. Move this information to "edge" data so only non-root nodes hold the "absolute info". Introduce an "edge" info struct, and use it with `indirect` enum case instead of `SyntaxBox`.

Previously the structure of `SyntaxData` was:
 * AbsoluteRawSyntax
   * RawSyntax
   * AbsoluteSyntaxInfo
 * Tagged union of
   * SyntaxArena
   * SyntaxBox
     * Parent SyntaxData

Now:
 * RawSyntax node
 * Tagged union of:
   * SyntaxData.Info.Root
     * SyntaxArena
   * SyntaxData.Info.Edge (indirect case)
     * Parent SyntaxData
     * AbsoluteSyntaxInfo

This reduces the size of `SyntaxData` to 2 words.

Also `SyntaxData.parent` now returns `SyntaxData` instead of `Syntax`. `SyntaxData` should not rely on `Syntax` API at all.

As for the performance, compared to pre #635,  this improves `testEmptyAnyVistorPerformance`/`testEmptyRewriterPerformance`/`testEmptyVisitorPerformance` by 5-10%.

### DIFF
--- a/Sources/SwiftSyntax/SyntaxData.swift
+++ b/Sources/SwiftSyntax/SyntaxData.swift
@@ -184,71 +184,78 @@ struct AbsoluteRawSyntax {
   }
 }
 
-/// Indirect wrapper for a `Syntax` node to avoid cyclic inclusion of the 
-/// `Syntax` struct in `SyntaxData`
-class SyntaxBox: CustomStringConvertible, 
-    CustomDebugStringConvertible, TextOutputStreamable {
-  let value: Syntax
-
-  init(_ value: Syntax) {
-    self.value = value
-  }
-
-  // SyntaxBox should be transparent in all descriptions
-
-  /// A source-accurate description of this node.
-  var description: String {
-    return value.description
-  }
-
-  /// Returns a description used by dump.
-  var debugDescription: String {
-    return value.debugDescription
-  }
-
-  /// Prints the raw value of this node to the provided stream.
-  /// - Parameter stream: The stream to which to print the raw tree.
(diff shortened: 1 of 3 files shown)


rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

Sources/SwiftSyntax/SyntaxData.swift
```diff
@@ -184,71 +184,78 @@ struct AbsoluteRawSyntax {
...
+
+    // For root node.
+    struct Root {
+      var arena: SyntaxArena
+    }
 
-  init(_ value: Syntax) {
-    self.value = value
+    // For non-root nodes.
+    struct Edge {
```
ahoppen, line 202: I don’t like the term `edge` here. To me the edge would be information that’s related to the parent-child relationship between two nodes. But `absoluteInfo` is information about this node in the entire tree.

It’s not a great name but `NonRoot` would make me stumble a lot less.

ahoppen (approved):
I’m not really comfortable with the name `edge`. Otherwise LGTM.

ahoppen:
rdar://99492992

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

PR #660, opened 2022-08-30
Revert "Use RawSyntax address as the 'rootId' of SyntaxIdentifier"

rintaro:
Reverts apple/swift-syntax#635

Revert while fixing race issues with CMake build change.


rintaro:
@swift-ci Please test

PR #661, opened 2022-08-30
Introduce `#_hasSymbol`

tshortli:
Introduces parsing and regenerates sources for the new `if #_hasSymbol()` condition.

https://github.com/apple/swift/pull/60852

Resolves rdar://99342017

### DIFF
--- a/Sources/SwiftParser/Statements.swift
+++ b/Sources/SwiftParser/Statements.swift
@@ -228,6 +228,11 @@ extension Parser {
     if self.at(any: [.poundAvailableKeyword, .poundUnavailableKeyword]) {
       return self.parsePoundAvailableConditionElement()
     }
+    
+    // Parse a #_hasSymbol condition if present.
+    if self.at(.poundHasSymbolKeyword) {
+      return self.parsePoundHasSymbolConditionElement()
+    }
 
     // Parse the basic expression case.  If we have a leading let/var/case
     // keyword or an assignment, then we know this is a binding.
@@ -351,6 +356,30 @@ extension Parser {
     case .macro:
       fatalError("Macros are not allowed in this position!")
     }
+  }
+  
+  /// Parse a `#_hasSymbol` condition.
+  ///
+  /// Grammar
+  /// =======
+  ///
+  ///     has-symbol-condition → '#_hasSymbol' '(' expr ')'
+  @_spi(RawSyntax)
+  public mutating func parsePoundHasSymbolConditionElement() -> RawSyntax {
+    let (unexpectedBeforeKeyword, keyword) = self.expect(.poundHasSymbolKeyword)
+    let (unexpectedBeforeLParen, lparen) = self.expect(.leftParen)
+    let expr = self.parseExpression()
+    let (unexpectedBeforeRParen, rparen) = self.expect(.rightParen)
+    return RawSyntax(RawHasSymbolConditionSyntax(
+      unexpectedBeforeKeyword,
+      hasSymbolKeyword: keyword,
+      unexpectedBeforeLParen,
+      leftParen: lparen,
+      expression: expr,
+      unexpectedBeforeRParen,
+      rightParen: rparen,
+      arena: self.arena)
+    )
   }
 }
 
--- a/Sources/generate-swift-syntax-builder/gyb_syntax_support/StmtNodes.py
+++ b/Sources/generate-swift-syntax-builder/gyb_syntax_support/StmtNodes.py
@@ -193,6 +193,7 @@ STMT_NODES = [
                              kind='MatchingPatternCondition'),
                        Child('OptionalBinding',
                              kind='OptionalBindingCondition'),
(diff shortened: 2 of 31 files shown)


tshortli:
https://github.com/apple/swift/pull/60852

@swift-ci please test

tshortli:
https://github.com/apple/swift/pull/60852

@swift-ci please test

ahoppen:
Could you also add parsing for `#_hasSymbol` to the new parser that lives in Sources/SwiftParser of this repo?

tshortli:
> Could you also add parsing for `#_hasSymbol` to the new parser that lives in Sources/SwiftParser of this repo?

I was afraid I might need to do that. Alright, I'm iterating a bit on the parsing anyways so I'm going to put this on hold until I get that squared away.

tshortli:
@swift-ci please test

tshortli:
@swift-ci please test

tshortli:
@ahoppen I've implemented parsing for `#_hasSymbol` in the parser implementation here now as well.

Sources/SwiftParser/Statements.swift
```diff
@@ -352,6 +357,30 @@ extension Parser {
...
+  /// Parse a `#_hasSymbol` condition.
+  ///
+  /// Grammar
+  /// =======
+  ///
+  ///     has-symbol-condition → '#_hasSymbol' '(' expr ')'
+  @_spi(RawSyntax)
+  public mutating func parsePoundHasSymbolConditionElement() -> RawSyntax {
+    assert(self.at(any: [.poundHasSymbolKeyword]))
+    let keyword = self.consumeAnyToken()
```
ahoppen, line 370: Could you change this to 
```suggestion
    let (unexpectedBeforeKeyword, keyword) = self.expect(.poundHasSymbolKeyword)
```

Should we want to support recovery in front of `#_hasSymbol`, that will allow us to do so more easily by changing
```diff
- if self.at(.poundHasSymbolKeyword) {
+ if self.canRecoverTo(.poundHasSymbolKeyword) != nil {
    return self.parsePoundHasSymbolConditionElement()
  }
```

Not saying that we should support recovery in this case but it’s good practice to prepare for that path.

Sources/generate-swift-syntax-builder/gyb_syntax_support/StmtNodes.py
```diff
@@ -242,6 +243,16 @@
              Child('RightParen', kind='RightParenToken'),
          ]),
 
+    # has-symbol-condition -> '#_hasSymbol' '(' expression ')'
+    Node('HasSymbolCondition', name_for_diagnostics="'#_hasSymbol' condition",
+         kind='Syntax',
+         children=[
+             Child('HasSymbolKeyword', kind='HasSymbolToken'),
+             Child('LeftParen', kind='LeftParenToken'),
+             Child('Expression', kind='Expr'),
```
ahoppen, line 252: I know that it’s also modeled like this in C++ and I don’t know exactly how `#_hasSymbol` works but is expression the right way to model the symbol? Wouldn’t `QualifiedDeclName` be better?

tshortli: I'm very much not sure about the best way to approach this, but my impression was that `QualifiedDeclName` is more limited than I need. I need to be able to unambiguously identify functions, methods, properties accessors, nominal types, Clang functions and global vars, and maybe even protocol conformances; basically any declaration that can be referenced in an expression. It seems like a way to accomplish this is to parse expressions and then accept a subset of them (post type checking) that can be recognized as identifying a single declaration. I recognize that this is a bit of an unusual approach. The advantages I see are that I can leverage existing type checking logic to do the declaration lookup in context. This is especially helpful when it comes to disambiguation since I can use coercion expressions to disambiguate overloaded functions and methods.

What do you think, given the requirements?

Sources/SwiftParser/Statements.swift
tshortli: Sure! I followed the patterns I found in the parsing for `if #available` - should that also be updated?

tshortli:
@swift-ci please test

Sources/generate-swift-syntax-builder/gyb_syntax_support/StmtNodes.py
ahoppen: Ah, I see. I think coercion is the feature that requires you to use an expression here. `QualifiedDeclName` wouldn’t be able to distinguish those AFAIK.

Do you think adding a test case for such a use case that requires expression parsing would make sense?

tshortli:
@swift-ci please test

Sources/generate-swift-syntax-builder/gyb_syntax_support/StmtNodes.py
tshortli: Yup, added one.

PR #662, opened 2022-08-30
Reapply: Use RawSyntax address as the 'rootId' of SyntaxIdentifier

rintaro:
(Reapply #635 without `_CSwiftSyntax` changes.)

---
We don't need to guarantee the uniqueness throughout the process lifetime. but only need to guarantee the uniqueness while the object is alive.


rintaro:
@swift-ci Please test

PR #614, opened 2022-08-19
Move SwiftSyntaxParser related code to its own module

rintaro:
* Remove `_SyntaxParserInterop.swift`
* Use `RawSyntax.toOpaque()`/`RawSyntax.fromOpaque(:_)` to interop
* Move `RawSyntax.createFromCSyntaxNode()` to the module
* Move SyntaxKind/RawTokenKind/RawTriviaPiece creation from C serialization code to the module
* Tweak some RawSyntax API visibility
* Refactor materialized token creation API to efficiently create a token from SyntaxParser modules
* Introduce `Syntax.withUnsafeRawSyntax(_:)` API

PR #664, opened 2022-08-30
Remove  `_CSwiftSyntax` module

rintaro:
Since we stopped using `AtomicCounter` (#662) , `SwiftSyntax` does not depend on `_CSwiftSyntax` anymore.
`CNodes` for `SwiftSyntaxParser` is provided by `_InternalSwiftSyntaxParser` module.


rintaro:
@swift-ci Please test

Sources/SwiftSyntaxParser/RawSyntax+CNodes.swift
```diff
@@ -10,7 +10,6 @@
 //
 //===----------------------------------------------------------------------===//
 
-@_implementationOnly import _CSwiftSyntax
```
rintaro, line 13: This is a line I forgot to remove in #614

PR #666, opened 2022-08-31
Make buildable node fields mutable and fix `withTrailingComma`

fwcd:
`withTrailingComma` currently discards the value for `leadingTrivia`. While adding the parameter would be an option, this PR refactors the function to use the
```swift
var result = self
result.field = value
return result
```
pattern. This also has the advantage that new optional fields cannot accidentally be forgotten.

To achieve this, the fields have to be mutable. Since we are dealing with value types with no further invariants, this should hopefully not be an issue though.


fwcd:
@swift-ci please test

ahoppen (approved):
Sounds good. Thanks for fixing this.

PR #672, opened 2022-08-31
Eliminate parser recovery based on errorTokens in CodeBlock

ahoppen:


ahoppen:
Included in https://github.com/apple/swift-syntax/pull/655

PR #673, opened 2022-08-31
Introduce a C function that performs parser consistency checks.

DougGregor:
Export a C function that can be used to perform various consistency
checks for each source file it processes. This will be used in the C++
compiler to improve testing coverage for the new parser.

### DIFF
--- a/Sources/SwiftParser/CompilerSupport.swift
+++ b/Sources/SwiftParser/CompilerSupport.swift
@@ -37,7 +37,7 @@ public func _parserConsistencyCheck(
       var bufferArray = [UInt8](buffer)
       bufferArray.append(0)
       if "\(sourceFile)" != String(cString: bufferArray) {
-        debugPrint(
+        print(
           "\(String(cString: filename)): error: file failed to round-trip")
         return 1
       }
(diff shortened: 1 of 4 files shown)


DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

Sources/SwiftParser/CompilerSupport.swift
```diff
@@ -0,0 +1,48 @@
...
+    // Parse the source file
+    let sourceFile = parser.parseSourceFile()
+
+    // Round-trip test.
+    if flags & 0x01 != 0 {
+      var bufferArray = [UInt8](buffer)
+      bufferArray.append(0)
+      if "\(sourceFile)" != String(cString: bufferArray) {
+        debugPrint(
+          "\(String(cString: filename)): error: file failed to round-trip")
```
ahoppen, line 41: Any reason why this is `debugPrint` and not `print`?

ahoppen (approved):

Sources/SwiftParser/CompilerSupport.swift
DougGregor: No, that's kinda silly. I'll fix it to use `print`.

DougGregor:
@swift-ci please test

PR #682, opened 2022-09-01
Tags are created with -a suffix

MEnnabah:


MEnnabah:
I just realised `-a` is probably part of the date.

PR #214, opened 2020-04-30
Introduce a `syntaxNodeType` property for all types conforming to `SyntaxProtocol`

akyrtzi:
It returns the underlying syntax node type. It is primarily intended as a debugging aid during development.

PR #683, opened 2022-09-01
Remove `SyntaxProtocol.syntaxNodeType` requirement

rintaro:
Instead, provide it in `extension SyntaxProtocol`.
Also, avoid creating existential or checking dynamic type for it. It can be statically determined by the `SyntaxKind`.


rintaro:
@swift-ci Please test

rintaro:
@akyrtzi You introduced this in https://github.com/apple/swift-syntax/pull/214. Do you have any concern with this change?

akyrtzi:
No, seems good 👍

PR #684, opened 2022-09-01
Eliminate `SyntaxNode`

rintaro:
`SyntaxNode` was introduced as a lighter `Syntax` when existentials were needed for creating `Syntax` instance.

Now that `Syntax`/`SyntaxData` is just a struct that holds one class (`SyntaxBox`). So it's not that heavy.
More importantly `SyntaxNode` was not safe to escape because it didn't retain the `SyntaxArena` of the `AbsoluteRawSyntax`

`SyntaxNode` was only used in `SyntaxCursor` for `IncrementalParseLookup`. Use `SyntaxData` instead.

(Conflicts with #659 . Will rebase after merging it)

(No noticeable performance regression with a test case in rdar://48511326)


rintaro:
@swift-ci Please test

Sources/SwiftSyntax/IncrementalParseTransition.swift
```diff
@@ -348,60 +348,53 @@ public struct IncrementalParseLookup {
...
-    var parents = ArraySlice(self.parents)
+  /// - Returns: `nil` if it run out of nodes to walk to.
+  var nextSibling: SyntaxData? {
     var node = self.node
-    while !parents.isEmpty {
-      if let sibling = node.nextSibling(parent: parents.last!, viewMode: viewMode) {
-        return sibling
+    while let parent = node.parent?.data {
+      var absRaw = node.absoluteRaw
+      if let sibling = absRaw.nextSibling(parent: parent.absoluteRaw, viewMode: viewMode) {
```
rintaro, line 371: Never mutated `var`
```suggestion
      if let sibling = node.absoluteRaw.nextSibling(parent: parent.absoluteRaw, viewMode: viewMode)
```

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

Sources/SwiftSyntax/Syntax.swift
```diff
@@ -766,3 +663,6 @@ extension RawSyntaxNodeProtocol {
     return Syntax(raw: raw).recursiveDescription
   }
 }
+
+@available(*, unavailable, message: "use 'Syntax' instead")
+public struct SyntaxNode {}
```
ahoppen, line 668: Is this useful in any way or just an artifact from your local development?

ahoppen (approved):
A great simplification. Thank you.

Sources/SwiftSyntax/Syntax.swift
rintaro: I left this only because `SyntaxNode` was a public type. Since the initializer was not public, I don't anticipate anyone is using this. But still.

PR #685, opened 2022-09-01
Exclude `gyb_syntax_support` from Swift target

fwcd:
Temporary Python compilation artifacts like `*.pyc` files tend to confuse the SPM build tool (i.e. cause a warning). Therefore this patch excludes `gyb_syntax_support` from `generate-swift-syntax-builder`, like `gyb_helpers`.


DougGregor (approved):

DougGregor:
Thank you!

PR #688, opened 2022-09-01
[Recovery 0a/7] Implement recovery in `eat`

ahoppen:
This is quite a mechanical change without any changed functionality.

---

At the moment this doesn’t change anything because all places that call `eat` have previous checks that make sure the parser is currently positioned at the expected token.

In the future, this will allow us to start parsing functions that `eat` a token at a different token and this function is responsible for eating all intermediate tokens until the expected one.

### DIFF
--- a/Sources/SwiftParser/Modifiers.swift
+++ b/Sources/SwiftParser/Modifiers.swift
@@ -51,11 +51,12 @@ extension Parser {
         let name = self.consumeAnyToken()
         let details: RawDeclModifierDetailSyntax?
         if self.at(.leftParen) {
-          let lparen = self.eat(.leftParen)
+          let (unexpectedBeforeLParen, lparen) = self.eat(.leftParen)
           assert(self.currentToken.isContextualKeyword("set"))
           let detail = self.consumeIdentifier()
           let (unexpectedBeforeRParen, rparen) = self.expect(.rightParen)
           details = RawDeclModifierDetailSyntax(
+            unexpectedBeforeLParen,
             leftParen: lparen,
             detail: detail,
             unexpectedBeforeRParen,
@@ -69,15 +70,19 @@ extension Parser {
         elements.append(RawDeclModifierSyntax(
           name: name, detail: details, arena: self.arena))
       case .staticKeyword:
-        let staticKeyword = self.eat(.staticKeyword)
+        let (unexpectedBeforeStaticKeyword, staticKeyword) = self.eat(.staticKeyword)
         elements.append(RawDeclModifierSyntax(
-          name: staticKeyword, detail: nil, arena: self.arena))
+          unexpectedBeforeStaticKeyword,
+          name: staticKeyword,
+          detail: nil,
+          arena: self.arena
+        ))
       case .classKeyword:
         // If 'class' is a modifier on another decl kind, like var or func,
         // then treat it as a modifier.
         do {
           var lookahead = self.lookahead()
-          lookahead.eat(.classKeyword)
+          lookahead.eatWithoutRecovery(.classKeyword)
           // When followed by an 'override' or CC token inside a class,
           // treat 'class' as a modifier in the case of a following CC
           // token, we cannot be sure there is no intention to override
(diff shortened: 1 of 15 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
rdar://99492995

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

Sources/SwiftParser/Modifiers.swift
```diff
@@ -43,15 +43,56 @@ extension Parser {
...
+        case .fileprivateKeyword: return .fileprivateKeyword
+        case .internalKeyword: return .internalKeyword
+        case .publicKeyword: return .publicKeyword
+        case .staticKeyword: return .staticKeyword
+        case .classKeyword: return .classKeyword
+        case .identifier: return .identifier
+        }
+      }
+
+      func accepts(lexeme: Lexer.Lexeme, parser: Parser) -> Bool {
```
CodaFi, line 67: This should live on an extension of the `Lookahead`.

ahoppen: I assume you are worried because we do lookahead in `accepts`. So was I and I removed that feature in one of my follow-up commits in https://github.com/apple/swift-syntax/pull/718. I cherry-picked that commit to this PR now.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

Sources/SwiftParser/Types.swift
```diff
@@ -790,19 +801,14 @@ extension Parser {
...
-    while (self.at(.inoutKeyword)
-           || self.currentToken.isContextualKeyword("__shared")
-           || self.currentToken.isContextualKeyword("__owned")
-           || self.currentToken.isContextualKeyword("isolated")
-           || self.currentToken.isContextualKeyword("_const")
-    ) && modifiersProgress.evaluate(currentToken) {
-      if self.at(.inoutKeyword) {
-        let inoutKeyword = self.eat(.inoutKeyword)
-        elements.append(RawSyntax(inoutKeyword))
+    while let token = self.consume(ifAny: .inoutKeyword, .identifier, where: { (lexeme, parser) in
```
CodaFi, line 804: This seems pretty messy.

Sources/SwiftParser/Statements.swift
```diff
@@ -642,22 +666,25 @@ extension Parser {
...
-    if self.at(.atSign) {
-      let at = self.eat(.atSign)
+    if let at = self.consume(if: .atSign) {
       let ident = self.consumeIdentifier()
 
       var tokenList = [RawTokenSyntax]()
       var loopProgress = LoopProgressCondition()
-      while self.at(.atSign) && loopProgress.evaluate(currentToken) {
-        tokenList.append(self.eat(.atSign))
+        while let atSign = self.consume(if: .atSign), loopProgress.evaluate(currentToken) {
```
CodaFi, line 674: Nit: Extraneous whitespace

Sources/SwiftParser/Types.swift
CodaFi, line 804: This seems pretty messy.

CodaFi:
I think this has grown into at least two PR: One that goes through and makes the simple eat -> expect transition, another that introduces the rest of the infrastructure here. It's hard for me to evaluate this much change in one place since it's no longer mechanical.

ahoppen:
I rebased this PR and made it only contain the first (mechanical) commit

Sources/SwiftParser/TokenConsumer.swift
```diff
@@ -79,7 +79,7 @@ extension TokenConsumer {
   ///
   /// - Parameter kind: The kind of token to consume.
   /// - Returns: A token of the given kind.
-  public mutating func eat(_ kind: RawTokenKind) -> Token {
+  public mutating func eatWithoutRecovery(_ kind: RawTokenKind) -> Token {
```
CodaFi, line 82: Let's move this utility into Parser and Lookup respectively if can no longer be shared. The new name doesn't read any better to me.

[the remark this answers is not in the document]
ahoppen: This method is going to be removed by https://github.com/apple/swift-syntax/pull/741 again, so it shouldn’t make much difference where it lives.

CodaFi (approved):

Issue #669, opened 2022-08-31
SwiftParser crash when parsing `#"""a`


omochi:
### Description

#"""a

### Steps to Reproduce

```
[omochi@omochi-iMacPro swift-syntax (main %<)]$ cat temp/crash1.swift
#"""a[omochi@omochi-iMacPro swift-syntax (main %<)]$ .build/debug/swift-parser-test verify-round-trip temp/crash1.swift 
SwiftParser/Expressions.swift:923: Assertion failed: existence of open/close delimiter should match
Illegal instruction: 4
```

Its similar to #668. But it emits another error.

I try to fix it.

ahoppen:
rdar://99373064

PR #690, opened 2022-09-02
Generate missing string closer quote and delimiter

### LINKED ISSUE #668: SwiftParser crash when parsing `#"""`
### Issue Kind

Parser Crash

### Source Code

```
#"""
```

### Description

```
[omochi@omochi-iMacPro swift-syntax (main %=)]$ cat temp/crash0.swift 
#"""[omochi@omochi-iMacPro swift-syntax (main %=)]$ .build/debug/swift-parser-test verify-round-trip temp/crash0.swift 
Swift/Collection.swift:711: Fatal error: Out of bounds: index >= endIndex
Illegal instruction: 4
```

I try to fix it

omochi:
This PR fix #668 #669.

If closer quote and delimiter is missing but opener ones are present,
generate missing token to fix round trip bug and emit missing diagnostics.

### DIFF
--- a/Sources/SwiftParser/Expressions.swift
+++ b/Sources/SwiftParser/Expressions.swift
@@ -935,11 +935,8 @@ extension Parser {
 
       if let openDelimiter = openDelimiter {
         return RawTokenSyntax(
-          kind: .rawStringDelimiter,
+          missing: .rawStringDelimiter,
           text: openDelimiter.tokenText,
-          leadingTriviaPieces: [],
-          trailingTriviaPieces: [],
-          presence: .missing,
           arena: arena
         )
       }
--- a/Sources/SwiftParser/Expressions.swift
+++ b/Sources/SwiftParser/Expressions.swift
@@ -925,24 +925,21 @@ extension Parser {
       text = text.dropFirst(closeQuote.tokenText.count)
     }
     /// Parse closing raw string delimiter if exist.
-    let closeDelimiter: RawTokenSyntax? = {
-      if let closeDelimiter = self.parseStringLiteralDelimiter(
-        at: .trailing,
-        text: text
-      ) {
-        return closeDelimiter
-      }
-
-      if let openDelimiter = openDelimiter {
-        return RawTokenSyntax(
-          missing: .rawStringDelimiter,
-          text: openDelimiter.tokenText,
-          arena: arena
-        )
-      }
-
-      return nil
-    }()
+    let closeDelimiter: RawTokenSyntax?
+    if let delimiter = self.parseStringLiteralDelimiter(
+      at: .trailing,
+      text: text
+    ) {
+      closeDelimiter = delimiter
+    } else if let openDelimiter = openDelimiter {
+      closeDelimiter = RawTokenSyntax(
+        missing: .rawStringDelimiter,
+        text: openDelimiter.tokenText,
+        arena: arena
+      )
+    } else {
+      closeDelimiter = nil
+    }
     assert((openDelimiter == nil) == (closeDelimiter == nil),
            "existence of open/close delimiter should match")
     if let closeDelimiter = closeDelimiter, !closeDelimiter.isMissing {
--- a/Sources/SwiftParser/Expressions.swift
+++ b/Sources/SwiftParser/Expressions.swift
@@ -907,7 +907,9 @@ extension Parser {
       at: openDelimiter != nil ? .leadingRaw : .leading,
       text: text
     ) ?? RawTokenSyntax(missing: .stringQuote, arena: arena)
-    text = text.dropFirst(openQuote.tokenText.count)
+    if !openQuote.isMissing {
+      text = text.dropFirst(openQuote.tokenText.count)
+    }
 
     /// Parse segments.
     let (segments, closeStart) = self.parseStringLiteralSegments(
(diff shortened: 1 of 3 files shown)


Sources/SwiftParser/Expressions.swift
```diff
@@ -907,7 +907,9 @@ extension Parser {
       at: openDelimiter != nil ? .leadingRaw : .leading,
       text: text
     ) ?? RawTokenSyntax(missing: .stringQuote, arena: arena)
-    text = text.dropFirst(openQuote.tokenText.count)
+    if !openQuote.isMissing {
```
omochi, line 910: I fixed logic here.

If opener quote missing token is generated, don't drop text.

Sources/SwiftParser/Expressions.swift
```diff
@@ -917,16 +919,37 @@ extension Parser {
     /// Parse close quote.
     let closeQuote = self.parseStringLiteralQuote(
       at: openDelimiter != nil ? .trailingRaw : .trailing,
-      text: text[closeStart...]
+      text: text
```
omochi, line 922: subscript is unnecessary here

Sources/SwiftParser/Expressions.swift
```diff
@@ -1060,7 +1083,10 @@ extension Parser {
       // position == .leadingRaw implies that we saw a `#` before the quote.
       // A multiline string literal must always start its contents on a new line.
       // Thus we are parsing somethign like #"""#, which is not a multiline string literal but a raw literal containing a single quote.
-      if position == .leadingRaw && text[index] == UInt8(ascii: "#") {
+      if position == .leadingRaw,
+         index < text.endIndex,
+         text[index] == UInt8(ascii: "#")
```
omochi, line 1088: This subscript access may access out of bounds.
I added checking of range.

Tests/SwiftParserTest/Expressions.swift
```diff
@@ -310,6 +310,36 @@ final class ExpressionTests: XCTestCase {
        """"""#^DIAG^#
        """##
      )
+
+    AssertParse(
+      ##"""
+      #"#^DQ^##^DD^#
```
omochi, line 316: Though I didn't report this input, it was also failed before.

Sources/SwiftParser/Expressions.swift
```diff
@@ -917,16 +919,37 @@ extension Parser {
...
+
+      if let openDelimiter = openDelimiter {
+        return RawTokenSyntax(
+          kind: .rawStringDelimiter,
+          text: openDelimiter.tokenText,
+          leadingTriviaPieces: [],
+          trailingTriviaPieces: [],
+          presence: .missing,
+          arena: arena
+        )
```
ahoppen, line 944: You could use
```suggestion
        return RawTokenSyntax(
          missing: .rawStringDelimiter,
          text: openDelimiter.tokenText,
          arena: arena
        )
```

If you add `text: SyntaxText? = nil,` to `RawTokenText.init(missing:arena:)` and apply the following change in the initializer.
```diff
- text: kind.defaultText ?? "",
+ text: text ?? kind.defaultText ?? "",
```

I think I’ve got at least two branches locally that do this as well…

Sources/SwiftParser/Expressions.swift
```diff
@@ -917,16 +919,37 @@ extension Parser {
...
+      }
+
+      return nil
+    }()
```
ahoppen, line 948: I’m personally not a huge fan of declaring a closure that you immediately call. I would prefer a paradigm like

```
let closeDelimiter: RawTokenSyntax?
if let parsedDelimiter = self.parseStringLiteralDelimiter(…) {
  closeDelimiter = parsedDelimiter
} else if let openDelimiter = openDelimiter {
  closeDelimiter = RawTokenSyntax(… missing ...)
} else {
  closeDelimiter = nil
}
```

That matches the style we use in the rest of the parser.

Tests/SwiftParserTest/Expressions.swift
ahoppen: I think you can just use a single marker in these tests because the two you have resolve to the same location currently.
```suggestion
      #"#^DIAG^#
```

If you name it `DIAG`, you also don’t need to specify the maker name in `DiagnosticSpec`.

ahoppen (commented):
Thanks for fixing these issues. The PR looks good, I’ve got a few stylistic comments.

Sources/SwiftParser/Expressions.swift
omochi: Actually I looked for `RawTokenText(missing:text:arena:)` first time 😄.
I added `text:` and changed to use it.

Sources/SwiftParser/Expressions.swift
omochi: Sorry I didn't check style well.
I changed it not to use closure.

Tests/SwiftParserTest/Expressions.swift
omochi: I changed to share single `#^DIAG^#`.

omochi:
@ahoppen Thanks for review. I changed styles to follow comments.

ahoppen:
@swift-ci Please test

ahoppen (approved):
Thanks for your fixes to the parser 👍

PR #692, opened 2022-09-02
Use `swift-argument-parser` in `generate-swift-syntax-builder`

fwcd:
This branch updates `generate-swift-syntax-builder` to use `swift-argument-parser` instead of ad-hoc parsing `CommandLine.arguments`.

Since the package already depends on `swift-argument-parser` for `swift-parser-test` I would not expect the introduction of this new target dependency to be an issue, but perhaps still worth checking whether there are any unexpected bootstrapping consequences to this.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -124,7 +124,7 @@ let package = Package(
     ),
     .executableTarget(
         name: "generate-swift-syntax-builder",
-        dependencies: ["SwiftSyntaxBuilder"],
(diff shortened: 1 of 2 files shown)


fwcd:
@swift-ci please test

ahoppen (approved):
I agree that there should not be any issues associated with using swift-argument-parser for generate-swift-syntax-builder. The important bit is that we would still be able to build SwiftParser without depending on swift-argument parser.

fwcd:
@swift-ci please test

ahoppen:
rdar://99493001

PR #695, opened 2022-09-02
Use `@_spi(Testing)` for `Format.indentTrivia`

fwcd:
...to provide slightly stronger encapsulation.


fwcd:
@swift-ci please test

ahoppen (approved):
LGTM. 

Are planning to also remove the underscore from `_format`? I just looked at it and we might not even need `@_spi(Testing)` here but could just make the method internal.

fwcd:
Sounds like a good idea!

PR #701, opened 2022-09-02
Simplify SyntaxClassifications implementation

rintaro:
Walking in a tree structure using `Iterator` is not a simple task - "next()" operation may walk into children, find a sibling, or find an ancestors' sibling. Also, this classification iterator was nested, token iterator and a iterator inside a token. Moreover it has a buffering mechanism that merges the consecutive same-kind elements.

These complexity can be simplified by using a super simple recursive visitor. Something like this pseudo code:
```swift
var classifications: [SyntaxClassifiedRange]

func visit(_ node: Node) {
  guard targetRange.contains(child) else { return }

  // Handle a token.
  if (node.isToken) {
    for piece node.leadingTrivia {
      classifications.append(classify(piece))
    }

    classifications.append(classify(node.tokenKind, node.tokenText))

    for piece node.trailingTrivia {
      classifications.append(classify(piece))
    }
    return
  }
  
  // Recursively visit children.
  for child in node.children {
    visit(child)
  }
}
```
This significantly simplify the implementation and improves the performance, but at the cost of some memory (i.e. now it eagerly construct an array)

### DIFF
--- a/Sources/SwiftSyntax/SyntaxClassifier.swift
+++ b/Sources/SwiftSyntax/SyntaxClassifier.swift
@@ -79,263 +79,6 @@ fileprivate struct TokenKindAndText {
   }
 }
 
-/// Represents a syntax node along with its absolute position in the tree and
-/// an optional classification.
-fileprivate struct AbsoluteNode {
-  let raw: RawSyntax
-  let position: AbsoluteSyntaxPosition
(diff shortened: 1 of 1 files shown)


ahoppen (approved):
Wow, that’s so much simpler. I really like it. 🤩

rintaro:
@swift-ci Please test

PR #702, opened 2022-09-02
Restore Xcode 13 Build

CodaFi:
The use of inferrable default parameters like this is a 5.7-only
feature. Divide AssertParse into an overload that defaults to
SourceFileSyntax parsing and a generic entrypoint that accepts
a closure.

This restores the ability to build with Xcode 13.


CodaFi:
@swift-ci test

PR #703, opened 2022-09-03
[🦩] Add proof of concept for ASTGen.

zoecarver:
Uses `SyntaxTransformVisitor` to start implementing ASTGen. Just a proof of concept for now.

Refs https://github.com/apple/swift/pull/60943


Sources/ASTGen/ASTGen.swift
```diff
@@ -0,0 +1,219 @@
...
+    let initializer = visit(node.initializer!)
+
+    return [pattern.only, initializer.only]
+  }
+
+  public func visit(_ node: VariableDeclSyntax) -> [UnsafeMutableRawPointer] {
+    let components = visit(node.bindings)
+    assert(components.count == 2)
+    let pattern = components.first!
+    let initializer = components.last!
```
zoecarver, line 100: This should probably get re-worked.

Sources/ASTGen/ASTGen.swift
```diff
@@ -0,0 +1,253 @@
...
+    fatalError("Use other overload.")
+  }
+  
+  public func visit(_ node: SourceFileSyntax) -> [UnsafeMutableRawPointer] {
+    let loc = self.base.advanced(by: node.position.utf8Offset).raw
+    var out = [UnsafeMutableRawPointer]()
+
+    for element in node.statements {
+      let swiftASTNodes = visit(element)
+      if element.item.is(StmtSyntax.self) {
```
ahoppen, line 44: I think we should encode information that elements in a source file can only be `StmtSyntax`, `ExprSyntax` or `DeclSyntax` in the syntax tree so we can do an exhaustive switch here.

The best idea that I’ve come up with is to code-gen enums like the following based on the `node_choices` of `CodeBlockItem.Item` in [CommonNodes.py](https://github.com/apple/swift-syntax/blob/main/Sources/generate-swift-syntax-builder/gyb_syntax_support/CommonNodes.py#L37-L43)
```swift
public struct CodeBlockItemSyntax /* CodeBlockItemSyntax conformances */ {
  enum Item {
    case stmt(StmtSyntax)
    case expr(ExprSyntax)
    case decl(DeclSyntax)
    case tokenList(TokenListSyntax)
    case nonEmptyTokenList(NonEmptyTokenListSyntax)
  
    func as(_ t: StmtSyntax.Type) -> StmtSyntax? { … }
    func as<T: StmtSyntaxProtocol>(_ t: T.Type) -> T? { … }
    func as(_ t: ExprSyntax.Type) -> ExprSyntax? { … }
    func as<T: ExprSyntaxProtocol>(_ t: T.Type) -> T? { … }
    func as(_ t: DeclSyntax.Type) -> DeclSyntax? { … }
    func as<T: DeclSyntaxProtocol>(_ t: T.Type) -> T? { … }
    func as(_ t: TokenList.Type) -> TokenListSyntax? { … }
    func as(_ t: nonEmptyTokenList.Type) -> NonEmptyTokenListSyntax? { … }
  }
  // existing code in CodeBlockItemSyntax
}
```

and change the type of `item` in CodeBlockItemSyntax from `Syntax` to `CodeBlockItemSyntax.Item`.
I think (haven’t tested) that this should be mostly (i.e. covering most common API use cases) source-comaptible with the current usage of these APIs.

That way you could exhaustivly switch over it here as
```swift
switch element.item {
    case stmt(StmtSyntax):
      out.append(SwiftTopLevelCodeDecl_createStmt(ctx, declContext, loc, swiftASTNodes, loc))
    case expr(ExprSyntax):
      out.append(SwiftTopLevelCodeDecl_createExpr(ctx, declContext, loc, swiftASTNodes, loc))
    case decl(DeclSyntax):
      out.append(swiftASTNodes)
    case tokenList(TokenListSyntax):
      fatalError("Should not be generated by SwiftParser")
    case nonEmptyTokenList(NonEmptyTokenListSyntax):
      fatalError("Should not be generated by SwiftParser")
}
```

I would really like to see this refactoring done before we fully start ASTGen. I don’t think it needs to block this PR but it’s something we should do ASAP.

A quick improvement without this refactoring: You can use `element.item.as(SyntaxEnum.self)` here and then switch over the enum cases.

Sources/ASTGen/ASTGen.swift
```diff
@@ -0,0 +1,253 @@
...
+      } else {
+        assert(element.item.is(DeclSyntax.self))
+        out.append(swiftASTNodes)
+      }
+    }
+    
+    return out
+  }
+
+  public func visit(_ node: FunctionCallExprSyntax) -> UnsafeMutableRawPointer {
```
ahoppen, line 57: I’m not really a fan that all these functions return `UnsafeMutableRawPointer`. I imagine it will be to easy to e.g. pass a `FunctionDecl` to a parameter that really expects an `Identifier` Can we do something more typed here? Like structs that wrap a `void *` or something like that?

Sources/ASTGen/ASTGen.swift
```diff
@@ -0,0 +1,253 @@
...
+
+extension UnsafePointer {
+  public var raw: UnsafeMutableRawPointer {
+    UnsafeMutableRawPointer(mutating: self)
+  }
+}
+
+struct ASTGenVisitor: SyntaxTransformVisitor {
+  let ctx: UnsafeMutableRawPointer
+  let base: UnsafePointer<CChar>
```
ahoppen, line 22: What do you think about renaming this to `sourceFileBufferStart?

Sources/ASTGen/ASTGen.swift
```diff
@@ -0,0 +1,253 @@
...
+  // TODO: this some how messes up the witness table when I uncomment it locally :/
+//  public func visit<T>(_ node: T?) -> [UnsafeMutableRawPointer]? {
+//    if let node = node { return visit(node) }
+//    return nil
+//  }
+  
+  @_disfavoredOverload
+  public func visit(_ node: SourceFileSyntax) -> UnsafeMutableRawPointer {
+    fatalError("Use other overload.")
+  }
```
ahoppen, line 36: Can’t you just delete this?

Sources/ASTGen/ASTGen.swift
```diff
@@ -0,0 +1,253 @@
...
+}
+
+extension UnsafePointer {
+  public var raw: UnsafeMutableRawPointer {
+    UnsafeMutableRawPointer(mutating: self)
+  }
+}
+
+struct ASTGenVisitor: SyntaxTransformVisitor {
+  let ctx: UnsafeMutableRawPointer
```
ahoppen, line 21: I assume this is a pointer to `ASTContext`? In the absence of proper typing, could you rename the variable to `astContext` to at least give some clue what this is?

Sources/ASTGen/ASTGen.swift
```diff
@@ -0,0 +1,253 @@
...
+
+  public func visit(_ node: FunctionCallExprSyntax) -> UnsafeMutableRawPointer {
+    let args = visit(node.argumentList)
+    // TODO: hack
+    let callee = visit(node.calledExpression)
+    return SwiftFunctionCallExpr_create(self.ctx, callee, args)
+  }
+
+  public func visit(_ node: IdentifierExprSyntax) -> UnsafeMutableRawPointer {
+    let loc = self.base.advanced(by: node.position.utf8Offset).raw
```
ahoppen, line 65: What do you think about extracting this to a helper function?

```swift
func loc<SyntaxType: SyntaxProtocol>(of node: SyntaxType) -> UnsafePointer<CChar> {
  return self.base.advanced(by: node.position.utf8Offset).raw
}
```

Sources/ASTGen/ASTGen.swift
```diff
@@ -0,0 +1,253 @@
...
+    let returnType: UnsafeMutableRawPointer?
+    if let output = node.signature.output {
+      returnType = visit(output.returnType)
+    } else {
+      returnType = nil
+    }
+    
+    let params = node.signature.input.parameterList.map { visit($0) }
+    return params.withBridgedArrayRef { ref in
+      FuncDecl_create(ctx, loc, false, loc, name, loc, false, nil, false, nil, loc, ref, loc, body, returnType, declContext)
```
ahoppen, line 204: I really wish we had argument labels here. Just an idea: You define `FuncDecl_create` on the C++ side and on the Swift side you define 

```swift
struct FuncDecl {
  var pointer: UnsafeMutableRawPointer

  init(<all the parameters>) {
    self.pointer = FuncDecl_create(<forward all the parameters>)
  }
}
```

The downside is that if you modify the parameters on `FuncDecl_create`, you also need to modify them on `FuncDecl.init`. But we could maybe also use `FuncDecl` as a typed wrapper instead of `UnsafeMutableRawPointer` in ASTGen.

And another thought: It might be good to namespace these types in an `enum AST` so they don’t conflict with types from SwiftSyntax or SwiftSyntaxBuilder.

Sources/ASTGen/ASTGen.swift
bnbarham: I was going to mention this on the C++ side, but since you have here - IMO it'd be good to define a bunch of opaque types (eg. `typedef void *swift_source_loc_t`) and use those instead of `void *` directly.

And another comment mentions labels - we could just add `swift_name` to the C functions, eg. `__attribute__((swift_name("FuncDecl_create(ctx:staticLoc:....)")))`

Sources/ASTGen/ASTGen.swift
DougGregor: I think this would be absolutely fantastic! The hardest thing about working with the syntax tree is that you don't know what cases to cover at any given point without working back from the grammar.

DougGregor: All that said, I don't know that *this* pull request is the place for this discussion. It's a general improvement to the ergonomics of the SwiftSyntax APIs.

DougGregor (commented):
I'll look more into details, but from a layering perspective I don't think this ASTGen library belongs in the swift-syntax package. It's too specific to the compiler itself, since it is layered on top of the compiler's (C++) AST representation, so it's a layering violation to have it here.

Sources/ASTGen/ASTGen.swift
zoecarver: No, we have a default implementation in the protocol that will be ambiguous.

zoecarver: We need the `_disfavoredOverload` attribute, so we need to add this overload.

Sources/ASTGen/ASTGen.swift
zoecarver: Great idea.

zoecarver:
Closing this patch as I don't think we need it anymore.

PR #704, opened 2022-09-03
Improve parsing of access level modifier

### LINKED ISSUE #631: Parser crash when parsing `private(`
### Issue Kind

Parser Crash

### Source Code

```swift
private(
```

### Description

Stack trace:
```
Thread 0 Crashed::  Dispatch queue: com.apple.main-thread
0   libswiftCore.dylib            	    0x7ff8187548a8 closure #1 in closure #1 in closure #1 in _assertionFailure(_:_:file:line:flags:) + 456
1   libswiftCore.dylib            	    0x7ff8187545bb closure #1 in closure #1 in _assertionFailure(_:_:file:line:flags:) + 267
2   libswiftCore.dylib            	    0x7ff8187543ad closure #1 in _assertionFailure(_:_:file:line:flags:) + 333
3   libswiftCore.dylib            	    0x7ff818753f25 _assertionFailure(_:_:file:line:flags:) + 309
4   libswiftCore.dylib            	    0x7ff8189dce51 specialized Collection._failEarlyRangeCheck(_:bounds:) + 673
5   swift-parser-test             	       0x1095c3ec9 protocol witness for Collection._failEarlyRangeCheck(_:bounds:) in conformance SyntaxText + 9
6   libswiftCore.dylib            	    0x7ff8188a609b Slice.subscript.getter + 523
7   libswiftCore.dylib            	    0x7ff81881f2a8 Collection.subscript.getter + 200
8   swift-parser-test             	       0x1093c2b1e Parser.parseStringLiteralSegments(_:_:_:) + 4782 (Expressions.swift:1167)
9   swift-parser-test             	       0x1093bcde5 Parser.parseStringLiteral() + 1269 (Expressions.swift:927)
10  swift-parser-test             	       0x1093b6564 Parser.parsePrimaryExpression(_:) + 740 (Expressions.swift:716)
11  swift-parser-test             	       0x1093b61fc Parser.parsePostfixExpression(_:forDirective:) + 124 (Expressions.swift:368)
12  swift-parser-test             	       0x1093b5830 Parser.parseUnaryExpression(_:forDirective:) + 672 (Expressions.swift:339)
13  swift-parser-test             	       0x1093b50cd Parser.parseSequenceExpressionElement(_:forDirective:) + 3613 (Expressions.swift:284)
14  swift-parser-test             	       0x1093b22e7 Parser.parseSequenceExpression(_:forDirective:inVarOrLet:) + 791 (Expressions.swift:94)
15  swift-parser-test             	       0x1093b1f5f Parser.parseExpression(_:) + 1775 (Expressions.swift:40)
16  swift-parser-test             	       0x10940d526 Parser.parseItem() + 2070 (TopLevel.swift:168)
17  swift-parser-test             	       0x10940b3ca Parser.parseCodeBlockItem() + 218 (TopLevel.swift:117)
18  swift-parser-test             	       0x10940b23c Parser.parseTopLevelCodeBlockItems() + 540 (TopLevel.swift:44)
19  swift-parser-test             	       0x10940af9f Parser.parseSourceFile() + 47 (TopLevel.swift:28)
20  swift-parser-test             	       0x1093f2857 closure #1 in static Parser.parse(source:parseTransition:filenameForDiagnostics:languageVersion:enableBareSlashRegexLiteral:) + 71 (Parser.swift:45)
21  swift-parser-test             	       0x1093f2aa0 partial apply for closure #1 in static Parser.parse(source:parseTransition:filenameForDiagnostics:languageVersion:enableBareSlashRegexLiteral:) + 16
22  libswiftCore.dylib            	    0x7ff8187661fc withExtendedLifetime<A, B>(_:_:) + 12
23  swift-parser-test             	       0x1093f1f0b static Parser.parse(source:parseTransition:filenameForDiagnostics:languageVersion:enableBareSlashRegexLiteral:) + 811 (Parser.swift:44)
24  swift-parser-test             	       0x1093f1ad8 closure #1 in static Parser.parse(source:parseTransition:filenameForDiagnostics:languageVersion:enableBareSlashRegexLiteral:) + 104 (Parser.swift:26)
25  swift-parser-test             	       0x1093f1bd2 partial apply for closure #1 in static Parser.parse(source:parseTransition:filenameForDiagnostics:languageVersion:enableBareSlashRegexLiteral:) + 50
26  swift-parser-test             	       0x1093f4a28 closure #1 in _SmallString.withUTF8<A>(_:) + 72
27  swift-parser-test             	       0x1093f4a70 partial apply for closure #1 in _SmallString.withUTF8<A>(_:) + 32
28  libswiftCore.dylib            	    0x7ff8187ed9b0 withUnsafeMutableBytes<A, B>(of:_:) + 32
29  libswiftCore.dylib            	    0x7ff8187ed987 withUnsafeBytes<A, B>(of:_:) + 23
30  swift-parser-test             	       0x1093f2194 String.withUTF8<A>(_:) + 372
31  swift-parser-test             	       0x1093f188e static Parser.parse(source:parseTransition:filenameForDiagnostics:languageVersion:enableBareSlashRegexLiteral:) + 382 (Parser.swift:25)
32  swift-parser-test             	       0x10a3d35c4 static VerifyRoundTrip.run(source:swiftVersion:enableBareSlashRegex:) + 196 (swift-parser-test.swift:102)
33  swift-parser-test             	       0x10a3d3445 VerifyRoundTrip.run() + 261 (swift-parser-test.swift:98)
34  swift-parser-test             	       0x10a3d486b protocol witness for ParsableCommand.run() in conformance VerifyRoundTrip + 27
35  swift-parser-test             	       0x1092f517c static ParsableCommand.main(_:) + 156 (ParsableCommand.swift:97)
36  swift-parser-test             	       0x1092f52d7 static ParsableCommand.main() + 39 (ParsableCommand.swift:107)
37  swift-parser-test             	       0x10a3d175f static SwiftParserTest.$main() + 47 (swift-parser-test.swift:56)
38  swift-parser-test             	       0x10a3df198 main + 24
39  dyld                          	    0x7ffb270182a9 start + 2345
```

omochi:
This PR fix #631.
It can round-trip correctly without crashing.

Not only that, I tried to improve recovering logic here.

For example, the logic recover such wrong inputs intuitively way.

```swift
private(get, set, didSet)
```

See my test cases for detail.
How do you think?

### DIFF
--- a/Sources/SwiftParser/Modifiers.swift
+++ b/Sources/SwiftParser/Modifiers.swift
@@ -175,14 +175,7 @@ extension Parser {
   }
 
   mutating func parseAccessLevelModifier() -> RawDeclModifierSyntax {
-    assert(
-      [
-        RawTokenKind.privateKeyword,
-        .fileprivateKeyword,
-        .internalKeyword,
-        .publicKeyword
-      ].contains(currentToken.tokenKind)
-    )
+    assert(atAny([.privateKeyword, .fileprivateKeyword, .internalKeyword, .publicKeyword]))
     let name = consumeAnyToken()
     let details: RawDeclModifierDetailSyntax?
     if let lparen = consume(if: .leftParen) {
@@ -190,26 +183,22 @@ extension Parser {
       var detail: RawTokenSyntax? = nil
       var afterDetail: [RawSyntax] = []
 
-      let rightParenPrecedence = TokenPrecedence(.rightParen)
-
-      // Find `set` keyword during eat all tokens as junk
-      // until reach to right-paren or other strong token.
-      while !at(.eof),
-            TokenPrecedence(currentToken.tokenKind) < rightParenPrecedence
-      {
-        if currentToken.isContextualKeyword("set") {
-          detail = consume(remapping: .contextualKeyword)
-          break
+      let (content, rightParen) = expect(.rightParen)
+      if let content = content {
+        for element in content.elements {
+          if let _ = detail {
+            afterDetail.append(element)
+          } else if var token = element.as(RawTokenSyntax.self),
+             token.isContextualKeyword("set")
+          {
+            token.tokenKind = .contextualKeyword
+            detail = token
+          } else {
+            beforeDetail.append(element)
+          }
         }
-        beforeDetail.append(RawSyntax(consumeAnyToken()))
       }
 
-      // Eat rest of junk and right-paren
-      let (junk, rightParen) = expect(.rightParen)
-      if let junk = junk {
-        afterDetail.append(contentsOf: junk.elements)
-      }
-      
       details = RawDeclModifierDetailSyntax(
         leftParen: lparen,
         beforeDetail.isEmpty ? nil :
--- a/Sources/SwiftParser/Modifiers.swift
+++ b/Sources/SwiftParser/Modifiers.swift
@@ -159,7 +159,7 @@ extension Parser {
   mutating func parseAccessLevelModifier() -> RawDeclModifierSyntax {
     let (unexpectedBeforeName, name) = expectAny(
       [.privateKeyword, .fileprivateKeyword, .internalKeyword, .publicKeyword],
-      default: .privateKeyword
+      default: .internalKeyword
     )
     let details: RawDeclModifierDetailSyntax?
     if let leftParen = consume(if: .leftParen) {
--- a/Sources/SwiftParser/Modifiers.swift
+++ b/Sources/SwiftParser/Modifiers.swift
@@ -186,43 +186,42 @@ extension Parser {
     let name = self.consumeAnyToken()
     let details: RawDeclModifierDetailSyntax?
     if let lparen = self.consume(if: .leftParen) {
-      var beforeDetail: [RawTokenSyntax] = []
+      var beforeDetail: [RawSyntax] = []
       var detail: RawTokenSyntax? = nil
-      while !atAny([.eof, .rightParen]) {
-        if currentToken.isContextualKeyword("set") {
-          detail = consume(remapping: .contextualKeyword)
-          break
+      var afterDetail: [RawSyntax] = []
+
+      let (content, rightParen) = expect(.rightParen)
+      if let content = content {
+        for element in content.elements {
+          if let _ = detail {
+            afterDetail.append(element)
+          } else if let token = element.as(RawTokenSyntax.self),
+                    [RawTokenKind.identifier, .contextualKeyword].contains(token.tokenKind),
+                    token.tokenText == "set"
+          {
+            detail = token
+          } else {
+            beforeDetail.append(element)
+          }
         }
-        beforeDetail.append(consumeAnyToken())
       }
-      var afterDetail: [RawTokenSyntax] = []
-      if let _ = detail {
-        while !atAny([.eof, .rightParen]) {
-          afterDetail.append(consumeAnyToken())
-        }
-      } else {
-        afterDetail = beforeDetail
-        beforeDetail = []
-      }
-      let rparen = consume(if: .rightParen) ??
-        RawTokenSyntax(missing: .rightParen, arena: arena)
+      
       details = RawDeclModifierDetailSyntax(
         leftParen: lparen,
         beforeDetail.isEmpty ? nil :
-        RawUnexpectedNodesSyntax(
-          elements: beforeDetail.map { RawSyntax($0) }, arena: arena
-        ),
+          RawUnexpectedNodesSyntax(
+            elements: beforeDetail, arena: arena
+          ),
         detail: detail ?? RawTokenSyntax(
           missing: .contextualKeyword,
           text: arena.intern("set"),
           arena: arena
         ),
         afterDetail.isEmpty ? nil :
-        RawUnexpectedNodesSyntax(
-          elements: afterDetail.map { RawSyntax($0) },
-          arena: arena
-        ),
-        rightParen: rparen,
+          RawUnexpectedNodesSyntax(
+            elements: afterDetail, arena: arena
+          ),
+        rightParen: rightParen,
         arena: self.arena
       )
     } else {
--- a/Sources/SwiftParser/Modifiers.swift
+++ b/Sources/SwiftParser/Modifiers.swift
@@ -47,29 +47,11 @@ extension Parser {
     var modifierLoopCondition = LoopProgressCondition()
     MODIFIER_LOOP: while modifierLoopCondition.evaluate(currentToken) {
       switch self.at(anyIn: DeclarationModifier.self) {
-      case (.privateKeyword, let handle)?,
-          (.fileprivateKeyword, let handle)?,
-          (.internalKeyword, let handle)?,
-          (.publicKeyword, let handle)?:
-        let name = self.eat(handle)
-        let details: RawDeclModifierDetailSyntax?
-        if let lparen = self.consume(if: .leftParen) {
-          let (unexpectedBeforeDetail, detail) = self.expectContextualKeyword("set")
-          let (unexpectedBeforeRParen, rparen) = self.expect(.rightParen)
-          details = RawDeclModifierDetailSyntax(
-            leftParen: lparen,
-            unexpectedBeforeDetail,
-            detail: detail,
-            unexpectedBeforeRParen,
-            rightParen: rparen,
-            arena: self.arena
-          )
-        } else {
-          details = nil
-        }
-
-        elements.append(RawDeclModifierSyntax(
-          name: name, detail: details, arena: self.arena))
+      case (.privateKeyword, _)?,
+          (.fileprivateKeyword, _)?,
+          (.internalKeyword, _)?,
+          (.publicKeyword, _)?:
+        elements.append(parseAccessLevelModifier())
       case (.staticKeyword, let handle)?:
         let staticKeyword = self.eat(handle)
         elements.append(RawDeclModifierSyntax(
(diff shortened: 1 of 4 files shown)


Sources/SwiftParser/Modifiers.swift
```diff
@@ -167,4 +148,63 @@ extension Parser {
...
+      var afterDetail: [RawTokenSyntax] = []
+      if let _ = detail {
+        while !atAny([.eof, .rightParen]) {
+          afterDetail.append(consumeAnyToken())
+        }
+      } else {
+        afterDetail = beforeDetail
+        beforeDetail = []
+      }
+      let rparen = consume(if: .rightParen) ??
```
CodaFi, line 182: Is `expect` not sufficient to recover to the right paren? I would expect commas wouldn't bind as tightly as parentheses here so you wouldn't have to do the extra token munging here.

omochi: @CodaFi

I made single `expect(.rightParen)` version here.
Is this match with your recommendation?

https://github.com/omochi/swift-syntax/pull/1/files

This implementation is simpler.
And safer when there is no stopper token. 
For example:

```swift
private(
#endif
```

`expect` stops to scanning at `#endif` without eating many tokens following from here.

But diagnostics for 

```swift
private(get, set, didSet)
```

is worse than before.
(see diff https://github.com/omochi/swift-syntax/pull/1/files#diff-ad19f5e399c23c3b244b7e592ced5c7b7013e224463b61962f131d4fbec70599R159-R161)

The reason of changing diagnostics is comma and right-paren have same `TokenPrecedence` level.
`expect` stops to search right-paren when see comma.
How do you think about this change of diagnostics?

If you say that previous diagnostics is better,
I try to achieve both using `expect` and better diagnostics.
I could be it with `expectAny([.rightParen, .comma)]` and some loops.

CodaFi:
I believe Alex has improved things here such that `expect` should work a bit better.

omochi:
@CodaFi 
I updated implementation. 
Please review again.
I am satisfied with all diagnostics I added in tests.

The new token precedence was useful here.
https://github.com/apple/swift-syntax/pull/704/files#diff-31d38b5d7995a4ac07a90fd878dae83e68a45f801e3ccfe213ec9ba73b6dc168R183
Thank you @ahoppen .

---

I tried to use only single `expect` without manual loop.
Unfortunately it couldn't.

For example with following, during find `set` keyword, starts at right of left paren,

```swift
private(get, set
```

`expect(.rightParen)` gives up at right of left paren since there are no expected token after that.
In such case,  parser should move to EoF (right of `set`) and found `set` in moving.

The behavior what i want is that
seek continuously until to meet right paren or some strong token like right paren.
This is a little deferent from `expect` and `canRecoverTo`.
So I wrote manual loop control as this.
https://github.com/apple/swift-syntax/pull/704/files#diff-31d38b5d7995a4ac07a90fd878dae83e68a45f801e3ccfe213ec9ba73b6dc168R172-R180

Sources/SwiftParser/Modifiers.swift
```diff
@@ -167,4 +148,66 @@ extension Parser {
...
+      
+      details = RawDeclModifierDetailSyntax(
+        leftParen: lparen,
+        beforeDetail.isEmpty ? nil :
+          RawUnexpectedNodesSyntax(
+            elements: beforeDetail, arena: arena
+          ),
+        detail: detail ?? RawTokenSyntax(
+          missing: .contextualKeyword,
+          text: arena.intern("set"),
```
ahoppen, line 196: `arena.intern("set")` is not necessary here because `"set"` is a `StaticString` that doesn’t get deallocated because it lives in the binary itself (instead of stack or heap memory). Thus, it doesn’t need to be managed by a `SyntaxArena`.

ahoppen (commented):
Sorry, I had this review pending and apparently didn’t submit it.

Once https://github.com/apple/swift-syntax/pull/718/commits/cf4605f18b7f303982a07d5da50b89b2b6732f7a makes it into `main`, I think we should be able to search for `set` by calling `expectContextualKeyword("set")`.

While we are waiting for that, I would prefer to have a single `expect` call. If the diagnostic is sub-optimal, could you add a `FIXME` to it so we can re-visit it once the extended recovery infrastructure from https://github.com/apple/swift-syntax/pull/718 is in `main`?

Sources/SwiftParser/Modifiers.swift
omochi: When I simply wrote `text: "set"` here,
I got following error:

```
SwiftSyntax/RawSyntax.swift:427: Assertion failed: token text must be managed by the arena, or known default text for the token
2022-09-08 12:00:12.816402+0900 swift-parser-test[59612:27324768] SwiftSyntax/RawSyntax.swift:427: Assertion failed: token text must be managed by the arena, or known default text for the token
```

It come from this assertion.
https://github.com/apple/swift-syntax/blob/1754a24d6f5aebe39cd066b5d17438621ea02009/Sources/SwiftSyntax/Raw/RawSyntax.swift#L427-L428

How do I write?

omochi:
@ahoppen I rewrote it as single `expect(.rightParen)` and find `set` from unexpected tokens.
Diagnostics issues are marked as `FIXME`.

To mimic `currentToken.isContextualKeyword` and `consume(remapping:)` pattern,
I added `isContextualKeyword` and setter of `tokenKind` on `RawTokenSyntax`.
But I don't think that its right way.

By the way, `expectContextualKeyword` in cf4605f18b7f303982a07d5da50b89b2b6732f7a looks what I really want.
I think that we need customizability for behavior of `expect` and this is just `expectImpl`.
Can I wait for merge that?

CodaFi:
@omochi Things have landed.

Sources/SwiftParser/Modifiers.swift
ahoppen: Ah, I see. Makes sense. I’ll probably remove that `assert` in another PR but you can just call `arena.intern` for now.

Sources/SwiftParser/Modifiers.swift
```diff
@@ -167,4 +148,55 @@ extension Parser {
...
+  mutating func parseAccessLevelModifier() -> RawDeclModifierSyntax {
+    assert(atAny([.privateKeyword, .fileprivateKeyword, .internalKeyword, .publicKeyword]))
+    let name = consumeAnyToken()
+    let details: RawDeclModifierDetailSyntax?
+    if let lparen = consume(if: .leftParen) {
+      var beforeDetail: [RawSyntax] = []
+      var detail: RawTokenSyntax? = nil
+      var afterDetail: [RawSyntax] = []
+
+      let (content, rightParen) = expect(.rightParen)
```
ahoppen, line 161: I’m not really a fan of taking apart the unexpected tokens here. Fortunately, with the new recovery API, you should be able to just expect the `set` keyword followed by a `)`. Setting the precedence of the `set` keyword to `weakBracketClose` will make sure that we can skip over weak punctuators (like `,` or `:`) in the details.

```swift
let unexpectedBeforeDetail: RawUnexpectedNodesSyntax?
let detail: RawTokenSyntax
if let setHandle = self.canRecoverToContextualKeyword("set", precedence: .weakBracketClose) {
  (unexpectedBeforeDetail, detail) = self.eat(setHandle)
} else {
  unexpectedBeforeDetail = nil
  detail = RawTokenSyntax(
    missing: .contextualKeyword,
    text: arena.intern("set"),
    arena: arena
  )
}
let (unexpectedBeforeRightParen, rightParen) = self.expect(.rightParen)

details = RawDeclModifierDetailSyntax(
  leftParen: lparen,
  unexpectedBeforeDetail,
  detail: detail,
  unexpectedBeforeRightParen,
  rightParen: rightParen,
  arena: arena
)
```

ahoppen (commented):
I’m sorry that this PR took so many iterations due to my recovery-related changes that would have conflicted with this one. But since it’s merged now, I think we can go ahead with this one. 

Left one suggestion inline. Also, I think you need to adjust the test cases to actually have a declaration that we can attach the modifier to. Adding `var x: Int` or something of the sort to the test cases should do the job.

I think your helpers might also not be needed anymore.

Sources/SwiftParser/Parser.swift
```diff
@@ -238,7 +238,11 @@ extension Parser {
       )
     }
     var lookahead = self.lookahead()
-    return lookahead.canRecoverTo([], contextualKeywords: [name])
+    return lookahead.canRecoverTo(
+        [],
+        contextualKeywords: [name],
+        recoveryPrecedence: precedence
```
omochi, line 244: Unused parameter `precedence` is expected to pass here.

Sources/SwiftParser/Modifiers.swift
```diff
@@ -173,4 +155,44 @@ extension Parser {
...
+    if let leftParen = consume(if: .leftParen) {
+      let unexpectedBeforeDetail: RawUnexpectedNodesSyntax?
+      let detail: RawTokenSyntax
+      if let setHandle = canRecoverToContextualKeyword("set", precedence: .weakBracketClose) {
+        (unexpectedBeforeDetail, detail) = eat(setHandle)
+      } else {
+        unexpectedBeforeDetail = nil
+        detail = RawTokenSyntax(
+          missing: .contextualKeyword,
+          text: "set",
```
omochi, line 174: I can use literal now. Thank you.

Tests/SwiftParserTest/Declarations.swift
```diff
@@ -212,6 +212,96 @@ final class DeclarationTests: XCTestCase {
...
+    AssertParse(
+      """
+      private(#^DIAG^#get: set) var a = 0
+      """,
+      diagnostics: [
+        DiagnosticSpec(message: "Unexpected text 'get:' in modifier")
+      ]
+    )
+
+    AssertParse(
```
omochi, line 264: It is reported input at #631.

omochi (commented):
@ahoppen @CodaFi Thank you for all support for me.

My new code becomes beautiful and perfect. (really almost same as Alex suggested)
New `expect` utilities family is very useful and have good semantics.

I found small bug at `canRecoverToContextualKeyword` and fixed it.

I added test cases to `var` keywords to run it correctly.
I saw changes for `parseDeclaration`.
Using `canRecoverTo` here is nice.

Tests/SwiftParserTest/Declarations.swift
```diff
@@ -212,6 +212,96 @@ final class DeclarationTests: XCTestCase {
...
+      """
+      private(#^DIAG^#+
+        set
+      ) var a = 0
+      """,
+      diagnostics: [
+        DiagnosticSpec(message: "Expected 'set' in modifier"),
+        DiagnosticSpec(message: "Expected ')' to end modifier"),
+        // FIXME: It should print `+` as detail of text.
+        DiagnosticSpec(message: "Unexpected text in variable")
```
ahoppen, line 242: Just FYI (not any request to change this in this PR): The unexpected text spans multiple lines here:
```
+
  set
```

and that’s why we print the generic “Unexpected text” instead of “Unexpected '+ set'” (which would look ugly if there’s a line break in the diagnostic message).

Sources/SwiftParser/Modifiers.swift
```diff
@@ -173,4 +155,44 @@ extension Parser {
       arena: self.arena
     )
   }
+
+  mutating func parseAccessLevelModifier() -> RawDeclModifierSyntax {
+    let (unexpectedBeforeName, name) = expectAny(
+      [.privateKeyword, .fileprivateKeyword, .internalKeyword, .publicKeyword],
+      default: .privateKeyword
```
ahoppen, line 162: It doesn’t really matter too much but I think `.internalKeyword` would be the better default here because it’ Swift’s default access level if you don’t specify anything.

ahoppen (commented):
Thanks for your patience. This looks great now. I’ve got one minor comment inline, otherwise this looks good to be merged.

Tests/SwiftParserTest/Declarations.swift
omochi: I see.
I think that escaped representation is best in such case.

For example,

```
Unexpected '+\n  set'
```

Sources/SwiftParser/Modifiers.swift
omochi: Ah thats right. I fixed it.

Tests/SwiftParserTest/Declarations.swift
omochi: I got an idea.
In this case, can parser split code into `+` and `\n  set` ?
Currently `\n  ` is trailing trivia of `+`.
If they are moved to leading trivia of `set`,
unexpected text is only `+` and `\n  set` is valid code.

ahoppen (approved):
Looks good. Thank you!

ahoppen:
@swift-ci Please test

Tests/SwiftParserTest/Declarations.swift
ahoppen: If you’ve got ideas, I would be very open to discuss them in follow-up PRs. One thing to keep in mind though, is that these lines might be >100 characters long and the unexpected text might span more than 2 lines.

Also, I don’t think we should be showing `\n` in the diagnostic message because it’s non-trivial to read and confusing to newcomers who don’t know about the `\n` escape sequence.

PR #705, opened 2022-09-03
Fixes #681 - Missing generic type in attrs

### LINKED ISSUE #681: Parse @differentiable attribute
### Issue Kind

Round-Trip Failure

### Source Code

```swift
@differentiable(reverse wrt,where T
```

### Description

_No response_

MEnnabah:
I believe this fixes 681. ~However, I think I'm not able to write tests properly.~

~For example, the following fails with error `Expected diagnostic on 1:37 but got 1:28`~
```swift
func testMissingGenericTypeToAttribute() {
  AssertParse(
    """
    @differentiable(reverse wrt,where T #^DIAG_1^#
    func podcastPlaybackSpeed() {
    }
    """,
    diagnostics: [
      DiagnosticSpec(locationMarker: "DIAG_1", message: "Expected '=' in same type requirement"),
    ]
  )
}
```

~I don't think this is true, because the diagnostic marker is not on 1:28. 
It would be great if this PR didn't get merged. Instead, i'd like to fix the test first, so any tips are welcome!~

Edit: Fixed the tests in [d40bae7](https://github.com/apple/swift-syntax/pull/705/commits/d40bae78b1861437810585497ce71630695f022f)

### DIFF
--- a/Tests/SwiftParserTest/Attributes.swift
+++ b/Tests/SwiftParserTest/Attributes.swift
@@ -18,17 +18,18 @@ final class AttributeTests: XCTestCase {
     )
   }
   
-  func testMissingGenericTypeToAttribute() {
-    AssertParse(
+func testMissingGenericTypeToAttribute() {
+  AssertParse(
     """
-    @differentiable(reverse wrt,where T #^DIAG_1^#
+    @differentiable(reverse wrt#^DIAG_1^#,where T#^DIAG_2^##^DIAG_3^#
     func podcastPlaybackSpeed() {
     }
     """,
     diagnostics: [
-      DiagnosticSpec(locationMarker: "DIAG_1", message: "Expected '=' in same type requirement"),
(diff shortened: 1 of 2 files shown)


CodaFi (approved):
Looks great, thanks!

CodaFi:
@swift-ci test

Tests/SwiftParserTest/Attributes.swift
```diff
@@ -17,4 +17,19 @@ final class AttributeTests: XCTestCase {
       ]
     )
   }
+  
+  func testMissingGenericTypeToAttribute() {
+    AssertParse(
+      """
+      @differentiable(reverse wrt#^DIAG_1^#,where T#^DIAG_2^##^DIAG_3^#
```
ahoppen, line 24: You don’t need to specify two location markers here. You could just write `#^DIAG_2^#` here and use that marker for the second and third `DiagnosticSpec`.

MEnnabah: Good one. I've submitted another PR.

PR #707, opened 2022-09-05
Improve token precedence classifications

ahoppen:
Motivated by https://github.com/apple/swift-syntax/pull/704 this adjusts token precedences a little bit to produce better diagnostics. Previously, we had different logic how we classified weak brackets (like `(`) and strong brackets (like `{`). Unify the two concepts and allow consuming weak punctuators (like `.` or `:`) while looking for a weak closing bracket and skipping strong punctuators (like `;`) while looking for a strong closing bracket.

### DIFF
--- a/Tests/SwiftParserTest/Statements.swift
+++ b/Tests/SwiftParserTest/Statements.swift
@@ -150,16 +150,16 @@ final class StatementTests: XCTestCase {
   func testAttributesOnStatements() {
     AssertParse(
       """
-      func test1() {#^END_FUNCTION^#
-        #^EXTRANEOUS^#@s return
+      func test1() {
+        #^TEST_1^#@s return
       }
       func test2() {
-        @unknown return
+        #^TEST_2^#@unknown return
       }
       """,
       diagnostics: [
-        DiagnosticSpec(locationMarker: "END_FUNCTION", message: "Expected '}' to end function"),
-        DiagnosticSpec(locationMarker: "EXTRANEOUS", message: "Extraneous code at top level")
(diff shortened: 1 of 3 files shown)


ahoppen:
@swift-ci Please test

Tests/SwiftParserTest/Statements.swift
```diff
@@ -150,57 +150,54 @@ final class StatementTests: XCTestCase {
...
       func test2() {
-        @unknown return
+        #^TEST_2^#@unknown return
       }
       """,
       diagnostics: [
-        DiagnosticSpec(locationMarker: "END_FUNCTION", message: "Expected '}' to end function"),
-        DiagnosticSpec(locationMarker: "EXTRANEOUS", message: "Extraneous code at top level")
+        DiagnosticSpec(locationMarker: "TEST_1", message: "Unexpected text '@s return' found in function"),
+        DiagnosticSpec(locationMarker: "TEST_2", message: "Unexpected text '@unknown return' found in function")
```
bnbarham, line 162: `return` is a fairly strong keyword (IMO), should we really consume it as well?

bnbarham (approved):

Tests/SwiftParserTest/Statements.swift
ahoppen: That would require more recovery logic in `parseStatement`. I filed https://github.com/apple/swift-syntax/issues/714 to do this later (after I have merged all my other recovery work).

ahoppen:
@swift-ci Please test macOS

PR #708, opened 2022-09-05
Remove underscore from `_format` and make it internal

fwcd:
As suggested in https://github.com/apple/swift-syntax/pull/695#pullrequestreview-1095997368.


fwcd:
@swift-ci please test

ahoppen (approved):

PR #710, opened 2022-09-05
Fixes #679: @specialize with invalid attribute

### LINKED ISSUE #679: Round-trip failure with @specialize attribute
### Issue Kind

Round-Trip Failure

### Source Code

```swift
@_specialize(e
```

### Description

_No response_

MEnnabah:
This closes #679. 

The solution suggests that the `@_specialize` parameter is missing a colon (i.e. `Expected ':' in attribute argument`). This is only true for parameters of `SpecializeParameter` enum. 

However, I believe this behaviour is incorrect in case of supplying a type as a parameter (E.g. `@_specialize(T == Int)`), because it won't have a colon.

The previous example is from [`@_specialize` attribute pitch][pitch], where it seems it's possible to define a parameter with a type only, as opposed to the current unit tests in [`SwiftParserTest`][SwiftParserTest] where the type parameters are only defined with a `where` clause (E.g. `@_specialize(where T.Element == Int)`).

So I'm not 100% sure about the possibility of supplying the type only without a `where` clause.

[pitch]: https://forums.swift.org/t/specialize-attribute/1853
[SwiftParserTest]: https://github.com/apple/swift-syntax/tree/main/Tests/SwiftParserTest

### DIFF
--- a/Sources/SwiftParser/Attributes.swift
+++ b/Sources/SwiftParser/Attributes.swift
@@ -493,9 +493,7 @@ extension Parser {
     // Parse optional "exported" and "kind" labeled parameters.
     while !self.at(.eof) && !self.at(.whereKeyword) {
       let ident = self.parseAnyIdentifier()
-      guard let knownParameter = SpecializeParameter(rawValue: ident.tokenText) else {
-        fatalError()
-      }
+      let knownParameter = SpecializeParameter(rawValue: ident.tokenText)
       let (unexpectedBeforeColon, colon) = self.expect(.colon)
 
       switch knownParameter {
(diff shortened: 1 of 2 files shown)


Sources/SwiftParser/Attributes.swift
```diff
@@ -573,6 +571,15 @@ extension Parser {
           trailingComma: comma,
           arena: self.arena
         )))
+      case .none:
+        let valueLabel = self.consumeAnyToken()
```
CodaFi, line 575: Let's try to parse a trailing comma here just in case the user wrote

```
@_specialize(e, exported: true)
```

or something.

MEnnabah: Thanks for the feedback Robert! I've pushed another commit.

Sources/SwiftParser/Attributes.swift
```diff
@@ -573,6 +571,17 @@ extension Parser {
           trailingComma: comma,
           arena: self.arena
         )))
+      case .none:
```
ahoppen, line 574: Nitpick: We use `case nil` in the rest of the codebase.

ahoppen (approved):
One nitpick, otherwise LGTM.

CodaFi:
@swift-ci test

CodaFi (approved):

PR #716, opened 2022-09-06
Parse an Optional Code Block in Deinits

CodaFi:
The body of a deinitializer is not strictly required a la
declarations in swiftinterface file.


CodaFi:
@swift-ci test

PR #717, opened 2022-09-06
Redo SwiftSyntax's Documentation

CodaFi:
Use the same GYB for the nodes to generate the documentation landing page for SwiftSyntax. The high-level goal here is to get the documentation neatly categorized, but not necessarily to have everything in its final position.

Since there are so many nodes, we ought to look into splitting out articles for each of the syntactic categories and the visitors. This would let us co-locate article files about each class too - e.g. the Syntax Visitor page can link out to an article on visiting and transforming syntax nodes.


ahoppen (approved):

CodaFi:
@swift-ci test

PR #719, opened 2022-09-07
Take Ownership of Swift Attribute Declarations

CodaFi:
Pull declaration and type attributes into the gyb infrastructure and generate them for the parser. To add a new attribute, add an entry to the DECL_ATTR_KINDS list.


Sources/SwiftParser/TypeAttribute.swift.gyb
```diff
@@ -0,0 +1,29 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+@_spi(RawSyntax) import SwiftSyntax
+
+extension Parser {
```
ahoppen, line 23: Why did you choose to define `TypeAttribute` as a nested type in `Parser`? I think they could be equally valuable e.g. for `Lookeahead` and I would define them as top level types.

(same for `DeclarationAttribute`)

Sources/generate-swift-syntax-builder/gyb_syntax_support/AttributeKinds.py
```diff
@@ -0,0 +1,871 @@
...
+#   serialized, then please place that flag on the last line too. For example:
+#     code=123),
+#     NotSerialized, code=321),
+# - Please sort attributes by serialization number.
+# - Please create a "NOTE" comment if a unique number is skipped.
+#
+# If you're adding a new underscored attribute here, please document it in
+# docs/ReferenceGuides/UnderscoredAttributes.md.
+DECL_ATTR_KINDS = [
+  DeclAttribute('_silgen_name', 'SILGenName',
```
ahoppen, line 200: Nit: We use four space for indentation in Python so all all the members of `DECL_ATTR_KINDS` should be indented by four spaces instead of two.

(same for `TYPE_ATTR_KINDS`)

Sources/generate-swift-syntax-builder/gyb_syntax_support/AttributeKinds.py
```diff
@@ -0,0 +1,871 @@
...
+  TypeAttribute('tuple'),
+
+  # Generated interface attributes
+  TypeAttribute('_opaqueReturnTypeOf'),
+]
+
+# Schema for DECL_ATTR:
+#
+# - Attribute name.
+# - Class name without the 'Attr' suffix (ignored for
```
ahoppen, line 175: Is there something missing after `(ignored for`?

Sources/generate-swift-syntax-builder/gyb_syntax_support/AttributeKinds.py
```diff
@@ -0,0 +1,871 @@
...
+#
+# SIMPLE_DECL_ATTR is the same, but the class becomes
+# SimpleDeclAttr<DAK_##NAME>.
+#
+# Please help ease code review/audits:
+# - Please indent once, not to the opening '('.
+# - Please place the "OnXYZ" flags together on the next line.
+# - Please place the non-OnXYZ flags together on the next to last line.
+# - Please place the unique number on the last line. If the attribute is NOT
+#   serialized, then please place that flag on the last line too. For example:
```
ahoppen, line 191: > If the attribute is NOT serialized, then please place that flag on the last line too

We aren’t adhering to this in this file. I would just remove this comment.

Sources/generate-swift-syntax-builder/gyb_syntax_support/AttributeKinds.py
```diff
@@ -0,0 +1,871 @@
...
+  TypeAttribute('unchecked'),
+  TypeAttribute('_typeSequence'),
+  TypeAttribute('_local'),
+  TypeAttribute('tuple'),
+
+  # Generated interface attributes
+  TypeAttribute('_opaqueReturnTypeOf'),
+]
+
+# Schema for DECL_ATTR:
```
ahoppen, line 172: This comment still references the C++ macro names (`DECL_ATTR` etc.). It should probably use the Python names (`DeclAttr` etc.)

Sources/generate-swift-syntax-builder/gyb_syntax_support/AttributeKinds.py
```diff
@@ -0,0 +1,871 @@
...
+
+class TypeAttribute(Attribute):
+    def __init__(self, name):
+        super().__init__(name, name)
+
+    def macro_value(self):
+        return "TYPE_ATTR(" + self.name + ")"
+
+class DeclAttribute(Attribute):
+    def __init__(self, name, klass, *options, code, swift_name=None):
```
ahoppen, line 20: What do you think about moving `code` and `swift_name` before `options`? That way the variadic options are the last option, which makes it easier to read IMHO.

Sources/generate-swift-syntax-builder/gyb_syntax_support/AttributeKinds.py
```diff
@@ -0,0 +1,871 @@
...
+
+    def macro_value(self):
+        return ""
+
+class TypeAttribute(Attribute):
+    def __init__(self, name):
+        super().__init__(name, name)
+
+    def macro_value(self):
+        return "TYPE_ATTR(" + self.name + ")"
```
ahoppen, line 17: I assume that you are using `macro_value` from a `.def.gyb` file. It’s a little unfortunate that we currently spoil the SwiftSyntax repo with C++ macro logic. I understand that it’s easier to implement this way but maybe the cleaner separation of concerns would be to move this code out of the swift-syntax repo and have a big `if` statement in `.def.gyb` like the following:

```python
% if attr is TypeAttribute:
  TYPE_ATTR(${self.name})
% elif attr is DeclAttribute:
  …
% else:
%   fatal_error("Unhandled attribute class")
```

(realized that we need to handle subclasses before superclasses in the `if` after writing it, but that should be just a matter of the `if` order)

Sources/generate-swift-syntax-builder/gyb_syntax_support/AttributeKinds.py
CodaFi: I like having the code last. It helps it stand out, and python requires you spell `code=` after the positional arguments here.

Sources/SwiftParser/TypeAttribute.swift.gyb
CodaFi: I'll move it in a follow-up. I want to scope this just to the python.

Sources/generate-swift-syntax-builder/gyb_syntax_support/AttributeKinds.py
ahoppen: I don’t have super strong opinions, so fine with me.

Sources/SwiftParser/TypeAttribute.swift.gyb
ahoppen: What do you mean by “I want to scope this just to the python.”

Sources/generate-swift-syntax-builder/gyb_syntax_support/AttributeKinds.py
ahoppen: What about `swift_name`. I think it makes more sense to move that after `name`.

Sources/SwiftParser/TypeAttribute.swift.gyb
CodaFi: These types are nested in extensions of the Parser type before this PR, I left them nested afterwards.

ahoppen: Ah, I see. Makes sense 👍

CodaFi:
Doug points out we can do this independently of the C++. I'm going to hold off on the compiler-side merge until people get some living-on knowledge here.

CodaFi:
@swift-ci test

CodaFi:
@swift-ci test

ahoppen:
> Doug points out we can do this independently of the C++. I'm going to hold off on the compiler-side merge until people get some living-on knowledge here.

Does this mean we are duplicating the definition of AttributeKinds.py and the .def file on the C++ side? If so, do we have any measures in place to make sure they stay in sync?

CodaFi:
The C++ PR will gather merge conflicts. I will rebase and merge them back here on Monday and that'll be that.

PR #729, opened 2022-09-08
Emulate the Structure of Pattern Binding Conditions

### LINKED ISSUE #724: OptionalBindingConditions contain ExpressionPatternSyntaxes instead of IdentifierPatternSyntaxes
### Issue Kind

### Source Code

```swift
if let baz = foo {}
```

In the old parser, this produced a tree like the following:

```
...
- OptionalBindingConditionSyntax
  - letKeyword
  - IdentifierPatternSyntax
    - identifier("baz")
  ...
```

In the new parser, this produces the following tree:

```
SourceFileSyntax children=2
      0: IfStmtSyntax children=3
        0: ifKeyword
        1: ConditionElementListSyntax children=1
          0: ConditionElementSyntax children=1
            0: OptionalBindingConditionSyntax children=3
              0: letKeyword
              1: ExpressionPatternSyntax children=1
                  0: identifier("baz")
              2: InitializerClauseSyntax children=2
                0: equal
                  0: identifier("foo")
        2: CodeBlockSyntax children=3
          0: leftBrace
          1: CodeBlockItemListSyntax
          2: rightBrace
  1: eof
```

I wasn't sure if this was an intentional change; it didn't seem to be, since non-optional bindings (such as `let baz = foo` on its own) still produce `IdentifierPatternSyntax` on the left-hand side.

### Description

_No response_

CodaFi:
The expression parser has to go out of its way to parse all the way down to primary expressions before realizing that the expression pattern it was after isn't actually an expression pattern. This involves setting a bit of ambient state, that we have to plumb down expression parsing.

This is pretty gross, but it emulates what the legacy parser was doing here.

Fixes #724

rdar://99669036

### DIFF
--- a/Sources/SwiftParser/Expressions.swift
+++ b/Sources/SwiftParser/Expressions.swift
@@ -70,7 +70,9 @@ extension Parser {
     // sequenced.
     var lastElement: RawExprSyntax
 
-    lastElement = self.parseSequenceExpressionElement(flavor, forDirective: forDirective)
+    lastElement = self.parseSequenceExpressionElement(flavor,
+                                                      forDirective: forDirective,
+                                                      inVarOrLet: inVarOrLet)
 
     var loopCondition = LoopProgressCondition()
     while loopCondition.evaluate(currentToken) {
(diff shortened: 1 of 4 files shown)


CodaFi:
@swift-ci test

PR #732, opened 2022-09-08
Parse Objective-C zero-argument selectors (e.g., `retain`).

DougGregor:


DougGregor:
@swift-ci please smoke test

DougGregor:
@swift-ci please test

Sources/SwiftParser/Attributes.swift
```diff
@@ -451,6 +451,16 @@ extension Parser {
 
       if self.currentToken.isIdentifier || self.currentToken.isKeyword {
         let name = self.consumeAnyToken()
+
+        // If we hit a ')' we may have a zero-argument selector.
+        if self.at(.rightParen) && elements.isEmpty {
```
CodaFi, line 456: This needs to move below the colon check here because ObjC is cursed

```
@objc(:::::::)
func foo() {}
```

Sources/SwiftParser/Attributes.swift
```diff
@@ -451,6 +451,16 @@ extension Parser {
...
+          elements.append(RawObjCSelectorPieceSyntax(
+            name: name,
+            colon: nil,
+            arena: self.arena))
+          continue
```
CodaFi, line 461: I believe you want either `break` or `return` here.

PR #733, opened 2022-09-08
Extend parser consistency checking to check for failed parses.

DougGregor:
Optionally diagnose unexpected and missing tokens as part of the parser consistency checking. Ignore diagnostics within `#if`, because we cannot reason about those yet.


DougGregor:
@swift-ci please test

PR #734, opened 2022-09-08
Miscellaneous parser fixes

DougGregor:
Fix a number of small issues encountered while parsing the standard library:
* Don't lex regex literals with unbalanced close parentheses in them (`/a)/`)
* Lex hexadecimal floating point literals
* Properly parse the `@rethrows` attribute
* Parse `#fileID`
* Parse `@differentiable` on types

### DIFF
--- a/Tests/SwiftParserTest/Attributes.swift
+++ b/Tests/SwiftParserTest/Attributes.swift
@@ -63,6 +63,9 @@ final class AttributeTests: XCTestCase {
       """
       @objc(zeroArg)
       class A { }
+
+      @objc(:::::)
+      func f(_: Int, _: Int, _: Int, _: Int, _: Int) { }
       """
     )
   }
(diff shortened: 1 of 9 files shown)


DougGregor:
@swift-ci please test

CodaFi (approved):

Tests/SwiftParserTest/Attributes.swift
```diff
@@ -66,4 +66,13 @@ final class AttributeTests: XCTestCase {
       """
     )
   }
+
+  func testRethrowsAttribute() {
+    AssertParse(
+      """
+      @rethrows
```
ahoppen, line 76: Is `@rethrows` a valid attribute? Do we need to handle `@try` or other keywords as well?

ahoppen: I see. In that case you probably want to remap `rethrows` to an identifier when consuming it, right?

DougGregor: Finally did this: https://github.com/apple/swift-syntax/pull/818

PR #735, opened 2022-09-08
Fixup Bare Slash Literal Lexing

### LINKED ISSUE #627: Parser incorrectly parses things in between two `/` characters as regex literal
```
min(reduced.count / 2, chunkSize / 2)
```

Produces the following error
1:19: Unexpected text '/ 2, chunkSize / 2' found in expression

It should parse correctly.

CodaFi:
Break the ambiguity with bare slash regex literals that have any amount of whitespace after the slash delimiter. These now lex as spaced binary operators as appropriate.

Fixes #627

rdar://98736315

### DIFF
--- a/Sources/SwiftParser/Lexer.swift
+++ b/Sources/SwiftParser/Lexer.swift
@@ -2015,6 +2015,20 @@ extension Lexer.Cursor {
       return nil
     }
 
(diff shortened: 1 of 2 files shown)


CodaFi:
@swift-ci test

PR #736, opened 2022-09-08
Improve diagnostic message if unexpected tokens are found at the start of a layout node

ahoppen:
This helps us produce better error messages once we implement recovery in front of declarations.

Unfortunately, we don’t have a test case that exercises the new paths yet.

### DIFF
--- a/Sources/SwiftParser/Diagnostics/ParserDiagnosticMessages.swift
+++ b/Sources/SwiftParser/Diagnostics/ParserDiagnosticMessages.swift
@@ -16,12 +16,20 @@ import SwiftSyntax
 let diagnosticDomain: String = "SwiftParser"
 
 extension SyntaxProtocol {
-  var nodeTypeNameForDiagnostics: String? {
+  /// Return a name of this syntax node that can be used to describe it in
+  /// diagnostics.
+  /// Nodes that mostly exist for the syntax tree's structure and don't have a
+  /// correspondence in the source code that's meeingful to the user by default
+  /// use the name of the parent node that encloses it. Pass `false` to `inherit`
+  /// to prevent this name inheritance.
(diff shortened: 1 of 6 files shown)


ahoppen:
@swift-ci Please test

CodaFi (approved):
Doug and I added some more tests that are messing with you here :)

ahoppen:
@swift-ci Please test

Sources/SwiftParser/Diagnostics/ParserDiagnosticMessages.swift
```diff
@@ -16,12 +16,20 @@ import SwiftSyntax
 let diagnosticDomain: String = "SwiftParser"
 
 extension SyntaxProtocol {
-  var nodeTypeNameForDiagnostics: String? {
+  /// Return a name of this syntax node that can be used to describe it in
+  /// diagnostics.
+  /// Nodes that mostly exist for the syntax tree's structure and don't have a
+  /// correspondence in the source code that's meeingful to the user by default
+  /// use the name of the parent node that encloses it. Pass `false` to `inherit`
+  /// to prevent this name inheritance.
```
bnbarham, line 24: I know this is already merged, but FWIW there's `meeingful` + that sentence is a little hard to parse. Maybe:
> Return the name of this syntax node to be used to describe it in diagnostics. If the node itself has no diagnostic name (eg. it exists for the syntax tree's structure), use the name of the enclosing parent node by default. Pass `inherit: false` to prevent this behavior, in which case `nil` will be returned instead.

ahoppen: https://github.com/apple/swift-syntax/pull/760

PR #750, opened 2022-09-08
Improve diagnostics for invalid identifiers

ahoppen:
The pattern here is that we’ll consume the invalid identifier as an `unknown` token and place it into the unexpected tokens in front of the identifier, which has been synthesized as missing. The diagnostic generator will detect this pattern and emit a matching diagnostic.


ahoppen:
@swift-ci Please test

CodaFi (approved):

PR #751, opened 2022-09-08
.swiftinterface-Related Fixes

CodaFi:
A stack of fixes that get us parsing the declarations in swift interface files correctly.

- Model the argument list to @_opaqueReturnTypeOf and parse it
- Parse attributes involving both qualified names and generics
- Rename some of the methods that got touched here to be more internally consistent.


CodaFi:
@swift-ci test

PR #755, opened 2022-09-09
Provide API for working with syntax text that includes malformed UTF-8.

DougGregor:
The `description` operation on a syntax node cannot represent invalid UTF-8. Document this limitation, and introduce `syntaxTextBytes`, which produces an array of bytes that accurately represents the source, even in the presence of malformed UTF-8.

Switch the parser consistency check's round-tripping validation over to use `syntaxTextBytes`, so that we correctly verify round-tripping of malformed UTF-8.

### DIFF
--- a/Sources/SwiftSyntax/Raw/RawSyntax.swift
+++ b/Sources/SwiftSyntax/Raw/RawSyntax.swift
@@ -267,6 +267,40 @@ extension RawSyntax {
   }
 }
 
+extension RawSyntax {
+  /// Enumerate all of the syntax text present in this node, and all
+  /// of its children, to give a source-accurate view of the bytes.
+  ///
+  /// Unlike `description`, this provides a source-accurate representation
+  /// even in the presence of malformed UTF-8 in the input source.
+  ///
+  /// The `SyntaxText` arguments passed to the visitor are only guaranteed
+  /// to be valid within that call. It is unsafe to escape the `SyntaxValue`
+  /// values outside of the closure.
+  public func withEachSyntaxText(body: (SyntaxText) throws -> Void) rethrows {
+    switch rawData.payload {
+    case .parsedToken(let dat):
+      if dat.presence == .present {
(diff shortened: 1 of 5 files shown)


DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test macOS

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please smoke test

DougGregor:
@swift-ci please clean test

DougGregor:
@swift-ci please test

Sources/SwiftSyntax/Raw/RawSyntax.swift
```diff
@@ -267,6 +267,50 @@ extension RawSyntax {
...
+    case .layout(let dat):
+      for case let child? in dat.layout {
+        try child.withEachSyntaxText(body: body)
+      }
+    }
+  }
+
+  /// Retrieve the syntax text as an array of bytes that models the input
+  /// source even in the presence of invalid UTF-8.
+  public var syntaxTextBytes: [UInt8] {
```
ahoppen, line 305: Could we use `Data` instead of `[UInt8]`?

DougGregor: No, we can't take a dependency on Foundation for anything that goes in a compiler. It's too large for any non-Darwin platform.

DougGregor:
@swift-ci please test

Sources/SwiftSyntax/Raw/RawSyntax.swift
ahoppen: And I assume that it’s alright to use Foundation in `swift-parser-test` because it’s only a testing utility?

DougGregor: Yes.

Issue #677, opened 2022-08-31
Round-trip failure with malformed property


DougGregor:
### Issue Kind

Round-Trip Failure

### Source Code

```swift
var{@
```

liamrosenfeld:
I tracked down the cause of the issue to the [parseAccessorIntroducer call](https://github.com/apple/swift-syntax/blob/b422b48289e1b5d4ef36442a05c3e42227248c05/Sources/SwiftParser/Declarations.swift#L1501). It is consuming the attribute but if there are no explicit accessors, the attribute is never added to the syntax tree. This is more apparent in this code snippit: 
```swift
var greeting: Int {
  @available(swift 5.0)
  func addOne(_ num: Int) -> Int {
    return num + 1
  }
  return addOne(1)
}
```
Where the round-tripped failure is:
```swift
var greeting: Int {
  func addOne(_ num: Int) -> Int {
    return num + 1
  }
  return addOne(1)
}
```
I'm currently working on a PR that fixes this.

CodaFi:
Yeah, I think you've got the root cause nailed down. In the C++ parser, it throws a backtracking scope around this point and when it sees that the `func` keyword after the attributes isn't one of the accessor kinds it backtracks all the way to the opening curly and tries again. We don't have a notion of backtracking scopes, but we can implement the checking part with Lookahead instead. The call to `parseAttributeList` in `parseAccessorIntroducer` should be guarded by the check.

liamrosenfeld:
I got it working using Lookahead, thanks for that recommendation. It seems the protocol parsing is dependent on the existing behavior so I am going to fix that now and then if there are no other regressions that need to be addressed after that, I will open up a PR.

liamrosenfeld:
I opened a pull request with a fix. Sorry for the repeated commits that referenced the issue above. Apparently, if you hit command z too many times in git tower, it can undo the creation of the remote without confirmation.

PR #756, opened 2022-09-09
Fix parsing attributes at the start of an accessor

liamrosenfeld:
Previously, it would consume an attribute without adding it to the tree if an identifier is not found. Now, it looks ahead to confirm there is an identifier before consuming.

Added two test cases to verify parseAccessorIntroducer

Fixes: #677


CodaFi:
Excellent. Thanks for tackling this

@swift-ci test

PR #762, opened 2022-09-09
Parallelize swift parse tests

DougGregor:
Refactor and parallelize the parser's self-test, making it ready to introduce a whole lot more code.

### DIFF
--- a/Tests/SwiftParserTest/ParserTests.swift
+++ b/Tests/SwiftParserTest/ParserTests.swift
@@ -3,17 +3,10 @@ import SwiftSyntax
 import SwiftParser
 
 public class ParserTests: XCTestCase {
-  func testSelfParse() throws {
-    // Allow skipping the self parse test in local development environments
-    // because it takes very long compared to all the other tests.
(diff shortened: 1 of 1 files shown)


DougGregor:
@swift-ci please test

Tests/SwiftParserTest/ParserTests.swift
```diff
@@ -1,8 +1,52 @@
...
+    let diagnostics = ParseDiagnosticsGenerator.diagnostics(for: parsed)
+    if !diagnostics.isEmpty {
+      var locationAndDiagnostics: [String] = []
+      let locationConverter = SourceLocationConverter(file: fileURL.lastPathComponent, tree: parsed)
+      for diag in diagnostics {
+        let location = diag.location(converter: locationConverter)
+        let message = diag.message
+        locationAndDiagnostics.append("\(location): \(message)")
+      }
+      XCTFail("""
```
ahoppen, line 26: Could you pass `file` and `line` from `runParserTests` to this `XCTFail` and the `AssertStringsEqualWithDiff` above?

ahoppen (commented):
Nice. Change. How much faster are the self parse tests now that they are running in parallel?

PR #763, opened 2022-09-09
Fix parsing of key-path expressions with something before a `[`

DougGregor:


DougGregor:
@swift-ci please test

Sources/SwiftParser/Expressions.swift
```diff
@@ -707,7 +707,12 @@ extension Parser {
 
     let expression: RawExprSyntax
     if (self.at(anyIn: Operator.self) != nil && self.currentToken.tokenText.count != 1) || self.peek().tokenKind == .leftSquareBracket {
-      let dot = self.consumePrefix(".", as: .period)
+      let dot: RawTokenSyntax
+      if self.currentToken.starts(with: ".") {
+        dot = self.consumePrefix(".", as: .period)
+      } else {
+        dot = self.consumeAnyToken()
```
ahoppen, line 714: What’s the source text of `dot` in this case? If it doesn’t start with `.` it doesn’t appear to be a dot to me 🤔

PR #764, opened 2022-09-09
Enable round-trip testing for the Swift compiler's testsuite 

DougGregor:
Perform parser round-trip testing for the Swift compiler's testsuite,
by looking for Swift source files in its `test` and `validation-test`
directories, recursively. There are a number of expected failures
here, which we handle by skipping those tests (for now).

### DIFF
--- a/Tests/SwiftParserTest/ParserTests.swift
+++ b/Tests/SwiftParserTest/ParserTests.swift
@@ -39,7 +39,11 @@ public class ParserTests: XCTestCase {
     let fileURLs = FileManager.default
       .enumerator(at: path, includingPropertiesForKeys: nil)!
       .compactMap({ $0 as? URL })
-      .filter({$0.pathExtension == "swift"})
+      .filter {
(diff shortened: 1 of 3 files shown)


DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

Tests/SwiftParserTest/ParserTests.swift
```diff
@@ -60,4 +73,95 @@ public class ParserTests: XCTestCase {
       name: "Self-parse tests", path: currentDir, checkDiagnostics: true
     )
   }
+
+  func testSwiftTestsuite() throws {
+    try XCTSkipIf(ProcessInfo.processInfo.environment["SKIP_SELF_PARSE"] == "1")
```
ahoppen, line 78: Is this test a no-op if you don’t have `swift` checked out next to `swift-syntax`? Could you document that you need the sibling checkout to run these tests?

PR #772, opened 2022-09-10
Fixup Regression in Return Statement Parsing

CodaFi:
When we introduced the at(any:) primitive, this condition got inverted so we were parsing

```
{
  return x
}
```

as

```
{
  return
  x
}
```


CodaFi:
@swift-ci test

Issue #754, opened 2022-09-08
Parser Mismatch for self rebind


jpsim:
I know you're all actively looking for source code round-trip failures but I'm not sure if you're looking for discrepancies between the C++ parser and the Swift parser. Here's one in case you are:

## Source Code

```swift
if let self = self { return }
```

<details>
<summary>Swift Parser Syntax Tree</summary>

```
▿ SourceFileSyntax
  - unexpectedBeforeStatements: nil
  ▿ statements: CodeBlockItemListSyntax
    ▿ CodeBlockItemSyntax
      - unexpectedBeforeItem: nil
      ▿ item: IfStmtSyntax
        - unexpectedBeforeIfKeyword: nil
        ▿ ifKeyword: ifKeyword
          - text: "if"
          ▿ leadingTrivia: []
            - pieces: 0 elements
          ▿ trailingTrivia: spaces(1)
            ▿ pieces: 1 element
              ▿ spaces(1)
                - spaces: 1
          - tokenKind: SwiftSyntax.TokenKind.ifKeyword
        - unexpectedBetweenIfKeywordAndConditions: nil
        ▿ conditions: ConditionElementListSyntax
          ▿ ConditionElementSyntax
            - unexpectedBeforeCondition: nil
            ▿ condition: OptionalBindingConditionSyntax
              - unexpectedBeforeLetOrVarKeyword: nil
              ▿ letOrVarKeyword: letKeyword
                - text: "let"
                ▿ leadingTrivia: []
                  - pieces: 0 elements
                ▿ trailingTrivia: spaces(1)
                  ▿ pieces: 1 element
                    ▿ spaces(1)
                      - spaces: 1
                - tokenKind: SwiftSyntax.TokenKind.letKeyword
              - unexpectedBetweenLetOrVarKeywordAndPattern: nil
              ▿ pattern: ExpressionPatternSyntax
                - unexpectedBeforeExpression: nil
                ▿ expression: IdentifierExprSyntax
                  - unexpectedBeforeIdentifier: nil
                  ▿ identifier: selfKeyword
                    - text: "self"
                    ▿ leadingTrivia: []
                      - pieces: 0 elements
                    ▿ trailingTrivia: spaces(1)
                      ▿ pieces: 1 element
                        ▿ spaces(1)
                          - spaces: 1
                    - tokenKind: SwiftSyntax.TokenKind.selfKeyword
                  - unexpectedBetweenIdentifierAndDeclNameArguments: nil
[...]

PR #774, opened 2022-09-10
Forward Consumption Handles to All Productions

### LINKED ISSUE #770: Failure to make progress in loop on invalid code
### Issue Kind

Parser Crash

### Source Code

```swift
protocol P{{}case
```

### Description

_No response_

CodaFi:
canRecoverTo and expect have different precedence models, so it's not enough to use one on one side and pair it with the other. In this case, it's because strong structure markers like braces cannot be overcome by the latter, but _can_ be overcome by the former. So we could potentially not make progress in some cases.

Fixes #770 and #754

rdar://99766200


CodaFi:
@swift-ci test

CodaFi:
@swift-ci test

CodaFi:
@swift-ci test

PR #782, opened 2022-09-12
Enable the creation of syntax nodes from a string interpolation

DougGregor:
Make the main categories of syntax nodes (`DeclSyntax`, `ExprSyntax`, `StmtSyntax`, `TypeSyntax`, and `PatternSyntax`) conform to `ExpressiblyByStringInterpolation`. The string literal portions are source text, and one can interpolate both additional source text and syntax nodes into that source text. The result will be parsed by the parser to produce a node of the given kind. For example, we can produce a `TypeSyntax` like this:

    let tupleSyntax: TypeSyntax = "(Int, name: String)"

And then use that in another production, for example to create a function declaration:

    let fnSyntax: DeclSyntax =
       "func \(name)(String) async throws -> \(tupleSyntax)"

This approach is intended to simplify code generation, by allowing code to generate Swift syntax trees in the most natural way possible---by writing the Swift code that generates them.

### DIFF
--- a/Sources/SwiftParser/CMakeLists.txt
+++ b/Sources/SwiftParser/CMakeLists.txt
@@ -24,6 +24,7 @@ add_library(SwiftParser STATIC
   RawTokenKindSubset.swift
   Recovery.swift
   Statements.swift
+  Syntax+StringInterpolation.swift
   TokenConsumer.swift
   TokenPrecedence.swift
   TopLevel.swift
--- a/Sources/SwiftParser/Syntax+StringInterpolation.swift
+++ b/Sources/SwiftParser/Syntax+StringInterpolation.swift
@@ -1,0 +1,124 @@
+@_spi(RawSyntax)
+import SwiftSyntax
+
+/// An individual interpolated syntax node.
+struct InterpolatedSyntaxNode {
+  let node: Syntax
+  let startIndex: Int
+  let endIndex: Int
+}
+
+/// The string interpolation type used for creating syntax nodes.
+public struct SyntaxStringInterpolation {
+  /// The source text in UTF-8.
+  ///
+  /// We use an array of UTF-8 for the representation of the source text
+  /// because that's what the parser uses, and we need the stable indices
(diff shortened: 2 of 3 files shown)


DougGregor:
@swift-ci please test

ktoso:
I'm loving this a lot 🥳 

Managed to get out of the source generation business a while ago, but it'll be back -- very happy we'll have this at our disposal then 👍

DougGregor:
Hmm, I think I should opt in `SourceFileSyntax` as well, because it's handy to generate whole source files like this.

jpsim:
Looks great. Here's how it simplifies some rewriting logic in SwiftLint:

### Before

```swift
let functionCall = FunctionCallExprSyntax(
    calledExpression: ExprSyntax(
        MemberAccessExprSyntax(
            base: node.first!.withoutTrivia(),
            dot: .periodToken(),
            name: .identifier("toggle"),
            declNameArguments: nil
        )
    ),
    leftParen: .leftParenToken(), argumentList: .init([]), rightParen: .rightParenToken(),
    trailingClosure: nil, additionalTrailingClosures: nil
)

let newNode = node
    .replacing(childAt: 0, with: ExprSyntax(functionCall))
    .removingLast()
    .removingLast()
    .withLeadingTrivia(node.leadingTrivia ?? .zero)
    .withTrailingTrivia(node.trailingTrivia ?? .zero)
```

### After

```swift
let newNode = node
    .replacing(childAt: 0, with: "\(node.first!.withoutTrivia()).toggle()")
    .removingLast()
    .removingLast()
    .withLeadingTrivia(node.leadingTrivia ?? .zero)
    .withTrailingTrivia(node.trailingTrivia ?? .zero)
```

---

This is for the `toggle_bool` rule that rewrites `someFlag = !someFlag` to `someFlag.toggle()`.

DougGregor:
> This is for the `toggle_bool` rule that rewrites `someFlag = !someFlag` to `someFlag.toggle()`.

Beautiful!

CodaFi (approved):

PR #783, opened 2022-09-12
Correct Typo in the Lexer

### LINKED ISSUE #780: Unspaced binary operator treated as PostfixUnaryExprSyntax instead of SequenceExprSyntax
### Issue Kind

### Source Code

```swift
myString==""
```

### Description

This produces an `PostfixUnaryExprSyntax` when it should instead produce a `SequenceExprSyntax`.

It treats the `==` as a postfix operator instead of an unspaced binary operator.

It should produce a similar syntax tree as the spaced operator equivalent of the expression:

```swift
myString == ""
```

CodaFi:
This should have been a CR, but was instead checking for ". This appears to be a typo.

Fixes #780

rdar://99804977

### DIFF
--- a/Sources/SwiftParser/Lexer.swift
+++ b/Sources/SwiftParser/Lexer.swift
@@ -237,7 +237,7 @@ extension Lexer.Cursor {
     }
 
     switch self.peek() {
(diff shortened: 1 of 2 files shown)


CodaFi:
@swift-ci test

jpsim (approved):
I can confirm this fixes #780 in my testing. Thanks for the quick fix!

PR #790, opened 2022-09-14
Newlines Prevent Bare Slash Regex Lexing

CodaFi:
Break bare slash regex parsing at newline boundaries as well. The lexing test here was trying to pair the / operator with the comment operator several lines below, which was catching the brace in the fray.


CodaFi:
@swift-ci test

PR #796, opened 2022-09-15
Assert the Linkage of SwiftSyntax and SwiftParser

CodaFi:
We want to keep this set as minimal as possible. To that end, provide a test on Darwin that the set of link-time dependencies remains known to us. If this set expands, this test will fail and we'll be able to investigate the dependency growth.

### DIFF
--- a/Tests/SwiftParserTest/Linkage.swift
+++ b/Tests/SwiftParserTest/Linkage.swift
@@ -9,10 +9,152 @@ final class LinkageTest: XCTestCase {
   // a SwiftSyntax maintainer to see if there's a way to avoid adding the
   // dependency.
   func testLinkage() throws {
+    guard let baseURL = try self.findEnclosingTestBundle() else {
+      XCTFail("Unable to determine path to enclosing xctest bundle")
+      return
+    }
+
+    try assertLinkage(of: "SwiftSyntax", in: baseURL, assertions: [
+      .library("-lobjc"),
+      .library("-lswiftCompatibility51", condition: .mayBeAbsent("Starting in Xcode 14 this library is not always autolinked")),
+      .library("-lswiftCompatibility56", condition: .mayBeAbsent("Starting in Xcode 14 this library is not always autolinked")),
+      .library("-lswiftCompatibilityConcurrency"),
+      .library("-lswiftCore"),
+      .library("-lswiftDarwin"),
+      .library("-lswiftSwiftOnoneSupport", condition: .when(configuration: .debug)),
+      .library("-lswift_Concurrency"),
+      .library("-lswift_StringProcessing", condition: .when(swiftVersionAtLeast: .v5_7)),
+    ])
+
+    try assertLinkage(of: "SwiftParser", in: baseURL, assertions: [
+      .library("-lobjc"),
+      .library("-lswiftCompatibility51", condition: .mayBeAbsent("Starting in Xcode 14 this library is not always autolinked")),
+      .library("-lswiftCompatibility56", condition: .mayBeAbsent("Starting in Xcode 14 this library is not always autolinked")),
+      .library("-lswiftCompatibilityConcurrency"),
+      .library("-lswiftCore"),
+      .library("-lswiftDarwin"),
+      .library("-lswiftSwiftOnoneSupport", condition: .when(configuration: .debug)),
+      .library("-lswift_Concurrency"),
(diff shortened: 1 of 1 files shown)


CodaFi:
@swift-ci test

Tests/SwiftParserTest/Linkage.swift
```diff
@@ -0,0 +1,130 @@
...
+        XCTFail("Unable to determine path to enclosing xctest bundle")
+        return
+      }
+
+      baseURL = baseURL.deletingLastPathComponent()
+      let swiftSyntaxURL = baseURL.appendingPathComponent("SwiftSyntax.o")
+      try assertLinkage(of: swiftSyntaxURL) { linkages in
+        XCTAssertEqual(linkages, [
+          .library("-lobjc"),
+          .library("-lswiftCompatibilityConcurrency"),
```
DougGregor, line 34: Should we allow-list all `-lswiftCompatibility` prefixes? For example, there's a `-lswiftCompatibility56` that just got introduced and will kick in sometimes.

Tests/SwiftParserTest/Linkage.swift
```diff
@@ -0,0 +1,130 @@
...
+          .library("-lswiftDarwin"),
+          .library("-lswiftSwiftOnoneSupport"),
+          .library("-lswift_Concurrency"),
+        ])
+      }
+      let swiftParserURL = baseURL.appendingPathComponent("SwiftParser.o")
+      try assertLinkage(of: swiftParserURL) { linkages in
+        XCTAssertEqual(linkages, [
+          .library("-lobjc"),
+          .library("-lswiftCompatibilityConcurrency"),
```
DougGregor, line 45: Same comment as above.

DougGregor (approved):
This is quite clever and very good hygiene for the project, thank you!

Tests/SwiftParserTest/Linkage.swift
```diff
@@ -0,0 +1,130 @@
...
+  // dependencies. If this set changes - in particular, if it grows - consult
+  // a SwiftSyntax maintainer to see if there's a way to avoid adding the
+  // dependency.
+  func testLinkage() throws {
+    for i in 0..<_dyld_image_count() {
+      let name = try XCTUnwrap(_dyld_get_image_name(i))
+      let path = String(cString: name)
+      guard path.hasSuffix("SwiftParserTest") else {
+        continue
+      }
```
ahoppen, line 17: IIUC we only need this loop to get the path of SwiftParserTest, right? In that case, could we extract the loop to a helper function to make the real test case simpler? I.e. something like

```swift
func getSwiftParserTestBundlePath() -> URL? {
  for i in 0..<_dyld_image_count() {
    let name = try XCTUnwrap(_dyld_get_image_name(i))
    let path = String(cString: name)
    guard path.hasSuffix("SwiftParserTest") else {
      continue
    }
    var baseURL = URL(fileURLWithPath: path)
    while !baseURL.pathComponents.isEmpty, baseURL.pathExtension != "xctest" {
      baseURL = baseURL.deletingLastPathComponent()
    }
    return baseURL
  }
  return nil
}
```

The way it’s currently written, I wonder if we pass the `guard` multiple times.

Tests/SwiftParserTest/Linkage.swift
```diff
@@ -0,0 +1,130 @@
...
+
+  private func assertLinkage(of object: URL, assertion: ([Linkage]) throws -> Void) throws {
+    var linkages = [Linkage]()
+    var lines = try self.readLoadCommands(in: object).makeIterator()
+    while let line = lines.next() {
+      guard line.starts(with: "Load command") else {
+        continue
+      }
+
+      guard
```
ahoppen, line 86: Could you add an example load command in as a comment here? I think that would make your string processing code easier to understand.

Tests/SwiftParserTest/Linkage.swift
```diff
@@ -0,0 +1,130 @@
+#if canImport(Darwin)
```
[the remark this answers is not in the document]
CodaFi: It's not sufficient, I just don't have a plan for Linux, and only a foggy one for Windows.

CodaFi:
@swift-ci test

CodaFi:
@swift-ci test

CodaFi:
@swift-ci test

CodaFi:
@swift-ci test

PR #800, opened 2022-09-16
[Performance] Minimize RawTokenKindSubset.allCases usage

rintaro:
`allCases` is not trivial because it creates an Array. We should avoid creating temporary array if possible.

This recovers recent performance regression (2.22s -> 1.31s)

Instead of using `allCases.map { $0.rawTokenKind }.contains(self.currentToken.tokenKind)`, make `RawTokenKindSubset.init(lexeme:)` requirement and `switch` the `tokenKind` in it.

### DIFF
--- a/Sources/SwiftParser/RawTokenKindSubset.swift
+++ b/Sources/SwiftParser/RawTokenKindSubset.swift
@@ -17,6 +17,9 @@
 protocol RawTokenKindSubset: CaseIterable {
   var rawTokenKind: RawTokenKind { get }
 
+  /// Creates an instance if `lexeme` satisfy the condition of this subset.
+  init?(lexeme: Lexer.Lexeme)
+
   /// If this returns a non-nil value, only tokens with this token text are matched.
   /// Must only return a non-nil value if `rawTokenKind` is `identifier` or `contextualKeyword`.
   var contextualKeyword: SyntaxText? { get }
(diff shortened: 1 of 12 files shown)


Sources/SwiftParser/RawTokenKindSubset.swift
```diff
@@ -50,37 +48,18 @@ extension RawTokenKindSubset {
   var precedence: TokenPrecedence? {
     return nil
   }
-
-  func accepts(lexeme: Lexer.Lexeme) -> Bool {
-    return true
-  }
-
-  static var allRawTokenKinds: [RawTokenKind] {
-    return self.allCases.map { $0.rawTokenKind }
```
rintaro, line 59: This was the main offender for the performance.

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

Sources/SwiftParser/RawTokenKindSubset.swift
CodaFi: :oof:

CodaFi (approved):

ahoppen (commented):
This is unfortunate. I was really hoping that we could get away without having to manually implement something like `init?(lexeme:)`. But looks like there’s no way around it 😬

PR #802, opened 2022-09-16
Relax Designated Type Parsing; Parse Pack Expansions

CodaFi:
- Designated type parsing is pretty buggy, so we need to relax the constraints to match it
- Parse pack expansion ellipses everywhere


CodaFi:
@swift-ci test

PR #804, opened 2022-09-16
Allow `swift-parser-test` to take input from stdin.

allevato:
For quick runs, I find it easier to paste some source into a heredoc on the command line vs. creating a named scratch file.


allevato:
@swift-ci test

PR #805, opened 2022-09-16
Generalize Lexeme State into Flags

CodaFi:
Generalize the current "is at start of line" bit into a small flag set. On top of that, pass the lexer's "is multiline string" flag up to the parser. This mirrors the behavior of the C++ parser and makes sure we don't have to spend time re-lexing the string literal to determine this for ourselves in a completely separate component.

The important part here is that we have a space to attach a future "error" bit to the tokens, rather than yielding plain `.unknown` back. This should give us the space to start issuing (what were formerly) lex-time diagnostics.


CodaFi:
@swift-ci test

PR #806, opened 2022-09-16
Use XCTFail over XCTAssertNil for substructure matching

bnbarham:
There's no reason to print the nil assertion message when matching tree substructures - we just want the diff message itself.


bnbarham:
@swift-ci please test

CodaFi (approved):

Issue #785, opened 2022-09-12
Data race in latest versions


jpsim:
### Description

There's a data race present when calling `SyntaxVisitor.visit(_:)` that wasn't present in the 0.50600.1 release.

<details>
<summary>Thread Sanitizer Report</summary>

```
==================
WARNING: ThreadSanitizer: Swift access race (pid=93661)
  Modifying access of Swift variable at 0x000109a50e70 by thread T13:
    #0 BumpPtrAllocator.allocate(byteCount:alignment:) BumpPtrAllocator.swift:98 (swiftlint:arm64+0x100658208)
    #1 specialized BumpPtrAllocator.allocate<A>(_:count:) BumpPtrAllocator.swift:129 (swiftlint:arm64+0x1006c0004)
    #2 specialized static RawSyntax.makeLayout(kind:uninitializedCount:arena:initializingWith:) RawSyntax.swift:634 (swiftlint:arm64+0x10066a498)
    #3 SyntaxData.replacingChild(_:at:) SyntaxData.swift:353 (swiftlint:arm64+0x100753894)
    #4 SyntaxData.withLeadingTrivia(_:) SyntaxData.swift:359 (swiftlint:arm64+0x1007526c0)
    #5 TokenSyntax.withoutTrivia() SyntaxOtherNodes.swift:142 (swiftlint:arm64+0x100755668)
    #6 OperatorUsageWhitespaceVisitor.violation(operatorToken:) OperatorUsageWhitespaceRule.swift:186 (swiftlint:arm64+0x100580fe8)
    #7 OperatorUsageWhitespaceVisitor.visitPost(_:) OperatorUsageWhitespaceRule.swift:145 (swiftlint:arm64+0x1005804b0)
    #8 SyntaxVisitor.visitImplTypeInitializerClauseSyntax(_:) SyntaxVisitor.swift:3781 (swiftlint:arm64+0x1008fc21c)
    #9 SyntaxVisitor.visit(_:) SyntaxVisitor.swift:5890 (swiftlint:arm64+0x1009068dc)
    #10 specialized SyntaxVisitor.visitChildren<A>(_:) SyntaxVisitor.swift:6244 (swiftlint:arm64+0x10090787c)
    #11 SyntaxVisitor.visitImplTypealiasDeclSyntax(_:) SyntaxVisitor.swift:3790 (swiftlint:arm64+0x1008fc2d4)
[...]

PR #807, opened 2022-09-16
SyntaxArena: thread safety

rintaro:
Use mutex to guard concurrent mutation to the arena.
But make a "single thread" mode so that we don't regress performance too much when we know the arena is not used by other threads.

(non `Darwin`/`Glibc` environment e.g. Windows implementation is currently a dummy)

https://github.com/apple/swift-syntax/issues/785
rdar://99812330

### DIFF
--- a/Sources/SwiftSyntax/SyntaxArena.swift
+++ b/Sources/SwiftSyntax/SyntaxArena.swift
@@ -13,14 +13,19 @@
 #if canImport(Darwin)
 import Darwin
 
-class ScopeGuard {
-  private var lock: os_unfair_lock
-  init() {
-    self.lock = os_unfair_lock()
+struct ScopeGuard {
+  private let lock: os_unfair_lock_t
+  init(allocator: BumpPtrAllocator) {
+    let storage = allocator.allocate(os_unfair_lock.self, count: 1).baseAddress!
+    storage.initialize(to: os_unfair_lock())
+    self.lock = os_unfair_lock_t(storage)
   }
+
+  func deinitialize() {}
+
   func withGuard<T>(body: () throws -> T) rethrows -> T {
-    os_unfair_lock_lock(&lock)
-    defer { os_unfair_lock_unlock(&lock)}
+    os_unfair_lock_lock(lock)
+    defer { os_unfair_lock_unlock(lock)}
     return try body()
   }
 }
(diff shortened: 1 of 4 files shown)


rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

jpsim:
Thank you, this certainly helps, although TSan now reports an access race in the guard code:

```

WARNING: ThreadSanitizer: Swift access race (pid=42514)
  Modifying access of Swift variable at 0x0001148c22b0 by thread T25:
    #0 ScopeGuard.withGuard<A>(body:) <null>:82026728 (SwiftLintFrameworkTests:arm64+0x126a334)
    #1 SyntaxArena.withGuard<A>(body:) <null>:82026728 (SwiftLintFrameworkTests:arm64+0x126b504)
    #2 SyntaxArena.allocateRawSyntaxBuffer(count:) <null>:82026728 (SwiftLintFrameworkTests:arm64+0x126c2fc)
    #3 static RawSyntax.makeLayout(kind:uninitializedCount:arena:initializingWith:) <null>:82026728 (SwiftLintFrameworkTests:arm64+0xffcd14)
    #4 RawSyntaxLayoutView.replacingChild(at:with:arena:) <null>:82026728 (SwiftLintFrameworkTests:arm64+0x10036c8)
    #5 SyntaxData.replacingChild(_:at:) <null>:82026728 (SwiftLintFrameworkTests:arm64+0x1285864)
```

Full report: https://buildkite.com/swiftlint/swiftlint/builds/3411#018354b5-95eb-4c14-ab21-f7d59113f241

Sources/SwiftSyntax/SyntaxArena.swift
```diff
@@ -114,7 +189,9 @@ public class SyntaxArena {
   /// Copies a `RawSyntaxData` to the memory this arena manages, and retuns the
   /// pointer to the destination.
   func intern(_ value: RawSyntaxData) -> UnsafePointer<RawSyntaxData> {
-    let allocated = allocator.allocate(RawSyntaxData.self, count: 1).baseAddress!
+    let allocated = lock.withGuard {
```
ahoppen, line 192: Same here, should this be `self.withGuard`?

Sources/SwiftSyntax/SyntaxArena.swift
```diff
@@ -10,11 +10,60 @@
...
+/// Dummy mutex that doesn't actually guard at all.
+class ScopeGuard {
+  init() {}
+  func withGuard<T>(body: () throws -> T) rethrows -> T {
+    return try body()
+  }
+}
+#endif
+
 public class SyntaxArena {
```
CodaFi, line 58: Can we allocate these locks inline with the arena blocks?

CodaFi (changes requested):
These ScopeGuards are not correct. We cannot use the materialization operator here despite the fact that we're indexing through a heap allocation for the member access - it's still a logical data race even if it's not always a physical data race. These locks need to be allocated out-of-line and the resulting address passed.

CodaFi (approved):

rintaro:
@jpsim I confirmed the failure locally. And the modified implementation resolves it. Could you check the new implementation?

rintaro:
@swift-ci Please test

jpsim (approved):
There are no more TSan failures in my testing. Thank you!

PR #803, opened 2022-09-16
Keep 'SyntaxArena's of rewritten children alive.

rintaro:

PR #808, opened 2022-09-16
SyntaxRewriter: Use an own arena to manage the modified layouts

rintaro:
Remove an usage of SyntaxArena.default


rintaro:
@swift-ci Please test

bnbarham (approved):
If you can think of a good way to test this to make sure we don't accidentally drop the arena that'd be a nice addition. That's more relevant for https://github.com/apple/swift-syntax/pull/803 but since that's already merged... 😅

rintaro:
This possibly causes retain cycle.

PR #810, opened 2022-09-17
Add a public API to look up an infix operator by name.

allevato:
Rather than limit myself in with an API specifically for the precedence group, returning the full `Operator` value seems more flexible/future-proof.

I added analogs for prefix and postfix operators, and then realized when writing tests that the `defaultOperators` set doesn't include those yet! So I stuck with just the infix API and we can revisit prefix/postfix later—it doesn't seem right to offer those public APIs if they don't have the expected behavior when called on `defaultOperators` yet.

### DIFF
--- a/Sources/SwiftOperators/OperatorTable.swift
+++ b/Sources/SwiftOperators/OperatorTable.swift
@@ -83,13 +83,19 @@ public struct OperatorTable {
 }
 
 extension OperatorTable {
+  /// Returns the ``Operator`` corresponding to the given infix operator, or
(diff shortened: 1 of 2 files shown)


allevato:
@swift-ci please test

DougGregor (approved):
Thank you! I agree that we should add prefix/postfix operator lookups once we've updated the standard and logical operator sets.

DougGregor:
I went ahead and added prefix/postfix operators and lookup functions in https://github.com/apple/swift-syntax/pull/819

PR #811, opened 2022-09-18
Build support library

DougGregor:
Create separate module for compiler support (`SwiftCompilerSupport`) that can pull in the various modules whose functionality will be needed by the compiler. We'll stage this in and eventually move the consistency checking code out of `SwiftParser` and into here.

While here, add CMake build rules for the `SwiftOperators` module as well. We'll need it for the compiler.


DougGregor:
@swift-ci please test

PR #812, opened 2022-09-18
Fix loop breaking condition for dictionary literal parsing.

sjavora:
This resolves a FIXME I found in the tests - the loop break previously required the key AND colon AND value to be missing, which I don't think is even possible to satisfy.

### DIFF
--- a/Sources/SwiftParser/Expressions.swift
+++ b/Sources/SwiftParser/Expressions.swift
@@ -1557,7 +1557,7 @@ extension Parser {
   mutating func parseCollectionElement(_ existing: CollectionKind?) -> CollectionKind {
     let key = self.parseExpression()
     switch existing {
-    case .array(_):
+    case .array:
       return .array(key)
     case nil:
       guard self.at(.colon) else {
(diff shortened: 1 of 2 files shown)


Sources/SwiftParser/Expressions.swift
```diff
@@ -1630,7 +1630,7 @@ extension Parser {
             valueExpression: value,
             trailingComma: comma,
             arena: self.arena)))
-          if key.is(RawMissingExprSyntax.self), colon.isMissing, value.is(RawMissingExprSyntax.self) {
+          if key.is(RawMissingExprSyntax.self) || colon.isMissing || value.is(RawMissingExprSyntax.self) {
```
ahoppen, line 1633: I am not sure if this is superior in general. IIUC if you have 
```swift
[
  1: "one",
  2: ,
  3: "three"
]
```
we wouldn’t be stopping dictionary literal parsing at the missing value for `2` and thus not add `3: "three"` to the string literal.

What do `key` and `value` contain when we synthesize the missing `colon`?

sjavora: Thanks for the pointer, I'll try to investigate further.

Interestingly, your example produces _no_ diagnostics for me with the original implementation 😅

ahoppen: Oh, I forgot that my example will only start producing diagnostics once https://github.com/apple/swift-syntax/pull/739 is merged.

ahoppen:
I forgot that I already have a fix for this up as part of one of my PRs: https://github.com/apple/swift-syntax/pull/739/commits/5731968baa69ff84ab36415cae2e5a0b7df2d6f6. Because that fixes the issue, I’m closing this PR. Sorry.

PR #813, opened 2022-09-18
Introduce separate header for SwiftCompilerSupport library.

DougGregor:
This is more staging to move the compiler over to SwiftCompilerSupport.


DougGregor:
@swift-ci please test

PR #826, opened 2022-09-21
Pin SwiftSyntaxBuilder version for generate-swiftsyntaxbuilder 

ahoppen:
This pretty much re-opens https://github.com/apple/swift-syntax/pull/580 on top of current `main`.

This resolves a bootstrapping issue that running `generate-swiftsyntaxbuilder` might modify `SwiftSyntaxBuilder`, which in turn makes it impossible to build and run `generate-swiftsyntaxbuilder`.


ahoppen:
@swift-ci Please test

ahoppen:
https://github.com/apple/swift/pull/61226

@swift-ci Please test

ahoppen:
https://github.com/apple/swift/pull/61226

@swift-ci Please test macOS

ahoppen:
https://github.com/apple/swift/pull/61226

@swift-ci Please test macOS

ahoppen:
https://github.com/apple/swift/pull/61226

@swift-ci Please test Linux

ahoppen:
https://github.com/apple/swift/pull/61226

@swift-ci Please test

build-script.py
```diff
@@ -861,6 +813,31 @@ def add_default_build_arguments(parser: argparse.ArgumentParser) -> None:
...
+                "--gyb-exec",
+                default=GYB_EXEC,
+                help="Path to the gyb tool (default: %(default)s).",
+            )
+
+            parser.add_argument(
+                "--gyb-only",
+                action="store_true",
+                help="""
+                Only generate gyb templates (and not generate-swift-syntax-builder's templates)
```
fwcd, line 833: I know this wasn't a change introduced in this PR, but I think Python doesn't automatically dedent multi-line strings like Swift, so we should probably use something like `.strip()` or `textwrap.dedent` to avoid ending up with strange formatting in the help output.

fwcd (commented):
Looks good, just some minor notes below!

Code-Generation/Package.swift
```diff
@@ -0,0 +1,46 @@
+// swift-tools-version:5.7
+
+import PackageDescription
+
+let package = Package(
+  name: "SwiftSyntaxGeneration",
```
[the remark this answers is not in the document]
ahoppen: Good catch. I renamed it.

build-script.py
ahoppen: Good idea. I’ll set up another PR for it.

ahoppen:
https://github.com/apple/swift/pull/61226

@swift-ci Please test

build-script.py
ahoppen: I just checked and these messages get automatically demented because we use `argparse.RawDescriptionHelpFormatter`.

fwcd: Ah interesting, didn't know about that, nice! 👍

PR #827, opened 2022-09-21
Roll Back ArgumentParser Dependency

CodaFi:
This matches the version number found in the rest of the Swift projects and also matches update-checkout which is pinned to version 1.0.3.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -201,7 +201,7 @@ let package = Package(
 if ProcessInfo.processInfo.environment["SWIFTCI_USE_LOCAL_DEPS"] == nil {
   // Building standalone.
   package.dependencies += [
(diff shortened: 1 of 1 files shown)


CodaFi:
@swift-ci test

jpsim:
Is there a plan to update this? It'd be nice to allow projects that depend on SwiftSyntax and ArgumentParser to be able to be on the latest released versions of both.

CodaFi:
I believe that this repository in particular can probably relax its upper bound on the argument parser to accommodate you. 

My understanding is the general problem is keeping the Windows builders happy a la https://github.com/apple/swift-driver/pull/1106.

jpsim:
Yes either loosening the upper bound or decoupling the CLI targets from the library targets in this package could also work, since most consumers of the SwiftSyntax libraries won’t need its CLI targets.

PR #830, opened 2022-09-22
Move `Format` from SwiftSyntaxBuilder into a separate package

ahoppen:
Depends on https://github.com/apple/swift-syntax/pull/826.

---

This way we can use `Format` to determine whether spaces should be inserted after tokens when creating Fix-Its in the parser.


Code-Generation/Sources/generate-swiftbasicformat/GenerateSwiftBasicFormat.swift
```diff
@@ -0,0 +1,50 @@
...
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import ArgumentParser
+import Foundation
+import SwiftSyntax
+import SwiftSyntaxBuilder
+
+/// SwiftSyntaxBuilder sources to be generated
```
fwcd, line 18:
```suggestion
/// SwiftBasicFormat sources to be generated
```

Code-Generation/Sources/generate-swiftbasicformat/GenerateSwiftBasicFormat.swift
```diff
@@ -0,0 +1,50 @@
...
+private let sourceTemplates: [(SourceFile, String)] = [
+  (formatFile, "Format.swift"),
+]
+
+@main
+struct GenerateSwiftSyntaxBuilder: ParsableCommand {
```
fwcd, line 24:
```suggestion
struct GenerateSwiftBasicFormat: ParsableCommand {
```

Code-Generation/Sources/generate-swiftbasicformat/GenerateSwiftBasicFormat.swift
```diff
@@ -0,0 +1,50 @@
...
+    let generatedURL = URL(fileURLWithPath: generatedPath)
+    let format = Format(indentWidth: 2)
+
+    try FileManager.default.createDirectory(
+      atPath: generatedURL.path,
+      withIntermediateDirectories: true,
+      attributes: nil
+    )
+
+    for (sourceFile, name) in sourceTemplates {
```
fwcd, line 41: Perhaps we could move all of this logic from `main` into a function taking the generation path/url and a `[(SourceFile, String)]` and share it with `generate-swiftsyntaxbuilder`?

ahoppen: Oh, I wanted to do that but forgot about it while I was working on this PR. Moved it to `Utils` now. Thanks!

ahoppen:
@swift-ci Please test

Sources/SwiftBasicFormat/generated/Format.swift
```diff
@@ -35,1865 +35,1865 @@ extension Format {
...
     var result = syntax
     let leadingTrivia = result.leadingTrivia ?? []
     if !leadingTrivia.isEmpty {
       result = result.withLeadingTrivia(leadingTrivia.addingSpacingAfterNewlinesIfNeeded())
     }
     return result
   }
-  func format(syntax: DeclNameArgumentSyntax) -> DeclNameArgumentSyntax {
+  public func format(syntax: DeclNameArgumentSyntax) -> DeclNameArgumentSyntax {
     var result = syntax
```
DougGregor, line 115: This code is so repetitive. Can we not put it into an extension of `SyntaxProtocol` somehow instead of stamping out a hundred copies?

build-script.py
```diff
@@ -343,27 +344,36 @@ def generate_gyb_files(
...
+            target, destination_dir
+        ]
+        
+        if verbose:
+            swiftpm_call.extend(["--verbose"])
+
+        env = dict(os.environ)
+        env["SWIFT_BUILD_SCRIPT_ENVIRONMENT"] = "1"
+        # Tell other projects in the unified build to use local dependencies
+        env["SWIFT_SYNTAX_PARSER_LIB_SEARCH_PATH"] = \
```
DougGregor, line 374: Can we use `SwiftParser` now for the code generation?

DougGregor (approved):

build-script.py
ahoppen: What do you mean by using `SwiftParser` for code generation? We are using `SwiftSyntaxBuilder` and I don’t see a reason why `SwiftSyntaxBuilder` shouldn’t be able to use `SwiftParser` as well.

Sources/SwiftBasicFormat/generated/Format.swift
ahoppen: Could definitely be something we can investigate in a follow-up PR.

build-script.py
DougGregor: I should have been more specific: can we remove the `env["SWIFT_SYNTAX_PARSER_LIB_SEARCH_PATH"]` bit?

ahoppen: Oh, good catch. We can absolutely do that because `CodeGeneration` doesn’t depend on neither the old nor the new parser.

ahoppen:
@swift-ci Please test

PR #834, opened 2022-09-23
Rename `Code-Generation` folder to `CodeGeneration`

fwcd:
To be consistent with the package name, I think this was an oversight in #826.


fwcd:
@swift-ci please test

ahoppen (approved):
Thanks. Looks good to me.

PR #838, opened 2022-09-23
Merge missing token diagnostics if they occur at the same source location

ahoppen:


Sources/SwiftParser/Diagnostics/ParserDiagnosticMessages.swift
```diff
@@ -131,79 +132,184 @@ public struct InvalidIdentifierError: ParserError {
...
+      } else {
+        parts.append(.node(missingNode))
+      }
+    }
+    let partDescriptions = parts.map({ $0.description })
+    if partDescriptions.count > 1 {
+      return "\(partDescriptions[0..<partDescriptions.count - 1].joined(separator: ", ")) and \(partDescriptions.last!)"
     } else {
-      message += " '\(missingToken.text)'"
+      return "\(partDescriptions.joined(separator: ", "))"
```
bnbarham, line 230: This is 0/1 (though would only be 0 if you remove the `!` as in my previous comment). So no need to join here.

Sources/SwiftParser/Diagnostics/ParserDiagnosticMessages.swift
```diff
@@ -131,79 +132,184 @@ public struct InvalidIdentifierError: ParserError {
...
-      case .rightAngle, .rightBrace, .rightParen, .rightSquareBracket:
-        if parent.children(viewMode: .fixedUp).last?.as(TokenSyntax.self) == missingToken {
-          message += " to end \(parentTypeName)"
+      default:
+        break
+      }
+    }
+    if let missingToken = missingNodes.last?.as(TokenSyntax.self), let commonParent = commonParent {
+      switch missingToken.tokenKind {
+      case .rightBrace, .rightAngle, .rightParen, .rightSquareBracket:
```
bnbarham, line 267: Add string quotes here? Should only ever have one if we've started a string, so they're always the end.

Sources/SwiftParser/Diagnostics/ParserDiagnosticMessages.swift
```diff
@@ -131,79 +132,184 @@ public struct InvalidIdentifierError: ParserError {
...
+
+  /// If applicable, returns a string that describes the node in which the missing nodes are expected.
+  private var parentContextClause: String? {
+    enum ContextType {
+      case toStart
+      case `in`
+      case toEnd
+    }
+    var nodePosition = ContextType.in
+
```
bnbarham, line 254: `guard let commonParent = commonParent, let parentTypeName = commonParent.nodeTypeNameForDiagnostics(allowSourceFile: false) else { return nil }`

Sources/SwiftParser/Diagnostics/ParserDiagnosticMessages.swift
```diff
@@ -131,79 +132,184 @@ public struct InvalidIdentifierError: ParserError {
...
-          message += " to end \(parentTypeName)"
+      default:
+        break
+      }
+    }
+    if let missingToken = missingNodes.last?.as(TokenSyntax.self), let commonParent = commonParent {
+      switch missingToken.tokenKind {
+      case .rightBrace, .rightAngle, .rightParen, .rightSquareBracket:
+        if commonParent.children(viewMode: .fixedUp).last?.as(TokenSyntax.self) == missingToken {
+          if nodePosition == .toStart {
```
bnbarham, line 269: `in` is the default, just `if nodePosition != .toStart`?

Sources/SwiftParser/Diagnostics/ParserDiagnosticMessages.swift
```diff
@@ -131,79 +132,184 @@ public struct InvalidIdentifierError: ParserError {
...
+            nodePosition = .toEnd
+          }
         }
       default:
-        message += " in \(parentTypeName)"
+        break
       }
     }
-    return message
+    if let commonParent = commonParent, let parentTypeName = commonParent.nodeTypeNameForDiagnostics(allowSourceFile: false) {
```
bnbarham, line 280: Can remove this check with the guard above (and then all the `commonParent` checks as well)

Sources/SwiftParser/Diagnostics/ParserDiagnosticMessages.swift
```diff
@@ -131,79 +132,184 @@ public struct InvalidIdentifierError: ParserError {
...
+    return message
+  }
+}
+
+public struct MissingAttributeArgument: ParserError {
+  /// The name of the attribute that's missing the argument, without `@`.
+  public let attributeName: TokenSyntax
+
+  public var message: String {
+    return "Expected argument for '@\(attributeName)' attribute"
```
bnbarham, line 310: `argument` or `arguments`?

Sources/SwiftSyntax/Syntax.swift
```diff
@@ -207,6 +207,31 @@ public extension SyntaxProtocol {
...
+      return nil
+    }
+    let siblings = NonNilRawSyntaxChildren(parent, viewMode: viewMode)
+    let nextSiblingIndex = siblings.index(after: self.index)
+    if let previousAbsoluteRawSibling = siblings[nextSiblingIndex...].last {
+      return Syntax(SyntaxData(previousAbsoluteRawSibling, parent: parent))
+    } else {
+      return nil
+    }
+  }
```
bnbarham, line 233: Probably worth adding some quick tests for these - this should be `.first`.

Tests/SwiftParserTest/Declarations.swift
```diff
@@ -319,8 +312,7 @@ final class DeclarationTests: XCTestCase {
       private(#^DIAG^#get, didSet var a = 0
       """,
       diagnostics: [
-        DiagnosticSpec(message: "Expected 'set' in modifier"),
-        DiagnosticSpec(message: "Expected ')' to end modifier"),
+        DiagnosticSpec(message: "Expected 'set)' to end modifier"),
         DiagnosticSpec(message: "Unexpected text 'get, didSet' in variable")
```
bnbarham, line 316: Probably worth specializing this one.

Tests/SwiftParserTest/Declarations.swift
```diff
@@ -702,7 +702,7 @@ final class DeclarationTests: XCTestCase {
       }
       """,
       diagnostics: [
-        DiagnosticSpec(locationMarker: "EXPECTED_DECL", message: "Expected declaration after '{' in struct"),
+        DiagnosticSpec(locationMarker: "EXPECTED_DECL", message: "Expected declaration in struct"),
```
bnbarham, line 705: I *think* I agree with doing this, especially since the location would be the "after" anyway (though I suppose not all IDEs display the actual column).

Tests/SwiftParserTest/Expressions.swift
```diff
@@ -395,8 +395,8 @@ final class ExpressionTests: XCTestCase {
       \#^AFTER_SLASH^#\(#^AFTER_PAREN^#
       """##,
       diagnostics: [
-        DiagnosticSpec(locationMarker: "AFTER_SLASH", message: #"Expected root and expression after '\' in key path"#),
-        DiagnosticSpec(locationMarker: "AFTER_PAREN", message: "Expected value after '(' in tuple"),
+        DiagnosticSpec(locationMarker: "AFTER_SLASH", message: #"Expected root and expression in key path"#),
```
bnbarham, line 398: Is there an issue up for key paths in general? `root` isn't required, but it looks like we're calling `parsePostfixExpression` if there's no `.`, that returns missing, and thus we get an expected "root" here where that makes no sense (the type isn't required, really I'd expect "Expected '.' and expression in key path" here).

Tests/SwiftParserTest/Expressions.swift
```diff
@@ -467,7 +467,7 @@ final class ExpressionTests: XCTestCase {
                 diagnostics: [
                   DiagnosticSpec(message: "Expected identifier in '#keyPath' expression"),
                   DiagnosticSpec(message: "Expected ')' to end '#keyPath' expression"),
-                  DiagnosticSpec(locationMarker: "MISSING_VALUE", message: "Expected value after ':' in function call"),
+                  DiagnosticSpec(locationMarker: "MISSING_VALUE", message: "Expected value in function call"),
```
bnbarham, line 470: Ideally this would be "argument" but that's not particularly easy with the current scheme 🤔.

Tests/SwiftParserTest/Expressions.swift
```diff
@@ -476,7 +476,7 @@ final class ExpressionTests: XCTestCase {
       "[(Int) -> #^DIAG^#throws Int]()",
       diagnostics: [
         // FIXME: We should suggest to move 'throws' in front of '->'
-        DiagnosticSpec(message: "Expected expression after '->' in array element"),
+        DiagnosticSpec(message: "Expected expression in array element"),
```
bnbarham, line 479: Ignoring the throws problem, this not being something like "expected return type" is also an issue.

Tests/SwiftParserTest/Expressions.swift
```diff
@@ -509,10 +509,10 @@ final class ExpressionTests: XCTestCase {
       let #^VAR_NAME^#:(#^DIAG_1^#..)->#^END^#
       """,
       diagnostics: [
-        DiagnosticSpec(locationMarker: "VAR_NAME", message: "Expected pattern after 'let' in variable"),
-        DiagnosticSpec(locationMarker: "DIAG_1", message: "Expected type after '(' in function type"),
+        DiagnosticSpec(locationMarker: "VAR_NAME", message: "Expected pattern in variable"),
+        DiagnosticSpec(locationMarker: "DIAG_1", message: "Expected type in function type"),
         DiagnosticSpec(locationMarker: "DIAG_1", message: "Unexpected text '..' in function type"),
-        DiagnosticSpec(locationMarker: "END", message: "Expected type after '->' in function type"),
+        DiagnosticSpec(locationMarker: "END", message: "Expected type in function type"),
```
bnbarham, line 515: Similar to the above, without `after '->'` this would really need to be `expected return type`.

Sources/SwiftParser/Diagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -131,7 +127,44 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
+        FixIt(message: InsertTokenFixIt(missingToken: token), changes: [
+          .makePresent(node: node)
+        ])
+      ]
+    } else {
+      position = node.endPosition
+      fixIts = []
+    }
+
+    // Find the previous sibling, skipping over siblings that don't actually contain any source text.
```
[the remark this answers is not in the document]
ahoppen: > Can't we just use `.sourceAccurate` here rather than checking for missing?

No, because then we wouldn’t find missing tokens, which are skipped in the `.sourceAccurate` traversal mode.

> Rather than grabbing previous siblings and replacing previous diagnostics, could we instead grab next siblings and just create the one diagnostic at this point? That would avoid looking through all diagnostics each time. IMO we also want to merge the fix-its, which that change would also make easier. 

Good idea. I thought that this would be more difficult than looking behind but it turns out that it’s actually simpler.

> We may want to start splitting these files up as well. Ie. the generation + message for this one could be defined in a `MissingSyntaxDiagnostic.swift` instead (it's a fair amount of code).

Good idea.

Sources/SwiftParser/Diagnostics/ParserDiagnosticMessages.swift
```diff
@@ -131,79 +132,184 @@ public struct InvalidIdentifierError: ParserError {
...
-        message += " after '\(previousToken.text)'"
-      }
-    }
-    if !hasNamedParent {
-      if let parent = missingNode.parent, let parentTypeName = parent.nodeTypeNameForDiagnostics(allowSourceFile: false) {
-        message += " in \(parentTypeName)"
-      }
+private extension String {
+  /// Remove any leading or trailing whitespace
+  func trimmingWhitespaces() -> String {
```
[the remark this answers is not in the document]
ahoppen: Yes, exactly. I wished we had functionality like this in the stdlib but apparently we don’t 😬 

Good catch with spaces vs whitespace

Sources/SwiftParser/Diagnostics/ParserDiagnosticMessages.swift
```diff
@@ -131,79 +132,184 @@ public struct InvalidIdentifierError: ParserError {
...
+    }
+
+    enum Part {
+      case sourceText(String)
+      case tokenWithoutDefaultText(RawTokenKind)
+      case node(Syntax)
+
+      var description: String {
+        switch self {
+        case .sourceText(let content):
```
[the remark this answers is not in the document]
ahoppen: It can’t take a `TokenSyntax` because we are merging these further below. But `[TokenSyntax]` is possible. I changed it to that.

Sources/SwiftParser/Diagnostics/ParserDiagnosticMessages.swift
ahoppen: I changed the entire `if` to a `switch` that covers the cases `0`, `1` and `>1`

Sources/SwiftParser/Diagnostics/ParserDiagnosticMessages.swift
ahoppen: Very good idea 👍

Sources/SwiftParser/Diagnostics/ParserDiagnosticMessages.swift
ahoppen: I refactored this logic quite a bit to make it simpler so it doesn’t use `ContextType` anymore.

Sources/SwiftParser/Diagnostics/ParserDiagnosticMessages.swift
ahoppen: I think based on the way we currently model attributes we don’t really know whether a particular attribute requires a single argument or multiple… Might be something to tackle in a follow-up PR.

Tests/SwiftParserTest/Declarations.swift
ahoppen: What would you expect there?

Sources/SwiftParser/Diagnostics/ParserDiagnosticMessages.swift
bnbarham: `argument(s)` 😅

Tests/SwiftParserTest/Attributes.swift
```diff
@@ -6,16 +6,19 @@ final class AttributeTests: XCTestCase {
   func testMissingArgumentToAttribute() {
     AssertParse(
       """
-      @_dynamicReplacement(#^DIAG_1^#
+      @_dynamicReplacement(#^DIAG^#
       func #^DIAG_2^#test_dynamic_replacement_for2() {
```
bnbarham, line 10: Can remove `DIAG_2`

Tests/SwiftParserTest/Attributes.swift
```diff
@@ -27,12 +30,15 @@ final class AttributeTests: XCTestCase {
...
-        DiagnosticSpec(locationMarker: "DIAG_2", message: "Expected '=' in same type requirement"),
-        DiagnosticSpec(locationMarker: "DIAG_2", message: "Expected right-hand type of same type requirement"),
-        DiagnosticSpec(locationMarker: "DIAG_2", message: "Expected ')' to end attribute"),
-      ]
+        DiagnosticSpec(locationMarker: "DIAG_1", message: "Expected ':' and parameters in '@differentiable' argument", fixIts: ["Insert ':' and parameters"]),
+        DiagnosticSpec(locationMarker: "DIAG_2", message: "Expected '=' and right-hand type in same type requirement", fixIts: ["Insert '=' and right-hand type"]),
+        DiagnosticSpec(locationMarker: "DIAG_2", message: "Expected ')' to end attribute", fixIts: ["Insert ')'"]),
+      ],
+      fixedSource: """
+        @differentiable(reverse wrt: <#syntax#>,where T = <#type#>)
```
bnbarham, line 38: Shouldn't this be 'parameters'? Not much better than 'syntax` though 🤔

Tests/SwiftParserTest/Declarations.swift
```diff
@@ -965,35 +955,35 @@ final class DeclarationTests: XCTestCase {
...
-                  DiagnosticSpec(locationMarker: "DIAG_2", message: "Expected '}' to end class"),
-                ])
+    AssertParse(
+      "#^DIAG_1^#}class C#^DIAG_2^#",
+      diagnostics: [
+        DiagnosticSpec(locationMarker: "DIAG_1", message: "Unexpected text '}' before class"),
+        DiagnosticSpec(locationMarker: "DIAG_2", message: "Expected member block in class"),
+      ],
+      fixedSource: """
+        }class C {}
```
bnbarham, line 965: Shouldn't this be removing the first `}`?

bnbarham (approved):

Sources/SwiftParser/Diagnostics/DiagnosticExtensions.swift
```diff
@@ -0,0 +1,38 @@
+//===--- DiagnosticExtensions.swift ---------------------------------------===//
```
bnbarham, line 1: Need to add all the new files to CMakeLists.txt?

ahoppen:
@swift-ci Please test

PR #840, opened 2022-09-26
Fix parsing of key paths that involve `.<integer literal>`.

DougGregor:


DougGregor:
@swift-ci please test

ahoppen (approved):

DougGregor:
We need a bigger rework of key path literal parsing, but this helps unblock a bunch of testing, so I'll merge it.

PR #842, opened 2022-09-26
A return followed by default has no expression associated with it

DougGregor:
Fixes an issue found while performing valid-parse testing in the compiler test suite


DougGregor:
@swift-ci please test

PR #845, opened 2022-09-26
Translate test case from swift/test/Parse to SwiftParserTest

ahoppen:
This translates the test cases from swift/test/Parse into XCTests that can be run in the swift-syntax repository.

All previously expected errors, warnings and notes are imported as TODOs, while the expected diagnostics are filled with those the parser currently produces.

In a first step, this performs a mechanical 1:1 translation. A second commit in this PR removes TODOs that area obviously semantic. Upcoming commits will need to address the TODOs by
- removing them (e.g. because the diagnostics are semantic and should be reported by ASTGen in the new scheme)
- matching them up with diagnostics that are already produced by the new parser
- Adjusting the new parser to produce the previously expected diagnostics.

Powering through these TODOs will be quite a bit of effort but it’s something that I think we need to do…

### DIFF
--- a/Tests/SwiftParserTest/translated/ActorTests.swift
+++ b/Tests/SwiftParserTest/translated/ActorTests.swift
@@ -1,0 +1,34 @@
+// This test file has been translated from swift/test/Parse/actor.swift
+
+import XCTest
+
+final class ActorTests: XCTestCase {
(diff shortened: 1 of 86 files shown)


ahoppen:
@swift-ci Please test

PR #846, opened 2022-09-26
Don't parse ".identifier" as part of the root of a key path expression

DougGregor:


DougGregor:
@swift-ci please test

DougGregor:
This is the wrong approach. I'm going to rewrite the grammar for key paths.

PR #847, opened 2022-09-27
Lowercase diagnostic and Fix-It messages and remove "after <previous token>" clause from missing node diagnostics

ahoppen:
- Remove "after <previous token>" clause from missing node diagnostics
  - The “after” clause doesn’t seem to be generally useful. We can always add it back in special cases where it’s useful
- Lowercase diagnostic and Fix-It messages
  - This matches the capitalization of messages in the compiler


ahoppen:
@swift-ci Please test

PR #854, opened 2022-09-28
Prepare SwiftSyntaxBuilder to be used in Code-Generation

ahoppen:
- Make all decl, expr, pattern, stmt and type nodes as well as some hand-picked types conform to `ExpressibleByStringInterpolation` by generating the conformances using gyb
- Move SyntaxExpressibleByStringInterpolation to SwiftSyntaxBuilder
  - I think creating syntax nodes from string using interpolation is mostly useful in contexts where you’re building syntax trees in the SwiftSyntaxBuilder style and thus these functions should live in SwiftSyntaxBuilder instead of SwiftParser. It also reduces the amount of code we have in SwiftParser
- Two SwiftSyntaxBuilder adjustments
  - Fix indentation issue if a parsed syntax node with comments is used inside a builder
  - Add calledExpression label to FunctionCallExpr convenience initializer because otherwise initializing a `FunctionCallExpr` from a string is ambiguous whether it should be parsed or call the string


ahoppen:
@swift-ci Please test

ahoppen:
@DougGregor Could you review the changes in 
- Parser.swift
- Patterns.swift
- gyb_syntax_support
- SyntaxExpressibleByStringInterpolationConformances.swift.gyb
- As well as the fact that I’m moving `ExpressibleByStringInterpolation` to `SwiftSyntaxBuilder`

@rintaro Could you take a look at 
- RawSyntaxNodeProtocol.swift
- RawSyntaxNodes.swift.gyb

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

PR #855, opened 2022-09-28
Improve error messages when using SyntaxStringInterpolation with invalid syntax code

ahoppen:
If parsing source text of a `SyntaxStringInterpolation` produced a tree with errors or did not consume all characters in the string literal (and would thus drop text), raise a fatal error.

To make sure parsing errors are displayed at the string literal itself and don’t navigate Xcode to the `SyntaxExpressibleByStringInterpolation` initializer that `fatalError`ed, split the initializer into a throwing variant and a `@_transparent` one that `fatalError`s when the throwing initializer encounters an error. Because `@_transparent` is also transparent in terms of source locations, this will make the crash appear at the start of the string literal.

## Example error messages

Screenshot 2022-09-29 at 14 18 42
Screenshot 2022-09-29 at 14 20 22

### DIFF
--- a/Sources/SwiftSyntaxBuilder/Syntax+StringInterpolation.swift
+++ b/Sources/SwiftSyntaxBuilder/Syntax+StringInterpolation.swift
@@ -1,5 +1,6 @@
 @_spi(RawSyntax) import SwiftSyntax
 @_spi(RawSyntax) import SwiftParser
+import SwiftDiagnostics
 
 /// An individual interpolated syntax node.
 struct InterpolatedSyntaxNode {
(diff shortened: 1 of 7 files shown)


Sources/SwiftSyntaxBuilder/Syntax+StringInterpolation.swift
```diff
@@ -91,24 +92,76 @@ public protocol SyntaxExpressibleByStringInterpolation:
...
 }
 
 extension SyntaxExpressibleByStringInterpolation {
   /// Initialize a syntax node by parsing the contents of the interpolation.
+  /// This function is marked `@_transparent` so that fatalErrors raised here
+  /// are reported at the string literal itself.
+  /// This makes debugging easier because Xcode will jump to the string literal
+  /// that had a parsing error instead of the initializer that raised the `fatalError`
+  @_transparent
   public init(stringInterpolation: SyntaxStringInterpolation) {
```
ahoppen, line 122: Does anyone have any concerns about this `@_transparent` hack? @DougGregor

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

Sources/SwiftSyntaxBuilder/Syntax+StringInterpolation.swift
DougGregor: Can we do this using default arguments with `#file` and `#line` instead? `@_transparent` is a messy feature that we'd prefer not spread much farther.

ahoppen: No, because Xcode’s debugger will still jump to the position where the `fatalError` was called without `@_transparent`. 😬

PR #856, opened 2022-09-28
Use parser integration of SwiftSyntaxBuilder in Code Generation

ahoppen:
This starts using the parser integration in SwiftSyntaxBuilder that was added by https://github.com/apple/swift-syntax/pull/848 in the Code-Generation package. IMO it simplifies a lot of the static boilerplate code generation that was overly verbose with SwiftSyntaxBuilder.

The only changes in the generated files are formatting-related, so there’s no functionality change in SwiftSyntaxBuilder.

### DIFF
--- a/CodeGeneration/Sources/generate-swiftsyntaxbuilder/BuildableCollectionNodesFile.swift
+++ b/CodeGeneration/Sources/generate-swiftsyntaxbuilder/BuildableCollectionNodesFile.swift
@@ -15,6 +15,7 @@ import SwiftSyntax
 import SwiftSyntaxBuilder
 import SyntaxSupport
 import Utils
+import SwiftBasicFormat
 
 let buildableCollectionNodesFile = SourceFile {
   ImportDecl(
@@ -45,11 +46,10 @@ let buildableCollectionNodesFile = SourceFile {
     ) {
       for (side, trivia) in zip(triviaSides, trivias) {
         VariableDecl(
-          leadingTrivia: .docLineComment("/// The \(side) trivia attached to this syntax node once built.") + .newline,
-          .var,
-          name: trivia,
-          type: "Trivia",
-          initializer: ArrayExpr()
+          """
+          /// The \(side) trivia attached to this syntax node once built.
+          var \(trivia): Trivia = []
+          """
         )
       }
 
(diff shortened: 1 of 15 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@fwcd I would like to hear what you think about this change in Code-Generation in general. I don’t think it’s necessary to review every single change.

ahoppen:
@swift-ci Please test

fwcd:
I like it, definitely more readable than before! The only thing I'm a bit worried about is that we lose some type-safety and that parsing might add some overhead to generation time. But since we usually run the generator immediately after compiling, the former should probably not be a big issue (though the IDE of course cannot help us much as before, e.g. also in making sure that we don't accidentally pass a declaration where an expression would be expected etc.).

ahoppen:
I think we still have most of the safety once https://github.com/apple/swift-syntax/pull/855 is merged and we report any errors in the parsed code snippets. I.e. if the code that’s passed to a `FunctionDecl` is not a function, you will get an error. 

The only place where we loose some compile-time safety is when we are performing string interpolation, because e.g. the following compiles but can’t produce valid code
```swift
let variable = VariableDecl("var x: Int")
let func = FunctionDecl("func \(variable)() {}")
```

But even then, parsing of the `FunctionDecl` should fail at generation-time and I think it should be pretty easy to spot the mistake because the `fatalError` will be raised where the `FunctionDecl` is being constructed.

ahoppen:
@swift-ci Please test macOS

CodeGeneration/Sources/generate-swiftsyntaxbuilder/BuildableCollectionNodesFile.swift
```diff
@@ -107,153 +102,80 @@ let buildableCollectionNodesFile = SourceFile {
...
-      }
-    }
   }
+  return InitializerDecl(
+    """
+    /// Creates a `\(type.buildableBaseName)` with the provided list of elements.
+    /// - Parameters:
+    ///   - elements: A list of `\(elementType.expressibleAsBaseName)`
+    public init(_ elements: \(ArrayType(elementType: elementType.expressibleAs))) {
+      self.elements = \(elementsInit)
```
DougGregor, line 118: Annoyingly, this uses implicit opening of existentials, so it fails with older toolchains.

DougGregor: (I have a fix)

PR #859, opened 2022-09-29
Generalize "InVarOrLetPattern" to "PatternContext"

CodaFi:
The legacy parser maintains a stack of contexts that determine whether a matching pattern is being parsed, and whether a let or var is the introducer. This turns on and off a particular production in expression parsing that helps to meld patterns like

`Optional<String>.none`

into shape depending on whether it comes after the keyword "case" or after "let" or "var".


CodaFi:
@swift-ci test

PR #862, opened 2022-09-29
Add `to match this opening <thing>` notes to the diagnostics

ahoppen:
For diagnostics like `expected '}' to end actor` issue a corresponding `to match this opening '{'`

This resolves 38 TODOs, down to 1025.

### DIFF
--- a/Sources/SwiftDiagnostics/Note.swift
+++ b/Sources/SwiftDiagnostics/Note.swift
@@ -1,4 +1,4 @@
-//===--- Diagnostics.swift ------------------------------------------------===//
+//===--- Note.swift -------------------------------------------------------===//
 //
 // This source file is part of the Swift.org open source project
 //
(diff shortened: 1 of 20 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

Tests/SwiftParserTest/Assertions.swift
```diff
@@ -156,6 +226,18 @@ func AssertDiagnostic<T: SyntaxProtocol>(
   if let highlight = spec.highlight {
     AssertStringsEqualWithDiff(diag.highlights.map(\.description).joined(), highlight, file: file, line: line)
   }
+  if let notes = spec.notes {
+    if diag.notes.count != notes.count {
+      XCTFail("""
+      Expected \(notes.count) notes but received \(diag.notes.count):
```
bnbarham, line 232: Could print the expected notes as well?

Tests/SwiftParserTest/Assertions.swift
```diff
@@ -246,9 +328,10 @@ func AssertParse<Node: RawSyntaxNodeProtocol>(
         Expected \(expectedDiagnostics.count) diagnostics but received \(diags.count):
```
bnbarham, line 328: Same comment as for notes, would be nice to print all expected as well.

Tests/SwiftParserTest/Expressions.swift
```diff
@@ -354,10 +360,10 @@ final class ExpressionTests: XCTestCase {
 
     AssertParse(
       ##"""
-      #"#^DIAG^#
+      #^NOTE^##"#^DIAG^#
```
bnbarham, line 363: Not used any more?

Sources/SwiftDiagnostics/Note.swift
```diff
@@ -0,0 +1,68 @@
...
+public struct Note: CustomDebugStringConvertible {
+  /// The node whose location the node is pointing.
+  public let node: Syntax
+
+  /// The position at which the location should be anchored.
+  /// By default, this is the start location of `node`.
+  public let position: AbsolutePosition
+
+  /// A description of what this note is pointing at.
+  public let noteMessage: NoteMessage
```
bnbarham, line 36: I'm still not sure having an ID for note/diagnostic is really worth it. But we can keep it for now 🤷

Tests/SwiftParserTest/Assertions.swift
ahoppen: I explicitly didn’t print the expected notes because the test failure will point to where the `DiagnosticSpec` is declared and the expected notes will be shown there anyway. So I think there is very little value in printing the expected notes. This matches what we are doing for diagnostics as well.

Sources/SwiftDiagnostics/Note.swift
ahoppen: I still think it’s very little overhead at this point and once a use case comes up we will be glad that we have so we don’t need to do string matching for the error message.

Tests/SwiftParserTest/Expressions.swift
ahoppen: https://github.com/apple/swift-syntax/pull/878

Tests/SwiftParserTest/Assertions.swift
bnbarham: I had a comment there too :P. I generally like having everything in the error, makes it easier to attach them to issues and figure out the problem at a glance (rather than going to the test, comparing the output to what's in the test, etc). I don't feel particularly strongly about this though.

ahoppen: I disagree but fair enough. If you want to add it, I won’t oppose you. 😜

PR #865, opened 2022-09-30
Rework key path parsing and grammar

DougGregor:
The key path grammar used for parsing was attempting to be an expression, but that doesn't match The Swift Programming Language documentation or the mental model. Move to a simpler model where a key path is an optional base type (the root type, a `TypeSyntax`) followed by a sequence of key path components. Implement bespoke parsing rules to handle the key path components, rather than trying to customize postfix expression parsing. This fixes a number of issues observed in the Swift test suite.


DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

PR #867, opened 2022-09-30
Don’t visit parts of the syntax tree that don’t contain any `SequenceExpr` in `SequenceFolder`

ahoppen:
Use a recursive bit to record whether a given subtree contains a `SequenceExpr` and skip any parts of the syntax tree that don’t contain `SequenceExpr`s in `SequenceFolder`.


ahoppen:
@swift-ci Please test

PR #868, opened 2022-09-30
Add a member on `SyntaxProtocol` that returns an expression to reconstruct a syntax tree

ahoppen:
My hope is that this can be used to create more substructure assertions in the test suite because currently, writing out the substructure by hand is quite laborious.

For example if I have 
```swift
/// Comment
func foo() -> String
func bar() -> Int
```

parsing this code into a syntax tree and printing `tree.initializer` emits

```swift
SourceFileSyntax(
  statements: CodeBlockItemListSyntax([
    CodeBlockItemSyntax(
      item: FunctionDeclSyntax(
        attributes: nil, 
        modifiers: nil, 
        funcKeyword: .funcKeyword(
          leadingTrivia: [
            .docLineComment("/// Comment"), 
            .newlines(1)
          ], 
          trailingTrivia: .space
        ), 
        identifier: .identifier("foo"), 
        genericParameterClause: nil, 
        signature: FunctionSignatureSyntax(
          input: ParameterClauseSyntax(
            leftParen: .leftParenToken(), 
            parameterList: FunctionParameterListSyntax([]), 
            rightParen: .rightParenToken(trailingTrivia: .space)
          ), 
          asyncOrReasyncKeyword: nil, 
          throwsOrRethrowsKeyword: nil, 
          output: ReturnClauseSyntax(
            arrow: .arrowToken(trailingTrivia: .space), 
            returnType: SimpleTypeIdentifierSyntax(
              name: .identifier("String"), 
              genericArgumentClause: nil
            )
          )
        ), 
        genericWhereClause: nil, 
        body: nil
      ), 
      semicolon: nil, 
      errorTokens: nil
    ), 
    CodeBlockItemSyntax(
      item: FunctionDeclSyntax(
        attributes: nil, 
        modifiers: nil, 
        funcKeyword: .funcKeyword(
          leadingTrivia: .newline, 
          trailingTrivia: .space
        ), 
        identifier: .identifier("bar"), 
        genericParameterClause: nil, 
        signature: FunctionSignatureSyntax(
          input: ParameterClauseSyntax(
            leftParen: .leftParenToken(), 
            parameterList: FunctionParameterListSyntax([]), 
            rightParen: .rightParenToken(trailingTrivia: .space)
          ), 
          asyncOrReasyncKeyword: nil, 
          throwsOrRethrowsKeyword: nil, 
          output: ReturnClauseSyntax(
            arrow: .arrowToken(trailingTrivia: .space), 
            returnType: SimpleTypeIdentifierSyntax(
              name: .identifier("Int"), 
              genericArgumentClause: nil
            )
          )
        ), 
        genericWhereClause: nil, 
        body: nil
      ), 
      semicolon: nil, 
      errorTokens: nil
    )
  ]), 
  eofToken: .eof()
)
```

The idea is that you can pick parts of this syntax tree to create a substructure assertion in `AssertParse`.

### DIFF
--- a/CodeGeneration/Sources/generate-swiftbasicformat/BasicFormatFile.swift
+++ b/CodeGeneration/Sources/generate-swiftbasicformat/BasicFormatFile.swift
@@ -25,7 +25,7 @@ let basicFormatFile = SourceFile {
   ClassDecl(modifiers: [Token.open], identifier: "BasicFormat", inheritanceClause: TypeInheritanceClause { InheritedType(typeName: "SyntaxRewriter") }) {
     VariableDecl("public var indentationLevel: Int = 0")
     VariableDecl("open var indentation: TriviaPiece { .spaces(indentationLevel * 4) }")
-    VariableDecl("private var indentedNewline: Trivia { Trivia(pieces: [.newlines(1), indentation]) }")
+    VariableDecl("public var indentedNewline: Trivia { Trivia(pieces: [.newlines(1), indentation]) }")
     VariableDecl("private var lastRewrittenToken: TokenSyntax?")
 
     for node in SYNTAX_NODES where !node.isBase {
--- a/Package.swift
+++ b/Package.swift
@@ -111,7 +111,7 @@ let package = Package(
     ),
     .target(
       name: "_SwiftSyntaxTestSupport",
-      dependencies: ["SwiftSyntax"]
+      dependencies: ["SwiftBasicFormat", "SwiftSyntax", "SwiftSyntaxBuilder"]
     ),
     .target(
       name: "SwiftParser",
@@ -143,8 +143,8 @@ let package = Package(
     ),
     .executableTarget(
       name: "swift-parser-cli",
-      dependencies: ["SwiftDiagnostics", "SwiftSyntax", "SwiftParser",
-                     "SwiftOperators",
+      dependencies: ["_SwiftSyntaxTestSupport", "SwiftDiagnostics",
+                     "SwiftSyntax", "SwiftParser", "SwiftOperators",
                      .product(name: "ArgumentParser", package: "swift-argument-parser")]
     ),
     .testTarget(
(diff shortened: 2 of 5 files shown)


Tests/SwiftParserTest/SyntaxProtocol+Initializer.swift
```diff
@@ -0,0 +1,167 @@
...
+      }
+    }
+  }
+}
+
+extension SyntaxProtocol {
+  /// Returns a Swift expression that, when parsed, constructs this syntax node
+  /// (or at least an expression that's very close to constructing this node, the addition of a few manual upcast by hand is still needed).
+  /// The intended use case for this is to print a syntax tree and create a substructure assertion from the generated expression.
+  var initializer: String {
```
bnbarham, line 90: If we could get this perfect, maybe we could include it in `swift-parser-test`? Either way it's probably useful to have there as a similar command to `dump-tree`. Not really sure where it should live in that case though.

ahoppen: The problem is that I haven’t found a good way to retrieve a member’s type at runtime that doesn’t require code-generating all that information (which is what I tried to avoid). I found this most useful when writing tests? What would your use cases for integrating it in `swift-parser-test` be?

bnbarham: Just to run `swift-parser-cli` on a file rather than adding a call to `.initializer` in the test, running, and grabbing from there instead.

ahoppen: The way I have used this so far was to use
```
e print(tree.initializer)
```

in the debugger. But I think we can move this to `_SwiftSyntaxTestSupport`, depend `swift-parser-cli` on `_SwiftSyntaxTestSupport` and add a subcommand for that.

ahoppen:
@swift-ci Please test

PR #870, opened 2022-09-30
Disallow multiple statements on the same line

ahoppen:
Before, the new parser allowed multiple declarations on the same line.

This produces a bunch more diagnostics but all of those are at locations at which we already produce diagnostics and where we need to improve recovery anyway.

### DIFF
--- a/Sources/SwiftDiagnostics/FixIt.swift
+++ b/Sources/SwiftDiagnostics/FixIt.swift
@@ -29,6 +29,8 @@ public struct FixIt {
   public enum Change {
     /// Replace `oldNode` by `newNode`.
     case replace(oldNode: Syntax, newNode: Syntax)
+    /// Remove the trailing trivia of the given token
+    case removeTrailingTrivia(TokenSyntax)
   }
 
   /// A description of what this Fix-It performs.
--- a/Sources/SwiftParser/Diagnostics/DiagnosticExtensions.swift
+++ b/Sources/SwiftParser/Diagnostics/DiagnosticExtensions.swift
@@ -36,3 +36,20 @@ extension FixIt.Change {
     )
   }
 }
+
+extension Array where Element == FixIt.Change {
+  /// Makes the `token` present, moving it in front of the previous token's trivia.
+  static func makePresentBeforeTrivia(token: TokenSyntax) -> [FixIt.Change] {
+    if let previousToken = token.previousToken(viewMode: .sourceAccurate) {
+      let presentToken = PresentMaker().visit(token).withTrailingTrivia(previousToken.trailingTrivia)
+      return [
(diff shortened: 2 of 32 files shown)


ahoppen:
@swift-ci Please test

CodaFi:
Can we do this in the diagnostics pass and not the parser?

ahoppen:
No, because the diagnostics generator needs to have a missing token it can diagnose. With https://github.com/apple/swift-syntax/pull/871 we won’t be walking any subtrees that don’t contain missing or unexpected nodes.

ahoppen:
The motivating example for this was that we accepted `func foo() -> Int async {}` as a valid parse by parsing a function declaration without a body `func foo() -> Int` followed by a call with a trailing closure `async {}` on the same line. But we should be rejecting this code.

CodaFi:
>  because the diagnostics generator needs to have a missing token it can diagnose

This seems like an implementation detail. It doesn't change the structure of the parse tree itself to insert this missing semicolon, it's just a post-hoc transformation and a diagnostic. To diagnose this you need only walk code block items and check the same conditions here.

ahoppen:
It might be an implementation detail but it's fundamental to the current diagnostics design. We currently have the following assumptions/invariants. I should probably write them down somewhere. 

1. A syntax tree has the `hasError` recursive bit set iff it contains a missing token or unexpected nodes. This is such a simple rule that I wouldn't want to complicate it with more special cases, like manually setting `hasError` for `CodeBlockItem`a that aren't on a new line. 
2. The diagnostics producer produces no errors iff `hasError` of that subtree is `false`. This allows us to skip subtrees in `ParseDiagnosticsProducer` and make it a lot more efficient. 
3. The `.fixedUp` view of a syntax tree (which contains missing tokens but no unexpected nodes) should represent how the parser interpreted the code and should be the basis of all clients that want a semantic-ish representation of the tree.
4. Rendering the `.fixedUp` view of a source file and parsing it again should result in a tree that's structurally equivalent to the original's tree `.fixedUp` view. (I will try adding an assertion for this to `AssertParse`. 

If we are unwilling to loosen any of these assumptions, the only way we could diagnose the "consecutive statements" issue, would be to diagnose it in ASTGen. But that seems really unfortunate for multiple reasons:
1. ASTGen really shouldn't need to worry about trivia and newlines
2. Adding this diagnostic surfaced parses that we previously missed (like the `func foo() -> Int async {}` case I mentioned above.

CodaFi (approved):
After further discussion, we have come to the conclusion that the semicolon _is_ a part of the grammar and as such a missing semicolon token node is warranted.

ahoppen:
@swift-ci Please test

PR #871, opened 2022-09-30
Skip subtrees that don’t contain any errors during diagnostic generation

ahoppen:
If a subtree doesn’t contain an error, we don’t need to walk it for diagnostics generation.


ahoppen:
@swift-ci Please test

PR #872, opened 2022-09-30
Respect Trailing Commas Better In Collection Element Parsing

CodaFi:
Match the C++ parser in its eagerness to process the next element after it sees a comma. This bug appears to have been masked by the rewrite of `atStartOfStatement`, which I reverted recently.


CodaFi:
@swift-ci test

DougGregor (approved):

PR #873, opened 2022-10-01
Small fixes for regular expression literals.

DougGregor:
Two small fixes for regular expression literals:
* A regular expression literal cannot end in a space or a tab
* A regular expression literal might show up as the name of a function, where it should be an operator. Treat it as an operator.


DougGregor:
@swift-ci please test

PR #877, opened 2022-10-04
Add convenience initializer for ClosureParam

kimdv:
Used in #479


kimdv:
How do I bump the version og the code gen cool? 
Is it a follow up PR?

kimdv:
@swift-ci please test

ahoppen:
You could try using the SwiftParser integration with SwiftSyntaxBuilder in https://github.com/apple/swift-syntax/pull/479, similar to what I‘ve don in https://github.com/apple/swift-syntax/pull/856. My impression is that with the parser integration in SwiftSyntaxBuilder, these convenience initializers are often no longer needed and I find the parser-integrated code a lot easier to read and write. 

If you think you still need this convenience initializer, bumping the CodeGeneration dependency commit would be a follow-up PR.

kimdv:
I did it with the new parser integration! WOW! What a powerfull feature 😍

PR #879, opened 2022-10-05
Remove ElseBlock and ElseIfContinuation nodes

hamishknight:
These don't appear to be used by anything, the if statement node should own these bits of syntax.


hamishknight:
@swift-ci please test

CodaFi (approved):

PR #880, opened 2022-10-05
Remove TODOs for diagnostics that are semantic

ahoppen:
All of the removed TODOs are semantic and should be diagnosed in ASTGen.


ahoppen:
@swift-ci Please test

Tests/SwiftParserTest/translated/UnknownPlatformTests.swift
```diff
@@ -8,10 +8,7 @@ final class UnknownPlatformTests: XCTestCase {
       """
```
bnbarham, line 8: Is it worth having any of the tests in this file? It'd be good to check we parse this correctly, but the diagnostics would be tested in what will likely be a separate if config module.

ahoppen: I could add a few substructure tests once https://github.com/apple/swift-syntax/pull/868 is merged. 😉

bnbarham: Removing both these files and having a structural test for `#if someCall(arg) && def || def2` is what I'd have for the parser test. Or break that down into a bunch 🤷.

ahoppen: Decided to move them to IfconfigExprTests but keep them because why not.

ahoppen:
@swift-ci Please test

PR #881, opened 2022-10-05
Add diagnostic to move 'async' and 'throws' in 'ArrowExpr' in front of '->'

ahoppen:
After parsing the `->` for an `ArrowExpr`, eat trailing `throws` or `async` and include them as unexpected nodes after the `->`. This is safe because the expression after `->` cannot start with `throws` or `async`.

This changes the syntax nodes to allow unexpected tokens after the last expected token. Previously @rintaro had concerns about unexpected nodes at the end of a node because this make it even more ambiguous where the unexpected tokens should be placed, but it’s super convenient for this diagnostic, so allowed it again.

### DIFF
--- a/Sources/SwiftParser/Expressions.swift
+++ b/Sources/SwiftParser/Expressions.swift
@@ -319,17 +319,42 @@ extension Parser {
         return nil
       }
     case (.arrow, _)?, (.throwsKeyword, _)?:
-      let asyncKeyword = self.consumeIfContextualKeyword("async")
+      var asyncKeyword = self.consumeIfContextualKeyword("async")
 
-      let throwsKeyword = self.consume(if: .throwsKeyword)
+      var throwsKeyword = self.consume(if: .throwsKeyword)
+
       let (unexpectedBeforeArrow, arrow) = self.expect(.arrow)
+
+      // Recover if effect modifiers are specified after '->' by eating them into
+      // unexpectedAfterArrow and inserting the corresponding effects modifier as
+      // missing into the ArrowExprSyntax. This reflect the semantics the user
+      // originally intended.
+      var effectModifiersAfterArrow: [RawTokenSyntax] = []
+      if let asyncAfterArrow = self.consumeIfContextualKeyword("async") {
+        effectModifiersAfterArrow.append(asyncAfterArrow)
+        if asyncKeyword == nil {
+          asyncKeyword = missingToken(.contextualKeyword, text: "async")
+        }
+      }
+      if let throwsAfterArrow = self.consume(if: .throwsKeyword) {
+        effectModifiersAfterArrow.append(throwsAfterArrow)
+        if throwsKeyword == nil {
+          throwsKeyword = missingToken(.throwsKeyword, text: nil)
+        }
+      }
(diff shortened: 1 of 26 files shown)


rintaro:
I'm fine with adding `UnexpectedAfterLastChild`

Sources/SwiftParser/Expressions.swift
```diff
@@ -319,17 +319,42 @@ extension Parser {
...
+      var throwsKeyword = self.consume(if: .throwsKeyword)
 
-      let throwsKeyword = self.consume(if: .throwsKeyword)
       let (unexpectedBeforeArrow, arrow) = self.expect(.arrow)
 
+      // Recover if effect modifiers are specified after '->' by eating them into
+      // unexpectedAfterArrow and inserting the corresponding effects modifier as
+      // missing into the ArrowExprSyntax. This reflect the semantics the user
+      // originally intended.
+      var effectModifiersAfterArrow: [RawTokenSyntax] = []
```
bnbarham, line 332: This only handles `throws async` and not eg.`throws throws async`. I've put up a very partial PR for function signatures (https://github.com/apple/swift-syntax/pull/889). Any thoughts on something like that instead? There's quite a few places we'll want to change this 😅 (if we decide this is in fact something the parser should handle).

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
https://github.com/apple/swift/pull/61482

@swift-ci Please test

ahoppen:
https://github.com/apple/swift/pull/61482

@swift-ci Please test

ahoppen:
https://github.com/apple/swift/pull/61482

@swift-ci Please test

ahoppen:
https://github.com/apple/swift/pull/61482

@swift-ci Please test macOS

Tests/SwiftParserTest/translated/TypeExprTests.swift
```diff
@@ -49,11 +34,7 @@ final class TypeExprTests: XCTestCase {
         static func meth() {} 
         func instMeth() {} 
       }
-      """,
-      diagnostics: [
-        // TODO: Old parser expected error on line 5: protocol methods must not have bodies
-        // TODO: Old parser expected error on line 6: protocol methods must not have bodies
-      ]
+      """
```
AnthonyLatsis, line 37: Why was this removed?

ahoppen: SwiftSyntax uses the same nodes for functions inside protocols and structs. Since protocols have optional bodies, function bodies are also optional for functions in in structs in SwiftSyntax. This has two advantages:
- It reduces the nodes we need in SwiftSyntax (we only need a single one and not one for protocols and one for all the other types)
- `.swiftinterface` files also have optional function bodies (even for functions in structs). This way, we can parse `.swiftinterface` files without any problems.

The idea is that the missing function body will be diagnosed by ASTGen further down in the compilation pipeline instead of by the parser.

AnthonyLatsis: I appreciate the detailed clarification!

PR #882, opened 2022-10-05
Use os_unfair_lock on Darwin platforms

rintaro:
https://github.com/apple/swift/pull/61460


rintaro:
PlatformMutex is now gone

PR #885, opened 2022-10-06
Detect and recover from missing type colon

flashspys:
This PR will make the Parser capable of detect and recover from a missing colon in a type declaration. The motivating example was the following:

```
protocol Foo {
    var i Int { get }
}

protocol Bar {
    
}

  --- Diagnostics Before 

 1 │ protocol Foo {
 2 │     var i Int { get }
   ∣           ╰─ unexpected text before protocol
 3 │ }
 4 │
 5 │ protocol Bar {
 6 │
 7 │ }
   ∣  ╰─ expected '}' to end protocol


  --- Diagnostics After
 
 1 │ protocol Foo {
 2 │     var i Int { get }
   ∣          ╰─ expected ':' in type annotation
 3 │ }
 4 │
```

### DIFF
--- a/Sources/SwiftParser/Patterns.swift
+++ b/Sources/SwiftParser/Patterns.swift
@@ -135,6 +135,8 @@ extension Parser {
         arena: self.arena
       )
     } else if lookahead.canParseType() && !self.currentToken.isAtStartOfLine {
+      // TODO: If we are in a for loop, dont do this. normally there an "in" is expected. Siehe failing tests
+        
       // Recovery if the user forgot to add ':'
       let result = self.parseResultType()
       type = RawTypeAnnotationSyntax(
--- a/Sources/SwiftParser/Patterns.swift
+++ b/Sources/SwiftParser/Patterns.swift
@@ -124,28 +124,27 @@ extension Parser {
     let pattern = self.parsePattern()
 
     // Now parse an optional type annotation.
-    guard let colon = self.consume(if: .colon) else {
-      
+    let colon = self.consume(if: .colon)
+    var lookahead = self.lookahead()
+    var type: RawTypeAnnotationSyntax?
+    if let colon = colon {
       let result = self.parseResultType()
-      if result.syntax.isMissingAllTokens {
-        return (pattern, nil)
-      }
-      
-      let type = RawTypeAnnotationSyntax(
-        colon: .init(missing: .colon, arena: self.arena),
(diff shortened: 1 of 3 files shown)


Sources/SwiftParser/Patterns.swift
```diff
@@ -120,20 +120,31 @@ extension Parser {
   /// =======
   ///
   ///     typed-pattern → pattern ':' attributes? inout? type
-  mutating func parseTypedPattern() -> (RawPatternSyntax, RawTypeAnnotationSyntax?) {
+  mutating func parseTypedPattern(allowRecover: Bool = true) -> (RawPatternSyntax, RawTypeAnnotationSyntax?) {
```
ahoppen, line 123: Nitpick: We call this parameter `allowRecovery` in the rest of the codebase. But I think in this case, I would be more specific and call it `allowRecoveryFromMissingColon`.

Sources/SwiftParser/Patterns.swift
```diff
@@ -120,20 +120,31 @@ extension Parser {
...
+    var lookahead = self.lookahead()
+    var type: RawTypeAnnotationSyntax?
+    if let colon = colon {
+      let result = self.parseResultType()
+      type = RawTypeAnnotationSyntax(
+        colon: colon,
+        type: result,
+        arena: self.arena
+      )
+    } else if allowRecover && lookahead.canParseType() && !self.currentToken.isAtStartOfLine {
```
ahoppen, line 137: I would swap the conditions because `!self.currentToken.isAtStartOfLine` is by far the cheaper check.

```suggestion
    } else if allowRecover && !self.currentToken.isAtStartOfLine && lookahead.canParseType() {
```

Sources/SwiftParser/Statements.swift
```diff
@@ -572,7 +572,7 @@ extension Parser {
         type = nil
       }
     } else {
-      (pattern, type) = self.parseTypedPattern()
+      (pattern, type) = self.parseTypedPattern(allowRecover: false)
```
CodaFi, line 575: Why not always recover?

flashspys: For the case the user forgot the `in` in `for item in list {…}` statements we would suggest the user to insert a `:` instead of an `in` because `list` would be assumed as the type.

ahoppen (approved):

ahoppen:
@swift-ci Please test

PR #886, opened 2022-10-06
Detect missing comma in function declaration

flashspys:
This PR will make the Parser capable of detect and recover from a missing comma in a function declaration. The motivating example was the following:

```
func a(foo: Bar foo2: Bar2) {}

 --- Before Diagnostics
 
 1 │ func a(foo: Bar foo2: Bar2) {}
   ∣                 ╰─ unexpected text 'foo2: Bar2' in parameter clause

 
 --- After Diagnostics
 
 1 │ func a(foo: Bar foo2: Bar2) {}
   ∣                 ╰─ expected ',' in function parameter

```

### DIFF
--- a/Sources/SwiftParser/Declarations.swift
+++ b/Sources/SwiftParser/Declarations.swift
@@ -1357,7 +1357,13 @@ extension Parser {
           defaultArgument = nil
         }
 
-        let trailingComma = self.consume(if: .comma)
+        var trailingComma = self.consume(if: .comma)
+        if trailingComma == nil {
+          var lookahead = self.lookahead()
+          if lookahead.consume(if: { $0.canBeArgumentLabel }, followedBy: { $0.tokenKind == .colon }) != nil {
+            trailingComma = missingToken(.comma, text: nil)
+          }
+        }
         keepGoing = trailingComma != nil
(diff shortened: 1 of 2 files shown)


Sources/SwiftParser/Declarations.swift
```diff
@@ -1357,7 +1357,13 @@ extension Parser {
           defaultArgument = nil
         }
 
-        let trailingComma = self.consume(if: .comma)
+        var trailingComma = self.consume(if: .comma)
+        if trailingComma == nil {
+          var lookahead = self.lookahead()
+          if lookahead.consume(if: { $0.canBeArgumentLabel }, followedBy: { $0.tokenKind == .colon }) != nil {
```
ahoppen, line 1363: Hmm, I’m not sure if this is the right approach because it break for anything but the most simple arguments. E.g. we wouldn’t enter this recovery case if the argument has modifiers or a `secondName`. I think you would need to implement a little more sophisticated logic in `Lookahead` and call that from here.

Also adding test a test case for `func foo(first second: Int third fourth: Int)` would be good.

kimdv:
Closing as https://github.com/swiftlang/swift-syntax/pull/3014 is merged

PR #887, opened 2022-10-06
Add new example: Migrate a Swift Codebase to the new if-let-Syntax

flashspys:
We used this piece of code to transform our codebase to the new if let Syntax. I think it is a nice beginner project and would fit as a nice, not too easy, not too hard, example project. I commented the source code as much as possible. Maybe someone has even some suggestions for improvement 😊 I'm for example not sure about the `var conditionCopy = condition`, maybe there is a more clean way to do this.

### DIFF
--- a/Examples/MigrateToNewIfLetSyntax.swift
+++ b/Examples/MigrateToNewIfLetSyntax.swift
@@ -2,11 +2,11 @@ import Foundation
 import SwiftSyntax
 import SwiftParser
 
-/// MigrateToNewIfLetSyntax will visit each if statement in the Syntax tree, and
-/// checks if the there is an if condition which is of the pre Swift 5.7 "if-let-style"
-/// and rewrites it to the new one
+/// MigrateToNewIfLetSyntax will visit each `if` statement in the syntax tree
+/// replacing all "old style" optional bindings by the new shorter syntax available
+/// since Swift 5.7.
 ///
-/// For example will it turn:
+/// For example, it will turn:
 /// ```
 /// if let foo = foo {
 ///   ...
@@ -18,20 +18,20 @@ import SwiftParser
 ///   ...
 /// }
 class MigrateToNewIfLetSyntax: SyntaxRewriter {
-  // Visit over all IfStmtSyntax nodes
+  // Visit all `if` statements.
   override func visit(_ node: IfStmtSyntax) -> StmtSyntax {
-    // For each node, visit all conditions in this node
+    // Visit all conditions in the node.
     let newConditions = node.conditions.enumerated().map { (index, condition) in
       var conditionCopy = condition
-      // check if the condition is a OptionalBindingConditionSyntax...
+      // Check if the condition is an optional binding ...
       if var binding = condition.condition.as(OptionalBindingConditionSyntax.self),
-         // ...and has an initializer
+         // ... and has an initializer ...
          let initializer = binding.initializer,
-         // and the name of the variable pattern and the initializer is the same, ignoring any whitespace
+         // ... and both sides of the assignment are the same identifiers.
          binding.pattern.withoutTrivia().description == initializer.value.withoutTrivia().description {
-        // Remove the initializer
+        // Remove the initializer ... 
         binding.initializer = nil
-        // And remove whitespace (the space before the comma) in if statements with multiple conditions
+        // ... and remove whitespace before the comma (in `if` statements with multiple conditions).
         if index != node.conditions.count - 1 {
           binding.pattern = binding.pattern.withoutTrailingTrivia()
         }


allevato:
If you'd like to contribute this as an opt-in rule for swift-format, that would also be a very welcome contribution!

CodaFi:
This is quite cool! My one comment here is that you might consider checking for an IdentifierPatternSyntax node and an IdentifierExprSyntax node rather than comparing their textual output directly.

Examples/MigrateToNewIfLetSyntax.swift
```diff
@@ -0,0 +1,59 @@
...
+        let fullPath = folderPath + "/" + fileURL
+        let tree = try! Parser.parse(source: String(data: FileManager.default.contents(atPath: fullPath)!, encoding: .utf8 )!)
+        let newTree = MigrateToNewIfLetSyntax().visit(tree)
+        try! newTree.description.write(toFile: fullPath, atomically: true, encoding: .utf8)
+      }
+    }
+  }
+}
+
+MigrateToNewIfLetSyntax.migrate(folderPath: "/path/to/folder")
```
SimplyDanny, line 59: Would be good if this transformation is executable as a command line tool like in the [existing example](https://github.com/apple/swift-syntax/blob/main/Examples/AddOneToIntegerLiterals.swift).

Examples/MigrateToNewIfLetSyntax.swift
```diff
@@ -0,0 +1,59 @@
...
+/// }
+/// ```
+/// into:
+/// ```
+/// if let foo {
+///   ...
+/// }
+class MigrateToNewIfLetSyntax: SyntaxRewriter {
+  // Visit over all IfStmtSyntax nodes
+  override func visit(_ node: IfStmtSyntax) -> StmtSyntax {
```
SimplyDanny, line 22: Only `if` statements are currently considered, but optional bindings are allowed in `guard` and `while` statements as well. For an example, it might be fine to cover only one case and leave the rest up to the motivated user, though. 😉

Shameless plug by the way: There already is such a [rule](https://github.com/realm/SwiftLint/blob/main/Source/SwiftLintFramework/Rules/Idiomatic/IfLetShadowingRule.swift) available in [SwiftLint](https://github.com/realm/SwiftLint).

Examples/MigrateToNewIfLetSyntax.swift
```diff
@@ -0,0 +1,59 @@
...
+    // For each node, visit all conditions in this node
+    let newConditions = node.conditions.enumerated().map { (index, condition) in
+      var conditionCopy = condition
+      // check if the condition is a OptionalBindingConditionSyntax...
+      if var binding = condition.condition.as(OptionalBindingConditionSyntax.self),
+         // ...and has an initializer
+         let initializer = binding.initializer,
+         // and the name of the variable pattern and the initializer is the same, ignoring any whitespace
+         binding.pattern.withoutTrivia().description == initializer.value.withoutTrivia().description {
```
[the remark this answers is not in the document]
flashspys: I see, but it's what Xcode reindent does 😄 One space more or less?

Examples/MigrateToNewIfLetSyntax.swift
flashspys: How are you even supposed to run this example without creating a new package? I tried `swift run ./Example/AddOneToIntegerLiterals.swift` without success because the modules are missing. Is there any way to tell swift that the packages are defined in the current folder?

SimplyDanny: You are right. They cannot be run easily. I prepared #900 to make this possible.

Examples/MigrateToNewIfLetSyntax.swift
```diff
@@ -0,0 +1,59 @@
...
+/// ```
+/// into:
+/// ```
+/// if let foo {
+///   ...
+/// }
+class MigrateToNewIfLetSyntax: SyntaxRewriter {
+  // Visit over all IfStmtSyntax nodes
+  override func visit(_ node: IfStmtSyntax) -> StmtSyntax {
+    // For each node, visit all conditions in this node
```
SimplyDanny, line 23:
```suggestion
    // Visit all conditions in the node.
```

Examples/MigrateToNewIfLetSyntax.swift
```diff
@@ -0,0 +1,59 @@
...
+    let newConditions = node.conditions.enumerated().map { (index, condition) in
+      var conditionCopy = condition
+      // check if the condition is a OptionalBindingConditionSyntax...
```
SimplyDanny, line 26:
```suggestion
      // Check if the condition is an optional binding ...
```

Examples/MigrateToNewIfLetSyntax.swift
```diff
@@ -0,0 +1,59 @@
...
+      if var binding = condition.condition.as(OptionalBindingConditionSyntax.self),
+         // ...and has an initializer
```
SimplyDanny, line 28:
```suggestion
         // ... and has an initializer ...
```

Examples/MigrateToNewIfLetSyntax.swift
```diff
@@ -0,0 +1,59 @@
...
+         let initializer = binding.initializer,
+         // and the name of the variable pattern and the initializer is the same, ignoring any whitespace
```
SimplyDanny, line 30:
```suggestion
         // ... and both sides of the assignment are the same identifiers.
```

Examples/MigrateToNewIfLetSyntax.swift
```diff
@@ -0,0 +1,59 @@
...
+         binding.pattern.withoutTrivia().description == initializer.value.withoutTrivia().description {
+        // Remove the initializer
```
SimplyDanny, line 32:
```suggestion
        // Remove the initializer ... 
```

Examples/MigrateToNewIfLetSyntax.swift
```diff
@@ -0,0 +1,59 @@
...
+        binding.initializer = nil
+        // And remove whitespace (the space before the comma) in if statements with multiple conditions
```
SimplyDanny, line 34:
```suggestion
        // ... and remove whitespace before the comma (in `if` statements with multiple conditions).
```

SimplyDanny (commented):
Allow me to be pedantic and suggest some style and formulation changes to better match the way the existing examples are documented.

Examples/MigrateToNewIfLetSyntax.swift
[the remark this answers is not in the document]
ahoppen: Sorry, I didn’t realize this was part of the `if` condition. Looking at it again, it seems just right.

flashspys:
I applied the suggestions, thank you very much @SimplyDanny. I think its ready to merge now

SimplyDanny (approved):
LGTM! Considering the decision against merging #900, I'm not sure if the `migrate` method should remain in the example, though.

Examples/MigrateToNewIfLetSyntax.swift
flashspys: > For an example, it might be fine to cover only one case and leave the rest up to the motivated user, though.

That was exactly my thought 😉

DougGregor:
@swift-ci please test

PR #889, opened 2022-10-07
DNM: Fix up effect specifiers in function signatures

bnbarham:
In progress, up for feedback/to comment in another PR.

### DIFF
--- a/Sources/SwiftParser/Declarations.swift
+++ b/Sources/SwiftParser/Declarations.swift
@@ -1440,26 +1440,18 @@ extension Parser {
       arena: self.arena)
   }
 
-  /// If a `throws` keyword appears right in front of the `arrow`, it is returned as `misplacedThrowsKeyword` so it can be synthesized in front of the arrow.
-  @_spi(RawSyntax)
-  public mutating func parseFunctionReturnClause() -> (returnClause: RawReturnClauseSyntax, misplacedThrowsKeyword: RawTokenSyntax?) {
-    let (unexpectedBeforeArrow, arrow) = self.expect(.arrow)
-    var misplacedThrowsKeyword: RawTokenSyntax? = nil
-    let unexpectedBeforeReturnType: RawUnexpectedNodesSyntax?
-    if let throwsKeyword = self.consume(ifAny: [.rethrowsKeyword, .throwsKeyword]) {
-      misplacedThrowsKeyword = throwsKeyword
-      unexpectedBeforeReturnType = RawUnexpectedNodesSyntax(elements: [RawSyntax(throwsKeyword)], arena: self.arena)
-    } else {
-      unexpectedBeforeReturnType = nil
-    }
+  /// If effect specifiers appear in front of the `arrow` or after the return type, they are returned so that they can be synthesized in front of the arrow.
+  mutating func parseReturnClause(_ handle: RecoveryConsumptionHandle) -> (returnClause: RawReturnClauseSyntax, misplacedSpecifiers: [SpecifierWithToken]) {
+    let (unexpectedBeforeArrow, arrow) = self.eat(handle)
+    let misplacedSpecifiers = parseAllEffectsSpecifiers()
     let result = self.parseResultType()
     let returnClause = RawReturnClauseSyntax(
       unexpectedBeforeArrow,
       arrow: arrow,
-      unexpectedBeforeReturnType,
+      RawUnexpectedNodesSyntax(misplacedSpecifiers, arena: self.arena),
       returnType: result,
       arena: self.arena)
-    return (returnClause, misplacedThrowsKeyword)
+    return (returnClause, misplacedSpecifiers)
   }
 }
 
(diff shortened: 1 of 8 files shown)


Sources/SwiftSyntax/Raw/RawSyntax.swift
```diff
@@ -774,3 +774,11 @@ extension RawSyntax {
   }
 }
 
+extension RawUnexpectedNodesSyntax {
```
ahoppen, line 777: I just added a similar extension as well but I think this should live in the parser module, because really it’s a parser-specific use case. I created a SyntaxUtils.swift file to hold it.

ahoppen (commented):
Looks good to me in general. I’ve got a few stylist ideas inline

Sources/SwiftParser/Declarations.swift
```diff
@@ -1763,6 +1732,69 @@ extension Parser {
...
+      return (nil, nil, nil)
+    }
+
+    var asyncOrReasync: RawTokenSyntax? = nil
+    var throwsOrRethrows: RawTokenSyntax? = nil
+    var unexpectedTokens = [RawTokenSyntax]()
+
+    // Not a fan of the number of places contextual + async/reasync is checked (or throws/rethrows for that matter). Can we use EffectSpecifier somehow?
+    if firstSpecifier.tokenKind == .contextualKeyword,
+       firstSpecifier.tokenText == "async" || firstSpecifier.tokenText == "reasync" {
```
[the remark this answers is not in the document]
bnbarham: This helps with this one, but not the ones `addMissingEffectSpecifiers`. I was thinking of adding another `init` to `EffectsSpecifier` that takes a `RawSyntaxToken` but... that doesn't seem much better either.

Sources/SwiftSyntax/Raw/RawSyntax.swift
bnbarham: Let me know when you've got that one in. No real rush on this PR. I haven't even done the diagnostics yet :)(

ahoppen: It's in now
https://github.com/apple/swift-syntax/blob/d30f30d95e9854b25ae8f51b6380ebec6fe41cb4/Sources/SwiftParser/SyntaxUtils.swift#L27-L35

Sources/SwiftParser/Declarations.swift
```diff
@@ -1601,9 +1587,11 @@ extension Parser {
 
     let indices = self.parseParameterClause(for: .indices)
 
+    let unexpectedBeforeArrow = RawUnexpectedNodesSyntax(parseAllEffectsSpecifiers(), arena: self.arena)
+
     let result: RawReturnClauseSyntax
-    if self.at(.arrow) {
-      result = self.parseFunctionReturnClause().returnClause
+    if let handle = self.canRecoverTo(.arrow) {
+      result = self.parseReturnClause(handle).returnClause
```
ahoppen, line 1594: You are dropping the effect specifiers here, right?

Sources/SwiftParser/Declarations.swift
```diff
@@ -1884,14 +1838,15 @@ extension Parser {
 
         // Next, parse effects specifiers. While it's only valid to have them
         // on 'get' accessors, we also emit diagnostics if they show up on others.
-        let asyncKeyword: RawTokenSyntax?
-        let throwsKeyword: RawTokenSyntax?
-        if self.at(anyIn: EffectsSpecifier.self) != nil {
-          asyncKeyword = self.parseEffectsSpecifier()
-          throwsKeyword = self.parseEffectsSpecifier()
+        var asyncOrReasync: RawTokenSyntax? = nil
+        var throwsOrRethrows: RawTokenSyntax? = nil
```
ahoppen, line 1842: I think the general paradigm in the rest of the compiler is to leave these uninitialized here 

```suggestion
        var asyncOrReasync: RawTokenSyntax?
        var throwsOrRethrows: RawTokenSyntax?
```

Sources/SwiftParser/Declarations.swift
```diff
@@ -2233,3 +2189,59 @@ extension Parser {
...
+
+  mutating func parseEffectsSpecifiers(atDecl: Bool) -> (asyncOrReasync: RawTokenSyntax?,
+                                                         throwsOrRethrows: RawTokenSyntax?,
+                                                         specifiersAfterThrows: [SpecifierWithToken]){
+    var initialSpecifiers = parseAllEffectsSpecifiers()
+
+    var asyncOrReasync: RawTokenSyntax?
+    var throwsOrRethrows: RawTokenSyntax?
+    if !initialSpecifiers.isEmpty {
+      let first = initialSpecifiers.removeFirst()
```
ahoppen, line 2229:
```suggestion
    if !initialSpecifiers.isEmpty, let first = initialSpecifiers.removeFirst()
```

Sources/SwiftParser/Declarations.swift
```diff
@@ -2233,3 +2189,59 @@ extension Parser {
...
+      if first.specifier == .asyncContextualKeyword || atDecl && first.specifier == .reasyncContextualKeyword {
+        asyncOrReasync = first.token
+
+        if !initialSpecifiers.isEmpty {
+          let second = initialSpecifiers.removeFirst()
```
ahoppen, line 2234:
```suggestion
        if !initialSpecifiers.isEmpty, let second = initialSpecifiers.removeFirst()
```

Sources/SwiftParser/Declarations.swift
```diff
@@ -2233,3 +2189,59 @@ extension Parser {
...
   }
 }
+
+extension Parser {
+  mutating func parseAllEffectsSpecifiers() -> [SpecifierWithToken] {
+    var specifiers = [SpecifierWithToken]()
+    var loopProgress = LoopProgressCondition()
+    while let (specifier, handle) = self.canRecoverTo(anyIn: EffectsSpecifier.self), loopProgress.evaluate(currentToken) {
+      let (unexpected, token) = self.eatKeepUnexpected(handle)
+      specifiers.append(contentsOf: unexpected.map { SpecifierWithToken(specifier: nil, token: $0) })
```
ahoppen, line 2199: I don’t really like how we are considering unexpected tokens as `SpecifierWithToken` here and setting `specifier` to `nil`. 

What do you think about changing `SpecifierWithToken` to 

```swift
struct ParsedEffectsSpecifier {
  let specifier: EffectsSpecifier?
  let unexpected: RawUnexpectedNodesSyntax?
  let token: RawTokenSyntax
 }
```

That way you also don’t need to modify `eat`.

Sources/SwiftParser/Declarations.swift
bnbarham: `removeFirst` isn't optional, so you can't have the let binding in the if. I'm guessing you're thinking of `dropFirst`, but we're modifying the underlying collection here.

Sources/SwiftParser/Declarations.swift
bnbarham: Same here

Sources/SwiftParser/Declarations.swift
bnbarham: Think I had something different here originally, but I can change these to `let` 👍

Sources/SwiftParser/Declarations.swift
bnbarham: Much better name, but I'm not sure I understand the `unexpected` suggestion. The issue here is that they all need to get merged into the single `RawUnexpectedNodeSyntax`.

Sources/SwiftParser/Declarations.swift
bnbarham: Weird that I can't see this in the review. But no, they're added to the return clause itself. They're only returned to be able to make `async/throws` missing when necessary. This is one of the many reasons I'm not a huge fan of this patch (though this particular returning of unexpected nodes was already here).

It really seems like we should just recover + place tokens in unexpected + have ASTGen handle these diagnostics. Though one issue there is that they wouldn't be parsed as contextual keywords if we just did regular recovery vs looked for the effects specifically.

`@unknown` is sort of an example of this, though it currently allows any attribute as the `@unknown` attribute and parses the rest as unexpected (ie. just recovers to the case/default).

ahoppen: Ah, never mind. I missed that `throws` also gets added to the unexpected tokens of the `returnClause`.

Sources/SwiftParser/Declarations.swift
ahoppen: We can add `RawUnexpectedNodeSyntax` as a child of `RawUnexpectedNodeSyntax`. Or, if you don’t like that, maybe it should be 

```swift 
struct ParsedEffectsSpecifier {
  let specifier: EffectsSpecifier?
  let unexpected: [RawTokenSyntax]
  let token: RawTokenSyntax
 }
```

Sources/SwiftParser/Declarations.swift
bnbarham: Yeah, but it's super easy to miss. Hence my problem with doing it this way :)

Sources/SwiftParser/Declarations.swift
bnbarham: Sure, makes sense.

PR #890, opened 2022-10-07
Use emojis to mark locations in test cases instead of the `#^NAME^#` syntax that’s hard to parse for a human

ahoppen:
Instead of the `#^NAME^#` syntax that can obfuscate the test case quite a bit, use the following emojis as markers: 0️⃣, 1️⃣, 2️⃣, 3️⃣, 4️⃣, 5️⃣, 6️⃣, 7️⃣, 8️⃣, 9️⃣, 🔟, ℹ️

If no explicit marker is specified, diagnostics refer to the marker 1️⃣ and notes to ℹ️.

---

For example, the test cases now look as follows

if #availableℹ️(OSX 1️⃣{

instead of 

if #available#^NOTE^#(OSX #^DIAG^#{

which I find a lot more readable.


ahoppen:
@swift-ci Please test

Sources/_SwiftSyntaxTestSupport/LocationMarkers.swift
```diff
@@ -19,31 +19,41 @@ fileprivate func findMarkedRanges(text: String) -> [Marker] {
...
-  let name = text[start.upperBound ..< end.lowerBound]
+  let markerRange = start ..< end
+  let name = text[start ..< end]
 
   // Expand to the whole line if the line only contains the marker
-  let lineRange = text.lineRange(for: start)
+  let lineRange = text.lineRange(for: start..<start)
   if text[lineRange].trimmingCharacters(in: .whitespacesAndNewlines) == text[markerRange] {
     return Marker(name: name, range: lineRange)
   }
```
bnbarham, line 45: The `if` is never true now. If we want to keep `Marker` with a `range` then `extractMarkers` should probably just take either a start/end delimiter *or* a set of characters to use as markers. Then the caller of `extractMarkers` decides what's used.

That would also allow us to just use the subset passed as markers in `AssertParse` (we could still limit it some allowable set though).

bnbarham (commented):
This is so much more readable on some of the messier examples. Thanks! We could likely rework the way "markers" is done (see comment) but that doesn't *really* matter right now.

Unfortunately it seems like various editors don't handle emoji's very well. Ie. Sublime/Vim (or possible iterm2 for the latter) fail to display these correctly. Should we pick something different 🤔?

Sources/_SwiftSyntaxTestSupport/LocationMarkers.swift
ahoppen: I think it can still be true because you can still have a line that only contains an emoji marker.
Also, I would like to keep this a range because we could reasonably decide to have markers again that consist of multiple emojis.


Also, I prefer the current approach of using a static list of markers because the lack of contextuality makes the test cases easier to reason about IMHO.

bnbarham: > I think it can still be true because you can still have a line that only contains an emoji marker.

Ah, I completely misread this. `start` used to be a range but it's now just a single character, so this is no different to before. Though looking at it all again I can't really remember *why* I did this 😅 

> Also, I would like to keep this a range because we could reasonably decide to have markers again that consist of multiple emojis.
> Also, I prefer the current approach of using a static list of markers because the lack of contextuality makes the test cases easier to reason about IMHO.

To be clear, I would also prefer to keep this all a range.

My suggestion was still for a static list, just one passed into `extractMarkers` rather than hardcoded in this file. It doesn't really matter though.

ahoppen: > Ah, I completely misread this. `start` used to be a range but it's now just a single index, so this is no different to before. Though looking at it all again I can't really remember _why_ I did this 😅

I think I remember. It allowed you to write test cases like

```swift
func foo() {
  1️⃣
  var foo = 2
}
```

that got translated to 

```swift
func foo() {
  var foo = 2
}
```

removing the entire marker line. That way you didn’t have to mess up indentation in following lines for the marker. Not sure if it really turned out to be useful though…

bnbarham: Oh I see. Probably more relevant when it's multiple characters 🤷

PR #894, opened 2022-10-07
Add diagnostic for try let and try var

ahoppen:
Add a diagnostic if the user wrote `try let foo = bar()`. 

Also support recovery at of statements (essentially adding `allowRecovery` to `atStartOfStatement`). This required a rewrite of `atStartOfStatement`.

### DIFF
--- a/Tests/SwiftParserTest/translated/TryTests.swift
+++ b/Tests/SwiftParserTest/translated/TryTests.swift
@@ -122,6 +122,17 @@ final class TryTests: XCTestCase {
     )
   }
 
+  func testTry11() {
+    AssertParse(
+      """
+      1️⃣try let singleLet = try foo()
+      """,
+      diagnostics: [
+        DiagnosticSpec(message: "'try' must be placed on the initial value expression", fixIts: ["remove redundant 'try'"]),
+      ], fixedSource: "let singleLet = try foo()"
+    )
+  }
+
   func testTry11a() {
     AssertParse(
       """
--- a/Tests/SwiftParserTest/translated/TryTests.swift
+++ b/Tests/SwiftParserTest/translated/TryTests.swift
@@ -125,38 +125,32 @@ final class TryTests: XCTestCase {
   func testTry11a() {
     AssertParse(
       """
-      try1️⃣ 2️⃣let singleLet = foo()
+      1️⃣try let singleLet = foo()
       """,
       diagnostics: [
-        // TODO: Old parser expected error on line 1: 'try' must be placed on the initial value expression, Fix-It replacements: 1 - 5 = '', 21 - 21 = 'try '
-        DiagnosticSpec(locationMarker: "1️⃣", message: "consecutive statements on a line must be separated by ';'"),
-        DiagnosticSpec(locationMarker: "2️⃣", message: "expected expression in 'try' expression"),
-      ]
+        DiagnosticSpec(message: "'try' must be placed on the initial value expression", fixIts: ["move 'try' after '='"]),
+      ], fixedSource: "let singleLet = try foo()"
     )
   }
 
   func testTry11b() {
     AssertParse(
       """
-      try1️⃣ 2️⃣var singleVar = foo()
+      1️⃣try var singleVar = foo()
       """,
       diagnostics: [
-        // TODO: Old parser expected error on line 1: 'try' must be placed on the initial value expression, Fix-It replacements: 1 - 5 = '', 21 - 21 = 'try '
-        DiagnosticSpec(locationMarker: "1️⃣", message: "consecutive statements on a line must be separated by ';'"),
-        DiagnosticSpec(locationMarker: "2️⃣", message: "expected expression in 'try' expression"),
-      ]
+        DiagnosticSpec(message: "'try' must be placed on the initial value expression", fixIts: ["move 'try' after '='"]),
+      ], fixedSource: "var singleVar = try foo()"
     )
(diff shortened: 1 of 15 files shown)


ahoppen:
@swift-ci Please test

Tests/SwiftParserTest/translated/TryTests.swift
```diff
@@ -125,71 +125,55 @@ final class TryTests: XCTestCase {
...
-      try1️⃣ 2️⃣let singleLet = foo()
+      1️⃣try let singleLet = foo()
       """,
       diagnostics: [
-        // TODO: Old parser expected error on line 1: 'try' must be placed on the initial value expression, Fix-It replacements: 1 - 5 = '', 21 - 21 = 'try '
-        DiagnosticSpec(locationMarker: "1️⃣", message: "consecutive statements on a line must be separated by ';'"),
-        DiagnosticSpec(locationMarker: "2️⃣", message: "expected expression in 'try' expression"),
-      ]
+        DiagnosticSpec(message: "'try' must be placed on the initial value expression", fixIts: ["move 'try' after '='"]),
+      ], fixedSource: "let singleLet = try foo()"
```
bnbarham, line 132: What happens if there's already a `try`? IIUC we'd wrap it again, ie. end up with `try try foo()`?

Tests/SwiftParserTest/translated/TryTests.swift
```diff
@@ -204,98 +188,76 @@ final class TryTests: XCTestCase {
...
-        DiagnosticSpec(locationMarker: "1️⃣", message: "unexpected text 'try' before variable"),
-        // TODO: Old parser expected error on line 3: 'try' must be placed on the initial value expression, Fix-It replacements: 3 - 7 = '', 23 - 23 = 'try '
-        // TODO: Old parser expected error on line 3: call can throw, but errors cannot be thrown out of a property initializer
-        DiagnosticSpec(locationMarker: "2️⃣", message: "unexpected text 'try' before variable"),
-        // TODO: Old parser expected error on line 4: expected declaration
-        DiagnosticSpec(locationMarker: "3️⃣", message: "unexpected text 'try' before function"),
-      ]
+        DiagnosticSpec(locationMarker: "1️⃣", message: "'try' must be placed on the initial value expression", fixIts: ["move 'try' after '='"]),
+        DiagnosticSpec(locationMarker: "2️⃣", message: "'try' must be placed on the initial value expression", fixIts: ["move 'try' after '='"]),
+        DiagnosticSpec(locationMarker: "3️⃣", message: "'try' cannot be used with 'func'"),
```
bnbarham, line 193: Little bit odd, IMO a missing expression makes more sense here. If we keep this then I'd expect `try` to be unexpected and thus removed though (where as it doesn't seem to be below).

Tests/SwiftParserTest/translated/TryTests.swift
```diff
@@ -204,98 +188,76 @@ final class TryTests: XCTestCase {
...
-      func test() throws -> Int {
-        try1️⃣ 2️⃣return
-      }
+      1️⃣try return
       """,
       diagnostics: [
-        // TODO: Old parser expected error on line 7: 'try' cannot be used with 'return'
-        DiagnosticSpec(locationMarker: "1️⃣", message: "consecutive statements on a line must be separated by ';'"),
-        DiagnosticSpec(locationMarker: "2️⃣", message: "expected expression in 'try' expression"),
+        DiagnosticSpec(message: "'try' cannot be used with 'return'"),
```
bnbarham, line 237: Hmmm... I'm not really sure what I'd expect here. This is what we currently do so I suppose it's fine.

This is similar to the `try while` case above. Having it expect `return try` and `while try` sort of makes more sense to me since if someone has written `try` they're probably going to call something that throws. I don't feel very strongly about this though, pretty edge case and removing also seems fine.

Tests/SwiftParserTest/translated/TryTests.swift
ahoppen: Good catch. I changed it to not add a second `try` and added a test case for it.

Tests/SwiftParserTest/translated/TryTests.swift
ahoppen: I think what we are trying to recover from here is that the user assumed that `try` would cover the entire variable assignment and didn't recognize that `try` can only be used on expressions. So, IMHO moving the `try` makes perfect sense to me. What would your suggestion for a diagnostic + fix-its be?

Tests/SwiftParserTest/translated/TryTests.swift
ahoppen: Yeah, not sure if it's really worth having a diagnostic for this but since we have it now, I don't think it does any harm.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

Tests/SwiftParserTest/translated/TryTests.swift
bnbarham: I'm only talking about (3) in this comment. (1) and (2) seem fine to me.

For 3 I'd just expect `expected expression` after the `try`. Though if it was on the same line as the `func` I'd probably expect `unexpected try, did you mean throws?` so... 🤷

ahoppen: Good idea 👍

PR #864, opened 2022-09-30
Fixup 'yield' Parsing

CodaFi:
- Fixup the modeling of multi-yields by adding custom nodes for a yield list 
- Fixup the parsing of yield by plumbing the coroutine context through
- Fixup statement lookahead to take all of this into account.

Fixes #828
Fixes rdar://100248814
Fixes https://github.com/apple/swift-collections/pull/188

PR #895, opened 2022-10-07
Decontextualize multi-'yield' Parsing

CodaFi:
Yield statements support the ability to yield multiple values out of the coroutine context. These are modeled like an argument list in the grammar and parse sort of like one (minus labels). But this means there's a pretty serious grammatical ambiguity with applies

func yield() {}

yield ()

Rather than contextualize the parse, we're going to ignore the contextuality of multi-yields. The swift parser will always unambiguously parse 'yield &' as a yield statement, regardless of context. It will similarly unambiguously parse 'yield (' as an apply of a function called 'yield' regardless of context. We will then rewrite this AST on the way out to ASTGen where we actually have the ability to make context-sensitive decisions about the parse.

Fixes a regression introduced by #864 in the parsing of the macOS 11.3 copy of the standard library's swift interface.

### DIFF
--- a/Sources/SwiftParser/Declarations.swift
+++ b/Sources/SwiftParser/Declarations.swift
@@ -14,7 +14,7 @@
 
 extension TokenConsumer {
   func atStartOfDeclaration(
-    context: Parser.ItemContext? = nil,
+    isAtTopLevel: Bool = false,
     allowRecovery: Bool = false
   ) -> Bool {
     if self.at(anyIn: PoundDeclarationStart.self) != nil {
@@ -58,7 +58,7 @@ extension TokenConsumer {
     if allowRecovery {
       declStartKeyword = subparser.canRecoverTo(
         anyIn: DeclarationStart.self,
-        recoveryPrecedence: context == .topLevel ? nil : .closingBrace
+        recoveryPrecedence: isAtTopLevel ? nil : .closingBrace
       )?.0
     } else {
       declStartKeyword = subparser.at(anyIn: DeclarationStart.self)?.0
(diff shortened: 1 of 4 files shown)


CodaFi:
@swift-ci test

Sources/SwiftParser/Declarations.swift
```diff
@@ -14,7 +14,7 @@
 
 extension TokenConsumer {
   func atStartOfDeclaration(
-    context: Parser.ItemContext? = nil,
```
DougGregor, line 17: I love this context going away!

Sources/SwiftParser/Statements.swift
```diff
@@ -1106,10 +1106,23 @@ extension Parser.Lookahead {
...
+            return true
+          case .leftParen:
+            // "yield (", by contrast, must be disambiguated with additional
+            // context. We always consider it an apply expression of a function
+            // called `yield` for the purposes of the parse.
+            return false
+          default:
+            // "yield" followed by any other token is likely a yield statement
+            // of some singular expression.
+            return true
```
DougGregor, line 1122: Hmm. Should we check whether the next token is on a new line?

```swift
x = yield
print("huh")
```

?

DougGregor (approved):

Sources/SwiftParser/Statements.swift
CodaFi: Ooh, good catch.

CodaFi:
```
yield
print("huh")
```

CodaFi:
@swift-ci test

CodaFi:
@swift-ci test

PR #898, opened 2022-10-08
Detect and recover from missing comma in tupel

flashspys:
This PR will make the Parser capable of detect and recover from a missing comma in tuples. The motivating example was the following:

```
var a: (foo: Foo bar: Bar)

 --- Before Diagnostics
 
 1 │ var a: (foo: Foo bar: Bar)
   ∣                  ╰─ unexpected text 'bar: Bar' in tuple type

 
 --- After Diagnostics
 
 1 │ var a: (foo: Foo bar: Bar)
   ∣                  ╰─ expected ',' in tuple type

```

The recover logic includes some interesting logic for the case `var x: (foo bar)`. In this case we could either recover to a missing comma or a missing colon. I checked if the first part of the tuple is uppercase to decide which one is missing.

### DIFF
--- a/Sources/SwiftParser/Types.swift
+++ b/Sources/SwiftParser/Types.swift
@@ -426,10 +426,36 @@ extension Parser {
           unexpectedBeforeColon = nil
           colon = nil
         }
+        
+        // In the case that the input is "(foo bar)" we have to decide whether we parse it as "(foo: bar)" or "(foo, bar)".
+        // As most people write identifiers lowercase and types capitalized, we decide on the first character of the first token
+        if let first = first,
+            second == nil,
+            colon?.isMissing == true,
+            first.tokenText.description.first?.isUppercase == true {
+          elements.append(RawTupleTypeElementSyntax(
+            inOut: nil,
+            name: nil,
+            secondName: nil,
+            unexpectedBeforeColon,
+            colon: nil,
+            type: RawTypeSyntax(RawSimpleTypeIdentifierSyntax(name: first, genericArgumentClause: nil, arena: self.arena)),
+            ellipsis: nil,
+            initializer: nil,
+            trailingComma: self.missingToken(.comma),
+            arena: self.arena
+          ))
+          keepGoing = true
+          continue
+        }
         // Parse the type annotation.
         let type = self.parseType(misplacedSpecifiers: misplacedSpecifiers)
         let ellipsis = self.currentToken.isEllipsis ? self.consumeAnyToken() : nil
-        let trailingComma = self.consume(if: .comma)
+        var trailingComma = self.consume(if: .comma)
+        if trailingComma == nil && self.withLookahead({ $0.canParseType() }) {
+          // If the next token does not close the tuple, it is very likely the user forgot the comma.
+          trailingComma = self.missingToken(.comma)
+        }
         keepGoing = trailingComma != nil
         elements.append(RawTupleTypeElementSyntax(
             inOut: nil,
--- a/Tests/SwiftParserTest/Types.swift
+++ b/Tests/SwiftParserTest/Types.swift
@@ -358,9 +358,86 @@ final class TypeParameterPackTests: XCTestCase {
     // We allow whitespace between the generic parameter and the '...', this is
     // consistent with regular variadic parameters.
     AssertParse(
+    """
+    func f1<T ...>(_ x: T ...) -> (T ...) {}
+    """)
+  }
+  
+  func testMissingCommaInType() throws {
+    AssertParse(
       """
-      func f1<T ...>(_ x: T ...) -> (T ...) {}
+      var foo: (Int)
       """)
+    
+    AssertParse(
+      """
+      var foo: (Int, Int)
+      """)
+    
+    AssertParse(
+      """
+      var foo: (bar: Int 1️⃣bar2: Int)
+      """,
+      diagnostics: [
+        DiagnosticSpec(message: "expected ',' in tuple type")
+      ])
+    
+    AssertParse(
+      """
+      var foo: (bar: Int 1️⃣Int)
+      """,
+      diagnostics: [
+        DiagnosticSpec(message: "expected ',' in tuple type")
+      ])
+    
+    AssertParse(
+      """
+      var foo: (a 1️⃣Int)
+      """,
+      diagnostics: [
+        DiagnosticSpec(message: "expected ':' in tuple type")
+      ])
+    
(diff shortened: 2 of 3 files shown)


Sources/SwiftParser/Types.swift
```diff
@@ -422,10 +422,37 @@ extension Parser {
...
+            second == nil,
+            colon?.isMissing == true,
+            first.tokenText.description.first?.isUppercase == true {
+          elements.append(RawTupleTypeElementSyntax(
+            inOut: nil,
+            name: nil,
+            secondName: nil,
+            unexpectedBeforeColon,
+            colon: nil,
+            type: RawTypeSyntax(RawSimpleTypeIdentifierSyntax(name: first, genericArgumentClause: nil, arena: self.arena)),
```
flashspys, line 438: In this case we have to "reparse" an already as identifier parsed token as a type. I'm not quite sure if there is a better way.

ahoppen: I don’t see any issues here in wrapping `first` in a `RawSimpleTypeIdentifierSyntax` but because we won’t be parsing generic types like `Array<Int>` correctly here. It would also be worth adding test cases for this, I think.

What I don’t understand here, is why we are wrapping `first` here. For `(foo Bar)` doesn’t this result in two tuple elements, one that only has a label and one that only has a type but no label? I think we should be forming  a single tuple element, that has a name `foo`, missing colon and type `Bar`.

ahoppen (commented):
Thank you. I like the idea of using capitalization to disambiguate type vs identifier 👍

Sources/SwiftParser/Types.swift
flashspys: > I don’t see any issues here in wrapping first in a RawSimpleTypeIdentifierSyntax but because we won’t be parsing generic types like Array<Int> correctly here.

As we visit this code only if first was set above, and this is only the case if `self.lookahead().startsParameterName(false)` is true, first must be an argument label (and thus, always a RawSimpleTypeIdentifierSyntax), this code does not apply for things like `Array<Int>`

> It would also be worth adding test cases for this, I think.

I added some test cases for generic types

> What I don’t understand here, is why we are wrapping first here. For (foo Bar) doesn’t this result in two tuple elements, one that only has a label and one that only has a type but no label?

For `(foo Bar)` we start to look for an argument label (`first`), an optional second argument label (`second`), and a colon. If we found first (`foo`), but no second and no colon, we are in the case where we must decide if `foo` is an identifier and a colon is missing or a type and a comma is missing. In the first case we don't visited the commented code and proceed to parse `Bar` as a type. But if we think a comma is missing we must convert first (already parsed as an argument label) to a type. `Bar` now has to be parsed as a type, too, but as a separate `RawTupleTypeElementSyntax`, thus we cannot proceed with the current loop here, but we have to `continue` the loop and start the "tuple element parsing" on `Bar` again, resulting in two separate `RawTupleTypeElementSyntax`, with  the first one missing a comma. To summarize: Is first uppercase, finish this `RawTupleTypeElementSyntax` directly, if not, proceed this `RawTupleTypeElementSyntax`.

> I think we should be forming a single tuple element, that has a name foo, missing colon and type Bar.

This is exactly what's happening here if this code is not visited, respectively, `foo` is not uppercase.

ahoppen: > As we visit this code only if first was set above, and this is only the case if `self.lookahead().startsParameterName(false)` is true, first must be an argument label (and thus, always a RawSimpleTypeIdentifierSyntax), this code does not apply for things like `Array<Int>`

Ah, I see. That makes sense. 👍 

> > It would also be worth adding test cases for this, I think.
> 
> I added some test cases for generic types

Thank you 🙏 

> For `(foo Bar)` we start to look for an argument label (`first`), an optional second argument label (`second`), and a colon. If we found first (`foo`), but no second and no colon, we are in the case where we must decide if `foo` is an identifier and a colon is missing or a type and a comma is missing. In the first case we don't visited the commented code and proceed to parse `Bar` as a type. But if we think a comma is missing we must convert first (already parsed as an argument label) to a type. `Bar` now has to be parsed as a type, too, but as a separate `RawTupleTypeElementSyntax`, thus we cannot proceed with the current loop here, but we have to `continue` the loop and start the "tuple element parsing" on `Bar` again, resulting in two separate `RawTupleTypeElementSyntax`, with the first one missing a comma. To summarize: Is first uppercase, finish this `RawTupleTypeElementSyntax` directly, if not, proceed this `RawTupleTypeElementSyntax`. 

Oh, I know what my misunderstanding was. I thought we were parsing tuples in patterns (as in `let (a, b) = foo()`) here, but this is for tuple types. In that case everything makes sense 👍

Sources/SwiftParser/Types.swift
```diff
@@ -426,10 +426,37 @@ extension Parser {
...
+          keepGoing = true
+          continue
+        }
         // Parse the type annotation.
         let type = self.parseType(misplacedSpecifiers: misplacedSpecifiers)
         let ellipsis = self.currentToken.isEllipsis ? self.consumeAnyToken() : nil
-        let trailingComma = self.consume(if: .comma)
+        var trailingComma = self.consume(if: .comma)
+        var lookahead = self.lookahead()
+        if trailingComma == nil && lookahead.canParseType() {
```
ahoppen, line 456: We’ve got a nice convenience function now that can simplify this:

```suggestion
        if trailingComma == nil && self.withLookahead({ $0.canParseType() }) {
```

Tests/SwiftParserTest/Types.swift
```diff
@@ -355,12 +355,89 @@ final class TypeParameterPackTests: XCTestCase {
...
   }
   func testParameterPacks28() throws {
-    // We allow whitespace between the generic parameter and the '...', this is
-    // consistent with regular variadic parameters.
+      // We allow whitespace between the generic parameter and the '...', this is
+      // consistent with regular variadic parameters.
+      AssertParse(
+    """
+    func f1<T ...>(_ x: T ...) -> (T ...) {}
+    """)
```
ahoppen, line 363: Indentation seems off here.

ahoppen (commented):
I’ve got two more minor comments inline. Also: Could you squash your commits?

Tests/SwiftParserTest/Types.swift
flashspys: Indeed, also the indention below. I changed it.

ahoppen:
@swift-ci Please test

PR #892, opened 2022-10-07
Add diagnostic if ':' is used in a typealias declaration instead of '='

ahoppen:

PR #899, opened 2022-10-08
Replace 'value' by 'type' as name for diagnostics in TypeInitializerClause

ahoppen:
As suggested in https://github.com/apple/swift-syntax/pull/892#discussion_r990358572.


ahoppen:
@swift-ci Please test

PR #902, opened 2022-10-09
Ensure that we handle all modifiers at the beginning of a declaration.

DougGregor:
We weren't handling contextual keywords as modifiers, nor properly dealing with those that don't have parenthesized arguments.

### DIFF
--- a/Sources/SwiftParser/Declarations.swift
+++ b/Sources/SwiftParser/Declarations.swift
@@ -12,6 +12,22 @@
 
 @_spi(RawSyntax) import SwiftSyntax
 
+extension DeclarationModifier {
+  var canHaveParenthesizedArgument: Bool {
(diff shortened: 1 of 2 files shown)


DougGregor:
@swift-ci please test

Sources/SwiftParser/Declarations.swift
```diff
@@ -30,13 +46,15 @@ extension TokenConsumer {
       _ = subparser.consumeAttributeList()
     }
 
-    if subparser.currentToken.isKeyword {
+    if subparser.currentToken.isKeyword ||
+        subparser.currentToken.tokenKind == .identifier {
```
ahoppen, line 50: Do we even need this `if` condition? AFAICT it should be handled by the `while` condition already.

PR #903, opened 2022-10-09
Diagnose if multiple declarations are on the same line

ahoppen:
Previously, we only diagnosed if two statements were on the same line. Do the same check for declarations now.

### DIFF
--- a/Sources/SwiftParser/Declarations.swift
+++ b/Sources/SwiftParser/Declarations.swift
@@ -150,17 +150,23 @@ extension Parser {
   public mutating func parseDeclaration() -> RawDeclSyntax {
     switch self.at(anyIn: PoundDeclarationStart.self) {
     case (.poundIfKeyword, _)?:
-      return RawDeclSyntax(self.parsePoundIfDirective { parser in
+      let directive = self.parsePoundIfDirective { parser in
         let parsedDecl = parser.parseDeclaration()
         let semicolon = parser.consume(if: .semicolon)
         return RawMemberDeclListItemSyntax(
           decl: parsedDecl,
           semicolon: semicolon,
           arena: parser.arena)
-      }
-                           syntax: { parser, elements in
+      } addSemicolonIfNeeded: { lastElement, newItemAtStartOfLine, parser in
+        if lastElement.semicolon == nil && !newItemAtStartOfLine {
+          return lastElement.withSemicolon(parser.missingToken(.semicolon, text: nil), arena: parser.arena)
+        } else {
+          return nil
+        }
+      } syntax: { parser, elements in
         return RawSyntax(RawMemberDeclListSyntax(elements: elements, arena: parser.arena))
-      })
+      }
+      return RawDeclSyntax(directive)
     case (.poundWarningKeyword, _)?, (.poundErrorKeyword, _)?:
       return self.parsePoundDiagnosticDeclaration()
     case nil:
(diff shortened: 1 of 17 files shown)


ahoppen:
@swift-ci Please test

Sources/SwiftParser/Declarations.swift
```diff
@@ -544,26 +550,48 @@ extension Parser {
...
+  public mutating func parseMemberDeclListItem() -> RawMemberDeclListItemSyntax? {
+    let decl: RawDeclSyntax
+    if self.at(.poundSourceLocationKeyword) {
+      decl = RawDeclSyntax(self.parsePoundSourceLocationDirective())
+    } else {
+      decl = self.parseDeclaration()
+    }
+
+    let semi = self.consume(if: .semicolon)
+    var trailingSemas: [RawTokenSyntax] = []
```
bnbarham, line 563: `Sema` -> `Semi`?

Could also just have this array and then use `trailingSemi.first` + `trailingSemi.dropFirst` below.

Sources/SwiftParser/TopLevel.swift
```diff
@@ -127,11 +119,21 @@ extension Parser {
     // "statements" and not "items".
     let item = self.parseItem(isAtTopLevel: isAtTopLevel)
     let semi = self.consume(if: .semicolon)
+    var trailingSemas: [RawTokenSyntax] = []
+    while let trailingSema = self.consume(if: .semicolon) {
+      trailingSemas.append(trailingSema)
+    }
```
bnbarham, line 125: Same thing as above here.

Tests/SwiftParserTest/translated/InitDeinitTests.swift
```diff
@@ -68,21 +68,21 @@ final class InitDeinitTests: XCTestCase {
...
+        deinit 3️⃣x()
       }
       """,
       diagnostics: [
         // TODO: Old parser expected error on line 2: expected '{' for deinitializer
         // TODO: Old parser expected error on line 3: deinitializers cannot have a name, Fix-It replacements: 10 - 12 = ''
         // TODO: Old parser expected error on line 3: expected '{' for deinitializer
-        DiagnosticSpec(locationMarker: "1️⃣", message: "unexpected text 'x' before deinitializer"),
+        DiagnosticSpec(locationMarker: "1️⃣", message: "consecutive declarations on a line must be separated by ';'"),
+        DiagnosticSpec(locationMarker: "2️⃣", message: "unexpected text 'x' before deinitializer"),
```
bnbarham, line 81: Are we missing marking `x` as handled by (1)?

Tests/SwiftParserTest/translated/TrailingSemiTests.swift
```diff
@@ -97,4 +88,84 @@ final class TrailingSemiTests: XCTestCase {
     )
   }
 
+  func testSingleTrailingSemaInStmt() {
```
bnbarham, line 91: Is the idea for the `translated` tests to eventually move them out and have them as just regular tests (possibly with all the methods renamed from `blah<Num>`?

Also `Sema` -> `Semi` again 😅 (this and the two below).

Sources/SwiftParser/Declarations.swift
ahoppen: Good catch with the typos. I prefer to consume the expected semicolon separately from the other ones though. Just a separation of normal parsing and recovery logic.

Tests/SwiftParserTest/translated/InitDeinitTests.swift
ahoppen: No, parsing of the `deinit` with the (1) marker ends before `x` and thus `x` is being considered unexpected text before the next `deinit`.

Tests/SwiftParserTest/translated/TrailingSemiTests.swift
ahoppen: Realistically, I don’t think anyone will ever have so much time that they want to rename all these tests cases. Whether we want to move them out of the `translated` folder, I’m not sure yet. There does seem to be value in indicating that these weren’t specifically crafted for the new parser. But then, I’m adding new tests to these files now, so maybe they really belong to the top-level once they have been audited. Maybe worth discussing once everything has been audited.

Tests/SwiftParserTest/translated/InitDeinitTests.swift
bnbarham: (1) is what I'm finding confusing here. If `x` is being treated as unexpected text then we shouldn't have two decls on the same line, so why do we have that diagnostic?

PR #904, opened 2022-10-09
Improve diagnostic if an editor placeholder is used to represent a closure

ahoppen:
If an editor placeholder is used where a closure is expected, parse it as a closure with missing braces and the editor placeholder as its only statement. Detect that pattern in the diagnostics generator and emit a corresponding diagnostic.

### DIFF
--- a/Sources/SwiftParser/Diagnostics/Utils.swift
+++ b/Sources/SwiftParser/Diagnostics/Utils.swift
@@ -22,3 +22,14 @@ extension String {
     return String(result)
   }
 }
+
+extension Collection {
+  /// If the collection contains a single element, return it, otherwise `nil`.
+  var only: Element? {
+    if self.count == 1 {
+      return self.first!
+    } else {
+      return nil
+    }
+  }
+}
--- a/Sources/SwiftParser/Lexer.swift
+++ b/Sources/SwiftParser/Lexer.swift
@@ -51,6 +51,12 @@ public struct Lexer {
       return self.flags.contains(.isMultilineStringLiteral)
     }
 
+    var isEditorPlaceholder: Bool {
+      return self.tokenKind == .identifier &&
+      self.tokenText.starts(with: SyntaxText("<#")) &&
+      self.tokenText.hasSuffix(SyntaxText("#>"))
+    }
+
     @_spi(RawSyntax)
     public init(
       tokenKind: RawTokenKind,
--- a/Tests/SwiftParserTest/translated/TrailingClosuresTests.swift
+++ b/Tests/SwiftParserTest/translated/TrailingClosuresTests.swift
@@ -1,5 +1,6 @@
 // This test file has been translated from swift/test/Parse/trailing_closures.swift
 
+import SwiftSyntax
 import XCTest
 
 final class TrailingClosuresTests: XCTestCase {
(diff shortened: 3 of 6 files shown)


ahoppen:
@swift-ci Please test

Sources/SwiftParser/Diagnostics/Utils.swift
```diff
@@ -22,3 +22,14 @@ extension String {
...
 }
+
+extension Collection {
+  /// If the collection contains a single element, return it, otherwise `nil`.
+  var only: Element? {
+    if self.count == 1 {
+      return self.first!
+    } else {
+      return nil
+    }
```
bnbarham, line 33: Ah yes, another `.only`. `if !isEmpty && index(after: startIndex) == endIndex` could be slightly better than doing a count for some collections (I think we discuss this every time there's a new `.only` 😆), though for all the ones we'll use internally here `count` is fine 🤷

Sources/SwiftParser/Lexer.swift
```diff
@@ -51,6 +51,12 @@ public struct Lexer {
       return self.flags.contains(.isMultilineStringLiteral)
     }
 
+    var isEditorPlaceholder: Bool {
```
bnbarham, line 54: `parseIdentifierExpression` has this exact check as well. Could use this there.

There's also a bunch of others (differing slightly):
  - `SyntaxClassifier.contextFreeClassification`
  - `Lexer.lexOperatorIdentifier` and `Lexer.tryLexEditorPlaceholder`
  - `SyntaxText.containsPlaceholderEnd`

Would be nice not to have so many 😅

Tests/SwiftParserTest/translated/TrailingClosuresTests.swift
```diff
@@ -145,30 +146,44 @@ final class TrailingClosuresTests: XCTestCase {
     )
   }
 
-  func testTrailingClosures13() {
+  func testTrailingClosures13a() {
```
bnbarham, line 149: Would be nice to actually give these names rather than just `a` and `b`

Sources/SwiftParser/Diagnostics/Utils.swift
ahoppen: Yes, it just happens to always be useful in all sorts of contexts… Good suggestion with the performance improvement though. I haven’t been part of any of these discussions yet. 😆

Sources/SwiftParser/Lexer.swift
ahoppen: The problem is that all of these work on slightly different types/have slightly different semantics:

- This implementation: Full text match on `Lexer.Lexeme`
- `SyntaxClassifier.contextFreeClassification`: Full text match on `RawTokenKind` + `SyntaxText`
- `Lexer.lexOperatorIdentifier` and `Lexer.tryLexEditorPlaceholder`: Substring match on `UInt8`
- `SyntaxText.containsPlaceholderEnd`: Implementation detail of the above

I tried unifying them but it isn’t really possible 😬

Tests/SwiftParserTest/translated/TrailingClosuresTests.swift
ahoppen: Yes, that would be nice. But most of the time, I wouldn’t even know how to name these tests (this one is easy, but a lot of the others are not…)

Sources/SwiftParser/Diagnostics/Utils.swift
bnbarham: It would only matter in cases where `count` is actually expensive, so not here. But thought I'd mention anyway :).

Sources/SwiftParser/Lexer.swift
bnbarham: Could at least unify `parseIdentifierExpression`

ahoppen: What would you like to unify `parseIdentifierExpression` with?

bnbarham: This one. They both use `SyntaxText` -
`parseIdentifierExpression`:
```
name.tokenText.hasPrefix("<#") && name.tokenText.hasSuffix("#>")
```
`isEditorPlaceholder`:
```
self.tokenText.starts(with: SyntaxText("<#")) &&
       self.tokenText.hasSuffix(SyntaxText("#>"))
```

`SyntaxText.isEditorPlaceholder`? 🤷

ahoppen: Ah, of course. Will do.

PR #905, opened 2022-10-09
Remove handled/semantic TODOs from TopLevelLibraryInvalidTests

ahoppen:
Merge ToplevelLibraryInvalidTests.swift and TopLevelLibraryTests.swift and remove semantic TODOs.


ahoppen:
@swift-ci Please test

Tests/SwiftParserTest/translated/ToplevelLibraryTests.swift
```diff
@@ -12,4 +12,28 @@ final class ToplevelLibraryTests: XCTestCase {
     )
   }
 
+  func testToplevelLibraryInvalid1() {
```
bnbarham, line 15: I'd be fine with just deleting this file as well. It doesn't seem particularly relevant for parsing and we should be testing all these elsewhere anyway.

PR #906, opened 2022-10-10
Only diagnose missing semicolons if the terminated node contains no errors

ahoppen:
If the item contains errors, the root cause is most likely something different and not the missing semicolon.


ahoppen:
@swift-ci Please test

PR #908, opened 2022-10-10
Remove `print-init` subcommand from `swift-parser-cli`

ahoppen:
`debugInitCall`, which is the backbone of `swift-parser-cli` is defined in `_SwiftSyntaxTestSupport`, which links against XCTest. If `swift-parser-cli` links against `_SwiftSyntaxTestSupport`, it also links against `XCTest`, causing its execution to fail with a dyld error outside of Xcode.

The solution would be to move `debugInitCall` into a new module that `swift-parser-cli` can link against but IMHO the `print-init` subcommand doesn’t provide sufficent value for that since you most likely want to use it in test cases where you can use a debugger command to get the `debugInitCall` result: `e print(tree.debugInitCall)`.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -142,8 +142,7 @@ let package = Package(
     ),
     .executableTarget(
       name: "swift-parser-cli",
-      dependencies: ["_SwiftSyntaxTestSupport", "SwiftDiagnostics",
-                     "SwiftSyntax", "SwiftParser", "SwiftOperators",
+      dependencies: ["SwiftDiagnostics", "SwiftSyntax", "SwiftParser", "SwiftOperators",
                      .product(name: "ArgumentParser", package: "swift-argument-parser")]
     ),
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test macOS

ahoppen:
@swift-ci Please test

bnbarham:
Unfortunate. I'd still be fine with it being a part of `SwiftSyntax` (behind SPI if you really care) - it seems like a generally useful function to me 🤷.

ahoppen:
It can’t be part of SwiftSyntax because it uses SwiftSyntaxBuilder. We could add it to the SwiftSyntaxBuilder module but that also really doesn’t seem like the right place to put it.

PR #909, opened 2022-10-10
Merge more nodes in “expected …” diagnostics

ahoppen:


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

Sources/SwiftParser/Diagnostics/MissingNodesError.swift
```diff
@@ -137,17 +140,42 @@ fileprivate extension TokenKind {
   }
 }
 
+/// Checks whether a node contains any tokens (missing or present)
+fileprivate class HasTokenChecker: SyntaxAnyVisitor {
```
bnbarham, line 144: We really need to define a better visitor interface for all this. Ie. one that returns a result + has stop.

Sources/SwiftParser/Diagnostics/MissingNodesError.swift
```diff
@@ -233,19 +260,30 @@ extension ParseDiagnosticsGenerator {
...
     }
 
+    /// Ancestors that don't contain any tokens are not very interesting to merge diagnostics (because there can't be any missing tokens we can merge them with).
+    /// Find the first ancestor that contains any tokens.
+    var ancestorWithTokens = node.parent
+    var index = node.index
+    while let unwrappedParent = ancestorWithTokens, !unwrappedParent.hasTokens {
+      ancestorWithTokens = unwrappedParent.parent
+      index = unwrappedParent.index
+    }
```
bnbarham, line 270: This is almost `ancestorOrSelf { $0.hasTokens }` but it gives the index of the child that you came from. Could add that to `ancestorOrSelf`?

Sources/SwiftParser/Diagnostics/ParserDiagnosticMessages.swift
```diff
@@ -153,9 +153,9 @@ public struct UnexpectedNodesError: ParserError {
       message += " '\(shortContent)'"
     }
     if let parent = unexpectedNodes.parent {
-      if let parentTypeName = parent.nodeTypeNameForDiagnostics, parent.children(viewMode: .sourceAccurate).first?.id == unexpectedNodes.id {
+      if let parentTypeName = parent.nodeTypeNameForDiagnostics(allowBlockNames: false), parent.children(viewMode: .sourceAccurate).first?.id == unexpectedNodes.id {
         message += " before \(parentTypeName)"
-      } else if let parentTypeName = parent.ancestorOrSelf(where: { $0.nodeTypeNameForDiagnostics != nil })?.nodeTypeNameForDiagnostics {
+      } else if let parentTypeName = parent.ancestorOrSelf(where: { $0.nodeTypeNameForDiagnostics(allowBlockNames: false) != nil })?.nodeTypeNameForDiagnostics(allowBlockNames: false) {
```
bnbarham, line 158: I wouldn't be against pulling out this into a `firstNodeTypeNameForDiagnostics`, we do this in a bunch of places.

Tests/SwiftParserTest/translated/DiagnoseDynamicReplacementTests.swift
```diff
@@ -21,9 +21,7 @@ final class DiagnoseDynamicReplacementTests: XCTestCase {
       """,
       diagnostics: [
         // TODO: Old parser expected error on line 1: expected '(' in '_dynamicReplacement' attribute
-        DiagnosticSpec(message: "expected '(' in attribute"),
-        DiagnosticSpec(message: "expected argument for '@_dynamicReplacement' attribute"),
-        DiagnosticSpec(message: "expected ')' to end attribute"),
+        DiagnosticSpec(message: "expected '(', attribute argument and ')' in attribute"),
```
bnbarham, line 24: Poor oxford comma :(. Am I the only one that likes it?

Tests/SwiftParserTest/Statements.swift
```diff
@@ -45,8 +45,7 @@ final class StatementTests: XCTestCase {
       }
       """,
       diagnostics: [
-        DiagnosticSpec(message: "expected expression in pattern"),
-        DiagnosticSpec(message: "expected '=' and expression in pattern matching"),
+        DiagnosticSpec(message: "expected expression, '=' and expression in pattern matching"),
```
bnbarham, line 48: Might be worth adding some more fix-it checks? Is this one`<#expression#> = <#expression#>`?

Sources/SwiftParser/Diagnostics/MissingNodesError.swift
ahoppen: I thought about it but decided against it because I think the index is not really useful in any other place and it would make `ancestorOfSelf` harder to use (you‘d need to extract a value from the tuple `ancestorOrSelf` would now return).

Sources/SwiftParser/Diagnostics/ParserDiagnosticMessages.swift
ahoppen: I just improved this to

```swift  
let parentTypeName = parent.ancestorOrSelf(mapping: { $0.nodeTypeNameForDiagnostics(allowBlockNames: false) })
```

in one of the commits I created locally today (and which I’ll create a PR for in the next couple of days). I think the new call is sufficiently simple that we don’t need to create a new method for it. WDYT?

Tests/SwiftParserTest/translated/DiagnoseDynamicReplacementTests.swift
ahoppen: I don’t have strong opinions and am not a native speaker, so I don’t know where people prefer it and where they don’t. If you want an Oxford comma, we can add it. It’s super trivial to change.

Tests/SwiftParserTest/Statements.swift
ahoppen: I think I didn’t bother adding these checks for this test case because it’s so broken that we’ll have a hard time producing a good diagnostic (let alone Fix-Its). I don’t even know what I would expect here. I can add more checks if you prefer though.

Sources/SwiftParser/Diagnostics/ParserDiagnosticMessages.swift
bnbarham: Sure, makes sense.

Tests/SwiftParserTest/Statements.swift
bnbarham: I didn't really mean this case specifically. Just that you added merging for more, but no checks that the generated fix-it in these cases made any sense. I suspect it should since it hasn't really change, would just be nice to make sure.

I was mostly wondering whether we do in fact use editor placeholders for "expected expression".

Tests/SwiftParserTest/translated/DiagnoseDynamicReplacementTests.swift
bnbarham: It's basically personal preference at this point honestly. Some people prefer it, some don't 🤷. *Personally* I think it makes these being a list more clear, not sure if others feel that way.

PR #910, opened 2022-10-10
Diagnose ';' statements

ahoppen:
Add diagnostics for code block items that only contain a semicolon.


ahoppen:
@swift-ci Please test

Sources/SwiftParser/Diagnostics/ParserDiagnosticMessages.swift
```diff
@@ -72,6 +72,7 @@ public enum StaticParserError: String, DiagnosticMessage {
   case editorPlaceholderInSourceFile = "editor placeholder in source file"
   case missingColonInTernaryExprDiagnostic = "expected ':' after '? ...' in ternary expression"
   case missingFunctionParameterClause = "expected argument list in function declaration"
+  case stanaloneSemicolonStatement = "';' statements are not allowed"
```
bnbarham, line 75:
```suggestion
  case standaloneSemicolonStatement = "';' statements are not allowed"
```

Also I know this is what the old diagnostic was, but could we add `standalone` to the start? We could also make some better suggestions depending on the context, ie. in a switch suggest `break`. Possibly `fallthrough` as well, though that's semantically different.

PR #912, opened 2022-10-10
Simplify the signatures for Parser.parse()

DougGregor:
The new parser doesn't throw, and doesn't use several of the parameters that the old one used.


DougGregor:
@swift-ci please test

ahoppen (approved):
Thank you 👍

PR #913, opened 2022-10-10
Remove TODOs from SwitchTests.swift that are semantic

ahoppen:


ahoppen:
@swift-ci Please test

Tests/SwiftParserTest/translated/SwitchTests.swift
```diff
@@ -1402,11 +1179,7 @@ final class SwitchTests: XCTestCase {
       }
       """,
       diagnostics: [
-        // TODO: Old parser expected error on line 2: switch must be exhaustive
-        // TODO: Old parser expected note on line 2: do you want to add a default clause?
-        // TODO: Old parser expected error on line 5: unknown attribute 'unknown'
         DiagnosticSpec(message: "expected declaration after attribute in switch case"),
```
bnbarham, line 1182: Seems like we should parse this as the start of a `default` clause rather than a decl?

Tests/SwiftParserTest/translated/SwitchTests.swift
```diff
@@ -259,9 +238,7 @@ final class SwitchTests: XCTestCase {
       }
       """,
       diagnostics: [
-        // TODO: Old parser expected error on line 2: all statements inside a switch must be covered by a 'case' or 'default'
         DiagnosticSpec(message: "unexpected text 'x = 1' in switch case"),
```
bnbarham, line 241: Maybe keep a TODO? It'd be nice to handle this as a missing case rather than unexpected text.

ahoppen: I improved the diagnostic here.

Tests/SwiftParserTest/translated/SwitchTests.swift
ahoppen: I improved the diagnostic.

PR #915, opened 2022-10-11
Diagnose default switch case with 'where' clause

ahoppen:


ahoppen:
@swift-ci Please test

PR #916, opened 2022-10-11
Address review comments from previous diagnostic improvement PRs

ahoppen:
Addresses review comments from @bnbarham from previous PRs that improved diagnostics.

### DIFF
--- a/Sources/_SwiftSyntaxTestSupport/LocationMarkers.swift
+++ b/Sources/_SwiftSyntaxTestSupport/LocationMarkers.swift
@@ -38,11 +38,6 @@ fileprivate func nextMarkedRange(text: String, from: String.Index) -> Marker? {
   let markerRange = start ..< end
   let name = text[start ..< end]
 
-  // Expand to the whole line if the line only contains the marker
-  let lineRange = text.lineRange(for: start..<start)
-  if text[lineRange].trimmingCharacters(in: .whitespacesAndNewlines) == text[markerRange] {
-    return Marker(name: name, range: lineRange)
-  }
   return Marker(name: name, range: markerRange)
 }
 
--- a/Tests/SwiftParserTest/translated/TryTests.swift
+++ b/Tests/SwiftParserTest/translated/TryTests.swift
@@ -193,23 +193,55 @@ final class TryTests: XCTestCase {
       """
       class TryDecl {
         1️⃣try let singleLet = foo()
-        2️⃣try var singleVar = foo()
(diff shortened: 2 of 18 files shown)


ahoppen:
@swift-ci Please test

Sources/_SwiftSyntaxTestSupport/LocationMarkers.swift
```diff
@@ -38,11 +38,6 @@ fileprivate func nextMarkedRange(text: String, from: String.Index) -> Marker? {
   let markerRange = start ..< end
   let name = text[start ..< end]
 
-  // Expand to the whole line if the line only contains the marker
-  let lineRange = text.lineRange(for: start..<start)
-  if text[lineRange].trimmingCharacters(in: .whitespacesAndNewlines) == text[markerRange] {
-    return Marker(name: name, range: lineRange)
-  }
   return Marker(name: name, range: markerRange)
```
bnbarham, line 41: FWIW after your explanation I wasn't expecting you to remove this. But we can always add it back if we need it :)

Tests/SwiftParserTest/translated/TryTests.swift
```diff
@@ -234,8 +266,8 @@ final class TryTests: XCTestCase {
       """,
       diagnostics: [
         DiagnosticSpec(locationMarker: "1️⃣", message: "'try' must be placed on the thrown expression", fixIts: ["move 'try' after 'throw'"]),
-        DiagnosticSpec(locationMarker: "2️⃣", message: "expected expression in 'try' expression"),
-      ], fixedSource: "throw try <#expression#>"
+        DiagnosticSpec(locationMarker: "2️⃣", message: "expected expression after 'try'"),
+      ], fixedSource: "throw try"
```
bnbarham, line 270: I actually like the placeholder here, though do prefer the "after" rather than "in ... expression".

Sources/_SwiftSyntaxTestSupport/LocationMarkers.swift
ahoppen: Thinking about it again, I couldn’t really find a good use case, so I though we might just simplify things.

Tests/SwiftParserTest/translated/TryTests.swift
ahoppen: I thought it provider very little value. But I don’t have strong opinions either, so I’m adding it back in.

PR #918, opened 2022-10-11
When inserting a token in front of the previous token's trivia, only override its trailing trivia if the previous trivia is not empty

ahoppen:
Otherwise, we might be dropping spaces that are necessary to form valid Swift source.


ahoppen:
@swift-ci Please test

Sources/SwiftSyntax/Syntax.swift
```diff
@@ -225,10 +225,12 @@ public extension SyntaxProtocol {
       return nil
     }
     let siblings = NonNilRawSyntaxChildren(parent, viewMode: viewMode)
-    for absoluteRaw in siblings[..<self.index].reversed() {
-      let child = Syntax(SyntaxData(absoluteRaw, parent: parent))
-      if let token = child.lastToken(viewMode: viewMode) {
-        return token
+    if self.index >= siblings.startIndex {
```
bnbarham, line 228: How does this condition happen?

Sources/SwiftParser/Diagnostics/DiagnosticExtensions.swift
```diff
@@ -66,10 +66,13 @@ extension FixIt.Changes {
   /// Makes the `token` present, moving it in front of the previous token's trivia.
   static func makePresentBeforeTrivia(token: TokenSyntax) -> Self {
     if let previousToken = token.previousToken(viewMode: .sourceAccurate) {
-      let presentToken = PresentMaker().visit(token).withTrailingTrivia(previousToken.trailingTrivia)
+      var presentToken = PresentMaker().visit(token).as(TokenSyntax.self)!
```
bnbarham, line 69: Is it ever valid to rewrite a `TokenSyntax` as anything other than `TokenSyntax`? Should `SyntaxRewriter` just return a `TokenSyntax` for `visit(TokenSyntax)` instead? This sort of applies to the `SyntaxRewriter` in general though - we can rewrite to completely non-sensical trees,

ahoppen: Yes, because there are children that are of type `Syntax`. And in these cases it would be valid to rewrite `TokenSyntax` to something else. But maybe that’s not really something that should be done in practice.

Sources/SwiftSyntax/Syntax.swift
bnbarham: Thanks 🙇

Sources/SwiftParser/Diagnostics/DiagnosticExtensions.swift
bnbarham: Having a look at the gyb, the only time that's actually possible is for unexpected nodes. Perhaps we can change this after Rintaro's PR is in where we then have an enum of the choices instead.

ahoppen: I think we can change it now already. I’ll do it in a follow-up PR. This change cleans up a few other places as well

ahoppen: The motivating example was that this will allow us to output `precedence group` in the following test case. 

```
AssertParse(
  """
  postfix operator ++:1️⃣
  """,
  diagnostics: [
    DiagnosticSpec(message: "expected precedence group in operator declaration"),
  ]
)
```

I don’t personally have a strong preference regarding identifier vs name.

bnbarham: Maybe we should just change the child name to `identifier` for the name cases? If other people agree. Otherwise 🤷

PR #923, opened 2022-10-11
Highlight unexpected range

ahoppen:


ahoppen:
@swift-ci Please test

Tests/SwiftParserTest/translated/SubscriptingTests.swift
```diff
@@ -394,8 +379,15 @@ final class SubscriptingTests: XCTestCase {
         }
       """,
       diagnostics: [
-        // TODO: Old parser expected error on line 2: expected '{' in subscript to specify getter and setter implementation
```
bnbarham, line 397: Not sure about removing this one. I'd maybe at least expect a missing property?

ahoppen: The problem is that all declarations have optional bodies in SwiftSyntax, so this is technically a correct parse. The only way I can think of how we could model this to produce an error in the parser, would be to have `SubscriptDecl` (with a required body, this node can be used for structs etc.) and `SubscriptDeclWithoutBody` (for usage in protocols). But that would also imply that we couldn’t parse `.swiftinterface` files anymore. So maybe this really is an ASTGen diagnostic.

PR #924, opened 2022-10-11
Add diagnosic that subscripts cannot have a name

ahoppen:


ahoppen:
@swift-ci Please test

Tests/SwiftParserTest/translated/SubscriptingTests.swift
```diff
@@ -429,11 +436,18 @@ final class SubscriptingTests: XCTestCase {
       }
       """,
       diagnostics: [
-        // TODO: Old parser expected error on line 2: subscripts cannot have a name, Fix-It replacements: 13 - 14 = ''
-        DiagnosticSpec(locationMarker: "1️⃣", message: "unexpected text 'x' before parameter clause"),
-        // TODO: Old parser expected error on line 5: subscripts cannot have a name, Fix-It replacements: 13 - 14 = ''
-        DiagnosticSpec(locationMarker: "2️⃣", message: "unexpected text 'x<T>' before parameter clause"),
-      ]
+        DiagnosticSpec(locationMarker: "1️⃣", message: "subscripts cannot have a name", fixIts: ["remove 'x'"]),
+        DiagnosticSpec(locationMarker: "2️⃣", message: "subscripts cannot have a name", fixIts: ["remove 'x<T>'"]),
```
bnbarham, line 440: Nice improvement, we only removed `x` before and not `x<T>`.

Tests/SwiftParserTest/translated/SubscriptingTests.swift
```diff
@@ -447,26 +461,18 @@ final class SubscriptingTests: XCTestCase {
...
         DiagnosticSpec(locationMarker: "4️⃣", message: "expected '->' and return type in subscript"),
-      ]
-    )
-  }
-
-  func testSubscripting25() {
-    AssertParse(
+      ], fixedSource: """
+      struct A11 {
+        subscript( x y : (Int) -> Int)  -> <#type#>{
```
bnbarham, line 471: Woaw. That's some pretty crazy (and arguably incorrect) recovery. I'm not sure how we could end up with the "ideal" recovery of `subscript(x y: Int) -> Int` though, we're always going to eat that `->` as part of the type. Given how unlikely this case is to come up, it's probably fine.

Sources/SwiftParser/Diagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -235,7 +235,7 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
     if let semicolon = node.semicolon, semicolon.presence == .present, node.item.isMissingAllTokens {
       addDiagnostic(node, .standaloneSemicolonStatement, fixIts: [
         FixIt(message: RemoveTokensFixIt(tokensToRemove: [semicolon]), changes: [
-          .makeMissing(node: semicolon)
+          .makeMissing(tokens: [semicolon])
```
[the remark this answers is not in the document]
ahoppen: Good catch. I’ll address this in a commit where I’m also dropping the `tokens: ` parameter label.

Tests/SwiftParserTest/translated/SubscriptingTests.swift
ahoppen: Yeah, I think getting the correct recovery would require a lot more context. Basically, we would need to tell type parsing that it shouldn’t parse the last `->`, but that would require a lot of lookahead. Also, I’m not sure if I care too much. This code is pretty broken.

bnbarham: Yeah I'm definitely of the same opinion 👍. I can't see someone actually writing this. As long as we produce *something* decl related here I'd say it's fine.

bnbarham: What's much more likely is probably `subscript(x y: Int -> Int {`, ie. they just forgot the final parenthesis. But yeah, I agree it would require too much lookahead.

PR #931, opened 2022-10-11
Diagnose body in operator declaration

ahoppen:


ahoppen:
@swift-ci Please test

Sources/SwiftParser/Diagnostics/ParserDiagnosticMessages.swift
```diff
@@ -76,6 +76,7 @@ public enum StaticParserError: String, DiagnosticMessage {
   case editorPlaceholderInSourceFile = "editor placeholder in source file"
   case expectedExpressionAfterTry = "expected expression after 'try'"
   case missingColonInTernaryExprDiagnostic = "expected ':' after '? ...' in ternary expression"
+  case operatorShouldBeDeclaredWithoutBody = "operator should no longer be declared with body"
```
bnbarham, line 79: IMO I'd remove the "no longer". Probably made sense at the time, but I don't think our messages should be related to previous language versions. `operator should not be declared with body`?

ahoppen: Good idea. I’ll change it.

Issue #893, opened 2022-10-07
Access race in Swift Parser


jpsim:
### Description

TSan reports an access race in `Unicode.Scalar.isAsciiIdentifierContinue`.

Tested with 4f2e1537f39583f635d4a343128749d650157537

```
WARNING: ThreadSanitizer: Swift access race (pid=48066)
  Modifying access of Swift variable at 0x00011740b090 by thread T4:
    #0 Unicode.Scalar.isAsciiIdentifierContinue.getter <null>:78717160 (IntegrationTests:arm64+0xac7288)
    #1 Unicode.Scalar.isValidIdentifierContinuationCodePoint.getter <null>:78717160 (IntegrationTests:arm64+0xb0a178)
    #2 Lexer.Cursor.lexIdentifier(_:) <null>:78717160 (IntegrationTests:arm64+0xb08a4c)
    #3 Lexer.Cursor.nextToken(_:) <null>:78717160 (IntegrationTests:arm64+0xb023e4)
    #4 static Lexer.tokenize(_:from:) <null>:78717160 (IntegrationTests:arm64+0xb0408c)
    #5 specialized static Parser.parse(source:parseTransition:filenameForDiagnostics:languageVersion:enableBareSlashRegexLiteral:) <null>:78717160 (IntegrationTests:arm64+0xb1d80c)
    #6 specialized String.withUTF8<A>(_:) <null>:78717160 (IntegrationTests:arm64+0xb1d9bc)
    #7 static Parser.parse(source:parseTransition:filenameForDiagnostics:languageVersion:enableBareSlashRegexLiteral:) <null>:78717160 (IntegrationTests:arm64+0xb1ab50)
    #8 SwiftLintFile.syntaxTree.getter <null>:78717160 (IntegrationTests:arm64+0xa09fc)
    #9 SwiftLintFile.parserDiagnostics.getter <null>:78717160 (IntegrationTests:arm64+0xa0508)
    #10 CollectedLinter.correct(using:) <null>:78717160 (IntegrationTests:arm64+0xdff80)
    #11 closure #1 in IntegrationTests.testSwiftLintAutoCorrects() <null>:78717160 (IntegrationTests:arm64+0xcf0c)
    #12 partial apply for closure #1 in IntegrationTests.testSwiftLintAutoCorrects() <null>:78717160 (IntegrationTests:arm64+0xcf74)
[...]

PR #932, opened 2022-10-11
[Parse] Avoid 'inout' operator passing in Character.Info testing

rintaro:
Passing `var` value with `&` operator triggers TSAN "modifying access" detection even though it doesn't actually mutates the value. That results "access race" sanitizer error in multi-thread environments.

Instead of using a tuple table, use 'switch' to test the character info. Also, eliminate unused character info value, and introduce 'IDENT_CONT'

https://github.com/apple/swift-syntax/issues/893 
rdar://100906576

### DIFF
--- a/Sources/SwiftParser/CharacterInfo.swift
+++ b/Sources/SwiftParser/CharacterInfo.swift
@@ -18,18 +18,10 @@ extension Character {
       self.rawValue = rawValue
     }
 
-    static let SPACE    = Character.Info(rawValue: 0x01)  // ' '
-    static let DIGIT    = Character.Info(rawValue: 0x02)  // 0-9
-    static let XLETTER  = Character.Info(rawValue: 0x04)  // a-f,A-F
-    static let UPPER    = Character.Info(rawValue: 0x08)  // A-Z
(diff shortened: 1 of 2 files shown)


rintaro:
@swift-ci Please test

Sources/SwiftParser/CharacterInfo.swift
```diff
@@ -42,167 +34,70 @@ extension Unicode.Scalar {
   /// to be allowed to appear in a starting position in a programming language
   /// identifier.
   var isAsciiIdentifierStart: Bool {
-    return self.testRawInfoTable { entry in
-      !entry.intersection([ .UPPER, .LOWER, .UNDER, ]).isEmpty
-    }
+    self.testCharacterInfo(.IDENT_CONT) && !self.isDigit && self != "$"
```
CodaFi, line 37: Maybe we should burn a bit for identifier starts too.

CodaFi (approved):
It's great to see this table continue to shrink to only the relevant bits.

rintaro:
@swift-ci Please test

PR #935, opened 2022-10-12
If a missing token has a child name, use that instead of the generic token kind

ahoppen:


ahoppen:
@swift-ci Please teste

bnbarham:
I prefer `identifier` personally. Did this change any others? Can't see any except identifier -> name, but I could just be missing them.

PR #938, opened 2022-10-13
Parse `return @unknown default` correctly

### LINKED ISSUE #933: Unexpected MissingExprSyntax when empty return precedes "@unknown default" in switch statement
### Issue Kind

### Source Code

```swift
public func bar_bad(foo: Int) {
  switch foo {
  case 0:
    return
  @unknown default:
    return
  }
}
```

### Description

This produces the following parse tree:

```
SourceFileSyntax children=2
      0: FunctionDeclSyntax children=5
        0: ModifierListSyntax children=1
          0: DeclModifierSyntax children=1
            0: publicKeyword
        1: funcKeyword
        2: identifier("bar_bad")
        3: FunctionSignatureSyntax children=1
          0: ParameterClauseSyntax children=3
            0: leftParen
            1: FunctionParameterListSyntax children=1
              0: FunctionParameterSyntax children=3
                0: identifier("foo")
                1: colon
                2: SimpleTypeIdentifierSyntax children=1
                  0: identifier("Int")
            2: rightParen
        4: CodeBlockSyntax children=3
          0: leftBrace
          1: CodeBlockItemListSyntax children=1
              0: SwitchStmtSyntax children=5
                0: switchKeyword
                  0: identifier("foo")
                2: leftBrace
                3: SwitchCaseListSyntax children=2
                  0: SwitchCaseSyntax children=2
                    0: SwitchCaseLabelSyntax children=3
                      0: caseKeyword
                      1: CaseItemListSyntax children=1
                        0: CaseItemSyntax children=1
                          0: ExpressionPatternSyntax children=1
                            0: IntegerLiteralExprSyntax children=1
                              0: integerLiteral("0")
                      2: colon
                    1: CodeBlockItemListSyntax children=1
                        0: ReturnStmtSyntax children=2
                          0: returnKeyword
                          1: MissingExprSyntax
                  1: SwitchCaseSyntax children=3
                    0: AttributeSyntax children=2
                      0: atSign
                      1: identifier("unknown")
                    1: SwitchDefaultLabelSyntax children=2
                      0: defaultKeyword
                      1: colon
                    2: CodeBlockItemListSyntax children=1
                        0: ReturnStmtSyntax children=1
                          0: returnKeyword
                4: rightBrace
          2: rightBrace
  1: eof
```

Notice how the `ReturnStmtSyntax` includes a `MissingExprSyntax` as a child. I think that's incorrect because the "empty return" is valid in this situation. If the return isn't empty, either by adding a semicolon or returning a value, or if the next case isn't `@unknown default` then the parser handles it correctly.

ahoppen:
Fixes #933
Resolves rdar://101060095

### DIFF
--- a/Sources/SwiftParser/Statements.swift
+++ b/Sources/SwiftParser/Statements.swift
@@ -24,6 +24,16 @@ extension TokenConsumer {
   }
 }
 
+extension Parser.Lookahead {
+  mutating func atStartOfSwitchCaseItem() -> Bool {
+    while self.consume(if: .atSign) != nil {
+      self.consume(if: .identifier)
+    }
+
+    return self.at(anyIn: SwitchCaseStart.self) != nil
+  }
+}
+
 extension Parser {
   /// Parse a statement.
   ///
@@ -1126,6 +1136,11 @@ extension Parser.Lookahead {
   /// - Note: This function must be kept in sync with `parseStatement()`.
   /// - Seealso: ``Parser/parseStatement()``
   public mutating func isStartOfStatement(allowRecovery: Bool = false) -> Bool {
+    if (self.at(anyIn: SwitchCaseStart.self) != nil || self.at(.atSign)) && withLookahead({ $0.atStartOfSwitchCaseItem() }) {
+      // We consider SwitchCaseItems statements so we don't parse the start of a new case item as trailing parts of an expresion.
+      return true
+    }
+
     _ = self.consume(if: .identifier, followedBy: .colon)
     let switchSubject: CanBeStatementStart?
     if allowRecovery {
--- a/Sources/SwiftParser/TokenConsumer.swift
+++ b/Sources/SwiftParser/TokenConsumer.swift
@@ -123,6 +123,11 @@ extension TokenConsumer {
       return consumeAnyToken()
     }
   }
+
+  public func withLookahead<T>(_ body: (_: inout Parser.Lookahead) -> T) -> T {
+    var lookahead = lookahead()
+    return body(&lookahead)
+  }
 }
 
 // MARK: Consuming tokens (`consume`)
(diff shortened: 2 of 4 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

Sources/SwiftParser/Statements.swift
```diff
@@ -906,7 +916,7 @@ extension Parser {
         .poundIfKeyword, .poundErrorKeyword, .poundWarningKeyword,
         .poundEndifKeyword, .poundElseKeyword, .poundElseifKeyword
       ])
-        && !self.atStartOfStatement() && !self.atStartOfDeclaration() {
+        && !self.atStartOfStatement() && !self.atStartOfDeclaration() && !self.withLookahead({ $0.atStartOfSwitchCaseItem() }) {
```
CodaFi, line 919: We should fix this in `atStartOfStatement`. This is ultimately a regression there because we dropped the clause that checks for `(@attr)* <identifier>`

ahoppen: Do you think that case labels should return `true` for `atStartOfStatement`? Because in my opinion these are not statements. If anything, I’d say that case labels are declarations, but even that I don’t think they are.

ahoppen:
@swift-ci Please test

Sources/SwiftParser/Statements.swift
CodaFi: I would say yeah they're not real statements, but if we intend to match the behavior of the legacy parser, then we have to consider them a part of that predicate so the rest of the parser picks this up too.

ahoppen:
@swift-ci Please test

CodaFi (approved):

PR #939, opened 2022-10-13
Diagnose identifier in operator declarations

ahoppen:
Diagnose if identifiers are used in operator declarations or if multiple operators occur in the operator name.


ahoppen:
@swift-ci Please test

PR #940, opened 2022-10-13
Improve recovery in precedence groups

ahoppen:


ahoppen:
@swift-ci Please test

Tests/SwiftParserTest/translated/OperatorDeclTests.swift
```diff
@@ -302,51 +300,91 @@ final class OperatorDeclTests: XCTestCase {
...
+      """
+    )
+  }
+
+
+  func testOperatorDecl15e() {
+    AssertParse(
+      """
       precedencegroup D {
-        assignment: 4️⃣no
```
bnbarham, line 327: Worth having a fix-it here to cover common cases? no/NO/FALSE/0/etc

ahoppen: Honestly, I don’t think so. If you are declaring precedence groups but haven’t understood how Swift specifies booleans, I don’t care about you, I think.

PR #941, opened 2022-10-13
Diagnose if a number is used in place of an identifier

ahoppen:


ahoppen:
@swift-ci Please test

Tests/SwiftParserTest/Declarations.swift
```diff
@@ -690,8 +690,8 @@ final class DeclarationTests: XCTestCase {
       }
       """,
       diagnostics: [
-        DiagnosticSpec(locationMarker: "1️⃣", message: "identifier can only start with a letter or underscore, not a number"),
-        DiagnosticSpec(locationMarker: "2️⃣", message: "identifier can only start with a letter or underscore, not a number"),
+        DiagnosticSpec(locationMarker: "1️⃣", message: "name can only start with a letter or underscore, not a number"),
```
bnbarham, line 693: As in the other PR, IMO `identifier` is more descriptive than `name` here.

ahoppen: Once I’ve merged my PR backlog, I will set up a PR that checks what happens if we remove the `name_for_diagnostics`. It’s probably easiest to discuss all the implications this has on there.

PR #946, opened 2022-10-13
Consider regex literals as identifier-like

ahoppen:
This was caught by consistency checking between the old and the new parser.


ahoppen:
@swift-ci Please test

bnbarham:
@swift-ci please test

bnbarham:
Going to merge to unblock CI. But we likely need many, many more regex literal tests 😅

PR #947, opened 2022-10-14
Add diagnostic for misplaced specifiers

ahoppen:


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

Sources/SwiftParser/Types.swift
```diff
@@ -783,8 +788,12 @@ extension Parser.Lookahead {
 
 extension Parser {
   @_spi(RawSyntax)
-  public mutating func parseTypeAttributeList() -> (RawTokenSyntax?, RawAttributeListSyntax?) {
-    let specifier = self.consume(ifAny: [.inoutKeyword], contextualKeywords: ["__shared", "__owned"])
+  public mutating func parseTypeAttributeList(misplacedSpecifiers: [RawTokenSyntax] = []) -> (RawTokenSyntax?, RawAttributeListSyntax?) {
+    var specifier: RawTokenSyntax? = self.consume(ifAnyIn: TypeSpecifier.self)
+    // We can only stick one specifier on this type. Let's pick the first one
```
bnbarham, line 793: I feel the same way about this PR as I do about mine. I'm not sure it's worth keeping around all the state in order to allow adding the "missing" specifier here.

Also note that specifiers are parsed into `RawAttributeListSyntax` below, without being marked as unexpected and without the eventual fix-it. The current diagnostic (in the actual compiler) isn't great about this either, we end up with a `'<specifier>' may only be used on parameters` diagnostic and only recover from a single specifier after the attribute list and after that lose the type.

Whatever we end up doing here, it would be nice to be consistent across All the Things (where "things" here are attributes/modifiers/specifiers). Ie. should we ever expect a single, or should we always allow a list? Do we diagnose moving them to their correct position, or just add them (parsed correctly) as unexpected nodes?

There are current clients (eg. SwiftLint) that check eg. the async keyword in a function declaration to determine if a function is async. Should that check pass if the async is in any position, or just the correct?

Tests/SwiftParserTest/translated/InvalidTests.swift
```diff
@@ -289,29 +256,56 @@ final class InvalidTests: XCTestCase {
...
+
+
+  func testInvalid16b() {
+    AssertParse(
+      """
+      func f2_43591(1️⃣inout inout b: Int) {}
+      """,
+      diagnostics: [
+        DiagnosticSpec(message: "'inout inout' before a parameter name is not allowed", fixIts: ["move 'inout inout' in front of type"]),
+      ], fixedSource: "func f2_43591(b: inout Int) {}"
```
bnbarham, line 275: That's somewhat confusing 😅. What happens when there's already an `inout` in this case? Does it say "remove redundant inout inout"?

Sources/SwiftParser/Declarations.swift
```diff
@@ -1387,7 +1392,7 @@ extension Parser {
 
         let type: RawTypeSyntax?
         if shouldParseType {
-          type = self.parseType()
+          type = self.parseType(misplacedSpecifiers: misplacedSpecifiers)
```
bnbarham, line 1395: This name is somewhat confusing since function types can also have specifiers, but this is for the type specifiers specifically.

PR #951, opened 2022-10-15
Allow leading dot syntax for StaticParserError

hamishknight:
Convert StaticParserError into a struct, and define the diagnostics in a `DiagnosticMessage where Self == StaticParserError` extension. This allows us to use leading dot syntax to pass a StaticParserError anywhere a DiagnosticMessage is expected. Also, do the same for StaticParserFixIt.

To compute the diagnostic ID, we can (ab)use `#function` to pick up the name of the property

### DIFF
--- a/Sources/SwiftParser/Diagnostics/DiagnosticExtensions.swift
+++ b/Sources/SwiftParser/Diagnostics/DiagnosticExtensions.swift
@@ -14,14 +14,18 @@ import SwiftDiagnostics
 import SwiftSyntax
 
 extension FixIt {
-  init(message: StaticParserFixIt, changes: Changes) {
-    self.init(message: message as FixItMessage, changes: changes)
-  }
-
   public init(message: FixItMessage, changes: [Changes]) {
     self.init(message: message, changes: FixIt.Changes(combining: changes))
   }
 
+  // These overloads shouldn't be needed, but are currently required for the
+  // Swift 5.5 compiler to handle non-trivial FixIt initializations using
+  // leading-dot syntax.
+  // TODO: These can be dropped once we require a minimum of Swift 5.6 to
+  // compile the library.
+  init(message: StaticParserFixIt, changes: Changes) {
+    self.init(message: message as FixItMessage, changes: changes)
+  }
   init(message: StaticParserFixIt, changes: [Changes]) {
     self.init(message: message as FixItMessage, changes: FixIt.Changes(combining: changes))
   }
--- a/Sources/SwiftParser/Diagnostics/ParseDiagnosticsGenerator.swift
+++ b/Sources/SwiftParser/Diagnostics/ParseDiagnosticsGenerator.swift
@@ -65,27 +65,6 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
     self.handledNodes.append(contentsOf: handledNodes)
   }
 
-  /// Produce a diagnostic.
-  func addDiagnostic<T: SyntaxProtocol>(
-    _ node: T,
-    position: AbsolutePosition? = nil,
-    _ message: StaticParserError,
-    highlights: [Syntax] = [],
-    notes: [Note] = [],
-    fixIts: [FixIt] = [],
-    handledNodes: [SyntaxIdentifier] = []
-  ) {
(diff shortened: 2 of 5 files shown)


hamishknight:
@swift-ci please test

hamishknight:
@swift-ci please test

hamishknight:
```
/Users/ec2-user/jenkins/workspace/swift-syntax-PR-macOS/branch-main/swift-syntax/Sources/SwiftParser/Diagnostics/ParseDiagnosticsGenerator.swift:442:5: error: type of expression is ambiguous without more context
    exchangeTokens(
    ^~~~~~~~~~~~~~~
```

:( wonder what swift version CI is using

hamishknight:
@swift-ci please test

hamishknight:
@swift-ci please test

hamishknight:
@swift-ci please test

hamishknight:
Looks like the CI is using Swift 5.5, which should be fine as it should have [SE-0299](https://github.com/apple/swift-evolution/blob/main/proposals/0299-extend-generic-static-member-lookup.md), but it seems the 5.5 constraint system is barfing when a closure is involved (works on 5.6 though). Adding a generic parameter to `exchangeTokens` & `removeToken` should work around the issue.

hamishknight:
@swift-ci please test

hamishknight:
@swift-ci please test

ahoppen (commented):
Neat idea 🤩 I’ve got a couple of additional suggestion/ideas:

- Could you add a test case that checks the diagnostic ID for one `StaticParserError`? I would like to see how the diagnostic ID looks like (basically making sure it contains no parentheses)
- I find it harder to find diagnostic messages in the new extension but I think that’s resolved by formatting the diagnostic messages as single lines, e.g.
```swift 
public static var allStatmentsInSwitchMustBeCoveredByCase: Self { .init("all statements inside a switch must be covered by a 'case' or 'default' label") }
```
- I think we can do the same for `StaticParserFixIt`.

hamishknight:
> Could you add a test case that checks the diagnostic ID for one StaticParserError? I would like to see how the diagnostic ID looks like (basically making sure it contains no parentheses)

We [already have one](https://github.com/apple/swift-syntax/blob/e5ebdb6e71893e1453ab751285f021ecd8a526ff/Tests/SwiftParserTest/DiagnosticInfrastructureTests.swift#L16) :)

> I find it harder to find diagnostic messages in the new extension but I think that’s resolved by formatting the diagnostic messages as single lines, e.g.

Hmm, I definitely prefer the current formatting. Putting them on a single line means that each diagnostic message starts at a different column (depending on how long the var name is), which make it hard to read IMO. It also makes it extra hard for people with smaller screens, e.g:

Screen Shot 2022-10-17 at 10 55 39


> I think we can do the same for `StaticParserFixIt`

👍

hamishknight:
@swift-ci please test

hamishknight:
Bleh, the 5.5 compiler really seems to struggle with these:

```
/Users/ec2-user/jenkins/workspace/swift-syntax-PR-macOS/branch-main/swift-syntax/Sources/SwiftParser/Diagnostics/ParseDiagnosticsGenerator.swift:192:43: error: missing argument label 'fromProtocol:' in call
              newNode: Syntax(TokenSyntax.identifier("`\(invalidIdentifier.text)`", leadingTrivia: invalidIdentifier.leadingTrivia, trailingTrivia: invalidIdentifier.trailingTrivia))
                                          ^
                              fromProtocol: 
/Users/ec2-user/jenkins/workspace/swift-syntax-PR-macOS/branch-main/swift-syntax/Sources/SwiftParser/Diagnostics/ParseDiagnosticsGenerator.swift:268:12: error: type 'FixIt.Change' has no member 'remove'
          .remove(unexpected: unexpected)
          ~^~~~~~
```

going to add back the `StaticParserFixIt` overloads on `FixIt`, which we can drop once 5.6 is the minimum required Swift version.

hamishknight:
@swift-ci please test

ahoppen (approved):

hamishknight:
@swift-ci please test

hamishknight:
@swift-ci please test

PR #952, opened 2022-10-16
Work around lack of implicit existential opening in older compilers

DougGregor:


DougGregor:
@swift-ci please test

Issue #928, opened 2022-10-11
Keypath parsing treats `.?.` as binary operator


allevato:
### Issue Kind

Parse of Valid Source Produced Invalid Syntax Tree

### Source Code

```swift
let x = \.?.foo
```

### Description

This produces the following parse tree:

```
SourceFileSyntax children=2
  0: CodeBlockItemListSyntax children=1
    0: CodeBlockItemSyntax children=1
      0: VariableDeclSyntax children=2
        0: letKeyword
        1: PatternBindingListSyntax children=1
          0: PatternBindingSyntax children=2
            0: IdentifierPatternSyntax children=1
              0: identifier("x")
            1: InitializerClauseSyntax children=2
              0: equal
              1: SequenceExprSyntax children=1
                0: ExprListSyntax children=3
                  0: KeyPathExprSyntax children=2
                    0: backslash
                    1: KeyPathComponentListSyntax
                  1: BinaryOperatorExprSyntax children=1
                    0: unspacedBinaryOperator(".???.")
                  2: IdentifierExprSyntax children=1
                    0: identifier("foo")
  1: eof
```

Note that the `KeyPathComponentListSyntax` is empty and the components are instead parsed as a binary operator and an identifier.

I saw a [comment in the parser](https://github.com/apple/swift-syntax/blob/main/Sources/SwiftParser/Expressions.swift#L869-L873) calling out the difficulty here that the keypath component is being lexed as an operator, but I couldn't immediately figure out the best way to resolve it.

jpsim:
Did you check before or after https://github.com/apple/swift-syntax/pull/896 landed? Seems very similar.

allevato:
This is occurring at current HEAD (f12d83150874c3c0abeb5a78bce0659b5564463e).

ahoppen:
rdar://101038025

PR #953, opened 2022-10-16
Parse (unspaced) binary operators within a key path, e.g., .?..

DougGregor:
Fixes https://github.com/apple/swift-syntax/issues/928.

### DIFF
--- a/Tests/SwiftParserTest/Expressions.swift
+++ b/Tests/SwiftParserTest/Expressions.swift
@@ -92,6 +92,45 @@ final class ExpressionTests: XCTestCase {
   func testKeypathExpression() {
     AssertParse(
       #"""
+      \.?.foo
+      """#,
+      substructure: Syntax(
+        CodeBlockItemListSyntax([
+          CodeBlockItemSyntax(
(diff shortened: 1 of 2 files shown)


DougGregor:
@swift-ci please test

Tests/SwiftParserTest/Expressions.swift
```diff
@@ -90,6 +90,45 @@ final class ExpressionTests: XCTestCase {
...
   func testKeypathExpression() {
+    AssertParse(
+      #"""
+      \.?.foo
+      """#,
+      substructure: Syntax(
+        CodeBlockItemListSyntax([
+          CodeBlockItemSyntax(
+            item: Syntax(
+              KeyPathExprSyntax(
```
ahoppen, line 101: IMO it would have been sufficient to start the `substructure` with `KeyPathExprSyntax` because the wrapping `CodeBlockItem` isn’t really interesting. But it doesn’t really do any harm either.

DougGregor: Okay, that makes sense.

PR #954, opened 2022-10-17
Diagnose if multiple identifiers are used for declaration names

ahoppen:


ahoppen:
@swift-ci Please test

PR #955, opened 2022-10-17
Remove semantic TODOs from MatchingPatternTests.swift and InvalidTests.swift

ahoppen:


ahoppen:
@swift-ci Please test

PR #957, opened 2022-10-17
Diagnose if `class` is being used as an inherited type in a generic parameter clause

ahoppen:


ahoppen:
@swift-ci Please test

PR #958, opened 2022-10-17
Improve diagnostic if colon and expression are missing in ternary expression

ahoppen:


ahoppen:
@swift-ci Please test

PR #959, opened 2022-10-17
Add diagnostic if deinits have names and parameters

ahoppen:


ahoppen:
@swift-ci Please test

Tests/SwiftParserTest/translated/InitDeinitTests.swift
```diff
@@ -121,9 +151,12 @@ final class InitDeinitTests: XCTestCase {
       }
       """,
       diagnostics: [
-        // TODO: Old parser expected error on line 2: no parameter clause allowed on deinitializer, Fix-It replacements: 9 - 18 = ''
-        DiagnosticSpec(message: "unexpected code '(a : Int) {}' in class"),
-      ]
+        DiagnosticSpec(message: "deinitializers cannot have parameters", fixIts: ["remove function signature"]),
+      ], fixedSource: """
+      class FooClassDeinitializerA {
+        deinit{}
```
bnbarham, line 157: Unfortunate that we remove the trailing trivia here. But adding it in would make the `deinit x(a: Int) {` case look silly instead 🤷.

Tests/SwiftParserTest/translated/InitDeinitTests.swift
```diff
@@ -403,4 +393,14 @@ final class InitDeinitTests: XCTestCase {
     )
   }
 
+  func testDeinitInSwiftinterfaceIsFollowedByFinalFunc() {
+    AssertParse(
```
bnbarham, line 397: Could you check the structure here? This could also be parsed as
```
deinit final
func foo()
```
depending on recovery. Do we already have:
```
class Foo {
  func foo() final
  func bar()
}
```
?

PR #961, opened 2022-10-17
Simplify signature of makeMissing and makePresent

ahoppen:


ahoppen:
@swift-ci Please test

PR #962, opened 2022-10-17
Allow calls to `init` without leading `self.` or `super.`

ahoppen:
ASTGen will need to diagnose the missing `self.` or `super.`

### DIFF
--- a/Sources/SwiftParser/Declarations.swift
+++ b/Sources/SwiftParser/Declarations.swift
@@ -31,6 +31,7 @@ extension DeclarationModifier {
 extension TokenConsumer {
   func atStartOfDeclaration(
     isAtTopLevel: Bool = false,
+    allowInitDecl: Bool = true,
     allowRecovery: Bool = false
   ) -> Bool {
     if self.at(anyIn: PoundDeclarationStart.self) != nil {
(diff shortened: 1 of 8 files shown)


ahoppen:
@swift-ci Please test

DougGregor (commented):
I'm surprised this is diagnosed in the syntax tree at all. Why doesn't this belong in semantic analysis?

ahoppen:
IIUC the root cause is that we always consider `init` as the start of a declaration. If you want to call `init` on `self`, you need to disambiguate by prefixing it with `self.`. At least that’s how the old parser operated and I wanted to maintain this behavior.

Given that `init` is only valid as a type member and not in code blocks, we could change the rules a little bit and only consider `init` as the start of a declaration if we are parsing a `MemberDeclList`. This would make the following a valid parse (it currently isn’t), which will need to be diagnosed in ASTGen. WDYT?
```swift
class Foo {
  init() {}
  init(param: Int) {
    init()
  }
}
```

CC @CodaFi

DougGregor:
> Given that `init` is only valid as a type member and not in code blocks, we could change the rules a little bit and only consider `init` as the start of a declaration if we are parsing a `MemberDeclList`. This would make the following a valid parse (it currently isn’t), which will need to be diagnosed in ASTGen.

I'd be okay with this. It's not context-sensitive in any real sense, and it would be better for recovery to have `init()` be an expression wherever we allow them.

ahoppen:
Updated the PR to allow calls to `init` without `self.` or `super.`

ahoppen:
@swift-ci Please test

DougGregor (approved):

PR #963, opened 2022-10-17
Allow postfix #if to be empty

ahoppen:


ahoppen:
@swift-ci Please test

PR #964, opened 2022-10-17
Check parents for `childNameForDiagnostics`

ahoppen:


ahoppen:
@swift-ci Please test

Tests/SwiftParserTest/translated/GuardTests.swift
```diff
@@ -9,18 +9,23 @@ final class GuardTests: XCTestCase {
...
 
+  func testGuard2() {
+    AssertParse(
+      """
+      func noCondition() {
+        guard 1️⃣else {}
+      }
+      """,
+      diagnostics: [
+        DiagnosticSpec(message: "expected conditions in 'guard' statement"),
```
bnbarham, line 27: I know we allow "conditions" but IMO "condition" makes more sense here, ie. we expect at least one *condition*, not necessarily multiple.

PR #969, opened 2022-10-17
Syntactic macro system

DougGregor:
Implement a simple, experimental syntactic macro system that allows syntax of the form `#macroName(optional-macro-args)` to be treated as a macro evaluation, where macros are implemented via syntactic transformations on a `SwiftSyntax` tree. 

There are a few example macros here, one of which is a `stringify` macro that expands its argument into a tuple containing both the value and the string form of the argument, e.g.,

```swift
let b = #stringify(x + y)
```

will evaluate to

```swift
let b = (x + y, #"x + y"#)
```

New macros can be introduced by creating a type conforming to one of the `*Macro` protocols, such as the `ExpressionMacro` protocol for types that use the `#` expression syntax. 

Although this library provides purely syntactic expansion, the idea is to have a semantic component to macros, such that Swift's type checker performs some checking on the macro arguments, then initiates the macro evaluation to produce the final code. 

The design here is discussed in [A Possible Vision for Macros in Swift](https://gist.github.com/DougGregor/4f3ba5f4eadac474ae62eae836328b71).


DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

Sources/SwiftParser/Declarations.swift
```diff
@@ -2336,4 +2338,57 @@ extension Parser {
...
+      args = parseArgumentListElements(pattern: .none)
+      (unexpectedBeforeRightParen, rightParen) = self.expect(.rightParen)
+    } else {
+      args = []
+      unexpectedBeforeRightParen = nil
+      rightParen = nil
+    }
+
+    // Parse the optional trailing closures.
+    let trailingClosure: RawClosureExprSyntax?
```
ahoppen, line 2368: How would a macro expansion declaration with a trailing closure look like?

Sources/SwiftParser/Expressions.swift
```diff
@@ -1187,57 +1148,61 @@ extension Parser {
...
-    let (unexpectedBeforeLeftParen, leftParen) = self.expect(.leftParen)
-    let arguments = self.parseArgumentListElements(pattern: .none)
-    let (unexpectedBeforeRightParen, rightParen) = self.expect(.rightParen)
-    return RawObjectLiteralExprSyntax(
-      identifier: poundKeyword,
-      unexpectedBeforeLeftParen,
+    let (unexpectedBeforeMacro, macro) = self.expectIdentifier()
+
+    // Parse the optional parenthesized argument list.
+    let leftParen = self.consume(if: .leftParen, where: { !$0.isAtStartOfLine })
```
ahoppen, line 1167: AFAICT `parseMacroExpansionExpr` and `parseMacroExpansionDecl` seem pretty similar to me. Do you think we could unify their parsing?

ahoppen (commented):
A few comments from reviewing the parser bits of this PR.

PR #972, opened 2022-10-18
Introduce a SyntaxRawRewriter that is more performant than SyntaxRewriter

ahoppen:
`SyntaxRawRewriter` is pretty much equivalent to `SyntaxRewriter` but the `visit` methods return raw syntax nodes, which makes `SyntaxRawRewriter` more performant than `SyntaxRewriter`.

The motivation for this is that `BasicFormat` needs to be written in terms of `SyntaxRawRewriter` to get it to an acceptable performance level.

### DIFF
--- a/Sources/SwiftSyntax/SyntaxRawRewriter.swift.gyb
+++ b/Sources/SwiftSyntax/SyntaxRawRewriter.swift.gyb
@@ -1,0 +1,279 @@
+%{
+  from gyb_syntax_support import *
+  # -*- mode: Swift -*-
+  # Ignore the following admonition it applies to the resulting .swift file only
+}%
+//// Automatically Generated From SyntaxRawRewriter.swift.gyb.
+//// Do Not Edit Directly!
+//===------------ SyntaxRawRewriter.swift ---------------------------------===//
+//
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

Sources/SwiftSyntax/SyntaxRawRewriter.swift.gyb
```diff
@@ -0,0 +1,279 @@
...
+          newLayout!.append(nil)
+        }
+        continue
+      }
+
+      // Build the Syntax node to rewrite
+      let absoluteRaw = AbsoluteRawSyntax(raw: child, info: info)
+      let data = SyntaxData(absoluteRaw, parent: syntaxNode)
+      
+      let rewritten = visit(data)
```
rintaro, line 234: Even at this point, we can't know the arena of `rewritten` is alive or not.
So it must be the rewriter (sub classes)'s responsibly to keep arena's alive until the rewriting completes.

ahoppen:
Thinking about this again, I realized that this performance optimization isn’t actually necessary, https://github.com/apple/swift-syntax/pull/977 is sufficient to get acceptable performance.

PR #973, opened 2022-10-18
Recover if `func` or `var` is missing for a member decl item

ahoppen:


ahoppen:
@swift-ci Please test

PR #974, opened 2022-10-18
Diagnose if initializer is used as type annotation

ahoppen:


ahoppen:
@swift-ci Please test

PR #976, opened 2022-10-19
Bridge Parser Diagnostics into the C++ Diagnostic Engine

CodaFi:


CodaFi:
apple/swift#61631

@swift-ci test

Sources/SwiftCompilerSupport/ConsistencyCheck.swift
```diff
@@ -39,7 +39,9 @@ extension Syntax {
 @_spi(SwiftCompiler)
 public func _parserConsistencyCheck(
   bufferPtr: UnsafePointer<UInt8>, bufferLength: Int,
-  filename: UnsafePointer<UInt8>, flags: CUnsignedInt
+  filename: UnsafePointer<UInt8>, flags: CUnsignedInt,
+  ctx: OpaquePointer,
+  hook: @convention(c) (Int, UnsafePointer<Int8>, OpaquePointer) -> Void
```
ahoppen, line 44: Could you add doc comments for `ctx` and `hook`, including what the parameters to `hook` are? Also, I think `hook` would probably be better named `report diagnostic` or something like that.

Sources/SwiftCompilerSupport/ConsistencyCheck.swift
```diff
@@ -73,7 +77,9 @@ public func _parserConsistencyCheck(
           continue
         }
 
-        print("\(String(cString: filename)): error: \(diag.debugDescription)")
+        diag.debugDescription.withCString { diagText in
+          hook(diag.location(converter: converter).offset, diagText, ctx)
```
ahoppen, line 81: I don’t think you actually need a `SourceLocationConverter` here because `AbsolutePosition` already contains a `utf8Offset`.

```suggestion
          hook(diag.position.utf8Offset, diagText, ctx)
```

Sources/SwiftCompilerSupport/ConsistencyCheck.swift
CodaFi: Sure, I'll do it on the C prototype though.

CodaFi:
https://github.com/apple/swift/pull/61631

@swift-ci test

ahoppen (approved):

DougGregor (approved):
LGTM. Should the round-trip failures go through this new diagnostic hook, too?

PR #977, opened 2022-10-19
Significantly improve performance of BasicFormat

ahoppen:
In `visit(TokenSyntax)`, we were modifying the token in place, which caused the entire syntax tree to be rewritten to its root. It’s a lot (6x) faster to just modify `leadingTrivia` and `trailingTrivia` and create a new `TokenSyntax` that doesn’t have a parent.


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

PR #980, opened 2022-10-19
Special case error messages if a single keyword is unexpected

ahoppen:


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

PR #982, opened 2022-10-19
Remove name_for_diagnostics='name'

ahoppen:
@bnbarham We chatted about this before. Do you think the new error messages are better? I don’t really have strong opinions.

### DIFF
--- a/gyb_syntax_support/StmtNodes.py
+++ b/gyb_syntax_support/StmtNodes.py
@@ -5,7 +5,7 @@ STMT_NODES = [
     # labeled-stmt -> label ':' stmt
     Node('LabeledStmt', name_for_diagnostics='labeled statement', kind='Stmt',
          children=[
-             Child('LabelName', kind='IdentifierToken', name_for_diagnostics='name'),
+             Child('LabelName', kind='IdentifierToken', name_for_diagnostics='label name'),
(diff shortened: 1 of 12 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

gyb_syntax_support/StmtNodes.py
```diff
@@ -5,7 +5,7 @@
     # labeled-stmt -> label ':' stmt
     Node('LabeledStmt', name_for_diagnostics='labeled statement', kind='Stmt',
          children=[
-             Child('LabelName', kind='IdentifierToken', name_for_diagnostics='name'),
+             Child('LabelName', kind='IdentifierToken', name_for_diagnostics='label name'),
```
bnbarham, line 8: Doesn't seem to be a test for this case?

bnbarham (commented):
Yeah, I like it better 👍

gyb_syntax_support/StmtNodes.py
ahoppen: I don’t actually think we ever hit this right now because we only parse a labeled statement if there is a label. But I think it doesn’t hurt either.

PR #984, opened 2022-10-19
Fix failing test case

ahoppen:
This test case started failing because two PRs raced.


ahoppen:
@swift-ci Please test

PR #985, opened 2022-10-19
Change trivia attribution rule to make comments trailing trivia unless separated by a newline

### LINKED ISSUE #438: [SR-9616] SwiftSyntax is not following trivia rule
|                  |                 |
   |------------------|-----------------|
   |Previous ID       | SR-9616      |
   |Radar             | None         |
   |Original Reporter | inamiy (JIRA User)      |
   |Type              | Bug    |

Attachment: [Download](https://user-images.githubusercontent.com/2727770/164963084-c9cbbe5e-57ed-4a2d-bb06-7f7926132905.gz)

      
<details>
  <summary>Environment</summary>

-   Apple Swift version 4.2.1 (swiftlang-1000.11.42 clang-1000.11.45.1)

-   [SwiftSyntax 0.40200.0](https://github.com/apple/swift-syntax/tree/0.40200.0)

</details>

   <details>
  <summary>Additional Detail from JIRA</summary>

   |                  |                 |
   |------------------|-----------------|
   |Votes             | 0         |
   |Component/s       | Compiler, SwiftSyntax    |
   |Labels            | Bug, Parser, Syntax        |
   |Assignee          | None      |
   |Priority          | Medium      |

   

   md5: b19e5c1a45f2ad0872b99a66f9211c27

  </details>

**relates to**:  
* [SR-2373](https://bugs.swift.org/browse/SR-2373) Trailing doc comment should not be attached to the next declaration  

**Issue Description:**

According to the [libSyntax documentation](https://github.com/apple/swift/tree/master/lib/Syntax):

> 1\. A token owns all of its trailing trivia up to, but not including, the next newline character.  
> 2. Looking backward in the text, a token owns all of the leading trivia up to and including the first contiguous sequence of newlines characters.

The following code should have leading / trailing trivias as follows:

``` swift
let x = 1 // This should be "1"'s trailing trivia
/// This should be `var`'s leading trivia
var y = 2
```

But actually, both comments are bundled as \`var\`'s leading trivia.

You can test the code at <https://swift-ast-explorer.kishikawakatsumi.com/> (screenshot attached).

By the way, in TypeScript, it seems to work as is:  
<https://github.com/basarat/typescript-book/blob/master/docs/compiler/ast-trivia.md>

ahoppen:
This changes the trivia attribution rule in the new parser to consider all trivia up to the next newline as trailing trivia. Previously, comments were always considered trailing trivia.

This was discussed last year in https://forums.swift.org/t/changing-comment-trivia-attribution-from-trailing-trivia-to-leading-trivia/50773.

My idea is that right now is probably one of the best times to change this because switching from the old to the new parser is an active source change for all clients.

Resolves rdar://68234477
Resolves rdar://95639215
Fixes #438

CC @allevato @jpsim

### DIFF
--- a/Sources/SwiftParser/Diagnostics/MissingNodesError.swift
+++ b/Sources/SwiftParser/Diagnostics/MissingNodesError.swift
@@ -28,14 +28,16 @@ fileprivate enum NodesDescriptionPart {
 
   func description(format: Bool) -> String? {
     switch self {
-    case .tokensWithDefaultText(let tokens):
-      let tokenContents: String
+    case .tokensWithDefaultText(var tokens):
       if format {
-        tokenContents = tokens.map({ BasicFormat().visit($0).description }).joined()
-      } else {
-        tokenContents = tokens.map(\.description).joined()
+        tokens = tokens.map({ BasicFormat().visit($0) })
       }
-      return "'\(tokenContents.trimmingWhitespace())'"
+      if !tokens.isEmpty {
+        tokens[0] = tokens[0].withLeadingTrivia([])
+        tokens[tokens.count - 1] = tokens[tokens.count - 1].withTrailingTrivia([])
+      }
+      let tokenContents = tokens.map(\.description).joined()
+      return "'\(tokenContents)'"
     case .tokenWithoutDefaultText(let token):
       if let childName = token.childNameInParent {
         return childName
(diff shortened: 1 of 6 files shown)


ahoppen:
@swift-ci Please test

CodaFi (approved):
We met to discuss the implications of this change and settled on accepting it. The forum thread contains a number of oddities that these rules still don't quite handle as well as one would hope, but the balance between "smarter" trivia attribution and an efficient trivia lexing scheme struck by this solution is better than the status quo in enough places that it's worth it.

This is also a behavior change that swiftlint and swift-format may have to adapt to.

CodaFi:
CC @jpsim and @allevato you may want to check this out and see if your respective projects behaviors change under this patch.

PR #986, opened 2022-10-19
Consume Balanced Braces in Trailing Closure Lookahead

### LINKED ISSUE #983: Failure to parse a one-line if
### Issue Kind

Parse of Valid Source Failed

### Source Code

```swift
enum Thing {
	case object([String: String?])
	case lol
}
var res = [String: Thing]()
var tmp = [String: String?]()
let includeSavedHints = false
/* This fails to parse with the new parser. */
if includeSavedHints {res["abc"] = .object(tmp.mapValues{ $0.flatMap{ $0 } ?? nil })}
```

### Description

I get the following error:
```
toto.swift: error: 9:86: expected code block in 'if' statement
<unknown>:0: error: new parser has failed consistency checking; please see errors above
```

CodaFi:
Consuming single tokens isn't enough in this position because, as this test case demonstrates, any true trailing closures inside of the braced block will end our lookahead early.

Fixes #983
rdar://101330805

### DIFF
--- a/Sources/SwiftParser/Expressions.swift
+++ b/Sources/SwiftParser/Expressions.swift
@@ -2294,7 +2294,6 @@ extension Parser.Lookahead {
     //   {...}()
     // by looking ahead for the ()'s, but this has been replaced by do{}, so this
     // probably isn't worthwhile.
-    //
     guard case .basic = flavor else {
       return true
     }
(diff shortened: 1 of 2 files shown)


CodaFi:
@swift-ci test

DougGregor (approved):

PR #988, opened 2022-10-19
Start Emitting Lexer Diagnostics

CodaFi:
The general idea here is the lexer will tag tokens with an error bit when it detects some recoverable inconsistency. In the diagnostic post-pass, we will visit these erroneous tokens and re-lex them with a diagnostic handler installed.

This approach stands in contrast with our current one of recovering with `unknown` tokens, which lose quite a bit of information and make the parse quite brittle given that the lexer can recover from every error case it handles today. I'm starting small with the numeric literal diagnostics since Strings are the real challenge here.

### DIFF
--- a/Sources/SwiftParser/Lexer.swift
+++ b/Sources/SwiftParser/Lexer.swift
@@ -204,6 +204,12 @@ extension Lexer {
     var input: UnsafeBufferPointer<UInt8>
     var previous: UInt8
 
+    @_spi(LexerDiagnostics)
+    public init(input: UnsafeBufferPointer<UInt8>, previous: UInt8) {
+      self.input = input
+      self.previous = previous
+    }
+
     public static func == (lhs: Cursor, rhs: Cursor) -> Bool {
       return lhs.input.baseAddress == rhs.input.baseAddress
     }
@@ -756,7 +762,8 @@ extension Lexer.Cursor {
 }
 
 extension Lexer.Cursor {
-  mutating func nextToken(
+  @_spi(LexerDiagnostics)
+  public mutating func nextToken(
     _ ContentStart: Lexer.Cursor,
     diagnosticHandler: ((Int, DiagnosticMessage) -> Void)? = nil
   ) -> Lexer.Lexeme {
@@ -1316,15 +1323,6 @@ extension Lexer.Cursor {
     assert((Unicode.Scalar(self.previous).isDigit || self.previous == UInt8(ascii: ".")),
            "Unexpected start")
 
-//    let expected_int_digit = { (loc: Lexer.Cursor, kind: ExpectedDigitKind) -> TokenKind in
-//      diagnose(loc, diag::lex_invalid_digit_in_int_literal, StringRef(loc, 1),
-//               (unsigned)kind)
-//      loc.advance(while: { $0.isValidIdentifierContinuationCodePoint })
-//      return .unknown(TokStart.adoptRange(upTo: loc) { stringBuf in
-//        String(decoding: stringBuf, as: UTF8.self)
-//      })
-//    }
-
     if !self.isAtEndOfFile && self.previous == UInt8(ascii: "0") && self.peek() == UInt8(ascii: "x") {
       return self.lexHexNumber(TokStart, diagnosticHandler)
     }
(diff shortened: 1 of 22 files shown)


CodaFi:
@swift-ci test

CodaFi:
@swift-ci test

CodaFi:
@swift-ci test

Sources/SwiftParser/Lexer.swift
```diff
@@ -737,14 +743,19 @@ extension Lexer.Cursor {
...
+  ) -> Lexer.Lexeme {
     // Leading trivia.
     let leadingTriviaStart = self
     let newlineInLeadingTrivia = self.lexTrivia(.leading)
 
     // Token text.
     let textStart = self
-    var (kind, flags) = self.lexImpl(ContentStart: ContentStart)
+    var (kind, flags) = self.lexImpl(
+      ContentStart: ContentStart,
```
ahoppen, line 757: Not really related to this PR, but I think `ContentStart` should be lowercased. There seem to be a few uppercase variables left in the parser that would be good to fix.

Sources/SwiftParser/Lexer.swift
```diff
@@ -1333,20 +1338,20 @@ extension Lexer.Cursor {
...
-        return (.unknown, [])
+        return (.integerLiteral, [ .isErroneous ])
       }
 
       self.advance(while: {
         $0 == Unicode.Scalar("0") || $0 == Unicode.Scalar("1") || $0 == Unicode.Scalar("_")
       })
 
-//      let tmp = self
+      let tmp = self
```
ahoppen, line 1350: Why do we need `tmp`? Can’t we use `self` below? Same in all the other cases.

Sources/SwiftParser/Diagnostics/LexerDiagnosticMessages.swift
```diff
@@ -0,0 +1,120 @@
...
+    switch self.kind {
+    case .digit(let digit):
+      return "'\(digit)' is not a valid digit in floating point exponent"
+    case .character(let char):
+      return "'\(char)' is not a valid first character in floating point exponent"
+    }
+  }
+}
+
+public struct InvalidDigitInIntegerliteral: LexerError {
```
ahoppen, line 98:
```suggestion
public struct InvalidDigitInIntegerLiteral: LexerError {
```

Sources/SwiftParser/Diagnostics/LexerDiagnosticMessages.swift
```diff
@@ -0,0 +1,120 @@
...
+    MessageID(domain: diagnosticDomain, id: "\(type(of: self)).\(self)")
+  }
+
+  public var severity: DiagnosticSeverity { .error }
+}
+
+public struct InvalidFloatingPointExponentDigit: LexerError {
+  public enum Kind {
+    case digit(Unicode.Scalar)
+    case character(Unicode.Scalar)
```
ahoppen, line 84: Mostly curious: Why didn’t you create separate `LexerError`s for these `Kind`s? AFAICT the different `Kind`s don’t share any code and that would also mean that separate diagnostic IDs would be used for the different error messages. Also, it would match what the parser is doing.

Sources/SwiftParser/Diagnostics/LexerDiagnosticMessages.swift
```diff
@@ -0,0 +1,120 @@
...
+  static var fixItID: MessageID {
+    return MessageID(domain: diagnosticDomain, id: "\(self)")
+  }
+
+  var fixItID: MessageID {
+    return Self.fixItID
+  }
+}
+
+public protocol LexerFixIt: FixItMessage {
```
ahoppen, line 51: Are we anticipating to use lexer notes and Fix-Its? If not, I think it’s fine to remove these for now (to keep things simpler) and add them once needed.

ahoppen (commented):
I don’t feel super happy with the `diagnosticHandler` in the lexer but I don’t have any better idea either, so probably it’s fine…

Sources/SwiftParser/Lexer.swift
CodaFi: That would be an off-by-one error

Sources/SwiftParser/Diagnostics/LexerDiagnosticMessages.swift
CodaFi: I didn't see any reason why these two exponent-related diagnostics should be separate strings. They ought to appear in the C++ as a select between these two messages - they do share a common tail in their error messages.

Tests/SwiftParserTest/LexerTests.swift
```diff
@@ -690,6 +690,48 @@ public class LexerTests: XCTestCase {
       ])
     }
   }
+
+  func testNumericLiteralDiagnostics() {
+    AssertParse(
+      """
+      var fl_l: Float = 0x1.0#^MISSING_HEX_EXPONENT_END^#
```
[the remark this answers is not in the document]
CodaFi: Yeah, it doesn't work. This PR was rebased but not updated 😓

CodaFi:
@swift-ci test

Sources/SwiftSyntax/Raw/RawSyntaxTokenView.swift
```diff
@@ -199,4 +199,14 @@ struct RawSyntaxTokenView {
...
   }
 
+  var hasLexerError: Bool {
+    switch raw.rawData.payload {
+    case .parsedToken(let dat):
+      return dat.hasLexerError
+    case .materializedToken(_):
+      return false
+    case .layout(_):
+      preconditionFailure("'hasError' is a token-only property")
```
ahoppen, line 209:
```suggestion
      preconditionFailure("'hasLexerError' is a token-only property")
```

Sources/SwiftParser/Diagnostics/LexerDiagnosticMessages.swift
ahoppen: 👍 Makes sense

Sources/SwiftParser/Lexer.swift
ahoppen: Ah, I missed the `advance` below. Makes sense 👍

ahoppen (approved):

CodaFi:
@swift-ci test

PR #991, opened 2022-10-20
Re-Export Targets in a Consumable Package Format

CodaFi:
Fixup the CMake build so it also writes out a find_package()-compatible module so that clients can consume that instead of having to directly include() the generated exported targets file.


CodaFi:
@swift-ci test

PR #998, opened 2022-10-21
Improve recovery of dollar identifiers

ahoppen:
I believe that the remaining diagnostics need to be diagnosed in the lexer.

### DIFF
--- a/Sources/SwiftParser/Types.swift
+++ b/Sources/SwiftParser/Types.swift
@@ -398,7 +398,9 @@ extension Parser {
       var keepGoing = true
       var loopProgress = LoopProgressCondition()
       while !self.at(any: [.eof, .rightParen]) && keepGoing && loopProgress.evaluate(currentToken) {
+        let unexpectedBeforeFirst: RawUnexpectedNodesSyntax?
         let first: RawTokenSyntax?
+        let unexpectedBeforeSecond: RawUnexpectedNodesSyntax?
         let second: RawTokenSyntax?
         let unexpectedBeforeColon: RawUnexpectedNodesSyntax?
         let colon: RawTokenSyntax?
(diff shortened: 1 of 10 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

Sources/SwiftParser/Types.swift
```diff
@@ -459,8 +465,9 @@ extension Parser {
         keepGoing = trailingComma != nil
         elements.append(RawTupleTypeElementSyntax(
             inOut: nil,
-            RawUnexpectedNodesSyntax(misplacedSpecifiers, arena: self.arena),
+            RawUnexpectedNodesSyntax(misplacedSpecifiers.map(RawSyntax.init) + (unexpectedBeforeFirst?.elements ?? []), arena: self.arena),
```
bnbarham, line 468: There's a bunch of places we're doing things like this now. Couldn't we just return `[RawSyntax]` any calls that return unexpected + make all the unexpected parameters `[RawSyntax]` instead? That would both avoid creating unexpected nodes that aren't removed + avoid looking into them to grab `elements`.

ahoppen: I don’t think I like the idea of changing the parameters of all raw nodes from `RawUnexpectedNodesSyntax` to `RawSyntax` (incidentally that would also prevent us from using all the convenience initializer on `RawUnexpectedNodesSyntax` we currently have (e.g. the one taking `[RawSyntax?]`). I put together https://github.com/apple/swift-syntax/pull/1005. I’m not entirely sure whether I like the change yet because it does add as bit of complexity on the declaration side, but the call side is rather nice. WDYT?

bnbarham: I've commented in that PR. Personally I'd find having all `[RawSyntax]` more convenient than the `[RawSyntax?]` convenience initializer. Though I haven't actually checked how many places that's being used, so maybe that wouldn't be the case. Anyway, that PR is better than the status quo so I'd be fine with merging that in.

PR #1000, opened 2022-10-21
Rename SwiftParser/Tests/Statements.swift to StatementTests.swift etc.

ahoppen:
This makes it easier to open the tests files using Xcode’s Open Quickly because they aren’t ambiguous with the corresponding files in `Sources/SwiftParser`

rdar://101334858


ahoppen:
@swift-ci Please test

PR #1003, opened 2022-10-21
Change SyntaxRewriter to disallow type node type conversions for all types that are not ExprSyntax, StmtSyntax …

ahoppen:
Intuitively, changing the kind of an expression in a rewriter makes sense but for all the other syntax kinds I can’t see any use case for changing their types.

Thus, add some type safety to `SyntaxRewriter` by requiring that all `visit` methods for nodes that don’t have a base kind like `ExprSyntax` return a node of the same type.

rdar://101355004


ahoppen:
Here’s the PR to update swift-format. In general it appears to clean things up in there: https://github.com/apple/swift-format/pull/433

bnbarham (approved):

ahoppen:
@swift-ci Please test

PR #1007, opened 2022-10-22
Make examples runnable

SimplyDanny:
Follow-up of #900.

### DIFF
--- a/Examples/.gitignore
+++ b/Examples/.gitignore
@@ -1,0 +1,2 @@
+.DS_Store
+/.build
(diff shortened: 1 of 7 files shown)


ahoppen (approved):
Thank you. It looks good to me. I think the apple/swift change is already necessary now because we’re building the example from `build-script` (thanks for adding that logic).

I just opened a PR with that change: https://github.com/apple/swift/pull/61684

Let’s see what CI has to say about it.

ahoppen:
https://github.com/apple/swift/pull/61684

@swift-ci Please test

PR #765, opened 2022-09-09
WIP `enum`erate `node_choices` and `element_choices`

rintaro:
NO NOT MERGE

PR #1010, opened 2022-10-23
If a node’s child has `node_choices` or a collection has `element_choices` use an enum to ensure there are only matching child nodes

ahoppen:
This is based on https://github.com/apple/swift-syntax/pull/765 but rebased on `main` and with all tests passing.

Previously, we relied on runtime checks to make sure only supported node types were used for a child with `node_choices`. If we explicitly model the set of supported nodes as an enum, we get compile-time safety.

rdar://101200491

### DIFF
--- a/CodeGeneration/Sources/Utils/SyntaxBuildableChild.swift
+++ b/CodeGeneration/Sources/Utils/SyntaxBuildableChild.swift
@@ -26,6 +26,18 @@ public extension Child {
     )
   }
 
+  var parameterBaseType: String {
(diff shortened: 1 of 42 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
https://github.com/apple/swift-stress-tester/pull/206

@swift-ci Please test macOS

ahoppen:
https://github.com/apple/swift-stress-tester/pull/206 

@swift-ci Please test

rintaro (approved):
Thank you for doing this!

PR #1011, opened 2022-10-24
Extend macro protocols with information about the macro types.

DougGregor:
Macros are intended to have typed inputs and outputs, so express those
via a signature and (optional) generic parameter clause to help express
that signature. This puts the entirety of the macro's "declaration" in
the conformance, so it can be queried by the macro system.


DougGregor:
@swift-ci please test

PR #1012, opened 2022-10-24
Add SwiftSyntaxBuilder convenience initializers that mix parser integration and result builders

ahoppen:
The most important use case of this is that we now allow things like

```swift
FunctionDecl("func fibonacci(_ n: Int) -> Int") {
  // result builder to generate body
}
```

Previously, the syntax tree for the signature had to be created manually.

--- 

While I was touching the convenience initializers, I merged them into a single file (which makes them easier to navigate IMO)

---

https://github.com/ahoppen/swift-syntax/commit/fe30a08fdb00b1a70ed8b33aec8cf8d0e2ad6e94 contains the corresponding changes to adopt these initializers in CodeGeneration.


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

Sources/SwiftSyntaxBuilder/ConvenienceInitializers.swift
```diff
@@ -0,0 +1,307 @@
...
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+@_spi(RawSyntax) import SwiftParser
+@_spi(RawSyntax) import SwiftSyntax
+
+extension BinaryOperatorExpr {
```
kimdv, line 17: Personally I like to use `// MARK: - BinaryOperatorExpr` etc to group code as it make it more readable IMO. 
What do you think about that?

Tests/SwiftSyntaxBuilderTest/ProtocolDeclTests.swift
```diff
@@ -16,15 +16,8 @@ import SwiftSyntaxBuilder
...
-    let input = ParameterClause {
-      FunctionParameter(firstName: .identifier("format"), colon: .colon, type: Type("Format"))
-      FunctionParameter(firstName: .identifier("leadingTrivia"), colon: .colon, type: OptionalType(wrappedType: Type("Trivia")))
-    }
-    let functionSignature = FunctionSignature(input: input, output: ReturnClause(returnType: returnType))
-
-    let buildable = ProtocolDecl(modifiers: [DeclModifier(name: .public)], identifier: "DeclListBuildable") {
-      FunctionDecl(identifier: .identifier("buildDeclList"), signature: functionSignature, body: nil)
+    let buildable = ProtocolDecl("public protocol DeclListBuildable") {
+      FunctionDecl("func buildDeclList(format: Format, leadingTrivia: Trivia?) -> [DeclSyntax]")
```
kimdv, line 20: 😍

Tests/SwiftSyntaxBuilderTest/SwitchTests.swift
```diff
@@ -10,16 +10,32 @@
...
-    self.init(digits: String(value))
-  }
-}
+final class SwitchTests: XCTestCase {
+  func testSwitch() {
+    let syntax = SwitchStmt(expression: Expr("count")) {
+      for num in 1..<3 {
+        SwitchCase("case \(num):") {
+          Expr("print(count)")
+        }
```
kimdv, line 23: So nice and easy to read! 🎉

kimdv (approved):

Sources/SwiftSyntaxBuilder/ConvenienceInitializers.swift
ahoppen: I don’t have strong opinions because each type as a single `extension` at the moment. Since you prefer them, I added them 😄

ahoppen:
@swift-ci Please test

PR #1013, opened 2022-10-24
Simplify SwiftBasicFormat by introducing `childrenKeyPaths`

ahoppen:
The core idea here is that we add `childrenKeyPaths` to all the syntax nodes, which allows us to explicitly name the parent-child relationship between a parent and a child. That way `BasicFormat` can have a list of all the children it should indent or put on a new line, reducing its implementation from 3,000 to 300 lines.

`childrenKeyPaths` should also be useful in a variety of other use cases:
- We can use it to improve `debugInitCall` to insert upcast conversions it is currently missing because it can use `childrenKeyPaths` to query the static variable type of a child
- I have a feeling that `childNameForDiagnostics` could be implemented more cleanly on top of `childrenKeyPaths`
- I will try to implement keyword completion on top of this infrastructur
- It could allow us to specify where nodes should be inserted into a syntax tree. Currently e.g. Fix-Its only support replacements.


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

PR #1017, opened 2022-10-25
Add a documentation requirement to the Macro protocol 

DougGregor:
With the `Macro` protocol being the way in which macros are declared,
with no in-source representation, we lost the ability to have a
documentation comment. Add that back with a new (defaulted)
requirement to provide documentation.


DougGregor:
@swift-ci please test

PR #1018, opened 2022-10-25
Handle object literal's rewriting of the first argument label

DougGregor:
The macro signature of the object literals is different from the initializer that's used for the rewritten version. Account for this in the macro's definition.


DougGregor:
@swift-ci please test

PR #1023, opened 2022-10-26
Don’t eat closing braces while trying to recover to a declaration start keyword

ahoppen:
Previously, when given 

```swift
class A {
  ^
}
class B {
}
```

recovery to the second `class` (because `^` is unexpected) would also consume the closing `}` nesting `B` inside `A`. That’s unexpected. Recovery of decl keywords inside member decl lists should not consume closing braces.

---

While doing this, also fix a few miscellaneous bugs in the test infrastructure.

### DIFF
--- a/Sources/_SwiftSyntaxTestSupport/SyntaxComparison.swift
+++ b/Sources/_SwiftSyntaxTestSupport/SyntaxComparison.swift
@@ -45,8 +45,8 @@ extension TreeDifference: CustomDebugStringConvertible {
   public var debugDescription: String {
     let includeTrivia = reason == .trivia
 
-    let expectedConverter = SourceLocationConverter(file: "Baseline.swift", source: baseline.description)
-    let actualConverter = SourceLocationConverter(file: "Actual.swift", source: node.description)
+    let expectedConverter = SourceLocationConverter(file: "Baseline.swift", tree: baseline.root)
+    let actualConverter = SourceLocationConverter(file: "Actual.swift", tree: node.root)
 
     let expectedDesc = baseline.debugDescription(includeTrivia: includeTrivia, converter: expectedConverter)
     let actualDesc = node.debugDescription(includeTrivia: includeTrivia, converter: actualConverter)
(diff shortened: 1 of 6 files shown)


ahoppen:
@swift-ci Please test

Sources/_SwiftSyntaxTestSupport/SyntaxComparison.swift
```diff
@@ -76,14 +76,17 @@ extension TreeDifference: CustomDebugStringConvertible {
                 """
     }
 
+    let baselineRootConverter = SourceLocationConverter(file: "Baseline.swift", source: baseline.root.description)
+    let nodeRootConverter = SourceLocationConverter(file: "Actual.swift", source: node.root.description)
```
bnbarham, line 80: May as well use these when printing the nodes above as well?

Tests/SwiftParserTest/ParserTests.swift
```diff
@@ -160,6 +160,8 @@ public class ParserTests: XCTestCase {
         || fileURL.absoluteString.contains("26773-swift-diagnosticengine-diagnose.swift")
         || fileURL.absoluteString.contains("parser-cutoff.swift")
         || fileURL.absoluteString.contains("26233-swift-iterabledeclcontext-getmembers.swift")
+        || fileURL.absoluteString.contains("26190-swift-iterabledeclcontext-getmembers.swift")
+        || fileURL.absoluteString.contains("26139-swift-abstractclosureexpr-setparams.swift")
```
bnbarham, line 164: These aren't even *that* bad, pretty surprising to me that they cause a stack overflow. Isn't even 100 nested braces!

Sources/_SwiftSyntaxTestSupport/SyntaxComparison.swift
ahoppen: Good idea 👍

Tests/SwiftParserTest/ParserTests.swift
ahoppen: The issue here appears to be debug builds. In release builds, these are just fine.

ahoppen:
@swift-ci please test

ahoppen:
@swift-ci Please test

PR #1016, opened 2022-10-25
Remove Foundation dependency in SwiftSyntaxBuilder

bnbarham:
Avoid using a `NSRegularExpression` to count the number of #'s. Walk through the contents of the string and count them directly.

Resolves rdar://101503176.

PR #1026, opened 2022-10-26
build: add dependency on Foundation

compnerd:
Now that Foundation is required for SwiftSyntax, add the ability to tie together build trees and bootstrap SwiftSyntax.  This is required to repair the Windows build.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,6 +23,9 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
 
 set(CMAKE_MACOSX_RPATH YES)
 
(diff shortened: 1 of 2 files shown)


compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

DougGregor:
No, we shouldn't need this. We accidentally grew one Foundation dependency and have removed it in https://github.com/apple/swift-syntax/pull/1016.

CodaFi (changes requested):
swift-syntax should never depend on Foundation. That would make bootstrapping for non-Darwin platforms very difficult.

compnerd:
This was fixed in #1027, #1028.

PR #1030, opened 2022-10-26
Restrict the maximum nesting level in the parser to avoid stack overflows

ahoppen:
The basic idea is that we start a new nesting level whenever we consume or synthesize an opening bracket (`<`, `{`, `(`, `[` or `#if`) and reduce it when we read the matching closing bracket. This should be a pretty simple indicator for the current nesting level that doesn’t require custom logins in productions. IMO the benefit here is that this notion of nesting level is easy to describe and explain.

Productions that are known to be able to overflow (I just took the ones that introduce `StructureMarkerRAII` in the C++ parser) can check `remainingTokensIfMaximumNestingLevelReached` and stop parsing if the maximum nesting level has been exceeded.

It turns out that a nesting level of 256 (which is the current maximum nesting level in the C++ parser) is totally fine in Release builds of the parser, but debug builds fail with nesting levels > 30-ish. 

---

Since lookahead cannot really bail out to prevent stack overflow, I rewrote `skipSingle` and `skipUntil` to work iteratively instead of recursively.

### DIFF
--- a/Sources/SwiftParser/Parser.swift
+++ b/Sources/SwiftParser/Parser.swift
@@ -26,11 +26,16 @@ extension Parser {
   }
 
   /// Parse the source code in the given string as Swift source file.
+  /// If `maximumNestingLevel` is set, the parser will stop if a nesting level
+  /// that is greater than this value is reached to avoid overflowing the stack.
+  /// The nesting level is increased whenever a bracketed expression like `(`
+  /// or `{` is stared.
   public static func parse(
     source: UnsafeBufferPointer<UInt8>,
+    maximumNestingLevel: Int? = nil,
     parseTransition: IncrementalParseTransition? = nil
   ) -> SourceFileSyntax {
-    var parser = Parser(source)
+    var parser = Parser(source, maximumNestingLevel: maximumNestingLevel)
     // Extended lifetime is required because `SyntaxArena` in the parser must
     // be alive until `Syntax(raw:)` retains the arena.
     return withExtendedLifetime(parser) {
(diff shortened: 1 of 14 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test macOS

Sources/SwiftParser/Parser.swift
```diff
@@ -122,6 +127,18 @@ public struct Parser: TokenConsumer {
...
+  /// `startNestingLevel` minus the number of tokens that `endNestingLevel`
+  /// which have been consumed so far.
+  public var nestingLevel: Int = 0
+
+  /// When this nesting level is exceeded, the parser should stop parsing.
+  public let maximumNestingLevel: Int
+
+  /// A default maximum nesting level that is used if the client didn't
+  /// explicitly specify one.
+  public static let defaultMaximumNestingLevel = 256
```
DougGregor, line 140: I know it's terrible, but should we do some kind of `#if DEBUG` check to decide whether to use the lower limit or higher limit?

Sources/SwiftParser/TopLevel.swift
```diff
@@ -13,6 +13,27 @@
 @_spi(RawSyntax) import SwiftSyntax
 
 extension Parser {
+  /// Consumes and returns all remaining tokens in the source file.
+  mutating func consumeRemaingingTokens() -> [RawSyntax] {
```
DougGregor, line 17: Typo `Remainging`

Sources/SwiftParser/TopLevel.swift
```diff
@@ -13,6 +13,27 @@
...
+    var extraneousTokens = [RawSyntax]()
+    while !self.at(.eof) {
+      extraneousTokens.append(RawSyntax(consumeAnyToken()))
+    }
+    return extraneousTokens
+  }
+
+  /// If the maximum nesting level has been reached, return the remaining tokens in the source file
+  /// as unexpected nodes that have the `isMaximumNestingLevelOverflow` bit set.
+  /// Check this in places that are likley to cause deep recursion and if this returns non-nil, abort parsing.
```
DougGregor, line 27: Typo `likley`

Tests/SwiftParserTest/ParserTests.swift
```diff
@@ -103,14 +105,7 @@ public class ParserTests: XCTestCase {
...
     runParserTests(
-      name: "Swift tests", path: testDir, checkDiagnostics: false,
-      shouldExclude: { fileURL in
-        false
-
-        // These tests overflow the parser.
-        || fileURL.absoluteString.contains("_overflow")
-        || fileURL.absoluteString.contains("parser-cutoff.swift")
-      }
+      name: "Swift tests", path: testDir, checkDiagnostics: false
```
DougGregor, line 108: This is so fantastic, thank you!

DougGregor (approved):
This is awesome, thank you!

Sources/SwiftParser/Parser.swift
ahoppen: That’s actually a good idea. The entire check is pretty horrible in itself, so a `#if DEBUG` check doesn’t make a huge difference anymore.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

PR #1031, opened 2022-10-27
Add SyntaxProtocol.token(at:) to find the token at an absolute position

DougGregor:
This is useful for finding the token at a particular absolute position.


DougGregor:
@swift-ci please test

ahoppen (approved):

Issue #1025, opened 2022-10-26
Crash when parsing initializer with unavailable attribute


jpsim:
### Issue Kind

Parser Crash

### Source Code

```swift
struct Math {
  public static let pi = 3.14
  @available(*, unavailable) init() {}
}
```

### Description

Tested with 79c8e55d2773c249342ae21abd8a2e0df88ebc5f (`main` as of today):

```swift
let source = """
  struct Math {
    public static let pi = 3.14
    @available(*, unavailable) init() {}
  }
  """
dump(Parser.parse(source: source))
```

Results in this fatal error unwrapping an optional:

> SwiftSyntax/SyntaxNodes.swift:24664: Fatal error: Unexpectedly found nil while unwrapping an Optional value

jpsim:
Seems to have been introduced in d3471fdb4b49c9f91343b587af8f531f8bbb4a4c

ahoppen:
rdar://101589498

ahoppen:
Ah, we are now taking advantage of the `node_choices` constraints with the commit you mentioned above but were aren’t actually enforcing that these constraints are upheld in the parser, causing the crash. I’m working on resolving this in https://github.com/apple/swift-syntax/pull/1032.

jpsim:
Glad to see you're on top of it.

Should d3471fdb4b49c9f91343b587af8f531f8bbb4a4c be reverted if it's preventing `main` from parsing valid syntax?

Especially if the fix from #1032 isn't going to be ready quickly?

ahoppen:
https://github.com/apple/swift-syntax/pull/1034 should fix the issue and I hope to land it today or at least tomorrow. I hope that’s sufficient.

jpsim:
Thanks for the update, that PR looks good on my end, and I'm not in a big rush to update on `main`, but the pace of development and breaking changes is fast enough that I'd rather not get too far behind.

jpsim:
There's still a `SIGTRAP` crash with ecd2ba9c7edc2e82904a3e32790ee0a1867059f7 compared to 4ad75a85f5f14caecdd6cdb41c4c9a090ce82340 though when parsing source files from the the `apple/swift` repo. I'll dig further into that to narrow it down.

jpsim:
Here's the file that's crashing: https://github.com/apple/swift/blob/main/stdlib/public/Distributed/DistributedActor.swift

jpsim:
[...]

PR #1032, opened 2022-10-27
WIP: Validate node_choices and token_choices

ahoppen:
The python files put structural restrictions on the allowed syntax node kinds for `node_choices`, `token_choices` and `text_choices` but we weren’t enforcing them anywhere during creation of raw nodes.

With https://github.com/apple/swift-syntax/pull/1010 we are actually making use of these restrictions and are failing if the invariant we impose on ourselves are not upheld, which caused https://github.com/apple/swift-syntax/issues/1025.

This is still very much WIP because most test cases are hitting one of the new assertion failures now.

---

Alternative/additional ideas:
- We should also have something like `SyntaxNodeChoices` for raw nodes to enforce the structural `node_choices` constraints at compile time.
- It would be really nice if we could also enforce the `token_choices` and `text_choices` constraints at compile time but I’m not sure if there’s a good way to model that.


ahoppen:
Closing in favor of https://github.com/apple/swift-syntax/pull/1034

PR #1033, opened 2022-10-27
Make sure that any keyword or identifier inserted by a Fix-It is separated from the previous one by trivia

ahoppen:
Previously, the now updated test case generated `for iin <#expression#> {}` when applying the Fix-It.


ahoppen:
@swift-ci Please test

PR #1034, opened 2022-10-27
Prevent incorrect `node_choices` and ignore `token_choices`

ahoppen:
The python files put structural restrictions on the allowed syntax node kinds for `node_choices`, `token_choices` and `text_choices` but we weren’t enforcing them anywhere during creation of raw nodes.

With https://github.com/apple/swift-syntax/pull/1010 we are actually making use of these restrictions and are failing if the invariant we impose on ourselves are not upheld, which caused https://github.com/apple/swift-syntax/issues/1025.

After some discussion with @rintaro and @bnbarham we agreed that we should:
- Make sure no incorrect node types are passed to nodes that have `node_choices` by introducing the equivalent of `SynaxChildChoices` in the `RawSyntax` world
- Disallow multiple different token kinds as different `node_choices`. All tokens should have the same `node_choice`. This way we don’t need to look at token kinds to determine which case of a `SyntaxChildEnum` a token belongs to
- Once we drop support of the C++ parser, we can delete `token_choices` and `text_choices`. Since they weren’t enforced, they were misleading and are not providing any real value. 

Fixes https://github.com/apple/swift-syntax/issues/1025
Resolves rdar://101589498


ahoppen:
@swift-ci Please test

Sources/SwiftSyntax/SyntaxNodes.swift.gyb.template
```diff
@@ -78,28 +78,7 @@ public struct ${node.name}: ${base_type}Protocol, SyntaxHashable {
...
 %         for choice in child.node_choices:
-%           if choice.is_token():
-%             token_choices += [choice]
-%           else:
-%             other_choices += [choice]
-%           end
-%         end
-%
-%         if token_choices:
-      if let tok = syntaxNode.as(TokenSyntax.self) {
```
rintaro, line 92: We should be able to have `init(_: TokenSyntax)` initializer now.

ahoppen: Good catch. Thanks

ahoppen:
@swift-ci Please test

Sources/SwiftSyntax/SyntaxNodes.swift.gyb.template
```diff
@@ -65,9 +65,7 @@ public struct ${node.name}: ${base_type}Protocol, SyntaxHashable {
 %           choice_node = NODE_MAP.get(choice.syntax_kind)
 %           # We don't add 'init(_: TokenSyntax)' because we need to check the
 %           # token kind.
-%           if choice.is_token():
-%             pass
```
rintaro, line 69: Please remove the comment too 🙏 
```
%           # We don't add 'init(_: TokenSyntax)' because we need to check the
%           # token kind.
```

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

PR #1038, opened 2022-10-28
Remove UserInaccesible from SetterAccess

bnbarham:
`__setter_access` is similar to `_custom` in that this name isn't actually used when printing. So even though it's double underscored, it should still be user accessible.


bnbarham:
@swift-ci please test

bnbarham:
@swift-ci please test

rintaro (approved):

PR #1048, opened 2022-11-03
Remove XXXRawSyntax.withYYY() methods

rintaro:
This should save some code size.

### DIFF
--- a/Sources/SwiftParser/Declarations.swift
+++ b/Sources/SwiftParser/Declarations.swift
@@ -167,7 +167,13 @@ extension Parser {
           arena: parser.arena)
       } addSemicolonIfNeeded: { lastElement, newItemAtStartOfLine, parser in
         if lastElement.semicolon == nil && !newItemAtStartOfLine {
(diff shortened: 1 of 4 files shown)


rintaro:
@swift-ci Please test

Sources/SwiftParser/Declarations.swift
```diff
@@ -167,7 +167,13 @@ extension Parser {
           arena: parser.arena)
       } addSemicolonIfNeeded: { lastElement, newItemAtStartOfLine, parser in
         if lastElement.semicolon == nil && !newItemAtStartOfLine {
-          return lastElement.withSemicolon(parser.missingToken(.semicolon, text: nil), arena: parser.arena)
```
bnbarham, line 170: Totally fine with this change for now. But one thing I was thinking of... What if we instead had copy initializers that take the previous node + optional params + leading/trailing trivia? For both the red + green tree. Then this would be eg.
```
return RawMemberDeclListItemSyntax(lastElement, semicolon: parser.missingToken(.semicolon, text: nil)))
```

CodaFi (approved):

PR #1050, opened 2022-11-03
Eliminate SyntaxArena.default

rintaro:
SyntaxArena.default was a global arena which just leaked. Eliminate it, and make all Syntax factory/modifier API use a new arena.

rdar://101894396

### DIFF
--- a/Sources/SwiftSyntax/Raw/RawSyntax.swift
+++ b/Sources/SwiftSyntax/Raw/RawSyntax.swift
@@ -219,7 +219,8 @@ extension RawSyntax {
   /// If the syntax tree did not contain a token and thus no trivia could be attached to it, `nil` is returned.
   /// - Parameters:
   ///   - leadingTrivia: The trivia to attach.
-  func withLeadingTrivia(_ leadingTrivia: Trivia) -> RawSyntax? {
+  ///   - arena: SyntaxArena to the result node data resides.
+  func withLeadingTrivia(_ leadingTrivia: Trivia, arena: SyntaxArena) -> RawSyntax? {
     switch view {
     case .token(let tokenView):
       return .makeMaterializedToken(
@@ -229,7 +230,7 @@ extension RawSyntax {
         arena: arena)
     case .layout(let layoutView):
       for (index, child) in layoutView.children.enumerated() {
-        if let replaced = child?.withLeadingTrivia(leadingTrivia) {
+        if let replaced = child?.withLeadingTrivia(leadingTrivia, arena: arena) {
           return layoutView.replacingChild(at: index, with: replaced, arena: arena)
         }
       }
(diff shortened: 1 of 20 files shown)


rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

jpsim:
Big +1 from me, this is pretty amazing.

In terms of memory usage, I'm seeing a 2-3x improvement over `main` with #1050, going from 465MB to 162MB in one benchmark.

In terms of time performance, SwiftLint sees a 19% to 47% (!!) overall processing time speedup, and that's considering other processing time not related to SwiftSyntax.

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

rintaro:
@jpsim I updated the PR. I don't think this would change the performance much, but could you test it again, just to make sure?

Sources/SwiftSyntax/Raw/RawSyntax.swift
```diff
@@ -219,7 +219,7 @@ extension RawSyntax {
   /// If the syntax tree did not contain a token and thus no trivia could be attached to it, `nil` is returned.
   /// - Parameters:
   ///   - leadingTrivia: The trivia to attach.
-  func withLeadingTrivia(_ leadingTrivia: Trivia) -> RawSyntax? {
+  func withLeadingTrivia(_ leadingTrivia: Trivia, arena: SyntaxArena) -> RawSyntax? {
```
bnbarham, line 222: Is this actually needed? We're always using the one arena for `RawSyntax` anyway aren't we? You also removed these in your other PR, so there'll be a conflict (?)

bnbarham (approved):

CodaFi (approved):
💯

jpsim:
I see similar performance with 28813884a1659f3cd2fa61b5b3e5d01e8c25bc34

Sources/SwiftSyntax/Raw/RawSyntax.swift
```diff
@@ -219,7 +219,7 @@ extension RawSyntax {
   /// If the syntax tree did not contain a token and thus no trivia could be attached to it, `nil` is returned.
   /// - Parameters:
   ///   - leadingTrivia: The trivia to attach.
```
jpsim, line 221: Not sure if you care about keeping parameter docstrings up to date, but if so this one’s missing.

rintaro:
@swift-ci Please test

Sources/SwiftSyntax/Raw/RawSyntax.swift
rintaro: Thanks for noticing. Updated.

rintaro:
@swift-ci Please test

PR #1051, opened 2022-11-04
[Macros] Fix rewrite implementation for _CompilerPluginSupport

DougGregor:
We were computing the location of the macro incorrectly, and unnecessarily stringifying the source buffer.


DougGregor:
@swift-ci please test

CodaFi (approved):

PR #1052, opened 2022-11-04
Add an Editor Extension for Testing Refactorings 

CodaFi:
Add an editor extension that scans itself for refactoring actions and
translates them into a form that Xcode can ingest via a source editor
plugin. At the moment this applies the selected transformation to every
input instance in the entire buffer.


CodaFi:
@swift-ci test

PR #1053, opened 2022-11-04
Eliminate PlatformMutex

rintaro:
Now that a new syntax arena is used for every mutation/node creation. And the arena is never reused for allocation once the transaction completes. So there's no chance that multiple threads allocates memory, or adding children to an arena at the same time as long as the user dealing with red-tree API only. That means mutex is now useless.

Of course there still be chances to cause data races when a user uses SyntaxArena or RawSyntax directly. But it's now user's responsibility to manage threads accordingly.


rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test macOS

Sources/SwiftSyntax/SyntaxText.swift
```diff
@@ -10,14 +10,6 @@
 //
 //===----------------------------------------------------------------------===//
 
-#if !SWIFT_SYNTAX_ALWAYS_SINGLE_THREADED
-#if canImport(Darwin)
-@_implementationOnly import Darwin
-#elseif canImport(Glibc)
-@_implementationOnly import Glibc
-#endif
-#endif
```
rintaro, line 19: Forgot to remove this. This should resolve "ambiguous `memcmp`" errors. But may affects Linkage tests. checking locally.

rintaro:
@swift-ci Please test

Sources/SwiftSyntax/SyntaxText.swift
rintaro: Updated the test.

bnbarham (approved):
I'm fairly surprised this doesn't actually improve the SwiftLint performance numbers (https://github.com/realm/SwiftLint/pull/4513). Is the `memcmp` slower? In general this seems like it should only improve things, I'm definitely all for removing the lock.

jpsim:
If you want to check why this is slowing down SwiftLint (which I agree is surprising), you can run the same benchmarks as CI locally:

```
git clone https://github.com/realm/SwiftLint.git
cd SwiftLint
git checkout update-swiftsyntax
./tools/oss-check
```

You do need [bazelisk](https://github.com/bazelbuild/bazelisk) installed.

I can also try to reduce a more isolated repro case.

jpsim:
I think I've narrowed it down to this file taking 3x longer to process with this PR vs main: https://github.com/wireapp/wire-ios/blob/develop/WireCommonComponents/Icons/Autogenerated/WireStyleKit.swift

jpsim:
```
$ hyperfine \
  "./swiftlint-main --lenient --no-cache --enable-all-rules --quiet /Wire/WireCommonComponents/Icons/Autogenerated/WireStyleKit.swift" \
  "./swiftlint-branch --lenient --no-cache --enable-all-rules --quiet /Wire/WireCommonComponents/Icons/Autogenerated/WireStyleKit.swift"
Benchmark 1: ./swiftlint-main --lenient --no-cache --enable-all-rules --quiet /Wire/WireCommonComponents/Icons/Autogenerated/WireStyleKit.swift
  Time (mean ± σ):      7.831 s ±  0.093 s    [User: 10.407 s, System: 21.420 s]
  Range (min … max):    7.744 s …  7.984 s    10 runs
 
Benchmark 2: ./swiftlint-branch --lenient --no-cache --enable-all-rules --quiet /Wire/WireCommonComponents/Icons/Autogenerated/WireStyleKit.swift
  Time (mean ± σ):     21.011 s ±  0.446 s    [User: 113.984 s, System: 0.199 s]
  Range (min … max):   20.285 s … 21.954 s    10 runs
 
Summary
  './swiftlint-main --lenient --no-cache --enable-all-rules --quiet /Wire/WireCommonComponents/Icons/Autogenerated/WireStyleKit.swift' ran
    2.68 ± 0.07 times faster than './swiftlint-branch --lenient --no-cache --enable-all-rules --quiet /Wire/WireCommonComponents/Icons/Autogenerated/WireStyleKit.swift'
```

Only difference is `swiftlint-main` was built with e1f771ea8ad8bc172de766e0963a76aeb7b08cef and `swiftlint-branch` with built with 392944016fcfc16ea07fe8658c2c5d23cffa6433.

jpsim:
[swift_syntax_bench.swift.zip](https://github.com/apple/swift-syntax/files/9944736/swift_syntax_bench.swift.zip)

ok this is a self-contained, minimal-ish repro. 200 lines. Compiled with `swift build -c release`.

With e1f771ea8ad8bc172de766e0963a76aeb7b08cef it runs in 4.2s.
With 392944016fcfc16ea07fe8658c2c5d23cffa6433 it runs in 9.6s.

```
hyperfine "./swift-syntax-bench-main" "./swift-syntax-bench-pr"
Benchmark 1: ./swift-syntax-bench-main
  Time (mean ± σ):      4.250 s ±  0.172 s    [User: 4.030 s, System: 7.793 s]
  Range (min … max):    4.042 s …  4.619 s    10 runs
 
Benchmark 2: ./swift-syntax-bench-pr
  Time (mean ± σ):      9.663 s ±  0.508 s    [User: 36.729 s, System: 0.047 s]
  Range (min … max):    8.987 s … 10.552 s    10 runs
 
Summary
  './swift-syntax-bench-main' ran
    2.27 ± 0.15 times faster than './swift-syntax-bench-pr'
```

rintaro:
Thank you JP! This is interesting. With `main`, pthread mutex related functions takes up most of the time. But in this PR, `swift_retain`/`swift_release` does. The number of retain/release calls _should_ be the same, but... 🤔 

* main
  Screen Shot 2022-11-05 at 7 47 36 PM

* pr-1053
  Screen Shot 2022-11-05 at 7 47 26 PM

bnbarham:
> Thank you JP! This is interesting. With `main`, pthread mutex related functions takes up most of the time. But with this PR, `swift_retain`/`swift_release` does. The number of retain/release calls _should_ be the same, but... 🤔

Thanks for reducing that so much JP. Did you record and compare reference counts between them @rintaro?

rintaro:
> Did you happen to record the reference counts

The only way to count `swift_retain` correctly is to run it with a symbolic breakpoint and count it, which is too slow.
So I tried logging (`print()`) right before `children.insert(other)` (Over 50% of retain is recored in `SyntaxArena.addChild(_:)`). The number of line was exactly the same between them

rintaro:
I still don't know why `swift_retain/release` time is such different between pre/post this PR. But probably, a hot pair of retain/release is somehow optimized away in `main`.
But anyway, I added a commit to reduce ref-counting traffic significantly, at least in this specific case. Testing locally...

jpsim (approved):
Optimizing that retain/release addressed the performance degradation I was seeing in SwiftLint.

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

CodaFi:
Hm, I'm seeing the same thing

@swift-ci clean test

CodaFi (approved):

CodaFi:
@swift-ci test Linux

rintaro:
@swift-ci Please clean test macOS

PR #1056, opened 2022-11-06
Use generic SyntaxProtocol initializer

bnbarham:
Rather than all syntax nodes taking a `Syntax`, instead make the initializer generic over a `SyntaxProtocol`. This is particularly useful now that each child has an enum for its kind.

### DIFF
--- a/Sources/SwiftSyntax/Syntax.swift
+++ b/Sources/SwiftSyntax/Syntax.swift
@@ -71,12 +71,6 @@ public struct Syntax: SyntaxProtocol, SyntaxHashable {
   public init<S: SyntaxProtocol>(_ syntax: S) {
     self = syntax._syntaxNode
   }
-
-  /// Create a `Syntax` node from a specialized optional syntax node.
-  public init?<S: SyntaxProtocol>(_ syntax: S?) {
-    guard let syntax = syntax else { return nil }
-    self = syntax._syntaxNode
-  }
   
   public init(fromProtocol syntax: SyntaxProtocol) {
     self = syntax._syntaxNode
@@ -111,21 +105,6 @@ public struct Syntax: SyntaxProtocol, SyntaxHashable {
 
   public static func ==(lhs: Syntax, rhs: Syntax) -> Bool {
     return lhs.data.nodeId == rhs.data.nodeId
-  }
-}
-
-// Casting functions to specialized syntax nodes.
-extension SyntaxProtocol {
-  public func `is`<S: SyntaxProtocol>(_ syntaxType: S.Type) -> Bool {
-    return self.as(syntaxType) != nil
-  }
-
-  public func `as`<S: SyntaxProtocol>(_ syntaxType: S.Type) -> S? {
-    return S.init(self._syntaxNode)
-  }
-
-  func cast<S: SyntaxProtocol>(_ syntaxType: S.Type) -> S {
-    return self.as(S.self)!
   }
 }
 
@@ -188,9 +167,9 @@ public protocol SyntaxProtocol: CustomStringConvertible,
   /// Do not retrieve this property directly. Use `Syntax(self)` instead.
   var _syntaxNode: Syntax { get }
 
-  /// Converts the given `Syntax` node to this type. Returns `nil` if the
+  /// Converts the given specialized node to this type. Returns `nil` if the
   /// conversion is not possible.
-  init?(_ syntaxNode: Syntax)
+  init?<S: SyntaxProtocol>(_ node: S)
 
   /// The statically allowed structure of the syntax node.
   static var structure: SyntaxNodeStructure { get }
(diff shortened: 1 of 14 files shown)


bnbarham:
@swift-ci please test

Sources/SwiftSyntax/Syntax.swift
```diff
@@ -71,12 +71,6 @@ public struct Syntax: SyntaxProtocol, SyntaxHashable {
   public init<S: SyntaxProtocol>(_ syntax: S) {
     self = syntax._syntaxNode
   }
-
-  /// Create a `Syntax` node from a specialized optional syntax node.
```
bnbarham, line 75: This is now defined by an extension for all SyntaxProtocol's

Sources/SwiftSyntax/Syntax.swift
```diff
@@ -202,6 +181,28 @@ public protocol SyntaxProtocol: CustomStringConvertible,
...
+    }
+    self.init(syntaxNode)
+  }
+
+  func `is`<S: SyntaxProtocol>(_ syntaxType: S.Type) -> Bool {
+    return self.as(syntaxType) != nil
+  }
+
+  func `as`<S: SyntaxProtocol>(_ syntaxType: S.Type) -> S? {
+    return S.init(self._syntaxNode)
```
rintaro, line 198: This could be `S(self)` (or `S.init(self)`, right?

Sources/SwiftSyntax/SyntaxNodes.swift.gyb.template
```diff
@@ -100,11 +100,9 @@ public struct ${node.name}: ${base_type}Protocol, SyntaxHashable {
...
   public let _syntaxNode: Syntax
 
-  /// Converts the given `Syntax` node to a `${node.name}` if possible. Returns
-  /// `nil` if the conversion is not possible.
-  public init?(_ syntax: Syntax) {
-    guard syntax.raw.kind == .${node.swift_syntax_kind} else { return nil }
-    self._syntaxNode = syntax
+  public init?<S: SyntaxProtocol>(_ syntaxNode: S) {
+    guard syntaxNode.raw.kind == .${node.swift_syntax_kind} else { return nil }
+    self._syntaxNode = Syntax(syntaxNode)
```
rintaro, line 105: Personally, `self._syntaxNode = syntaxNode._syntaxNode` looks more clear what this is doing.

Sources/SwiftSyntax/SyntaxNodes.swift.gyb.template
```diff
@@ -100,11 +100,9 @@ public struct ${node.name}: ${base_type}Protocol, SyntaxHashable {
 %
   public let _syntaxNode: Syntax
 
-  /// Converts the given `Syntax` node to a `${node.name}` if possible. Returns
-  /// `nil` if the conversion is not possible.
-  public init?(_ syntax: Syntax) {
-    guard syntax.raw.kind == .${node.swift_syntax_kind} else { return nil }
-    self._syntaxNode = syntax
+  public init?<S: SyntaxProtocol>(_ syntaxNode: S) {
```
rintaro, line 103: `SyntaxBaseNodes.swift.gyb` has `init<S: ${node.name}Protocol>(_ syntax: S)`. The parameter name should be consistent. Same for `init?<S: SyntaxProtocol>(_:)` in `SyntaxBaseNodes.swift.gyb`

Sources/SwiftSyntax/SyntaxOtherNodes.swift
```diff
@@ -16,16 +16,11 @@
...
-    guard syntax.raw.kind == .unknown else { return nil }
-    self._syntaxNode = syntax
+  public init?<S: SyntaxProtocol>(_ syntaxNode: S) {
+    guard syntaxNode.raw.kind == .unknown else { return nil }
+    self._syntaxNode = Syntax(syntaxNode)
   }
 
-  /// Creates an `UnknownSyntax` node from the given `SyntaxData`. This assumes
-  /// that the `SyntaxData` is of the correct kind. If it is not, the behaviour
-  /// is undefined.
```
rintaro, line 28: Is this intentional? I'm not against this change, but could you do this for other types syntax nodes too?

Sources/SwiftSyntax/SyntaxBaseNodes.swift.gyb
```diff
@@ -59,12 +59,6 @@ public struct ${node.name}: ${node.name}Protocol, SyntaxHashable {
     self.init(syntax._syntaxNode.data)
   }
 
-  /// Create a `${node.name}` node from a specialized optional syntax node.
-  public init?<S: ${node.name}Protocol>(_ syntax: S?) {
-    guard let syntax = syntax else { return nil }
-    self.init(syntax)
```
rintaro, line 65: I'm not sure we should delete this.
This exists for delegating to specialized non-nil `init(_:)`. For example,
```
func test(declOpt: FuncDeclSyntax?) -> DeclSyntax? {
  return DeclSyntax(declOpt)
}
```
should call `DeclSyntax.init<S: DeclSyntaxProtocol>(_: S)`, but not `.init?<S: SyntaxProtocol>(_: S)`.
I don't think `SyntaxProtocol.init?<S: SyntaxProtocol>(_: S)` would do it.

Sources/SwiftSyntax/Syntax.swift
bnbarham: Ah, sure. I just moved these without really thinking about them. Thanks

Sources/SwiftSyntax/SyntaxBaseNodes.swift.gyb
bnbarham: Ah, I suppose we'd be skipping the fast path in this case yeah. Fair enough, I'll add it back in.

Sources/SwiftSyntax/SyntaxNodes.swift.gyb.template
bnbarham: Which do you prefer?

Sources/SwiftSyntax/SyntaxOtherNodes.swift
bnbarham: I was removing comments where the protocol already had them since they seemed mostly pointless in that case. This one is unintentional though because this is just defined by each node. If we want to get rid of it we could define an internal conformance I guess, but I'll leave it for now.

Sources/SwiftSyntaxBuilder/generated/ResultBuilders.swift
```diff
@@ -264,6 +264,11 @@ public struct AttributeListBuilder {
   public static func buildExpression(_ expression: CustomAttribute) -> Self.Component {
     return buildExpression(.init (expression))
   }
+  /// If declared, provides contextual type information for statement
```
bnbarham, line 267: I haven't gone looking for the commit that missed adding this 🤷

bnbarham:
@swift-ci please test

rintaro (approved):

CodaFi (approved):

Issue #829, opened 2022-09-22
`default` is not remapped to an identifier


ahoppen:
### Issue Kind

Other

### Source Code

```swift
data[position, default: 0]
```

### Description

Produces the following syntax treee
```
SourceFileSyntax children=2
  0: CodeBlockItemListSyntax children=1
    0: CodeBlockItemSyntax children=1
      0: SubscriptExprSyntax children=4
        0: IdentifierExprSyntax children=1
          0: identifier("data")
        1: leftSquareBracket
        2: TupleExprElementListSyntax children=2
          0: TupleExprElementSyntax children=2
            0: IdentifierExprSyntax children=1
              0: identifier("position")
            1: comma
          1: TupleExprElementSyntax children=3
            0: defaultKeyword
            1: colon
            2: IntegerLiteralExprSyntax children=1
              0: integerLiteral("0")
        3: rightSquareBracket
  1: eof
```

`defaultKeyword` should be remapped to an `identifier` in this case.

ahoppen:
rdar://100262158

bnbarham:
Fixed in https://github.com/apple/swift-syntax/pull/1057

PR #1057, opened 2022-11-06
Add remapping  to identifier when parsing argument label

ibrahimoktay:
Fixes the issue #829 by remapping the `default keyword` to identifier.

### DIFF
--- a/Sources/SwiftParser/Names.swift
+++ b/Sources/SwiftParser/Names.swift
@@ -31,6 +31,9 @@ extension Parser {
         self.missingToken(.identifier, text: nil)
       )
     } else {
+      if let wildcardToken = self.consume(if: .wildcardKeyword) {
+        return (nil, wildcardToken)
+      }
       return (nil, self.consumeAnyToken(remapping: .identifier))
     }
   }
--- a/Sources/SwiftParser/Names.swift
+++ b/Sources/SwiftParser/Names.swift
@@ -31,7 +31,7 @@ extension Parser {
         self.missingToken(.identifier, text: nil)
       )
     } else {
-      return (nil, self.consumeAnyToken())
+      return (nil, self.consumeAnyToken(remapping: .identifier))
     }
   }
 }
--- a/Tests/SwiftParserTest/DeclarationTests.swift
+++ b/Tests/SwiftParserTest/DeclarationTests.swift
@@ -73,6 +73,22 @@ final class DeclarationTests: XCTestCase {
     )
 
     AssertParse("func /^ (lhs: Int, rhs: Int) -> Int { 1 / 2 }")
+    
+    AssertParse(
+      """
+      func name(_ default: Int) {}
+      """,
+      substructure: Syntax(FunctionParameterSyntax(
+        attributes: nil,
+        modifiers: nil,
+        firstName: .wildcardKeyword(),
+        secondName: .identifier("default"),
+        colon: TokenSyntax.colonToken(),
+        type: TypeSyntax(SimpleTypeIdentifierSyntax(name: TokenSyntax.identifier("Int"), genericArgumentClause: nil)),
+        ellipsis: nil,
+        defaultArgument: nil,
+        trailingComma: nil))
+    )
   }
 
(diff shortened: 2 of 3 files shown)


CodaFi:
Awesome. There's probably a few other places we're failing to do the remapping but this will stamp out a whole class of bugs.

@swift-ci test

CodaFi (approved):

CodaFi:
@swift-ci test Linux

CodaFi:
@swift-ci test

bnbarham:
@ibrahimoktay thanks for fixing this! The failure is because a dependent project (swift-stress-tester) had a test that now fails with your fix. It happened to have a FIXME there already, so I just updated it.

@CodaFi these need to be merged together.

ibrahimoktay:
> @ibrahimoktay thanks for fixing this! The failure is because a dependent project (swift-stress-tester) had a test that now fails with your fix. It happened to have a FIXME there already, so I just updated it.
> 
> @CodaFi these need to be merged together.

Thanks for helping @bnbarham I tried to check the details but I was not able to understand what was failing. Is there a way to see where it fails ?

CodaFi:
sourcekit-lsp uses token kinds for classification so I'll go update those tests as well.

CodaFi:
Ah, this is a proper failure. This patch now makes us remap underscores as identifiers. They ought to remain as keywords.

bnbarham:
> Thanks for helping @bnbarham I tried to check the details but I was not able to understand what was failing. Is there a way to see where it fails ?

In the build logs, though they might not have been super easy to find if you looked after I started the tests again (you could find it by going to https://ci.swift.org/job/swift-syntax-PR-macOS/ and looking for the last failure for this PR)  - https://ci.swift.org/job/swift-syntax-PR-macOS/1075/console
```
Test Case '-[StressTesterToolTests.ActionGeneratorTests testCompletionExpectedResults]' started.
/Users/ec2-user/jenkins/workspace/swift-syntax-PR-macOS/branch-main/swift-stress-tester/SourceKitStressTester/Tests/StressTesterToolTests/ActionGeneratorTests.swift:97: error: -[StressTesterToolTests.ActionGeneratorTests testCompletionExpectedResults] : XCTAssertEqual failed: ("["data", "position", "default:"]") is not equal to ("["data", "position"]")
Test Case '-[StressTesterToolTests.ActionGeneratorTests testCompletionExpectedResults]' failed (0.518 seconds).
```

The latest failure is in sourcekit-lsp in `testSemanticTokensForFunctionSignatures` (search for "1 failure" in https://ci.swift.org/job/swift-syntax-PR-macOS/1080/consoleText). It's a little harder to understand in the format there, but if you split out the ranges onto lines then you'll see the difference is:
```
SourceKitLSP.SyntaxHighlightingToken(range: Range(Position(line: 0, utf16index: 15)..<Position(line: 0, utf16index: 16)), kind: SourceKitLSP.SyntaxHighlightingToken.Kind.identifier, ...),
```
vs
```
SourceKitLSP.SyntaxHighlightingToken(range: Range(Position(line: 0, utf16index: 15)..<Position(line: 0, utf16index: 16)), kind: SourceKitLSP.SyntaxHighlightingToken.Kind.keyword, modifiers: ...), 
```

And if you look at the test case, that range is an `_`, which is what @CodaFi mentioned above. So `parseArgumentLabel` needs to leave underscore as a keyword rather than remapping to an identifier, which we should add a test for here as well.

ibrahimoktay:
I have kept the remapping and just excluded the `.wildcard` which seemed to be a better approach to me (might fix other possible issues). Would you prefer to remap only if it is `.defaultKeyword` so that It could be safer?  @CodaFi @bnbarham

bnbarham:
> I have kept the remapping and just excluded the `.wildcard` which seemed to be a better approach to me (might fix other possible issues). Would you prefer to remap only if it is `.defaultKeyword` so that It could be safer? @CodaFi @bnbarham

LGTM, thanks.

bnbarham:
Thanks @ibrahimoktay!

PR #1062, opened 2022-11-07
Add Binary Operators to the Exception List for 'yield'

CodaFi:
This is a stopgap that prevents us from parsing yield followed by an operator as a yield statement. This is not a fantastic solution in the long run because

```
_modify {
  yield += x
}
```

is meant to actually be a parse-time error under the current rules. I suspect the proper solution is to have an UnresolvedYieldSyntax node and make either ASTGen or operator folding do the right thing here. For now, fix the regression in parsing.


DougGregor (approved):

CodaFi:
@swift-ci test

PR #1064, opened 2022-11-08
[NFC] Rename `SyntaxData.replacingChild(_:at:arena:)`

rintaro:
to `SyntaxData.replacingChild(at:with:arena)`. This is consistent with `RawSyntax`.


rintaro:
@swift-ci Please test

PR #1067, opened 2022-11-09
Escape newlines in `StringLiteralExprSyntax(contents:)` initializer.

DougGregor:


DougGregor:
@swift-ci please test

bnbarham (approved):
Could add a `StringLiteralTests` test:
```
  func testNewlines() {
    AssertBuildResult(
      StringLiteralExpr(content: "linux\nwindows\r\n"), "linux\\nwindows\\r\\n"
    )
  }
```

LGTM otherwise though

EDIT: Actually, what about `\u`? Both in the general case and because there's a bunch of other newline characters

DougGregor:
Great suggestions, @bnbarham , I've updated the PR accordingly.

DougGregor:
@swift-ci please test

Sources/SwiftSyntaxBuilder/ConvenienceInitializers.swift
```diff
@@ -192,6 +192,28 @@ extension MemberAccessExpr {
...
+    var result = ""
+    var input = self[...]
+    while let firstNewline = input.firstIndex(where: { $0.isNewline }) {
+      result += input[..<firstNewline]
+      if input[firstNewline] == "\r" {
+        result += "\\r"
+      } else if input[firstNewline] == "\r\n" {
+        result += "\\r\\n"
+      } else {
+        result += "\\n"
```
bnbarham, line 207: Replacing is probably fair enough, though you could also do `result += input[firstNewLine].unicodeScalars.map({ $0.escaped(asASCII: true) }).joined())` 🤷

PR #1070, opened 2022-11-10
Add Refactoring Actions to Reformat Integer Literals

CodaFi:


CodaFi:
@swift-ci test

allevato:
Strategy-wise, I'm curious how you all envision this new package which is likely to contain rules that overlap with those already in swift-format; for example, this is similar (but not identical) to https://github.com/apple/swift-format/blob/main/Sources/SwiftFormatRules/GroupNumericLiterals.swift.

One possibility is that I can imagine swift-format being a frontend for these, by us creating an adaptor for refactoring actions so that they can be applied by our pipeline and controlled in someone's configuration.

But mainly, I want to make sure we're not duplicating work or having two similar but still different implementations of mostly the same concept in both repositories.

CodaFi:
I'd very much like for SwiftRefactor to become the de-facto host for syntactic SwiftSyntax transformations. I anticipate this library having a broad set of clients - from IDEs powered by sourcekit-lsp turning these into [code actions](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_codeAction) to code-correcting formatters like swift-format and SwiftLint acting as porcelain over the raw refactoring actions.

PR #1071, opened 2022-11-10
[Macros] Add notion of owning module and supplemental modules for macros

DougGregor:
Macros are associated with a module, which is the module in which they would be defined if the macro could be defined within the language itself. We call this the owning module. It will need to be imported in any Swift source file that wants to use the macro. Introduce an "owningModule" requirement to the Macro protocol to describe this.

The signatures of macros can make use of types in the owning module, but might also require other modules (e.g., that the owning module depends on). Introduce a "supplementalSignatureModules" requirement to the Macro protocol, with a default implementation of `[]` since this is unlikely to be used often.

Both the owning and supplemental modules will be imported when processing the macro signature.


DougGregor:
@swift-ci please test

PR #1076, opened 2022-11-14
Add `*Syntax.parse` methods for the nodes that we support parsing

bnbarham:
Adds a new `SyntaxParseable` protocol with a single `parse(from:)` method. Any nodes with `parser_function` defined implement this protocol.

Other updates:
  - `SyntaxExpressibleByStringInterpolation` nodes to use these new methods
  - Parser tests to check against `*Syntax` rather than `Raw*Syntax` nodes
  - Builder methods to to take nodes (which are expressible by string literals) rather than `String`

### DIFF
--- a/Sources/SwiftParser/Parser+Entry.swift.gyb
+++ b/Sources/SwiftParser/Parser+Entry.swift.gyb
@@ -1,0 +1,76 @@
+%{
+  from gyb_syntax_support import *
+  # -*- mode: Swift -*-
+  # Ignore the following admonition it applies to the resulting .swift file only
+}%
+//// Automatically Generated From Entry.swift.gyb.
+//// Do Not Edit Directly!
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2022 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+@_spi(RawSyntax) import SwiftSyntax
+
+extension Parser {
+  /// Parse the source code in the given string as Swift source file. See
+  /// `Parser.init` for more details.
+  public static func parse(
+    source: String,
+    parseTransition: IncrementalParseTransition? = nil
+  ) -> SourceFileSyntax {
+    var parser = Parser(source)
+    return SourceFileSyntax.parse(from: &parser)
+  }
+
+  /// Parse the source code in the given string as Swift source file. See
+  /// `Parser.init` for more details.
+  public static func parse(
+    source: UnsafeBufferPointer<UInt8>,
+    maximumNestingLevel: Int? = nil,
+    parseTransition: IncrementalParseTransition? = nil
+  ) -> SourceFileSyntax {
+    var parser = Parser(source, maximumNestingLevel: maximumNestingLevel)
+    return SourceFileSyntax.parse(from: &parser)
+  }
+}
+
+public protocol SyntaxParseable: SyntaxProtocol {
+  static func parse(from parser: inout Parser) -> Self
+}
+
+% for node in SYNTAX_NODES:
+%   if node.parser_function:
+extension ${node.name}: SyntaxParseable {
+  public static func parse(from parser: inout Parser) -> Self {
+    let node = parser.${node.parser_function}()
+    return parser.parseRemainder(into: node).syntax
+  }
+}
+
+%   end
+% end
+fileprivate extension Parser {
+  mutating func parseRemainder<R: RawSyntaxNodeProtocol>(into: R) -> R {
+    guard !into.raw.kind.isSyntaxCollection, let layout = into.raw.layoutView else {
+      assertionFailure("Only support parsing of non-collection layout nodes")
+      return into
+    }
+
+    let remainingTokens = self.consumeRemainingTokens()
+    if remainingTokens.isEmpty {
+      return into
+    }
+
+    let unexpected = RawUnexpectedNodesSyntax(elements: remainingTokens, arena: self.arena)
+    let withUnexpected = layout.replacingChild(at: layout.children.count - 1, with: unexpected.raw, arena: self.arena)
+    return R.init(withUnexpected)!
+  }
+}
--- a/Sources/SwiftParser/gyb_generated/Parser+Entry.swift
+++ b/Sources/SwiftParser/gyb_generated/Parser+Entry.swift
@@ -1,0 +1,130 @@
+//// Automatically Generated From Entry.swift.gyb.
+//// Do Not Edit Directly!
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2022 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+@_spi(RawSyntax) import SwiftSyntax
+
+extension Parser {
+  /// Parse the source code in the given string as Swift source file. See
+  /// `Parser.init` for more details.
+  public static func parse(
+    source: String,
+    parseTransition: IncrementalParseTransition? = nil
+  ) -> SourceFileSyntax {
+    var parser = Parser(source)
+    return SourceFileSyntax.parse(from: &parser)
+  }
+
+  /// Parse the source code in the given string as Swift source file. See
+  /// `Parser.init` for more details.
+  public static func parse(
+    source: UnsafeBufferPointer<UInt8>,
+    maximumNestingLevel: Int? = nil,
+    parseTransition: IncrementalParseTransition? = nil
+  ) -> SourceFileSyntax {
+    var parser = Parser(source, maximumNestingLevel: maximumNestingLevel)
+    return SourceFileSyntax.parse(from: &parser)
+  }
+}
+
+public protocol SyntaxParseable: SyntaxProtocol {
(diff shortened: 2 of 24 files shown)


bnbarham:
@swift-ci please test

Sources/SwiftParser/gyb_generated/Parser+Entry.swift
```diff
@@ -0,0 +1,128 @@
...
+//===----------------------------------------------------------------------===//
+
+@_spi(RawSyntax) import SwiftSyntax
+
+extension Parser {
+  /// Parse the source code in the given string as Swift source file. See
+  /// `Parser.init` for more details.
+  public static func parse(
+    source: String,
+    maximumNestingLevel: Int? = nil
```
bnbarham, line 22: Whoops, I did not mean to remove parseTransition here

bnbarham:
@swift-ci please test

bnbarham:
@swift-ci please test

Sources/SwiftParser/Parser+Entry.swift.gyb
```diff
@@ -0,0 +1,75 @@
...
+public protocol SyntaxParseable: SyntaxProtocol {
+  static func parse(from parser: inout Parser) -> Self
+}
+
+% for node in SYNTAX_NODES:
+%   if node.parser_function:
+extension ${node.name}: SyntaxParseable {
+  public static func parse(from parser: inout Parser) -> ${node.name} {
+    let node = parser.${node.parser_function}()
+    return parser.parseRemainder(into: node).syntax
```
rintaro, line 54: Are we 100% sure that `parser` wouldn't be deinitialized before calling `syntax`?

Sources/SwiftSyntaxParser/SyntaxParser.swift
```diff
@@ -122,7 +122,7 @@ public enum SyntaxParser {
     enableBareSlashRegexLiteral: Bool?,
     diagnosticHandler: ((Diagnostic) -> Void)?
   ) throws -> SourceFileSyntax {
-    return Parser.parse(source: source, parseTransition: parseTransition)
+    return Parser.parse(source: source)
```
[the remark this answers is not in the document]
bnbarham: Sorry, left over from a previous change. Thanks.

Sources/SwiftParser/Parser+Entry.swift.gyb
```diff
@@ -0,0 +1,74 @@
...
+}
+
+public protocol SyntaxParseable: SyntaxProtocol {
+  static func parse(from parser: inout Parser) -> Self
+}
+
+% for node in SYNTAX_NODES:
+%   if node.parser_function:
+extension ${node.name}: SyntaxParseable {
+  public static func parse(from parser: inout Parser) -> ${node.name} {
```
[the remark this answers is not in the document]
bnbarham: Sure

Sources/SwiftSyntax/Raw/RawSyntaxLayoutView.swift
```diff
@@ -166,11 +166,5 @@ public struct RawSyntaxLayoutView {
   public var children: RawSyntaxBuffer {
     layoutData.layout
   }
-
-  /// The number of children, `present` or `missing`, in this node.
-  @_spi(RawSyntax)
-  public var numberOfChildren: Int {
-    return children.count
-  }
```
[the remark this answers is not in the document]
bnbarham: Yeah, I can put in another PR if you want. It's related because I used `children.count` and noticed this var when I was completing `children` 😅

Sources/SwiftParser/Parser+Entry.swift.gyb
bnbarham: It would seem super broken to me if it could since it's passed in as an `inout`. I suppose theoretically we could know that it's not used after this call + the last use is on the `parseRemainder`, but... 🤷 I'm happy to add a `withExtendedLifetime` if you think it needs it.

bnbarham:
@swift-ci please test

Sources/SwiftParser/Parser+Entry.swift.gyb
rintaro: My concern was when this function is inlined within the module, the optimizer might decide to release it early. But as far as I tried, it doesn't seem to be the case. So let's keep this as is.

rintaro (approved):

Sources/SwiftParser/Parser+Entry.swift.gyb
```diff
@@ -0,0 +1,75 @@
...
+      return into
+    }
+
+    let remainingTokens = self.consumeRemainingTokens()
+    if remainingTokens.isEmpty {
+      return into
+    }
+
+    let unexpected = RawUnexpectedNodesSyntax(elements: remainingTokens, arena: self.arena)
+    let withUnexpected = layout.replacingChild(at: layout.children.count - 1, with: unexpected.raw, arena: self.arena)
```
ahoppen, line 72: This breaks if the node is a syntax collection because syntax collections are not interleaved with unexpected nodes. Could you use the syntax node structure (I don't recall exactly what I ended up calling it, it was one of my latest PRs) to assert that the node we are parsing is a layout node and has unexpected nodes as the last child?

Sources/SwiftSyntaxBuilder/SyntaxNodeWithBody.swift
```diff
@@ -14,13 +14,18 @@ import SwiftSyntax
 
 // MARK: - HasCodeBlock
 
+// TODO: Consider removing `HasTrailingCodeBlock` and friends
+// It'd be much nicer if we didn't have any builder APIs taking strings. We
+// should just make convenvience inits where needed such that we could write:
+//   `IfStmt("if foo")` -> `IfStmt(conditions: [.expression("foo")])`
```
ahoppen, line 20: This is what we had before I introduced `HasTrailingCodeBlock` and it was super verbose and I didn't like it. If you want, you can take a look at old versions of how classes or Strucks were being constructed in CodeGeneration.

Tests/SwiftParserTest/Assertions.swift
```diff
@@ -316,32 +309,68 @@ func AssertParse(
...
+///     asserts the parse was successful (ie. it has no diagnostics). Note
+///     that `DiagnosticsSpec` uses the location marked by `1️⃣` by default.
+///   - applyFixIts: Applies only the fix-its with these messages.
+///   - fixedSource: Asserts that the source after applying fix-its matches
+///     this string.
+func AssertParse<S: SyntaxProtocol>(
   _ markedSource: String,
-  _ parseSyntax: (inout Parser) -> Node,
+  _ parse: (String) -> S,
   substructure expectedSubstructure: Syntax? = nil,
```
ahoppen, line 374: Seeing this, I'm wondering whether we can get rid of the custom entry point here altogether. It didn't really turn out to be super useful IMO.

bnbarham: Which one? I was going to change these methods to a struct + `check` or similar after this PR.

Sources/SwiftSyntaxBuilder/SyntaxNodeWithBody.swift
bnbarham: I'll take a look, but if it's too verbose we should fix that. `IfStmt("if foo")` seems really weird to me, we've said it's an `if` so why are we writing `if` again 😅? In general I'd just really like to steer away from methods that take a `String` and parse it, without it being obvious that's what they're doing.

Sources/SwiftParser/Parser+Entry.swift.gyb
bnbarham: Sure, if I can find it :). FWIW these are only generated for layout nodes, I think I used to actually specifically check that in the gyb but then removed it since no non-layout nodes have parse functions anyway.

Sources/SwiftSyntaxBuilder/SyntaxNodeWithBody.swift
bnbarham: I assume you're referring to https://github.com/apple/swift-syntax/commit/2c04f3416a098b55d76080f3b977e0122fdb49f7? To take some examples from there -

Old:
```
SwitchCase(label: SwitchCaseLabel(caseItems: [CaseItem(pattern: ExpressionPattern(expression: MemberAccessExpr(name: "contextualKeyword")))]))
```
New:
```
SwitchCase("case .contextualKeyword:")
```
Suggested:
```
SwitchCase(pattern: ".contextualKeyword") { }

Old:
```
ExtensionDecl(
       leadingTrivia: node.documentation.isEmpty
         ? []
         : .docLineComment("/// \(node.documentation)") + .newline,
       extendedType: Type(node.type.shorthandName),
       inheritanceClause: TypeInheritanceClause { InheritedType(typeName: Type("ExpressibleByArrayLiteral")) }
     )
```
New:
```
ExtensionDecl("\(docComment)extension \(node.type.shorthandName): ExpressibleByArrayLiteral")
```
Suggested:
```
ExtensionDecl(
  leadingTrivia: docComment,
  extendedType: "\(node.type.shorthandName)",
  inheritances: ["ExpressibleByArrayLiteral"])
```

bnbarham:
@swift-ci please test

bnbarham:
@swift-ci please test

bnbarham:
@swift-ci please test

Sources/SwiftSyntaxBuilder/SyntaxNodeWithBody.swift
ahoppen: I think the function declarations were one of the worst in the old API but maybe you're right that we can switch if every child node of these is passable from a string literal. Your examples definitely look good to me.

Tests/SwiftParserTest/Assertions.swift
ahoppen: I was thinking that `parse` here could just always be `SourceFileSyntax.parse`. I'm still not 100% sure about it though.

bnbarham: Oh I see. We currently expose a few parse methods (eg. decl/statement/expr/pattern/memberdecl), so IMO it probably makes sense to keep these still.

bnbarham:
@swift-ci please test

bnbarham:
@swift-ci please test

PR #1077, opened 2022-11-15
Fix member type/expression parsing

bnbarham:
Few fixes to member parsing:
  - Parse `.Type` and `.Protocol` as `MetatypeTypeSyntax`
  - Parse keywords as identifiers
  - Generate a missing identifier rather than unexpected nodes if the next
    token is on a new line
  - Diagnose extraneous whitespace after `.` if there was no leading
    whitespace
  - Remove `.prefixPeriod` since it is no longer used


bnbarham:
@swift-ci please test

bnbarham:
Note that this depends on https://github.com/apple/swift-syntax/pull/1076 (which is included in this PR right now) and the FIXME in https://github.com/apple/swift/pull/62092should be able to be removed after this (looks like there's already a `MetatypeTypeSyntax` visit).

bnbarham:
@swift-ci please test

bnbarham:
@swift-ci please test

bnbarham:
@swift-ci please test

bnbarham:
@swift-ci please test

bnbarham:
@swift-ci please test

Sources/SwiftParser/Types.swift
```diff
@@ -194,15 +194,21 @@ extension Parser {
...
-    // '.Type', '.Protocol', '?', '!', and '[]' still leave us with type-simple.
     var loopCondition = LoopProgressCondition()
     while loopCondition.evaluate(currentToken) {
-      if !stopAtFirstPeriod, let (period, type) = self.consume(
-        if: { [.period, .prefixPeriod].contains($0.tokenKind) },
-        followedBy: { $0.isContextualKeyword(["Type", "Protocol"])}
-      ) {
-        base = RawTypeSyntax(RawMetatypeTypeSyntax(
-          baseType: base, period: period, typeOrProtocol: type, arena: self.arena))
+      if !stopAtFirstPeriod, self.at(.period) || self.at(.prefixPeriod) {
```
rintaro, line 199: This allows `A.Type.B` (which is currently not parsed). And the structure would be like `(memberType (metatypeType (typeIdentifier)))`. I'm totally OK with it, but we should at least have a test case?

Sources/SwiftParser/Types.swift
```diff
@@ -255,45 +263,22 @@ extension Parser {
...
-      }
-
-      if stopAtFirstPeriod {
-        keepGoing = nil
-      } else {
-        keepGoing = self.consume(if: .period) ?? self.consume(if: .prefixPeriod)
-      }
-    } while keepGoing != nil && loopProgress.evaluate(currentToken)
+  mutating func parseTypeIdentifier() -> RawTypeSyntax {
+    if self.at(.anyKeyword) {
```
ahoppen, line 275: Do we also need to handle `some` here?

Sources/SwiftParser/Types.swift
```diff
@@ -194,15 +194,21 @@ extension Parser {
...
-      if !stopAtFirstPeriod, let (period, type) = self.consume(
-        if: { [.period, .prefixPeriod].contains($0.tokenKind) },
-        followedBy: { $0.isContextualKeyword(["Type", "Protocol"])}
-      ) {
-        base = RawTypeSyntax(RawMetatypeTypeSyntax(
-          baseType: base, period: period, typeOrProtocol: type, arena: self.arena))
+      if !stopAtFirstPeriod, self.at(.period) || self.at(.prefixPeriod) {
+        let period = self.consumeAnyToken()
+        if self.atContextualKeyword("Type") || self.atContextualKeyword("Protocol") {
+          let typeOrProtocol = self.consumeAnyToken(remapping: .contextualKeyword)
```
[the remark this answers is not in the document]
bnbarham: Yeah, @rintaro thought it'd be better to avoid the extra ARC traffic from the temporary array creation.

Sources/SwiftParser/Types.swift
bnbarham: I don't think so, but this seems a little odd to me actually - this `Any` not `any` in which case is `Any` really a keyword in this context (ends up being put into a simple type identifier as `anyKeyword`)? Also, it seems like it would be better to still parse generics in this context for recovery purposes - thoughts?

Sources/SwiftParser/Types.swift
```diff
@@ -194,15 +194,21 @@ extension Parser {
...
+          base = RawTypeSyntax(RawMetatypeTypeSyntax(
+            baseType: base, period: period, typeOrProtocol: typeOrProtocol,
+            arena: self.arena))
+        } else {
+          let (name, generics) = self.parseTypeNameWithGenerics()
```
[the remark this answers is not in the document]
bnbarham: Sure 👍 Nice catch.

Sources/SwiftParser/Types.swift
```diff
@@ -194,15 +194,21 @@ extension Parser {
...
+          base =  RawTypeSyntax(RawMemberTypeIdentifierSyntax(
+            baseType: base, period: period, name: name,
```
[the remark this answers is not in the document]
bnbarham: I'll have a look

Sources/SwiftParser/Types.swift
bnbarham: Sure, I was pretty lazy with non-metatype test additions (ie. I didn't add any). Do we want a diagnostic for the `A.Type.B` case, ie. any member type with a metatype suggesting to add backpacks?

Sources/SwiftParser/Types.swift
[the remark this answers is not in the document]
bnbarham: This wasn't working because we actually parse `.` as an unknown token. I'll fix that up and add a couple diagnostics to cover the `Foo. member` and `Foo.Type.Bar` cases since the grammar doesn't allow either of those.

Sources/SwiftParser/Types.swift
ahoppen: Oh, maybe this should be called `captialAnyKeyword`. But nothing for this PR.

Makes sense to me. I won’t stop you from implementing better recovery.

Sources/SwiftParser/Types.swift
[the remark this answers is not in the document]
bnbarham: Yeah, probably wouldn't hurt. I'd do that here but the PR is now... much larger than I intended it to be :P

bnbarham:
@swift-ci please test

bnbarham:
So I went down a rabbit hole here 😅 ... if you could have another look that'd be great @rintaro / @ahoppen 🙇

Sources/SwiftParser/Parser.swift
```diff
@@ -585,3 +585,24 @@ extension SyntaxText {
     try body(UnsafeBufferPointer<UInt8>(start: self.baseAddress, count: self.count))
   }
 }
+
+extension Parser {
+  /// Consume a period as part of a member expression/type. Create unexpected +
+  /// missing tokens if there's trailing whitespace no leading whitespace
```
ahoppen, line 591: there's trailing whitespace **and** no leading whitespace.

---

Do we also need to diagnose the case that is leading whitespace but no trailing whitespace?

Sources/SwiftParser/Parser.swift
```diff
@@ -585,3 +585,24 @@ extension SyntaxText {
...
+  /// (whether from `leading` or from the leading whitspace of the current
+  /// token).
+  mutating func consumeMemberPeriod<R: RawSyntaxNodeProtocol>(leading: R?) -> (unexpected: RawUnexpectedNodesSyntax?, period: RawTokenSyntax) {
```
ahoppen, line 594: Would this better be named `previousNode`. Leading sounds too much like the variable contains trivia to me.

Sources/SwiftParser/Expressions.swift
```diff
@@ -512,13 +512,19 @@ extension Parser {
...
+    name: RawTokenSyntax,
+    declNameArgs: RawDeclNameArgumentsSyntax?,
+    generics: RawGenericArgumentClauseSyntax?
   ) {
-    assert(self.at(any: [.period, .prefixPeriod]))
-    let period = self.consumeAnyToken(remapping: .period)
+    assert(self.at(.period))
+    let (unexpectedPeriod, period) = self.consumeMemberPeriod(leading: leading)
+    if currentToken.isAtStartOfLine {
+      let missingIdentifier = RawTokenSyntax(missing: .identifier, arena: self.arena)
```
ahoppen, line 525: If you want to, you can shorten this to something like

```suggestion
      let missingIdentifier = missingToken(.identifier)
```

Sources/SwiftParserDiagnostics/MissingNodesError.swift
```diff
@@ -290,6 +290,51 @@ public struct InsertTokenFixIt: ParserFixIt {
...
+        fixIts = []
+      }
+      addDiagnostic(
+        invalidToken,
+        InvalidIdentifierError(invalidIdentifier: invalidToken, missingIdentifier: node),
+        fixIts: fixIts,
+        handledNodes: [unexpectedTokens.id]
+      )
+    } else if node.tokenKind == .period && invalidToken.tokenKind == .period {
+      let fixIt =  FixIt(message: .removeExtraneousWhitespace, changes: [
```
ahoppen, line 325: Do we need to check first whether the invalid token actually had more trivia than the missing one?

Tests/SwiftParserTest/MemberRefTests.swift
```diff
@@ -0,0 +1,226 @@
...
+      "MyType.1️⃣",
+      { TypeSyntax.parse(from: &$0) },
+      substructure: Syntax(MemberTypeIdentifierSyntax(
+        baseType: SimpleTypeIdentifierSyntax(
+          name: .identifier("MyType")
+        ),
+        period: .periodToken(),
+        name: .identifier("", presence: .missing)
+      )),
+      diagnostics: [DiagnosticSpec(message: "expected member type in member type")])
```
ahoppen, line 42: Uh, that diagnostic message doesn't seem ideal. Do you know if we can improve it by updating some `name_for_diagnostic`s? Doesn't have to be this PR.

Tests/SwiftParserTest/MemberRefTests.swift
```diff
@@ -0,0 +1,226 @@
...
+        substructure: expected)
+    }
+  }
+
+  func testInvalidWhitespace() {
+    let expected = "MyType.Nested"
+
+    let cases = [
+      "MyType1️⃣. Nested",
+      "MyType1️⃣./*bar*/Nested",
```
ahoppen, line 72: What about

```swift
"MyType .Nested",
"MyType/*bar*/.Nested"
```

Sources/SwiftParser/Expressions.swift
```diff
@@ -512,13 +512,19 @@ extension Parser {
...
+    period: RawTokenSyntax,
+    name: RawTokenSyntax,
+    declNameArgs: RawDeclNameArgumentsSyntax?,
+    generics: RawGenericArgumentClauseSyntax?
   ) {
-    assert(self.at(any: [.period, .prefixPeriod]))
-    let period = self.consumeAnyToken(remapping: .period)
+    assert(self.at(.period))
+    let (unexpectedPeriod, period) = self.consumeMemberPeriod(leading: leading)
+    if currentToken.isAtStartOfLine {
```
rintaro, line 524: As of Swift 5.7,
```
foo  .
bar
```
i.e `"foo .\nbar"` is a valid member access syntax. I think this changes the behavior.

Sources/SwiftParser/Parser.swift
bnbarham: Leading and no trailing is fine, happens all the time with newlines:
```
foo
  .bar
  ```

File: Sources/SwiftParser/Expressions.swift (hunk 8 above)
Comment (bnbarham) [reply Sources/SwiftParser/Expressions.swift:525]: Ah, sure, makes sense I guess. Though `RawTokenSyntax(missing: .identifier, arena: self.arena)` isn't that bad to me 😅

File: Tests/SwiftParserTest/MemberRefTests.swift (hunk 11 above)
Comment (bnbarham) [reply Tests/SwiftParserTest/MemberRefTests.swift:72]: They're both valid, but I can check them above 👍

File: Sources/SwiftParser/Expressions.swift (hunk 12 above)
Comment (bnbarham) [reply Sources/SwiftParser/Expressions.swift:524]: Ah, whoops. Thanks. Guess I'll check for the previous node's trivia here as well :(

File: Sources/SwiftParserDiagnostics/MissingNodesError.swift (hunk 9 above)
Comment (bnbarham) [reply Sources/SwiftParserDiagnostics/MissingNodesError.swift:325]: I don't think so, this pattern should never happen except in the case of the extraneous whitespace right? Unless you made your own tree I guess but... do we care about that case?

File: Sources/SwiftParser/Parser.swift (hunk 7 above)
Comment (bnbarham) [reply Sources/SwiftParser/Parser.swift:594]: Sure, I can do that.

**Comment by bnbarham** [comment]
@swift-ci please test

**Comment by bnbarham** [comment]
https://github.com/apple/swift/pull/62561

@swift-ci please test

**Review by rintaro — APPROVED** [review]

File: Sources/SwiftParser/Parser.swift [hunk 13]
```diff
@@ -585,3 +585,34 @@ extension SyntaxText {
...
+    if beforePeriodWhitespace || !afterPeriodWhitespace  {
+      // Valid, parse as-is
+      return (nil, period, false)
+    }
+
+    // Invalid, extraneous whitespace. Have callers synthesize a missing
+    // member if there's a newline after the period.
+    return (RawUnexpectedNodesSyntax(elements: [period.raw], arena: arena),
+            RawTokenSyntax(missing: .period, arena: arena),
+            afterContainsAnyNewline)
```
ahoppen, line 616: Personal preference (ignore if you disagree) but I find conjunctions easier to read than disjunctions.

```suggestion
    if !beforePeriodWhitespace && afterPeriodWhitespace  {
      // Invalid, extraneous whitespace. Have callers synthesize a missing
      // member if there's a newline after the period.
      return (RawUnexpectedNodesSyntax(elements: [period.raw], arena: arena),
              RawTokenSyntax(missing: .period, arena: arena),
              afterContainsAnyNewline)
    } else {
      // Valid, parse as-is
      return (nil, period, false)
    }

```

Sources/SwiftParserDiagnostics/MissingNodesError.swift
```diff
@@ -331,15 +392,20 @@ extension ParseDiagnosticsGenerator {
     }
 
     let changes = missingNodes.enumerated().map { (index, missingNode) -> FixIt.Changes in
-      if index == 0, let token = missingNode.as(TokenSyntax.self), token.tokenKind.isPunctuation == true, token.previousToken(viewMode: .sourceAccurate)?.tokenKind.isPunctuation == false {
+      let previousTokenKind = missingNode.previousToken(viewMode: .sourceAccurate)?.tokenKind
+      if index == 0,
+         let token = missingNode.as(TokenSyntax.self),
+         (token.tokenKind.isPunctuation && previousTokenKind?.isPunctuation == false
+          || !token.tokenKind.isPunctuation && previousTokenKind?.isPunctuation == true) {
```
ahoppen, line 399: Style idea (again not sure if you agree)

```suggestion
         let previousTokenIsPunctionation = previousTokenKind?.isPunctuation,
         token.tokenKind.isPunctuation != previousTokenIsPunctuation) {
```

You could even inline `previousTokenKind` into the `if`

Sources/SwiftParserDiagnostics/MissingNodesError.swift
```diff
@@ -290,7 +290,68 @@ public struct InsertTokenFixIt: ParserFixIt {
...
+        .makeMissing(invalidToken, transferTrivia: false),
+        .makePresent(node)
+      ]
+
+      if let identifier = node.nextToken(viewMode: .all),
+         identifier.rawTokenKind == .identifier,
+         identifier.presence == .missing {
+        // The extraneous whitespace caused a missing identifier, output a
+        // diagnostic for it instead. Insert non-whitespace trivia after the
+        // missing identifier.
```
ahoppen, line 336: It took me a while to understand what “instead” was referring to. 

```suggestion
        // The extraneous whitespace caused a missing identifier, output a
        // diagnostic inserting it instead of a diagnostic to fix the trivia around the period. Insert non-whitespace trivia after the
        // missing identifier.
```

Tests/SwiftParserTest/MemberRefTests.swift
```diff
@@ -0,0 +1,233 @@
...
+      "MyType/*foo*/./*bar*/Nested",
+      "MyType\n.  Nested",
+      "MyType\n.\nNested"
+    ]
+    for source in cases {
+      AssertParse(
+        source,
+        { TypeSyntax.parse(from: &$0) },
+        substructure: expected)
+    }
```
ahoppen, line 68: What do you think about adding `#line` as a key to `cases` like in the following? That way Xcode will point you to the test case that failed instead of the `AssertParse` call that is being repeated.

https://github.com/apple/swift-syntax/blob/fd2f742c7bdac60f27685ebdacce146706f1d8ef/Tests/SwiftSyntaxBuilderTest/TriviaTests.swift#L49-L64

Sources/SwiftParser/Parser.swift
bnbarham: Yeah, I think this is one area where we very much have different styles. I much prefer less nesting + early returns. I can swap to using the conjunction if you really prefer but personally prefer as is (ie. larger part not being nested).

Tests/SwiftParserTest/MemberRefTests.swift
bnbarham: Sure

Sources/SwiftParserDiagnostics/MissingNodesError.swift
bnbarham: Uh... yeah, I can just pull `previousTokenKind` in. Wonder how late it was when I wrote this 😆. Thanks, good suggestion!

bnbarham:
@swift-ci please test

Sources/SwiftParser/Parser.swift
ahoppen: I’m fine to leave it as-is.

bnbarham:
@swift-ci please test

Sources/SwiftParserDiagnostics/MissingNodesError.swift
bnbarham: @ahoppen I ended up having to change this. This broke eg. required keywords following a `{`. I now specifically check if the current token is a punctuation and the next token is an identifier. That would also break if we ever had a required identifier after `{` but I don't think that ever happens? Open to other ideas here, this seems pretty hacky right now.

bnbarham:
@swift-ci please test

bnbarham:
https://github.com/apple/swift-format/pull/463

@swift-ci please test

Sources/SwiftParser/Types.swift
AnthonyLatsis: @rintaro We are working toward modeling and diagnosing stuff with `A.Type.B` in Sema (apple/swift#62775, apple/swift#62319). I am going to add tests for these to the Swift parser in #1171.

Sources/SwiftParser/Types.swift
```diff
@@ -844,42 +857,9 @@ extension Parser.Lookahead {
...
-    } while self.consume(if: .comma) != nil && loopProgress.evaluate(currentToken)
-
-    // Check for the terminating '>'.
-    guard self.currentToken.starts(with: ">") else {
-      return false
-    }
-    self.consumePrefix(">", as: .rightAngle)
-
-    return true
-  }
```
AnthonyLatsis, line 878: @bnbarham Was this removed on purpose?

rintaro: @AnthonyLatsis Yes, in `SwiftParser` we decided not to support old style protocol composition (there's no `parseOldStyleProtocolComposition()`.) So keeping `canParseOldStyleProtocolComposition()` doesn't make sense.

Sources/SwiftParserDiagnostics/MissingNodesError.swift
ahoppen: Sounds reasonable to me.

Sources/SwiftParser/Types.swift
bnbarham: Thanks @AnthonyLatsis 🙇. Sorry, I didn't realize you were making changes for this in the Swift parser as well. Sounds like we'll be able to have ASTGen just create the naive AST and then have Sema diagnose.

PR #1081, opened 2022-11-16
Temporarily workaround CMake dependencies bug

bnbarham:
Workaround a CMake bug in order to make sure that libraries are rebuilt when their dependencies change.


bnbarham:
https://github.com/apple/swift/pull/62129

@swift-ci please test

bnbarham:
https://github.com/apple/swift/pull/62129

@swift-ci please test

Sources/SwiftParser/CMakeLists.txt
```diff
@@ -47,6 +48,15 @@ set_target_properties(SwiftParser PROPERTIES
   INTERFACE_INCLUDE_DIRECTORIES
   "${CMAKE_Swift_MODULE_DIRECTORY}")
 
+#file(TOUCH ${CMAKE_CURRENT_BINARY_DIR}/SwiftParser-complete.swift)
```
etcwilde, line 51: Remove this?

Sources/CMakeLists.txt
```diff
@@ -6,6 +6,32 @@
...
+# cmake generation for Swift adds an order only dependency, which matches how C-family languages
+# works. In that case, however, ninja (and presumably other generators) will rebuild on header
+# changes. That's not the case for Swift, and thus if A -> B, A is not being rebuilt when the
+# ABI/API of B changes.
+#
+# For now workaround this by touching a file whenever B is rebuilt and then compiling that file as
+# part of A. Ideally this file would be generated by each of the targets, but that dependency didn't
+# seem to be being tracked.
+function(add_forced_dependencies TARGET)
+  cmake_parse_arguments(ARGS "" "" "PUBLIC" ${ARGN})
```
etcwilde, line 18: The naming here feels a bit weird. I'm not getting that it's linking from the function name, and the `PUBLIC` seems to come out of nowhere.

Sources/CMakeLists.txt
```diff
@@ -6,6 +6,32 @@
...
+
+  foreach(DEPENDENCY ${ARGS_PUBLIC})
+    target_link_libraries(${TARGET} PUBLIC
```
etcwilde, line 21: You shouldn't need to add these separately. `target_link_libraries(${TARGET} PUBLIC ${ARGS_PUBLIC})` should work.

Sources/CMakeLists.txt
```diff
@@ -6,6 +6,32 @@
...
+      ${DEPENDENCY}
+    )
+
+    add_custom_command(OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/forced-${DEPENDENCY}-dep.swift
```
etcwilde, line 25: Then this part can be done in the for-loop.

etcwilde (changes requested):
It's terrible, but I think it works. A few style comments.
Also, can you add reference to rdar://102202478 in a comment? Thanks.

CC @compnerd any thoughts/feedback?

Sources/SwiftParser/CMakeLists.txt
bnbarham: Thanks

Sources/CMakeLists.txt
bnbarham: So I totally agree about the `PUBLIC`, but I was basically just aiming at making it a 1:1 replacement for `target_link_libraries` (at least for the case we use it). I don't *really* mind either way. Do you have a strong opinion here?

How about I rename it to `force_target_link_libraries`?

Sources/CMakeLists.txt
bnbarham: Ah yeah, sure. For some background this used to be `add_forced_dependency` singular and I changed it to take multiple right before pushing :P. Hence all the force pushes.

Sources/CMakeLists.txt
bnbarham: All this is already in the for-loop, or you just mean keep it there? Sure.

bnbarham:
https://github.com/apple/swift/pull/62129

@swift-ci please test

etcwilde (approved):
Yeah, I think the new changes are fine given that the goal is to delete the function.

Sources/CMakeLists.txt
compnerd: I actually think that this change is far too invasive for what it intends to do.  I think that we should be more clever about this workaround.  Please rename this function `target_link_librares`.  We can subsequently, at the end of the function passthrough to the underlying implementation of `target_link_libraries`.

Something like this should do the trick:

```
function(target_link_libraries target)
  cmake_parse_arguments(TLL_ARGS "" "" "PUBLIC" ${ARGN})

  foreach(dependency ${TLL_PUBLIC})
    ...
  endforeach()

  _target_link_libraries(${target} PUBLIC ${TLL_PUBLIC} ${TLL_ARGN})
endfunction()
```

This way, in the future, the revert of this will just drop this one function and everything goes back to normal.

compnerd (changes requested):

Sources/CMakeLists.txt
bnbarham: I had no idea you could do that. I've updated the PR to do this instead, but I'll leave the Swift PR as is - once the function is redefined, it's redefined from that point on (not scoped to a subdirectory). This is fine for swift-syntax, but not for the swift-side change (and that only touches the one file anyway). Thanks for the suggestion @compnerd!

bnbarham:
@swift-ci please test

PR #1084, opened 2022-11-17
Take Ownership of the Incremental Parse Utility

CodaFi:
Technically we aren't using this, but we haven't decided what to do about the old incremental parsing tests. For now, move these tools into its last consumer - swift-syntax - to unblock the build.


CodaFi:
@swift-ci test

PR #1087, opened 2022-11-17
Add generic argument clause to macro expansion exprs/decls

DougGregor:


DougGregor:
@swift-ci please test

PR #1089, opened 2022-11-18
Drop Node, Token, and Trivia Serialization Codes from the Gyb

CodaFi:
The re-coring (apple/swift-syntax#1041) removed the last client, and deleting libSyntax (apple/swift#62145) removed the service, we have no need to retain these serialization codes. Drop them like a stone.


CodaFi:
@swift-ci test

CodaFi:
apple/swift#62284

@swift-ci test

PR #1090, opened 2022-11-19
Rework string interpolation for SwiftSyntaxBuilder literals

beccadax:
The current design for interpolating into syntax literals makes it far too easy to accidentally stick random strings in the middle of your Swift code. It also encourages half-baked escaping of data. Let's not do that—instead, we'll:

* Require interpolations to have the `raw:` label if they are just stringified and inserted directly into the source code.

```swift
let number = Int.random(in: 1...10)
let varRef1: ExprSyntax = "myVar\(number)"         // deprecated
let varRef2: ExprSyntax = "myVar\(raw: number)"    // use this instead
```

* Allow interpolations to use the `literal:` label to convert a value to a Swift literal and insert it into the source code, with any escaping or other edge cases handled automatically.

```swift
let printCall: ExprSyntax = "print(\(literal: ["Hello", "\\", "World" ]))"
// `print(["Hello", #"\"#, "World"])` or some such
```

Based on a [discussion about macros generating literals](https://forums.swift.org/t/pitch-expression-macros/61499/16) in the expression macro pitch.


Sources/SwiftSyntaxBuilder/Syntax+StringInterpolation.swift
```diff
@@ -108,19 +108,35 @@ extension SyntaxStringInterpolation: StringInterpolationProtocol {
   }
 
   // Append a value of any CustomStringConvertible type as source text.
+  @available(*, deprecated, renamed: "appendInterpolation(raw:)", message: "use '\\(raw: <value>)' to interpolate a plain string directly into Swift code")
```
bnbarham, line 111: This is a new API anyway, so I'd say it's fine if you delete it. But also okay if you want to leave it in. It's possible that swift-format/lint have already updated to use it.

bnbarham (approved):
Very cool, thanks Becca!

Sources/SwiftSyntaxBuilder/Syntax+StringInterpolation.swift
```diff
@@ -152,6 +168,26 @@ enum SyntaxStringInterpolationError: Error, CustomStringConvertible {
...
+///      // `expr1` is a syntax tree for `print("Hello, world!")`
+///
+/// Note that quote marks are automatically added around the contents; you don't have to write them yourself. The conformance will automatically ensure the contents are correctly escaped, possibly by using raw literals or other language features:
+///
+///      let msPath = "c:\\windows\\system32"
+///      let expr2 = ExprSyntax("open(\(literal: msPath))")
+///      // `expr2` might be a syntax tree for `open(#"c:\windows\system32"#)`
+///      // or for `open("c:\\windows\\system32")`.
+public protocol ExpressibleByLiteralSyntax {
+  func makeLiteralSyntax() -> ExprSyntaxProtocol
```
DougGregor, line 188: Should we introduce an associated type here to express the result type more completely, e.g.,

```swift
  associatedtype LiteralSyntaxNode: ExprSyntaxProtocol
  func makeLiteralSyntax() -> LiteralSyntaxNode
```

so that (e.g.) `String`'s `makeLiteralSyntax` produces a `StringLiteralExprSyntax`?

Sources/SwiftSyntaxBuilder/Syntax+StringInterpolation.swift
```diff
@@ -192,3 +228,215 @@ extension SyntaxExpressibleByStringInterpolation {
...
+extension Substring: ExpressibleByLiteralSyntax {
+  public func makeLiteralSyntax() -> ExprSyntaxProtocol {
+    // TODO: Choose whether to use a single-line or multi-line literal.
+    let quote = TokenSyntax.stringQuote
+
+    // Select a raw delimiter long enough that we won't need to escape quotes or backslashes.
+    // Locate backslashes and quotes...
+    let problemIndices = allIndices(where: #"\""#.contains(_:))
+    // Count adjacent hashes and compute the largest number (-1 = no problem chars)...
+    let maxPoundCount = problemIndices.reduce(-1) { prevCount, i in
```
DougGregor, line 255: FWIW, we have an implementation like this over in `ConvenienceInitializers.swift`.

DougGregor (commented):
I love the direction of this. It makes it far less error-prone to produce Swift code via string interpolation, and does so in an elegant way.

Sources/SwiftSyntaxBuilder/Syntax+StringInterpolation.swift
beccadax: I have some concerns about how that implementation handles non-interpolation backslashes in raw literals, but I suppose I should actually fix the thing if it's wrong instead of reimplementing it.

Sources/SwiftSyntaxBuilder/Syntax+StringInterpolation.swift
beccadax: For floats, I ended up needing to return some things that weren't strictly literal expressions, like `-.infinity`. Would I just say that floats map to `ExprSyntax` or something?

Sources/SwiftSyntaxBuilder/Syntax+StringInterpolation.swift
bnbarham: See my comment above, it's not clear to me that we actually want/need that initializer any more.

Sources/SwiftSyntaxBuilder/Syntax+StringInterpolation.swift
```diff
@@ -192,3 +228,215 @@ extension SyntaxExpressibleByStringInterpolation {
...
+        return nil
+      }
+      i = index(after: newI)
+      return newI
+    }
+  }
+}
+
+extension Substring: ExpressibleByLiteralSyntax {
+  public func makeLiteralSyntax() -> ExprSyntaxProtocol {
```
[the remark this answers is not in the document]
beccadax: I think of `makeLiteralSyntax()` as purely an implementation detail of `ExpressibleByLiteralSyntax`, not something users would call directly. If we want to expose this functionality for use outside of interpolations (probably a good idea!), I'd put an initializer on `ExprSyntax` or something and tell people to call that.

Sources/SwiftSyntaxBuilder/Syntax+StringInterpolation.swift
```diff
@@ -192,3 +228,215 @@ extension SyntaxExpressibleByStringInterpolation {
...
+    // TODO: Choose whether to use a single-line or multi-line literal.
```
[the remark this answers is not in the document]
beccadax: If we did multiline, I would hope we did indent it.

Sources/SwiftSyntaxBuilder/Syntax+StringInterpolation.swift
beccadax: The deprecation exists both to correct existing code, and to tell people what to do if they're writing brand-new code and naïvely use a plain interpolation. Granted, it should probably mention `\(literal:)` now that that's an option.

Sources/SwiftSyntaxBuilder/Syntax+StringInterpolation.swift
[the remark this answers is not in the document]
beccadax: Also, looking at the `StringLiteralExpr` initializer in particular, it has some advanced capabilities that this doesn't; for instance, you can specify a raw delimiter, and if you do then escape sequences in the `content` are not escaped. I assume we want to hang on to those capabilities.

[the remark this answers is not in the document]
bnbarham: I just put this comment on `Substring.makeLiteralSyntax` because it's where I saw it doing the much of the implementation of the `StringLiteralExpr` convenience initializer. I too would consider  `makeLiteralSyntax` an implementation detail.

> for instance, you can specify a raw delimiter, and if you do then escape sequences in the content are not escaped

That's true, but then you could also just use the regular constructor if you *really* want that (providing a `StringLiteralSegments`), so I'm still not convinced it adds much over `ExpressibleByLiteralSyntax`. Its main use case was escaping the literal, IMO.

Anyway, I don't mind having both, but I would strongly prefer them to use the same escaping implementation (totally fine if you think yours handles more cases/is better).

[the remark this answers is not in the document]
beccadax: I'm working on bringing the existing implementation up to snuff so that the new methods can use that.

Sources/SwiftSyntaxBuilder/Syntax+StringInterpolation.swift
beccadax: I haven't done this because of both floats and the new `Optional` conformance (which sometimes generates `.some(nil)` and such), but we can discuss it if you think it's important.

Sources/SwiftSyntaxBuilder/Syntax+StringInterpolation.swift
beccadax: I've fixed up the convenience initializer and ripped out this separate implementation.

Sources/SwiftSyntaxBuilder/Syntax+StringInterpolation.swift
```diff
@@ -373,3 +385,47 @@ extension Dictionary: ExpressibleByLiteralSyntax where Key: ExpressibleByLiteral
...
+        // TODO: Can't we have something nicer than this? `MemberAccessExpr(name: "some").makeCall(wrapped)`?
+        return FunctionCallExpr(
+          calledExpression: MemberAccessExpr(name: "some"),
+          leftParen: .leftParen,
+          argumentList: TupleExprElementList {
+            TupleExprElement(expression: wrappedExpr)
+          },
+          rightParen: .rightParen
+        )
+      }
```
bnbarham, line 459: There's certainly more work we need to do here. I believe the current idea is to use result builders + more convenience initializers to make this a little nicer. Right now you could have:
```
FunctionCallExpr(
  calledExpression: MemberAccessExpr(name: "some"),
  leftParen: .leftParen,
  rightParen: .rightParen) {
  TupleExprElement(expression: wrappedExpr)
}
```

If we had default parens and didn't require a `TupleExprElement` when there's no label, this could be:
```
FunctionCallExpr(calledExpression: MemberAccessExpr(name: "some")) {
  wrappedExpr
}
```

And `calledExpression` isn't really necessary either, in which case we'd end up with:
```
FunctionCallExpr(MemberAccessExpr(name: "some")) { wrappedExpr }
```

I've also added a TODO in https://github.com/apple/swift-syntax/pull/1076/files#diff-d0bdd506f33bc1e41fd5d3e79fe44bcee342a0d21590d6aac116dfd479fa0f60R17 which lists some more we need.

FWIW I had been planning on adding defaults for parameters that take a single token this week sometime, but I haven't started that yet 😅 (how is it already Tuesday!).

beccadax:
@swift-ci please test

Sources/SwiftSyntaxBuilder/Syntax+StringInterpolation.swift
beccadax: Tangential, but: Have you considered adding something like this?

```swift
extension ExprSyntaxProtocol {
  /// A proxy used to construct a function, method, or initializer call.
  ///
  ///      let fn: Expr = ...
  ///      let zeroArgCall = fn.call()
  ///      let mixedCall = fn.call(Expr(literal: "hello"), at: Expr(literal: 0))
  public var call: FunctionCallCalleeProxy {
    FunctionCallCalleeProxy(calledExpression: self)
  }

  /// A proxy used to construct a member access.
  ///
  ///     let arrayVarExpr: Expr = ...
  ///     let countExpr = arrayVarExpr.member.count
  ///     let insertExpr = arrayVarExpr
  ///                        .member.insert
  ///                        .call(Expr(literal: "hello"), at: Expr(literal: 0))
  public var member: MemberAccessBaseProxy {
    MemberAccessBaseProxy(base: self)
  }
}

@dynamicMemberLookup
public struct MemberAccessBaseProxy {
  var base: Expr?

  public subscript(dynamicMember name: String) -> MemberAccessExpr {
    MemberAccessExpr(base: base, name: name)
  }
}

@dynamicCallable
public struct FunctionCallCalleeProxy {
  var calledExpression: Expr

  public func dynamicallyCall(withKeywordArguments args: KeyValuePairs<String, ExprSyntaxProtocol>) -> FunctionCallExpr {
    FunctionCallExpr(
      calledExpression: calledExpression,
      leftParen: .leftParen,
      argumentList: TupleExprElementList {
        for arg in args {
          TupleExprElement(
            label: arg.key.isEmpty ? nil : arg.key,
            expression: Expr(fromProtocol: arg.value)
          )
        }
      },
      rightParen: .rightParen,
      trailingClosure: nil,
      additionalTrailingClosures: nil
    )
  }
}
```

beccadax:
@swift-ci please test

beccadax:
@swift-ci please test

beccadax:
@swift-ci please test

Sources/SwiftSyntaxBuilder/Syntax+StringInterpolation.swift
```diff
@@ -129,6 +159,58 @@ enum SyntaxStringInterpolationError: Error, CustomStringConvertible {
...
+  ///
+  ///     let expr2 = Expr(literal: [0+1, 1+1, 2+1])
+  ///     // `expr2` is a syntax tree for `[1, 2, 3]`.
+  ///
+  /// Or interpolated into a Swift source code literal with the syntax
+  /// `\(literal: <value>)`:
+  ///
+  ///     let greeting = "Hello, world!"
+  ///     let expr1 = ExprSyntax("print(\(literal: greeting))")
+  ///     // `expr1` is a syntax tree for `print("Hello, world!")`
```
ahoppen, line 210: This documentation also appears in the definition of the `ExpressibleByLiteralSyntax` type. Is that repetition intentional? If it is, that’s fine with me but I think it might be unintentional.

Sources/SwiftSyntaxBuilder/Syntax+StringInterpolation.swift
```diff
@@ -160,3 +242,197 @@ extension SyntaxExpressibleByStringInterpolation {
...
+extension String: ExpressibleByLiteralSyntax {
+  public func makeLiteralSyntax() -> ExprSyntaxProtocol {
+    // TODO: Use a multi-line literal if there are more than N inner newlines.
+    StringLiteralExpr(content: self)
+  }
+}
+
+extension ExpressibleByLiteralSyntax where Self: BinaryInteger {
+  public func makeLiteralSyntax() -> ExprSyntaxProtocol {
+    // TODO: Radix selection? Thousands separators?
```
ahoppen, line 263: Personally, I think we can just remove this `TODO`. I can’t think of any situation where we could intelligently pick a radix other than 10 and thousands separators are also up to personal style and I don’t think we should assume that clients want them.

(same below for float literals)

Sources/SwiftSyntaxBuilder/Syntax+StringInterpolation.swift
```diff
@@ -160,3 +242,197 @@ extension SyntaxExpressibleByStringInterpolation {
...
+
+    return ArrayExpr(
+      leftSquare: .leftSquareBracket,
+      elements: ArrayElementList {
+        for elemSyntax in elemSyntaxes {
+          ArrayElement(expression: elemSyntax)
+        }
+      },
+      rightSquare: .rightSquareBracket
+    )
```
ahoppen, line 356: Couldn’t this just be `Array(elemSyntaxes).makeLiteralSyntax()`?

(similar below for `Dictionary` and `KeyValuePairs`)

Sources/SwiftSyntaxBuilder/Syntax+StringInterpolation.swift
ahoppen: There is this convenience initializer: https://github.com/apple/swift-syntax/blob/715f12517277d914c667ae09e734fa5ff3e6cd70/Sources/SwiftSyntaxBuilder/ConvenienceInitializers.swift#L111-L116

I think it should allow you to write it as follows (written in GitHub so not tested):

```swift
return FunctionCallExpr(callee: MemberAccessExpr(name: "some")) {
  TupleExprElement(expression: wrappedExpr)
}
```
 
---

As to your suggestion for `dynamicMemberLookup` and `dynamicCallable`, I think they read nice if you know what you’re doing but if you’re trying to understand unknown code, they seem a little to magic to me. I would like to try and get as far as possible with the convenience initializers first. If we decide that we can’t go any further on that front and it’s still too verbose, we can reconsider the dynamic calls.

Tests/SwiftSyntaxBuilderTest/StringInterpolation.swift
```diff
@@ -93,6 +107,121 @@ final class StringInterpolationTests: XCTestCase {
...
+    AssertStringsEqualWithDiff(d.description, #"print("Hello, world!\n")"#)
+
+    let e: ExprSyntax = "print(\(literal: "\"Hello\", world!\n"))"
+    AssertStringsEqualWithDiff(e.description, ##"print(#""Hello", world!\#n"#)"##)
+  }
+
+  func testInterpolationLiteralInt() {
+    func test<T: ExpressibleByIntegerLiteral & ExpressibleByLiteralSyntax>(with ty: T.Type, unsigned: Bool = false) {
+      let a: ExprSyntax = "print(\(literal: 42 as T))"
+      AssertStringsEqualWithDiff(a.description, #"print(42)"#, additionalInfo: String(describing: ty))
```
ahoppen, line 130: To make sure Xcode displays any test failures on the `test` line, could you pass down `line` here?

```suggestion
    func test<T: ExpressibleByIntegerLiteral & ExpressibleByLiteralSyntax>(with ty: T.Type, unsigned: Bool = false, file: StaticString = #file, line: UInt = #line) {
      let a: ExprSyntax = "print(\(literal: 42 as T))"
      AssertStringsEqualWithDiff(a.description, #"print(42)"#, additionalInfo: String(describing: ty), file: file, line: line)
```

(same for the other `AssertStringsEqualWithDiff` calls and the `test` functions in the `FloatLiteral` test)

beccadax: I actually did it this way initially, but I found that when a test case failed, I usually cared more about the value (e.g. `42`) than the type. Each of them includes the type in its `additionalInfo:` so that you can figure out exactly what failed in any case.

Sources/SwiftSyntaxBuilder/Syntax+StringInterpolation.swift
beccadax: That API ends up requiring callee erasure to `ExprSyntax`, so it's actually:

```suggestion
        return FunctionCallExpr(callee: ExprSyntax(MemberAccessExpr(name: "some"))) {
          TupleExprElement(expression: wrappedExpr)
        }
```

Which is messier but still an improvement.

beccadax: Ended up adding an overload that took `ExprSyntaxProtocol`. (I couldn't just do that to the original because that would break code literals.)

beccadax:
@swift-ci smoke test

Sources/SwiftSyntaxBuilder/Syntax+StringInterpolation.swift
bnbarham: Can't we just have the one `init` that's generic over `ExprSyntaxProtocol`?

beccadax:
@swift-ci test

Sources/SwiftSyntaxBuilder/Syntax+StringInterpolation.swift
beccadax: No. There are places where we say `FunctionCallExpr(callee: "some code fragment")`, and these need a concrete `ExprSyntax` overload to keep the compiler from inferring their type (invalidly) as `String`.

bnbarham: I assume you'd be against `"some code fragment" as ExprSyntax` 😅? If we have the generic rather than `ExprSyntaxProtocol` do we pick the `ExprSyntax` overload in that case? I'd still prefer that over taking `ExprSyntaxProtocol`. Little annoying that we'd need both of these wherever we're using `some` if we want the function to be able to take a literal though 🤔.

beccadax: I just meant that there are existing uses in the package, like SyntaxProtocol+Initializer.swift:65, that fail if I change it to `ExprSyntaxProtocol`. I assumed we didn't want those to break, but we could break them if you'd prefer.

And no, using a generic parameter instead of an existential doesn't help with the code literal issue. In both cases, the type checker doesn't consider types conforming to the protocol as candidates (and would have no way to know it should use `ExprSyntax` instead of some other conforming type even if it did).

bnbarham: I'm fine with the break, not sure how @ahoppen or @rintaro feel. It is... not as pretty 🤷I don't want this to block your PR, we can always change it later.

beccadax:
Failure was unrelated and a fix has apparently been merged.

beccadax:
@swift-ci test

Sources/SwiftSyntaxBuilder/Syntax+StringInterpolation.swift
ahoppen: I’m totally fine with all breaks of SwiftSyntaxBuilder. We break it all the time at the moment so one more breakage doesn’t make a difference. And I would prefer generic functions over existential wherever possible.
 
---

Maybe this example just shows that we need 

```swift
extension ExprSyntaxProtocol {
  var asBaseType: ExprSyntax { … }
}
```
which @bnbarham suggested before.

Sources/SwiftSyntaxBuilder/Syntax+StringInterpolation.swift
ahoppen: Oh, of course. Makes sense 👍

Tests/SwiftSyntaxBuilderTest/StringInterpolation.swift
ahoppen: If multiple tests fail I just dislike having multiple assertion failures on the same line but if you made this deliberate decision, I’m happy with it.

PR #1092, opened 2022-11-24
Refactor `Syntax` initializer

bnbarham:
Use generics rather than existentials, add `nil` defaults for *all*
optional nodes, and add a default for tokens where there's only a single
choice. Allows removing the default initializer in SwiftSyntaxBuilder,
since the regular initializer covers it entirely.

There's a small hack that adds a second initializer when a parameter
with a generic type has a empty default, since a
`Optional<BaseType>.none` default doesn't work in Swift 5.6. This can be
removed and the property initializer updated once 5.7 is our minimum
version.

### DIFF
--- a/Sources/SwiftSyntax/Raw/RawSyntax.swift
+++ b/Sources/SwiftSyntax/Raw/RawSyntax.swift
@@ -692,9 +692,24 @@ extension RawSyntax {
   static func makeLayout<C: Collection>(
     kind: SyntaxKind,
     from collection: C,
-    arena: SyntaxArena
+    arena: SyntaxArena,
+    leadingTrivia: Trivia? = nil,
+    trailingTrivia: Trivia? = nil
   ) -> RawSyntax where C.Element == RawSyntax? {
-    .makeLayout(kind: kind, uninitializedCount: collection.count, arena: arena) {
+    if leadingTrivia != nil || trailingTrivia != nil {
+      var layout = Array(collection)
+      if let leadingTrivia = leadingTrivia,
+         let idx = layout.firstIndex(where: {$0 != nil}) {
+        layout[idx] = layout[idx]!.withLeadingTrivia(leadingTrivia, arena: arena)
+      }
+      if let trailingTrivia = trailingTrivia,
+         let idx = layout.lastIndex(where: {$0 != nil}) {
+        layout[idx] = layout[idx]!.withTrailingTrivia(trailingTrivia, arena: arena)
+      }
+      return .makeLayout(kind: kind, from: layout, arena: arena)
+    }
+
+    return .makeLayout(kind: kind, uninitializedCount: collection.count, arena: arena) {
       _ = $0.initialize(from: collection)
     }
   }
--- a/Sources/SwiftSyntax/SyntaxNodes.swift.gyb.template
+++ b/Sources/SwiftSyntax/SyntaxNodes.swift.gyb.template
@@ -113,35 +113,60 @@ public struct ${node.name}: ${base_type}Protocol, SyntaxHashable {
     self._syntaxNode = Syntax(data)
   }
 
-  public init(
-%     for (index, child) in enumerate(node.children):
-%       comma = ',' if index != len(node.children) - 1 else ''
-%       param_type = child.name if child.node_choices else child.type_name
-%       if child.is_optional:
-%         param_type = param_type + "?"
-%       if child.is_unexpected_nodes():
-    _ ${child.swift_name}: ${param_type} = nil${comma}
-%       else:
-    ${child.swift_name}: ${param_type}${comma}
-%       end
-%     end
-  ) {
+  ${node.generate_initializer_decl(optional_base_as_missing=False, current_indentation="  ")} {
+%     if node.children:
     let layout: [RawSyntax?] = [
-%     for child in node.children:
-%       if child.is_optional:
+%       for child in node.children:
+%         if child.is_optional:
       ${child.swift_name}?.raw,
-%       else:
+%         else:
       ${child.swift_name}.raw,
+%         end
 %       end
-%     end
     ]
+%     end
     let data: SyntaxData = withExtendedLifetime(SyntaxArena()) { arena in
-      let raw = RawSyntax.makeLayout(kind: SyntaxKind.${node.swift_syntax_kind},
-        from: layout, arena: arena)
+%     if node.children:
+      let raw = RawSyntax.makeLayout(
+        kind: SyntaxKind.${node.swift_syntax_kind}, from: layout, arena: arena,
+        leadingTrivia: leadingTrivia, trailingTrivia: trailingTrivia)
+%     else:
+      let raw = RawSyntax.makeEmptyLayout(kind: SyntaxKind.${node.swift_syntax_kind}, arena: arena)
+%     end
       return SyntaxData.forRoot(raw)
     }
     self.init(data)
   }
+%     if node.has_optional_base_type_child():
+
+% # TODO: Remove when we no longer support compiling in Swift 5.6. Change the
+% # above constructor to use `Optional<BaseType.none` instead.
+  /// This initializer exists solely because Swift 5.6 does not support
+  /// `Optional<ConcreteType>.none` as a default value of a generic parameter.
+  /// The above initializer thus defaults to `nil` instead, but that means it
+  /// is not actually callable when either not passing the defaulted parameter,
+  /// or passing `nil`.
+  ///
+  /// Hack around that limitation using this initializer, which takes a
+  /// `Missing*` syntax node instead. `Missing*` is used over the base type as
+  /// the base type would allow implicit conversion from a string literal,
+  /// which the above initializer doesn't support.
+  ${node.generate_initializer_decl(optional_base_as_missing=True, current_indentation="  ")} {
+    self.init(
+      leadingTrivia: leadingTrivia,
+%     for child in node.children:
+%       if child.has_optional_base_type():
+      ${child.swift_name}: Optional<${child.type_name}>.none,
+%       elif child.is_unexpected_nodes():
+      ${child.swift_name},
+%       else:
+      ${child.swift_name}: ${child.swift_name},
+%       end
+%     end
+      trailingTrivia: trailingTrivia
+    )
+  }
+%     end
 %     for (idx, child) in enumerate(node.children):
 %       child_node = NODE_MAP.get(child.syntax_kind)
 %
--- a/Sources/SwiftSyntax/gyb_generated/syntax_nodes/SyntaxPatternNodes.swift
+++ b/Sources/SwiftSyntax/gyb_generated/syntax_nodes/SyntaxPatternNodes.swift
@@ -31,13 +31,9 @@ public struct MissingPatternSyntax: PatternSyntaxProtocol, SyntaxHashable {
     self._syntaxNode = Syntax(data)
   }
 
-  public init(
-  ) {
-    let layout: [RawSyntax?] = [
-    ]
+  public init() {
     let data: SyntaxData = withExtendedLifetime(SyntaxArena()) { arena in
-      let raw = RawSyntax.makeLayout(kind: SyntaxKind.missingPattern,
-        from: layout, arena: arena)
+      let raw = RawSyntax.makeEmptyLayout(kind: SyntaxKind.missingPattern, arena: arena)
       return SyntaxData.forRoot(raw)
     }
     self.init(data)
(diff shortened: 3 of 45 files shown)


Sources/SwiftSyntax/gyb_generated/syntax_nodes/SyntaxStmtNodes.swift
```diff
@@ -2960,13 +2984,33 @@ public struct ReturnStmtSyntax: StmtSyntaxProtocol, SyntaxHashable {
...
     }
     self.init(data)
   }
 
+  public init(
+    leadingTrivia: Trivia? = nil,
+    _ unexpectedBeforeReturnKeyword: UnexpectedNodesSyntax? = nil,
+    returnKeyword: TokenSyntax = .returnKeyword(),
+    _ unexpectedBetweenReturnKeywordAndExpression: UnexpectedNodesSyntax? = nil,
+    _ unexpectedAfterExpression: UnexpectedNodesSyntax? = nil,
```
bnbarham, line 3000: Not sure how I feel about this. Should we just have the one and if so, which one? I think my preference would be to skip `unexpectedBetweenReturnKeywordAndExpression`.

Sources/SwiftSyntax/Raw/RawSyntax.swift
```diff
@@ -700,11 +700,20 @@ extension RawSyntax {
...
-    .makeLayout(kind: kind, uninitializedCount: collection.count, arena: arena) {
+    var raw: RawSyntax = .makeLayout(kind: kind, uninitializedCount: collection.count, arena: arena) {
       _ = $0.initialize(from: collection)
     }
+    if let leadingTrivia = leadingTrivia {
+      raw = raw.withLeadingTrivia(leadingTrivia + (raw.formLeadingTrivia() ?? []), arena: arena) ?? raw
+    }
+    if let trailingTrivia = trailingTrivia {
+      raw = raw.withTrailingTrivia((raw.formTrailingTrivia() ?? []) + trailingTrivia, arena: arena) ?? raw
+    }
```
bnbarham, line 715: Is there a better way to do this Rintaro?

Sources/_SwiftSyntaxTestSupport/SyntaxProtocol+Initializer.swift
```diff
@@ -58,40 +58,40 @@ private class InitializerExprFormat: BasicFormat {
 }
 
 private extension TriviaPiece {
-  var initializerExpr: ExprSyntaxProtocol {
+  var initializerExpr: ExprSyntax {
     let (label, value) = Mirror(reflecting: self).children.first!
     switch value {
     case let value as String:
-      return FunctionCallExpr(callee: ".\(label!)") {
+      return ExprSyntax(FunctionCallExpr(callee: ".\(label!)") {
```
bnbarham, line 65: I'm tempted to add a `.asBaseType` (a default method on `Decl/Stmt/Expr/PatternSyntaxProtocol`) or something, thoughts?

ahoppen: If we want this as a postfix notation (and I think it might be nice). I think it should be `as(ExprSynax.self)`. But don’t most of these conversions now become unnecessary that we have the generic initializers?

Sources/SwiftSyntax/SyntaxNodes.swift.gyb.template
```diff
@@ -113,35 +113,54 @@ public struct ${node.name}: ${base_type}Protocol, SyntaxHashable {
...
-%       if child.is_optional:
-%         param_type = param_type + "?"
-%       if child.is_unexpected_nodes():
-    _ ${child.swift_name}: ${param_type} = nil${comma}
-%       else:
-    ${child.swift_name}: ${param_type}${comma}
-%       end
-%     end
-  ) {
+  ${node.generate_initializer_decl(True, "  ")} {
```
ahoppen, line 116: I think this is a lot clearer with argument labels (same below).

```suggestion
  ${node.generate_initializer_decl(include_optional_base=True, current_indentation="  ")} {
```

gyb_syntax_support/Node.py
```diff
@@ -123,3 +123,57 @@ def is_token(self):
...
+      skip_next = False
+      if self.children:
+        params.append("leadingTrivia: Trivia? = nil")
+      for child in self.children:
+        if skip_next:
+          skip_next = False
+          continue
+
+        if child.has_optional_base_type() and not include_optional_base:
+          skip_next = True
```
ahoppen, line 147: This is to skip the unexpected nodes, right? Could you add a comment for that?

ahoppen (commented):
Thanks a lot, this is a great simplification 🙏 

Do you know if the generic initializers affect code size or performance in either direction?

Sources/_SwiftSyntaxTestSupport/SyntaxProtocol+Initializer.swift
bnbarham: It would have to be `cast(ExprSyntax.self)` since `as` returns optional and it would be nice if we didn't have to specify `ExprSyntax` at all since we know this node's base type.

The conversion is necessary *because* of the generic initializers, unless the callsite uses `fromProtocol` instead 🤷. Unless I'm missing something, which is entirely possible.

gyb_syntax_support/Node.py
bnbarham: Yep, will do.

Sources/SwiftSyntax/SyntaxNodes.swift.gyb.template
bnbarham: Heh, yeah fair enough. Thanks

bnbarham (commented):
I can check performance/size next week

bnbarham:
> I can check performance/size next week

There's no real performance difference, but it seems to vary by ~1s (out of ~16s) each time I test so 🤷 Sizes have dropped slightly:
```
                            Pre      Post
libIDEUtils.a               32016096 31767448
libSwiftDiagnostics.a       32158792 31910144
libSwiftOperators.a         41534824 41288040
libSwiftParser.a            40506392 40257752
libSwiftParserDiagnostics.a 42527096 42279544
libSwiftRefactor.a          40836936 40589096
libSwiftSyntax.a            31890912 31642272
libSwiftSyntaxBuilder.a     48013032 46971680
libSwiftSyntaxParser.a      40854664 40606016
lib_SwiftSyntaxMacros.a     48306464 47265344

```

bnbarham:
@swift-ci please test

ahoppen (approved):
Thanks for checking performance and size.

bnbarham:
@swift-ci please test

bnbarham:
@swift-ci please test

Sources/SwiftSyntaxBuilder/ConvenienceInitializers.swift
```diff
@@ -141,12 +141,11 @@ extension FloatLiteralExprSyntax: ExpressibleByFloatLiteral {
...
 
 extension FunctionCallExpr {
-  // Need an overload that's explicitly `ExprSyntax` for code literals to work.
   /// A convenience initializer that allows passing in arguments using a result builder
   /// instead of having to wrap them in a `TupleExprElementList`.
   /// The presence of the parenthesis will be inferred based on the presence of arguments and the trailing closure.
-  public init(
-    callee: ExprSyntax,
+  public init<C: ExprSyntaxProtocol>(
+    callee: C,
```
bnbarham, line 148: Note: this means that we can't do `FunctionCallExpr(callee: "foo"...)` any more. Do we want to leave in the extra `ExprSyntax` overload so that doesn't have to be `FunctionCallExprSyntax(callee: ExprSyntax("Foo")...)`?

If so... maybe we should have those initializers in general, ie. the generic init + overload taking base types as well. Thoughts @ahoppen / @rintaro?

bnbarham: I've replaced all our calls for now. Let me know what you think (all wrapped in `ExprSyntax` now).

bnbarham:
@swift-ci please test

bnbarham:
@swift-ci please test

Sources/SwiftSyntax/gyb_generated/syntax_nodes/SyntaxPatternNodes.swift
```diff
@@ -154,13 +148,41 @@ public struct EnumCasePatternSyntax: PatternSyntaxProtocol, SyntaxHashable {
...
+        leadingTrivia: leadingTrivia, trailingTrivia: trailingTrivia)
       return SyntaxData.forRoot(raw)
     }
     self.init(data)
   }
 
+  public init(
+    leadingTrivia: Trivia? = nil,
+    _ unexpectedBeforeType: UnexpectedNodesSyntax? = nil,
+    type: MissingTypeSyntax? = nil,
```
bnbarham, line 162: This is an example of the hack - `Missing*` is used so that clients don't start depending on being able to pass a literal directly (which they could do if this was `TypeSyntax`). I've also removed the literal conformance from `Missing`, since it doesn't make sense to be able to construct missing nodes from a literal (it would currently always fatal error).

Once we support 5.7 as the minimum, we want to remove this initializer and update the previous to use `= Optional<TypeSyntax>.none` instead.

ahoppen: I don't quite understand why we need this initializer at all. Shouldn't clients be able to use the initializer that's generic on `<T: TypeSyntaxProtocol>`?

bnbarham: Calling the other constructor with the defaulted `type` will result in an error about being unable to infer `T`. We could fix this by specifying `type: T? = Optional<TypeSyntax>.none`, but that fails in Swift 5.6.

bnbarham:
@swift-ci please test

Sources/SwiftSyntax/gyb_generated/syntax_nodes/SyntaxPatternNodes.swift
ahoppen: Oh, this is so stupid.

Could you add your message above as a doc-comment somewhere?

bnbarham: I couldn't think of anywhere nice to add it as a doc comment. There is already a comment in `generate_default_initialization`. Is that fine or do you want something else?

bnbarham:
@swift-ci please test

bnbarham:
@swift-ci please test

Sources/SwiftSyntax/gyb_generated/syntax_nodes/SyntaxPatternNodes.swift
bnbarham: @ahoppen I've added a doc comment to each of these initializers.

Sources/SwiftSyntax/gyb_generated/syntax_nodes/SyntaxPatternNodes.swift
```diff
@@ -104,13 +102,51 @@ public struct EnumCasePatternSyntax: PatternSyntaxProtocol, SyntaxHashable {
...
+    caseName: TokenSyntax,
+    _ unexpectedBetweenCaseNameAndAssociatedTuple: UnexpectedNodesSyntax? = nil,
+    associatedTuple: TuplePatternSyntax? = nil,
+    _ unexpectedAfterAssociatedTuple: UnexpectedNodesSyntax? = nil,
+    trailingTrivia: Trivia? = nil
+  ) {
+    self.init(
+      leadingTrivia: leadingTrivia,
+      unexpectedBeforeType,
+      type: Optional<TypeSyntax>.none,
```
ahoppen, line 139: You are dropping `unexpectedBetweenTypeAndPeriod` here. AFAICT you should remove `skip_next` in `SyntaxNodes.swift.gyb.template`.

ahoppen (approved):
I’ve reviewed the entire PR again and have one more comment. Otherwise, let’s 🚢it

Sources/SwiftSyntax/gyb_generated/syntax_nodes/SyntaxPatternNodes.swift
bnbarham: Argh. Thanks. I thought I removed all that but apparently not.

bnbarham:
@swift-ci please test

PR #1098, opened 2022-12-02
[CMake] Build with library evolution and generate Swift interfaces.

DougGregor:
Enable library evolution for the CMake build of swift-syntax, so that plugins can be built with different compilers than the Swift compiler itself. Introduce `@frozen` in a few places to try to dodge what appears to be a miscompile in the older Swift 5.5 that's currently used in CI.


DougGregor:
@swift-ci please test

PR #1099, opened 2022-12-02
Don't drop token presence

bnbarham:
`with(Leading|Trailing)Trivia` was dropping the token's source presence for materialized tokens. Make sure to include it.


bnbarham:
@swift-ci please test

bnbarham:
Yes, we should have a test for this. But given we want to change the `with...` for Syntax anyway, I figure I can write some tests when we do that change.

FWIW this ends up causing every builder-created node to ignore the token presence, as all builder inits use with*Trivia.

ahoppen (approved):

PR #1100, opened 2022-12-02
[CMake] Only install .swiftinterface files from the module.

DougGregor:


DougGregor:
@swift-ci please test

PR #1103, opened 2022-12-02
Add a `detach()` to syntax nodes

bnbarham:
Allow clients to detach (ie. re-root) a syntax tree.

Also remove `RawSyntaxToSyntax` since this should only be used by this parser, which can just use the `Syntax` constructor directly instead.

### DIFF
--- a/Sources/SwiftSyntax/Syntax.swift
+++ b/Sources/SwiftSyntax/Syntax.swift
@@ -228,6 +228,12 @@ extension SyntaxProtocol {
     return _syntaxNode.data.raw
   }
 
+  /// Return this subtree with this node as the root, ie. detach this node
+  /// from its parent.
+  public func detach() -> Self {
+    return Syntax(raw: self.raw).cast(Self.self)
+  }
+
   public var kind: SyntaxKind {
     return raw.kind
   }
--- a/Tests/SwiftSyntaxTest/SyntaxTests.swift
+++ b/Tests/SwiftSyntaxTest/SyntaxTests.swift
@@ -10,8 +10,8 @@
 //
 //===----------------------------------------------------------------------===//
 
-import XCTest
 import SwiftSyntax
+import XCTest
 
 public class SyntaxTests: XCTestCase {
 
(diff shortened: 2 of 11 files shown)


rintaro (approved):

bnbarham:
@swift-ci please test

Tests/SwiftSyntaxTest/SyntaxTests.swift
```diff
@@ -25,4 +25,19 @@ public class SyntaxTests: XCTestCase {
     XCTAssertTrue(TokenListSyntax([TokenSyntax.funcKeyword(presence: .missing)]).hasError)
     XCTAssertFalse(TokenListSyntax([TokenSyntax.funcKeyword(presence: .present)]).hasError)
   }
+
+  public func testDetach() {
+    let s = StructDeclSyntax(
+      attributes: nil,
```
ahoppen, line 31: Not really related to this PR but can’t we omit these `nil` arguments now?

Sources/SwiftSyntax/Syntax.swift
```diff
@@ -228,6 +228,10 @@ extension SyntaxProtocol {
     return _syntaxNode.data.raw
   }
 
+  public func detach() -> Self {
```
ahoppen, line 231: Could you add a doc comment to explain what this method does?

ahoppen (commented):
Looks good to me but I don’t see any use cases for `detach`. What are you planning to use it for?

Tests/SwiftSyntaxTest/SyntaxTests.swift
bnbarham: No, I haven't merged that PR yet.

bnbarham:
> Looks good to me but I don’t see any use cases for `detach`. What are you planning to use it for?

Cases where you want to pass around a subtree without access to the rest of the tree.

Sources/SwiftSyntax/Syntax.swift
bnbarham: Sure

bnbarham:
@swift-ci please test

ahoppen (approved):

PR #1104, opened 2022-12-02
[Macros] Rework the ExpressionMacro API based on pitch feedback

DougGregor:
Make a number of changes to the API for expression macros:

* `MacroEvaluationContext` gets renamed to `MacroExpansionContext`
* Remove the source location converter from `MacroExpansionContext`, because one shouldn't have access to the whole source file
* Detach a macro expansion from its parent nodes, because one shouldn't have access to the whole source file
* Pass `MacroExpansionContext` inout
* Remove `MacroResult`; instead, allow diagnostics to be emitted via the macro expansion context
* Add a simplistic `MacroExpansionContext.createUniqueLocalName()`
* Rename `ExpressionMacro.apply` to `ExpressionMacro.expand`


Sources/_SwiftSyntaxMacros/ExpressionMacro.swift
```diff
@@ -19,11 +19,12 @@ import _CompilerPluginSupport
 
 /// Describes a macro that is explicitly expanded as an expression.
 public protocol ExpressionMacro: Macro {
-  /// Evaluate a macro described by the given macro expansion expression
+  /// Expand a macro described by the given macro expansion expression
   /// within the given context to produce a replacement expression.
-  static func apply(
-    _ macro: MacroExpansionExprSyntax, in context: MacroEvaluationContext
-  ) -> MacroResult<ExprSyntax>
+  static func expand(
```
rxwei, line 24: IMO the method should start with a noun phrase to be in line with the API guidelines, e.g. `expansion(for:in:)`, because the method is not meant to have side effects.

DougGregor: A later commit introduces the potential for side effects in the macro evaluation context, e.g., for emitting diagnostics and creating new unique local names.

DougGregor:
https://github.com/apple/swift/pull/62381

@swift-ci please test

Sources/_SwiftSyntaxMacros/ExpressionMacro.swift
DougGregor: I've thought about this more, and you're totally right. PR incoming to call it `expansion(for:in:)`.

PR #1108, opened 2022-12-05
Update CodeGeneration to latest SwiftSyntax version

ahoppen:
This updates the version lock of CodeGeneration to the latest SwiftSyntax version.

CodeGeneration still relies quite a bit on raw string interpolation. I’m hoping that we can remove some of these if we stop generating the node definitions using gyb and can thus improve the underlying data structure in the SyntaxSupport module.


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

PR #1110, opened 2022-12-05
Fix a crash in diagnostic generation if a primary associated type is not ended with a `>`

ahoppen:
In the new test case, we were parsing `:` as a right angle, which caused a crash when we were retrieving the token’s kind in diagnostic generation.


ahoppen:
@swift-ci Please test

Sources/SwiftParser/Nominals.swift
```diff
@@ -258,7 +258,12 @@ extension Parser {
           arena: self.arena))
       } while keepGoing != nil && loopProgress.evaluate(currentToken)
     }
-    let rangle = self.consumeAnyToken(remapping: .rightAngle)
+    let rangle: RawTokenSyntax
```
bnbarham, line 261: What's the precedence of `:` compared to `>`? Could we use `expect` here instead?

Tests/SwiftParserTest/DeclarationTests.swift
```diff
@@ -1333,6 +1333,19 @@ final class DeclarationTests: XCTestCase {
...
   }
+
+  func testPrimaryAssociatedTypeNotTerminatedWithAngleBracket() {
+    AssertParse(
+      "protocol1️⃣<2️⃣:3️⃣",
+      diagnostics: [
+        DiagnosticSpec(locationMarker: "1️⃣", message: "expected identifier in protocol"),
+        DiagnosticSpec(locationMarker: "2️⃣", message: "expected name in primary associated type clause"),
+        DiagnosticSpec(locationMarker: "2️⃣", message: "expected '>' to end primary associated type clause"),
+        DiagnosticSpec(locationMarker: "3️⃣", message: "expected type in type"),
```
bnbarham, line 1344: Similar to my PR - `expected type in type` is uh... not great 😅. I'm fine if you just want to add a TODO here

bnbarham: FWIW I think I would just not expect this particular diagnostic, the rest seem fine. If we ate the `:` we'd get an "unexpected :" instead which might make more sense.

bnbarham (approved):

Sources/SwiftParser/Nominals.swift
ahoppen: I don’t think we currently have a way to `expect` the `>` as a token prefix. The problem here is that sometimes the `>` can be part of a bigger token.

Tests/SwiftParserTest/DeclarationTests.swift
ahoppen: I adjusted a name and now it’s `expected type in inherited type`. I think that makes sense.

PR #1111, opened 2022-12-05
Import swift/test/Parse/recovery.swift

ahoppen:
I realized that I skipped importing this test case because my import script hade some unicode problems with it. Import it now. This shows a few more opportunities where we can improve diagnostics.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #1113, opened 2022-12-05
Format the swift-syntax repository using swift-format

ahoppen:
This does not format the generated files because it’s harder to debug generator changes if the newly generated files need to be formatted to get a meaningful diff.

This PR is intended to spark the discussion whether we want to format all files in SwiftSyntax using `swift-format`. If we agree that we should do this, I will add a verification pass in CI to assure that all files are formatted. 

I personally think that we should do this. In most cases I agree with the new format and not having to care about formatting when writing the code is a big win IMO.

rdar://103251585


.swift-format
```diff
@@ -0,0 +1,9 @@
+{
+    "version": 1,
+    "lineLength": 1000,
```
allevato, line 3: Is 1000 intentional here, rather than 100?

ahoppen: Yes, we (or rather I) don’t want a maximum line length.

ahoppen: Or is there a way to set the maximum line length to inifinity?

bnbarham: Is there a way to provide comment line lengths? I'm not sure why, but I do prefer *comments* having a max length. Maybe I'm weird though.

allevato: We don't currently support that (because who would be crazy enough to want such a thing 😁 ).

But it's a reasonable feature request, since it still ensures that indentation and other rules are consistently applied while honoring the author's own line breaks, but otherwise not forcing any new breaks.

allevato: > Is there a way to provide comment line lengths? I'm not sure why, but I do prefer _comments_ having a max length. Maybe I'm weird though.

Even when regular line lengths are enforced by swift-format, comments aren't, since we don't want to break weird cases (like cool ASCII art diagrams).

I'd like a more principled way of addressing this, but I'd probably tie it to using swift-markdown to reflow comments and requiring anyone doing the weird edge cases to put them in code blocks that wouldn't get reformatted.

CodaFi:
The styling with the trailing `)` on a new line looks odd to me. I think we ought to hew to the Swift standard library's whitespace conventions where possible.

.swift-format
bnbarham: Seems reasonable to me, thanks for the info 🙇

Sources/SwiftParser/CharacterInfo.swift
```diff
@@ -70,35 +70,30 @@ extension Unicode.Scalar {
   ) -> Bool {
     let info: Character.Info
     switch self.value {
-    case
-      // '0'-'9'
-      48, 49, 50, 51, 52, 53, 54, 55, 56, 57:
+    case  // '0'-'9'
+    48, 49, 50, 51, 52, 53, 54, 55, 56, 57:
```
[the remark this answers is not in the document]
ahoppen: @allevato Do you know if this is a bug in swift-format or whether this formatting is intentional? 

We can always turn off formatting for some of these lines with `// swift-format-ignore` as well.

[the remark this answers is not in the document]
allevato: This looks like a bug; we have some logic that tries to determine whether a comment is an end-of-line comment or some other inline comment so that we can do something that looks reasonable with it, and it looks like it's falling over in this case. At the _very least_ we shouldn't be dedenting that next line.

Sources/SwiftSyntax/SyntaxText.swift
```diff
@@ -88,8 +91,7 @@ public struct SyntaxText {
     guard !self.isEmpty && !other.isEmpty else {
       return self.isEmpty && other.isEmpty
     }
-    return (other.baseAddress! <= self.baseAddress! &&
-            self.baseAddress! + count <= other.baseAddress! + other.count)
+    return (other.baseAddress! <= self.baseAddress! && self.baseAddress! + count <= other.baseAddress! + other.count)
```
bnbarham, line 94: @allevato is this one intended because of the line length, or a bug?

allevato: This is because the current implementation only allows breaking on the left-hand side of an infix operator (with the exception of assignment operators, which may only break on the right). So we remove the break after the `&&` and then continue laying out the rest of the line, which never breaks again because of the high column limit.

With a lower line length, this would "auto-correct" because we would see that the `&&` and the text after it doesn't fit and we would force the break before the `&&` instead, but the high column limit is interfering here.

bnbarham: Ah so:
```
    return (other.baseAddress! <= self.baseAddress!
            && self.baseAddress! + count <= other.baseAddress! + other.count)
```
would have remained split?

allevato: Correct.

bnbarham: Is there a setting for that? I personally like them the other way 😅

allevato: There isn't a configuration option for that today. It probably wouldn't be difficult to add, but to date we haven't had the resources to maintain a large matrix of configurability options and make sure they all work well together, so we mainly prioritize the default style that we use within Google. (PRs welcome 😄)

bnbarham: Makes sense, thanks 🙇

ahoppen:
https://github.com/apple/swift/pull/62584

@swift-ci Please test

ahoppen:
https://github.com/apple/swift/pull/62584

@swift-ci Please test

ahoppen:
https://github.com/apple/swift/pull/62584

@swift-ci Please test

ahoppen:
https://github.com/apple/swift/pull/62584

@swift-ci Please test

ahoppen:
https://github.com/apple/swift/pull/62584

@swift-ci Please test

PR #1109, opened 2022-12-05
Move gyb file to SwiftSyntaxBuilder cod gen

kimdv:
- [x] Depends on #1116

It's the last gyb file in `Sources/SwiftSyntaxBuilder` 🎉

PR #1116, opened 2022-12-06
Remove leading and trailing space from `&`

kimdv:
In #1109 `return Self.parse(from: &parser)` is parsed to `return Self.parse(from: & parser)`.


kimdv:
@swift-ci please test

ahoppen (approved):

PR #1122, opened 2022-12-07
[CMake] Build libraries with only the install rpath.

DougGregor:
This avoids having extraneous rpaths pointing into build directories.


DougGregor:
@swift-ci please test

PR #1125, opened 2022-12-10
Improve diagnostics of `#available` and `#unavailable`

ahoppen:

PR #1126, opened 2022-12-10
Require a version on availability version restriction entries 🚥 #1125

ahoppen:
Depends on https://github.com/apple/swift-syntax/pull/1125.

The relevant commit is https://github.com/apple/swift-syntax/pull/1126/commits/fd2f4fb1cf57e6f47b4f75864a85c730eb45ac9c.

---

AFAICT we always require a version after the platform in `#available`. The C++ parser was a little complicated here, making assumptions but I couldn't find a case in which the version was truly optional.


ahoppen:
OK, we can’t do this because the availability macros can contain versions like `_iOS8Aligned` that are defined using `-define-availability` and those don’t take a version.

PR #1130, opened 2022-12-10
[Macros] Adjust macro declaration syntax to match the latest pitch. 

DougGregor:
Instead of macro definitions always being defined via
`= ModuleName.TypeName`, parse the more general `= expression`, which
is already captured by the `InitializerClause` grammar production.
Additionally, make the definition optional, which accounts for
built-in macros.


DougGregor:
@swift-ci please test

PR #1134, opened 2022-12-11
[Variadic Generics] Add a new type syntax node and parsing support for explicit pack references.

hborla:
* Add a new type node, `PackReferenceType`, for pack references spelled with a contextual `each` keyword.
* Add parsing support to produce `PackReferenceTypeSyntax` when the contextual `each` keyword is applied to a type.

### DIFF
--- a/gyb_syntax_support/TypeNodes.py
+++ b/gyb_syntax_support/TypeNodes.py
@@ -123,6 +123,14 @@ TYPE_NODES = [
              Child('Ellipsis', kind='EllipsisToken')
          ]),
 
+    # pack-reference-type -> 'each' type
+    Node('PackReferenceType', name_for_diagnostics='pack reference', kind='Type',
+         children=[
+             Child('EachKeyword', kind='ContextualKeyworkToken',
+                   text_choices=['each'], is_optional=False),
+             Child('PackType', kind='Type')
+         ]),
+
     # tuple-type-element -> identifier? ':'? type-annotation ','?
     Node('TupleTypeElement', name_for_diagnostics=None, kind='Syntax',
          traits=['WithTrailingComma'],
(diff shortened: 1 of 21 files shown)


hborla:
@swift-ci please test

gyb_syntax_support/TypeNodes.py
```diff
@@ -123,6 +123,14 @@
              Child('Ellipsis', kind='EllipsisToken')
          ]),
 
+    # pack-reference-type -> 'each' type
+    Node('PackReferenceType', name_for_diagnostics='pack reference', kind='Type',
```
ahoppen, line 127: This node is identical to the `ConstrainedSugarType`. What do you think about using the same node type for it instead of introducing a new node? I'm open to naming suggestions.

ahoppen (commented):
Could you add a couple of test cases? Interesting test cases that come to my mind are
- the example in your PR description
- `each MyType` as a parameter / return type (I assume that this should be invalid)

hborla:
> Could you add a couple of test cases? Interesting test cases that come to my mind are

I have test cases in my PR to the Swift repo here: https://github.com/apple/swift/pull/62509/files#diff-b4b3fbedbd5cb01a2a671c3e6e039e0bdc87f29ae3744e6aa6a713b32bb9ad04. These should exercise this code through parser validation. Do I need to add test cases here in addition to those?

gyb_syntax_support/TypeNodes.py
hborla: Hmm. You're right that the syntactic structure is the same (though it's parsed slightly differently), but conceptually a pack element type is very different from a constrained sugar type. Do you think that most tooling operating on these type syntax nodes would want to treat pack elements the same as constraint types? The only reason why I added a new syntax node is because I added a new `TypeRepr` on the Swift side to make type resolution easier, since the resolution logic is completely different.

hborla:
@ahoppen I added some test cases in https://github.com/apple/swift-syntax/pull/1134/commits/91035df80766e94bd838b6203651f9b51f7c849d! How do they look?

hborla:
@swift-ci please test

hborla:
@swift-ci please test

ahoppen (commented):
Thanks for adding the test cases. LGTM 👍

ahoppen (approved):

hborla:
@swift-ci please test

PR #1135, opened 2022-12-11
Remove `__FILE__` et al as keywords and syntax nodes.

DougGregor:
`__FILE__`, `__COLUMN__`, et al were keywords prior to Swift 2, and have only been kept around as keywords to provide a custom error message to rewrite to `#file`, `#column`, etc. It's incorrect to keep them as keywords, and that error should be implemented differently if we care about it.

These keywords were the only things keeping the syntax nodes for `#file` et al around; the correct spellings like `#file` are now macro expansion expression syntax nodes. So, remove these syntax nodes as well.


DougGregor:
https://github.com/apple/swift/pull/62512

@swift-ci please test

DougGregor:
https://github.com/apple/swift/pull/62512

@swift-ci please test

PR #1137, opened 2022-12-12
Fix trailing space on as keyword

kimdv:
Currently the parser makes `ReturnStmt("return self.asProtocol(SyntaxProtocol.self) as? DeclSyntaxProtocol")` into `return self.asProtocol(SyntaxProtocol.self) as ? DeclSyntaxProtocol`


kimdv:
@swift-ci please test

Sources/SwiftBasicFormat/generated/BasicFormat.swift
```diff
@@ -136,8 +136,16 @@ open class BasicFormat: SyntaxRewriter {
...
-    switch tokenKind {
+  open func requiresTrailingSpace(_ token: TokenSyntax) -> Bool {
+    if let nextTokenKind = token.nextToken(viewMode: .sourceAccurate)?.tokenKind {
+      switch nextTokenKind {
+      case .exclamationMark, .postfixQuestionMark: 
+        return false
+      default: 
+        break 
+      }
+    }
```
ahoppen, line 147: I don’t think this is true in general. For example, think about `return !myBool`. We need the exclamation mark here. 

I think we should just omit the space if the current token kind is in a hardcoded list. For now AFAICT that list should contain `as` and `init`.

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test

CodeGeneration/Sources/generate-swiftbasicformat/BasicFormatFile.swift
```diff
@@ -136,8 +136,21 @@ let basicFormatFile = SourceFile {
...
 
-    FunctionDecl("open func requiresTrailingSpace(_ tokenKind: TokenKind) -> Bool") {
-      SwitchStmt(expression: Expr("tokenKind")) {
+    FunctionDecl("open func requiresTrailingSpace(_ token: TokenSyntax) -> Bool") {
+      SwitchStmt("""
+        switch (token.tokenKind, token.nextToken(viewMode: .sourceAccurate)?.tokenKind) {
+        case (.asKeyword, .exclamationMark),
+             (.asKeyword, .postfixQuestionMark),
+             (.initKeyword, .postfixQuestionMark),
+             (.tryKeyword, .exclamationMark),
```
kimdv, line 145: Also added for `try!`

kimdv:
@swift-ci please test

ahoppen (approved):
Thank you. This looks good 👍

CodeGeneration/Sources/generate-swiftbasicformat/BasicFormatFile.swift
stevapple: Should `(.initKeyword, .leftParen)` be in the list to produce `init()` without the additional whitespace?

ahoppen: Yes. If you want to open a pull request for it, that would be a very worthwhile addition.

PR #1145, opened 2022-12-14
Disable two code blocks to make the package compiler on iOS-like platforms

ahoppen:
The targets that failed to build weren’t important for iOS but it’s good hygine to make the entire package compile on iOS-like platforms as well.

rdar://103294088


bnbarham (approved):

ahoppen:
@swift-ci Please test Linux

PR #1147, opened 2022-12-15
Fix trailing space on `init` keyword

stevapple:
Minor fix-up to #1137 which solves the problem of `BasicFormat` formatting `init()` into `init ()`.

PR #1148, opened 2022-12-15
Mark function parameters and tuple elements as indented

stevapple:
RT, this will allow `SwiftBasicFormat` to indent a multiline

- `TupleTypeElementList`
- `TupleExprElementList `
- `FunctionParameterList`

more nicely.


stevapple:
One of the tests depends on the style update in #1147, so that should be tested with & merged ahead

Tests/SwiftSyntaxBuilderTest/FunctionTests.swift
```diff
@@ -42,7 +42,55 @@ final class FunctionTests: XCTestCase {
...
+    let builder = FunctionDecl("""
+      func test(
+        _ p1: Int,
+        p2: Int,
+        _ p3: Int,
+        p4: Int,
+        _ p5: Int
+      ) -> Int {
+        return p1 + p2 + p3 + p4 + p5
+      }
```
kimdv, line 56: will this also format it like this if I make it like this?

```
func test(_ p1: Int, p2: Int, _ p3: Int, p4: Int, _ p5: Int ) -> Int {
         return p1 + p2 + p3 + p4 + p5
       }
```

stevapple: Nope, all these changes works for source input that already has `.newline` leading trivia, eg:
```swift
func test(_ p1: Int, p2: Int,
_ p3: Int, p4: Int, _ p5: Int
    ) -> Int {
         return p1 + p2 + p3 + p4 + p5
       }
```
should be formatted as
```swift
func test(_ p1: Int, p2: Int,
    _ p3: Int, p4: Int, _ p5: Int
) -> Int {
    return p1 + p2 + p3 + p4 + p5
}
```

kimdv: Could we add a test to ensure it then works that way. 

And also that 
```
func test(_ p1: Int, p2: Int, _ p3: Int, p4: Int, _ p5: Int) -> Int {
    return p1 + p2 + p3 + p4 + p5
}
```

also becomes

```
func test(_ p1: Int, p2: Int, _ p3: Int, p4: Int, _ p5: Int) -> Int {
    return p1 + p2 + p3 + p4 + p5
}
```

stevapple: It looks like `SwiftBasicFormat` will respect additional spaces, so those are not getting away unless we land a specific patch.

Going to add the common single-line case as a test, which just works out of the box:)

kimdv:
@swift-ci please test

kimdv (approved):

Tests/SwiftSyntaxBuilderTest/FunctionTests.swift
ahoppen: Yes, not removing whitespace is currently a limitation of how `BasicFormat` currently works. Without this rule we would over-indent some nodes that are being constructed with SwiftSyntaxBuilder and that thus already have indentation. If anyone is interested in trying to remove that restriction, I’d be open to a proposal. 

But I think for now this is fine. If you feed stupid formatting you’ll get stupid formatting out – I think this is fine.

ahoppen (approved):
Thanks you for the fix and all the test cases. Looks good to me 👍

PR #1144, opened 2022-12-14
Update swift-syntax revision

kimdv:
- [x] Depends on #1113

PR #1136, opened 2022-12-12
Move gyb files to codegen

kimdv:
- [x] Depends on #1128
- [x] Depends on #1144 
- [x] Depends on #1113

PR #1149, opened 2022-12-15
Move generation to a single executable

kimdv:
I open this PR with some initial ideas so we can start discuss on the solution.

- [x] Depends on #1113
- [x] Depends on #1136
- [x] Depends on #1144
- [x] Depends on #1151


CodeGeneration/Sources/generate-swiftsyntax/GenerateSwiftSyntax.swift
```diff
@@ -17,18 +17,75 @@ import Utils
...
+    try generateSwiftSyntax()
+    try generateSwiftSyntaxBuilder()
+  }
+  
+  private func generateSwiftBasicFormat() throws {
+    try generateFiles(
+    templates: [
+      (basicFormatFile, "BasicFormat.swift"),
+    ],
+    inSource: "SwiftBasicFormat")
```
ahoppen, line 39: What do you think about having something like:

```swift
let templates: [(builderNode: SourceFileSyntax, filename: String, module: String)] = [
  (basicFormatFile, "BasicFormat.swift", "SwiftBasicFormat"),
  (syntaxClassificationFile, "SyntaxClassification.swift", "IDEUtils")
]
```

and just generating everything from a single function?

build-script.py
```diff
@@ -517,26 +504,37 @@ def verify_code_generated_files(
...
+    )
+    self_swiftparser_generated_dir = os.path.join(
+        self_temp_dir, "Sources", "SwiftParser", "generated"
+    )
+    self_swiftsyntax_generated_dir = os.path.join(
+        self_temp_dir, "Sources", "SwiftSyntax", "generated"
+    )
+    self_swiftsyntaxbuilder_generated_dir = os.path.join(
+        self_temp_dir, "Sources", "SwiftSyntaxBuilder", "generated"
+    )
```
ahoppen, line 533: Since e you are refactoring this: I think this can become a loop now.

```python
self_temp_dir = tempfile.mkdtemp()

run_code_generation(…)

for module in ["SwiftBasicFormat", "IDEUtils", …]:
  self_generated_dir = os.path.join(self_temp_dir, "Sources", module, "generated")
  user_generated_dir = os.path.join(<something>, module, "generated")
  check_generated_files_match(self_generated_dir, user_generated_dir)
```

ahoppen (commented):
Much cleaner. Thank you.

One additional thing I’ve got on my wish list (doesn’t have to be part of this PR): Add a parameter to `generate-swiftsyntax` to only generate a single file. I would have found this useful while debugging why a file wasn’t generated the way I expected.

ahoppen (approved):

ahoppen:
@swift-ci Please test

PR #1154, opened 2022-12-16
Prepend/append trivia passed to RawSyntax.makeLayout instead of overriding it

ahoppen:
Previously, the first/last node's leading/trailing trivia was overriden. That’s surprising. Instead, we want to prepend/append the new trivia.

Discovered by @kimdv here: https://github.com/apple/swift-syntax/pull/1151/files#r1050708880

### DIFF
--- a/Sources/SwiftSyntax/Raw/RawSyntax.swift
+++ b/Sources/SwiftSyntax/Raw/RawSyntax.swift
@@ -714,12 +714,12 @@ extension RawSyntax {
       if let leadingTrivia = leadingTrivia,
         let idx = layout.firstIndex(where: { $0 != nil })
       {
-        layout[idx] = layout[idx]!.withLeadingTrivia(leadingTrivia, arena: arena)
+        layout[idx] = layout[idx]!.withLeadingTrivia(leadingTrivia + (layout[idx]?.formLeadingTrivia() ?? []), arena: arena)
       }
       if let trailingTrivia = trailingTrivia,
         let idx = layout.lastIndex(where: { $0 != nil })
       {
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

bnbarham:
Personally I find it somewhat odd to add rather than replace here. It's also unclear to me from the referenced PR how replacing is causing issues there. It should only replace if leading/trailing was set to non-nil, which it doesn't seem to be?

As an aside, while looking at this, I noticed that swift-format is formatting the buildable nodes in an ... interesting way:
```
  public init (leadingTrivia: Trivia? = nil, ..., @MemberDeclListBuilder membersBuilder: () -> MemberDeclListSyntax = {
    MemberDeclListSyntax([])
  }, trailingTrivia: Trivia? = nil) {
```

ahoppen:
What’s happening is that we start off with an `IntegerLiteralExprSyntax` that contains both leading and trailing trivia. 

```swift
// Inner leading
42
// Inner trailing
```

Now, when I’m wrapping the `IntegerLiteralExprSyntax` in an `ExpressionPatternSyntax` and passing the `leadingTrivia` and `trailingTrivia`, my initial mental model would be that the inner trivia (passed before) are attached to the `IntegerLiteralExprSyntax` while the outer trivia that I’m passing now are attached to the `ExpressionPatternSyntax`. Thus they shouldn’t override each other. But since we attach trivia to tokens, we need to attach both to the `42` token. Still, I think maintaining the naïve mental model is the least surprising approach here and thus I’d prefer to combine the trivia rather than dropping whatever leading trivia the first token in the node currently has.

PR #1155, opened 2022-12-16
Drop example/builtin macros from _SwiftSyntaxMacros and simplify API.

DougGregor:
Slim down the public API of the `_SwiftSyntaxMacros` module considerably:
* Drop the example/builtin macros, moving them into the tests
* Drop the "expand-macros" operation from swift-parser-cli
* Take `MacroSystem` and all of its various errors out of the public API
* Instead, introduce an `expand(macros:in:)` function on `SyntaxProtocol` to make it easy to do the syntactic expansions.

While here, fix up diagnostic propagation in `expand(macros:in:)`.


DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

PR #1157, opened 2022-12-17
Fix warnings and docs in the syntax builder sample

TiagoMaiaL:
There are some warnings in `CodeGenerationUsingSwiftSyntaxBuilder.swift`. This MR fixes them:
Screenshot 2022-12-16 at 23 55 28

This MR also fixes the docs, applying the right indentation, as printed by the executable:
```swift
struct Person {
    var lastName: String
    func withLastName(_ lastName: String) -> Person {
        var result = self
        result.lastName = lastName
        return result
    }
    var firstName: String
    func withFirstName(_ firstName: String) -> Person {
        var result = self
        result.firstName = firstName
        return result
    }
    var age: Int
    func withAge(_ age: Int) -> Person {
        var result = self
        result.age = age
        return result
    }
}
```


kimdv (approved):
Thanks 🥳

kimdv:
@swift-ci please test

ahoppen (approved):
Thank you!

PR #1158, opened 2022-12-19
Move gyb files

kimdv:
Yet another PR moving gyb files to swift code gen ☠️

- [x] Depends on #1161
- [x] Depends on #1162
- [ ] Depends on #1164

I'm not able to move the Trivia.swift.gyb because of `Swift/Character.swift:177: Fatal error: Can't form a Character from a String containing more than one extended grapheme cluster`
Even we have it in raw strings using `#` it's not working.


kimdv:
@swift-ci please test

ahoppen:
Thanks for killing the remaining gyb. 

When the dependencies are merged, could you open separate PRs for the individual files? That makes reviewing them easier.

You can also open a PR for Character.swift and I’ll try to figure out what’s going wrong.

kimdv:
Closing in favour of individual PRs

PR #1160, opened 2022-12-19
Eliminate SwiftCompilerSupport module.

DougGregor:
This module is no longer needed by the Swift compiler, because we have subsumed its functionality into the compiler's ASTGen module.


DougGregor:
@swift-ci please test

PR #1164, opened 2022-12-21
Fix spacing around `<`, `>`, `*`, `!` and `?`

### LINKED ISSUE #853: TryExpr with questionOrExclamationMark does not build correct syntax.
### Description

In case of TryExpr with `questionOrExclamationMark`,  `SwiftSyntaxBuilder` should build syntax:  `try! a.b()` instead of `try !a.b()`

### Steps to Reproduce

Add  `questionOrExclamationMark` DoStmtTests:
```swift
let buildable = DoStmt(
      body: CodeBlock(statementsBuilder: {
          CodeBlockItem(item: TryExpr(questionOrExclamationMark: .exclamationMark, expression: FunctionCallExpr(MemberAccessExpr(base: "a", name: "b"))))
      }),
      catchClauses: [
        CatchClause(CatchItemList {
          CatchItem(pattern: "Error1")
          CatchItem(pattern: "Error2")
        }) {
          FunctionCallExpr("print") {
            TupleExprElement(expression: StringLiteralExpr("Known error"))
          }
        },
        CatchClause(CatchItemList {
          CatchItem(
            pattern: "Error3", whereClause: WhereClause(guardResult: MemberAccessExpr(base: "error", name: "isError4")))
        }) {
          ThrowStmt(expression: MemberAccessExpr(base: "Error4", name: "error3"))
        },
        CatchClause {
          FunctionCallExpr("print") {
            TupleExprElement(expression: "error")
          }
        }
      ])
```

kimdv:
This fixes so we don't get `var newLayout: ContiguousArray < RawSyntax ? > ?`when doing `VariableDecl("var newLayout: ContiguousArray<RawSyntax?>?")`.

If you can think of other test cases that I nee to add, to ensure we have expected behaviour different places feel free to comment then.

this PR also fixes #853


kimdv:
@swift-ci please test

kimdv:
this PR also fixes #853

CodeGeneration/Sources/generate-swiftbasicformat/BasicFormatFile.swift
```diff
@@ -122,6 +122,19 @@ let basicFormatFile = SourceFile {
...
 
     FunctionDecl("open func requiresLeadingSpace(_ token: TokenSyntax) -> Bool") {
+      SwitchStmt("""
+        switch (token.tokenKind, token.previousToken(viewMode: .sourceAccurate)?.tokenKind) {
+        case (.leftAngle, .identifier(_)),
+             (.rightAngle, .identifier(_)),
+             (.rightAngle, .postfixQuestionMark):
+          return false
+        case (.spacedBinaryOperator(let `operator`), .leftParen):
+          return `operator` != "*"
```
ahoppen, line 132: What’s the motivation for this special case?

CodeGeneration/Sources/generate-swiftbasicformat/BasicFormatFile.swift
```diff
@@ -122,6 +122,19 @@ let basicFormatFile = SourceFile {
     }
 
     FunctionDecl("open func requiresLeadingSpace(_ token: TokenSyntax) -> Bool") {
+      SwitchStmt("""
+        switch (token.tokenKind, token.previousToken(viewMode: .sourceAccurate)?.tokenKind) {
```
ahoppen, line 126: What do you think about flipping the order of the elements in this tuple here so we are matching agains the token kinds in the same order as they appear in source code?

ahoppen (commented):
I’ve got a few test cases that I think would be interesting and suspect that some of them will break with your changes 😬 Sorry. 
- `myOptional!.someProperty`
- `lhs < rhs`
- `lhs > rhs`
- `(lhs1, lhs2) < (rhs1, rhs2)`
- `MyGeneric<String>!` (an implicitly unwrapped generic type)
- `!(true)`

If you want to fix the formatting of `<` in generics, I think it might need to be explicitly done inside the generic parameter clause. I don’t think we’ve got any precedence for that though.

kimdv:
Add test cases for 

- [ ] `myOptional!.someProperty`
- [ ] `lhs < rhs`
- [ ] `lhs > rhs`
- [ ] `(lhs1, lhs2) < (rhs1, rhs2)`
- [ ] `MyGeneric<String>!` (an implicitly unwrapped generic type)
- [ ] `!(true)`

kimdv:
@ahoppen I've added tests cases with the examples you suggested. It seems to work. 😬
If needed I can try to run code gen with the latest commit, and see what it changes and open a PR so we can see it?

CodeGeneration/Sources/generate-swiftsyntax/templates/basicformat/BasicFormatFile.swift
```diff
@@ -122,6 +122,20 @@ let basicFormatFile = SourceFile {
     }
 
     FunctionDecl("open func requiresLeadingSpace(_ token: TokenSyntax) -> Bool") {
+      SwitchStmt("""
+        switch (token.previousToken(viewMode: .sourceAccurate)?.tokenKind, token.tokenKind) {
+        case (.identifier(_), .leftAngle),
+             (.identifier(_), .rightAngle),
+             (.postfixQuestionMark, .leftAngle),
+             (.postfixQuestionMark, .rightAngle):
```
ahoppen, line 130: Don’t we need the same for 
- exclamation marks for implicitly unwrapped optional types (`MyGeneric<Foo?>`)
- right square for array and dictionary types (`MyGeneric<[Foo]>`)

Incidentally, what breaks if we unconditionally don’t require leading space for `<` and `>`?

CodeGeneration/Sources/generate-swiftsyntax/templates/basicformat/BasicFormatFile.swift
```diff
@@ -141,11 +155,20 @@ let basicFormatFile = SourceFile {
         switch (token.tokenKind, token.nextToken(viewMode: .sourceAccurate)?.tokenKind) {
         case (.asKeyword, .exclamationMark),
              (.asKeyword, .postfixQuestionMark),
+             (.exclamationMark, .period),
              (.initKeyword, .leftParen),
              (.initKeyword, .postfixQuestionMark),
+             (.leftAngle, .identifier(_)),
+             (.rightAngle, .leftParen),
+             (.rightAngle, .postfixQuestionMark),
+             (.postfixQuestionMark, .leftParen),
```
ahoppen, line 164: Do we need to do the same for `exclamationMark`, e.g. for `myOptionalClosure!()`

ahoppen (commented):
Sorry, I’ve got some more thoughts inline. I hope that we can converge on a solution soon.

CodeGeneration/Sources/generate-swiftsyntax/templates/basicformat/BasicFormatFile.swift
```diff
@@ -141,11 +155,20 @@ let basicFormatFile = SourceFile {
...
+             (.postfixQuestionMark, .rightAngle),
+             (.postfixQuestionMark, .rightParen),
              (.tryKeyword, .exclamationMark),
              (.tryKeyword, .postfixQuestionMark):
           return false
+        case (.spacedBinaryOperator(let `operator`), .comma):
+          return `operator` != "*"
```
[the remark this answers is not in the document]
kimdv: https://github.com/apple/swift-syntax/pull/1164#discussion_r1060621893

I only found this case. Will not affect more than necessary

CodeGeneration/Sources/generate-swiftsyntax/templates/basicformat/BasicFormatFile.swift
```diff
@@ -141,11 +155,20 @@ let basicFormatFile = SourceFile {
...
              (.asKeyword, .postfixQuestionMark),
+             (.exclamationMark, .period),
              (.initKeyword, .leftParen),
              (.initKeyword, .postfixQuestionMark),
+             (.leftAngle, .identifier(_)),
+             (.rightAngle, .leftParen),
+             (.rightAngle, .postfixQuestionMark),
+             (.postfixQuestionMark, .leftParen),
+             (.postfixQuestionMark, .rightAngle),
+             (.postfixQuestionMark, .rightParen),
```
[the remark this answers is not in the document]
kimdv: Do you mean add a comment with an example?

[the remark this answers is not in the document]
ahoppen: Yes, exactly.

CodeGeneration/Sources/generate-swiftsyntax/templates/basicformat/BasicFormatFile.swift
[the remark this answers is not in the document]
ahoppen: I would prefer to decide on whether the operator is `*` or not.

Either this rule works in all situations in which operators occur, in which case it should also apply for e.g. `+`. If it breaks other operators in expression positions, it will also be broken for `*`.

CodeGeneration/Sources/generate-swiftsyntax/templates/basicformat/BasicFormatFile.swift
kimdv: If we remove trailing and leading space from `<` and `>` the test suite pass

CodeGeneration/Sources/generate-swiftsyntax/templates/basicformat/BasicFormatFile.swift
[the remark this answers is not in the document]
kimdv: So you suggest we do it for all operators?

CodeGeneration/Sources/generate-swiftsyntax/templates/basicformat/BasicFormatFile.swift
```diff
@@ -139,12 +150,19 @@ let basicFormatFile = SourceFile {
...
-             (.tryKeyword, .postfixQuestionMark):
+        case (.asKeyword, .exclamationMark), // Ensures there is not space in `as!`
+             (.asKeyword, .postfixQuestionMark), // Ensures there is not space in `as?`
+             (.exclamationMark, .leftParen), // Ensures there is not space in `!(`
+             (.exclamationMark, .period), // Ensures there is not space in `!.`
+             (.initKeyword, .leftParen), // Ensures there is not space in `init(`
+             (.initKeyword, .postfixQuestionMark), // Ensures there is not space in `init?`
+             (.postfixQuestionMark, .leftParen), // Ensures there is not space in `?(`
+             (.postfixQuestionMark, .rightAngle), // Ensures there is not space in `?<`
+             (.postfixQuestionMark, .rightParen), // Ensures there is not space in `?(`
```
ahoppen, line 161: Especially for the cases that don't contain keywords I think an example that shows a little bit more context could be useful. E.g. for this one
```suggestion
             (.postfixQuestionMark, .rightParen), // Ensures there is not space in `init?()`
```

CodeGeneration/Sources/generate-swiftsyntax/templates/basicformat/BasicFormatFile.swift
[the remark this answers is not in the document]
ahoppen: Yes, that's what I'm suggesting. Because otherwise it will be surprising why `*` is formatted differently than `+`.

CodeGeneration/Sources/generate-swiftsyntax/templates/basicformat/BasicFormatFile.swift
ahoppen: Do you mean the test suite doesn't pass? What exactly break?

kimdv: Nothing, all the tests are green ✅

CodeGeneration/Sources/generate-swiftsyntax/templates/basicformat/BasicFormatFile.swift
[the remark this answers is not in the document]
kimdv: Fixed. We do now do the same for alle spaced binary operators

CodeGeneration/Sources/generate-swiftsyntax/templates/basicformat/BasicFormatFile.swift
ahoppen: In that case, I would prefer to just change to set `requires_leading_space` to `False` for `<` and `>`.

ahoppen: I see that you already changed that. In that case, I think you can remove the special cases here.

ahoppen (approved):
One comment that a few of the special cases regarding angle brackets are superfluous now, otherwise LGTM.

CodeGeneration/Sources/generate-swiftsyntax/templates/basicformat/BasicFormatFile.swift
kimdv: The code was outdated. It should be alright now.

kimdv:
@swift-ci please test

CodeGeneration/Sources/generate-swiftsyntax/templates/basicformat/BasicFormatFile.swift
kimdv: @ahoppen will you resolve this if you agree?

CodeGeneration/Sources/generate-swiftsyntax/templates/basicformat/BasicFormatFile.swift
```diff
@@ -146,12 +155,19 @@ let basicFormatFile = SourceFile {
...
-             (.tryKeyword, .postfixQuestionMark):
+        case (.asKeyword, .exclamationMark), // Ensures there is not space in `as!`
+             (.asKeyword, .postfixQuestionMark), // Ensures there is not space in `as?`
+             (.exclamationMark, .leftParen), // Ensures there is not space in `myOptionalClosure!()`
+             (.exclamationMark, .period), // Ensures there is not space in `myOptionalBar!.foo()`
+             (.initKeyword, .leftParen), // Ensures there is not space in `init()`
+             (.initKeyword, .postfixQuestionMark), // Ensures there is not space in `init?`
+             (.postfixQuestionMark, .leftParen), // Ensures there is not space in `init?()`
+             (.postfixQuestionMark, .rightAngle), // Ensures there is not space in `ContiguousArray<RawSyntax?>`
+             (.postfixQuestionMark, .rightParen), // Ensures there is not space in `myOptionalClosure?()`
```
ahoppen, line 166: I think those are also no longer necessary

ahoppen (approved):

CodeGeneration/Sources/generate-swiftsyntax/templates/basicformat/BasicFormatFile.swift
ahoppen: I see they are necessary to remove the trailing space on `postfixQuestionMark`

PR #1169, opened 2022-12-25
Remove classification gyb

kimdv:
First step for removing the dependency on gyb files


ahoppen (commented):
Don’t you also need to delete Classification.swift.gyb?

kimdv:
> Don’t you also need to delete Classification.swift.gyb?

🤦‍♂️

ahoppen (approved):

kimdv:
Will o

PR #1173, opened 2022-12-28
Explicitly model all supported contextual keywords in the `contextualKeyword` `RawTokenKind`

ahoppen:
The main change here is that the `contextualKeyword` `TokenKind` and `RawTokenKind` no longer have a string as an associated value but an enum that enumerates all known contextual keywords.

In follow-up commits, this allows us to rename `contextualKeyword` to `keyword` and merge all the other known keywords into the `Keyword` enum that currently enumerates all contextual keywords. That way we can remove the distinction between contextual keywords and keywords we currently have.

I’m also thinking that we can use this additional structure to guarantee a little more type safety in the parser, for example by generating `RawTokenKindSubset`s for children that have `text_choices` set, but I still need to look into that.

### DIFF
--- a/Sources/SwiftParser/Modifiers.swift
+++ b/Sources/SwiftParser/Modifiers.swift
@@ -44,7 +44,7 @@ extension Parser {
         // treat 'class' as a modifier in the case of a following CC
         // token, we cannot be sure there is no intention to override
         // or witness something static.
-        if lookahead.atStartOfDeclaration() || lookahead.atContextualKeyword("override") {
+        if lookahead.atStartOfDeclaration() || lookahead.at(.contextualKeyword(.override)) {
           let classKeyword = self.eat(handle)
           elements.append(
             RawDeclModifierSyntax(
@@ -59,30 +59,31 @@ extension Parser {
         }
       case (.unowned, _)?:
         elements.append(self.parseUnownedModifier())
-      case (.final, _)?,
-        (.required, _)?,
-        (.optional, _)?,
-        (.lazy, _)?,
-        (.dynamic, _)?,
-        (.infix, _)?,
-        (.prefix, _)?,
-        (.postfix, _)?,
-        (.__consuming, _)?,
-        (.mutating, _)?,
-        (.nonmutating, _)?,
-        (.convenience, _)?,
-        (.override, _)?,
-        (.weak, _)?,
-        (.indirect, _)?,
-        (.isolated, _)?,
-        (.async, _)?,
-        (.nonisolated, _)?,
-        (.distributed, _)?,
-        (._const, _)?,
-        (._local, _)?,
-        (.__setter_access, _)?,
-        (.reasync, _)?:
-        elements.append(self.parseSimpleModifier())
+      case (.final, let handle)?,
+        (.required, let handle)?,
+        (.optional, let handle)?,
+        (.lazy, let handle)?,
+        (.dynamic, let handle)?,
+        (.infix, let handle)?,
+        (.prefix, let handle)?,
+        (.postfix, let handle)?,
+        (.__consuming, let handle)?,
+        (.mutating, let handle)?,
+        (.nonmutating, let handle)?,
+        (.convenience, let handle)?,
+        (.override, let handle)?,
+        (.weak, let handle)?,
+        (.indirect, let handle)?,
+        (.isolated, let handle)?,
+        (.async, let handle)?,
+        (.nonisolated, let handle)?,
+        (.distributed, let handle)?,
+        (._const, let handle)?,
+        (._local, let handle)?,
+        (.__setter_access, let handle)?,
+        (.reasync, let handle)?:
+        let keyword = self.eat(handle)
+        elements.append(RawDeclModifierSyntax(name: keyword, detail: nil, arena: self.arena))
       case (.rethrows, _)?:
         fallthrough
       case nil:
@@ -94,11 +95,6 @@ extension Parser {
 }
 
 extension Parser {
-  mutating func parseSimpleModifier() -> RawDeclModifierSyntax {
-    let keyword = self.consumeAnyToken(remapping: .contextualKeyword)
-    return RawDeclModifierSyntax(name: keyword, detail: nil, arena: self.arena)
-  }
-
   mutating func parseModifierDetail() -> RawDeclModifierDetailSyntax {
     let (unexpectedBeforeLeftParen, leftParen) = self.expect(.leftParen)
     let detailToken = self.consumeAnyToken()
@@ -114,7 +110,7 @@ extension Parser {
   }
 
   mutating func parseUnownedModifier() -> RawDeclModifierSyntax {
-    let (unexpectedBeforeKeyword, keyword) = self.expectContextualKeyword("unowned")
+    let (unexpectedBeforeKeyword, keyword) = self.expect(.contextualKeyword(.unowned))
 
     let detail: RawDeclModifierDetailSyntax?
     if self.at(.leftParen) {
@@ -132,7 +128,7 @@ extension Parser {
   }
 
   mutating func parsePackageAccessLevelModifier() -> RawDeclModifierSyntax {
-    let (unexpectedBeforeName, name) = self.expectContextualKeyword("package")
+    let (unexpectedBeforeName, name) = self.expect(.contextualKeyword(.package))
     let details = self.parseAccessModifierDetails()
     return RawDeclModifierSyntax(
       unexpectedBeforeName,
@@ -143,7 +139,7 @@ extension Parser {
   }
 
   mutating func parseOpenAccessLevelModifier() -> RawDeclModifierSyntax {
-    let (unexpectedBeforeName, name) = self.expectContextualKeyword("open")
+    let (unexpectedBeforeName, name) = self.expect(.contextualKeyword(.open))
     let details = self.parseAccessModifierDetails()
     return RawDeclModifierSyntax(
       unexpectedBeforeName,
@@ -174,12 +170,12 @@ extension Parser {
 
     let unexpectedBeforeDetail: RawUnexpectedNodesSyntax?
     let detail: RawTokenSyntax
-    if let setHandle = canRecoverToContextualKeyword("set", precedence: .weakBracketClose) {
+    if let setHandle = canRecoverTo(.contextualKeyword(.set), recoveryPrecedence: .weakBracketClose) {
       (unexpectedBeforeDetail, detail) = eat(setHandle)
     } else {
       unexpectedBeforeDetail = nil
       detail = RawTokenSyntax(
-        missing: .contextualKeyword,
+        missing: .contextualKeyword(.set),
         text: "set",
         arena: arena
       )
--- a/Sources/SwiftParser/Types.swift
+++ b/Sources/SwiftParser/Types.swift
@@ -232,7 +232,7 @@ extension Parser {
     stopAtFirstPeriod: Bool = false
   ) -> RawTypeSyntax {
     var base: RawTypeSyntax
-    switch self.currentToken.tokenKind {
+    switch self.currentToken.rawTokenKind {
     case .capitalSelfKeyword,
       .anyKeyword,
       .identifier:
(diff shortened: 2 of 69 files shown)


Sources/SwiftParser/Modifiers.swift
```diff
@@ -59,30 +59,31 @@ extension Parser {
...
-        (.isolated, _)?,
-        (.async, _)?,
-        (.nonisolated, _)?,
-        (.distributed, _)?,
-        (._const, _)?,
-        (._local, _)?,
-        (.__setter_access, _)?,
-        (.reasync, _)?:
-        elements.append(self.parseSimpleModifier())
+      case (.final, let handle)?,
```
bnbarham, line 62: There's a few places where this is the case, but IMO it'd be better to assign kind/handle first and then switch over kind rather than having to do (.kind, let handle) or (.kind, _) everywhere. Not really relevant to this, I just noticed it while going through the changes.

Sources/SwiftParser/Types.swift
```diff
@@ -141,7 +141,7 @@ extension Parser {
   @_spi(RawSyntax)
   public mutating func parseSimpleOrCompositionType() -> RawTypeSyntax {
     // 'each' is a contextual keyword for a pack reference.
-    if let each = consume(ifAny: [], contextualKeywords: ["each"]) {
+    if let each = consume(ifAny: [.contextualKeyword(.each)]) {
```
bnbarham, line 144: Nit: could just be `if:`

Sources/SwiftParser/Types.swift
```diff
@@ -263,8 +263,8 @@ extension Parser {
             )
           )
           break
-        } else if self.atContextualKeyword("Type") || self.atContextualKeyword("Protocol") {
-          let typeOrProtocol = self.consumeAnyToken(remapping: .contextualKeyword)
+        } else if self.at(.contextualKeyword(.Type)) || self.at(.contextualKeyword(.Protocol)) {
+          let typeOrProtocol = self.consume(if: .contextualKeyword(.Type)) ?? self.consume(if: .contextualKeyword(.Protocol))!
```
bnbarham, line 267: IMO we should add a 2 and 3 arg consume to avoid the array creation and then use that here. The only non-2/3 kind calls should probably be subsets anyway.

Sources/SwiftSyntax/Raw/RawSyntax.swift
```diff
@@ -526,6 +526,9 @@ extension RawSyntax {
     initializingLeadingTriviaWith: (UnsafeMutableBufferPointer<RawTriviaPiece>) -> Void,
     initializingTrailingTriviaWith: (UnsafeMutableBufferPointer<RawTriviaPiece>) -> Void
   ) -> RawSyntax {
+    if let defaultText = kind.defaultText {
+      assert(text.isEmpty || kind.defaultText == text)
+    }
```
bnbarham, line 531: How would you feel about `assert(kind.defaultText == nil || text.isEmpty || kind.defaultText == text)`?

Sources/SwiftSyntax/gyb_generated/TokenKind.swift
```diff
@@ -810,131 +925,6 @@ public enum TokenKind: Hashable {
     case .yield: return "kw_yield"
     }
   }
-
-  var sourceLength: SourceLength {
```
bnbarham, line 814: Was this not used?

bnbarham (commented):
Cool stuff, the simplification of `consume*` is really nice! I'm not a huge fan of the huge switch for `SyntaxText` and then the conversion to `String` in `TokenKind`, but given we need it for `defaultText` in places ... 🤔.

CodeGeneration/Sources/generate-swiftparser/templates/DeclarationModifierFile.swift
```diff
@@ -24,20 +24,23 @@ let declarationModifierFile = SourceFile {
...
     for attribute in DECL_MODIFIER_KINDS {
-      EnumCaseDecl("case \(raw: attribute.swiftName) = \"\(raw: attribute.name)\"")
+      EnumCaseDecl("case \(raw: attribute.swiftName)")
     }
     InitializerDecl("init?(lexeme: Lexer.Lexeme)") {
-      SwitchStmt(switchKeyword: .switch, expression: Expr("lexeme.tokenKind")) {
-        for attribute in DECL_MODIFIER_KINDS where attribute.swiftName.hasSuffix("Keyword") {
-          SwitchCase("case .\(raw: attribute.swiftName):") {
-            SequenceExpr("self = .\(raw: attribute.swiftName)")
+      SwitchStmt(switchKeyword: .switch, expression: Expr("lexeme")) {
```
kimdv, line 32: Think we could remove the `.switch` keyword

gyb_syntax_support/Token.py
```diff
@@ -11,7 +11,7 @@ class Token(object):
     def __init__(self, name, kind, name_for_diagnostics,
                  unprefixed_kind=None, text=None, classification='None',
                  is_keyword=False, requires_leading_space=False,
-                 requires_trailing_space=False):
+                 requires_trailing_space=False, associated_value_class=None):
```
kimdv, line 14: Should we also add this to the swift child file

gyb_syntax_support/Child.py
```diff
@@ -130,6 +130,9 @@ def generate_default_initialization(self):
       if self.token.text:
         return f" = .{self.token.swift_kind()}Token()"
       if self.text_choices and len(self.text_choices) == 1:
-        return f" = .{self.token.swift_kind()}(\"{self.text_choices[0]}\")"
+        if self.token.associated_value_class:
+          return f" = .{self.token.swift_kind()}(.{self.text_choices[0]})"
+        else:
+          return f" = .{self.token.swift_kind()}(\"{self.text_choices[0]}\")"
```
kimdv, line 136: And this, is it also needed in the Child.swift

Sources/SwiftSyntax/Tokens.swift.gyb
```diff
@@ -56,6 +56,20 @@ extension TokenSyntax {
...
+    trailingTrivia: Trivia = [],
+    presence: SourcePresence = .present
+  ) -> TokenSyntax {
+    return TokenSyntax(
+      .${token.swift_kind()}(value),
+      leadingTrivia: leadingTrivia,
+      trailingTrivia: trailingTrivia,
+      presence: presence
+    )
+  }
```
kimdv, line 72: Note to self: Remember this in #1158

Sources/SwiftSyntax/TokenKind.swift.gyb
```diff
@@ -26,7 +26,9 @@ public enum TokenKind: Hashable {
 %
 %   # Tokens that don't have a set text have an associated value that
 %   # contains their text.
-%   if not token.text:
```
kimdv, line 29: Note to self: Remember in #1158

Sources/SwiftSyntax/SyntaxTraits.swift.gyb
```diff
@@ -29,7 +29,7 @@ public protocol ${trait.trait_name}Syntax: SyntaxProtocol {
 %       ret_type += '?'
 %   end
   var ${child.swift_name}: ${ret_type} { get }
-  func with${child.name}(_ newChild: ${child.type_name}?) -> Self
```
kimdv, line 32: Note to self: Remember in #1158

kimdv (commented):
Added some online questions and notes to my self

Sources/SwiftSyntax/SyntaxNodes.swift.gyb.template
```diff
@@ -233,12 +233,12 @@ public struct ${node.name}: ${base_type}Protocol, SyntaxHashable {
   /// Returns a copy of the receiver with its `${child.swift_name}` replaced.
   /// - param newChild: The new `${child.swift_name}` to replace the node's
   ///                   current `${child.swift_name}`, if present.
-  public func with${child.name}(_ newChild: ${child_type}?) -> ${node.name} {
+  public func with${child.name}(_ newChild: ${child_type}${'?' if child.is_optional else ''}) -> ${node.name} {
```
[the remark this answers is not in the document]
ahoppen: `make_missing_child` is still being used for SyntaxFactory so we can’t delete it just yet 😬

Sources/SwiftParser/Modifiers.swift
ahoppen: The nice thing about this approach is that it guarantees we have a handle if we match a kind without having to force-unwrap it (remember there’s a default case at the bottom that doesn’t have a handle).

Sources/SwiftParser/Types.swift
ahoppen: Thanks

Sources/SwiftParser/Types.swift
ahoppen: I agree. I’ll re-consider it after I have had some time to play around automatic generation of `RawTokenKindSubset` from `text_choices` tomorrow.

Sources/SwiftSyntax/Raw/RawSyntax.swift
ahoppen: Much better

Sources/SwiftSyntax/gyb_generated/TokenKind.swift
ahoppen: Nope

CodeGeneration/Sources/generate-swiftparser/templates/DeclarationModifierFile.swift
```diff
@@ -24,20 +24,23 @@ let declarationModifierFile = SourceFile {
...
+        for attribute in DECL_MODIFIER_KINDS {
+          if attribute.swiftName.hasSuffix("Keyword") {
```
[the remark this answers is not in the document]
ahoppen: Nice catch. They were indeed the same.

CodeGeneration/Sources/generate-swiftparser/templates/DeclarationModifierFile.swift
ahoppen: Thanks

gyb_syntax_support/Token.py
ahoppen: I’ll add it in a follow-up PR that also starts generating Keywords.swift using CodeGeneration: https://github.com/apple/swift-syntax/pull/1174

gyb_syntax_support/Child.py
ahoppen: Same, will be added by https://github.com/apple/swift-syntax/pull/1174

ahoppen:
@swift-ci Please test

Sources/SwiftParser/Modifiers.swift
bnbarham: There's no reason that has to be in the switch though:
```
guard let (kind, handle) = self.at(anyIn: DeclarationModifier.self) else {
  break MODIFIER_LOOP
}

switch kind {
...
}
```

Or am I missing something obvious?

ahoppen: No, that would also work. I think that’s just a matter of style now. If there were fewer cases in the `RawTokenKindSubset`, I would prefer a single `switch` instead of the `guard` + `switch` combination. And then I think it’s nice to always use the same style when you’re switching over `RawTokenKindSubset`s.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

Sources/SwiftParser/Modifiers.swift
bnbarham: Sure, it's definitely style. I'll win you over eventually!

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
https://github.com/apple/swift-stress-tester/pull/218
https://github.com/apple/swift-format/pull/464

@swift-ci Please test

ahoppen:
https://github.com/apple/swift-stress-tester/pull/218
https://github.com/apple/swift-format/pull/464

@swift-ci Please test

bnbarham (approved):

ahoppen:
https://github.com/apple/swift-stress-tester/pull/218
https://github.com/apple/swift-format/pull/464

@swift-ci Please test

PR #1180, opened 2023-01-03
Assert that tokens in the syntax tree match `token_choices` and `text_choices`

ahoppen:
We currently specify `token_choices` and `text_choices` in the Python files but don't enforce that the tokens in the syntax tree actually match these – effectively we are ignoring these properties.

Add assertions that the actual tokens in the syntax tree match the specification and fix a few cases where these assertions were violated.

rdar://103448567


Sources/SwiftSyntax/Raw/gyb_generated/RawSyntaxNodes.swift
```diff
@@ -8531,6 +8680,7 @@ public struct RawAccessorDeclSyntax: RawDeclSyntaxNodeProtocol {
     _ unexpectedAfterBody: RawUnexpectedNodesSyntax? = nil,
     arena: __shared SyntaxArena
   ) {
+    assert(accessorKind.tokenKind == .keyword(.get) || accessorKind.tokenKind == .keyword(.set) || accessorKind.tokenKind == .keyword(.didSet) || accessorKind.tokenKind == .keyword(.willSet) || accessorKind.tokenKind == .keyword(.unsafeAddress) || accessorKind.tokenKind == .keyword(.addressWithOwner) || accessorKind.tokenKind == .keyword(.addressWithNativeOwner) || accessorKind.tokenKind == .keyword(.unsafeMutableAddress) || accessorKind.tokenKind == .keyword(.mutableAddressWithOwner) || accessorKind.tokenKind == .keyword(.mutableAddressWithNativeOwner) || accessorKind.tokenKind == .keyword(._read) || accessorKind.tokenKind == .keyword(._modify), "Received \(accessorKind.tokenKind)")
```
bnbarham, line 8683: These should overlap with the subset right? Wonder if we should autogenerate subsets when there's > 1 kind. I imagine that would be a lot though, maybe only > 3 and then `at` can have overloads for up to 3 kinds or a subset?

Tests/SwiftParserTest/AttributeTests.swift
```diff
@@ -27,7 +27,7 @@ final class AttributeTests: XCTestCase {
         DiagnosticSpec(message: "expected ')' to end attribute", fixIts: ["insert ')'"]),
       ],
       fixedSource: """
-        @_dynamicReplacement(for: <#identifier#>)
+        @_dynamicReplacement(for : <#identifier#>)
```
bnbarham, line 30: What's the whitespace for?

ahoppen: Oh, I wanted to fix that. It’s because `for` is a keyword now and keywords always have trailing space in BasicFormat at the moment. Improving that is a slightly larger change in BasicFormat that I would like to do in a follow-up PR.

Sources/SwiftSyntax/Raw/gyb_generated/RawSyntaxNodes.swift
ahoppen: That’s the next item on my To-Do list after this PR is merge: “Generate RawTokenKindSubset”. I’m still not entirely sure how exactly we want to decide whether to generate the `RawTokenKindSubset`, but it’s definitely something we should do.

Tests/SwiftParserTest/AttributeTests.swift
ahoppen: Here we go: https://github.com/apple/swift-syntax/pull/1298

bnbarham (approved):

ahoppen:
@swift-ci Please test

PR #1185, opened 2023-01-04
Remove stringInterpolationAnchor

CodaFi:
This token has a sordid history in the legacy parser being used to represent both the start and end of a string interpolation segment

```
\(...)
^    ^
```

SwiftSyntax has always interpreted this as just the closing delimiter

```
\(...)
     ^
```

Which isn't correct. The new lexer does not ever lex for these, and the new parser does not re-lex for them either, so let's just drop it for now. If we want the structure of a string interpolation segment's delimiters, we'd need new modeling for it anyways.


CodaFi:
@swift-ci test

ahoppen (approved):

CodaFi:
@swift-ci test macOS

Issue #1184, opened 2023-01-04
Odd corruption when constructing new nodes


allevato:
### Description

I discovered this while playing with expression macros in a development snapshot, and was also able to reproduce it with a standalone binary built against the toolchain's host copy of SwiftSyntax. cc @DougGregor 

I was **not** able to reproduce this against SwiftSyntax `main`, so this is somehow specific to the toolchain's copy of the library. I browsed briefly through the recent PRs but nothing immediately stood out to me as possibly fixing this problem. It's odd since I'm using a trunk snapshot and not a 5.8 snapshot, so `main` should be almost identical.

I've uploaded a repro project here that has the right build settings to link against a trunk toolchain's SwiftSyntax: https://github.com/allevato/swift-syntax-corruption-repro

The tl;dr is that this produces an incorrect structure:

```swift
  let array = [1, 2, 3]
  var elements = [ArrayElementSyntax]()
  for element in array {
    // BUG: This causes "[3,3,3,]" to be printed below.
    elements.append(
      ArrayElementSyntax(
        expression: "\(literal: element)" as ExprSyntax,
        trailingComma: .commaToken()))
  }
  let arrayLiteral = ArrayExprSyntax(
    elements: ArrayElementListSyntax(elements))

  print(arrayLiteral)
```

while this one works correctly:

```swift
  let array = [1, 2, 3]
  var elements = [ArrayElementSyntax]()
  for element in array {
    // WORKAROUND: Hoisting the node out into its own local
    // variable causes "[1,2,3,]" to be printed below
    // (correct).
    let elementExpr = "\(literal: element)" as ExprSyntax
    elements.append(
      ArrayElementSyntax(
        expression: elementExpr,
        trailingComma: .commaToken()))
  }
  let arrayLiteral = ArrayExprSyntax(
    elements: ArrayElementListSyntax(elements))

  print(arrayLiteral)
```

Depending on what I was doing (in the macro case), I was also observing either bizarre crashes or infinite recursion when trying to print the node, so there seems to be some kind of memory or structural corruption occurring here.

ahoppen:
rdar://103878350

ahoppen:
Thanks for finding this. This was indeed a memory corruption bug. I’m fixing it in https://github.com/apple/swift-syntax/pull/1190.

PR #1190, opened 2023-01-04
Fix memory corruption when constructing syntax nodes from different arenas

ahoppen:
The initializers of the syntax nodes didn’t retain the parameters until the end of the initializer call. This means that their arenas could be deallocated before they were added as children to the newly created arena, causing memory corruption.

Fixes https://github.com/apple/swift-syntax/issues/1184 
Fixes rdar://103878350


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

DougGregor:
@swift-ci please test macOS

PR #1191, opened 2023-01-05
Refactor the lexer to store lexer errors in the tokens instead of having a diagnosticsHandler

ahoppen:
Instead of marking a token as having an error and relying on re-lexing to emit a diagnostic, store the type of error and the offset at which it occurred in the token itself.

This is a more sound design IMO and will continue to work even if we introduce state into the lexer (e.g. whether we are currently in a string literal or inside a conflict marker range).

### DIFF
--- a/Sources/SwiftParser/Lexer.swift
+++ b/Sources/SwiftParser/Lexer.swift
@@ -11,7 +11,6 @@
 //===----------------------------------------------------------------------===//
 
 @_spi(RawSyntax) import SwiftSyntax
-import SwiftDiagnostics
 
 /// A lexical analyzer for the Swift programming language.
 ///
@@ -33,7 +32,6 @@ public struct Lexer {
 
       public static let isAtStartOfLine = Flags(rawValue: 1 << 0)
       public static let isMultilineStringLiteral = Flags(rawValue: 1 << 1)
-      public static let isErroneous = Flags(rawValue: 1 << 2)
 
       public var debugDescription: String {
         var descriptionComponents: [String] = []
@@ -43,9 +41,6 @@ public struct Lexer {
         if self.contains(.isMultilineStringLiteral) {
           descriptionComponents.append("isMultilineStringLiteral")
         }
-        if self.contains(.isErroneous) {
-          descriptionComponents.append("isErroneous")
-        }
         return "[\(descriptionComponents.joined(separator: ", "))]"
       }
     }
@@ -53,6 +48,7 @@ public struct Lexer {
     @_spi(RawSyntax)
     public var tokenKind: RawTokenKind
     public var flags: Lexeme.Flags
+    public var error: LexerError?
     var start: UnsafePointer<UInt8>
     public var leadingTriviaByteLength: Int
     public var textByteLength: Int
@@ -75,6 +71,7 @@ public struct Lexer {
     public init(
       tokenKind: RawTokenKind,
       flags: Flags,
+      error: LexerError?,
       start: UnsafePointer<UInt8>,
       leadingTriviaLength: Int,
       textLength: Int,
@@ -82,6 +79,7 @@ public struct Lexer {
     ) {
       self.tokenKind = tokenKind
       self.flags = flags
+      self.error = error
       self.start = start
       self.leadingTriviaByteLength = leadingTriviaLength
       self.textByteLength = textLength
@@ -154,6 +152,7 @@ extension Lexer {
           self.nextToken = Lexeme(
             tokenKind: .eof,
             flags: [],
+            error: nil,
             start: self.cursor.pointer,
             leadingTriviaLength: 0,
             textLength: 0,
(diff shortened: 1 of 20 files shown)


ahoppen:
@swift-ci Please test

Sources/SwiftParser/Lexer.swift
```diff
@@ -1448,47 +1497,58 @@ extension Lexer.Cursor {
...
     // 0x[0-9a-fA-F][0-9a-fA-F_]*
     _ = self.advance()
-    guard !self.isAtEndOfFile, Unicode.Scalar(self.peek()).isHexDigit else {
-      if !self.isAtEndOfFile {
-        diagnosticHandler?(TokStart.distance(to: self), InvalidDigitInIntegerLiteral(kind: .hex(Unicode.Scalar(self.peek()))))
-      }
+    guard !self.isAtEndOfFile else {
       self.advance(while: { $0.isValidIdentifierContinuationCodePoint })
-      return (.integerLiteral, [.isErroneous])
+      return LexerResult(.integerLiteral)
```
CodaFi, line 1522: This appears to accept the literal `0x`

Sources/SwiftParser/Lexer.swift
```diff
@@ -768,22 +774,51 @@ extension Lexer.Cursor {
   }
 }
 
+struct LexerResult {
```
CodaFi, line 777: Can we make this an inner type `Lexer.Result`?

Sources/SwiftSyntax/LexerError.swift
```diff
@@ -0,0 +1,31 @@
...
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// If the token has a lexical error, this defines the type of the error.
+/// `lexerErrorOffset` in the token will specify at which offset the error
+/// occurred.
+public enum LexerError {
```
bnbarham, line 16: Any thoughts on something like:
```
public struct LexerError {
  public enum Kind {
    case ...
  }

  public let kind: Kind
  public let byteOffset: UInt16 // No need for alias, it's only defined in one place now

  public init(...) { ... }
}
```

And then just having the one `public var lexerError: LexerError` field and a
```
public var hasLexerError: Bool {
  return lexerError != nil
}
```
?

This would also simplify all the comments that mention "if there's an error, `lexerErrorOffset` will specify the offset".

As an aside, I'd personally be fine with `UInt8` as the offset type. Do people really have more than 255 characters for a token?

Sources/SwiftParser/Lexer.swift
ahoppen: Sure. I’m not sure how best to spell the return values though. Something like `return Result(.leftSquareBracket)` would be the most idiomatic in Swift but to me that looks like we’re returning the `Result` type from the stdlib. Thus I’ve gone with the fully-qualified `return Lexer.Result(.leftSquareBracket)` for now.

Sources/SwiftSyntax/LexerError.swift
ahoppen: That’s a much better idea. No idea why I modeled it the way I did…

> As an aside, I'd personally be fine with UInt8 as the offset type. Do people really have more than 255 characters for a token?

String literals are currently single tokens and those can reasonably be bigger than 255 characters, I think.

bnbarham: Oh, unfortunate.

Sources/SwiftSyntax/SyntaxOtherNodes.swift
```diff
@@ -169,6 +169,11 @@ public struct TokenSyntax: SyntaxProtocol, SyntaxHashable {
   public func childNameForDiagnostics(_ index: SyntaxChildrenIndex) -> String? {
     return nil
   }
+
+  /// If the token has a lexial error, the type of the error.
```
bnbarham, line 173:
```suggestion
  /// If the token has a lexical error, the type of the error.
```

Sources/SwiftParser/Lexer.swift
bnbarham: Could name it `LexedToken` or something, not sure that's much better though 🤷

Sources/SwiftParser/Lexer.swift
bnbarham: This one is still an issue too

Sources/SwiftSyntax/Raw/RawSyntax.swift
```diff
@@ -160,7 +161,7 @@ extension RawSyntax {
     switch view {
     case .token(let tokenView):
       var recursiveFlags: RecursiveRawSyntaxFlags = []
-      if tokenView.hasLexerError || tokenView.presence == .missing {
+      if tokenView.lexerError != nil || tokenView.presence == .missing {
```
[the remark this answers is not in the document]
ahoppen: I prefer to remove it.

Sources/SwiftParser/Lexer.swift
ahoppen: Since this is not a regression, I just opened https://github.com/apple/swift-syntax/issues/1214 so we can fix it in a follow-up PR.

Sources/SwiftParser/Lexer.swift
ahoppen: I don’t think that’s an improvement.

bnbarham: Yeah me neither :). I'm happy with what you changed it to.

bnbarham (approved):

ahoppen:
@swift-ci Please test

PR #1192, opened 2023-01-05
Produce separate tokens for raw string delimiters and string quotes in the lexer

ahoppen:
The eventual goal of this change is that we no longer need to re-lex string literals from the parser to separate them into their components. Instead, the lexer should just produce the lexemes that will later be put into the syntax tree as tokens.

The downside of this is that the lexer now needs to carry state and know whether it is lexing a string literal. On the upside, the string literal parser could be significantly simplified and the diagnostics got better without any further changes.

### DIFF
--- a/Sources/SwiftParser/Lexer.swift
+++ b/Sources/SwiftParser/Lexer.swift
@@ -31,15 +31,11 @@ public struct Lexer {
       }
 
       public static let isAtStartOfLine = Flags(rawValue: 1 << 0)
-      public static let isMultilineStringLiteral = Flags(rawValue: 1 << 1)
 
       public var debugDescription: String {
         var descriptionComponents: [String] = []
         if self.contains(.isAtStartOfLine) {
           descriptionComponents.append("isAtStartOfLine")
-        }
-        if self.contains(.isMultilineStringLiteral) {
-          descriptionComponents.append("isMultilineStringLiteral")
         }
         return "[\(descriptionComponents.joined(separator: ", "))]"
       }
@@ -56,11 +52,6 @@ public struct Lexer {
 
     var isAtStartOfLine: Bool {
       return self.flags.contains(.isAtStartOfLine)
-    }
-
-    var isMultilineStringLiteral: Bool {
-      assert(self.rawTokenKind == .stringLiteral)
-      return self.flags.contains(.isMultilineStringLiteral)
     }
 
     var isEditorPlaceholder: Bool {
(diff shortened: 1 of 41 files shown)


Sources/SwiftParser/Lexer.swift
```diff
@@ -1104,11 +1240,10 @@ extension Lexer.Cursor {
           // The only case we reach here is unterminated single line string in the
           // interpolation. For better recovery, go on after emitting an error.
           //          diagnose(CurPtr, diag::lex_unterminated_string)
-          wasErroneous = true
           continue
         } else {
           //          diagnose(TokStart, diag::lex_unterminated_string)
-          return (.unknown, [])
+          return LexerResult(.stringLiteralContents, newMode: .normal)
```
CodaFi, line 1246: This change is dangerous without possibly corresponding changes to string literal contents parsing. We need to make sure that its scanning and indexing logic can tolerate an early eof and strings like `"\(xyz \v"`

ahoppen: It seems to be working… At least the existing test cases didn't catch anything. I haven't banged it harder than that though

Tests/SwiftParserTest/translated/PoundAssertTests.swift
```diff
@@ -18,8 +18,12 @@ final class PoundAssertTests: XCTestCase {
   func testPoundAssert1() {
     AssertParse(
       """
-      #assert(true, 123)
-      """
+      #assert(true, 1️⃣123)
+      """,
+      diagnostics: [
+        DiagnosticSpec(message: "expected string literal in '#assert' directive"),
+        DiagnosticSpec(message: "unexpected code '123' in '#assert' directive"),
```
bnbarham, line 25: Two diagnostics here is a little weird. Maybe we should just eat the next unexpected after a missing token and then have the fix-it remove the unexpected? I haven't checked whether that would make other cases worse though.

Tests/SwiftParserTest/translated/RawStringErrorsTests.swift
```diff
@@ -21,10 +21,7 @@ final class RawStringErrorsTests: XCTestCase {
       let _ = "foo\(#"bar"#1️⃣#)baz"
       """###,
       diagnostics: [
-        // TODO: Old parser expected error on line 1: too many '#' characters in closing delimiter
-        // TODO: Old parser expected error on line 1: expected ',' separator
-        // TODO: Old parser expected error on line 1: expected expression in list of expressions
-        DiagnosticSpec(message: "unexpected code '#' in string literal")
+        DiagnosticSpec(message: "too many '#' characters in closing delimiter")
```
bnbarham, line 24: Way better!

Tests/SwiftParserTest/translated/RawStringErrorsTests.swift
```diff
@@ -73,16 +67,13 @@ final class RawStringErrorsTests: XCTestCase {
   func testRawStringErrors5() {
     AssertParse(
       #####"""
-      let _ = ###"invalid"###1️⃣#2️⃣#3️⃣#4️⃣
+      let _ = ###"invalid"###1️⃣###
```
bnbarham, line 70: This is a duplicate of 3 isn't it?

Tests/SwiftParserTest/translated/StringLiteralEofTests.swift
```diff
@@ -123,18 +130,14 @@ final class StringLiteralEofTests: XCTestCase {
...
-        DiagnosticSpec(message: #"unexpected code '"bar' in string literal"#),
-        DiagnosticSpec(locationMarker: "2️⃣", message: #"expected ')' in string literal"#),
-        DiagnosticSpec(locationMarker: "3️⃣", message: #"unexpected code '' in string literal"#),
-        DiagnosticSpec(locationMarker: "4️⃣", message: #"expected '"""' to end string literal"#),
-        // TODO: Old parser expected error on line 3: cannot find ')' to match opening '(' in string interpolation
-        // TODO: Old parser expected error on line 3: unterminated string literal
+        DiagnosticSpec(locationMarker: "1️⃣", message: #"expected '"' to end string literal"#),
+        DiagnosticSpec(locationMarker: "1️⃣", message: "unexpected code in string literal"),
+        DiagnosticSpec(locationMarker: "2️⃣", message: "expected ')' in string literal"),
+        DiagnosticSpec(locationMarker: "3️⃣", message: #"expected '"""' to end string literal"#),
```
bnbarham, line 140: Bit of a weird case I suppose, but it's a little strange that we don't treat this as `bar` being part of the string within the interpolation. Though having a string within an interpolation seems fairly unusual anyway, so... maybe doesn't actually matter.

I think ideally we'd have just two diagnostics: one for the missing `")` (or two, whatever) and another for the missing `"""`.

Tests/SwiftParserTest/translated/UnclosedStringInterpolationTests.swift
```diff
@@ -26,99 +26,90 @@ final class UnclosedStringInterpolationTests: XCTestCase {
...
+      _ = "mid == \(pete1️⃣"2️⃣
       """##,
       diagnostics: [
-        // TODO: Old parser expected error on line 1: cannot find ')' to match opening '(' in string interpolation
-        // TODO: Old parser expected error on line 1: unterminated string literal
-        DiagnosticSpec(message: "expected expression"),
-        DiagnosticSpec(message: #"extraneous code '"mid == \(pete"' at top level"#),
+        DiagnosticSpec(locationMarker: "1️⃣", message: #"unexpected code '"' in string literal"#),
+        DiagnosticSpec(locationMarker: "2️⃣", message: "expected ')' in string literal"),
+        DiagnosticSpec(locationMarker: "2️⃣", message: #"expected '"' to end string literal"#),
```
bnbarham, line 34: It'd be nice to handle this one better, ie. don't eat the `"` as unexpected.  That makes 7 worse, but IMO this would be by far the more common case.

Tests/SwiftParserTest/translated/UnclosedStringInterpolationTests.swift
```diff
@@ -26,99 +26,90 @@ final class UnclosedStringInterpolationTests: XCTestCase {
...
+      let theGoat = "kanye \("1️⃣
       """##,
       diagnostics: [
-        // TODO: Old parser expected error on line 1: cannot find ')' to match opening '(' in string interpolation
-        // TODO: Old parser expected error on line 1: unterminated string literal
-        DiagnosticSpec(message: "expected expression in variable"),
-        DiagnosticSpec(message: #"extraneous code '"kanye \("' at top level"#),
+        DiagnosticSpec(message: #"expected '"' to end string literal"#),
+        DiagnosticSpec(message: "expected ')' in string literal"),
+        DiagnosticSpec(message: #"expected '"' to end string literal"#),
```
bnbarham, line 47: Same as 2, though honestly I'm not sure it's worth keeping both. That's true for most of the `testUncloseStringInterpolation*` TBH.

Sources/SwiftParser/Lexer.swift
```diff
@@ -492,6 +529,58 @@ extension Lexer.Cursor {
     return false
   }
 
+  mutating func lexStringQuote() -> LexerResult {
+    func newMode(currentMode: LexerCursorMode, kind: StringLiteralKind) -> LexerCursorMode {
```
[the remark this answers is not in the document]
ahoppen: Good idea 👍 I renamed it to `state`.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

bnbarham (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test macOS

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

PR #1167, opened 2022-12-22
[Variadic Generics] Change pack expansion types and expressions to use the `repeat` syntax.

hborla:
We're explaining patterned pack expansion as using _repetition patterns_. Given a concrete pack substitution, the pattern is repeated for each element in the substituted pack. This syntax expresses this idea directly using the `repeat` keyword, and `each` contextual keywords applied to pack references:

```swift
func tuplify<T...>(_ t: repeat each T) -> (repeat each T) {
  return (repeat each t)
}
```

The contextual `each` keyword was added for types and expressions in https://github.com/apple/swift-syntax/pull/1134 and https://github.com/apple/swift-syntax/pull/1146, respectively. This change replaces postfix ellipsis with a prefix `repeat` keyword for pack expansion types and expressions.

For now, pack expansions with closure patterns must wrap the closure in parenthesis. I'll investigate adding more sophisticated lookahead in a follow-up PR.

PR #1193, opened 2023-01-05
Generate SyntaxAnyVisitor.swift

ahoppen:
SyntaxAnyVisitor.swift was missing from the template list in GenerateSwiftSyntax.


ahoppen:
@swift-ci Please test

kimdv (approved):

kimdv:
Should we do something like gyb, where we delete files before generating ?

To avoid this?

hborla:
Thank you! Merging this to unblock https://github.com/apple/swift-syntax/pull/1167

PR #1195, opened 2023-01-06
[Macros] Introduce (Freestanding)DeclarationMacro and implement expansion

DougGregor:
Freestanding declaration macros are spelled in the same manner as expression macros, with a leading `#`. Introduce a protocol with their expansion operation and implement expansion semantics in the macro system for testing purposes.

Rework the `myError` macro used for testing to be a freestanding declaration macro, which better matches the actual semantics of `#error`.


DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

PR #1196, opened 2023-01-07
[Macros] Implement peer declaration macros with expansion

DougGregor:
Peer declaration macros are applied to a declaration via an attribute,
but are responsible for generating "peer" declarations that go
alongside the declaration to which the attribute is attached.
Implement the `addCompletionHandler` example, which takes:

    @addCompletionHandler
    func f(a: Int, for b: String, _ value: Double) async -> String { }

and adds the following peer declaration:

    func f(a: Int, for b: String, _ value: Double, completionHandler: (String) -> Void) {
      Task {
        completionHandler(await f(a: a, for: b, value))
      }
    }

### DIFF
--- a/Sources/_SwiftSyntaxMacros/ExpressionMacro.swift
+++ b/Sources/_SwiftSyntaxMacros/ExpressionMacro.swift
@@ -12,7 +12,6 @@
 
 import SwiftSyntax
 import SwiftParser
-import SwiftDiagnostics
 
 /// Describes a macro that is explicitly expanded as an expression.
 public protocol ExpressionMacro: Macro {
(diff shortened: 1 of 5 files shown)


DougGregor:
@swift-ci please test

Sources/_SwiftSyntaxMacros/ExpressionMacro.swift
```diff
@@ -12,7 +12,6 @@
 
 import SwiftSyntax
 import SwiftParser
```
rxwei, line 14: Looks like SwiftParser isn't needed here either

DougGregor: You're right, the macro protocols have a bunch of extra imports. I've cleaned them up locally.

DougGregor:
@swift-ci please test macOS

DougGregor:
@swift-ci please test macOS

DougGregor:
@swift-ci please test macOS

PR #1201, opened 2023-01-07
Move SyntaxVisitor from gyb to codegen

kimdv:
Part of https://github.com/apple/swift-syntax/pull/1158


CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxVisitorFile.swift
```diff
@@ -0,0 +1,139 @@
...
+    EnumCaseDecl("""
+      /// The visitor should visit the descendents of the current node.
+      case visitChildren
+      """)
+    
+    EnumCaseDecl("""
+      /// The visitor should avoid visiting the descendents of the current node.
+      case skipChildren
+      """)
+  }
```
ahoppen, line 33: I think this could be simplified to 
```suggestion
  EnumDecl("""
    /// The enum describes how the SyntaxVistor should continue after visiting
    /// the current node.
    public enum SyntaxVisitorContinueKind {
      /// The visitor should visit the descendents of the current node.
      case visitChildren
      /// The visitor should avoid visiting the descendents of the current node.
      case skipChildren
    }
    """)
```

ahoppen (approved):
One minor suggestion, otherwise LGTM.

kimdv:
@swift-ci please test

PR #1208, opened 2023-01-10
Generalize macro expansion for macros written as attributes

DougGregor:
Add 'attributed' trait for nodes that can have attributes, and use it to generalize macro expansion for macros written as attributes.


DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

ahoppen (approved):

PR #1217, opened 2023-01-11
Update swift syntax

kimdv:

PR #1218, opened 2023-01-11
Fix spacings around `!`and `?`

kimdv:
Fixes some spacing regressions found in #1217


kimdv:
@swift-ci please test

ahoppen (commented):
This makes me wonder whether the better default would be to mark `exclamationMark` and `postfixQuestionMark` as not having a trailing space and selectively enabling it in cases where we want a trailing space.

I’m thinking that in the most cases the exclamation and question mark will be appended to an identifier and since identifiers don’t have a trailing space, `!` and `?` also shouldn’t have a trailing space. I might be wrong though.

kimdv:
I can see what happens 😬

kimdv:
@swift-ci please test

CodeGeneration/Sources/generate-swiftsyntax/templates/basicformat/BasicFormatFile.swift
```diff
@@ -157,18 +157,16 @@ let basicFormatFile = SourceFile {
...
-             (.postfixQuestionMark, .rightAngle), // Ensures there is not space in `ContiguousArray<RawSyntax?>`
-             (.postfixQuestionMark, .rightParen), // Ensures there is not space in `myOptionalClosure?()`
              (.tryKeyword, .exclamationMark), // Ensures there is not space in `try!`
              (.tryKeyword, .postfixQuestionMark): // Ensures there is not space in `try?`
           return false
         case (.spacedBinaryOperator, .comma):
           return false
+        case (.exclamationMark, .identifier), // Ensures there is not space in `try! foo`
+             (.postfixQuestionMark, .identifier): // Ensures there is not space in `try? foo`
+          return true
```
ahoppen, line 169: Wouldn’t it make more sense to do 
```swift
switch (token.kind, token.parent?.kind) {
  case (.colon, .dictionaryExpr): // Covers the case at the top of this function
    return false
  case (.exclamationMark, .tryExpr), // Ensures there is a space in `try! foo`
       (.postfixQuestionMark, .tryExpr): // Ensures there is a space in `try? foo`
    return true
}
```

That should also cover cases like `try! (foo())` or `try! !functionThatThrows()`.

ahoppen (commented):
I like this much better.

kimdv:
@swift-ci please test

CodeGeneration/Sources/generate-swiftsyntax/templates/basicformat/BasicFormatFile.swift
ahoppen: Yes, exactly. Looks good!

ahoppen (approved):

PR #1219, opened 2023-01-11
Save 2 bytes in `RawSyntaxData.ParsedToken`

ahoppen:
Introduction of `LexerError` incrased the `ParsedToken` size by 2 bytes compared to storing the kind and offset separately. `MemoryLayoutTest` didn’t fail because we only run them for debug builds but CI only creates release builds.

Reclaim those two bytes by storing kind and offset  as two fields in `RawSyntaxData.ParsedToken` again and transparently mapping them to a `LexerError` on access.

### DIFF
--- a/Sources/SwiftSyntax/Raw/RawSyntax.swift
+++ b/Sources/SwiftSyntax/Raw/RawSyntax.swift
@@ -55,8 +55,28 @@ internal struct RawSyntaxData {
 
     var presence: SourcePresence
 
-    /// If the token has a lexial error, the type of the error.
-    var lexerError: LexerError?
+    /// Store the members of `LexerError` individually so the compiler can pack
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

rintaro:
Comment on https://github.com/apple/swift-syntax/pull/1191/files
https://github.com/apple/swift-syntax/blob/8e6ae6eef6259c1f50692b509f776e66564ae189/Sources/SwiftSyntax/LexerError.swift#L36-L38
What would we do for lexer errors in leading trivia? Maybe this should be "relative to the tokens `position`"?

rintaro (commented):
I'm OK with this change, but as long as the size/stride is smaller or equal to other `RawSyntaxData` payloads (i.e.  `Layout` and `MaterializedToken`) there's no performance benefits.

PR #1220, opened 2023-01-11
[Macros] Introduce member attribute macros.

hborla:
For each member of the declaration the macro attribute is attached to, the macro can produce a set of attributes to attach to that member. For example:

```swift
@declaration(memberAttribute)
macro wrapStoredProperties: Void

@wrapStoredProperties
struct Point {
  var x: Int
  var y: Int

  var description: String { "" }
}

// Expanded to

struct Point {
  @Wrapper var x: Int
  @Wrapper var y: Int

  var description: String { "" }
}
```

The member attribute macro expansion is given the declaration it's attached to along with a member declaration, and it returns a set of attributes to attach to the member. The expansion method is called for each member in the `attachedTo` declaration.

### DIFF
--- a/Sources/_SwiftSyntaxMacros/CMakeLists.txt
+++ b/Sources/_SwiftSyntaxMacros/CMakeLists.txt
@@ -14,6 +14,7 @@ add_swift_host_library(_SwiftSyntaxMacros
   Macro.swift
   MacroExpansionContext.swift
   MacroSystem.swift
+  MemberAttributeMacro.swift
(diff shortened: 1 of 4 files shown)


hborla:
@swift-ci please test

DougGregor (approved):

hborla:
@swift-ci please test

hborla:
@swift-ci please test

hborla:
@swift-ci please test

PR #1223, opened 2023-01-11
Fix order of of entries in CODEOWNERS

ahoppen:
The last item has *most* precedence not *least*. Entries should be ordered the other way.


ahoppen:
@swift-ci Please test

DougGregor (approved):

ahoppen:
@swift-ci Please test macOS

PR #1224, opened 2023-01-12
Remap identifiers to keywords in `consume(ifAny:)`

ahoppen:
I don’t have an example that exercises this at the moment but if we are expecting a contextual keyword, we should be remapping it to a keyword in `consume(ifAny:)`

Also fix `debugInitCall` producing invalid code for keyword tokens that now have an associated value.

### DIFF
--- a/Sources/SwiftParser/TokenConsumer.swift
+++ b/Sources/SwiftParser/TokenConsumer.swift
@@ -181,8 +181,17 @@ extension TokenConsumer {
     ifAny kinds: [RawTokenKind],
     allowTokenAtStartOfLine: Bool = true
   ) -> Token? {
-    if self.at(any: kinds, allowTokenAtStartOfLine: allowTokenAtStartOfLine) {
-      return self.consumeAnyToken()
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

Sources/SwiftParser/TokenConsumer.swift
```diff
@@ -175,8 +175,14 @@ extension TokenConsumer {
     ifAny kinds: [RawTokenKind],
     where condition: (Lexer.Lexeme) -> Bool = { _ in true }
   ) -> Token? {
-    if self.at(any: kinds, where: condition) {
-      return self.consumeAnyToken()
+    for kind in kinds {
+      if case RawTokenKindMatch(kind) = self.currentToken {
```
bnbarham, line 179: Still need to check the condition

ahoppen: Good catch. Thank you

ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #1225, opened 2023-01-12
[Macros] Implement a type wrapper transformation as a macro.

hborla:
The type wrapper macro struct conforms to `MemberDeclarationMacro`, `MemberAttributeMacro`, and `AccessorDeclarationMacro`. Each macro capability provides a different part of the type wrapper transform:

* `MemberDeclarationMacro` adds the backing `var _storage` variable.
* `MemberAttributeMacro` applies macro attributes to each stored property inside the wrapped type, which are recursively expanded.
* `AccessorDeclarationMacro` uses the macros applied to stored properties to add `get` and `set` accessors, turning those stored properties into computed properties that indirect access through `_storage`.

### DIFF
--- a/Tests/SwiftSyntaxMacrosTest/MacroSystemTests.swift
+++ b/Tests/SwiftSyntaxMacrosTest/MacroSystemTests.swift
@@ -468,6 +468,74 @@ public struct WrapStoredProperties: MemberAttributeMacro {
   }
 }
 
+struct CustomTypeWrapperMacro {}
+
+extension CustomTypeWrapperMacro: MemberDeclarationMacro {
+  static func expansion(
+    of node: AttributeSyntax,
+    attachedTo declaration: DeclSyntax,
+    in context: inout MacroExpansionContext
+  ) throws -> [DeclSyntax] {
+    return [
+      """
+
+        var _storage: Wrapper<Self>
+      """
+    ]
+  }
+}
+
+extension CustomTypeWrapperMacro: MemberAttributeMacro {
+  static func expansion(
+    of node: AttributeSyntax,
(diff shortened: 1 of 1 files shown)


Tests/SwiftSyntaxMacrosTest/MacroSystemTests.swift
```diff
@@ -467,6 +467,74 @@ public struct WrapStoredProperties: MemberAttributeMacro {
...
+      return []
+    }
+
+    if identifier.text == "_storage" { return [] }
+
+    return [
+      """
+
+          get {
+            _storage[wrappedKeyPath: \\.\(identifier)]
```
xedin, line 525: Just FYI that type wrapper is a bit more then just `wrappedKeyPath:`, it also detects whether it’s attached to a reference type and if so, adds `wrappedSelf:` argument and always `storageKeyPath:` as the last argument which points to `\$Storage.<prop-name>`.

hborla: I know, this is just a proof of concept. The macro can generate whatever it wants in the accessors. For example, if the macro is attached to an actor or class, it could apply a different macro attribute to each of the stored properties to direct the recursive expansion to generate special accessors in that case. It could also just choose to put whatever code is in the subscript implementation directly into the accessors

xedin: Sounds good!

hborla:
@swift-ci please test

hborla:
@swift-ci please test

hborla:
@swift-ci please test

PR #1227, opened 2023-01-13
Refactor the lexer to make it easier to understand, maintain and more swifty

ahoppen:
Honestly, I don’t know how best to review this. Essentially, I’ve gone through the entire lexer (without string lexing because I want to refactor that anyway) and made the following changes 
- `peek` is now safe by default and returns `nil` if we reached the end of the file. There is `peek(matches:)` and `peek(doesntMatch:)` to check if the current token is one of the expected
- Split the lexer into multiple files
- Re-arranges some methods to be grouped more logically 
- All the `lex*` methods now expect to be placed at the first character they are lexing (previously they expected the first character to already be lexed). This means we could get rid of the `tokStart` parameter. Also I think it’s more natural this way
- Added a few more assertions
- Made all variables lowercase
- Added some documentation
- Enabled argument labels
- Plus probably a few miscellaneous more that I forgot

There should be no functionality change in this PR.

### DIFF
--- a/Sources/SwiftParser/Lexer.swift
+++ b/Sources/SwiftParser/Lexer.swift
@@ -243,7 +243,7 @@ public enum LexerCursorState {
     case .inStringLiteral(kind: let kind, delimiterLength: _):
       switch kind {
       case .singleLine, .singleQuote:
-        return !cursor.isAtEndOfFile && (cursor.peek() == UInt8(ascii: "\r") || cursor.peek() == UInt8(ascii: "\n"))
+        return cursor.peek(matches: "\r", "\n")
       case .multiLine: return false
       }
     case .afterStringLiteral: return false
@@ -288,13 +288,6 @@ extension Lexer {
       return self.pointer.distance(to: other.pointer)
     }
 
-    func peek(at offset: Int = 0) -> UInt8 {
-      assert(!self.isAtEndOfFile)
-      assert(offset >= 0)
-      assert(offset < self.input.count)
-      return self.input[offset]
-    }
-
     var isAtEndOfFile: Bool {
       return self.input.isEmpty
     }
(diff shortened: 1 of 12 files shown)


ahoppen:
@swift-ci Please test

Sources/SwiftParser/Lexer/Cursor.swift
```diff
@@ -331,11 +331,35 @@ extension Lexer.Cursor {
...
+      return true
+    } else {
+      return false
+    }
+  }
+
+  /// If the current character is in `matching`, advance the cursor and return `true`.
+  /// Otherwise, this is a no-op and returns `false`.
+  @_disfavoredOverload  // favor the stamped out copies
+  mutating func advance(matching characters: CharacterByte...) -> Bool {
```
bnbarham, line 356: What's the max number of characters we pass here (and similar for the peek case)? I'd prefer to just not have this one at all, I don't think we ever want to pay the cost of the ARC traffic.

Sources/SwiftParser/Lexer.swift
```diff
@@ -310,6 +303,91 @@ extension Lexer {
   }
 }
 
+// MARK: - Peeking
+
+extension Lexer.Cursor {
+  func peek(at offset: Int = 0) -> UInt8? {
+    assert(offset >= 0)
+    guard offset < self.input.count else {
```
bnbarham, line 219: It would be good to check the performance cost of running this on every single peek, ie. do we still need an unsafe version for when we know we're not at EOF?

Sources/SwiftParser/Lexer.swift
```diff
@@ -310,6 +303,91 @@ extension Lexer {
...
+    init(integerLiteral value: UInt8) {
+      self.value = value
+    }
+  }
+
+  // MARK: Positive matches
+
+  /// Stamped out copy of `peek(at:matches:)` in case we are matching a single
+  /// character to avoid construction of an array.
+  func peek(at offset: Int = 0, matches character: PeekMatch) -> Bool {
```
bnbarham: `advance` has `matching` which IMO reads better. You could also make `doesntMatch` `unless` instead. It sort of sounds like a condition, but not as much as `if` and `when` would for the positive case 🤷.

Sources/SwiftParser/Lexer/Cursor.swift
ahoppen: It’s three. I just stamped out three copies.

Sources/SwiftParser/Lexer.swift
ahoppen: Discussed offline to rename this to `is(at:)`

Sources/SwiftParser/Lexer.swift
ahoppen: Performance measurements of `testSelfParsePerformance` (release builds):

- Current `main`: 6.061 s
- With https://github.com/apple/swift-syntax/pull/1192: 6.285 (+3.7%)
- This PR: 6.330 (+ 0.7%)

I think this is acceptable given the added safety this gives us.

<details>
<summary>Raw data</summary>

```
main

Test Case '-[SwiftParserTest.ParserTests testSelfParsePerformance]' measured [Time, seconds] average: 6.062, relative standard deviation: 1.030%, values: [6.188462, 6.127593, 6.119602, 6.032848, 6.075716, 6.015036, 5.986559, 6.063117, 6.017368, 5.995290], performanceMetricID:com.apple.XCTPerformanceMetric_WallClockTime, baselineName: "", baselineAverage: , polarity: prefers smaller, maxPercentRegression: 10.000%, maxPercentRelativeStandardDeviation: 10.000%, maxRegression: 0.100, maxStandardDeviation: 0.100
Test Case '-[SwiftParserTest.ParserTests testSelfParsePerformance]' measured [Time, seconds] average: 6.061, relative standard deviation: 0.619%, values: [6.136058, 6.019346, 6.036355, 6.052679, 6.034555, 6.046183, 6.101317, 6.109356, 6.043739, 6.031751], performanceMetricID:com.apple.XCTPerformanceMetric_WallClockTime, baselineName: "", baselineAverage: , polarity: prefers smaller, maxPercentRegression: 10.000%, maxPercentRelativeStandardDeviation: 10.000%, maxRegression: 0.100, maxStandardDeviation: 0.100


lex-string-delimiters

Test Case '-[SwiftParserTest.ParserTests testSelfParsePerformance]' measured [Time, seconds] average: 6.298, relative standard deviation: 0.360%, values: [6.314035, 6.311459, 6.294189, 6.284121, 6.263693, 6.275958, 6.329505, 6.316271, 6.323488, 6.268774], performanceMetricID:com.apple.XCTPerformanceMetric_WallClockTime, baselineName: "", baselineAverage: , polarity: prefers smaller, maxPercentRegression: 10.000%, maxPercentRelativeStandardDeviation: 10.000%, maxRegression: 0.100, maxStandardDeviation: 0.100
Test Case '-[SwiftParserTest.ParserTests testSelfParsePerformance]' measured [Time, seconds] average: 6.272, relative standard deviation: 0.664%, values: [6.369508, 6.263280, 6.288378, 6.267686, 6.229722, 6.224330, 6.228591, 6.259549, 6.308582, 6.278303], performanceMetricID:com.apple.XCTPerformanceMetric_WallClockTime, baselineName: "", baselineAverage: , polarity: prefers smaller, maxPercentRegression: 10.000%, maxPercentRelativeStandardDeviation: 10.000%, maxRegression: 0.100, maxStandardDeviation: 0.100


This PR

Test Case '-[SwiftParserTest.ParserTests testSelfParsePerformance]' measured [Time, seconds] average: 6.340, relative standard deviation: 0.672%, values: [6.456374, 6.327926, 6.361954, 6.321446, 6.342133, 6.335581, 6.328410, 6.287928, 6.318708, 6.322288], performanceMetricID:com.apple.XCTPerformanceMetric_WallClockTime, baselineName: "", baselineAverage
Test Case '-[SwiftParserTest.ParserTests testSelfParsePerformance]' measured [Time, seconds] average: 6.319, relative standard deviation: 0.740%, values: [6.443643, 6.314616, 6.300575, 6.312995, 6.347714, 6.276697, 6.307492, 6.329889, 6.280654, 6.279295], performanceMetricID:com.apple.XCTPerformanceMetric_WallClockTime, baselineName: "", baselineAverage: , polarity: prefers smaller, maxPercentRegression: 10.000%, maxPercentRelativeStandardDeviation: 10.000%, maxRegression: 0.100, maxStandardDeviation: 0.100
```

</details>

bnbarham: I'm somewhat surprised by the ~4% fro the delimiters split. Any idea why that's so high?

ahoppen: I just measured again using Instruments (and a command line tool that contains the code from `testSelfParsePerformance` instead of `testSelfParsePerformance` itself) and couldn’t find a significant difference between any of the branches. Not sure what introduced the variation before.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

PR #1228, opened 2023-01-13
Introduce if/switch expressions

hamishknight:
Introduce IfExprSyntax and SwitchExpr syntax, which replace IfStmtSyntax and SwitchStmtSyntax, and can represent both if/switch expressions, as well as regular if/switch statements when wrapped in an ExpressionStmtSyntax. Implement the parsing logic such that we start parsing IfExprSyntax and SwitchExprSyntax in expression position.


hamishknight:
https://github.com/apple/swift/pull/63022
https://github.com/apple/sourcekit-lsp/pull/672
https://github.com/apple/swift-format/pull/452
https://github.com/apple/swift-stress-tester/pull/213

@swift-ci please test

hamishknight:
https://github.com/apple/swift/pull/63022
https://github.com/apple/sourcekit-lsp/pull/672
https://github.com/apple/swift-format/pull/452
https://github.com/apple/swift-stress-tester/pull/213

@swift-ci please test

Tests/SwiftRefactorTest/MigrateToNewIfLetSyntax.swift
```diff
@@ -17,102 +17,113 @@ import SwiftSyntaxBuilder
 import XCTest
 import _SwiftSyntaxTestSupport
 
+@discardableResult
+func testRefactorIfLet(
+  _ syntax: ExprSyntax,
+  expected: ExprSyntax,
+  file: StaticString = #file,
+  line: UInt = #line
+) throws -> ExprSyntax {
```
ahoppen, line 26: I think it would be cleaner to name this `AssertRefactorIfLet` and not have it return a value. Since we are asserting that `refactored` is `expected` callers don’t actually need the return value and can use whatever they are passing to `expected` instead.

Tests/SwiftParserTest/ExpressionTests.swift
```diff
@@ -897,3 +897,154 @@ final class MemberExprTests: XCTestCase {
     }
   }
 }
+
+final class StatementExpressionTests: XCTestCase {
+  func testIfExprInCoercion() {
```
ahoppen, line 902: Do you think it would make sense to add a few `substructure` assertions here? It seems like we could easily get the structure wrong here in some cases (although I didn’t look too closely). I’ll leave it up to you to decide where substructure assertions might make sense.

Tests/SwiftParserTest/ExpressionTests.swift
```diff
@@ -897,3 +897,154 @@ final class MemberExprTests: XCTestCase {
...
+    AssertParse(
+      """
+      func foo() {
```
ahoppen, line 905: I don’t think you actually need to have wrap the expression in a function declaration in all these test cases. Top-level should also allow expressions.

Sources/SwiftParser/TopLevel.swift
```diff
@@ -189,6 +189,35 @@ extension Parser {
...
+  private mutating func parseStatementItem() -> RawCodeBlockItemSyntax.Item {
+    let stmt = self.parseStatement()
+
+    // Special case: An 'if' or 'switch' statement followed by an 'as' must
+    // be an if/switch expression in a coercion.
+    // We could also achieve this by more eagerly attempting to parse an 'if'
+    // or 'switch' as an expression when in statement position, but that
+    // could result in less useful recovery behavior.
+    if at(.keyword(.as)), let stmtExpr = stmt.as(RawExpressionStmtSyntax.self) {
+      let expr = stmtExpr.expression
```
ahoppen, line 201: Can be shortened slightly

```suggestion
    if at(.keyword(.as)), let expr = stmt.as(RawExpressionStmtSyntax.self).expression {
```

Sources/SwiftParser/TopLevel.swift
```diff
@@ -189,6 +189,35 @@ extension Parser {
...
   }
 
+  private mutating func parseStatementItem() -> RawCodeBlockItemSyntax.Item {
+    let stmt = self.parseStatement()
+
+    // Special case: An 'if' or 'switch' statement followed by an 'as' must
+    // be an if/switch expression in a coercion.
+    // We could also achieve this by more eagerly attempting to parse an 'if'
+    // or 'switch' as an expression when in statement position, but that
+    // could result in less useful recovery behavior.
```
ahoppen, line 199: Wouldn’t this also apply to `is`. Something stupid like

```swift
if true { 1 } else { 2 } is Int
```

Or is `as` the only thing where we allow a postfix expression after an `if` expression?

Sources/SwiftParser/RawTokenKindSubset.swift
```diff
@@ -612,6 +612,26 @@ enum AwaitTryMove: RawTokenKindSubset {
...
 }
 
+enum IfOrSwitch: RawTokenKindSubset {
+  case ifKeyword
+  case switchKeyword
+
+  init?(lexeme: Lexer.Lexeme) {
+    switch lexeme {
+    case RawTokenKindMatch(.keyword(.if)): self = .ifKeyword
+    case RawTokenKindMatch(.keyword(.switch)): self = .switchKeyword
```
ahoppen, line 622: This can be slightly shorter

```suggestion
    case RawTokenKindMatch(.if): self = .ifKeyword
    case RawTokenKindMatch(.switch): self = .switchKeyword
```

Sources/SwiftParser/Expressions.swift
```diff
@@ -18,6 +18,11 @@ extension TokenConsumer {
     case (.awaitTryMove, let handle)?:
       var backtrack = self.lookahead()
       backtrack.eat(handle)
+
+      // These can be parsed as expressions with try/await.
+      if backtrack.at(anyIn: IfOrSwitch.self) != nil {
+        return true
+      }
```
ahoppen, line 25: Am I missing something or does `atStartOfExpression` return `false` for `if` expressions?

Sources/SwiftParser/Expressions.swift
```diff
@@ -510,6 +527,21 @@ extension Parser {
...
+      return RawExprSyntax(
+        parseSwitchExpression(switchHandle: .constant(.keyword(.switch)))
+      )
+    }
+    if self.at(.keyword(.if)) {
+      return RawExprSyntax(
+        parseIfExpression(ifHandle: .constant(.keyword(.if)))
+      )
+    }
+
```
ahoppen, line 544: We could probably unify this with `repeat` above into a `switch`

```suggestion
    // Try parse an 'if' or 'switch' as an expression. Note we do this here in
    // parseUnaryExpression as we don't allow postfix syntax to hang off such
    // expressions to avoid ambiguities such as postfix '.member', which can
    // currently be parsed as a static dot member for a result builder.
    switch self.currentToken.rawTokenKind {
      // Same for `repeat`
      case .keyword(.switch):
        return RawExprSyntax(
          parseSwitchExpression(switchHandle: .constant(.keyword(.switch)))
        )
      case .keyword(.if):
        return RawExprSyntax(
          parseIfExpression(ifHandle: .constant(.keyword(.if)))
        )
```

Sources/SwiftParser/Expressions.swift
```diff
@@ -2515,6 +2547,11 @@ extension Parser.Lookahead {
       return false
     }
 
+    // If this is the start of a switch body, this isn't a trailing closure.
+    if self.peek().rawTokenKind == .keyword(.case) {
+      return false;
```
ahoppen, line 2552:
```suggestion
      return false
```

Sources/SwiftParser/TopLevel.swift
hamishknight: `as` is currently the only thing we're going to allow as a postfix expression (we don't allow `as?` or `as!`, although we'll parse them here, but we'll diagnose them with a semantic error). We could allow `is` here too, but it will be diagnosed with a semantic error instead of a parse error.

Sources/SwiftParser/Expressions.swift
hamishknight: Good catch! Though it currently doesn't make a difference as `atStartOfExpression` is currently only be used to distinguish expressions from statements in brace element position, and we prefer to parse `if`/`switch` as statements there. We ought to still handle the case tho.

Sources/SwiftParser/Expressions.swift
hamishknight: Heh, been writing too much C++

Sources/SwiftParser/TopLevel.swift
ahoppen: OK, that makes sense to me. Just wanted to make sure we’re not missing anything.

hamishknight:
How's this @ahoppen?

ahoppen (approved):
LGTM

hamishknight:
https://github.com/apple/swift/pull/63022
https://github.com/apple/sourcekit-lsp/pull/672
https://github.com/apple/swift-format/pull/452
https://github.com/apple/swift-stress-tester/pull/213

@swift-ci please test

hamishknight:
https://github.com/apple/swift/pull/63022
https://github.com/apple/sourcekit-lsp/pull/672
https://github.com/apple/swift-format/pull/452
https://github.com/apple/swift-stress-tester/pull/213

@swift-ci please test

hamishknight:
https://github.com/apple/swift/pull/63022
https://github.com/apple/sourcekit-lsp/pull/672
https://github.com/apple/swift-format/pull/452
https://github.com/apple/swift-stress-tester/pull/213

@swift-ci please test

hamishknight:
https://github.com/apple/swift/pull/63022
https://github.com/apple/sourcekit-lsp/pull/672
https://github.com/apple/swift-format/pull/452
https://github.com/apple/swift-stress-tester/pull/213

@swift-ci please test

hamishknight:
https://github.com/apple/swift/pull/63022
https://github.com/apple/sourcekit-lsp/pull/672
https://github.com/apple/swift-format/pull/452
https://github.com/apple/swift-stress-tester/pull/213

@swift-ci please test

hamishknight:
https://github.com/apple/swift/pull/63022
https://github.com/apple/sourcekit-lsp/pull/672
https://github.com/apple/swift-format/pull/452
https://github.com/apple/swift-stress-tester/pull/213

@swift-ci please test macOS

hamishknight:
https://github.com/apple/swift/pull/63022
https://github.com/apple/sourcekit-lsp/pull/672
https://github.com/apple/swift-format/pull/452
https://github.com/apple/swift-stress-tester/pull/213

@swift-ci please test

hamishknight:
https://github.com/apple/swift/pull/63022
https://github.com/apple/sourcekit-lsp/pull/672
https://github.com/apple/swift-format/pull/452
https://github.com/apple/swift-stress-tester/pull/213

@swift-ci please test

hamishknight:
https://github.com/apple/swift/pull/63022
https://github.com/apple/sourcekit-lsp/pull/672
https://github.com/apple/swift-format/pull/452
https://github.com/apple/swift-stress-tester/pull/213

@swift-ci please test

hamishknight:
https://github.com/apple/swift/pull/63022
https://github.com/apple/sourcekit-lsp/pull/672
https://github.com/apple/swift-format/pull/452
https://github.com/apple/swift-stress-tester/pull/213

@swift-ci please test

gsl-anthonymerle:
Hello @hamishknight,

I have a question regarding the current way of what differentiates a switch expression from a switch statement:
By reading at your PR's description, I understood that the swift-syntax's differentiation between the two is that a switch statement is enclosed in a `ExpressionStmtSyntax` while a switch expression isn't.
I wonder if my understanding is correct or not because a lot of examples given to showcase switch expression is on of the following:
```
var myComputedVar: Bool {
    switch someEnum {
        case 1: true
        case 2: false
    }
}
```
```
func myFunc() -> Bool {
    switch someEnum {
        case 1: true
        case 2: false
    }
}
```

But I found out that, looking at what swift-syntax creates, these aren't actually switch expression but statements because they are being enclosed in a `ExpressionStmtSyntax`.

If my understanding of the rule to differentiate between the two, is the above actually expected?

Note: Keeping the above examples, to make them considered as switch expressions, I have to write then as follow:
```
var myComputedVar: Bool {
    let result = switch someEnum {
        case 1: true
        case 2: false
    }
    return result
}
```
```
func myFunc() -> Bool {
    let result = switch someEnum {
        case 1: true
        case 2: false
    }
    return result
}
```

Thank you in advance for your enlightenment 🙏 
Anthony.

hamishknight:
> But I found out that, looking at what swift-syntax creates, these aren't actually switch expression but statements because they are being enclosed in a ExpressionStmtSyntax.

Indeed, this is because _syntactically_ they appear in statement/expression position, and therefore are considered to be statements by swift-syntax (same as if they weren't the only element in the function body). The compiler is then responsible for determining whether they are _semantically_ expressions, and handles converting into a single expression return. For e.g a `switch` in the RHS of a binding, it is unambiguously in expression position and is therefore not wrapped in ExpressionStmtSyntax.

gsl-anthonymerle:
I see, after your explanation it makes more sense syntactically wise 🙏 

I have a follow-up question if you don't mind then: I do want to catch the cases of where switch are _syntactically_ statements but compiled as expressions (as in my previous examples). Do you think I could kind of infer that by looking in the syntactic context of the switch ? (for example, if the switch is enclosed in a computed property and there is nothing else than this switch, etc.) Or would that be a rabbit hole and I would risk to have lots of false-positives because the syntax itself isn't enough to handle all cases ?

gsl-anthonymerle:
I see, indeed I'm afraid as well to dive into a big hole trying to replicate what the compiler is doing but with the syntax only.
Thank you for all you enlightenment @hamishknight 

Best regards,
Anthony

PR #1183, opened 2023-01-04
Remove the typealiases in SwiftSyntaxBuilder that refer to syntax nodes without the `Syntax` suffix

ahoppen:
In the past, before SwiftSyntaxBuilder shared the same syntax nodes with SwiftSyntax, it made sense for it to have its own syntax nodes and thus drop the `Syntax` suffix.

But since SwiftSyntaxBuilder is now just providing convenience initializers, it’s confusing to have two names to refer to the same type - the style was already pretty inconsitent in this code base. To make things consistent, drop the typealiases.

PR #1229, opened 2023-01-13
Fix a build failure due to PRs racing

ahoppen:
#1183 and #1218 raced to create a build failure.


ahoppen:
@swift-ci Please test

PR #1231, opened 2023-01-14
Remove `TokenSyntax` shorthand static members for SwiftSyntaxBuilder

ahoppen:
In an effort to unify SwiftSyntax and SwiftSyntaxBuilder I think we should remove the shorthand static members to create tokens in SwiftSyntaxBuilder that drop the `Token` suffix.

In a follow-up PR I will investigate whether it makes sense to drop the `Token` suffix from the token static members in SwiftSyntax.

rdar://104090850


kimdv (approved):
I like it, a bit more clean and readable 🚀

ahoppen:
@swift-ci Please test

PR #1232, opened 2023-01-14
Return `nil` as text in `decomposeToRaw` for keywords

ahoppen:
This should have always returned `nil`, like for the other keywords but was missed when initially implemented.


ahoppen:
@swift-ci Pleases test

ahoppen:
@swift-ci Please test

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/TokenKindFile.swift
```diff
@@ -409,7 +409,7 @@ let tokenKindFile = SourceFileSyntax {
         for token in SYNTAX_TOKENS {
           if token.associatedValueClass != nil {
             SwitchCaseSyntax("case .\(raw: token.swiftKind)(let assoc):") {
```
kimdv, line 411: `(let assoc)` is nto used. Could that be omitted?

kimdv (approved):

PR #1233, opened 2023-01-15
[WIP] Support parsing `@_package` attribute

stevapple:
This is a very, very early (partial) demo implementation of [SwiftPM support for Swift scripts](https://forums.swift.org/t/pitch-swiftpm-support-for-swift-scripts-revision/46717), which expects feedback from both the compiler & `SwiftSyntax` team as I'm new to implement a compiler-related feature.


gyb_syntax_support/AttributeNodes.py
```diff
@@ -199,6 +201,46 @@
...
+         The arguments for the `@_package` attribute imitating `PackageDescription`
+         ''',
+         children=[
+             Child('LocationLabel', kind='IdentifierToken',
+                   text_choices=['url', 'path', 'id'],
+                   description='The location label.'),
+             Child('LocationColon', kind='ColonToken'),
+             Child('Location', kind='StringLiteralToken',
+                   description='The location/identifier of package.'),
+             Child('LocReqComma', kind='CommaToken',
```
ahoppen, line 220: Since it will be the name of the child in the SwiftSyntax API, I would prefer non-abbreviated names like `LocationRequirementComma` or `CommaBetweenLocationAndRequirement` (I think I prefer the latter, don’t have super strong opinions though).

(same below for the next comma)

Tests/SwiftParserTest/AttributeTests.swift
```diff
@@ -511,4 +511,9 @@ final class AttributeTests: XCTestCase {
     AssertParse("func takeRepeater(_ f: @MainActor @Sendable @escaping () -> Int) {}")
     AssertParse("takeRepesater { @MainActor @Sendable () -> Int in 0 }")
   }
+
+  func testPackageAttribute() {
+    AssertParse(#"@_package(path: "../my-package") import A"#)
+    AssertParse(#"@_package(url: "https://example.com/package.git", from: "0.0.1") @_exported import A"#)
```
ahoppen, line 517: Do you think you could also add a few test cases for invalid Swift code? Ideas that come to my mind are 
- misspelled argument labels
- missing commas or colons
- Using identifiers instead of string literals

Sources/SwiftParser/Attributes.swift
```diff
@@ -869,6 +876,152 @@ extension Parser {
...
+    }
+  }
+
+  mutating func parsePackageAttributeArguments() -> RawPackageAttributeArgumentsSyntax {
+    // Parsing package location.
+    let locationLabel = self.consume(ifAnyIn: PackageLocationLabel.self) ?? missingToken(.identifier)
+    let (unexpectedBeforeLocationColon, locationColon) = self.expect(.colon)
+    let (unexpectedBeforeLocation, location) = self.expect(.stringLiteral)
+    // Parsing package requirement.
+    let (unexpectedBeforeLocReqComma, locReqComma): (RawUnexpectedNodesSyntax?, RawTokenSyntax?)
```
ahoppen, line 952: I don’t think you need these type annotations.

(same below)

Sources/SwiftParser/Attributes.swift
```diff
@@ -869,6 +876,152 @@ extension Parser {
...
+    if locationLabel.tokenKind != .keyword(.path) {
+      (unexpectedBeforeLocReqComma, locReqComma) = self.expect(.comma)
+      if let label = self.consume(ifAnyIn: PackageRequirementLabel.self) {
+        (unexpectedBeforeRequirementLabel, requirementLabel) = (nil, label)
+        (unexpectedBeforeRequirementColon, requirementColon) = self.expect(.colon)
+      } else {
+        (unexpectedBeforeRequirementLabel, requirementLabel) = (nil, nil)
+        (unexpectedBeforeRequirementColon, requirementColon) = (nil, nil)
+      }
+      requirement = self.parseExpression()
```
ahoppen, line 965: Shouldn’t this also be in the `if let label` branch?

---

Open question: 

Do we want to allow arbitrary expressions here? If we only want to allow e.g. `"1.1.0"` or `"1.1.0"..<"1.2.0"` it might make sense to create a new syntax node for that kind of version requirement.

Sources/SwiftParser/Attributes.swift
```diff
@@ -869,6 +876,152 @@ extension Parser {
...
+        unexpectedBeforeRequirementLabel,
+        requirementLabel: requirementLabel,
+        unexpectedBeforeRequirementColon,
+        requirementColon: requirementColon,
+        requirement: requirement,
+        reqProdComma: nil,
+        productLabel: nil,
+        productColon: nil,
+        productName: nil,
+        self.remainingTokensIfMaximumNestingLevelReached(),
```
ahoppen, line 990: You shouldn’t pass this. `remainingTokensIfMaximumNestingLevelReached` is to prevent stack overflow in the parser and should always be combined with an early return. If you’re interested, search for `remainingTokensIfMaximumNestingLevelReached` in this project to see how it’s being used. But since `@_package` can’t be the source of deep recursion (like e.g. expression parsing can), you don’t need it.

(same below)

Sources/SwiftParser/Attributes.swift
```diff
@@ -869,6 +876,152 @@ extension Parser {
...
+        arena: self.arena
+      )
+    }
+    let (unexpectedBeforeReqProdComma, reqProdComma) = self.expect(.comma)
+    let (unexpectedBeforeProductLabel, productLabel) = self.expect(.keyword(.product))
+    let (unexpectedBeforeProductColon, productColon) = self.expect(.colon)
+    let (unexpectedBeforeProductName, productName) = self.expect(.stringLiteral)
```
ahoppen, line 997: Just a heads-up: This will need to be `parseStringLiteral` and you need to change the type of `product` to `StringLiteralExpr` in `AttributeNodes.py once https://github.com/apple/swift-syntax/pull/1192 is merged.

ahoppen (commented):
The implementation looks good to me overall. I’ve got a few comments inline.

gyb_syntax_support/AttributeNodes.py
stevapple: done👌

Sources/SwiftParser/Attributes.swift
stevapple: Well, the current proposal handles requirement like this:
* If the given location is a `path`, there should be no requirement; otherwise we always have one.
* If the label is `from`, it accepts a semantic version in string literal;
* If the label is `branch`, it accepts a branch name in string literal;
* If the label is `revision`, it accepts a commit hash in string literal.
* If there's no label, it accepts either a range expression (eg. `"1.0.0"..<"1.1.0"`) or a member access expression (eg. `.upToNextMinor(from: "1.0.0")`).

I haven't decided if we want to simplify the interface (making it less like `PackageDescription`), so we're now accepting arbitrary expressions here.

stevapple: Since this is only a draft, there're plenty of rooms for refactors, like:
* Extracting the logic of `@_package` into a separate file like `@available` does;
* Implementing fully semantical parsing logic including expression diagnostic and even semantics versioning;
* Consuming all parameters at a whole (just like a tuple) and implementing semantical error handling.

WDYT?

Sources/SwiftParser/Attributes.swift
stevapple: Thanks for pointing out this! Fixed~

Sources/SwiftParser/Attributes.swift
```diff
@@ -869,6 +876,152 @@ extension Parser {
   }
 }
 
+extension Parser {
+  mutating func parsePackageAttribute() -> RawAttributeSyntax {
```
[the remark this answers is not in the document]
stevapple: Removed👌

Sources/SwiftParser/Attributes.swift
stevapple: It seems they're still required. I'll extract the parsing logic to independent methods to keep this one clean. WIP

Tests/SwiftParserTest/AttributeTests.swift
stevapple: I didn't care of diagnostics for now, so some false syntax may be accepted by the current implementation. The error messages also seemed not helpful at all...

Anyway I'll do that some time. Thanks for your suggestion:)

Sources/SwiftParser/Attributes.swift
ahoppen: > * Extracting the logic of `@_package` into a separate file like `@available` does;

I don’t think starting a new file makes sense for ~100 LOC. If things grow further, 

> * Implementing fully semantical parsing logic including expression diagnostic and even semantics versioning

What do you mean by “fully semantic parsing logic including expression diagnostics”. 

> * Introducing additional node types for package attribute;

There’s always a trade-off here between expressibility and trying to keep the syntax tree simple. The main consideration would be if there are invariants in the syntax tree that clients might care about but which are not expressed in the syntax tree at the moment (e.g. if two options are mutually exclusive). I don’t see anything like that at the moment, so maybe it’s fine as-is.

> * Consuming all parameters at a whole (just like a tuple) and implementing semantical error handling.
My gut f

stevapple: Now I've refactored the support and it grows to ~120 loc. Parser implementation should look better with new nodes added.

After checking `PackageDescription` source, I found that wildcard requirement was deprecated in SwiftPM 5.6, so for a new feature it makes sense to drop the support. This means although the node is a plain `Expr` at the time, we can further assert it to be a range expression ( `semantic-version-low ( '..<' | '...' ) semantic-version-high`).

Tests/SwiftParserTest/AttributeTests.swift
stevapple: I adds some more valid test cases and I believe the parsing is robust enough now. It's time to dive into invalid cases...

Sources/SwiftParser/Attributes.swift
ahoppen: If we just allow `semantic-version-low ( '..<' | '...' ) semantic-version-high`, I think it would make sense to represent the version requirement in the syntax tree as three nodes `VersionTuple`, `BinaryOperator` and `VersionTuple`.

stevapple: `VersionTuple` doesn't fully mimic a semantic version (which is rather complex and better handed over to other tools). I’m going to use `StringLiteralExpr` instead. Thanks for your suggestions on `BinaryOperator`, but I didn’t have idea on how to diagnose wrong operator here🤔 Should that be handled elsewhere?

ahoppen: Handling the semantic version somewhere else sounds good to me. The binary operator could be anything. There isn’t any way to restrict the kinds of operators we allow in the SwiftSyntax tree.

PR #1240, opened 2023-01-16
In SwiftSyntaxBuilder change String parameters to a custom string interpolation protocol

ahoppen:
The current mixed String + result builder initializers in SwiftSyntaxBuilder allow people to construct string using all sorts of dynamic features (like concatenation) and then pass them into the initializer. That’s not an API paradigm we want to support.

Instead, these initializers should take a custom type that is expressible by string literal, similar to how expressibility by string literals is implemented for syntax nodes. This stops people from doing string transformations on the strings and only allows string interpolation.

rdar://104090423

### DIFF
--- a/Sources/SwiftSyntaxBuilder/Syntax+StringInterpolation.swift
+++ b/Sources/SwiftSyntaxBuilder/Syntax+StringInterpolation.swift
@@ -140,7 +140,7 @@ extension SyntaxStringInterpolation: StringInterpolationProtocol {
 /// Syntax nodes that can be formed by a string interpolation involve source
 /// code and interpolated syntax nodes.
 public protocol SyntaxExpressibleByStringInterpolation:
-  ExpressibleByStringInterpolation, SyntaxProtocol
+  ExpressibleByStringInterpolation
 where Self.StringInterpolation == SyntaxStringInterpolation {
   init(stringInterpolationOrThrow stringInterpolation: SyntaxStringInterpolation) throws
 }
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

kimdv (approved):

Sources/SwiftSyntaxBuilder/Syntax+StringInterpolation.swift
```diff
@@ -140,7 +140,7 @@ extension SyntaxStringInterpolation: StringInterpolationProtocol {
 /// Syntax nodes that can be formed by a string interpolation involve source
 /// code and interpolated syntax nodes.
 public protocol SyntaxExpressibleByStringInterpolation:
-  ExpressibleByStringInterpolation, SyntaxProtocol
```
bnbarham, line 143: I'm surprised nothing is using this 🤔

Sources/SwiftSyntaxBuilder/SyntaxNodeWithBody.swift
```diff
@@ -37,7 +55,7 @@ public protocol HasTrailingCodeBlock {
 }
 
 public extension HasTrailingCodeBlock where Self: SyntaxExpressibleByStringInterpolation {
```
bnbarham, line 57: There's a TODO just above this that you could remove now if you want

bnbarham (approved):

Sources/SwiftSyntaxBuilder/SyntaxNodeWithBody.swift
ahoppen: Good point. Removed it

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

PR #1241, opened 2023-01-16
When running CodeGeneration, remove files from templates that no longer exist

ahoppen:
This would have caught https://github.com/apple/swift-syntax/pull/1193.


ahoppen:
@swift-ci Please test

kimdv (approved):

PR #1242, opened 2023-01-16
[WIP] Allow try before yield

airspeedswift:
This adds the ability to parse a `try` before a `yield` statement, as part of a path to dropping the underscore on `_modify`.

This approach seems to work OK for statements, and keeps the current tests for rogue `try`s in front of various statement  passing  (though it needs more work to better handle `try` in front of additional kinds of statement). But allowing `try` to start a statement like this causes all sorts of recovery breakage for other things so this isn't ready to go yet.


Sources/SwiftParser/Statements.swift
```diff
@@ -1241,6 +1280,8 @@ extension Parser.Lookahead {
       .switchKeyword?,
       .poundAssertKeyword?:
       return true
+    case .tryKeyword?:
+      return true
```
airspeedswift, line 1284: seems like tightening this up will resolve some of the recovery knock-on effects elsewhere

Sources/SwiftParser/Statements.swift
```diff
@@ -1240,6 +1280,10 @@ extension Parser.Lookahead {
       .switchKeyword?,
       .poundAssertKeyword?:
       return true
+    case .tryKeyword?:
+      // strip the try and then check if the rest is a statement
+      self.eat(.keyword(.try))
+      return isStartOfStatement(allowRecovery: allowRecovery)
```
airspeedswift, line 1286: this seems to resolve all but one of the recovery regressions

Sources/SwiftParser/TokenPrecedence.swift
```diff
@@ -159,7 +159,7 @@ public enum TokenPrecedence: Comparable {
       // Secondary parts of control-flow constructs
       .keyword(.case), .keyword(.catch), .keyword(.default), .keyword(.else),
       // Return-like statements
-      .keyword(.break), .keyword(.continue), .keyword(.fallthrough), .keyword(.return), .keyword(.throw), .keyword(.yield),
+      .keyword(.break), .keyword(.continue), .keyword(.fallthrough), .keyword(.return), .keyword(.throw), .keyword(.yield), .keyword(.try),
```
airspeedswift, line 162: Maybe this is iffy, but without it it was giving `CanBeStatementStart`'s recovery precedence that of an expression as calculated [here](https://github.com/apple/swift-syntax/blob/b931534a93a21e1ceda00495ab6cb5ff4d8bd649/Sources/SwiftParser/Recovery.swift#L104). An alternative could to give `RawTokenKindSubset` a customization point and have the one for `CanBeStatementStart` exclude `try`.

airspeedswift:
@swift-ci please test

airspeedswift:
@swift-ci please test

Sources/SwiftParser/Statements.swift
```diff
@@ -73,56 +73,59 @@ extension Parser {
...
         )
       )
+    } else {
+      return parseStatementAfterLabel()
     }
+  }
 
-    let optLabel = self.parseOptionalStatementLabel()
+  @_spi(RawSyntax)
+  public mutating func parseStatementAfterLabel(
```
ahoppen, line 91: I don’t think this function needs to be public.

Sources/SwiftParser/Statements.swift
```diff
@@ -408,11 +411,14 @@ extension Parser {
   ///
   ///     throw-statement → 'throw' expression
   @_spi(RawSyntax)
-  public mutating func parseThrowStatement(throwHandle: RecoveryConsumptionHandle) -> RawThrowStmtSyntax {
+  public mutating func parseThrowStatement(
+    throwHandle: RecoveryConsumptionHandle,
+    with tryPrefix: (RawUnexpectedNodesSyntax?, RawTokenSyntax)?
```
ahoppen, line 416: What do you think about naming the elements of the tuple so we don’t have the `.0` and `.1` access below?

Also, I would prefer to name the parameter `tryPrefix` (instead of the external argument name `with` because nothing really suggests what `with` should be).

---

```suggestion
    tryPrefix: (unexpectedBeforeTry: RawUnexpectedNodesSyntax?, tryToken: RawTokenSyntax)?
```

(same for the other functions)

Sources/SwiftParser/Statements.swift
ahoppen: That’s how I would have implemented it as well. What’s the recovery recession that still remains?

Sources/SwiftParser/TokenPrecedence.swift
ahoppen: Raising the precedence of `try` seems totally correct to me. If `try` can start a statement now, we shouldn’t be eating it while looking for a statement keyword.

ahoppen (commented):
Looks good to me overall. I’ve got a few naming/style comments and I think a few test cases would be good.

Sources/SwiftParser/Statements.swift
airspeedswift: it was a bug fixed by the precedence chance where it was doing ")blah return foo" and not recovering to the return

ahoppen: Ah, great 👍

Sources/SwiftParser/Statements.swift
ahoppen: I can agree with your argument for the recovery handle but for the `tryPrefix`, I’m still not sure. If the `tryPrefix` took a parameter of type `RawTryTokenSyntax` (which we don’t have), I would be on your side. But as it stands `RawTokenSyntax` is completely ambiguous as to which token it is and if I consider the public signature of this function, nothing indicates that you should pass a the `try` token here. Or, putting it another way: How would the API look like if we also allowed an `await` prefix? You surely don’t want to have two `with` parameters,  right?

airspeedswift:
@swift-ci please test

airspeedswift:
@swift-ci please test

airspeedswift:
@swift-ci please test

PR #1243, opened 2023-01-17
Create separate `lexImpl` methods for the different lexer states

ahoppen:
This adds an additional level of safety because we can assert which characters we are expecting at each state.

It also makes it easier to e.g. extract string lexing into a `StringLexer` in the future, should we want to do that in the future.


ahoppen:
@swift-ci Please test

rintaro (approved):

ahoppen:
@swift-ci Please test

PR #1248, opened 2023-01-19
Don't fail LoopProgressCondition when consuming zero-length string segment

### LINKED ISSUE #1247: Parser crash with skipSingle and string literals
### Issue Kind

Parser Crash

### Source Code

```swift
if p{""
```

### Description

_No response_

ahoppen:
The loop has made progress if either
 - The parser is now pointing at a different location in the source file
 - The parser is still pointing at the same position in the source file but now has a different token kind (and thus consumed a zero-length token like an empty string interpolation

The second condition is new.

Fixes #1247
rdar://104413423

### DIFF
--- a/Sources/SwiftParser/Lookahead.swift
+++ b/Sources/SwiftParser/Lookahead.swift
@@ -307,6 +307,13 @@ extension Parser.Lookahead {
     return skip(initialState: .skipSingle)
   }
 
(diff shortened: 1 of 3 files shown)


ahoppen:
@swift-ci Please test

DougGregor (approved):

ahoppen:
@swift-ci Please test

PR #1250, opened 2023-01-19
Lex string literal segments in the lexer and compose them in the parser

ahoppen:
Instead of parsing the entire string literal as a single string literal token, directly produce the lexemes that will become tokens in the parser.


ahoppen:
@swift-ci Please test

Sources/SwiftParser/StringLiterals.swift
```diff
@@ -0,0 +1,126 @@
...
+    /// Parse segments.
+    var segments: [RawStringLiteralSegmentsSyntax.Element] = []
+    var loopProgress = LoopProgressCondition()
+    while loopProgress.evaluate(self.currentToken) {
+      if let stringSegment = self.consume(if: .stringSegment) {
+        segments.append(.stringSegment(RawStringSegmentSyntax(content: stringSegment, arena: self.arena)))
+      } else if let backslash = self.consume(if: .backslash) {
+        let (unexpectedBeforeDelimiter, delimiter) = self.parseStringDelimiter(openDelimiter: openDelimiter)
+        let (unexpectedBeforeLeftParen, leftParen) = self.expect(.leftParen)
+        let expressions = RawTupleExprElementListSyntax(elements: self.parseArgumentListElements(pattern: .none), arena: self.arena)
```
rintaro, line 65: How newlines in the string interpolation are handled? e.g.
```
let a = "test \(label:
foo()
```
This should be missing expression after `label:`, missing `)`, and missing `"`. But I think `foo()` is parsed as the expression, no?

Sources/SwiftParser/StringLiterals.swift
```diff
@@ -0,0 +1,126 @@
...
+
+        // For recovery, eat anything up to the next token that either starts a new string segment or terminates the string.
+        // This allows us to skip over extraneous identifiers etc. in an unterminated string interpolation.
+        var unexpectedBeforeRightParen: [RawTokenSyntax] = []
+        var unexpectedProgress = LoopProgressCondition()
+        while !self.at(any: [.rightParen, .stringSegment, .backslash, openQuote.tokenKind, .eof]) && unexpectedProgress.evaluate(self.currentToken) {
```
rintaro, line 71: In single line string literals, this should stop at newline IMO.

Sources/SwiftParser/Lexer/Cursor.swift
```diff
@@ -165,24 +250,34 @@ extension Lexer.Cursor {
...
       result = lexAfterRawStringDelimiter(delimiterLength: delimiterLength)
     case .inStringLiteral(kind: let stringLiteralKind, delimiterLength: let delimiterLength):
-      result = lexInStringLiteral(kind: stringLiteralKind, delimiterLength: delimiterLength)
+      result = lexInStringLiteral(stringLiteralKind: stringLiteralKind, delimiterLength: delimiterLength)
     case .afterStringLiteral(isRawString: _):
       result = lexAfterStringLiteral()
     case .afterClosingStringQuote:
       result = lexAfterClosingStringQuote()
+    case .afterBackslashOfStringInterpolation(stringLiteralKind: let stringLiteralKind):
+      result = lexAfterBackslashOfStringInterpolation(stringLiteralKind: stringLiteralKind)
```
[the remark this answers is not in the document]
ahoppen: Good idea 👍 Thanks.

Sources/SwiftParser/StringLiterals.swift
ahoppen: We are leaving the `inStringInterpolationState` state when hitting a newline because of

https://github.com/apple/swift-syntax/pull/1250/files#diff-253b01bc981faa185173c1697784a41bf1b586e91ec7b4de806dfd2646db52c3R135

I added test additional test cases for this

Sources/SwiftParser/StringLiterals.swift
ahoppen: The lexer automatically produces a `stringSegment` at the end of a single-line string literal, even if it is not terminated (see the new test `testNewlineInInterpolationOfSingleLineString` in LexerTest.swift), so we will stop at that.

ahoppen:
@swift-ci Please test

PR #1253, opened 2023-01-20
Remove the with<childName> functions on syntax nodes

ahoppen:
Instead, provide a single `with` function that takes a key path of the child you want to replace. This significatnly reduces the amout of code that is being generated for SwiftSyntax.

While at it, also remove the `without(Leading|Trailing)?Trivia` functions.

@DougGregor I changed a few calls of `token.withoutTrivia()` in macros to `token.text`, which should perform the same thing while also being slightly more efficient.

### DIFF
--- a/CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxTraitsFile.swift
+++ b/CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxTraitsFile.swift
@@ -24,9 +24,21 @@ let syntaxTraitsFile = SourceFileSyntax {
       """) {
       
       for child in trait.children {
-        VariableDeclSyntax("var \(raw: child.swiftName): \(raw: child.typeName)\(raw: child.isOptional ? "?" : "") { get }")
-        FunctionDeclSyntax("func with\(raw: child.name)(_ newChild: \(raw: child.typeName)\(raw: child.isOptional ? "?" : "")) -> Self")
+        VariableDeclSyntax("var \(raw: child.swiftName): \(raw: child.typeName)\(raw: child.isOptional ? "?" : "") { get set }")
       }
+    }
+
+    ExtensionDeclSyntax("public extension \(trait.traitName)Syntax") {
+      FunctionDeclSyntax("""
+      /// Without this function, the `with` function defined on `SyntaxProtocol`
+      /// does not work on existentials of this protocol type.
+      @_disfavoredOverload
+      func with<T>(_ keyPath: WritableKeyPath<\(raw: trait.traitName)Syntax, T>, _ newChild: T) -> \(raw: trait.traitName)Syntax {
+        var copy: \(raw: trait.traitName)Syntax = self
+        copy[keyPath: keyPath] = newChild
+        return copy
+      }
+      """)
     }
     
     ExtensionDeclSyntax("public extension SyntaxProtocol") {
--- a/Sources/SwiftParserDiagnostics/SyntaxExtensions.swift
+++ b/Sources/SwiftParserDiagnostics/SyntaxExtensions.swift
@@ -71,7 +71,7 @@ extension SyntaxProtocol {
   /// diagnostic message), return that.
   /// Otherwise, return a generic message that describes the tokens in this node.
   var shortSingleLineContentDescription: String {
-    let contentWithoutTrivia = self.withoutLeadingTrivia().withoutTrailingTrivia().description
+    let contentWithoutTrivia = self.trimmedDescription
     if self.children(viewMode: .sourceAccurate).allSatisfy({ $0.as(TokenSyntax.self)?.tokenKind == .rightBrace }) {
       if self.children(viewMode: .sourceAccurate).count == 1 {
         return "brace"
--- a/Sources/SwiftSyntax/SyntaxNodes.swift.gyb.template
+++ b/Sources/SwiftSyntax/SyntaxNodes.swift.gyb.template
@@ -194,7 +194,14 @@ public struct ${node.name}: ${base_type}Protocol, SyntaxHashable {
       return ${child_type}(childData!)
     }
     set(value) {
-      self = with${child.name}(value)
+      let arena = SyntaxArena()
+  %       if child.is_optional:
+      let raw = value?.raw
+  %       else:
+      let raw = value.raw
+  %       end
+      let newData = data.replacingChild(at: ${idx}, with: raw, arena: arena)
+      self = ${node.name}(newData)
     }
   }
 %
@@ -229,23 +236,6 @@ public struct ${node.name}: ${base_type}Protocol, SyntaxHashable {
     return ${node.name}(newData)
   }
 %       end
-
-%       # ===================
-%       # Replacing children
-%       # ===================
-  /// Returns a copy of the receiver with its `${child.swift_name}` replaced.
-  /// - param newChild: The new `${child.swift_name}` to replace the node's
-  ///                   current `${child.swift_name}`, if present.
-  public func with${child.name}(_ newChild: ${child_type}${'?' if child.is_optional else ''}) -> ${node.name} {
-    let arena = SyntaxArena()
-%       if child.is_optional:
-    let raw = newChild?.raw
-%       else:
-    let raw = newChild.raw
-%       end
-    let newData = data.replacingChild(at: ${idx}, with: raw, arena: arena)
-    return ${node.name}(newData)
-  }
 %     end
 
   public static var structure: SyntaxNodeStructure {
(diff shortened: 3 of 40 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
https://github.com/apple/swift-format/pull/477
https://github.com/apple/swift-stress-tester/pull/224

@swift-ci Please test

Sources/SwiftSyntax/SyntaxNodes.swift.gyb.template
```diff
@@ -194,7 +194,14 @@ public struct ${node.name}: ${base_type}Protocol, SyntaxHashable {
...
     set(value) {
-      self = with${child.name}(value)
+      let arena = SyntaxArena()
+  %       if child.is_optional:
+      let raw = value?.raw
+  %       else:
+      let raw = value.raw
+  %       end
+      let newData = data.replacingChild(at: ${idx}, with: raw, arena: arena)
+      self = ${node.name}(newData)
```
bnbarham, line 204: Ah, I was hoping we could use `SyntaxNodeStructure` and the need for this entirely. We don't currently have the reverse mapping (ie. keypath -> index), but maybe that would be useful. That would be even more codegen though, not less 😅.

bnbarham (approved):

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxTraitsFile.swift
```diff
@@ -24,10 +24,22 @@ let syntaxTraitsFile = SourceFileSyntax {
...
     }
+
+    ExtensionDeclSyntax("public extension \(trait.traitName)Syntax") {
+      FunctionDeclSyntax("""
+      /// Without this function, the `with` function defined on `SyntaxProtocol`
+      /// does not work on existentials of this protocol type.
+      @_disfavoredOverload
+      func with<T>(_ keyPath: WritableKeyPath<\(raw: trait.traitName)Syntax, T>, _ newChild: T) -> \(raw: trait.traitName)Syntax {
+        var copy: \(raw: trait.traitName)Syntax = self
+        copy[keyPath: keyPath] = newChild
```
bnbarham, line 38: `layoutKeyPath` or something instead? `keyPath: keyPath` just looks weird 😅

rintaro:
I know we decided to do `with()` accepting keypaths. But maybe resurrecting `Cursor` (but with associated values) might be an option?
For example:
```swift
extension FunctionDeclSyntax {
  enum Cursor {
    case attributes(AttributeListSyntax)
    case modifiers(ModifierListSyntax?)
    ...
  }
  func with(_ values: Cursor...) -> Self {
    ...
    for cursor in values {
      switch cursor {
        case .attributes(let val): layout[0] = val.raw
        case .modifiers(let val): layout[1] = val?.raw
      }
    }
    ...
  }
}
```
So you can call `foo = bar.with(.attribute(newAttr), .signature(rewritten))`

rintaro:
A concern accepting KeyPaths on `Self` is that it need to accept any properties including that is not meant to. For example:
```swift
newFn = fn.with(\.self, otherFn)
```
It works, but not sure we should accept this.

Also do we want to accept multi-level paths? `fn.with(\.signature.input, newInput)`. This looks cool, so I think the answer is yes.

Sources/SwiftParserDiagnostics/SyntaxExtensions.swift
```diff
@@ -71,7 +71,7 @@ extension SyntaxProtocol {
   /// diagnostic message), return that.
   /// Otherwise, return a generic message that describes the tokens in this node.
   var shortSingleLineContentDescription: String {
-    let contentWithoutTrivia = self.withoutLeadingTrivia().withoutTrailingTrivia().description
+    let contentWithoutTrivia = self.with(\.leadingTrivia, []).with(\.trailingTrivia, []).description
```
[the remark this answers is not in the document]
ahoppen: Yeah, it seems wasteful but it seems like a fairly narrow use-case and I don’t want to blow up our API surface for such use cases.

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxTraitsFile.swift
ahoppen: Well, it’s not just layouts right. `keyPath` can also point to trivia.

Sources/SwiftSyntax/SyntaxNodes.swift.gyb.template
ahoppen: How would you want to get rid of these using `SyntaxNodeStructure`?

ahoppen:
> I know we decided to do with() accepting keypaths. But maybe resurrecting Cursor (but with associated values) might be an option?

That would just be trading some code generation for another. What I like about the key path approach is that it doesn’t require code generation.

> A concern accepting KeyPaths on `Self` is that it need to accept any properties including that is not meant to. For example:
> 
> ```swift
> newFn = fn.with(\.self, otherFn)
> ```
> 
> It works, but not sure we should accept this.
> 
> Also do we want to accept multi-level paths? `fn.with(\.signature.input, newInput)`. This looks cool, so I think the answer is yes.


I actually really liked how the with function generalized from layouts to trivia as well. And it’s really just a shorthand to write three lines in one, so I’m totally happy if it generalizes to multi-level key paths. And I don’t see any problem with `fn.with(\.self, otherFn)` working as well. Sure, it’s silly but it does exactly what you expect, right?

ahoppen:
https://github.com/apple/swift/pull/63160
https://github.com/apple/swift-format/pull/477
https://github.com/apple/swift-stress-tester/pull/224

@swift-ci Please test

ahoppen:
https://github.com/apple/swift/pull/63160
https://github.com/apple/swift-format/pull/477
https://github.com/apple/swift-stress-tester/pull/224

@swift-ci Please test

ahoppen:
https://github.com/apple/swift/pull/63160
https://github.com/apple/swift-format/pull/477
https://github.com/apple/swift-stress-tester/pull/224

@swift-ci Please test macOS

Sources/SwiftSyntax/SyntaxNodes.swift.gyb.template
bnbarham: You can't right, we need the reverse mapping (key path -> index).

Sources/SwiftParserDiagnostics/SyntaxExtensions.swift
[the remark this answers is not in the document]
bnbarham: Doesn't seem *that* narrow. Can leave for now, we can always add later.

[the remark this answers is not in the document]
jpsim: Drive-by comment: it's useful for SwiftLint, we currently have 13 occurrences of `withoutTrivia().description` https://github.com/search?q=repo%3Arealm%2FSwiftLint%20withoutTrivia().description&type=code

Maybe there's a better way to accomplish this, but that's what we're doing now.

ahoppen:
https://github.com/apple/swift/pull/63160
https://github.com/apple/swift-format/pull/477
https://github.com/apple/swift-stress-tester/pull/224

@swift-ci Please test macOS

ahoppen:
https://github.com/apple/swift/pull/63160
https://github.com/apple/swift-format/pull/477
https://github.com/apple/swift-stress-tester/pull/224

@swift-ci Please test macOS

Sources/SwiftParserDiagnostics/SyntaxExtensions.swift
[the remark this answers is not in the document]
ahoppen: OK, you convinced me. I added `trimmedDescription`. I decided against naming it `descriptionWithoutTrivia` because that sounds like we would also remove trivia in the middle of the node.

ahoppen:
https://github.com/apple/swift/pull/63160
https://github.com/apple/swift-format/pull/477
https://github.com/apple/swift-stress-tester/pull/224

@swift-ci Please test macOS

ahoppen:
https://github.com/apple/swift/pull/63160
https://github.com/apple/swift-format/pull/477
https://github.com/apple/swift-stress-tester/pull/224

@swift-ci Please test macOS

ahoppen:
https://github.com/apple/swift/pull/63160
https://github.com/apple/swift-format/pull/477
https://github.com/apple/swift-stress-tester/pull/224

@swift-ci Please test macOS

ahoppen:
https://github.com/apple/swift/pull/63160
https://github.com/apple/swift-format/pull/477
https://github.com/apple/swift-stress-tester/pull/224

@swift-ci Please test Linux

ahoppen:
https://github.com/apple/swift/pull/63160
https://github.com/apple/swift-format/pull/477
https://github.com/apple/swift-stress-tester/pull/224

@swift-ci Please test

ahoppen:
https://github.com/apple/swift/pull/63160
https://github.com/apple/swift-format/pull/477
https://github.com/apple/swift-stress-tester/pull/224

@swift-ci Please test macOS

PR #1255, opened 2023-01-20
Diagnose incorrect indentation in multi-line string literals

ahoppen:
In contrast to the C++ parser, we diagnose the indentation in the parser instead of the lexer. This is necessary since the lexer doesn’t know the expected indentation because it can’t/shouldn’t look ahead to the closing quote to know the expected indentation. Thus, the lexer requires relatively few changes.

The bulk of the changes are in SwiftParser/StringLiterals.swift. Here, we post-process the tokens produced by the lexer, validating indentation. The basic ideas are:
- The text of the string segments should only contain text that’s actually part of the string literal. Indentation will be stripped away and re-classified as trivia.
- Similarly the following part of the string become trivia
  - The newline that separates the opening `"""` and the first line
  - The last newline before the closing `"""`
  - Escaped newlines (lines ending with `\`)
- If we discover an indentation error in one of the tokens, we mark that token as having a `insufficientIndentationInMultilineStringLiteral` lexer error – really lexer error is misnamed here now, maybe token error would be better but renaming that would be a follow-up PR

During diagnostics generation `MultiLineStringLiteralIndentatinDiagnosticsGenerator` collects all errors in a multi-line string literal, so we can show errors like “incorrect indentation in the next 3 lines”, which requires a little bit of context between the errors inside the string literal.

The remaining changes are fairly straightforward.


Sources/SwiftParser/StringLiterals.swift
```diff
@@ -44,6 +148,243 @@ extension Parser {
...
+          return true
+        }
+      },
+      "String segement produced by the lexer should not have unexpected text or trivia because we would drop it during post-processing"
+    )
+
+    // -------------------------------------------------------------------------
+    // Variables
+
+    var middleSegments = allSegments
```
bnbarham, line 195: Seems like we could avoid removing from the start and then inserting later if we instead just kept separate arrays here, ie.
```
let multilineSegments = allSegments.dropFirst().dropLast() // this is a slice rather than the array
let expectedStart = allSegments.count > 1 ? allSegments.first : nil
let expectedIndentation = allSegments.last()

var processed = [RawStringLiteralSegmentsSyntax.Element]()
```

I can understand not wanting to do that though as it makes it a little easier to accidentally drop a segment, I still think it's cleaner but... up to you. I'd change the `removeLast` to just a `popLast` in either case (no `popFirst`, I assume because they didn't want to make that easy to do 🤷).

IMO it also wouldn't hurt to have a comment at the start of this function explaining the expected segments, ie. newline, the actual string segments, ending indentation + that no trivia is expected (the assertion does cover that though).

Sources/SwiftParser/StringLiterals.swift
```diff
@@ -44,6 +148,243 @@ extension Parser {
...
+        if case .stringSegment(let segment) = $0 {
+          return segment.unexpectedBeforeContent == nil
+            && segment.unexpectedAfterContent == nil
+            && segment.content.leadingTriviaByteLength == 0
+            && segment.content.trailingTriviaByteLength == 0
+        } else {
+          return true
+        }
+      },
+      "String segement produced by the lexer should not have unexpected text or trivia because we would drop it during post-processing"
```
bnbarham, line 189:
```suggestion
      "String segment produced by the lexer should not have unexpected text or trivia because we would drop it during post-processing"
```

Sources/SwiftParser/StringLiterals.swift
```diff
@@ -44,6 +148,243 @@ extension Parser {
...
+    let indentation: SyntaxText
+    let indentationTrivia: [RawTriviaPiece]
+
+    var unexpectedBeforeOpenQuote: [RawTokenSyntax] = []
+    var openQuote = openQuote
+    var unexpectedBeforeCloseQuote: [RawTokenSyntax] = []
+    var closeQuote = closeQuote
+
+    // -------------------------------------------------------------------------
+    // Check close quote is on new line
```
bnbarham, line 208: I'd personally pull this out into a separate function. Then all the `closeDelimiterOnNewLine`s can just be returns, which IMO is easier to follow. It would then just become:
```
guard let lastStringSegment = reclassifyNewline(multilineSegments.last) else {
        // The closing delimiter doesn't start on a new line and thus it doesn't
        // make sense to try and extract indentation from it.

        unexpectedBeforeCloseQuote = [closeQuote]
        closeQuote = RawTokenSyntax(missing: closeQuote.tokenKind, leadingTriviaPieces: [.newlines(1)], arena: self.arena)
        return (unexpectedBeforeOpenQuote, openQuote, allSegments, unexpectedBeforeCloseQuote, closeQuote)
}
```

Sources/SwiftParser/StringLiterals.swift
```diff
@@ -44,6 +148,243 @@ extension Parser {
...
+    // Condition for the loop below that indicates whether the segment we are
+    // iterating over is on a new line.
+    var isSegmentOnNewLine: Bool
+
+    if let newlineTrivia = firstSegment?.content.tokenText.triviaPieceIfNewline {
+      openQuote = openQuote.extendingTrailingTrivia(by: [newlineTrivia], arena: self.arena)
+      isSegmentOnNewLine = true
+    } else {
+      if let firstSegment = firstSegment {
+        middleSegments.insert(.stringSegment(firstSegment), at: 0)
```
bnbarham, line 297: Following above this would then be `processed.append(.stringSegment(firstSegment))`.

Sources/SwiftParser/StringLiterals.swift
```diff
@@ -44,6 +148,243 @@ extension Parser {
...
+    }
+
+    // -------------------------------------------------------------------------
+    // Check open quote followed by newline
+
+    // Condition for the loop below that indicates whether the segment we are
+    // iterating over is on a new line.
+    var isSegmentOnNewLine: Bool
+
+    if let newlineTrivia = firstSegment?.content.tokenText.triviaPieceIfNewline {
```
bnbarham, line 292: Probably worth adding a test for:
```
_ = """<some whitespace>
"""
```

This handles that properly, I just don't remember seeing one for it.

Sources/SwiftParser/StringLiterals.swift
```diff
@@ -44,6 +148,243 @@ extension Parser {
...
+    // Check indentation of segments and escaped newlines at end of segment
+
+    let expressionIndentationChecker = StringLiteralExpressionIndentationChecker(expectedIndentation: indentation, arena: self.arena)
+
+    for (index, segment) in middleSegments.enumerated() {
+      switch segment {
+      case .stringSegment(var segment):
+        // We are not considering leading trivia for indentation computation.
+        // If these assertions are violated, we can probably lift them but we
+        // would need to check the produce the expected results.
```
bnbarham, line 314:
```
        // would need to check the produce the expected results.
```

`to check the produce` 😅. `we` maybe? IMO I'd probably just skip the comment and the assertion since we already have this assertion at the start of the function (for all segments).

Sources/SwiftParser/StringLiterals.swift
```diff
@@ -44,6 +148,243 @@ extension Parser {
...
+          if segment.content.tokenText.hasPrefix(indentation) {
+            segment = RawStringSegmentSyntax(
+              segment.unexpectedBeforeContent,
+              content: segment.content.reclassifyAsLeadingTrivia(indentationTrivia, arena: self.arena),
+              segment.unexpectedAfterContent,
+              arena: self.arena
+            )
+          } else {
+            let actualIndentation = segment.content.tokenText.prefix(while: { $0 == UInt8(ascii: " ") || $0 == UInt8(ascii: "\t") })
+            let actualIndentationTriva = TriviaParser.parseTrivia(SyntaxText(rebasing: actualIndentation), position: .leading)
```
bnbarham, line 328:
```suggestion
            let actualIndentationTrivia = TriviaParser.parseTrivia(SyntaxText(rebasing: actualIndentation), position: .leading)
```

Sources/SwiftParser/StringLiterals.swift
```diff
@@ -44,6 +148,243 @@ extension Parser {
...
+            let content = segment.content.reclassifyAsLeadingTrivia(
```
bnbarham, line 329: Could this just take the length to reclassify and do the trivia parsing itself?

Sources/SwiftParser/StringLiterals.swift
```diff
@@ -44,6 +148,243 @@ extension Parser {
...
+              segment.unexpectedAfterContent,
+              arena: self.arena
+            )
+          }
+        }
+
+        isSegmentOnNewLine = segment.content.tokenText.newlineSuffix != nil
+
+        // If the segment has a `\` in front of its trailing newline, that newline
+        // is not part of the reprsented string and should be trivia.
```
bnbarham, line 346:
```suggestion
        // is not part of the represented string and should be trivia.
```

Sources/SwiftParser/StringLiterals.swift
```diff
@@ -44,6 +148,243 @@ extension Parser {
...
+        }
+        if let backslashNewlineSuffix = backslashNewlineSuffix {
+          segment = RawStringSegmentSyntax(
+            segment.unexpectedBeforeContent,
+            content: segment.content.reclassifyAsTrailingTrivia(backslashNewlineSuffix, arena: self.arena),
+            segment.unexpectedAfterContent,
+            arena: self.arena
+          )
+        }
+        middleSegments[index] = .stringSegment(segment)
```
bnbarham, line 366: Would become an append (and the one below).

Sources/SwiftParserDiagnostics/MultiLineStringLiteralDiagnoticsGenerator.swift
```diff
@@ -0,0 +1,154 @@
...
+    // Determine kind and position of the diagnonstic
+    var kind: InvalidIndentationInMultiLineStringLiteralError.Kind = .insufficientIdentation
+    var position = token.positionAfterSkippingLeadingTrivia
+
+    let tokenLeadingTrivia = token.leadingTrivia
+
+    let indentationStartIndex = tokenLeadingTrivia.pieces.lastIndex(where: { $0.isNewline })?.advanced(by: 1) ?? tokenLeadingTrivia.startIndex
+    let preIndentationTrivia = Trivia(pieces: tokenLeadingTrivia[0..<indentationStartIndex])
+    let indentationTrivia = Trivia(pieces: tokenLeadingTrivia[indentationStartIndex...])
+    var positionOffset = preIndentationTrivia.reduce(0, { $0 + $1.sourceLength.utf8Length })
```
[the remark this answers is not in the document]
ahoppen: It does indeed. Got into the habit of using `reduce` because in the other cases I was working with `[RawTriviaPiece]`.

Tests/SwiftParserTest/Assertions.swift
```diff
@@ -440,16 +440,32 @@ func AssertDiagnostic<T: SyntaxProtocol>(
...
 /// parsing the resulting `String` as a `SourceFileSyntax`.
 func AssertParse(
   _ markedSource: String,
   substructure expectedSubstructure: Syntax? = nil,
   substructureAfterMarker: String = "START",
-  substructureCheckTrivia: Bool = false,
   diagnostics expectedDiagnostics: [DiagnosticSpec] = [],
   applyFixIts: [String]? = nil,
   fixedSource expectedFixedSource: String? = nil,
+  options: AssertParseOptions = [],
```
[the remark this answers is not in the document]
ahoppen: I would be leading slightly towards having a `SubstructureSpec` with a marker + expected substructure so you could also check multiple substructures. But the need didn’t really come up yet.

Tests/SwiftParserTest/ExpressionTests.swift
```diff
@@ -900,6 +905,76 @@ final class ExpressionTests: XCTestCase {
...
+        """
+        line 1 \
+        line 2
+        """
+      """#,
+      substructure: Syntax(
+        StringLiteralExprSyntax(
+          openQuote: .multilineStringQuoteToken(leadingTrivia: .spaces(2), trailingTrivia: .newline),
+          segments: StringLiteralSegmentsSyntax([
+            .stringSegment(StringSegmentSyntax(content: .stringSegment("line 1 ", leadingTrivia: .spaces(2), trailingTrivia: [.unexpectedText("\\"), .newlines(1)]))),
```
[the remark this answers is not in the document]
ahoppen: It should indeed. `TriviaParser` needed to be taught about backslashes.

Tests/SwiftParserTest/translated/MultilineErrorsTests.swift
```diff
@@ -16,255 +16,355 @@ import XCTest
...
+            Eleven
+            Mu
+            """
+        """#
     )
   }
 
   func testMultilineErrors5() {
-    AssertParse(
+    // \t is not the same as an actual tab for de-indentation
```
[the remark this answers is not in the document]
ahoppen: I just removed the “for de-indentation” part of the comment

Tests/SwiftParserTest/translated/MultilineErrorsTests.swift
```diff
@@ -410,49 +582,44 @@ final class MultilineErrorsTests: XCTestCase {
...
       diagnostics: [
-        // TODO: Old parser expected error on line 1: multi-line string literal content must begin on a new line
-        // TODO: Old parser expected error on line 1: escaped newline at the last line is not allowed, Fix-It replacements: 8 - 9 = ''
-      ]
+        DiagnosticSpec(message: "multi-line string literal content must begin on a new line")
+      ],
+      fixedSource: #"""
+        _ = """
+          \
+          """
```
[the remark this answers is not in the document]
ahoppen: Yes but if you start off with
```
""" \ 
"""
```

I think I don’t care about you. If you want, you can create a specialized diagnostics for it 😜

Tests/SwiftParserTest/translated/MultilineErrorsTests.swift
```diff
@@ -493,4 +660,92 @@ final class MultilineErrorsTests: XCTestCase {
...
+          \(
+          1
+          +
+          2
+          )
+          """
+        """#
+    )
+  }
```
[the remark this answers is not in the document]
ahoppen: I added a test case for that. We accept
```
"foo\(test("""
bar
""") )"
```

just like the C++ parser does

Sources/SwiftSyntax/Raw/RawSyntaxTokenView.swift
```diff
@@ -129,6 +129,34 @@ public struct RawSyntaxTokenView {
...
+  @_spi(RawSyntax)
+  public func leadingTrivia<T>(_ body: (SyntaxText) -> T) -> T {
+    switch raw.rawData.payload {
+    case .parsedToken(let dat):
+      return body(dat.leadingTriviaText)
+    case .materializedToken(let dat):
+      var leadingTriviaStr = Trivia(pieces: dat.leadingTrivia.map(TriviaPiece.init)).description
+      return leadingTriviaStr.withSyntaxText(body)
+    case .layout(_):
+      preconditionFailure("'tokenTrailingRawTriviaPieces' is called on non-token raw syntax")
```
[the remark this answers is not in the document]
ahoppen: Whoops. I updated all the `preconditionFailure` in the file

Sources/SwiftSyntax/Raw/RawSyntax.swift
```diff
@@ -818,12 +873,14 @@ extension RawSyntax: CustomReflectable {
   }
 }
 
-enum RawSyntaxView {
+@_spi(RawSyntax)
```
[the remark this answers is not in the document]
ahoppen: Nothing does. I made these internal again.

Sources/SwiftParser/StringLiterals.swift
ahoppen: Good idea. I also split `postProcessIndentationAndEscapedNewlineOfMiddleSegments` into its own function.

Sources/SwiftParser/StringLiterals.swift
ahoppen: Added a test case

Sources/SwiftParser/StringLiterals.swift
ahoppen: This one is particularly checking that we haven’t introduced leading trivia in one of the post-processing steps before.

I originally also had an assertion that we don’t have any unexpected text here and needed to lift that, which was OK, it just reminded me that something changed.

Sources/SwiftParser/StringLiterals.swift
ahoppen: No, we can't because `TriviaParser` lives in the SyntaxParser module whereas these functions are defined in the SwiftSyntax module. And if we wanted to pull them up into SyntaxParser, we would need to make scary implementation details (`rawData`, `RawTriviaPieceBuffer`, designated factory methods to create raw nodes) public.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

Tests/SwiftParserTest/Assertions.swift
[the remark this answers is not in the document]
bnbarham: Ah yeah, that would work too.

Tests/SwiftParserTest/translated/MultilineErrorsTests.swift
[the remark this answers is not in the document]
bnbarham: If it doesn't take too much time, sure :). Worst case we would then just have another fix it right now, so 🤷

ahoppen:
@swift-ci Please test

Sources/SwiftParser/StringLiterals.swift
bnbarham: Yeah, that one would have to be appended at the end :P. I was expecting this answer, so it's all good 👍

Sources/SwiftParser/StringLiterals.swift
bnbarham: Thought about this more and spoke to Rintaro. Few questions:
1. Could we move this into `RawSyntaxTokenView` since these only apply to tokens?
2. To avoid a whole bunch of code, how would you feel if we always gave back parsed tokens instead? I'd even be fine with just asserting on the materialized case. But this way we could completely avoid having to parse the trivia + just pass in the length as above.

Sources/SwiftSyntax/Raw/RawSyntaxNodeProtocol.swift
```diff
@@ -225,7 +245,179 @@ public struct RawTokenSyntax: RawSyntaxNodeProtocol {
...
+  /// The caller is responsible to delete preceeding trivia from the tree to
+  /// maintain source-fidelity.
+  public func extendingLeadingTrivia(by extendedTrivia: [RawTriviaPiece], arena: SyntaxArena) -> RawTokenSyntax {
+    let extendedTriviaByteLength = extendedTrivia.reduce(0, { $0 + $1.byteLength })
+    switch raw.rawData.payload {
+    case .parsedToken(let dat):
+      assert(
+        String(syntaxText: SyntaxText(baseAddress: dat.wholeText.baseAddress?.advanced(by: -extendedTriviaByteLength), count: extendedTriviaByteLength))
+          == Trivia(pieces: extendedTrivia.map(TriviaPiece.init)).description
+      )
```
rintaro, line 265: This is a little scary considering this is `SwiftSyntax` (instead of `Parser`)

This always succeeds in the _current_ parser, but in general "Assuming that text representing `extendedTrivia` preceeds this token," should be more like "Assuming that all `extendedTrivia` text and the entire `self` text are consecutive in a single _buffer_". Maybe: `text` -> `SyntaxText`? Or we might want to have a fallback to `MaterializedToken` path if that's not the case.
Future incremental parsing might break this assertion too, if we introduce more usages of these functions.

rintaro: To clarify, I'm not saying we should support such situation unnecessarily. I just think we need to clarify what is supported.

ahoppen: I went with @bnbarham’s suggestion and just create materialized tokens for the re-classified cases. This simplifies parsing and allows us to get rid of all these scary methods. https://github.com/apple/swift-syntax/pull/1255/commits/2deaafd0df2d67440cd41513bccf76d2664d1d19

Tests/SwiftSyntaxTest/RawSyntaxTests.swift
```diff
@@ -152,4 +152,66 @@ final class RawSyntaxTests: XCTestCase {
       XCTAssertEqual(barIdent.description, "\nopen ")
     }
   }
+
+  func testExtendingTrivia() {
+    // We are only testing materialized token here because parsed token should
+    // be covered pretty well in the parser. Materialized tokens are less common
+    // and need dedicated testing.
+    withExtendedLifetime(SyntaxArena()) { arena in
```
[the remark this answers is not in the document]
ahoppen: I don’t think that’ correct. `withExtendedLifetime` only extends to the end of the closure. We could put the `withExtendedLifetime` at the end of the test case but AFAIK the indented design of this function is that you put the code that assumes the variable is alive inside the closure.

Sources/SwiftSyntax/Raw/RawSyntaxNodeProtocol.swift
bnbarham: Heh, to be clear my suggestion was to use parsed token and not parse the trivia at all. I don't feel super strongly about this though. Using materialized tokens are a little more heavyweight, I do think it's worth the trade off here. Are you okay with this @rintaro?

Tests/SwiftSyntaxTest/RawSyntaxTests.swift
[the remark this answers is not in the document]
bnbarham: Heh, sorry. I forgot to add the `defer` :)
```
defer withExtendedLifetime(arena) {}
```

ahoppen:
https://github.com/apple/swift-format/pull/480

@swift-ci Please test

PR #1256, opened 2023-01-23
Implement basic DiagnosticFormatter output coloring support

simba909:
As the title implies this PR implements some rudimentary output coloring in the `DiagnosticFormatter` using ANSI color codes. As of this PR, only errors and warnings are colored (using bold red or yellow text respectively). It should be fairly trivial to add further coloring so if you see any low-hanging fruit please let me know!

To get colored output, append the `--colorized` flag to an invocation of `swift-parser-cli print-diags`:
```sh
$ swift-parser-cli print-diags source.swift --colorize
```

Some notes:
 - Being a first-time contributor (hi! 👋🏻 ) I'm not sure about code style in terms of type nesting or visibility of introduced types. I ran the formatter against the code before committing though so that _should_ be good.
 - I did not find any cases of `DiagnosticMessage`s with the `.warning` severity in the codebase, making it hard to test this coloring case. Is there anything I'm missing here?

Lastly, I'm more than happy to make any changes to this code. Excited to finally be contributing directly to the Swift core libraries 🎉

### DIFF
--- a/Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
+++ b/Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
@@ -21,10 +21,28 @@ public struct DiagnosticsFormatter {
   }
 
   /// Number of lines which should be printed before and after the diagnostic message
-  static let contextSize = 2
+  public let contextSize: Int
+
+  /// Whether to colorize formatted diagnostics.
+  public let colorize: Bool
+
+  public init(contextSize: Int = 2, colorize: Bool = false) {
+    self.contextSize = contextSize
+    self.colorize = colorize
+  }
+
+  public static func annotatedSource<SyntaxType: SyntaxProtocol>(
+    tree: SyntaxType,
+    diags: [Diagnostic],
+    contextSize: Int = 2,
+    colorize: Bool = false
+  ) -> String {
+    let formatter = DiagnosticsFormatter(contextSize: contextSize, colorize: colorize)
+    return formatter.annotatedSource(tree: tree, diags: diags)
+  }
 
   /// Print given diagnostics for a given syntax tree on the command line
-  public static func annotatedSource<SyntaxType: SyntaxProtocol>(tree: SyntaxType, diags: [Diagnostic], colorize: Bool = false) -> String {
+  public func annotatedSource<SyntaxType: SyntaxProtocol>(tree: SyntaxType, diags: [Diagnostic]) -> String {
     let slc = SourceLocationConverter(file: "", tree: tree)
 
     // First, we need to put each line and its diagnostics together
@@ -41,7 +59,7 @@ public struct DiagnosticsFormatter {
     let rangesToPrint = annotatedSourceLines.enumerated().compactMap { (lineIndex, sourceLine) -> Range<Int>? in
       let lineNumber = lineIndex + 1
       if !sourceLine.diagnostics.isEmpty {
-        return Range<Int>(uncheckedBounds: (lower: lineNumber - Self.contextSize, upper: lineNumber + Self.contextSize + 1))
+        return Range<Int>(uncheckedBounds: (lower: lineNumber - contextSize, upper: lineNumber + contextSize + 1))
       }
       return nil
     }
@@ -103,25 +121,21 @@ public struct DiagnosticsFormatter {
         }
 
         for diag in diags.dropLast(1) {
-          if colorize {
-            annotatedSource.append("\(preMessage)├─ \(self.colorize(diag.diagMessage))\n")
-          } else {
-            annotatedSource.append("\(preMessage)├─ \(diag.message)\n")
-          }
+          annotatedSource.append("\(preMessage)├─ \(colorizeIfRequested(diag.diagMessage))\n")
         }
-
-        if colorize {
-          annotatedSource.append("\(preMessage)╰─ \(self.colorize(diags.last!.diagMessage))\n")
-        } else {
-          annotatedSource.append("\(preMessage)╰─ \(diags.last!.message)\n")
-        }
+        annotatedSource.append("\(preMessage)╰─ \(colorizeIfRequested(diags.last!.diagMessage))\n")
       }
     }
     return annotatedSource
   }
 
-  /// Annotates the given ``DiagnosticMessage`` with an appropriate ANSI color code and returns the result as a printable string.
-  static func colorize(_ message: DiagnosticMessage) -> String {
+  /// Annotates the given ``DiagnosticMessage`` with an appropriate ANSI color code (if the value of the `colorize`
+  /// property is `true`) and returns the result as a printable string.
+  private func colorizeIfRequested(_ message: DiagnosticMessage) -> String {
+    guard colorize else {
+      return message.message
+    }
+
     switch message.severity {
     case .error:
       let annotation = ANSIAnnotation(color: .red, trait: .bold)
--- a/Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
+++ b/Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
@@ -133,9 +133,11 @@ public struct DiagnosticsFormatter {
       return message.message
     }
   }
+}
 
-  enum ANSIColor: UInt8 {
-    case reset = 0
+struct ANSIAnnotation {
+  enum Color: UInt8 {
+    case normal = 0
     case black = 30
     case red = 31
     case green = 32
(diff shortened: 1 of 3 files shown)


Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
```diff
@@ -103,12 +103,76 @@ public struct DiagnosticsFormatter {
         }
 
         for diag in diags.dropLast(1) {
-          annotatedSource.append("\(preMessage)├─ \(diag.message)\n")
+          if colorize {
+            annotatedSource.append("\(preMessage)├─ \(self.colorize(diag.diagMessage))\n")
+          } else {
+            annotatedSource.append("\(preMessage)├─ \(diag.message)\n")
+          }
```
ahoppen, line 110: What do you think about making all the functions on `DiagnosticFormatter` non-`static`. That way `colorize: Bool` could be a property on `DiagnsoticsFormatter`, you could have `func colorizeIfWanted(_ message: DiagnosticMessage)` and this would become the following where we don’t need to repeat the `preMessage` and box drawing character.

```suggestion
          annotatedSource.append("\(preMessage)├─ \(self.colorizeIfRequested(diag.diagMessage))\n")
```

As an added bonus, this would allow callers of the `DiagnosticFormatter` to also control the context size.

The entry to the diagnostic formatter could still be a static function that looks something like the following:

```swift
public static func annotatedSource<SyntaxType: SyntaxProtocol>(tree: SyntaxType, diags: [Diagnostic], contextSize: Int = 2, colorize: Bool = false) -> String {
  let formatter = DiagnosticFormatter(contextSize: contextSize, colorize: colorize)
  return formatter.annotatedSource(tree: tree, diags: diags)
}
```

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
```diff
@@ -103,12 +103,76 @@ public struct DiagnosticsFormatter {
...
+    case cyan = 36
+    case white = 37
+  }
+
+  enum ANSITrait: UInt8 {
+    case bold = 1
+    case underline = 4
+  }
+
+  struct ANSIAnnotation {
```
ahoppen, line 154: I think you could make `ANSIAnnotation` a top-level the and nest `ANSIColor` and `ANSITrait` in it. It somehow feels weird to me to nest this in the `DiagnosticsFormatter`.

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
```diff
@@ -103,12 +103,76 @@ public struct DiagnosticsFormatter {
...
+    case green = 32
+    case yellow = 33
+    case blue = 34
+    case magenta = 35
+    case cyan = 36
+    case white = 37
+  }
+
+  enum ANSITrait: UInt8 {
+    case bold = 1
```
ahoppen, line 150: If you’ve got the `reset` color, I think you should also have `ANSITrait.reset`. Or alternatively remove `ANSIColor.reset` and use nil to represent resetting the color. I think I slightly prefer using `nil` but don’t really have strong opinions. It would just be good to have this be consistent.

ahoppen (commented):
This is great stuff, thank you! I’ve got a few comments inline to tweak the code design but otherwise this looks good to me.

And you are right, the parser doesn’t currently output any warnings. The warning diagnostic kind just exists for completeness sake at the moment. It’s fine if we don’t test warning coloring at the moment.

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
simba909: Yeah I think this is a good improvement; the static-ness actually caught me off guard initially but I didn't want to make too many changes so left it where it was. I'll refactor the `DiagnosticFormatter` in another commit to this PR 👍🏻

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
simba909: Agree with this as well. What made me go for nesting was that I wasn't sure how useful this "supporting" type would be to other parts of the codebase but I'm more than happy to decouple it. Slight bike-shedding question but does that warrant its own file or should it stay within `DiagnosticFormatter.swift`? If I'm nesting the color and trait types anyways, would you agree that the `ANSI` prefix becomes kinda superfluous and that it can be dropped?

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
simba909: Agree on the consistency note. What about introducing the reset (I've noticed the alternative naming `normal`) case on `ANSITrait` and having that as the default? Maybe it would make more sense to have that and `ANSIColor.reset` just be called `normal` in that case though.

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
ahoppen: Don't be afraid to refactor all the things if it makes the API or implementation simpler. 😄

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
ahoppen: I don't have strong opinions since I don't anticipate any further users of it in SwiftDiagnostics. I'll leave it up to you.

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
ahoppen: Yes, I think having `ANSITrait.reset` would be equally good. And `normal` sounds better to me as well. Makes it sound less like a state machine that you need to reset.

simba909:
Thank you @ahoppen for the review 👏🏻  I've implemented the requested changes, please have another look when you have the time

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
```diff
@@ -103,12 +121,73 @@ public struct DiagnosticsFormatter {
...
+    self.trait = trait
+  }
+
+  func applied(to message: String) -> String {
+    // Resetting after the message ensures that we don't color unintended lines in the output
+    return "\(code)\(message)\(ANSIAnnotation.reset.code)"
+  }
+
+  /// The "reset" ANSI code used to unset any previously added annotation.
+  static var reset: ANSIAnnotation {
```
ahoppen, line 190: To be consistent with `normal` in `Color` and `Trait`, should this also be named `normal`?

ahoppen (commented):
This looks good. I’ve got one more naming suggestion, other than that I think it’s ready to be merged. 

It would be nice if we could auto-detect if the terminal supports color but from a quick investigation I haven’t found an easy way to check that, so I think the explicit option is the best option for now.

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
simba909: Sure, that sounds good to me 👍🏻

ahoppen (approved):
Thanks for this improvement and welcome to the club of Swift contributors.

ahoppen:
@swift-ci Please test

DougGregor:
This is fantastic @simba909 , thank you! I'll wire it up in the compiler soon.

DougGregor:
@swift-ci Please test

DougGregor:
@swift-ci Please test macOS

DougGregor:
@swift-ci Please test macOS

ahoppen:
@simba909 In case you are interested in working a little more on this and want to auto-detect if we should colorize the output: I just noticed that swift-format is determining whether to enable coloring depending on whether stderr is connected to a terminal

https://github.com/apple/swift-format/blob/c5bec02db7c09adde5485df0f51100edd286c387/Sources/swift-format/Utilities/StderrDiagnosticPrinter.swift#L49-L62

`isTTY` is defined here

https://github.com/apple/swift-tools-support-core/blob/58bb41c30c02ea4f46afdd31d90339ed0058c35f/Sources/TSCBasic/TerminalController.swift#L109-L122

I think we could to the same for swift-syntax. I would like to avoid pulling in swift-tools-support-core as a dependency though, so we would need to copy the necessary methods to swift-syntax.

simba909:
@ahoppen I'd love to! Let me see what I can do.

DougGregor:
I added the compiler-side code. I do think we should prepend strings like "warning: " and "error: " in the diagnostic output, because not everyone can see color. We might want to steal the colors and scheme used currently in the compiler.

simba909:
@ahoppen and @DougGregor: I've had a go at implementing auto-detection of a terminal environment in #1280, which also adds the string prefixing suggestions by Doug.

PR #1257, opened 2023-01-23
Fix performance regressions introduced by my recent changes

ahoppen:
This improves parsing performance by ~2.5x. Parsing is still about 1.5x slower compared to before https://github.com/apple/swift-syntax/pull/1173 but given that I introduced quite a bit of necessary state and the fact that we still parse all files in [MovieSwiftUI](https://github.com/Dimillian/MovieSwiftUI) in less than 15ms I hope that’s acceptable. If we want to improve performance further, we just need to do some more investigation.

It’s probably best to review this commit by commit. All commits are standalone and have a description of what they do. I can split any commit that’s controversial to its own PR.

### DIFF
--- a/Sources/SwiftParser/Lexer/Cursor.swift
+++ b/Sources/SwiftParser/Lexer/Cursor.swift
@@ -80,7 +80,7 @@ extension Lexer.Cursor {
     /// `(` that opens the string interpolation.
     ///
     /// `stringInterpolationStart` points to the first character inside the interpolation.
-    case inStringInterpolation(stringLiteralKind: StringLiteralKind, parenCount: Int, stringInterpolationStart: Lexer.Cursor)
+    case inStringInterpolation(stringLiteralKind: StringLiteralKind, parenCount: Int)
 
     /// We have parsed a string interpolation segment and are now expecting the closing `)`.
     case afterStringInterpolation
@@ -95,7 +95,7 @@ extension Lexer.Cursor {
       case .afterStringLiteral: return nil
       case .afterClosingStringQuote: return nil
       case .inStringInterpolationStart: return nil
-      case .inStringInterpolation(let stringLiteralKind, _, _):
+      case .inStringInterpolation(let stringLiteralKind, _):
         // Single line strings cannot span multiple lines, so we don't want to
         // consume any newline inside a string interpolation either.
         switch stringLiteralKind {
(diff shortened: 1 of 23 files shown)


ahoppen:
@swift-ci Please test

Sources/SwiftParser/Lexer/Cursor.swift
```diff
@@ -133,11 +133,61 @@ extension Lexer.Cursor {
...
+    }
+
+    mutating func perform(stateTransition: Lexer.StateTransition, stateAllocator: BumpPtrAllocator) {
+      switch stateTransition {
+      case .push(newState: let newState):
+        if let topState = topState {
+          if let stateStack = stateStack {
+            let newStateStack = stateAllocator.allocate(State.self, count: stateStack.count + 1)
+            let (_, existingStateStackEndIndex) = newStateStack.initialize(from: stateStack)
+            newStateStack[existingStateStackEndIndex] = topState
```
rintaro, line 166: I don't think `topState` should be in `newStateStack`.
for top-normal: `topState=nil`, `stateStack=nil`
for top-some: `topState=.some`, `stateStack=nil`
for non-top: `topState=.some`, `stateStack=[...]`

Sources/SwiftParser/Lexer/Cursor.swift
```diff
@@ -133,11 +133,61 @@ extension Lexer.Cursor {
...
+    var currentState: State {
+      return topState ?? .normal
+    }
+
+    mutating func perform(stateTransition: Lexer.StateTransition, stateAllocator: BumpPtrAllocator) {
+      switch stateTransition {
+      case .push(newState: let newState):
+        if let topState = topState {
+          if let stateStack = stateStack {
+            let newStateStack = stateAllocator.allocate(State.self, count: stateStack.count + 1)
```
rintaro, line 164: It's a little unfortunate we don't reuse previously-allocated buffer. For example
`push`, `push`, `pop`, `push`. operation should reuse the buffer allocated at the second `push`, instead of allocating a whole new buffer.

Sources/SwiftParser/Lexer/Cursor.swift
```diff
@@ -133,11 +133,61 @@ extension Lexer.Cursor {
...
+            self.stateStack = UnsafeBufferPointer(newStateStack)
+          } else {
+            let newStateStack = stateAllocator.allocate(State.self, count: 1)
+            newStateStack[0] = topState
+            self.stateStack = UnsafeBufferPointer(newStateStack)
+          }
+        }
+        topState = newState
+      case .replace(newState: let newState):
+        topState = newState
```
rintaro, line 176: _If_ we place the `topState` into `stateStack`, we need to replace the `stateStack.last` too.
otherwise `push` `replace` `push` `pop` operation doesn't restore the replaced state.

Sources/SwiftParser/Lexer/Cursor.swift
ahoppen: That’s exactly what we do (or I’m completely misunderstanding what you’re trying to say). If you don’t count the bottom `normal` state as an entry in the stack size we have

- Stack size 0: `topState=nil`, `stateStack=nil`
- Stack size 1: `topState=.some`, `stateStack=nil`
- Stack size >1 `topState=.some`, `stateStack=[…]`

If we are in this `if`, then we have stack size >=1 (since `topState != nil`). So what we do is to transfer `topState` into `newStateStack` so that we can assign `newState` to `topState` (which now lives in `stateStack`)

Sources/SwiftParser/Lexer/Cursor.swift
ahoppen: Yes, but how do you know that you have already seen this exact stack? We could try to create a dictionary for it, but given that the bump allocator doesn’t allocate more than 250 byte for any file in MovieSwiftUI, I’m not sure if there’s much to be gained here.

Sources/SwiftParser/Lexer/Cursor.swift
ahoppen: When I implemented this, I measured and found that storing `topState` inline gives the best performance because that way we don’t need to allocate a new `stateStack` for every string literal (only for every interpolation, which I believe should be rarer). I also tried storing the top two states inline (so that we only need to allocate memory in the bump allocator for string literals inside interpolations) but that was slightly slower.

Sources/SwiftParser/Lexer/Cursor.swift
rintaro: Sorry I misread your PR. Ignore this.

Sources/SwiftParser/Lexer/Cursor.swift
rintaro: Ditto. Please ignore this.

rintaro (approved):

Sources/SwiftParser/Lexer/Cursor.swift
rintaro: I am convinced that we can't reuse the memory region since other  (copied) cursor might still have a reference to it.

PR #1263, opened 2023-01-25
Fix an assertion failure if a string literal inside string interpolation is unterminated

ahoppen:
We don't eat newlines as leading trivia in string interpolation of single line strings but  `lexNormal` expects newlines to already be eaten. If we reach a newline inside string interpolation of a single-line string, emit an empty string segment to indicate to the parser that the string has ended and pop out of string interpolation.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #1261, opened 2023-01-24
Add init with `PartialSyntaxNodeString` on `SwitchStmtSyntax`

kimdv:

PR #1264, opened 2023-01-25
Add node with body to variable decl

kimdv:
Depends on #1261

### DIFF
--- a/Sources/SwiftSyntaxBuilder/SyntaxNodeWithBody.swift
+++ b/Sources/SwiftSyntaxBuilder/SyntaxNodeWithBody.swift
@@ -123,3 +123,17 @@ public extension SwitchStmtSyntax {
     self.cases = casesBuilder()
   }
 }
+
+// MARK: - VariableDeclSyntax
+// VariableDeclSyntax is a special scenario as it don't have body or members
+// So we cannot conform to `HasTrailingCodeBlock` or `HasTrailingMemberDeclBlock`
+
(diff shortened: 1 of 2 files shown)


Sources/SwiftSyntaxBuilder/SyntaxNodeWithBody.swift
```diff
@@ -123,3 +123,16 @@ public extension SwitchStmtSyntax {
...
+// MARK: - VariableDeclSyntax
+// VariableDeclSyntax is a special scenario as it don't have body or members
+// So we cannot conform to `HasTrailingCodeBlock` or `HasTrailingMemberDeclBlock`
+
+public extension VariableDeclSyntax {
+  init(_ signature: PartialSyntaxNodeString, @CodeBlockItemListBuilder accessor: () -> CodeBlockItemListSyntax) {
+    self = "\(signature) {}"
+    var binding: PatternBindingSyntax? = self.bindings.last
+    binding?.accessor = .getter(CodeBlockSyntax(statements: accessor()))
+    bindings = PatternBindingListSyntax([binding].compactMap { $0 })
```
ahoppen, line 136: I would add an `assert(self.bindings.count == 1)` after `self = "\(signature) {}" ` because we would be dropping all but the last binding here.

ahoppen (approved):

kimdv:
@swift-ci please test

PR #1267, opened 2023-01-25
Remove pound keywords token kinds

ahoppen:
These are no longer being used, so we can clean up the list of token kinds a little further.


ahoppen:
@swift-ci Please test

gyb_syntax_support/Token.py
```diff
@@ -227,17 +227,6 @@ def macro_name(self):
...
-    PoundKeyword('PoundKeyPath', 'keyPath', text='#keyPath'),
-    PoundKeyword('PoundLine', 'line', text='#line'),
-    PoundKeyword('PoundSelector', 'selector', text='#selector'),
-    PoundKeyword('PoundFile', 'file', text='#file'),
-    PoundKeyword('PoundFileID', 'fileID', text='#fileID'),
-    PoundKeyword('PoundFilePath', 'filePath', text='#filePath'),
-    PoundKeyword('PoundColumn', 'column', text='#column'),
-    PoundKeyword('PoundFunction', 'function', text='#function'),
-    PoundKeyword('PoundDsohandle', 'dsohandle', text='#dsohandle'),
     PoundKeyword('PoundAssert', 'assert', text='#assert'),
```
DougGregor, line 230: `#assert` should be removable as well.

DougGregor (approved):
Thank you!

gyb_syntax_support/Token.py
ahoppen: We are still using it here

https://github.com/apple/swift-syntax/blob/3ca3c0bc15124009119d490e9aba998332f5dff5/Sources/SwiftParser/Statements.swift#L119-L120

DougGregor: Alright, let's get rid of that later. It can be parsed as a freestanding macro expression.

PR #1268, opened 2023-01-25
[Macros] Freestanding macros expand to CodeBlockItemList; parse top-level macro expansion decls

rxwei:
- Change the return type of `FreestandingDeclarationMacro.expanion(of:in:)` to ``CodeBlockItemListSyntax` to allow expressions, statements, and declarations.
- Parse freestanding macro expansions as declarations rather than expressions when it's not part of a statement or expression (i.e. "top-level").

### DIFF
--- a/Sources/_SwiftSyntaxMacros/CodeItemMacro.swift
+++ b/Sources/_SwiftSyntaxMacros/CodeItemMacro.swift
@@ -16,5 +16,5 @@ public protocol CodeItemMacro: FreestandingMacro {
   static func expansion(
     of node: MacroExpansionDeclSyntax,
     in context: inout MacroExpansionContext
-  ) throws -> [CodeBlockItemSyntax]
+  ) throws -> CodeBlockItemListSyntax
 }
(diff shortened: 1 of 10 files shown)


Sources/_SwiftSyntaxMacros/CodeItemMacro.swift
```diff
@@ -16,5 +16,5 @@ public protocol CodeItemMacro: FreestandingMacro {
   static func expansion(
     of node: MacroExpansionDeclSyntax,
     in context: inout MacroExpansionContext
-  ) throws -> [CodeBlockItemSyntax]
+  ) throws -> CodeBlockItemListSyntax
```
ahoppen, line 19: I discussed this with @DougGregor yesterday. If you are intending to merge these items with other items, this should return an array IMO because arrays are easier to work with (you can append, insert, delete, …). IMO you should only create a `CodeBlockItemListSyntax` once you think the item list is complete.

(similar in DeclarationMacros.swift)

rxwei: Good point! In that case this PR is no longer necessary. `DeclarationMacro` still returns decls, `ExpressionMacro` still returns exprs, and `CodeItemMacro` returns `[CodeBlockItemSyntax]`

PR #1269, opened 2023-01-25
Don’t automaticaly conform all decl/stmt/etc. nodes to `ExpressibleByStringLiteral`

ahoppen:
This removes the implicit expressibility by string literals for all decl/stmt/... nodes, which parsed the node as a decl/stmt/... and then cast it to the expected type.

rdar://104090218


ahoppen:
@swift-ci Please test

Sources/SwiftSyntaxBuilder/SyntaxNodeWithBody.swift
```diff
@@ -12,6 +12,8 @@
 
 import SwiftSyntax
 
+// MARK: - Error
+
```
kimdv, line 16: Is there a reason for this mark?

Sources/SwiftSyntaxBuilder/SyntaxNodeWithBody.swift
```diff
@@ -39,16 +41,22 @@ extension SyntaxStringInterpolation {
...
+    guard let castedStmt = stmt.as(Self.self) else {
+      throw SyntaxStringInterpolationError.producedInvalidNodeType(expectedType: IfStmtSyntax.self, actualNode: stmt)
+    }
+    self = castedStmt
     self.body = CodeBlockSyntax(statements: bodyBuilder())
   }
 }
 
-extension CatchClauseSyntax: HasTrailingCodeBlock {}
+//extension CatchClauseSyntax: HasTrailingCodeBlock {}
```
kimdv, line 59: Should this be added or removed?

ahoppen: Oh, I forgot about this. It should be added. Thanks

Sources/SwiftSyntaxBuilder/SyntaxNodeWithBody.swift
ahoppen: No, I ended not adding an error in this file.

kimdv (approved):

CodeGeneration/Sources/SyntaxSupport/Node.swift
```diff
@@ -24,6 +24,7 @@ public class Node {
   public let description: String?
   public let baseKind: String
   public let parserFunction: String?
+  public let expressibleByStringLiteral: Bool
```
bnbarham, line 27: I'm not really sure I see the point of separating these. If we think something is worth having a public parse method for then it's probably also worth being expressible by string interpolation as well. I imagine this bumps up the count of non-base nodes though :(.

Sources/SwiftSyntaxBuilder/SyntaxNodeWithBody.swift
```diff
@@ -39,16 +39,31 @@ extension SyntaxStringInterpolation {
...
 }
 
-extension CatchClauseSyntax: HasTrailingCodeBlock {}
+extension CatchClauseSyntax: HasTrailingCodeBlock {
+  public init(_ signature: PartialSyntaxNodeString, @CodeBlockItemListBuilder bodyBuilder: () -> CodeBlockItemListSyntax) throws {
+    let stmt = CatchClauseSyntax("\(signature) {}")
+    guard let castedStmt = stmt.as(Self.self) else {
+      throw SyntaxStringInterpolationError.producedInvalidNodeType(expectedType: IfStmtSyntax.self, actualNode: stmt)
+    }
+    self = castedStmt
```
bnbarham, line 63: It is already a `CatchClauseSyntax` 😅

Sources/SwiftSyntaxBuilder/SyntaxNodeWithBody.swift
```diff
@@ -39,16 +39,31 @@ extension SyntaxStringInterpolation {
...
+public extension HasTrailingCodeBlock where Self: StmtSyntaxProtocol {
+  init(_ signature: PartialSyntaxNodeString, @CodeBlockItemListBuilder bodyBuilder: () -> CodeBlockItemListSyntax) throws {
+    let stmt = StmtSyntax("\(signature) {}")
+    guard let castedStmt = stmt.as(Self.self) else {
+      throw SyntaxStringInterpolationError.producedInvalidNodeType(expectedType: IfStmtSyntax.self, actualNode: stmt)
+    }
+    self = castedStmt
     self.body = CodeBlockSyntax(statements: bodyBuilder())
   }
 }
```
bnbarham, line 55: Still not a huge fan of this one, but if you really want it `expectedType: IfStmtSyntax.self` is likely not what you intended :P. `Self.self`? Also odd that we use `StmtSyntax`, I assume that's because the majority of these are statements. Could split it into `StmtHasTrailing...` and `DeclHasTrailing...`?

Sources/SwiftSyntaxBuilder/SyntaxNodeWithBody.swift
```diff
@@ -39,16 +39,31 @@ extension SyntaxStringInterpolation {
...
 
-extension CatchClauseSyntax: HasTrailingCodeBlock {}
+extension CatchClauseSyntax: HasTrailingCodeBlock {
+  public init(_ signature: PartialSyntaxNodeString, @CodeBlockItemListBuilder bodyBuilder: () -> CodeBlockItemListSyntax) throws {
```
bnbarham, line 58: `signature` seems weird to use here (and for the others).

CodeGeneration/Sources/SyntaxSupport/Node.swift
ahoppen: My initial idea was that we could keep the list of exceptions for non-base nodes really small (<5 let’s say). But I’ve got a feeling that I’m loosing that battle in this PR already. I’ve removed that property again.

Sources/SwiftSyntaxBuilder/SyntaxNodeWithBody.swift
ahoppen: Who doesn’t love copy-paste errors

Sources/SwiftSyntaxBuilder/SyntaxNodeWithBody.swift
ahoppen: > Still not a huge fan of this one, but if you really want it `expectedType: IfStmtSyntax.self` is likely not what you intended :P. `Self.self`? 

Oh, absolutely. Yes. Again, copy-paste error from the `IfStmt` below.

Also odd that we use `StmtSyntax`, I assume that's because the majority of these are statements. Could split it into `StmtHasTrailing...` and `DeclHasTrailing...`?

Sources/SwiftSyntaxBuilder/SyntaxNodeWithBody.swift
ahoppen: What do you think about `header`? Since the rest is the body and HTML has header and body? Or do you have better ideas?

ahoppen:
@swift-ci Please test

Sources/SwiftSyntaxBuilder/SyntaxNodeWithBody.swift
bnbarham: I like that better than `signature`, yeah.

bnbarham (approved):

ahoppen:
@swift-ci Please test

ahoppen:
https://github.com/apple/swift/pull/63251

@swift-ci Please test macOS

PR #1270, opened 2023-01-25
Shorten getters and setters of syntax nodes

ahoppen:
This makes the `Syntax*Nodes.swift` more succinct and easier to read


kimdv (approved):

ahoppen:
@swift-ci Please test

bnbarham (approved):
Nice 🤩

Sources/SwiftSyntax/SyntaxNodes.swift.gyb.template
```diff
@@ -187,21 +187,14 @@ public struct ${node.name}: ${base_type}Protocol, SyntaxHashable {
 %       end
   public var ${child.swift_name}: ${child_type}${optional_mark} {
     get {
```
bnbarham, line 189: Could go even furtherrrr if you wanted to, ie. get/set on the same line as the body. Or does swift-format not like that?

PR #1274, opened 2023-01-26
Store lexer error offset relative to leading trivia start

ahoppen:
We need this so we can diagnose invalid UTF-8 in a token’s leading trivia. Also, it seems cleaner in general to store the lexer error relative to the token’s leading trivia start instead of the token’s text start.

### DIFF
--- a/Sources/SwiftParser/Lexer/Cursor.swift
+++ b/Sources/SwiftParser/Lexer/Cursor.swift
@@ -255,13 +255,15 @@ extension Lexer {
   struct Result {
     let tokenKind: RawTokenKind
     let flags: Lexer.Lexeme.Flags
-    let error: LexerError?
+    /// The error kind and the cursor pointing to the character at which the
+    /// error occurred
+    let error: (kind: LexerError.Kind, position: Lexer.Cursor)?
     let stateTransition: StateTransition?
 
     init(
       _ tokenKind: RawTokenKind,
       flags: Lexer.Lexeme.Flags = [],
-      error: LexerError? = nil,
+      error: (kind: LexerError.Kind, position: Cursor)? = nil,
       stateTransition: StateTransition? = nil
     ) {
       self.tokenKind = tokenKind
@@ -335,10 +337,14 @@ extension Lexer.Cursor {
       flags.insert(.isAtStartOfLine)
     }
 
+    let error = result.error.map { error in
+      return LexerError(error.kind, byteOffset: cursor.distance(to: error.position))
+    }
+
     return .init(
       tokenKind: result.tokenKind,
       flags: flags,
-      error: result.error,
+      error: error,
       start: leadingTriviaStart.pointer,
       leadingTriviaLength: leadingTriviaStart.distance(to: textStart),
       textLength: textStart.distance(to: trailingTriviaStart),
@@ -1194,11 +1200,11 @@ extension Lexer.Cursor {
       let oConsumed = self.advance(matching: "o")  // Consome 'o'
       assert(zeroConsumed && oConsumed)
       if let peeked = self.peek(), peeked < UInt8(ascii: "0") || peeked > UInt8(ascii: "7") {
-        let errorOffset = tokenStart.distance(to: self)
+        let errorPos = self
         self.advance(while: { $0.isValidIdentifierContinuationCodePoint })
         return Lexer.Result(
           .integerLiteral,
-          error: LexerError(.invalidOctalDigitInIntegerLiteral, byteOffset: errorOffset)
+          error: (.invalidOctalDigitInIntegerLiteral, errorPos)
         )
       }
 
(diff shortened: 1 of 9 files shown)


ahoppen:
@swift-ci Please test

Sources/SwiftSyntax/Raw/RawSyntaxTokenView.swift
```diff
@@ -65,6 +65,29 @@ public struct RawSyntaxTokenView {
     }
   }
 
+  @_spi(RawSyntax)
+  public func wholeText<T>(_ body: (SyntaxText) -> T) -> T {
```
bnbarham, line 69: Could we instead have a function to get the `UnicodeScalar` for a particular offset?

bnbarham (approved):

Sources/SwiftSyntax/Raw/RawSyntaxTokenView.swift
ahoppen: The current diagnostics only reference a single character but I could imagine that future diagnostics might reference an entire substring, e.g. from the offset to the end of the token, so I think the added flexibility of this API is good.

ahoppen:
@swift-ci Please test macOS

Sources/SwiftSyntax/Raw/RawSyntaxTokenView.swift
bnbarham: That's true... in that case Doug added a `withEachSyntaxText` that you could use instead. It's a ... little more involved, but would avoid the temporary buffer creation (well, we still make it for leading/trailing).

Alternatively given this is only ever used in the error case, you could just use `syntaxTextBytes` (which uses `withEachSyntaxText` but writes it into an array).

This function is just sort of half way, which I don't really see the point of. Either we care about performance, in which case we should use `withEachSyntaxText`. Or we don't and thus `syntaxTextBytes` is simpler.

ahoppen: Using `syntaxTextBytes` sounds good to me.

ahoppen:
@swift-ci Please test

Sources/SwiftParserDiagnostics/LexerDiagnosticMessages.swift
```diff
@@ -98,9 +98,11 @@ public extension SwiftSyntax.LexerError {
   /// `tokenText` is the entire text of the token in which the `LexerError`
   /// occurred, including trivia.
   @_spi(RawSyntax)
-  func diagnostic(wholeText: SyntaxText) -> DiagnosticMessage {
+  func diagnostic(wholeTextBytes: [UInt8]) -> DiagnosticMessage {
     var scalarAtErrorOffset: UnicodeScalar {
-      Unicode.Scalar(wholeText[Int(self.byteOffset)])
+      // Fall back to the Unicode replacement character U+FFFD in case we can't
+      // lex the unicode character at `byteOffset`. It's the best we can do
+      Unicode.Scalar.lexing(from: wholeTextBytes[Int(self.byteOffset)...]) ?? UnicodeScalar("�")
```
bnbarham, line 105: The other option would be to display the hex value.

Sources/SwiftParser/Lexer/Cursor.swift
```diff
@@ -674,62 +674,7 @@ extension Lexer.Cursor {
...
-    // Number of bits in the value, ignoring leading zeros.
-    let numBits = 32 - charValue.leadingZeroBitCount
-    if numBits <= 5 + 6 {
-      return encodedBytes == 2 ? Unicode.Scalar(charValue) : nil
-    }
-    if numBits <= 4 + 6 + 6 {
-      return encodedBytes == 3 ? Unicode.Scalar(charValue) : nil
-    }
-    return encodedBytes == 4 ? Unicode.Scalar(charValue) : nil
+    return Unicode.Scalar.lexing(advance: { self.advance() }, peek: { self.peek(at: 0) })
```
bnbarham, line 677: Any slowdown on this or is it optimized it fine?

ahoppen: Thanks for asking. I needed to add another `@inline(__always)` to fix a performance regression.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

Sources/SwiftParserDiagnostics/LexerDiagnosticMessages.swift
ahoppen: Let’s stick with the replacement character. I don’t really expect anyone to hit this anyway and if they do, I don’t think a hex value would help them any more.

rintaro (approved):

PR #1277, opened 2023-01-26
Parse `0.2` after `…` as float literal, not member access

ahoppen:
Deciding whether `0.2` should be lexed as a float literal or a member access is a little more difficult than just looking at the previous character because `0.2` might be preceeded by an operator like `…` or `.^.`, in which case it should be lexed as a float literal and not a member access.

We might be able to do some disambiguation magic on whether the character before the period is also an operator continuation point but that seems fairly brittle to me. The sanest way of doing this, is to store the previously lexed token’s kind in the cursor and checking that.

I measured and did not see a performance regregssion when parsing MovieSwiftUI.

rdar://103273988

### DIFF
--- a/Sources/SwiftParser/Lexer/Cursor.swift
+++ b/Sources/SwiftParser/Lexer/Cursor.swift
@@ -201,6 +201,8 @@ extension Lexer {
   struct Cursor {
     var input: UnsafeBufferPointer<UInt8>
     var previous: UInt8
+    /// If we have already lexed a token, the kind of the previously lexed token
+    var previousTokenKind: RawTokenBaseKind?
     private var stateStack: StateStack = StateStack()
 
     init(input: UnsafeBufferPointer<UInt8>, previous: UInt8) {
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

Sources/SwiftParser/Lexer/Cursor.swift
```diff
@@ -335,6 +337,8 @@ extension Lexer.Cursor {
       flags.insert(.isAtStartOfLine)
     }
 
+    self.previousTokenKind = result.tokenKind.base
```
bnbarham, line 340: There's `Cursor.backUp` where `previousTokenKind` wouldn't be correct any more. We also don't set the state in there, but I suppose the assumption there is that we're always in `normal` when splitting tokens.

ahoppen: `Cursor.backUp` is still correct because the previous token still has the same kind if we `resetForSplit`, which is the only caller of this function (and should remain to be the only caller). 

If we have just lexed `<..` as a binary operator, `previousToken` is `binaryOperator`. If we now back up by two characters to consume only `<`, we are now placed at the first `.` but the previous token continues to be of kind `binaryOperator`. So no change of `previousTokenKind`  is necessary here.

bnbarham (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Linux

ahoppen:
@swift-ci Please test Linux

Issue #778, opened 2022-09-11
Unterminated string literal causes mis-parse of subsequent nodes


DougGregor:
### Issue Kind

Other

### Source Code

```swift
"This is unterminated
x
```

### Description

When we parse the above, we get unexpected node syntax:

```
SourceFileSyntax children=3
  0: CodeBlockItemListSyntax
  1: UnexpectedNodesSyntax children=2
    0: unknown("\"This is unterminated")
    1: identifier("x")
  2: eof
```

rather than what we'd hope for, which is a string literal with missing close quote and an identifier expression.

ahoppen:
rdar://99789794

CodaFi:
This issue is symptomatic of the C++ parser's token design leaking into ours. A string literal ought to have its delimiters and quotes as separate tokens so we can recover by `expect`ing them in the parser. Instead the entire blob is handed to us via the lexer which has ostensibly done some amount of validation on it. Then we're left to re-lex and re-parse it to get the interpolation segments. I'd much prefer the lexer stay out of the business of structural validation like this so we're not repeating it.

ahoppen:
This has been fixed by https://github.com/apple/swift-syntax/pull/1192. I’m adding a test case for the example mentioned above in https://github.com/apple/swift-syntax/pull/1279.

PR #1279, opened 2023-01-26
Add substructure test case for unterminated string literal

ahoppen:
Adds a test case for apple/swift-syntax#778


ahoppen:
@swift-ci Please test

PR #1280, opened 2023-01-26
Implement support for automatically coloring output in DiagnosticsFormatter

simba909:
This builds on the work in #1256 to add support for automatically enabling output coloring as suggested by @ahoppen. As of this change when invoking `$ swift-parser-cli print-diags` in a terminal it will enable output coloring by default. The `--colorize` flag introduced in #1256 is still supported to force output colors, if desired.

API-wise this should be a non-breaking change since it only adds overloads to instantiate a `DiagnosticsFormatter` with a new `ColorMode` parameter (shamelessly stolen from swift-format).

Some notes:
 - This implementation only copies the absolutely minimum required components from swift-format and swift-tools-support-core to enable the feature. I'm happy to discuss moving more things over from TSC / laying the groundwork for further development if that improves the end result.
   - In particular, TSC wraps `stderr` with a few wrapper types to simplify reading/writing to it. This doesn't seem to be used for detecting a TTY environment which is why I left it out for now.
 - I noticed that swift-tools-support-core exposes a `TSCLibc` module that re-exports symbols from C standard libraries on the various platforms (Glibc for Linux, CRT on Windows and Darwin.C on Apple platforms) but opted to "inline" the use of `stderr` for now. Don't have the resources at the moment to test this cross platform and so would more than appreciate any pointers here.
 - The implementation in TSCBasic uses a `ProcessEnv` type to read properties from the environment (relevant in this case is the `ProcessEnv.vars["TERM"]` call to check whether it's set to `"dumb"`) on non-Windows platforms. I've opted to remove support for detecting this case for now since copying `ProcessEnv` into swift-syntax seems out of scope but please do let me know if there's anything I'm missing here.

Additionally, I've taken the liberty of adding "error:" and "warning:" prefixes to formatted output when coloring is enabled to help with accessibility, as suggested by @DougGregor in #1256. I realize it's slightly out of scope but it's a nice addition in my opinion, let me know if this is better suited for a separate PR.

### DIFF
--- a/Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
+++ b/Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
@@ -139,10 +139,11 @@ public struct DiagnosticsFormatter {
     switch message.severity {
     case .error:
       let annotation = ANSIAnnotation(color: .red, trait: .bold)
-      return annotation.applied(to: message.message)
+      return annotation.applied(to: "error: \(message.message)")
     case .warning:
-      let annotation = ANSIAnnotation(color: .yellow)
-      return annotation.applied(to: message.message)
+      let color = ANSIAnnotation(color: .yellow)
+      let prefix = color.withTrait(.bold).applied(to: "warning: ")
+      return prefix + color.applied(to: message.message)
     case .note:
       return message.message
     }
@@ -179,6 +180,10 @@ struct ANSIAnnotation {
   init(color: Color, trait: Trait = .normal) {
     self.color = color
     self.trait = trait
+  }
+
+  func withTrait(_ trait: Trait) -> Self {
+    return ANSIAnnotation(color: self.color, trait: trait)
   }
 
   func applied(to message: String) -> String {
--- a/Sources/swift-parser-cli/TerminalUtils.swift
+++ b/Sources/swift-parser-cli/TerminalUtils.swift
@@ -1,0 +1,51 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2022 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+#if canImport(Glibc)
+import Glibc
+#elseif os(Windows)
+import CRT
+#else
+import Darwin.C
+#endif
+
+#if os(Android)
+typealias FILEPointer = OpaquePointer
+#else
+typealias FILEPointer = UnsafeMutablePointer<FILE>
+#endif
+
+enum TerminalHelper {
+  static var isConnectedToTerminal: Bool {
+    return isTTY(stderr)
+  }
+
+  /// Checks if passed file pointer is a tty.
+  static func isTTY(_ filePointer: FILEPointer) -> Bool {
+    return terminalType(filePointer) == .tty
+  }
+
+  /// The type of terminal.
+  enum TerminalType {
+    /// The terminal is a TTY.
+    case tty
+
+    /// The terminal is a file stream.
+    case file
+  }
+
+  /// Computes the terminal type of the stream.
+  static func terminalType(_ filePointer: FILEPointer) -> TerminalType {
+    let isTTY = isatty(fileno(filePointer)) != 0
+    return isTTY ? .tty : .file
+  }
+}
--- a/Sources/swift-parser-cli/swift-parser-cli.swift
+++ b/Sources/swift-parser-cli/swift-parser-cli.swift
@@ -198,7 +198,7 @@ class PrintDiags: ParsableCommand {
   @Flag(name: .long, help: "Perform sequence folding with the standard operators")
   var foldSequences: Bool = false
 
-  @Flag(name: .long, help: "Colorize output with ANSI color codes")
+  @Flag(name: .long, help: "Force output coloring with ANSI color codes")
   var colorize: Bool = false
 
   func run() throws {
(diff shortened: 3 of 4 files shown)


DougGregor:
@swift-ci please test

simba909:
Looks like tests are failing because I forgot to add `TerminalUtils.swift` to `CMakeLists`, I've pushed a new commit that fixes this.

DougGregor:
@swift-ci please test

DougGregor:
Ah, the remaining macOS failure is because we've been trying not to introduce a Darwin dependency into the parts of swift-syntax that get linked into the compiler.

DougGregor:
I have a concrete suggestion. How about we move the TTY detection logic down into `swift-parser-cli`, which is allowed to have more dependencies (on Foundation, system packages like Darwin, etc.), and have the `SwiftDiagnostics` formatter code only take a `Bool` for whether the output should be colorized or not.

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
```diff
@@ -139,10 +179,10 @@ public struct DiagnosticsFormatter {
     switch message.severity {
     case .error:
       let annotation = ANSIAnnotation(color: .red, trait: .bold)
-      return annotation.applied(to: message.message)
+      return annotation.applied(to: "error: \(message.message)")
     case .warning:
       let annotation = ANSIAnnotation(color: .yellow)
-      return annotation.applied(to: message.message)
+      return annotation.applied(to: "warning: \(message.message)")
```
DougGregor, line 185: Thank you! Should we bold the `error` and `warning` bits to make them stand out more?

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
```diff
@@ -26,21 +41,46 @@ public struct DiagnosticsFormatter {
...
     self.contextSize = contextSize
     self.colorize = colorize
   }
 
+  public init(contextSize: Int = defaultContextSize, colorMode: ColorMode) {
+    var colorize: Bool
+
+    switch colorMode {
+    case .auto:
+      colorize = TerminalHelper.isConnectedToTerminal
```
DougGregor, line 54: This is the bit that I think we should life into `swift-parser-cli`, because we're trying to keep the `SwiftDiagnostics` library (and everything that gets linked into the compiler proper) with minimal dependencies.

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
simba909: Good idea. The `error` text should already be bold (as part of the message that gets annotated), I'll add the necessary changes for warnings as well 👍🏻

simba909:
@DougGregor thank you for reviewing! I've moved the TTY detection logic and made the `warning:` prefix bold.

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor (approved):
Awesome, thanks!

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
```diff
@@ -139,10 +139,11 @@ public struct DiagnosticsFormatter {
     switch message.severity {
     case .error:
       let annotation = ANSIAnnotation(color: .red, trait: .bold)
-      return annotation.applied(to: message.message)
+      return annotation.applied(to: "error: \(message.message)")
     case .warning:
-      let annotation = ANSIAnnotation(color: .yellow)
-      return annotation.applied(to: message.message)
+      let color = ANSIAnnotation(color: .yellow)
+      let prefix = color.withTrait(.bold).applied(to: "warning: ")
```
ahoppen, line 145: Just to be clear: It’s intentional that errors have both the prefix and the the message in bold but warnings only have a bold prefix? I’m fine with either way but just want to make sure the current design is a conscious decision.

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
```diff
@@ -139,10 +139,11 @@ public struct DiagnosticsFormatter {
...
+      return prefix + color.applied(to: message.message)
     case .note:
       return message.message
```
ahoppen, line 148: Should we also add a prefix for notes?

Sources/swift-parser-cli/TerminalUtils.swift
```diff
@@ -0,0 +1,51 @@
...
+#if os(Android)
+typealias FILEPointer = OpaquePointer
+#else
+typealias FILEPointer = UnsafeMutablePointer<FILE>
+#endif
+
+enum TerminalHelper {
+  static var isConnectedToTerminal: Bool {
+    return isTTY(stderr)
+  }
```
ahoppen, line 30: Since we are printing the errors to stdout (instead of stderr like swift-format), we should be checking whether stdout is connected to a terminal here. For example, stdout could be connected to a terminal while stderr is not if you do `swift-parser-cli … 2> /tmp/errors.txt`.

And, to be clear, we should probably name the method to mention that is checking `stdout` instead of `stderr`.

```suggestion
  static var isStdoutConnectedToTerminal: Bool {
    return isTTY(stdout)
  }
```

Sources/swift-parser-cli/TerminalUtils.swift
```diff
@@ -0,0 +1,51 @@
...
+#elseif os(Windows)
+import CRT
+#else
+import Darwin.C
+#endif
+
+#if os(Android)
+typealias FILEPointer = OpaquePointer
+#else
+typealias FILEPointer = UnsafeMutablePointer<FILE>
```
ahoppen, line 24: Should this be named `FilePointer` to look a little more swift? It’s bad enough to have `FILE` scream at you in all caps.

Sources/swift-parser-cli/swift-parser-cli.swift
```diff
@@ -198,17 +198,22 @@ class PrintDiags: ParsableCommand {
   @Flag(name: .long, help: "Perform sequence folding with the standard operators")
   var foldSequences: Bool = false
 
-  @Flag(name: .long, help: "Colorize output with ANSI color codes")
+  @Flag(name: .long, help: "Force output coloring with ANSI color codes")
```
ahoppen, line 201: You could make this a flag with `inversion: .prefixedNo` like in https://github.com/apple/swift-format/blob/072768f1af50989996e137919f9f6be71e8294e6/Sources/swift-format/Subcommands/LintFormatOptions.swift#L62-L69 so you can also force coloring off if you want to.

---

And now that I think about it, we should name the flag `color-diagnostics` (just rename the variable to `colorDiagnostics`) to match the name of the flag in the Swift Compiler.

```suggestion
  @Flag(name: .long, inversion: .prefixedLo, help: "Enable or disable coloring with ANSI color codes")
  let colorDiagnostics: Bool?
```

ahoppen (commented):
Thank you. It’s nice to see colors automatically enabled.

DougGregor:
@ahoppen are you okay with those suggestions being addressed in a follow-up PR?

DougGregor:
I'm going to merge because this is progress, and we can address feedback in a follow-up.

PR #1286, opened 2023-01-27
Fix two assertion failures related to invalid UTF-8

ahoppen:
- Fix an incorrect assertion in Cursor.swift
  - This assertion returned `false` if we reached the end of the file, while it should return `true`.
- Make SourceLocationConverter handle invalid UTF-8
  - The `SourceLocationConverter` previously worked on `String`, which is lossy if the source contains invalid UTF-8. Make it work on `SyntaxText` and `RawTriviaPiece` so it’s source accurate.

While at it, also fix a few things in `swift-parser-cli`:
- Show diagnostics from operator folding when running `swift-parser-cli print-diags`
- Run the diagnostic formatter during round-trip testing
  - This allows us to reduce crashes in the diagnostic formatter.
- Output reduced test case as bytes instead of a string
  - This allows us to reduce invalid UTF-8, which were lossily converted when creating a `String`.

### DIFF
--- a/Sources/SwiftParser/Lexer/Cursor.swift
+++ b/Sources/SwiftParser/Lexer/Cursor.swift
@@ -2102,7 +2102,7 @@ extension Lexer.Cursor {
   /// valid operator start, advance the cursor by what can be considered a
   /// lexeme.
   mutating func lexUnknown() -> UnknownCharactersClassification {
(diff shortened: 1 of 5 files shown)


ahoppen:
@swift-ci Please test

DougGregor (approved):

PR #1287, opened 2023-01-27
Add throws to result builder

kimdv:
Depends on #1294


CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxTraitsFile.swift
```diff
@@ -28,7 +28,7 @@ let syntaxTraitsFile = SourceFileSyntax {
       }
     }
 
-    ExtensionDeclSyntax("public extension \(trait.traitName)Syntax") {
+    ExtensionDeclSyntax("public extension \(raw: trait.traitName)Syntax") {
```
kimdv, line 31: There was a warning. Unrelated to this PR tough

kimdv:
@swift-ci please test

Sources/SwiftSyntaxBuilder/generated/BuildableNodes.swift
```diff
@@ -17,397 +17,440 @@ import SwiftSyntax
 
 extension AccessorDeclSyntax {
   /// A convenience initializer that allows initializing syntax collections using result builders
-  public init(leadingTrivia: Trivia? = nil, unexpectedBeforeAttributes: UnexpectedNodesSyntax? = nil, attributes: AttributeListSyntax? = nil, unexpectedBetweenAttributesAndModifier: UnexpectedNodesSyntax? = nil, modifier: DeclModifierSyntax? = nil, unexpectedBetweenModifierAndAccessorKind: UnexpectedNodesSyntax? = nil, accessorKind: TokenSyntax, unexpectedBetweenAccessorKindAndParameter: UnexpectedNodesSyntax? = nil, parameter: AccessorParameterSyntax? = nil, unexpectedBetweenParameterAndEffectSpecifiers: UnexpectedNodesSyntax? = nil, effectSpecifiers: DeclEffectSpecifiersSyntax? = nil, unexpectedBetweenEffectSpecifiersAndBody: UnexpectedNodesSyntax? = nil, @CodeBlockItemListBuilder bodyBuilder: () -> CodeBlockItemListSyntax? = {
+  @_disfavoredOverload
```
kimdv, line 20: @ahoppen,  I've tried to add `@_disfavoredOverload` to the function, as we get compile errors in the test with doing 
```swift
    let carriateReturnsStruct = StructDeclSyntax(
      leadingTrivia: [
        .docLineComment("/// A nested struct"),
        .carriageReturns(1),
        .docLineComment("/// with multi line comment where the newline is a CR"),
        .carriageReturns(1),
      ],
      structKeyword: .keyword(.struct),
      identifier: "CarriateReturnsStruct"
    )
```

etc. 

If I understand it correctly is it because it chose the result builder init method instead of the "normal" one

ahoppen: I don’t understand why `@_disfavoredOverload` would be necessary here. But I haven’t looked into it either. Will take a look on Monday.

kimdv: I think a it's because it cannot chose the init method that doesn't throw.

ahoppen: I just debugged and I believe this is due to https://github.com/apple/swift/issues/63285. 

The gist is that `rethrows` doesn’t work property with throwing closures that have default values.

The solution would be to stamp out two copies of the initializer. One that is the one we have currently and the other one is what you generated but with `throws` instead of `rethrows` (essentially we are just implementing the `rethrows` behavior with the two overloads ourselves.

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntaxbuilder/BuildableNodesFile.swift
```diff
@@ -82,9 +87,11 @@ private func createConvenienceInitializer(node: Node) -> InitializerDeclSyntax?
         }
       }
       builderParameters.append(FunctionParameterSyntax(
-        "@\(builderInitializableType.resultBuilderBaseName) \(child.swiftName)Builder: () -> \(builderInitializableType.syntax) = \(defaultArgument)",
+        "@\(builderInitializableType.resultBuilderBaseName) \(child.swiftName)Builder: () \(addRethrows ? "throws" : "") -> \(builderInitializableType.syntax) = \(defaultArgument)",
         for: .functionParameters
       ))
+
+      willBuilderThrow = true
```
ahoppen, line 94: I don't think we need this variable. AFAICT it always has the same value as `shouldCreateInitializer`. And if we get to the point where we're checking the variable, we know that `shouldCreateInitializer` is true because there's a `guard`.

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntaxbuilder/BuildableNodesFile.swift
```diff
@@ -45,10 +49,11 @@ private func convertFromSyntaxProtocolToSyntaxType(child: Child) -> ExprSyntax {
 }
 
 /// Create a builder-based convenience initializer, if needed.
-private func createConvenienceInitializer(node: Node) -> InitializerDeclSyntax? {
+private func createConvenienceInitializer(node: Node, addRethrows: Bool) -> InitializerDeclSyntax? {
```
ahoppen, line 52: I would Name the parameter `allowBuildersToThrow`

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntaxbuilder/BuildableNodesFile.swift
```diff
@@ -103,6 +110,7 @@ private func createConvenienceInitializer(node: Node) -> InitializerDeclSyntax?
 
   return InitializerDeclSyntax(
     leadingTrivia: .docLineComment("/// A convenience initializer that allows initializing syntax collections using result builders") + .newline,
+    // attributes: [.attribute(AttributeSyntax(attributeName: TypeSyntax("_disfavoredOverload"), trailingTrivia: .newline))],
```
ahoppen, line 113: No longer needed

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test

kimdv:
Rebased after #1294 was merged @ahoppen 
Single initialiser is now possible

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test

PR #1289, opened 2023-01-28
Improve modeling of `Child` in CodeGeneration

ahoppen:
Instead of having `tokenChoices`, `textChoices`, `nodeChoices` and `collectionElementName` in `Child`, which are all pretty much mutually exclusive with each other, model the kind of a child as an enum that be either of
- A single node kind
- A list of node choices
- A syntax collection
- A list of possible tokens

The syntax in the Python definition files is a little awkward with this because you specify keywords e.g. as `Keyword|async` but since I anticipate that they will soon be gone anyway once we've moved all code generation to use SwiftSyntaxBuilder, I don't think we need to put too much effort into making the Python files look nice.

---

This is a prerequisite for https://github.com/apple/swift-syntax/pull/1180, which will assert that tokens in a syntax tree match the specified token kind and text choices.


ahoppen:
@swift-ci Please test

kimdv (approved):

CodeGeneration/Sources/SyntaxSupport/Child.swift
```diff
@@ -10,28 +10,79 @@
...
     return lowercaseFirstWord(name: name)
   }
 
+  public var syntaxKind: String {
+    switch kind {
+    case .node(kind: let kind):
+      return kind
+    case .nodeChoices:
+      return "syntax"
+    case .collection(kind: let kind, collectionElementName: _):
```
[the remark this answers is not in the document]
ahoppen: We are using it in `SyntaxNodes.swift.gyb.template`. So when we transition that to CodeGeneration, we’ll need `collectionElementName`.

ahoppen:
@swift-ci Please test

CodeGeneration/Sources/SyntaxSupport/Child.swift
```diff
@@ -10,28 +10,79 @@
...
+    case .keyword: return true
+    case .token: return false
+    }
+  }
+}
+
+public enum ChildKind {
+  /// The child always contains a node of the given `kind`.
+  case node(kind: String)
+  /// The child always contains a child that matches one of the `choices`.
```
bnbarham, line 30:
```suggestion
  /// The child always contains a node that matches one of the `choices`.
```

But also, why separate this and the single kind case but not the single token vs tokens case?

CodeGeneration/Sources/SyntaxSupport/gyb_generated/AvailabilityNodes.swift
```diff
@@ -81,14 +66,11 @@ public let AVAILABILITY_NODES: [Node] = [
        kind: "Syntax",
        children: [
          Child(name: "Platform",
-               kind: "IdentifierToken",
+               kind: .token(choices: [.token(tokenKind: "IdentifierToken")]),
                description: "The name of the OS on which the availability should berestricted or 'swift' if the availability should berestricted based on a Swift version.",
```
bnbarham, line 70: I'm not actually sure what these are used for. But it appears we're not adding a space when we drop the newlines and indentation (separate to this PR, just noticed it here).

gyb_syntax_support/Child.py
```diff
@@ -46,16 +46,17 @@ def __init__(self, name, kind, name_for_diagnostics=None, description=None, is_o
...
+                choice = full_choice[:pipe_index]
+                choice_text = full_choice[(pipe_index+1):]
             token = SYNTAX_TOKEN_MAP[choice]
-            self.token_choices.append(token)
-
-        # A list of valid text for tokens, if specified.
-        # This will force validation logic to check the text passed into the
-        # token against the choices.
-        self.text_choices = text_choices or []
+            self.token_choices.append((token, choice_text))
```
bnbarham, line 59: Is there a case where `token != KeywordToken` and `kind != KeywordToken`? Can't we just check/use `kind` instead?

CodeGeneration/Sources/SyntaxSupport/Child.swift
ahoppen: Because even if we have multiple `TokenChoice`s, the kind of the child is a `TokenSyntax`. But if there are multiple layout node choices, we need to create an enum for those choices. Also note that `nodeChoices` has `Child` (not a list of `String` the represent the different kinds) as the array kind, which means that `nodeChoices` would be a collection or list of tokens.

CodeGeneration/Sources/SyntaxSupport/gyb_generated/AvailabilityNodes.swift
ahoppen: What exactly do you mean?

gyb_syntax_support/Child.py
ahoppen: What exactly are you suggesting? Also, this is going to be removed in a month or so anyway when all gyb files are migrated to CodeGeneration so maybe it doesn’t matter too much anyway.

ahoppen:
@swift-ci Please test

kimdv:
@swift-ci please test macOS

ahoppen:
@swift-ci Please test macOS

bnbarham (approved):

PR #1290, opened 2023-01-28
Don’t fatalError when expressing syntax nodes by string literals

ahoppen:
We should not crash the process if the source code that you use to create a syntax node using string interpolation is invalid. Instead we should:

- Not check for errors in the normal string interpolation case and just return a syntax tree that has the `hasError` flag set when there are syntax errors
- In either SwiftSyntaxBuilder offer `init(validating node: Self) throws` that check that `node` has no sytnax errors and throws otherwise. This allows us write e.g. `try wDeclSyntax(validating: "struct Foo {}")` to create a `DeclSyntax` that’s guaranteed to not have any syntax errors.

I’m not sure yet whether CodeGeneration should use `init(validating:)` with `try!` sprayed all over the place or have something like `init(checking:)` that `fatalError`s if the syntax is not valid.

rdar://104423126


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

kimdv (approved):

bnbarham:
Whoops, pretend my https://github.com/apple/swift-syntax/pull/1291 comments were here 😅

bnbarham (approved):

ahoppen:
@swift-ci Please test

PR #1291, opened 2023-01-28
Fix a crash when string interpolation creates a MissingStmtSyntax

ahoppen:
We were crashing here because string interpolation created a `MissingStmtSyntax`, which doesn’t have any children. It then tried to assign the remaining tokens as unexpected tokens to the last child of the `MissingStmtSyntax`, which failed (because there are no children).

The fix here is to just add an `unexpected` child to the missing nodes. I’ve got a gut feeling that that might be useful anyway.

### DIFF
--- a/Tests/SwiftSyntaxBuilderTest/StringInterpolationTests.swift
+++ b/Tests/SwiftSyntaxBuilderTest/StringInterpolationTests.swift
@@ -504,4 +504,22 @@ final class StringInterpolationTests: XCTestCase {
       )
     }
   }
+
+  func testInvalidSyntax2() {
+    let invalid = StmtSyntax("struct Foo {}")
+    XCTAssert(invalid.hasError)
+
+    XCTAssertThrowsError(try StmtSyntax(validating: "struct Foo {}")) { error in
+      AssertStringsEqualWithDiff(
+        String(describing: error),
+        """
+
+        1 │ struct Foo {}
+          ∣ │            ╰─ expected statement
+          ∣ ╰─ unexpected code 'struct Foo {}' before statement
+
+        """
+      )
+    }
+  }
 }
--- a/Sources/SwiftSyntax/Raw/gyb_generated/RawSyntaxNodes.swift
+++ b/Sources/SwiftSyntax/Raw/gyb_generated/RawSyntaxNodes.swift
@@ -203,12 +203,20 @@ public struct RawMissingSyntax: RawSyntaxNodeProtocol {
   }
 
   public init(
+    _ unexpected: RawUnexpectedNodesSyntax? = nil,
     arena: __shared SyntaxArena
   ) {
-    let raw = RawSyntax.makeEmptyLayout(kind: .missing, arena: arena)
+    let raw = RawSyntax.makeLayout(
+      kind: .missing, uninitializedCount: 1, arena: arena) { layout in
+      layout.initialize(repeating: nil)
+      layout[0] = unexpected?.raw
+    }
     self.init(raw: raw)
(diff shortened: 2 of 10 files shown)


Tests/SwiftSyntaxBuilderTest/StringInterpolationTests.swift
```diff
@@ -485,4 +486,40 @@ final class StringInterpolationTests: XCTestCase {
...
+  func testInvalidSyntax() {
+    let invalid = DeclSyntax("return 1")
+    XCTAssert(invalid.hasError)
+
+    XCTAssertThrowsError(try DeclSyntax(validating: "return 1")) { error in
+      AssertStringsEqualWithDiff(
+        String(describing: error),
+        """
+
+        1 │ return 1
```
bnbarham, line 499: I'd prefer not testing the formatted string here. That's going to be a pain if we ever change it 😅

Sources/SwiftSyntaxBuilder/generated/SyntaxExpressibleByStringInterpolationConformances.swift
```diff
@@ -19,8 +19,10 @@ import SwiftParser
 import SwiftParserDiagnostics
 
 extension SyntaxParseable {
-  public init(stringInterpolationOrThrow stringInterpolation: SyntaxStringInterpolation) throws {
-    self = try performParse(source: stringInterpolation.sourceText, parse: { parser in 
+  public typealias StringInterpolation = SyntaxStringInterpolation
```
bnbarham, line 22: Why wasn't this needed before?

Sources/SwiftSyntaxBuilder/ValidatingSyntaxNodes.swift
```diff
@@ -0,0 +1,58 @@
...
+}
+
+extension Trivia {
+  /// If `trivia` has contains no unexpected trivia, return `trivia`, otherwise
+  /// throw an error with diagnostics describing the unexpected trivia.
+  public init(validating trivia: Trivia) throws {
+    self = trivia
+    if pieces.contains(where: { $0.isUnexpected }) {
+      var diagnostics: [Diagnostic] = []
+      let tree = SourceFileSyntax(statements: [], eofToken: .eof(leadingTrivia: self))
```
bnbarham, line 41: Little bit weird, but I suppose this is what we were doing before.

bnbarham (approved):

Sources/SwiftSyntaxBuilder/generated/SyntaxExpressibleByStringInterpolationConformances.swift
ahoppen: I honestly don’t know and gave up trying to figure out why we need it now after a couple of minutes.

Sources/SwiftSyntaxBuilder/ValidatingSyntaxNodes.swift
ahoppen: It does the job of presenting a nice diagnostic so 🤷

Tests/SwiftSyntaxBuilderTest/StringInterpolationTests.swift
ahoppen: We already have tests that check the diagnostic format in `SwiftDiagnosticTests` and I really want to check that we produce a nice-looking error message here. 

So: I would-refer to keep it.

ahoppen:
@swift-ci Please test

PR #1273, opened 2023-01-26
Update swift-syntax

kimdv:
Depends on #1287

PR #1293, opened 2023-01-29
Move syntax node template from gyb to code gen

kimdv:
Depends on #1273, #1294


kimdv:
@swift-ci please test

CodeGeneration/Sources/SyntaxSupport/Child.swift
```diff
@@ -69,12 +70,25 @@ public class Child {
     syntaxKind == "UnexpectedNodes"
   }
 
+  /// Returns `True` if this child's type is one of the base syntax kinds and
```
ahoppen, line 73: Nit `True` -> `true`. Same in `hasOptionalBaseType`

CodeGeneration/Sources/SyntaxSupport/Child.swift
```diff
@@ -118,4 +133,46 @@ public class Child {
       assert(choice.nodeChoices.isEmpty, "node choice \(choice.name) cannot have further choices")
     }
   }
+
+  /// If the type has a default value, return an expression of the form
+  /// ` = default_value` that can be used as the default value to for a
+  /// function parameter. Otherwise, return an empty string.
+  func generateDefaultInitialization() -> String {
```
ahoppen, line 140: I think it would be nicer to return an `ExprSyntax` here.

CodeGeneration/Sources/SyntaxSupport/Node.swift
```diff
@@ -145,4 +155,61 @@ public class Node {
     // For SyntaxCollections make sure that the elementName is set.
     assert(!isSyntaxCollection || elementName != nil || element != "")
   }
+
+  public func generateInitializerDecl(optionalBaseAsMissing: Bool, currentIndentation: String = "") -> String {
```
ahoppen, line 159: Again, I think we should return an `ExprSyntax` here.

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodeFile.swift
```diff
@@ -0,0 +1,359 @@
...
+/// This gyb-file generates the syntax nodes for SwiftSyntax. To keep the generated
+/// files at a managable file size, it is to be invoked multiple times with the
+/// global variable EMIT_KIND set to a base kind listed in
+/// gyb_syntax_support/kinds.py:SYNTAX_BASE_KINDES. It then only emits those syntax
+/// nodes whose base kind are that specified kind.
+func syntaxNode(emitKind: String) -> SourceFileSyntax {
+  SourceFileSyntax(leadingTrivia: .docLineComment(generateCopyrightHeader(for: "generate-swiftsyntax"))) {
+    for node in SYNTAX_NODES {
+      if node.isBase {
+        // Handled in SyntaxNodesBase.swift.gyb
```
ahoppen, line 27: Nit: `SyntaxNodesBase.swift.gyb` -> `SyntaxNodesBaseFile.swift`

(similar below).

Also, we could reduce nesting level by 1 by having `continue` in these branches and then the `else` block could come after this `if` statement.

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodeFile.swift
```diff
@@ -0,0 +1,359 @@
...
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+import SwiftSyntaxBuilder
+import SyntaxSupport
+import Utils
+
+/// This gyb-file generates the syntax nodes for SwiftSyntax. To keep the generated
+/// files at a managable file size, it is to be invoked multiple times with the
+/// global variable EMIT_KIND set to a base kind listed in
```
ahoppen, line 20: `EMIT_KIND` no longer is a global variable but a parameter.

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodeFile.swift
```diff
@@ -0,0 +1,359 @@
...
+        // We are actually handling this node now
+        let nodeDoc = node.description.map { "/// \($0)" }
+        try! StructDeclSyntax("""
+        // MARK: - \(raw: node.name)
+
+        \(raw: nodeDoc ?? "")
+        public struct \(raw: node.name): \(raw: node.baseType.syntaxBaseName)Protocol, SyntaxHashable
+        """) {
+          for child in node.children {
+            if !child.nodeChoices.isEmpty {
```
ahoppen, line 42: You can write this as

```swift
for child in node.children where !child.nodeChoices.isEmpty
```

---

I think it makes sense to extract the generation of the `SyntaxChildChoices` enum to a separate function to improve readability.

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodeFile.swift
```diff
@@ -0,0 +1,359 @@
...
+                      TupleExprElementSyntax(
+                        expression: ArrayExprSyntax {
+                          for choice in child.nodeChoices {
+                            ArrayElementSyntax(
+                              expression: ExprSyntax(".node(\(raw: choice.typeName).self)"))
+                          }
+                        }
+                      )
+                    }
+                  )
```
ahoppen, line 102: I’m not sure if it works but can we write this as something like

```swift
let choices = ArrayExprSyntax {
  for choice in child.nodeChoices {
    ArrayElementSyntax(
      expression: ExprSyntax(".node(\(raw: choice.typeName).self)"))
  }
}
ReturnStmtSyntax("return .choices(\(choices))")
```

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodeFile.swift
```diff
@@ -0,0 +1,359 @@
...
+
+                DeclSyntax("init(_ data: SyntaxData) { self.init(Syntax(data))! }")
+
+                for choice in child.nodeChoices {
+                  if let choiceNode = SYNTAX_NODE_MAP[choice.syntaxKind], choiceNode.isBase {
+                    DeclSyntax("""
+                        public init<Node: \(raw: choiceNode.name)Protocol>(_ node: Node) {
+                          self = .\(raw: choice.swiftName)(\(raw: choiceNode.name)(node))
+                        }
+                        """)
```
ahoppen, line 66: Investing that swift-format doesn’t complain about this but the closing quote of this literal is indented by 4 spaces (same below) and the one after is not indented at all. I think I would expect 2 space of indentation.

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodeFile.swift
```diff
@@ -0,0 +1,359 @@
...
+          try! InitializerDeclSyntax("\(raw: node.generateInitializerDecl(optionalBaseAsMissing: false, currentIndentation: "  "))") {
+            let parameters = node.children.map { $0.swiftName }.joined(separator: ", ")
+
+            VariableDeclSyntax(
+              leadingTrivia: [
+                .docLineComment("// Extend the lifetime of all parameters so their arenas don't get destroyed"),
+                .newlines(1),
+                .docLineComment("// before they can be added as children of the new arena."),
+                .newlines(1)
+              ],
```
ahoppen, line 140: Trivia is now expressible by string literals, so this can just be

```swift
leadingTrivia: """
// Extend the lifetime of all parameters so their arenas don't get destroyed
// before they can be added as children of the new arena.
"""
```

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodeFile.swift
```diff
@@ -0,0 +1,359 @@
...
+              letOrVarKeyword: .keyword(.let)) {
+                PatternBindingSyntax(
+                  pattern: PatternSyntax("data"),
+                  typeAnnotation: TypeAnnotationSyntax(type: TypeSyntax("SyntaxData")),
+                  initializer: InitializerClauseSyntax(
```
ahoppen, line 145: Wow, this variable declaration is really deeply nested and long. What do you think about a pattern like the following? I wrote this in GitHub and it might be buggy but I think it gets the idea across.

```swift
let layoutVariable = <lines 164 - 180>
let rawVariable = <lines 183 - 191>
let withExtendedLifetimeParameters = <lines 149 - 151>
let closureSignature = <lines 154 - 161>
let initializer = ExprSyntax("""
  withExtendedLifetime(\(withExtendedLifetimeParameters)) { \(closureSignature) in
    \(layoutVariable)
    \(rawVariable)
    return SyntaxData.forRoot(raw)
  """)
```

---

Also, I just noticed we have a convenience initializer on `VariableDecl` that takes an `initializer` so I t think you don’t need to mess around with the `PatternBindingSyntax`.

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodeFile.swift
```diff
@@ -0,0 +1,359 @@
...
+                  )
+                )
+              }
+
+            ExprSyntax("self.init(data)")
+          }
+
+          if node.hasOptionalBaseTypeChild {
+            // TODO: Remove when we no longer support compiling in Swift 5.6. Change the
+            // above constructor to use `Optional<BaseType.none` instead.
```
ahoppen, line 205: Nit, missing `>` after `none`.

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodeFile.swift
```diff
@@ -0,0 +1,359 @@
...
+
+            let childType = child.nodeChoices.isEmpty ? child.typeName : child.name
+            let type = child.isOptional ? TypeSyntax("\(raw: childType)?") : TypeSyntax("\(raw: childType)")
+            let childDoc = child.description.map { dedented(string: $0) }.map { Trivia.docLineComment("/// \($0)") + .newline }
+
+            VariableDeclSyntax(
+              leadingTrivia: childDoc ?? [],
+              modifiers: [DeclModifierSyntax(name: .keyword(.public))],
+              name: PatternSyntax("\(raw: child.swiftName)"),
+              type: TypeAnnotationSyntax(type: type)) {
```
ahoppen, line 248: It might be worth adding a new convenience initializer to `VariableDecl` that takes multiple `AccessorDeclSyntax` using a builder. Doesn’t have to be for this PR though.

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodeFile.swift
```diff
@@ -0,0 +1,359 @@
...
+                  }
+                  let newData = data.replacingChild(at: \(raw: index), with: collection, arena: arena)
+                  return \(raw: node.name)(newData)
+                }
+                """)
+            }
+          }
+
+          try! VariableDeclSyntax("public static var structure: SyntaxNodeStructure") {
+            ReturnStmtSyntax(
```
ahoppen, line 297: Similar to a case above I think this can be simplified as

```swift
let layout = ArrayExprSyntax {
  for child in node.children {
    ArrayElementSyntax(
      expression: ExprSyntax(#"\Self.\#(raw: child.swiftName)"#))
  }
}
StmtSyntax("return .layout(\(layout))")
```

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodeFile.swift
kimdv: We cannot use `continue` due to `Closure containing control flow statement cannot be used with result builder 'CodeBlockItemListBuilder'`

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodeFile.swift
kimdv: I think it should be after `BaseType` and before the `.`

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodeFile.swift
kimdv: With the suggested change above we are always initialising and empty array lige `let layout: [RawSyntax?] = []`

Not sure what performance impact it will have.

ahoppen: What do mean? AFAICT this should just be moving code around in this file but still  generate the exact same code.

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodeFile.swift
ahoppen: Oh, yes. I only noticed the missing `>` and didn’t actually read the line 😬

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodeFile.swift
ahoppen: We could do the following instead and merge all the comments in the `if` branches into a single one above the loop.

```swift
    for node in SYNTAX_NODES where !node.isBase && node.collectionElement.isEmpty && node.baseKind == emitKind {
```

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodeFile.swift
kimdv: Before `<lines 164 - 180>` was inside an if. So when `node.children` was empty is was **not** created. 
Now we get a bit of difference. See `MissingExprSyntax` etc. 

```swift
  public init() {
    // Extend the lifetime of all parameters so their arenas don't get destroyed
    // before they can be added as children of the new arena.
    let data: SyntaxData = withExtendedLifetime((SyntaxArena(), ())) { (arena, _) in 
      let layout: [RawSyntax?] = []
      let raw = RawSyntax.makeEmptyLayout(kind: SyntaxKind.missingExpr, arena: arena)
      return SyntaxData.forRoot(raw)
    }
    self.init(data)
  }
```

Layout is not used

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodeFile.swift
kimdv: We can use the one already there 🥳

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodeFile.swift
kimdv: Fixed in #1301

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodeFile.swift
```diff
@@ -0,0 +1,432 @@
...
+                pattern: PatternSyntax("data"),
+                typeAnnotation: TypeAnnotationSyntax(type: TypeSyntax("SyntaxData")),
+                initializer: InitializerClauseSyntax(
+                  value: initializer
+                )
+              )
+            }
+
+          ExprSyntax("self.init(data)")
+        }
```
ahoppen, line 116: I couldn’t resist the temptation to play around with this to get rid of `layout` where it’s not needed. This is what I came up with. WDYT?

```swift
try! InitializerDeclSyntax("\(raw: generateInitializerDecl(children: node.children, optionalBaseAsMissing: false, currentIndentation: "  "))") {
  let parameters = node.children.map { $0.swiftName }.joined(separator: ", ")

  let closureSignature = ClosureSignatureSyntax(
    input: .input(
      ParameterClauseSyntax {
        FunctionParameterSyntax(firstName: .identifier("arena"))
        FunctionParameterSyntax(firstName: .wildcardToken())
      }
    )
  )
  let layoutList = ArrayExprSyntax {
    for child in node.children {
      ArrayElementSyntax(expression: MemberAccessExprSyntax(
        base: child.type.optionalChained(expr: ExprSyntax("\(raw: child.swiftName)")),
        dot: .periodToken(),
        name: "raw"
      ))
    }
  }

  let initializer = FunctionCallExprSyntax(
    calledExpression: ExprSyntax("withExtendedLifetime"),
    leftParen: .leftParenToken(),
    argumentList: TupleExprElementListSyntax {
      TupleExprElementSyntax(expression: ExprSyntax("(SyntaxArena(), (\(raw: parameters)))"))
    },
    rightParen: .rightParenToken(),
    trailingClosure: ClosureExprSyntax(signature: closureSignature) {
      if node.children.isEmpty {
        DeclSyntax("let raw = RawSyntax.makeEmptyLayout(kind: SyntaxKind.\(raw: node.swiftSyntaxKind), arena: arena)")
      } else {
        DeclSyntax("let layout: [RawSyntax?] = \(layoutList)")
        DeclSyntax("""
            let raw = RawSyntax.makeLayout(
              kind: SyntaxKind.\(raw: node.swiftSyntaxKind), from: layout, arena: arena,
              leadingTrivia: leadingTrivia, trailingTrivia: trailingTrivia)
            """)
      }
    StmtSyntax("return SyntaxData.forRoot(raw)")
  })

  VariableDeclSyntax(
    leadingTrivia: """
      // Extend the lifetime of all parameters so their arenas don't get destroyed
      // before they can be added as children of the new arena.

      """,
    .let,
    name: PatternSyntax("data"),
    type: TypeAnnotationSyntax(type: TypeSyntax("SyntaxData")),
    initializer: InitializerClauseSyntax(value: initializer)
  )

  ExprSyntax("self.init(data)")
}
```

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodeFile.swift
```diff
@@ -0,0 +1,432 @@
...
+                    DictionaryElementSyntax(
+                      leadingTrivia: .newline,
+                      keyExpression: ExprSyntax(#""\#(raw: child.swiftName)""#),
+                      valueExpression: ExprSyntax("\(raw: child.swiftName).map(Syntax.init)?.asProtocol(SyntaxProtocol.self) as Any"))
+                  } else {
+                    DictionaryElementSyntax(
+                      leadingTrivia: .newline,
+                      keyExpression: ExprSyntax(#""\#(raw: child.swiftName)""#),
+                      valueExpression: ExprSyntax("Syntax(\(raw: child.swiftName)).asProtocol(SyntaxProtocol.self)"))
+                  }
```
ahoppen, line 259: It’s a little unfortunate to repeat everything except the `valueExpression`. I’m not a huge fan of the ternary operator but maybe it’s clearer here.

```suggestion
                  DictionaryElementSyntax(
                    leadingTrivia: .newline,
                    keyExpression: ExprSyntax(#""\#(raw: child.swiftName)""#),
                    valueExpression: child.isOptional
                    ? ExprSyntax("\(raw: child.swiftName).map(Syntax.init)?.asProtocol(SyntaxProtocol.self) as Any")
                    : ExprSyntax("Syntax(\(raw: child.swiftName)).asProtocol(SyntaxProtocol.self)")
                  )
```

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodeFile.swift
```diff
@@ -0,0 +1,432 @@
...
+}
+
+private func generateInitializerDecl(children: [Child], optionalBaseAsMissing: Bool, currentIndentation: String = "") -> DeclSyntax {
+    if children.isEmpty {
+        return "public init()"
+    }
+
+    var genericParamNames: [String: Int] = [:]
+    var genericParams: [String] = []
+    var params = ["leadingTrivia: Trivia? = nil"]
```
ahoppen, line 338: Not part of this PR but maybe this could be a reason to make `FunctionParameterSyntax` `ExpressibleByStringInterpolation` so `params` can be of type `[FunctionParameterSyntax]`.

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodeFile.swift
```diff
@@ -0,0 +1,432 @@
...
+            expression: ExprSyntax(".node(\(raw: choice.typeName).self)"))
+        }
+      }
+
+      StmtSyntax("return .choices(\(choices))")
+    }
+  }
+}
+
+private func generateInitializerDecl(children: [Child], optionalBaseAsMissing: Bool, currentIndentation: String = "") -> DeclSyntax {
```
ahoppen, line 331: I would expect that BasicFormat handles indentation so we shouldn’t need `currentIndentation` and all the indentation handling below.

---

Also, I would rename this to `generateInitializerDeclHeader` because it’s not generating the full initializer decl, just its header without the body. This could also return a `PartialSyntaxNodeString` instead of a `DeclSyntax` to indicate that the returned value doesn’t have a body.

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodeFile.swift
```diff
@@ -0,0 +1,432 @@
...
+      public init\(raw: generics)(
+      \(raw: paramIndentation)\(raw: params.joined(separator: separator))
+      \(raw: currentIndentation))
+      """
+}
+
+/// If the type has a default value, return an expression of the form
+/// ` = default_value` that can be used as the default value to for a
+/// function parameter. Otherwise, return an empty string.
+private func generateDefaultInitialization(for child: Child) -> String {
```
ahoppen, line 395: What do you think about making this an extension on `Child`?

```swift
fileprivate extension Child {
  var defaultInitialization: String { … }
}
```

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodeFile.swift
kimdv: Copied and pasted!

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodeFile.swift
kimdv: Should we also make the `generateInitializerDecl` above as an extension?
Then we don't need to pass the children list.

ahoppen: Yes, I think that would make sense.

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodeFile.swift
kimdv: Did the same for he generic parameters

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodeFile.swift
```diff
@@ -0,0 +1,494 @@
...
+fileprivate extension Node {
+  func generateInitializerDeclHeader(optionalBaseAsMissing: Bool) -> PartialSyntaxNodeString {
+    if children.isEmpty {
+      return "public init()"
+    }
+
+    var genericParamNames: [String: Int] = [:]
+    var genericParams: [String] = []
+
+    func createFunctionParameterSyntax(for child: Child, optionalBaseAsMissing: Bool) -> FunctionParameterSyntax {
```
kimdv, line 373: It's not possible to store paramType inside a result builder. 
So made this function here to have the logic of creating a function parameter syntax

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodeFile.swift
```diff
@@ -0,0 +1,494 @@
...
+          /// is undefined.
+          internal init(_ data: SyntaxData) {
+            assert(data.raw.kind == .\(raw: node.swiftSyntaxKind))
+            self._syntaxNode = Syntax(data)
+          }
+          """
+        )
+
+        try! InitializerDeclSyntax("\(node.generateInitializerDeclHeader(optionalBaseAsMissing: false))") {
+          let parameters = node.children.map { $0.swiftName }.joined(separator: ", ")
```
ahoppen, line 69: I know that I wrote this but reading it again, we should be using syntax nodes here as well:

```swift
let parameters = FunctionParameterListSyntax {
  for child in node.children {
    FunctionParameterSyntax(firstName: child.swiftName)
  }
}
```

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodeFile.swift
```diff
@@ -0,0 +1,494 @@
...
+                  for child in node.children {
+                    DictionaryElementSyntax(
+                      leadingTrivia: .newline,
+                      keyExpression: ExprSyntax(#""\#(raw: child.swiftName)""#),
+                      valueExpression: child.isOptional
+                        ? ExprSyntax("\(raw: child.swiftName).map(Syntax.init)?.asProtocol(SyntaxProtocol.self) as Any")
+                        : ExprSyntax("Syntax(\(raw: child.swiftName)).asProtocol(SyntaxProtocol.self)")
+                    )
+                  }
+                }
```
ahoppen, line 282: This is nested pretty deep. What do you think about extracting the `DictionaryExprSyntax` into a variable?

```swift
let children = DictionaryExprSyntax { …
```

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodeFile.swift
```diff
@@ -0,0 +1,494 @@
...
+        """
+    }
+  }
+}
+
+fileprivate extension Child {
+  /// If the type has a default value, return an expression of the form
+  /// ` = default_value` that can be used as the default value to for a
+  /// function parameter. Otherwise, return an empty string.
+  var defaultInitialization: String {
```
ahoppen, line 446: This could return `InitializerClauseSyntax?` instead of a string for a little more structure.

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodeFile.swift
ahoppen: Nice.

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodeFile.swift
```diff
@@ -0,0 +1,503 @@
...
+            /// the base type would allow implicit conversion from a string literal,
+            /// which the above initializer doesn't support.
+            \(node.generateInitializerDeclHeader(optionalBaseAsMissing: true))
+            """
+          ) {
+            FunctionCallExprSyntax(callee: ExprSyntax("self.init")) {
+              TupleExprElementSyntax(label: "leadingTrivia", expression: ExprSyntax("leadingTrivia"))
+              for child in node.children {
+                if child.hasOptionalBaseType {
+                  TupleExprElementSyntax(label: child.swiftName, expression: ExprSyntax("Optional<\(raw: child.typeName)>.none"))
```
ahoppen, line 156: I think the generated source could would be a fair amount more readable if all these parameters are on separate lines. Should be doable by just adding a leading newline to all the `TupleExprElementSyntax` here.

ahoppen (approved):
One comment regarding the format of the generated code, otherwise LGTM. Thanks

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodeFile.swift
kimdv: Also added newline in the init signature right above

ahoppen:
@swift-ci Please test

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test

build-script.py
```diff
@@ -244,61 +235,13 @@ def generate_gyb_file(gyb_file: str) -> None:
...
-
-
 # If `temp_directories` is True, creates a dictionary that maps every source dir in
 # `source_dirs` to a unique temporary directory.
 # If `temp_directories` is False, it maps each source dir to the corresponding
 # gyb_generated directory.
 def gyb_dir_mapping(temp_directories: bool) -> Dict[str, str]:
     source_dirs = [
         SYNTAXSUPPORT_DIR,
-        SWIFTSYNTAX_DIR,
```
kimdv, line 301: @ahoppen the build was failing because of 
```
diff: /home/build-user/swift-syntax/Sources/SwiftSyntax/gyb_generated: No such file or directory
FAIL: Gyb-generated files committed to repository do not match generated ones. Please re-generate the gyb-files using the following command, open a PR to the SwiftSyntax project and merge it alongside the main PR.$ swift-syntax/build-script.py generate-source-code --toolchain /path/to/toolchain.xctoolchain/usr
```

https://ci.swift.org/job/swift-syntax-PR-Linux/1674/console

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test

PR #1294, opened 2023-01-29
Add missing parameter

kimdv:


CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxTraitsFile.swift
```diff
@@ -28,7 +28,7 @@ let syntaxTraitsFile = SourceFileSyntax {
       }
     }
 
-    ExtensionDeclSyntax("public extension \(trait.traitName)Syntax") {
+    ExtensionDeclSyntax("public extension \(raw: trait.traitName)Syntax") {
```
kimdv, line 31: Fixed warning

Sources/SwiftParser/Lexer/Cursor.swift
```diff
@@ -1298,7 +1298,6 @@ extension Lexer.Cursor {
   }
 
   mutating func lexHexNumber() -> Lexer.Result {
-    let tokStart = self
```
kimdv, line 1301: Fixed a warning

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntaxbuilder/BuildableNodesFile.swift
```diff
@@ -103,7 +103,8 @@ private func createConvenienceInitializer(node: Node) -> InitializerDeclSyntax?
 
   return InitializerDeclSyntax(
     leadingTrivia: .docLineComment("/// A convenience initializer that allows initializing syntax collections using result builders") + .newline,
-    modifiers: [DeclModifierSyntax(name: .keyword(.public))],
+    attributes: [.attribute(AttributeSyntax("@_disfavoredOverload"))],
```
kimdv, line 106: Without this, we get a `Ambiguous use of 'init'`

kimdv:
@swift-ci please test

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntaxbuilder/BuildableNodesFile.swift
ahoppen: I just tried it and we only need this if the built type is optional. Only in that case is the parameter is defaulted in both the builder initializer and the normal initializer in SwiftSyntax.

Since these convenience initializers only exist for the result builders now, I think we should not default their value, which also resolves the ambiguity. I just tried that, it requires a few updates to the test cases but IMO they all provide more clarity.

CodeGeneration/Sources/SyntaxSupport/Node.swift
```diff
@@ -117,8 +117,13 @@ public class Node {
...
           ),
-          child,
-        ]
+          child
+        ] + (i == children.count - 1 ? [Child(
+          name: "UnexpectedAfter\(child.name)",
+          kind: "UnexpectedNodes",
+          isOptional: true,
+          collectionElementName: "UnexpectedAfter\(child.name)"
+        )] : [])
```
ahoppen, line 126: I would prefer to add this after the `flatMap` is completed, i.e. have something like `children.enumerated().flatMap { … } + [<unexpectedAfter>]`. I think that reads better.

kimdv:
@swift-ci please test

Tests/SwiftSyntaxBuilderTest/StructTests.swift
```diff
@@ -45,7 +45,8 @@ final class StructTests: XCTestCase {
         .carriageReturns(1),
       ],
       structKeyword: .keyword(.struct),
-      identifier: "CarriateReturnsStruct"
+      identifier: "CarriateReturnsStruct",
+      membersBuilder: {}
```
ahoppen, line 49: It’s not really important but for the empty cases, I would not use the result builders, it somehow seems cleaner to me that way.
```suggestion
      members: MemberDeclBlockSyntax(members: [])
```

Same for the other test cases.

ahoppen (approved):

kimdv:
@swift-ci please test

Issue #1292, opened 2023-01-27
[Source compatibility suite] SwiftLint failing to build - new Swift parser generated errors for code that C++ parser accepted


justice-adams-apple:

ahoppen:
Reduced reproducer:

```swift
#file == $0.path
```

@DougGregor we seem to be parsing `#file` as a declaration macro and then we don’t know what to do with the trailing `== $0.path`

```swift
_ = #file == $0.path
```

parses fine.

ahoppen:
rdar://104782355

DougGregor:
Thank you! I'm working on a fix for this.

PR #1296, opened 2023-01-30
[Parser] Go back to parsing macro expansions as expressions where possible

DougGregor:
The change to eagerly parsing macro expansions as declarations rather
than expressions caused a regression for expressions that start with a
macro expansion, e.g., `#fileID == "Module/filename.swift"`. Rather
than perform extensive lookahead to disambiguate, go back to parsing
as a macro expansion expression except in places where only a
declaration is allowed (e.g., a member list).

To accommodate this for freestanding macro implementations, generalize
the `CodeItemMacro` and `DeclarationMacro` expansion operations to
deal with an arbitrary `FreestandingMacroExpansionSyntax`, as we had
already done for `ExpressionMacro`.

Fixes https://github.com/apple/swift-syntax/issues/1292.


DougGregor:
@swift-ci please test

ahoppen (approved):
Thank you

PR #1298, opened 2023-01-31
Allow node children to override whether a token needs a leading/trailing space

ahoppen:
The motivating example was that we should not add a space after `for` in `@_dynamicReplacement(for:…)`.

IMO, this also cleans up a few rules in BasicFormat.


kimdv (approved):

bnbarham (approved):
Shouldn't this need a fix in `Tests/SwiftParserTest/AttributeTests.swift`?

ahoppen:
That test will only be broken by https://github.com/apple/swift-syntax/pull/1180

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

PR #1301, opened 2023-02-01
Format codegen files

kimdv:
Codegen was not formatted. IMO we should do that to keep the code consistent


kimdv:
This should properly be coordinated with some of the other PR's before merge

ahoppen:
I deliberately decided not to format the generated files because it simplifies development: If you make changes to `CodeGenerateion` and re-generate the files, you aren’t hit with a wall of changes because now the files are no longer formatter.

If we want to format those files (which I think is a good idea), IMO the formatting should be done in `CodeGeneration` itself at generation time. That way running CodeGeneration will always generate formatted files. For this, CodeGeneration can depend on swift-format, similarly to how it depends on swift-syntax.

kimdv:
I don't format the generated files, but the code generation project 😅

Like template files etc.

format.py
```diff
@@ -87,7 +87,7 @@ def find_swiftformat(swift_format: str) -> Path:
 def get_files_to_format() -> List[Path]:
     package_dir = Path(__file__).parent
     files_to_format: List[Path] = []
-    for swift_file_path in (package_dir / "Sources").glob('**/*.swift'):
+    for swift_file_path in (package_dir).glob('**/Sources/**/*.swift'):
```
ahoppen, line 90: Just a thought: Could we just do the following and then get rid of the `for` loop for `Tests`. AFAICT that would also format the examples, which would be good as well.

```suggestion
    for swift_file_path in package_dir.glob('**/*.swift'):
```

ahoppen (commented):
Oh, I should have looked at the PR more closely, sorry. `CodeGeneration` itself should certainly be formatted by `swift-format`. Thanks for catching and fixing this. I wonder whether we can generalize this a little further to also include the examples (see inline comment).

CodeGeneration/Package.swift
```diff
@@ -5,10 +5,10 @@ import PackageDescription
 let package = Package(
   name: "CodeGeneration",
   platforms: [
-      .macOS(.v10_15),
-    ],
+    .macOS(.v10_15)
+  ],
```
kimdv, line 9: Right now we also include `Package.swift` files. Not sure if we should do that.

ahoppen: Why not. I don’t see any reason why it shouldn’t be formatted.

format.py
```diff
@@ -86,22 +86,21 @@ def find_swiftformat(swift_format: str) -> Path:
...
+    # Don't format gyb_generated files
+    files_to_exclude.update(package_dir.glob('**/gyb_generated/**/*.swift'))
+    # Don't generate lit_tests files
+    files_to_exclude.update(package_dir.glob('**/lit_tests/**/*.swift'))
+    # Don't format generated files
+    files_to_exclude.update(package_dir.glob('**/generated/**/*.swift'))
+    # Don't format .build folder and content
+    files_to_exclude.update(package_dir.glob('**/.build/**/*.swift'))
+    # Don't format test input files
+    files_to_exclude.update(package_dir.glob('**/Inputs/**/*.swift'))
```
ahoppen, line 101: This looks really nice 😍

ahoppen (approved):
Thanks for catching that we didn’t format all of these files and fixing it.

kimdv:
@swift-ci plese test

kimdv:
@swift-ci please test

Issue #1221, opened 2023-01-11
swift should offer fixit for "#elif"


ahoppen:
### Description

Swift offers a fixit for #else if -> #elseif.

We should also offer a fixit for #elif, eg:

```
error: use of unknown directive '#elif'
                    #elif arch(arm64)
                    ^
```
---

```
error: unexpected 'if' keyword following '#else' conditional compilation directive; did you mean '#elseif'?
                    #else if arch(arm64)
                    ~~~~~~^~
                    #elseif
```

rdar://103506087

TiagoMaiaL:
I've started investigating this issue. I've used the following code to check on the parser diagnostics:
```swift
        let sourceFile = Parser.parse(source: """
            #if arch(arm64)
            debugPrint("arm64")
            #elif arch(x86_64)
            debugPrint("x86_64")
            #endif
        """)
        
        let diagnostics = ParseDiagnosticsGenerator.diagnostics(for: sourceFile)
        print(diagnostics)
        print(diagnostics.flatMap(\.fixIts).map(\.message.message))
```

The above code prints the following:
```
[3:6: consecutive statements on a line must be separated by ';']
["insert \';\'"]
```

> Swift offers a fixit for #else if -> #elseif.

I've tried using ´#else if´ instead of ´#elseif´ in the code above, but here's the emitted diagnostic (which is different from the message in this issue):
```
[3:22: expected '{' in 'if' statement, 4:21: expected '}' to end 'if' statement]
["insert \'{\'", "insert \'}\'"]
```

@ahoppen Am I looking at the right place? Should I be looking for `fixIts` in `ParseDiagnosticsGenerator`?

Thanks in advance.

bnbarham:
> Swift offers a fixit for #else if -> #elseif.

Is referring to the C++ parser in the Swift compiler. It outputs a useful diagnostic on a `#elseif`. This bug is about adding a similar diagnostic to this parser + also handling `elif` as the current diagnostics for these cases aren't great (as you saw).

https://github.com/apple/swift-syntax/blob/main/Sources/SwiftParser/SwiftParser.docc/FixingBugs.md talks about this a little. In this case `parsePoundIfDirective` will need to be updated to produce an unexpected node for the `#else` + `if` and then a missing node for `#elseif`. Similar for the `#elif` case, ie. an unexpected for `#elif` and then missing `#elseif`. After that you'll want to add a custom diagnostic in the diagnostics generator.

[...]

PR #1302, opened 2023-02-01
Produce meaningful diagnostic for '#else if' typo

TiagoMaiaL:
The purpose of this PR is to fix the first part of issue https://github.com/apple/swift-syntax/issues/1221: produce a meaningful diagnostic for the common `#else if` typo.

I've followed the tips given by @bnbarham and the ones from the [fixingBugs document](https://github.com/apple/swift-syntax/blob/main/Sources/SwiftParser/SwiftParser.docc/FixingBugs.md). Here's what I did so far:
1. Added a test inside `ifconfigExprTests` (it's currently failing because the diagnostics include more errors than the one I'm expecting, I must be doing something wrong)
2. Generated an unexpected token for the `#else if` typo, and a missing one for `#elseif`, in `parsePoundIfDirective`, inside `Directives.swift` (I'm not sure if I'm using the right methods)
3. Added code to generate correct diagnostics for `IfConfigDeclSyntax`, in `ParseDiagnosticsGenerator.swift`

Any tip or help to get this right is really appreciated. Thank you in advance.

### DIFF
--- a/Sources/SwiftParser/Directives.swift
+++ b/Sources/SwiftParser/Directives.swift
@@ -106,7 +106,7 @@ extension Parser {
       while let poundIfHandle = firstIteration ? self.canRecoverTo(.poundIfKeyword) : self.canRecoverTo(anyIn: IfConfigContinuationClauseStartKeyword.self)?.handle,
         loopProgress.evaluate(self.currentToken)
       {
-        let (unexpectedBeforePoundIf, poundIf) = self.eat(poundIfHandle)
+        var (unexpectedBeforePoundIf, poundIf) = self.eat(poundIfHandle)
         firstIteration = false
         // Parse the condition.
         let condition: RawExprSyntax?
@@ -114,7 +114,13 @@ extension Parser {
         case .poundIfKeyword, .poundElseifKeyword:
           condition = RawExprSyntax(self.parseSequenceExpression(.basic, forDirective: true))
         case .poundElseKeyword:
-          condition = nil
+          if let ifToken = self.consume(if: .init(.if, allowAtStartOfLine: false)) {
+            unexpectedBeforePoundIf = RawUnexpectedNodesSyntax(combining: unexpectedBeforePoundIf, poundIf, ifToken, arena: self.arena)
+            poundIf = self.missingToken(.poundElseifKeyword)
+            condition = RawExprSyntax(self.parseSequenceExpression(.basic, forDirective: true))
+          } else {
+            condition = nil
+          }
         default:
           preconditionFailure("The loop condition should guarantee that we are at one of these tokens")
         }
--- a/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
+++ b/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
@@ -668,6 +668,37 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
     return .visitChildren
   }
 
+  public override func visit(_ node: IfConfigDeclSyntax) -> SyntaxVisitorContinueKind {
+    for clause in node.clauses where clause.hasError {
+      if let unexpectedBeforePoundKeyword = clause.unexpectedBeforePoundKeyword,
+        clause.poundKeyword.tokenKind == .poundElseifKeyword,
+        clause.poundKeyword.presence == .missing
+      {
+        let unexpectedTokens =
+          unexpectedBeforePoundKeyword
+          .suffix(2)
+          .compactMap { $0.as(TokenSyntax.self) }
+        if unexpectedTokens.map(\.tokenKind) == [.poundElseKeyword, .keyword(.if)] {
+          addDiagnostic(
+            unexpectedBeforePoundKeyword,
+            StaticParserError.unexpectedPoundElseSpaceIf,
+            fixIts: [
+              FixIt(
+                message: ReplaceTokensFixIt(replaceTokens: unexpectedTokens, replacement: clause.poundKeyword),
+                changes: [
+                  .makeMissing(unexpectedBeforePoundKeyword, transferTrivia: false),
+                  .makePresent(clause.poundKeyword, leadingTrivia: unexpectedBeforePoundKeyword.leadingTrivia),
+                ]
+              )
+            ],
+            handledNodes: [unexpectedBeforePoundKeyword.id, clause.poundKeyword.id]
+          )
+        }
+      }
+    }
+    return .visitChildren
+  }
+
   public override func visit(_ node: InitializerClauseSyntax) -> SyntaxVisitorContinueKind {
     if shouldSkip(node) {
       return .skipChildren
--- a/Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift
+++ b/Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift
@@ -197,6 +197,9 @@ extension DiagnosticMessage where Self == StaticParserError {
   public static var tryOnInitialValueExpression: Self {
     .init("'try' must be placed on the initial value expression")
   }
+  public static var unexpectedPoundElseSpaceIf: Self {
+    .init("unexpected 'if' keyword following '#else' conditional compilation directive; did you mean '#elseif'?")
+  }
   public static var unexpectedSemicolon: Self {
     .init("unexpected ';' separator")
   }
--- a/Tests/SwiftParserTest/translated/IfconfigExprTests.swift
+++ b/Tests/SwiftParserTest/translated/IfconfigExprTests.swift
@@ -428,6 +428,35 @@ final class IfconfigExprTests: XCTestCase {
     )
   }
 
+  func testIfconfigExpr31() {
+    AssertParse(
+      """
+      #if arch(x86_64)
+        debugPrint("x86_64")
+      1️⃣#else if arch(arm64)
+        debugPrint("arm64")
+      #else
+        debugPrint("Some other architecture.")
+      #endif
+      """,
+      diagnostics: [
+        DiagnosticSpec(
+          message: "unexpected 'if' keyword following '#else' conditional compilation directive; did you mean '#elseif'?",
+          fixIts: ["replace '#else if' with '#elseif'"]
+        )
+      ],
+      fixedSource: """
+        #if arch(x86_64)
+          debugPrint("x86_64")
+        #elseif arch(arm64)
+          debugPrint("arm64")
+        #else
+          debugPrint("Some other architecture.")
+        #endif
+        """
+    )
+  }
+
   func testUnknownPlatform1() {
     AssertParse(
       """


Sources/SwiftParser/Directives.swift
```diff
@@ -86,14 +86,20 @@ extension Parser {
...
         let condition: RawExprSyntax?
         switch poundIf.tokenKind {
         case .poundIfKeyword, .poundElseifKeyword:
           condition = RawExprSyntax(self.parseSequenceExpression(.basic, forDirective: true))
         case .poundElseKeyword:
+          if self.currentToken.rawTokenKind == .keyword(.if),
+              let poundElseHandle = self.canRecoverTo(.poundElseKeyword) {
+            (unexpectedBetweenPoundKeywordAndCondition, _) = self.eat(poundElseHandle)
+            poundIf = self.missingToken(.poundElseifKeyword)
+          }
```
ahoppen, line 102: I see three issues with this approach

- We eat all tokens up to the next the next `#else` as unexpected tokens when you call `self.eat(poundElseHandle)`. We should be treating the `#else if` as if it was spelled as `#elseif` parse anything in between the `#else if` and `#else` as normal elements in the `IfConfigClauseSyntax`
- AFAICT we are dropping the `#else` you parse with `self.eat(poundElseHandle)` on the floor and thus loosing [source fidelity](https://swiftpackageindex.com/apple/swift-syntax/main/documentation/swiftparser/parserdesign#Source-Fidelity). I’m surprised this isn’t caught by the test case you added.

I think what you want to do instead is:
- Check whether the currentToken is an `if` keyword (you can use `self.at(.keyword(.if))` for that, which is slightly shorter than what you wrote but functionally equivalent)
- If you discover that case, you should shove the `#else` and `if` into `unexpectedBeforePoundIf` so they are not lost in the syntax tree and we maintain source fidelity
- Synthesize the `poundIf` as a `poundElseifKeyword` like you already do.

Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift
```diff
@@ -360,6 +360,12 @@ public struct TryCannotBeUsed: ParserError {
   }
 }
 
+public struct PoundElseSpaceIfTypo: ParserError {
+  public var message: String {
+    "error: unexpected 'if' keyword following '#else' conditional compilation directive; did you mean '#elseif'?"
+  }
+}
+
```
ahoppen, line 368: Since this error doesn’t contain any variables, you can also put it in `StaticParserErrors`, which is a little more succinct.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -647,6 +647,30 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
     }
     return .visitChildren
   }
+  
+  public override func visit(_ node: IfConfigDeclSyntax) -> SyntaxVisitorContinueKind {
+    for clause in node.clauses where clause.hasError {
+      if let unexpected = clause.unexpectedBetweenPoundKeywordAndCondition,
+         clause.poundKeyword.tokenKind == .poundElseifKeyword,
+         clause.poundKeyword.presence == .missing {
```
ahoppen, line 655: With my comment above, I think you should also check that `unexpected` contains `#else` and `if` as its last two elements to make sure we really have the `#else if` case and didn’t randomly recover to `#elseif` by e.g. skipping an unexpected identifier.

Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift
```diff
@@ -542,3 +548,24 @@ public struct ReplaceTokensFixIt: ParserFixIt {
     "replace \(nodesDescription(replaceTokens, format: false)) by '\(replacement.text)'"
   }
 }
+
+public struct ReplacePoundElseSpaceIfTypo: ParserFixIt {
```
ahoppen, line 552: Fix-Its in SwiftParser describe the action they are performing (for example see how the messages above say e.g. `insert newline`). So this message should be `replace '#else if' by '#elseif'`. We have that exact message above, so you should be able to reuse that exact message in the diagnostic generator by calling the initializer of `ReplaceTokensFixIt`. You can search the diagnostic generator for examples how to do that.

Tests/SwiftParserTest/translated/IfconfigExprTests.swift
```diff
@@ -427,6 +427,28 @@ final class IfconfigExprTests: XCTestCase {
...
+        debugPrint("Some other architecture.")
+      #endif
+      """,
+      diagnostics: [
+        DiagnosticSpec(message: """
+        error: unexpected 'if' keyword following '#else' conditional compilation directive; did you mean '#elseif'?
+                            #else if arch(arm64)
+                            ~~~~~~^~
+                            #elseif
+        """)
```
ahoppen, line 448: The message of the diagnostic doesn’t include the `error: ` prefix, it’s just `unexpected 'if' keyword following '#else' conditional compilation directive; did you mean '#elseif'?`. The fix-its also aren’t part of the message and should be specified in the `fixIts` parameter to `DiagnsoticSpec`. So this should be something like the following (written in GitHub so it might not be 100% correct)

```suggestion
        DiagnosticSpec(message: "unexpected 'if' keyword following '#else' conditional compilation directive; did you mean '#elseif'?", fixIts: ["replace '#else if' by '#elseif"])
```

ahoppen (commented):
Hey @TiagoMaiaL,

thanks for working on this. I just took a look at your changes and noticed a few things to which I put comments inline. Can you take a look whether they make sense to you. If they don’t, feel free to reach out again.

Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift
TiagoMaiaL: Done

Sources/SwiftParser/Directives.swift
TiagoMaiaL: Regarding your suggestions:
1. > Check whether the currentToken is an `if`

I'm using the `consume(if:)` method, which returns a `RawTokenSyntax` I can inform as unexpected (item 2):
```swift
if let ifToken = self.consume(if: .keyword(.if), allowTokenAtStartOfLine: false) {
```

2. > you should shove the `#else` and `if` into `unexpectedBeforePoundIf`

The only way I figured out how to do this was by using `RawUnexpectedNodesSyntax` combining initializers:
```swift
unexpectedBeforePoundIf = RawUnexpectedNodesSyntax(combining: unexpectedBeforePoundIf, poundIf, ifToken, arena: self.arena)
```

Is this acceptable?

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
TiagoMaiaL: Here's how I'm doing it (if there's a more succinct and readable way, I'm all ears):
```swift
let unexpectedTokens = unexpectedBeforePoundKeyword
  .flatMap { $0.tokens(viewMode: .sourceAccurate) }
let lastTwoTokenKinds = unexpectedTokens
  .suffix(2)
  .map(\.tokenKind)
        
if lastTwoTokenKinds == [.poundElseKeyword, .keyword(.if)] {
  var fixedClause = clause
  fixedClause.poundKeyword = .poundElseifKeyword()
          
  addDiagnostic(
    unexpectedBeforePoundKeyword,
    StaticParserError.unexpectedPoundElseSpaceIf,
    fixIts: [
      FixIt(
        message: ReplaceTokensFixIt(replaceTokens: unexpectedTokens, replacement: fixedClause.poundKeyword),
        changes: [.replace(oldNode: Syntax(clause), newNode: Syntax(fixedClause))]
      )
    ]
  )
}
```

Tests/SwiftParserTest/translated/IfconfigExprTests.swift
TiagoMaiaL: I've included some more `DiagnosticSpec`s because the parser is producing more than the one I previously added (`"unexpected 'if' keyword following '#else' conditional compilation directive; did you mean '#elseif'?"`). I think it's correct.

Is this ok?

Sources/SwiftParser/Directives.swift
ahoppen: Yes, both solutions not only acceptable but the right way to go. 👍

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
ahoppen: You can shorten the first lines a little bit by doing the following but other than that, it’s the same way I would do it.

```swift
let unexpected = unexpectedBeforePoundKeyword.suffix(2).map { $0.as(TokenSyntax.self)?.tokenKind }
```

---

In the Fix-It you are currently replacing the entire node, which won’t render nicely in editors (e.g. the command line output will underline the entire `#else if` including the condition and suggest to replace it while keeping the condition. Instead, you should make the `poundKeyword` present and the unexpected nodes missing like the following.

```swift
[
  .makeMissing(unexpectedBeforePoundKeyword),
  .makePresent(clause.poundKeyword)
]
```

TiagoMaiaL: Thank you for the suggestion. I've just pushed the changes.

Tests/SwiftParserTest/translated/IfconfigExprTests.swift
TiagoMaiaL: With the `handledNodes` adjustment you suggested above, the extra diagnostics are gone. I've adjusted the test accordingly.

TiagoMaiaL:
@swift-ci please test

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -647,6 +647,32 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
+  
+  public override func visit(_ node: IfConfigDeclSyntax) -> SyntaxVisitorContinueKind {
+    for clause in node.clauses where clause.hasError {
+      if let unexpectedBeforePoundKeyword = clause.unexpectedBeforePoundKeyword,
+         clause.poundKeyword.tokenKind == .poundElseifKeyword,
+         clause.poundKeyword.presence == .missing {
+        let unexpectedTokens = unexpectedBeforePoundKeyword
+          .suffix(2)
+          .compactMap { $0.as(TokenSyntax.self) }
+        if unexpectedTokens.map(\.tokenKind) == [.poundElseKeyword, .keyword(.if)] {
```
ahoppen, line 659: I think the two `if`s can again be merged

```suggestion
          .compactMap { $0.as(TokenSyntax.self) },
          unexpectedTokens.map(\.tokenKind) == [.poundElseKeyword, .keyword(.if)] {
```

Tests/SwiftParserTest/translated/IfconfigExprTests.swift
```diff
@@ -427,6 +427,26 @@ final class IfconfigExprTests: XCTestCase {
...
+      #else
+        debugPrint("Some other architecture.")
+      #endif
+      """,
+      diagnostics: [
+        DiagnosticSpec(
+          message: "unexpected 'if' keyword following '#else' conditional compilation directive; did you mean '#elseif'?",
+          fixIts: ["replace '#else if' by '#elseif'"]
+        ),
+      ]
```
ahoppen, line 447: And one more request for the test. You can check that applying the Fix-It actually produces the expected source by  adding a `fixedSource:` parameter to `AssertParse`.

ahoppen (commented):
Thanks. This is looking great now. I’ve got two more minor comments inline. After you addressed those, we can ship this.

And in case you’re wondering why your test request didn’t trigger anything: Only people with [Commit Access](https://www.swift.org/contributing/#commit-access) can trigger CI testing.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
TiagoMaiaL: The problem I see here is that the `unexpectedTokens` constant isn't part of the if statement above, and it's not `Optional`, so I can't include all of them in the same if statement:
```swift
  if let unexpectedBeforePoundKeyword = clause.unexpectedBeforePoundKeyword,
     clause.poundKeyword.tokenKind == .poundElseifKeyword,
     clause.poundKeyword.presence == .missing { // End of first if statement
       let unexpectedTokens = unexpectedBeforePoundKeyword // Constant declaration (non-optional)
         .suffix(2)
         .compactMap { $0.as(TokenSyntax.self) }
       if unexpectedTokens.map(\.tokenKind) == [.poundElseKeyword, .keyword(.if)] { // Second if statement
         // ...
       }
```

Tests/SwiftParserTest/translated/IfconfigExprTests.swift
TiagoMaiaL: Just did it. The `fixedSource` is being generated with a wrong indentation (it has a space before `#elseif`). Maybe this is happening because of the `locationMarker`?
```swift
      fixedSource: """
        #if arch(x86_64)
          debugPrint("x86_64")
         #elseif arch(arm64)
          debugPrint("arm64")
        #else
          debugPrint("Some other architecture.")
        #endif
        """
```

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
ahoppen: Oh, sorry, I misread that on GitHub. You’re absolutely right. We can’t merge the `if`s in that case.

Tests/SwiftParserTest/translated/IfconfigExprTests.swift
ahoppen: Oh, good thing we checked it then. I think what’s happening is that the diagnostic generator is thinking that you want to replace the `if` by `#elseif` and is thus retaining the trailing space of the `if`. I haven’t tried it but you should be able to fix this by either setting `transferTrivia: false` in `makeMissing` or `leadingTrivia: []` in `makePresent`. The former would be the preferred solution.

TiagoMaiaL: Thanks for the tip. 

`transferTrivia` alone didn't fix it, as it removed the newline in the source code. The empty array for `leadingTrivia` in `makePresent` wasn't enough either. Here's what made it work:
```swift
               [
                  .makeMissing(unexpectedBeforePoundKeyword, transferTrivia: false),
                  .makePresent(clause.poundKeyword, leadingTrivia: unexpectedBeforePoundKeyword.leadingTrivia)]
               ]
```

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
ahoppen: Great. That’s exactly the right solution.

ahoppen:
@swift-ci Please test

ahoppen (approved):

ahoppen:
Oh, one last request: Could you rebase and squash your commits on top of `main`? It just makes for a nicer git history.

Sources/SwiftParser/Directives.swift
```diff
@@ -106,15 +106,21 @@ extension Parser {
...
+        var (unexpectedBeforePoundIf, poundIf) = self.eat(poundIfHandle)
         firstIteration = false
         // Parse the condition.
         let condition: RawExprSyntax?
         switch poundIf.tokenKind {
         case .poundIfKeyword, .poundElseifKeyword:
           condition = RawExprSyntax(self.parseSequenceExpression(.basic, forDirective: true))
         case .poundElseKeyword:
-          condition = nil
+          if let ifToken = self.consume(if: .init(.if, allowAtStartOfLine: false)) {
```
TiagoMaiaL, line 117: @ahoppen I had to change this line to use the new `consume` API (allowsAtStartOfLine got removed as a parameter). Is this method usage correct?

ahoppen:
Thanks for squashing the commits. Let’s 🚢it.

@swift-ci Please test

TiagoMaiaL:
@ahoppen I've fixed the formatting issues in the code by running `format.py`. The tests are passing. 
Was formatting the only issue regarding this PR?

ahoppen:
Let’s run CI again and see. If it got to checking the formatting, everything should be good.

@swift-ci Please test

TiagoMaiaL:
Thank you for your help, @ahoppen

PR #1310, opened 2023-02-03
Allow keyword init as a `@derivative` argument

bnbarham:


bnbarham:
@swift-ci please test

bnbarham:
I'd be happy to get this in as a quick stop-gap. But IMO we probably want to parse these differently. `parseDeclNameRef` in the old parser parses these as identifiers rather than keywords - https://github.com/apple/swift/blob/2b542cba37421f36c2a63358ebd109c2fb0ef53d/lib/Parse/ParseExpr.cpp#L2181. We're also completely ignoring `keywordsUsingSpecialNames`, though the documentation there is somewhat confusing 🤔.

ahoppen:
@swift-ci Please test

bnbarham:
@swift-ci please test

PR #1311, opened 2023-02-03
Disable assertions that check TokenKind

ahoppen:
I only now noticed that normal CI doesn't have assertions enabled and debug CI seems to fail in a variety of places with this assertion.

Disable the assertion for now. The path forward is to enable assertions in SwiftSyntax for Release+Assert builds and then re-apply this fix, fixing any issues that CI now discovers.


ahoppen:
@swift-ci Please test

PR #1315, opened 2023-02-06
Fix formatting for init followed by a paren

kimdv:


Sources/SwiftBasicFormat/generated/BasicFormat.swift
```diff
@@ -174,12 +174,13 @@ open class BasicFormat: SyntaxRewriter {
...
-    case (.keyword(.as), .exclamationMark), // Ensures there is not space in `as!`
-     (.keyword(.as), .postfixQuestionMark), // Ensures there is not space in `as?`
-     (.exclamationMark, .leftParen), // Ensures there is not space in `myOptionalClosure!()`
+    case (.exclamationMark, .leftParen), // Ensures there is not space in `myOptionalClosure!()`
      (.exclamationMark, .period), // Ensures there is not space in `myOptionalBar!.foo()`
+     (.keyword(.as), .exclamationMark), // Ensures there is not space in `as!`
+     (.keyword(.as), .postfixQuestionMark), // Ensures there is not space in `as?`
      (.keyword(.`init`), .leftParen), // Ensures there is not space in `init()`
      (.keyword(.`init`), .postfixQuestionMark), // Ensures there is not space in `init?`
+     (.keyword(.`init`), .rightParen), // Ensures there is not space in `.map(Foo.init)`
```
kimdv, line 183: Maybe we should make it more general? So when I have a something `["foo", "bar", "baz"].map(MyStruct.someStaticMethod)`

ahoppen: I just merged https://github.com/apple/swift-syntax/pull/1298, which should allow you to override `requires_trailing_space` for `Name` in a `MemberAccessExpr`. 

Which just got me thinking whether `init` shouldn’t require a trailing space by default. I can’t think of a place where it does require a trailing space right now.

kimdv:
@swift-ci please test

Tests/SwiftSyntaxBuilderTest/ReturnStmsTests.swift
```diff
@@ -45,6 +37,10 @@ final class ReturnStmtTests: XCTestCase {
         StmtSyntax("return !myBool"),
         "return !myBool"
       ),
+      #line: (
+        StmtSyntax("return data.child(at: 2, parent: Syntax(self)).map(UnexpectedNodesSyntax.init)"),
+        "return data.child(at: 2, parent: Syntax(self)).map(UnexpectedNodesSyntax.init)"
```
kimdv, line 42: Added a test case to ensure don't add a leading/trailing space in this example

ahoppen (approved):
Nice

PR #1316, opened 2023-02-06
Update swift syntax

kimdv:
Updated SwiftSyntax and did some cleaning


kimdv:
@swift-ci please test

kimdv:
@swift-ci please test

Sources/SwiftSyntaxBuilder/generated/BuildableNodes.swift
```diff
@@ -17,8 +17,7 @@ import SwiftSyntax
...
+  public init(leadingTrivia: Trivia? = nil, unexpectedBeforeAttributes: UnexpectedNodesSyntax? = nil, attributes: AttributeListSyntax? = nil, unexpectedBetweenAttributesAndModifier: UnexpectedNodesSyntax? = nil, modifier: DeclModifierSyntax? = nil, unexpectedBetweenModifierAndAccessorKind: UnexpectedNodesSyntax? = nil, accessorKind: TokenSyntax, unexpectedBetweenAccessorKindAndParameter: UnexpectedNodesSyntax? = nil, parameter: AccessorParameterSyntax? = nil, unexpectedBetweenParameterAndEffectSpecifiers: UnexpectedNodesSyntax? = nil, effectSpecifiers: DeclEffectSpecifiersSyntax? = nil, unexpectedBetweenEffectSpecifiersAndBody: UnexpectedNodesSyntax? = nil, unexpectedAfterBody: UnexpectedNodesSyntax? = nil, @CodeBlockItemListBuilder bodyBuilder: () throws -> CodeBlockItemListSyntax?, trailingTrivia: Trivia? = nil) rethrows {
```
ahoppen, line 20: Not related to this PR but it would be nice if all those parameters also were on separate lines. What we could do is to add a rule to `CodeGenerationFormat` that puts the parameters on separate lines if there are more than, say, 3.

I did something similar for `debugInitCall` where I put parameters on separate lines if there is more than 1

https://github.com/apple/swift-syntax/blob/b8ce4f6f91eac8c1d24660ca53ab84c49adcc645/Sources/_SwiftSyntaxTestSupport/SyntaxProtocol%2BInitializer.swift#L46-L48

ahoppen (approved):
Thank you

Sources/SwiftSyntaxBuilder/generated/BuildableNodes.swift
kimdv: We might d the same for `ArrayExprSyntax`

ahoppen: Yes, that also seems reasonable to me. And if you mention `ArrayExprSyntax`, `DictionaryExprSyntax` also seems like a good candidate.

kimdv:
@swift-ci please test linux

PR #1317, opened 2023-02-07
[Macros] Use a type parameter conforming to DeclSyntaxProtocol in the PeerMacro protocol requirement.

hborla:
I believe this is the last step in generalizing all `AttachedMacro` protocol refinements to use type parameters instead of `DeclSyntax` in the `expansion` requirement.


hborla:
@swift-ci please test

DougGregor (approved):

hborla:
@swift-ci please test macOS

PR #1318, opened 2023-02-07
[Macros] generate diagnostics from thrown error

elizablock:
Macro expansion methods throw an error to indicate that the expansion failed. This patch allows the error thrown to add full-fledged diagnostics, rather than autogenerating a diagnostic based on the error's description.

It adds an error protocol:
```swift
public protocol DiagnosticsProviding: Error {
  var diagnostics: [Diagnostic] { get }
}
```

...and a utility method on `MacroExpansionContext` to add diagnostics from an error. The implementation of this method checks if the error is diagnostic providing; if not, it generates the same diagnostic as before. If so, it adds all the provided diagnostics.

### DIFF
--- a/Sources/SwiftDiagnostics/Diagnostic.swift
+++ b/Sources/SwiftDiagnostics/Diagnostic.swift
@@ -71,3 +71,9 @@ public struct Diagnostic: CustomDebugStringConvertible {
     return "\(location): \(message)"
   }
 }
+
+public protocol DiagnosticsProviding: Error {
+  /// The diagnostics provided by this error.
+  /// At least one diagnostic should have `severity == .error`.
+  var diagnostics: [Diagnostic] { get }
+}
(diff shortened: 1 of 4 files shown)


elizablock:
@swift-ci please test

ahoppen (approved):
This looks good to me. I’m not a huge fan of the name `DiagnosticProviding` but I don’t have any better names so it’s good with me.

elizablock:
> This looks good to me. I’m not a huge fan of the name `DiagnosticProviding` but I don’t have any better names so it’s good with me.

Happy to rename ... `DiagnosticError`? `DiagnosticGenerating`?

elizablock:
@swift-ci please test

ahoppen:
What about naming it `DiagnosticsError`? If we do that, we can also assert in the initializer of `DiagnosticsError` that at least one diagnostics with severity `error` is provided.

PR #1322, opened 2023-02-07
Remove trailing whitespace in multi-line string literals of tests

ahoppen:
In mulit-line string literal tests these trailing whitespaces are really hard to spot but influence the test significantly. To make them more obvious, use `\u{20}`, which is visible in the editor and will be replaced by a normal space in `AssertParse`

In all other cases where the trailing whitespace doesn’t matter, remove it.

### DIFF
--- a/Tests/SwiftParserTest/translated/AsyncTests.swift
+++ b/Tests/SwiftParserTest/translated/AsyncTests.swift
@@ -112,7 +112,7 @@ final class AsyncTests: XCTestCase {
           get {
             return 0
           }
-          set async { 
+          set async {
           }
         }
       }
--- a/Tests/SwiftParserTest/translated/AvailabilityQueryTests.swift
+++ b/Tests/SwiftParserTest/translated/AvailabilityQueryTests.swift
@@ -61,7 +61,7 @@ final class AvailabilityQueryTests: XCTestCase {
   func testAvailabilityQuery5a() {
     AssertParse(
       """
-      if !1️⃣#available(OSX 10.52, *) { 
+      if !1️⃣#available(OSX 10.52, *) {
       }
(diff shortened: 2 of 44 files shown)


ahoppen:
@swift-ci Please test

bnbarham:
How many tests actually need this? Rintaro suggested that we could just have this inline on each of the tests instead. I'd also prefer it not be replaced with anything (ie. replace with empty string) and just be a marker for "keep this whitespace".

allevato:
FWIW, in swift-format we just write literal `\u{0020}` in any strings where we want to trailing space to be explicit, visible, and not removed automatically by tooling, and that doesn't require any custom logic for replacement: https://github.com/apple/swift-format/blob/3cd655ff323bd22be523dd33b5e1cd0b14f6bf0a/Tests/SwiftFormatPrettyPrintTests/CommentTests.swift#L651

ahoppen:
I like the idea of using `\u{20}` and adopted that.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

bnbarham (approved):
The commit comment doesn't *really* apply now, but looks good otherwise.

> which is visible in the editor and will be replaced by a normal space in `AssertParse`

It's not really "replaced" by `AssertParse` any more  - just by the compiler 😅.

ahoppen:
Oh, I changed the commit message, just forgot to update the PR description.

ahoppen:
@swift-ci Please test macOS

bnbarham:
> Oh, I changed the commit message, just forgot to update the PR description.

I was referring to the commit message as well, FWIW.

PR #1324, opened 2023-02-08
Fix colon followed by right paren spacing

kimdv:


kimdv:
@swift-ci please test

ahoppen (commented):
I think what you want to do, is mark the colon in `DeclNameArgument` as not requiring a space so that we also don’t insert a space after `raw:` in `Foo.init(raw:otherParam:)`.

kimdv:
Nice, thanks Alex!

kimdv:
@swift-ci please test

ahoppen (approved):
Great

PR #1327, opened 2023-02-08
Replace DiagnosticsProviding protocol with concrete type

elizablock:
As discussed in #1318, a nicer design for providing diagnostics via thrown error is to expose a concrete error type:
```swift
public struct DiagnosticsError: Error {
  public var diagnostics: [Diagnostic]
}
```

This soft-asserts in init if none of the diagnostics has severity == `.error`, and removes the burden from users of creating their own conforming error type just to throw diagnostics.


elizablock:
@swift-ci please test

ahoppen (approved):
Thank you. Looks good to me.

PR #1333, opened 2023-02-09
Preparation to switch RawTokenKind to be a trivial enum again

ahoppen:
This moves a little more users of `RawTokenKind` to `TokenSpec`, which will make it easier to switch `RawTokenKind` to be a trivial enum again in a follow-up PR.

Performance-wise there is no regression associated with this change.


ahoppen:
@swift-ci Please test

Sources/SwiftParser/Attributes.swift
```diff
@@ -66,7 +67,7 @@ extension Parser {
     init?(lexeme: Lexer.Lexeme) {
       switch lexeme {
       case TokenSpec(._alignment): self = ._alignment
-      case TokenSpec(._backDeploy): self = .backDeployed
+      case TokenSpec(._backDeploy): self = ._backDeploy
```
bnbarham, line 70: Whoops!

bnbarham (approved):

PR #1334, opened 2023-02-09
Make `RawTokenKind` a trivial enum

ahoppen:
This changes `RawTokenKind` to be a trivial enum again, after I introduced an associated value for `keyword` a few months ago. To avoid repeatedly comparing strings or accessing `Keyword.defaultText` and thus regressing in performance, before matching a lexeme against a bunc of `TokenSpec` that refer to keywords, we pre-compute the keyword it might represent by wrapping it in `PrepareForKeywordMatch`.

Performance-wise I have seen slight improvements on on the order of ~2%-3% locally but nothing that I would eyeball as being statistically relevant - but there doesn’t seem to be a performance regression.

rdar://104659543

### DIFF
--- a/Sources/SwiftParser/Modifiers.swift
+++ b/Sources/SwiftParser/Modifiers.swift
@@ -166,7 +166,7 @@ extension Parser {
       }
 
       init?(lexeme: Lexer.Lexeme) {
-        switch lexeme {
+        switch PrepareForKeywordMatch(lexeme) {
         case TokenSpec(.private): self = .private
         case TokenSpec(.fileprivate): self = .fileprivate
         case TokenSpec(.internal): self = .internal
@@ -200,11 +200,7 @@ extension Parser {
       (unexpectedBeforeDetail, detail) = eat(setHandle)
     } else {
       unexpectedBeforeDetail = nil
-      detail = RawTokenSyntax(
-        missing: .keyword(.set),
-        text: "set",
-        arena: arena
-      )
+      detail = missingToken(.set)
     }
     let (unexpectedBeforeRightParen, rightParen) = expect(.rightParen)
 
--- a/Sources/SwiftParser/TokenSpec.swift
+++ b/Sources/SwiftParser/TokenSpec.swift
@@ -12,6 +12,32 @@
 
 @_spi(RawSyntax) import SwiftSyntax
 
+/// Pre-computes the keyword a lexeme might represent. This makes matching
+/// a lexeme that has been converted into `PrepareForKeyword` match cheaper to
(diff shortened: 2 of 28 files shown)


ahoppen:
@swift-ci Please test

Sources/SwiftParser/TokenSpec.swift
```diff
@@ -12,6 +12,32 @@
 
 @_spi(RawSyntax) import SwiftSyntax
 
+/// Pre-computes the keyword a lexeme might represent. This makes matching
+/// a lexeme that has been converted into `PrepareForKeyword` match cheaper to
+/// match against multiple `TokenSpec` that assume a keyword.
```
bnbarham, line 17: This would remove the main advantage I see in not just converting to keyword in the lexer - ie. for identifier cases, we don't need to do the string -> keyword conversion. But with `PrepareForKeywordMatch` we're now ignoring that. So why not just always convert to a keyword in the lexer once and simplify the rest?

Sources/SwiftParser/Modifiers.swift
```diff
@@ -201,7 +201,7 @@ extension Parser {
     } else {
       unexpectedBeforeDetail = nil
       detail = RawTokenSyntax(
-        missing: .keyword(.set),
+        missing: .keyword,
         text: "set",
```
bnbarham, line 205: `Keyword.set.defaultText`? (and the try/etc ones)

Or alternatively have a `RawTokenSyntax(mising:arena:)` where `missing` is a `Keyword` that does automatically.

ahoppen: Good idea. I added `missingToken(Keyword)` as a method on `Parser`.

Sources/SwiftParser/TokenSpec.swift
ahoppen: My impression is that the entire “how should we represent things that might be keywords in the lexer” discussion is related to this PR but it’s fundamentally an orthogonal issue. 

IIUC we all agree that `RawTokenKind` should be a trivial enum, so I would propose that we merge this PR and continue the discussion of how keywords should be represented in a follow-up PR. WDYT?

ahoppen:
@swift-ci Please test

Sources/SwiftParser/TokenSpec.swift
bnbarham: To be clear, I'm not arguing against merging this PR. I was mostly responding to the comment in the other PR about this addressing it 😅

bnbarham (approved):

ahoppen:
@swift-ci Please test

PR #1335, opened 2023-02-11
Fix multiline string when using string interpolation

kimdv:
Added some tests cases that fail and a single that pass

### DIFF
--- a/CodeGeneration/Sources/generate-swiftsyntax/templates/basicformat/BasicFormatFile.swift
+++ b/CodeGeneration/Sources/generate-swiftsyntax/templates/basicformat/BasicFormatFile.swift
@@ -92,11 +92,14 @@ let basicFormatFile = SourceFileSyntax(leadingTrivia: generateCopyrightHeader(fo
         if requiresTrailingSpace(node) && trailingTrivia.isEmpty {
           trailingTrivia += .space
         }
-        if let keyPath = getKeyPath(Syntax(node)), requiresLeadingNewline(keyPath), !(leadingTrivia.first?.isNewline ?? false) {
+        if let keyPath = getKeyPath(Syntax(node)), requiresLeadingNewline(keyPath), !(leadingTrivia.first?.isNewline ?? false), !shouldOmitNewline(node) {
           leadingTrivia = .newline + leadingTrivia
         }
-        leadingTrivia = leadingTrivia.indented(indentation: indentation)
-        trailingTrivia = trailingTrivia.indented(indentation: indentation)
+        var isOnNewline: Bool = (lastRewrittenToken?.trailingTrivia.pieces.last?.isNewline == true)
+        if case .stringSegment(let text) = lastRewrittenToken?.tokenKind {
+          isOnNewline = isOnNewline || (text.last?.isNewline == true)
+        }
+        leadingTrivia = leadingTrivia.indented(indentation: indentation, isOnNewline: isOnNewline)
         let rewritten = TokenSyntax(
           node.tokenKind,
           leadingTrivia: leadingTrivia,
@@ -106,6 +109,25 @@ let basicFormatFile = SourceFileSyntax(leadingTrivia: generateCopyrightHeader(fo
         lastRewrittenToken = rewritten
         putNextTokenOnNewLine = false
         return rewritten
+      }
+      """
+    )
+
+    DeclSyntax(
+      """
+      /// If this returns `true`, ``BasicFormat`` will not wrap `node` to a new line. This can be used to e.g. keep string interpolation segments on a single line.
+      /// - Parameter node: the node that is being visited
+      /// - Returns: returns true if newline should be omitted
+      open func shouldOmitNewline(_ node: TokenSyntax) -> Bool {
+        var ancestor: Syntax = Syntax(node)
+        while let parent = ancestor.parent {
+          ancestor = parent
+          if ancestor.is(ExpressionSegmentSyntax.self) {
+            return true
+          }
+        }
+
+        return false
       }
       """
     )
--- a/Sources/SwiftBasicFormat/Trivia+Indented.swift
+++ b/Sources/SwiftBasicFormat/Trivia+Indented.swift
@@ -13,18 +13,24 @@
 import SwiftSyntax
 
 extension Trivia {
-  func indented(indentation: TriviaPiece) -> Trivia {
+  /// Makes sure each newline of this trivia is followed by `indentation`. If this is not the case, the existing indentation is extended to `indentation`.
+  /// `isOnNewline` determines whether the trivia starts on a new line. If this is the case, the function makes sure that the returned trivia starts with `indentation`.
+  func indented(indentation: TriviaPiece, isOnNewline: Bool = false) -> Trivia {
     var indentedPieces: [TriviaPiece] = []
     for (index, piece) in self.enumerated() {
-      let nextPiece = index < pieces.count - 1 ? pieces[index + 1] : nil
-      indentedPieces.append(piece)
-      if piece.isNewline {
-        switch (nextPiece, indentation) {
-        case (.spaces(let nextPieceSpaces)?, .spaces(let indentationSpaces)):
+      let previousPieceIsNewline: Bool
+      if index == 0 {
+        previousPieceIsNewline = isOnNewline
+      } else {
+        previousPieceIsNewline = pieces[index - 1].isNewline
+      }
+      if previousPieceIsNewline {
+        switch (piece, indentation) {
+        case (.spaces(let nextPieceSpaces), .spaces(let indentationSpaces)):
           if nextPieceSpaces < indentationSpaces {
             indentedPieces.append(.spaces(indentationSpaces - nextPieceSpaces))
           }
-        case (.tabs(let nextPieceTabs)?, .tabs(let indentationTabs)):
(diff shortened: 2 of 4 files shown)


Sources/SwiftBasicFormat/Trivia+Indented.swift
```diff
@@ -13,25 +13,33 @@
...
-        case (.tabs(let nextPieceTabs)?, .tabs(let indentationTabs)):
+        case (.tabs(let nextPieceTabs), .tabs(let indentationTabs)):
           if nextPieceTabs < indentationTabs {
             indentedPieces.append(.tabs(indentationTabs - nextPieceTabs))
           }
         default:
           indentedPieces.append(indentation)
         }
       }
+      indentedPieces.append(piece)
```
ahoppen, line 39: Actually, I’m thinking we might be able to avoid the entire subtraction logic because anywhere we subtract, we will just insert the subtracted number of spaces/tabs later on. So we should be able to change this to something like

```swift
case (.spaces(let nextPieceSpaces), .spaces(let indentationSpaces)):
   if nextPieceSpaces < indentationSpaces {
     indentedPieces.append(indentation)
   } else {
     indentedPieces.append(piece)
   }
 case (.tabs(let nextPieceTabs), .tabs(let indentationTabs)):
   if nextPieceSpaces < indentationSpaces {
     indentedPieces.append(indentation)
   } else {
     indentedPieces.append(piece)
   }
 default:
   indentedPieces.append(indentation + piece)
 }
}
```

I haven’t tried it though.

Tests/SwiftSyntaxBuilderTest/StringLiteralTests.swift
```diff
@@ -170,4 +170,182 @@ final class StringLiteralTests: XCTestCase {
...
+      #"\(nonNilErrors.map({ "- \($0.description)" }).joined(separator: "\n"))"#
+    )
+  }
+
+  func testStringLiteralSegmentWithCode() {
+    let buildable = StringLiteralSegmentsSyntax {
+      StringSegmentSyntax(content: .stringSegment(#"Error validating child at index \(index) of \(nodeKind):"#), trailingTrivia: .newline)
+      StringSegmentSyntax(content: .stringSegment(#"Node did not satisfy any node choice requirement."#), trailingTrivia: .newline)
+      StringSegmentSyntax(content: .stringSegment(#"Validation failures:"#), trailingTrivia: .newline)
+      StringSegmentSyntax(content: .stringSegment(#"\(nonNilErrors.map({ "- \($0.description)" }).joined(separator: "\n"))"#))
```
ahoppen, line 203: If you want to form the syntax tree as it’s being parsed, this should be a `ExpressionSegmentSyntax` instead of a `StringSegmentSyntax`. Similarly, the interpolations in the first line should be `ExpressionSegmentSyntax` nodes. I’d be fine to just use tests that invoke the parser, like `testStringLiteralInExpr` if that’s simpler.

Tests/SwiftSyntaxBuilderTest/StringLiteralTests.swift
```diff
@@ -170,4 +170,182 @@ final class StringLiteralTests: XCTestCase {
...
+        """#
+      ,
+      closeQuote: .multilineStringQuoteToken(leadingTrivia: .newline)
+    )
+
+    AssertBuildResult(
+      buildable,
+      ##"""
+      #"""
+      assertionFailure("""\#n  Error validating child at index \(index) of \(nodeKind):\#n  Node did not satisfy any node choice requirement.\#n  Validation failures:\#n  \(nonNilErrors.map({ "- \($0.description)" }).joined(separator: "\n") )\#n  """, file: file, line: line)
```
ahoppen, line 262: Why do you use escaped `\#n` instead of just inserting a newline into the string literal?

CodeGeneration/Sources/generate-swiftsyntax/templates/basicformat/BasicFormatFile.swift
```diff
@@ -110,6 +115,22 @@ let basicFormatFile = SourceFileSyntax(leadingTrivia: generateCopyrightHeader(fo
       """
     )
 
+    DeclSyntax(
+      """
+      open func shouldOmitNewline(_ node: TokenSyntax) -> Bool {
```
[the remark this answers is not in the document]
kimdv: Can change. But the idea was to have it open, so other people can override it. 
If we change it, I think we should make I public or private

Tests/SwiftSyntaxBuilderTest/StringLiteralTests.swift
kimdv: It was because the outcome is that. 

```
testMultiLineStringWithFunctions(): failed - Actual output (+) differed from expected output (-):
–assertionFailure("""
–Error validating child at index \(index) of \(nodeKind):
–Node did not satisfy any node choice requirement.
–Validation failures:
–  \(nonNilErrors.map({ "- \($0.description)" }).joined(separator: "\n") )
–""", file: file, line: line)
+#"""
+assertionFailure("""\#n  Error validating child at index \(index) of \(nodeKind):\#n  Node did not satisfy any node choice requirement.\#n  Validation failures:\#n  \(nonNilErrors.map({ "- \($0.description)" }).joined(separator: "\n") )\#n  """, file: file, line: line)
+"""#
```

Not sure why?

CodeGeneration/Sources/generate-swiftsyntax/templates/basicformat/BasicFormatFile.swift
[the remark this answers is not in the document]
ahoppen: Oh, I didn’t realize it was `open`. With it being open, the naming makes sense. I think a little doc comment would help though.

Tests/SwiftSyntaxBuilderTest/StringLiteralTests.swift
ahoppen: Oh, interesting. So the actual output contains `\#n`. That shouldn’t be the case.

But that might be a side effect of you creating a single string segment that contains multiple lines and a string interpolation segment. The issue might just disappear if you construct the `StringLiteralExprSyntax` by parsing a string.

kimdv: Will remove the test as I understand and the usage of the API it looks like wrong usage. 
And I think tests should show correct examples

CodeGeneration/Sources/generate-swiftsyntax/templates/basicformat/BasicFormatFile.swift
```diff
@@ -92,11 +92,14 @@ let basicFormatFile = SourceFileSyntax(leadingTrivia: generateCopyrightHeader(fo
         if requiresTrailingSpace(node) && trailingTrivia.isEmpty {
           trailingTrivia += .space
         }
-        if let keyPath = getKeyPath(Syntax(node)), requiresLeadingNewline(keyPath), !(leadingTrivia.first?.isNewline ?? false) {
-          leadingTrivia = .newline + leadingTrivia
+        if let keyPath = getKeyPath(Syntax(node)), requiresLeadingNewline(keyPath), !(leadingTrivia.first?.isNewline ?? false), !shouldOmitNewline(node) {
+            leadingTrivia = .newline + leadingTrivia
```
ahoppen, line 96: Incorrect indentation?

CodeGeneration/Sources/generate-swiftsyntax/templates/basicformat/BasicFormatFile.swift
```diff
@@ -110,6 +113,25 @@ let basicFormatFile = SourceFileSyntax(leadingTrivia: generateCopyrightHeader(fo
...
+      """
+      /// This return true if a ``TokenSyntax`` should have leading newline ``Trivia``.
+      /// - Parameter node: the node that is being visited
+      /// - Returns: returns true if newline should be omitted
+      open func shouldOmitNewline(_ node: TokenSyntax) -> Bool {
+        var ancestor: Syntax = Syntax(node)
+        while let parent = ancestor.parent {
+          ancestor = parent
+          if ancestor.is(ExpressionSegmentSyntax.self) {
+              return true
```
ahoppen, line 126:
```suggestion
            return true
```

CodeGeneration/Sources/generate-swiftsyntax/templates/basicformat/BasicFormatFile.swift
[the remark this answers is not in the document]
ahoppen: What do you think about trying to explain what the method can be used for?

E.g.

```
If this returns true, `BasicFormat` will not wrap `node` to a new line. This can be used to e.g. keep string interpolation segments on a single line.
```

ahoppen (approved):

CodeGeneration/Sources/generate-swiftsyntax/templates/basicformat/BasicFormatFile.swift
kimdv: You have hawk eyes! 👀

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test linux

kimdv:
@swift-ci please test macOS

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test Linux

PR #1340, opened 2023-02-12
Enable a mode in the parser in which it inspects alternative token choices to mutate test cases

ahoppen:
The basic idea is that the parser records which `TokenSpec`s it checked for at individual offsets in the source. We can then use that information to generate new, interesting test cases by replacing a token by the one of the `TokenSpec` we checked for. This technique has found 7 bugs in the parser and I’m expecting it to find quite a few more once we assert that tokens have one of the expected kinds. I’m fixing the bugs this technique found in https://github.com/apple/swift-syntax/pull/1341.

Gathering of that information is hidden behind a conditional compilation flag because just performing the check of whether we want to record alternative token choices inflicts a 6% performance regression, which doesn’t provide any value except when we are running SwiftParserTest.

### DIFF
--- a/Sources/SwiftParser/TokenSpec.swift
+++ b/Sources/SwiftParser/TokenSpec.swift
@@ -44,7 +44,8 @@ struct PrepareForKeywordMatch {
 /// marked `@inline(__always)` so the compiler inlines the `RawTokenKind` we are
 /// matching against and is thus able to rule out one of the branches in
 /// `matches(rawTokenKind:text:)` based on the matched kind.
-struct TokenSpec {
+@_spi(RawSyntax)
+public struct TokenSpec {
   /// The kind we expect the token that we want to consume to have.
   /// This can be a keyword, in which case the `TokenSpec` will also match an
   /// identifier with the same text as the keyword and remap it to that keyword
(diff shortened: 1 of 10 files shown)


ahoppen:
@swift-ci Please test

Sources/SwiftParser/TokenSpec.swift
```diff
@@ -160,6 +161,28 @@ struct TokenSpec {
       atStartOfLine: lexeme.isAtStartOfLine
     )
   }
+
+  /// Returns a `TokenKind` that will most likely be parsed as a token that
+  /// matches this `TokenSpec`.
+  ///
+  /// IMPORTANT: Should only be used when generating tokens during the
+  /// modifiation of test cases. This should never be used in the parser itself.
```
bnbarham, line 169:
```suggestion
  /// modification of test cases. This should never be used in the parser itself.
```

Tests/SwiftParserTest/Assertions.swift
```diff
@@ -555,7 +556,18 @@ func AssertParse<S: SyntaxProtocol>(
   var (markerLocations, source) = extractMarkers(markedSource)
   markerLocations["START"] = 0
 
-  let tree: S = parse(source)
+  let enableTestCaseMutation = false
+  // FIXME: Currently, tests are failing when we enable test case mutation.
+  // Once all of those issues are fixed, this should become
+  // let enableTestCaseMutation = ProcessInfo.processInfo.environment["SKIP_LONG_TESTS"] != "1"
```
bnbarham, line 562: Out of interest, how much longer does this take to run?

bnbarham (approved):
Cool idea!

Tests/SwiftParserTest/Assertions.swift
ahoppen: Oh, it’s quite significant. It increases test time from 39s to 250s if you are not skipping long tests (`SKIP_LONG_TESTS=1`)

bnbarham: How many mutations happen? Given it's mutating one at a time (concurrently, but each one just has the one change), That's maybe not as bad as I thought it would be 😅. Wonder if it's worth batching some together.

ahoppen: > How many mutations happen?

I didn’t measure but I imagine there are on average 3 to 5 different choices for each token. I thought about batching them together, but you really need to be clever so one doesn’t shadow the other. And trying to be clever is usually a bad idea in my experience.

ahoppen:
https://github.com/apple/swift/pull/65174

@swift-ci Please test

PR #1342, opened 2023-02-13
Enable assertions in release builds

ahoppen:
This introduces two new assertion functions: `syntaxAssert` which is always enabled in debug and release builds and `debugAssert` which has the same behavior as `assert` from the stdlib of only being enabled in debug builds. It also re-defines `assert` to mark it as deprecated and nudge towards using `syntaxAssert`.

With a little tweaking of enabling a few assertions that provide little value only in debug build, I was able to keep the overhead of enabling assertions at ~2% when parsing MovieSwiftUI.


ahoppen:
@swift-ci Please test

kimdv (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

Sources/SwiftSyntax/Assert.swift
```diff
@@ -0,0 +1,51 @@
...
+
+/// An assertion that is active in DEBUG builds, just like `Swift.assert` and
+/// additionally if assertions are explicitly requested by setting the
+/// `SWIFTSYNTAX_ENABLE_ASSERTIONS` conditional compilation flag.
+/// Use this instead of `precondition` in places where the assertion has a
+/// non-trivial cost but provides little value in release builds.
+@_transparent
+public func assert(_ condition: @autoclosure () -> Bool, _ message: @autoclosure () -> String = String(), file: StaticString = #file, line: UInt = #line) {
+  #if SWIFTSYNTAX_ENABLE_ASSERTIONS
+  precondition(condition(), message(), file: file, line: line)
```
bnbarham, line 34: You could change this to fatalError rather than precondition so we get messages still. Or not, probably doesn't really matter.

bnbarham (approved):

utils/group.json
```diff
@@ -55,6 +55,7 @@
     "TokenDiagnostic.swift",
   ],
   "Internal": [
+    "Assert.swift",
```
bnbarham, line 58: Needs to be added to CMakeLists.txt as well

bnbarham (approved):

ahoppen:
@swift-ci Please test

PR #1343, opened 2023-02-13
Clean code gen

kimdv:
This is a PR that grew a bit. If it need to be split, I'll do so! 

I think it's easier to review commit by commit.


kimdv:
Maybe we should create a little tool for rewriting the headers 😎

Sources/SwiftSyntax/generated/TokenKind.swift
```diff
@@ -1,5 +1,19 @@
 
 
+//// Automatically Generated by generate-swiftsyntax
```
ahoppen, line 3: Not part of this PR but do you know what generates these two newlines at the start of each file?

ahoppen (commented):
Thanks. Could you extract the commit that removes force try to a separate PR because I’ve got opinions on that one. All the other commits are looking good to me and we can merge them in one PR.

Regarding force try: Instead of throwing, I think we should `fatalError` as close to the point where the syntax error was introduced as possible instead of throwing an error from the `main` function because that makes fixing the bugs easier. My vision for this has been that:
1. For each function that throws (i.e. the ones in SyntaxNodeWithBody.swift) we add a variant in `CodeGeneration` that doesn’t throw but has a label on the `PartialSyntaxNodeString` parameter that’s named `validating` or something of the sort. When the original function throws, this function `fatalError`s
2. In CodeGeneration we add an initializer `init(validating node: Self)` to `SyntaxProtocol`. This initializer `fatalError`s whenever `node` has a syntax error. And we’d use this initializer whenever we are creating a syntax node from string interpolation in CodeGeneration

That way we would restore the behavior prior to https://github.com/apple/swift-syntax/pull/1290 in CodeGeneration because that kind of behavior was actually really nice for code generation IMO.

Sources/SwiftSyntax/generated/TokenKind.swift
kimdv: I have no clue! Other files it's just one new line

kimdv: I can investigate

ahoppen: I think none of them should contain a leading newline before the comment. If you could investigate, that’d be great but as I said, it’s not a top priority.

ahoppen (approved):

PR #1346, opened 2023-02-14
Don’t override existing unexpected tokens when parsing remainder in `SyntaxParsable`

ahoppen:
Before, this was causing a round-trip failure because we we overriding the unexpected `{}` in the operator declaration by the tokens that were consumed by `consumeRemainder`.

### DIFF
--- a/Sources/SwiftParser/generated/Parser+Entry.swift
+++ b/Sources/SwiftParser/generated/Parser+Entry.swift
@@ -151,7 +151,15 @@ fileprivate extension Parser {
       return into
     }
 
-    let unexpected = RawUnexpectedNodesSyntax(elements: remainingTokens, arena: self.arena)
+    let existingUnexpected: [RawSyntax]
+    if let unexpectedNode = layout.children[layout.children.count - 1] {
+      assert(unexpectedNode.is(RawUnexpectedNodesSyntax.self))
+      existingUnexpected = unexpectedNode.as(RawUnexpectedNodesSyntax.self).elements
+    } else {
+      existingUnexpected = []
+    }
+    let unexpected = RawUnexpectedNodesSyntax(elements: existingUnexpected + remainingTokens, arena: self.arena)
+
     let withUnexpected = layout.replacingChild(at: layout.children.count - 1, with: unexpected.raw, arena: self.arena)
     return R.init(withUnexpected)!
   }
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

Sources/SwiftParser/generated/Parser+Entry.swift
```diff
@@ -146,11 +146,15 @@ fileprivate extension Parser {
...
 
-    let remainingTokens = self.consumeRemainingTokens()
+    var remainingTokens = self.consumeRemainingTokens()
     if remainingTokens.isEmpty {
       return into
     }
 
+    if let existingUnexpected = layout.children[layout.children.count - 1] {
+      assert(existingUnexpected.is(RawUnexpectedNodesSyntax.self))
+      remainingTokens.insert(contentsOf: existingUnexpected.as(RawUnexpectedNodesSyntax.self)!.elements, at: 0)
```
bnbarham, line 156: Could just use `.cast` here. FWIW I'd prefer something like:
```
var remainingTokens = self.consumeRemainingTokens()
if remainingTokens.isEmpty {
  return into
}

let existingUnexpected: [...]
if let unexpectedNode = layout.children[layout.children.count - 1] {
  assert(...) // we could probably move this assert into the cast so that we get a nice message rather than the force unwrap one
  existingUnexpected = unexpectedNode.cast(RawUnexpectedNodesSyntax.self).elements
} else {
  existingUnexpected = []
}
let unexpected = RawUnexpectedNodesSyntax(elements: existingUnexpected + remainingTokens, arena: self.arena)
```

But either is fine.

bnbarham (approved):

Sources/SwiftParser/generated/Parser+Entry.swift
ahoppen: > Could just use .cast here

Nope. `cast` does not exist on `RawSyntax`. I’m not sure if this use case is worth adding `case`. `as(…)!` works equally fine.

ahoppen:
@swift-ci Please test

Sources/SwiftParser/generated/Parser+Entry.swift
bnbarham: The split still gets me :P!

PR #1347, opened 2023-02-15
Remove attribute kind gyb

kimdv:


kimdv:
@swift-ci please test

ahoppen:
Just for my understanding: Did you essentially just remove the `.gyb` files, move the `AttributeKind.swift` from `gyb_generated` to the top level and run swift-format on it?

ahoppen (approved):
Nice. One less `.gyb` file 🥳

ahoppen:
Sounds good to me. I’m pretty sure we aren’t using these parameters anywhere so we should be able to just remove them, so I don’t really care either way.

kimdv:
I've made them private. 
IMO I thinks it's better to have them, as it reduce the chances of typos that's hard to see when debugging

kimdv:
> Sounds good to me. I’m pretty sure we aren’t using these parameters anywhere so we should be able to just remove them, so I don’t really care either way.

Or are we not using the `options` anywhere?

ahoppen:
Nope, I don’t think so. These files are more or less 1:1 copies from the corresponding files in apple/swift and they are used in apple/swift.

kimdv:
Will merge now. 

I can look into them after this PR

PR #1348, opened 2023-02-17
Improve performance of `format.py` by 2.5x

ahoppen:
We were spending a significant part of the time in `format.py` by running the `glob` queries for the files to exclude. Switch those to work based on the paths in the `files_to_format` list, improving performance of the script from ~5s to ~2s.


kimdv (approved):
Nice!

kimdv:
@swift-ci please test

PR #1350, opened 2023-02-18
Add "grouped diagnostics" that allows rendering across source files 

DougGregor:
Extend the diagnostics renderer to handle a set of "grouped"
diagnostics, where the diagnostics occur across different source files
that are related through source code generation, for example because
one is a macro expansion buffer that was triggered within the other.
These inner source files are shown as nested source files in the
output, underneath the location of where the outer source file where
they logically occur.

Introduce a few other small improvements to the colorization in the 
diagnostics formatter that became more apparent when we do this:
* Underline highlighted ranges in the source text
* Make the boxes and line numbers cyan so they feel separate from the code

Here's an example from the tests:

    === main.swift ===
    1 │ let pi = 3.14159
    2 │ #myAssert(pi == 3)
      │ ╰─ in expansion of macro 'myAssert' here
      ╭-── #myAssert ──────────────────────────────────────────────────────
      │1 │ let __a = pi
      │2 │ let __b = 3
      │3 │ if #invertedEqualityCheck(__a, __b) {
      │  │    ╰─ in expansion of macro 'invertedEqualityCheck' here
      │  ╭-── #invertedEqualityCheck ─────────────────────────────────────
      │  │1 │ !(__a == __b)
      │  │  │       ╰─ no matching operator '==' for types 'Double' and 'Int'
      │  ╰-───────────────────────────────────────────────────────────────
      │4 │   fatalError("assertion failed: pi != 3")
      │5 │ }
      ╰-───────────────────────────────────────────────────────────────────
    3 │ print("hello"
      │              ╰─ expected ')' to end function call

### DIFF
--- a/Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
+++ b/Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
@@ -14,10 +14,19 @@ import SwiftSyntax
 
 public struct DiagnosticsFormatter {
 
-  /// A wrapper struct for a source line and its diagnostics
+  /// A wrapper struct for a source line, its diagnostics, and any
+  /// non-diagnostic text that follows the line.
   private struct AnnotatedSourceLine {
     var diagnostics: [Diagnostic]
     var sourceString: String
+
+    /// Non-diagnostic text.
+    var suffixText: String
+
+    /// Whether this line is free of annotations.
+    var isEmpty: Bool {
+      return diagnostics.isEmpty && suffixText.isEmpty
+    }
   }
 
   /// Number of lines which should be printed before and after the diagnostic message
@@ -42,7 +51,12 @@ public struct DiagnosticsFormatter {
   }
 
   /// Print given diagnostics for a given syntax tree on the command line
-  public func annotatedSource<SyntaxType: SyntaxProtocol>(tree: SyntaxType, diags: [Diagnostic]) -> String {
+  func annotatedSource<SyntaxType: SyntaxProtocol>(
+    fileName: String?,
+    tree: SyntaxType, diags: [Diagnostic],
+    indentString: String,
+    suffixText: [(AbsolutePosition, String)]
+  ) -> String {
     let slc = SourceLocationConverter(file: "", tree: tree)
 
     // First, we need to put each line and its diagnostics together
@@ -52,19 +66,33 @@ public struct DiagnosticsFormatter {
       let diagsForLine = diags.filter { diag in
         return diag.location(converter: slc).line == (sourceLineIndex + 1)
       }
-      annotatedSourceLines.append(AnnotatedSourceLine(diagnostics: diagsForLine, sourceString: sourceLine))
+      let suffixText = suffixText.compactMap { (position, text) in
+        if slc.location(for: position).line == (sourceLineIndex + 1) {
+          return text
+        }
+
+        return nil
+      }.joined()
+
+      annotatedSourceLines.append(AnnotatedSourceLine(diagnostics: diagsForLine, sourceString: sourceLine, suffixText: suffixText))
     }
 
     // Only lines with diagnostic messages should be printed, but including some context
     let rangesToPrint = annotatedSourceLines.enumerated().compactMap { (lineIndex, sourceLine) -> Range<Int>? in
       let lineNumber = lineIndex + 1
-      if !sourceLine.diagnostics.isEmpty {
+      if !sourceLine.isEmpty {
         return Range<Int>(uncheckedBounds: (lower: lineNumber - contextSize, upper: lineNumber + contextSize + 1))
       }
       return nil
     }
 
     var annotatedSource = ""
+
+    // If there was a filename, add it first.
+    if let fileName = fileName {
+      annotatedSource.append(indentString)
+      annotatedSource.append("=== \(fileName) ===\n")
+    }
 
     /// Keep track if a line missing char should be printed
     var hasLineBeenSkipped = false
@@ -89,11 +117,14 @@ public struct DiagnosticsFormatter {
 
       // If necessary, print a line that indicates that there was lines skipped in the source code
       if hasLineBeenSkipped && !annotatedSource.isEmpty {
-        let lineMissingInfoLine = String(repeating: " ", count: maxNumberOfDigits) + " ┆"
+        let lineMissingInfoLine = indentString + String(repeating: " ", count: maxNumberOfDigits) + " ┆"
         annotatedSource.append("\(lineMissingInfoLine)\n")
       }
       hasLineBeenSkipped = false
 
+      // add indentation
+      annotatedSource.append(indentString)
+      
       // print the source line
       annotatedSource.append("\(linePrefix)\(annotatedLine.sourceString)")
 
@@ -111,7 +142,7 @@ public struct DiagnosticsFormatter {
 
       for (column, diags) in diagsPerColumn {
         // compute the string that is shown before each message
-        var preMessage = String(repeating: " ", count: maxNumberOfDigits) + " ∣"
+        var preMessage = indentString + String(repeating: " ", count: maxNumberOfDigits) + " ∣"
         for c in 0..<column {
           if columnsWithDiagnostics.contains(c) {
             preMessage.append("│")
@@ -125,8 +156,27 @@ public struct DiagnosticsFormatter {
         }
         annotatedSource.append("\(preMessage)╰─ \(colorizeIfRequested(diags.last!.diagMessage))\n")
       }
+
+      // Add suffix text.
+      annotatedSource.append(annotatedLine.suffixText)
+      if annotatedSource.last != "\n" {
+        annotatedSource.append("\n")
+      }
     }
     return annotatedSource
+  }
+
+  /// Print given diagnostics for a given syntax tree on the command line
+  public func annotatedSource<SyntaxType: SyntaxProtocol>(
+    tree: SyntaxType, diags: [Diagnostic]
+  ) -> String {
+    return annotatedSource(
+      fileName: nil,
+      tree: tree,
+      diags: diags,
+      indentString: "",
+      suffixText: []
+    )
   }
 
   /// Annotates the given ``DiagnosticMessage`` with an appropriate ANSI color code (if the value of the `colorize`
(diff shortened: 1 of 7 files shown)


DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

Sources/SwiftDiagnostics/GroupedDiagnostics.swift
```diff
@@ -0,0 +1,205 @@
...
+      let extraLengthNeeded = targetLineLength - padding.count - sourceFile.displayName.count - 5
+      let boxSuffix: String
+      if extraLengthNeeded > 0 {
+        boxSuffix = String(repeating: "─", count: extraLengthNeeded)
+      } else {
+        boxSuffix = ""
+      }
+
+      prefixString = padding + "╭-── " + sourceFile.displayName + boxSuffix + "\n"
+      suffixString = padding + "╰-──" + String(repeating: "─", count: sourceFile.displayName.count + 1) + boxSuffix + "\n"
```
ahoppen, line 182: I’d use string interpolation here.

Tests/SwiftParserTest/translated/StringLiteralEofTests.swift
```diff
@@ -105,7 +105,7 @@ final class StringLiteralEofTests: XCTestCase {
             foo
             \(<#expression#>)"""
         """##
-      // FIXME: The closing delimiter should be put on the new line
+        // FIXME: The closing delimiter should be put on the new line
```
ahoppen, line 108: I think your local build of swift-format is too old. The formatting of this line changed with https://github.com/apple/swift-format/pull/480.

Tests/SwiftDiagnosticsTest/GroupDiagnosticsFormatterTests.swift
```diff
@@ -0,0 +1,205 @@
...
+        │  │1 │ !(__a == __b)
+        │  │  ∣       ╰─ no matching operator '==' for types 'Double' and 'Int'
+        │  ╰-─────────────────────────────────────────────────────────────────
+        │4 │   fatalError("assertion failed: pi != 3")
+        │5 │ }
+        ╰-────────────────────────────────────────────────────────────────────
+      3 │ print("hello"
+        ∣              ╰─ expected ')' to end function call
+
+      """
```
ahoppen, line 202: The debug output looks a little bit too squashed to me. What do you think about adding an empty l line before and after starting a new source file buffer and adda leading space after the `|` of the inline source buffer. I.e.

```
      === main.swift ===
       1 │ let pi = 3.14159
       2 │ #myAssert(pi == 3)
         ∣ ╰─ in expansion of macro 'myAssert' here

         ╭-── #myAssert────────────────────────────────────────────────────────
         │ 1 │ let __a = pi
         │ 2 │ let __b = 3
         │ 3 │ if #invertedEqualityCheck(__a, __b) {
         │   ∣    ╰─ in expansion of macro 'invertedEqualityCheck' here
         │ 
         │   ╭-── #invertedEqualityCheck────────────────────────────────────────
         │   │ 1 │ !(__a == __b)
         │   │   ∣       ╰─ no matching operator '==' for types 'Double' and 'Int'
         │   ╰-─────────────────────────────────────────────────────────────────
         │ 
         │ 4 │   fatalError("assertion failed: pi != 3")
         │ 5 │ }
         ╰-────────────────────────────────────────────────────────────────────

       3 │ print("hello"
         ∣              ╰─ expected ')' to end function call

```

Tests/SwiftDiagnosticsTest/GroupDiagnosticsFormatterTests.swift
```diff
@@ -0,0 +1,205 @@
...
+
+final class GroupedDiagnosticsFormatterTests: XCTestCase {
+  func testGroupingForMacroExpansion() {
+    var group = GroupedDiagnostics()
+
+    // Main source file.
+    let (mainSourceID, mainSourceMarkers) = group.addTestFile(
+      """
+      let pi = 3.14159
+      0️⃣#myAssert(pi == 3)
```
ahoppen, line 84: Not really important but in all the other tests we start counting diagnostics with 1️⃣ not 0️⃣.

ahoppen (commented):
This looks nice! I’ve got a few minor comments inline.

Sources/SwiftDiagnostics/GroupedDiagnostics.swift
```diff
@@ -0,0 +1,197 @@
...
+    /// The set of diagnostics that are emitted within this source file.
+    var diagnostics: [Diagnostic] = []
+  }
+
+  /// The set of source files for which diagnostics will be emitted.
+  var sourceFiles: [SourceFile] = []
+
+  /// Mapping from the root source file syntax nodes to the corresponding
+  /// source file IDs.
+  var rootIndexes: [SourceFileSyntax: SourceFileID] = [:]
```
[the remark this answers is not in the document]
DougGregor: Maybe? It felt more fragile than if we made this explicit.

Tests/SwiftParserTest/translated/StringLiteralEofTests.swift
DougGregor: Ah, that makes sense.

Tests/SwiftDiagnosticsTest/GroupDiagnosticsFormatterTests.swift
DougGregor: This is an easy change we can do at any point, but... this looks way too spaced out for me.  In the colorized version, I've made the boxes and line numbers cyan, which helps distinguish things more.

DougGregor:
@swift-ci please test

Tests/SwiftDiagnosticsTest/GroupDiagnosticsFormatterTests.swift
ahoppen: Ah, I didn’t look at the colorized output yet. If you’ve got a preference, let’s keep it as-is for now.

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
```diff
@@ -41,9 +90,108 @@ public struct DiagnosticsFormatter {
...
+
+      return startColumn..<endColumn
+    }.sorted { (lhs, rhs) in
+      lhs.lowerBound < rhs.lowerBound
+    }.mergingOverlappingRanges()
+
+    // Map the column ranges into index ranges within the source string itself.
+    let sourceString = annotatedLine.sourceString
+    let highlightIndexRanges: [Range<String.Index>] = highlightRanges.map { highlightRange in
+      let startIndex = sourceString.index(sourceString.startIndex, offsetBy: highlightRange.lowerBound - 1)
```
ahoppen, line 156: I’m not 100% sure but I don’t think this is Unicode correct. The columns are about UTF-8 byte offsets whereas `String.Index` is about Unicode characters/grapheme clusters/something along those lines. IIRC there’s a way to convert between the two but I don’t remember what it is. Adding a few tests with emojis would certainly be good.

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
```diff
@@ -195,4 +406,14 @@ struct ANSIAnnotation {
...
   }
+
+  /// Annotation used for the outline and line numbers of a buffer.
+  static var bufferOutline: ANSIAnnotation {
+    ANSIAnnotation(color: .cyan, trait: .normal)
+  }
+
+  /// Annotation used for highlighting source text.
+  static var sourceHighlight: ANSIAnnotation {
+    ANSIAnnotation(color: .white, trait: .underline)
```
ahoppen, line 417: Will this also print the text as white in a terminal with white background and thus make it invisible? Shouldn’t it be

```suggestion
    ANSIAnnotation(color: .normal, trait: .underline)
```

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
DougGregor: Oh, I think you're right. I should be computing indices via the UTF-8 view of the source string.

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
DougGregor: Yeah, I wanted to do that, but if I. use `.normal` here, Terminal doesn't show the underline.

ahoppen: Hmm, but having invisible test in a white terminal seems like a really bad problem to me (I’m one of those weirdos who doesn’t use a dark terminal 😆).

Sources/SwiftDiagnostics/GroupedDiagnostics.swift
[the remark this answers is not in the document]
DougGregor: The mapping is only used in cases where the client isn't supplying a buffer ID, so you never need to depend on it. We could either drop the mapping internally (and have clients always supply the buffer ID when adding a diagnostic) or drop the buffer ID (and rely on the mapping everywhere), but this does feel like a weird middle ground.

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
DougGregor: Yes, I got this fixed up in https://github.com/apple/swift-syntax/pull/1365

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
DougGregor: Addressed this in https://github.com/apple/swift-syntax/pull/1365

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
```diff
@@ -41,9 +90,108 @@ public struct DiagnosticsFormatter {
...
+  ///
+  /// - Parameters:
+  ///   - suffixTexts: suffix text to be printed at the given absolute
+  ///                  locations within the source file.
+  func annotatedSource<SyntaxType: SyntaxProtocol>(
+    fileName: String?,
+    tree: SyntaxType,
+    diags: [Diagnostic],
+    indentString: String,
+    suffixTexts: [(AbsolutePosition, String)],
```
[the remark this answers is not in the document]
DougGregor: This is a little nicer, so I went ahead and did it in https://github.com/apple/swift-syntax/pull/1365

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
ahoppen: Thanks!

PR #1352, opened 2023-02-18
Add missing code for parser entry template and fix `build-script.py`

kimdv:
In #1346 code was modified on the file and not in the template

### DIFF
--- a/CodeGeneration/Sources/generate-swiftsyntax/templates/swiftparser/ParserEntryFile.swift
+++ b/CodeGeneration/Sources/generate-swiftsyntax/templates/swiftparser/ParserEntryFile.swift
@@ -87,7 +87,15 @@ let parserEntryFile = SourceFileSyntax(leadingTrivia: generateCopyrightHeader(fo
           return into
         }
 
-        let unexpected = RawUnexpectedNodesSyntax(elements: remainingTokens, arena: self.arena)
+        let existingUnexpected: [RawSyntax]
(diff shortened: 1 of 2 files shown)


kimdv:
@swift-ci please test

kimdv:
The error was properly introduced in #1149 where we refactored the code-gen into a single executable. 

Not sure why the CI didn't catch that before

kimdv:
@swift-ci plese test

kimdv:
@swift-ci please test

keith:
@kimdv looks like things don't build without this, is it good to merge?

kimdv:
> @kimdv looks like things don't build without this, is it good to merge?

For me yea. But we need approval from @ahoppen I think 🤔

keith:

keith:
Let's merge and do post-commit review if needed to unblock folks

gottesmm:
@Keith this broke the build: https://ci.swift.org/job/oss-swift-incremental-RA-macos/2331/. I am going to try reverting it.

keith:
dang, sorry, thanks! looks like we need to run the integration tests before merging again

ahoppen:
> Not sure why the CI didn't catch that before

We don’t verify that the CodeGenerated files match the templates because running code generation requires us to check out an older version of SwiftSyntax (the one specified in the Package.swift of CodeGeneration). But CI doesn’t have internet access at that point to perform that checkout… So it’s a little tricky to do.

PR #1355, opened 2023-02-22
Fix leading newline in source files

kimdv:


CodeGeneration/Sources/Utils/CodeGenerationFormat.swift
```diff
@@ -36,7 +36,11 @@ public class CodeGenerationFormat: BasicFormat {
   }
 
   public override func visit(_ node: CodeBlockItemSyntax) -> CodeBlockItemSyntax {
-    if node.parent?.parent?.is(SourceFileSyntax.self) == true, !node.item.is(ImportDeclSyntax.self) {
+    // First item in the `CodeBlockItemListSyntax` don't need a newline or identation if the parent is a `SourceFileSyntax`.
+    // We want to group imports so newline between them should be omitted
+    let shouldOmitNewline = node.parent?.as(CodeBlockItemListSyntax.self)?.first == node || node.item.is(ImportDeclSyntax.self)
```
kimdv, line 41: We could also add this for `EnumCaseDeclSyntax`

So 
```swift
enum Foo {
case bar

case baz
}
```

becomes

```swift
enum Foo {
case bar
case baz
}
```

ahoppen: Sounds reasonable to me. To make sure the expression doesn’t get too large, it might make sense to refactor it to a separate function, and also move the `node.parent?.parent?.is(SourceFileSyntax.self)` check to that function.

Also, I think `shouldOmitNewline` is not really the right name here because there will still be a newline before the `import`, just not two. So the function should probably be named `shouldBeSeparatedByTwoNewlines` or something like that.

kimdv:
I would like to include this: https://github.com/apple/swift-syntax/pull/1335 before merging

CodeGeneration/Sources/Utils/CodeGenerationFormat.swift
```diff
@@ -64,9 +56,17 @@ public class CodeGenerationFormat: BasicFormat {
     }
   }
 
+  public override func visit(_ node: MemberDeclBlockSyntax) -> MemberDeclBlockSyntax {
+    if node.members.count == 0 {
+      return node.with(\.leftBrace, .leftBraceToken())
+    } else {
+      return super.visit(node)
+    }
+  }
```
kimdv, line 65: Just moved this so it's sorted alphabetically by the node type

ahoppen (approved):
Thank you. Looks much nicer now.

kimdv:
@swift-ci plese test

kimdv:
@swift-ci please test

PR #1356, opened 2023-02-22
[Macros] Introduce a `ConformanceMacro` protocol.

hborla:


hborla:
@swift-ci please test

PR #1358, opened 2023-02-23
Produce a meaningful diagnostic for "#elif" typo

TiagoMaiaL:
## Purpose

To fix the second part of issue https://github.com/apple/swift-syntax/issues/1221: emit a proper diagnostic and fixit for the common `#elif` typo.

## Implementation details

I've considered two approaches:
1. Add a new `elif` case to the `TokenKind` enums, and later on use it to generate the proper diagnostic
2. Use the existing `pound` `TokenKind`  to determine if the user has a typo in the code and generate the diagnostic

I chose the second option because it didn't require changes to the `TokenKind` enums. I'm now adding unexpected tokens for the `.pound` + `.identifier` cases. If the identifier text is `elif`, I synthesize a `poundElseIf` token. I don't know how we could also check for `.pound` + `.keywords`, or if this would be needed.

For the diagnostic generations, I'm adding the proper fixit for the `elif` case, and if the identifier is any other word, I just add a generic diagnostic for it.

Any tip or help to get this right is really appreciated. Thank you in advance.

### DIFF
--- a/Sources/SwiftParser/Directives.swift
+++ b/Sources/SwiftParser/Directives.swift
@@ -16,11 +16,13 @@ extension Parser {
   private enum IfConfigContinuationClauseStartKeyword: TokenSpecSet {
     case poundElseifKeyword
     case poundElseKeyword
+    case pound
 
     var spec: TokenSpec {
       switch self {
       case .poundElseifKeyword: return .poundElseifKeyword
       case .poundElseKeyword: return .poundElseKeyword
+      case .pound: return TokenSpec(.pound, recoveryPrecedence: .openingPoundIf)
       }
     }
 
@@ -28,6 +30,7 @@ extension Parser {
       switch PrepareForKeywordMatch(lexeme) {
       case TokenSpec(.poundElseifKeyword): self = .poundElseifKeyword
       case TokenSpec(.poundElseKeyword): self = .poundElseKeyword
+      case TokenSpec(.pound): self = .pound
       default: return nil
       }
     }
@@ -100,56 +103,64 @@ extension Parser {
     }
 
     var clauses = [RawIfConfigClauseSyntax]()
-    do {
-      var firstIteration = true
-      var loopProgress = LoopProgressCondition()
-      while let poundIfHandle = firstIteration ? self.canRecoverTo(.poundIfKeyword) : self.canRecoverTo(anyIn: IfConfigContinuationClauseStartKeyword.self)?.handle,
-        loopProgress.evaluate(self.currentToken)
-      {
-        var (unexpectedBeforePoundIf, poundIf) = self.eat(poundIfHandle)
-        firstIteration = false
-        // Parse the condition.
-        let condition: RawExprSyntax?
-        switch poundIf.tokenKind {
-        case .poundIfKeyword, .poundElseifKeyword:
+
+    // Parse #if
+    let (unexpectedBeforePoundIfKeyword, poundIfKeyword) = self.expect(.poundIfKeyword)
+    let condition = RawExprSyntax(self.parseSequenceExpression(.basic, forDirective: true))
+
+    clauses.append(
+      RawIfConfigClauseSyntax(
+        unexpectedBeforePoundIfKeyword,
+        poundKeyword: poundIfKeyword,
+        condition: condition,
+        elements: syntax(&self, parseIfConfigClauseElements(parseElement, addSemicolonIfNeeded: addSemicolonIfNeeded)),
+        arena: self.arena
+      )
+    )
+
+    // Proceed to parse #if continuation clauses (#elseif, #else, check #elif typo, #endif)
+    var loopProgress = LoopProgressCondition()
+    LOOP: while let (match, handle) = self.canRecoverTo(anyIn: IfConfigContinuationClauseStartKeyword.self), loopProgress.evaluate(self.currentToken) {
+      var unexpectedBeforePoundKeyword: RawUnexpectedNodesSyntax?
+      var poundKeyword: RawTokenSyntax
+      let condition: RawExprSyntax?
+
+      switch match {
+      case .poundElseifKeyword:
+        (unexpectedBeforePoundKeyword, poundKeyword) = self.eat(handle)
+        condition = RawExprSyntax(self.parseSequenceExpression(.basic, forDirective: true))
+      case .poundElseKeyword:
+        (unexpectedBeforePoundKeyword, poundKeyword) = self.eat(handle)
+        if let ifToken = self.consume(if: .init(.if, allowAtStartOfLine: false)) {
+          unexpectedBeforePoundKeyword = RawUnexpectedNodesSyntax(combining: unexpectedBeforePoundKeyword, poundKeyword, ifToken, arena: self.arena)
+          poundKeyword = self.missingToken(.poundElseifKeyword)
           condition = RawExprSyntax(self.parseSequenceExpression(.basic, forDirective: true))
-        case .poundElseKeyword:
-          if let ifToken = self.consume(if: .init(.if, allowAtStartOfLine: false)) {
-            unexpectedBeforePoundIf = RawUnexpectedNodesSyntax(combining: unexpectedBeforePoundIf, poundIf, ifToken, arena: self.arena)
-            poundIf = self.missingToken(.poundElseifKeyword)
-            condition = RawExprSyntax(self.parseSequenceExpression(.basic, forDirective: true))
-          } else {
-            condition = nil
-          }
-        default:
-          preconditionFailure("The loop condition should guarantee that we are at one of these tokens")
+        } else {
+          condition = nil
         }
-
-        var elements = [Element]()
-        do {
-          var elementsProgress = LoopProgressCondition()
-          while !self.at(.eof) && !self.at(.poundElseKeyword, .poundElseifKeyword, .poundEndifKeyword) && elementsProgress.evaluate(currentToken) {
-            let newItemAtStartOfLine = self.currentToken.isAtStartOfLine
-            guard let element = parseElement(&self, elements.isEmpty), !element.isEmpty else {
-              break
-            }
-            if let lastElement = elements.last, let fixedUpLastItem = addSemicolonIfNeeded(lastElement, newItemAtStartOfLine, &self) {
-              elements[elements.count - 1] = fixedUpLastItem
-            }
-            elements.append(element)
+      case .pound:
+        if self.atElifTypo() {
+          (unexpectedBeforePoundKeyword, poundKeyword) = self.eat(handle)
+          guard let elif = self.consume(if: TokenSpec(.identifier, allowAtStartOfLine: false)) else {
+            preconditionFailure("The current token should be an identifier, guaranteed by the `atElifTypo` check.")
           }
+          unexpectedBeforePoundKeyword = RawUnexpectedNodesSyntax(combining: unexpectedBeforePoundKeyword, poundKeyword, elif, arena: self.arena)
+          poundKeyword = self.missingToken(.poundElseifKeyword)
+          condition = RawExprSyntax(self.parseSequenceExpression(.basic, forDirective: true))
+        } else {
+          break LOOP
         }
-
-        clauses.append(
-          RawIfConfigClauseSyntax(
-            unexpectedBeforePoundIf,
-            poundKeyword: poundIf,
-            condition: condition,
-            elements: syntax(&self, elements),
-            arena: self.arena
-          )
-        )
       }
+
+      clauses.append(
+        RawIfConfigClauseSyntax(
+          unexpectedBeforePoundKeyword,
+          poundKeyword: poundKeyword,
+          condition: condition,
+          elements: syntax(&self, parseIfConfigClauseElements(parseElement, addSemicolonIfNeeded: addSemicolonIfNeeded)),
+          arena: self.arena
+        )
+      )
     }
 
     let (unexpectedBeforePoundEndIf, poundEndIf) = self.expect(.poundEndifKeyword)
@@ -159,6 +170,40 @@ extension Parser {
       poundEndif: poundEndIf,
       arena: self.arena
     )
+  }
+
+  private mutating func atElifTypo() -> Bool {
+    guard self.at(TokenSpec(.pound)), self.currentToken.trailingTriviaText.isEmpty else {
+      return false
+    }
+    var lookahead = self.lookahead()
+    lookahead.consumeAnyToken()  // consume `#`
+    guard lookahead.at(TokenSpec(.identifier, allowAtStartOfLine: false)), lookahead.currentToken.tokenText == "elif", lookahead.currentToken.leadingTriviaText.isEmpty else {
+      return false  // `#` and `elif` must not be separated by trivia
+    }
+    lookahead.consumeAnyToken()  // consume `elif`
+    // We are only at a `elif` typo if it’s followed by an identifier for the condition.
+    // `#elif` or `#elif(…)` could be macro invocations.
+    return lookahead.at(TokenSpec(.identifier, allowAtStartOfLine: false))
+  }
+
+  private mutating func parseIfConfigClauseElements<Element: RawSyntaxNodeProtocol>(
+    _ parseElement: (_ parser: inout Parser, _ isFirstElement: Bool) -> Element?,
+    addSemicolonIfNeeded: (_ lastElement: Element, _ newItemAtStartOfLine: Bool, _ parser: inout Parser) -> Element?
+  ) -> [Element] {
+    var elements = [Element]()
+    var elementsProgress = LoopProgressCondition()
+    while !self.at(.eof) && !self.at(.poundElseKeyword, .poundElseifKeyword, .poundEndifKeyword) && !self.atElifTypo() && elementsProgress.evaluate(currentToken) {
+      let newItemAtStartOfLine = self.currentToken.isAtStartOfLine
+      guard let element = parseElement(&self, elements.isEmpty), !element.isEmpty else {
+        break
+      }
+      if let lastElement = elements.last, let fixedUpLastItem = addSemicolonIfNeeded(lastElement, newItemAtStartOfLine, &self) {
+        elements[elements.count - 1] = fixedUpLastItem
+      }
+      elements.append(element)
+    }
+    return elements
   }
 }
 
--- a/Tests/SwiftParserTest/translated/IfconfigExprTests.swift
+++ b/Tests/SwiftParserTest/translated/IfconfigExprTests.swift
@@ -465,6 +465,102 @@ final class IfconfigExprTests: XCTestCase {
     )
   }
 
+  func testIfConfigExpr32() {
+    assertParse(
+      """
+      #if arch(x86_64)
+        debugPrint("x86_64")
+      1️⃣#elif arch(arm64)
+        debugPrint("arm64")
+      #else
+        debugPrint("Some other architecture.")
+      #endif
+      """,
+      diagnostics: [
+        DiagnosticSpec(
+          message: "use of unknown directive '#elif'",
+          fixIts: ["replace '#elif' with '#elseif'"]
+        )
+      ],
+      fixedSource: """
+        #if arch(x86_64)
+          debugPrint("x86_64")
+        #elseif arch(arm64)
+          debugPrint("arm64")
+        #else
+          debugPrint("Some other architecture.")
+        #endif
+        """
+    )
+  }
+
+  func testIfConfigExpr33() {
+    assertParse(
+      """
+      #if arch(x86_64)
+      #line
+      #endif
+      """
+    )
+  }
+
+  // FIXME: Parsing should generate diagnostics - https://github.com/apple/swift-syntax/issues/1395
+  func testIfConfigExpr34() {
+    assertParse(
+      """
+      #if MY_FLAG
+      #
+      elif
+      #endif
+      """
+    )
+  }
+
+  // FIXME: Parsing should generate diagnostics - https://github.com/apple/swift-syntax/issues/1395
+  func testIfConfigExpr35() {
+    assertParse(
+      """
+      #if MY_FLAG
+      # elif
+      #endif
+      """
+    )
+  }
+
+  func testIfConfigExpr36() {
+    assertParse(
+      """
+      switch x {
+        1️⃣#()
+      #if true
+        2️⃣bar()
+      #endif
+        case .A, .B:
+          break
+      }
+      """,
+      diagnostics: [
+        DiagnosticSpec(message: "unexpected code '#()' before conditional compilation clause"),
+        DiagnosticSpec(
+          locationMarker: "2️⃣",
+          message: "all statements inside a switch must be covered by a 'case' or 'default' label",
+          fixIts: ["insert label"]
+        ),
+      ],
+      fixedSource: """
+        switch x {
+          #()
+        #if true
+        case <#identifier#>:
+          bar()
+        #endif
+          case .A, .B:
+            break
+        }
+        """
+    )
+  }
+
   func testUnknownPlatform1() {
     assertParse(
       """
--- a/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
+++ b/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
@@ -865,10 +865,18 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
           unexpectedBeforePoundKeyword
           .suffix(2)
           .compactMap { $0.as(TokenSyntax.self) }
+        var diagnosticMessage: DiagnosticMessage?
+
         if unexpectedTokens.map(\.tokenKind) == [.poundElseKeyword, .keyword(.if)] {
+          diagnosticMessage = StaticParserError.unexpectedPoundElseSpaceIf
+        } else if unexpectedTokens.first?.tokenKind == .pound, unexpectedTokens.last?.text == "elif" {
+          diagnosticMessage = UnknownDirectiveError(unexpected: unexpectedBeforePoundKeyword)
+        }
+
+        if let diagnosticMessage = diagnosticMessage {
           addDiagnostic(
             unexpectedBeforePoundKeyword,
-            StaticParserError.unexpectedPoundElseSpaceIf,
+            diagnosticMessage,
             fixIts: [
               FixIt(
                 message: ReplaceTokensFixIt(replaceTokens: unexpectedTokens, replacements: [clause.poundKeyword]),


Sources/SwiftParser/Directives.swift
```diff
@@ -121,14 +124,23 @@ extension Parser {
           } else {
             condition = nil
           }
+        case .pound:
```
ahoppen, line 127: Your implementation currently also constructs a `#elseif` for everything other than `#elif`. For example if you have

```swift
#if MY_FLAG
#line
#endif
```

I think this would treat `#line` just like `#elseif`, producing a diagnostic, even though the code is valid.

Also, we need to consider that someone might have declared a macro named `elif` that doesn’t take any parameters and thus even

```swift
#if MY_FLAG
#elif
#endif
```

might be valid Swift (it’s not yet because [SE-0382](https://github.com/apple/swift-evolution/blob/main/proposals/0382-expression-macros.md) hasn’t been accepted yet, but it will probably be valid soon).

So, what I think you need to do is, first thing in the loop, even before `self.eat(poundIfHandle)`, you need to check that 
(a) the `#` is followed by `elif` and 
(b) that `elif` is followed by another identifier which is on the same line as `elif`. 

(b) should ensure that we can disambiguate between the macro case and `#elif` which was meant to be `#elseif`

ahoppen (commented):
Thank you. I think you are going in the right direction but macros are making this harder because `#elif` could be a macro invocation. See my comment inline.

Tests/SwiftParserTest/translated/IfconfigExprTests.swift
```diff
@@ -457,6 +457,45 @@ final class IfconfigExprTests: XCTestCase {
...
+  
+  func testIfConfigExpr33() {
+    AssertParse(
+      """
+      #if arch(x86_64)
+      #line
+      #endif
+      """
+    )
+  }
```
ahoppen, line 497: I just ran this test locally and don’t think it passes.

TiagoMaiaL: Yes, it doesn't. I added it to make sure I'm not breaking anything. If you think it doesn't make sense, I can remove it.

Sources/SwiftParser/Directives.swift
TiagoMaiaL: Thank you for the tip. I'm trying to accomplish what you suggested. I'll submit the code once it's ready.

TiagoMaiaL: @ahoppen I'm doing the check in the following code:
```swift
         var atElifTypo: Bool {
           guard self.at(.pound) else {
             return false
           }

           lookahead.consumeAnyToken()

           if lookahead.at(.identifier), lookahead.currentToken.tokenText == "elif" {
             lookahead.consumeAnyToken()
           } else {
             return false
           }

           return lookahead.at(.identifier)
         }
```

The tests are passing, although I'm not sure if this is enough.

ahoppen (commented):
Sorry for the delay. I started a review but forgot to submit the comments 😬

Sources/SwiftParser/Directives.swift
```diff
@@ -106,29 +109,58 @@ extension Parser {
       while let poundIfHandle = firstIteration ? self.canRecoverTo(.poundIfKeyword) : self.canRecoverTo(anyIn: IfConfigContinuationClauseStartKeyword.self)?.handle,
         loopProgress.evaluate(self.currentToken)
       {
-        var (unexpectedBeforePoundIf, poundIf) = self.eat(poundIfHandle)
-        firstIteration = false
-        // Parse the condition.
+        var unexpectedBeforePoundIf: RawUnexpectedNodesSyntax?
+        var poundIf: Parser.Token
```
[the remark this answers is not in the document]
TiagoMaiaL: The types do not match when I use `Lexer.Lexeme`. Am I missing something?
Screenshot 2023-03-08 at 20 24 00

Sources/SwiftParser/Directives.swift
```diff
@@ -106,29 +109,58 @@ extension Parser {
...
+          } else {
+            return false
+          }
+
+          return lookahead.at(.identifier)
+        }
+
+        if atElifTypo {
+          (unexpectedBeforePoundIf, poundIf) = self.eat(poundIfHandle)
+          guard let identifier = self.consume(if: TokenSpec(.identifier, allowAtStartOfLine: false)) else {
```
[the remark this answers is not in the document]
TiagoMaiaL: Hey @ahoppen, I've added the following tests and they are passing, but I think they shouldn't:
```swift
  func testIfConfigExpr34() {
    AssertParse(
      """
      #if MY_FLAG
      #
      elif
      #endif
      """
    )
  }
  
  func testIfConfigExpr35() {
    AssertParse(
      """
      #if MY_FLAG
      # elif
      #endif
      """
    )
  }
```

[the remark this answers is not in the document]
ahoppen: Could you share what kind of syntax trees these produce? Either with `swift-parser-cli print-tree /path/to/file/containing/snippet.swift` or by setting a breakpoint in `AssertParse` and running `e print(tree.recursiveDescription)`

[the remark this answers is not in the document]
TiagoMaiaL: Sure. For the code in `testIfConfigExpr34`, it prints the following tree:
```swift
SourceFileSyntax children=2
      0: IfConfigDeclSyntax children=2
        0: IfConfigClauseListSyntax children=1
          0: IfConfigClauseSyntax children=3
            0: poundIfKeyword
              0: identifier("MY_FLAG")
            2: CodeBlockItemListSyntax children=1
                0: MacroExpansionExprSyntax children=3
                  0: pound
                  1: identifier("elif")
                  2: TupleExprElementListSyntax
        1: poundEndifKeyword
  1: eof
```

And for the code in `testIfConfigExpr35` it prints this:
```swift
SourceFileSyntax children=2
      0: IfConfigDeclSyntax children=2
        0: IfConfigClauseListSyntax children=1
          0: IfConfigClauseSyntax children=3
            0: poundIfKeyword
              0: identifier("MY_FLAG")
            2: CodeBlockItemListSyntax children=1
                0: MacroExpansionExprSyntax children=3
                  0: pound
                  1: identifier("elif")
                  2: TupleExprElementListSyntax
        1: poundEndifKeyword
  1: eof
```

They look like the same output.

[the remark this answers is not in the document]
ahoppen: OK, nice. This looks correct as far as your PR is concerned since we are parsing `#elif` as a macro expansion. I filed https://github.com/apple/swift-syntax/issues/1395 to keep track of the issue. It’s similar to what you are doing here so  if you want to make another contribution, I think it would be a good candidate for you.

Sources/SwiftParser/Directives.swift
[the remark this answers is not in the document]
ahoppen: Oh, sorry, I mixed things up. I meant `RawTokenSyntax` instead of `Lexeme`.

Sources/SwiftParser/Directives.swift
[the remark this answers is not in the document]
ahoppen: Could you add a FIXME to these test cases mentioning that they should generate a diagnostic and link to #1395?

Sources/SwiftParser/Directives.swift
```diff
@@ -106,29 +109,54 @@ extension Parser {
...
+          lookahead.consumeAnyToken()
+          guard lookahead.at(identifierSpec), lookahead.currentToken.tokenText == "elif", lookahead.currentToken.leadingTriviaText.isEmpty else {
+            return false
+          }
+          lookahead.consumeAnyToken()
+          return lookahead.at(identifierSpec)
+        }
+
+        if atElifTypo {
+          (unexpectedBeforePoundIf, poundIf) = self.eat(poundIfHandle)
```
ahoppen, line 130: `poundIf` really is only a pound, so I would rename the variable.

```suggestion
          (unexpectedBeforePound, pound) = self.eat(poundIfHandle)
```

Sources/SwiftParser/Directives.swift
```diff
@@ -106,29 +109,54 @@ extension Parser {
...
+          guard self.currentToken.leadingTriviaText.isEmpty, let identifier = self.consume(if: TokenSpec(.identifier, allowAtStartOfLine: false)) else {
```
ahoppen, line 131: I don’t think we need to check `self.currentToken.leadingTriviaText.isEmpty` here since we also aren’t checking that `poundIf` contains no trailing trivia.

ahoppen (commented):
This is looking good now. I’ve got a couple stylistic comments inline, otherwise this is ready to be merged.

Sources/SwiftParser/Directives.swift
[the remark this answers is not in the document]
TiagoMaiaL: Nice @ahoppen, thank you. I'll start working on it.

ahoppen (approved):
Thank you. Let’s 🚢it

ahoppen:
@swift-ci Please test

Sources/SwiftParser/Directives.swift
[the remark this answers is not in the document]
ahoppen: Awesome, thanks!

ahoppen:
@swift-ci Please test

TiagoMaiaL:
@ahoppen I'm not sure why this is failing. Here's what I'm seeing from the pipeline:
Screenshot 2023-03-09 at 14 45 00

I'm also seeing some tests fail on main:
Screenshot 2023-03-09 at 19 18 06

I'm using Xcode 14.0.1, and MacOS 13.0

ahoppen:
`Backtracing/CrashAsync.swift` is currently a known failure that also fails on `main`. You can see the current status of main on https://ci.swift.org/job/oss-swift-pr-test-macoss/. So that’s nothing you need to worry about. It should be resolved soon.

`Parse/ConditionalCompilation/language_version.swift` appears to be a real failure caused by your changes. You should be able to reproduce it by
- Cloning https://github.com/apple/swift/
- Building your branch of SwiftSyntax and running `swift-parser-cli dump-parse path/to/swift/Parse/ConditionalCompilation/language_version.swift`
- You should also be able to create a smaller reproducer of the problem using `swift-parser-cli reduce path/to/swift/Parse/ConditionalCompilation/language_version.swift`
- You should then add a test case with the reduced source outputted by the command above and fix wherever it’s crashing

Those other test cases look odd. I can’t reproduce them on my machine. Can you expand the test failures and send me the difference between the expected and generated source?

TiagoMaiaL:
Hey @ahoppen, I've performed the steps you suggested above, and I've created some issues as a result. 

The failing tests on main are being described by issue https://github.com/apple/swift-syntax/issues/1410

Regarding the parsing error on `Parse/ConditionalCompilation/language_version.swift`, issue https://github.com/apple/swift-syntax/issues/1409 directly affects this branch. What was supposed to be parsed as a type, is  being parsed as a macro, and later on we have a single `pound (#)` char that's being parsed in `parsePoundIfDirective` and hitting the precondition in the default case of the `switch`.

Any ideas on how to proceed?

ahoppen:
Hey @TiagoMaiaL,
I replied to your comments on the two issues. I think both of them might behaving as expected. Regarding the `Parse/ConditionalCompilation/language_version.swift` failure is the following command crashing/hitting an assertion for you (I would expect that based on the test failure) or does it just output parser failures?
```bash
swift-parser-cli dump-parse path/to/swift/Parse/ConditionalCompilation/language_version.swift
```

ahoppen:
@swift-ci Please test

TiagoMaiaL:
> Hey @TiagoMaiaL, I replied to your comments on the two issues. I think both of them might behaving as expected. Regarding the `Parse/ConditionalCompilation/language_version.swift` failure is the following command crashing/hitting an assertion for you (I would expect that based on the test failure) or does it just output parser failures?
> 
> ```shell
> swift-parser-cli dump-parse path/to/swift/Parse/ConditionalCompilation/language_version.swift
> ```

Hey @ahoppen, `dump-parse` is not a valid sub-command. I believe you meant `print-tree`, right? If I run this command on my branch, the parser hits the `preconditionFailure` on `Directives.swift, line 125:
```swift
          (unexpectedBeforePound, pound) = self.eat(poundIfHandle)
          firstIteration = false
          switch pound.tokenKind {
          case .poundIfKeyword, .poundElseifKeyword:
            // ...
          case .poundElseKeyword:
            // ...
          default:
            // pound.tokenKind is simply `#`.
            preconditionFailure("The loop condition should guarantee that we are at one of these tokens")
          }
```

This is happening when we parse the following source from that file:
```swift
let x: @#$()%&*)@#$(%&*
```

The parser interprets the `#` as a macro expansion, which I think is not right. The code then exits the elements parsing loop (on line 159), with the `currentToken` being a `pound`, which enters the `preconditionFailure` branch in the `default` switch case.

On `main`, the parser doesn't crash. Here's the relevant portion of the `print-tree` output:
```
    17: CodeBlockItemSyntax children=1
      0: StructDeclSyntax children=3
        0: keyword(SwiftSyntax.Keyword.struct)
        1: identifier("S")
        2: MemberDeclBlockSyntax children=3
          0: leftBrace
          1: MemberDeclListSyntax children=5
            0: MemberDeclListItemSyntax children=2
              0: IfConfigDeclSyntax children=2
                0: IfConfigClauseListSyntax children=2
                  0: IfConfigClauseSyntax children=3
                    0: poundIfKeyword
                    1: FunctionCallExprSyntax children=4
                        0: identifier("swift")
                      1: leftParen
                      2: TupleExprElementListSyntax children=1
                        0: TupleExprElementSyntax children=1
                          0: PrefixOperatorExprSyntax children=2
                            0: prefixOperator(">=")
                            1: FloatLiteralExprSyntax children=1
                              0: floatingLiteral("2.2")
                      3: rightParen
                    2: MemberDeclListSyntax children=1
                      0: MemberDeclListItemSyntax children=1
                        0: VariableDeclSyntax children=2
                          0: keyword(SwiftSyntax.Keyword.let)
                          1: PatternBindingListSyntax children=1
                            0: PatternBindingSyntax children=2
                              0: IdentifierPatternSyntax children=1
                                0: identifier("x")
                              1: TypeAnnotationSyntax children=2
                                0: colon
                                1: SimpleTypeIdentifierSyntax children=1
                                  0: identifier("Int")
                  1: IfConfigClauseSyntax children=2
                    0: poundElseKeyword
                    1: MemberDeclListSyntax children=2
                      0: MemberDeclListItemSyntax children=2
                        0: VariableDeclSyntax children=2
                          0: keyword(SwiftSyntax.Keyword.let)
                          1: PatternBindingListSyntax children=1
                            0: PatternBindingSyntax children=2
                              0: IdentifierPatternSyntax children=1
                                0: identifier("x")
                              1: TypeAnnotationSyntax children=2
                                0: colon
                                1: AttributedTypeSyntax children=2
                                  0: AttributeListSyntax children=1
                                    0: AttributeSyntax children=2
                                      0: atSign
                                      1: MissingTypeSyntax
                                  1: MissingTypeSyntax
                        1: semicolon MISSING
                      1: MemberDeclListItemSyntax children=1
                        0: MacroExpansionDeclSyntax children=6
                          0: pound
                          1: UnexpectedNodesSyntax children=1
                            0: dollarIdentifier("$")
                          2: identifier("") MISSING
                          3: leftParen
                          4: TupleExprElementListSyntax
                          5: rightParen
                1: poundEndifKeyword MISSING
              1: semicolon MISSING
            1: MemberDeclListItemSyntax children=2
              0: FunctionDeclSyntax children=3
                0: keyword(SwiftSyntax.Keyword.func) MISSING
                1: binaryOperator("%&*")
                2: FunctionSignatureSyntax children=1
                  0: ParameterClauseSyntax children=3
                    0: leftParen MISSING
                    1: FunctionParameterListSyntax
                    2: rightParen
              1: semicolon MISSING
            2: MemberDeclListItemSyntax children=2
              0: MissingDeclSyntax children=1
                0: AttributeListSyntax children=1
                  0: AttributeSyntax children=2
                    0: atSign
                    1: MissingTypeSyntax
              1: semicolon MISSING
            3: MemberDeclListItemSyntax children=2
              0: MacroExpansionDeclSyntax children=6
                0: pound
                1: UnexpectedNodesSyntax children=1
                  0: dollarIdentifier("$")
                2: identifier("") MISSING
                3: leftParen
                4: TupleExprElementListSyntax children=1
                  0: TupleExprElementSyntax children=1
                    0: MissingExprSyntax
                5: rightParen MISSING
              1: semicolon MISSING
            4: MemberDeclListItemSyntax children=1
              0: FunctionDeclSyntax children=3
                0: keyword(SwiftSyntax.Keyword.func) MISSING
                1: binaryOperator("%&*")
                2: FunctionSignatureSyntax children=1
                  0: ParameterClauseSyntax children=3
                    0: leftParen MISSING
                    1: FunctionParameterListSyntax
                    2: rightParen MISSING
          2: rightBrace MISSING
  1: UnexpectedNodesSyntax children=20
    0: poundEndifKeyword
    1: rightBrace
    2: poundIfKeyword
    3: identifier("swift")
    4: leftParen
    5: prefixOperator(">=")
    6: floatingLiteral("2.2")
    7: rightParen
    8: keyword(SwiftSyntax.Keyword.var)
    9: identifier("zzz")
    10: equal
    11: stringQuote
    12: stringSegment("zzz")
    13: stringQuote
    14: poundElseKeyword
    15: keyword(SwiftSyntax.Keyword.var)
    16: identifier("zzz")
    17: equal
    18: identifier("zzz")
    19: poundEndifKeyword
  2: eof
```

ahoppen:
Yes, that makes sense to me. I think it’s fine that the parser parsers the `#` as a macro expansion. There isn’t really any better way to parse it.

The problem here is that you have loosened the loop condition by adding `pound` to `IfConfigContinuationClauseStartKeyword` and now the precondition is wrong just as it’s saying. It might be sufficient to just `break` out of the loop for `case .pound` because you have handled the `#elif` case above and we haven’t consumed any tokens when we `break` when we’re at `pound`.

So, I think you would need to add a label `LOOP` before `while let` in line 109 and the following to the `switch`. 
```swift
case .pound:
  break LOOP
```

I haven’t tested it though.

TiagoMaiaL:
Hey @ahoppen, thank you for the help. I did as you suggested, and tests are passing. I noted, however, that the `swift-parser-cli reduce` command prints the following on my branch:
`#if@#%`

As opposed to `main`, which prints the following:
`Error: Source file passed to reduce subcommand does not fail to roundtrip`

Any guidance on how to fix this is appreciated. Thank you.

ahoppen:
What you are describing seems correct to me.

`swift-parser-cli reduce` will only do something useful to a test that crashes the compiler or fails to round-trip. If this is the case, it produces a smaller test case that has the same problem. So when `swift-parser-cli reduce` outputs `Error: Source file passed to reduce subcommand does not fail to roundtrip` on `main`, it’s telling you that it didn’t find one of these issues.

With your branch, it created the reduced `#if@#%` test case. You should be able to add an `AssertParse` test case to the test suite for that. It doesn’t really matter what diagnostics that test case produces (since it’s just a silly example) as long as it doesn’t crash the parser or fail to round-trip.

Sources/SwiftParser/Directives.swift
```diff
@@ -106,29 +109,56 @@ extension Parser {
...
+          case .poundElseKeyword:
+            if let ifToken = self.consume(if: .init(.if, allowAtStartOfLine: false)) {
+              unexpectedBeforePound = RawUnexpectedNodesSyntax(combining: unexpectedBeforePound, pound, ifToken, arena: self.arena)
+              pound = self.missingToken(.poundElseifKeyword)
+              condition = RawExprSyntax(self.parseSequenceExpression(.basic, forDirective: true))
+            } else {
+              condition = nil
+            }
+          case .pound:
             condition = nil
```
TiagoMaiaL, line 152: @ahoppen I managed to fix the parser round-trip failure by setting `condition` to `nil` here. As I said before, I'm not sure if this is correct.

I've also added the reduced test case to ensure this parses without any round-trip failures or crashes.

Tests/SwiftParserTest/translated/IfconfigExprTests.swift
```diff
@@ -455,6 +455,82 @@ final class IfconfigExprTests: XCTestCase {
...
+    assertParse(
+      """
+      #if MY_FLAG
+      # elif
+      #endif
+      """
+    )
+  }
+
+  func testIfConfigExpr36() {
```
TiagoMaiaL, line 520: I've also added this reduced test case to ensure the parser parses without any round-trip failures or crashes.

Sources/SwiftParser/Directives.swift
ahoppen: I suspect the fix here is not because you set `condition` to `nil` here but because you added the `case .pound` and thus you’re not hitting the `preconditionFailure` in the `default` case. Because AFAICT `condition` is always `nil` when we hit this statement anyway.

---

I still think that the right implementation here would be to `break` out of the outer loop here by adding a label to the loop and breaking that. You can see that if you add a substructure test for `testIfConfigExpr36` – I usually do that by setting a breakpoint in `assertParse` and execute `e print(tree.debugInitCall(includeTrivia: false)` in the debugger and then adding the relevant output (here it would be the `IfConfigDeclSyntax` to the `substructure` parameter of `assertParse` + fix the remaining compilation errors.

In that case, you’ll have a `IfConfigClauseSyntax` where `poundKeyword` is a standalone `poundToken` while we expect `poundKeyword` to be one of `#if`, `#else`, etc.

Sources/SwiftParser/Directives.swift
```diff
@@ -106,29 +109,56 @@ extension Parser {
...
+          var lookahead = self.lookahead()
+          lookahead.consumeAnyToken()
+          guard lookahead.at(identifierSpec), lookahead.currentToken.tokenText == "elif", lookahead.currentToken.leadingTriviaText.isEmpty else {
+            return false
+          }
+          lookahead.consumeAnyToken()
+          return lookahead.at(identifierSpec)
+        }
+
+        if atElifTypo {
```
ahoppen, line 129: I have just been thinking: You should be able to write this in terms of `consume(if:followedBy:)`. That way you wouldn’t need the `atElifTypo` variable. Something like (written in GitHub, so might have some syntax errors):

```swift
if let (pound, identifier) = self.consume(if: { $0.tokenKind == .pound && $0.trailingTrivia.isEmpty }, followedBy: { $0.tokenKind == .identifier && $0.tokenText == "elif" && $0.leadingTrivia.isEmpty })
```

Sources/SwiftParser/Directives.swift
ahoppen: This is still incorrect. If we have a standalone `#` instead of a `#elseif`, we will happily eat it as the `#elseif` token and continue parsing. You need to break out of the loop if you see a standalone `#` that is not a `#elif`.

I think it will break in the following example, which AFAICt will not emit an error even though it should (haven’t tested though)

```swift
#if true
# someCondition
#endif
```

---

Also, I think you can move your `atElifTypo` inside this switch case. That way it‘s more performant because we don’t need to check for `#elif` in valid Swift code.

Sources/SwiftParser/Directives.swift
TiagoMaiaL: I managed to add the following: 
```swift
        if let (_pound, identifier) = self.consume(
          if: {
            $0.rawTokenKind == .pound && $0.trailingTriviaText.isEmpty
          }, followedBy: {
            $0.rawTokenKind == .identifier && $0.tokenText == "elif" && $0.leadingTriviaText.isEmpty
          }) { 
            // ...
          }
```

The problem is that we still use the `atElifTypo` in the while loop on line 149, for parsing the elements in between.

TiagoMaiaL:
> Also, I think you can move your atElifTypo inside this switch case. That way it‘s more performant because we don’t need to check for #elif in valid Swift code.

You mean moving the check inside the `.pound case`?

Sources/SwiftParser/Directives.swift
ahoppen: Ah, I didn’t realize that. In that case I agree that we still need `atElifTypo`. Thanks for clarifying.

ahoppen:
> Did you mean moving the check inside the `.pound case`?

Yes, exactly

TiagoMaiaL:
> > Did you mean moving the check inside the `.pound case`?
> 
> Yes, exactly

In that case we would need to change `atElifTypo` to start the check by expecting an identifier, as the previous `pound` would have been consumed. At the same time, if we do change the computed variable, the check on line 161 wouldn't work, as in that scenario we could have `pound` as the current token. I think we would need to change it to something like this:
```swift
... && !(self.at(.pound) && atElifTypo) && ...
```

ahoppen:
Oh, I see now. The real problem here is that the condition of the `while` loop is just really complicated. If it was just a `while let poundIfHandle = self.canRecoverTo`, we could switch over the returned `match` and eat the handle inside each `case `, like we do everywhere else in the parser.

Could you try the following?
- As a first step, try refactoring the pound if directive parsing to simplify the condition (not taking care of `#elif` yet)
  - Add `poundIf` too `IfConfigContinuationClauseStartKeyword` (and rename that type appropriately)
  - Remove the ternary operator from the loop and just have `while let (match, handle) = self.canRecoverTo(….)`
  - Switch over the match and push `self.eat` into each case
  - If we see a `#if` but `firstIteration == false`, we should break out of the loop without consuming `#if`
- Once that’s passing, you should be able to add `pound` to `IfConfigContinuationClauseStartKeyword` (or whatever you renamed it to before) and perform the `atElifTypo` check there.

Sources/SwiftParser/Directives.swift
```diff
@@ -103,32 +109,62 @@ extension Parser {
...
+
+        switch match {
+        case .poundIfKeyword:
+          if !firstIteration {
+            break LOOP
+          }
+          firstIteration = false
+          (unexpectedBeforePound, pound) = self.eat(handle)
+          condition = RawExprSyntax(self.parseSequenceExpression(.basic, forDirective: true))
+        case .poundElseifKeyword:
```
ahoppen, line 138: If we are in the first iteration, we should break out of the loop here to maintain the previous behavior. I.e. add

```swift
if firstIteration {
  break LOOP
}
```

(same for all other cases that aren’t `poundIfKeyword`)

Sources/SwiftParser/Directives.swift
```diff
@@ -103,32 +109,62 @@ extension Parser {
...
       var loopProgress = LoopProgressCondition()
-      while let poundIfHandle = firstIteration ? self.canRecoverTo(.poundIfKeyword) : self.canRecoverTo(anyIn: IfConfigContinuationClauseStartKeyword.self)?.handle,
-        loopProgress.evaluate(self.currentToken)
-      {
-        var (unexpectedBeforePoundIf, poundIf) = self.eat(poundIfHandle)
-        firstIteration = false
-        // Parse the condition.
+      LOOP: while let (match, handle) = self.canRecoverTo(anyIn: PoundIfDirectiveKeywords.self), loopProgress.evaluate(self.currentToken) {
+        var unexpectedBeforePound: RawUnexpectedNodesSyntax?
+        var pound: RawTokenSyntax
```
ahoppen, line 114: What do you think about naming this `poundKeyword` instead of `pound`. Because in almost all cases, it won’t be a `#` by itself. Same for `unexpectedBeforePound` -> `unexpectedBeforePoundKeyword`.

Sources/SwiftParser/Directives.swift
```diff
@@ -103,32 +109,62 @@ extension Parser {
...
+            return false
+          }
+          let identifierSpec = TokenSpec(.identifier, allowAtStartOfLine: false)
+          var lookahead = self.lookahead()
+          lookahead.consumeAnyToken()
+          guard lookahead.at(identifierSpec), lookahead.currentToken.tokenText == "elif", lookahead.currentToken.leadingTriviaText.isEmpty else {
+            return false
+          }
+          lookahead.consumeAnyToken()
+          return lookahead.at(identifierSpec)
```
ahoppen, line 127: I would duplicate `identifierSpec` because really it’s just coincidence that both the `elif` and the following identifier need to both be identifiers and on the same line. I’d also add a few comments

```suggestion
          var lookahead = self.lookahead()
          lookahead.consumeAnyToken() // consume `#`
          guard lookahead.at(TokenSpec(.identifier, allowAtStartOfLine: false)), lookahead.currentToken.tokenText == "elif", lookahead.currentToken.leadingTriviaText.isEmpty else {
            return false // `#` and `elif` must not be separated by trivia
          }
          lookahead.consumeAnyToken() // consume `elif`
          // We are only at a `elif` typo if it’s followed by an identifier for the condition. `#elif` or `#elif(…)` could be macro invocations.
          return lookahead.at(TokenSpec(.identifier, allowAtStartOfLine: false)) 
```

Sources/SwiftParser/Directives.swift
```diff
@@ -103,32 +109,62 @@ extension Parser {
...
             condition = RawExprSyntax(self.parseSequenceExpression(.basic, forDirective: true))
           } else {
             condition = nil
           }
-        default:
-          preconditionFailure("The loop condition should guarantee that we are at one of these tokens")
+        case .pound:
+          if atElifTypo {
+            (unexpectedBeforePound, pound) = self.eat(handle)
+            guard let identifier = self.consume(if: TokenSpec(.identifier, allowAtStartOfLine: false)) else {
```
ahoppen, line 153: Nitpick to make it a little easier to read.

```suggestion
            guard let elif = self.consume(if: TokenSpec(.identifier, allowAtStartOfLine: false)) else {
```

ahoppen (commented):
Thanks. That looks a lot cleaner to me now. I’ve got a few minor comments inline. If tests are passing, I think we’re getting close to merging this now.

Sources/SwiftParser/Directives.swift
```diff
@@ -103,32 +109,72 @@ extension Parser {
...
       var firstIteration = true
       var loopProgress = LoopProgressCondition()
-      while let poundIfHandle = firstIteration ? self.canRecoverTo(.poundIfKeyword) : self.canRecoverTo(anyIn: IfConfigContinuationClauseStartKeyword.self)?.handle,
-        loopProgress.evaluate(self.currentToken)
-      {
-        var (unexpectedBeforePoundIf, poundIf) = self.eat(poundIfHandle)
-        firstIteration = false
-        // Parse the condition.
+      LOOP: while let (match, handle) = self.canRecoverTo(anyIn: PoundIfDirectiveKeywords.self), loopProgress.evaluate(self.currentToken) {
+        var unexpectedBeforePound: RawUnexpectedNodesSyntax?
```
ahoppen, line 113:
```suggestion
        var unexpectedBeforePoundKeyword: RawUnexpectedNodesSyntax?
```

Sources/SwiftParser/Directives.swift
```diff
@@ -103,32 +109,72 @@ extension Parser {
...
+        var poundKeyword: RawTokenSyntax
         let condition: RawExprSyntax?
-        switch poundIf.tokenKind {
-        case .poundIfKeyword, .poundElseifKeyword:
+        var atElifTypo: Bool {
+          guard self.at(TokenSpec(.pound)), self.currentToken.trailingTriviaText.isEmpty else {
+            return false
+          }
+          var lookahead = self.lookahead()  // consume `#`
+          lookahead.consumeAnyToken()
```
ahoppen, line 121:
```suggestion
          var lookahead = self.lookahead()
          lookahead.consumeAnyToken()  // consume `#`
```

ahoppen (commented):
Two minor comments, then this will be ready to go 🚀

ahoppen (approved):
Let’s run it through CI again 🤞🏽

ahoppen:
@swift-ci Please test

ahoppen:
CI is running a few additional tests that are disabled locally by default because they have significant performance impact on the parser. Essentially what these tests do, is to modify the source passed to `assertParse` in interesting™ ways and making sure we don’t hit assertions or round-trip failures after that. (https://github.com/apple/swift-syntax/pull/1340 has some information). The easiest way to enable them in Xcode is to change Package.swift as follows

```diff
--- a/Package.swift
+++ b/Package.swift
@@ -22,14 +22,14 @@ if ProcessInfo.processInfo.environment["SWIFT_BUILD_SCRIPT_ENVIRONMENT"] != nil
     ]),
   ]
 }
-if ProcessInfo.processInfo.environment["SWIFTSYNTAX_ENABLE_RAWSYNTAX_VALIDATION"] != nil {
+if ProcessInfo.processInfo.environment["SWIFTSYNTAX_ENABLE_RAWSYNTAX_VALIDATION"] != nil || true {
   swiftSyntaxSwiftSettings += [
     .define("SWIFTSYNTAX_ENABLE_RAWSYNTAX_VALIDATION")
   ]
 }
 
 var swiftParserSwiftSettings: [SwiftSetting] = []
-if ProcessInfo.processInfo.environment["SWIFTPARSER_ENABLE_ALTERNATE_TOKEN_INTROSPECTION"] != nil {
+if ProcessInfo.processInfo.environment["SWIFTPARSER_ENABLE_ALTERNATE_TOKEN_INTROSPECTION"] != nil || true {
   swiftParserSwiftSettings += [
     .define("SWIFTPARSER_ENABLE_ALTERNATE_TOKEN_INTROSPECTION")
   ]
```

If you then run `testBogusSwitchStatement` without `SKIP_LONG_TESTS=1` you will hit the assertion failure. In one of the stack frames further down you can print the source code that hits the assertion  using `e print(mutatedSource)`. In this case it’s the following (which you can probably reduce a bit further using `swift-parser-cli reduce`).

```swift
switch x {
  #()
#if true
  bar()
#endif
  case .A, .B:
    break
}
```

TiagoMaiaL:
Thanks for the information, @ahoppen. I did some comparisons between `main` and my branch, using a test case involving the code you shared, and here's what happens:
1. on `main`, the tokens `#()` are parsed as `unexpectedBeforePoundKeyword`
2. on my branch, these tokens (`#()`) are skipped due to the `break LOOP` on line 159, inside `case .pound` (after checking if this was the `firstIteration`)

If I try to recover parsing  `poundIfKeyword`, it works normally. So I was considering two approaches:
1. Having the `case .poundIfKeyword` as the last case in the switch, and then `fallthrough` to it, instead of breaking the loop in the `.pound` case
2. Having an internal function that tries to parse the initial `.poundIfKeyword` when we hit the other cases, if `firstIteration` is `true`

My understanding is limited, so I would like to see if you have some suggestions, if possible.

ahoppen:
Oh, I see what’s going wrong. My suggestion of how to unify the loop condition was simply wrong. I’m sorry for that. But I’m making up suggestions of how to fix this as we go as well and don’t know the final outcome myself… 

Your analysis was correct, in `parseSwitchCases` we assume that we will make progress because we can recover to a `#if` but then we bail out of `parsePoundIfDirective` without consuming anything because we execute `break LOOP` after hitting the standalone `#`.

---

I did some testing myself now and am fairly confident that the following should now really work (it does in my little prototype implementation I did locally at least). The idea is to just handle all the parsing of the initial `#if` block before the loop. That way the first call inside `parsePoundIfDirective` will be `self.expect(.poundIf)`, which will guarantee to make progress as expected by `parseSwitchCases`:
- I would extract the parsing of the elements into a standalone function  `parseIfconfigClauseElements`. That way it can be shared between the parsing of the initial `#if` block and any consecutive blocks https://github.com/apple/swift-syntax/blob/36b0f23c4b0a3baa928b71aeacf08c17e998838a/Sources/SwiftParser/Directives.swift#L129-L141
- Start by `self.expect(.poundIfKeyword)` followed by parsing the condition and the elements (using the extracted function), outside the loop and add the first element to `clauses`
- For the loop, remove `#if` from `PoundIfDirectiveKeywords` again, since it’s handled before now (and rename that type back)
- And since you are touching this already:  As you encounter `do` statements, you can just remove them and flatten their contents into the surrounding function. I don’t think they provide any benefit here.

---

> I'll try to run the tests the way you suggested. When you say "running without `SKIP_LONG_TESTS=1`", where do I set this configuration?

https://github.com/apple/swift-syntax/blob/main/CONTRIBUTING.md#xctests under `Tip` has some documentation on this environment variable. But it looks like you haven’t set it so far, so there’s nothing you needed to change.

ahoppen (approved):
Nice. This reads so much better better with the `#if` parsing extracted than what we currently have in `main`. 😍

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

TiagoMaiaL:
@ahoppen If I'm not mistaken, the CI failures don't seem related to the changes in this PR. Is there anything I can do?

ahoppen:
Oh, yes. I think we just hit bad timing where CI picked up one of the PRs related to https://github.com/apple/swift-syntax/pull/1496 but not the other one. I’ll just trigger CI again.

ahoppen:
@swift-ci Please test Windows

PR #1359, opened 2023-02-23
[Macros] Add executable compiler plugin support library

rintaro:
`SwiftCompilerPlugin` is a helper library to write an executable compiler plugin.

Usage:
  ```swift
  import SwiftCompilerPlugin
  import SwiftSyntaxMacros

  @main
  struct MyPlugin: CompilerPlugin {
    var providingMacros: [Macro.Type] = [MyMacro.self]
  }

  struct MyMacro: ExpressionMacro {
    ...
  }
  ```


rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

rintaro:
https://github.com/apple/swift/pull/63888
@swift-ci Please test

rintaro:
https://github.com/apple/swift/pull/63888
@swift-ci Please test

rintaro:
@ahoppen Let me know if you have any concern with the `build-script.py` changes.

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

Sources/SwiftCompilerPlugin/CompilerPlugin.swift
```diff
@@ -0,0 +1,246 @@
...
+  let outputStream: FileHandle
+
+  func sendMessage(_ message: TX) throws {
+    // Encode the message as JSON.
+    let payload = try JSONEncoder().encode(message)
+
+    // Write the header (a 64-bit length field in little endian byte order).
+    var count = UInt64(payload.count).littleEndian
+    let header = Swift.withUnsafeBytes(of: &count) { Data($0) }
+    assert(header.count == 8)
```
ahoppen, line 184: Just a few thought that crossed my mind. I’m fine with the way it is if you think it’s the best solution.

SourceKit-LSP uses the following HTTP-style header, which could allow us to add more options in the future (e.g. encoding of the message). I can’t really think of any options we want to add here but having the flexibility could come in handy in the future.

```
Content-Length: <some number in ASCII>\r\n\r\n
```

---

Alternatively, everything else in the communication channel is human readable, so I’m wondering whether it makes sense to also make this header human-readable. That might make debugging the plugin easier. I’m not sure if it’s worth it though.

Sources/SwiftCompilerPlugin/Diagnostics.swift
```diff
@@ -0,0 +1,125 @@
...
+  case macroTypeNotFound = "macro expanding type not found"
+  case freestandingMacroSyntaxIsNotMacro = "macro syntax couldn't be parsed"
+}
+
+extension MacroExpansionError: DiagnosticMessage {
+  var message: String {
+    self.rawValue
+  }
+  var diagnosticID: SwiftDiagnostics.MessageID {
+    .init(domain: "\(type(of: self))", id: "\(self)")
```
ahoppen, line 27: Should we include the name of the module in the ID? We do that in the parser.

Sources/SwiftCompilerPlugin/PluginMessages.swift
```diff
@@ -0,0 +1,134 @@
...
+    case memberAttribute
+    case member
+    case peer
+    case conformance
+  }
+
+  struct SourceLocation: Codable {
+    var fileID: String
+    var fileName: String
+    var offset: Int
```
ahoppen, line 81: I assume this `offset` and `column` are UTF-8 byte offsets. Could you add doc comments for this? I’ve wondered way too many times before SwiftSyntax added comments for these all over the place.

Same for all the other `offset` fields in this file.

Sources/SwiftCompilerPlugin/PluginMacroExpansionContext.swift
```diff
@@ -0,0 +1,253 @@
...
+import SwiftOperators
+import SwiftSyntax
+import SwiftSyntaxMacros
+
+/// Manages known source code combined with their filename/fileID. This can be
+/// used to get line/column from a syntax node in the managed source code.
+class SourceManager {
+  class KnownSourceSyntax {
+    struct Location {
+      var offset: Int
```
ahoppen, line 24: Same here, could you add that these are UTF-8 byte offsets.

Sources/SwiftCompilerPlugin/PluginMacroExpansionContext.swift
```diff
@@ -0,0 +1,253 @@
...
+      var line: Int
+      var column: Int
+      var fileID: String
```
ahoppen, line 27: Could you add a comment what `fileID` is and how it relates to `fileName`?

Sources/SwiftCompilerPlugin/PluginMessages.swift
```diff
@@ -0,0 +1,134 @@
...
+    case freeStandingDeclaration
+    case accessor
+    case memberAttribute
+    case member
+    case peer
+    case conformance
+  }
+
+  struct SourceLocation: Codable {
+    var fileID: String
```
ahoppen, line 79: Could you add a comment what `fileID` is and how it relates to `fileName`?

Sources/SwiftCompilerPlugin/PluginMacroExpansionContext.swift
```diff
@@ -0,0 +1,253 @@
...
+      node = Syntax(TypeSyntax.parse(from: &parser))
+    case .pattern:
+      node = Syntax(PatternSyntax.parse(from: &parser))
+    case .attribute:
+      node = Syntax(AttributeSyntax.parse(from: &parser))
+    }
+    do {
+      node = try OperatorTable.standardOperators.foldAll(node)
+    } catch let error {
+      // TODO: Error handling.
```
ahoppen, line 83: Just to make sure: You know that this is left over? If operator folding fails, it might be OK to just leave the unresolved nodes in the tree, i.e. just keep the `catch` clause empty.

Sources/SwiftCompilerPlugin/PluginMacroExpansionContext.swift
```diff
@@ -0,0 +1,253 @@
...
+
+    let localPosition = node.position(at: kind)
+    let localLocation = base.locationConverter.location(for: localPosition)
+
+    let positionOffset = base.location.offset
+    let lineOffset = base.location.line - 1
+    let columnOffset = localLocation.line == 1 ? base.location.column : 0
+
+    return SourceLocation(
+      line: localLocation.line! + lineOffset,
```
ahoppen, line 160: I forgot why `SourceLocation` has an optional `line` and `column` but are we sure that these are non-nil here? If yes, could you add a doc comment explaining why we can assume they are non-nil here?

Sources/SwiftCompilerPlugin/PluginMacroExpansionContext.swift
```diff
@@ -0,0 +1,253 @@
...
+      column: localLocation.column! + columnOffset,
+      offset: localLocation.offset + positionOffset,
+      file: file
+    )
+  }
+}
+
+fileprivate extension Syntax {
+  /// get a position in the node by `PositionInSyntaxNode`.
```
ahoppen, line 169: Nit

```suggestion
  /// Get a position in the node by `PositionInSyntaxNode`.
```

Sources/SwiftCompilerPlugin/PluginMacroExpansionContext.swift
```diff
@@ -0,0 +1,253 @@
...
+    }
+    let localPosition = node.position(at: kind)
+    let positionOffset = base.location.offset
+    return SourcePosition(
+      fileName: base.location.fileName,
+      utf8Offset: localPosition.advanced(by: positionOffset).utf8Offset
+    )
+  }
+
+  /// Get position range of `node` in the known root nodes.
```
ahoppen, line 117: Did you mean

```suggestion
  /// Get range of `node` in the known root nodes.
```

Sources/SwiftCompilerPlugin/PluginMacroExpansionContext.swift
```diff
@@ -0,0 +1,253 @@
...
+  ///
+  /// The expansion discriminator is combined with the `uniqueNames` counters
+  /// to produce unique names.
+  private var discriminator: String
+
+  private var sourceManger: SourceManager
+
+  /// Counter for each of the uniqued names.
+  ///
+  /// Used in conjunction with `expansionDiscriminator`.
```
ahoppen, line 196: I think `expansionDiscriminator` no longer exists.

Sources/SwiftCompilerPlugin/PluginMacroExpansionContext.swift
```diff
@@ -0,0 +1,253 @@
...
+  private var uniqueNames: [String: Int] = [:]
```
ahoppen, line 197: Wouldn’t `nextUnusedDiscriminatorIndex` be a better name?

Sources/SwiftCompilerPlugin/PluginMacroExpansionContext.swift
```diff
@@ -0,0 +1,253 @@
...
+    resultString += "\(name.count)\(name)"
+
+    // Mangle the operator for unique macro names.
+    resultString += "fMu"
+
+    // Mangle the index.
+    if uniqueIndex > 0 {
+      resultString += "\(uniqueIndex - 1)"
+    }
+    resultString += "_"
```
ahoppen, line 232: Does this need to match any mangling logic in apple/swift? If so, I think we should add a doc comment explaining it. If not: Why do we need the `fMu` mangling prefix?

Sources/SwiftCompilerPlugin/Macros.swift
```diff
@@ -0,0 +1,254 @@
...
+//===----------------------------------------------------------------------===//
+
+import SwiftDiagnostics
+import SwiftSyntax
+import SwiftSyntaxMacros
+
+/// Implementation for `CompilerPlugin` macro related request processing.
+extension CompilerPlugin {
+  public  // @testable
+    func resolveMacro(moduleName: String, typeName: String) -> Macro.Type?
```
ahoppen, line 20: What I’ve done before (and what I’m relatively happy with) to make methods public for testing is to define the real function as internal and then add a public thunk around it that is prefixed with `_` and marked as only being public for testing purposes.

Sources/SwiftCompilerPlugin/Macros.swift
```diff
@@ -0,0 +1,254 @@
...
+        .expandFreestandingMacroResult(
+          expandedSource: "",
+          diagnostics: [diag]
+        )
+      )
+      return
+    }
+
+    let macroDef = self.resolveMacro(macro)
+    ///
```
ahoppen, line 68: Leftover comment?

Sources/SwiftCompilerPlugin/Macros.swift
```diff
@@ -0,0 +1,254 @@
...
+        PluginMessage.Diagnostic(from: $0, in: sourceManager)
+      }
+
+      let resultMessage = PluginToHostMessage.expandFreestandingMacroResult(
+        expandedSource: rewritten.description,
+        diagnostics: diagnostics
+      )
+
+      try pluginHostConnection.sendMessage(resultMessage)
+      return
```
ahoppen, line 96: Could we share this with `ExpressionMacro`?

Sources/SwiftCompilerPlugin/Macros.swift
```diff
@@ -0,0 +1,254 @@
...
+    parentDeclSyntax: PluginMessage.Syntax?
+  ) throws {
+    let sourceManager = SourceManager()
+    let context = PluginMacroExpansionContext(
+      sourceManager: sourceManager,
+      discriminator: discriminator
+    )
+
+    guard let macroDefinition = resolveMacro(macro) else {
+      fatalError("macro type not found: \(macro.moduleName).\(macro.typeName)")
```
ahoppen, line 125: Could we emit a diagnostic here similar to freestanding macros? Alternatively, I think this should have `FIXME` to highlight we need to fix this in the future.

Same for quite a few `fatalError`s below.

Sources/SwiftCompilerPlugin/Macros.swift
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    let customAttributeNode = sourceManager.add(customAttributeSyntax).cast(AttributeSyntax.self)
```
ahoppen, line 128: Mostly out of interest: Why is this named `customAttributeSyntax`. The `CustomAttributeSyntax` type in SwiftSyntax has been renamed to `AttributeSyntax`.

Sources/SwiftCompilerPlugin/Macros.swift
```diff
@@ -0,0 +1,254 @@
...
+    switch (macroDefinition, macroRole) {
+    case (let attachedMacro as AccessorMacro.Type, .accessor):
+      let accessors = try attachedMacro.expansion(
+        of: customAttributeNode,
+        providingAccessorsOf: declarationNode,
+        in: context
+      )
+      expanded =
+        accessors
+        .map { $0.trimmedDescription }
```
ahoppen, line 141: I think it would read better in one line but if you disagree, that’s also fine. Similar below.

Sources/SwiftCompilerPlugin/Macros.swift
```diff
@@ -0,0 +1,254 @@
...
+      guard
+        let parentDeclSyntax = parentDeclSyntax,
+        let parentDeclGroup = sourceManager.add(parentDeclSyntax).asProtocol(DeclGroupSyntax.self)
+      else {
+        fatalError("parentDecl is mandatory for MemberAttributeMacro")
+      }
+
+      // Local function to expand a member atribute macro once we've opened up
+      // the existential.
+      func expandMemberAttributeMacro<Node: DeclGroupSyntax>(
```
ahoppen, line 153: Couldn’t the content of this just be a trailing closure to `_openExistential`? Same below.

Tests/SwiftCompilerPluginTest/CompilerPluginTests.swift
```diff
@@ -0,0 +1,70 @@
...
+    let plugin = MyPlugin()
+
+    let registeredMacro = plugin.resolveMacro(
+      moduleName: "SwiftCompilerPluginTest",
+      typeName: "RegisteredMacro"
+    )
+    XCTAssertNotNil(registeredMacro)
+    XCTAssertTrue(registeredMacro == RegisteredMacro.self)
+
+    /// Test the plugin doesn't provide macros other than ``
```
ahoppen, line 61: Did you forget to write something in the backticks?

lit_tests/compiler_plugin_basic.swift
```diff
@@ -0,0 +1,29 @@
+// REQUIRES: platform=Darwin
```
ahoppen, line 1: Hmm, I was really hoping that we could get rid of the lit-based tests soon so that SwiftSyntax doesn’t have any dependency on tools in apple/swift. But it’s the best tool for the job here, so I think it’s fine for now.

Sources/SwiftCompilerPlugin/PluginMacroExpansionContext.swift
rintaro: cc: @DougGregor I copied this from `ASTGen`'s `SourceManager.MacroExpansionContext`. What's the requirement of this `createUniqueName`?

Sources/SwiftCompilerPlugin/Macros.swift
rintaro: [Aligning with `ASTGen`](https://github.com/apple/swift/blob/a0e1810f627add51fbbe3ddf2c3ce9a9c6e3afdc/lib/ASTGen/Sources/ASTGen/Macros.swift#L677-L695) cc: @DougGregor

Sources/SwiftCompilerPlugin/Macros.swift
rintaro: Just be cause ASTGen call it `customAttrNode` https://github.com/apple/swift/blob/a0e1810f627add51fbbe3ddf2c3ce9a9c6e3afdc/lib/ASTGen/Sources/ASTGen/Macros.swift#L438-L445

Sources/SwiftCompilerPlugin/PluginMacroExpansionContext.swift
rintaro: Aligning with ASTGen https://github.com/apple/swift/blob/a0e1810f627add51fbbe3ddf2c3ce9a9c6e3afdc/lib/ASTGen/Sources/ASTGen/SourceManager%2BMacroExpansionContext.swift#L21-L24)
If we rename it, we should do both.

Sources/SwiftCompilerPlugin/PluginMacroExpansionContext.swift
rintaro: `SourceLocation.line/column` are nil [only if the location converter was not supplied](https://github.com/apple/swift-syntax/blob/7fa6666bb8bde61ce3af2389a2731538091c2948/Sources/SwiftSyntax/SourceLocation.swift#L82-L91). And yes I am sure these cannot be nil, if the `offset` is out of the range, it's clamped to the range of the file.

And yes, I can add a comment.

Sources/SwiftCompilerPlugin/CompilerPlugin.swift
rintaro: This basic message format is from SwiftPM's plugin mechanism (https://github.com/apple/swift-package-manager/blob/48192bbdd2800a38a82fa658a244e70b3167ec5d/Sources/PackagePlugin/Plugin.swift#L30-L31). Although the compiler plugin mechanism doesn't need to follow it, I thought it's good to have the same messaging format among the toolchain.

Sources/SwiftCompilerPlugin/Diagnostics.swift
rintaro: Hm yeah, maybe `SwiftCompilerPlugin.MacroExpansionError` would make more sense. Is `domain` a user-visible value?

(fwiw, I just followed [`ParserDiagnosticMessages`](https://github.com/apple/swift-syntax/blob/main/Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift#L81))

Sources/SwiftCompilerPlugin/CompilerPlugin.swift
ahoppen: Sure, makes sense to me. I just wanted to hear your thoughts on my ideas.

Sources/SwiftCompilerPlugin/PluginMacroExpansionContext.swift
ahoppen: I think renaming both would make sense but it doesn’t need to be this PR.

Sources/SwiftCompilerPlugin/Macros.swift
ahoppen: I would assume that’s because ASTGen might pre-date the rename `CustomAttributeSyntax` -> `AttributeSyntax`. I think we should just call it `attributeNode` here (and ideally also rename `ASTGen` but that’s not as important IMO).

rintaro: Sounds good. Since this name is also used in `PluginMessage`, we need to change ASTGen too. I will open a PR to `swift` repo.

Sources/SwiftCompilerPlugin/Diagnostics.swift
rintaro: Ah, I think I misread something, so do you think this should be `.init(domain: "MacroExpansionError", id: "\(self)")`, or anything else?

ahoppen: Yes, exactly

Sources/SwiftCompilerPlugin/PluginMacroExpansionContext.swift
rintaro: Done.

Sources/SwiftCompilerPlugin/PluginMessages.swift
rintaro: Done

Sources/SwiftCompilerPlugin/PluginMacroExpansionContext.swift
rintaro: Added a no-op error handler to make it non-`throws`

Sources/SwiftCompilerPlugin/PluginMacroExpansionContext.swift
rintaro: Update the comment to clarify this is a range of "position" (filename + UTF-8 offset) not range of "location"  (file, line, col, and offset)

Sources/SwiftCompilerPlugin/PluginMacroExpansionContext.swift
rintaro: Renamed `discriminator` to `expansionDiscriminator` to align with ASTGen.

rintaro:
https://github.com/apple/swift/pull/63952
@swift-ci Please test

rintaro:
https://github.com/apple/swift/pull/63958
@swift-ci Please smoke test

rintaro:
https://github.com/apple/swift/pull/63958
@swift-ci Please test

rintaro:
https://github.com/apple/swift/pull/63958
@swift-ci Please test macOS

rintaro:
https://github.com/apple/swift/pull/63958
@swift-ci Please test

Sources/SwiftCompilerPlugin/PluginMacroExpansionContext.swift
ahoppen: Ah, makes sense. Thanks

ahoppen (approved):

PR #1360, opened 2023-02-24
[Macros] Correct the where clause syntax type in `ConformanceMacro`.

hborla:
`WhereClauseSyntax` is used for expression conditions in statements; `GenericWhereClauseSyntax` lists generic requirements for conditional conformances.


hborla:
@swift-ci please test

PR #1361, opened 2023-02-25
Accept access level attributes on imports

xymus:
This feature will be protected by an experimental flag on the compiler side.


xymus:
@swift-ci Please test

ahoppen (approved):

PR #1366, opened 2023-02-28
[CMake] Workaround stale object files causing rebuilds

bnbarham:
Consider the case of A -> B. The previous workaround will add a `forced-B-dep.swift` to A that is touched if B rebuilds. So if B is rebuilt, A is also rebuilt. But A itself has no real changes, and so none of its object files are built and none of the mtimes are updated.

Thus, A is then rebuilt again on the next run because its input (`forced-B-dep.swift`) is newer than the corresponding object file. To prevent this, remove all object files before calling the build (which is actually a link step that both compiles and links).


bnbarham:
@swift-ci please test

bnbarham:
@drodriguez I haven't tested this thoroughly, but a quick test of rebuilding swift-syntax itself with a comment change does work.

@artemcm I'll probably put this hack in either way, but it seems like maybe we should just update the object file mtime here? In general I don't think it's particularly uncommon for build systems to rebuild if the input mtime is newer than the output. Even if CMake was fixed to properly add a dependency to the swiftmodule here, this would still be a problem when there's actual new code that doesn't cause any object file changes.

etcwilde:
For my notes; it's because the object file is recorded in the Ninja build graph as an output of the compilation (along with the archive). Since the object file doesn't get updated, ninja continues to see it as an out-of-date product, and goes in for the rebuild. The driver should likely `touch` the files saying that they are in fact up-to-date to appease the Ninjas.

drodriguez:
The library where I found the problem does not use `add_swift_host_library`, though. It uses a naked `add_library`, so if I am not mistaken, this approach will not apply there? https://github.com/apple/swift/blob/f20eaac5be2f216dbbe2ddb44909ccb548ece048/lib/ASTGen/CMakeLists.txt#L21

What I tried was adding two `COMMAND` to https://github.com/apple/swift/blob/f20eaac5be2f216dbbe2ddb44909ccb548ece048/lib/ASTGen/CMakeLists.txt#L79. That way the object is only deleted if the dependency changes (leaving the rest of the fake files untouched). I was "hardcoding" a little the paths, but I think something can be done with `TARGET_OBJECTS` that I did not realized before.

```
COMMAND ${CMAKE_COMMAND} -E remove $<FILTER:$<TARGET_OBJECTS:${TARGET}>,INCLUDE,forced-${DEPENDENCY}-dep.swift>>
```

I imagine that forcing CMake to rebuild all the objects is not completely bad, but it will do a lot more work than necessary. Small-ish libraries might be fine, but basically every change in one of the leafs might cascade a full rebuild of everything (where now it is mostly skipped, since those object files are considered valid as long as their files do not have changes).

bnbarham:
> The library where I found the problem does not use add_swift_host_library, though. It uses a naked add_library, so if I am not mistaken, this approach will not apply there? https://github.com/apple/swift/blob/f20eaac5be2f216dbbe2ddb44909ccb548ece048/lib/ASTGen/CMakeLists.txt#L21

ASTGen depends on the various SwiftSyntax libraries, it's needed in both places. I'll put the swift one up tomorrow. It was just easier to start here since the build only takes a few seconds :)

> I imagine that forcing CMake to rebuild all the objects is not completely bad, but it will do a lot more work than necessary. Small-ish libraries might be fine, but basically every change in one of the leafs might cascade a full rebuild of everything (where now it is mostly skipped, since those object files are considered valid as long as their files do not have changes).

Unfortunately there's not really any way around it right now. Like I said, this isn't just the `forced-*-dep.swift `files - it is *any* Swift file where the change doesn't make it into the object file (eg. a comment). Downstream there shouldn't be any impact here since dependencies only care about the archive, not the individual objects.

Having said that, we could maybe just do a touch in the POST_BUILD instead, though we'd probably want to touch the actual archive at the same time so that it's timestamp isn't older than the object files.

bnbarham:
@swift-ci please test

DougGregor:
@swift-ci please test

PR #1368, opened 2023-02-28
Add newline for pund-keywords in `IfConfigDeclSyntax`

kimdv:


Tests/SwiftSyntaxBuilderTest/IfConfigDeclSyntaxTests.swift
```diff
@@ -0,0 +1,70 @@
...
+        )
+      },
+      poundEndif: .poundEndifKeyword(leadingTrivia: .newline)
+    )
+
+    AssertBuildResult(
+      buildable,
+      """
+
+      #if DEBUG
```
kimdv, line 58: The only "downside" is that it also counts for the `#if` and not only `#elseIf`

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test

Tests/SwiftSyntaxBuilderTest/IfConfigDeclSyntaxTests.swift
ahoppen: This should probably be fixed if you add a check to `shouldOmitNewline` that returns `true` if to token is the first in the source file, i.e. if `node.previousToken(view: .sourceAccurate) == nil`

kimdv:
@swift-ci please test

ahoppen (approved):

PR #1369, opened 2023-02-28
[Macros] Introduce AbstractSourceLocation and its APIs

DougGregor:
Introduce `AbstractSourceLocation` as a macro-focused API that uses opaque syntax rather than direct values. Update the MacroExpansionContext APIs for location access to use this type, leaving the old versions in place to allow for some staging for clients.

This aligns the location APIs with what is proposed in SE-0382.


DougGregor:
@swift-ci please test

PR #1370, opened 2023-02-28
Fix a misparser where we would classify a call to 'open' as a DeclModifier

ahoppen:
While I was working on this, I also noticed that we forgot to pass `isAtTopLevel` in a few places. They didn’t really matter here but I made sure to correctly pass the flag regardless.

### DIFF
--- a/Sources/SwiftParser/Declarations.swift
+++ b/Sources/SwiftParser/Declarations.swift
@@ -60,7 +60,9 @@ extension TokenConsumer {
         modifierProgress.evaluate(subparser.currentToken)
       {
         subparser.eat(handle)
-        if subparser.at(.leftParen) && modifierKind.canHaveParenthesizedArgument {
+        if modifierKind != .open && subparser.at(.leftParen) && modifierKind.canHaveParenthesizedArgument {
+          // When determining whether we are at a declaration, don't consume anything in parentheses after 'open'
+          // so we don't consider a function call to open as a decl modifier. This matches the C++ parser.
           subparser.consumeAnyToken()
           subparser.consume(to: .rightParen)
         }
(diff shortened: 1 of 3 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/SwiftParser/Declarations.swift
```diff
@@ -60,7 +60,9 @@ extension TokenConsumer {
         modifierProgress.evaluate(subparser.currentToken)
       {
         subparser.eat(handle)
-        if subparser.at(.leftParen) && modifierKind.canHaveParenthesizedArgument {
+        if modifierKind != .open && subparser.at(.leftParen) && modifierKind.canHaveParenthesizedArgument {
```
bnbarham, line 63: Why only for `open`?

Tests/SwiftParserTest/DeclarationTests.swift
```diff
@@ -286,7 +286,11 @@ final class DeclarationTests: XCTestCase {
       fileprivate fileprivate(set) var fileprivateProp = 0
       private private(set) var privateProp = 0
       internal(set) var defaultProp = 0
-      """
+      """,
+      diagnostics: [
```
bnbarham, line 290: `open open(set) var openProp = 0` is valid isn't it? Pointless, but valid 😅?

Sources/SwiftParser/Declarations.swift
ahoppen: Because all the other access modifiers are lexer classified keywords and thus can’t be used as function base names.

Tests/SwiftParserTest/DeclarationTests.swift
ahoppen: Yes, I agree that it should be valid but it’s rejected with the same error in the C++ parser.

Also, related, I would imagine that

```swift
func foo() {
  var open = 1
  open 
  var myVar = 2
}
```

is valid but it’s rejected in the C++ parser and the safest thing I can think of doing is to also reject it in the new parser. Any new heuristics to disambiguate `open` as an identifier vs decl modifier might introduce new source breakage.

bnbarham: Ah. It fails top level but doesn't fail when it's a member. Ie.
```
class Foo {
  open open(set) var openProp = 0
}
```
is fine.

PR #1374, opened 2023-03-01
[Diagnostics formatter] Consistently apply diagnostic kinds

DougGregor:
Address two issues regarding the diagnostic type indicators in formatted diagnostics.

1. When colors were disabled, we wouldn't get any of the diagnostic type indicators because the "colorize" check came too early. Make sure they're consistently applied whether colorized or not.
2. There was no "note:" prefix, so it wasn't clear which things were notes.


DougGregor:
@swift-ci please smoke test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

Issue #1262, opened 2023-01-24
Bad diagnostics for bogus effects specifier before arrow or brace in function 


ahoppen:
### Issue Kind

Bad Diagnostic Produced

### Source Code

```swift
func foo() bogus -> Int {}
() bogus reasync -> Int
```

### Description

Instead of using `at` to check for the `->` or `{`, we should check whether we can recover to it on the same line, possibly reducing the recovery precedence when parsing function types because they could be followed by an identifier.

ahoppen:
rdar://104621275

Rajveer100:
Hi @ahoppen, so with the grinding going on `swift compiler` already, I am also looking forward to contributing towards other apple repos like this one, so is there any additional required setup specifically for repos apart from `swift compiler,` as well as other potential ones which I could help and solve issues.

StevenWong12:
Hi @ahoppen, I've just started investigating this issue. I ran the parse cli to print the diagnosics for the code snippet above and I got the following diagnostics

```
1 │ func test() bogus -> Int {}
  ∣            ╰─ error: consecutive statements on a line must be separated by ';'
```

From my understanding, it should produce a more useful diagnostics message for that e.g.

```
1 │ func test() bogus -> Int {}
  ∣            ╰─ error: unknown effect specifier
```

Have I got that right? 

Thanks in advance!

ahoppen:
Yes, exactly. 

The first step would be to get it to produce something like the following by improving recovery inside there parser itself

```
1 │ func test() bogus -> Int {}
  ∣            ╰─ error: unexpected code 'bogus' in function declaration
```

As a second step you can then improve the diagnostic to `unknown effect specifier 'bogus'` in the diagnostic generator.

StevenWong12:
Great! Thanks for your advice. I am about to try it out and hope this will be my first contribution to this repo!

ahoppen:
Fixed by https://github.com/apple/swift-syntax/pull/1381

PR #1381, opened 2023-03-02
Produce more useful diagnostics for bogus specifier before arrow and leftbrace

StevenWong12:
## Purpose
To produce more useful diagnostics for bogus specifier before arrow and leftbrace as the first step to solve issue #1262

## Implementation details
Replace ``at`` with recovery function and a new testcase.

Additionally, to pass the ``testMacroDecl`` testcase, I made an edit about ``TokenPrecedence`` to get the correct precedence of the ``macro`` keyword.

I also modified some of the other testcases  but I am not sure if they're reasonable.

If there is anything I can do for this PR, please tell me. Many thanks!

### DIFF
--- a/Sources/SwiftParser/Declarations.swift
+++ b/Sources/SwiftParser/Declarations.swift
@@ -1375,7 +1375,8 @@ extension Parser {
     var effectSpecifiers = self.parseDeclEffectSpecifiers()
 
     let output: RawReturnClauseSyntax?
-    if let _ = self.canRecoverTo(TokenSpec(.arrow)) {
+
+    if self.at(.arrow) || self.canRecoverTo(TokenSpec(.arrow, recoveryPrecedence: .weakPunctuator, allowAtStartOfLine: false)) != nil {
       output = self.parseFunctionReturnClause(effectSpecifiers: &effectSpecifiers, allowNamedOpaqueResultType: true)
     } else {
       output = nil
--- a/Sources/SwiftParser/Declarations.swift
+++ b/Sources/SwiftParser/Declarations.swift
@@ -1375,7 +1375,7 @@ extension Parser {
     var effectSpecifiers = self.parseDeclEffectSpecifiers()
 
     let output: RawReturnClauseSyntax?
-    if self.at(.arrow) {
+    if let _ = self.canRecoverTo(TokenSpec(.arrow)) {
       output = self.parseFunctionReturnClause(effectSpecifiers: &effectSpecifiers, allowNamedOpaqueResultType: true)
     } else {
       output = nil
--- a/Sources/SwiftParser/Declarations.swift
+++ b/Sources/SwiftParser/Declarations.swift
@@ -1376,7 +1376,7 @@ extension Parser {
 
     let output: RawReturnClauseSyntax?
 
-    if self.at(.arrow) || self.canRecoverTo(TokenSpec(.arrow, recoveryPrecedence: .weakPunctuator, allowAtStartOfLine: false)) != nil {
+    if self.at(.arrow) || self.canRecoverTo(TokenSpec(.arrow, allowAtStartOfLine: false)) != nil {
       output = self.parseFunctionReturnClause(effectSpecifiers: &effectSpecifiers, allowNamedOpaqueResultType: true)
     } else {
       output = nil
--- a/Sources/SwiftParser/TokenPrecedence.swift
+++ b/Sources/SwiftParser/TokenPrecedence.swift
@@ -101,8 +101,8 @@ public enum TokenPrecedence: Comparable {
 
   @_spi(RawSyntax)
   public init(_ lexeme: Lexer.Lexeme) {
-    if let kw = Keyword(lexeme.tokenText) {
-      self.init(kw)
+    if lexeme.rawTokenKind == .keyword {
+      self.init(Keyword(lexeme.tokenText)!)
     } else {
       self.init(nonKeyword: lexeme.rawTokenKind)
     }
--- a/Sources/SwiftParser/TokenPrecedence.swift
+++ b/Sources/SwiftParser/TokenPrecedence.swift
@@ -216,7 +216,7 @@ public enum TokenPrecedence: Comparable {
       // Access modifiers
       .fileprivate, .internal, .private, .public, .static,
       // Functions
-      .deinit, .func, .`init`, .subscript, 
+      .deinit, .func, .`init`, .subscript,
       // Marco
       .macro,
       // Variables
--- a/Sources/SwiftParser/TokenPrecedence.swift
+++ b/Sources/SwiftParser/TokenPrecedence.swift
@@ -101,8 +101,8 @@ public enum TokenPrecedence: Comparable {
 
   @_spi(RawSyntax)
   public init(_ lexeme: Lexer.Lexeme) {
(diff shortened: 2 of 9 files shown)


StevenWong12:
@swift-ci please test

Sources/SwiftParser/TokenPrecedence.swift
```diff
@@ -101,8 +101,8 @@ public enum TokenPrecedence: Comparable {
 
   @_spi(RawSyntax)
   public init(_ lexeme: Lexer.Lexeme) {
-    if lexeme.rawTokenKind == .keyword {
-      self.init(Keyword(lexeme.tokenText)!)
+    if let kw = Keyword(lexeme.tokenText) {
+      self.init(kw)
```
ahoppen, line 105: This is not correct because it looses the distinction of e.g. a token that has text `open` e.g. in `self.open` and is an identifier and `open` that is considered a keyword e.g. `open func foo() {}`.

Also, performance-wise, iterating over all different keyword text choices is a lot more expensive than doing a scalar type comparison for the lexeme's `rawTokenKind`.

Sources/SwiftParser/Declarations.swift
```diff
@@ -1371,7 +1371,7 @@ extension Parser {
     var effectSpecifiers = self.parseDeclEffectSpecifiers()
 
     let output: RawReturnClauseSyntax?
-    if self.at(.arrow) {
+    if let _ = self.canRecoverTo(TokenSpec(.arrow)) {
```
ahoppen, line 1374: The general style in this code base is the following if we aren't using the result of `canRecoverTo`
```suggestion
    if self.canRecoverTo(TokenSpec(.arrow)) != nil {
```

same for `{`

---

Also, can you make sure that the arrow is on the same line the current token by passing `allowAtStartOfLine: false` to the `TokenSpec`? I think it's not as much of an issue here but I've commented on the `{` why this check is important for correct parsing.

Sources/SwiftParser/TopLevel.swift
```diff
@@ -98,7 +98,7 @@ extension Parser {
   /// This function is used when parsing places where function bodies are
   /// optional - like the function requirements in protocol declarations.
   mutating func parseOptionalCodeBlock(allowInitDecl: Bool = true) -> RawCodeBlockSyntax? {
-    guard self.at(.leftBrace) else {
+    guard let _ = self.canRecoverTo(TokenSpec(.leftBrace)) else {
```
ahoppen, line 101: I can't give you an exact counter-example but I'm a little hesitant to allow recovering to a left brace on an arbitrary line because that could cause us to skip over arbitrary identifiers to reach a `{` that is really completely unrelated to the optional code block we want to parse. E.g. if you have

```swift
var x: Int
somethingThatTakesATrailingClosure {}
```

we don't want to consume `somethingThatTakesATrailingClosure` as unexpected to recover to the `{`. I haven't checked this exact example but I hope the idea comes across. I would feel a lot more comfortable if we only allowed recovery to `{` on the same line because that way the `{` can't belong to a different statement since statements on the same line need to be separated by `;` and `;` has a higher precedence than `{`.

I.e. I'd suggest to change this to something like the following

```suggestion
    guard self.at(.leftBrace) || self.canRecoverTo(TokenSpec(.leftBrace, allowAtStartOfLine: false)) != nil else {
```

same above when looking for `->`

ahoppen (commented):
Thank you for your contribution @StevenWong12. Your changes look good to me, I've got a few small comments to make sure we don't accidentally parse valid code as invalid with the new recovery.

StevenWong12:
Thanks for your detailed suggestions! @ahoppen  Although the code snippet above was correctly parsed, I think it is indeed a little risky to recover to a left brace on an arbitrary line so I will make changes as your suggestions about checking ``->`` and ``{``.

As for the modification about ``init`` function of ``TokenPrecedence``, my original intention was to solve the problems when parsing codes containing multi-line macro declaration with function like signatures e.g.
```
macro m2(_: Int) = A.M2
macro m3(a b: Int) -> Int = A.M3
```

Since the marco signature parsing method reuses the logic of function signature parsing, the tokens from ``=`` to ``)`` in the second line will be recovered to ``->`` . Specifically, the second ``macro`` cannot be recognized as a keyword because it is considered contextual and will be initialized with ``exprKeyword `` precedence.

About this issue, I came across two solutions:
1. Reduce the recovery precedence of ``->`` to ``exprKeyword`` inside function signature parsing;
2. Add a case for ``RawTokenKind`` to specify this kind of "contextual keyword" and assign to them when initializing lexeme.

Solution 1 is pretty straightforward but, from my perspective, not pretty fit the context, since the precedence of ``macro`` is supposed to be ``declKeyword``
Solution 2 may bring extra wordload but we can explicitly assign a token precedence to these " contextual keywords" in ``TokenPrecedence.swift``

For now, I cannot decide which solution to implement yet, if there is anything I missed or misunderstood, please let me know.

Thanks in advance!

StevenWong12:
@swift-ci Please test

ahoppen:
I just debugged this myself and the problem seems to be that `canRecoveryTo` is not respecting `allowAtStartOfLine` when skipping over tokens. The following changes should fix that an remove the need to change the recovery precedence of the arrow. I think one other test case needs to be updated but that’s fine.


```diff
diff --git a/Sources/SwiftParser/Declarations.swift b/Sources/SwiftParser/Declarations.swift
index b3368e3c..0d56ef09 100644
--- a/Sources/SwiftParser/Declarations.swift
+++ b/Sources/SwiftParser/Declarations.swift
@@ -1376,7 +1376,7 @@ extension Parser {
 
     let output: RawReturnClauseSyntax?
 
-    if self.at(.arrow) || self.canRecoverTo(TokenSpec(.arrow, recoveryPrecedence: .weakPunctuator, allowAtStartOfLine: false)) != nil {
+    if self.at(.arrow) || self.canRecoverTo(TokenSpec(.arrow, allowAtStartOfLine: false)) != nil {
       output = self.parseFunctionReturnClause(effectSpecifiers: &effectSpecifiers, allowNamedOpaqueResultType: true)
     } else {
       output = nil
diff --git a/Sources/SwiftParser/Recovery.swift b/Sources/SwiftParser/Recovery.swift
index 21b4f7f7..4e48f45d 100644
--- a/Sources/SwiftParser/Recovery.swift
+++ b/Sources/SwiftParser/Recovery.swift
@@ -70,9 +70,10 @@ extension Parser.Lookahead {
     let initialTokensConsumed = self.tokensConsumed
 
     let recoveryPrecedence = min(spec1.recoveryPrecedence, spec2.recoveryPrecedence, spec3.recoveryPrecedence)
+    let shouldSkipOverNewlines = recoveryPrecedence.shouldSkipOverNewlines && spec1.allowAtStartOfLine && spec2.allowAtStartOfLine && spec3.allowAtStartOfLine
 
     while !self.at(.eof) {
-      if !recoveryPrecedence.shouldSkipOverNewlines, self.currentToken.isAtStartOfLine {
+      if !shouldSkipOverNewlines, self.currentToken.isAtStartOfLine {
         break
       }
       let matchedSpec: TokenSpec?

```

StevenWong12:
Cool! Thanks for your help @ahoppen ! The changes have been committed.😀

ahoppen (approved):
Thank you for your contribution @StevenWong12!

Issue #1373, opened 2023-03-01
Improve diagnostics marked with TODO in `Tests/SwiftParser/translated`


ahoppen:
There are ~150 TODOs left in the test cases where diagnostics from the C++ parser are superior to the ones generated by SwiftParser. Those should be fixed.

Some of these TODOs are marked as “good first issue” and fixing those should be a good starting point to get started with the swift-syntax codebase.

To make sure we don’t have two people working on the same TODO, I suggest that anyone who starts working on a TODO posts the file name of that TODO in this issue to let others know.

ahoppen:
rdar://106083939

bbvch13531:
Hello! I'm currently working on [testAvailabilityQuery34](https://github.com/apple/swift-syntax/blob/540f0086be3111f45bd1c2d91a6cbc79eadda7fe/Tests/SwiftParserTest/translated/AvailabilityQueryTests.swift#L403) in `AvailabilityQueryTests.swift`

Could I create a new issue and if there are few questions about this TODO? @ahoppen

kimdv:
Hi! 

I'm looking into [`testIdentifiers7`](https://github.com/apple/swift-syntax/blob/main/Tests/SwiftParserTest/translated/IdentifiersTests.swift#L82)

ahoppen:
> Hey @ahoppen, what are the additional codebase setup requirements apart from the `swift-compiler` setup which I already have, so I can start contributing to this repo too!

SwiftSyntax is (mostly) a standalone repo for which you don't need to build the Swift compiler yourself. You should be able to just open it in Xcode or in any other editor that supports LSP with SourceKit-LSP such as VSCode with the [Swift Extension](https://github.com/swift-server/vscode-swift).

ahoppen:
To anyone who starts working on an issue, if you've gotten so far that you can open a PR (even if there are still things that don't work) we can use that to discuss further questions. If you need help before you've made any significant code changes, I agree that opening a new issues is a good way to keep the discussions focussed.

CippoX:
Hi everyone, I'm working on [`testRecovery64`](https://github.com/apple/swift-syntax/blob/07c08da983b2815a0acc8e3a477573e4c663b158/Tests/SwiftParserTest/translated/RecoveryTests.swift#L824)

ahoppen:
@CippoX, @kimdv is already working on `testRecovery64` here https://github.com/apple/swift-syntax/pull/1382

CippoX:
[...]

PR #1382, opened 2023-03-02
Add diagnostic for unexpected second identifier

kimdv:
Part of fixing an TODOs in here: https://github.com/apple/swift-syntax/issues/1373

I just wanna ask if I'm on the right path.

### DIFF
--- a/Sources/SwiftParser/Patterns.swift
+++ b/Sources/SwiftParser/Patterns.swift
@@ -217,9 +217,22 @@ extension Parser {
       while !self.at(.eof, .rightParen) && keepGoing && loopProgress.evaluate(currentToken) {
         // If the tuple element has a label, parse it.
         let labelAndColon = self.consume(if: .identifier, followedBy: .colon)
-        let (label, colon) = (labelAndColon?.0, labelAndColon?.1)
+        var (label, colon) = (labelAndColon?.0, labelAndColon?.1)
+
+        /// If we have something like `x SomeType`, use the indication that `SomeType` starts with a capital letter (and is thus probably a type name)
+        /// as an indication that the user forgot to write the colon instead of forgetting to write the comma to separate two elements.
+        if label == nil, colon == nil, peek().rawTokenKind == .identifier, peek().tokenText.isStartingWithUppercase {
+          label = consumeAnyToken()
+          colon = self.missingToken(.colon)
+        }
+
         let pattern = self.parsePattern()
-        let trailingComma = self.consume(if: .comma)
+        var trailingComma = self.consume(if: .comma)
+
+        if trailingComma == nil && self.at(TokenSpec(.identifier, allowAtStartOfLine: false)) {
+          trailingComma = self.missingToken(RawTokenKind.comma)
+        }
+
         keepGoing = trailingComma != nil
         elements.append(
           RawTuplePatternElementSyntax(
--- a/Sources/SwiftParser/SyntaxUtils.swift
+++ b/Sources/SwiftParser/SyntaxUtils.swift
@@ -88,6 +88,14 @@ extension SyntaxText {
   var isEditorPlaceholder: Bool {
     return self.starts(with: SyntaxText("<#")) && self.hasSuffix(SyntaxText("#>"))
   }
+
+  var isStartingWithUppercase: Bool {
+    if !self.isEmpty, let firstCharacterByte = self.baseAddress?.pointee {
+      return 65 <= firstCharacterByte && firstCharacterByte <= 90
+    } else {
+      return false
+    }
+  }
 }
 
 extension RawTokenKind {
--- a/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
+++ b/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
@@ -327,7 +327,7 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
         FixIt(
           message: .joinIdentifiers,
           changes: [
-            FixIt.MultiNodeChange(.replace(oldNode: Syntax(previousToken), newNode: Syntax(TokenSyntax(.identifier(joined), presence: .present)))),
+            FixIt.MultiNodeChange(.replace(oldNode: Syntax(previousToken), newNode: Syntax(TokenSyntax(.identifier(joined), trailingTrivia: tokens.last?.trailingTrivia ?? [], presence: .present)))),
             .makeMissing(tokens),
           ]
         )
@@ -338,7 +338,7 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
           FixIt(
             message: .joinIdentifiersWithCamelCase,
             changes: [
-              FixIt.MultiNodeChange(.replace(oldNode: Syntax(previousToken), newNode: Syntax(TokenSyntax(.identifier(joinedUsingCamelCase), presence: .present)))),
+              FixIt.MultiNodeChange(.replace(oldNode: Syntax(previousToken), newNode: Syntax(TokenSyntax(.identifier(joinedUsingCamelCase), trailingTrivia: tokens.last?.trailingTrivia ?? [], presence: .present)))),
               .makeMissing(tokens),
             ]
           )
--- a/Tests/SwiftParserTest/TypeTests.swift
+++ b/Tests/SwiftParserTest/TypeTests.swift
@@ -15,15 +15,17 @@
 import XCTest
 
 final class TypeTests: XCTestCase {
-
   func testMissingColonInType() {
     assertParse(
       """
       var foo 1️⃣Bar = 1
       """,
       diagnostics: [
-        DiagnosticSpec(message: "expected ':' in type annotation")
-      ]
+        DiagnosticSpec(message: "expected ':' in type annotation", fixIts: ["insert ':'"])
+      ],
+      fixedSource: """
+        var foo: Bar = 1
+        """
     )
   }
 
@@ -65,7 +67,6 @@ final class TypeTests: XCTestCase {
   }
 
   func testClosureSignatures() {
-
     assertParse(
       """
       simple { [] str in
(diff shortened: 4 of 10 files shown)


Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -810,6 +810,19 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
   }
 
+    public override func visit(_ node: PatternBindingSyntax) -> SyntaxVisitorContinueKind {
+        if shouldSkip(node) {
+            return .skipChildren
+        }
+
+        if let unexpected = (node.pattern.as(TuplePatternSyntax.self)?.unexpectedBetweenElementsAndRightParen),
+            let previousToken = node.previousToken(viewMode: .sourceAccurate) {
+            addDiagnostic(unexpected, SpaceSeparatedIdentifiersError(firstToken: previousToken, additionalTokens: []))
```
kimdv, line 820: This will not use `constant declaration` but variable. 
We could add `nameForDiagnostics` where it's possible to add a string a string. 

And we could also add `helperText` where it's we can add `"is there an accidental break?"`

Or should we add `is there an accidental break?` for all `SpaceSeparatedIdentifiersError`

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -810,6 +810,24 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
     return .visitChildren
   }
 
+    public override func visit(_ node: PatternBindingSyntax) -> SyntaxVisitorContinueKind {
+        if shouldSkip(node) {
+            return .skipChildren
+        }
+
+        if let unexpected = (node.pattern.as(TuplePatternSyntax.self)?.unexpectedBetweenElementsAndRightParen),
+            let previousToken = node.previousToken(viewMode: .sourceAccurate) {
```
ahoppen, line 819: I think you should check that `unexpected` only consists of a single tokens, which is an identifier. Otherwise we also emit this diagnostic if we skipped over something completely unrelated. Same below.

Also, we should add the Fix-Its to join the identifiers together like we do in the other places where we emit this diagnostic.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -878,6 +878,84 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
+                FixIt(
+                  message: .joinIdentifiersWithCamelCase,
+                  changes: [
+                    [.replace(oldNode: Syntax(previousToken), newNode: Syntax(TokenSyntax(.identifier(joinedUsingCamelCase), presence: .present)))],
+                    .makeMissing(tokens),
+                  ]
+                )
+              )
+            }
+          }
```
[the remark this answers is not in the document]
kimdv: What is the common pattern? Have multiple fix its or just a single? 

Because if you type `fooBar` like in `TypeTests.testMissingColonInType` then a wrong space is as common error as a missing `:` IMO.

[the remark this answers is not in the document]
ahoppen: That’s a very good question. Our infrastructure does support multiple Fix-Its, as with the joining of identifiers but the tricky thing is that we need to pick one “primary” fix of how we want to represent the incorrect code in the fixed up syntax tree. So, in most cases I would expect there to be a single Fix-it.

An idea I have discussed with @flashspys in the context of https://github.com/apple/swift-syntax/pull/886 (but IIRC there’s no PR for that) is that we could check if the next identifier starts with a lowercase or uppercase letter. If it starts with a lowercase letter, then it’s likely a variable name and you want to insert a comma and if it starts with an uppercase letter it’s likely a type name and you’re likely missing a colon.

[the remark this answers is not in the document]
kimdv: Added your suggestion. It sounds reasonable for me, that we suggest adding a comma for `foo bar` and removing the space for `foo Bar`

Tests/SwiftParserTest/translated/RecoveryTests.swift
```diff
@@ -767,19 +775,36 @@ final class RecoveryTests: XCTestCase {
...
+      #line: (
+        "join the identifiers together",
+        """
+        enum EEEE<T> where T : Multi {
+          case a a
+          case b
+        }
+        """
+      )
+    ]
```
ahoppen, line 788: Is there any benefit of the `testCases` array here if it only contains a single entry?

Tests/SwiftParserTest/translated/RecoveryTests.swift
```diff
@@ -809,9 +844,12 @@ final class RecoveryTests: XCTestCase {
       let (efg 1️⃣hij, foobar) = (5, 6)
```
ahoppen, line 844: Maybe I wasn’t clear in my previous comment. What I thought was that the parser (not `ParseDiagnosticGenerator`) will synthesize a missing comma for this case, ie. parsing its s `(efg, his, foobar)`

If `hij` had a capital `H`, i.e. if the test case was `(efg Hij, foobar)`  the user was more likely to have forgotten the colon because `Hij` is by most conventions a type since it starts with a capital letter, so the parser should synthesize a missing colon and parse it as `(efg: Hij, foobar)`. That way I think we don’t actually need any special logic in `ParseDiagnosticGenerator` (or at least not much).

Sources/SwiftParser/SyntaxUtils.swift
```diff
@@ -88,6 +88,10 @@ extension SyntaxText {
   var isEditorPlaceholder: Bool {
     return self.starts(with: SyntaxText("<#")) && self.hasSuffix(SyntaxText("#>"))
   }
+
+  var isStartingWithSuppercase: Bool {
+    return description.first?.isUppercase ?? false
```
ahoppen, line 93: I don’t think we need to render the entire buffer into a `Swift.String` to decide whether the first character is uppercase. You should be able to do the following based on the ASCII value of the first character.

```swift
if let firstCharacterByte = buffer.first {
  return 65 <= firstCharacterByte && firstCharacterByte <= 90
}
```

---

Also there’s a typo. But I like `Suppercase` 🍲

Sources/SwiftParser/Patterns.swift
```diff
@@ -217,9 +217,20 @@ extension Parser {
       while !self.at(.eof, .rightParen) && keepGoing && loopProgress.evaluate(currentToken) {
         // If the tuple element has a label, parse it.
         let labelAndColon = self.consume(if: .identifier, followedBy: .colon)
-        let (label, colon) = (labelAndColon?.0, labelAndColon?.1)
+        var (label, colon) = (labelAndColon?.0, labelAndColon?.1)
+
+        if colon == nil, peek().rawTokenKind == .identifier, peek().tokenText.isStartingWithSuppercase {
+          label = consumeAnyToken()
```
ahoppen, line 223: What happens if `label != nil` here. Would we be dropping the existing label in that case?

Tests/SwiftParserTest/TypeTests.swift
```diff
@@ -15,16 +15,30 @@
...
+        """
+        var foo 1️⃣Bar = 1
+        """,
+        diagnostics: [
+          DiagnosticSpec(message: "expected ':' in type annotation", fixIts: ["insert ':'"], line: line),
+          DiagnosticSpec(
+            message: "found an unexpected second identifier in variable; is there an accidental break?",
+            fixIts: ["join the identifiers together"],
+            line: line
+          ),
```
ahoppen, line 35: I just wouldn’t bother generating this diagnostic. I think `expected ':' in type annotation` is a perfectly adequate diagnostic for the test case. IIUC that would allow you to remove most of the code you wrote in ParseDiagnsoticGenerator.swift

Sources/SwiftParser/SyntaxUtils.swift
kimdv: It could also be moved, but I'm not sure if that's correct tough. It doesn't quite fit I think.

Tests/SwiftParserTest/TypeTests.swift
kimdv: I'm not sure I understand this. If I remove the code in `ParseDiagnsoticGenerator.swift`, it will fail as we are not generating the expected diagnostics.

Sources/SwiftParser/Patterns.swift
kimdv: Yes it will. Fixed it

Sources/SwiftParser/Patterns.swift
```diff
@@ -217,9 +217,20 @@ extension Parser {
       while !self.at(.eof, .rightParen) && keepGoing && loopProgress.evaluate(currentToken) {
         // If the tuple element has a label, parse it.
         let labelAndColon = self.consume(if: .identifier, followedBy: .colon)
-        let (label, colon) = (labelAndColon?.0, labelAndColon?.1)
+        var (label, colon) = (labelAndColon?.0, labelAndColon?.1)
+
+        if label == nil, colon == nil, peek().rawTokenKind == .identifier, peek().tokenText.isStartingWithUppercase {
+          label = consumeAnyToken()
```
ahoppen, line 223: I think a comment explaining why we check for `isStartingWithUppercase` would be good here. Something like 

```swift
/// If we have something like `x SomeType`, use the indication that `SomeType` starts with a capital letter (and is thus probably a type name) as an indication that the user forgot to write the colon instead of forgetting to write the comma to separate two elements.
```

ahoppen (approved):
Thank you. Looks good to me. Just one request for a comment

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test windows

kimdv:
@swift-ci please test windows

kimdv:
@swift-ci please test windows

kimdv:
It looks like windows is failing, can't figure out why?
Can you help @compnerd ?

kimdv:
@swift-ci please test windows

PR #1383, opened 2023-03-02
Improved diagnostics for misplaced AttributeList in variable declaration

flashspys:
Previously the error messages for this code were quite bad. Now they are great!

```swift
struct A {
  var @State name: String
}
```

Part of this PR are some minor changes:
- The fixit message is shown in the diagnostics formatter
- The SyntaxRewriter now can use a given arena 
- Do not print colors if run within Xcode (debatable, feel free to remove)

Best regards from [Cocoaheads](https://cocoaheads.de) 😉

### DIFF
--- a/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
+++ b/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
@@ -262,7 +262,7 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
       unexpectedTokenCondition: { EffectSpecifier(token: $0) != nil },
       correctTokens: [effectSpecifiers?.asyncSpecifier, effectSpecifiers?.throwsClause?.throwsSpecifier],
       message: { EffectsSpecifierAfterArrow(effectsSpecifiersAfterArrow: $0) },
-      moveFixIt: { MoveTokensInFrontOfFixIt(movedTokens: $0, inFrontOf: .arrow) },
+      moveFixIt: { MoveNodesInFrontOfFixIt(movedNodes: $0, inFrontOf: .arrow) },
       removeRedundantFixIt: { RemoveRedundantFixIt(removeTokens: $0) }
     )
   }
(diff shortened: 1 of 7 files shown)


CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxRewriterFile.swift
```diff
@@ -30,7 +30,19 @@ let syntaxRewriterFile = SourceFileSyntax(leadingTrivia: generateCopyrightHeader
...
-    DeclSyntax("public init() {}")
+    DeclSyntax(
+      """
+      let arena: SyntaxArena?
+      """
+    )
+
+    DeclSyntax(
+      """
+      public init(arena: SyntaxArena? = nil) {
```
ahoppen, line 41: Actually, I think it would be better if all nodes created by a `SyntaxRewriter` live in the same `SyntaxArena`. Creating a `SyntaxArena` is fairly costly because it creates a new bump pointer allocator and it’s a waste to create a new one for every node. So, what I’m saying is that `arena` can be non-optional and this could be

```suggestion
      public init(arena: SyntaxArena = SyntaxArena()) {
```

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -1102,6 +1102,29 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
       moveFixIt: { MoveTokensAfterFixIt(movedTokens: $0, after: .equal) },
       removeRedundantFixIt: { RemoveRedundantFixIt(removeTokens: $0) }
     )
+
+    if let attributes = node.attributes,
+      attributes.isMissingAllTokens,
+      let unexpected = node.unexpectedBetweenBindingKeywordAndBindings
```
ahoppen, line 1108: I think we should also check that `unexpected` only contains a `AttributeListSyntax`. Otherwise we might emit this diagnostic if we just recover to the bindings and stuff any other node into `unexpectedBetweenBindingKeywordAndBindings`. I.e. I’d add

```suggestion
      let unexpected = node.unexpectedBetweenBindingKeywordAndBindings,
      unexpected.only?.is(AttributeListSyntax.self) ?? false
```

ahoppen (commented):
Thank you, Felix. Just a few minor comments but overall this looks very good to me.

Und viele Grüße aus Cupertino an CocoaHeads Aachen 🎉

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxRewriterFile.swift
flashspys: I tried that, but unfortunately the tests are failing. The `MacroApplication` SyntaxRewriter crashes internally.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test and merge

ahoppen:
@swift-ci Please test

flashspys:
The SwiftSyntaxBuilder tests were failing, I fixed the part I could. They are still failing on my machine with some really weird error messages, but maybe this is already known?

ahoppen:
@swift-ci Please test

ahoppen:
Picking this up. Rebased on `main` and removed the changes to not print ASCII colors in Xcode terminal as well as displaying a hammer to indicate the presence of a Fix-It.

ahoppen:
@swift-ci Please test

bnbarham (approved):

ahoppen:
@swift-ci Please test Windows

PR #1384, opened 2023-03-02
[Variadic Generics] type parameter pack syntax change: T... -> each T

sophiapoirier:


sophiapoirier:
apple/swift#64039

@swift-ci please test

sophiapoirier:
https://github.com/apple/swift/pull/64039

@swift-ci please test

hborla (approved):

Sources/SwiftParser/Declarations.swift
```diff
@@ -469,14 +469,14 @@ extension Parser {
       repeat {
         let attributes = self.parseAttributeList()
 
-        let (unexpectedBeforeName, name) = self.expectIdentifier()
-        if attributes == nil && unexpectedBeforeName == nil && name.isMissing && elements.isEmpty {
+        // Parse the 'each' keyword for a type parameter pack 'each T'.
+        let each = self.consume(if: .keyword(.each))
+
+        let (unexpectedBetweenEachAndName, name) = self.expectIdentifier()
+        if attributes == nil && unexpectedBetweenEachAndName == nil && name.isMissing && elements.isEmpty {
```
ahoppen, line 476: You need to check that `each` is `nil` here as well, otherwise we could drop `each` and not round-trip. A test case that currently fails should be if you open a generic argument list, write `each` and then reach the end of the file, i.e. 

```swift
func foo<each
```

sophiapoirier: `each` is supposed to be optional/nullable because the absence of `each` indicates a standard generic type parameter and the presence of `each` indicates a variadic type parameter pack. This was the case before my PR as well except prior to my syntax change, it was postfix `ellipsis` that made the distinction rather than the preceding `each`. Admittedly I do not know what is meant by "we could drop `each` and not round-trip"?

slavapestov: swift-syntax preserves each token in the source file in the parse tree as it was written, even if you hav something invalid like `each` not followed by a type parameter.

sophiapoirier: Thank you @ahoppen , I have addressed this in my follow-up PR https://github.com/apple/swift-syntax/pull/1393

PR #1386, opened 2023-03-05
Add @retroactive attribute

harlanhaskins:
This adds a new node and re-generates Swift code for the @retroactive attribute (added in https://github.com/apple/swift/pull/36068)

### DIFF
--- a/CodeGeneration/Sources/SyntaxSupport/AttributeKinds.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/AttributeKinds.swift
@@ -75,6 +75,7 @@ public let TYPE_ATTR_KINDS = [
   TypeAttribute(name: "noDerivative"),
   TypeAttribute(name: "async"),
   TypeAttribute(name: "Sendable"),
+  TypeAttribute(name: "retroactive"),
   TypeAttribute(name: "unchecked"),
   TypeAttribute(name: "_local"),
   TypeAttribute(name: "_noMetadata"),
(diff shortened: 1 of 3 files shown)


DougGregor:
You shouldn't need to do this at all. It'll be handled via the normal custom-attribute mechanism, and the compiler can recognize it as special.

harlanhaskins:
Oh good! Will I still need to teach ASTGen to lower this to TAK_retroactive or will that Just Happen?

DougGregor:
> Oh good! Will I still need to teach ASTGen to lower this to TAK_retroactive or will that Just Happen?

We haven't gotten to mapping over attributes in ASTGen, so you'll just need to handle this in the C++ parser implementation.

harlanhaskins:
Gotcha -- well, this comes in through the same parse step that @unchecked does so this parses fine in the C++ parser

ahoppen:
You do actually need to add it to disambiguate how parameters are being parsed. If there is no special handling and you have `@retroactive () -> Void`, the `()` will be parsed as arguments to `@retroactive` and then we don’t know what to do with the `-> Void`. Known compiler attributes don’t parse arguments so we won’t parse `()` as arguments.

DougGregor:
> You do actually need to add it to disambiguate how parameters are being parsed. If there is no special handling and you have `@retroactive () -> Void`, the `()` will be parsed as arguments to `@retroactive` and then we don’t know what to do with the `-> Void`.

I consider this a feature. It means new attributes can be handled correctly by older parsers, even if they don't know what they are. The set of known compiler attributes that don't parse the following `()` should be dictated by history and not expanded going forward.

PR #1387, opened 2023-03-05
[Macros] Specify introduced names for example member macro.

hborla:
This is required in https://github.com/apple/swift/pull/64097


hborla:
@swift-ci please test

PR #1394, opened 2023-03-09
Generate the SwiftSyntax docc index using CodeGeneration

ahoppen:
It turns out we need this index because docc generates the left sidebar that contains from it.


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

CodaFi (approved):

CodeGeneration/Sources/generate-swiftsyntax/GenerateSwiftSyntax.swift
```diff
@@ -55,49 +72,51 @@ struct GenerateSwiftSyntax: ParsableCommand {
...
+        GeneratedFileSpec([swiftSyntaxBuilderDir, "generated", "BuildableNodes.swift"], buildableNodesFile),
+        GeneratedFileSpec([swiftSyntaxBuilderDir, "generated", "ResultBuilders.swift"], resultBuildersFile),
+        GeneratedFileSpec([swiftSyntaxBuilderDir, "generated", "SyntaxExpressibleByStringInterpolationConformances.swift"], syntaxExpressibleByStringInterpolationConformancesFile),
       ]
       + BASE_KIND_FILES.map { baseKind in
-        TemplateSpec(sourceFile: syntaxNode(emitKind: baseKind.key), module: swiftSyntaxDir, filename: "syntaxNodes/\(baseKind.value)")
-      }
+        GeneratedFileSpec([swiftSyntaxDir, "generated", "syntaxNodes", baseKind.value], syntaxNode(emitKind: baseKind.key))
+      } + [
+        GeneratedFileSpec([swiftSyntaxDir, "Documentation.docc", "generated", "SwiftSyntax.md"], swiftSyntaxDoccIndex)
```
kimdv, line 116: Nit pick and don't need to be part of this PR, personally I would have a `private let generatedPath: String = "generated"` next to  `swiftSyntaxDir`, `swiftSyntaxBuilderDir` and to so on.

kimdv (approved):

kimdv:
And does DocC not support this kind of thing out of the box?

CodeGeneration/Sources/generate-swiftsyntax/GenerateSwiftSyntax.swift
ahoppen: Good suggestion. I updated it.

ahoppen:
> And does DocC not support this kind of thing out of the box?

I’m not sure but in either case it’s good if we can sort the types into groups ourselves rather than listing them all together, which is, what I suspect docc would do by default.

ahoppen:
@swift-ci Please test

kimdv:
@swift-ci please test macOS

ahoppen:
@swift-ci Please test macOS

PR #1397, opened 2023-03-09
Improve diagnostics message for init declaration

StevenWong12:
Fixed part of TODOs in #1373 

Also fixed a TODO in ``InitDeinitTests.swift``

### DIFF
--- a/Sources/SwiftParser/Declarations.swift
+++ b/Sources/SwiftParser/Declarations.swift
@@ -1059,15 +1059,7 @@ extension Parser {
     }
 
     // Parse the signature.
-    let signature = self.parseFunctionSignature()
-
-    // mark the unexpected result type of initializer
-    // and replace the ‘output‘ of signature with nil
-    var unexpectedResultType: RawUnexpectedNodesSyntax?
-    if let output = signature.output {
-      signature.raw = signature.layoutView.replacingChild(at: 5, with: nil, arena: self.arena)
-      unexpectedResultType = RawUnexpectedNodesSyntax([output.raw], arena: self.arena)
-    }
+    let signature = self.parseFunctionSignature(allowOutput: false)
 
     let whereClause: RawGenericWhereClauseSyntax?
     if self.at(.keyword(.where)) {
@@ -1086,7 +1078,6 @@ extension Parser {
       optionalMark: failable,
       genericParameterClause: generics,
       signature: signature,
-      unexpectedResultType,
       genericWhereClause: whereClause,
       body: items,
       arena: self.arena
@@ -1398,12 +1389,12 @@ extension Parser {
   }
 
   @_spi(RawSyntax)
-  public mutating func parseFunctionSignature() -> RawFunctionSignatureSyntax {
+  public mutating func parseFunctionSignature(allowOutput: Bool = true) -> RawFunctionSignatureSyntax {
     let input = self.parseParameterClause(for: .functionParameters)
 
     var effectSpecifiers = self.parseDeclEffectSpecifiers()
 
-    let output: RawReturnClauseSyntax?
+    var output: RawReturnClauseSyntax?
 
     /// Only allow recovery to the arrow with exprKeyword precedence so we only
     /// skip over misplaced identifiers and don't e.g. recover to an arrow in a 'where' clause.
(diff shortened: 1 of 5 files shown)


Sources/SwiftParser/Declarations.swift
```diff
@@ -1041,6 +1041,14 @@ extension Parser {
     // Parse the signature.
     let signature = self.parseFunctionSignature()
 
+    // mark the unexpected result type of initializer
+    // and replace the ‘output‘ of signature with nil
+    var unexpectedResultType: RawUnexpectedNodesSyntax?
+    if let output = signature.output {
+      signature.raw = signature.layoutView.replacingChild(at: 5, with: nil, arena: self.arena)
```
ahoppen, line 1048: Replacing a child by its index is super fragile because the index might change if a new child is added to `RawFunctionSignatureSyntax`. What I would prefer is to add a parameter to `parseFunctionSignature` like `allowOutput` that, if set to `false` just doesn’t parse an output type in the function signature.

Tests/SwiftParserTest/translated/RecoveryTests.swift
```diff
@@ -1751,9 +1753,11 @@ final class RecoveryTests: XCTestCase {
       init? 1️⃣c(_ d: Int) {}
       """,
       diagnostics: [
-        // TODO: (good first issue) Old parser expected error on line 1: initializers cannot have a name, Fix-It replacements: 9 - 10 = ''
-        DiagnosticSpec(message: "unexpected code 'c' before parameter clause")
-      ]
+        DiagnosticSpec(message: "initializers cannot have a name", fixIts: ["remove 'c'"])
+      ],
+      fixedSource: """
+        init? (_ d: Int) {}
```
ahoppen, line 1759: Ideally, we would also remove the space here so that we get `init?(_ d: Int)`. I think there is a `FixIt.Change` to remove trailing trivia from a token and you could apply that to the previous token of the removed identifier (`previousToken(viewMode: .sourceAccurate)`).

ahoppen (commented):
Thank you @StevenWong12, this PR is going and the right direction and the improvements in the tests are looking good. I’ve put a few comments inline.

Sources/SwiftParser/Declarations.swift
StevenWong12: Hey @ahoppen , I've just made a change about this.

Instead of just not parsing the output type, I make the output nil and create an unexpected node in ``parseFunctionSignature`` (not using ``replacingChild``), which is controlled by the flag ``allowOutput``.

If I leave the output type unparsed, there will be some corner cases, e.g. 
```
init(_ foo: T) -> Int where T: Comparable {}
```

in which the 'where' clause will be left unparsed and recognized as a part of the unexpected node, which is not.  I've also added this snippet as a new testcase.

BTW, if I promote the recovery precedence of ``where`` and try to recover to it will bring other problems.

I wonder whether this change is reasonable? Thanks in advance!

Sources/SwiftParser/Declarations.swift
```diff
@@ -1369,23 +1369,32 @@ extension Parser {
...
     } else {
       output = nil
     }
 
+    var unexpectedAfterOutput: RawUnexpectedNodesSyntax?
+    if !allowOutput,
+      let unexpectedOutput = output
+    {
+      output = nil
+      unexpectedAfterOutput = RawUnexpectedNodesSyntax([unexpectedOutput.raw], arena: self.arena)
```
ahoppen, line 1390: I don’t think you need the `.raw`. If it doesn’t compile without, could you share the error message you are getting?

ahoppen (approved):
Nice. Thank you. This looks good 👍

ahoppen:
@swift-ci Please test

ahoppen:
@StevenWong12 I approved but when I hit approve I forgot that I still had an open comment. If you could address https://github.com/apple/swift-syntax/pull/1397#discussion_r1132737221, I’ll merge the PR.

StevenWong12:
Hey @ahoppen, I think the .raw is redundant. Already removed it😀.

ahoppen:
Thank you.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

PR #1399, opened 2023-03-09
Reduce recovery precedence of arrow in function type to exprKeyword

ahoppen:
Only allow recovery to the arrow with exprKeyword precedence so we only skip over misplaced identifiers and don't e.g. recover to an errow in a 'where' clause.

CC @StevenWong12 I didn’t think about this case when reviewing your PR.

rdar://106457182

### DIFF
--- a/Sources/SwiftParser/Declarations.swift
+++ b/Sources/SwiftParser/Declarations.swift
@@ -1376,7 +1376,9 @@ extension Parser {
 
     let output: RawReturnClauseSyntax?
 
-    if self.at(.arrow) || self.canRecoverTo(TokenSpec(.arrow, allowAtStartOfLine: false)) != nil {
+    /// Only allow recovery to the arrow with exprKeyword precedence so we only
+    /// skip over misplaced identifiers and don't e.g. recover to an arrow in a 'where' clause.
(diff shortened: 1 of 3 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

bnbarham (approved):

Sources/SwiftParser/Declarations.swift
```diff
@@ -1376,7 +1376,9 @@ extension Parser {
 
     let output: RawReturnClauseSyntax?
 
-    if self.at(.arrow) || self.canRecoverTo(TokenSpec(.arrow, allowAtStartOfLine: false)) != nil {
+    /// Only allow recovery to the arrow with exprKeyword precedence so we only
+    /// skip over misplaced identifiers and don't e.g. recover to an errow in a 'where' clause.
```
kimdv, line 1380: Is there a typo in `errow` ?

ahoppen: You are right, thank you.

ahoppen:
@swift-ci Please test

PR #1401, opened 2023-03-10
Use CRLF in diagnostics on Windows

etcwilde:
Windows has formatting issues in the new diagnostics because we're using the wrong raw newlines and they need to be carriage-return/line-feeds. I added an internal static member on `String`, which is kind of gross, for the platform newline string. Then went through and replaced all of the `\n` characters that I saw getting appended to string in the SwiftDiagnostics directory.

```swift
      // If the line did not end with \n (e.g. the last line), append it manually
      if annotatedSource.last != "\n" {
        annotatedSource.append(String.newline)
      }
```

I left it in the condition because the last character will be `\n` in both cases if there is a newline, and I wanted to avoid even more `#if os` checks. We can change that if needed/wanted, of course.

This is more of a nerd-snipe PR than a "I'm totally ready to go" PR and I'm fully expecting comments to improve and swift-if the code.

### DIFF
--- a/Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
+++ b/Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
@@ -49,6 +49,14 @@ extension Sequence where Element == Range<Int> {
   }
 }
 
+extension String {
+#if os(Windows)
+    internal static let newline = "\r\n"
+#else
+    internal static let newline = "\n"
+#endif
+}
+
 public struct DiagnosticsFormatter {
 
   /// A wrapper struct for a source line, its diagnostics, and any
@@ -227,7 +235,7 @@ public struct DiagnosticsFormatter {
     // If there was a filename, add it first.
     if let fileName = fileName {
       let header = colorizeBufferOutline("===")
-      annotatedSource.append("\(indentString)\(header) \(fileName) \(header)\n")
+      annotatedSource.append("\(indentString)\(header) \(fileName) \(header)\(String.newline)")
     }
 
     /// Keep track if a line missing char should be printed
@@ -254,7 +262,7 @@ public struct DiagnosticsFormatter {
       // If necessary, print a line that indicates that there was lines skipped in the source code
       if hasLineBeenSkipped && !annotatedSource.isEmpty {
         let lineMissingInfoLine = indentString + String(repeating: " ", count: maxNumberOfDigits) + " \(colorizeBufferOutline("┆"))"
-        annotatedSource.append("\(lineMissingInfoLine)\n")
+        annotatedSource.append("\(lineMissingInfoLine)\(String.newline)")
       }
       hasLineBeenSkipped = false
 
@@ -274,7 +282,7 @@ public struct DiagnosticsFormatter {
 
       // If the line did not end with \n (e.g. the last line), append it manually
       if annotatedSource.last != "\n" {
-        annotatedSource.append("\n")
+        annotatedSource.append(String.newline)
       }
 
       let columnsWithDiagnostics = Set(annotatedLine.diagnostics.map { $0.location(converter: slc).column ?? 0 })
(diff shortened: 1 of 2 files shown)


etcwilde:
@swift-ci please test windows

etcwilde:
@swift-ci please test

DougGregor (approved):
This looks fine to me, but I'm quite afraid that it's going to be easy to forget and screw it up again.

etcwilde:
> This looks fine to me, but I'm quite afraid that it's going to be easy to forget and screw it up again.

It might be good to go through the diagnostic string at the point of printing and “fixup” new lines (of either persuasion) to be whatever they should be for the platform.

compnerd:
Thanks for the patch @etcwilde!  Unfortunately, from some quick local testing, it appears that it does not resolve the issue:

```
==== S:/SourceCache/swift-project/swift/stdlib/public/Concurrency/SourceCompatibilityShims.swift:57:2 ====

 56 | @_alwaysEmitIntoClient

 57 | @available(*, deprecated, renamed: "withTaskCancellationHandler(operation:onCancel:)")

    |  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

 58 | public func withTaskCancellationHandler<T>(


```

etcwilde:
> Unfortunately, from some quick local testing, it appears that it does not resolve the issue:

Fascinating. It's happening on the header lines. From chatting with @compnerd, the 4 equals signs are not actually there and were some artifact of copy-pasting. From a screenshot, there are only three, as expected. So we have:

```
=== S:/SourceCache/swift-project/swift/stdlib/public/Concurrency/SourceCompatibilityShims.swift:57:2 ===

 56 | @_alwaysEmitIntoClient
```

```swift
    // If there was a filename, add it first.
    if let fileName = fileName {
      let header = colorizeBufferOutline("===")
      annotatedSource.append("\(indentString)\(header) \(fileName) \(header)\(String.newline)")
    }
```

From this, we know that the newline isn't happening between `header` and the `fileName`, therefore the final `header` before the newline shouldn't be the cause of the issue. The second newline has to be getting inserted after the explicit one on that line.

The next place where annotatedSource is touched is down here;

```swift
      // If necessary, print a line that indicates that there was lines skipped in the source code
      if hasLineBeenSkipped && !annotatedSource.isEmpty {
        let lineMissingInfoLine = indentString + String(repeating: " ", count: maxNumberOfDigits) + " \(colorizeBufferOutline("┆"))"
        annotatedSource.append("\(lineMissingInfoLine)\(String.newline)")
      }
      hasLineBeenSkipped = false

      // add indentation
      annotatedSource.append(indentString)

      // print the source line
      annotatedSource.append(linePrefix)
```

So the first string added to the `annotatedSource` after the header will be `indentString`, which is a parameter to the function. If that is prefaced with a newline, we'd see an extra newline. It's also the leader for the line-number lines, which could be inserting the newline there? Will need to hop on Windows to see what's happening there.

compnerd:
image

This is without the patch ... I think that @bnbarham has a thought that is worth investigating - this could be the C++ driver.

bnbarham:
> This is without the patch ... I think that @bnbarham has a thought that is worth investigating - this could be the C++ driver.

Yeah, does Swift handle printing \n properly on Windows? If so, maybe it's fine for swift-syntax to be giving back \n and we should handle that back in C++ land?

etcwilde:
> EDIT: Actually, it just occurred to me that we don't build earlyswiftsyntax on Windows yet. Which means SWIFT_SWIFT_PARSER isn't set and we're not even using the swift-syntax implementation. So... maybe this one is actually fine? Seems to be output fine from the tests at least.

Yeah, I'm starting to think that you're right here. I could see the extra newlines coming from somewhere in all of the `printNewline`s in `lib/swift/Frontend/PrintingDiagnosticConsumer.cpp`.

Issue #1385, opened 2023-03-03
Swift error reporting style removed complete `path:line:col` information


ktoso:
### Description

The new error formatting does the following:

```
=== /Users/ktoso/code/swift-project/swift/stdlib/public/core/SIMDVector.swift ===
     ┆
  61 │
  62 │ /// A type that can be used as an element in a SIMD vector.
  63 │ public protocol SIMDScalar {
     ∣                 ├─ warning: protocol 'SIMDScalar' should be declared to refine 'Decodable' due to a same-type constraint on 'Self'
     ∣                 ├─ warning: protocol 'SIMDScalar' should be declared to refine 'Encodable' due to a same-type constraint on 'Self'
     ∣                 ╰─ warning: protocol 'SIMDScalar' should be declared to refine 'Hashable' due to a same-type constraint on 'Self'
```

which is missing a crucial feature of the previous format, including the:

```
/Users/ktoso/code/swift-project/swift/stdlib/public/core/SIMDVector.swift:61:16
```

information as one string where the error was located.

This is important because terminals and IDEs are able to understand this format and e.g. in iterm2 you can click such "path:line:col" to open up your editor at the exact line where the error was simplifying editing by a lot when working from the terminal (which always do for example).

Including this in the "header" where the fullpath is would be good enough I think 👍 

Thanks in advance for bringing back this info the new formatter is quite pleasant, looking forward to it!

ahoppen:
Hmm, I’m not sure if the header is the right place to put it because a source file rendered by this diagnostic formatter could show multiple errors at different lines. But I don’t know where else we could put it either.

ktoso:
wouldn't one expect to have one error per one listing tbh?

But even if multiple perhaps at least offer the first one as that'd jump to the "right location" mostly?

ahoppen:
rdar://106434704

PR #1402, opened 2023-03-10
[Diagnostic formatter] Add ":<line>" to filename header.

DougGregor:
When printing the filename header for diagnostics, also include the line at which the displqy starts using the standard `:<line number>` format, which is recognized by various tools to quickly go to that source file and line.

Fixes https://github.com/apple/swift-syntax/issues/1385.


DougGregor:
@swift-ci please test

ktoso (approved):
Sweet, thanks!

ktoso:
Thinking about this some more i think we can do even better, though this already is an improvement so let's merge :) 
I'll follow up with some ideas soon 👍

PR #1403, opened 2023-03-10
[CompilerPlugin] Factor out message handling logic to a module

rintaro:
Create a module `_SwiftCompilerPluginImpl` that abstracts the low level message connection/serialization.

`CompilerPlugin` is now an implementation of the message connection, and adapts the concrete type to the message handler.

This make it possible to make a plugin with a different messaging connection and serialization.


rintaro:
@swift-ci Please smoke test

rintaro:
@swift-ci Please test

bnbarham (approved):
Not a huge fan of `Impl`. `CompilerPluginAPI` or something?

Package.swift
```diff
@@ -43,6 +43,7 @@ let package = Package(
     .library(name: "SwiftSyntaxParser", type: .static, targets: ["SwiftSyntaxParser"]),
     .library(name: "SwiftSyntaxBuilder", type: .static, targets: ["SwiftSyntaxBuilder"]),
     .library(name: "SwiftSyntaxMacros", type: .static, targets: ["SwiftSyntaxMacros"]),
+    .library(name: "_SwiftCompilerPluginBase", type: .static, targets: ["_SwiftCompilerPluginImpl"]),
```
ahoppen, line 46: The library is named differently than the target. Is that intentional?

ahoppen (commented):
What do you think about naming the module `SwiftCompilerPluginMessageHandling` or something like that. I think that’s more descriptive than `Impl`. Also, I’m not sure whether we should underscore it because it seems equally useful to users as the other modules in swift-syntax.

rintaro:
Thank for naming idea. Let me go with `SwiftCompilerPluginMessageHandling`

PR #1405, opened 2023-03-12
[NFC] Rename `TokenConsumptionHandle.missing` to `.tokenIsMissing`

WowbaggersLiquidLunch:
This clarifies what is missing at the use site. Also according the the API design guidelines:

> **Uses of Boolean methods and properties should read as assertions about the receiver** when the use is nonmutating, e.g. `x.isEmpty`, `line1.intersects(line2)`.

### DIFF
--- a/Sources/SwiftParser/Declarations.swift
+++ b/Sources/SwiftParser/Declarations.swift
@@ -1658,7 +1658,7 @@ extension Parser {
 
   /// Parse an accessor.
   mutating func parseAccessorDecl() -> RawAccessorDeclSyntax {
-    let forcedHandle = TokenConsumptionHandle(spec: .keyword(.get), missing: true)
+    let forcedHandle = TokenConsumptionHandle(spec: .keyword(.get), tokenIsMissing: true)
     let introducer = parseAccessorIntroducer(forcedKind: (.get, forcedHandle))!
     return parseAccessorDecl(introducer: introducer)
   }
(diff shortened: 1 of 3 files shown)


WowbaggersLiquidLunch:
@swift-ci please test

WowbaggersLiquidLunch:
@swift-ci please test

WowbaggersLiquidLunch:
@swift-ci please test

ahoppen (commented):
Renaming `missing` -> `isMissing` makes sense to me. I’m not sure if we need to include `token` in there because at least to me it’s obvious that we’re talking about a token since only tokens can be missing in SwiftSyntax and this is a *Token*ConsumptionHandle. What did you think might have been missing? If it was ambiguous to you, I’m happy to change my mind.

WowbaggersLiquidLunch:
> this is a _Token_ConsumptionHandle

When I was first reading the source, all I saw was that all the handles were named just `handle`s. It wasn't obvious to me that they were instances of _Token_ConsumptionHandle. I only figured out that it's the token that's missing when I jumped to definition and read the documentation. I wasn't able to deduce it without reading the documentation, because I didn't know what the handles were for to begin with. So it would be confusing and doesn't seem obvious to me (and especially to new contributors I guess) that the receiver/subject of `isMissing` is the token not the handle.

ahoppen (approved):
Ah, with that explanation the rename makes sense to me. Thanks.

PR #1406, opened 2023-03-13
[NFC] Simplify some value bindings

WowbaggersLiquidLunch:
This makes the bindings a little easier to read imo.

Also this might pave a little way for implementing [the pitched `is case`](https://forums.swift.org/t/proposal-draft-for-is-case-pattern-match-boolean-expressions/58260), if we decide to ban

```swift
foo is case is Bar
```

for being too silly.

### DIFF
--- a/Sources/SwiftParser/Patterns.swift
+++ b/Sources/SwiftParser/Patterns.swift
@@ -242,28 +242,29 @@ extension Parser {
   mutating func parseMatchingPattern(context: PatternContext) -> RawPatternSyntax {
     // Parse productions that can only be patterns.
     switch self.at(anyIn: MatchingPatternStart.self) {
-    case (.varKeyword, let handle)?,
-      (.letKeyword, let handle)?,
-      (.inoutKeyword, let handle)?:
-      let bindingKeyword = self.eat(handle)
-      let value = self.parseMatchingPattern(context: .bindingIntroducer)
-      return RawPatternSyntax(
-        RawValueBindingPatternSyntax(
-          bindingKeyword: bindingKeyword,
-          valuePattern: value,
-          arena: self.arena
+    case let .some((patternStart, handle)):
+      switch patternStart {
+      case .varKeyword, .letKeyword, .inoutKeyword:
+        let bindingKeyword = self.eat(handle)
+        let value = self.parseMatchingPattern(context: .bindingIntroducer)
+        return RawPatternSyntax(
+          RawValueBindingPatternSyntax(
+            bindingKeyword: bindingKeyword,
+            valuePattern: value,
+            arena: self.arena
+          )
         )
-      )
-    case (.isKeyword, let handle)?:
-      let isKeyword = self.eat(handle)
-      let type = self.parseType()
-      return RawPatternSyntax(
-        RawIsTypePatternSyntax(
-          isKeyword: isKeyword,
-          type: type,
-          arena: self.arena
+      case .isKeyword:
+        let isKeyword = self.eat(handle)
+        let type = self.parseType()
+        return RawPatternSyntax(
+          RawIsTypePatternSyntax(
+            isKeyword: isKeyword,
+            type: type,
+            arena: self.arena
+          )
         )
-      )
+      }
     case nil:
       // matching-pattern ::= expr
       // Fall back to expression parsing for ambiguous forms. Name lookup will


WowbaggersLiquidLunch:
@swift-ci please test

ahoppen (commented):
I understand that code style is always a little subjective but I disagree with you here. I think at the old version is more readable because it has one fewer nesting level. Which part of the old implementation did you find hard to read? Maybe we can find some style that works for both of us.

WowbaggersLiquidLunch:
> Which part of the old implementation did you find hard to read?

I suppose it wasn't hard to read. It just felt a little weird to me that `handle` is bound to 3 times in the case label. Trading one more nesting level for a "single-source-of-truth" binding to `handle` and fewer `?` feels better to me. This also puts the 3 cases on the same line without anything else in between, so it's clear (to me) from a single glance what the case matches. It is indeed subjective which side of the trade-off one prefers. 

> Maybe we can find some style that works for both of us.

What if the 3 patterns are indented to align vertically? That probably would improve the "from a single glance" aspect imo.

---

Another reason for this PR is that I wanted to land in some small preparations for implementing `is case`, so only the most relevant changes would be parked in a PR pending for evolution approval. I'm partially considering banning (and sort of expecting during the review that some might want to ban) `foo is case is Bar` in the implementation. If we do ban it, then it might be better to diagnose it like this:

```swift
switch self.at(...) {
case .some(patternStart, handle):
  if context == .matching {
    addDiagnostics(...)
  }
  switch patternStart {...}
case .nil:
  ...
}
``` 

than this:

```swift
switch self.at(...) {
case .varKeyword, .letKeyword, .inoutKeyword:
  if context == .matching {
    addDiagnostics(...)
  }
  ...
case .isKeyword:
  if context == .matching {
    addDiagnostics(...)
  }
  ...
case .nil:
  ...
}
```

But I might be jumping way too ahead here.

ahoppen:
> I suppose it wasn't hard to read. It just felt a little weird to me that `handle` is bound to 3 times in the case label. Trading one more nesting level for a "single-source-of-truth" binding to `handle` and fewer `?` feels better to me. This also puts the 3 cases on the same line without anything else in between, so it's clear (to me) from a single glance what the case matches. It is indeed subjective which side of the trade-off one prefers.

I don’t really have a problem with binding `handle` three times because the compiler will complain if you forget to bind it in one of the `case`s.

> What if the 3 patterns are indented to align vertically? That probably would improve the "from a single glance" aspect imo.

I agree that that would help. I’m wondering whether we should fix that in `swift-format`. CC @allevato 


> Another reason for this PR is that I wanted to land in some small preparations for implementing `is case`, so only the most relevant changes would be parked in a PR pending for evolution approval. I'm partially considering banning (and sort of expecting during the review that some might want to ban) `foo is case is Bar` in the implementation. If we do ban it, then it might be better to diagnose it like this:

I don’t fully understand what you’re trying to say here. Also where are these examples coming from?

allevato:
> I agree that that would help. I’m wondering whether we should fix that in `swift-format`. CC @allevato

`case` clauses in general have been tricky to format well since they terminate with a `:` instead of a braced structure, but I suppose that's not the problem here. I guess there are two options to have the desired effect:

1. Force a break immediately after `case`:
   ```
    case
      (.varKeyword, let handle)?,
      (.letKeyword, let handle)?,
      (.inoutKeyword, let handle)?:
    ```

2. Use a keyword-length-based indentation:
   ```
    case (.varKeyword, let handle)?,
         (.letKeyword, let handle)?,
         (.inoutKeyword, let handle)?:
    ```

We've resisted #2 in the past in favor of having a uniform indentation step throughout. This isn't really much different than the issue you have with `if`/`guard` statements that have multiple conditions:

```
if someLongCondition,
  someOtherConditionThatDoesntQuiteLineUp
{
```

where breaking after the `if` feels particularly egregious because it's just two characters. But for `case` and `guard`, it feels slightly more reasonable.

The least invasive change to make would be to change [this](https://github.com/apple/swift-format/blob/f2b3e6c0c9b2ae8629a2aba0224a698ecfa8c6a7/Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift#L734) `space` to `break`, which would allow you to insert a newline after `case` and the formatter wouldn't remove it as it does today. If someone wants to make that change, I'm happy to review it.

Longer-term, it might be worth revisiting the rigid uniform indentation step, but I don't want to do that quite yet since it would have much broader implications to how the formatting algorithm works.

WowbaggersLiquidLunch:
> I don’t fully understand what you’re trying to say here. Also where are these examples coming from?

I think I can explain my rationale better when I have the PR for `is case` ready (probably in a few hours or a longest maybe 2 days).

Though I admit that the more I think about it, the less merit I see in pushing for this PR.

ahoppen:
OK, let’s just put this PR on hold then for now and re-evaluate it when you’ve got your other PR up.

WowbaggersLiquidLunch:
It turned out I misunderstood how diagnostics is produced, so my second reason for this PR doesn't stand. I don't really want to change the code for minor aesthetic reasons alone, so I'm closing this PR. Thanks for the review!

ahoppen:
Aesthetic changes are always welcome, even if they are minor, so don't feel like you shouldn't create PRs like this in the future. We just need to agree that the change is for the better 😉

PR #1413, opened 2023-03-16
[CMake] Fix detecting -disable-implicit-string-processing-module-import

rintaro:
The check command used to always fail because it fails to load stdlib. Workaround it by adding '-parse-stdlib' to the check command.


rintaro:
@swift-ci Please test

ahoppen (approved):
I trust you if you say that this fixes things.

PR #1414, opened 2023-03-17
[DNM][SE-NNNN] Parse `is case`

WowbaggersLiquidLunch:
A small piece of the puzzle.

This PR allows SwiftSyntax to parse `is case` pattern patching expressions, [as pitched on the forums](https://forums.swift.org/t/proposal-draft-for-is-case-pattern-match-boolean-expressions/58260).

2 questions for reviewers:

1. Currently as implemented, this PR allows `is case` to be used with value-binding like this:
   
   ```swift
   foo is case let .bar(baz)
   ```
   
   My intention is to allow value-bindings like this only inside conditions:
   
   ```swift
   guard foo is case let .bar(baz) // good, allowed
   
   let a = b is case let c         // bad, diallowed
   ```
   
   Is it possible to diagnose this on the syntactic level in this package? My understanding so far is that this can only be done during semantic analysis.
   
2. Is this parser used along with the old one written in C++ in the compiler, or is this the only one used? Do I need to re-implement the same logic in the C++ version too?


Sources/SwiftParser/Expressions.swift
```diff
@@ -248,6 +249,8 @@ extension Parser {
   ///     type-casting-operator → 'as' type
   ///     type-casting-operator → 'as' '?' type
   ///     type-casting-operator → 'as' '!' type
+  ///     avo-pattern-matching-expression → avo-pattern-matching-operator pattern
+  ///     avo-pattern-matching-operator  →  'is' 'case'
```
WowbaggersLiquidLunch, line 253: (A call for suggestions for better names, with a detour into some linguistic stuff)

With this new syntax included, we have 3 pattern matching constructs in Swift:

- `~=`
- `case`
- `is case`

They all have the same capabilities, with their only difference being the "word" order:

```swift
pattern ~= value
case pattern = value
value is case pattern
```

So it seems to me the names of the operators and expressions has to reflect this difference in order to be distinct from each other. However, the dominant order alone isn't enough to differentiate between them, because both `pattern ~= value` and `value is case pattern` have the subject-verb-object structure (if we take both operators `~=` and `is case` as a form of the English word "match").

In the comments here and that in `CodeGeneration/Sources/SyntaxSupport/ExprNodes.swift`, I'm assuming the matching is always done by the value, making it the agent. And by this logic, I have tentatively named the `is case` operator as the "agent-verb-object pattern matching operator", and the corresponding expression as "agent-verb-object pattern matching expression. This also makes `~=` "subject-verb-agent", and `case` "verb-object-agent".

However, this assumption is incorrect, because the value doesn't initiate/perform the action, so it isn't the agent.

<details>

<summary>Even longer linguistic detours (sorry)</summary>

We can first observe that all 4 sentences below are perfectly natural English sentences:

> The value matches the pattern.
> The pattern matches the value.
> The pattern is matched by the value.
> The value is matched by the pattern.

These 4 sentences make it as if both "value" and "pattern" can be the agent. However, if we use "match" as a ditransitive verb, we can see that neither the value nor the pattern is the agent.

> The computer/runtime matches the value to the pattern.

Instead, they're both themes.

</details>

Anyway, it would be great if anyone could suggest some better names for the operators and expressions.

WowbaggersLiquidLunch:
@swift-ci please test

ahoppen:
> 1. Currently as implemented, this PR allows `is case` to be used with value-binding like this:
>    ```swift
>    foo is case let .bar(baz)
>    ```
>      
>    My intention is to allow value-bindings like this only inside conditions:
>    ```swift
>    guard foo is case let .bar(baz) // good, allowed
>    
>    let a = b is case let c         // bad, diallowed
>    ```
>    Is it possible to diagnose this on the syntactic level in this package? My understanding so far is that this can only be done during semantic analysis.

I haven’t read the entire proposal pitch but allowing value bindings in `is case` seems a bit like overkill to me since we can already do that using `if case .base(let bad) = foo`, so my impression from just skimming over Jordan’s pitch was to not allow value bindings in `is case` so that `is case` always just evaluates to a boolean.

> 2. Is this parser used along with the old one written in C++ in the compiler, or is this the only one used? Do I need to re-implement the same logic in the C++ version too?

The compiler still uses the C++ parser so you’ll need to implement the same in the C++ parser as well.

Sources/SwiftSyntax/generated/Keyword.swift
```diff
@@ -735,7 +735,7 @@ public enum Keyword: UInt8, Hashable {
     }
   }
   
-  /// Whether the token kind is switched from being an identifier to a keyword in the lexer.
+  /// Whether the token kind is switched from being an identifier to being an identifier to a keyword in the lexer.
```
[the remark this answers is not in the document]
ahoppen: This is generate from here

https://github.com/apple/swift-syntax/blob/808d4038fe1ef15ab2334f6095e7d8a55c36486c/CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/KeywordFile.swift#L66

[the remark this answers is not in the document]
WowbaggersLiquidLunch: Opened #1421 for this.

Sources/SwiftParser/Expressions.swift
WowbaggersLiquidLunch: renamed to "new pattern-matching [operator|expression]" for now in d1afd93.

WowbaggersLiquidLunch:
> I haven’t read the entire proposal pitch but allowing value bindings in `is case` seems a bit like overkill to me since we can already do that using `if case .base(let bad) = foo`, so my impression from just skimming over Jordan’s pitch was to not allow value bindings in `is case` so that `is case` always just evaluates to a boolean.

Yes, in Jordan's pitch the intention is not allowing value-binding in `is case`. But [some](https://forums.swift.org/t/proposal-draft-for-is-case-pattern-match-boolean-expressions/58260/15) people [in](https://forums.swift.org/t/proposal-draft-for-is-case-pattern-match-boolean-expressions/58260/16) the pitch thread pointed out (and mentioned [in the pitch itself](https://forums.swift.org/t/proposal-draft-for-is-case-pattern-match-boolean-expressions/58260/16), though in comparison to `~=` not `case let`) that \<value\> `is case let`/`var`/`inout` \<pattern\> is ergonomic, better for code completion, and potentially help with further refining matches. I agree with these points. So I'm thinking if it's not a lot extra work to support value-binding, I'll default to having it.

Sources/SwiftParser/Expressions.swift
WowbaggersLiquidLunch: After a discussion with Jordan, it now seems right to me to just call it "pattern-matching boolean expression", as it is in the pitch. Updated in 2bd28cc.

WowbaggersLiquidLunch:
@swift-ci please test

ahoppen:
> Yes, in Jordan's pitch the intention is not allowing value-binding in `is case`. But [some](https://forums.swift.org/t/proposal-draft-for-is-case-pattern-match-boolean-expressions/58260/15) people [in](https://forums.swift.org/t/proposal-draft-for-is-case-pattern-match-boolean-expressions/58260/16) the pitch thread pointed out (and mentioned [in the pitch itself](https://forums.swift.org/t/proposal-draft-for-is-case-pattern-match-boolean-expressions/58260/16), though in comparison to `~=` not `case let`) that <value> `is case let`/`var`/`inout` <pattern> is ergonomic, better for code completion, and potentially help with further refining matches. I agree with these points. So I'm thinking if it's not a lot extra work to support value-binding, I'll default to having it.

Thanks for the explanation and sorry for not reading the entire thread. I imagine this is something that needs to be discussed in the Swift Evolution proposal, so I won’t leave any comment on personal preference here.

PR #1415, opened 2023-03-17
[Doc] Add a note pointing to where the enum-case-pattern is parsed

WowbaggersLiquidLunch:
This helps prevent contributors from getting lost.

### DIFF
--- a/Sources/SwiftParser/Patterns.swift
+++ b/Sources/SwiftParser/Patterns.swift
@@ -44,6 +44,8 @@ extension Parser {
   ///     as-pattern → pattern 'as' type
   ///
   ///     expression-pattern → expression
+  ///
+  /// - Note: enum-case-pattern is parsed in ``parseGuardStatement(guardHandle:)``.
   @_spi(RawSyntax)
   public mutating func parsePattern() -> RawPatternSyntax {
     enum ExpectedTokens: TokenSpecSet {


WowbaggersLiquidLunch:
@swift-ci please smoke test

Sources/SwiftParser/Patterns.swift
```diff
@@ -44,6 +44,8 @@ extension Parser {
   ///     as-pattern → pattern 'as' type
   ///
   ///     expression-pattern → expression
+  ///
+  /// - Note: enum-case-pattern is parsed in ``parseGuardStatement(guardHandle:)``.
```
ahoppen, line 48: Is this true though? `parseGuardStatement` parses `guard whatever else {}` and not `enum-case-pattern`.

WowbaggersLiquidLunch: Is `parseGuardedCasePattern` the right one? I think I meant to write that, but wrote a wrong one. Sorry!

ahoppen: Actually, I don’t think there is a dedicated function that parses `enum-case-pattern`. It’s parsed in `parseSequenceExpression` if the `pattern` parameter is  `matching`.

These grammar definitions were mostly copied from the C++ parser and weren’t even up-to-date in that codebase. Deleting them would probably be the best thing to do as they are just misleading if they are inaccurate IMO.

ahoppen: I just opened https://github.com/apple/swift-syntax/pull/1420 to remove these `Grammar` comment sections altogether.

WowbaggersLiquidLunch: What you say makes some sense to me. Although, I don't know if deleting all of them is the best approach. Anyway, I think I'll close this PR for now, because it seems there needs to be a more general solution than leaving a note on a single production rule.

PR #1416, opened 2023-03-17
Output number of instructions executed during performance test in `swift-parser-cli performance-test`

ahoppen:
Instruction count is usually a far more stable value than time when it comes to performance measurements

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -48,6 +48,7 @@ let package = Package(
     .library(name: "SwiftRefactor", type: .static, targets: ["SwiftRefactor"]),
   ],
   targets: [
+    .target(name: "_InstructionCounter"),
     .target(
       name: "SwiftBasicFormat",
       dependencies: ["SwiftSyntax"],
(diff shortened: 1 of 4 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

kimdv:
@swift-ci please test Windows

Package.swift
```diff
@@ -48,6 +48,7 @@ let package = Package(
     .library(name: "SwiftRefactor", type: .static, targets: ["SwiftRefactor"]),
   ],
   targets: [
+    .target(name: "_InstructionCounter"),
```
kimdv, line 51: I'm a bit curious on why those files are in a new target and not just in the CLI target?

ahoppen: Because SwiftPM does not support mixed-language targets, so you can’t have Swift and C in the same target.

PR #1419, opened 2023-03-17
Don't allow single-line string literals to span multiple lines

hamishknight:
There were a few cases where we could parse tokens that had leading trivia, causing us to incorrectly accept some single-line string literals that spanned multiple lines. Fix up these cases to ensure the tokens we parse don't have leading trivia.

### DIFF
--- a/Sources/SwiftParser/StringLiterals.swift
+++ b/Sources/SwiftParser/StringLiterals.swift
@@ -490,7 +490,7 @@ extension Parser {
           unexpectedBeforeRightParen.append(self.consumeAnyToken())
         }
         let rightParen = self.expectWithoutRecovery(.rightParen)
-        if rightParen.isMissing, case .inStringInterpolation = self.currentToken.cursor.currentState {
+        if case .inStringInterpolation = self.currentToken.cursor.currentState {
           // The parser has more knowledge that we have reached the end of the
           // string interpolation now, even if we haven't seen the closing ')'.
           // For example, consider the following code
--- a/Sources/SwiftParser/StringLiterals.swift
+++ b/Sources/SwiftParser/StringLiterals.swift
@@ -475,11 +475,15 @@ extension Parser {
     var segments: [RawStringLiteralSegmentsSyntax.Element] = []
     var loopProgress = LoopProgressCondition()
     while loopProgress.evaluate(self.currentToken) {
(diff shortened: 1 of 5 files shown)


hamishknight:
@swift-ci please test

Sources/SwiftParser/StringLiterals.swift
```diff
@@ -475,11 +475,15 @@ extension Parser {
     var segments: [RawStringLiteralSegmentsSyntax.Element] = []
     var loopProgress = LoopProgressCondition()
     while loopProgress.evaluate(self.currentToken) {
+      // If we encounter a token with leading trivia, we're no longer in the
+      // string literal.
+      guard currentToken.leadingTriviaText.isEmpty else { break }
```
ahoppen, line 480: Technically the condition here is that the leading trivia contains a newline, right? You are just using the implication here that `leadingTrivia.isEmpty` is only `true` if it contains a newline. I’d prefer to check for a newline here, just to be safe should we change the trivia attribution rule. So, wouldn’t `!currentToken.isAtStartOfLine` be better?

ahoppen (commented):
Thank you 🙏🏽

Sources/SwiftParser/StringLiterals.swift
hamishknight: > Technically the condition here is that the leading trivia contains a newline, right?

Yes, but we shouldn't accept any other form of trivia either, as string literals can't contain trivia outside of interpolations. I'm happy to change this to just check for a newline, though I think in that case we should then assert that we don't have any other kind of trivia. Does that sound reasonable?

> just to be safe should we change the trivia attribution rule

Could you elaborate on what you mean by this?

ahoppen: > Yes, but we shouldn't accept any other form of trivia either, as string literals can't contain trivia outside of interpolations. I'm happy to change this to just check for a newline, though I think in that case we should then assert that we don't have any other kind of trivia. Does that sound reasonable?

Ah, yes that makes sense. Thanks for the explanation.

> Could you elaborate on what you mean by this?

Currently the lexer only puts newlines in leading trivia, never in trailing trivia, so non-empty leading trivia implies that there is a newline inside the trivia. But with your explanation above this doesn’t matter.

PR #1420, opened 2023-03-17
Remove `Grammar` doc comments

ahoppen:
These grammars were most of the time incomplete, incorrect and in the case of recovery wrong by definition. Instead of having misleading doc comments, let’s delete them entirely.

### DIFF
--- a/Sources/SwiftParser/Availability.swift
+++ b/Sources/SwiftParser/Availability.swift
@@ -14,11 +14,6 @@
 
 extension Parser {
   /// Parse a list of availability arguments.
-  ///
-  /// Grammar
-  /// =======
-  ///
-  ///     availability-arguments → availability-argument | availability-argument , availability-arguments
   mutating func parseAvailabilitySpecList() -> RawAvailabilitySpecListSyntax {
     var elements = [RawAvailabilityArgumentSyntax]()
     do {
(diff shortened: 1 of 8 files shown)


ahoppen:
@swift-ci Please test

WowbaggersLiquidLunch:
As incorrect as some of them are, I did find them quite useful. These comments helped me trace down what I needed (to use/change) for #1414. I'm not really opposed to this PR, but I do kind of worry that once they're deleted, the correct ones won't be added back. 

Also as far as I understand, it's sometimes necessary for the parser's logic to deviate from the grammar, in order to produce good diagnostics. I wonder if there is a good way to document both the correct grammar and the looser grammar the parser uses, and the relation between their production rules?

ahoppen:
Thinking more about this, the path forward here should be to generate a grammar for the Swift language from the syntax nodes because they will automatically be kept up-to-date if the layout of the syntax tree changes.

ahoppen:
@WowbaggersLiquidLunch I’m generating the grammar in https://github.com/apple/swift-syntax/pull/1771. Do you think that would be a sufficient replacement of the current comments?

WowbaggersLiquidLunch:
> @WowbaggersLiquidLunch I’m generating the grammar in https://github.com/apple/swift-syntax/pull/1771. Do you think that would be a sufficient replacement of the current comments?

I think #1771 is a good approach in general. I left some comments in the PR there. 

One of the comments I want to highlight is that the new _Grammar_ section uses a lot of the parser's type/protocol names instead of the actual grammar components' names. I don't know if it's possible to generate it, but it's be really helpful to have a translation table between the types/protocols and the syntactic categories.

ahoppen:
@swift-ci Please test

bnbarham (approved):

ahoppen:
@swift-ci Please test

PR #1423, opened 2023-03-19
fixed testAvailabilityQuery34 and testAvailabilityQueryUnavailability28

CippoX:
This is my first fix attempt for this repo. I hope I'm heading in the right direction.


Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -452,6 +452,18 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
         message: { _ in .joinConditionsUsingComma },
         moveFixIt: { ReplaceTokensFixIt(replaceTokens: $0, replacement: trailingComma) }
       )
+    } else if let unexpected = (node.condition.as(AvailabilityConditionSyntax.self)?.unexpectedBetweenAvailabilitySpecAndRightParen) {
```
kimdv, line 455: Not sure if you need the left and right paren around 
```suggestion
    } else if let unexpected = node.condition.as(AvailabilityConditionSyntax.self)?.unexpectedBetweenAvailabilitySpecAndRightParen {
```

CippoX: Ty

ahoppen (commented):
Thanks for the pull request. This is looking good already. I’d prefer to implement the recovery inside the parser instead of `ParseDiagnosticGenerator.swift` though because that allows us to form a SwiftSyntax tree that still represents the code the user intended to write. 

So, wherever we currently parse the `>=` as unexpected tokens, I would consume the `>=` as unexpected text and then continue parsing  `10.51` as a version.

Could you try that? If something doesn’t make sense to you, let me know.

CippoX:
Hope I understood the request

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -440,6 +440,23 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
     return .visitChildren
   }
 
+  public override func visit(_ node: AvailabilityVersionRestrictionSyntax) -> SyntaxVisitorContinueKind {
+    if shouldSkip(node) {
+      return .skipChildren
+    }
+    if let unexpected = node.unexpectedBetweenPlatformAndVersion {
```
ahoppen, line 447: `unexpected` could theoretically be anything if we changed the parser, so just to be sure, I’d suggest to check that `unexpected` really only contains a `>=` operator here:

```swift
unexpected.onlyToken(where: { $0.tokenKind == .binaryOperator(">=") }) != nil
```

ahoppen (commented):
One minor comment, otherwise this looks excellent to me. Thank you very much.

CippoX:
I had a little problem with source control 😅 now everything should be ok

ahoppen (approved):
Great. I’ve got one final request: Could you squash the commits in this PR. It just makes for a nicer Git history.

CippoX:
@ahoppen I'm sorry but looks like something went wrong, can you help me

PR #1424, opened 2023-03-19
NFC: Introduce Lexer.Cursor.Position

hamishknight:
This just holds the position info, which is useful for clients that don't care about the extra state in Cursor (e.g the state stack). LexingDiagnostic is one such client, and regex literal lexing will eventually be another.

### DIFF
--- a/Sources/SwiftParser/Lexer/Cursor.swift
+++ b/Sources/SwiftParser/Lexer/Cursor.swift
@@ -193,15 +193,15 @@ extension Lexer.Cursor {
   struct LexingDiagnostic {
     let kind: TokenDiagnostic.Kind
     /// The position in the token at which the diagnostic is.
-    let position: Lexer.Cursor
+    let position: Lexer.Cursor.Position
 
     init(_ kind: TokenDiagnostic.Kind, position: Lexer.Cursor) {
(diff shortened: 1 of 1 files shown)


hamishknight:
@swift-ci  please test

hamishknight:
@swift-ci please test

hamishknight:
@swift-ci please test macOS

hamishknight:
@ahoppen Sorry, I realize now the PR description is ambiguously worded, when I mentioned the StateStack, I'm referring to it as state that certain clients don't care about (and so should use `Cursor.Position` instead). AFAIK the StateStack does not currently store Cursors, so there shouldn't be any work to do there. That being said, one of the reasons I'm adding this is to avoid storing a whole `Cursor` as the endpoint for a regex literal lexeme, which _will_ need to be stored in the state stack (I'm planning on doing some benchmarks to see whether storing that state inline vs indirectly is better).

If you're interested in where I am currently with the regex work, I have a WIP branch https://github.com/hamishknight/swift-syntax/compare/lets-go-back-to-the-island

ahoppen:
Ah, I see. Makes sense 👍🏽

Issue #1165, opened 2022-12-21
Make CodeBlockItemListSyntax expressible by string literal


bradfol:
### Description

For the Syntax Builder it would be very convenient to be able to do something like the following to create a `CodeBlockItemList` from one multi-line string:

```
CodeBlockItemList("""
  let one = object.methodOne()
  let two = object.methodTwo()
""")
```

The `FunctionDecl` currently has behavior that allows for something similar, where it can parse the `CodeBlock` body of the function as well as the declaration:

```
FunctionDecl("""
func aFunction() {
  let one = object.methodOne()
  let two = object.methodTwo()
  // ...
}
""")
```
But I can't find a way to create a `CodeBlockItemList` from a multiline string otherwise.

ahoppen:
rdar://103618439

StevenWong12:
Hi @ahoppen, could you please assign this issue to me? I've already started to work with it recently.

StevenWong12:
Thank you @kimdv!

bradfol:
I did find a bit of a workaround that I've been using, which is to rely on the interpolation string conversion for `SourceFileSyntax`

```
let sourceFile: SourceFile = """
    let one = object.methodOne()
    """
```

ahoppen:
I should have probably commented on this issue at some point, sorry. 

The current idea is that you don't create a `CodeBlockItemList` from one big string literal but every item from its own string literal. You should be able to do the following

```swift
CodeBlockItemList {
  "let one = object.methodOne()"
  "let two = object.methodTwo()"
}
```

I'm happy to change that but it would be good to see a motivating example that wouldn't work with the result builder on `CodeBlockItemList` before we make another node expressible by string literals

bradfol:
> `CodeBlockItemList` from [...] every item from its own string literal

That would be great!

ahoppen:
Great. I just renamed the issue and all that would need to be done for this is to make `CodeBlockItemListSyntax` conform to `SyntaxExpressibleByStringInterpolation`.

ahoppen:
In the meantime what you can do is the following

```swift
CodeBlockItemListSyntax {
  DeclSyntax("let one = object.methodOne()")
  DeclSyntax("let two = object.methodTwo()")
}
```

PR #1426, opened 2023-03-21
Make CodeBlockItemListSyntax expressible by string literal

StevenWong12:
This PR is to enable creating a ``CodeBlockItemList`` from one multi-line string, which is mentioned in #1165.

BTW, I also noticed that there are two linked TODOs in ``ConvenienceInitializers.swift`` and ``SyntaxExpressibleByStringInterpolationConformances.swift``
```swift
// TODO: This should be inlined in SyntaxParseable.init(stringInterpolation:),
// but is currently used in `ConvenienceInitializers.swift`.
// See the corresponding TODO there.
func performParse<SyntaxType: SyntaxProtocol>(

// TODO: We should split FunctionParameter into separate nodes
extension FunctionParameterSyntax {
```

I'm a little confused about ``We should split FunctionParameter into separate nodes``. Some examples or elaborations would be of much help!


Tests/SwiftSyntaxBuilderTest/StringInterpolationTests.swift
```diff
@@ -460,4 +460,22 @@ final class StringInterpolationTests: XCTestCase {
...
+      """
+    XCTAssertTrue(codeBlockItemList.is(CodeBlockItemListSyntax.self))
+    XCTAssertEqual(
+      codeBlockItemList.description,
+      """
+      let one = object.methodOne()
+      let two = object.methodTwo()
+      func foo() {}
+      """
+    )
```
kimdv, line 479: We have an `Assert` that can be used instead. 
In it you can provide `codeBlockItemList` and a string with the expected string

kimdv: Sorry for not being clear enough. 
I was thing more like this `AssertBuildResult`
https://github.com/apple/swift-syntax/blob/main/Tests/SwiftSyntaxBuilderTest/ArrayExprTests.swift#L24

StevenWong12: Thanks for your comments! Since the ``StringInterpolationTests.swift`` is used for testing the *string interpolation*, I've add another class to test the build result in ``CodeBlockItemListTests.swift``. I wonder whether this is reasonable? @kimdv

kimdv: Both looks good for me. 😁

ahoppen:
As I mentioned in https://github.com/apple/swift-syntax/issues/1165#issuecomment-1471192005, I would prefer to make `CodeBlockItemSyntax` expressible by string literal instead of `CodeBlockItemListSyntax`. I think that generalizes better because you can then still use result builder constructs such as loops to create the final `CodeBlockItemListSyntax`.

Also, regarding the implementation, instead of adding `parserCodeBlockItemListForBuilder` you should make `CodeBlockItemSyntax` conform to `SyntaxParsable` in `Parser+Entry.swift`.

> I'm a little confused about `We should split FunctionParameter into separate nodes`. Some examples or elaborations would be of much help!

We currently use `FunctionParameterSyntax` for a variety of parameters like function parameters, enum parameters and closure parameters and they all have different needs. Closure parameters for example don’t require a type and enum parameters no parameter label. Thus all the children of `FunctionParameterSyntax` are optional. If we had a separate node for each of these uses we could get rid of `ParameterSubject` and make the children that are required for each of these parameter kinds non-optional.

StevenWong12:
Hi @ahoppen, thanks for your comments!

Sorry for misunderstanding your comments in #1165. I have updated the codes according to your suggestions. To prevent the undefined behavior of parsing a code block item, I added an assertion in the ``Parser+Entry.swift``. I wonder whether this is reasonable?

As for ``FunctionParameterSyntax``, I would like to make a try to separate it into different nodes. Should I create an issue for that to follow up?

Thanks in advance!

Sources/SwiftParser/generated/Parser+Entry.swift
```diff
@@ -65,6 +65,17 @@ extension CatchClauseSyntax: SyntaxParseable {
...
+  public static func parse(from parser: inout Parser) -> Self {
+    guard let node = parser.parseCodeBlockItem() else {
+      assertionFailure("Invalid code block item")
+      return Syntax(raw: RawSyntax(parser.parseCodeBlockItem()!)).cast(Self.self)
+    }
+    let raw = RawSyntax(parser.parseRemainder(into:node))
+    return Syntax(raw: raw).cast(Self.self)
+  }
+}
+
```
ahoppen, line 78: I think this will get overridden once you re-generate this file using the following command.
```bash
swift run --package-path CodeGeneration generate-swiftsyntax Sources
```

You will need to do this by adjusting `CodeGeneration/Sources/ParserEntryFile.swift`, probably by providing a `parserFunction` to `CodeBlockItem` in CommonNodes.swift

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftparser/ParserEntryFile.swift
```diff
@@ -59,17 +59,34 @@ let parserEntryFile = SourceFileSyntax(leadingTrivia: generateCopyrightHeader(fo
...
-            let node = parser.\(raw: parserFunction)()
-            let raw = RawSyntax(parser.parseRemainder(into: node))
-            return Syntax(raw: raw).cast(Self.self)
+      if node.syntaxKind == "CodeBlockItem" {
+        DeclSyntax(
+          """
+          extension \(raw: node.name): SyntaxParseable {
+            public static func parse(from parser: inout Parser) -> Self {
+              guard let node = parser.\(raw: parserFunction)() else {
+                assertionFailure("Invalid code block item")
```
ahoppen, line 68: We should not crash when parsing an invalid code block item. Instead, we should form an invalid syntax node. What you could do instead when `parseCodeBlockItem` returns `nil` is to create a `CodeBlockItemSyntax` that contains a `MissingExprSyntax` – or any of the other `Missing*Syntax` nodes, I don’t have a strong preference here.

---

Also, I would prefer not to add a special-case in this code generation loop. Instead, what I think would be the better solution is to have a function `parseNonOptionalCodeBlockItem`, which does what I described above, at the bottom of Parser+Entry.swift and then use that as the `parserFunction`.

StevenWong12: Awesome, thanks for your suggestions!

I've updated the codes and added corresponding test cases. If there is anything I can do better for this PR, please feel free to let me know.

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftparser/ParserEntryFile.swift
```diff
@@ -101,5 +101,23 @@ let parserEntryFile = SourceFileSyntax(leadingTrivia: copyrightHeader) {
...
+      mutating func parseNonOptionalCodeBlockItem() -> RawCodeBlockItemSyntax {
+        guard let node = self.parseCodeBlockItem() else {
+          // The missing item is not neccessary to be a declaration,
+          // which is just a placeholder here
+          let invalidCodeBlockItemNode = RawCodeBlockItemSyntax(
+            item: .decl(RawDeclSyntax(RawMissingDeclSyntax(attributes: nil, modifiers: nil,arena: self.arena))),
+            semicolon: nil,
+            arena: self.arena
+          )
+          return invalidCodeBlockItemNode
```
ahoppen, line 116: There’s no real need for the variable here

```suggestion
          return RawCodeBlockItemSyntax(
            item: .decl(RawDeclSyntax(RawMissingDeclSyntax(attributes: nil, modifiers: nil,arena: self.arena))),
            semicolon: nil,
            arena: self.arena
          )
```

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftparser/ParserEntryFile.swift
```diff
@@ -100,5 +100,22 @@ let parserEntryFile = SourceFileSyntax(leadingTrivia: copyrightHeader) {
...
     )
+
+    DeclSyntax(
+      """
+      mutating func parseNonOptionalCodeBlockItem() -> RawCodeBlockItemSyntax {
+        guard let node = self.parseCodeBlockItem(isAtTopLevel: false, allowInitDecl: true) else {
+          // The missing item is not neccessary to be a declaration,
+          // which is just a placeholder here
+          return RawCodeBlockItemSyntax(
+            item: .decl(RawDeclSyntax(RawMissingDeclSyntax(attributes: nil, modifiers: nil,arena: self.arena))),
```
ahoppen, line 111:
```suggestion
            item: .decl(RawDeclSyntax(RawMissingDeclSyntax(attributes: nil, modifiers: nil, arena: self.arena))),
```

ahoppen (approved):
I just noticed one missing space, otherwise LGTM.

ahoppen:
@swift-ci Please test macOS

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Tests/SwiftSyntaxBuilderTest/StringInterpolationTests.swift
```diff
@@ -408,6 +408,26 @@ final class StringInterpolationTests: XCTestCase {
...
+  func testCodeBlockItemInterpolation() {
+    let codeBlockItem: CodeBlockItemSyntax =
+      """
+      func foo() {
+        return bar
+      }
+      """
+
+    XCTAssertTrue(codeBlockItem.is(CodeBlockItemSyntax.self))
+    AssertStringsEqualWithDiff(
```
ahoppen, line 420: This function has been converted to lowercase. Could you update it and make sure that tests build and pass locally?

StevenWong12: Ohhh...I selected ``swift-parser-cli`` as the scheme and it just skipped ``SwiftSyntaxBuilderTest``. Sorry for that.

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test Windows

StevenWong12:
I think there is something wrong with the Windows CI. Could you please trigger that again? @kimdv

kimdv:
@swift-ci please test Windows

StevenWong12:
Thank you 😁

PR #1429, opened 2023-03-22
Add missing `switch` & for-each expression and `while` & `if` condition error

kimdv:


Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -1025,6 +1025,18 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
     return .visitChildren
   }
 
+    public override func visit(_ node: SwitchExprSyntax) -> SyntaxVisitorContinueKind {
+        if shouldSkip(node) {
+            return .skipChildren
+        }
+
+        if node.expression.is(MissingExprSyntax.self) && !node.cases.isEmpty {
+            addDiagnostic(node.expression, .missingExpressionInSwitchStatement, handledNodes: [node.expression.id])
```
kimdv, line 1034: Would it make sense to add a fix-it with an editor placeholder?

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -1025,6 +1025,22 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
+      return .skipChildren
+    }
+
+    if node.expression.is(MissingExprSyntax.self) && !node.cases.isEmpty {
+      addDiagnostic(node.expression, .missingExpressionInSwitchStatement, handledNodes: [node.expression.id])
+    }
+
+    if node.cases.isEmpty && !node.leftBrace.isMissingAllTokens && !node.rightBrace.isMissingAllTokens {
+      addDiagnostic(node.cases, .switchExprBodyEmpty, handledNodes: [node.cases.id])
+    }
```
kimdv, line 1039: Should here be a fixit here?
If yes, what could it be?

Tests/SwiftParserTest/translated/RecoveryTests.swift
```diff
@@ -621,82 +621,84 @@ final class RecoveryTests: XCTestCase {
...
     )
   }
 
   func testRecovery50() {
     AssertParse(
       """
-      switch { 42 } {
+      switch 1️⃣{ 42 } {
         case _: return
       }
```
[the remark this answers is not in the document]
ahoppen: I would expect this as anything else is really confusing to read. It also matches the current parser.

https://github.com/apple/swift/blob/e47f72c35208db108383ee14bf36791ac256df5a/lib/Parse/ParseStmt.cpp#L2404-L2407

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -1025,6 +1025,22 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
+      return .skipChildren
+    }
+
+    if node.expression.is(MissingExprSyntax.self) && !node.cases.isEmpty {
+      addDiagnostic(node.expression, .missingExpressionInSwitchStatement, handledNodes: [node.expression.id])
+    }
+
+    if node.cases.isEmpty && !node.leftBrace.isMissingAllTokens && !node.rightBrace.isMissingAllTokens {
+      addDiagnostic(node.cases, .emptySwitchExprBody, handledNodes: [node.cases.id])
+    }
```
ahoppen, line 1039: This only works if the `SwiftExprSyntax` has another error because otherwise the `hasError` bit is not set on the `SwitchExprSyntax` and thus we’re skipping the node. You can verify that we are not seeing this issue if the expression is not missing in the `switch`. So, we should consider this diagnostic semantic and it should be diagnosed at a later stage since it doesn’t prevent us from producing a well-formed SwiftSyntax tree.

That is, it’s OK if they are not emitted by SwiftParser even if they were by the old parser.

ahoppen (commented):
Thanks. If you want, you can do the same for `if` and `else if`

kimdv:
> Thanks. If you want, you can do the same for `if` and `else if`

Also added `for` and `while` loops

Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift
```diff
@@ -161,9 +167,18 @@ extension DiagnosticMessage where Self == StaticParserError {
...
+  }
+  public static var missingConditionInWhileStatement: Self {
+    .init("missing condition in 'while' statement")
+  }
   public static var missingConformanceRequirement: Self {
     .init("expected ':' or '==' to indicate a conformance or same-type requirement")
   }
+  public static var missingExpressionInSwitchStatement: Self {
+    .init("expected expression in 'switch' statement")
+  }
```
kimdv, line 181: Should we add a single error with this error message?

ahoppen: Yes, I think so

Sources/SwiftParser/Expressions.swift
```diff
@@ -2222,7 +2239,15 @@ extension Parser {
   ) -> RawSwitchExprSyntax {
     let (unexpectedBeforeSwitchKeyword, switchKeyword) = self.eat(switchHandle)
 
-    let subject = self.parseExpression(.basic)
+    // if there is no expression, like `switch { default: return false }` then left brace will be parsed as
+    // an `RawClosureExprSyntax`.
```
ahoppen, line 2243: Just an idea to make this clearer.

```suggestion
    // if there is no expression, like `switch { default: return false }` then left brace would parsed as
    // a `RawClosureExprSyntax` in the condition, which is most likely not what the user meant. 
    // Create a missing condition instead and use the `{` for the start of the body.
```

Same below

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
ahoppen: In that case we would still skip the parent of the `SwitchExprSyntax`. It’s just not something we can represent in `ParserDiagnosticGenerator` and it will need to be diagnostic at a later stage.

kimdv: Okay, will remove and add the todo again.
I still think it is something we should handle somewhere 😁

ahoppen: You can also remove the TODO. I have removed many of those that we consider semantic. Just missed this one.

Sources/SwiftParser/Statements.swift
```diff
@@ -616,7 +631,14 @@ extension Parser {
 
     let (unexpectedBeforeInKeyword, inKeyword) = self.expect(.keyword(.in))
 
-    let expr = self.parseExpression(.basic)
+    // if there is no expression, like `switch { default: return false }` then left brace will be parsed as
+    // an `RawClosureExprSyntax`.
```
ahoppen, line 635: Could you also update this comment with my suggestion.

Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift
```diff
@@ -330,6 +333,34 @@ public struct MissingAttributeArgument: ParserError {
...
+    }
+}
+
+public struct MissingExpressionInStatement: ParserError {
+    let firstToken: TokenSyntax
+
+    public var message: String {
+        if let name = firstToken.parent?.ancestorOrSelf(mapping: {
+            $0.nodeTypeNameForDiagnostics(allowBlockNames: false)
+        }) {
```
ahoppen, line 356: Can't you just pass in the node that contains the `nodeTypeNameForDiagnostics` here (i.e. `IfExpr`, `SwitchExpr` etc.) instead of having to rely on `ancestorOrSelf`?

Tests/SwiftParserTest/translated/RecoveryTests.swift
```diff
@@ -317,11 +311,11 @@ final class RecoveryTests: XCTestCase {
   func testRecovery24() {
     assertParse(
       """
-      while { true }() { //  expected-error 2 {{consecutive statements on a line must be separated by ';'}} expected-error {{closure expression is unused}} expected-note {{did you mean to use a 'do' statement?}} {{20-20=do }} expected-warning {{boolean literal is unused}}
+      while 1️⃣{ true }() { //  expected-error 2 {{consecutive statements on a line must be separated by ';'}} expected-error {{closure expression is unused}} expected-note {{did you mean to use a 'do' statement?}} {{20-20=do }} expected-warning {{boolean literal is unused}}
```
ahoppen, line 314: While you're editing this, would you mind also removing the 
```
//  expected-error 2 {{consecutive statements on a line must be separated by ';'}} expected-error {{closure expression is unused}} expected-note {{did you mean to use a 'do' statement?}} {{20-20=do }} expected-warning {{boolean literal is unused}}
```

Looks like I forgot that when auditing the translated test cases.

Tests/SwiftParserTest/translated/RecoveryTests.swift
```diff
@@ -613,90 +593,92 @@ final class RecoveryTests: XCTestCase {
       switch1️⃣
       """,
       diagnostics: [
-        DiagnosticSpec(message: "expected expression and '{}' to end 'switch' statement")
+        DiagnosticSpec(locationMarker: "1️⃣", message: "expected expression and '{}' to end 'switch' statement")
```
ahoppen, line 596: You don't need to specify a location Marker if there's only `1`.

Tests/SwiftParserTest/translated/RecoveryTests.swift
kimdv: Like this?

ahoppen (approved):

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test Windows

PR #1431, opened 2023-03-23
Tweak `.gitignore` to always ignore `.swiftpm`

hamishknight:
By default Xcode saves an xctestplan directly in the `.swiftpm` directory if you change any of its default values (e.g when enabling code coverage). As such, ignore the entire `.swiftpm` directly, not just `.swiftpm/xcode`. Also ignore it for any child directories, avoiding the need to have separate `.gitignore`s for child SwiftPM projects.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,8 +2,8 @@
 .build/
 # Ignore generated Xcode projects
 *.xcodeproj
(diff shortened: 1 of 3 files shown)


hamishknight:
@swift-ci please test

ahoppen (approved):

PR #1433, opened 2023-03-23
Enable two swift-format rules that are now satisfied

ahoppen:
I disabled all of those rules to make the introduction of swift-format easiser. Now I went through them again and these rules are satisfied. All the other rules that are disabled have good reasons for being diabled, so I want to keep them disabled.

rdar://103919185


ahoppen:
@swift-ci Please test

kimdv (approved):

PR #1339, opened 2023-02-12
Add a conditional compilation flag to verify token choices

ahoppen:
Verification of token choice is currently causing assertion failures where the list of expected token choices in the syntax tree doesn't match those the parser generates.

I’d still like to get the verification code into `main` so I can conditionally enable it locally while working on the issues, so add it behind a conditional compilation flag.

PR #1436, opened 2023-03-23
Fix multiple cases where the parser produced token kinds that didn’t match the token kinds in the syntax node definitions

ahoppen:
The main problem here was that we were accepting `self` and `Self` in `expectIdentifier` but not handling that in the syntax tree. I went through every call of `expectIdentifier` and checked what the C++ parser does to maintain the behavior. Where we need to accept `self` or `Self` for compatibility reasons, it is now remapped to an identifier.

In other cases, the syntax node definitions just needed to be updated.

These problems were found by the verification added in https://github.com/apple/swift-syntax/pull/1339.

### DIFF
--- a/Sources/SwiftParser/Types.swift
+++ b/Sources/SwiftParser/Types.swift
@@ -271,7 +271,7 @@ extension Parser {
             )
           )
         } else {
-          let (name, generics) = self.parseTypeNameWithGenerics(include: .keywords)
+          let (name, generics) = self.parseTypeNameWithGenerics(allowKeywordAsName: true)
           base = RawTypeSyntax(
             RawMemberTypeIdentifierSyntax(
               baseType: base,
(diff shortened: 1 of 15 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

Sources/SwiftParser/Types.swift
```diff
@@ -337,8 +337,15 @@ extension Parser {
     )
   }
 
-  mutating func parseTypeNameWithGenerics(include flags: DeclNameOptions = []) -> (RawTokenSyntax, RawGenericArgumentClauseSyntax?) {
-    let (name, _) = self.parseDeclNameRef(flags)
+  mutating func parseTypeNameWithGenerics(allowKeywordAsName: Bool) -> (RawTokenSyntax, RawGenericArgumentClauseSyntax?) {
+    let name: RawTokenSyntax
+    if let identOrSelf = self.consume(if: .identifier, .keyword(.Self), .keyword(.Self)) {
```
bnbarham, line 342: `Self` twice

CodeGeneration/Sources/SyntaxSupport/ExprNodes.swift
```diff
@@ -551,7 +551,7 @@ public let EXPR_NODES: [Node] = [
     children: [
       Child(
         name: "Identifier",
-        kind: .token(choices: [.token(tokenKind: "IdentifierToken")])
+        kind: .token(choices: [.token(tokenKind: "IdentifierToken"), .keyword(text: "self"), .keyword(text: "Self"), .keyword(text: "init")])
```
bnbarham, line 554: Editor placeholder is always <##> so ... seems weird we'd need this.

Sources/SwiftParser/Types.swift
ahoppen: Oh, one of them should’ve been lowercase `self`.

CodeGeneration/Sources/SyntaxSupport/ExprNodes.swift
ahoppen: You’re right, we don’t actually need it here.

ahoppen:
@swift-ci Please test Windows

bnbarham (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #1438, opened 2023-03-24
Lowercase all `Assert` methods

ahoppen:
I think they were uppercase because they used to be called `XCTAssert*` but really it’s against all naming guidelines to start a Swift function with an uppercase letter.


ahoppen:
@swift-ci Please test

bnbarham (approved):

kimdv (approved):

PR #1441, opened 2023-03-24
If no explicit highlight range was given, highlight the node the diagnostic is anchored at instead of highlighting nothing

ahoppen:
Especially for IDEs we should always report a highlight range whenever possible because it looks nicer if a token can be underlined instead of just underlining the character at which the diagnostic is anchored, which is what many IDEs default to.

rdar://104113111


ahoppen:
@swift-ci Please smoke test

kimdv (approved):

rintaro (approved):

PR #1443, opened 2023-03-25
Add diagnostic for equal with missing trailing or leading space

kimdv:


Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -771,6 +771,23 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
+
+      addDiagnostic(
+        node.equal,
+        StaticTokenError.equalMustHaveConsistentWhitespaceOnBothSides,
+        fixIts: [
+          FixIt(message: .insertWhitespace, changes: .makePresent(node.equal, leadingTrivia: leadingTrivia, trailingTrivia: .space))
+        ],
+        handledNodes: [node.equal.id]
+      )
+    }
```
kimdv, line 789: I'm not sure how to add the fix it. 
Is this the right way?

Tests/SwiftParserTest/translated/RecoveryTests.swift
```diff
@@ -1962,14 +1962,21 @@ final class RecoveryTests: XCTestCase {
...
+        let _1️⃣= 4
+        let _ =2️⃣5
       }
       """,
       diagnostics: [
-        // TODO: Old parser expected error on line 3: '=' must have consistent whitespace on both sides, Fix-It replacements: 8 - 8 = ' '
-        // TODO: Old parser expected error on line 4: '=' must have consistent whitespace on both sides, Fix-It replacements: 10 - 10 = ' '
-      ]
+        DiagnosticSpec(locationMarker: "1️⃣", message: "'=' must have consistent whitespace on both sides"),
+        DiagnosticSpec(locationMarker: "2️⃣", message: "'=' must have consistent whitespace on both sides"),
```
kimdv, line 1971: This position is wrong. We get 9 but expect 10. Not sure I'm finding the correct error position in the Cursor.swift

kimdv: Deleting derived data fixed it

Tests/SwiftParserTest/translated/RecoveryTests.swift
```diff
@@ -1978,14 +1985,22 @@ final class RecoveryTests: XCTestCase {
...
       }
       """,
       diagnostics: [
-        // TODO: Old parser expected error on line 3: '=' must have consistent whitespace on both sides, Fix-It replacements: 13 - 13 = ' '
-        // TODO: Old parser expected error on line 4: '=' must have consistent whitespace on both sides, Fix-It replacements: 20 - 20 = ' '
-      ]
+        DiagnosticSpec(locationMarker: "1️⃣", message: "'=' must have consistent whitespace on both sides"),
+        DiagnosticSpec(locationMarker: "2️⃣", message: "'=' must have consistent whitespace on both sides"),
+      ],
+      applyFixIts: ["insert whitespace", "insert whitespace"],
```
ahoppen, line 1996: Do you need to explicitly specify `applyFixIts`? By default it applies all.

Tests/SwiftParserTest/translated/RecoveryTests.swift
```diff
@@ -1962,14 +1962,21 @@ final class RecoveryTests: XCTestCase {
...
       """,
       diagnostics: [
-        // TODO: Old parser expected error on line 3: '=' must have consistent whitespace on both sides, Fix-It replacements: 8 - 8 = ' '
-        // TODO: Old parser expected error on line 4: '=' must have consistent whitespace on both sides, Fix-It replacements: 10 - 10 = ' '
-      ]
+        DiagnosticSpec(locationMarker: "1️⃣", message: "'=' must have consistent whitespace on both sides"),
+        DiagnosticSpec(locationMarker: "2️⃣", message: "'=' must have consistent whitespace on both sides"),
+      ],
+      fixedSource: """
+        // <rdar://problem/22387625> QoI: Common errors: 'let x= 5' and 'let x =5' could use Fix-its
```
ahoppen, line 1974: If you want, you can extract this comment from the string literal. It’s just an artifact of the translation that it ended up in the literal instead of being a real comment.

Sources/SwiftParserDiagnostics/LexerDiagnosticMessages.swift
```diff
@@ -198,6 +202,15 @@ public extension SwiftSyntax.TokenDiagnostic {
...
       ]
+    case .equalMustHaveConsistentWhitespaceOnBothSides:
+      let hasLeadingSpace = token.previousToken(viewMode: .all)?.trailingTrivia.contains(where: { $0.isSpaceOrTab }) ?? false
+      let fixedToken =
+        token
+        .with(\.leadingTrivia, hasLeadingSpace ? [] : .space)
+        .with(\.trailingTrivia, .space)
+      return [
+        FixIt(message: .insertWhitespace, changes: [[.replace(oldNode: Syntax(token), newNode: Syntax(fixedToken))]])
+      ]
```
ahoppen, line 213: You should be able to use `replaceLeadingTrivia` and `replaceTrailingTrivia` instead of replacing the entire token. That should show up nicer in IDE’s which will then really only highlight the replaced trivia instead of also assuming that the token is replaced (by itself).

Also, I’d make sure that you only replace trivia if it’s really needed. I.e. if leading trivia already has a space, there’s no need to replace it.

---

But I agree that the entire `FixIt.Change` API is a mess and I want to improve that pretty soon-ish.

ahoppen (commented):
Thanks. This looks great!

PR #1446, opened 2023-03-25
Available cannot be used as an expression

CippoX:
@ahoppen I'm not sure how to diagnose an UnexpectedNodesSyntax with 2 token, should an UnexpectedNodesSyntax have two tokens? Is the resulting tree correct?

```swift
 if #available(*) == false {
 }
```

```
0: ConditionElementSyntax children=2
  0: AvailabilityConditionSyntax children=4
    0: poundAvailableKeyword
    1: leftParen
    2: AvailabilitySpecListSyntax children=1
       0: AvailabilityArgumentSyntax children=1
          0: binaryOperator("*")
     3: rightParen
  1: UnexpectedNodesSyntax children=2
    0: binaryOperator("==")
    1: keyword(SwiftSyntax.Keyword.false)
```


ahoppen:
Yes, that seems as expected to me. You’d need to check that the `UnexpectedNodesSyntax` contains two elements, one of which is `==` and the second one is `false`.

CippoX:
Thanks, that's what I thought, just wanted to be sure.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -470,6 +470,28 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
         message: { _ in .joinConditionsUsingComma },
         moveFixIt: { ReplaceTokensFixIt(replaceTokens: $0, replacement: trailingComma) }
       )
+    } else if let unexpected = node.unexpectedBetweenConditionAndTrailingComma,
+      let availability = node.condition.as(AvailabilityConditionSyntax.self),
+      let comparisonOperator = unexpected.oneTokenSatisfying(satisfying: { $0.tokenKind == .binaryOperator("==") }),
```
CippoX, line 475: I thought it might be convenient to create an ad-hoc function . Let me know if there's a better way to do it.

Sources/SwiftParserDiagnostics/SyntaxExtensions.swift
```diff
@@ -39,6 +39,16 @@ extension UnexpectedNodesSyntax {
       return nil
     }
   }
+
+  /// If this only contains one tokens satisfying `condition`, return that token, otherwise return `nil`.
+  func oneTokenSatisfying(satisfying condition: (TokenSyntax) -> Bool) -> TokenSyntax? {
```
CippoX, line 44: I thought it might be convenient to create an ad-hoc function . Let me know if there's a better way to do it.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -470,6 +470,28 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
+      let falseKeyword = unexpected.oneTokenSatisfying(satisfying: { $0.tokenKind == .keyword(.false) })
```
ahoppen, line 476: But this would also be `true` if `unexpected` is `false ==` (different order) or `== false some more garbage` right?

What do you think about adding a `twoTokens(first: (TokenSyntax) -> Bool, second: (TokenSyntax) -> Bool) -> Bool` function similar to `onlyToken` in SyntaxExtensions.swift?

PR #1447, opened 2023-03-25
Add diagnostic for missing conditional binding keyword

kimdv:


Sources/SwiftParser/Statements.swift
```diff
@@ -219,15 +219,15 @@ extension Parser {
   ///     optional-binding-condition → 'let' pattern initializer? | 'var' pattern initializer? |
   ///                                  'inout' pattern initializer?
   @_spi(RawSyntax)
-  public mutating func parseConditionElement() -> RawConditionElementSyntax.Condition {
+  public mutating func parseConditionElement(lastBindingKind: RawTokenSyntax?) -> RawConditionElementSyntax.Condition {
```
ahoppen, line 222: Could you add a doc comment describing what `lastBindingKind` does?

Sources/SwiftParser/Statements.swift
```diff
@@ -255,7 +254,14 @@ extension Parser {
       let pattern = self.parseMatchingPattern(context: .matching)
       kind = .pattern(caseKeyword, pattern)
     } else {
-      let letOrVar = self.consumeAnyToken()
+      let letOrVar: RawTokenSyntax
+
+      if self.at(.identifier) {
+        letOrVar = RawTokenSyntax(missing: lastBindingKind!.tokenKind, text: lastBindingKind!.tokenText, arena: self.arena)
```
ahoppen, line 260: Could you either add a comment describing why `lastBindingKind` cannot be `nil` here or (and I’d probably prefer that) add `let lastBindingKind = lastBindingKind` to the `if` condition?

---

I think you don’t need to specify `text`

---

Could you add a comment that we’re in the case where the the user forgot to write `let` or `var`?

Sources/SwiftParser/Statements.swift
```diff
@@ -255,7 +254,14 @@ extension Parser {
...
+      } else {
+        letOrVar = self.consumeAnyToken()
```
ahoppen, line 262: I know this has been `consumeAnyToken` before but if you’ve got time could you try if you could change this to `self.expect(.let, .var)`. That just makes it a lot clearer what we expect `letOrVar` to be. `consumeAnyToken` is pretty much an anti-pattern in 90% of all cases that I would like to get rid of in general.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -913,6 +913,28 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
+        ExpectedBindingKeyword(token: node.bindingKeyword),
+        fixIts: [
+          FixIt(
+            message: InsertFixIt(tokenToBeInserted: node.bindingKeyword),
+            changes: [.makePresent(node.bindingKeyword)]
+          )
+        ],
+        handledNodes: [node.bindingKeyword.id]
+      )
+    }
```
ahoppen, line 933: Out of curiosity: What’s the error message if we don’t have this custom one? My expectation is that it would be equally good and we could save ourselves some code. But I might be wrong.

kimdv: They are more or less the same.

kimdv: Removed this

Sources/SwiftParser/Statements.swift
kimdv: Something like this?

ahoppen: Looks good 👍🏽

ahoppen (approved):

kimdv:
@swift-ci please test

PR #1448, opened 2023-03-26
Add diagnostics for missing modifier in operator declaration

StevenWong12:
Part of #1373


Tests/SwiftParserTest/translated/OperatorDeclTests.swift
```diff
@@ -167,7 +167,9 @@ final class OperatorDeclTests: XCTestCase {
       prefix operator %%+
       """,
       diagnostics: [
-        DiagnosticSpec(message: "unexpected code before operator declaration")
+        DiagnosticSpec(message: "operator must be declared as 'prefix', 'postfix', or 'infix'"),
+        // TODO: should be "unexpected code before operator declaration" here
+        DiagnosticSpec(message: "unexpected code in operator declaration"),
```
StevenWong12, line 172: I wonder whether there is a method can be used to mark a SyntaxCollection as a missing node when parsing? If not, does it make sense to create one? So that we can ignore the missing ``modifierList`` here in ``sourceAccurate`` mode and produce unexpected code **before** operator declaration rather than **in** declaration.

StevenWong12: Actually, the cause of the problem I mentioned above is that when we try to parse a modifier with a semicolon in front of, both the semicolon and the modifier will be parsed into an `UnexpectedNodeSyntax`. Although we could produce the correct diagnostic message, but the AST seems not correct to me, since the modifier is not unexpected. Maybe we could solve this problem in terms of recovery rather than the methods I mentioned above?

Sources/SwiftParser/Declarations.swift
```diff
@@ -1964,9 +1964,16 @@ extension Parser {
     } else {
       unexpectedAtEnd = nil
     }
+
+    var modifiers: RawModifierListSyntax? = attrs.modifiers
+    if modifiers == nil, unexpectedBeforeOperatorKeyword == nil {
```
StevenWong12, line 1969: I check the ``unexpectedBeforeOperatorKeyword`` here to keep ``testOperatorDecl10`` produce the same diagnostics as the old parser. Not sure whether this makes sense🤔

ahoppen (commented):
I haven’t looked at your changes in detail yet but I think the way to fix this is to change how `OperatorDecl` is represented. It shouldn’t have an arbitrary `ModifierList` but instead require a single child that can be either of the following tokens: `prefix`, `infix`, `postfix`. And then `parseOperatorDeclaration` wouldn’t actually be passed a `DeclAttributes` (which is essentially attributes + modifiers). Instead, it would require some handling in `parseDecl` earlier on.

Does this make sense to you?

StevenWong12:
My changes just handled the case that missing modifier for operator declaration. Your comments reminded me that the wrong modifier here is also not handled yet and I have updated the codes. 😀

As for the implementation, I prefer to parse the modifier list in ``parseDecl`` and handle the misused or missing modifiers in ``parseOperatorDeclaration``. Then we could leave other work to ``ParseDiagnosticsGenerator``.

Tests/SwiftParserTest/translated/OperatorDeclTests.swift
```diff
@@ -480,6 +480,41 @@ final class OperatorDeclTests: XCTestCase {
...
+    )
+  }
+  
+  func testOperatorDecl125() {
+    assertParse(
+    """
+    prefix 1️⃣postfix operator ** : PrecedenceGroup
+    """,
+    diagnostics: [
+        DiagnosticSpec(message: "unexpected code 'postfix' in operator declaration")
```
StevenWong12, line 501: The old parser generates *'postfix' contradicts previous modifier 'prefix'* here. If my changes work, I could optimize the diagnostics message here later.

StevenWong12:
I have refactored how ``OperatorDecl`` is presented, with a ``TokenSyntax`` to represent ``infix``, ``prefix`` or ``postfix`` and get it checked in ``RawSyntaxValidation.swift``. 

According to the [doc](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/declarations#Operator-Declaration), we can not have attributes in operator declarations, so I have also removed the attributes in ``OperatorDeclSyntax``.

As for SwiftParser,  
> Instead, it would require some handling in parseDecl earlier on

I think handling in the early stage of ``parseDecl`` will require an additional call of ``canRecoverTo`` to help us determine whether this declaration needs ``DeclAttributes``, which will bring some unnecessary overhead, since only operator declaration and precedence group declaration do not need ``DeclAttributes``. 

So I make the attributes and modifiers list parsed as before, and get them handled in ``parseOperatorDeclarationIntroducer``. Does it make sense to you? @ahoppen 

If there is anything I missed, please feel free to let me know.

Thanks in advance!

Tests/SwiftParserTest/translated/OperatorDeclTests.swift
```diff
@@ -167,7 +167,8 @@ final class OperatorDeclTests: XCTestCase {
       prefix operator %%+
       """,
       diagnostics: [
-        DiagnosticSpec(message: "unexpected code before operator declaration")
+        DiagnosticSpec(message: "operator must be declared as 'prefix', 'postfix', or 'infix'"),
```
ahoppen, line 170: Why do we get an error message that the operator needs to be declared as `prefix` here? It is, right?

Sources/SwiftParser/Declarations.swift
```diff
@@ -1877,12 +1877,55 @@ extension Parser {
...
+      operatorModifier: modifier,
+      unexpectedBetweenModifierAndOperatorKeyword: unexpectedBetweenModifierAndOperatorKeyword,
+      operatorKeyword: operatorKeyword
+    )
+
+  }
+
   @_spi(RawSyntax)
   public mutating func parseOperatorDeclaration(
     _ attrs: DeclAttributes,
```
ahoppen, line 1923: Instead of retroactively taking `attrs` apart to find the operator, I think it would be easier if `parseOperatorDeclaration` doesn’t take any `attrs` at all. You would just need to make sure that we call into `parseOperatorDeclaration` before the attributes are getting parsed in `parseDeclaration`. I would at least like to see how that looks like. If it’s truly awful, I’m fine with reverting to this design as a last resort.

StevenWong12:
> You would just need to make sure that we call into parseOperatorDeclaration before the attributes are getting parsed in parseDeclaration
To achieve this, I made a most straightforward way: make an extra call of ``canRecoverTo``. This indeed makes the ``parseDeclaration`` and ``parseOperatorDecl`` more simple. 

Considering I call the extra ``canRecoverTo`` before ``parseAttributeList`` and ``parseModifierList``, I add two extra test cases (``testOperatorDecl124`` and ``testOperatorDecl125``) to make sure the output fall into expectation.

Also, since I call the extra ``canRecoverTo`` every time when we parse a declaration, I wonder whether this will cause a certain degree of performance loss?

Sources/SwiftParser/Declarations.swift
```diff
@@ -211,15 +211,22 @@ extension Parser {
...
+    // If we are inside a memberDecl list, we don't want to eat closing braces (which most likely close the outer context)
+    // while recoverying to the declaration start.
+    let recoveryPrecedence = inMemberDeclList ? TokenPrecedence.closingBrace : nil
+
+    switch self.canRecoverTo(anyIn: DeclarationStart.self, overrideRecoveryPrecedence: recoveryPrecedence) {
+    case (.operatorKeyword, _)?:
+      return RawDeclSyntax(parseOperatorDeclaration())
+    default:
+      break
+    }
```
ahoppen, line 223: You should be able to simplify this to

```suggestion
    if self.canRecoverTo(TokenSpec(.operator, recoveryPrecedence: recoveryPrecedence) != nil {
      return RawDeclSyntax(parseOperatorDeclaration())
    }
```

Sources/SwiftParser/Lookahead.swift
```diff
@@ -179,6 +179,17 @@ extension Parser.Lookahead {
...
+    guard self.at(anyIn: DeclarationModifier.self) != nil else {
+      return false
+    }
+    while self.consume(ifAnyIn: DeclarationModifier.self) != nil {
+      self.skipSingle()
+    }
+
+    return true
+  }
+
```
ahoppen, line 192: Is this still needed?

Sources/SwiftParser/SyntaxUtils.swift
```diff
@@ -41,6 +41,13 @@ extension RawUnexpectedNodesSyntax {
   init?<SyntaxType: RawSyntaxNodeProtocol>(_ nodes: [SyntaxType?], arena: __shared SyntaxArena) {
     self.init(nodes.compactMap({ $0 }), arena: arena)
   }
+
+  init?<SyntaxType: RawSyntaxNodeProtocol>(_ nodes: [SyntaxType]?, arena: __shared SyntaxArena) {
+    guard let nodes = nodes else {
+      return nil
+    }
+    self.init(nodes, arena: arena)
+  }
```
ahoppen, line 50: Is this still needed?

ahoppen (commented):
Yes, this looks much simpler now. I’ve got a few local comments inline, after that it’s ready to be merged.

Sources/SwiftParser/Declarations.swift
StevenWong12: If we only try to recover to ``.operator``, there will be some corner case e.g.
```swift
enum foo {
  case operator
}
```
where ``case`` is with ``stmtKeyword`` precedence and ``operator`` is with ``declKeyword`` precedence, which leads to some broken diagnostic messages.

On the other hand, I think have a switch control flow here can help us extend to those declarations that do not need ``DeclAttr`` later, such as ``precedenceGroupDecl``. WDYT?

ahoppen: That makes sense to me. You have convinced me. 👍🏽

StevenWong12:
Sorry for committing some test codes before. I have already removed them.😁

ahoppen (dismissed):
Thanks. Let’s get this merged.

ahoppen:
@swift-ci Please test

ahoppen:
https://github.com/apple/swift/blob/main/test/attr/accessibility.swift failed to round-trip in the parser. Could you investigate what’s going wrong here? A good place to start is
1. Clone https://github.com/apple/swift
2. Build `swift-parser-cli` in this repo
3. Run `swift-parser-cli reduce path/to/test/attr/accessibility.swift`. That should give you a reduced test case
4. Create an `assertParse` test case from the output of the step above and fix that.

StevenWong12:
I think the problem lies in that ``private`` modifier has ``declKeyword`` precedence so that we can not recover to ``operator`` until modifiers are parsed. 

My solution is to add a new ``declModifier`` token precedence group whose priority just lower than ``declKeyword``. Also, I add a ``overrideRecoveryPrecedence`` param for ``expect`` to promote the token precedence of operator modifiers when we try to recover to them and exclude other modifiers.

Actually, this problem is supposed to be detected by ``testOperatorDecl125`` but I find that the recovery precedence of ``mutating`` keyword is not explicitly assigned 😂 .

kimdv:
@swift-ci please test

StevenWong12:
[Last CI test](https://ci.swift.org/job/swift-syntax-PR-macOS/2096/consoleFull#135254046d6fdb6cb-f376-4f2e-8bce-d31c7304698b) indicates that 
 > TEST 'Swift(macosx-x86_64) :: attr/attr_dynamic.swift' FAILED
 
 but the error seems a little odd to me 
 > /Users/ec2-user/jenkins/workspace/swift-syntax-PR-macOS/branch-main/swift/test/attr/attr_dynamic.swift:1:1: error: unexpected error produced: new Swift parser generated errors for code that C++ parser accepted
// RUN: %target-typecheck-verify-swift -enable-objc-interop
^
/Users/ec2-user/jenkins/workspace/swift-syntax-PR-macOS/branch-main/swift/test/attr/attr_dynamic.swift:10:1: error: unexpected error produced: unexpected code 'dynamic' before operator declaration
dynamic prefix operator +!+  // expected-error{{'dynamic' modifier cannot be applied to this declaration}} {{1-9=}}

since we should produce a unexpected error here.

ahoppen:
The problem here is that `'dynamic' modifier cannot be applied to this declaration` was diagnosed by the type checker, not parser in the C++ world. Thus, the C++ parser actually accepted the following code and it was only diagnosed at type checking time.

```swift
dynamic prefix operator +!+ 
```

Now you are diagnosing the error at parse time, which causes the error: `new Swift parser generated errors for code that C++ parser accepted`. I need to think about how best to fix or work around this and will come back to you in the next couple of days.

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
```diff
@@ -1283,24 +1283,14 @@ public let DECL_NODES: [Node] = [
...
-        isOptional: true
-      ),
-      Child(
-        name: "Modifiers",
-        kind: .collection(kind: "ModifierList", collectionElementName: "Modifier"),
-        nameForDiagnostics: "modifiers",
-        description: "The declaration modifiers applied to the 'operator' declaration.",
-        isOptional: true,
-        classification: "Attribute"
+        name: "OperatorModifier",
```
ahoppen, line 1290: The correct term for this would be `Fixity`

ahoppen (commented):
After talking to @rintaro about it, I think this needs two adjustments:

1. @rintaro rightfully noticed that the extra `canRecoverTo` check in `parseDeclaration` might have a non-trivial performance cost and since operator declarations are not very common, we shouldn’t need to pay that. So, I changed my mind and think that `parseOperator` should indeed be taking `attrs` as a parameter and be taking it apart to find the fixity of the operator declaration. The overall redesign of `OperatorDeclSyntax` should stay the same though. Sorry for my change of mind.
2. To fix the CI failure that I explained in my previous comment, I think we should update the C++ parser to diagnose missing fixity of an operator. Do you want to try that @StevenWong12? I can give you a few pointers. If you don’t feel comfortable doing that, I can also take care of it.

StevenWong12:
Oh, I see. I think this is not a big deal for me since I have implemented it before.😄

> I think we should update the C++ parser to diagnose missing fixity of an operator

I'd like to give it a try. And some pointers will be much appreciated!

StevenWong12:
I opened a [PR](https://github.com/apple/swift/pull/65224) to modify the C++ parser. Does these changes make sense to you? @ahoppen 

If that pr works, I can update this pr later.

StevenWong12:
May need to re-run CodeGeneration after #1567 is merged

StevenWong12:
> May need to re-run CodeGeneration after https://github.com/apple/swift-syntax/pull/1567 is merged

Done!

kimdv:
@swift-ci please test

ahoppen:
@StevenWong12 Could you rebase this and resolve the merge conflicts? Now that the Swift PR is merged, I think we can give this another shot.

StevenWong12:
Hey @ahoppen, I've rebased this PR. Could you please trigger the CI?

Sources/SwiftParser/Declarations.swift
```diff
@@ -1697,12 +1697,91 @@ extension Parser {
...
+        unexpectedBeforeFixity = RawUnexpectedNodesSyntax(combining: unexpectedBeforeFixity, RawUnexpectedNodesSyntax(unexpectedBeforeFixityElements, arena: self.arena), arena: self.arena)
+      }
+
+      if !unexpectedBeforeOperatorElements.isEmpty {
+        unexpectedBeforeOperatorKeyword = RawUnexpectedNodesSyntax(unexpectedBeforeOperatorElements, arena: self.arena)
+      } else {
+        unexpectedBeforeOperatorKeyword = nil
+      }
+
+    }
```
StevenWong12, line 1765: To produce better diagnostic when fixity could not be parsed into `attrs.modifiers`, e.g. 
```swift
; // ';' separator here has higher precedence and prefix is a contextual keyword
prefix operator ++
```
I added these codes to extract the fixity from `unexpectedBeforeOperatorKeyword`, which looks IMO awful. Do you have better idea about this?

Sources/SwiftParser/Declarations.swift
```diff
@@ -1697,12 +1697,91 @@ extension Parser {
...
+      }
+    }
+
+    var (unexpectedBeforeOperatorKeyword, operatorKeyword) = self.expect(.keyword(.operator))
+
+    // Sometimes we could not parse the fixity into attr.modifiers
+    // so we need extract it from unexpectedBeforeOperatorKeyword manually
+    if fixity.presence == .missing,
+      let elements = unexpectedBeforeOperatorKeyword?.elements
+    {
```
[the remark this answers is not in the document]
StevenWong12: Yeah, that's another way I thought before. But there are some corner cases in the way when I tried to recover to `DeclarationModifier` with `declKeyword` precedence and I could not find a proper workaround with them.  

1. 
```swift
enum foo {
  case a
  static let b
}
```

The modifier `static` here has  `declKeyword` precedence and `case` has `stmtKeyword` precedence, which makes the whole `case a` as an unexpected node.

2. If we promote the `case` to `declKeyword` precedence, codes like
```swift
switch {
  @case
}
```
will generate some broken diagnostic message since `case` has a higher precedence than `}`

Do you have any idea about this?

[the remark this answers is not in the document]
ahoppen: OK, I’ve got one more idea: What you want to do is to raise the recovery precedence of `case` in this case. So, `case` needs to be part of the `TokenSpecSet` that you pass to `canRecoveryTo`. Since `case` is not a declaration modifier though, the way I would go about doing this, is to define a union set of `DeclarationStart` and `DeclarationModifier` just like we do for `ExpressionStart`. You should then be able to check if you can recover to this union set and if you recover to a case in `DeclarationStart`, you just `break` the loop.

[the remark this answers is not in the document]
StevenWong12: Cool! This works perfectly! 🤩 Already submitted this change.

Sources/SwiftParser/Declarations.swift
```diff
@@ -1697,12 +1697,55 @@ extension Parser {
...
+
+    var unexpectedBeforeFixity = RawUnexpectedNodesSyntax(combining: attrs.modifiers?.elements.compactMap { $0.unexpectedBeforeName }, RawUnexpectedNodesSyntax(attrs.attributes?.elements ?? [], arena: self.arena), arena: self.arena)
+
+    var fixity: RawTokenSyntax?
+    var leftModifiers: [RawDeclModifierSyntax?] = []
+
+    if let modifiers = attrs.modifiers?.elements {
+      for modifier in modifiers {
+        if fixity == nil,
+          fixityText.contains(modifier.name.tokenText)
```
ahoppen, line 1720: I would prefer to use `TokenSpec` instead of comparing the token text, just because it does a bit more keyword handling and is generally considered the entry point to check if a token is of a specific kind. Something like

```swift
switch modifier.name {
case TokenSpec(.prefix), TokenSpec(.infix), TokenSpec(.postfix):
  // do something
}
```

Sources/SwiftParser/TokenSpecSet.swift
```diff
@@ -144,7 +144,6 @@ enum ContextualDeclKeyword: TokenSpecSet {
   case dynamic
   case final
   case indirect
-  case infix
```
ahoppen, line 147: Why did you need to remove `infix` from `ContextualDeclKeyword`?

Sources/SwiftParser/Declarations.swift
```diff
@@ -1697,12 +1697,55 @@ extension Parser {
...
+    var fixity: RawTokenSyntax
+    var unexpectedBeforeOperatorKeyword: RawUnexpectedNodesSyntax?
+    var operatorKeyword: RawTokenSyntax
+  }
+
+  mutating func parseOperatorDeclIntroducer(_ attrs: DeclAttributes, _ handle: RecoveryConsumptionHandle) -> OperatorDeclIntroducer {
+
+    let fixityText = [Keyword.prefix, .infix, .postfix].map({ $0.defaultText })
+
+    var unexpectedBeforeFixity = RawUnexpectedNodesSyntax(combining: attrs.modifiers?.elements.compactMap { $0.unexpectedBeforeName }, RawUnexpectedNodesSyntax(attrs.attributes?.elements ?? [], arena: self.arena), arena: self.arena)
```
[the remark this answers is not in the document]
StevenWong12: > The following should give you RawUnexpectedNodesSyntax? which conforms to UnexpectedNodesCombinable

The attributes in `attrs` is actually optional, so we could not just initialize a `RawUnexpectedNodesSyntax` with it, since that requires a non-optional value.

> Any modifiers after the fixity get added to leftModifiers

I think we could just make modifiers after the fixity parsed into `unexpectedBeforeOperatorKeyword`. It should make it easier.

Sources/SwiftParser/Modifiers.swift
```diff
@@ -18,36 +18,38 @@ extension Parser {
     var elements = [RawDeclModifierSyntax]()
     var modifierLoopCondition = LoopProgressCondition()
     MODIFIER_LOOP: while modifierLoopCondition.evaluate(currentToken) {
-      switch self.at(anyIn: DeclarationModifier.self) {
-      case (.private, _)?,
-        (.fileprivate, _)?,
-        (.internal, _)?,
-        (.public, _)?:
+      switch self.canRecoverTo(anyIn: DeclarationStart.self, overrideRecoveryPrecedence: .declKeyword) {
```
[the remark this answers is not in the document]
StevenWong12: I think this should be the easiest way. Since we get the `spec` from the associated value in `declarationModifier` and `declarationKeyword`, override the token precedence in `DeclarationStart` may require us to add a new initializer to `TokenSpec` which init from other `TokenSpec` and pass the new token precedence to it. Is there anything I missed?

Sources/SwiftParser/TokenSpecSet.swift
StevenWong12: That was a wrong modification, it has been modified.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -1052,6 +1052,15 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
     if shouldSkip(node) {
       return .skipChildren
     }
+
+    if node.fixity.presence == .missing {
+      addDiagnostic(
+        node.fixity,
+        .missingModifierInOperatorDeclaration,
+        handledNodes: [node.fixity.id]
```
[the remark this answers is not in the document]
StevenWong12: I think that's ok. Do we need to add all possible Fix-Its on this type of missing node in the future? e.g. the mandatory `_version` or `_underlyingVersion` label of the second parameter of `canImport`.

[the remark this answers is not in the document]
ahoppen: > Before this, my thought was not to add the Fix-Its since we could not infer what the user miss here.

As long as there is a reasonably short list of possible options, I think offering Fix-Its is just a nice thing to do. And we don’t need to know what the user wanted to write because there will be three different Fix-Its, one for each fixity, and the user can choose which one they want to apply.

And yes, I agree that offering the same for `_version` and `_underlyingVersion` would also be nice.

Sources/SwiftParser/Declarations.swift
```diff
@@ -1694,11 +1694,57 @@ extension Parser {
...
+
+    if let modifiers = attrs.modifiers?.elements,
+      let firstFixityIndex = modifiers.firstIndex(where: { isFixity($0) })
+    {
+      let targetModifier = modifiers[firstFixityIndex]
+      fixity = targetModifier.name
+
+      let beforeFixity = Array(modifiers[0..<firstFixityIndex])
+
+      unexpectedBeforeFixity = RawUnexpectedNodesSyntax(combining: unexpectedBeforeFixity, RawUnexpectedNodesSyntax(beforeFixity, arena: self.arena), targetModifier.unexpectedBeforeName, arena: self.arena)
```
ahoppen, line 1729: This looks a lot better now. You are still dropping `targetModifier.unexpectedBetweenNameAndDetail`, `targetModifier.detail` and `targetModifier.unexpectedAfterDetail` though. Those should be added to `unexpectedBeforeOperatorKeyword`.

Sources/SwiftParser/Declarations.swift
```diff
@@ -1694,11 +1694,57 @@ extension Parser {
...
+    }
+
+    var unexpectedBeforeFixity = RawUnexpectedNodesSyntax(attrs.attributes?.elements ?? [], arena: self.arena)
+
+    var fixity: RawTokenSyntax?
+
+    if let modifiers = attrs.modifiers?.elements,
+      let firstFixityIndex = modifiers.firstIndex(where: { isFixity($0) })
+    {
+      let targetModifier = modifiers[firstFixityIndex]
```
ahoppen, line 1724: What about naming this variable `fixityModifier`? Because that’s really what it is, I think. `target` is kind of ambiguous IMO.

Sources/SwiftParser/Modifiers.swift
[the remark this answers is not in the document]
ahoppen: The thing is that every other call that tries to recover to a `DeclarationModifier` or `DeclarationStart` would also need to specify the `overrideRecoveryPrecedence`. And really, the increased recovery precedence is not because of the context we are recovering here but it’s inherent to the keywords we try to recover to, all of which are declaration keywords in this case. 

I briefly checked locally and adjusting the recovery precedence of the `DeclarationModifier`s seems to work. So, I think we should continue doing that.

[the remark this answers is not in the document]
StevenWong12: > the increased recovery precedence is not because of the context we are recovering here

That makes sense👍. I've moved the precedence override to `DeclarationModifier`.

Sources/SwiftParser/Declarations.swift
StevenWong12: Good catch! These just really easy to be left.

Sources/SwiftParser/TokenPrecedence.swift
```diff
@@ -219,7 +219,7 @@ enum TokenPrecedence: Comparable {
       // Operator stuff
       .operator, .precedencegroup,
       // Misc
-      .import:
+      .import, .__consuming:
```
ahoppen, line 222: You adding `__consuming` here actually got me thinking if we really need to specify a recovery precedence in `DeclarationModifier` (sorry for not noticing this earlier). I just checked and all keywords in there except for the following are exclusively used for declarations, so we should be able to just set their recovery precedence in `TokenPrecedence`.

- `weak` 
- `unowned`
- `isolated`
- `rethrows`
- `async`
- `reasync`
- `_const`
- `consuming`
- `borrowing`

Just for reference: What I did was right click on all of the `Keyword.something` references and select Find -> “Find Selected Symbol in Workspace” and then judging on where the keyword was used. If it was only used in `DeclarationModifier`, it’s unconditionally a declaration keyword.

Sources/SwiftParser/Declarations.swift
```diff
@@ -1694,11 +1694,62 @@ extension Parser {
...
+    {
+      let fixityModifier = modifiers[firstFixityIndex]
+      fixity = fixityModifier.name
+
+      leftInFixityModifier = RawUnexpectedNodesSyntax(combining: fixityModifier.unexpectedBetweenNameAndDetail, RawUnexpectedNodesSyntax([fixityModifier.detail], arena: self.arena), fixityModifier.unexpectedAfterDetail, arena: self.arena)
+
+      let beforeFixity = Array(modifiers[0..<firstFixityIndex])
+
+      unexpectedBeforeFixity = RawUnexpectedNodesSyntax(combining: unexpectedBeforeFixity, RawUnexpectedNodesSyntax(beforeFixity, arena: self.arena), fixityModifier.unexpectedBeforeName, arena: self.arena)
+    }
```
ahoppen, line 1733: If `modifiers.firstIndex(where: { isFixity($0) })` returns `nil`, we drop the modifiers. To avoid that, you need to add an `else` branch to the `if` statement that adds all modifiers to `unexpectedBeforeFixity or `unexpectedBeforeOperatorKeyword`. Which one you choose doesn’t really matter here because there are no present tokens between these two unexpected sets.

Sources/SwiftParser/TokenPrecedence.swift
StevenWong12: I added `__consuming` is because it's `isLexerClassifiedKeyword=true`. It seems assigned precedence to most of the modifier keywords there also make sense. Already done this.

Sources/SwiftParser/Declarations.swift
StevenWong12: Good catch! I also added a test case for this.

ahoppen:
@swift-ci Please test

ahoppen:
@StevenWong12 Looks like you modified generated files that would be overwritten when CodeGeneration is run again. Could you make sure that your changes are properly generated using CodeGeneration. https://github.com/apple/swift-syntax/blob/main/CodeGeneration/README.md describes how to run CodeGeneration.

StevenWong12:
Oh, I didn't notice `DeclarationModifier.swift` is a generated file. Already synced changes in CodeGeneration.

ahoppen (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@StevenWong12 Testing failed. Assuming that you ran the test locally, they probably failed in the test case mutation mode. I’m adding documentation of how to run these tests in https://github.com/apple/swift-syntax/pull/1696

StevenWong12:
Solved! 

There is another thing that came to my mind that we accept `RawDocComment` as an attribute for operator declaration in the C++ parser, do we have to do something corresponding here? But I am not quite sure how `RawDocComment` is modeled in `SwiftParser`.
___

Just out of curiosity: Is this mutation test basically the same as the fuzzing target in #1329?

Sources/SwiftParser/Declarations.swift
```diff
@@ -1718,6 +1718,7 @@ extension Parser {
 
     var fixity: RawTokenSyntax?
     var leftInFixityModifier: RawUnexpectedNodesSyntax?
+    var modifiersAfterFixity: RawUnexpectedNodesSyntax?
```
ahoppen: I don’t think these need to be separate variables. We can just combine the in the `if`-branch already. And I think variable-name wise, `afterFixity` would be a good name to match `beforeFixity`.

Sources/SwiftParser/Declarations.swift
```diff
@@ -1694,11 +1694,65 @@ extension Parser {
...
+    if let modifiers = attrs.modifiers?.elements {
+      if let firstFixityIndex = modifiers.firstIndex(where: { isFixity($0) }) {
+        let fixityModifier = modifiers[firstFixityIndex]
+        fixity = fixityModifier.name
+
+        let beforeFixity = Array(modifiers[0..<firstFixityIndex])
+
+        unexpectedBeforeFixity = RawUnexpectedNodesSyntax(combining: unexpectedBeforeFixity, RawUnexpectedNodesSyntax(beforeFixity, arena: self.arena), fixityModifier.unexpectedBeforeName, arena: self.arena)
+
+        unexpectedAfterFixity = RawUnexpectedNodesSyntax(combining: RawUnexpectedNodesSyntax(combining: fixityModifier.unexpectedBetweenNameAndDetail, RawUnexpectedNodesSyntax([fixityModifier.detail], arena: self.arena), fixityModifier.unexpectedAfterDetail, arena: self.arena), RawUnexpectedNodesSyntax(Array(modifiers[modifiers.index(after: firstFixityIndex)...]), arena: self.arena), arena: self.arena)
```
ahoppen, line 1731: To make this more readable, what do you think about adding a new overload of `RawUnexpectedNodesSyntax` that takes four parameters? Also, I think putting the parameters on separate lines would make this a lot more readable

```suggestion
        unexpectedAfterFixity = RawUnexpectedNodesSyntax(
          combining: fixityModifier.unexpectedBetweenNameAndDetail,
          RawUnexpectedNodesSyntax([fixityModifier.detail], arena: self.arena),
          fixityModifier.unexpectedAfterDetail,
          RawUnexpectedNodesSyntax(Array(modifiers[modifiers.index(after: firstFixityIndex)...]), arena: self.arena),
          arena: self.arena
        )
```

Tests/SwiftParserTest/DeclarationTests.swift
```diff
@@ -2163,4 +2163,31 @@ final class DeclarationTests: XCTestCase {
...
+        1️⃣^
+      }
+      unowned 2️⃣B 3️⃣{
+      }4️⃣
+      """,
+      diagnostics: [
+        DiagnosticSpec(locationMarker: "1️⃣", message: "unexpected code before modifier"),
+        DiagnosticSpec(locationMarker: "2️⃣", message: "expected 'func' in function", fixIts: ["insert 'func'"]),
+        DiagnosticSpec(locationMarker: "3️⃣", message: "expected parameter clause in function signature", fixIts: ["insert parameter clause"]),
+        DiagnosticSpec(locationMarker: "4️⃣", message: "expected '}' to end class", notes: [NoteSpec(locationMarker: "5️⃣", message: "to match this opening '{'")], fixIts: ["insert '}'"]),
```
ahoppen, line 2180: You can use ℹ️ instead of 5️⃣ for the marked source and then you don’t need to specify a `locationMarker` for the `NoteSpec` (because it defaults to ℹ️)

ahoppen (commented):
We’re nearly there

ahoppen:
Oh, and two more things: Looks like there is a merge conflict and can you squash your commits and force push? It makes for a nicer commit history IMO.

StevenWong12:
Done!

ahoppen:
@swift-ci Please test

ahoppen (approved):

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

ahoppen:
Sorry @StevenWong12, looks like introduced another conflict by merging https://github.com/apple/swift-syntax/pull/1668. Could you rebase one more time?

StevenWong12:
Done!

I noticed that there is a regression test in `swift-format` failed and I opened a [pr](https://github.com/apple/swift-format/pull/536) to sync this change.

kimdv:
https://github.com/apple/swift-format/pull/536

@swift-ci please test

StevenWong12:
Could you please trigger the test on Windows? @kimdv

ahoppen:
And another merge conflict, this time from https://github.com/apple/swift-syntax/pull/1711. Sorry that you continue to need to rebase this PR.

kimdv:
When rebased I'll merge tomorrow if it's not merged before or if it's not okay @ahoppen  😁

StevenWong12:
That's not a big deal. 😄

Already rebased.

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test windows

kimdv:
https://github.com/apple/swift-format/pull/536

@swift-ci please test

kimdv:
> I think maybe we should include the pr([apple/swift-format#536](https://github.com/apple/swift-format/pull/536)) in `swift-format` to trigger the test like yesterday.
> 
> Could you please trigger that again? @kimdv

Doh! Sorry about that. 
Triggered it again

PR #1449, opened 2023-03-26
Add diagnostic for array type

kimdv:


Sources/SwiftParser/Types.swift
```diff
@@ -1045,7 +1045,21 @@ extension Parser {
...
+        result = RawTypeSyntax(
+          RawArrayTypeSyntax(
+            leftSquareBracket: missingToken(.leftSquareBracket),
+            elementType: result,
+            unexpectedBeforeRSquareBracket,
+            rightSquareBracket: rightSquareBracket,
+            arena: self.arena
+          )
+        )
+      }
```
kimdv, line 1060: I'm not sure if I'm doing this at the right place. 

I've also been looking into to putting this into `parseSimpleType`

kimdv: @ahoppen

ahoppen: This is exactly the right place to handle it IMO. It’s also where the old parser handled it. https://github.com/apple/swift/blob/316dfaffa1a387dd0ab1b0e80f0e98c8cea09d7a/lib/Parse/ParseType.cpp#L646

In `parseSimpleType` you get into trouble because the `]` could belong to a subscript or an outer array type.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -405,6 +405,23 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
   }
 
+  public override func visit(_ node: ArrayTypeSyntax) -> SyntaxVisitorContinueKind {
+    if shouldSkip(node) {
+      return .skipChildren
+    }
+
+    if node.leftSquareBracket.presence == .missing {
+      addDiagnostic(
+        node.leftSquareBracket,
```
ahoppen, line 415: Can you anchor the diagnostic at `node.rightSquareBracket`. Since it’s talking about the unexpected `]`, I think it should also point to the `]`.

ahoppen (approved):
LGTM overall, just one comment on where the diagnostic should be anchored.

Sources/SwiftParser/Types.swift
kimdv: We get an strange error with `"let a2: Set<Int1️⃣]>"`

The `var result = self.parseType()` return this tree. 
It look like it can't find the ending `>`

```
▿ Set<Int
  ▿ raw : .layout(simpleTypeIdentifier byteLength=7 descendantCount=8)
    - 0 : nil
    - 1 : .parsedToken(identifier wholeText="Set" textRange=0..<3)
    - 2 : nil
    ▿ 3 : .layout(genericArgumentClause byteLength=4 descendantCount=6)
      - 0 : nil
      - 1 : .parsedToken(leftAngle wholeText="<" textRange=0..<1)
      - 2 : nil
      ▿ 3 : .layout(genericArgumentList byteLength=3 descendantCount=3)
        ▿ 0 : .layout(genericArgument byteLength=3 descendantCount=2)
          - 0 : nil
          ▿ 1 : .layout(simpleTypeIdentifier byteLength=3 descendantCount=1)
            - 0 : nil
            - 1 : .parsedToken(identifier wholeText="Int" textRange=0..<3)
            - 2 : nil
            - 3 : nil
            - 4 : nil
          - 2 : nil
          - 3 : nil
          - 4 : nil
      - 4 : nil
      - 5 : .materializedToken(rightAngle text=">" numLeadingTrivia=0 byteLength=1)
      - 6 : nil
    - 4 : nil
```

kimdv: The fixed code ends look like `let a2: [Set<Int>]>`

kimdv: `"let a4: Int1️⃣]?"` also looks like it fails, because of the `?`. 
Should I try to refactor so we can reuse the optional or force unwrap inside `parseSimpleType`

kimdv: There wan't a way to look if a raw then had an error. 
So I've added one

Sources/SwiftSyntax/Syntax.swift
```diff
@@ -246,9 +246,9 @@ public extension SyntaxProtocol {
   /// Whether the tree contained by this layout has any
   ///  - missing nodes or
   ///  - unexpected nodes or
-  ///  - tokens with a `TokenDiagnostic` of severity `error`
+  ///  - tokens with a ``TokenDiagnostic`` of severity ``TokenDiagnostic/Severity-swift.enum/error``.
```
kimdv, line 249: Changed to support DocC

kimdv:
@swift-ci please test

Sources/SwiftParser/Types.swift
```diff
@@ -1052,7 +1052,35 @@ extension Parser {
...
+
+      if !self.currentToken.isAtStartOfLine {
+        if self.at(.postfixQuestionMark) {
+          result = RawTypeSyntax(self.parseOptionalType(result))
+        }
+        if self.at(.exclamationMark) {
+          result = RawTypeSyntax(self.parseImplicitlyUnwrappedOptionalType(result))
+        }
+      }
```
ahoppen, line 1081: Should this be a loops so that we also parse two trailing question marks?

Also, I would prefer to use `self.at(TokenSpec(.postfixQuestionMark, allowAtStartOfLine: false))` instead of the nested `if` statements because it works better with the fuzzing I want to introduce in https://github.com/apple/swift-syntax/pull/1340.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -410,6 +410,23 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
     return .visitChildren
   }
 
+  public override func visit(_ node: ArrayTypeSyntax) -> SyntaxVisitorContinueKind {
+    if shouldSkip(node) {
+      return .skipChildren
+    }
+
+    if node.leftSquareBracket.presence == .missing {
```
ahoppen, line 418: We only want to emit this if the right bracket is present, right?

Sources/SwiftSyntax/Raw/RawSyntaxNodeProtocol.swift
```diff
@@ -46,6 +46,14 @@ public extension RawSyntaxNodeProtocol {
...
     return raw.byteLength == 0
   }
+
+  /// Whether the tree contained by this layout has any
+  ///  - missing nodes or
+  ///  - unexpected nodes or
+  ///  - tokens with a `TokenDiagnostic` of severity `error`
+  var hasError: Bool {
+    return raw.recursiveFlags.contains(.hasError)
+  }
```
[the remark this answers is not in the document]
ahoppen: Yes, that’s good. 👍🏽

Sources/SwiftParser/Types.swift
kimdv: Not sure I understand?
Can you give an example so I also can add it as a test case.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
ahoppen: Yes, that’s what I’m thinking.

Sources/SwiftParser/Types.swift
kimdv: Added.

Not sure if the name of the while loop is good enough 🫣

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test Windows

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test windows

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test windows

kimdv:
@ahoppen can you review this again?

ahoppen (approved):
LGTM

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test windows

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test windows

kimdv:
@swift-ci please test windows

kimdv:
@swift-ci please test windows

kimdv:
@swift-ci please test windows

PR #1450, opened 2023-03-26
Add diagnostic for wrong inheritance

kimdv:


Sources/SwiftParser/Nominals.swift
```diff
@@ -180,7 +180,13 @@ extension Parser {
     }
 
     let inheritance: RawTypeInheritanceClauseSyntax?
-    if self.at(.colon) {
+    if self.at(.colon)
+      || (self.at(.leftParen)
+        && self.withLookahead {
+          $0.consume(if: .leftParen)
+          return $0.canParseType() && ($0.at(.rightParen, .keyword(.where), .leftBrace) || $0.at(.eof))
+        })
```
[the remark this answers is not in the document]
kimdv: Function added. 

> Also: Do we actually ned to check for `where`, `leftBrace`, and `eof` here? I would expect Python-style inheritance clauses to always have a closing paren.

This is copied from the old parser. 
We can remove it and see if test cases work.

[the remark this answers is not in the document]
kimdv: The function needed to be marked at mutating because `self.at(leftParen)` is marked with mutating 

But why?
It is not mutating, is it?

Sources/SwiftParser/Nominals.swift
```diff
@@ -246,10 +262,20 @@ extension Parser {
         )
       } while keepGoing != nil && loopProgress.evaluate(currentToken)
     }
+
+    let unexpectedAfterInheritedTypeCollection: RawUnexpectedNodesSyntax?
+
+    if let rightBrace = self.consume(if: .rightParen) {
```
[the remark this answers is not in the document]
kimdv: Good idea!

Tests/SwiftParserTest/translated/RecoveryTests.swift
```diff
@@ -1570,9 +1570,12 @@ final class RecoveryTests: XCTestCase {
       class WrongInheritanceClause11️⃣(Int) {}
       """,
       diagnostics: [
-        // TODO: Old parser expected error on line 1: expected ':' to begin inheritance clause, Fix-It replacements: 30 - 31 = ': ', 34 - 35 = ''
-        DiagnosticSpec(message: "unexpected code '(Int)' in class")
-      ]
+        DiagnosticSpec(message: "expected ':' to begin inheritance clause",
+                       fixIts: ["replace '()' with ':'"])
```
kimdv, line 1574: Not sure if this replacement is correct or it should be `replace '(Int)' with `: Int``

Sources/SwiftParser/Nominals.swift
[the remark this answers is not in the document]
ahoppen: For some reason calling a non-mutating function from a mutating function has a non-trivia performance overhead (which it shouldn’t have). Thus, `at` is also marked as `mutating`.

Tests/SwiftParserTest/translated/RecoveryTests.swift
ahoppen: I think it’s fine. Especially if the type is long, the current Fix-It message might be more readable.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -1269,6 +1269,53 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
     return handleEffectSpecifiers(node)
   }
 
+  public override func visit(_ node: TypeInheritanceClauseSyntax) -> SyntaxVisitorContinueKind {
+    if shouldSkip(node) {
+      return .skipChildren
+    }
+
+    if let unexpected = node.unexpectedBeforeColon,
+      let leftParen = unexpected.tokens(satisfying: { $0.tokenKind == .leftParen }).first
```
ahoppen, line 1278: Since you are making all the unexpected nodes missing, I think this should be 

```suggestion
      let leftParen = unexpected.onlyToken(where: { $0.tokenKind == .leftParen })
```

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -1269,6 +1269,53 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
+        
+        var replaceTokens = [leftParen]
+        
+        if let rightParen = node.unexpectedAfterInheritedTypeCollection?.tokens(satisfying: { $0.tokenKind == .rightParen }).first,
+           let lastToken = node.inheritedTypeCollection.lastToken(viewMode: .sourceAccurate)
+        {
+            handledNodes += [rightParen.id]
+            changes += [
+                .makeMissing(rightParen),
+                FixIt.MultiNodeChange(.replaceTrailingTrivia(token: lastToken, newTrivia: .space)),
```
ahoppen, line 1298: I think adding the space is only necessary if the last token doesn’t have any trailing trivia. Otherwise, we might be overriding existing trailing trivia, which could be comments the user wants to keep. If you want, you can add a convenience function like `ensuringTrailingSpace` to `MultiNodeChange` for this. We might want to use that in other, existing cases as well, I haven’t looked through the code though.

kimdv: We will use that in other places. I found many examples in #1549. So I was already there thinking about adding a method somewhere we can do this easily.

kimdv: Something like this?

Sources/SwiftCompilerPluginMessageHandling/Diagnostics.swift
```diff
@@ -126,3 +135,14 @@ extension PluginMessage.Diagnostic {
...
+extension TriviaPiece {
+  var isSpaceOrTab: Bool {
+    switch self {
+    case .spaces, .tabs:
+      return true
+    default:
+      return false
+    }
+  }
+}
```
kimdv, line 148: This is copied. 
Is this fine or should we move it to `TriviaPiece` and make it public?

Sources/SwiftCompilerPluginMessageHandling/Diagnostics.swift
```diff
@@ -109,8 +109,17 @@ extension PluginMessage.Diagnostic {
...
             text = newTrivia.description
+          case .ensuringTrailingSpace(let token):
+            if !token.trailingTrivia.contains(where: { $0.isSpaceOrTab }) {
+              range = sourceManager.range(
+                of: Syntax(token),
+                from: .beforeTrailingTrivia,
+                to: .afterTrailingTrivia
+              )
+              text = token.trailingTrivia.appending(.space).description
+            }
```
kimdv, line 120: Is there a test package for `SwiftCompilerPluginMessageHandling`? 
Don't look like there is one.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
ahoppen: I think this might be fixed with my upcoming changes to BasicFormat (there’s one final PR that’s not up yet). I would like to see if we actually need the explicit insertion of trivia once that’s merged.

Sources/SwiftCompilerPluginMessageHandling/Diagnostics.swift
ahoppen: I’m adding this in https://github.com/apple/swift-syntax/pull/1558, so once that’s merged, I don’t think you need to add it anymore 😉

Sources/SwiftCompilerPluginMessageHandling/Diagnostics.swift
ahoppen: This used to be part of `SwiftCompilerPlugin` and I think it’s still tested as part of `SwiftCompilerPluginTests`. If you want to write tests for it, feel free to add a test target.

Sources/SwiftCompilerPluginMessageHandling/Diagnostics.swift
```diff
@@ -126,3 +126,14 @@ extension PluginMessage.Diagnostic {
     }
   }
 }
+
+extension TriviaPiece {
```
ahoppen, line 130: Do we still need this?

Sources/SwiftParser/Nominals.swift
```diff
@@ -304,10 +314,22 @@ extension Parser {
         )
       } while keepGoing != nil && loopProgress.evaluate(currentToken)
     }
+
+    let unexpectedAfterInheritedTypeCollection: RawUnexpectedNodesSyntax?
+
+    // If it contains a left paren, then it might be a python style inheritance clause.
+    // Left paren have been replaced with colon
+    if unexpectedBeforeColon?.containsToken(where: { $0.tokenKind == .leftParen }) == true, let rightParen = self.consume(if: .rightParen) {
```
ahoppen, line 322: I’m thinking that the code might read easier if we store a boolean variable whether we were in a Python-style inheritance clause instead of inspecting `unexpectedBeforeColon`.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -1286,6 +1286,51 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
+      let leftParen = unexpected.onlyToken(where: { $0.tokenKind == .leftParen })
+    {
+
+      var handledNodes: [SyntaxIdentifier] = [
+        leftParen.id,
+        node.colon.id,
+      ]
+
+      var changes: [FixIt.MultiNodeChange] = [
+        .makePresent(node.colon, trailingTrivia: .space),
```
ahoppen, line 1304: Do you still need to specify `trailingTrivia` with the new `BasicFormat`?

Sources/SwiftParser/Nominals.swift
kimdv: Do you think as an extension on the type or?

ahoppen: No just a local variable. Like, above you do

```swift
    var isPythonStyleInheritanceClause: Bool
    // Parse python style inheritance clause and replace parentheses with a colon
    if let leftParen = self.consume(if: .leftParen) {
      unexpectedBeforeColon = RawUnexpectedNodesSyntax([leftParen], arena: self.arena)
      colon = missingToken(.colon)
      isPythonStyleInheritanceClause = true
    } else {
      (unexpectedBeforeColon, colon) = self.expect(.colon)
      isPythonStyleInheritanceClause = false
    }
```

kimdv:
@swift-ci please test

ahoppen (approved):

PR #1453, opened 2023-03-27
Fix missing diagnostics for multiple expressions on the same line in closures, switches, and getters

### LINKED ISSUE #1452: Missing diagnostic for expressions on the same line in closures, switches, and get accessors
### Issue Kind

Other

### Source Code

```swift
let x = {a b c}
var i: Int {a b c}
switch x { case y: a b c }
```

### Description

These cases should produce diagnostics that multiple expressions cannot appear on the same line, but don't.

hamishknight:
Switch loops that used `parseCodeBlockItem` without semicolon checking over to using `parseCodeBlockItemList`.

Resolves #1452
rdar://107265306

### DIFF
--- a/Sources/SwiftParser/Declarations.swift
+++ b/Sources/SwiftParser/Declarations.swift
@@ -1768,20 +1768,14 @@ extension Parser {
             )
           }
 
-          var body = [RawCodeBlockItemSyntax]()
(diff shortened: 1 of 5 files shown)


hamishknight:
@swift-ci please test

PR #1454, opened 2023-03-27
Split DeclEffectSpecifiers into AccessorEffectSpecifiers and FunctionEffectSpecifiers

StevenWong12:
Part of #1373


Tests/SwiftParserTest/translated/EffectfulPropertiesTests.swift
```diff
@@ -188,12 +188,13 @@ final class EffectfulPropertiesTests: XCTestCase {
...
       """
       var bad2 : Int {
-        get reasync { 0 }
+        get 1️⃣reasync { 0 }
         set reasync { }
       }
       """,
       diagnostics: [
         // TODO: Old parser expected error on line 2: expected '{' to start getter definition
+        DiagnosticSpec(message: "only function declarations may be marked 'reasync'; did you mean 'async'?")
```
StevenWong12, line 197: I think it is reasonable that we produce a similar diagnostics message here as ``rethrows``, since they're, in terms of syntax analysis, basically identical, although the message is not the same as the old parser. What do you think? @ahoppen

ahoppen: Yes, I agree. Your diagnostic is way better than the old one. You can remove the TODO.

Sources/SwiftParser/SyntaxUtils.swift
```diff
@@ -80,6 +80,19 @@ extension RawUnexpectedNodesSyntax {
   init?<T1: UnexpectedNodesCombinable, T2: UnexpectedNodesCombinable, T3: UnexpectedNodesCombinable>(combining syntax1: T1, _ syntax2: T2, _ syntax3: T3, arena: __shared SyntaxArena) {
     self.init(syntax1.elements + syntax2.elements + syntax3.elements, arena: arena)
   }
+  
+  init?<T1: UnexpectedNodesCombinable, T2: UnexpectedNodesCombinable>(conbining optionalSyntax1: T1?, _ optionalSyntax2: T2?, arena: __shared SyntaxArena) {
```
ahoppen, line 84:
```suggestion
  init?<T1: UnexpectedNodesCombinable, T2: UnexpectedNodesCombinable>(combining optionalSyntax1: T1?, _ optionalSyntax2: T2?, arena: __shared SyntaxArena) {
```

ahoppen (commented):
Conceptually this boils down to a similar problem that I think we discussed in one of your other PRs: As long as we can’t represent that the effect specifier of an accessor cannot be `reasync` or `rethrows`, then no client (and “client” here includes later stages of the compiler) can assume that `asyncSpecifier` of an `AccessorDecl` cannot be `reasync` and thus needs to diagnose it themselves. 

What needs to be done here is that we need to split `DeclEffectSpecifiers` into two different nodes: `AccessorEffectSpecifiers` and `FunctionEffectSpecifiers` where `FunctionEffectSpecifiers` does not allow `rethrows` and `reasync`. For this you need to modify the declaration of `DeclEffectSpecifiers` in `CodeGeneration` and run `generate-swiftsyntax` to regenerate the code. Do you think you could try that?

StevenWong12:
Thanks for your comments! I think I only considered ``SwiftParserDiagnostics`` as the client of the parser, which makes the changes in my PRs not friendly to other clients. I think your suggestions make sense and I will make a try to modify my PRs.

StevenWong12:
> `AccessorEffectSpecifiers` and `FunctionEffectSpecifiers` where `FunctionEffectSpecifiers` does not allow rethrows and reasync. 

Just want to make sure if there is a typo and it should be 
> where `AccessorEffectSpecifiers` does not allow rethrows and reasync

ahoppen:
Yes, that was a typo indeed. Good catch

StevenWong12:
I have took ``AccessorEffectSpecifiers`` apart from ``DeclEffectSpecifiers`` and renamed ``DeclEffectSpecifiers`` to ``FunctionEffectSpecifiers``. It seems that this split also solves the TODOs in ``EffectfulPropertiesTests.swift`` without much add-ons.😀

Tests/SwiftParserTest/DeclarationTests.swift
```diff
@@ -756,6 +759,7 @@ final class DeclarationTests: XCTestCase {
     )
   }
 
+  // TODO:
```
ahoppen, line 762: Is there anything left to do here or is this just a lingering comment?

ahoppen (approved):
Thanks. This looks good. I just think you have a lingering TODO comment in your codebase.

StevenWong12:
Oh, I forgot to delete that. I've removed it and squashed my changes into a single commit.

kimdv:
@swift-ci please test

ahoppen:
@swift-ci Please test Windows

PR #1457, opened 2023-03-28
Remove check for `endOfFile` in `advance`

hamishknight:
Instead, let the lexer produce `.eof` itself. This will be needed for regex literal lexing, where the lexer may still want to produce an empty regex literal pattern when at the end of the file, before the `.eof` token.


hamishknight:
@swift-ci please test

hamishknight:
@ahoppen FWIW it seems this is a minor performance win too, parsing MovieSwiftUI:

- time: ~16ms -> ~15ms
- instructions: ~128,670,000 -> ~128,400,000

PR #1459, opened 2023-03-28
Use a different unicode character for the corner diagnostics

Azoy:
The current character there wasn't aligning properly and this new one aligns like intended.


Azoy:
@swift-ci please test

PR #1461, opened 2023-03-28
add parsing for ~ (the 'without' operator)

kavon:
see the primary PR https://github.com/apple/swift/pull/64699 for more details. This companion PR implements the parsing for the "without operator" `~` used for conformance suppression. For example, `~Copyable`.

rdar://106775103

### DIFF
--- a/CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
@@ -906,6 +906,14 @@ public let DECL_NODES: [Node] = [
       "WithTrailingComma"
     ],
     children: [
+      /// Indicates whether the 'without' operator was applied to the type to
+      /// indicate the suppression of implicit conformance to this type.
+      /// This child stores the token representing the 'without' operator.
+      Child(
+        name: "HasWithout",
+        kind: .token(choices: [.token(tokenKind: "PrefixOperatorToken")]),
+        isOptional: true
+      ),
       Child(
         name: "TypeName",
         kind: .node(kind: "Type")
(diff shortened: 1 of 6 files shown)


kavon:
@swift-ci please test

Tests/SwiftParserTest/DeclarationTests.swift
```diff
@@ -1551,6 +1551,16 @@ final class DeclarationTests: XCTestCase {
...
   }
+
+  func testSuppressedImplicitConformance() {
+    assertParse(
+      """
+      struct Hello: ~Copyable {}
+
+      enum Whatever: Int, ~ Hashable, Equatable {}
+      """
+    )
```
ahoppen, line 1562: Could you create separate `assertParse` checks for these two cases and make sure they are parsed correctly (i.e. actually have the `~` as a separate token that’s not part of the identifier?

Sources/SwiftParser/Nominals.swift
```diff
@@ -228,12 +229,16 @@ extension Parser {
             )
           )
         } else {
+          if self.currentToken.starts(with: "~") {
+            withoutToken = self.consumePrefix("~", as: .prefixOperator)
+          }
```
ahoppen, line 234: Is the `~` part of the identifier? If so, that would surprise me. I would assume that you should be able to do 

```swift
withoutToken = self.consumeIfContextualPunctuator("~")
```

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
```diff
@@ -906,6 +906,14 @@ public let DECL_NODES: [Node] = [
       "WithTrailingComma"
     ],
     children: [
+      /// Indicates whether the 'without' operator was applied to the type to
+      /// indicate the suppression of implicit conformance to this type.
+      /// This child stores the token representing the 'without' operator.
+      Child(
+        name: "HasWithout",
```
ahoppen, line 913: I would prefer this to be named `WithoutTilde`. This actually contains the tilde itself, so it’s not a boolean variable, which the `Has` implies.

kavon:
thanks alex for the review. i've addressed them in https://github.com/apple/swift-syntax/pull/1559

PR #1462, opened 2023-03-28
Only output number of executed instructions in `swift-parser-cli` on macOS

ahoppen:
`libproc.h` is not available on iOS and I’m not sure how it behaves on Linux. Since `swift-parser-cli` is only a testing utility, take the safe approach and only output the number of executed instructions on macOS.

### DIFF
--- a/Sources/_InstructionCounter/src/InstructionsExecuted.c
+++ b/Sources/_InstructionCounter/src/InstructionsExecuted.c
@@ -10,7 +10,16 @@
 //
 //===----------------------------------------------------------------------===//
 
+#if __APPLE__
+#include <TargetConditionals.h>
+#if TARGET_OS_MAC && !TARGET_OS_IPHONE
+#define TARGET_IS_MACOS 1
+#endif
+#endif
+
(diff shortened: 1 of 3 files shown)


ahoppen:
@swift-ci Please test

Sources/_InstructionCounter/src/InstructionsExecuted.c
```diff
@@ -21,3 +30,8 @@ uint64_t getInstructionsExecuted() {
   }
   return 0;
 }
+#else
+uint64_t getInstructionsExecuted() {
```
bnbarham, line 34: Would be nice to see if we can do something cross platform here, but fine for now I suppose 🤷

bnbarham (approved):

Sources/_InstructionCounter/src/InstructionsExecuted.c
ahoppen: Yeah, I agree. But let’s wait to see if a need comes up first since this is just a testing utility anyway.

PR #1464, opened 2023-03-29
Fixed testAvailabilityQueryUnavailability34a

CippoX:
I'm clearly doing something wrong because Source Control keeps breaking down. Problems aside, I've implemented the changes you suggested and now it looks way better. Unfortunately I am still having trouble fixing this edge case 
`if #available(*) == false && other_condition`

### DIFF
--- a/Sources/SwiftParser/Statements.swift
+++ b/Sources/SwiftParser/Statements.swift
@@ -187,13 +187,14 @@ extension Parser {
     var loopProgress = LoopProgressCondition()
     repeat {
       let condition = self.parseConditionElement(lastBindingKind: elements.last?.condition.as(RawOptionalBindingConditionSyntax.self)?.bindingKeyword)
-      let unexpectedBeforeKeepGoing: RawUnexpectedNodesSyntax?
+      var unexpectedBeforeKeepGoing: RawUnexpectedNodesSyntax? = nil
+      if let equalOperator = self.consumeIfContextualPunctuator("=="), let falseKeyword = self.consume(if: .keyword(.false)) {
+        unexpectedBeforeKeepGoing = RawUnexpectedNodesSyntax([equalOperator, falseKeyword], arena: self.arena)
+      }
       keepGoing = self.consume(if: .comma)
       if keepGoing == nil, let andOperator = self.consumeIfContextualPunctuator("&&") {
-        unexpectedBeforeKeepGoing = RawUnexpectedNodesSyntax([andOperator], arena: self.arena)
+        unexpectedBeforeKeepGoing = RawUnexpectedNodesSyntax(combining: unexpectedBeforeKeepGoing, andOperator, arena: self.arena)
         keepGoing = missingToken(.comma)
-      } else {
-        unexpectedBeforeKeepGoing = nil
       }
       elements.append(
         RawConditionElementSyntax(
--- a/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
+++ b/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
@@ -14,6 +14,22 @@ import SwiftDiagnostics
 import SwiftParser
 @_spi(RawSyntax) import SwiftSyntax
 
+fileprivate func getTokens(between first: TokenSyntax, and second: TokenSyntax) -> [TokenSyntax] {
+  var tokens: [TokenSyntax] = []
+  var currentToken = first
+
+  while currentToken != second {
+    tokens.append(currentToken)
+    guard let nextToken = currentToken.nextToken(viewMode: .sourceAccurate) else {
+      assertionFailure("second Token must occur after first Token")
+      return tokens
+    }
+    currentToken = nextToken
+  }
+  tokens.append(second)
+  return tokens
+}
+
 fileprivate extension TokenSyntax {
   /// Assuming this token is a `poundAvailableKeyword` or `poundUnavailableKeyword`
   /// returns the opposite keyword.
@@ -239,7 +255,7 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
           unexpectedTokenCondition: isOfSameKind,
           correctTokens: [specifier],
           message: { _ in misspelledError },
-          moveFixIt: { ReplaceTokensFixIt(replaceTokens: $0, replacement: specifier) },
+          moveFixIt: { ReplaceTokensFixIt(replaceTokens: $0, replacements: [specifier]) },
           removeRedundantFixIt: { RemoveRedundantFixIt(removeTokens: $0) }
         )
       }
@@ -438,7 +454,7 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
         unexpectedTokenCondition: { $0.text == "||" },
         correctTokens: [node.trailingComma],
         message: { _ in .joinPlatformsUsingComma },
-        moveFixIt: { ReplaceTokensFixIt(replaceTokens: $0, replacement: trailingComma) }
+        moveFixIt: { ReplaceTokensFixIt(replaceTokens: $0, replacements: [trailingComma]) }
       )
     }
     return .visitChildren
@@ -467,13 +483,42 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
     if shouldSkip(node) {
       return .skipChildren
     }
+    if let unexpected = node.unexpectedBetweenConditionAndTrailingComma,
+      let availability = node.condition.as(AvailabilityConditionSyntax.self),
+      let (_, falseKeyword) = unexpected.twoTokens(
+        firstSatisfying: { $0.tokenKind == .binaryOperator("==") },
+        secondSatisfying: { $0.tokenKind == .keyword(.false) }
+      )
+    {
+      // Diagnose #available used as an expression
+      let negatedAvailabilityKeyword = availability.availabilityKeyword.negatedAvailabilityKeyword
+      let negatedCoditionElement = ConditionElementSyntax(
+        condition: .availability(availability.with(\.availabilityKeyword, negatedAvailabilityKeyword)),
+        trailingComma: node.trailingComma
+      )
+      if let negatedAvailability = negatedCoditionElement.condition.as(AvailabilityConditionSyntax.self) {
+        addDiagnostic(
+          unexpected,
+          AvailabilityConditionAsExpression(availabilityToken: availability.availabilityKeyword, negatedAvailabilityToken: negatedAvailabilityKeyword),
+          fixIts: [
+            FixIt(
+              message: ReplaceTokensFixIt(replaceTokens: getTokens(between: availability.availabilityKeyword, and: falseKeyword), replacements: getTokens(between: negatedAvailability.availabilityKeyword, and: negatedAvailability.rightParen)),
+              changes: [
+                .replace(oldNode: Syntax(node), newNode: Syntax(negatedCoditionElement))
+              ]
+            )
+          ],
+          handledNodes: [unexpected.id]
+        )
+      }
+    }
     if let trailingComma = node.trailingComma {
       exchangeTokens(
         unexpected: node.unexpectedBetweenConditionAndTrailingComma,
         unexpectedTokenCondition: { $0.text == "&&" },
         correctTokens: [node.trailingComma],
         message: { _ in .joinConditionsUsingComma },
-        moveFixIt: { ReplaceTokensFixIt(replaceTokens: $0, replacement: trailingComma) }
+        moveFixIt: { ReplaceTokensFixIt(replaceTokens: $0, replacements: [trailingComma]) }
       )
     }
     return .visitChildren
@@ -563,7 +608,7 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
         .expectedCommaInWhereClause,
         fixIts: [
           FixIt(
-            message: ReplaceTokensFixIt(replaceTokens: [token], replacement: .commaToken()),
+            message: ReplaceTokensFixIt(replaceTokens: [token], replacements: [.commaToken()]),
             changes: [
               .makeMissing(token),
               .makePresent(trailingComma),
@@ -698,7 +743,7 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
         .typeParameterPackEllipsis,
         fixIts: [
           FixIt(
-            message: ReplaceTokensFixIt(replaceTokens: [unexpectedEllipsis], replacement: .keyword(.each)),
+            message: ReplaceTokensFixIt(replaceTokens: [unexpectedEllipsis], replacements: [.keyword(.each)]),
             changes: [
               .makeMissing(unexpected),
               .makePresent(each, trailingTrivia: .space),
@@ -714,7 +759,7 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
         unexpectedTokenCondition: { $0.tokenKind == .keyword(.class) },
         correctTokens: [inheritedTypeName],
         message: { _ in StaticParserError.classConstraintCanOnlyBeUsedInProtocol },
-        moveFixIt: { ReplaceTokensFixIt(replaceTokens: $0, replacement: inheritedTypeName) }
+        moveFixIt: { ReplaceTokensFixIt(replaceTokens: $0, replacements: [inheritedTypeName]) }
       )
     }
     return .visitChildren
@@ -745,7 +790,7 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
             NegatedAvailabilityCondition(avaialabilityCondition: availability, negatedAvailabilityKeyword: negatedAvailabilityKeyword),
             fixIts: [
               FixIt(
-                message: ReplaceTokensFixIt(replaceTokens: [operatorToken, availability.availabilityKeyword], replacement: negatedAvailabilityKeyword),
+                message: ReplaceTokensFixIt(replaceTokens: [operatorToken, availability.availabilityKeyword], replacements: [negatedAvailabilityKeyword]),
                 changes: [
                   .replace(oldNode: Syntax(conditionElement), newNode: Syntax(negatedCoditionElement))
                 ]
@@ -777,7 +822,7 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
             StaticParserError.unexpectedPoundElseSpaceIf,
             fixIts: [
               FixIt(
-                message: ReplaceTokensFixIt(replaceTokens: unexpectedTokens, replacement: clause.poundKeyword),
+                message: ReplaceTokensFixIt(replaceTokens: unexpectedTokens, replacements: [clause.poundKeyword]),
                 changes: [
                   .makeMissing(unexpectedBeforePoundKeyword, transferTrivia: false),
                   .makePresent(clause.poundKeyword, leadingTrivia: unexpectedBeforePoundKeyword.leadingTrivia),
@@ -821,7 +866,7 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
         .expectedAssignmentInsteadOfComparisonOperator,
         fixIts: [
           FixIt(
-            message: ReplaceTokensFixIt(replaceTokens: [.binaryOperator("==")], replacement: node.equal),
+            message: ReplaceTokensFixIt(replaceTokens: [.binaryOperator("==")], replacements: [node.equal]),
             changes: [.makeMissing(unexpected), .makePresent(node.equal, leadingTrivia: [])]
           )
         ],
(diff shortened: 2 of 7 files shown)


Sources/SwiftParserDiagnostics/SyntaxExtensions.swift
```diff
@@ -39,6 +39,18 @@ extension UnexpectedNodesSyntax {
       return nil
     }
   }
+
+  /// If this only contains two tokens, one satisfying `firstCondition`, and one satisfying `secondCondition`,
+  /// return these tokens as a tuple, otherwise return `nil`. The function is not order sensitive
+  func twoTekens(firstSatisfying firstCondition: (TokenSyntax) -> Bool, secondSatisfying secondCondition: (TokenSyntax) -> Bool) -> (first: TokenSyntax, second: TokenSyntax)? {
```
ahoppen, line 45: Typo
```suggestion
  func twoTokens(firstSatisfying firstCondition: (TokenSyntax) -> Bool, secondSatisfying secondCondition: (TokenSyntax) -> Bool) -> (first: TokenSyntax, second: TokenSyntax)? {
```

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -463,6 +463,35 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
+            message: ReplaceTokensFixIt(replaceTokens: [availabilityKeyword], replacement: negatedAvailabilityKeyword),
+            changes: [
+              .replace(oldNode: Syntax(availabilityKeyword), newNode: Syntax(negatedAvailabilityKeyword)),
+              .replaceTrailingTrivia(token: availability.rightParen, newTrivia: .zero),
+            ]
+          ),
+          FixIt(
+            message: RemoveNodesFixIt([comparisonOperator, falseKeyword]),
+            changes: .makeMissing([comparisonOperator, falseKeyword])
+          ),
```
ahoppen, line 490: This should be a single `FixIt` with all the changes combined. As it stands right now we offer you two options to resolve the issue: Either to replace `#available` by `#unavailable` or by removing `== false`. But obviously, to resolve the error, you need to perform both changes.

Sources/SwiftParserDiagnostics/SyntaxExtensions.swift
```diff
@@ -39,6 +39,18 @@ extension UnexpectedNodesSyntax {
       return nil
     }
   }
+
+  /// If this only contains two tokens, one satisfying `firstCondition`, and one satisfying `secondCondition`,
+  /// return these tokens as a tuple, otherwise return `nil`. The function is not order sensitive
```
ahoppen, line 44: >  The function is not order sensitive

I think it should be order sensitive, that’s the whole point of the function, right? See https://github.com/apple/swift-syntax/pull/1446/files#r1150941427

CippoX:
@ahoppen I've fixed everything you suggested, the fact that I had to introduce two new constructs `getTokenRange` and `ReplaceWithMultipleTokensFixIt`, looks a bit odd to me, is there a batter way?

CippoX:
The only way I see to fix this specific case `if #available(*) == false && other_condition`, is to assign `unexpectedBeforeKeepGoing` to another variable like
```swift
public struct RawConditionElementSyntax: RawSyntaxNodeProtocol {
  ...
  public var unexpectedInsteadOfTrailingComma: RawUnexpectedNodesSyntax? {...}
  ...
}
```

Sources/SwiftParserDiagnostics/SyntaxExtensions.swift
```diff
@@ -40,12 +40,13 @@ extension UnexpectedNodesSyntax {
...
+  /// return these tokens as a tuple, otherwise return `nil`.
+  func twoTokens(firstSatisfying firstCondition: (TokenSyntax) -> Bool,
+                 secondSatisfying secondCondition: (TokenSyntax) -> Bool) -> (first: TokenSyntax, second: TokenSyntax)? {
     let firstTokens = tokens(satisfying: firstCondition)
     let secondTokens = tokens(satisfying: secondCondition)
-    if firstTokens.count == 1, secondTokens.count == 1, firstTokens.first != secondTokens.first {
+    if firstTokens.count == 1, secondTokens.count == 1, self.first?.as(TokenSyntax.self) == firstTokens.first, firstTokens.first != secondTokens.first {
       return (firstTokens[0], secondTokens[0])
     } else {
       return nil
```
ahoppen: All of this seem fairly roundabout. Wouldn’t it be simpler to do something like the following?

```swift
let sourceAccurateChildren = self.children(viewMode: .sourceAccurate)
guard sourceAccurateChildren.count == 2 else {
  return nil
}
guard firstCondition(sourceAccurateChildren[0]) && secondCondition(sourceAccurateChildren[1]) else {
  return nil
}
return (sourceAccurateChildren[0], sourceAccurateChildren[1])
```

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -465,32 +476,32 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
+        condition: .availability(availability.with(\.availabilityKeyword, negatedAvailabilityKeyword)),
+        trailingComma: node.trailingComma
       )
+      if let negatedAvailability = negatedCoditionElement.condition.as(AvailabilityConditionSyntax.self) {
+        addDiagnostic(
+          unexpected,
+          AvailabilityConditionAsExpression(availabilityToken: availability.availabilityKeyword, negatedAvailabilityToken: negatedAvailabilityKeyword),
+          fixIts: [
+            FixIt(
+              message: ReplaceWithMultipleTokensFixIt(replaceTokens: getTokenRange(availability.availabilityKeyword, falseKeyword), replacements: getTokenRange(negatedAvailability.availabilityKeyword, negatedAvailability.rightParen)),
```
ahoppen, line 496: Why do you need to do `getTokenRange` here? Can’t you just do `[availability.availabilityKeyword, comparisonOperator, falseKeyword]`?

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -724,7 +735,7 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
         if let prefixOperatorExpr = node.parent?.as(PrefixOperatorExprSyntax.self),
           let operatorToken = prefixOperatorExpr.operatorToken,
           operatorToken.text == "!",
-          let conditionElement = prefixOperatorExpr.parent?.as(ConditionElementSyntax.self)
+           let conditionElement = prefixOperatorExpr.parent?.as(ConditionElementSyntax.self)
```
ahoppen: I think you accidentally added a space here.

Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift
```diff
@@ -611,3 +611,13 @@ public struct ReplaceTokensFixIt: ParserFixIt {
...
+
+public struct ReplaceWithMultipleTokensFixIt: ParserFixIt {
+  public let replaceTokens: [TokenSyntax]
+  
+  public let replacements: [TokenSyntax]
+  
+  public var message: String {
+    "replace \(nodesDescription(replaceTokens, format: false)) with \(nodesDescription(replacements, format: false))"
+  }
+}
```
ahoppen: Instead of introducing a new Fix-It you could just generalize the existing `ReplaceTokensFixIt` to allow multiple tokens for `replacement`.

Tests/SwiftParserTest/translated/AvailabilityQueryUnavailabilityTests.swift
```diff
@@ -432,6 +432,8 @@ final class AvailabilityQueryUnavailabilityTests: XCTestCase {
       """
     )
   }
+  
+
```
ahoppen: Accidentally added those newlines?

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
CippoX: I've already tried what you suggested and in this case the diagnostic seems to be
`replace '#available== false' with '#unavailable(*)'`
instead of
`replace '#available(*) == false' with '#unavailable(*)'`

ahoppen: Oh, I see. I forgot about the `(*)`. In that case is a very good solution. I’ve got a few minor comments on it:
- It’s not really returning a range, so I would name it `getTokens(between: TokenSyntax, and: TokenSyntax)`
- I would make it a top-level function instead of a member of `ParseDiagnosticGenerator`. Also I would mark it as `fileprivate` so we explicitly introduce new users of it
- You can also use this function in `visit(_ node: ForInStmtSyntax)` and thereby remove the `FIXME` in there.

Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift
ahoppen: You’ve hereby been granted permission 😄

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
CippoX: Thank you, indeed that `FIXME` gave me the idea, I was already looking at it, that would have been my next question :)

CippoX:
If you don't mind, I would like to resolve the `visit(_ node: ForInStmtSyntax)` `FIXME` in another PR.

ahoppen (commented):
Thank you. The PR looks veery good now. I’ve got three local comments but they should be straightforward to fix. After this we can merge it. 🎉

Sources/SwiftParser/Statements.swift
```diff
@@ -187,13 +187,14 @@ extension Parser {
...
     repeat {
       let condition = self.parseConditionElement()
-      let unexpectedBeforeKeepGoing: RawUnexpectedNodesSyntax?
+      var unexpectedBeforeKeepGoing: RawUnexpectedNodesSyntax? = nil
+      if let equalOperator = self.consumeIfContextualPunctuator("=="), let falseKeyword = self.consume(if: .keyword(.false)) {
+        unexpectedBeforeKeepGoing = RawUnexpectedNodesSyntax([equalOperator, falseKeyword], arena: self.arena)
+      }
       keepGoing = self.consume(if: .comma)
       if keepGoing == nil, let andOperator = self.consumeIfContextualPunctuator("&&") {
         unexpectedBeforeKeepGoing = RawUnexpectedNodesSyntax([andOperator], arena: self.arena)
```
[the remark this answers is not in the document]
CippoX: Wouldn't this break the assumption that `UnexpectedNodesSyntax` contains only two tokens, which is necessary to diagnose `== false`?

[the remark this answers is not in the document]
ahoppen: Yes, it would, but round-trippability of the source file is a hard requirement we cannot break. And to be frank, I think it’s fairly unlikely that somebody writes the example above, so I don’t care too much. Having said that, if you do find a good way to produce good diagnostics for the case, I wouldn’t object.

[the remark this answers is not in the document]
CippoX: Fixing this particular case seems not trivial to me, so since `UnexpectedNodesSyntax` can contain more than two tokens, should the function `twoTokens` do not check this requirement?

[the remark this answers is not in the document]
CippoX: Or should I just let the parser produce `unexpected code '== false &&' in 'if' statement` & `expected ',' in 'if' statement`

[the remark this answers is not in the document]
ahoppen: I think producing `unexpected code '== false &&' in 'if' statement` & `expected ',' in 'if' statement` would be good enough for now.

ahoppen (approved):
Thank you for improving the diagnostic. This looks good.

ahoppen:
@CippoX Could you rebase your PR onto `main`. There appears is a merge conflict.

ahoppen:
CI failed in one test case, I was curious what it was, debugged it and this diff should fix the problem. Could you apply it to your PR?

```diff
diff --git a/Sources/SwiftParserDiagnostics/MissingNodesError.swift b/Sources/SwiftParserDiagnostics/MissingNodesError.swift
index 9e17832a6..7faffba2d 100644
--- a/Sources/SwiftParserDiagnostics/MissingNodesError.swift
+++ b/Sources/SwiftParserDiagnostics/MissingNodesError.swift
@@ -112,13 +112,17 @@ func nodesDescription<SyntaxType: SyntaxProtocol>(_ nodes: [SyntaxType], format:
 func nodesDescriptionAndCommonParent<SyntaxType: SyntaxProtocol>(_ nodes: [SyntaxType], format: Bool) -> (commonAncestor: Syntax?, description: String) {
   let missingSyntaxNodes = nodes.map(Syntax.init)
 
-  // If all tokens in the parent are missing, return the parent type name.
-  if let commonAncestor = findCommonAncestor(missingSyntaxNodes),
+  // If all tokens in the parent are missing, return the parent type name unless
+  // we are replacing by a single token that has explicit text, in which case we
+  // return that explicit text.
+  if !isIdentifierWithCustomName,
+    let commonAncestor = findCommonAncestor(missingSyntaxNodes),
     commonAncestor.isMissingAllTokens,
     let firstToken = commonAncestor.firstToken(viewMode: .all),
     let lastToken = commonAncestor.lastToken(viewMode: .all),
     missingSyntaxNodes.contains(Syntax(firstToken)),
-    missingSyntaxNodes.contains(Syntax(lastToken))
+    missingSyntaxNodes.contains(Syntax(lastToken)),
+    nodes.only?.as(TokenSyntax.self)?.text != nil
   {
     if let nodeTypeName = commonAncestor.nodeTypeNameForDiagnostics(allowBlockNames: true) {
       return (commonAncestor, nodeTypeName)
```

CippoX:
sure

CippoX:
@ahoppen where is `isIdentifierWithCustomName` defined?

ahoppen:
Oh, sorry, copy-pasted the wrong diff. It should have been

```diff
diff --git a/Sources/SwiftParserDiagnostics/MissingNodesError.swift b/Sources/SwiftParserDiagnostics/MissingNodesError.swift
index 9e17832a6..7faffba2d 100644
--- a/Sources/SwiftParserDiagnostics/MissingNodesError.swift
+++ b/Sources/SwiftParserDiagnostics/MissingNodesError.swift
@@ -112,13 +112,17 @@ func nodesDescription<SyntaxType: SyntaxProtocol>(_ nodes: [SyntaxType], format:
 func nodesDescriptionAndCommonParent<SyntaxType: SyntaxProtocol>(_ nodes: [SyntaxType], format: Bool) -> (commonAncestor: Syntax?, description: String) {
   let missingSyntaxNodes = nodes.map(Syntax.init)
 
  if let commonAncestor = findCommonAncestor(missingSyntaxNodes),
-  // If all tokens in the parent are missing, return the parent type name.
+  // If all tokens in the parent are missing, return the parent type name unless
+  // we are replacing by a single token that has explicit text, in which case we
+  // return that explicit text.
     commonAncestor.isMissingAllTokens,
     let firstToken = commonAncestor.firstToken(viewMode: .all),
     let lastToken = commonAncestor.lastToken(viewMode: .all),
     missingSyntaxNodes.contains(Syntax(firstToken)),
-    missingSyntaxNodes.contains(Syntax(lastToken))
+    missingSyntaxNodes.contains(Syntax(lastToken)),
+    nodes.only?.as(TokenSyntax.self)?.text != nil
   {
     if let nodeTypeName = commonAncestor.nodeTypeNameForDiagnostics(allowBlockNames: true) {
       return (commonAncestor, nodeTypeName)
```

CippoX:
Done everything, and thanks for debugging it!

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

CippoX:
Is `testInvalid29` the test that was causing the issue? Because if I understood the log correctly and that's the problem, the bug seems to be caused by the way I've changed the `ReplaceTokensFixIt` function.

So I suggest checking if `replacements` contains one element and then providing the appropriate FixIt. The alternative would be restoring the old function and introducing a new function that takes multiple `replacements`.

ahoppen:
> Is `testInvalid29` the test that was causing the issue? Because if I understood the log correctly and that's the problem, the bug seems to be caused by the way I've changed the `ReplaceTokensFixIt` function.

Yes, correct.

> So I suggest checking if `replacements` contains one element and then providing the appropriate FixIt. The alternative would be restoring the old function and introducing a new function that takes multiple `replacements`.

I thought that’s essentially what I did with my diff but looks like I was wrong and didn’t run all tests to verify, sorry about that. I don’t think we should add a new function if you could investigate in finding some way to make the current one behave in both scenarios, that’d be great.

CippoX:
@ahoppen I've pushed my my idea, let me know

CippoX:
I know my solution is very amateur, but it looks quite hard to change `nodesDescriptionAndCommonParent` without breaking something else.

ahoppen:
I just looked into it further and it looks like my previous solution was really close, I just forgot to check for empty string instead of `nil`. The following diff should get all tests passing and has a generic implementation in `nodesDescriptionAndCommonParent`

```diff
diff --git a/Sources/SwiftParserDiagnostics/MissingNodesError.swift b/Sources/SwiftParserDiagnostics/MissingNodesError.swift
index b5f1215e..3e6b9f34 100644
--- a/Sources/SwiftParserDiagnostics/MissingNodesError.swift
+++ b/Sources/SwiftParserDiagnostics/MissingNodesError.swift
@@ -112,6 +112,13 @@ func nodesDescription<SyntaxType: SyntaxProtocol>(_ nodes: [SyntaxType], format:
 func nodesDescriptionAndCommonParent<SyntaxType: SyntaxProtocol>(_ nodes: [SyntaxType], format: Bool) -> (commonAncestor: Syntax?, description: String) {
   let missingSyntaxNodes = nodes.map(Syntax.init)
 
+  let isOnlyTokenWithNonMissingText: Bool
+  if let token = nodes.only?.as(TokenSyntax.self) {
+    isOnlyTokenWithNonMissingText = token.text != ""
+  } else {
+    isOnlyTokenWithNonMissingText = false
+  }
+
   // If all tokens in the parent are missing, return the parent type name unless
   // we are replacing by a single token that has explicit text, in which case we
   // return that explicit text.
@@ -120,7 +127,8 @@ func nodesDescriptionAndCommonParent<SyntaxType: SyntaxProtocol>(_ nodes: [Synta
     let firstToken = commonAncestor.firstToken(viewMode: .all),
     let lastToken = commonAncestor.lastToken(viewMode: .all),
     missingSyntaxNodes.contains(Syntax(firstToken)),
-    missingSyntaxNodes.contains(Syntax(lastToken))
+    missingSyntaxNodes.contains(Syntax(lastToken)),
+    !isOnlyTokenWithNonMissingText
   {
     if let nodeTypeName = commonAncestor.nodeTypeNameForDiagnostics(allowBlockNames: true) {
       return (commonAncestor, nodeTypeName)
diff --git a/Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift b/Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift
index 52198c2b..fd2af952 100644
--- a/Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift
+++ b/Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift
@@ -614,10 +614,6 @@ public struct ReplaceTokensFixIt: ParserFixIt {
   public let replacements: [TokenSyntax]
 
   public var message: String {
-    if replacements.count == 1 {
-      return "replace \(nodesDescription(replaceTokens, format: false)) with '\(replacements[0].text)'"
-    } else {
-      return "replace \(nodesDescription(replaceTokens, format: false)) with \(nodesDescription(replacements, format: false))"
-    }
+    return "replace \(nodesDescription(replaceTokens, format: false)) with \(nodesDescription(replacements, format: false))"
   }
 }

```

CippoX:
Thanks!

ahoppen (approved):
Thank you. This looks great now!

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
Looks like this PR raced with the one that changed `replacement` to `replacements` (I can’t find it right now). Could you rebase and make sure that everything builds locally?

kimdv:
@swift-ci please test

CippoX:
@kimdv please wait, looks there's something wrong 🥲

CippoX:
Honestly, I'm not sure what happened, but everything seems to be fine now. I've also removed a warning introduced by my previous PR.

kimdv:
Let's see what the CI says 🤞

CippoX:
Finally 🎉

PR #1467, opened 2023-03-30
Added missing fix its

kimdv:
Not sure if this is needed, but I tought it might be a good idea to compare the fixits we expect and what we actually get.

So just got some examples of where it was missing.
There is still a few hundred places missing.

Also tought that I adding `fixedSource` so we ensure they works in as many scenarios as possible.

### DIFF
--- a/Tests/SwiftParserTest/Assertions.swift
+++ b/Tests/SwiftParserTest/Assertions.swift
@@ -242,7 +242,7 @@ struct DiagnosticSpec {
   let notes: [NoteSpec]?
   /// If not `nil`, assert that the diagnostic contains fix-its with these messages.
   /// Use the `fixedSource` parameter on `AssertParse` to check that applying the Fix-It yields the expected result.
-  let fixIts: [String]?
+  let fixIts: [String]
 
   /// The file and line at which this `DiagnosticSpec` was created, so that assertion failures can be reported at its location.
   let file: StaticString
@@ -255,7 +255,7 @@ struct DiagnosticSpec {
     severity: DiagnosticSeverity = .error,
     highlight: String? = nil,
     notes: [NoteSpec]? = nil,
-    fixIts: [String]? = nil,
+    fixIts: [String] = [],
     file: StaticString = #file,
     line: UInt = #line
   ) {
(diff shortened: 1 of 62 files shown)


Tests/SwiftParserTest/Assertions.swift
```diff
@@ -594,6 +594,7 @@ func assertParse<S: SyntaxProtocol>(
     )
   } else {
     for (diag, expectedDiag) in zip(diags, expectedDiagnostics) {
+      XCTAssertEqual(diag.fixIts.map { $0.message.message }, expectedDiag.fixIts ?? [], file: file, line: expectedDiag.line)
```
ahoppen, line 597: We should be checking this exact same thing in
https://github.com/apple/swift-syntax/blob/05356239a2b5b3f5c54653a70ec47cb17d5afd14/Tests/SwiftParserTest/Assertions.swift#L437-L446

kimdv:
An assertion message could look like `testSwitch3(): XCTAssertEqual failed: ("["insert \'{}\'"]") is not equal to ("[]")`
Should that be improved ?

Tests/SwiftParserTest/Assertions.swift
kimdv: Changed the logic a bit.

ahoppen:
I think that’s good enough but if you want to improve it, I won’t stop you.

kimdv:
@ahoppen we need a bit of planning when merging this. 
When it's approved I will rebase on latest main and trigger CI right after. 

I think we then should wait a bit on merging other things, to avoid test start failing because of missing fix-its.

kimdv:
@swift-ci please test Windows

Tests/SwiftParserTest/Assertions.swift
```diff
@@ -434,7 +434,18 @@ func assertDiagnostic<T: SyntaxProtocol>(
       }
     }
   }
-  if let fixIts = spec.fixIts {
+
+  if diag.fixIts.count != (spec.fixIts?.count ?? 0) {
```
ahoppen, line 438: Should we just make `spec.fixIts` non-optional and default to `[]`? That way we don’t need to handle with optionals here.

ahoppen (approved):
Thank you. Adding all those expected Fix-Its must have taken you some time 😆 It’s a great change to better testing. I like it. 

One minor comment, otherwise you can try landing it over the weekend.

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test windows

PR #1469, opened 2023-03-30
[Macros] Centralize "unwrapping" of errors into diagnostics.

DougGregor:
We had a few different implementations of "take an error thrown from a
macro and turn it into diagnostics". Improve the implementation in
`MacroExpansionContext.addDiagnostics` to properly handle both
`DiagnosticsError` and `DiagnosticsMessage`, as well as fall back
gracefully if either of those is missing an error diagnostic. Then,
use that implementation for the compiler plugin message handler.

Another part of fixing rdar://107289985, where diagnostics provided by
throwing `DiagnosticsError` aren't coming through well.

### DIFF
--- a/Sources/SwiftSyntaxMacros/MacroExpansionContext.swift
+++ b/Sources/SwiftSyntaxMacros/MacroExpansionContext.swift
@@ -189,22 +189,23 @@ private struct ThrownErrorDiagnostic: DiagnosticMessage {
 extension MacroExpansionContext {
   /// Add diagnostics from the error thrown during macro expansion.
   public func addDiagnostics<S: SyntaxProtocol>(from error: Error, node: S) {
-    guard let diagnosticsError = error as? DiagnosticsError else {
-      diagnose(
-        Diagnostic(
-          node: Syntax(node),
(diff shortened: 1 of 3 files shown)


DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test macOS

Sources/SwiftSyntaxMacros/MacroExpansionContext.swift
```diff
@@ -189,22 +189,23 @@ private struct ThrownErrorDiagnostic: DiagnosticMessage {
...
-    for diagnostic in diagnosticsError.diagnostics {
+    // Emit the diagnostics.
+    for diagnostic in diagnostics {
       diagnose(diagnostic)
     }
 
-    // handle possibility that none of the diagnostics was an error
-    if !diagnosticsError.diagnostics.contains(
+    // Handle possibility that none of the diagnostics was an error.
+    if !diagnostics.contains(
```
bnbarham, line 208: Nice, this is actually used now for the `DiagnosticMessage` case 😅!

bnbarham (approved):

PR #1470, opened 2023-03-30
Simplify MacroDecl syntax to match the accepted proposals 

DougGregor:
* *Explanation*: Early macro proposals differentiated syntax between "function-like" and "value-like" macros. However, we settled on always making macros "function-like", so remove the "value-like" syntax, which causes the parser and syntax tree to be more complex than necessary (and not represent the actual language grammar).
* Scope: Parser and syntax tree are affected, but no clients ever used this syntax.
* Risk: Very low; it's eliminating parsing and syntax tree support for an experimental syntax that is no longer in use.
* Reviewer: @ahoppen


DougGregor:
@swift-ci please test

PR #1471, opened 2023-03-30
[Macros] Centralize "unwrapping" of errors into diagnostics

DougGregor:
We had a few different implementations of "take an error thrown from a
macro and turn it into diagnostics". Improve the implementation in
`MacroExpansionContext.addDiagnostics` to properly handle both
`DiagnosticsError` and `DiagnosticsMessage`, as well as fall back
gracefully if either of those is missing an error diagnostic. Then,
use that implementation for the compiler plugin message handler.

Another part of fixing rdar://107289985, where diagnostics provided by
throwing `DiagnosticsError` aren't coming through well.


DougGregor:
@swift-ci please test

PR #1484, opened 2023-04-02
Add diagnostic for dictionary

kimdv:
Depends on #1449 and #1382

### DIFF
--- a/Sources/SwiftParser/Types.swift
+++ b/Sources/SwiftParser/Types.swift
@@ -1069,6 +1069,30 @@ extension Parser {
             arena: self.arena
           )
         )
+      } else if self.at(.colon) {
+        var lookahead = self.lookahead()
+
+        // We only want to continue with a dictionary if we can parse a colon and a simpletype.
+        // Otherwise we can get a wrong diagnostic if we get a Python-style function declaration.
+        guard lookahead.consume(if: .colon) != nil && lookahead.canParseSimpleType() else {
+          return result
+        }
+
+        let (unexpectedBeforeColon, colon) = self.expect(.colon)
+        let secondType = self.parseSimpleType()
+        let rightSquareBracket = self.consume(if: .rightSquareBracket) ?? self.missingToken(.rightSquareBracket)
+
+        result = RawTypeSyntax(
+          RawDictionaryTypeSyntax(
+            leftSquareBracket: self.missingToken(.leftSquareBracket),
+            keyType: result,
+            unexpectedBeforeColon,
+            colon: colon,
+            valueType: secondType,
+            rightSquareBracket: rightSquareBracket,
+            arena: self.arena
+          )
+        )
       }
 
       var loopProgress = LoopProgressCondition()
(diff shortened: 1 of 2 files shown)


Tests/SwiftParserTest/translated/RecoveryTests.swift
```diff
@@ -810,6 +810,7 @@ final class RecoveryTests: XCTestCase {
   }
 
   func testRecovery63() {
+
```
ahoppen, line 813: Superfluous newline?

Sources/SwiftParser/Types.swift
```diff
@@ -1069,6 +1069,20 @@ extension Parser {
             arena: self.arena
           )
         )
+      } else if let colon = self.consume(if: .colon) {
```
ahoppen, line 1072: I’m a little worried that we enter into this path and give a weird diagnostic if somebody tries to do a Python-style function declaration like the following. It would probably be better to only recover to a dictionary type if we have colon followed by a type. And yes, I know that this doesn’t match the behavior of the old parser.

```swift
func foo() -> Int:
  // some Code
```

Tests/SwiftParserTest/translated/RecoveryTests.swift
```diff
@@ -831,7 +832,8 @@ final class RecoveryTests: XCTestCase {
           message: "expected ':' in type annotation",
           fixIts: ["insert ':'"]
         ),
-        DiagnosticSpec(locationMarker: "3️⃣", message: #"unexpected code ': Int = ""' before function"#),
+        DiagnosticSpec(
+            locationMarker: "3️⃣", message: #"unexpected code ': Int = ""' before function"#),
```
ahoppen, line 836: Unrelated change?

Sources/SwiftParser/Types.swift
kimdv: Something like this. 
Also added a test case for the scenario above.

kimdv:
Depends on #1580

Sources/SwiftParser/Types.swift
```diff
@@ -1069,6 +1069,30 @@ extension Parser {
             arena: self.arena
           )
         )
+      } else if self.at(.colon) {
+        var lookahead = self.lookahead()
+
+        // We only want to continue with a dictionary if we can parse a colon and a simpletype.
+        // Otherwise we can get a wrong diagnostiv if we get a Python-style function declaration.
```
ahoppen, line 1076:
```suggestion
        // Otherwise we can get a wrong diagnostic if we get a Python-style function declaration.
```

ahoppen (approved):

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test linux

ahoppen (approved):

kimdv:
@swift-ci please test linux

kimdv:
@swift-ci please test Windows

Issue #1473, opened 2023-03-31
BasicFormat indents closure parameter


bnbarham:
```
func testClosureParam() {
  var input = Parser("""
    foo(someClosure: { _ in
        return 1
    })
    """)
  let call = ExprSyntax.parse(from: &input)

  let formatted = call.formatted()
  assertStringsEqualWithDiff(formatted.description, """
    foo(someClosure: { _ in
        return 1
    })
    """)
}
```

Produces:
```
failed - Actual output (+) differed from expected output (-):
–foo(someClosure: { _ in
–    return 1
–})
+foo(someClosure: { _ in 
+        return 1
+    })
```

Note the double indent on `return 1` and indent of `}`. Presumably we're always indenting on newlines within parameters of call expressions and then again from the code block.

ahoppen:
rdar://107455223

PR #1485, opened 2023-04-02
Add convert to trailing closure and editor placeholder refactorings

bnbarham:
Adds three new refactorings:
  - `ConvertToTrailingClosures`
  - `ExpandEditorPlaceholder`
  - `ExpandEditorPlaceholders`

`ExpandEditorPlaceholders` is a combination of `ExpandEditorPlaceholder` and `ConvertToTrailingClosures`, ie. it first expands any function-typed closures at the end of a call using `ExpandEditorPlaceholder` and then runs `ConvertToTrailingClosures` on that call.

Resolves rdar://107532856.


bnbarham:
@swift-ci please test

bnbarham:
This has a couple of differences with the existing placeholder expansion:
  - Typed placeholders within strings are converted the same way as they would be outside of a string. They're currently expanded as is, ie. `<#T##display##type#>` would be expanded to `T##display##type#`.
  - Missing types in the type syntax would cause placeholders to be expanded as is. This now falls back as much as possible instead, see `ExpandEditorPlaceholderTest.testEmpty`. IMO this makes more sense, but happy to change if anyone feels strongly about it.

BasicFormat still needs some fixes here, but we can do that after. The main one is the missing space between `}` and the label for multiple trailing closures. There's also the indentation, but our current expectation is that clients would fix up indentation on any returned refactoring anyway.

Sources/SwiftRefactor/ExpandEditorPlaceholder.swift
```diff
@@ -0,0 +1,339 @@
...
+/// foo(closure1: <#T##(Int) -> String##(Int) -> String##(_ someInt: Int) -> String#>, normalArg: <#T##Int#>, closure2: { ... }) { someInt in
+///   <#T##String#>
+/// } closure2: { someInt in
+///   <#T##String#>
+/// }
+/// ```
+///
+/// Expansion on `closure1` and `normalArg` is the same as `ExpandEditorPlaceholder`.
+public struct ExpandEditorPlaceholders: RefactoringProvider {
+  public static func refactor(syntax token: TokenSyntax, in context: Void) -> ExprSyntax? {
```
bnbarham, line 117: This is a bit of a weird API at the moment. We take a `TokenSyntax` but may return a `FunctionCallExprSyntax` if we convert the call itself to trailing closure syntax. Would love to hear opinions on this 😅

bnbarham:
@swift-ci please test

bnbarham:
@swift-ci please test

bnbarham:
@swift-ci please test

Sources/SwiftRefactor/ExpandEditorPlaceholder.swift
hamishknight: Any reason not to have it be like `CallToTrailingClosures` and have it take the function call with a starting index?

bnbarham: Just that it seemed a little weird to force clients to check whether they have a `TokenSyntax` within a `FunctionCallExprSyntax` themselves. Though having said that, I'm not sure there's any way at the moment for the client to know that the *parent* was replaced rather than the `TokenSyntax` they passed. So maybe it is better to go the `CallToTrailingClosures` way.

Sources/SwiftRefactor/CallToTrailingClosures.swift
```diff
@@ -0,0 +1,115 @@
...
+    var closures = [(label: TokenSyntax, closure: ClosureExprSyntax)]()
+    for arg in argumentList.dropFirst(startAt) {
+      guard let closure = arg.expression.as(ClosureExprSyntax.self) else {
+        closures.removeAll()
+        continue
+      }
+
+      // Trailing comma won't exist any more, move its trivia to the end of
+      // the closure instead
+      closures.append((arg.label ?? .wildcardToken(), closure.with(\.trailingTrivia, (closure.trailingTrivia ?? []) + (arg.trailingComma?.trailingTrivia ?? []))))
```
ahoppen, line 67: Technically, you would also need to move the trailing comma’s leading trivia, right? I think you could end up with a comment as leading trivia if you have the comma on a newline like the following where the comma is on a newline (haven’t verified though). Maybe just retain the comments from the trivia?
```swift
foo(1
/*some comment*/,2)
```

Sources/SwiftRefactor/CallToTrailingClosures.swift
```diff
@@ -0,0 +1,115 @@
...
+        .with(\.leftParen, nil)
+        .with(\.rightParen, nil)
+    } else {
+      // Remove the last comma and move its trivia to after the right paren.
+      // Also move all the trailing comma trivia of each closure. The right
+      // paren's trivia was moved to the very end of the call above.
+      let last = argList.last!
+      converted =
+        converted
+        .with(\.rightParen, TokenSyntax.rightParenToken(trailingTrivia: last.trailingComma?.trailingTrivia ?? []))
```
ahoppen, line 98: Again you should be putting the last comma’s leading trivia somewhere.

Sources/SwiftRefactor/ExpandEditorPlaceholder.swift
```diff
@@ -0,0 +1,347 @@
...
+  case typed(text: Substring, type: TypeSyntax)
+
+  init?(token: TokenSyntax) {
+    guard ExpandEditorPlaceholder.isPlaceholder(token.text) else {
+      return nil
+    }
+
+    var text = token.text.dropFirst(2).dropLast(2)
+
+    if !text.hasPrefix("T##") {
```
ahoppen, line 286: Is `T##` really the thing we want to be looking for here? I thought that the first type is the one that will be used during type checking and it can be any type. But I might be completely wrong.

Sources/SwiftRefactor/ExpandEditorPlaceholder.swift
```diff
@@ -0,0 +1,347 @@
...
+      self = .basic(text: text)
+      return
+    }
+
+    // Have type text, see if we also have expansion text
+
+    let expansionText: Substring
+    (typeText, expansionText) = split(typeText, separatedBy: "##")
+    if expansionText.isEmpty {
+      if typeText.isEmpty {
```
ahoppen, line 308: Can this ever happen? Aren’t we checking this in line 297 already?

Sources/SwiftRefactor/ExpandEditorPlaceholder.swift
```diff
@@ -0,0 +1,347 @@
...
+    )
+  }
+}
+
+extension TupleTypeElementSyntax {
+  /// Return a token to use as the parameter name in the expanded closure.
+  /// We prefer the argument name if there is one and it isn't a wildcard,
+  /// falling back to the label with the same conditions, and finally just the
+  /// placeholder-ed type otherwise.
+  fileprivate func expansionNameToken() -> TokenSyntax {
```
ahoppen, line 203: This could just be a computed property, right?

Sources/SwiftRefactor/ExpandEditorPlaceholder.swift
```diff
@@ -0,0 +1,347 @@
...
+
+    // Single statement for the body - the placeholder-ed type if non-void and
+    // 'code' otherwise.
+    let ret = output.returnType.description
+    let placeholder: String
+    if ret == "Void" || ret == "()" {
+      placeholder = ExpandEditorPlaceholder.wrapInTypePlaceholder("code", type: "Void")
+    } else {
+      placeholder = ExpandEditorPlaceholder.wrapInTypePlaceholder(ret, type: ret)
+    }
```
ahoppen, line 180: Do we even need the `if` here? The only time it does anything is translate `()` to `Void` and I would imagine that all parts of the compiler that deal with placeholders know that `()` and `Void` are equivalent.

Tests/SwiftRefactorTest/ExpandEditorPlaceholderTests.swift
```diff
@@ -0,0 +1,116 @@
...
+  expected: ExprSyntax?,
+  file: StaticString = #file,
+  line: UInt = #line
+) throws {
+  let token: TokenSyntax
+  if wrap {
+    token = "\(raw: ExpandEditorPlaceholder.wrapInPlaceholder(placeholder))"
+  } else {
+    let expr: ExprSyntax = "\(raw: placeholder)"
+    token = try XCTUnwrap(expr.as(EditorPlaceholderExprSyntax.self)?.identifier)
```
ahoppen, line 94:
```suggestion
    token = try XCTUnwrap(expr.as(EditorPlaceholderExprSyntax.self)?.identifier, file: file, line: line)
```

Sources/SwiftRefactor/CallToTrailingClosures.swift
bnbarham: Yes, There's probably a few places like this - I was waiting for https://github.com/apple/swift-syntax/pull/1478 (which was waiting on the other). I'll fix that up and then use it here.

Sources/SwiftRefactor/ExpandEditorPlaceholder.swift
bnbarham: The main purpose isn't for making `()` `Void`, it's to make `Void` show `code` and not `Void` which IMO is a nice little change 😅.

Sources/SwiftRefactor/ExpandEditorPlaceholder.swift
bnbarham: `T## is mandatory for both typechecking and expansion. Without `T##` typechecking treats it as `Void`. As an example, `foo(<#(Int) -> (Int)#>)` is not expanded to a closure (or trailing closure).

Sources/SwiftRefactor/ExpandEditorPlaceholder.swift
bnbarham: If you'd prefer 🤷

Sources/SwiftRefactor/ExpandEditorPlaceholder.swift
ahoppen: Interesting. You learn something new every day.

Sources/SwiftRefactor/ExpandEditorPlaceholder.swift
ahoppen: I would, not super strongly though.

Sources/SwiftRefactor/ExpandEditorPlaceholder.swift
ahoppen: Ah, I see.

Sources/SwiftRefactor/ExpandEditorPlaceholder.swift
ahoppen: Ah, I missed the `"code"` vs `ret` distinction. Makes sense 👍🏽

Sources/SwiftRefactor/CallToTrailingClosures.swift
bnbarham: > Maybe just retain the comments from the trivia?

I'd rather keep whitespace I think, ie.
```
foo(1 /* some comments */
/*more comments*/, { _ in })
```
->
```
foo(1/* some comments */
/*more comments*/) { _ in }
```

Rather than:
```
foo(1/* some comments *//*more comments*/) { _ in }
```

ahoppen: I was just thinking to strip whitespace to avoid the need to do more complicated trivia merging. But it’s good with me if you’re planning to update this once #1478 is merged.

DougGregor:
@swift-ci please test Windows

Sources/SwiftRefactor/ExpandEditorPlaceholder.swift
bnbarham: My current plan here is to add a new protocol `EditRefactoringProvider`. It would have a single:
```
  static func refactor(syntax: Self.Input, in context: Self.Context) -> [Edit]
```

`RefactoringProvider` would itself be an `EditRefactoringProvider`, providing a default implementation of `refactor` that removes the input node and replaces it with the returned node.

Then `ExpandEditorPlaceholders` would no longer be a `RefactoringProvider`, but rather a `EditRefactoringProvider`. Possibly `ExpandEditorPlaceholder` too, haven't decided on that one yet. It really does just return text, so it probably shouldn't be a `RefactoringProvider`.

bnbarham:
@swift-ci please test

bnbarham:
@swift-ci please test

Sources/SwiftRefactor/CallToTrailingClosures.swift
```diff
@@ -0,0 +1,123 @@
...
+/// Convert a call with inline closures to one that uses trailing closure
+/// syntax. Returns `nil` if there's already trailing closures or there are no
+/// closures within the call. Pass `conversionStart` to specify an index to
+/// start the conversion from, ie. to skip converting closures before
+/// `conversionStart`
+///
+/// ## Before
+/// ```
+/// someCall(closure1: (String) -> Int, closure2: (String) -> Int)
+/// ```
```
hamishknight, line 25: Shouldn't the arguments be closure exprs here?

hamishknight (approved):

Sources/SwiftRefactor/CallToTrailingClosures.swift
```diff
@@ -0,0 +1,123 @@
...
+    var argList = Array(argumentList.dropLast(closures.count))
+    if argList.isEmpty {
+      converted =
+        converted
+        .with(\.leftParen, nil)
+        .with(\.rightParen, nil)
+    } else {
+      // Remove the last comma and move its trivia to after the right paren.
+      // Also move all the trailing comma trivia of each closure. The right
+      // paren's trivia was moved to the very end of the call above.
```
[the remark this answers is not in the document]
bnbarham: Thanks!

bnbarham:
@swift-ci please test

bnbarham:

Sources/SwiftRefactor/CallToTrailingClosures.swift
```diff
@@ -0,0 +1,131 @@
...
+/// ## After
+/// ```
+/// someCall { arg in
+///   return 1
+/// } closure2: { arg in
+///   return 2
+/// }
+/// ```
+public struct CallToTrailingClosures: SyntaxRefactoringProvider {
+  public struct Options {
```
ahoppen, line 40: You could just name this struct `Context` and then you wouldn’t need the `typealias` below.

Sources/SwiftRefactor/CallToTrailingClosures.swift
```diff
@@ -0,0 +1,131 @@
...
+        closure = closure.with(\.trailingTrivia, closure.trailingTrivia.merging(triviaOf: comma))
+      }
+      closures.append((arg.label ?? .wildcardToken(), closure))
+    }
+
+    guard !closures.isEmpty else {
+      return nil
+    }
+
+    let trailingClosure = closures.first!.closure
```
ahoppen, line 82: You are dropping the label’s trivia here (and the colon’s trivia based on my comment above). I think we should merge the trivia of those tokens into the `trailingClosure`’s leading trivia.

Sources/SwiftRefactor/CallToTrailingClosures.swift
```diff
@@ -0,0 +1,131 @@
...
+
+    // Remove parens if there's no non-closure arguments left. Just remove the
+    // last comma otherwise and move its trivia to after the right paren (ie.
+    // before the first trailing closure).
+    var argList = Array(argumentList.dropLast(closures.count))
+    if argList.isEmpty {
+      converted =
+        converted
+        .with(\.leftParen, nil)
+        .with(\.rightParen, nil)
```
ahoppen, line 101: Again, dropping the trivia of the `leftParen` and `rightParen`. Should we attach it to the first trailing closure instead? I wonder whether it would make sense to have a function like the following that removes a token and moves its trailing trivia to adjacent tokens. That way you wouldn’t really need to worry about moving the trivia for the tokens you’re removing somewhere. I haven’t tested it thoroughly though.

```swift
class TransferTrivia: SyntaxRewriter {
  var tokenToRemove: TokenSyntax

  init(tokenToRemove: TokenSyntax) {
    self.tokenToRemove = tokenToRemove
  }

  override func visitAny(_ node: Syntax) -> Syntax? {
    if !node.byteRange.intersects(tokenToRemove.byteRange) {
      // Nothing to rewrite. Skip this subtree
      return node
    }
    return nil
  }

    if token.nextToken(viewMode: .sourceAccurate) == tokenToRemove {
      let leadingTriviaToTransfer = (tokenToRemove.leadingTrivia + tokenToRemove.trailingTrivia).prefix(while: { !$0.isNewline })
      return token.with(\.trailingTrivia, token.trailingTrivia + Trivia(pieces: leadingTriviaToTransfer))
    } else if token.previousToken(viewMode: .sourceAccurate) == tokenToRemove {
      let trailingTriviaToTransfer = (tokenToRemove.leadingTrivia + tokenToRemove.trailingTrivia).drop(while: { !$0.isNewline })
      return token.with(\.leadingTrivia, token.leadingTrivia + Trivia(pieces: trailingTriviaToTransfer))
    } else {
      return token
    }
  }
}

extension SyntaxProtocol {
  func removingRetainingTrivia(_ keyPath: WritableKeyPath<Self, TokenSyntax?>) -> Self {
    guard let token = self[keyPath: keyPath] else {
      return self
    }
    let transferredTrivia = TransferTrivia(tokenToRemove: token).rewrite(Syntax(self)).cast(Self.self)
    return transferredTrivia.with(keyPath, nil)
  }
}
```

Tests/SwiftRefactorTest/ExpandEditorPlaceholdersTests.swift
```diff
@@ -0,0 +1,116 @@
...
+import SwiftParser
+import SwiftRefactor
+import SwiftSyntax
+import SwiftSyntaxBuilder
+import XCTest
+import _SwiftSyntaxTestSupport
+
+fileprivate let closurePlaceholder = ExpandEditorPlaceholder.wrapInPlaceholder("T##closure##() -> Void")
+fileprivate let voidPlaceholder = ExpandEditorPlaceholder.wrapInPlaceholder("T##code##Void")
+fileprivate let intPlaceholder = ExpandEditorPlaceholder.wrapInPlaceholder("T##Int##Int")
```
ahoppen, line 23: I don’t think you’ve got any tests that for `type-for-expansion-string`. Could you add some?

Sources/SwiftRefactor/ExpandEditorPlaceholder.swift
```diff
@@ -0,0 +1,346 @@
...
+}
+
+/// If a function-typed placeholder is the argument to a non-trailing closure
+/// call, expands it and any adjacent function-typed placeholders to trailing
+/// closures on that call. All other placeholders will expand as per
+/// `ExpandEditorPlaceholder`.
+///
+/// ## Before
+/// ```swift
+/// foo(closure1: <#T##(Int) -> String##(Int) -> String##(_ someInt: Int) -> String#>, normalArg: <#T##Int#>, closure2: { ... }, closure3: <#T##(Int) -> String##(Int) -> String##(_ someInt: Int) -> String#>, closure4: <#T##(Int) -> String##(Int) -> String##(_ someInt: Int) -> String#>
```
ahoppen, line 112: This is pretty awful to read. Could you put the parameters onto different lines?

bnbarham: Heh, I didn't notice because they're placeholders in the editor 😅, but sure.

Sources/SwiftRefactor/CallToTrailingClosures.swift
bnbarham: It's only dropping `leftParen`, `rightParen` is added below. I don't think adding to the next token is quite what we want, ie. I would expect the trivia of `)` to end up after the call.

Sources/SwiftRefactor/CallToTrailingClosures.swift
```diff
@@ -0,0 +1,131 @@
...
+//
+// Copyright (c) 2014 - 2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftBasicFormat
```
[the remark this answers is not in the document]
bnbarham: Indeed. I'm ... surprised this works right now, but it definitely does. 😕

Sources/SwiftRefactor/RefactoringProvider.swift
```diff
@@ -12,30 +12,43 @@
...
-/// refactoring action that wishes to exchange the leading trivia of a node
-/// would call `with(\.leadingTrivia, _:)` against its input syntax and return
-/// the resulting syntax node. For compound syntax nodes, entire sub-trees
-/// can be added, exchanged, or removed by calling the corresponding `with`
-/// API.
+/// A refactoring expressed as textual edits on the original syntax tree. In
+/// general clients should prefer `SyntaxRefactoringProvider` where possible.
+public protocol EditRefactoringProvider {
+  /// The type of syntax this refactoring action accepts.
+  associatedtype Input: SyntaxProtocol = SourceFileSyntax
```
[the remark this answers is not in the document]
bnbarham: I think having to specify these is... not great. I've opened https://github.com/apple/swift/issues/66004, referenced that in `SyntaxRefactoringProvider`, and changed `SourceFileSyntax` to `MissingSyntax`.

Sources/SwiftRefactor/ExpandEditorPlaceholder.swift
```diff
@@ -0,0 +1,346 @@
...
+  /// { someInt in
+  ///   <#T##String#>
+  /// }
+  /// ```
+  fileprivate var closureExpansion: ClosureExprSyntax {
+    let closureSignature: ClosureSignatureSyntax?
+    if !arguments.isEmpty {
+      var args = arguments.dropLast().map { ClosureParamSyntax(name: $0.expansionNameToken(), trailingComma: .commaToken()) }
+      args.append(ClosureParamSyntax(name: arguments.last!.expansionNameToken()))
+      closureSignature = ClosureSignatureSyntax(input: .simpleInput(ClosureParamListSyntax(args)))
```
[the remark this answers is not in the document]
bnbarham: Nice, thanks. I wasn't including SwiftSyntaxBuilder before but... guess I can.

Sources/SwiftRefactor/RefactoringProvider.swift
```diff
@@ -12,30 +12,43 @@
...
+
+  /// Perform the refactoring action on the provided syntax node.
+  ///
+  /// - Parameters:
+  ///   - syntax: The syntax to transform.
+  ///   - context: Contextual information used by the refactoring action.
+  /// - Returns: Textual edits that describe how to apply the result of the
+  ///            refactoring action on locations within the original tree. An
+  ///            empty array if the refactoring could not be performed.
+  static func textRefactor(syntax: Self.Input, in context: Self.Context) -> [Edit]
```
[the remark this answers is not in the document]
ahoppen: If you feel strongly about it, keep `Self`. I would just remove it.

Tests/SwiftRefactorTest/ExpandEditorPlaceholdersTests.swift
bnbarham: There's tests for it in `ExpandEditorPlaceholderTest`. `ExpandEditorPlaceholders` just calls the other, so I don't really see the point of testing it all here again. I can add one if you want though.

Sources/SwiftRefactor/CallToTrailingClosures.swift
```diff
@@ -0,0 +1,131 @@
...
+
+    var closures = [(label: TokenSyntax, closure: ClosureExprSyntax)]()
+    for arg in argumentList.dropFirst(startAt) {
+      guard var closure = arg.expression.as(ClosureExprSyntax.self) else {
+        closures.removeAll()
+        continue
+      }
+
+      // Trailing comma won't exist any more, move its trivia to the end of
+      // the closure instead
```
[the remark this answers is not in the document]
bnbarham: Nice catch. I'll just keep the original arg 🙇

Tests/SwiftRefactorTest/RefactorTestUtils.swift
```diff
@@ -0,0 +1,123 @@
...
+    XCTFail("Expected nil result, actual:\n\(refactored.description)", file: file, line: line)
+    return
+  }
+
+  assertStringsEqualWithDiff(
+    refactored.description,
+    expected.description,
+    file: file,
+    line: line
+  )
```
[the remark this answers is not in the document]
bnbarham: Could have both 🤷 I'd rather not re-write all the refactoring tests in this PR though

Tests/SwiftRefactorTest/CallToTrailingClosureTests.swift
```diff
@@ -0,0 +1,229 @@
...
+      }, 1, { label2 in
+          return 2
+      }, { label3 in
+          return 3
+      }, named: { label4 in
+          return 4
+      })
+      """
+
+    // TODO: BasicFormat is pretty messed up here
```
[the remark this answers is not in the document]
bnbarham: I already have one, I can reference it here though (https://github.com/apple/swift-syntax/issues/1473)

Tests/SwiftRefactorTest/RefactorTestUtils.swift
```diff
@@ -0,0 +1,123 @@
...
+  context: R.Context,
+  provider: R.Type,
+  expected: [Edit],
+  file: StaticString = #file,
+  line: UInt = #line
+) throws {
+  let edits = R.textRefactor(syntax: input, in: context)
+  guard !edits.isEmpty else {
+    if !expected.isEmpty {
+      XCTFail("Refactoring produced empty result, expected:\n\(expected)", file: file, line: line)
```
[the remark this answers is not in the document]
bnbarham: Eh, I don't mind it in cases like this. But I'll remove it just for you :)

Sources/SwiftRefactor/RefactoringProvider.swift
```diff
@@ -84,3 +96,80 @@ extension RefactoringProvider where Context == Void {
...
+
+  /// Convenience function to create a textual deletion the given node and its
+  /// trivia.
+  public static func remove<T: SyntaxProtocol>(_ node: T) -> Edit {
+    return Edit(start: node.position, end: node.endPosition, replacement: "")
+  }
+}
+
+extension Edit: CustomStringConvertible {
+  public var description: String {
```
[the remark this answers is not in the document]
bnbarham: Seems useful as a description in general to me. Any reason to prefer debug here?

bnbarham:
@swift-ci please test

Sources/SwiftRefactor/RefactoringProvider.swift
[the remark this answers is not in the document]
ahoppen: I usually try to use `description` only if there is a single, obvious string representation of the type because it’s fairly easy to use `description` in code. And then `description` also needs to provide some sort of stability guarantees IMO. `debugDescription` just makes a lot fewer guarantees IMO and it’s sufficient here.

Tests/SwiftRefactorTest/CallToTrailingClosureTests.swift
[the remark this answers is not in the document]
ahoppen: Ah, thanks

Tests/SwiftRefactorTest/RefactorTestUtils.swift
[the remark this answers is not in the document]
ahoppen: That’s a very fair point.  👍🏽

Sources/SwiftRefactor/CallToTrailingClosures.swift
```diff
@@ -0,0 +1,143 @@
...
+    guard !closures.isEmpty else {
+      return nil
+    }
+
+    // First trailing closure won't have label/colon. Transfer their trivia.
+    var trailingClosure = closures.first!.closure
+      .with(\.leadingTrivia, Trivia()
+        .merging(triviaOf: closures.first!.original.label)
+        .merging(triviaOf: closures.first!.original.colon)
+        .merging(closures.first!.closure.leadingTrivia))
```
ahoppen, line 86: Are you dropping the trivia of `closures.first!.trailingComma` or am I missing something?

Sources/SwiftRefactor/RefactoringProvider.swift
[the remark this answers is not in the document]
bnbarham: Fair enough.

Sources/SwiftRefactor/CallToTrailingClosures.swift
bnbarham: It's added above. FWIW I added quite a few more comments in the trivia test that hopefully catches all the cases now 😅. Worth checking those as well - IMO there's a few odd spaces that we should probably remove, but I think it's good enough for now.

bnbarham:
@swift-ci please test

bnbarham:
@swift-ci please test

bnbarham:

bnbarham:
@swift-ci please test

bnbarham:

bnbarham:
@swift-ci please test

bnbarham:
@swift-ci please test

bnbarham:
https://github.com/apple/swift-stress-tester/pull/238

@swift-ci please test

ahoppen (approved):

bnbarham:
@swift-ci please test

bnbarham:
https://github.com/apple/swift-stress-tester/pull/238

@swift-ci please test

bnbarham:
@swift-ci please test

bnbarham:
https://github.com/apple/swift-stress-tester/pull/238

@swift-ci please test

bnbarham:
https://github.com/apple/swift-stress-tester/pull/238

@swift-ci please test

bnbarham:
https://github.com/apple/swift-stress-tester/pull/238

@swift-ci please test

PR #1477, opened 2023-03-31
Make leading/trailing trivia required fields

bnbarham:
Rather than having the possibility of `nil` trivia and requiring all clients to check every use, just return empty trivia when there are no nodes present.

PR #1487, opened 2023-04-03
[5.9] Make leading/trailing trivia required fields

bnbarham:
* Explanation: Rather than having the possibility of `nil` trivia and requiring all clients to check every use, just return empty trivia when there are no nodes present.
* Scope: All swift-syntax clients using `leading`/`trailingTrivia`
* Risk: Extremely low, uses generally created an empty `Trivia` anyway 
* Testing: No additional tests
* Original PR: https://github.com/apple/swift-syntax/pull/1477


bnbarham:
https://github.com/apple/swift-format/pull/500
https://github.com/apple/swift/pull/64874

@swift-ci please test

DougGregor (approved):

PR #1489, opened 2023-04-04
Change included header for uint64_t in InstructionsExecuted.h

ahoppen:
The Linux build on swiftpackageindex.com said it couldn’t find uint64_t in InstructiosnExecuted.h. I think we need to include `stdint.h` instead. Let’s try that.


ahoppen:
@swift-ci Please test

rintaro (approved):

PR #1460, opened 2023-03-28
Slightly clean up `FixIt.Change` and `FixIt.Changes`

ahoppen:
The existing API with `FixIt.Change` and `FixIt.Changes` and the fact that we passed `[FixIt.Changes]` around what fairly weird.

Ideally, we would only have `FixIt.Change` and make it powerful enough to represent all the trivia transfer that `FixIt.Changes` is doing. But given that that will be a larger effort, let’s make some smaller changes now that reduce the badness of the public API impact:

- Move `FixIt.Changes` to `SwiftParserDiagnostics`, make it internal and rename it to `MultiNodeChange`. That way this type is no longer exposed in the public API, so we can iterate on it or remove it without breaking clients. The only thing that remains exposed is `FixIt.Change`, which we want to continue to support. We will probably add more cases to it in the future.
- Make `FixIt.Changes` no longer conform to `ExpressibleByArrayLiteral`. That just makes everything a lot more explicit and easier to follow.
- Remove some unecessary initializations of `FixIt.Changes` where `FixIt.Change` was sufficient.
- Make `presence` on `TokenSyntax` settable so you can do `token.with(\.presence, .missing)`
- Slightly unrelated: Rename SyntaxOtherNodes.swift to TokenSyntax.swift because that’s the only node it contains.

PR #1496, opened 2023-04-05
[5.9] Slightly clean up `FixIt.Change` and `FixIt.Changes`

ahoppen:
Cherry-pick https://github.com/apple/swift-syntax/pull/1460 to release/5.9

---

The existing API with `FixIt.Change` and `FixIt.Changes` and the fact that we passed `[FixIt.Changes]` around what fairly weird.

Ideally, we would only have `FixIt.Change` and make it powerful enough to represent all the trivia transfer that `FixIt.Changes` is doing. But given that that will be a larger effort, let’s make some smaller changes now that reduce the badness of the public API impact:

- Move `FixIt.Changes` to `SwiftParserDiagnostics`, make it internal and rename it to `MultiNodeChange`. That way this type is no longer exposed in the public API, so we can iterate on it or remove it without breaking clients. The only thing that remains exposed is `FixIt.Change`, which we want to continue to support. We will probably add more cases to it in the future.
- Make `FixIt.Changes` no longer conform to `ExpressibleByArrayLiteral`. That just makes everything a lot more explicit and easier to follow.
- Remove some unecessary initializations of `FixIt.Changes` where `FixIt.Change` was sufficient.
- Make `presence` on `TokenSyntax` settable so you can do `token.with(\.presence, .missing)`
- Slightly unrelated: Rename SyntaxOtherNodes.swift to TokenSyntax.swift because that’s the only node it contains.

### DIFF
--- a/Sources/SwiftCompilerPluginMessageHandling/Diagnostics.swift
+++ b/Sources/SwiftCompilerPluginMessageHandling/Diagnostics.swift
@@ -84,7 +84,7 @@ extension PluginMessage.Diagnostic {
     self.fixIts = syntaxDiag.fixIts.compactMap {
       PluginMessage.Diagnostic.FixIt(
         message: $0.message.message,
-        changes: $0.changes.changes.compactMap {
+        changes: $0.changes.compactMap {
           let range: SourceManager.SourceRange?
           let text: String
           switch $0 {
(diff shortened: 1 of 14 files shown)


ahoppen:
https://github.com/apple/swift/pull/64960

@swift-ci Please test

DougGregor (approved):

PR #1497, opened 2023-04-06
fix XCTSkip warning

andrewjl:
Remove code after 'throw' will never be executed warning


ahoppen:
@swift-ci Please test

PR #1498, opened 2023-04-06
Fix wrong diagnostic for generics

### LINKED ISSUE #1483: Empty generic parameter clause produces no diagnostic
### Issue Kind

Bad Diagnostic Produced

### Source Code

```swift
// Can be parsed and produces no diagnostics.
func f< >() {}

// Can be parsed. Printing the node or asking for the diagnostics leads to a crash.
func f<>() {} 
```

### Description

In both examples, the "old" parser reports "Expected an identifier to name generic parameter" on `>` which is helpful, but not the real issue.

Best would be a message like "Generic parameter clause cannot be empty" or similar. It should be reported in both examples. The `ParseDiagnosticsGenerator` should not crash in the second example.

kimdv:
Fixes #1483 

Opened to see if I was on the right path

### DIFF
--- a/Sources/SwiftParser/Declarations.swift
+++ b/Sources/SwiftParser/Declarations.swift
@@ -465,7 +465,7 @@ extension Parser {
     }
 
     precondition(self.currentToken.starts(with: "<"))
-    let langle = self.consumeAnyToken(remapping: .leftAngle)
+    let langle = self.consumePrefix("<", as: .leftAngle)
     var elements = [RawGenericParameterSyntax]()
     do {
       var keepGoing: RawTokenSyntax? = nil
@@ -477,7 +477,7 @@ extension Parser {
         var each = self.consume(if: .keyword(.each))
 
         let (unexpectedBetweenEachAndName, name) = self.expectIdentifier(allowSelfOrCapitalSelfAsIdentifier: true)
-        if attributes == nil && each == nil && unexpectedBetweenEachAndName == nil && name.isMissing && elements.isEmpty {
+        if attributes == nil && each == nil && unexpectedBetweenEachAndName == nil && name.isMissing && elements.isEmpty && !self.currentToken.starts(with: ">") {
           break
         }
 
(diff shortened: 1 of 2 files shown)


kimdv:
@swift-ci please test

ahoppen:
I think we should not diagnose this in the parser. An generic parameter list can be represented perfectly well in the SwiftSyntax tree and should thus be diagnosed by later stages of the compilation pipeline.

kimdv:
@ahoppen I'm not sure I understand.
When debugging this, I looked into the old parser and try to make the same as the old parser. 

I think the reason is, because here the current token is `<>` (a binary operator).
https://github.com/apple/swift-syntax/blob/ac4f8f3ee2d3866152bcd555a62cb8317ea3cb11/Sources/SwiftParser/Declarations.swift#L467

The line after we then consume the token and remap it to `<`.
Then it crash here because text is **not** empty, it is `<>`.
https://github.com/apple/swift-syntax/blob/ac4f8f3ee2d3866152bcd555a62cb8317ea3cb11/Sources/SwiftSyntax/generated/TokenKind.swift#L982

So what I try to do, is only consume the `<` and then move the lexer to `>`.

Sources/SwiftParser/Declarations.swift
```diff
@@ -465,7 +465,7 @@ extension Parser {
     }
 
     precondition(self.currentToken.starts(with: "<"))
-    let langle = self.consumeAnyToken(remapping: .leftAngle)
+    let langle = self.consumeStartingCharacterOfCurrentToken(as: .leftAngle)
```
ahoppen, line 468: Can’t you use `consumePrefix` here?

ahoppen (commented):
Forget what I said. I was stupid and didn’t read your PR properly, I’m sorry. 

I thought you had added code in `ParseDiagnosticGenerator` that detected empty generic argument lists but that’s just not what you did. Looking at your PR again, it looks good, I’m just wondering why you can’t use `consumePrefix`.

Sources/SwiftParser/Declarations.swift
kimdv: Doh! Sorry. Using `consumePrefix` instead.

ahoppen (approved):

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test windows

PR #1499, opened 2023-04-06
[bazel] Add Macro related libraries

keith:


keith:
@swift-ci please test

PR #1508, opened 2023-04-09
Implement StringLiteralExprSyntax/contentValue.

NikolaiRuhe:
SwiftSyntax parses string literals as `StringLiteralExprSyntax`. When the literal is static (does not contain interpolation segments) it can be useful to access the actual string value from the syntax tree.

The source-accurate representation in `StringLiteralExprSyntax` differs from run-time value:

- Escaped character sequences (`\n`, `\u{1F600}`) are not evaluated.
- Some strings are split into segments (i.e. multiline strings).

`StringLiteralExprSyntax/contentValue` performs the necessary unescaping and concatenation of segments.

### DIFF
--- a/Sources/SwiftParser/StringLiteralRepresentedLiteralValue.swift
+++ b/Sources/SwiftParser/StringLiteralRepresentedLiteralValue.swift
@@ -1,0 +1,105 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+@_spi(RawSyntax) import SwiftSyntax
+
+extension StringLiteralExprSyntax {
+
+  /// Returns the string value of the literal as the parsed program would see
+  /// it: Multiline strings are combined into one string, escape sequences are
+  /// resolved.
+  ///
+  /// Returns nil if the literal contains interpolation segments.
+  public var representedLiteralValue: String? {
+    // Currently the implementation relies on properly parsed literals.
+    guard !hasError else { return nil }
+    guard let stringLiteralKind else { return nil }
+
+    // Concatenate unescaped string literal segments. For example multiline
+    // strings consist of multiple segments. Abort on finding string
+    // interpolation.
+    var result = ""
+    for segment in segments {
+      switch segment {
+      case .stringSegment(let stringSegmentSyntax):
+        stringSegmentSyntax.appendUnescapedLiteralValue(
+          stringLiteralKind: stringLiteralKind,
+          delimiterLength: delimiterLength,
+          to: &result
+        )
+      case .expressionSegment:
+        // Bail out if there are any interpolation segments.
+        return nil
+      }
+    }
+
+    return result
+  }
+
+  fileprivate var stringLiteralKind: StringLiteralKind? {
+    switch openQuote.tokenKind {
+    case .stringQuote:
+      return .singleLine
+    case .multilineStringQuote:
+      return .multiLine
+    case .singleQuote:
+      return .singleQuote
+    default:
+      return nil
+    }
+  }
+
+  fileprivate var delimiterLength: Int {
+    openDelimiter?.text.count ?? 0
+  }
+}
+
+extension StringSegmentSyntax {
+  fileprivate func appendUnescapedLiteralValue(
+    stringLiteralKind: StringLiteralKind,
+    delimiterLength: Int,
+    to output: inout String
+  ) {
+    precondition(!hasError, "appendUnescapedLiteralValue relies on properly parsed literals")
+
+    var text = content.text
+    text.withUTF8 { buffer in
+      var cursor = Lexer.Cursor(input: buffer, previous: 0)
+
+      // Put the cursor in the string literal lexing state. This is just
+      // defensive as it's currently not used by `lexCharacterInStringLiteral`.
+      let state = Lexer.Cursor.State.inStringLiteral(kind: stringLiteralKind, delimiterLength: delimiterLength)
+      let transition = Lexer.StateTransition.push(newState: state)
+      cursor.perform(stateTransition: transition, stateAllocator: BumpPtrAllocator(slabSize: 256))
+
+      while true {
+        let lex = cursor.lexCharacterInStringLiteral(
+          stringLiteralKind: stringLiteralKind,
+          delimiterLength: delimiterLength
+        )
+
+        switch lex {
+        case .success(let scalar):
+          output.append(Character(scalar))
+        case .validatedEscapeSequence(let character):
+          output.append(character)
+        case .endOfString, .error:
+          // We get an error at the end of the string because
+          // `lexCharacterInStringLiteral` expects the closing quote.
+          // We can assume the error just signals the end of string
+          // because we made sure the token lexed fine before.
+          return
+        }
+      }
+    }
+  }
+}
--- a/Tests/SwiftParserTest/StringLiteralRepresentedLiteralValueTests.swift
+++ b/Tests/SwiftParserTest/StringLiteralRepresentedLiteralValueTests.swift
@@ -1,0 +1,318 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import XCTest
+import SwiftSyntax
+import SwiftParser
+import SwiftSyntaxBuilder
+
+/// Test ``StringLiteralExprSyntax/representedLiteralValue``.
+///
+/// Most tests are expressed by a single function call that compares the
+/// run-time String value against the parsed node's `representedLiteralValue`.
+public class StringLiteralRepresentedLiteralValueTests: XCTestCase {
+
+  func testIntro() {
+    test(
+      #"""
+      This test suite verifies the correctness of ``StringLiteralExprSyntax/\
+      representedLiteralValue.`` It does so by comparing the run-time String
+      value of various string literals to the value returned from
+      `representedLiteralValue`.
+
+      The SwiftSyntax parsed representation `StringLiteralExprSyntax` contains
+      the source-accurate representation of the string literal. This
+      representation differs from the runtime value:
+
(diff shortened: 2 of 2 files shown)


Sources/SwiftParser/StringLiterals.swift
```diff
@@ -598,3 +598,105 @@ public extension RawTriviaPiece {
...
+  fileprivate var stringLiteralKind: StringLiteralKind {
+    switch openQuote.tokenKind {
+    case .stringQuote:
+      return .singleLine
+    case .multilineStringQuote:
+      return .multiLine
+    case .singleQuote:
+      return .singleQuote
+    default:
+      fatalError()
```
CodaFi, line 646: Please add an error message here. Or better yet, make this return an optional.

Sources/SwiftParser/StringLiterals.swift
```diff
@@ -598,3 +598,105 @@ public extension RawTriviaPiece {
...
+
+extension StringLiteralExprSyntax {
+
+  /// Returns the string value of the literal as the parsed program would see
+  /// it: Multiline strings are combined into one string, escape sequences are
+  /// resolved.
+  ///
+  /// Returns nil if the literal contains interpolation segments.
+
+  public var contentValue: String? {
```
CodaFi, line 612: We need to be very very careful about trafficking in `String`s in the syntax tree. The documentation should be very clear about the limitations of this API with respect to the caveats in its implementation and the [source fidelity concerns laid out here](https://github.com/apple/swift-syntax/blob/ac4f8f3ee2d3866152bcd555a62cb8317ea3cb11/Sources/SwiftParser/SwiftParser.docc/ParserDesign.md?plain=1#L56).

CodaFi:
I'm curious why this in `SwiftParser` and not, say, `SwiftSyntaxBuilder`.

Tests/SwiftParserTest/StringLiteralContentValueTests.swift
```diff
@@ -0,0 +1,318 @@
...
+    }
+
+    XCTAssertEqual(contentValue, expected.description, file: file, line: line)
+  }
+
+  static let literals = try! StringLiteralCollector()
+
+  /// Helper class to find string literals in this source file
+  class StringLiteralCollector: SyntaxVisitor {
+    var locationConverter: SourceLocationConverter? = nil
```
ahoppen, line 290: Does this need to be optional?

Tests/SwiftParserTest/StringLiteralContentValueTests.swift
```diff
@@ -0,0 +1,318 @@
...
+      Line 1\
+      .still on Line 1
+      """#
+    )
+  }
+
+  // MARK: literal value not available
+
+  func testMissingQuoteStringLiteral() throws {
+    let sut = StringLiteralExprSyntax(#""a"# as ExprSyntax)!
```
ahoppen, line 253: Why did you name it `sut`? `stringLiteral` would be the more natural name for me. 

Same below.

Sources/SwiftParser/StringLiterals.swift
```diff
@@ -598,3 +598,105 @@ public extension RawTriviaPiece {
     }
   }
 }
+
+// MARK: - StringLiteralExprSyntax/contentValue
+
+extension StringLiteralExprSyntax {
```
ahoppen, line 604: Since this really is functionality that’s separate from the parser, I think it should live in a separate file. That way it’s nice and isolated.

ahoppen (commented):
Thanks, this is looking great to me. A very clever use of the lever to lex the escaped string characters. 

@CodaFi IIUC this lives in `SwiftParser` because it access the lever, which isn’t public (and shouldn’t be). We could define this as SPI in SwiftParser and then offer it as public API in `SwiftSyntaxBuilder` but I think it shouldn’t live in `SwiftSyntaxBuilder` because it’s not about building a syntax tree. So, I think `SwiftParser` is indeed the correct location for this property.

NikolaiRuhe:
Thanks @CodaFi and @ahoppen for the feedback🙏. I’ll reply to it and fix code when I get back from vacation in a week.

Sources/SwiftParser/StringLiterals.swift
NikolaiRuhe: Agreed, thanks.

Sources/SwiftParser/StringLiterals.swift
NikolaiRuhe: I had trouble coming up with a proper name.

I considered a couple of candidates like `stringValue`, `runtimeValue`, `content`, etc. The final decision for `contentValue` came from some existing code in ConvenienceInitializers.swift which basically does the opposite: Building a `StringLiteralExprSyntax` from an unescaped `String` value. This initializer uses the label `content:` for the value. Just `content` seemed to conflicting, though.

I'm also open for better names. I have the feeling `unescapedValue` only matches partly, leaving out the concatenation part, and is difficult to find. But I'll follow your advice here.

Sources/SwiftParser/StringLiterals.swift
```diff
@@ -598,3 +598,105 @@ public extension RawTriviaPiece {
...
+    precondition(!hasError, "impl. relies on properly parsed literals")
+
+    var text = content.text
+
+    // Fast path for literals without escape sequences.
+    // TODO: Check if this isn't actually slower.
+    guard text.contains("\\") else {
+      output.append(text)
+      return
+    }
```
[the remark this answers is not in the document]
NikolaiRuhe: Removed.

Tests/SwiftParserTest/StringLiteralContentValueTests.swift
```diff
@@ -0,0 +1,318 @@
...
+/// Test ``StringLiteralExprSyntax/contentValue``.
+///
+/// Most tests are expressed by a single function call that compares the
+/// run-time String value against the parsed node's `contentValue`.
+
+public class StringLiteralContentValueTests: XCTestCase {
+
+  func testIntro() {
+    test(
+      #"""
```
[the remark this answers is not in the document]
NikolaiRuhe: I do like to have a less "unit", more "integration" style test. Instead of using "lorem ipsum ..." I just typed something that was worth reading.

Sources/SwiftParser/StringLiterals.swift
NikolaiRuhe: Agreed.

Tests/SwiftParserTest/StringLiteralContentValueTests.swift
NikolaiRuhe: I'm used to [name the test subject `sut`](https://en.wikipedia.org/wiki/System_under_test) in unit tests. I did not notice that it's not common in SwiftSyntax. Replaced.

Tests/SwiftParserTest/StringLiteralContentValueTests.swift
NikolaiRuhe: It's only used temporarily during `walk` and can be freed afterwards. Not sure if this is a good reason.

Tests/SwiftParserTest/StringLiteralContentValueTests.swift
```diff
@@ -0,0 +1,318 @@
...
+      This test suite verifies the correctness of ``StringLiteralExprSyntax/\
+      contentValue.`` It does so by comparing the run-time String value of various
+      string literals to the value returned from `contentValue`.
+
+      The Swift-Syntax parsed representation `StringLiteralExprSyntax` contains
```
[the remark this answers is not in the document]
NikolaiRuhe: Fixed.

NikolaiRuhe:
> I'm curious why this in `SwiftParser` and not, say, `SwiftSyntaxBuilder`.

My understanding of SwiftSyntax' module structure is limited. But I think it could actually live in the plain SwiftSyntax module. The functionality is neither about parsing nor building but really just a derived value from the syntax node. It could potentially be useful from manually initialized nodes.

The reason I put it in SwiftParser is because I'm reusing private API from the parser module: `Lexer/lexCharacterInStringLiteral`.

NikolaiRuhe:
I'm not sure about the proper flow: Shall I squash the two commits? Or shall I even rebase onto current main? How does the PR proceed from here?

Thanks, again!

Tests/SwiftParserTest/StringLiteralContentValueTests.swift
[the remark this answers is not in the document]
ahoppen: It’s not something I would have done but I see your reasoning. Makes sense to me 👍🏽

Tests/SwiftParserTest/StringLiteralContentValueTests.swift
ahoppen: Oh, I wasn’t aware of that term. Thanks for changing it.

Sources/SwiftParser/StringLiteralContentValue.swift
```diff
@@ -0,0 +1,106 @@
...
+@_spi(RawSyntax) import SwiftSyntax
+
+extension StringLiteralExprSyntax {
+
+  /// Returns the string value of the literal as the parsed program would see
+  /// it: Multiline strings are combined into one string, escape sequences are
+  /// resolved.
+  ///
+  /// Returns nil if the literal contains interpolation segments.
+
```
ahoppen, line 22: Superfluous newline?

Tests/SwiftParserTest/StringLiteralContentValueTests.swift
```diff
@@ -0,0 +1,318 @@
...
+import XCTest
+import SwiftSyntax
+import SwiftParser
+import SwiftSyntaxBuilder
+
+/// Test ``StringLiteralExprSyntax/contentValue``.
+///
+/// Most tests are expressed by a single function call that compares the
+/// run-time String value against the parsed node's `contentValue`.
+
```
ahoppen, line 22: Superfluous newline?

Sources/SwiftParser/StringLiterals.swift
ahoppen: I think the main reason why I don’t like `contentValue` is that `content` is already a term that’s used in SwiftSyntax, e.g. `StringLiteralExprSyntax` has a `contentLength` property that’s completely unrelated to `contentValue`. And that’s pretty confusing to me.

Alternative idea: What do you think about `representedLiteralValue`?

Sources/SwiftParser/StringLiteralContentValue.swift
NikolaiRuhe: fixed

Tests/SwiftParserTest/StringLiteralContentValueTests.swift
NikolaiRuhe: fixed

NikolaiRuhe:
Yes, naming indeed can be very difficult—and is at the same time _so_ important. It's a bit like an app's icon on the app store. The first thing you see and it should be able to explain what it does in a glimpse. Nobody reads the docs 😢.

In that sense I do like `representedLiteralValue`. It works unchanged on `IntegerLiteralExprSyntax` and `FloatLiteralExprSyntax` (upcoming PR?).

Thanks for the help and the review!

ahoppen (approved):
Great. Thank you. Let’s merge it.

ahoppen:
@swift-ci Please test

kimdv:
@swift-ci please test windows

Issue #1472, opened 2023-03-31
BasicFormat does not add whitespace between end of previous closure and argument name for multi-trailing closures


bnbarham:
```  
func testMultiTrailingClosure() {
  var input = Parser("""
    foo { _ in
    }anotherClosure: { _ in
    }
    """)
  let call = ExprSyntax.parse(from: &input)

  let formatted = call.formatted()
  assertStringsEqualWithDiff(formatted.description, """
    foo { _ in
    } anotherClosure: { _ in
    }
    """)
}
```

Produces:
```
failed - Actual output (+) differed from expected output (-):
–foo { _ in
–} anotherClosure: { _ in
+foo { _ in 
+}anotherClosure: { _ in 
 }
```
ie. extra space after `in` and no space after `}`.

ahoppen:
Fixed by https://github.com/apple/swift-syntax/pull/1569.

PR #1511, opened 2023-04-09
Add test cases for spaces between attributes

### LINKED ISSUE #1504: `BasicFormat` doesn't add spaces between attributes
### Description

When using `AttributeListBuilder`, the build result (formatted by `BasicFormat`) doesn't have spaces between each attributes, which is ill-formed in Swift.

### Steps to Reproduce

```swift
let builder = AttributeListSyntax {
  AttributeSyntax("@inlinable")
  AttributeSyntax("@discardableResult")
}
print(builder) // @inlinable@discardableResult
```

kimdv:
Add test cases for #1504

### DIFF
--- a/Tests/SwiftSyntaxBuilderTest/FunctionTests.swift
+++ b/Tests/SwiftSyntaxBuilderTest/FunctionTests.swift
@@ -165,6 +165,47 @@ final class FunctionTests: XCTestCase {
       ),
       #line: (
         DeclSyntax(
+          """
+          @discardableResult
+          public func foo() -> String {
+            return "foo"
+          }
+          """
+        ),
+        """
+        @discardableResult
+        public func foo() -> String {
+          return "foo"
+        }
+        """
+      ),
+      #line: (
+        DeclSyntax(
+          FunctionDeclSyntax(
+            attributes: [.attribute(AttributeSyntax("@inline(__always)")), .attribute(AttributeSyntax("@discardableResult"))],
+            modifiers: [DeclModifierSyntax(name: .keyword(.public))],
+            identifier: TokenSyntax.identifier("foo"),
+            signature: FunctionSignatureSyntax(
+              input: ParameterClauseSyntax(
+                parameterList: FunctionParameterListSyntax {}
+              ),
+              output: ReturnClauseSyntax(
+                returnType: SimpleTypeIdentifierSyntax(name: .identifier("String"))
+              )
+            ),
+            bodyBuilder: {
+              StmtSyntax(#"return "foo""#)
+            }
+          )
+        ),
+        """
+        @inline(__always) @discardableResult public func foo() -> String {
+            return "foo"
+        }
+        """
+      ),
+      #line: (
+        DeclSyntax(
           FunctionDeclSyntax(
             modifiers: [DeclModifierSyntax(name: .keyword(.public)), DeclModifierSyntax(name: .keyword(.static))],
             identifier: TokenSyntax.identifier("=="),
--- a/CodeGeneration/Sources/SyntaxSupport/CommonNodes.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/CommonNodes.swift
@@ -16,8 +16,7 @@ public let COMMON_NODES: [Node] = [
     name: "CodeBlockItemList",
     nameForDiagnostics: nil,
     kind: "SyntaxCollection",
-    element: "CodeBlockItem",
-    elementsSeparatedByNewline: true
+    element: "CodeBlockItem"
   ),
 
   // code-block-item = (decl | stmt | expr) ';'?
--- a/CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
@@ -111,8 +111,7 @@ public let DECL_NODES: [Node] = [
     name: "AccessorList",
     nameForDiagnostics: nil,
     kind: "SyntaxCollection",
-    element: "AccessorDecl",
-    elementsSeparatedByNewline: true
+    element: "AccessorDecl"
   ),
 
   // (value)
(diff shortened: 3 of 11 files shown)


Sources/SwiftBasicFormat/generated/BasicFormat.swift
```diff
@@ -245,6 +247,8 @@ open class BasicFormat: SyntaxRewriter {
      (.postfixQuestionMark, .rightAngle), // Ensures there is not a space in `ContiguousArray<RawSyntax?>`
      (.postfixQuestionMark, .rightParen): // Ensures there is not a space in `myOptionalClosure?()`
       return false
+    case (.identifier, .keyword):
```
stevapple, line 250: Add a comment explaining what it’s going to solve?

CodeGeneration/Sources/SyntaxSupport/AttributeNodes.swift
```diff
@@ -19,7 +19,8 @@ public let ATTRIBUTE_NODES: [Node] = [
     element: "Syntax",
     elementName: "Attribute",
     elementChoices: ["Attribute", "IfConfigDecl"],
-    omitWhenEmpty: true
+    omitWhenEmpty: true,
+    elementsSeparatedByNewline: true
```
stevapple, line 23: We may need `elementsSeparatedBySpace` instead… Splitting the attribute list with newlines looks terrible, especially when they’re applied to a declaration without leading newline trivia.

Tests/SwiftSyntaxBuilderTest/FunctionTests.swift
```diff
@@ -163,6 +163,45 @@ final class FunctionTests: XCTestCase {
...
+        """
+        @discardableResult public func foo() -> String {
+            return "foo"
+        }
+        """
+      ),
+      #line: (
+        DeclSyntax(
+            FunctionDeclSyntax(
+                attributes: [.attribute(AttributeSyntax("@discardableResult"))],
```
stevapple, line 183: Try multiple attributes here?

stevapple (changes requested):

kimdv:
Do you think we should add newline or a whitespace between @stevapple?

stevapple:
I would prefer whitespaces to newlines for wherever it's applicable (aka. legal in Swift).

kimdv:
I've been looking into some swift code and cannot conclude what is the best practice to do or if the `BasicFormat` should support it. 
IMO a newline is fine so it will generate valid swift and we can use what we already have. 

Then we can also avoid having the special handling `case (.identifier, .keyword):` inside `BasicFormat`

What do you think of that?

stevapple:
Newlines are not very bad, but the current implementation will produce
```swift
@inlinable
@discardableResult
@available(*, deprecated, message: "Use print directly instead") public func hello() -> Int {
    print("Hello world!")
    return 0
}
```
which looks really weird.

The main reason to prefer whitespace is that, when we want to a newline we can always add `.with(\.trailingTrivia, .newline)`, but if we default to newline the reverse just won’t work.

stevapple:
Separating attributes with newlines maybe more troublesome than it seems… After some investigation.

```swift

let buildable = FunctionDeclSyntax(
  identifier: "hello",
  signature: .init(
    input: .init(
      parameterList: [
        FunctionParameterSyntax(
          attributes: [.attribute("@NameBuilder")],
          firstName: "name",
          type: AttributedTypeSyntax(
            attributes: [.attribute("@escaping"), .attribute("@Sendable")],
            baseType: FunctionTypeSyntax(
              arguments: [],
              output: .init(returnType: TypeSyntax("String"))
            )
          )
        )
      ]
    )
  ),
  body: .init(
    statements: [
      .init(item: .decl(#"print("Hello, \(name())!")"#))
    ]
  )
)
print(buildable.formatted())
```

Gives:
```swift
func hello(
    @NameBuilder
    name: 
    @escaping
    @Sendable
    () -> String) {print("Hello, \(name())!")
}
```

Which is, IMO, valid but almost unacceptable.

kimdv:
I see.. hmm.
Then we either should handle case by case or only add white space and let users add new line.

stevapple:
I would suggest the latter, leaving case-by-case formatting for `swift-format`.

ahoppen:
> At least we should have a way to opt out of these new lines, which I didn’t see at the moment.

Just for the record, I don’t think we should add many customization options to `BasicFormat`. It is called `BasicFormat` for a reason: It’s goal is to format a SwiftSyntax tree in a way that, most importantly, parses and, secondly, doesn’t look terrible. I am fine with requiring users to subclass `BasicFormat` if they want to customize it.

stevapple:
@ahoppen I didn’t meant to add the option to `BasicFormat` actually. In existing APIs we already can use explicit newlines over spaces by setting the trivia manually, without touching `BasicFormat` at all.

You can see my use case in https://github.com/teco-project/teco-code-generators, which is used to generate ~1 million lines of nice-looking code with `SwiftSyntaxBuilder`, with little modification to `BasicFormat`.

kimdv:
@ahoppen changed this to add some test cases that for #1504

kimdv:
@swift-ci please test

Tests/SwiftSyntaxBuilderTest/SourceFileTests.swift
```diff
@@ -31,13 +31,12 @@ final class SourceFileTests: XCTestCase {
     assertBuildResult(
       source,
       """
-      import Foundation
-      import UIKit
+      import Foundation 
+      import UIKit
```
stevapple, line 35: Looks like a regression here. Is this intentional?

stevapple (approved):
Looks good overall!

stevapple:
Should there be test cases for #1472 and #1551 too?

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test windows

Tests/SwiftSyntaxBuilderTest/TriviaTests.swift
```diff
@@ -83,12 +83,14 @@ final class TriviaTests: XCTestCase {
           AttributeSyntax(attributeName: TypeSyntax("A")).with(\.trailingTrivia, .space)
           AttributeSyntax(attributeName: TypeSyntax("B")).with(\.trailingTrivia, .space)
         },
-        "@A @B "
+        """
+        @A @B
```
ahoppen, line 87: I would prefer this string literal to continue being single line so you can actually see the trailing whitespace.

ahoppen (approved):

kimdv:
@swift-ci please test

PR #1517, opened 2023-04-12
Address default parameter regressions and add tests for `ForInStmtSyntax`

### LINKED ISSUE #1515: Various `ForInStmtSyntax` builder regressions
### Description

We've found a series of regressions when building `ForInStmtSyntax`, including:
- `awaitKeyword` is defaulted to `.keyword(.await)` instead of `nil`, which is unexpected for most users;
- `BasicFormat` fails to add spacing after `awaitKeyword`;
- `BasicFormat` fails to add newline after left brace.

### Steps to Reproduce

```swift 
let buildable = ForInStmtSyntax(pattern: PatternSyntax("element"), sequenceExpr: ExprSyntax("array")) {
  DeclSyntax("print(element)")
}
print(buildable.formatted())
```
Outputs:
```swift
for awaitelement in array {print(element)
}
```
Expected:
```swift
for element in array {
    print(element)
}
```

stevapple:
Fixes #1515


ahoppen:
The problem here is that for some reason we are generating the `.keyword(.await)` default parameter for the initializer even though the child is marked as optional. I’m not sure why that is but I think this would be the thing that needs fixing. The child declaration as `kind: .token(choices: [.keyword(text: "await")]),` is correct.

Could you look into what’s causing us to generate `.keyword(.await)` instead of `nil`, which is what I would expect?

stevapple:
It turns out to be a pretty large fix… We used to have two different implementations of `defaultInitialization`, and both are not really correct. One of them caused the `awaitKeyword` regression in `SwiftSyntaxBuilder`, while the other synthesized wrong defaults for a series of identifier and operator tokens in `SwiftSyntax`.

ahoppen (approved):
Thank you 🙏🏽 This is so much better. And it fixes the case where we incorrectly generate `IdentifierToken` default values 😍

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

kimdv:
Looks like there is some formatting issues.
Can you format it again? 😄

stevapple:
> Can you format it again? 😄

Done:)

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test windows

PR #1520, opened 2023-04-12
Fix escape handling for emoji `#️⃣` in string literals

kishikawakatsumi:
This pull request addresses an issue with the escape handling when counting the number of `#` characters in string literals. Previously, the code didn't consider the emoji `#️⃣` presence. This resulted in incorrect escape counts.

For example, the string literal generated from the string `foo "#️⃣"bar` would be `##"foo "#️⃣"bar "##`, but the current implementation generates `#"foo "#️⃣"bar "#`. This literal will result in a compile error because the escape is incomplete.


ahoppen (approved):
Thanks. This looks good to me

ahoppen:
@swift-ci Please test

ahoppen:
Thanks for contributing to Swift @kishikawakatsumi!

PR #1526, opened 2023-04-13
Don't guard assertions behind `#if DEBUG`

ahoppen:
Some assertions were still guarded by a `#if DEBUG`. Since we are now also enabling assertions in release builds if `SWIFTSYNTAX_ENABLE_ASSERTIONS` is set, this is no longer correct. Convert most of them to plain preconditions. I have measured and could not detect a performance difference when parsing MovieSwiftU.


kimdv (approved):

ahoppen:
@swift-ci Please test

ahoppen:
I just tested and couldn’t find a significant performance regression when linting MovieSwiftUI using `swift-format`.

PR #1535, opened 2023-04-13
[bazel] Add remaining libraries

keith:
These were the only ones not already added, in case folks need them


keith:
@swift-ci please test

PR #1537, opened 2023-04-14
Update RawSyntaxValidationFile.swift

ahoppen:
I only updated the generated version of it, not the original. Whoops.


ahoppen:
@swift-ci Please test

kimdv (approved):

PR #1540, opened 2023-04-14
Improved Diagnostic for C-style for loop

CippoX:
Improved Diagnostic for C-style for loop using `getTokens` introduced by #1464

### DIFF
--- a/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
+++ b/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
@@ -30,6 +30,31 @@ fileprivate func getTokens(between first: TokenSyntax, and second: TokenSyntax) 
   return tokens
 }
 
+fileprivate func getChildren(between first: Syntax, and second: Syntax) -> [Syntax] {
+  let parent = first.parent
+  let children = parent?.children(viewMode: .sourceAccurate)
+  let map = children?.compactMap({ $0.as(Syntax.self) })
+  var res: [Syntax] = []
+  
+  var inTheLoop = false
+  
+  for m in map! {
+    if inTheLoop {
+      if m == second {
+        res.append(m)
+        break
+      } else {
+        res.append(m)
+      }
+    } else if m == first {
+      res.append(m)
+      inTheLoop = true
+    }
+  }
+  
+  return res
+}
+
 fileprivate extension TokenSyntax {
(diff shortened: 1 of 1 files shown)


kimdv:
@swift-ci please test

CippoX:
I have to think about the right approach

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -662,33 +662,21 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
-          Syntax(node.unexpectedBetweenTypeAnnotationAndInKeyword),
-          Syntax(node.inKeyword),
-          Syntax(node.unexpectedBetweenInKeywordAndSequenceExpr),
-          Syntax(node.sequenceExpr),
-          Syntax(node.unexpectedBetweenSequenceExprAndWhereClause),
-          Syntax(node.whereClause),
-          Syntax(node.unexpectedBetweenWhereClauseAndBody),
-          Syntax(unexpectedCondition),
-        ] as [Syntax?]).compactMap({ $0 }),
+        highlights: (getTokens(between: firstToken, and: lastToken).map { Optional(Syntax($0)) }).compactMap({ $0 }),
```
ahoppen, line 672: Why do you need to explicitly spell out `Optional` here?

CippoX: I was thinking about replicating `as [Syntax?]`,
However, the test failing is `testColoringWithHighlights()`. The problem seems to be that with `getTokens` we are creating an array of tokens, which is not what was happening previously. Do you think I should find a way to walk through `node`'s children?

CippoX:
What about this solution? Obviously, it's very primitive, if you think that's the right approach I'll refine it

ahoppen:
If we need to write a new function for this, I don’t think it’s worth it. The C-style for loop diagnostic is a bit of a weirdo anyway and unless a new use case for `getNodes(between:and:)` comes up, I don’t think it’s worth introducing it.

CippoX:
@ahoppen seems reasonable, what about this last attempt? Otherwise, I'll just close the PR

ahoppen:
This makes implicit assumptions about the number of children of a for loop that we are not interested in, which might change  if we changed the structure of for loops. I’d just close the PR, sorry.

PR #1543, opened 2023-04-14
Fix debugInitCall

ahoppen:
I broke `debugInitCall` when I removed the custom mirrors for syntax nodes. Fix it again, also fixing some more issues along the way that previously caused us to generate Swift code that didn’t compile.

### DIFF
--- a/CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/ChildNameForKeyPathFile.swift
+++ b/CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/ChildNameForKeyPathFile.swift
@@ -20,7 +20,8 @@ let childNameForKeyPathFile = SourceFileSyntax(leadingTrivia: copyrightHeader) {
     """
     /// If the keyPath is one from a layout structure, return the property name
     /// of it.
-    internal func childName(_ keyPath: AnyKeyPath) -> String?
+    @_spi(RawSyntax)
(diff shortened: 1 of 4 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

kimdv:
@swift-ci please test windows

bnbarham (approved):
I'd personally prefer not having the runtime hackery by making the assumption that `SyntaxChildChoices` is within a type and then:
  1. Not supporting printing `SyntaxChildChoices` directly (can override to print something to that effect)
  2. Pass down the parent type name when making the recursive call

But it's just a debug call and only in _SwiftSyntaxTestSupport so I don't feel *super* strongly about that. Any thoughts @rintaro?

ahoppen:
I found a fairly elegant solution without calling into the runtime.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

bnbarham:
> I found a fairly elegant solution without calling into the runtime.

Ah, I forgot we had `parent` :)

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test macOS

PR #1544, opened 2023-04-14
Remove SwiftSyntaxParser module

ahoppen:
This removes the `SwiftSyntaxParser` module, which was just a wrapper around `SwiftParser` to ease the transition for clients that previously used `SwiftSyntaxParser`. Everybody should be using `SwiftParser` by now, so remove the deprecated module.

This migrates all test cases that still provide value to other modules.


ahoppen:
@swift-ci Please test macOS

keith (approved):

Tests/SwiftSyntaxParserTest/Inputs/closure.swift
```diff
@@ -1,3 +0,0 @@
-// A closure without a signature. The test will ensure it stays the same after
```
bnbarham, line 1: I haven't looked through all these, but I assume you moved ones you thought necessary and deleted otherwise?

bnbarham (approved):
💃

Tests/SwiftSyntaxParserTest/Inputs/closure.swift
ahoppen: Yes, most of them became initializers calls to create syntax trees.

Tests/SwiftSyntaxTest/SyntaxVisitorTests.swift
```diff
@@ -0,0 +1,248 @@
...
+  /// func foo() {
+  ///   public func foo() {
+  ///     func foo() {
+  ///       /*Unknown token */0xG
+  ///     }
+  ///   }
+  /// }
+  /// ```
+  ///
+  /// The source file is hard-coded so this test case doesn't need to depend on the parser.
```
bnbarham, line 38: Is there some reason not to depend on the parser here?

Tests/SwiftSyntaxTest/SyntaxVisitorTests.swift
```diff
@@ -0,0 +1,248 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import XCTest
+import SwiftSyntax
+import SwiftParser
+import SwiftSyntaxBuilder
+import _SwiftSyntaxTestSupport
+
+extension Parser {
```
bnbarham, line 19: Is this used by another file somewhere?

Tests/SwiftSyntaxTest/SyntaxVisitorTests.swift
ahoppen: It’s just that this test is in `SwiftSyntaxTest` and I don’t think we should be depending on the parser from `SwiftSyntaxTest` because it’s lower level. Also, it didn’t make any sense for me to move this file to `SwiftParserTest` because that’s not where it belongs, so I ended up with this solution, which I don’t find super terrible.

Tests/SwiftSyntaxTest/SyntaxVisitorTests.swift
ahoppen: No, I can make it file private.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

PR #1548, opened 2023-04-14
Add notes to assertion

kimdv:
As in #1467 as tought that it might also makes sense to force testing for notes. 

Let me hear what you think before I continue fixing the other failing 223 test cases @ahoppen

### DIFF
--- a/Tests/SwiftParserTest/DeclarationTests.swift
+++ b/Tests/SwiftParserTest/DeclarationTests.swift
@@ -132,11 +132,20 @@ final class DeclarationTests: XCTestCase {
         """
     )
     assertParse(
-      "class B<where g1️⃣",
+      "class Bℹ️<where g1️⃣",
       diagnostics: [
-        DiagnosticSpec(message: "expected ':' or '==' to indicate a conformance or same-type requirement"),
-        DiagnosticSpec(message: "expected '>' to end generic parameter clause", fixIts: ["insert '>'"]),
-        DiagnosticSpec(message: "expected member block in class", fixIts: ["insert member block"]),
+        DiagnosticSpec(
+          message: "expected ':' or '==' to indicate a conformance or same-type requirement"
+        ),
+        DiagnosticSpec(
+          message: "expected '>' to end generic parameter clause",
+          notes: [NoteSpec(message: "to match this opening '<'")],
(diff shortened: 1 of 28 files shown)


ahoppen:
That would be great step for better testing. 🥇

Tests/SwiftParserTest/DeclarationTests.swift
```diff
@@ -737,22 +745,26 @@ final class DeclarationTests: XCTestCase {
...
     )
   }
 
   func testExpressionMember() {
     assertParse(
       """
-      struct S {1️⃣
-        /2️⃣ ###line 25 "line-directive.swift"3️⃣
+      struct S ℹ️{1️⃣
+        🔟/2️⃣ ###line 25 "line-directive.swift"3️⃣
```
kimdv, line 757: Before I continue @ahoppen I got into an issue here. 

We normally use ℹ️ for Note  locationmarkers.
But in this case here we have multiple locations. 

I saw that you in #1550 add line and column. 
1. We could start to use same approach here, then we never hit this again. 
The bad, as it makes it hard to see when we receive things like 2:37.
2. We can also just use another number but then we mix with where we expect the diagnosis.

ahoppen: > I saw that you in https://github.com/apple/swift-syntax/pull/1550 add line and column.

My thinking here was that this is for an assertion function that authors of macros (outside of the swift-syntax repository) would use. Since I think the location marker concept has a non-trivial learning curve, I resorted to just using line/columns which everybody understands. For contributors of swift-syntax, I think this learning curve is acceptable because it *significantly* improves our tests.

---

I think there are already some test cases that contain multiple notes and we just the normal `1️⃣` etc. markers for those. I think that’s fine. 

Also wherever possible, I think tests should be split so they ideally only emit a single note. But that’s not how all our existing tests are structured (mostly because the my migration script wasn’t able to split them to a finer-grained level) and it’s not always possible.

kimdv:
@ahoppen this is ready for review now 😅

ahoppen (approved):
Thanks 🙏🏽

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test windows

PR #1550, opened 2023-04-15
Add a public function that can be used to test macro expansions

ahoppen:
Testing macros should be as easy as possible and we should thus make the  `assertMacroExpansion` function public.

While making the function public, I made the following changes:
- Changed `diagnosticMessages` to expect `DiagnsoticSpec` similar to the parser tests. The only difference is that they take line and columns instead of location markers because the location markers have a slighgly high learning curve IMO.

Slightly unrelated, I also made the following changes:
- Made a few minor improvements to `SwiftParserTest/Assertions.swift`
- Fixed a bug in MacroApplication where the wrong syntax node was passed as the `node` parameter` to `addDiagnostics`, leading to a highlight that was bigger than expected.


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test macOS

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test macOS

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Linux

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Package.swift
```diff
@@ -56,6 +56,7 @@ let package = Package(
     .library(name: "SwiftSyntax", type: .static, targets: ["SwiftSyntax"]),
     .library(name: "SwiftSyntaxBuilder", type: .static, targets: ["SwiftSyntaxBuilder"]),
     .library(name: "SwiftSyntaxMacros", type: .static, targets: ["SwiftSyntaxMacros"]),
+    .library(name: "SwiftSyntaxMacrosTestSupport", type: .static, targets: ["SwiftSyntaxMacrosTestSupport"]),
```
DougGregor, line 59: We should add this to the CMake build as well?

Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
```diff
@@ -0,0 +1,304 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import _SwiftSyntaxTestSupport
+import SwiftBasicFormat
+import SwiftDiagnostics
+import SwiftParser
+import SwiftSyntax
+import SwiftSyntaxMacros
+import XCTest
```
DougGregor, line 19: ... or is this why we don't have it in the CMake build?

Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
```diff
@@ -0,0 +1,304 @@
...
+///   - testFileName: The name of the test file name to use.
+public func assertMacroExpansion(
+  _ originalSource: String,
+  expandedSource: String,
+  diagnostics: [DiagnosticSpec] = [],
+  macros: [String: Macro.Type],
+  testModuleName: String = "TestModule",
+  testFileName: String = "test.swift",
+  file: StaticString = #file,
+  line: UInt = #line
```
DougGregor, line 272: This is a really nice API 👍

DougGregor (approved):

Package.swift
ahoppen: I don’t think there’s any reason why this should be in the toolchain. So no need to add it to the CMake build…

ahoppen:
Note to self: This requires the BasicFormat improvements I’m merging in other PRs.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
```diff
@@ -0,0 +1,305 @@
...
+    assertStringsEqualWithDiff(
+      highlightedCode.trimmingTrailingWhitespace(),
+      highlight.trimmingTrailingWhitespace(),
+      "highlight does not match",
+      file: spec.originatorFile,
+      line: spec.originatorLine
+    )
+  }
+  if diag.notes.count != spec.notes.count {
+    XCTFail(
```
bnbarham, line 221: Can we output both notes, ie. actual and expected? Same with fix-it's below.

Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
```diff
@@ -0,0 +1,305 @@
...
+    )
+  } else {
+    for (note, expectedNote) in zip(diag.notes, spec.notes) {
+      assertNote(note, in: tree, expected: expectedNote)
+    }
+  }
+  if diag.fixIts.count != spec.fixIts.count {
+    XCTFail(
+      """
+      Expected \(spec.fixIts.count) Fix-Its but received \(diag.fixIts.count):
```
bnbarham, line 237: This is in a few places - IMO I wouldn't capitalize `fix-its`.

Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
```diff
@@ -0,0 +1,305 @@
...
+
+/// Assert that expanding the given macros in the original source produces
+/// the given expanded source code.
+///
+/// - Parameters:
+///   - originalSource: The original source code, which is expected to contain
+///     macros in various places (e.g., `#stringify(x + y)`).
+///   - expandedSource: The source code that we expect to see after performing
+///     macro expansion on the original source.
+///   - diagnostics:
```
bnbarham, line 258: Missing comment

Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
```diff
@@ -0,0 +1,305 @@
...
+///   - macros: The macros that should be expanded, provided as a dictionary
+///     mapping macro names (e.g., `"stringify"`) to implementation types
+///     (e.g., `StringifyMacro.self`).
+///   - testModuleName: The name of the test module to use.
+///   - testFileName: The name of the test file name to use.
```
bnbarham, line 263: Is it useful to expose these? It doesn't actually matter for anything right?

Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
```diff
@@ -0,0 +1,305 @@
...
+public func assertMacroExpansion(
+  _ originalSource: String,
+  expandedSource: String,
+  diagnostics: [DiagnosticSpec] = [],
+  macros: [String: Macro.Type],
+  testModuleName: String = "TestModule",
+  testFileName: String = "test.swift",
+  indentationWidth: Trivia = .spaces(4),
```
bnbarham, line 271: What are your thoughts on `indentWidth` vs `indentationWidth`?

Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
```diff
@@ -0,0 +1,305 @@
...
+
+  assertStringsEqualWithDiff(
+    expandedSourceFile.description.trimmingTrailingWhitespace(),
+    expandedSource.trimmingTrailingWhitespace(),
+    file: file,
+    line: line
+  )
+
+  if context.diagnostics.count != diagnostics.count {
+    XCTFail(
```
bnbarham, line 292: As with notes/fix-its, IMO printing both expected and actuals useful.

Tests/SwiftSyntaxMacrosTest/MacroSystemTests.swift
```diff
@@ -721,51 +673,56 @@ public let testMacros: [String: Macro.Type] = [
 ]
 
 final class MacroSystemTests: XCTestCase {
+  private let indentationWidth: Trivia = .spaces(2)
```
bnbarham, line 676: `static`?

Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
ahoppen: I don’t have strong feelings. Do you?

Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
ahoppen: You could use them in diagnostic messages, right?

Tests/SwiftSyntaxMacrosTest/MacroSystemTests.swift
ahoppen: But then I need to access it with `Self.indentationWidth` which is a little to verbose for my taste.

Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
ahoppen: I just checked. SwiftSyntax general prefers `indentation` over `indent`, so I’d prefer to stick with `indentation` for consistency.

PR #1557, opened 2023-04-18
Removed TODOs in testRecovery102/103/104

CippoX:
Fixing testRecovery102/103/104

### DIFF
--- a/Tests/SwiftParserTest/translated/RecoveryTests.swift
+++ b/Tests/SwiftParserTest/translated/RecoveryTests.swift
@@ -1273,7 +1273,6 @@ final class RecoveryTests: XCTestCase {
       }
       """,
       diagnostics: [
-        // TODO: Old parser expected error on line 2: array types are now written with the brackets around the element type, Fix-It replacements: 17 - 17 = '[', 20 - 21 = ''
         DiagnosticSpec(message: "unexpected code '[0]' in function")
       ]
     )
@@ -1289,7 +1288,6 @@ final class RecoveryTests: XCTestCase {
       }
       """,
       diagnostics: [
-        // TODO: Old parser expected error on line 2: array types are now written with the brackets around the element type, Fix-It replacements: 17 - 17 = '[', 20 - 21 = ''
         DiagnosticSpec(message: "unexpected code '[0_1]' in function")
       ]
     )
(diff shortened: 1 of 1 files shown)


CippoX:
@ahoppen I was wondering if this was the right way to solve it.

CippoX:
I also considered creating an ad-hoc node to represent this common error, but it seems excessive

Sources/SwiftParser/Types.swift
```diff
@@ -224,13 +224,23 @@ extension Parser {
...
+    case TokenSpec(.identifier):
+      var lookahead = self.lookahead()
+      lookahead.consumeAnyToken()
+      if !isRawKeyPathExprSyntax,
+         let _ = lookahead.consume(if: .leftSquareBracket, followedBy: .integerLiteral),
+         lookahead.currentToken.rawTokenKind == .rightSquareBracket {
+        base = RawTypeSyntax(self.parseCollectionType())
+      } else {
+        base = self.parseTypeIdentifier()
+      }
```
ahoppen, line 241: Hmm, this is some non-trivial lookahead overhead that we need to pay for *every* type and the diagnostic is really not that important. Swift has been spelling arrays as `[Int]` instead of `Int[]` since beta 2 or 3 of Swift 1.0 C-programmers will stumble across bigger differences in the Swift language than the spelling of array types. With the current implementation, I think the performance cost is just not worth the benefit. 

I think either we find a solution to emit this diagnostic without doing lookahead or we just ditch it (i.e. just remove the TODO). And I wouldn’t be terribly opposed to just ditching it. I’ll leave it up to you how much effort you want to put into finding a non-lookahead solution.

CippoX: Having thought about it quite a bit, I believe that a non-lookahead solution is beyond my capability. The only thing i cold find was peeking for `[` in order to reduce the number of lookaheads. Do you want me to remove the TODOs?

ahoppen: Yes, I think just removing the TODOs is reasonable.

kimdv (approved):

kimdv:
@swift-ci please test

ahoppen (approved):
Thanks. And sorry for all the work you did only to arrive at the simplest solution that doesn’t even change any code.

ahoppen:
@swift-ci Please test Windows

kimdv:
@swift-ci please test windows

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

CippoX:
No problem. I’ll look for something else to work on in my free time.

PR #1558, opened 2023-04-19
Improve implementation of BasicFormat (yet again)

ahoppen:
This re-writes the implementation of BasicFormat once again to better handle user-indented code.


ahoppen:
@swift-ci Please test

Sources/SwiftBasicFormat/Trivia+FormatExtensions.swift
```diff
@@ -0,0 +1,99 @@
...
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
+extension Trivia {
+  /// Removes all blanks that is trailing before a newline trivia,
+  /// effectively making sure that lines don't end with a blank
+  func trimingTrailingBlanksBeforeNewline(isBeforeNewline: Bool) -> Trivia {
```
kimdv, line 18: This is really nice! 
Looking forward to how many bytes we can save in Swift Syntax 😆

Sources/SwiftBasicFormat/BasicFormat.swift
```diff
@@ -0,0 +1,278 @@
...
+  /// The trivia by which tokens should currently be indented.
+  public var currentIndentationLevel: Trivia {
+    return indentationStack.last!
+  }
+
+  /// For every token that is being put on a new line but did not have
+  /// user-specified indentation, the generated indentation.
+  ///
+  /// This is used as a reference-point to indent user-indented code.
+  var anchorPoints: [TokenSyntax: Trivia] = [:]
```
kimdv, line 32: Should this also be public?
Or is it meant to be internal.

Sources/SwiftBasicFormat/BasicFormat.swift
```diff
@@ -0,0 +1,278 @@
...
+
+  public init(indentationIncrement: Trivia = .spaces(4), initialIndentation: Trivia = []) {
```
kimdv, line 34: Not sure if I like the default value is `[]`. I think it's a but hard to understand maybe for new developers.

But on the the other hand I cannot find any other name than `.zero`, `.none`, `.empty`. And I'm not convinced they fit either.

kimdv (approved):
This is looking really great! 😍

Huge improvement IMO

Sources/SwiftBasicFormat/BasicFormat.swift
```diff
@@ -0,0 +1,278 @@
...
+    } else if requiresLeadingBlank(token) {
+      // Add a leading space if the token requires it unless
+      //  - it already starts with a blank or
+      //  - the previous token ends with a blank after the rewrite
+      if !(leadingTrivia.first?.isBlank ?? false) && !previousTokenWillEndWithBlank {
+        leadingTrivia += .space
+      }
+    }
+
+    if leadingTrivia.indentation(isOnNewline: previousTokenWillEndInNewline) == [] {
```
bnbarham, line 220: If `containsIndentation` and `indentation` matched up, this could just be a contains check. It would be nice if we could remove anchor points, but I can't think of a fast way to do that (we can remove all tokens in the node in the post visit but... that doesn't seem worth the performance hit).

Sources/SwiftBasicFormat/BasicFormat.swift
```diff
@@ -0,0 +1,278 @@
...
+    var trailingTriviaIndentation = self.currentIndentationLevel
+
+    // If the trivia contain user-defined indentation, find their anchor point
+    // and indent the token relative to that anchor point.
+    if leadingTrivia.containsIndentation(isOnNewline: previousTokenWillEndInNewline),
+      let anchorPointIndentation = self.anchorPointIndentation(for: token)
+    {
+      leadingTriviaIndentation = anchorPointIndentation
+    }
+    if trailingTriviaAndNextTokensLeadingTriviaWhitespace.containsIndentation(isOnNewline: previousTokenWillEndInNewline),
```
bnbarham, line 247: Do we really care about the trailing? There shouldn't be a newline there anyway. That would also let us get rid of all the `isOnNewline`.

bnbarham (commented):
Nice, seems to work pretty well. Might be good to add a `stripIndentation` so the BasicFormat indentation can be preferred even if the string interpolation is being used with whatever indentation an author had there.

Sources/SwiftBasicFormat/BasicFormat.swift
ahoppen: I made it private

Sources/SwiftBasicFormat/BasicFormat.swift
ahoppen: We used to have `Trivia.zero` but IIRC @bnbarham removed it because it was redundant with the empty array. Do you have strong options on whether we should add it (or `.empty`) back in again?

---

In line 36: Well, `indentationStack` is an array of trivia again. And arrays of arrays aren’t that uncommon, so I think that should be fine…

Sources/SwiftBasicFormat/BasicFormat.swift
```diff
@@ -0,0 +1,278 @@
...
+  /// How much indentation should be added at a new indentation level.
+  public let indentationIncrement: Trivia
+
+  /// As we reach a new indendation level, its indentation will be added to the
+  /// stack. As we exit that indentation level, the indendation will be popped.
+  public var indentationStack: [Trivia]
+
+  /// The trivia by which tokens should currently be indented.
+  public var currentIndentationLevel: Trivia {
+    return indentationStack.last!
```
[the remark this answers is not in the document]
ahoppen: Good hint. I’ll add an assertion to `popIndentationLevel` that will break if you remove the last indentation level. Then we can safely unwrap here and trap closer to the actual cause of the failure.

Sources/SwiftBasicFormat/Trivia+FormatExtensions.swift
```diff
@@ -0,0 +1,99 @@
...
+      }
+    }
+    return false
+  }
+
+  /// Returns the first indentation level of the first non-empty line in this
+  /// trivia or `nil` if there is no newline.
+  /// If `isOnNewline` is `true`, then the trivia is preceeded by a newline
+  /// character and any indentation at its start will be returned.
+  func indentation(isOnNewline: Bool) -> Trivia? {
```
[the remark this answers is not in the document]
ahoppen: I unified the implementations now.

Sources/SwiftBasicFormat/BasicFormat.swift
```diff
@@ -0,0 +1,278 @@
...
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
+open class BasicFormat: SyntaxRewriter {
+  /// How much indentation should be added at a new indentation level.
+  public let indentationIncrement: Trivia
```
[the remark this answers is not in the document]
ahoppen: I knew there was a better name for this but I couldn’t remember it last night 😆

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/TriviaPiecesFile.swift
```diff
@@ -264,4 +264,57 @@ let triviaPiecesFile = SourceFileSyntax(leadingTrivia: copyrightHeader) {
...
+        }
+        SwitchCaseSyntax("default:") {
+          StmtSyntax("return false")
+        }
+      }
+    }
+
+    DeclSyntax(
+      """
+      public var isIndentationWhitespace: Bool {
```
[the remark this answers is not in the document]
ahoppen: I renamed these accessors and got rid of the term “blank” entirely because I found it ambiguous:
- `isNewline`: Obvious, though I have very little opinion on whether this should include form feeds and vertical tabs. I think it doesn’t make a difference in practice and if `Swift.String` considers them newlines, I don’t see a reason why we shouldn’t
- `isSpaceOrTab`: Let me know if you have a better name for it, I couldn’t find one and this is very clear IMO
- `isWhitespace`: `isSpaceOrTab || isNewline`

Tests/SwiftBasicFormatTest/BasicFormatTests.swift
```diff
@@ -0,0 +1,193 @@
...
+        // ABC
+        someFunc(a: 1,
+                 b: 1)
+      }
+      """
+
+    assertFormatted(source: source, expectedFormatted: source)
+  }
+
+  func testAlreadyIndentedWithComment2() {
```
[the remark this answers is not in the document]
ahoppen: I went with the following definition of a trivia’s indentation now, which is reasonable IMO. And by that definition, this is not indented…
```swift
  /// Returns the indentation of the last trivia piece in this trivia that is
  /// not a whitespace.
```

Sources/SwiftBasicFormat/BasicFormat.swift
```diff
@@ -0,0 +1,278 @@
...
+      }
+      if let ancestorsParent = ancestor.parent, childrenSeparatedByNewline(ancestorsParent) {
+        return true
+      }
+    }
+
+    return token.requiresLeadingNewline
+  }
+
+  open func childrenSeparatedByNewline(_ node: Syntax) -> Bool {
```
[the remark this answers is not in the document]
ahoppen: I don’t have strong opinions here. But in an effort to decrease API surface, I’ve made it private

Sources/SwiftBasicFormat/BasicFormat.swift
ahoppen: We have newlines as trailing trivia for multi-line string literals. In that case the newline after the `"""` is trailing trivia of the quotes.

Sources/SwiftBasicFormat/BasicFormat.swift
```diff
@@ -0,0 +1,278 @@
...
+      return previousToken.trailingTrivia.pieces.last?.isNewline ?? false
+    }()
+
+    lazy var nextTokenWillStartWithBlank: Bool = {
+      guard let nextToken = token.nextToken(viewMode: .sourceAccurate) else {
+        return false
+      }
+      return nextToken.leadingTrivia.first?.isBlank ?? false
+        || requiresLeadingBlank(nextToken)
+        || requiresLeadingNewline(nextToken)
```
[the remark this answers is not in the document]
ahoppen: Haha, I just realized I could just remove this property altogether. It wasn’t even called.

Sources/SwiftBasicFormat/BasicFormat.swift
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    var leadingTriviaIndentation = self.currentIndentationLevel
+    var trailingTriviaIndentation = self.currentIndentationLevel
+
+    // If the trivia contain user-defined indentation, find their anchor point
+    // and indent the token relative to that anchor point.
+    if leadingTrivia.containsIndentation(isOnNewline: previousTokenWillEndInNewline),
+      let anchorPointIndentation = self.anchorPointIndentation(for: token)
+    {
```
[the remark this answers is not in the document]
ahoppen: Yes it is, not sure why though…

Sources/SwiftBasicFormat/BasicFormat.swift
```diff
@@ -0,0 +1,278 @@
...
+      break
+    }
+
+    return token.requiresTrailingSpace
+  }
+
+  // MARK: - Formatting a token
+
+  open override func visit(_ token: TokenSyntax) -> TokenSyntax {
+    lazy var previousTokenWillEndWithBlank: Bool = {
```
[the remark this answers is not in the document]
ahoppen: No we can’t because they access e.g. `BasicFormat.requiresTrailingWhitespace`.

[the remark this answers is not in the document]
bnbarham: Ah. Unfortunate, `private` to `BasicFormat` then 😅. Unless there's actually a performance benefit on the lazy here.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Linux

Sources/SwiftBasicFormat/BasicFormat.swift
```diff
@@ -18,12 +18,15 @@ open class BasicFormat: SyntaxRewriter {
...
   /// stack. As we exit that indentation level, the indendation will be popped.
-  public var indentationStack: [Trivia]
+  /// `isUserDefined` is `true` if the indentation was inferred from something
+  /// the user provided manually instead of being inferred from the nesting
+  /// level.
+  public var indentationStack: [(indentation: Trivia, isUserDefined: Bool)]
 
   /// The trivia by which tokens should currently be indented.
   public var currentIndentationLevel: Trivia {
     // `popIndentationLevel` guarantees that there is always one item on the stack.
```
bnbarham, line 28: `decreaseIndentationLevel` now

Sources/SwiftBasicFormat/BasicFormat.swift
```diff
@@ -39,16 +42,16 @@ open class BasicFormat: SyntaxRewriter {
...
 
-  public func pushIndentationLevel(increasingIndentationBy: Trivia) {
-    indentationStack.append(currentIndentationLevel + increasingIndentationBy)
+  public func increaseIndentationLevel() {
+    indentationStack.append((currentIndentationLevel + indentationWidth, false))
   }
 
-  public func popIndentationLevel() {
+  public func decreaseIndentationLevel() {
     precondition(indentationStack.count > 1, "Popping more indentation levels than have been pushed")
```
bnbarham, line 55: I know this should always be the case since we push on init, but given this is going to be running all the time I would personally prefer this just only remove if count > 1. Then this can be `assert` instead.

Same with the `!` down below in `visitPre` - fallback to increasing indentation level normally if it's `nil` (and assert).

Sources/SwiftBasicFormat/BasicFormat.swift
```diff
@@ -60,16 +63,17 @@ open class BasicFormat: SyntaxRewriter {
         !tokenIndentation.isEmpty
       {
         // If the first token in this block is indented, infer the indentation level from it.
-        pushIndentationLevel(increasingIndentationBy: tokenIndentation)
+        let lastNonUserDefinedIndentation = indentationStack.last(where: { !$0.isUserDefined })!.indentation
+        indentationStack.append((indentation: lastNonUserDefinedIndentation + tokenIndentation, isUserDefined: true))
```
bnbarham, line 67: It's so disconcerting not seeing the same `increase` in both branches. Could `increase` take an indent level and if so add `isUserDefined`?
```
public func increaseIndentationLevel(to indentation: Trivia? = nil) {
  if let indentation = indentation {
    indentationStack.append((indentation, true))
  } else {
    indentationStack.append((currentIndentationLevel + indentationWidth, false))
  }
}
```

Sources/SwiftBasicFormat/BasicFormat.swift
```diff
@@ -197,7 +201,25 @@ open class BasicFormat: SyntaxRewriter {
         // don't add a leading newline to the file.
         return true
       }
-      return previousToken.trailingTrivia.endsWithNewline
+      if previousToken.trailingTrivia.pieces.last?.isNewline ?? false {
```
bnbarham, line 204: Why not `if previousToken.trailingTrivia.endsWithNewline`? Just matching up with `stringSegment`? If so, I don't think anything else uses `endsWithNewline` so you may as well delete it.

ahoppen: Because I missed it while changing things to `endsWithWhitespace`. Probably through a series of a rebases or something of the sort. Changed it now.

PR #1561, opened 2023-04-20
Fix a couple cases where the formatting in CodeGeneration was off

ahoppen:
With the new BasicFormat from https://github.com/apple/swift-syntax/pull/1558, these cases will be formatted weirdly, so I want to update them now.

### DIFF
--- a/Sources/SwiftSyntax/generated/syntaxNodes/SyntaxDeclNodes.swift
+++ b/Sources/SwiftSyntax/generated/syntaxNodes/SyntaxDeclNodes.swift
@@ -89,7 +89,8 @@ public struct AccessorDeclSyntax: DeclSyntaxProtocol, SyntaxHashable {
           arena: arena, 
           leadingTrivia: leadingTrivia, 
           trailingTrivia: trailingTrivia
-        )
+        
+      )
       return SyntaxData.forRoot(raw)
     }
     self.init(data)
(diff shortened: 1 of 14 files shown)


ahoppen:
@swift-ci Please test

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxDeclNodes.swift
```diff
@@ -89,7 +89,8 @@ public struct AccessorDeclSyntax: DeclSyntaxProtocol, SyntaxHashable {
           arena: arena, 
           leadingTrivia: leadingTrivia, 
           trailingTrivia: trailingTrivia
-        )
+        
+      )
```
bnbarham, line 93: Weird, but I assume the new will fix this?

bnbarham (approved):

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxTypeNodes.swift
```diff
@@ -71,7 +71,8 @@ public struct ArrayTypeSyntax: TypeSyntaxProtocol, SyntaxHashable {
           arena: arena, 
           leadingTrivia: leadingTrivia, 
           trailingTrivia: trailingTrivia
-        )
+        
+      )
```
kimdv, line 75: What is adding this empty newline. I can't see it in the template file

kimdv (approved):

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxDeclNodes.swift
ahoppen: I’m not entirely sure yet what’s going wrong here. Still need to investigate but I didn’t think it would make sense to investigate in the current `BasicFormat`.

PR #1563, opened 2023-04-21
Explicitly specify how many last versions of Swift we want to support

ahoppen:


ahoppen:
@swift-ci Please test

kimdv (approved):

PR #1564, opened 2023-04-21
Change version dependency on `swift-argument-parser` to from `upToNextMinor` to `upToNextMajor`

ahoppen:
This will make SwiftSyntax more tolerant with regard to which swift-argument-parser version it needs, resulting in fewer version conflicts for users of SwiftSyntax.


ahoppen:
@swift-ci Please test

MaxDesiatov (approved):

kimdv (approved):

PR #1566, opened 2023-04-21
Fix for some wrong attribute fix its

kimdv:
The code is looking the same and there is a lot more attributes that properly have the same issue. 

So I tought if it made sense to make something more generic that work in all cases without implementing it for each attribute.

### DIFF
--- a/Sources/SwiftParser/StringLiterals.swift
+++ b/Sources/SwiftParser/StringLiterals.swift
@@ -487,7 +487,7 @@ extension Parser {
       // string literal.
       guard currentToken.leadingTriviaText.isEmpty else { break }
 
-      if let stringSegment = self.consume(if: .stringSegment) {
+      if let stringSegment = self.consume(if: .stringSegment, TokenSpec(.identifier, remapping: .stringSegment)) {
         segments.append(.stringSegment(RawStringSegmentSyntax(content: stringSegment, arena: self.arena)))
       } else if let backslash = self.consume(if: .backslash) {
         let (unexpectedBeforeDelimiter, delimiter) = self.parsePoundDelimiter(.rawStringDelimiter, matching: openDelimiter)
--- a/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
+++ b/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
@@ -993,6 +993,35 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
     return handleMissingSyntax(node, additionalHandledNodes: [node.placeholder.id])
   }
 
+  override open func visit(_ node: OriginallyDefinedInArgumentsSyntax) -> SyntaxVisitorContinueKind {
+    if shouldSkip(node) {
+      return .skipChildren
+    }
+    if let token = node.unexpectedBetweenModuleLabelAndColon?.onlyToken(where: { $0.tokenKind.isIdentifier }),
+      node.moduleLabel.presence == .missing
+    {
+      addDiagnostic(
+        node,
+        MissingNodesError(missingNodes: [Syntax(node.moduleLabel)]),
+        fixIts: [
+          FixIt(
+            message: ReplaceTokensFixIt(
+              replaceTokens: [token],
+              replacements: [node.moduleLabel]
+            ),
+            changes: [
+              FixIt.MultiNodeChange.makeMissing(token),
+              FixIt.MultiNodeChange.makePresent(node.moduleLabel),
+            ]
+          )
+        ],
+        handledNodes: [node.moduleLabel.id, token.id]
+      )
+    }
+
+    return .visitChildren
+  }
+
   public override func visit(_ node: OperatorDeclSyntax) -> SyntaxVisitorContinueKind {
     if shouldSkip(node) {
       return .skipChildren
--- a/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
+++ b/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
@@ -1376,6 +1376,36 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
     return .visitChildren
   }
 
+  public override func visit(_ node: UnavailableFromAsyncArgumentsSyntax) -> SyntaxVisitorContinueKind {
+    if shouldSkip(node) {
+      return .skipChildren
+    }
+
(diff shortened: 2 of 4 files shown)


Sources/SwiftParser/StringLiterals.swift
```diff
@@ -487,7 +487,7 @@ extension Parser {
       // string literal.
       guard currentToken.leadingTriviaText.isEmpty else { break }
 
-      if let stringSegment = self.consume(if: .stringSegment) {
+      if let stringSegment = self.consume(if: .stringSegment, .identifier) {
```
ahoppen, line 490: This is not correct. `StringLiteralSegmentsSyntax` declares that the string segment must have a token kind `stringSegment` 

https://github.com/apple/swift-syntax/blob/d3c4f2461b75f05ad76a76d2c44da4e1a2350172/CodeGeneration/Sources/SyntaxSupport/ExprNodes.swift#L1399

This will fail if RawSyntax validation is enabled (which it is in CI).

https://github.com/apple/swift-syntax/blob/6bb19efac5ae49ec2de7d0b1d040b8e1e0a76de2/Package.swift#L27

I think it should be fine if you remapped the token to the `stringSegment` token kind.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -1285,6 +1314,36 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
+    if shouldSkip(node) {
+      return .skipChildren
+    }
+
+    if let token = node.unexpectedBetweenMessageLabelAndColon?.onlyToken(where: { $0.tokenKind.isIdentifier }),
+      node.messageLabel.presence == .missing
+    {
+      addDiagnostic(
+        node,
+        MissingNodesError(missingNodes: [Syntax(node.messageLabel)]),
```
ahoppen, line 1327: Why do we need this special diagnostic handling? Wouldn’t we get the same from the default handling?

kimdv: It did replace "" in front of the the identifier instead of wrapping the identifier in "".
So `message: bar` will become `message: ""bar` instead of `message: "bar"`

Sources/SwiftParser/StringLiterals.swift
kimdv: So I should consume and remap to `stringSegment `?

ahoppen: Yes, that’s what I’d do

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
ahoppen: That seems like we’re parsing it incorrectly. I would expect `bar` to be in the `segments` child of `StringLiteralExprSyntax` with the second quote after it. But it seems like `bar` got stuck somewhere after `closeQuote` if it produces `message: ""bar` as the fixed source.

kimdv: Ah oh! 
I was wrong here. 
This fixes that `@_unavailableFromAsync(otherLabel: "abc")` bot becomes `@_unavailableFromAsync(otherLabel message: "abc")`

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -993,6 +993,35 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
     return handleMissingSyntax(node, additionalHandledNodes: [node.placeholder.id])
   }
 
+  override open func visit(_ node: OriginallyDefinedInArgumentsSyntax) -> SyntaxVisitorContinueKind {
+    if shouldSkip(node) {
+      return .skipChildren
+    }
+    if let token = node.unexpectedBetweenModuleLabelAndColon?.onlyToken(where: { $0.tokenKind.isIdentifier }),
+      node.moduleLabel.presence == .missing
```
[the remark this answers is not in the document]
kimdv: Hmm.
How should we do this.

I was thinking a lot about it, as this was not the only case.

We could of course start with a private helper method, so there is less duplicates addDiagnostic code.
But that does not handle it in all upcoming cases.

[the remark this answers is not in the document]
ahoppen: Can’t we just add a case here that detects if the invalid and missing tokens are both identifiers and if they are, add the new diagnostic complaining about the spelling error? Or am I missing something?

https://github.com/apple/swift-syntax/blob/a6486790b46a4469bfb2b09d7ff6463aaecc40e5/Sources/SwiftParserDiagnostics/MissingTokenError.swift#L31-L32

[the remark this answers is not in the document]
kimdv: Just did some digging in this case. 
First the token that is missing is a `.keyword(.module)`. We should add that case and future ones we find, that is still handles one place, so for me it's a fine trade-off. 

Secons, when parsing `@_originallyDefinedIn(modulename: "foo", OSX 13.13)` the `moduleName` is consumed by `self.expect(.colon)` and not `self.expect(.keyword(.module))`.

https://github.com/apple/swift-syntax/blob/a6486790b46a4469bfb2b09d7ff6463aaecc40e5/Sources/SwiftParser/Attributes.swift#L985-L987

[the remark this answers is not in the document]
ahoppen: Oh, I see. In that case we will need to do the special handling.

Sources/SwiftParser/StringLiterals.swift
```diff
@@ -487,7 +487,8 @@ extension Parser {
       // string literal.
       guard currentToken.leadingTriviaText.isEmpty else { break }
 
-      if let stringSegment = self.consume(if: .stringSegment) {
+      if self.at(.stringSegment, .identifier) {
+        let stringSegment = self.consumeAnyToken(remapping: .stringSegment)
```
ahoppen, line 491: I think this can just be


```suggestion
      if let stringSegment = self.consume(if: .stringSegment, TokenSpec(.identifier, remapping: .stringSegment)) {
```

ahoppen (approved):

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test

ahoppen:
@swift-ci Please test Windows

PR #1567, opened 2023-04-21
Update CodeGeneration

ahoppen:


Sources/SwiftSyntax/generated/Tokens.swift
```diff
@@ -14,713 +14,713 @@
...
-        .prefixOperator(text), 
-        leadingTrivia: leadingTrivia, 
-        trailingTrivia: trailingTrivia, 
-        presence: presence
-      
+      .prefixOperator(text),
+      leadingTrivia: leadingTrivia,
+      trailingTrivia: trailingTrivia,
+      presence: presence
+
```
kimdv, line 520: There are multiple places with with white space.
Also think that Doug mentioned it.

kimdv (approved):

Sources/SwiftSyntax/generated/Tokens.swift
bnbarham: @ahoppen this one seems pretty weird to me:
1. I thought we were stripping trailing whitespace
2. We shouldn't be adding a newline here. Seems like that was broken before and after though

kimdv:
@ahoppen how much white space did we remove? 😅😆

Sources/SwiftSyntax/generated/Tokens.swift
ahoppen: This is the formatter subclass of `BasicFormat` that lives in CodeGeneration messing things up.

PR #1568, opened 2023-04-21
Some more improvements to BasicFormat + misc changes

ahoppen:
This is basically a grab bag of changes that I want to do before my final PR to improve BasicFormat. Splitting it so it’s easier to review:

Contains a couple of bug fixes:
- Fix an access to freed memory when modifying token nodes 
  - When a token node is modified, we would not add its arena as a child to the new arena. This is unsafe because the newly created token might still have its text stored in the original arena.
- Fix a bug where allTokens would not include the last token if it is missing

Minor new features
- Add a walk method to SyntaxRewriter
  - Does the same as `visit` but also re-attaches the rewritten node into the parent tree (I will need this in the next PR)

Biggest change:
- Add placeholder children to all missing syntax nodes
  - It’s much easier to get correct formatting in the new BasicFormat if the `MissingExprSyntax` etc. node has a token inside of it. Because if it does, we know whether we should e.g. add a space after `try` in front of a `MissingExprNode`. This also required a few small changes to SwiftParserDiagnostics

### DIFF
--- a/CodeGeneration/Sources/SyntaxSupport/CommonNodes.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/CommonNodes.swift
@@ -144,6 +144,7 @@ public let COMMON_NODES: [Node] = [
   Node(
     name: "MissingDecl",
     nameForDiagnostics: "declaration",
+    description: "In case the source code is missing a declaration, this node stands in place of the missing declaration.",
     kind: "Decl",
     traits: [
       "Attributed"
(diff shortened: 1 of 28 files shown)


ahoppen:
@swift-ci Please test

kimdv (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@bnbarham @rintaro I’m merging this now because I’ve got two more PRs that depend on this and which I would like to get merged ASAP as well. I‘ll address review comments in a follow-up PR.

PR #1570, opened 2023-04-21
Relax version requirement for SwiftSyntax

ahoppen:
This will make SwiftSyntax more tolerant with regard to which swift-argument-parser version it needs, resulting in fewer version conflicts for users of SwiftSyntax.


kimdv (approved):

ahoppen:
@swift-ci Please test

PR #1573, opened 2023-04-23
[5.9]https://github.com/apple/swift-syntax/pull/1449

kimdv:
Cherry picks https://github.com/apple/swift-syntax/pull/1449


kimdv:
@swift-ci please test

Sources/SwiftParser/Types.swift
```diff
@@ -1064,7 +1064,37 @@ extension Parser {
...
     } else {
-      return self.parseType()
+      var result = self.parseType()
+
+      guard !result.hasError else {
+        return result
+      }
+
+      if self.at(.rightSquareBracket) {
+        let (unexpectedBeforeRSquareBracket, rightSquareBracket) = self.expect(.rightSquareBracket)
```
ahoppen, line 1074: Actually looking at this code again, I think we should only do this recovery if the `]` is not on a new line. And I also realized that we could write this as follows. If you could incorporate those changes in `main`, that would be great. 

```suggestion
      if let rightSquareBracket = self.consume(if: .rightSquareBracket) {
```

And since I just found another thing to comment on and this PR only provide value in a very niche use case, I’d prefer not to cherry-pick it to 5.9.

ahoppen (changes requested):

PR #1574, opened 2023-04-23
Add diagnostic for `canImport` expression

StevenWong12:
Fixes part of #1373

This PR should solve all the TODOs in ``testIfconfigExpr.swift``.


StevenWong12:
There are 4 TODOs left in ``testIfconfigExpr.swift`` related to the value of ``_version``, e.g. 

```swift
#if canImport(A, _version: 2.2.2.2.2) // TODO: Old parser expected error on line 1: trailing components of version '2.2.2.2' are ignored
#if canImport(A, _version: >=2.2) // TODO: Old parser expected error on line 1: cannot parse module version '>=2.2'
```

I think it's reasonable to leave these diagnostics to the type check stage, since 1) Leaving an ``ExprSyntax`` there totally makes sense; 2) We can make use of the version check tools in llvm-project (`VersionTuple::tryParse`) to help us generating diagnostics instead of parsing the version string on our own. What do you think? @ahoppen

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -13,6 +13,7 @@
 import SwiftDiagnostics
 import SwiftParser
 @_spi(RawSyntax) import SwiftSyntax
+import Foundation
```
kimdv, line 16: Is this needed?

Tests/SwiftParserTest/Assertions.swift
```diff
@@ -12,7 +12,7 @@
 
 import XCTest
 @_spi(RawSyntax) import SwiftSyntax
-@_spi(Testing) @_spi(RawSyntax) import SwiftParser
+@_spi(Testing)@_spi(RawSyntax) import SwiftParser
```
kimdv, line 15: Is this the formatter doing that?

Tests/SwiftParserTest/DeclarationTests.swift
```diff
@@ -11,7 +11,7 @@
 //===----------------------------------------------------------------------===//
 
 @_spi(RawSyntax) import SwiftSyntax
-@_spi(Testing) @_spi(RawSyntax) import SwiftParser
+@_spi(Testing)@_spi(RawSyntax) import SwiftParser
```
kimdv, line 14: Same as previous comment

Tests/SwiftParserTest/translated/IfconfigExprTests.swift
```diff
@@ -312,12 +312,12 @@ final class IfconfigExprTests: XCTestCase {
...
     assertParse(
       """
-      #if canImport(A, unknown: 2.2)
+      #if canImport(A, 1️⃣unknown: 2.2)
         let a = 1
       #endif
       """,
       diagnostics: [
-        // TODO: Old parser expected error on line 1: 2nd parameter of canImport should be labeled as _version or _underlyingVersion
+        DiagnosticSpec(message: "2nd parameter of canImport should be labeled as _version or _underlyingVersion")
```
kimdv, line 320: Is there fix it's for those labels?

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
StevenWong12: Already removed😄

Tests/SwiftParserTest/DeclarationTests.swift
StevenWong12: I think these changes are made by the formatter. Already discard them.

Tests/SwiftParserTest/translated/IfconfigExprTests.swift
StevenWong12: I think we cannot infer whether the user miss a `_version` or `_underlyingVersion` here, so I prefer not to add a fixit here.

Sources/SwiftParser/Expressions.swift
```diff
@@ -2544,6 +2561,67 @@ extension Parser {
...
+  mutating func parseCanImportExpression(_ canImportIdentifierExpr: RawIdentifierExprSyntax) -> RawExprSyntax {
+    let (unexpectedBeforeLeftParen, leftParen) = self.expect(.leftParen)
+
+    let (unexpectedBeforeImportPath, importPath) = self.expect(.identifier)
+
+    let comma = self.consume(if: .comma)
+    var versionTupleExpr: RawTupleExprElementSyntax?
+
+    if comma != nil {
+      versionTupleExpr = self.parseTupleExprElement(pattern: .none)
```
ahoppen, line 2575: Instead of parsing a `TupleExprSyntax` and then taking it apart again, I think it would be easier to just write the parsing code for `canImport` arguments themselves.

Sources/SwiftParser/Expressions.swift
```diff
@@ -762,6 +762,14 @@ extension Parser {
         continue
       }
 
+      // canImport expression
+      if let leadingIdentifier = leadingExpr.as(RawIdentifierExprSyntax.self),
+        leadingIdentifier.identifier.tokenText == "canImport"
+      {
```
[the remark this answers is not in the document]
StevenWong12: Instead of adding an `ExprFlavor`, I add a `forDirective` for `parsePrimaryExpression ` to branch into expression parsing there.

Sources/SwiftParser/Expressions.swift
StevenWong12: Already refactored this part.😄

Sources/SwiftParser/Expressions.swift
```diff
@@ -1291,6 +1298,29 @@ extension Parser {
...
+      (.falseKeyword, let handle)?:
+      let tok = self.eat(handle)
+      return RawExprSyntax(
+        RawBooleanLiteralExprSyntax(
+          booleanLiteral: tok,
+          arena: self.arena
+        )
+      )
+    case (.identifier, _)?:
+      return RawExprSyntax(self.parseIdentifierExpression())
```
ahoppen, line 1318: Do we even need the `true`, `false` and identifier case in this switch? IIUC we are falling back to normal primary expression parsing if this method returns `nil`. And if those switch cases are removed, we might also be able to remove `parsePrimaryExprForDirective` altogether and call straight into `parseCanImportExpression` from `parsePrimaryExpression` if we’re parsing for a directive and are at `canImport`.

Also, we might not need the `CompilationCondition` `TokenSpecSet` at all.

Sources/SwiftParser/Expressions.swift
```diff
@@ -2544,6 +2574,52 @@ extension Parser {
   }
 }
 
+// MARK: Platform Condition
+extension Parser {
+  mutating func parseCanImportExpression(_ canImportKeyword: RawTokenSyntax) -> RawExprSyntax {
+
```
ahoppen, line 2580: Superfluous newline

Sources/SwiftParser/Expressions.swift
```diff
@@ -2544,6 +2574,52 @@ extension Parser {
   }
 }
 
+// MARK: Platform Condition
+extension Parser {
+  mutating func parseCanImportExpression(_ canImportKeyword: RawTokenSyntax) -> RawExprSyntax {
```
ahoppen, line 2579: The general paradigm in the parser is to pass the `TokenConsumptionHandle` for the `canImport` keyword instead of the keyword itself. Doesn’t really change much implementation-wise though.

Tests/SwiftParserTest/translated/IfconfigExprTests.swift
ahoppen, line 1: Could you add `fixedSource` parameters to these `assertParse` calls to make sure that applying the Fix-Its produces the correct result?

ahoppen (commented):
Thank you, that’s much cleaner. Also I didn’t realize we already had `forDirective` everywhere. Using that was a good idea!

Sources/SwiftParser/Expressions.swift
StevenWong12: I added these since I referred to the [the doc](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/statements#Conditional-Compilation-Block) when I investigated the conditional compiler syntax. 

I think if we want to modify some logic for these ifconfig-specific expressions later, we can change in `parsePrimaryExprForDirective` without influence the general expression parsing. Does it make sense?

ahoppen: I have just been thinking and at least the `swift` and `compiler` compilation conditions should also be represented by a dedicated node because they accept a version which cannot be represented properly inside a function call (would you like to look into that as well?). So, considering that this switch will soon have two cases, I think it makes sense to have it. Unless we make it exhaustive and not fall through to general primary expression parsing, I don’t think there’s value in having cases for boolean literal and identifier parsing in here.

StevenWong12: Oh, I see. That makes sense to me! I'd like to make `swift` and `compiler` dedicated nodes in a follow-up pr and remove the boolean literal and identifier parsing in `parsePrimaryExprForDirective `. 

Btw, I wonder if it is reasonable to delay the version parsing to type check stage? If yes, I also would like to do it.😁

ahoppen: Which version diagnostic do you mean exactly?

StevenWong12: In `IfconfigExprTests.swift` there are 4 TODOs I didn't touched such as
```swift
func testIfconfigExpr28() {
    assertParse(
      """
      #if canImport(A, _version: >=2.2)
        let a = 1
      #endif
      """,
      diagnostics: [
        // TODO: Old parser expected error on line 1: cannot parse module version '>=2.2'
      ]
    )
  }
  
  func testIfconfigExpr30() {
    assertParse(
      #"""
      #if canImport(A, _version: "20A301")
        let a = 1
      #endif
      """#,
      diagnostics: [
        // TODO: Old parser expected error on line 1: cannot parse module version '20A301'
      ]
    )
  }
```

StevenWong12: I just look through the type-checking codes and I think I got some misunderstandings about it before...So, just ignore my thoughts above.

Instead, I refactored the version tuple parsing in ``Availability.swift`` to get the version parsed and diagnosed in parsing stage.

Tests/SwiftParserTest/translated/IfconfigExprTests.swift
```diff
@@ -330,33 +330,35 @@ final class IfconfigExprTests: XCTestCase {
...
-      #if canImport(A, 2.2, 1.1)
+      #if canImport(A, 1️⃣2.22️⃣, 1.1)
         let a = 1
       #endif
       """,
       diagnostics: [
-        // TODO: Old parser expected error on line 1: canImport can take only two parameters
+        DiagnosticSpec(locationMarker: "1️⃣", message: "2nd parameter of canImport should be labeled as _version or _underlyingVersion"),
+        // TODO: Old parse produce "canImport can take only two parameters" here
+        DiagnosticSpec(locationMarker: "2️⃣", message: "trailing components of version code '2.2' are ignored", severity: .warning),
```
StevenWong12, line 361: In my implementation, I couldn't find a clean workaround to determine whether the trailing components are extra parameters here. Do you have any idea? Or could we just use the diagnostic message here?

CodeGeneration/Sources/SyntaxSupport/AvailabilityNodes.swift
```diff
@@ -118,42 +118,54 @@ public let AVAILABILITY_NODES: [Node] = [
     ]
   ),
 
-  // version-tuple -> integer-literal
-  //                | float-literal
-  //                | float-literal '.' integer-literal
+  // version-tuple-element -> '.' float-literal |
+  //                          '.' interger-literal
```
ahoppen, line 122: This is completely out of date since we no longer parse float literals into a version tuple. I think you can just remove it.

CodeGeneration/Sources/SyntaxSupport/AvailabilityNodes.swift
```diff
@@ -118,42 +118,54 @@ public let AVAILABILITY_NODES: [Node] = [
...
+        name: "Major",
+        kind: .token(choices: [.token(tokenKind: "IntegerLiteralToken")]),
+        description: "The major version."
       ),
       Child(
-        name: "Patch",
-        kind: .token(choices: [.token(tokenKind: "IntegerLiteralToken")]),
-        description: "The patch version if specified.",
+        name: "SubVersions",
+        kind: .collection(kind: "VersionList", collectionElementName: "VersionTuppleElement"),
```
ahoppen, line 167:
```suggestion
        kind: .collection(kind: "VersionList", collectionElementName: "VersionTupleElement"),
```

CodeGeneration/Sources/SyntaxSupport/AvailabilityNodes.swift
```diff
@@ -118,42 +118,54 @@ public let AVAILABILITY_NODES: [Node] = [
...
-  //                | float-literal
-  //                | float-literal '.' integer-literal
+  // version-tuple-element -> '.' float-literal |
+  //                          '.' interger-literal
+
   Node(
-    name: "VersionTuple",
-    nameForDiagnostics: "version tuple",
-    description: "A version number of the form major.minor.patch in which the minor and patch part may be omitted.",
+    name: "VersionTupleElement",
```
ahoppen, line 125: What do you think about naming this `VersionComponent` and the list `VersionComponentList`? And then rename `subversions` in your implementation to `components`. Subversion still always reminds me of the version control system.

Otherwise, I think this is a great improvement.

Sources/SwiftParser/Availability.swift
```diff
@@ -261,44 +261,30 @@ extension Parser {
   ///     platform-version → decimal-digits
   ///     platform-version → decimal-digits '.' decimal-digits
   ///     platform-version → decimal-digits '.' decimal-digits '.' decimal-digits
-  mutating func parseVersionTuple() -> RawVersionTupleSyntax {
+  mutating func parseVersionTuple(maxSubverNum: Int = 2) -> RawVersionTupleSyntax {
```
ahoppen, line 264: I would call this maxComponentCount, which should IMO also include the major version component. I.e. `1.2.3` has a component count of 3 to me.

CodeGeneration/Sources/SyntaxSupport/AvailabilityNodes.swift
```diff
@@ -118,42 +118,54 @@ public let AVAILABILITY_NODES: [Node] = [
...
     kind: "Syntax",
     children: [
       Child(
-        name: "Major",
-        kind: .token(choices: [.token(tokenKind: "IntegerLiteralToken")]),
-        description: "The major version."
-      ),
-      Child(
-        name: "MinorPeriod",
+        name: "period",
```
ahoppen, line 131: These names should be capitalized. Otherwise we get a lowercase name in the unexpected names. 

And since you just hit this. Would you like to also add an assertion that the first letter of a child’s name is uppercase?

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -514,6 +514,44 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
+      if versionTuple.major.presence == .missing {
+        addDiagnostic(
+          versionTuple,
+          CannotParseVersionTuple(versionTuple: unexpectedVersionTuple),
+          handledNodes: [versionTuple.id, unexpectedVersionTuple.id]
+        )
+      } else {
+        addDiagnostic(
+          unexpectedVersionTuple,
+          TrailingVersionAreIgnored(version: versionTuple),
```
ahoppen, line 546: Should we just consume any trailing components of a version in `parseVersionTuple` and have a generalized diagnostic that applies to all version tuples everywhere?

Tests/SwiftParserTest/AvailabilityTests.swift
```diff
@@ -121,8 +121,7 @@ final class AvailabilityTests: XCTestCase {
       substructure: Syntax(
         VersionTupleSyntax(
           major: .integerLiteral("10"),
-          minorPeriod: .periodToken(),
-          minor: .integerLiteral("0")
+          subVersions: VersionListSyntax([VersionTupleElementSyntax(period: .periodToken(), number: .integerLiteral("0"))])
```
ahoppen, line 124: Do you even need to specify `periodToken` here? Shouldn’t it be defaulted because it’s non-optional? Same for the other test cases.

Tests/SwiftParserTest/translated/IfconfigExprTests.swift
```diff
@@ -369,59 +371,64 @@ final class IfconfigExprTests: XCTestCase {
...
     assertParse(
       #"""
-      #if canImport(A, _version: "")
+      #if canImport(A, _version: 1️⃣"")
         let a = 1
       #endif
       """#,
       diagnostics: [
-        // TODO: Old parser expected error on line 1: _version argument cannot be empty
+        DiagnosticSpec(message: "cannot parse version code '\"\"'")
```
ahoppen, line 392: Where is the `code` coming from here?

Tests/SwiftParserTest/translated/IfconfigExprTests.swift
```diff
@@ -455,6 +462,19 @@ final class IfconfigExprTests: XCTestCase {
...
+    assertParse(
+      """
+        arch()
+        canImport(a, b, c)
+        os(bogus)
+        targetEnvironment(foo, bar)
+        compiler(a)
+        swift(foo)
+      """
+    )
```
ahoppen, line 475: Could you create separate `asserParse` calls for all of these? If they fail, I find it’s a lot easier to tell what’s going wrong if the code inside `assertParse` is as minimal as possible.

Tests/SwiftParserTest/translated/IfconfigExprTests.swift
StevenWong12: I think the `code` here is because I used `shortSingleLineContentDescription` for the node, already removed it.

CodeGeneration/Sources/SyntaxSupport/AvailabilityNodes.swift
StevenWong12: I think these names are better. Already renamed these nodes.

Sources/SwiftParser/Availability.swift
```diff
@@ -261,44 +261,46 @@ extension Parser {
...
+
+      var components: [RawVersionComponentSyntax] = []
+      for _ in 1..<maxComponentCount {
+        guard let period = self.consume(if: .period) else {
+          break
+        }
+        let version = self.expectDecimalIntegerWithoutRecovery()
+
+        components.append(RawVersionComponentSyntax(period: period, number: version, arena: self.arena))
       }
```
ahoppen, line 281: What do you think about unifying parsing of `components` and `trailingComponents` Something like the following

```suggestion
      var components: [RawVersionComponentSyntax] = []
      for i in 1… {
        guard let period = self.consume(if: .period) else {
          break
        }
        let version = self.expectDecimalIntegerWithoutRecovery()

        let versionComponent = RawVersionComponentSyntax(period: period, number: version, arena: self.arena)
        if i < maxComponentCount {
          components.append(versionComponent)
        } else {
          trailingComponents.append(versionComponent)
        }
      }
```

CodeGeneration/Sources/SyntaxSupport/AvailabilityNodes.swift
```diff
@@ -118,42 +118,51 @@ public let AVAILABILITY_NODES: [Node] = [
...
-        isOptional: true
+        description: "The number of this pair"
       ),
+    ]
+  ),
+
+  // version-list -> version-tuple-element version-list?
+  Node(
+    name: "VersionComponentList",
+    nameForDiagnostics: "version list",
```
ahoppen, line 144: “version list” is a big misleading here IMO because it’s not a list of versions, it’s just the components of a single version. I would just not provide a `nameForDiagnositcs` and then the name for diagnostics will default to the parent, which is `version tuple`

CodeGeneration/Sources/SyntaxSupport/AvailabilityNodes.swift
```diff
@@ -118,42 +118,51 @@ public let AVAILABILITY_NODES: [Node] = [
...
-        kind: .token(choices: [.token(tokenKind: "IntegerLiteralToken")]),
-        description: "The major version."
-      ),
-      Child(
-        name: "MinorPeriod",
+        name: "Period",
         kind: .token(choices: [.token(tokenKind: "PeriodToken")]),
-        description: "If the version contains a minor number, the period separating the major from the minor number.",
-        isOptional: true
+        description: "The period of this pair"
```
ahoppen, line 131:
```suggestion
        description: "The period of this version component"
```

CodeGeneration/Sources/SyntaxSupport/AvailabilityNodes.swift
```diff
@@ -118,42 +118,51 @@ public let AVAILABILITY_NODES: [Node] = [
...
       ),
       Child(
-        name: "Minor",
+        name: "Number",
         kind: .token(choices: [.token(tokenKind: "IntegerLiteralToken")]),
-        description: "The minor version if specified.",
-        isOptional: true
+        description: "The number of this pair"
```
ahoppen, line 136:
```suggestion
        description: "The version number of this component"
```

CodeGeneration/Sources/SyntaxSupport/AvailabilityNodes.swift
```diff
@@ -118,42 +118,51 @@ public let AVAILABILITY_NODES: [Node] = [
...
-  //                | float-literal
-  //                | float-literal '.' integer-literal
+  // version-tuple-element -> '.' interger-literal
   Node(
-    name: "VersionTuple",
-    nameForDiagnostics: "version tuple",
-    description: "A version number of the form major.minor.patch in which the minor and patch part may be omitted.",
+    name: "VersionComponent",
+    nameForDiagnostics: "version tuple element",
+    description: "An element to represent a dot and number pair",
```
ahoppen, line 125:
```suggestion
    description: "An element to represent a single component in a version, like `.1`.",
```

Sources/SwiftParser/Availability.swift
```diff
@@ -261,44 +261,46 @@ extension Parser {
   ///     platform-version → decimal-digits
   ///     platform-version → decimal-digits '.' decimal-digits
   ///     platform-version → decimal-digits '.' decimal-digits '.' decimal-digits
-  mutating func parseVersionTuple() -> RawVersionTupleSyntax {
+  mutating func parseVersionTuple(maxComponentCount: Int = 3) -> RawVersionTupleSyntax {
```
ahoppen, line 264: I just changed my mind and would prefer not to default this to any value. I think every call site of this function should be forced to explicitly think about how many components it wants to support, just so it’s a conscious decision.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -531,6 +531,44 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
+    if let label = node.label,
+      label.presence == .missing
+    {
+      addDiagnostic(
+        label,
+        .canImportWrongSecondParameterLabel,
+        handledNodes: [label.id]
+      )
+
+      [node.unexpectedBetweenLabelAndColon?.id, node.colon?.id, node.versionTuple?.id].compactMap { $0 }.forEach { handledNodes.append($0) }
```
ahoppen, line 548:
```suggestion
      handledNodes.append(contentsOf: [node.unexpectedBetweenLabelAndColon?.id, node.colon?.id, node.versionTuple?.id].compactMap { $0 })
```

reads nice IMO because it shows the action that is being performed

CodeGeneration/Sources/SyntaxSupport/AvailabilityNodes.swift
```diff
@@ -118,42 +118,51 @@ public let AVAILABILITY_NODES: [Node] = [
...
-  // version-tuple -> integer-literal
-  //                | float-literal
-  //                | float-literal '.' integer-literal
+  // version-tuple-element -> '.' interger-literal
   Node(
-    name: "VersionTuple",
-    nameForDiagnostics: "version tuple",
-    description: "A version number of the form major.minor.patch in which the minor and patch part may be omitted.",
+    name: "VersionComponent",
+    nameForDiagnostics: "version tuple element",
```
ahoppen, line 124: I would not specify a `nameForDiagnostics` for this one. See my comment a bit further down for an explanation.

Sources/SwiftParser/Availability.swift
StevenWong12: I think that looks cleaner. I will make a change on it.

CodeGeneration/Sources/SyntaxSupport/AvailabilityNodes.swift
```diff
@@ -118,42 +118,50 @@ public let AVAILABILITY_NODES: [Node] = [
...
+    kind: "SyntaxCollection",
+    element: "VersionComponent",
+    omitWhenEmpty: true
+  ),
+
+  // version-tuple -> integer-literal version-list?
+  Node(
+    name: "VersionTuple",
+    nameForDiagnostics: "version tuple",
+    description: "A version number of the form major.minor.patch in which the minor and patch part may be omitted.",
```
ahoppen, line 154: Sorry for not catching this earlier. Only realized that this description is no longer up-to-date. 

Suggestion

```suggestion
    description: "A version number like `1.2.0`. Only the first version component is required. There might be an arbitrary number of following components.",
```

CodeGeneration/Sources/SyntaxSupport/AvailabilityNodes.swift
```diff
@@ -118,42 +118,50 @@ public let AVAILABILITY_NODES: [Node] = [
...
+        name: "Major",
+        kind: .token(choices: [.token(tokenKind: "IntegerLiteralToken")]),
+        description: "The major version."
       ),
       Child(
-        name: "Patch",
-        kind: .token(choices: [.token(tokenKind: "IntegerLiteralToken")]),
-        description: "The patch version if specified.",
+        name: "Components",
+        kind: .collection(kind: "VersionComponentList", collectionElementName: "VersionComponent"),
```
ahoppen, line 164: Again, sorry for not catching it earlier. I’m trying to generally improve documentation of syntax nodes, it would be nice to add one here. I’d suggest

```swift
description: "Any version components that are not the major version . For example, for `1.2.0`, this will contain `.2.0`"
```

CodeGeneration/Sources/SyntaxSupport/ExprNodes.swift
```diff
@@ -185,6 +185,51 @@ public let EXPR_NODES: [Node] = [
     ]
   ),
 
+  // the canImport expr in if config expression
+  Node(
+    name: "CanImportExpr",
+    nameForDiagnostics: "'canImport' expression in if config expression",
```
ahoppen, line 191: I’d just change this to the following. It’s a bit more concise and doesn’t make as many assumptions about where `CanImportExpr` could also be used.

```suggestion
    nameForDiagnostics: "'canImport' expression",
```

ahoppen (commented):
Very nice. I’ve got a few minor comments regarding documentation, otherwise this looks good to me. 👍🏽

StevenWong12:
I've updated the documentation according to your suggestion.😀

kimdv:
@swift-ci please test

ahoppen:
The test failure is unrelated to your changes @StevenWong12. The [status bot](https://ci.swift.org/job/oss-swift-pr-test-macoss/) is also failing with the same error. I’ll re-trigger CI once the status bot is passing.

ahoppen:
@swift-ci Please test

CodeGeneration/Sources/SyntaxSupport/ExprNodes.swift
```diff
@@ -185,6 +185,51 @@ public let EXPR_NODES: [Node] = [
...
 
+  // the canImport expr in if config expression
+  Node(
+    name: "CanImportExpr",
+    nameForDiagnostics: "'canImport' expression",
+    kind: "Expr",
+    children: [
+      Child(
+        name: "CanImportKeyword",
+        kind: .token(choices: [.token(tokenKind: "IdentifierToken")])
```
ahoppen, line 196: CI is failing in its extended syntax tree validation because this is a `canImport` keyword instead of an identifier. To reproduce the failure locally, you need to enable raw syntax validation. The easiest way to do that in Xcode is to edit `Package.swift` as follows

```diff
- if ProcessInfo.processInfo.environment["SWIFTSYNTAX_ENABLE_RAWSYNTAX_VALIDATION"] != nil {
+ if ProcessInfo.processInfo.environment["SWIFTSYNTAX_ENABLE_RAWSYNTAX_VALIDATION"] != nil || true {
```

The fix should be to change this line to 
```suggestion
        kind: .token(choices: [.keyword(text: "canImport")])
```

StevenWong12: Oh, I see. I didn't notice the validation is now controlled by a standalone flag rather than `DEBUG`.

ahoppen: Yeah, we do that so if you build swift-syntax in debug for macros, you don’t get that extended validation, which could crash your macro. We really just want to do it in CI to catch issues in the syntax tree layout.

ahoppen (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

ahoppen:
Looks like the PR conflicted with another one. Could rebase and resolve the conflicts?

StevenWong12:
I've got the conflicts resolved.

kimdv:
@swift-ci please test

CodeGeneration/Sources/SyntaxSupport/AvailabilityNodes.swift
```diff
@@ -118,42 +118,51 @@ public let AVAILABILITY_NODES: [Node] = [
     ]
   ),
 
-  // version-tuple -> integer-literal
-  //                | float-literal
-  //                | float-literal '.' integer-literal
+  // version-tuple-element -> '.' interger-literal
```
CodaFi, line 121: Nit: `integer-literal`

CodeGeneration/Sources/SyntaxSupport/ExprNodes.swift
```diff
@@ -185,6 +185,51 @@ public let EXPR_NODES: [Node] = [
...
+      Child(
+        name: "ImportPath",
+        kind: .token(choices: [.token(tokenKind: "IdentifierToken")])
+      ),
+      Child(
+        name: "Comma",
+        kind: .token(choices: [.token(tokenKind: "CommaToken")]),
+        isOptional: true
+      ),
+      Child(
```
CodaFi, line 211: Rather than have a string of optional components, can we split the version information into its own syntax element and make that optional?

StevenWong12: That makes sense to me. I've extracted the optional part of `canImport` to a separate node.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

PR #1579, opened 2023-04-24
Remove all token children that allowed arbitrary keywords

ahoppen:
Allowing arbitrary keywords for a token was a bit of an anti-pattern. We should always explicitly specify which keywords we support. With token kind validation, this has pointed us to parser bugs before.

rdar://107156958


ahoppen:
@swift-ci Please test

bnbarham (approved):

kimdv (approved):

ahoppen:
@swift-ci Please test

PR #1580, opened 2023-04-25
Fix wrong diagnostic parse typed pattern

kimdv:
While digging in #1484 I found this.

### DIFF
--- a/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
+++ b/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
@@ -331,7 +331,7 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
     } else if node.first?.as(TokenSyntax.self)?.tokenKind.isIdentifier == true,
       let previousToken = node.previousToken(viewMode: .sourceAccurate),
       previousToken.tokenKind.isIdentifier,
-      previousToken.parent?.is(DeclSyntax.self) == true
+      previousToken.parent?.is(DeclSyntax.self) == true || previousToken.parent?.is(IdentifierPatternSyntax.self) == true
     {
       // If multiple identifiers are used for a declaration name, offer to join them together.
       let tokens =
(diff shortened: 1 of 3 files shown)


Tests/SwiftParserTest/translated/RecoveryTests.swift
```diff
@@ -869,6 +876,23 @@ final class RecoveryTests: XCTestCase {
...
+      private var a 1️⃣b : Int = ""
+      """,
+      diagnostics: [
+        DiagnosticSpec(
+            locationMarker: "1️⃣",
+            message: "unexpected code 'b' before type annotation"
+        )
+      ],
+      fixedSource: """
+        private var a b : Int = ""
```
kimdv, line 891: Should we in this case merge `a` and `b`

ahoppen: I think that would make sense.

kimdv: Fixed and rebased in main

kimdv:
@swift-ci please test

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -1414,6 +1414,36 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
+      return .skipChildren
+    }
+
+    if let identifierPattern = node.pattern.as(IdentifierPatternSyntax.self),
+      let unexpectedBeforeColon = node.typeAnnotation?.unexpectedBeforeColon,
+      let token = unexpectedBeforeColon.onlyToken(where: { $0.tokenKind.isIdentifier })
+    {
+      addDiagnostic(
+        unexpectedBeforeColon,
+        SpaceSeparatedIdentifiersError(
```
ahoppen, line 1428: Do you know why we need to do special handling here and why this is not caught by the general logic in `visit(UnexpectedNodesSyntax)`?

kimdv: The reason was that `previousToken.parent?.is(DeclSyntax.self) == true`. Parent here is an `IdentifierPatternSyntax`. 
I added the pattern as I think it's okay to merge an identifier with the previous identifier pattern.

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test windows

ahoppen (approved):
Thanks. That’s much cleaner now

PR #1585, opened 2023-04-25
[5.9] Add diagnostic for wrong inheritance

kimdv:
* **Explanation**: When there was a Python style inheritance the parser did show an "unexpected code" error. This will show a correct error and fix-it 
* **Scope**: Swift parser and diagnostics
* **Risk**: Low
* **Testing**: Added test cases that converts Python style inheritance to Swift inheritance 
* **Issue**: N/A
* **Reviewer**: @ahoppen on https://github.com/apple/swift-syntax/pull/1450

### DIFF
--- a/Sources/SwiftParser/Nominals.swift
+++ b/Sources/SwiftParser/Nominals.swift
@@ -240,7 +240,7 @@ extension Parser {
     }
 
     let inheritance: RawTypeInheritanceClauseSyntax?
-    if self.at(.colon) {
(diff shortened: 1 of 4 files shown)


ahoppen:
Looking at this PR again, I’d prefer not to cherry-pick it to 5.9. The change seems potentially risky (because `isAtPythonStyleInheritanceClause` is not super trivial) and the use case is very marginal.

PR #1586, opened 2023-04-25
[5.9] Add a public function that can be used to test macro expansions

ahoppen:
Cherry-pick https://github.com/apple/swift-syntax/pull/1550 to release/5.9

---

Testing macros should be as easy as possible and we should thus make the  `assertMacroExpansion` function public.

While making the function public, I made the following changes:
- Changed `diagnosticMessages` to expect `DiagnsoticSpec` similar to the parser tests. The only difference is that they take line and columns instead of location markers because the location markers have a slighgly high learning curve IMO.

Slightly unrelated, I also made the following changes:
- Made a few minor improvements to `SwiftParserTest/Assertions.swift`
- Fixed a bug in MacroApplication where the wrong syntax node was passed as the `node` parameter` to `addDiagnostics`, leading to a highlight that was bigger than expected.


DougGregor (approved):

ahoppen:
@swift-ci Please test

DougGregor (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #1588, opened 2023-04-26
redo 'without' operator parsing so it's within parseType to match C++ parser

kavon:
I decided to change the parsing on the C++ side to instead look for suppression in `parseType` as it yielded a better implementation. As such, "suppressed" inherited entries are now gone and you can now have general suppressed types being parsed.

### DIFF
--- a/CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
@@ -1028,14 +1028,6 @@ public let DECL_NODES: [Node] = [
       "WithTrailingComma"
     ],
     children: [
-      /// Indicates whether the 'without' operator was applied to the type to
-      /// indicate the suppression of implicit conformance to this type.
-      /// This child stores the token representing the 'without' operator.
-      Child(
-        name: "WithoutTilde",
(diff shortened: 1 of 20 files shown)


kavon:
@swift-ci please test

kavon:
@swift-ci please test

ahoppen (commented):
I am not following the evolution proposal but do we really want to support `~` in any place that supports types. As it stands right now, once we integrate the new parser into the compiler, ASTGen will need to diagnose that `~` cannot be used for every syntactic position that supports types.

kavon:
Eventually it will need to be supported in `where` clauses and in parsing generic parameters like `<T: ~Copyable>`. At the moment TypeCheckType.cpp flags those appearances as illegal, so I think ASTGen can safely just map it over to a `SuppressedTyprRepr` as in https://github.com/apple/swift/pull/64699

kavon:
So you can think of this as an improvement in terms of diagnostics, since now we give a good diagnostic complaining about uses of `~Type` in other places, as people are bound to try it elsewhere. Without this change they'd get really bizarre parsing errors. Those diagnostics happen during type checking, though, not parsing.

kavon:
@swift-ci please test windows

Issue #1528, opened 2023-04-13
Add documentation for all syntax nodes and their children


ahoppen:
SwiftSyntax’s nodes are currently mostly lacking any documentation. It would be a great community effort to add a description to to every node and every child. Writing documentation for some nodes and children will be easier than for other. The expectation is that if there are any non-obvious things to point out, the documentation should do so. I am happy to help if the semantics of a node or a child are not clear.

To make sure we don’t have multiple PRs adding documentation for the same node please add a comment to this issue with the node(s) you are working on. If you have commit access to this repository, you can also assign a node modify the issue and assign a node to yourself by marking it as ticked and adding your name behind it.

https://github.com/apple/swift-syntax/pull/1527 contains an example of how the documentation of a syntax node could look like.

- [ ] `AccessorBlockSyntax`
- [ ] `AccessorDeclSyntax`
- [ ] `AccessorEffectSpecifiersSyntax`
- [ ] `AccessorParameterSyntax`
- [ ] `AccessPathComponentSyntax`
- [ ] `ActorDeclSyntax`
- [ ] `ArrayElementSyntax`
- [ ] `ArrayExprSyntax`
- [ ] `ArrayTypeSyntax`
- [ ] `ArrowExprSyntax`
- [ ] `AsExprSyntax`
- [ ] `AssignmentExprSyntax`
- [ ] `AssociatedtypeDeclSyntax`
- [ ] `AttributedTypeSyntax`
- [ ] `AttributeSyntax`
- [ ] `AvailabilityArgumentSyntax`
- [ ] `AvailabilityConditionSyntax`
- [ ] `AvailabilityEntrySyntax`
- [ ] `AvailabilityLabeledArgumentSyntax`
- [ ] `AvailabilityVersionRestrictionListEntrySyntax`
- [ ] `AvailabilityVersionRestrictionSyntax`
- [ ] `AwaitExprSyntax`
- [ ] `BackDeployedAttributeSpecListSyntax`
- [ ] `BinaryOperatorExprSyntax`
- [ ] `BooleanLiteralExprSyntax`
- [ ] `BorrowExprSyntax`
- [ ] `BreakStmtSyntax`
- [ ] `CaseItemSyntax`
- [ ] `CatchClauseSyntax`
- [ ] `CatchItemSyntax`
- [ ] `ClassDeclSyntax`
- [ ] `ClassRestrictionTypeSyntax`
- [ ] `ClosureCaptureItemSpecifierSyntax`
[...]

PR #1589, opened 2023-04-26
SyntaxSupport Node Documentation: StructDecl

andrewjl:
#1528

### DIFF
--- a/CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
@@ -1805,6 +1805,63 @@ public let DECL_NODES: [Node] = [
   Node(
     name: "StructDecl",
     nameForDiagnostics: "struct",
+    description: """
+          A struct declaration like the following.
+
+          ```swift
+          struct SomeStruct {
+            let someMember: String
+            var anotherMember: Int
+
+            func foo() {
+              print(someMember)
+            }
+
+            mutating func bar() {
+              anotherMember = 42
+            }
+          }
+          ```
+
+          A struct declaration may be declared without any members.
+
+          ```swift
+          struct EmptyStruct {
+
+          }
+          ```
+
+          A struct declaration may include a type inheritance clause listing
+          one or more protocols the struct conforms to.
+
+          The example below uses Hashable and Equatable protocols whose members
+          are automatically synthesized by the compiler if the struct contains
+          stored members that are themselves `Hashable` and `Equatable`.
+
+          ```swift
+          struct AdvancedStruct: Hashable, Equatable {
+            let someMember: String
+            var anotherMember: Int
+          }
+          ```
+
+          A struct declaration may include a generic parameter clause as well
+          as a generic where clause.
+
+          ```swift
+          struct Stack<Element> {
+            var items: [Element] = []
+
+            mutating func push(_ item: Element) {
+              items.append(item)
+            }
+
+            mutating func pop() -> Element {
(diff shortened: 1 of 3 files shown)


CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
```diff
@@ -1824,43 +1836,51 @@ public let DECL_NODES: [Node] = [
...
         kind: .node(kind: "GenericParameterClause"),
         nameForDiagnostics: "generic parameter clause",
+        description: "The generic parameters of the struct declaration.",
         isOptional: true
       ),
       Child(
         name: "InheritanceClause",
         kind: .node(kind: "TypeInheritanceClause"),
         nameForDiagnostics: "type inheritance clause",
+        description: "The struct declaration inheritance clause. It may contain one or more type identifiers.",
```
andrewjl, line 1870: I took cues from[ this section](https://github.com/apple/swift-book/blob/main/TSPL.docc/ReferenceManual/Types.md#type-inheritance-clause) of TSPL here. 

Is it necessary to call out that a struct can only conform to one or more protocols and can't actually inherit from another type?

ahoppen: I don’t think we should put any restrictions on here saying that the inherited types can only be protocols because that can’t be ensured at parse time (the name resolution happens in the name lookup phase after type checking). As such, you could have a perfectly valid syntax tree where a struct inherits from multiple other structs. 

Mentioning protocols in more of an example-style to aid understanding would be a good option IMO.

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
```diff
@@ -1805,6 +1805,63 @@ public let DECL_NODES: [Node] = [
...
+            var items: [Element] = []
+
+            mutating func push(_ item: Element) {
+              items.append(item)
+            }
+
+            mutating func pop() -> Element {
+              return items.removeLast()
+            }
+         }
```
ahoppen, line 1862: That’s nice and verbose. I like it 🤩

ahoppen:
When you’re ready, you will need to re-generate files as described in https://swiftpackageindex.com/apple/swift-syntax/main/documentation/swiftsyntax/changingswiftsyntax. Let me know once you’ve done that and I’ll kick off CI for this PR.

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
andrewjl: Got it, thanks for clarifying! Added some extra illustrative examples instead of discussing restrictions.

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
andrewjl: Great!

andrewjl:
> When you’re ready, you will need to re-generate files as described in https://swiftpackageindex.com/apple/swift-syntax/main/documentation/swiftsyntax/changingswiftsyntax. Let me know once you’ve done that and I’ll kick off CI for this PR.

Done!

ahoppen (approved):
Great. Thanks!

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

kimdv:
@swift-ci please test

PR #1592, opened 2023-04-27
Add diagnostic for missing `0` in RawFloatLiteralExprSyntax

kimdv:


Sources/SwiftParser/Expressions.swift
```diff
@@ -1263,6 +1263,29 @@ extension Parser {
...
+      // Special case ".<integer_literal>" like ".4".  This isn't valid, but the
+      // developer almost certainly meant to use "0.4".  Diagnose this, and
+      // recover as if they wrote that.
+      if let integerLiteral = self.consume(if: .integerLiteral) {
+        return RawExprSyntax(
+          RawFloatLiteralExprSyntax(
+            floatingDigits: RawTokenSyntax(
+              missing: .floatingLiteral,
+              arena: self.arena
+            ),
```
ahoppen, line 1276: Can’t you create a floating literal token with the `0.4` text here already instead of having to construct it in `ParseDiagnosticGenerator`? `arena.intern` might be a useful method for that.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -678,6 +678,39 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
     return .visitChildren
   }
 
+  public override func visit(_ node: FloatLiteralExprSyntax) -> SyntaxVisitorContinueKind {
+    if shouldSkip(node) {
+      return .skipChildren
+    }
+    if node.floatingDigits.presence == .missing,
+      let tokens = node.unexpectedAfterFloatingDigits?.onlyTokens(satisfying: { $0.tokenKind == .period || $0.tokenKind.isIntegerLiteral })
```
ahoppen, line 686: I think you can use `twoTokens` instead of `onlyTokens` here to ensure that we really only match `period` followed by an integer literal.

Sources/SwiftParser/Expressions.swift
kimdv: I was thinking of that.
But not sure if we should do that inside the parser tough.

How should I add `"0"` and then append the dot token and the internet literal.

And another thing is that if we start to recover here how do we add a fit it and diagnostic ?

ahoppen: Can’t you just do something like the following? And the diagnostic just stays exactly the same, it just picks up the float literal contents from the missing token instead of constructing it itself.

```swift
let text = arena.intern("0" + String(syntaxText: tokenText) + String(syntaxText: integerLiteral.tokenText))
...
floatingDigits: missingToken(.floatingLiteral, text: text, arena: self.arena)
```

kimdv: This is not possible. I get an error `Cannot convert value of type 'String' to expected argument type 'SyntaxText'`

ahoppen: Ah, you need to expose the following as `@_spi(RawSyntax)`

 https://github.com/apple/swift-syntax/blob/7f1f40fe6e1dccef90010072f3ff9d60444a0ade/Sources/SwiftSyntax/SyntaxArena.swift#L82

kimdv: I've not fully understund the usage of `@_spi(RawSyntax)`
Can you please explain what it exactly does?

ahoppen: Oh, OK. So SPI stands for “System Programming Interface”. The idea is that you might want to expose a function to selected clients but not the general public. The idea is that you mark your function as `@_spi(RawSyntax)` public and then it will only be accessible to clients that `@_spi(RawSyntax) import SwiftSyntax`. 

The interface of the SPI will be in `.private.swiftinterface` files that can be stripped away when a library is distributed, basically making the SPI unavailable for third-party clients. 

https://github.com/apple/swift/blob/main/docs/ReferenceGuides/UnderscoredAttributes.md#_spispiname has a little bit more information.

Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift
```diff
@@ -306,6 +306,14 @@ public struct IdentifierNotAllowedInOperatorName: ParserError {
   }
 }
 
+public struct InvalidFloatLiteralMissingLeadingZero: ParserError {
+  public let integerLiteral: TokenSyntax
```
ahoppen, line 310: What do you think about naming this `decimalDigits`? `integerLiteral` is a bit confusing to me since this is an error about floating point literals.

ahoppen (approved):

PR #1595, opened 2023-04-28
Fix a precondition failure if a generic parameter clause had an additional closing right angle

ahoppen:
rdar://108624264

PR #1614, opened 2023-04-28
[5.9] Fix a precondition failure if a generic parameter clause had an additional closing right angle

ahoppen:
* **Explanation**: If a generic argument clause has an additional closing `>`, we performed a misparse, causing a precondition failure because we were consuming `>>` as a right angle, which doesn’t match the expected text.
* **Scope**: Parsing of generic argument argument clauses
* **Risk**: Very low. The old code was obviously wrong
* **Testing**: Added regression test
* **Issue**: rdar://108624264
* **Reviewer**: @bnbarham on https://github.com/apple/swift-syntax/pull/1595

### DIFF
--- a/Sources/SwiftParser/Declarations.swift
+++ b/Sources/SwiftParser/Declarations.swift
@@ -544,7 +544,7 @@ extension Parser {
 
     let rangle: RawTokenSyntax
     if self.currentToken.starts(with: ">") {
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

kimdv (approved):

PR #1616, opened 2023-04-28
Fix diagnostic for wrong `where` in if statement

kimdv:


Tests/SwiftParserTest/translated/RecoveryTests.swift
```diff
@@ -1911,20 +1911,25 @@ final class RecoveryTests: XCTestCase {
...
         // TODO: Old parser expected error on line 4: expected ',' joining parts of a multi-clause condition, Fix-It replacements: 15 - 21 = ','
-        DiagnosticSpec(locationMarker: "1️⃣", message: "expected '{' in 'if' statement", fixIts: ["insert '{'"]),
-        DiagnosticSpec(locationMarker: "1️⃣", message: "unexpected code 'where y == 0,' before variable"),
-        DiagnosticSpec(locationMarker: "2️⃣", message: "expected '}' to end 'if' statement", fixIts: ["insert '}'"]),
-      ]
+        DiagnosticSpec(message: "expected ',' joining parts of a multi-clause condition",
+                       fixIts: ["replace 'where' with ','"])
+      ],
+      fixedSource: """
+      if let y = x , y == 0, let z = x {
```
kimdv, line 1928: It would be nice to remove the space before the replaced `,`.

Is there a nice way to do?

ahoppen: I think you might just need to change `.all` to `.fixedUp` here. Not sure which other consequences on tests that might have though.

https://github.com/apple/swift-syntax/blob/3d53f9d01d7da1556e0eeca323f8244b82f89748/Sources/SwiftParserDiagnostics/DiagnosticExtensions.swift#LL170C19-L170C19

kimdv: Added a commit that shows the difference. 
It fixed something, but multiline string literal is not working correctly tough 

210f4feaec872d3b88f439dce8642f8097153857

ahoppen: Ah, I think you’re hitting the multiline string literal formatting issues that, I think, we also saw in your PR that introduced `fixedSource` everywhere.

What I would try is the following (more might be necessary):
- Introduce `requiresTrailingNewline` to `BasicFormat` and only return `true` for opening multiline string delimiters
- In `BasicFormat.visitPre` when entering a multi-line string literal, pick up the user defined indentation from the closing `"""`

Hopefully this will fix formatting issues inside multiline string literals in general.

kimdv: https://github.com/apple/swift-syntax/pull/1623

kimdv:
@ahoppen this is ready to review now 😁

ahoppen (approved):
Thanks. LGTM

ahoppen:
@swift-ci Please test

PR #1617, opened 2023-04-28
Add documentation for initializers that build a node from a header and a result builder

ahoppen:
rdar://108643660

### DIFF
--- a/Sources/SwiftSyntaxBuilder/SyntaxNodeWithBody.swift
+++ b/Sources/SwiftSyntaxBuilder/SyntaxNodeWithBody.swift
@@ -40,6 +40,25 @@ extension SyntaxStringInterpolation {
 public protocol HasTrailingCodeBlock {
   var body: CodeBlockSyntax { get set }
 
+  /// Constructs a syntax node where `header` builds the text of the node before the body in braces and `bodyBuilder` is used to build the node’s body.
+  ///
+  /// For example, you can construct
+  ///
+  /// ```swift
+  /// while x < 5 {
+  ///   x += 1
+  /// }
+  /// ```
+  ///
+  /// using this call
+  ///
+  /// ```swift
+  /// try WhileStmtSyntax("while x < 5") {
+  ///   ExprSyntax("x += 1")
+  /// }
+  /// ```
+  ///
+  /// Throws an error if `header` defines a different node type than the type the initializer is called on. E.g. if calling `try IfStmtSyntax("while x < 5") {}`
   init(_ header: PartialSyntaxNodeString, @CodeBlockItemListBuilder bodyBuilder: () throws -> CodeBlockItemListSyntax) rethrows
 }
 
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

kimdv (approved):

ahoppen:
@swift-ci Please test Windows

Sources/SwiftSyntaxBuilder/SyntaxNodeWithBody.swift
```diff
@@ -158,6 +226,18 @@ extension SwitchCaseSyntax {
 // So we cannot conform to `HasTrailingCodeBlock` or `HasTrailingMemberDeclBlock`
 
 public extension SwitchExprSyntax {
+  /// Constructs an `switch` expression where `header` builds the text before the opening `{` and `casesBuilder` can be used to build the case items.
```
bnbarham, line 229:
```suggestion
  /// Constructs a `switch` expression where `header` builds the text before the opening `{` and `casesBuilder` can be used to build the case items.
```

Sources/SwiftSyntaxBuilder/SyntaxNodeWithBody.swift
```diff
@@ -158,6 +226,18 @@ extension SwitchCaseSyntax {
...
+  ///
+  /// For example, the following switch expression hase header `switch direction` and two case items for `up` and `down`
```
bnbarham, line 231:
```suggestion
  /// For example, the following switch expression has header `switch direction` and two case items for `up` and `down`
```

Sources/SwiftSyntaxBuilder/SyntaxNodeWithBody.swift
```diff
@@ -132,6 +173,30 @@ public extension IfExprSyntax {
...
+  /// if x == 1 {
+  ///     return "one
+  /// } else if x == 2 {
+  ///     return "two
+  /// } else {
+  ///     return "many
+  /// }
+  /// ```
+  ///
+  /// You can use
```
bnbarham, line 189: I like how this one has an actual example of how to construct it. How would you feel about adding something similar to the others? If not that's fine, some documentation is better than none :). But a lot of the `For example the following ... has ...` could just be code instead.

Sources/SwiftSyntaxBuilder/SyntaxNodeWithBody.swift
```diff
@@ -95,6 +113,16 @@ extension InitializerDeclSyntax: HasTrailingOptionalCodeBlock {}
 public protocol HasTrailingMemberDeclBlock {
   var memberBlock: MemberDeclBlockSyntax { get set }
 
+  /// Constructs a syntax node where `header` builds the text of the node before the members in braces and `membersBuilder` is used to list the node’s members.
+  ///
+  /// For example the following function has header `struct Point` and the body will contain two `DeclSyntax` with text `var x: Int` and `var y: Int`, respecitively.
```
bnbarham, line 118:
```suggestion
  /// For example the following function has header `struct Point` and the body will contain two `DeclSyntax` with text `var x: Int` and `var y: Int`, respectively.
```

Sources/SwiftSyntaxBuilder/SyntaxNodeWithBody.swift
ahoppen: Added more examples

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

PR #1618, opened 2023-04-28
Migrate build script to Swift

### LINKED ISSUE #1465: Rewrite `build-script.py` in Swift
`build-script.py` is the only remaining Python file in SwiftSyntax. It would be nice to rewrite it in Swift so the only language used in SwiftSyntax is Swift (+ a tiny bit of C in `Sources/_InstructionCounter`).

kimdv:
Fixes #1465


kimdv:
Open PR to show the progress. 

I'm already open for feedback but it's still WIP

SwiftSyntaxCLI/Package.swift
```diff
@@ -0,0 +1,53 @@
...
+      dependencies: [
+        "InstructionCounter",
+        .product(name: "ArgumentParser", package: "swift-argument-parser"),
+        .product(name: "SwiftDiagnostics", package: "swift-syntax"),
+        .product(name: "SwiftOperators", package: "swift-syntax"),
+        .product(name: "SwiftParser", package: "swift-syntax"),
+        .product(name: "SwiftParserDiagnostics", package: "swift-syntax"),
+        .product(name: "SwiftSyntax", package: "swift-syntax"),
+      ]
+    ),
```
kimdv, line 33: @ahoppen should we have a single executable or two? 

We can split it as I've done with the build an test etc.

ahoppen:
One high-level comment: The re-implementation of the build script should not be living in the swift-syntax package, i.e. the package that’s defined by the root-level `Package.swift`. If it is, everybody depending on swift-syntax could depend on the build script, which we definitely don’t want.

Instead, I think the build script should live in the `CodeGeneration` package (which would then obviously need to be renamed). That way we could also remove the `generate-swiftsyntax` executable and instead invoke the source generation code directly from the rewritten build-script.

ahoppen:
One high-level comment: I think the build script should live in the `CodeGeneration` package (which would then obviously need to be renamed). That way we could also remove the `generate-swiftsyntax` executable and instead invoke the source generation code directly from the rewritten build-script.

I would like to keep `swift-parser-cli` separate from the build script because they serve to drastically different purposes. The build script’s purpose is to build SwiftSyntax in CI and provide the glue code with the build script to the build-script in the compiler repo. As such, it’s only interesting during build time. `swift-parser-cli` is a runtime tool that can be super useful to inspect SwiftSyntax trees and e.g. triage bugs. For example, if someone is hitting a SwiftSyntax crash but can’t share their source code, you might suggest that they run `swift-parser-cli reduce` and just attach the reduced source code (which with all likelihood contains no sensitive information). As such, I see `swift-parser-cli` also useful for adopters of SwiftSyntax.

kimdv:
> One high-level comment: I think the build script should live in the `CodeGeneration` package (which would then obviously need to be renamed). That way we could also remove the `generate-swiftsyntax` executable and instead invoke the source generation code directly from the rewritten build-script.

Cool. I'll try to merge CodeGeneration package and this one into a single Package.

> I would like to keep `swift-parser-cli` separate from the build script because they serve to drastically different purposes. The build script’s purpose is to build SwiftSyntax in CI and provide the glue code with the build script to the build-script in the compiler repo. As such, it’s only interesting during build time. `swift-parser-cli` is a runtime tool that can be super useful to inspect SwiftSyntax trees and e.g. triage bugs. For example, if someone is hitting a SwiftSyntax crash but can’t share their source code, you might suggest that they run `swift-parser-cli reduce` and just attach the reduced source code (which with all likelihood contains no sensitive information). As such, I see `swift-parser-cli` also useful for adopters of SwiftSyntax.

Just to be clear, I would revert the commit or will you still move it out of the swift-syntax package but just in it's own package? (As now. It is it's own executable)
My tought was to move it out, so the swift-syntax package don't have any dependency on swift-argument-parser.

ahoppen:
> Just to be clear, I would revert the commit or will you still move it out of the swift-syntax package but just in it's own package? (As now. It is it's own executable) My tought was to move it out, so the swift-syntax package don't have any dependency on swift-argument-parser.

Yes, I think moving `swift-parser-cli` to a different package does make sense. I was just thrown off by the PR title and thought that `swift-parser-cli` would be the vehicle for migrating build-script and that’s definitely not what we want.

kimdv:
> > Just to be clear, I would revert the commit or will you still move it out of the swift-syntax package but just in it's own package? (As now. It is it's own executable) My tought was to move it out, so the swift-syntax package don't have any dependency on swift-argument-parser.
> 
> Yes, I think moving `swift-parser-cli` to a different package does make sense. I was just thrown off by the PR title and thought that `swift-parser-cli` would be the vehicle for migrating build-script and that’s definitely not what we want.

Will remove this PR so it focus on migrate the build script only then

ahoppen:
I think a separate PR for moving `swift-parser-cli` would be great, but that’s a different topic.

SwiftSyntaxCLI/Sources/swift-syntax-cli/common/Builder.swift
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    args += additionalArguments
+
+    let process = Process()
+    let outputPipe = Pipe()
+    process.environment = env
+    process.executableURL = URL(fileURLWithPath: arguments.toolchain).appendingPathComponent("bin").appendingPathComponent("swift")
+    process.arguments = args
+    process.standardOutput = outputPipe
```
ahoppen, line 94: I think it would be nice to wrap the `Process` API behind a custom abstraction layer, mostly because I don’t like how it works (for example how much overhead you need to write to get `stdout` and `stderr` or how you manually need to implement `PATH` lookup).

An API that I would like would be something like the following. I have not implemented the PATH lookup in it yet or tested it thoroughly but I think having something like this will make invoking other terminal commands a lot simpler everywhere.

```swift
/// Error thrown if a process terminates with a non-zero exit code.
struct NonZeroExitCodeError: Error {
  let stdout: String
  let stderr: String
  let exitCode: Int
}

/// Error thrown if `stdout` or `stderr` could not be decoded as UTF-8.
struct FailedToDecodeUTF8Error: Error {
  let data: Data
}

/// If `name` is not a fully qualified path, looks an executable with that name
/// up in `PATH` and runs it with `args`. If `additionalEnvironment` is specified,
/// the current environment will be augmented by these values.
///
/// Throws in either of these conditions:
///  - The executable could not be found or could not be run
///  - Stdout or stderr weren't valid UTF-8
///  - The process executes with a non-zero exit code.
@discardableResult
func runProcess(name: String, args: [String], additionalEnvironment: [String: String] = [:]) throws -> (stdout: String, stderr: String) {
  let process = Process()

  // Set up stdout
  let stdoutPipe = Pipe()
  let stdoutHandle = stdoutPipe.fileHandleForReading
  var stdoutData = Data()
  stdoutHandle.readabilityHandler = { pipe in
    stdoutData += pipe.availableData
  }
  process.standardOutput = stdoutPipe

  // Set up stderr
  let stderrPipe = Pipe()
  let stderrHandle = stderrPipe.fileHandleForReading
  var stderrData = Data()
  stderrHandle.readabilityHandler = { pipe in
    stderrData += pipe.availableData
  }
  process.standardError = stderrPipe

  // Set up environment
  var environment = ProcessInfo.processInfo.environment
  for (key, value) in additionalEnvironment {
    environment[key] = value
  }
  process.environment = environment

  // Set up arguments
  process.executableURL = /* do the path lookup dance I mentioned in a different comment */
  process.arguments = args

  // Run the process
  try process.run()
  process.waitUntilExit()

  guard let stdoutString = String(data: stdoutData, encoding: .utf8) else {
    throw FailedToDecodeUTF8Error(data: stdoutData)
  }
  guard let stderrString = String(data: stderrData, encoding: .utf8) else {
    throw FailedToDecodeUTF8Error(data: stderrData)
  }

  if process.terminationStatus != 0 {
    throw NonZeroExitCodeError(stdout: stdoutString, stderr: stderrString, exitCode: Int(process.terminationStatus))
  }

  return (stdoutString, stderrString)
}
```

SwiftSyntaxCLI/Sources/swift-syntax-cli/common/Builder.swift
```diff
@@ -0,0 +1,120 @@
...
+
+    let outputHandle = outputPipe.fileHandleForReading
+    outputHandle.waitForDataInBackgroundAndNotify()
```
ahoppen, line 97: Why do you need to call `waitForDataInBackgroundAndNotify`? I’ve never seen this needed before (which isn’t to say that it’s not needed).

SwiftSyntaxCLI/Sources/swift-syntax-cli/common/Paths.swift
```diff
@@ -0,0 +1,59 @@
+//
+//  File.swift
+//
+//
+//  Created by Kim de Vos on 26/04/2023.
+//
```
ahoppen, line 6: Wrong header?

SwiftSyntaxCLI/Sources/swift-syntax-cli/common/Paths.swift
```diff
@@ -0,0 +1,59 @@
...
+
+import Foundation
+
+enum Paths {
+  static var modules: [String] {
+    ["SwiftBasicFormat", "SwiftIDEUtils", "SwiftParser", "SwiftSyntax", "SwiftSyntaxBuilder"]
+  }
```
ahoppen, line 13: Technically, these aren’t paths, right? Also, this doesn’t include all modules. Given that you’re only using it in `VerifySourceCode`, I think it should just be specified there.

SwiftSyntaxCLI/Sources/swift-syntax-cli/common/Paths.swift
```diff
@@ -0,0 +1,59 @@
...
+      .appendingPathComponent("Examples")
+  }
+
+  static var codeGenerationDir: URL {
+    packageDir
+      .appendingPathComponent("CodeGeneration")
+  }
+
+  static var workspaceDir: URL {
+    packageDir
```
ahoppen, line 36: I think `workspaceDir` should be the folder containing `swift-syntax`. So this should be `packageDir.deletingLastPathComponent()`. And you don’t need to call `deletingLastPathComponent` in `llvmDir`.

ahoppen (commented):
Thanks for doing this, @kimdv. I think the Swift implementation is really so much easier to read and work with. I’ve got a few comments though. Just to warn you, once you’ve addressed those, I might come up with another big-ish set of comments, but we’re definitely heading in the right direction.

SwiftSyntaxCLI/Sources/swift-syntax-cli/common/Builder.swift
bnbarham: Could just grab ProcessRunner.swift from the stress tester. It also handles adding a wait for getting the rest of stdout/stderr.

ahoppen:
One thought just came to me: With the current design, you won't be able to build the build script (or `swift-Syntax-cli` or whatever the name ends up being) if SwiftSyntax doesn't build itself, because the build script depends on SwiftSyntax. I think that's quite unfortunate and very surprising. 

With that in mind, I think CodeGeneration will need to continue being its own package so that the build script itself can continue to be free of a SwiftSyntax dependency.

SwiftSyntaxCLI/Sources/swift-syntax-cli/commands/Test.swift
```diff
@@ -0,0 +1,169 @@
...
+
+import ArgumentParser
+import Foundation
+
+struct Test: ParsableCommand, Builder {
+  @OptionGroup
+  var arguments: BuilderArguments
+
+  @Flag(help: "Don't run lit-based tests")
+  var skipLitTests: Bool = false
```
rauhul, line 21: fly-by comment, I don't know if you have an existing cli you need to match, but imo these sort of options read better as:
```
@Flag(help: "Run lit-based tests", inversion: .enableDisable)
var litTests: Bool = true
```

ahoppen: We’d like to maintain the behavior of `build-script.py` for now. Also, I’d like to get rid of the lit-based test anyway, so this flag can hopefully be defined away soon anyway.

SwiftSyntaxCLI/Sources/swift-syntax-cli/common/Builder.swift
kimdv: I picked the `ProcessRunner.swift` and trimmed it a bit so it just had what we needed.

ahoppen: Great. It’s probably the better choice as well. Thanks @bnbarham for remembering me that it existed.

kimdv:
@ahoppen I think I've resolved all feedback. Ready for round 2 😁

SwiftSyntaxDevUtils/Sources/swift-syntax-dev-utils/commands/Build.swift
```diff
@@ -0,0 +1,27 @@
...
+struct Build: ParsableCommand, Builder {
+  @OptionGroup
+  var arguments: BuilderArguments
+
+  func run() throws {
+    try buildProduct(productName: "SwiftSyntax-all")
+
+    // Build examples
+    try buildExample(exampleName: "AddOneToIntegerLiterals")
+    try buildExample(exampleName: "CodeGenerationUsingSwiftSyntaxBuilder")
```
ahoppen, line 25: Since https://github.com/apple/swift-syntax/pull/1660 you can build `Examples-all` to build all examples.

SwiftSyntaxDevUtils/Sources/swift-syntax-dev-utils/common/ScriptExectutionError.swift
```diff
@@ -0,0 +1,15 @@
...
+//
+// Copyright (c) 2014 - 2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+struct ScriptExectutionError: Error {
```
ahoppen, line 13: I would also make this convertible to `CustomStringConvertible`. It makes the errors show up a little nicer on the command line.

SwiftSyntaxDevUtils/README.md
ahoppen, line 1: This file needs to be updated to `swift-syntax-dev-utils` instead of `SwiftSyntaxCLI`.

SwiftSyntaxDevUtils/Package.swift
```diff
@@ -0,0 +1,33 @@
+// swift-tools-version:5.7
+
+import PackageDescription
+import Foundation
+
+let package = Package(
+  name: "SwiftSyntaxDevUtils",
```
ahoppen, line 7: We recently renamed the `swift-syntax` package to `swift-syntax`. For consistency, this should also need to be named `swift-syntax-dev-utils`.

SwiftSyntaxDevUtils/Sources/swift-syntax-dev-utils/commands/Test.swift
```diff
@@ -0,0 +1,172 @@
...
+  private func findSwiftpmBinPath(packageDir: URL) throws -> String {
+    return try invokeSwiftPM(
+      action: "build",
+      packageDir: packageDir,
+      additionalArguments: ["--show-bin-path"],
+      additionalEnvironment: [:]
+    )
+  }
+
+  private func findProductBinPath() throws -> URL {
```
ahoppen, line 163: Could you add a short comment which path this points to, ie. an example path and/or what the path contains.

SwiftSyntaxDevUtils/Sources/swift-syntax-dev-utils/commands/Test.swift
```diff
@@ -0,0 +1,172 @@
...
+          repo.
+          Refer to README.md for more information.
+          """
+      )
+    }
+
+    let productBinPath = try findProductBinPath()
+    let examplesBinPath = try findExamplesBinPath()
+
+    let litTestHelperExec = productBinPath.appendingPathComponent("/lit-test-helper")
```
ahoppen, line 80: Do we need the slash here? I don’t think so.

SwiftSyntaxDevUtils/Sources/swift-syntax-dev-utils/commands/Test.swift
```diff
@@ -0,0 +1,172 @@
...
+
+    // Print all failures
+    litCall += ["--verbose"]
+    // Don't show all commands if verbose is not enabled
+    if !arguments.verbose {
+      litCall += ["--succinct"]
+    }
+
+    guard let pythonExec = Paths.pythonExec else {
+      throw ScriptExectutionError(message: "Didn't find python executable")
```
ahoppen, line 105:
```suggestion
      throw ScriptExectutionError(message: "Didn't find python3 executable")
```

SwiftSyntaxDevUtils/Sources/swift-syntax-dev-utils/common/Builder.swift
```diff
@@ -0,0 +1,102 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+
+protocol Builder {
```
ahoppen, line 15: To be consistent: Name this `BuildCommand`?

SwiftSyntaxDevUtils/Sources/swift-syntax-dev-utils/common/Builder.swift
```diff
@@ -0,0 +1,102 @@
...
+      additionalEnvironment["SWIFTSYNTAX_ENABLE_RAWSYNTAX_VALIDATION"] = "1"
+    }
+
+    if arguments.enableTestFuzzing {
+      additionalEnvironment["SWIFTPARSER_ENABLE_ALTERNATE_TOKEN_INTROSPECTION"] = "1"
+    }
+
+    // Tell other projects in the unified build to use local dependencies
+    additionalEnvironment["SWIFTCI_USE_LOCAL_DEPS"] = "1"
+    additionalEnvironment["SWIFT_SYNTAX_PARSER_LIB_SEARCH_PATH"] = arguments.toolchain.appendingPathComponent("/lib/swift/macos").path
```
ahoppen, line 93: This should be calling `appendingPathComponent` multiple times for every component.

SwiftSyntaxDevUtils/Sources/swift-syntax-dev-utils/common/ProcessRunner.swift
```diff
@@ -0,0 +1,117 @@
...
+
+  public init(
+    executableURL: URL,
+    arguments: [String],
+    additionalEnvironment: [String: String] = [:]
+  ) {
+    process = Process()
+    process.executableURL = executableURL
+    process.arguments = arguments
+    process.environment = additionalEnvironment.merging(ProcessInfo.processInfo.environment) { (current, _) in current }
```
ahoppen, line 29: This prefers the value from `additionalEnvironment`, right? In that case, I think `current` should be named `additional`.

ahoppen (commented):
Looks pretty good now. Have you tried whether this also works on Linux? I’m a little worried because we’re using `Foundation` quite a bit and I don’t know what the current state of (corelibs-)Foundation on Linux.

SwiftSyntaxDevUtils/README.md
kimdv: That readme was wrong. We don't have the code gen in that. It's a package for it self.

kimdv:
> One thought just came to me: With the current design, you won't be able to build the build script (or `swift-Syntax-cli` or whatever the name ends up being) if SwiftSyntax doesn't build itself, because the build script depends on SwiftSyntax. I think that's quite unfortunate and very surprising.
> 
> With that in mind, I think CodeGeneration will need to continue being its own package so that the build script itself can continue to be free of a SwiftSyntax dependency.

I was thinking, if it not can be a single package. Why does it work with two? 
If `swift-syntax-dev-utils` calls `CodeGen`. 

If there is a code error in SwiftSyntax, it will fail no matter what? 
Or am I missing something?

kimdv:
> Looks pretty good now. Have you tried whether this also works on Linux? I’m a little worried because we’re using `Foundation` quite a bit and I don’t know what the current state of (corelibs-)Foundation on Linux.

No I haven't tried it on Linux. I'm not a Linux user. 
I'll see if I can spin up a container and build it. Or do you have an better idea?

ahoppen:
> I was thinking, if it not can be a single package. Why does it work with two? If `swift-syntax-dev-utils` calls `CodeGen`.
> 
> If there is a code error in SwiftSyntax, it will fail no matter what? Or am I missing something?

We have two different build steps now:

Build step 1 builds the build script (or as we now call it `swift-dev-utils`) itself. Previously no build was necessary here because the build script was written in Python. Building the build script should be very fast and, most importantly, shouldn’t build SwiftSyntax itself. 

Building SwiftSyntax should only happen when the build script is actually *run*. If we already build SwiftSyntax while building the build script, there’s no point in having a build script after all. 

Does this make sense?

> No I haven't tried it on Linux. I'm not a Linux user.
> I'll see if I can spin up a container and build it. Or do you have a better idea?

You could also just fire up CI and see what it says. I’m a little concerned at the moment because 
a) I’m not sure how much of Foundation that we use in `swift-dev-utils` is available on Linux and Windows. But it might be fine
b) I’m not entirely sure if a swift compiler is installed on the Windows and Linux CI nodes at the moment. Remember that we are always building a Swift compiler as part of the CI run, but we cannot use that compiler to run the build script because Foundation is built after swift-syntax. However, this is an issue that will need to be fixed soon anyway because otherwise we can’t build SwiftParser into the compiler to be used with macros. So this might just delay this PR for a bit.

bnbarham:
>b) I’m not entirely sure if a swift compiler is installed on the Windows and Linux CI nodes at the moment. Remember that we are always building a Swift compiler as part of the CI run, but we cannot use that compiler to run the build script because Foundation is built after swift-syntax. However, this is an issue that will need to be fixed soon anyway because otherwise we can’t build SwiftParser into the compiler to be used with macros. So this might just delay this PR for a bit.

There is no host swift compiler (yet). Having said that, can't this use the just-built compiler?

ahoppen:
> There is no host swift compiler (yet). Having said that, can't this use the just-built compiler?

We are using Foundation and the just-built toolchain doesn’t have Foundation built at this point yet.

stevapple:
> I’m not entirely sure if a swift compiler is installed on the Windows and Linux CI nodes at the moment.

For Windows specifically, we’re not even using `build-script.py` at this moment. The Windows build script calls CMake directly. SwiftSyntax is now built as one of the last two packages on Windows, even after SwiftPM, which means it’s possible to build `SwiftSyntaxDevUtils` at least for now.

Saleem had complained about the Windows CI short of disk storage, and adding a host toolchain could make things worse. Bootstrapping Swift on Windows is also blocked by some `SILOptimization` bugs, AFAIK macros still not working.

kimdv:
@swift-ci please test

build-script.py
ahoppen, line 1: I think we should sunset `build-script.py` in this PR to make sure that the Swift build script actually works. I see two ways of doing that. For that, I think we should update https://github.com/apple/swift/blob/main/utils/swift_build_support/swift_build_support/products/swiftsyntax.py to call directly into `swift-syntax-dev-utils`, either using `swift run` or using `swift build` and then running the executable (I’d prefer the first)

kimdv:
https://github.com/apple/swift/pull/66573

@swift-ci please test

kimdv:
https://github.com/apple/swift/pull/66573

@swift-ci please test

kimdv:
https://github.com/apple/swift/pull/66573

@swift-ci please test

kimdv:
https://github.com/apple/swift/pull/66573

@swift-ci please test

kimdv:
https://github.com/apple/swift/pull/66573

@swift-ci please test

kimdv:
@ahoppen can you see what the failure is. 
Have looked trough the logs and can't find it.

ahoppen:
I can’t find an error message either. What seems to happen, is that `swift run swift-syntax-dev-utils verify-source-code` returns with a non-zero exit code without writing to stdout or stderr. What I would try, is the following
- Setting the `SWIFTCI_USE_LOCAL_DEPS` environment variable when running `swift-syntax-dev-utils` to make sure we don’t access the internet to get swift-argument-parser.
- Add a `print` and `printerr` (that function exists somewhere in the codebase) statement at the start of the `verify-source-code` command and see if that message shows up in the log. If it doesn’t, there might be some problem with building swift-syntax-dev-utils altogether
- In the end you might not have a better choice than to run `apple/swift/utils/build-script` locally and make it invoke `swift-syntax-dev-utils` to try and reproduce the failure locally. If you do, you might be able to see more information or run `swift-syntax-dev-utils` in exactly the same way that `apple/swift/utils/build-script` runs it, which would hopefully yield some additional insights.

ahoppen:
Forgot what I said above, I found the failure

You are calling 
```
.../toolchain-macosx-x86_64/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr` 
```

but `usr` is a directory and you’re missing `bin/swift`

kimdv:
https://github.com/apple/swift/pull/66573

@swift-ci please test

kimdv:
https://github.com/apple/swift/pull/66573

@swift-ci please test

kimdv:
https://github.com/apple/swift/pull/66573

@swift-ci please test

kimdv:
https://github.com/apple/swift/pull/66573

@swift-ci please test

kimdv:
https://github.com/apple/swift/pull/66573

@swift-ci please test

kimdv:
https://github.com/apple/swift/pull/66573

@swift-ci please test

SwiftSyntaxDevUtils/Sources/swift-syntax-dev-utils/common/Paths.swift
```diff
@@ -0,0 +1,94 @@
...
+    return
+      envSearchPaths
+      .first(where: { $0.path.contains("python3") })
+  }
+
+  static var diffExec: URL? {
+    return
+      envSearchPaths
+      .first(where: { $0.path.contains("diff") })
+  }
```
ahoppen, line 68: `envSearchPaths` will be the contents of `echo $PATH` in a shell, which is the directory containing executables. Thus, the path itself doesn’t include `diff`. You need to implement something equivalent to `lookupExecutableInPath`

https://github.com/apple/swift-tools-support-core/blob/ef2db24dbe7e39ebfabca2c10905909708c5738b/Sources/TSCBasic/misc.swift#L217-L246

SwiftSyntaxDevUtils/README.md
```diff
@@ -0,0 +1 @@
+# swift-syntax-dev-utils
```
ahoppen, line 1: What do you think about adding a line like

> Scripts to help build swift-syntax in CI. In most cases, you should not need to run these scripts yourself. The [Contributing Guide](../Contributing.md) contains information of how to build swift-syntax locally.

SwiftSyntaxDevUtils/Sources/swift-syntax-dev-utils/commands/Test.swift
```diff
@@ -0,0 +1,163 @@
...
+
+    let process = ProcessRunner(
+      executableURL: pythonExec,
+      arguments: litCall
+    )
+
+    try process.run()
+  }
+
+  private func runXctests() throws {
```
ahoppen, line 103: Nit, I would capitalize `C` and `T`

```suggestion
  private func runXCTests() throws {
```

SwiftSyntaxDevUtils/Sources/swift-syntax-dev-utils/common/BuildArguments.swift
```diff
@@ -0,0 +1,69 @@
...
+
+  @Flag(
+    help:
+      """
+      When constructing RawSyntax nodes validate that their layout matches that
+      defined in `CodeGeneration` and that TokenSyntax nodes have a `tokenKind`
+      matching the ones specified in `CodeGeneration`.
+      """
+  )
+  var enableRawsyntaxValidation: Bool = false
```
ahoppen, line 56: Nit

```suggestion
  var enableRawSyntaxValidation: Bool = false
```

build-script.py
ahoppen, line 1: We don’t need build-script.py anymore after this PR is merged, right? So we should be able to just delete it.

kimdv:
https://github.com/apple/swift/pull/66573

@swift-ci please test

kimdv:
https://github.com/apple/swift/pull/66573

@swift-ci please test

SwiftSyntaxDevUtils/Sources/swift-syntax-dev-utils/commands/VerifySourceCode.swift
```diff
@@ -0,0 +1,67 @@
...
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import ArgumentParser
+import Foundation
+
+fileprivate var modules: [String] {
+  ["SwiftBasicFormat", "SwiftIDEUtils", "SwiftParser", "SwiftSyntax", "SwiftSyntaxBuilder"]
```
ahoppen, line 17: The modules that contain generated files have changed.

```suggestion
  ["SwiftParser", "SwiftParserDiagnsotics", "SwiftSyntax", "SwiftSyntaxBuilder"]
```

kimdv:
https://github.com/apple/swift/pull/66573

@swift-ci please test

SwiftSyntaxDevUtils/Sources/swift-syntax-dev-utils/commands/VerifySourceCode.swift
```diff
@@ -0,0 +1,67 @@
...
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import ArgumentParser
+import Foundation
+
+fileprivate var modules: [String] {
+    ["SwiftParser", "SwiftParserDiagnsotics", "SwiftSyntax", "SwiftSyntaxBuilder"]
```
ahoppen, line 17: Typo

```suggestion
    ["SwiftParser", "SwiftParserDiagnostics", "SwiftSyntax", "SwiftSyntaxBuilder"]
```

kimdv:
https://github.com/apple/swift/pull/66573

@swift-ci please test

SwiftSyntaxDevUtils/Sources/swift-syntax-dev-utils/common/BuildArguments.swift
```diff
@@ -0,0 +1,69 @@
...
+  var multirootDataFile: URL?
+
+  @Flag(help: "Disable sandboxes when building with SwiftPM")
+  var disableSandbox: Bool = false
+
+  @Flag(help: "Build in release mode.")
+  var release: Bool = false
+
+  @Flag(
+    help:
```
ahoppen, line 49: To remain compatible with `build-script.py`

```suggestion
  @Flag(
    name: .customLong("enable-rawsyntax-validation"),
    help:
```

kimdv:
https://github.com/apple/swift/pull/66573

@swift-ci please test

SwiftSyntaxDevUtils/Sources/swift-syntax-dev-utils/common/Paths.swift
```diff
@@ -0,0 +1,110 @@
...
+    llvmDir
+      .appendingPathComponent("utils")
+      .appendingPathComponent("lit")
+      .appendingPathComponent("lit.py")
+  }
+
+  static var python3Exec: URL? {
+    return
+      envSearchPaths
+      .first(where: { $0.path.contains("python3") })
```
ahoppen, line 61: You need to do the same trick you did for `diff` here.

kimdv:
https://github.com/apple/swift/pull/66573

@swift-ci please test

kimdv:
https://github.com/apple/swift/pull/66573

@swift-ci please test

kimdv:
https://github.com/apple/swift/pull/66573

@swift-ci please test

kimdv:
https://github.com/apple/swift/pull/66573

@swift-ci please test

ahoppen:
To debug the latest failure (and in general), I think it would be good if `swift-syntax-dev-utils` would print the commands for all subprocesses it runs if it has `--verbose`, similar to 

https://github.com/apple/swift-syntax/blob/861333cd66fa80e6dc0d0d11d0c05e218418e1b8/build-script.py#L95-L96C22

kimdv:
https://github.com/apple/swift/pull/66573

@swift-ci please test

kimdv:
https://github.com/apple/swift/pull/66573

@swift-ci please test

kimdv:
https://github.com/apple/swift/pull/66573

@swift-ci please test

kimdv:
https://github.com/apple/swift/pull/66573

@swift-ci please test

SwiftSyntaxDevUtils/Sources/swift-syntax-dev-utils/commands/Test.swift
```diff
@@ -0,0 +1,160 @@
...
+
+  private func runXCTests() throws {
+    logSection("Running XCTests")
+    var swiftpmCallArguments: [String] = []
+
+    if arguments.verbose {
+      swiftpmCallArguments += ["--verbose"]
+    }
+
+    swiftpmCallArguments += ["--test-product", "SwiftSyntaxPackageTests"]
```
ahoppen, line 115: We renamed the package from `SwiftSyntax` to `swift-syntax` at some point

```suggestion
    swiftpmCallArguments += ["--test-product", "swift-syntaxPackageTests"]
```

kimdv:
https://github.com/apple/swift/pull/66573

@swift-ci please test

kimdv:
https://github.com/apple/swift/pull/66573

@swift-ci please test

SwiftSyntaxDevUtils/Sources/swift-syntax-dev-utils/common/Paths.swift
```diff
@@ -0,0 +1,111 @@
...
+    let pathArg = "Path"
+    #else
+    let pathArg = "PATH"
+    #endif
+
+    return ProcessInfo.processInfo.environment[pathArg]
+  }
+
+  private static func lookupExecutable(for filename: String) -> URL? {
+    return (envSearchPaths + envSearchPaths.map { $0.appendingPathComponent(filename) })
```
ahoppen, line 92: I just saw the same failure on Linux as well for my format.swift script. Fix is to change this to

```suggestion
    return envSearchPaths.map { $0.appendingPathComponent(filename) }
```

ahoppen (commented):
🥳 macOS is passing. Linux can’t be far

kimdv:
https://github.com/apple/swift/pull/66573

@swift-ci please test

kimdv:
https://github.com/apple/swift/pull/66573

@swift-ci please test windows

PR #1623, opened 2023-05-01
Fix wrong formatting of multiline string literal

### LINKED ISSUE #1603: `"""` should be inserted on a newline
### Description

testMultilineErrors26 https://github.com/apple/swift-syntax/pull/1549#discussion_r1179897323
testMultilinePoundDiagnosticArgRdar411547971 https://github.com/apple/swift-syntax/pull/1549#discussion_r1179897767

### Steps to Reproduce

_No response_

kimdv:
Fixes #1603 

I tried you comment here: https://github.com/apple/swift-syntax/pull/1616#discussion_r1180813550

But it didn't work for me. 
What I figured out what because if it have an open `"""` but not a closing `"""` then the formatter will only format on the closing one. 
It sounds to me that works as intended. Only visit the tokens we fix.

### DIFF
--- a/Sources/SwiftBasicFormat/BasicFormat.swift
+++ b/Sources/SwiftBasicFormat/BasicFormat.swift
@@ -143,35 +143,49 @@ open class BasicFormat: SyntaxRewriter {
     return node.requiresIndent
   }
 
-  /// Whether a leading newline on `token` should be added.
-  open func requiresLeadingNewline(_ token: TokenSyntax) -> Bool {
-    // We don't want to add newlines inside string interpolation
-    if isInsideStringInterpolation(token) {
+  open func requiresNewline(between first: TokenSyntax?, and second: TokenSyntax?) -> Bool {
+    // We don't want to add newlines inside string interpolation.
+    // When first or second `TokenSyntax` is a multiline quote we want special handling
+    // even if it's inside a string interpolation, because it still requires newline
+    // after open quote and before close quote.
+    if let first,
+      isInsideStringInterpolation(first),
+      first.tokenKind != .multilineStringQuote,
+      second?.tokenKind != .multilineStringQuote
+    {
+      return false
+    } else if let second {
+      if second.requiresLeadingNewline {
+        return true
+      }
+
+      var ancestor: Syntax = Syntax(second)
+      while let parent = ancestor.parent {
+        ancestor = parent
+        if ancestor.position != second.position || ancestor.firstToken(viewMode: viewMode) != second {
+          break
+        }
+        if let ancestorsParent = ancestor.parent, childrenSeparatedByNewline(ancestorsParent) {
+          return true
+        }
+        switch ancestor.keyPathInParent {
+        case \IfConfigClauseSyntax.elements:
+          return true
+        default:
+          break
+        }
+      }
+    }
+
+    switch (first?.tokenKind, second?.tokenKind) {
+    case (.multilineStringQuote, .multilineStringQuote),  // string interpolation segment inside a multi-line string literal
+      (.multilineStringQuote, .stringSegment),  // empty multi-line string literal
+      (.stringSegment, .multilineStringQuote),  // segment starting a multi-line string literal
+      (.rightParen, .multilineStringQuote):  // ending a multi-line string literal that has a string interpolation segment at its end
+      return true
+    default:
       return false
     }
-
-    if token.requiresLeadingNewline {
-      return true
-    }
-
-    var ancestor: Syntax = Syntax(token)
-    while let parent = ancestor.parent {
-      ancestor = parent
-      if ancestor.position != token.position || ancestor.firstToken(viewMode: viewMode) != token {
-        break
-      }
-      if let ancestorsParent = ancestor.parent, childrenSeparatedByNewline(ancestorsParent) {
-        return true
-      }
-      switch ancestor.keyPathInParent {
-      case \IfConfigClauseSyntax.elements:
-        return true
-      default:
-        break
-      }
-    }
-
-    return false
   }
 
   open func requiresWhitespace(between first: TokenSyntax?, and second: TokenSyntax?) -> Bool {
@@ -298,10 +312,7 @@ open class BasicFormat: SyntaxRewriter {
       if previousToken.trailingTrivia.endsWithNewline {
         return true
       }
-      if case .stringSegment(let segment) = previousToken.tokenKind, segment.last?.isNewline ?? false {
-        return true
-      }
-      return false
+      return previousToken.isStringSegmentWithLastCharacterBeingNewline
     }()
 
     lazy var previousTokenIsStringLiteralEndingInNewline: Bool = {
@@ -310,10 +321,7 @@ open class BasicFormat: SyntaxRewriter {
         // don't add a leading newline to the file.
         return true
       }
-      if case .stringSegment(let segment) = previousToken.tokenKind, segment.last?.isNewline ?? false {
-        return true
-      }
-      return false
+      return previousToken.isStringSegmentWithLastCharacterBeingNewline
     }()
 
     lazy var nextTokenWillStartWithWhitespace: Bool = {
@@ -321,7 +329,7 @@ open class BasicFormat: SyntaxRewriter {
         return false
       }
       return nextToken.leadingTrivia.startsWithWhitespace
-        || (requiresLeadingNewline(nextToken) && isMutable(nextToken))
+        || (requiresNewline(between: token, and: nextToken) && isMutable(nextToken))
     }()
 
     lazy var nextTokenWillStartWithNewline: Bool = {
@@ -329,7 +337,7 @@ open class BasicFormat: SyntaxRewriter {
         return false
       }
       return nextToken.leadingTrivia.startsWithNewline
-        || (requiresLeadingNewline(nextToken) && isMutable(nextToken))
+        || (requiresNewline(between: token, and: nextToken) && isMutable(nextToken) && !token.trailingTrivia.endsWithNewline && !token.isStringSegmentWithLastCharacterBeingNewline)
     }()
 
     /// This token's trailing trivia + any spaces or tabs at the start of the
@@ -342,7 +350,7 @@ open class BasicFormat: SyntaxRewriter {
     var leadingTrivia = token.leadingTrivia
     var trailingTrivia = token.trailingTrivia
 
-    if requiresLeadingNewline(token) {
+    if requiresNewline(between: previousToken, and: token) {
       // Add a leading newline if the token requires it unless
       //  - it already starts with a newline or
       //  - the previous token ends with a newline
@@ -405,5 +413,16 @@ open class BasicFormat: SyntaxRewriter {
     }
 
     return token.detach().with(\.leadingTrivia, leadingTrivia).with(\.trailingTrivia, trailingTrivia)
+  }
+}
+
+fileprivate extension TokenSyntax {
+  var isStringSegmentWithLastCharacterBeingNewline: Bool {
+    switch self.tokenKind {
+    case .stringSegment(let segment):
+      return segment.last?.isNewline ?? false
+    default:
+      return false
+    }
   }
 }
--- a/Sources/SwiftBasicFormat/BasicFormat.swift
+++ b/Sources/SwiftBasicFormat/BasicFormat.swift
@@ -178,9 +178,10 @@ open class BasicFormat: SyntaxRewriter {
     }
 
     switch (first?.tokenKind, second?.tokenKind) {
-    case (.multilineStringQuote, .multilineStringQuote),  // string interpolation segment inside a multi-line string literal
-      (.multilineStringQuote, .stringSegment),  // empty multi-line string literal
-      (.stringSegment, .multilineStringQuote),  // segment starting a multi-line string literal
+    case (.multilineStringQuote, .backslash),  // string interpolation segment inside a multi-line string literal
+      (.multilineStringQuote, .multilineStringQuote),  // empty multi-line string literal
+      (.multilineStringQuote, .stringSegment),  // segment starting a multi-line string literal
+      (.stringSegment, .multilineStringQuote),  // ending a multi-line string literal that has a string interpolation segment at its end
       (.rightParen, .multilineStringQuote):  // ending a multi-line string literal that has a string interpolation segment at its end
       return true
     default:
--- a/Sources/SwiftBasicFormat/BasicFormat.swift
+++ b/Sources/SwiftBasicFormat/BasicFormat.swift
@@ -296,6 +296,12 @@ open class BasicFormat: SyntaxRewriter {
     let previousToken = self.previousToken ?? token.previousToken(viewMode: viewMode)
     let nextToken = token.nextToken(viewMode: viewMode)
 
+    /// In addition to existing trivia of `previousToken`, also considers
+    /// `previousToken` as ending with whitespace if it and `token` should be
+    /// separated by whitespace.
+    /// It does not consider whether a newline should be added between
+    /// `previousToken` and the `token` because that newline should be added to
+    /// the next token's trailing trivia.
     lazy var previousTokenWillEndWithWhitespace: Bool = {
       guard let previousToken = previousToken else {
         return false
@@ -304,6 +310,8 @@ open class BasicFormat: SyntaxRewriter {
         || (requiresWhitespace(between: previousToken, and: token) && isMutable(previousToken))
     }()
 
+    /// This method does not consider any posssible mutations to `previousToken`
+    /// because newlines should be added to the next token's leading trivia.
     lazy var previousTokenWillEndWithNewline: Bool = {
       guard let previousToken = previousToken else {
         // Assume that the start of the tree is equivalent to a newline so we
@@ -325,6 +333,10 @@ open class BasicFormat: SyntaxRewriter {
       return previousToken.isStringSegmentWithLastCharacterBeingNewline
     }()
 
+    /// Also considers `nextToken` as starting with a whitespace if a newline
+    /// should be added to it. It does not check whether `token` and `nextToken`
+    /// should be separated by whitespace because the whitespace should be added
+    /// to the `token`’s leading trivia.
     lazy var nextTokenWillStartWithWhitespace: Bool = {
       guard let nextToken = nextToken else {
         return false
@@ -333,6 +345,8 @@ open class BasicFormat: SyntaxRewriter {
         || (requiresNewline(between: token, and: nextToken) && isMutable(nextToken))
     }()
 
+    /// Also considers `nextToken` as starting with a leading newline if `token`
+    /// and `nextToken` should be separated by a newline.
     lazy var nextTokenWillStartWithNewline: Bool = {
       guard let nextToken = nextToken else {
         return false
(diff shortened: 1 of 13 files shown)


Tests/SwiftParserTest/ExpressionTests.swift
```diff
@@ -577,7 +577,8 @@ final class ExpressionTests: XCTestCase {
         )
       ],
       fixedSource: ##"""
-        """""""
+        """"
+        """
```
kimdv, line 581: In this case it should actually be

```swift
"""
"
"""
```

That's properly the parser not doing things 100% correctly.

Tests/SwiftParserTest/translated/UnclosedStringInterpolationTests.swift
```diff
@@ -126,7 +126,8 @@ final class UnclosedStringInterpolationTests: XCTestCase {
       fixedSource: ##"""
         _ = """
         \(
-        """""")"""
+        """ """)
+        """
```
kimdv, line 130: Here we require a whitespace because it's a multiline literal. 
A whitespace is a newline or a space.

Sources/SwiftBasicFormat/BasicFormat.swift
```diff
@@ -249,9 +255,10 @@ open class BasicFormat: SyntaxRewriter {
     case \ExpressionSegmentSyntax.backslash,
       \ExpressionSegmentSyntax.rightParen,
       \DeclNameArgumentSyntax.colon,
-      \StringLiteralExprSyntax.openQuote,
       \RegexLiteralExprSyntax.openSlash:
       return false
+    case \StringLiteralExprSyntax.openQuote:
+      return first?.tokenKind == .multilineStringQuote
```
ahoppen, line 261: This seems a little odd to me. A multiline string quote should be followed by a newline, not just whitespace. I think we might need to introduce `requiresTrailingNewline` for that.

Tests/SwiftParserTest/translated/AvailabilityQueryUnavailabilityTests.swift
```diff
@@ -103,7 +103,7 @@ final class AvailabilityQueryUnavailabilityTests: XCTestCase {
         )
       ],
       fixedSource: """
-        if #unavailable(OSX 10.51) , #unavailable(OSX 10.52) {
+        if #unavailable(OSX 10.51), #unavailable(OSX 10.52) {
```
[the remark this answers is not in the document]
kimdv: There wasn't an issue yet

Sources/SwiftBasicFormat/BasicFormat.swift
```diff
@@ -171,9 +171,31 @@ open class BasicFormat: SyntaxRewriter {
...
+    switch syntaxToken.keyPathInParent {
+    case \StringLiteralExprSyntax.closeQuote:
+      return token.tokenKind == .multilineStringQuote
+    default:
+      return false
+    }
   }
 
+    /// Whether a trailing newline on `token` should be added.
+    open func requiresTrailingNewline(_ token: TokenSyntax) -> Bool {
```
ahoppen, line 184: I think you will also need to check this in `previousTokenWillEndWithWhitespace`.

Sources/SwiftBasicFormat/BasicFormat.swift
```diff
@@ -171,9 +171,31 @@ open class BasicFormat: SyntaxRewriter {
...
+        // We don't want to add newlines inside string interpolation
+        if isInsideStringInterpolation(token) {
+            return false
```
ahoppen, line 187: 4 space indentation 👀

ahoppen (commented):
I’m not sure if it changes anything test-wise but since you are adding a trailing newline, that should also be reflected in `previousTokenWillEndWithWhitespace`. I.e. the return there should be replaced by the following (slightly refactored to improve readability.

```swift
      if previousToken.trailingTrivia.endsWithWhitespace {
        return true
      }
      if isMutable(previousToken) && (requiresWhitespace(between: previousToken, and: token) || requiresTrailingNewline(previousToken)) {
        return true
      }
      return false
```

ahoppen (approved):
Nice. This ended up being a pretty small change after all 😆

Unrelated to this PR, I think there’s still an error where we don’t insert the newline after the opening `"""`:

```swift

assertBuildResult(
  StringLiteralExprSyntax(openQuote: .multilineStringQuoteToken(), content: "a", closeQuote: .multilineStringQuoteToken()),
  #"""
  """
  a
  """
  """#
)
```

Would you like to look into that as well? If you don’t want to look at it straight away, I will file an issue for it.

ahoppen:
@swift-ci Please test

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test linux

kimdv:
@swift-ci please test Windows

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test

kimdv:
I don't understand why it says I haven't generated files.
When running locally there is nothing new ? 🤨

kimdv:
> I don't understand why it says I haven't generated files.
> When running locally there is nothing new ? 🤨

Found the reason. 
The reason that is that I had build and run code generation before. It did **not** look for a never version. 
Looks a bit like a bug to me. 

Tried to resolve and update but got those errors
```zsh
kimdevos@dk-kimdevos12-8752 CodeGeneration % swift package resolve
Updating /Users/kimdevos/repository/kimdv/swift-src/swift-syntax
Updated /Users/kimdevos/repository/kimdv/swift-src/swift-syntax (0.47s)
error: could not find a branch named ‘HEAD’ in /Users/kimdevos/repository/kimdv/swift-src/swift-syntax
kimdevos@dk-kimdevos12-8752 CodeGeneration % swift package update 
Updating /Users/kimdevos/repository/kimdv/swift-src/swift-syntax
Updated /Users/kimdevos/repository/kimdv/swift-src/swift-syntax (0.17s)
error: could not find a branch named ‘HEAD’ in /Users/kimdevos/repository/kimdv/swift-src/swift-syntax
```

I deleted the `.build` and re-run the `swift package resolve` and it worked fine

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test windows

ahoppen:
> Found the reason. The reason that is that I had build and run code generation before. It did **not** look for a never version. Looks a bit like a bug to me.

I think that’s intentional. Unless you run `swift package update`, it won’t update your package version. Could you add a comment to our docs where we suggest `swift run --package-path CodeGeneration generate-swiftsyntax Sources` to run `swift package update` beforehand?

> I deleted the `.build` and re-run the `swift package resolve` and it worked fine.
> 
> Looks a bit like a bug in SPM?

I was able to reproduce this as well but then couldn’t reproduce it again by creating a new commit and running `swift package update`. If you’ve got steps to reliably reproduce this, could you file an issue for SwiftPM?

Sources/SwiftBasicFormat/BasicFormat.swift
```diff
@@ -171,7 +171,29 @@ open class BasicFormat: SyntaxRewriter {
...
+      return false
+    }
+
+    let syntaxToken: Syntax = Syntax(token)
+    switch syntaxToken.keyPathInParent {
+    case \StringLiteralExprSyntax.openQuote:
+      return token.tokenKind == .multilineStringQuote
+    default:
+      return false
+    }
```
ahoppen, line 196: I think you could just define `requiresTrailingNewline` in terms of `requiresNewline(between:and:)`: `requiresTrailingNewline(token)` is equivalent to to `requiresNewline(between: token, and: nextToken)`

Same for `requiresLeadingNewline`. That way we have a single source of truth for defining whether two tokens should be separated by newlines.

Sources/SwiftBasicFormat/BasicFormat.swift
```diff
@@ -362,6 +393,12 @@ open class BasicFormat: SyntaxRewriter {
       anchorPoints[token] = currentIndentationLevel
     }
 
+    if requiresTrailingNewline(token) {
+      if !trailingTrivia.endsWithNewline && nextToken?.leadingTrivia.startsWithNewline == false {
```
ahoppen, line 397: Shouldn’t you be checking for `nextTokenWillStartWithNewline` here to make sure we’re not adding both a trailing newline here and a leading newline to the next token?
```suggestion
      if !trailingTrivia.endsWithNewline && !nextTokenWillStartsWithNewline {
```

ahoppen (commented):
I think you also need to modify `previousTokenWillEndWithNewline` to return `true` if `previousToken` is mutable and requires a trailing newline. I also wrote some documentation for the current behavior of these variables. Could you include those. Hopefully, they also explain a little more about how they operate. Reading the code again, I understand that it was a little magic.

<details><summary>New comments</summary>

```diff
diff --git a/Sources/SwiftBasicFormat/BasicFormat.swift b/Sources/SwiftBasicFormat/BasicFormat.swift
index 752c6f83..7d74a05d 100644
--- a/Sources/SwiftBasicFormat/BasicFormat.swift
+++ b/Sources/SwiftBasicFormat/BasicFormat.swift
@@ -276,6 +276,12 @@ open class BasicFormat: SyntaxRewriter {
     let previousToken = self.previousToken ?? token.previousToken(viewMode: viewMode)
     let nextToken = token.nextToken(viewMode: viewMode)
 
+    /// In addition to existing trivia of `previousToken`, also considers
+    /// `previousToken` as ending with whitespace if it and `token` should be
+    /// separated by whitespace.
+    /// It does not consider whetehr a newline should be added between
+    /// `previousToken` and the `token` because that newline should be added to
+    /// the next token's trailing trivia.
     lazy var previousTokenWillEndWithWhitespace: Bool = {
       guard let previousToken = previousToken else {
         return false
@@ -284,6 +290,8 @@ open class BasicFormat: SyntaxRewriter {
         || (requiresWhitespace(between: previousToken, and: token) && isMutable(previousToken))
     }()
 
+    /// This method does not consider any posssible mutations to `previousToken`
+    /// because newlines should be added to the next token's leading trivia.
     lazy var previousTokenWillEndWithNewline: Bool = {
       guard let previousToken = previousToken else {
         // Assume that the start of the tree is equivalent to a newline so we
@@ -311,6 +319,10 @@ open class BasicFormat: SyntaxRewriter {
       return false
     }()
 
+    /// Also considers `nextToken` as starting with a whitespace if a newline
+    /// should be added to it. It does not check whether `token` and `nextToken`
+    /// should be separated by whitespace because the whitespace should be added
+    /// to the `token`’s leading trivia.
     lazy var nextTokenWillStartWithWhitespace: Bool = {
       guard let nextToken = nextToken else {
         return false
@@ -319,6 +331,8 @@ open class BasicFormat: SyntaxRewriter {
         || (requiresLeadingNewline(nextToken) && isMutable(nextToken))
     }()
 
+    /// Also considers `nextToken` as starting with a leading newline if `token`
+    /// and `nextToken` should be separated by a newline.
     lazy var nextTokenWillStartWithNewline: Bool = {
       guard let nextToken = nextToken else {
         return false
```

</details>

Sources/SwiftBasicFormat/BasicFormat.swift
```diff
@@ -393,7 +430,7 @@ open class BasicFormat: SyntaxRewriter {
     trailingTrivia = trailingTrivia.indented(indentation: trailingTriviaIndentation, isOnNewline: false)
 
     leadingTrivia = leadingTrivia.trimmingTrailingWhitespaceBeforeNewline(isBeforeNewline: false)
-    trailingTrivia = trailingTrivia.trimmingTrailingWhitespaceBeforeNewline(isBeforeNewline: nextTokenWillStartWithNewline)
+    trailingTrivia = trailingTrivia.trimmingTrailingWhitespaceBeforeNewline(isBeforeNewline: nextTokenWillStartWithNewline || trailingTrivia.startsWithNewline)
```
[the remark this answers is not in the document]
kimdv: If we don't do it then 
```swift
      #"""
      """
      \(gen(xx) { (x) in
          return """
          case
      """
      })
      """
      """#
```

becomes 

```swift
      #"""
      """
      \(gen(xx) { (x) in
          return """
              case
      """
      })
      """
      """#
```

What I can see it is because it's moving spacing in front of the string segment that contains the `case`

Sources/SwiftBasicFormat/BasicFormat.swift
```diff
@@ -171,7 +171,50 @@ open class BasicFormat: SyntaxRewriter {
...
+          (.rightParen, .identifier),
+          (.rightParen, .keyword(.async)),
+          (.rightSquareBracket, _),
+          (_, .colon),
+          (_, .eof),
+          (_, .equal):
+          return false
+      default:
+          return true
+      }
```
kimdv, line 217: @ahoppen did you think on something like this?

Note: This is not the final result. Did just try to pass the first couple of test cases. 
Not sure if we should default to false and then only return true if needed. 

And if we should try to get the previous code from `requireLeadingNewline` into this method or not.

ahoppen: I was thinking more of something like this that’s very much based on `requiresLeadingNewline` and actually replaces `requiresLeadingNewline`. 

```swift
open func requiresNewline(between first: TokenSyntax?, and second: TokenSyntax?) -> Bool {
  // NOTE: From requiresLeadingNewline

  // We don't want to add newlines inside string interpolation
  if isInsideStringInterpolation(first) && token.tokenKind != .multilineStringQuote {
    return false
  }

  // NOTE: From AnyKeyPath.requiresLeadingTrivia, which would become obsolete
  switch second.keyPathInParent {
  case \AccessorBlockSyntax.rightBrace:
    return true
  case \ClosureExprSyntax.rightBrace:
    return true
  case \CodeBlockSyntax.rightBrace:
    return true
  case \IfConfigClauseSyntax.poundKeyword:
    return true
  case \IfConfigDeclSyntax.poundEndif:
    return true
  case \MemberDeclBlockSyntax.rightBrace:
    return true
  case \SwitchExprSyntax.rightBrace:
    return true
  default:
    break
  }

  // NOTE: Whatever you need to add for your PR
  switch first.keyPathInParent {
  case \StringLiteralExprSyntax.openQuote where first.tokenKind == .multilineStringQuote:
    return true
  default:
    break
  }

  // NOTE: From requiresLeadingNewline
  var ancestor: Syntax = Syntax(second)
  while let parent = ancestor.parent {
    ancestor = parent
    if ancestor.position != second.position || ancestor.firstToken(viewMode: viewMode) != second {
      break
    }
    if let ancestorsParent = ancestor.parent, childrenSeparatedByNewline(ancestorsParent) {
      return true
    }
    switch ancestor.keyPathInParent {
    case \IfConfigClauseSyntax.elements:
      return true
    default:
      break
    }
  }

  return false
}
```

Tests/SwiftParserTest/ExpressionTests.swift
```diff
@@ -668,7 +671,8 @@ final class ExpressionTests: XCTestCase {
         DiagnosticSpec(message: ##"expected '"""#' to end string literal"##, fixIts: [##"insert '"""#'"##])
       ],
       fixedSource: ##"""
-        #"""a"""#
+        #"""a
+        """#
```
kimdv, line 675: Here is a bug here too. 
We should in the parser make an error (If it's not already there) and apply a correct fix-it.

kimdv:
@swift-ci please test

Sources/SwiftBasicFormat/BasicFormat.swift
```diff
@@ -143,35 +143,50 @@ open class BasicFormat: SyntaxRewriter {
...
-    if isInsideStringInterpolation(token) {
+  open func requiresNewline(between first: TokenSyntax?, and second: TokenSyntax?) -> Bool {
+    // We don't want to add newlines inside string interpolation.
+    // When first or second `TokenSyntax` is a multiline quote we want special handling
+    // event it's inside a string interpolation, because it still requires newline
+    // after open quote and before close quote.
+    if let first,
+      first.tokenKind != .multilineStringQuote,
+      second?.tokenKind != .multilineStringQuote,
+      isInsideStringInterpolation(first)
```
ahoppen, line 154: I would swap the order of these checks because I think that way it reads the same way as the comment above.

Sources/SwiftBasicFormat/BasicFormat.swift
```diff
@@ -143,35 +143,50 @@ open class BasicFormat: SyntaxRewriter {
...
-
-    var ancestor: Syntax = Syntax(token)
-    while let parent = ancestor.parent {
-      ancestor = parent
-      if ancestor.position != token.position || ancestor.firstToken(viewMode: viewMode) != token {
-        break
-      }
-      if let ancestorsParent = ancestor.parent, childrenSeparatedByNewline(ancestorsParent) {
+    } else if let second {
+      if second.requiresLeadingNewline {
```
ahoppen, line 158: Definitely not part of this PR, but eventually I would like to remove `requiresLeadingNewline` from CodeGen altogether and just list the tokens that require a leading newline in this function. Same for `childrenSeparatedByNewline`. I don’t think we necessarily need to CodeGen it, it could just be implemented manually in this file.

Sources/SwiftBasicFormat/BasicFormat.swift
```diff
@@ -143,35 +143,50 @@ open class BasicFormat: SyntaxRewriter {
...
       }
     }
 
-    return false
+    switch (first?.tokenKind, second?.tokenKind) {
+    case (.multilineStringQuote, .backslash),
+      (.multilineStringQuote, .multilineStringQuote),
+      (.multilineStringQuote, .stringSegment),
+      (.stringSegment, .multilineStringQuote),
+      (.rightParen, .multilineStringQuote):
```
ahoppen, line 185: Just to follow the paradigm of `tokenSeparatedByNewline`, I think it would be good to add comment to at least part of them (it took me a bit of thinking to see why `multilineStringQuote` and `backslash` need to be separated by a newline, only to realize that it’s because of a string interpolation segment starting a multi-line string literal

```suggestion
    case (.multilineStringQuote, .backslash), // string interpolation segment inside a multi-line string literal
      (.multilineStringQuote, .multilineStringQuote), // empty multi-line string literal
      (.multilineStringQuote, .stringSegment), // segment starting a multi-line string literal
      (.stringSegment, .multilineStringQuote), // ending a multi-line string literal on a new line
      (.rightParen, .multilineStringQuote): // ending a multi-line string literal that has a string interpolation segment at its end
```

ahoppen (approved):
Wohooo. Nice. This looks really good now. I only have a few remarks mostly about naming and documentation.

Sources/SwiftBasicFormat/BasicFormat.swift
kimdv: The reason I did this was of performance.

kimdv: Swapped it. 
If it should be another order I can do a follow up PR later.

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test windows

Sources/SwiftBasicFormat/BasicFormat.swift
ahoppen: I was thinking about that as well but in >99% of all cases first and second won’t be multi-line string quotes, so we hit `isInsideStringInterpolation` basically all the time anyway.

kimdv: That was my thought. If `first` or `second` not multi line string, then is fails before is calls `isInsideStringInterpolation`

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test windows

PR #1625, opened 2023-05-01
[5.9] Fix debugInitCall

ahoppen:
* **Explanation**: I broke `debugInitCall` when I removed the custom mirrors for syntax nodes. Fix it again, also fixing some more issues along the way that previously caused us to generate Swift code that didn’t compile.
* **Scope**: A method that is only intended for debugging purposes
* **Risk**: Zero, this methods is never called in any production code
* **Testing**: Manual 
* **Issue**: rdar://108756044
* **Reviewer**: @bnbarham on https://github.com/apple/swift-syntax/pull/1543


ahoppen:
@swift-ci Please test

DougGregor (approved):

PR #1627, opened 2023-05-02
Update SwiftSyntax to Swift 5.7

ahoppen:
Now that Xcode on CI nodes has been updated, we don’t need to support Swift 5.5 and 5.6 anymore.

- Increase swift-tools-version to 5.7 
- Convert to shorthand `if let` notation
  - I was able to do this using the script from https://github.com/apple/swift-syntax/pull/887
- Migrate generics to opaque parameter types (aka. `some` types)
  - This was quite a lot of manual work
- Remove `@frozen` attributes
  - These only existed as a workaround for a miscompute on Swift 5.5 and are no longer needed
- Remove functions that only existed for Swift 5.6 compatibility 
  - No longer needed

### DIFF
--- a/CodeGeneration/Sources/SyntaxSupport/TokenSpec.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/TokenSpec.swift
@@ -48,7 +48,7 @@ public class TokenSpec {
     self.name = name
     self.kind = kind
     self.nameForDiagnostics = nameForDiagnostics
-    if let unprefixedKind = unprefixedKind {
+    if let unprefixedKind {
       self.unprefixedKind = unprefixedKind
     } else {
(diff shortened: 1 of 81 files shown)


ahoppen:
@swift-ci Please test

kimdv (approved):
WUHU!

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test macOS

ahoppen:
@swift-ci Please test macOS

ahoppen:
Since CI keeps failing for me on this one and I can’t reproduce the failure locally, I’m splitting this up to at least pinpoint which commit is causing the CI failure.

PR #1630, opened 2023-05-02
Add documentation for syntax nodes and their children

whiteio:
As part of #1528, documentation needs added to some syntax nodes. This PR adds documentation to the following syntax nodes and their children:
- AssociatedtypeDeclSyntax
- ClassDeclSyntax
- ImportDeclSyntax
- IsExprSyntax
- ProtocolDeclSyntax

### DIFF
--- a/CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
@@ -209,6 +209,31 @@ public let DECL_NODES: [Node] = [
   Node(
     name: "AssociatedtypeDecl",
     nameForDiagnostics: "associatedtype declaration",
+    description: """
+      An associated type declaration like the following.
+
+      ```swift
+      associatedtype Item
+      ```
+
+      An associated type declaration may contain a type initializer clause which represents a default type assignment for the associated type.
+
+      ```swift
+      associatedtype Item = Int
+      ```
+
+      An associated type declaration may be declared with an inheritance clause which specifies the required conformances.
+
+      ```swift
+      associatedtype Iterator: IteratorProtocol
+      ```
+
+      A generic where clause may be present, here is an example which shows an associated type containing an inheritance clauses and a generic where clause.
+
+      ```swift
+      associatedtype Iterator: IteratorProtocol where Iterator.Element == Item
+      ```
+      """,
     kind: "Decl",
     traits: [
       "IdentifiedDecl",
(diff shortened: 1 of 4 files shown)


CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
```diff
@@ -1001,26 +1071,31 @@ public let DECL_NODES: [Node] = [
...
         kind: .collection(kind: "AttributeList", collectionElementName: "Attribute"),
         nameForDiagnostics: "attributes",
+        description: "Attributes attached to the import declaration.",
         isOptional: true
       ),
       Child(
         name: "Modifiers",
         kind: .collection(kind: "ModifierList", collectionElementName: "Modifier"),
         nameForDiagnostics: "modifiers",
+        description: "Modifiers attached to the import declaration",
```
ahoppen, line 1081: AFAICT these modifiers aren’t used by any language feature yet. It’s fine that we still parse them but I think it’s worth pointing that out.

```suggestion
        description: "Modifiers attached to the import declaration. Currently, no modifiers are supported by Swift.",
```

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
```diff
@@ -1001,26 +1071,31 @@ public let DECL_NODES: [Node] = [
...
       Child(
         name: "ImportTok",
-        kind: .token(choices: [.keyword(text: "import")])
+        kind: .token(choices: [.keyword(text: "import")]),
+        description: "The `import` keyword for this declaration."
       ),
       Child(
         name: "ImportKind",
         kind: .token(choices: [.keyword(text: "typealias"), .keyword(text: "struct"), .keyword(text: "class"), .keyword(text: "enum"), .keyword(text: "protocol"), .keyword(text: "var"), .keyword(text: "let"), .keyword(text: "func"), .keyword(text: "inout")]),
+        description: "The kind of declaration being imported.",
```
ahoppen, line 1092:
```suggestion
        description: "The kind of declaration being imported. For example, a struct can be imported from a ",
```

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
```diff
@@ -1001,26 +1071,31 @@ public let DECL_NODES: [Node] = [
...
         isOptional: true
       ),
       Child(
         name: "Path",
-        kind: .collection(kind: "AccessPath", collectionElementName: "PathComponent")
+        kind: .collection(kind: "AccessPath", collectionElementName: "PathComponent"),
+        description: "The access path to the module, submodule or symbol being imported."
```
ahoppen, line 1098: We’re not really happy with the name `AccessPath` here. I would just cut the word `access` here.

https://github.com/apple/swift-syntax/blob/2179d9fe6335c54e2fb13419c1186ac549d407c2/CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift#L33

I filed https://github.com/apple/swift-syntax/issues/1632 to rename the type.

CodeGeneration/Sources/SyntaxSupport/ExprNodes.swift
```diff
@@ -905,19 +905,29 @@ public let EXPR_NODES: [Node] = [
   Node(
     name: "IsExpr",
     nameForDiagnostics: "'is'",
+    description: """
+      An `is` expression like the following.
+
+      ```swift
+      value is Double
+      ```
```
ahoppen, line 913: I would add the following

> This node is only generated after operators are folded using the `SwiftOperators` library. Beforehand, the parser does not know the precedences of operators and thus represents `is` by an `UnresolvedIsExpr`.

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
```diff
@@ -1726,43 +1810,51 @@ public let DECL_NODES: [Node] = [
         name: "Attributes",
         kind: .collection(kind: "AttributeList", collectionElementName: "Attribute"),
         nameForDiagnostics: "attributes",
+        description: "Attributes attached to the protocol declaration.",
```
ahoppen, line 1813: Could you add examples of attributes and modifiers here as well? I think they could be helpful to understand the difference between the two.

ahoppen (commented):
Thanks a lot. That’s so much better. I have a few suggestions to add illustrative examples, mostly to explain the difference between attributes and modifiers. My thinking is that for many developers using the library, the raw concept of e.g. `private` is clear and those help map keywords they know to the terms used within SwiftSyntax. If you could apply those, that would be great!

whiteio:
@ahoppen Thanks for the quick feedback! I've added in your suggestions.

ahoppen (approved):
Thanks for writing the documentation and making SwiftSyntax more approachable for newcomers.

ahoppen:
@swift-ci Please test

PR #1635, opened 2023-05-04
Verify that .spi.yml is up-to-date in swift-syntax-dev-utils

ahoppen:
Motivated by https://github.com/apple/swift-syntax/pull/1634

---

.spi.yml got out of sync of the libraries we added to Package.swift. Add a verification script to make sure that they match. The idea is that for every library that every library that is exposed by SwiftSyntax should be documentated on swiftpackageindex.com.

rdar://108901461

PR #1636, opened 2023-05-04
Update Package.swift and `.spi.yml` to make them match

ahoppen:
Package.swift and `.spi.yml` have drifted apart and didn’t actually match anymore. The targets declared in `.spi.yml` should match the libraries declared in the package manifest. To fix this, update `.spi.yml` and also declare `SwiftBasicFormat` as a library in `Package.swift`. BasicFormat is being used in sufficently many places by now that we can’t really consider it an implementation detail of SwiftSyntax anymore.

I’m adding a script that verifies that Package.swift and `.spi.yml` match in https://github.com/apple/swift-syntax/pull/1635.

rdar://108902357

### DIFF
--- a/.spi.yml
+++ b/.spi.yml
@@ -8,12 +8,15 @@ builder:
     - documentation_targets:
       # First item in the list is the "landing" (default) target
       - SwiftSyntax
-      - IDEUtils
       - SwiftBasicFormat
+      - SwiftCompilerPlugin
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

DougGregor (approved):

kimdv (approved):

PR #1637, opened 2023-05-04
Run CodeGeneration with local dependencies

ahoppen:
If we want to verify that the code-generated files match, we need to use local dependencies because CI doesn’t have internet access.

rdar://108902933


ahoppen:
@swift-ci Please test

build-script.py
```diff
@@ -141,6 +141,7 @@ def run_code_generation(
     env = dict(os.environ)
     env["SWIFT_BUILD_SCRIPT_ENVIRONMENT"] = "1"
     env["SWIFTSYNTAX_ENABLE_RAWSYNTAX_VALIDATION"] = "1"
+    env["SWIFTCI_USE_LOCAL_DEPS"] = "1"
```
kimdv, line 144: What happens if we run this locally ?

PR #1638, opened 2023-05-04
Rename `AccessPathSyntax` to `ImportPathSyntax`

### LINKED ISSUE #1632: Rename `AccessPathSyntax` to `ImportPathSyntax`
Address this Fix-It.

https://github.com/apple/swift-syntax/blob/2179d9fe6335c54e2fb13419c1186ac549d407c2/CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift#L33

Also rename any properties named `accessPath`. To keep existing clients building, provide a type alias that maps `AccessPathSyntax` to `ImportPathSyntax` and deprecated methods for `accessPath` to SwiftSyntaxCompatibility.swift

whiteio:
`AccessPathSyntax` should be renamed to `ImportPathSyntax`. This PR renames `AccessPath` to `ImportPath`, `AccessPathComponent` to `ImportPathComponent`, generates the code and adds the following type aliases to keep existing clients building:

* `AccessPathBuilder` -> `ImportPathBuilder`
* `AccessPathSyntax` -> `ImportPathSyntax`
* `AccessPathComponentSyntax` -> `ImportPathComponentSyntax`

Resolves #1632

### DIFF
--- a/Sources/SwiftSyntax/generated/SyntaxEnum.swift
+++ b/Sources/SwiftSyntax/generated/SyntaxEnum.swift
@@ -16,8 +16,6 @@
 @frozen // FIXME: Not actually stable, works around a miscompile
 public enum SyntaxEnum {
   case token(TokenSyntax)
-  case accessPathComponent(AccessPathComponentSyntax)
-  case accessPath(AccessPathSyntax)
   case accessorBlock(AccessorBlockSyntax)
   case accessorDecl(AccessorDeclSyntax)
   case accessorEffectSpecifiers(AccessorEffectSpecifiersSyntax)
@@ -153,6 +151,8 @@ public enum SyntaxEnum {
   case implementsAttributeArguments(ImplementsAttributeArgumentsSyntax)
   case implicitlyUnwrappedOptionalType(ImplicitlyUnwrappedOptionalTypeSyntax)
   case importDecl(ImportDeclSyntax)
+  case importPathComponent(ImportPathComponentSyntax)
+  case importPath(ImportPathSyntax)
   case inOutExpr(InOutExprSyntax)
   case infixOperatorExpr(InfixOperatorExprSyntax)
   case inheritedTypeList(InheritedTypeListSyntax)
@@ -293,10 +293,6 @@ public extension Syntax {
     switch raw.kind {
     case .token:
       return .token(TokenSyntax(self)!)
-    case .accessPathComponent:
-      return .accessPathComponent(AccessPathComponentSyntax(self)!)
-    case .accessPath:
-      return .accessPath(AccessPathSyntax(self)!)
     case .accessorBlock:
       return .accessorBlock(AccessorBlockSyntax(self)!)
     case .accessorDecl:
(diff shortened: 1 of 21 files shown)


Sources/SwiftSyntax/generated/SyntaxEnum.swift
```diff
@@ -16,8 +16,6 @@
 @frozen // FIXME: Not actually stable, works around a miscompile
 public enum SyntaxEnum {
   case token(TokenSyntax)
-  case accessPathComponent(AccessPathComponentSyntax)
-  case accessPath(AccessPathSyntax)
```
ahoppen, line 20: Note to myself (no action required on your side) since this is the first time we are trying to do a rename while maintaining source compatibility as much as possible and I’m trying to assess where source breakages might occur. 

We can’t rename a type without also renaming the case in `SyntaxEnum`. I think that it might be fine because there are way fewer people switching over a `SyntaxEnum` than there are using the type names.

Sources/SwiftSyntax/generated/SyntaxKind.swift
```diff
@@ -16,8 +16,6 @@
 @frozen // FIXME: Not actually stable, works around a miscompile
 public enum SyntaxKind {
   case token
-  case accessPathComponent
-  case accessPath
```
ahoppen, line 20: Same here, just a note to myself, no action required on your side: This is not source compatible if we are renaming a syntax node’s type name. We might be able to mitigate this by defining a static member on `SyntaxKind` as follows

```swift
public extension SyntaxKind {
  @available(*, deprecated, renamed: "importPath")
  var accessPath: SyntaxKind { .importPath }
}
```

This way we will only break clients that are exhaustively switching over `SyntaxKind`, which is fairly unlikely since the enum is so huge.

Sources/SwiftSyntaxBuilder/SwiftSyntaxBuilderCompatibility.swift
```diff
@@ -0,0 +1,17 @@
...
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+// This file provides compatiblity aliases to keep dependents of SwiftSyntaxBuilder building.
+// All users of the declarations in this file should transition away from them ASAP.
+
+@available(*, deprecated, renamed: "ImportPathBuilder")
```
ahoppen, line 16: Thanks for noticing this 👍🏽

ahoppen (approved):
Thanks. This looks great to me. ✅ 

I left a few review comments for myself because this will somewhat become the blueprint of which compatibility functions we need to create when renaming a syntax node kind and I want to keep track of where we intentionally continue to break source compatibility (which is fine, we just want to minimize the impact as much as possible).

ahoppen:
@swift-ci Please test

whiteio:
@ahoppen The failing CI checks don't seem to be related to the changes in the PR, could you re-run the checks please?

kimdv:
@swift-ci please test

PR #1640, opened 2023-05-05
Increase swift-tools-version to 5.7

ahoppen:
We no longer need to support building SwiftSyntax with Swift <5.7 so we can increase the tools version.

rdar://108753223


ahoppen:
@swift-ci Please test

Issue #1607, opened 2023-04-28
Wrong spacing around `:` in Fix-It for ternary expressions


kimdv:
### Description

https://github.com/apple/swift-syntax/pull/1549#discussion_r1179887649
Test case `testVariableDeclarations`

ahoppen:
Tracked in Apple’s issue tracker as rdar://108656823

PR #1641, opened 2023-05-08
Insert a space before colon in ternary expression in Fix-It

gibachan:
This pull request addresses #1607

### DIFF
--- a/Sources/SwiftBasicFormat/BasicFormat.swift
+++ b/Sources/SwiftBasicFormat/BasicFormat.swift
@@ -237,10 +237,9 @@ open class BasicFormat: SyntaxRewriter {
       (nil, _):
       return false
     case (_, .colon):
-      let isTernaryExpr = second?.parent?.is(UnresolvedTernaryExprSyntax.self) == true
-      || second?.parent?.is(TernaryExprSyntax.self) == true
-
-      if !isTernaryExpr {
+      if second?.keyPathInParent != \TernaryExprSyntax.colonMark
+        && second?.keyPathInParent != \UnresolvedTernaryExprSyntax.colonMark
+      {
         return false
       }
     case (.leftAngle, _) where second?.tokenKind != .rightAngle:  // `<` and `>` need to be separated by a space because otherwise they become an operator
--- a/Sources/SwiftBasicFormat/BasicFormat.swift
+++ b/Sources/SwiftBasicFormat/BasicFormat.swift
@@ -223,7 +223,6 @@ open class BasicFormat: SyntaxRewriter {
       (.singleQuote, .rawStringDelimiter),  // closing raw string delimiter should never be separate by a space
       (.stringQuote, .rawStringDelimiter),  // closing raw string delimiter should never be separate by a space
       (.stringSegment, _),
-      (_, .colon),
       (_, .comma),
       (_, .ellipsis),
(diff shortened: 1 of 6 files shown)


kimdv:
@swift-ci please test

ahoppen (commented):
@gibachan, welcome to SwiftSyntax’s development. 

I would prefer to adjust the formatting in `BasicFormat` instead of `ParseDiagnsoticsGenerator` because that way we also get the correct formatting you you are constructing a ternary operator e.g. from `SwiftSyntaxBuilder`, which is what people will do when writing macros. 

In that case, it should work both when the operators are folded and when they haven’t been folded yet (see https://swiftpackageindex.com/apple/swift-syntax/main/documentation/swiftoperators).

An unfolded ternary operator would be the following

```swift
SequenceExprSyntax {
  BooleanLiteralExprSyntax(true)
  UnresolvedTernaryExprSyntax(firstChoice: IntegerLiteralExprSyntax(1))
  IntegerLiteralExprSyntax(0)
}
```

and a folded ternary operator would be

```swift
TernaryExprSyntax(
  conditionExpression: BooleanLiteralExprSyntax(true),
  firstChoice: IntegerLiteralExprSyntax(1),
  secondChoice: IntegerLiteralExprSyntax(0)
)
```

The code that you’d like to change is most likely in here https://github.com/apple/swift-syntax/blob/cecb0340a95f00f8f2725c912b494266ed7b4a80/Sources/SwiftBasicFormat/BasicFormat.swift#L177

gibachan:
Thank you for your feedback. I'll take a look into it!

Sources/SwiftBasicFormat/BasicFormat.swift
```diff
@@ -237,6 +236,13 @@ open class BasicFormat: SyntaxRewriter {
       (_, nil),
       (nil, _):
       return false
+    case (_, .colon):
+      let isTernaryExpr = second?.parent?.is(UnresolvedTernaryExprSyntax.self) == true
+      || second?.parent?.is(TernaryExprSyntax.self) == true
+
+      if !isTernaryExpr {
+        return false
+      }
```
ahoppen, line 245: You can write this slightly  shorter by using `keyPathInParent`, just like we do in the second `switch` statement below.

```suggestion
    case (_, .colon):
      if second?.keyPathInParent != \TernaryExprSyntax.colonMark
        && second?.keyPathInParent != \UnresolvedTernaryExprSyntax.colonMark
      {
        return false
      }
```

ahoppen (commented):
Thanks. This looks great! I’ve got one slight comment on how to improve implementation, otherwise this is ready to be merged. 👍🏽

Sources/SwiftBasicFormat/BasicFormat.swift
gibachan: Great! Thank you 👍

ahoppen (approved):
Thank you for your contribution, @gibachan. This looks good to me. Once CI passes, I’ll merge the PR.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

PR #1642, opened 2023-05-08
Migrate SwiftSyntax to shorthand `let` notation

ahoppen:
Now that we no longer need to support Swift <5.7, we can use shorthand `let` syntax. I performed the migration using the script from https://github.com/apple/swift-syntax/pull/887.


ahoppen:
@swift-ci Please test

kimdv (approved):

PR #1643, opened 2023-05-08
Fix wrong fix-it in modifier for variable decl

kimdv:


Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -1481,6 +1481,40 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
     if shouldSkip(node) {
       return .skipChildren
     }
+
+    if let modifiers = node.modifiers, modifiers.hasError {
+      for modifier in modifiers {
+        if let detail = modifier.detail, detail.detail != .keyword(.set) {
+          let unexpectedBetweenLeftParenAndDetail = detail.unexpectedBetweenLeftParenAndDetail?.tokens(viewMode: .all)
+          let unexpectedBetweenDetailAndRightParen = detail.unexpectedBetweenDetailAndRightParen?.tokens(viewMode: .all)
+          let unexpectedTokens = (unexpectedBetweenLeftParenAndDetail?.map { $0 } ?? []) + (unexpectedBetweenDetailAndRightParen?.map { $0 } ?? [])
```
kimdv, line 1490: Didn't know if there was a more nice way to get all unexpected tokens.

Tests/SwiftParserTest/DeclarationTests.swift
```diff
@@ -410,12 +421,17 @@ final class DeclarationTests: XCTestCase {
...
-      private(1️⃣get, set2️⃣, didSet) var a = 0
+      private(1️⃣get, set, didSet) var a = 0
       """,
       diagnostics: [
-        DiagnosticSpec(locationMarker: "1️⃣", message: "unexpected code 'get,' in modifier"),
-        DiagnosticSpec(locationMarker: "2️⃣", message: "unexpected code ', didSet' in modifier"),
-      ]
+        DiagnosticSpec(
+          message: "expected 'set' in modifier",
+          fixIts: ["replace 'get, , didSet' with 'set'"]
```
kimdv, line 429: This fix-it message looks a bit strange for me.

Tests/SwiftParserTest/DeclarationTests.swift
```diff
@@ -410,12 +421,17 @@ final class DeclarationTests: XCTestCase {
...
+        )
+      ],
+      fixedSource: """
+        private(set) var a = 0
+        """
     )
 
     assertParse(
```
kimdv, line 437: This assert and fixedSource is not correct either. 
But that is in the parser and not the diagnostic generator. 

Think that is a PR for it self

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -1481,6 +1481,40 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
+
+          // If there is no unexpected tokens it means we miss a paren or set keyword.
+          // So we just skip the handling here
+          guard !unexpectedTokens.isEmpty else {
+            return .visitChildren
```
ahoppen, line 1495: I think we still want to continue to the handling below (`let missingTries = …`), so we shouldn’t `return` here.

---

To remove the force unwrap below, this check could also be 

```swift
guard let firstUnexpected = unexpected tokens.first else { … }
```

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
ahoppen: I think conceptually this is the right approach because you only want the unexpected tokens between the parentheses. I think you can slightly simplify it code-wise to

```swift
[detail.unexpectedBetweenLeftParenAndDetail, detail.unexpectedBetweenDetailAndRightParen]
  .compactMap { $0?.tokens(viewMode: .all) }
```

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -1481,6 +1481,40 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
     if shouldSkip(node) {
       return .skipChildren
     }
+
+    if let modifiers = node.modifiers, modifiers.hasError {
+      for modifier in modifiers {
+        if let detail = modifier.detail, detail.detail != .keyword(.set) {
```
ahoppen, line 1487: Why do we need to check for `detail.detail != .keyword(.set)` here?

---

You could decrease the nesting level a bit by doing

```swift
guard let detail = modifier.detail, detail.detail != .keyword(.set) else {
  continue
}
```

instead.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -1481,6 +1481,40 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
+            return .visitChildren
+          }
+
+          addDiagnostic(
+            unexpectedTokens.first!,
+            MissingNodesError(missingNodes: [Syntax(detail.detail)]),
+            fixIts: [
+              FixIt(
+                message: ReplaceTokensFixIt(
+                  replaceTokens: unexpectedTokens.map { $0 },
```
ahoppen, line 1504: Why do you need `map { $0 }` here?

Tests/SwiftParserTest/DeclarationTests.swift
```diff
@@ -374,17 +373,29 @@ final class DeclarationTests: XCTestCase {
...
       """
       private(1️⃣get: set) var a = 0
       """,
       diagnostics: [
-        DiagnosticSpec(message: "unexpected code 'get:' in modifier")
-      ]
+        DiagnosticSpec(
+          message: "expected 'set' in modifier",
+          fixIts: ["replace 'get:' with 'set'"]
```
ahoppen, line 393: It would be nice if the Fix-It said: `remove 'get:'` instead

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
ahoppen, line 1504: Why do you need `map { $0 }` here?

Tests/SwiftParserTest/DeclarationTests.swift
ahoppen: I agree. Maybe this can be fixed at the same time as my comment above that the Fix-It should say “remove” instead of “replace”. So this could become `remove 'get,' and ', didSet'`

kimdv: I used the `RemoveNodesFixIt` but it doesn't include the "and". It format it as a comma separated list. 
We could make a new one or update the existing one. I think the latter would be the best.

ahoppen: I wanted to know what was going wrong here and you nerd-sniped me into this. Fixing it in https://github.com/apple/swift-syntax/pull/1651. With that PR it will say `remove 'get,' and ', didSet'`.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -1525,6 +1525,47 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
+
+        // If there is no unexpected tokens it means we miss a paren or set keyword.
+        // So we just skip the handling here
+        guard let firstUnexpected = unexpectedTokens.first else {
+          continue
+        }
+
+        let fixItMessage: ParserFixIt
+
+        if unexpectedTokens.count == 1 {
```
ahoppen, line 1547: How is whether we replace or remove tokens related to the number of unexpected tokens. Shouldn’t this be the following?

```suggestion
        if detail.detail.presence == .missing {
```

And that got me thinking. If `detail.detail` is present, we shouldn’t be emitting a `MissingNodesError` at all. In that case we should probably just `continue` here and emit a defaulted `unexpected code` diagnostic.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -1525,6 +1525,47 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
+
+    if let modifiers = node.modifiers, modifiers.hasError {
+      for modifier in modifiers {
+        guard let detail = modifier.detail else {
+          continue
+        }
+
+        let unexpectedTokens: [TokenSyntax] = [detail.unexpectedBetweenLeftParenAndDetail, detail.unexpectedBetweenDetailAndRightParen]
+          .compactMap { $0?.tokens(viewMode: .all) }
+          .flatMap { $0 }
```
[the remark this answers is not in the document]
kimdv: I don't think so. 

We want a `[TokenSyntax]` and `.compactMap { $0?.tokens(viewMode: .all) }` returns a `TokenSequence`.

When using the flatMap it results in a `Type of expression is ambiguous without more context`.

Even if I change to this below it fails with same error message
```swift 
.flatMap { tokenSequence -> TokenSyntax? in
              return tokenSequence?.tokens(viewMode: .all)
          }
```

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
kimdv: It's because if we have `private(get)` I want to replace it with `(set)`. 
Or should we then remove get and the left/right paren.

Tests/SwiftParserTest/DeclarationTests.swift
```diff
@@ -374,17 +373,29 @@ final class DeclarationTests: XCTestCase {
       private(1️⃣get, set) var a = 0
       """,
       diagnostics: [
-        DiagnosticSpec(message: "unexpected code 'get,' in modifier")
-      ]
+        DiagnosticSpec(
+          message: "expected 'set' in modifier",
+          fixIts: ["remove 'get,'"]
+        )
+      ],
```
kimdv, line 380: If we continue if `detail.detail.presence == .present` then we get `unexpected code 'get,' in modifier` and there is no `"remove 'get,'"` Fix-It

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
kimdv: https://github.com/apple/swift-syntax/pull/1643/files#r1190729627

kimdv: https://github.com/apple/swift-syntax/pull/1643#discussion_r1190729627

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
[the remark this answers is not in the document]
ahoppen: Ah, I forgot that it was returning a `TokenSequence` instead of an `Array`. You are right.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
ahoppen: I think this raises the general question of whether we should create a Fix-It to remove unexpected tokens. So far I have not done so because my thinking was that most likely the user wrote this code for a reason and most likely the solution isn’t to just delete it. If you took that view to the extreme, for every function that has a compilation error, you could offer to just delete the function body. 

So: I think it’s perfectly consistent to produce a ` unexpected code 'get,' in modifier` diagnostic without a Fix-It. 

But I’m open to have a discussion about this decision, so far it’s just been what I thought best and my thoughts above are the only opinion that has influenced the decision so far.

kimdv: I see a downside, that the missing `set` will automatically have a `Insert 'set'` Fix-It. 
This will result `private(get)` becomes `private(setget)`. Not sure what one is not better as we then produce more invalid code

ahoppen: Does it though? In case `set` is missing, we will still enter this branch because `detail.detail.presence == .missing` is `true`. Also, in the `private(get)` test case, the Fix-It is `replace 'get' with 'set'` and the fixed source is `private(set)`.

kimdv: Fixed it now. 
Can you please review again @ahoppen ?

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -1525,6 +1525,47 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
+        // If there is no unexpected tokens it means we miss a paren or set keyword.
+        // So we just skip the handling here
+        guard let firstUnexpected = unexpectedTokens.first else {
+          continue
+        }
+
+        let fixItMessage: ParserFixIt
+
+        if detail.detail.presence == .missing {
+          fixItMessage = ReplaceTokensFixIt(replaceTokens: [firstUnexpected], replacements: [detail.detail])
```
ahoppen, line 1548: Just one minor comment: We should be able to replace all `unexpectedTokens` not jus the `firstUnexpected`

```suggestion
          fixItMessage = ReplaceTokensFixIt(replaceTokens: unexpectedTokens, replacements: [detail.detail])
```

ahoppen (approved):

kimdv:
@swift-ci please test

PR #1644, opened 2023-05-08
support parsing `discard` in place of `_forget`

kavon:
SE-390 concluded with choosing the keyword `discard` rather than `forget`. This commit begins the process of deprecating the old keyword on the swift-syntax side.

rdar://108859077

### DIFF
--- a/Sources/SwiftParser/Statements.swift
+++ b/Sources/SwiftParser/Statements.swift
@@ -55,7 +55,7 @@ extension Parser {
   ///     statement → control-transfer-statement ';'?
   ///     statement → defer-statement ';'?
   ///     statement → do-statement ';'?
-  ///     statement → forget-statement ';'?
+  ///     statement → discard-statement ';'?
   ///
   ///     loop-statement → for-in-statement
   ///     loop-statement → while-statement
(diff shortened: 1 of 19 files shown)


kavon:
@swift-ci please test

Sources/SwiftParser/Statements.swift
```diff
@@ -124,8 +124,10 @@ extension Parser {
       return label(self.parseContinueStatement(continueHandle: handle), with: optLabel)
     case (.fallthroughKeyword, let handle)?:
       return label(self.parseFallthroughStatement(fallthroughHandle: handle), with: optLabel)
-    case (.forgetKeyword, let handle)?:
-      return label(self.parseForgetStatement(forgetHandle: handle), with: optLabel)
+    case (.forgetKeyword, let handle)?:  // NOTE: support for deprecated _forget
+      fallthrough
+    case (.discardKeyword, let handle)?:
+      return label(self.parseDiscardStatement(discardHandle: handle), with: optLabel)
```
ahoppen, line 130: The more swifty way to do write this would be 

```suggestion
     case (.forgetKeyword, let handle)?, 
         (.discardKeyword, let handle)?:
      return label(self.parseDiscardStatement(discardHandle: handle), with: optLabel)
```

Sources/SwiftParser/Statements.swift
```diff
@@ -1009,22 +1013,24 @@ extension Parser.Lookahead {
         // yield statement of some singular expression.
         return !self.peek().isAtStartOfLine
       }
-    case .forgetKeyword?:
+    case .forgetKeyword?:  // NOTE: support for deprecated _forget
+      fallthrough
+    case .discardKeyword?:
```
ahoppen, line 1018: Same here. `fallthrough` is discouraged in Swift

```suggestion
    case .discardKeyword?, .forgetKeyword?:  // NOTE: support for deprecated _forget
```

ahoppen (commented):
I’ve got two minor code style comments. Could you address them in a follow-up PR?

PR #1646, opened 2023-05-09
Migrate generics to opaque parameter types where possible

ahoppen:
Now that we no longer need to support Swift <5.7, we can use opaque parameter types.


ahoppen:
@swift-ci Please test

kimdv (approved):

Issue #1395, opened 2023-03-09
Macro expansions allow a space (or even newline) between `#` and macro name, as well as `@` and attribute name


ahoppen:
### Issue Kind

Other

### Source Code

```swift
# myMacroName
```

```swift
#
myMacroName
```

```swift
@ MyAttribute
func foo() {}
```

```swift
@
MyAttribute 
func foo() {}
```

### Description

The above examples parse as macro expansion or attribute but shouldn’t be valid. We need to check that the `#` or `@` doesn’t contain trailing trivia and the identifier doesn’t contain leading trivia to parse a macro expansion/attribute.

ahoppen:
rdar://106464071

TiagoMaiaL:
@ahoppen Is it possible to assign this issue to me? I'm already working on it on my free time.

Matejkob:
Is it steel an issue? 
![Uploading Screenshot 2024-04-05 at 1.01.48 PM.png…]()

ahoppen:
No, this was fixed by https://github.com/apple/swift-syntax/pull/2466. Thanks for noticing 👍🏽

PR #1650, opened 2023-05-10
[Parser] Attributes on MacroExpansionDeclSyntax

rintaro:
* Add `attributes` and `modifiers` to `MacroExpansionDeclSyntax`
* Diagnose whitespaces between `#` and the macro name
* Attach attributes to `MacroExpansionDeclSyntax`

rdar://107386648


Sources/SwiftParser/Declarations.swift
```diff
@@ -2023,9 +2038,29 @@ extension Parser {
...
-    let (unexpectedBeforeMacro, macro) = self.expectIdentifier()
+  mutating func parseMacroExpansionDeclaration(
+    _ attrs: DeclAttributes,
+    _ handle: RecoveryConsumptionHandle
+  ) -> RawMacroExpansionDeclSyntax {
+
+    let (unexpectedBeforePound, poundKeyword) = self.eat(handle)
+    // Don't allow space between '#' and the macro name.
+    if poundKeyword.trailingTriviaByteLength != 0 || self.currentToken.leadingTriviaByteLength != 0 {
+      return RawMacroExpansionDeclSyntax(
```
rintaro, line 2049: The intent here is that parse `#` followed by whitespaces as `#<missing identifier>`, and let the caller do the recovery.

Sources/SwiftParser/Expressions.swift
```diff
@@ -1394,8 +1394,21 @@ extension Parser {
...
+        argumentList: .init(elements: [], arena: self.arena),
+        rightParen: nil,
+        trailingClosure: nil,
+        additionalTrailingClosures: nil,
+        arena: self.arena
+      )
+    }
     let poundKeyword = self.consumeAnyToken()
-    let (unexpectedBeforeMacro, macro) = self.expectIdentifier()
+    let (unexpectedBeforeMacro, macro) = self.expectIdentifier(keywordRecovery: true)
```
rintaro, line 1411: e.g. `#class` is invalid but should be recoverable.

rintaro:
@swift-ci Please test

Tests/SwiftParserTest/translated/IfconfigExprTests.swift
```diff
@@ -500,26 +500,49 @@ final class IfconfigExprTests: XCTestCase {
...
+      """,
+      diagnostics: [
+        DiagnosticSpec(
+          message: "expected identifier in macro expansion",
+          fixIts: ["insert identifier"]
+        )
+      ],
+      fixedSource: """
+        #if MY_FLAG
+        #<#identifier#>elif
```
bnbarham, line 543: Lost the space here 🤔?

Tests/SwiftParserTest/DeclarationTests.swift
```diff
@@ -1427,6 +1427,125 @@ final class DeclarationTests: XCTestCase {
     )
   }
 
+  func testAttributedMacroExpansionDeclaration() {
```
bnbarham, line 1430: Could add a test when split by a newline rather than space? And also one for with both attributes and modifiers?

Sources/SwiftParser/Expressions.swift
bnbarham: Could you add a test case for `#<keyword>`?

bnbarham:
@TiagoMaiaL note that this fixes part of https://github.com/apple/swift-syntax/issues/1395 (ie. whitespace between `#` and the identifier, but not `@`).

Tests/SwiftParserTest/translated/IfconfigExprTests.swift
rintaro: Yeah.. no idea what happened..

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

Sources/SwiftParser/Declarations.swift
```diff
@@ -29,18 +29,25 @@ extension DeclarationModifier {
 }
 
 extension TokenConsumer {
+  mutating func atStartOfFreestandingMacroExpansion() -> Bool {
+    if !self.at(.pound) {
+      return false
+    }
+    if self.peek().rawTokenKind != .identifier && !self.peek().isLexerClassifiedKeyword {
```
ahoppen, line 36: I would like to minimize the use of `isLexerClassifiedKeyword` in the hope that one day we can get rid of it. I think `!self.peek().rawTokenKind != .keyword` should work equally well here because you don’t care about all the pound keywords and the only lexemes that have a `RawTokenKind` of `keyword` are the lever classified keywords.

Sources/SwiftParser/Declarations.swift
```diff
@@ -29,18 +29,25 @@ extension DeclarationModifier {
...
+      return false
+    }
+    if self.currentToken.trailingTriviaByteLength != 0 || self.peek().leadingTriviaByteLength != 0 {
```
ahoppen, line 39: Could you add a comment:

```swift
// The `#` and the macro’s name must not be separated by a space
```

Sources/SwiftParser/Declarations.swift
ahoppen: I think we should still continue macro attribute parsing if the identifier is at the same line as the `#`. I would expect that to yield better recovery.

Sources/SwiftParser/Expressions.swift
```diff
@@ -1394,8 +1394,21 @@ extension Parser {
     pattern: PatternContext,
     flavor: ExprFlavor
   ) -> RawMacroExpansionExprSyntax {
+    if !atStartOfFreestandingMacroExpansion() {
+      return RawMacroExpansionExprSyntax(
+        poundToken: self.consumeAnyToken(),
```
ahoppen, line 1399: Could you change this to call `self.expect(.pound)`? `consumeAnyToken` have historically always been sources of bugs where we could consume a token for `poundToken` that wasn’t actually a `#`.

Sources/SwiftParser/TokenSpecSet.swift
```diff
@@ -396,20 +399,17 @@ enum OperatorLike: TokenSpecSet {
 
 enum PoundDeclarationStart: TokenSpecSet {
```
ahoppen, line 400: Do we even need `PoundDeclarationStart` now if it only contains a single case?

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxDeclNodes.swift
```diff
@@ -3772,7 +3784,7 @@ public struct MacroExpansionDeclSyntax: DeclSyntaxProtocol, SyntaxHashable {
     self.init(data)
   }
   
-  public var unexpectedBeforePoundToken: UnexpectedNodesSyntax? {
+  public var unexpectedBeforeAttributes: UnexpectedNodesSyntax? {
```
ahoppen, line 3787: Note to self: If we want to maintain source stability, we would need to add a deprecated compatibility accessor for this property. I think it’s fine for this PR though.

Tests/SwiftParserTest/DeclarationTests.swift
```diff
@@ -1427,6 +1427,208 @@ final class DeclarationTests: XCTestCase {
...
+    assertParse(
```
ahoppen, line 1431: Could you add separate `assertParse` calls for these three cases? It makes it a lot easier to understand and debug the test cases IMO.

Same for all the other test cases, could you only have a single macro expansion in each `assertParse` function? That way the `substructure` can also be reduced to the `MacroExpansionDeclSyntax` node.

Tests/SwiftParserTest/translated/IfconfigExprTests.swift
ahoppen: To start, this gets parsed as follows

```
SourceFileSyntax
IfConfigDeclSyntax
├─clauses: IfConfigClauseListSyntax
│ ╰─[0]: IfConfigClauseSyntax
│   ├─poundKeyword: poundIfKeyword
│   ├─condition: IdentifierExprSyntax
│   │ ╰─identifier: identifier("MY_FLAG")
│   ╰─elements: CodeBlockItemListSyntax
│     ├─[0]: CodeBlockItemSyntax
│     │ ├─item: MacroExpansionExprSyntax
│     │ │ ├─poundToken: pound
│     │ │ ├─macro: identifier("") MISSING
│     │ │ ╰─argumentList: TupleExprElementListSyntax
│     │ ╰─semicolon: semicolon MISSING
│     ╰─[1]: CodeBlockItemSyntax
│       ╰─item: IdentifierExprSyntax
│         ╰─identifier: identifier("elif")
╰─poundEndif: poundEndifKeyword
```

When inserting the missing identifier, we see that `#` and identifiers don’t need to be separated by a space, so we remove the space. This is really useful for example if you have `someCall(1 2)` where a comma is missing. In this case you want to insert it after immediate after the `1` and not the space.

So now to the next question: Why don’t we add a space after `<#identifier#>`. For this, we look at the fixed up view of the tree and discover that the missing identifier is followed by a semicolon in that view. And identifier and semicolons don’t need to be separated by spaces. So no space that needs to be inserted here. We just suppress the missing semicolon diagnostic because the code item already has an error. Because most of the time the missing semicolon diagnostic is actually really misleading.

I think fix here would be to improve how we recover here in general so that `elif` doesn’t end up in a separate `CodeBlockItem` anymore.

Sources/SwiftParser/Declarations.swift
rintaro: What would you expect as the parsed tree. How do we embed "hasError"?

ahoppen: For example if you have `# elif`, we should have `unexpectedBeforeMacroName = ' elif'` and `macroName` should be a missing identifier with text `elif`. That’s also what we do when parsing `a. b` (where whitespace around the `.` is not balanced).

rintaro: Thanks. Will do.

ahoppen (approved):
@rintaro and I chatted offline that he will address the comments in a follow-up PR

ahoppen:
Addressing the review comments in https://github.com/apple/swift-syntax/pull/1665.

PR #1654, opened 2023-05-12
[Macros] Introduce SwiftSyntaxMacroExpansion module

rintaro:
This module is a unified macro expansion logic shared between 'swift' ASTGen and SwiftCompilerPluginMessageHandling


rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

rintaro:
https://github.com/apple/swift/pull/65882
@swift-ci Please test

bnbarham (approved):

Package.swift
```diff
@@ -178,6 +179,12 @@ let package = Package(
       exclude: ["CMakeLists.txt"]
     ),
 
+    .target(
+      name: "SwiftSyntaxMacroExpansion",
+      dependencies: ["SwiftSyntax", "SwiftSyntaxMacros"],
+      exclude: ["CMakeLists.txt"]
+    ),
+
```
ahoppen, line 187: Since this is a new target, could you add a new `MARK` section for it? Each `MARK` section only contains the target + its test target right now

Sources/SwiftSyntaxMacroExpansion/MacroExpansion.swift
```diff
@@ -0,0 +1,203 @@
...
+
+/// Expand `@attached(XXX)` macros.
+public func expandAttachedMacro<Context: MacroExpansionContext>(
+  definition: Macro.Type,
+  macroRole: MacroRole,
+  attributeNode: AttributeSyntax,
+  declarationNode: DeclSyntax,
+  parentDeclNode: DeclSyntax?,
+  in context: Context
+) -> [String]? {
```
ahoppen, line 67: Would it maybe be cleaner to make this method `throw` instead of returning an optional? I think requiring callers to add the error message to the context’s diagnostics isn’t too much to ask.

Sources/SwiftSyntaxMacroExpansion/MacroExpansion.swift
```diff
@@ -0,0 +1,203 @@
...
+      return expandedSyntax.formattedExpansion(definition.formatMode)
+    }
+    return try _openExistential(node, do: _expand)
+  } catch {
+    context.addDiagnostics(from: error, node: node)
+    return nil
+  }
+}
+
+/// Expand `@attached(XXX)` macros.
```
ahoppen, line 59: It would be nice to also document all the parameters.

Sources/SwiftSyntaxMacroExpansion/MacroExpansion.swift
```diff
@@ -0,0 +1,203 @@
...
+      return accessors.map {
+        $0.formattedExpansion(definition.formatMode)
+      }
+
+    case (let attachedMacro as MemberAttributeMacro.Type, .memberAttribute):
+      guard
+        let parentDeclGroup = parentDeclNode?.asProtocol(DeclGroupSyntax.self)
+      else {
+        // Compiler error: 'parentDecl' is mandatory for MemberAttributeMacro.
+        throw MacroExpansionError(description: "parent decl group is nil")
```
ahoppen, line 85: Any reason why you went to string-based error messages instead of enum cases. I personally prefer enums cases because it doesn’t bury the message somewhere in the source code and enum’s you can actually check programmatically if you want.

Sources/SwiftSyntaxMacroExpansion/MacroExpansion.swift
rintaro: I think there's no reason _not_ to catch exceptions, turn it to a diagnostic and emit it to the context in this function. The whole point of this PR is to minimize the duplicated logic. Why do you think making it `throws` is cleaner?

Package.swift
rintaro: Huh, I didn't realize that, thank you, will do.

Sources/SwiftSyntaxMacroExpansion/MacroExpansion.swift
ahoppen: I think I just have a disliking of making `nil` indicate the error case because it doesn’t allow the client to check what went wrong. Since this is `public` anyone could call it to try expanding a macro node and for their use case, it might be useful to know what went wrong, even though for our use cases it doesn’t matter.

That being said, I think I’m fine with the current approach if we add a comment saying that upon failure (aka. if this returns `nil`) a diagnostic is guaranteed to be added to `context`.

PR #1656, opened 2023-05-12
Support for isolated and async deinit

nickolas-pohilets:
See also https://github.com/apple/swift/pull/60057

### DIFF
--- a/Sources/SwiftParser/Declarations.swift
+++ b/Sources/SwiftParser/Declarations.swift
@@ -1045,11 +1045,36 @@ extension Parser {
     _ handle: RecoveryConsumptionHandle
   ) -> RawDeinitializerDeclSyntax {
     let (unexpectedBeforeDeinitKeyword, deinitKeyword) = self.eat(handle)
+
     var unexpectedNameAndSignature: [RawSyntax?] = []
-    unexpectedNameAndSignature.append(self.consume(if: TokenSpec(.identifier, allowAtStartOfLine: false)).map(RawSyntax.init))
-    if self.at(TokenSpec(.leftParen, allowAtStartOfLine: false)) {
-      unexpectedNameAndSignature.append(RawSyntax(parseFunctionSignature()))
+
+    // async is a contextual keyword
+    // must be parsed before attempting to parse identifier
+    var effectSpecifiers = parseDeinitEffectSpecifiers()
+
+    if effectSpecifiers == nil {
+      if let identifier = self.consume(if: TokenSpec(.identifier, allowAtStartOfLine: false)).map(RawSyntax.init) {
+        unexpectedNameAndSignature.append(identifier)
+      }
+      effectSpecifiers = parseDeinitEffectSpecifiers()
     }
+    if effectSpecifiers == nil && self.at(TokenSpec(.leftParen, allowAtStartOfLine: false)) {
+      let input = parseParameterClause(RawParameterClauseSyntax.self) { parser in
+        parser.parseFunctionParameter()
+      }
+      unexpectedNameAndSignature.append(RawSyntax(input))
+
+      effectSpecifiers = parseDeinitEffectSpecifiers()
+    }
+
+    var unexpectedAfterAsync: [RawSyntax?] = []
+    /// Only allow recovery to the arrow with exprKeyword precedence so we only
+    /// skip over misplaced identifiers and don't e.g. recover to an arrow in a 'where' clause.
+    if self.canRecoverTo(TokenSpec(.arrow, recoveryPrecedence: .exprKeyword)) != nil {
+      let output = self.parseFunctionReturnClause(effectSpecifiers: &effectSpecifiers, allowNamedOpaqueResultType: true)
+      unexpectedAfterAsync.append(RawSyntax(output))
+    }
+
     let items = self.parseOptionalCodeBlock()
     return RawDeinitializerDeclSyntax(
       attributes: attrs.attributes,
@@ -1057,6 +1082,8 @@ extension Parser {
       unexpectedBeforeDeinitKeyword,
       deinitKeyword: deinitKeyword,
       RawUnexpectedNodesSyntax(unexpectedNameAndSignature, arena: self.arena),
+      effectSpecifiers: effectSpecifiers,
+      RawUnexpectedNodesSyntax(unexpectedAfterAsync, arena: arena),
       body: items,
       arena: self.arena
     )
@@ -1065,7 +1092,9 @@ extension Parser {
 
 extension Parser {
   /// If a `throws` keyword appears right in front of the `arrow`, it is returned as `misplacedThrowsKeyword` so it can be synthesized in front of the arrow.
-  mutating func parseFunctionReturnClause(effectSpecifiers: inout (some RawEffectSpecifiersTrait)?, allowNamedOpaqueResultType: Bool) -> RawReturnClauseSyntax {
+  mutating func parseFunctionReturnClause(effectSpecifiers: inout (some RawMisplacedEffectSpecifiersTrait)?, allowNamedOpaqueResultType: Bool)
+    -> RawReturnClauseSyntax
+  {
     let (unexpectedBeforeArrow, arrow) = self.expect(.arrow)
     let unexpectedBeforeReturnType = self.parseMisplacedEffectSpecifiers(&effectSpecifiers)
     let result: RawTypeSyntax
@@ -1149,7 +1178,7 @@ extension Parser {
 
     /// Only allow recovery to the arrow with exprKeyword precedence so we only
     /// skip over misplaced identifiers and don't e.g. recover to an arrow in a 'where' clause.
-    if self.at(.arrow) || self.canRecoverTo(TokenSpec(.arrow, recoveryPrecedence: .exprKeyword)) != nil {
+    if self.canRecoverTo(TokenSpec(.arrow, recoveryPrecedence: .exprKeyword)) != nil {
       output = self.parseFunctionReturnClause(effectSpecifiers: &effectSpecifiers, allowNamedOpaqueResultType: true)
     } else {
       output = nil
--- a/Sources/SwiftParser/Specifiers.swift
+++ b/Sources/SwiftParser/Specifiers.swift
@@ -134,7 +134,7 @@ public enum EffectSpecifier: TokenSpecSet {
 
 /// Raw syntax nodes don't have traits (because usually we don't need them).
 /// Specify the effect specifiers trait manually as a one off.
-protocol RawEffectSpecifiersTrait {
+protocol RawMisplacedEffectSpecifiersTrait {
   /// The token kinds that should be consumed as misspelled `asyncSpecifier`.
   /// Should be a subset of ``AsyncEffectSpecifier``.
   associatedtype MisspelledAsyncTokenKinds: TokenSpecSet
@@ -151,10 +151,21 @@ protocol RawEffectSpecifiersTrait {
   /// Should be a subset of ``ThrowsEffectSpecifier``.
   associatedtype CorrectThrowsTokenKinds: TokenSpecSet
 
-  var unexpectedBeforeAsyncSpecifier: RawUnexpectedNodesSyntax? { get }
   var asyncSpecifier: RawTokenSyntax? { get }
-  var unexpectedBetweenAsyncSpecifierAndThrowsSpecifier: RawUnexpectedNodesSyntax? { get }
   var throwsSpecifier: RawTokenSyntax? { get }
+
+  init(
+    asyncSpecifier: RawTokenSyntax?,
+    throwsSpecifier: RawTokenSyntax?,
+    arena: __shared SyntaxArena
+  )
+
+  func withMisplaced(async misplacedAsyncKeyword: RawTokenSyntax?, throws misplacedThrowsKeyword: RawTokenSyntax?, arena: __shared SyntaxArena) -> Self
+}
+
+protocol RawEffectSpecifiersTrait: RawMisplacedEffectSpecifiersTrait {
+  var unexpectedBeforeAsyncSpecifier: RawUnexpectedNodesSyntax? { get }
+  var unexpectedBetweenAsyncSpecifierAndThrowsSpecifier: RawUnexpectedNodesSyntax? { get }
   var unexpectedAfterThrowsSpecifier: RawUnexpectedNodesSyntax? { get }
 
   init(
@@ -168,6 +179,21 @@ protocol RawEffectSpecifiersTrait {
 }
 
 extension RawEffectSpecifiersTrait {
+  init(
+    asyncSpecifier: RawTokenSyntax?,
+    throwsSpecifier: RawTokenSyntax?,
+    arena: __shared SyntaxArena
+  ) {
+    self.init(
+      nil,
+      asyncSpecifier: asyncSpecifier,
+      nil,
+      throwsSpecifier: throwsSpecifier,
+      nil,
+      arena: arena
+    )
+  }
+
   func withMisplaced(async misplacedAsyncKeyword: RawTokenSyntax?, throws misplacedThrowsKeyword: RawTokenSyntax?, arena: __shared SyntaxArena) -> Self {
     return Self.init(
       self.unexpectedBeforeAsyncSpecifier,
@@ -415,7 +441,122 @@ extension RawAccessorEffectSpecifiersSyntax: RawEffectSpecifiersTrait {
       }
     }
   }
+}
 
+extension RawDeinitEffectSpecifiersSyntax: RawMisplacedEffectSpecifiersTrait {
+  enum MisspelledAsyncTokenKinds: TokenSpecSet {
+    case await
+    case reasync
+
+    init?(lexeme: Lexer.Lexeme) {
+      switch PrepareForKeywordMatch(lexeme) {
+      case TokenSpec(.await, allowAtStartOfLine: false): self = .await
+      case TokenSpec(.reasync): self = .reasync
+      default: return nil
+      }
+    }
+
+    var spec: TokenSpec {
+      switch self {
+      case .await: return TokenSpec(.await, allowAtStartOfLine: false)
+      case .reasync: return .keyword(.reasync)
+      }
+    }
+  }
+
+  enum CorrectAsyncTokenKinds: TokenSpecSet {
+    case async
+
+    init?(lexeme: Lexer.Lexeme) {
+      switch PrepareForKeywordMatch(lexeme) {
+      case TokenSpec(.async): self = .async
+      default: return nil
+      }
+    }
+
+    var spec: TokenSpec {
+      switch self {
+      case .async: return .keyword(.async)
+      }
+    }
+  }
+
+  enum MisspelledThrowsTokenKinds: TokenSpecSet {
+    case `rethrows`
+    case `try`
+    case `throw`
+    case `throws`
+
+    init?(lexeme: Lexer.Lexeme) {
+      switch PrepareForKeywordMatch(lexeme) {
+      case TokenSpec(.rethrows): self = .rethrows
+      case TokenSpec(.try, allowAtStartOfLine: false): self = .try
+      case TokenSpec(.throw, allowAtStartOfLine: false): self = .throw
+      case TokenSpec(.throws): self = .throws
+      default: return nil
+      }
+    }
+
+    var spec: TokenSpec {
+      switch self {
+      case .rethrows: return .keyword(.rethrows)
+      case .try: return TokenSpec(.try, allowAtStartOfLine: false)
+      case .throw: return TokenSpec(.throw, allowAtStartOfLine: false)
+      case .throws: return .keyword(.throws)
+      }
+    }
+  }
+
+  enum CorrectThrowsTokenKinds: TokenSpecSet {
+    // Uninhabited
+
+    init?(lexeme: Lexer.Lexeme) {
+      return nil
+    }
+
+    var spec: TokenSpec {
+      switch self {
+      }
+    }
+  }
+
+  var throwsSpecifier: RawTokenSyntax? { nil }
+
+  init(
+    asyncSpecifier: RawTokenSyntax?,
+    throwsSpecifier: RawTokenSyntax?,
+    arena: __shared SwiftSyntax.SyntaxArena
+  ) {
+    // `throwsSpecifier` should never be present because `parseMisplacedEffectSpecifiers()` only creates missing tokens
+    // and `CorrectThrowsTokenKinds` is an empty `TokenSpecSet`.
+    //
+    // We don't want to insert missing `throws` for deinit case,
+    // so if `parseMisplacedEffectSpecifiers()` creates one it will be discarded here.
+    precondition(throwsSpecifier?.isMissing ?? true)
+    self.init(
+      nil,
+      asyncSpecifier: asyncSpecifier,
+      nil,
+      arena: arena
+    )
+  }
+
+  func withMisplaced(async misplacedAsyncKeyword: RawTokenSyntax?, throws misplacedThrowsKeyword: RawTokenSyntax?, arena: SyntaxArena)
+    -> RawDeinitEffectSpecifiersSyntax
+  {
+    // `throwsSpecifier` should never be present because `parseMisplacedEffectSpecifiers()` only creates missing tokens
+    // and `CorrectThrowsTokenKinds` is an empty `TokenSpecSet`.
+    //
+    // We don't want to insert missing `throws` for deinit case,
+    // so if `parseMisplacedEffectSpecifiers()` creates one it will be discarded here.
+    precondition(throwsSpecifier?.isMissing ?? true)
+    return Self.init(
+      self.unexpectedBeforeAsyncSpecifier,
+      asyncSpecifier: self.asyncSpecifier ?? misplacedAsyncKeyword,
+      self.unexpectedAfterAsyncSpecifier,
+      arena: arena
+    )
+  }
 }
 
 extension TokenConsumer {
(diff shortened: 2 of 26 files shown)


Sources/SwiftParser/Declarations.swift
```diff
@@ -1120,10 +1120,39 @@ extension Parser {
...
+          asyncSpecifier: effects.asyncSpecifier,
+          effects.unexpectedBetweenAsyncSpecifierAndThrowsSpecifier,
+          throwsSpecifier: effects.throwsSpecifier,
+          effects.unexpectedAfterThrowsSpecifier,
+          arena: self.arena
+        )
+      }
+      unexpectedAfterAsync.append(contentsOf: sig.unexpectedBetweenEffectSpecifiersAndOutput?.elements ?? [])
+      if let output = sig.output {
+          unexpectedAfterAsync.append(RawSyntax(output))
```
ahoppen, line 1154: Incorrect  indentation. Could you run `./format.py`?

Sources/SwiftParser/Specifiers.swift
```diff
@@ -415,7 +415,120 @@ extension RawAccessorEffectSpecifiersSyntax: RawEffectSpecifiersTrait {
...
+      case .reasync: return .keyword(.reasync)
+      }
+    }
+  }
+  
+  enum MisspelledThrowsTokenKinds: TokenSpecSet {
+    case `try`
+    case `throw`
+    case `rethrows`
+    case `throws`
```
ahoppen, line 445: Could you order them alphabetically? It just makes it easier to read IMO. Same in the switches below

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -168,12 +168,9 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
   ) {
-    guard let incorrectContainer = unexpected,
-      let misplacedTokens = incorrectContainer.onlyPresentTokens(satisfying: unexpectedTokenCondition)
-    else {
-      // If there are no unexpected nodes or the unexpected contain multiple tokens, don't emit a diagnostic.
-      return
-    }
+    guard let incorrectContainer = unexpected else { return }
+    let misplacedTokens = incorrectContainer.presentTokens(satisfying: unexpectedTokenCondition)
+    if misplacedTokens.isEmpty { return }
```
ahoppen, line 173: Out of curiosity: What motivated this change?

Sources/SwiftParser/Specifiers.swift
```diff
@@ -415,7 +415,120 @@ extension RawAccessorEffectSpecifiersSyntax: RawEffectSpecifiersTrait {
       }
     }
   }
+}
+
+extension RawDeinitEffectSpecifiersSyntax {
+  enum MisspelledAsyncSpecifiers: TokenSpecSet {
```
ahoppen, line 421: Oh, I just noticed that `MisspelledAsyncSpecifiers` is inconsistently named with `MisspelledThrowsTokenKinds`. I’m fixing that in https://github.com/apple/swift-syntax/pull/1767. You’ll probably need to rebase once that PR is merged.

Sources/SwiftParser/Specifiers.swift
```diff
@@ -415,7 +415,120 @@ extension RawAccessorEffectSpecifiersSyntax: RawEffectSpecifiersTrait {
...
+  }
+  
+  public init(
+    _ unexpectedBeforeAsyncSpecifier: RawUnexpectedNodesSyntax? = nil,
+    asyncSpecifier: RawTokenSyntax?,
+    _ unexpectedBetweenAsyncSpecifierAndThrowsSpecifier: RawUnexpectedNodesSyntax? = nil,
+    throwsSpecifier: RawTokenSyntax?,
+    _ unexpectedAfterThrowsSpecifier: RawUnexpectedNodesSyntax? = nil,
+    arena: __shared SyntaxArena
+  ) {
```
ahoppen, line 474: Also, is it really worth having this entire dance of re-arranging the tokens? It seems overly complicated to me and possibly doesn’t provide too much value. AFAICT the only place where it’s used is when parsing a function signature after `deinit` and honestly, if you put a function signature after `deinit`, put a `throws` into it and then also do that in the wrong order, I’m not sure if that’s really a recovery case we need to add a >100-line recovery path for. 

If we do want to keep it, here are some more specific comments

I think it would be worth adding a doc-comment here saying that `throwsSpecifier` will be stored as an unexpected token because `RawDeinitEffectSpecifiersSyntax` doesn’t actually have an  throws specifier.

Could you also mark it as `@_spi(RawSyntax)`?

Sources/SwiftParser/Specifiers.swift
```diff
@@ -415,7 +415,120 @@ extension RawAccessorEffectSpecifiersSyntax: RawEffectSpecifiersTrait {
...
+    // Missing async keyword was created, meaning that throws and async are swapped
+    if asyncSpecifier?.isMissing ?? false {
+      var unexpectedBeforeAsync: [RawSyntax?] = []
+      var asyncKeyword: RawTokenSyntax? = asyncSpecifier?.raw.as(RawTokenSyntax.self)
```
ahoppen, line 478: Is `?.raw.as(RawTokenSyntax.self)` needed for anything? AFAICT it’s a no-op.

Sources/SwiftParser/Specifiers.swift
```diff
@@ -520,6 +633,65 @@ extension Parser {
...
+  mutating func parseDeinitEffectSpecifiers() -> RawDeinitEffectSpecifiersSyntax? {
+    // Note that parseEffectSpecifiers() consumes deinit name as unexpected token
+    // But we want it to be handled on the higher level.
+    // So we parseEffectSpecifiers() is not reused here.
+    
+    var unexpectedBeforeAsync: [RawSyntax?] = []
+    var asyncKeyword: RawTokenSyntax?
+    var unexpectedAfterAsync: [RawSyntax?] = []
+    
+    while true {
```
ahoppen, line 646: This also seems overly complicated. Can’t we just eat consume any misplaced throws specifiers into `unexpectedBeforeAsync` and then just do `self.consume(if: .async)`?

Alternatively, could we keep your hand-written initializer for `RawDeinitEffectSpecifiersSyntax`, make it conform to `RawEffectSpecifiersTrait` and use `parseEffectSpecifiers`?

nickolas-pohilets: I was trying to follow the logic of `parseEffectSpecifiers()` and handle the same recovery cases, including misspelled and duplicated `async`. If I understood you correctly, this simplification would not parse such cases, or would it?

I've tried the latter, but it consumes deinit name into `unexpectedBeforeAsyncSpecifier`.

Sources/SwiftParser/Specifiers.swift
nickolas-pohilets: I think this could be simplified if I don't reuse entire `parseFunctionSignature()`, but only smaller functions, like `parseParameterClause()`. I can pass dummy `inout RawFunctionEffectSpecifiersSyntax?` to `parseFunctionReturnClause()` and discard the value. Or don't attempt to parse output part at all. WDYT?

Sources/SwiftParser/Declarations.swift
nickolas-pohilets: Sure. Is it ok to run it once as a separate commit, or should I patch every commit?

Sources/SwiftParser/Specifiers.swift
nickolas-pohilets: Done

Sources/SwiftParser/Specifiers.swift
nickolas-pohilets: Updated

Sources/SwiftParser/Specifiers.swift
nickolas-pohilets: I don't see any tests for deinit with output type, so I guess dropping support for it should be ok. I've pushed an update, please take a look.

Sources/SwiftParser/Specifiers.swift
nickolas-pohilets: I would then combine all the loops into one.

nickolas-pohilets: I've pushed an update, please take a look

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -788,19 +785,78 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
+    return .visitChildren
+  }
+
+  public override func visit(_ node: DeinitEffectSpecifiersSyntax) -> SyntaxVisitorContinueKind {
+    let throwsTokens: [TokenKind] = [
+      .keyword(.throws),
+      .keyword(.rethrows),
+      .keyword(.try),
+      .keyword(.throw),
+    ]
```
ahoppen, line 810: Can’t you use the `ThrowsEffectSpecifier` token spec set here instead of re-defining the list?

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -788,19 +785,78 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
+        .deinitCannotThrow,
+        fixIts: [
+          FixIt(message: RemoveNodesFixIt(unexpectedThrows), changes: .makeMissing(unexpectedThrows))
         ],
-        handledNodes: [signature.id]
+        handledNodes: unexpectedThrows.map(\.id)
       )
     }
 
+    let isOfSameKind = { AsyncEffectSpecifier(token: $0) != nil }
```
ahoppen, line 831: `isOfSameKind` is a little misleading here. I would just call it: `isAsyncEffectSpecifier` (and maybe also rename `asThrowingToken` -> `asThrowsEffectSpecifier` for consistency).

Sources/SwiftParser/Declarations.swift
```diff
@@ -1120,10 +1120,26 @@ extension Parser {
...
+    // must be parsed before attempting to parse identifier
+    var effectSpecifiers = parseDeinitEffectSpecifiers()
+
+    if effectSpecifiers == nil {
+      if let identifier = self.consume(if: TokenSpec(.identifier, allowAtStartOfLine: false)).map(RawSyntax.init) {
+        unexpectedNameAndSignature.append(identifier)
+      }
+      effectSpecifiers = parseDeinitEffectSpecifiers()
+    }
+    if effectSpecifiers == nil && self.at(.leftParen) && !self.currentToken.isAtStartOfLine {
```
ahoppen, line 1136: I know that you just copied this code but it can be slightly simplified:

```suggestion
    if effectSpecifiers == nil && self.at(TokenSpec(.leftParen, allowAtStartOfLine: false)) {
```

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
ahoppen: Ah, I see. Could you update the doc comment to match the new behavior?

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -788,19 +785,78 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
+        ],
+        handledNodes: [params.id]
+      )
+    }
+
+    return .visitChildren
+  }
+
+  public override func visit(_ node: DeinitEffectSpecifiersSyntax) -> SyntaxVisitorContinueKind {
+    let throwsTokens: [TokenKind] = [
```
ahoppen, line 805: You’re missing

```swift
if shouldSkip(node) {
  return .skipChildren
}
```

at the start of the `visit` function. Unfortunately, I don’t know of a good way to automatically add it.

Sources/SwiftParser/Declarations.swift
```diff
@@ -1120,10 +1120,26 @@ extension Parser {
...
+      let input = parseParameterClause(RawParameterClauseSyntax.self) { parser in
+        parser.parseFunctionParameter()
+      }
```
ahoppen, line 1139: If you’re changing this from `parseFunctionSignature` to `parseParameterClause`, I think we aren’t recovering the following as nicely anymore. Can’t we continue to use `parseFunctionSignature` here anymore? 

```swift
deinit() -> Int {}
```

And also: Could you add a test case for a `deinit` with a return value?

Tests/SwiftParserTest/translated/InitDeinitTests.swift
```diff
@@ -438,4 +438,361 @@ final class InitDeinitTests: XCTestCase {
...
+      fixedSource: """
+        class FooClassDeinitializerA {
+          deinit async {}
+        }
+        """
+    )
+  }
+
+  func testAsyncDeinit() {
+    // TODO: Investigate why this produces no diagnostics
```
ahoppen, line 789: This is expected for now. `async` is parsed as a modifier like `public` because you can have an `async var x: Int`.

ahoppen (commented):
This looks really good. I’ve got a few minor comments but all of them should be fairly minor.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
nickolas-pohilets: Done

Sources/SwiftParser/Declarations.swift
nickolas-pohilets: Applied

Tests/SwiftParserTest/translated/InitDeinitTests.swift
nickolas-pohilets: Updated comment

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
nickolas-pohilets: Done

Sources/SwiftParser/Declarations.swift
nickolas-pohilets: `parseFunctionSignature()` uses `parseFunctionEffectSpecifiers()` which we don't want. I don't think it is worth to abstract `parseFunctionSignature()` over `parseFunctionEffectSpecifiers()` vs `parseDeinitEffectSpecifiers()`. It would be simpler to use `parseParameterClause()` + `parseDeinitEffectSpecifiers()` + `parseFunctionReturnClause()` directly from `parseDeinitializerDeclaration()`.

As a bonus it would allow recovering from `deinit -> Int {}` which `parseFunctionSignature()` cannot.

Note that using `RawDeinitEffectSpecifiersSyntax` with`parseFunctionReturnClause()` still requires conformance to  `RawEffectSpecifiersTrait` - that's still a bit ugly, but at least now without any logic.

nickolas-pohilets: Updated, please take a look

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
nickolas-pohilets: Updated. Please take a look

Sources/SwiftParser/Declarations.swift
```diff
@@ -1120,18 +1120,45 @@ extension Parser {
...
+      unexpectedNameAndSignature.append(RawSyntax(input))
+
+      effectSpecifiers = parseDeinitEffectSpecifiers()
     }
+
+    var unexpectedAfterAsync: [RawSyntax?] = []
+    /// Only allow recovery to the arrow with exprKeyword precedence so we only
+    /// skip over misplaced identifiers and don't e.g. recover to an arrow in a 'where' clause.
+    if self.at(.arrow) || self.canRecoverTo(TokenSpec(.arrow, recoveryPrecedence: .exprKeyword)) != nil {
```
ahoppen, line 1148: `self.at(.arrow)` implies `self.canRecoverTo(TokenSpec(.arrow, recoveryPrecedence: .exprKeyword)) != nil` so I think you can just remove `self.at(.arrow)`.

Sources/SwiftParser/Declarations.swift
ahoppen: Oh, nice. That looks good to me.

Sources/SwiftParser/Specifiers.swift
```diff
@@ -415,7 +415,108 @@ extension RawAccessorEffectSpecifiersSyntax: RawEffectSpecifiersTrait {
...
+      case .rethrows: return .keyword(.rethrows)
+      case .try: return TokenSpec(.try, allowAtStartOfLine: false)
+      case .throw: return TokenSpec(.throw, allowAtStartOfLine: false)
+      case .throws: return .keyword(.throws)
+      }
+    }
+  }
+
+  enum CorrectThrowsTokenKinds: TokenSpecSet {
+    // Uninhabited
```
ahoppen, line 486: We don’t have any precedence for empty `TokenSpecSet` and `canRecoveryTo` has a precondition that the spec set is not empty. 

Could you change 

https://github.com/apple/swift-syntax/blob/8a759449c646c72b9392b151236204a5fad03831/Sources/SwiftParser/Recovery.swift#L133

to 

```swift
if specSet.allCases.isEmpty {
  return nil
}
```

Sources/SwiftParser/Specifiers.swift
```diff
@@ -415,7 +415,108 @@ extension RawAccessorEffectSpecifiersSyntax: RawEffectSpecifiersTrait {
...
+  init(
+    _ unexpectedBeforeAsyncSpecifier: RawUnexpectedNodesSyntax?,
+    asyncSpecifier: RawTokenSyntax?,
+    _ unexpectedBetweenAsyncSpecifierAndThrowsSpecifier: RawUnexpectedNodesSyntax?,
+    throwsSpecifier: RawTokenSyntax?,
+    _ unexpectedAfterThrowsSpecifier: RawUnexpectedNodesSyntax?,
+    arena: __shared SwiftSyntax.SyntaxArena
+  ) {
+    // Inserted missing throws is discarded
+    assert(throwsSpecifier?.isMissing ?? true)
```
ahoppen, line 511: Could you change this to a `precondition`?

https://github.com/apple/swift-syntax/blob/8a759449c646c72b9392b151236204a5fad03831/Sources/SwiftSyntax/Assert.swift#L13-L22

Sources/SwiftParser/Specifiers.swift
```diff
@@ -415,7 +415,108 @@ extension RawAccessorEffectSpecifiersSyntax: RawEffectSpecifiersTrait {
...
+    assert(unexpectedAfterThrowsSpecifier == nil)
```
ahoppen, line 512: This worries me a little bit because I’m not entirely sure that we really always end up with this being `nil` but can’t come up with a counter example from the top of my head. I would feel more comfortable if we just combine it with `unexpectedBetweenAsyncSpecifierAndThrowsSpecifier`.

```swift
RawUnexpectedNodesSyntax(combining: unexpectedBetweenAsyncSpecifierAndThrowsSpecifier, unexpectedAfterThrowsSpecifier, arena: arena),
```

Sources/SwiftParser/Declarations.swift
nickolas-pohilets: I've copied this from `parseFunctionSignature()`. Should it be updated there as well? Could it be that this pre-check is a performance optimization? Any other reasons it might be needed there?

Sources/SwiftParser/Specifiers.swift
nickolas-pohilets: Done

Sources/SwiftParser/Specifiers.swift
nickolas-pohilets: Done

Sources/SwiftParser/Specifiers.swift
nickolas-pohilets: At this point we have only a single case (as noted in the extension comment), where we know that `unexpectedAfterThrowsSpecifier` cannot appear. Combining two groups of unexpected syntax looses information, which was affecting how `exchangeTokens()` behaves. Not sure if it still does after loosening the checks. But I would postpone combining or any other handling logic until other cases appear.

Sources/SwiftParser/Specifiers.swift
nickolas-pohilets: Actually, in the single use case where this conformance is used, `canRecoverTo()` is not used, only `at()` which seems to be perfectly fine with empty spec sets. So, I guess changes in `canRecoverTo()` are not needed. But I don't mind to do them just in case. @ahoppen, WDYT?

nickolas-pohilets:
Rebased after https://github.com/apple/swift-syntax/pull/1767

Sources/SwiftParser/Specifiers.swift
ahoppen: I think allowing empty `TokenSpecSet` makes total sense in `canRecoverTo` and I‘d prefer to be safe here instead of discovering a stupid crash about it later if we for some reason start to use `canRecoverTo` in the future.

Sources/SwiftParser/Declarations.swift
ahoppen: If you could also update the call in `parseFunctionSignature`, that would be great. I don’t see how `self.at(.arrow)` could help performance in any significant way.

Sources/SwiftParser/Specifiers.swift
ahoppen: Did you forget to push? Still appears to be an `assert` to me.

Sources/SwiftParser/Specifiers.swift
ahoppen: I’m mostly thinking that the assumption that `unexpectedAfterThrowsSpecifier` is empty might change in the future. And especially the recovery cases are not hit as frequently in CI but more infrequently while code is being edited, so we might not actually notice that the assertion starts being violated. Thus, I’d rather be safe here. 

And I don’t see how `RawUnexpectedNodesSyntax(combining:)` can change the layout of the node if `unexpectedAfterThrowsSpecifier` is `nil` because it just grabs the elements of `unexpectedBetweenAsyncSpecifierAndThrowsSpecifier` and puts them in a new `RawUnexpectedNodesSyntax`.

nickolas-pohilets: IMO, this entire conformance is hacky, and I would advise against introducing any new cases relying on it. I think even with `RawUnexpectedNodesSyntax(combining:)` implementation of the initialiser would be not robust enough to work in the general case. It is still not possible to pass `RawDeinitEffectSpecifiersSyntax` to `parseEffectSpecifiers()`.

Cleaner solution would be create a new protocol just for the `parseMisplacedEffectSpecifiers()`. I've pushed an updated. @ahoppen, WDYT?

Sources/SwiftParser/Specifiers.swift
```diff
@@ -415,7 +441,114 @@ extension RawAccessorEffectSpecifiersSyntax: RawEffectSpecifiersTrait {
...
+  }
+
+  var throwsSpecifier: RawTokenSyntax? { nil }
+
+  init(
+    asyncSpecifier: RawTokenSyntax?,
+    throwsSpecifier: RawTokenSyntax?,
+    arena: __shared SwiftSyntax.SyntaxArena
+  ) {
+    // Inserted missing throws is discarded
```
ahoppen, line 530: Just as I’m reading this. I would add the following comment
```
`throwsSpecifier` should never be present because `CorrectThrowsTokenKinds` is an empty `TokenSpecSet`.
```

Sources/SwiftParser/Declarations.swift
ahoppen: I think you only updated the existing call and forgot to update this one 😉

ahoppen (commented):
I really like the newly introduced protocol 😍 That’s a lot cleaner than everything we had before. Thanks for the idea.

I’ve got two really minor comments, otherwise this looks good to me now. Thanks for your great collaboration on this PR.

Sources/SwiftParser/Specifiers.swift
nickolas-pohilets: Not even because of that, but because `parseMisplacedEffectSpecifiers()` only creates missing tokens

ahoppen: Ah, I see. Thanks for clearing up the misunderstanding. I think it would be good to mention that as a comment.

Sources/SwiftParser/Declarations.swift
nickolas-pohilets: 🤦‍♂️

nickolas-pohilets:
Should this MR and https://github.com/apple/swift/pull/60057 be merged in any specific order? I was getting failing tests in the main repo without changes in this one, so I guess this MR needs to be merged first. Or neither can be merged until proposal is approved?

ahoppen (approved):
Nice. Thanks for your effort! One last request: Could you squash your commits? It just makes for a nicer commit history.

nickolas-pohilets:
@swift-ci, please test

nickolas-pohilets:
@swift-ci please test

nickolas-pohilets:
@ahoppen, could you please trigger CI? Looks like I don’t have rights. Or what am I doing wrong?

ahoppen:
Yes, you need push access to trigger CI for security reasons. Triggering it now.

@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

nickolas-pohilets:
CI failed because of this:
> FAIL: code-generated files committed to repository do not match generated ones. 
When I try to regenerate them, I get the following error:
```
CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/KeywordFile.swift:18:48: error: type 'TokenSyntax' has no member 'leftSquareToken'
let lookupTable = ArrayExprSyntax(leftSquare: .leftSquareToken(trailingTrivia: .newline)) {
                                              ~^~~~~~~~~~~~~~~
```
Reproduces on fresh `main` branch. @ahoppen, looks like `main` is broken.

nickolas-pohilets:
`git clean -fdx` helped. Could it be that `./build-script.py generate-source-code` incorrectly caches code generator binary?

I've pushed an update, let's re-try CI

ahoppen:
Yes, we’re currently seeing issues with CodeGeneration not updating correctly. We haven’t had the chance to 100% pinpoint the failure point. There was a discussion going on here: https://github.com/apple/swift-syntax/pull/1741#discussion_r1229816046

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

PR #1657, opened 2023-05-12
[Macros] Don't fold operators in unnecessary nodes

rintaro:
For attached macros, operator folding should only be done for the attribute syntax node, but not for the target declaration node and the parent decl node. This aligns with the compiler's ASTGen.


rintaro:
@swift-ci Please test

bnbarham (approved):

DougGregor (approved):

PR #1658, opened 2023-05-13
[Parser] Emit diagnostics for `@ AttributeName` and `@\nAttributeName`

TiagoMaiaL:
# Goal

To fix the second part of issue https://github.com/apple/swift-syntax/issues/1395. The first one (for unexpected whitespace in between the macro declaration), is being fixed by https://github.com/apple/swift-syntax/issues/1395.

## Notes

There are some `TODOs` I've placed to seek assistance from reviewers. Specifically, I don't know how I would synthesize a `RawTypeSyntax`, as I didn't find an easy way to access its text (without the leading trivia). Any help is appreciated. Thank you.

### DIFF
--- a/Sources/SwiftParser/Attributes.swift
+++ b/Sources/SwiftParser/Attributes.swift
@@ -166,8 +166,34 @@ extension Parser {
   }
 
   mutating func parseAttribute(argumentMode: AttributeArgumentMode, parseArguments: (inout Parser) -> RawAttributeSyntax.Argument) -> RawAttributeListSyntax.Element {
-    let (unexpectedBeforeAtSign, atSign) = self.expect(.atSign)
-    let attributeName = self.parseType()
+    var (unexpectedBeforeAtSign, atSign) = self.expect(.atSign)
+    var attributeName = self.parseType()
+    var unexpectedBetweenAtSignAndAttributeName: RawUnexpectedNodesSyntax?
+
+    if atSign.trailingTriviaByteLength != 0 {
+      unexpectedBeforeAtSign = RawUnexpectedNodesSyntax(
+        combining: unexpectedBeforeAtSign,
+        atSign,
+        arena: self.arena
+      )
+      atSign = RawTokenSyntax(
+        kind: .atSign,
+        text: atSign.tokenText,
+        leadingTriviaPieces: atSign.leadingTriviaPieces,
+        presence: .missing,
+        arena: self.arena
+      )
+    } else if attributeName.raw.leadingTriviaByteLength != 0 {
+      unexpectedBetweenAtSignAndAttributeName = RawUnexpectedNodesSyntax(attributeName)
+      // `withLeadingTrivia` only returns `nil` if there is no token in `attributeName`.
+      // But since `attributeName` has leadingTriviaLength != 0 there must be trivia and thus a token.
+      // So we can safely force-unwrap here.
+      attributeName = attributeName
+        .raw
+        .withLeadingTrivia([], arena: self.arena)!
+        .as(RawTypeSyntax.self)!
+    }
+
     let shouldParseArgument: Bool
     switch argumentMode {
     case .required:
@@ -185,6 +211,7 @@ extension Parser {
         RawAttributeSyntax(
           unexpectedBeforeAtSign,
           atSignToken: atSign,
+          unexpectedBetweenAtSignAndAttributeName,
           attributeName: attributeName,
           unexpectedBeforeLeftParen,
           leftParen: leftParen,
@@ -199,6 +226,7 @@ extension Parser {
         RawAttributeSyntax(
           unexpectedBeforeAtSign,
           atSignToken: atSign,
+          unexpectedBetweenAtSignAndAttributeName,
           attributeName: attributeName,
           leftParen: nil,
           argument: nil,
--- a/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
+++ b/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
@@ -457,7 +457,51 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
         ],
         handledNodes: [argument.id]
       )
-      return .visitChildren
+    }
+    if let unexpectedAtSign = node.unexpectedBeforeAtSignToken?.onlyToken(where: { $0.tokenKind == .atSign && !$0.trailingTrivia.isEmpty }),
+      node.atSignToken.presence == .missing
+    {
+      addDiagnostic(
+        unexpectedAtSign,
+        position: unexpectedAtSign.endPositionBeforeTrailingTrivia,
+        StaticParserError.invalidWhitespaceBetweenAttributeAtSignAndIdentifier,
+        fixIts: [
+          FixIt(
+            message: StaticParserFixIt.removeExtraneousWhitespace,
+            changes: [
+              .makeMissing(unexpectedAtSign, transferTrivia: false),
+              .makePresent(node.atSignToken),
+            ]
+          )
+        ],
+        handledNodes: [
+          node.id,
+          unexpectedAtSign.id,
+          node.atSignToken.id,
+        ]
+      )
+    } else if node.attributeName.isMissingAllTokens,
+      let unexpectedBetweenAtSignTokenAndAttributeName = node.unexpectedBetweenAtSignTokenAndAttributeName,
+      unexpectedBetweenAtSignTokenAndAttributeName.trailingTriviaLength.utf8Length != 0
+    {
+      addDiagnostic(
+        unexpectedBetweenAtSignTokenAndAttributeName,
+        StaticParserError.invalidWhitespaceBetweenAttributeAtSignAndIdentifier,
+        fixIts: [
+          FixIt(
+            message: StaticParserFixIt.removeExtraneousWhitespace,
+            changes: [
+              .makeMissing(unexpectedBetweenAtSignTokenAndAttributeName, transferTrivia: false),
+              .makePresent(node.attributeName),
+            ]
+          )
+        ],
+        handledNodes: [
+          node.id,
+          unexpectedBetweenAtSignTokenAndAttributeName.id,
+          node.attributeName.id,
+        ]
+      )
     }
     return .visitChildren
   }
--- a/Tests/SwiftParserTest/AttributeTests.swift
+++ b/Tests/SwiftParserTest/AttributeTests.swift
@@ -631,4 +631,43 @@ final class AttributeTests: XCTestCase {
       """
     )
   }
+
+  func testInvalidWhitespaceBetweenAtSignAndIdenfifierIsDiagnosed() {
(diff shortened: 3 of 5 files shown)


Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -456,7 +456,50 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
         ],
         handledNodes: [argument.id]
       )
-      return .visitChildren
+    }
+    if node.hasError,
+       let unexpectedBeforeAtSignTokens = node.unexpectedBeforeAtSignToken,
+       let unexpectedBeforeAtSign = unexpectedBeforeAtSignTokens.lastToken(viewMode: .sourceAccurate) {
```
ahoppen, line 462: Wit this implementation you would be discarding any tokens before the `@` sign, but the diagnostic message only complains about whitespace. What I would do instead is to check that `unexpectedBeforeAtSignTokens` only contains a single `@` token as follows:

```swift
let unexpectedBeforeAtSign = unexpectedBeforeAtSignTokens.onlyToken(where: { $0.tokenKind == .atSign && !$0.trailingTrivia.isEmpty })
```

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -456,7 +456,50 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
+              ]
+            )
+          ],
+          handledNodes: [
+            node.id,
+            unexpectedBeforeAtSignTokens.id,
+            node.atSignToken.id
+          ]
+        )
+      } else if unexpectedBeforeAtSign.tokenKind == .identifier(unexpectedBeforeAtSign.text) {
```
ahoppen, line 483: This condition seems odd to me. Why would we have an identifier *before* the `@`?

ahoppen:
@TiagoMaiaL Just wanted to let you know that I’m making a similar change in https://github.com/apple/swift-syntax/pull/1665 and you might be able to also use `handleExtraneousWhitespaceError` from it for your PR.

Tests/SwiftParserTest/AttributeTests.swift
```diff
@@ -631,4 +631,43 @@ final class AttributeTests: XCTestCase {
...
+      """
+      @
+      MyAttribute
+      func foo() {}
+      """,
+      diagnostics: [
+        DiagnosticSpec(
+          message: "extraneous whitespace after '@' is not permitted",
+          fixIts: ["remove whitespace"]
+        )
```
ahoppen, line 665: Did you forget to add a diagnostic marker `1️⃣` for this `DiagnosticSpec`? I would expect the test to fail as-is.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -456,7 +456,47 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
         ],
         handledNodes: [argument.id]
       )
-      return .visitChildren
+    }
+    if let unexpectedAtSign = node.unexpectedBeforeAtSignToken?.onlyToken(where: { $0.tokenKind == .atSign && !$0.trailingTrivia.isEmpty }) {
+      addDiagnostic(
+        unexpectedAtSign,
+        position: node.unexpectedBeforeAtSignToken?.endPosition.advanced(by: -1),
```
ahoppen, line 463: I think this only works if there is a single space after the `@`. If there are more spaces, this will show the error message before the last space. I think what you want is. `node.unexpectedBeforeAtSignToken.endPositionBeforeTrailingTrivia`.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -456,7 +456,47 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
+              node.attributeName.isMissingAllTokens {
+      addDiagnostic(
+        unexpectedIdentifier,
+        StaticParserError.invalidWhitespaceBetweenAttributeAtSignAndIdentifier,
+        fixIts: [
+          FixIt(
+            message: StaticParserFixIt.removeExtraneousWhitespace,
+            changes: [
+              .makeMissing(unexpectedIdentifier, transferTrivia: false),
+              .makePresent(node.attributeName), // TODO: Provide synthesized attributeName as a replacement.
```
ahoppen, line 490: What do you mean by this TODO?

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -456,7 +456,47 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
+          )
+        ],
+        handledNodes: [
+          node.id,
+          unexpectedAtSign.id,
+          node.atSignToken.id
+        ]
+      )
+    } else if let unexpectedIdentifier = node.unexpectedBetweenAtSignTokenAndAttributeName?.onlyToken(where: { $0.tokenKind.isIdentifier && !$0.leadingTrivia.isEmpty }),
+              node.attributeName.isMissingAllTokens {
```
ahoppen, line 481: The `onlyToken(where: { $0.tokenKind.isIdentifier })` check fails when the type for the attribute has generic arguments, e.g. `@MyType<Int>`. I think all you need to check here is that `node.unexpectedBetweenAtSignTokenAndAttributeName?.node.leadingTriviaLength != 0` and then make the entire `node.unexpectedBetweenAtSignTokenAndAttributeName` missing, even if it contains multiple tokens.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -456,7 +456,49 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
         ],
         handledNodes: [argument.id]
       )
-      return .visitChildren
+    }
+    if let unexpectedAtSign = node.unexpectedBeforeAtSignToken?.onlyToken(where: { $0.tokenKind == .atSign && !$0.trailingTrivia.isEmpty }) {
+      addDiagnostic(
+        unexpectedAtSign,
+        position: node.unexpectedBeforeAtSignToken?.endPositionBeforeTrailingTrivia,
```
ahoppen, line 463: Oh, one thing that just occurred to me. We don’t actually need to optionally chaining because we know that `node.unexpectedBeforeAtSignToken` only contains `unexpectedAtSign`.

```suggestion
        position: unexpectedAtSign.endPositionBeforeTrailingTrivia,
```

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -456,7 +456,49 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
         ],
         handledNodes: [argument.id]
       )
-      return .visitChildren
+    }
+    if let unexpectedAtSign = node.unexpectedBeforeAtSignToken?.onlyToken(where: { $0.tokenKind == .atSign && !$0.trailingTrivia.isEmpty }) {
```
ahoppen, line 460: Could you also check that `node.atSignToken` is missing? If it’s not `.makePresent` will hit an assertion failure.

ahoppen (commented):
Two more minor comments I saw while reading it again, otherwise LGTM.

ahoppen (approved):
Thank you

ahoppen:
@swift-ci Please test

Sources/SwiftParser/Attributes.swift
```diff
@@ -166,8 +166,34 @@ extension Parser {
...
+      )
+    } else if attributeName.raw.leadingTriviaByteLength != 0 {
+      unexpectedBetweenAtSignAndAttributeName = RawUnexpectedNodesSyntax(attributeName)
+      // `withLeadingTrivia` only returns `nil` if there is no token in `attributeName`.
+      // But since `attributeName` has leadingTriviaLength != 0 there must be trivia and thus a token.
+      // So we can safely force-unwrap here.
+      attributeName = attributeName
+        .raw
+        .withLeadingTrivia([], arena: self.arena)!
+        .as(RawTypeSyntax.self)!
```
ahoppen, line 194: I suspect this is where the round-trip failures is coming from. You have already stored `attributeName` in `unexpectedBetweenAtSignAndAttributeName` and now you’re adding it again with `SourcePresence.present`. Although it doesn’t quite match the failure message, we have the node twice in the tree now, once with trivia and once without trivia.

The best solution here is a little bit ugly but it’s the best I can come up with: We want to shove the first token of the `attributeName` (which contains the leading trivia) into `unexpectedBetweenAtSignAndAttributeName` and instead synthesize a new missing token that doesn’t have leading trivia for the first token in `attributeName`. 

Getting the first token to put into `unexpectedBetweenAtSignAndAttributeName` should be as easy as doing `attributeName.firstToken(viewMode: .sourceAccurate)`.

To replace the first token in `attributeName` you need to write a `SyntaxRewriter`. On the first token that it encounters, it should generate a corresponding missing token without leading trivia. After that, it doesn’t rewrite any further tokens. `SyntaxRewriter` operates on the `Syntax`, not `RawSyntax` level, so you will need to convert `attributeName` to a `Syntax` and extract the `RawSyntax` from the rewritten node. Also, you need to make sure that the rewritten first token is allocated in the same `SyntaxArena` as all the other tokens. https://github.com/apple/swift-syntax/pull/1383 contains a change to `SyntaxRewriter` that allows that.

TiagoMaiaL: > Getting the first token to put into unexpectedBetweenAtSignAndAttributeName should be as easy as doing attributeName.firstToken(viewMode: .sourceAccurate).

The only way I managed to do this was to first declare `RawSyntax.firstToken` as `@_spi(RawSyntax) public` and then doing the following:
```swift
    } else if attributeName.raw.leadingTriviaByteLength != 0,
      let firstAttributeNameToken = attributeName.raw.firstToken(viewMode: .sourceAccurate)
    {
      unexpectedBetweenAtSignAndAttributeName = RawUnexpectedNodesSyntax(
        firstAttributeNameToken.withPresence(.present, arena: self.arena)
      )
      ...
```

ahoppen: Can you push your changes and then I’ll take a look at them.

ahoppen:
@TiagoMaiaL Are you still working on this? If so, could you rebase on top of `main` and look at the round-trip failure?

adamcarter93:
Hey all, is this still open?

I'm just looking for a nice first issue and looks like this is done but not merged and it's been a while so don't want to step on any toes! 🦶🏻

ahoppen:
This should have been fixed by https://github.com/apple/swift-syntax/pull/2466

PR #1660, opened 2023-05-13
Created an Examples-all target to include all the example targets, and updated build_script.py.

### LINKED ISSUE #1655: Make sure that `build-script.py` builds all examples
Currently, `build-script.py` only builds the `AddOneToIntegerLiterals` and `CodeGenerationUsingSwiftSyntaxBuilder` examples. We should build all examples. Since we need to explicitly specify a product or target to build because we are using a multi root data file, we should an `Examples-all` target similar to what we do in `SwiftSyntax`

https://github.com/apple/swift-syntax/blob/17e83de1e15c809f554bd32ec1e67129872ec26b/Package.swift#L284-L296

When we change this, we also don’t need to build `ExamplePlugin` from the test command anymore.

TTOzzi:
Resolve #1655 

Created the Examples-all target in the same way as SwiftSyntax-all and modified build_script.py to build all the targets in the Examples package with it.

The command to build only the ExamplePlugin in the test command was left in because I think the ExamplePlugin is being used in Swift-syntax's compiler_plugin_basic.swift, but I'll delete it if it's not needed.

https://github.com/apple/swift-syntax/blob/d14176d74bff98aa877bbf284a8bccd4bb02c833/build-script.py#L516-L545

https://github.com/apple/swift-syntax/blob/d14176d74bff98aa877bbf284a8bccd4bb02c833/lit_tests/compiler_plugin_basic.swift#L14-L18

### DIFF
--- a/Examples/Package.swift
+++ b/Examples/Package.swift
@@ -40,3 +40,19 @@ let package = Package(
     ),
   ]
 )
+
+// This is a fake target that depends on all targets in the package.
+// We need to define it manually because the `Examples-Package` target doesn't exist for `swift build`.
+
+package.targets.append(
+  .target(
+    name: "Examples-all",
+    dependencies: package.targets.compactMap {
+      if $0.type == .test {
+        return nil
+      } else {
+        return .byName(name: $0.name)
+      }
+    }
+  )
+)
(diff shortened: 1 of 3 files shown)


ahoppen (approved):
Thank you @TTOzzi, this is looking great. And thanks for noticing that `ExamplePlugin` is used by `compiler_plugin_basic.swift` 👍🏽 

Once CI passes, I’ll merge this PR.

ahoppen:
@swift-ci Please test

PR #1661, opened 2023-05-14
Fix newline parsing at trailing trivia

### LINKED ISSUE #1626: Statements that are separated by a newline inside a block comment are not accepted
### Issue Kind

Parse of Valid Source Failed

### Source Code

```swift
let a = 2/*
*/let b = 3
```

### Description

The source code should parse fine but it doesn’t because the newline inside the block comment is trailing trivia to `2` and thus not considered to determine whether the second `let` is on a new line.

TTOzzi:
Resolve #1626 

Unlike leadingTrivia, trailingTrivia was not using the newlinePresence value of triviaResult.
Modified the logic to store the trailingTrivia newlinePresence value from the previous token and set isAtStartOfLine flag when parsing the next token.

### DIFF
--- a/Sources/SwiftParser/Lexer/Cursor.swift
+++ b/Sources/SwiftParser/Lexer/Cursor.swift
@@ -320,8 +320,8 @@ extension Lexer {
     /// If `tokenKind` is `.keyword`, the kind of keyword produced, otherwise
     /// `nil`.
     let keywordKind: Keyword?
-    /// Indicates whether the lexed token text contains a newline.
-    let newlinePresence: Lexer.Cursor.NewlinePresence
+    /// Indicates whether the end of the lexed token text contains a newline.
+    let trailingNewlinePresence: Lexer.Cursor.NewlinePresence
 
     private init(
       _ tokenKind: RawTokenKind,
@@ -330,7 +330,7 @@ extension Lexer {
       stateTransition: StateTransition?,
       trailingTriviaLexingMode: Lexer.Cursor.TriviaLexingMode?,
       keywordKind: Keyword?,
-      newlinePresence: Lexer.Cursor.NewlinePresence
+      trailingNewlinePresence: Lexer.Cursor.NewlinePresence
     ) {
       self.tokenKind = tokenKind
       self.flags = flags
@@ -338,7 +338,7 @@ extension Lexer {
       self.stateTransition = stateTransition
       self.trailingTriviaLexingMode = trailingTriviaLexingMode
       self.keywordKind = keywordKind
-      self.newlinePresence = newlinePresence
+      self.trailingNewlinePresence = trailingNewlinePresence
     }
 
     /// Create a lexer result. Note that keywords should use `Result.keyword`
@@ -349,7 +349,7 @@ extension Lexer {
       error: Cursor.LexingDiagnostic? = nil,
       stateTransition: StateTransition? = nil,
       trailingTriviaLexingMode: Lexer.Cursor.TriviaLexingMode? = nil,
-      newlinePresence: Lexer.Cursor.NewlinePresence = .absent
+      trailingNewlinePresence: Lexer.Cursor.NewlinePresence = .absent
     ) {
       precondition(tokenKind != .keyword, "Use Result.keyword instead")
       self.init(
@@ -359,7 +359,7 @@ extension Lexer {
         stateTransition: stateTransition,
         trailingTriviaLexingMode: trailingTriviaLexingMode,
         keywordKind: nil,
-        newlinePresence: newlinePresence
+        trailingNewlinePresence: trailingNewlinePresence
       )
     }
 
@@ -372,7 +372,7 @@ extension Lexer {
         stateTransition: nil,
         trailingTriviaLexingMode: nil,
         keywordKind: kind,
-        newlinePresence: .absent
+        trailingNewlinePresence: .absent
       )
     }
   }
@@ -448,7 +448,7 @@ extension Lexer.Cursor {
       flags.insert(.isAtStartOfLine)
     }
 
-    self.previousLexemeTrailingNewlinePresence = result.newlinePresence
+    self.previousLexemeTrailingNewlinePresence = result.trailingNewlinePresence
 
     if let stateTransition = result.stateTransition {
       self.stateStack.perform(stateTransition: stateTransition, stateAllocator: stateAllocator)
@@ -1905,7 +1905,7 @@ extension Lexer.Cursor {
           if character == UInt8(ascii: "\r") {
             _ = self.advance(matching: "\n")
           }
-          return Lexer.Result(.stringSegment, error: error, newlinePresence: .present)
+          return Lexer.Result(.stringSegment, error: error, trailingNewlinePresence: .present)
         } else {
           // Single line literals cannot span multiple lines.
           // Terminate the string here and go back to normal lexing (instead of `afterStringLiteral`)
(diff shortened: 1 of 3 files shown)


Sources/SwiftParser/Lexer/Cursor.swift
```diff
@@ -242,6 +256,9 @@ extension Lexer {
 
     /// If we have already lexed a token, the kind of the previously lexed token
     var previousTokenKind: RawTokenKind?
+    
+    /// If we have already lexed a token, the `NewlinePresence` of the previously lexed token
+    var previousTokenNewlinePresence: NewlinePresence?
```
ahoppen, line 261: Just a naming suggestion: I’d like to make it absolutely clear that this is about whether the previous token’s *trailing* trivia has a newline and avoid any confusion of whether the previous token had a newline inside its leading trivia. So I’d go with

```suggestion
    
    /// If we have already lexed a token, stores whether the previous lexeme‘s trailing trivia contains a newline.
    var previousLexemeTrailingTrivaNewlinePresence: NewlinePresence?
```

ahoppen (commented):
Thanks for figuring this out @TTOzzi. I’ve got a couple questions/comments inline. Otherwise, this looks very good to me 👍🏽

Sources/SwiftParser/Lexer/Cursor.swift
```diff
@@ -433,23 +450,30 @@ extension Lexer.Cursor {
...
       self.stateStack.perform(stateTransition: stateTransition, stateAllocator: stateAllocator)
     }
-
+    
+    var flags = result.flags
+    if newlineInLeadingTrivia == .present {
+      flags.insert(.isAtStartOfLine)
+    }
+    if let previousTokenNewlinePresence, previousTokenNewlinePresence == .present,
+       !currentState.isParsingMultilineString {
```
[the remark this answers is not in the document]
TTOzzi: In an effort to minimize side effects, I focused on maintaining the existing test cases, which led me to misunderstand the behavior of the lexer. As you mentioned, it is correct that the line1, line2, and whitespace tokens, excluding the `"""` character, are really on new lines.
I will try to make the modifications quickly within this week. Thank you for your prompt feedback 🙂

Sources/SwiftParser/Lexer/Cursor.swift
TTOzzi: Great suggestion! I will try to make the modifications along with the changes mentioned above. Thank you.

Sources/SwiftParser/Lexer/Cursor.swift
```diff
@@ -242,6 +242,9 @@ extension Lexer {
 
     /// If we have already lexed a token, the kind of the previously lexed token
     var previousTokenKind: RawTokenKind?
+    
+    /// If we have already lexed a token, stores whether the previous lexeme‘s ending contains a newline.
+    var previousLexemeTrailingNewlinePresence: NewlinePresence?
```
TTOzzi, line 247: Hello @ahoppen!
When I modified the parsing logic so that newlines in multiline strings also have the isAtStartOfLine flag, I used this property together.
So I changed the name to something that could include both trailing trivia and multiline string parsing, but would it be better to split them up and have them as separate properties? 🤔 
May I ask for your opinions?

ahoppen: That’s a good name. I like it!

Sources/SwiftParser/Lexer/Cursor.swift
```diff
@@ -433,28 +445,24 @@ extension Lexer.Cursor {
     if let stateTransition = result.stateTransition {
       self.stateStack.perform(stateTransition: stateTransition, stateAllocator: stateAllocator)
     }
-
+
```
ahoppen, line 448: I think `swift-format` will complain about this whitespace. Could you run `format.py`? https://github.com/apple/swift-syntax/blob/main/CONTRIBUTING.md#formatting

Sources/SwiftParser/Lexer/Cursor.swift
```diff
@@ -402,6 +405,15 @@ extension Lexer.Cursor {
...
     }
+    
+    var flags: Lexer.Lexeme.Flags = []
+    if newlineInLeadingTrivia == .present {
+      flags.insert(.isAtStartOfLine)
+    }
+    if let previousLexemeTrailingNewlinePresence, previousLexemeTrailingNewlinePresence == .present {
+      flags.insert(.isAtStartOfLine)
+    }
+    self.previousLexemeTrailingNewlinePresence = nil
```
ahoppen, line 416: Resetting to `nil` doesn’t really make sense here, right? If we follow the same meaning as for `previousTokenKind`, `nil` means that there hasn’t been a previous token but in this case there has been. I think the better implementation would be to have in `else` branch in the `if` statement in which you set `self.previousLexemeTrailingNewlinePresence` below and in it set `self.previousLexemeTrailingNewlinePresence = .absent`.

If you do that, I think you can also move this entire code block below the `Token text` lexing section and initialize `flags` with `result.flags` and avoid the `union` call when forming the result.

Sources/SwiftParser/Lexer/Cursor.swift
TTOzzi: Oops, I missed that!
I'll make sure to run it before every commit.
Thanks for pointing that out 🙇

Sources/SwiftParser/Lexer/Cursor.swift
TTOzzi: As you said, it seems like it should have a non-nil value when lexing the previous token.
However, I'm concerned that it would be redundant to assign .absent to each else in the branch that sets previousLexemeTrailingNewlinePresence.

https://github.com/apple/swift-syntax/pull/1661/files#diff-253b01bc981faa185173c1697784a41bf1b586e91ec7b4de806dfd2646db52c3R1900

The code above makes it look like I need to put previousLexemeTrailingNewlinePresence = .absent in the lex logic for each case.
https://github.com/apple/swift-syntax/blob/main/Sources/SwiftParser/Lexer/Cursor.swift#L410-L431

Why not just leave the code where it is, but assign absent as the default after the value is used?
I think we can get some reasonable logic with little modification.

```suggestion
    } else {
      self.previousLexemeTrailingNewlinePresence = .absent
    }
```

TTOzzi: Additionally, if we move this block of code below the token text lexing section, the value assigned in the token text lexing logic will affect the flag value of the current token, so it would be difficult to move it 🤔

ahoppen: Yes, that’s just what I was just about to suggest

TTOzzi: Similar to `TriviaResult`, I also added a `newlinePresence` property to `Lexer.Result` and changed it so that modifications to the `previousLexemeTrailingNewlinePresence` value can only be made within the `nextToken` method.
The flag initialize logic was also moved under the token text lexing section.

ahoppen (approved):
Thanks. This looks good ✅

ahoppen:
@swift-ci Please test

TTOzzi:
I appreciate your kind feedback. Thank you @ahoppen!

ahoppen:
@TTOzzi Could you also open a PR with this changes targeting the `release/5.9` branch? Since this fixes a proper parsing bug, I would like to include it in the Swift 5.9 release.

TTOzzi:
@ahoppen Of course, I can do it! When should I create the PR until?

ahoppen:
There is no clear deadline at the moment. Earlier is always better than later though 😉

TTOzzi:
Ok. It's currently working hours here, so I'll create it after I finish work and leave for the day 😉

ahoppen:
Anytime this week is equally good

PR #1662, opened 2023-05-15
Add helper function for getting the source edits of two strings

StevenWong12:
This change should make it easier to test two different source by getting `SourceEdit` with the helper function

### DIFF
--- a/Sources/_SwiftSyntaxTestSupport/SourceEditsTestUtils.swift
+++ b/Sources/_SwiftSyntaxTestSupport/SourceEditsTestUtils.swift
@@ -1,0 +1,76 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+import XCTest
+
+/// Returns the `ConcurrentEdits`s to transition from `base` to `new`.
+public func getConcurrentEdits(from base: String, to new: String) -> ConcurrentEdits {
+  let diffCollection = new.difference(from: base)
+
+  let diffArray: [(offset: Int, isInsert: Bool)] =
+    diffCollection
+    .map { diff in
+      switch diff {
+      case .remove(offset: let offset, element: _, associatedWith: _):
+        return (offset: offset, isInsert: false)
+      case .insert(offset: let offset, element: _, associatedWith: _):
+        return (offset: offset, isInsert: true)
+      }
+    }
+    .sorted {
+      // Change.remove should prior to Change.insert
+      if $0.offset < $1.offset {
+        return true
+      } else if $0.offset == $1.offset {
+        return !$0.isInsert
+      } else {
+        return false
+      }
+    }
+
+  let sourceEdits = diffArray.map({
+    return $0.isInsert ? SourceEdit(offset: $0.offset, length: 0, replacementLength: 1) : SourceEdit(offset: $0.offset, length: 1, replacementLength: 0)
+  })
+
+  return ConcurrentEdits(fromSequential: sourceEdits)
+}
+
+/// Apply the given edits to `testString` and return the resulting string.
+/// `concurrent` specifies whether the edits should be interpreted as being
+/// applied sequentially or concurrently.
+public func applyEdits(
+  _ edits: [SourceEdit],
+  concurrent: Bool,
+  to testString: String,
+  replacementChar: Character = "?"
+) -> String {
+  guard let replacementAscii = replacementChar.asciiValue else {
+    fatalError("replacementChar must be an ASCII character")
+  }
+  var edits = edits
+  if concurrent {
+    XCTAssert(ConcurrentEdits._isValidConcurrentEditArray(edits))
+
+    // If the edits are concurrent, sorted and not overlapping (as guaranteed by
+    // the check above, we can apply them sequentially to the string in reverse
+    // order because later edits don't affect earlier edits.
+    edits = edits.reversed()
+  }
+  var bytes = Array(testString.utf8)
+  for edit in edits {
+    assert(edit.endOffset <= bytes.count)
+    bytes.removeSubrange(edit.offset..<edit.endOffset)
+    bytes.insert(contentsOf: [UInt8](repeating: replacementAscii, count: edit.replacementLength), at: edit.offset)
+  }
+  return String(bytes: bytes, encoding: .utf8)!
+}
--- a/Sources/_SwiftSyntaxTestSupport/SourceEditsTestUtils.swift
+++ b/Sources/_SwiftSyntaxTestSupport/SourceEditsTestUtils.swift
@@ -39,7 +39,11 @@ public func getConcurrentEdits(from base: String, to new: String) -> ConcurrentE
     }
 
   let sourceEdits = diffArray.map({
-    return $0.isInsert ? SourceEdit(offset: $0.offset, length: 0, replacementLength: 1) : SourceEdit(offset: $0.offset, length: 1, replacementLength: 0)
+    if $0.isInsert {
+      return SourceEdit(offset: $0.offset, length: 0, replacementLength: 1)
+    } else {
+      return SourceEdit(offset: $0.offset, length: 1, replacementLength: 0)
+    }
   })
 
   return ConcurrentEdits(fromSequential: sourceEdits)
--- a/Tests/SwiftParserTest/IncrementalParsingTests.swift
+++ b/Tests/SwiftParserTest/IncrementalParsingTests.swift
@@ -13,34 +13,43 @@
 import XCTest
 import SwiftSyntax
 import SwiftParser
+import _SwiftSyntaxTestSupport
 
 public class IncrementalParsingTests: XCTestCase {
-
   public func testIncrementalInvalid() {
     let original = "struct A { func f() {"
-    let step: (String, (Int, Int, String)) =
-      ("struct AA { func f() {", (8, 0, "A"))
+    let newSource = "struct AA { func f() {"
 
-    var tree = Parser.parse(source: original)
-    let sourceEdit = SourceEdit(range: ByteSourceRange(offset: step.1.0, length: step.1.1), replacementLength: step.1.2.utf8.count)
-    let lookup = IncrementalParseTransition(previousTree: tree, edits: ConcurrentEdits(sourceEdit))
-    tree = Parser.parse(source: step.0, parseTransition: lookup)
-    XCTAssertEqual("\(tree)", step.0)
+    let concurrentEdits = getConcurrentEdits(from: original, to: newSource)
+
+    let oldTree = Parser.parse(source: original)
+    let lookup = IncrementalParseTransition(previousTree: oldTree, edits: concurrentEdits)
+    let newTree = Parser.parse(source: newSource, parseTransition: lookup)
+
+    XCTAssertEqual("\(newTree)", newSource)
   }
 
   public func testReusedNode() throws {
     try XCTSkipIf(true, "Swift parser does not handle node reuse yet")
 
-    let original = "struct A {}\nstruct B {}\n"
-    let step: (String, (Int, Int, String)) =
-      ("struct AA {}\nstruct B {}\n", (8, 0, "A"))
+    let original =
+      """
+      struct A {}
+      struct B {}
+      """
+
+    let newSource =
+      """
+      struct A {}
+      struct B {}
+      """
 
     let origTree = Parser.parse(source: original)
-    let sourceEdit = SourceEdit(range: ByteSourceRange(offset: step.1.0, length: step.1.1), replacementLength: step.1.2.utf8.count)
+    let concurrentEdits = getConcurrentEdits(from: original, to: newSource)
     let reusedNodeCollector = IncrementalParseReusedNodeCollector()
-    let transition = IncrementalParseTransition(previousTree: origTree, edits: ConcurrentEdits(sourceEdit), reusedNodeDelegate: reusedNodeCollector)
-    let newTree = Parser.parse(source: step.0, parseTransition: transition)
-    XCTAssertEqual("\(newTree)", step.0)
+    let transition = IncrementalParseTransition(previousTree: origTree, edits: concurrentEdits, reusedNodeDelegate: reusedNodeCollector)
+    let newTree = Parser.parse(source: newSource, parseTransition: transition)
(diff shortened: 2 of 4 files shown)


Tests/SwiftParserTest/IncrementalParsingTests.swift
```diff
@@ -16,31 +16,85 @@ import SwiftParser
...
+    let firstDiff = diffs.first!
+    XCTAssert(firstDiff == SourceEdit(offset: 8, length: 0, replacementLength: 1))
+  }
+
+  public func testDiffOfTwoSameStrings() {
+    let s1 = "0123456"
+    let s2 = "0123456"
+
+    let diffs = getDiffOfTwoString(s1, s2)
+    XCTAssert(diffs.count == 0)
```
ahoppen, line 35:
```suggestion
    XCTAssert(diffs.isEmpty)
```

Tests/SwiftParserTest/IncrementalParsingTests.swift
```diff
@@ -16,31 +16,85 @@ import SwiftParser
...
+  }
+
+  public func testDiffOfTwoStrings() {
+    let s1 = "0123456"
+    let s2 = "x12456yz"
+
+    // should return [(0, 1, 0), (0, 0, 1), (3, 1, 0), (6, 0, 1), (7, 0, 1)]
```
ahoppen, line 42: Do you think this comment provides much value? It’s exactly what you are specifying below in `expectedDiffs` and in my experience these comments just get out-of-sync. 😆

Tests/SwiftParserTest/IncrementalParsingTests.swift
```diff
@@ -61,3 +115,46 @@ public class IncrementalParsingTests: XCTestCase {
...
 }
+
+// MARK: helper functions
+extension IncrementalParsingTests {
+  /// return the diff of two strings with SourceEdit
+  /// example:
+  ///  - s1: 0123456
+  ///  - s2: x12456yz
+  ///  should return [(0, 1, 0,), (0, 0, 1), (3, 1, 0), (6, 0, 1), (7, 0, 1)]
+  ///  the above edits are **serial**
```
ahoppen, line 126: Instead of explicitly specifying that these edits are serial, what do you think about returning `ConcurrentEdits` here? That way the Swift type system will make sure we aren’t mixing up sequential and concurrent edits.

Tests/SwiftParserTest/IncrementalParsingTests.swift
```diff
@@ -61,3 +115,46 @@ public class IncrementalParsingTests: XCTestCase {
     XCTAssertEqual(origStructB.parent!.id, reusedNode.parent!.id)
   }
 }
+
+// MARK: helper functions
+extension IncrementalParsingTests {
```
ahoppen, line 120: Since you aren’t using any members (not that there are any) of `IncrementalParsingTests`, I would just make `getDiffOfTwoString` a top-level function. That’s also what we do with all the help functions in `_SwiftSyntaxTestSupport`.

Tests/SwiftParserTest/IncrementalParsingTests.swift
StevenWong12: That comment was used to remind myself what `expectedDiffs` should be when writing test case lol. Already dropped it.😄

Tests/SwiftParserTest/IncrementalParsingTests.swift
StevenWong12: Since our goal is to use this function to provide `ConcurrentEdits` for `IncrementalParseTransition`, that makes sense to me.👍

Tests/SwiftParserTest/IncrementalParsingTests.swift
```diff
@@ -16,31 +16,85 @@ import SwiftParser
...
+
+    let expectedDiffs: [SourceEdit] = [
+      SourceEdit(offset: 0, length: 1, replacementLength: 0),
+      SourceEdit(offset: 0, length: 0, replacementLength: 1),
+      SourceEdit(offset: 3, length: 1, replacementLength: 0),
+      SourceEdit(offset: 6, length: 0, replacementLength: 1),
+      SourceEdit(offset: 7, length: 0, replacementLength: 1),
+    ]
+
+    XCTAssert(diffs == expectedDiffs)
```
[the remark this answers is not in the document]
StevenWong12: I think `applyEdits` is written in `SwiftSyntaxTest` and we're in `SwiftParserTest` here so I could not directly call it. If that's necessary I could move it to elsewhere.

Tests/SwiftParserTest/IncrementalParsingTests.swift
```diff
@@ -14,33 +14,108 @@ import XCTest
...
     let origTree = Parser.parse(source: original)
-    let sourceEdit = SourceEdit(range: ByteSourceRange(offset: step.1.0, length: step.1.1), replacementLength: step.1.2.utf8.count)
+    let concurrentEdits = getConcurrentEdits(from: original, to: newSource)
     let reusedNodeCollector = IncrementalParseReusedNodeCollector()
-    let transition = IncrementalParseTransition(previousTree: origTree, edits: ConcurrentEdits(sourceEdit), reusedNodeDelegate: reusedNodeCollector)
-    let newTree = Parser.parse(source: step.0, parseTransition: transition)
-    XCTAssertEqual("\(newTree)", step.0)
+    let transition = IncrementalParseTransition(previousTree: origTree, edits: concurrentEdits, reusedNodeDelegate: reusedNodeCollector)
+    let newTree = Parser.parse(source: newSource, parseTransition: transition)
+    XCTAssertEqual("\(newTree)", "\(newSource)")
```
ahoppen, line 118: AFAICT there’s no need to wrap `newSource` in a string literal

```suggestion
    XCTAssertEqual("\(newTree)", newSource)
```

utils/incrparse/incr_transfer_round_trip.py
ahoppen, line 1: The tests in `lit_tests/incrParse` are still using these python files, so we can’t delete them just yet. If you want to run them locally, you can do so as described here. You just need to replace `build` by `test` (and I think you might need to build the compiler to get the `FileCheck` executable to verify the files, but maybe there’s some other way of getting it, eg. through Homebrew, I’m not sure)

https://github.com/apple/swift-syntax/blob/94959f0693610c3f51fc8823c2f38bb00b064c26/CONTRIBUTING.md?plain=1#L10-L21

Tests/SwiftParserTest/IncrementalParsingTests.swift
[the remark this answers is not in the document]
ahoppen: Actually, now that I think about it, it would make sense to move `getConcurrentEdits` to `_SwiftSyntaxTestSupport` and move all of your new tests in `Tests/SwiftSyntaxTestSupportTests` because they are really testing that this helper function actually works. You can then also move `applyEdits` to `_SwiftSyntaxTestSupport` and use it from your new tests.

Tests/SwiftParserTest/IncrementalParsingTests.swift
```diff
@@ -14,33 +14,108 @@ import XCTest
 import SwiftSyntax
 import SwiftParser
 
+/// Returns the `ConcurrentEdits`s to transition from `base` to `new`.
+func getConcurrentEdits(from base: String, to new: String) -> ConcurrentEdits {
```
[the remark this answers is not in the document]
StevenWong12: Since I moved `getConcurrentEdits` and `applyEdits` to a standalone file, I made both of them `public` to be called in other test files.

utils/incrparse/incr_transfer_round_trip.py
StevenWong12: Oh, I see. Maybe I could translate those tests in `IncrementalParsingTests` later when I added some functionality related to them.

Sources/_SwiftSyntaxTestSupport/SourceEditsTestUtils.swift
```diff
@@ -0,0 +1,76 @@
...
+        return true
+      } else if $0.offset == $1.offset {
+        return !$0.isInsert
+      } else {
+        return false
+      }
+    }
+
+  let sourceEdits = diffArray.map({
+    return $0.isInsert ? SourceEdit(offset: $0.offset, length: 0, replacementLength: 1) : SourceEdit(offset: $0.offset, length: 1, replacementLength: 0)
```
ahoppen, line 42: I would make this an `if` statement. It’s just easier to read:

```suggestion
    if $0.isInsert {
      return SourceEdit(offset: $0.offset, length: 0, replacementLength: 1)
    } else {
      SourceEdit(offset: $0.offset, length: 1, replacementLength: 0)
    }
```

---

Also, I think for the future we want this to be able to merge `SourceEdit`s to create `SourceEdit`s that either have `length` or `replacementLength` that’s created.

I.e. `base = "0123"` and `new = "abc0123"` should create a single `SourceEdit(offset: 0, length: 0, replacementLength: 3)`.

Tests/SwiftParserTest/IncrementalParsingTests.swift
```diff
@@ -13,34 +13,43 @@
...
-      ("struct AA {}\nstruct B {}\n", (8, 0, "A"))
+    let original =
+      """
+      struct A {}
+      struct B {}
+      """
+
+    let newSource =
+      """
+      struct A {}
```
ahoppen, line 43: `newSource` should have `struct AA` instead of `struct A`

```suggestion
      struct AA {}
```

Tests/SwiftSyntaxTestSupportTest/SourceEditsTestUtilsTest.swift
```diff
@@ -0,0 +1,54 @@
...
+//===----------------------------------------------------------------------===//
+
+import XCTest
+import SwiftSyntax
+import _SwiftSyntaxTestSupport
+
+public class SourceEditsUtilTest: XCTestCase {
+  public func testDiffOfTwoStringsSimple() throws {
+    let s1 = "struct A { func f() {"
+    let s2 = "struct AA { func f() {"
```
ahoppen, line 20: Just to have more descriptive names, could you name these strings `base` and `new` like in `getConcurrentEdits`? Same for all the other test cases.

Tests/SwiftSyntaxTestSupportTest/SourceEditsTestUtilsTest.swift
```diff
@@ -0,0 +1,54 @@
...
+      SourceEdit(offset: 3, length: 1, replacementLength: 0),
+      SourceEdit(offset: 7, length: 0, replacementLength: 2),
+    ]
+
+    XCTAssertEqual(diffs.edits, expectedDiffs)
+
+    let s3 = applyEdits(expectedDiffs, concurrent: true, to: s1)
+
+    XCTAssertEqual(s3, "?12456??")
+  }
```
ahoppen, line 53: Could you add one more test case that has an insertion at the start and an insertion at the end. That would make sure that `Collection.difference` actually contains sequential edits instead of concurrent edits.

```swift
let base = "0123"
let new = "a0123b"
```

Sources/_SwiftSyntaxTestSupport/SourceEditsTestUtils.swift
StevenWong12: I think for the incremental parsing test, we could just merge them into a `ConcurrentEdits` to help us test. Do you have some use cases about merging `SourceEdit`s into a big one?

ahoppen: Oh, I didn’t realize that the conversion into `ConcurrentEdits` merges them. In that case, it’s good, I think.

Tests/SwiftSyntaxTestSupportTest/SourceEditsTestUtilsTest.swift
ahoppen: Sorry if I wasn’t clear here. What I was looking for is a test case just like the one you have in `testDiffOfTwoStrings` (which I think has a very good test structure), just with the inputs I gave you. So basically:

```swift
  public func testStringDifferenceReturnSequentialEdits() throws {
    let base = "0123"
    let new = "a0123bc"

    let diffs = getConcurrentEdits(from: base, to: new)
    XCTAssertEqual(diffs.edits, [
        SourceEdit(offset: 0, length: 0, replacementLength: 1),
        SourceEdit(offset: 4, length: 0, replacementLength: 2)
      ])

    let appliedDiffsBase = applyEdits(diffs.edits, concurrent: true, to: base)

    XCTAssertEqual(appliedDiffsBase, "?0123??")
  }
```

Which seems to pass, so that’s good ✅

StevenWong12: I think this `replacementLength > 0` case is covered in `testDiffOfTwoStrings`. Do you think we should keep the current version of `testStringDifferenceReturnSequentialEdits` (which makes sure `String.difference` correctly produce a string of sequential edits)?

ahoppen:
I just tried running some random tests on the `getConcurrentEdits` and it looks like there are still some bugs in it.

For example `base = "0a1ab"` and `new = "bb01d"` seems to produce incorrect edits. The following test fails

```swift
let base = "0a1ab"
let new =  "bb01d"

let diffs = getConcurrentEdits(from: base, to: new)

let appliedDiffsBase = applyEdits(diffs.edits, concurrent: true, to: base)

XCTAssertEqual(appliedDiffsBase, "??aa?")
```
 
---

The automated test that I used to find this failure was the following (very hacky, the code itself is not ready to be submitted to the repo)

```swift
func randomString(length: Int) -> String {
  let letters = "abcd"
  return String((0..<length).map{ _ in letters.randomElement()! })
}

public func testFoo() {
  for i in 0..<100_000 {
    let startLength = Int.random(in: 0..<6)
    let endLength = Int.random(in: 0..<6)

    let base = randomString(length: startLength)
    let new = randomString(length: endLength)

    let edits = getConcurrentEdits(from: base, to: new)
    let editsApplied = applyEdits(edits.edits, concurrent: true, to: base)

    assert(new.count == editsApplied.count)
    for (n, e) in zip(new, editsApplied) {
      assert(e == "?" || n == e)
    }
  }
}
```

StevenWong12:
I found `String.Difference` sometimes produces some unexpected outputs. e.g.
```swift
"class".difference(from: "struct")
```
The function seems to center on the `c` in `"struct"` and view other characters as deletions and `"lass"` as insertions, which is not we expect.

I opened another [draft pr](https://github.com/apple/swift-syntax/pull/1684) to use markers to mark source edits and generate original and edited source (which is a bit like the previous one in the old parser), I think this should be easier to mark and test, also avoid being trapped in the underlying logic of `String.Difference`.  Maybe we could close this pr and turn to that one. What do you think?

ahoppen:
Thanks for catching that. With that explanation, I agree that being explicit about the edits makes a lot of sense. 👍🏽 So, I prefer https://github.com/apple/swift-syntax/pull/1684

StevenWong12:
Cool. Let's just close this and continue in the other pr.

PR #1666, opened 2023-05-15
[Macros] Copy attrs/modifiers on MacroExpansionDecl to expanded decls 

rintaro:
As per the SE-0397 amendment, copy attributes and modifiers on `MacroExpansionDecl` to the expanded declarations.

Preparation:
* Rename `AttributedSyntax` to `WithAttributesSyntax`.
* Keep `AttributedSyntax` as a typealias to  `WithAttributeSyntax`
* Introduce `WithModifiersSyntax`
* Add `init(_: [Element])` to `SyntaxCollection`'s requirements

### DIFF
--- a/Sources/SwiftParser/CMakeLists.txt
+++ b/Sources/SwiftParser/CMakeLists.txt
@@ -27,6 +27,7 @@ add_swift_host_library(SwiftParser
   Specifiers.swift
   Statements.swift
   StringLiterals.swift
+  StringLiteralRepresentedLiteralValue.swift
   SyntaxUtils.swift
   TokenConsumer.swift
   TokenPrecedence.swift
--- a/Sources/SwiftSyntaxMacroExpansion/MacroExpansion.swift
+++ b/Sources/SwiftSyntaxMacroExpansion/MacroExpansion.swift
@@ -1,5 +1,5 @@
 import SwiftSyntax
-import SwiftSyntaxMacros
+@_spi(MacroExpansion) import SwiftSyntaxMacros
 
 public enum MacroRole {
   case expression
@@ -31,7 +31,13 @@ public func expandFreestandingMacro(
         expandedSyntax = try Syntax(exprMacroDef.expansion(of: node, in: context))
 
       case let declMacroDef as DeclarationMacro.Type:
-        let rewritten = try declMacroDef.expansion(of: node, in: context)
+        var rewritten = try declMacroDef.expansion(of: node, in: context)
+        // Copy attributes and modifiers to the generated decls.
+        if let expansionDecl = node.as(MacroExpansionDeclSyntax.self) {
+          rewritten = rewritten.map {
+            $0.applying(attributes: expansionDecl.attributes, modifiers: expansionDecl.modifiers)
+          }
+        }
         expandedSyntax = Syntax(
           CodeBlockItemListSyntax(
             rewritten.map {
--- a/Sources/SwiftSyntaxMacroExpansion/MacroExpansion.swift
+++ b/Sources/SwiftSyntaxMacroExpansion/MacroExpansion.swift
@@ -13,11 +13,23 @@ public enum MacroRole {
 }
 
 /// Simple diagnostic message
-private struct MacroExpansionError: Error, CustomStringConvertible {
-  let description: String
+private enum MacroExpansionError: String, Error, CustomStringConvertible {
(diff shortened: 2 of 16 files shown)


rintaro:
@swift-ci Please test

Sources/SwiftParser/CMakeLists.txt
```diff
@@ -27,6 +27,7 @@ add_swift_host_library(SwiftParser
   Specifiers.swift
   Statements.swift
   StringLiterals.swift
+  StringLiteralRepresentedLiteralValue.swift
```
rintaro, line 30: Oh this is now unrelated to this PR. I wanted to use this feature (`representedLiteralValue `) in `swift` repo test cases, but ended up with not using it because the compiler doesn't support `@freestanding(declaration, names: arbitrary)` macros.

But this was missing anyway.

Sources/SwiftSyntaxMacroExpansion/MacroExpansion.swift
```diff
@@ -201,3 +211,35 @@ fileprivate extension SyntaxProtocol {
...
+    return C(elems)
+  }
+  var node = node
+  if let attributes = attributes,
+    let withAttrs = node.asProtocol(WithAttributesSyntax.self)
+  {
+    node = withAttrs.with(
+      \.attributes,
+      _combine(attributes, withAttrs.attributes)
+    ).as(DeclSyntax.self)!
```
ahoppen, line 234: You can use `cast(DeclSyntax.self)` to remove the need for the `!`.

Same below.

Sources/SwiftSyntaxMacros/MacroSystem.swift
```diff
@@ -468,6 +476,40 @@ extension MacroApplication {
   }
 }
 
+// FIXME: This was copied from 'SwiftSyntaxMacroExpansion' module.
+// All expansion logics should be unified.
```
ahoppen, line 480: Can’t you make this `@_spi(something) public` and then use it from `SwiftSyntaxMacroExpansion`. `SwiftSyntaxMacroExpansion` already depends on `SwiftSyntaxMacros` AFAICT.

Tests/SwiftSyntaxMacrosTest/MacroSystemTests.swift
```diff
@@ -1037,4 +1058,39 @@ final class MacroSystemTests: XCTestCase {
...
   }
+
+  func testDeclsFromStringLiterals() {
+    assertMacroExpansion(
+      #"""
+      struct S {
+        public #decls(
+          """
+          static func foo() {
+          print("value") }
```
ahoppen, line 1069: Is this intentionally weirdly formatted?

Tests/SwiftSyntaxMacrosTest/MacroSystemTests.swift
```diff
@@ -1037,4 +1058,39 @@ final class MacroSystemTests: XCTestCase {
       indentationWidth: indentationWidth
     )
   }
+
+  func testDeclsFromStringLiterals() {
+    assertMacroExpansion(
```
ahoppen, line 1063: As with my comment in your other PR, I would really prefer this to be three different `assertMacroExpansion` calls, each only having a single `#decl` macro expansion. It just makes debugging so much easier if the test case fails.

Test 1

```swift
struct S {
  public #decls(
    """
    static func foo() {
    print("value") }
    """,
    "struct Inner {\n\n}"
  )
}
```

Test 2

```swift
struct S {
  @attr static #decls("var value1 = 1", "typealias A = B")
}
```

Test 3

```swift
@attribute
@otherAttribute(x: 1) #decls("@moreAttibute var global = 42")
```

Tests/SwiftSyntaxMacrosTest/MacroSystemTests.swift
rintaro: Yes, I wanted to confirm things are correctly expanded from string literals.

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

Tests/SwiftSyntaxMacrosTest/MacroSystemTests.swift
ahoppen: OK 👍🏽

ahoppen (approved):

PR #1668, opened 2023-05-16
Drop `Keyword` suffix from `TokenSpecSet`

ahoppen:
Since we started modelling keywords as a single `RawTokenKind` there hasn’t been an `ifKeyword` etc. anymore. Instead we always refered to them as `Keyword.if` or `.if` everywhere. We should be consistent about this in `TokenSpecSet` as well, which also shouldn’t have a `Keyword` suffix on the cases.

### DIFF
--- a/Sources/SwiftParser/TokenSpecSet.swift
+++ b/Sources/SwiftParser/TokenSpecSet.swift
@@ -74,40 +74,40 @@ enum AccessorKind: TokenSpecSet {
 }
 
 enum CanBeStatementStart: TokenSpecSet {
-  case breakKeyword
-  case continueKeyword
-  case deferKeyword
-  case doKeyword
-  case fallthroughKeyword
-  case forKeyword
-  case forgetKeyword  // NOTE: support for deprecated _forget
-  case discardKeyword
-  case guardKeyword
-  case ifKeyword
-  case repeatKeyword
-  case returnKeyword
-  case switchKeyword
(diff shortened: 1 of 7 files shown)


ahoppen:
@swift-ci Please test

kimdv (approved):
Nice! 
A bit cleaner to read

Sources/SwiftParser/TokenSpecSet.swift
```diff
@@ -74,86 +74,86 @@ enum AccessorKind: TokenSpecSet {
...
-  case switchKeyword
-  case throwKeyword
-  case whileKeyword
+  case `break`
+  case `continue`
+  case `defer`
+  case `do`
+  case `fallthrough`
+  case `for`
+  case forget  // NOTE: support for deprecated _forget
```
bnbarham, line 83: `_forget`?

bnbarham (approved):

Sources/SwiftParser/TokenSpecSet.swift
ahoppen: Yes. Thanks. I missed that

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test macOS

ahoppen:
@swift-ci Please test Windows

PR #1669, opened 2023-05-16
[copy-operator] Add parsing support for copy operator.

gottesmm:
I followed the same model that we used for consume, namely we need an identifier
to copy. This ensures that if we have a function called copy or a variable
called copy, we do not break source stability.

rdar://101862423

### DIFF
--- a/Sources/SwiftParser/Expressions.swift
+++ b/Sources/SwiftParser/Expressions.swift
@@ -481,11 +481,12 @@ extension Parser {
       // what we have done for the consume keyword.
       switch self.peek() {
       case TokenSpec(.identifier, allowAtStartOfLine: false),
-           TokenSpec(.dollarIdentifier, allowAtStartOfLine: false),
-           TokenSpec(.self, allowAtStartOfLine: false):
+        TokenSpec(.dollarIdentifier, allowAtStartOfLine: false),
+        TokenSpec(.self, allowAtStartOfLine: false):
         break
       default:
-        break EXPR_PREFIX // Break out of `outer switch` on failure.
+        // Break out of `outer switch` on failure.
+        break EXPR_PREFIX
       }
 
       let copyTok = self.eat(handle)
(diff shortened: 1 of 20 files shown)


gottesmm:
Swift PR: apple/swift#65935

gottesmm:
https://github.com/apple/swift/pull/65935

@swift-ci test

Sources/SwiftParser/Expressions.swift
```diff
@@ -474,18 +474,51 @@ extension Parser {
...
+      // and the self check.
+      if next.isAtStartOfLine {
+        break
+      }
+
+      if next.rawTokenKind != .identifier,
+         next.rawTokenKind != .dollarIdentifier,
+         !(TokenSpec(.`self`) ~= next) {
+        break
+      }
```
ahoppen, line 494: I’d prefer to use a `TokenSpec` here for all the cases and then just `switch` over `next`. It’s just more consistent with the rest of the code base and the other parsing primitives like `at(anyIn:)` or `canRecoverTo()`

```suggestion 
      switch self.peek() {
      case TokenSpec(.identifier, allowAtStartOfLine: false),
        TokenSpec(.dollarIdentifier, allowAtStartOfLine: false),
        TokenSpec(.self, allowAtStartOfLine: false):
        break
      default:
        break EXPR_PREFIX // And add an EXPR_PREFIX to the outer `switch`
      }
```

Sources/SwiftParser/TokenSpecSet.swift
```diff
@@ -482,6 +482,7 @@ enum AwaitTryMove: TokenSpecSet {
   case _borrowKeyword
   case tryKeyword
   case consumeKeyword
+  case copyKeyword
```
ahoppen, line 485: This `TokenSpecSet` has long since outgrown its name of `AwaitTryMove`. What do you think about renaming it to `ExpressionPrefix`? I’m open to better naming suggestions.

gottesmm:
@ahoppen I decided to also fix consume in this commit as well. I did the same thing as we did with copy/added tests/etc.

gottesmm:
https://github.com/apple/swift/pull/65935

@swift-ci test

gottesmm:
https://github.com/apple/swift/pull/65935

@swift-ci test

gottesmm:
@swift-ci test windows platform

PR #1671, opened 2023-05-16
[5.9] Fix newline parsing at trailing trivia

TTOzzi:
* **Explanation**: Previously, we would not consider newlines inside comments as making a token start on a new line. Because of this we would reject the following, saying that the two elements should be separated by a semicolon
```swift
let a = 1 /*
*/let b = 2
```
* **Scope**: Parsing of tokens that start with a multi-line comment
* **Risk**: Relatively low
* **Reviewer**:  rdar://108762344 / https://github.com/apple/swift-syntax/issues/1626

### DIFF
--- a/Sources/SwiftParser/Lexer/Cursor.swift
+++ b/Sources/SwiftParser/Lexer/Cursor.swift
@@ -321,6 +321,8 @@ extension Lexer {
     /// If `tokenKind` is `.keyword`, the kind of keyword produced, otherwise
     /// `nil`.
     let keywordKind: Keyword?
+    /// Indicates whether the lexed token text contains a newline.
+    let newlinePresence: Lexer.Cursor.NewlinePresence
 
     private init(
       _ tokenKind: RawTokenKind,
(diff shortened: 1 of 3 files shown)


Tests/SwiftParserTest/Assertions.swift
```diff
@@ -12,7 +12,7 @@
 
 import XCTest
 @_spi(RawSyntax) import SwiftSyntax
-@_spi(Testing)@_spi(RawSyntax) import SwiftParser
+@_spi(Testing) @_spi(RawSyntax) import SwiftParser
```
TTOzzi, line 15: I ran format.py before opening the PR, and it made this change 🤔 
Can I include this in this PR?

Tests/SwiftParserTest/DeclarationTests.swift
```diff
@@ -11,7 +11,7 @@
 //===----------------------------------------------------------------------===//
 
 @_spi(RawSyntax) import SwiftSyntax
-@_spi(Testing)@_spi(RawSyntax) import SwiftParser
+@_spi(Testing) @_spi(RawSyntax) import SwiftParser
```
TTOzzi, line 14: This is also the same.

Tests/SwiftParserTest/Assertions.swift
ahoppen: Ah, that’s because the 5.9 `swift-format` doesn’t add a space here. It shouldn’t be included in the PR or CI will complain that the `release/5.9` branch is not formatted correctly according to the `release/5.9` branch of `swift-format`.

bnbarham (approved):

PR #1672, opened 2023-05-16
Make CodeGeneration a lot less string-based

ahoppen:
Because of its Python legacy the CodeGeneration tool has been heavily string-based. This PR fixes that at least to some degree.

The main motivating factors here were:
- Instead of using strings to specify the node type of a child node, define a `SyntaxNodeKind` enum that contains a case for each syntax node. This way you can be sure that you’re not referring to a non-existent syntax node (like we did for e.g. `InOutToken`, which we forgot to change to `.token("inout")`)
- Refactor `Node` so that it contains two initializers: One for collections and one for layout nodes instead of having one initializer that has a bunch of optional and defaulted arguments.
- Change a bunch of properties from returning a `String` to returning a `TypeSyntax` or `TokenSyntax` so we don‘t need to use the `raw:` interpolation style for them.

In general, the API of `Node` and `SyntaxNodeKind` is what I’m happy with now. The other files can still do with some cleanup.

All the changes resulted in nearly no functionality changes of the generated code, so I think this PR doesn’t need a super thorough review, just skimming over it should be sufficient. It’s really repetitive after all.


CodeGeneration/Sources/SyntaxSupport/Child.swift
```diff
@@ -123,19 +118,26 @@ public class Child {
 
   /// Whether this child has syntax kind `UnexpectedNodes`.
   public var isUnexpectedNodes: Bool {
-    syntaxKind == "UnexpectedNodes"
+    switch kind {
+    case .collection(kind: .unexpectedNodes, collectionElementName: _):
```
kimdv, line 122: Would it make sense to just omit `collectionElementName `

CodeGeneration/Sources/SyntaxSupport/CommonNodes.swift
```diff
@@ -13,34 +13,34 @@
...
 
   // code-block-item = (decl | stmt | expr) ';'?
   Node(
-    name: "CodeBlockItem",
+    kind: .codeBlockItem,
+    base: .syntax,
     nameForDiagnostics: nil,
-    description: "A CodeBlockItem is any Syntax node that appears on its own line inside a CodeBlock.",
-    kind: "Syntax",
+    documentation: "A CodeBlockItem is any Syntax node that appears on its own line inside a CodeBlock.",
```
kimdv, line 27: Not part of this, but we could add ticks around `CodeBlockItem` and `CodeBlock` (and properly also many other places) to add doc linking

kimdv (approved):
So nice! WUHU! 😍

Have been thinking a lot on doing this, but you were faster.

Sources/SwiftSyntaxBuilder/generated/BuildableCollectionNodes.swift
```diff
@@ -14,126 +14,126 @@
 
 import SwiftSyntax
 
-/// `AccessorList` represents a collection of `AccessorDeclSyntax`
+/// `AccessorListSyntax` represents a collection of `AccessorDeclSyntax`
```
bnbarham, line 17: These docs don't seem particularly useful to me. I'd expect them on the type itself rather than the extension (and even there... fairly dubious about their value).

bnbarham (approved):

Sources/SwiftSyntaxBuilder/generated/BuildableCollectionNodes.swift
ahoppen: I won’t argue with you about that. I hope that we’ll write proper documentation soon and then these will just automatically go away.

ahoppen:
@swift-ci Please test

PR #1676, opened 2023-05-18
[5.9][Macros] Don't fold operators in unnecessary nodes

rintaro:
Cherry-pick #1657 into `release/5.9`

* **Explanation**: For attached macros, operator folding should only be done for the attribute syntax node, but not for the target declaration syntax and the parent decl syntax. This aligns with the compiler's ASTGen.
* **Scope**: Macro executable plugin
* **Risk**: Low. The change is small and straightforward, this is a new feature so nothing should depends on the old behavior
* **Testing**: Passes current test suite
* **Issue**: rdar://109494050
* **Reviwer**: Ben Barham (@bnbarham), Doug Gregor (@DougGregor)

### DIFF
--- a/Sources/SwiftCompilerPluginMessageHandling/Macros.swift
+++ b/Sources/SwiftCompilerPluginMessageHandling/Macros.swift
@@ -12,6 +12,7 @@
 
 import SwiftBasicFormat
 import SwiftDiagnostics
+import SwiftOperators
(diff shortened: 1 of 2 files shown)


rintaro:
@swift-ci Please test

DougGregor (approved):

PR #1677, opened 2023-05-18
Fix diagnostic of ternary operator missing only colon

### LINKED ISSUE #1624: Wrong diagnostics for ternary expression when only the colon is missing
### Description

When diagnostics are generated for ternary expressions where the `:` is missing but both of the first and second choice expressions are present, the incorrect diagnostic message is used and the wrong Fix-It is applied.

### Steps to Reproduce

Example: 
```swift
foo ? 1 2
```

This results in the following diagnostic message and Fix-It:
```
expected ':' and expression after '? ...' in ternary expression
```
```swift
foo ? 1 : <#expression#>2
```

As the expression isn't missing in the example I expected the following diagnostic message and Fix-It:
```
expected ':' after '? ...' in ternary expression
```
```swift
foo ? 1 : 2
```

TTOzzi:
Resolve #1624 

Previously, in the ternary operator, the absence of a colon always resulted in parsing it as the right-hand side expression being missing.
Added logic to check if the next token is the same kind as the token before the colon when parsing a ternary operator, if there is no colon.

### DIFF
--- a/Sources/SwiftParser/Expressions.swift
+++ b/Sources/SwiftParser/Expressions.swift
@@ -317,7 +317,13 @@ extension Parser {
         arena: self.arena
       )
 
-      return (RawExprSyntax(op), nil)
+      let rhs: RawExprSyntax?
+      if colon.isMissing, currentToken.isAtStartOfLine {
+        rhs = RawExprSyntax(RawMissingExprSyntax(arena: self.arena))
+      } else {
+        rhs = nil
+      }
+      return (RawExprSyntax(op), rhs)
 
     case (.equal, let handle)?:
       switch pattern {
--- a/Sources/SwiftParser/Expressions.swift
+++ b/Sources/SwiftParser/Expressions.swift
@@ -320,7 +320,8 @@ extension Parser {
       let rhs: RawExprSyntax?
       if colon.isMissing {
         if let previousTokenKind = currentToken.cursor.previousTokenKind,
-           self.at(TokenSpec(previousTokenKind)) {
+          self.at(TokenSpec(previousTokenKind))
+        {
           rhs = nil
         } else {
           rhs = RawExprSyntax(RawMissingExprSyntax(arena: self.arena))
--- a/Sources/SwiftParser/Expressions.swift
+++ b/Sources/SwiftParser/Expressions.swift
@@ -318,9 +318,13 @@ extension Parser {
       )
 
       let rhs: RawExprSyntax?
-      if colon.isMissing,
-         currentToken.rawTokenKind != currentToken.cursor.previousTokenKind {
-        rhs = RawExprSyntax(RawMissingExprSyntax(arena: self.arena))
+      if colon.isMissing {
+        if let previousTokenKind = currentToken.cursor.previousTokenKind,
+           self.at(TokenSpec(previousTokenKind)) {
+          rhs = nil
+        } else {
+          rhs = RawExprSyntax(RawMissingExprSyntax(arena: self.arena))
+        }
       } else {
         rhs = nil
       }
--- a/Sources/SwiftParser/Expressions.swift
+++ b/Sources/SwiftParser/Expressions.swift
@@ -318,10 +318,8 @@ extension Parser {
       )
 
       let rhs: RawExprSyntax?
-      if colon.isMissing {
-        // If the colon is missing there's not much more structure we can
-        // expect out of this expression sequence. Emit a missing expression
-        // to end the parsing here.
+      if colon.isMissing,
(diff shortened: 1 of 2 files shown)


Sources/SwiftParser/Expressions.swift
```diff
@@ -319,10 +319,13 @@ extension Parser {
 
       let rhs: RawExprSyntax?
       if colon.isMissing {
-        // If the colon is missing there's not much more structure we can
-        // expect out of this expression sequence. Emit a missing expression
-        // to end the parsing here.
-        rhs = RawExprSyntax(RawMissingExprSyntax(arena: self.arena))
+        if let previousTokenKind = currentToken.cursor.previousTokenKind,
+          self.at(TokenSpec(previousTokenKind))
```
ahoppen, line 323: This will only catch cases where the two tokens that should be separated by the colon are of the same kind. It won’t catch cases like `foo ? 1 someVar`. 

But actually, I’m thinking there’s no reason to ever set `rhs` to a `MissingExprSyntax` (maybe there never was, I don’t know) and just setting it to `nil` seems like the cleaner fix to me.

TTOzzi: Oh, right!
I just checked the behavior of the Swift Compliler in Xcode and it doesn't seem to check for expressions that should be after the colon when there is no colon.
스크린샷 2023-05-20 오후 2 05 15스크린샷 2023-05-20 오후 2 11 06

One difference is that the Xcode Swift Compiler does not suggest adding an expression in the following cases 🤔 
스크린샷 2023-05-20 오후 2 19 21
I think the behavior of SwiftSyntax is more accurate in this case, so I'll keep it.

TTOzzi: In addition, in the case below, we need to check the type of both expressions of the colon, which is not possible with just the tokenKind check.
스크린샷 2023-05-20 오후 2 38 44
Is there an already implemented function to check for type mismatches?

ahoppen: Oh, I now see why you checked for the token kind. 

Let me give you some compiler background: Compilers are usually divided in multiple stages. The interesting stages here are lexing, parsing, name lookup and type checking. 

The purpose of lexing is to take the contents of a file and form tokens from it. For example, it knows that two words separated by a space are two different identifiers instead of one. It also takes care of classifying comments. These are the files in SwiftParser/Lexer

Parsing than takes this stream of tokens and forms a tree from them. And if it can't form a tree from the tokens, the parser will throw an error. Parsing is still what we call syntactical, that is, it doesn't have a clue about types or which declarations a variable, function, etc. references. This is the part of the code base you are working on. 

The next step is name lookup. Name lookup determines the declarations a variable/function/… reference might refer to. This stage can, for example, already take into account when one variable shadows another. Name lookup still doesn't concern itself with type information. If a function is overloaded, for example, it just returns an overload set of functions that might be applicable.

And after that there's type checking, which takes the structure from parsing and the information from type checking to check that all the types match. Because of Swift's rules that you can overload functions based on their return types and other type inference features, this is a whole lot more complicated beast. 

Having said that, name lookup and type checking are not implemented in the swift-syntax repo yet, so there is no way you can check that the types of the arguments match.

Sources/SwiftParser/Expressions.swift
```diff
@@ -317,16 +317,7 @@ extension Parser {
         arena: self.arena
       )
 
-      let rhs: RawExprSyntax?
-      if colon.isMissing {
```
ahoppen, line 321: Contrary to my first comment, I just had a thought. What do we produce if we have the following?

```
condition ? 1
someOtherVariable
```

I think in this case we should be suggesting to add `: <#expression#>` and not just a standalone colon. To achieve that you might need to revert to the old code and just add an additional check for whether the next token is on the same line or a new line.

Sources/SwiftParser/Expressions.swift
TTOzzi: Ah, I see! Now it all makes sense.
I sincerely appreciate your detailed explanation. 
Thanks to you, I was able to understand it easily 🙇‍♂️

Sources/SwiftParser/Expressions.swift
TTOzzi: In the current state of the project, it returns the diagnostic "expected ':' after '? ...' in ternary expression." 
The Swift compiler in Xcode works the same way.
As you mentioned, I think it is appropriate to suggest adding ": <#expression#>" to the code. 
I will proceed to work on it further and add an commit in this PR!

ahoppen (approved):
Nice. This looks very good to me

ahoppen:
@swift-ci Please test

Issue #1608, opened 2023-04-28
`:` is not applied in the correct place for function signature Fix-It


kimdv:
### Description

https://github.com/apple/swift-syntax/pull/1549#discussion_r1179896824
Test case `testInvalid13`

ahoppen:
Tracked in Apple’s issue tracker as rdar://108656824

TiagoMaiaL:
I want to start working in this issue. Is it ok if I get assigned to it? cc: @ahoppen @kimdv

ahoppen:
It’s yours now. Thanks for looking into it.

In my opinion the key here is to detect that we are missing a type annotation and checking whether the internal parameter name starts with an uppercase letter, which usually indicates that it is meant to be a type name. We are already doing something similar here

https://github.com/apple/swift-syntax/blob/d14176d74bff98aa877bbf284a8bccd4bb02c833/Sources/SwiftParser/Types.swift#L551

TiagoMaiaL:
I think we can close this issue 🎉 

Thank you for your assistance, @ahoppen!

PR #1679, opened 2023-05-19
[Parser] Improve FixIt for function parameter with missing type (`name1 Name2`)

TiagoMaiaL:
# Goal

To solve issue https://github.com/apple/swift-syntax/issues/1608.

## Notes

There are some `TODOs` in the code I want to solve but I didn't know how. Any help is appreciated. Thank you.

### DIFF
--- a/Sources/SwiftParser/Parameters.swift
+++ b/Sources/SwiftParser/Parameters.swift
@@ -92,10 +92,30 @@ extension Parser {
     let modifiers = parseParameterModifiers(isClosure: false)
     let misplacedSpecifiers = parseMisplacedSpecifiers()
 
-    let names = self.parseParameterNames()
+    var names = self.parseParameterNames()
     let (unexpectedBeforeColon, colon) = self.expect(.colon)
 
-    let type = self.parseType(misplacedSpecifiers: misplacedSpecifiers)
+    let type: RawTypeSyntax
+
+    if colon.presence == .missing, let secondName = names.secondName, secondName.tokenText.isStartingWithUppercase {
+      // Synthesize the secondName parameter as a type node.
+      type = RawTypeSyntax(
+        RawSimpleTypeIdentifierSyntax(
+          name: secondName,
+          genericArgumentClause: nil,
+          arena: self.arena
+        )
+      )
+      names = ParameterNames(
+        unexpectedBeforeFirstName: names.unexpectedBeforeFirstName,
+        firstName: names.firstName,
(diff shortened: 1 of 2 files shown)


Sources/SwiftParser/Parameters.swift
```diff
@@ -92,10 +92,39 @@ extension Parser {
...
+    if colon.presence == .missing, let secondName = names.secondName, secondName.tokenText.isStartingWithUppercase {
+      // Synthesize the secondName parameter as a type node.
+      type = RawTypeSyntax(
+        RawSimpleTypeIdentifierSyntax(
+          name: RawTokenSyntax(
+            kind: .identifier,
+            text: secondName.tokenText,
+            presence: .missing,
+            arena: self.arena
+          ),
```
ahoppen, line 109: You should be able to just use `secondName` here. There’s no need to construct a new `RawTokenSyntax`. That way, you also retain any trivia that’s attached to `secondName`. 

```suggestion
          name: secondName,
```

If you do that, I think you also don’t need to add `names.secondName` to `unexpectedBeforeColon`. And you might also not need to write any custom code in `ParseDiagnsoticGenerator` at all.

Tests/SwiftParserTest/translated/InvalidTests.swift
```diff
@@ -305,14 +305,17 @@ final class InvalidTests: XCTestCase {
   }
 
   func testInvalid13() {
+    // TODO: Determine if the comment below should be removed.
     // https://github.com/apple/swift/issues/43190
     // Crash with invalid parameter declaration
+    
+    // TODO: Remove the trailing trivia from the firstName node (`s` and space).
```
ahoppen, line 312: This might actually resolve itself if you apply my other comment above and if you don’t need special handling in `ParseDiagnosticGenerator`.

Sources/SwiftParser/Parameters.swift
TiagoMaiaL: Worked pretty well, thank you.

TiagoMaiaL:
Do we need the comment in `InvalidTests.testInvalid13`?
```swift
  func testInvalid13() {
    // https://github.com/apple/swift/issues/43190
    // Crash with invalid parameter declaration
```

ahoppen (approved):
Thank you for your contribution @TiagoMaiaL. This looks very good to me!

ahoppen:
@swift-ci Please test

PR #1682, opened 2023-05-20
[CMake] Updates to allow inclusion using FetchContent

bnbarham:
Various updates that allow swift-syntax to be included using FetchContent.

Mostly this is just not replacing `target_link_libraries` since that is a global replacement, but we also now use a couple of variables from the swift if they're set (eg. `SWIFT_HOST_LIBRARIES_DEST_DIR` and `SWIFT_HOST_MODULE_TRIPLE`).


cmake/modules/AddSwiftHostLibrary.cmake
```diff
@@ -6,22 +6,44 @@
...
+      )
+    target_sources(${TARGET} PRIVATE
+      ${CMAKE_CURRENT_BINARY_DIR}/forced-${DEPENDENCY}-dep.swift
+    )
+  endforeach()
+endfunction()
+
 # Add a new host library with the given name.
-function(add_swift_host_library name)
+function(add_swift_syntax_library name)
```
etcwilde, line 35: Is this function only being used for SwiftSyntax, or will we want to pull in other things this way too?

etcwilde: Oh, this is the swift-syntax repo, don't mind me!

bnbarham: It's only for swift-syntax, though theoretically we could replace the swift functions with functions defined here instead. 🤷

bnbarham:
Updated to remove export if not a top level project + add to SWIFT_EXPORTS otherwise.

DougGregor (approved):
This all looks reasonable to me

Sources/SwiftIDEUtils/CMakeLists.txt
```diff
@@ -6,12 +6,13 @@
 # See http://swift.org/LICENSE.txt for license information
 # See http://swift.org/CONTRIBUTORS.txt for Swift project authors
 
-add_swift_host_library(SwiftIDEUtils
+add_swift_syntax_library(SwiftIDEUtils
+  generated/SyntaxClassification.swift
```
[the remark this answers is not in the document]
bnbarham: No, this will have been a bad conflict resolution

PR #1684, opened 2023-05-21
Use marker to mark `SourceEdit`s

StevenWong12:
This change is used to help us mark the changes in source file, which makes incremental parsing test easier.

See comments in https://github.com/apple/swift-syntax/pull/1662#issuecomment-1556110495

### DIFF
--- a/Sources/_SwiftSyntaxTestSupport/SourceEditsTestUtils.swift
+++ b/Sources/_SwiftSyntaxTestSupport/SourceEditsTestUtils.swift
@@ -1,0 +1,94 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+import XCTest
+
+/// Get `ConcurrentEdits` in source whose edited zones are marked with markers
+/// Also extract the markers from source to get original source and edited source
+///
+/// `⏩️` is  *start marker*, `⏸️` is *separate marker*, `⏪️` is *end marker*
+/// Contents between `⏩️` and `⏸️` are source text that before modification, contents
+/// betwwen `⏸️` and `⏪️` are source text that after modification
+/// i.e. `⏩️foo⏸️bar⏪️`, the original source is `foo` and the edited source is `bar`
+public func getEditsAndSources(_ source: String) -> (edits: ConcurrentEdits, orignialSource: Substring, editedSource: Substring) {
+  var editedSource = Substring()
+  var originalSource = Substring()
+  var concurrentEdits: [SourceEdit] = []
+
+  var lastStartIndex = source.startIndex
+  while let startIndex = source[lastStartIndex...].firstIndex(where: { $0 == "⏩️" }),
+    let separateIndex = source[startIndex...].firstIndex(where: { $0 == "⏸️" }),
+    let endIndex = source[separateIndex...].firstIndex(where: { $0 == "⏪️" })
+  {
+
+    originalSource += source[lastStartIndex..<startIndex]
+    let edit = SourceEdit(
+      offset: originalSource.utf8.count,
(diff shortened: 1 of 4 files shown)


Sources/_SwiftSyntaxTestSupport/SourceEditsTestUtils.swift
```diff
@@ -0,0 +1,99 @@
...
+    editedSource = afterSource
+  }
+
+  return (ConcurrentEdits(fromSequential: sequentialEdits), extractOriginalSource(from: source), editedSource)
+}
+
+fileprivate func nextMarkedEdits(_ source: String) -> (edit: SourceEdit, editedSource: String)? {
+  guard let startIndex = source.firstIndex(where: { $0.isStartMarker }),
+    let separateIndex = source.firstIndex(where: { $0.isSeperateMarker }),
+    let endIndex = source.firstIndex(where: { $0.isEndMarker })
```
ahoppen, line 33: Technically, I think you want to find the first separate marker after the start marker, and end marker after the separate marker. Otherwise, you can get into weird issues of ranges where the end location is before the start location when someone uses an incorrect format string like `⏩️class⏪️struct⏸️ foo {`.

```suggestion
    let separateIndex = source[startIndex...].firstIndex(where: { $0.isSeperateMarker }),
    let endIndex = source[separateIndex...].firstIndex(where: { $0.isEndMarker })
```

ahoppen (commented):
With your comment on the other PR (https://github.com/apple/swift-syntax/pull/1662#issuecomment-1556110495), I think being explicit about the edits makes a lot of sense. Thanks for doing this. I have a few comments inline but the design looks good overall.

Sources/_SwiftSyntaxTestSupport/SourceEditsTestUtils.swift
```diff
@@ -0,0 +1,98 @@
...
+/// Get `ConcurrentEdits` in source whose edited zones are marked with markers
+/// Also extract the markers from source to get original source and edited source
+///
+/// `⏩️` is  *start marker*, `⏸️` is *separate marker*, `⏪️` is *end marker*
+/// Contents between `⏩️` and `⏸️` are source text that before modification, contents
+/// betwwen `⏸️` and `⏪️` are source text that after modification
+/// i.e. `⏩️foo⏸️bar⏪️`, the original source is `foo` and the edited source is `bar`
+public func getEditsAndSources(_ source: String) -> (edits: ConcurrentEdits, orignialSource: Substring, editedSource: Substring) {
+  var editedSource = source.prefix(0)
+  var originalSource = source.prefix(0)
```
ahoppen, line 25: I assume you want to create empty `Substring`s here, right? Can’t you just do the following?

```suggestion
  var editedSource = Substring()
  var originalSource = Substring()
```

Sources/_SwiftSyntaxTestSupport/SourceEditsTestUtils.swift
```diff
@@ -0,0 +1,98 @@
...
+  var concurrentEdits: [SourceEdit] = []
+
+  var lastStartIndex = source.startIndex
+  while let startIndex = source[lastStartIndex...].firstIndex(where: { $0.isStartMarker }),
+    let separateIndex = source[startIndex...].firstIndex(where: { $0.isSeperateMarker }),
+    let endIndex = source[separateIndex...].firstIndex(where: { $0.isEndMarker })
```
ahoppen, line 31: Out of curiosity: Why did you choose to move `isStartMarker` etc into an extension of `Character` instead of just inlining that check here? Just seems like an unnecessary abstraction layer to me if we aren’t calling the computed properties anywhere else.

ahoppen (commented):
I’ve got a few code style comments, otherwise this looks very good to me.

Sources/_SwiftSyntaxTestSupport/SourceEditsTestUtils.swift
StevenWong12: I thought that maybe these computed properties would be called elsewhere then. But it's just an unnecessary abstraction layer now. Already changed that.

ahoppen:
@swift-ci Please test

ahoppen (approved):
Great. Let’s get this merged

ahoppen:
CI failed for an unrelated reason. I’ll re-trigger it tomorrow

ahoppen:
@swift-ci Please test macOS

ahoppen:
@swift-ci Please test Windows

PR #1685, opened 2023-05-21
Port incremental parsing ability to `CodeBlockItem`

StevenWong12:
I want to make this pr as a start effort to port incremental parsing ability to `SwiftParser`.

### DIFF
--- a/Sources/SwiftParser/IncrementalParseTransition.swift
+++ b/Sources/SwiftParser/IncrementalParseTransition.swift
@@ -11,6 +11,30 @@
 //===----------------------------------------------------------------------===//
 
 @_spi(RawSyntax) import SwiftSyntax
+
+extension Parser {
+  mutating func loadCurrentSyntaxNodeFromCache(for kind: SyntaxKind) -> Syntax? {
+    guard parseLookup != nil else {
+      return nil
+    }
+
+    let currentOffset = self.lexemes.getOffsetToStart(self.currentToken)
+    if let node = parseLookup!.lookUp(currentOffset, kind: kind) {
+      self.lexemes.advance(by: node.byteSize, currentToken: &self.currentToken)
+      return node
+    }
+
+    return nil
+  }
+
+  mutating func registerNodeForIncrementalParse(node: RawSyntax, startToken: Lexer.Lexeme) {
+    lookaheadRanges.registerNodeForIncrementalParse(
+      node: node,
+      lookaheadLength: lexemes.lookaheadTracker.pointee.furthestOffset - self.lexemes.getOffsetToStart(startToken)
+    )
+  }
+}
+
 /// Accepts the re-used ``Syntax`` nodes that `IncrementalParseTransition`
 /// determined they should be re-used for a parse invocation.
 ///
@@ -20,13 +44,12 @@
 /// This is also used for testing purposes to ensure incremental reparsing
 /// worked as expected.
 public protocol IncrementalParseReusedNodeDelegate {
-  /// Accepts the range and ``Syntax`` node of skipped source region.
+  /// Accepts ``Syntax`` node of skipped source region.
   ///
   /// - Parameters:
-  ///   - range: The source region of the currently parsed source.
   ///   - previousNode: The node from the previous tree that is associated with
   ///                   the skipped source region.
-  func parserReusedNode(range: ByteSourceRange, previousNode: Syntax)
+  func parserReusedNode(previousNode: Syntax)
 }
 
 /// An implementation of `IncrementalParseReusedNodeDelegate` that just collects
@@ -34,12 +57,12 @@ public protocol IncrementalParseReusedNodeDelegate {
 public final class IncrementalParseReusedNodeCollector:
   IncrementalParseReusedNodeDelegate
 {
-  public var rangeAndNodes: [(ByteSourceRange, Syntax)] = []
+  public var nodes: [Syntax] = []
 
   public init() {}
 
-  public func parserReusedNode(range: ByteSourceRange, previousNode: Syntax) {
-    rangeAndNodes.append((range, previousNode))
+  public func parserReusedNode(previousNode: Syntax) {
+    nodes.append(previousNode)
   }
 }
 
@@ -48,6 +71,7 @@ public final class IncrementalParseReusedNodeCollector:
 public final class IncrementalParseTransition {
   fileprivate let previousTree: SourceFileSyntax
   fileprivate let edits: ConcurrentEdits
+  fileprivate let lookaheadRanges: LookaheadRanges
   fileprivate let reusedDelegate: IncrementalParseReusedNodeDelegate?
 
   /// - Parameters:
@@ -59,17 +83,19 @@ public final class IncrementalParseTransition {
   public init(
     previousTree: SourceFileSyntax,
     edits: ConcurrentEdits,
+    lookaheadRanges: LookaheadRanges,
     reusedNodeDelegate: IncrementalParseReusedNodeDelegate? = nil
   ) {
     self.previousTree = previousTree
     self.edits = edits
+    self.lookaheadRanges = lookaheadRanges
     self.reusedDelegate = reusedNodeDelegate
   }
 }
 
 /// Provides a mechanism for the parser to skip regions of an incrementally
 /// updated source that was already parsed during a previous parse invocation.
-public struct IncrementalParseLookup {
+struct IncrementalParseLookup {
   fileprivate let transition: IncrementalParseTransition
   fileprivate var cursor: SyntaxCursor
 
@@ -100,8 +126,7 @@ public struct IncrementalParseLookup {
   /// - Returns: A ``Syntax`` node from the previous parse invocation,
   ///            representing the contents of this region, if it is still valid
   ///            to re-use. `nil` otherwise.
-  @_spi(RawSyntax)
-  public mutating func lookUp(_ newOffset: Int, kind: SyntaxKind) -> Syntax? {
+  fileprivate mutating func lookUp(_ newOffset: Int, kind: SyntaxKind) -> Syntax? {
     guard let prevOffset = translateToPreEditOffset(newOffset) else {
       return nil
     }
@@ -109,14 +134,13 @@ public struct IncrementalParseLookup {
     let node = cursorLookup(prevPosition: prevPosition, kind: kind)
     if let delegate = reusedDelegate, let node {
       delegate.parserReusedNode(
-        range: ByteSourceRange(offset: newOffset, length: node.byteSize),
         previousNode: node
       )
     }
     return node
   }
 
-  mutating fileprivate func cursorLookup(
+  fileprivate mutating func cursorLookup(
     prevPosition: AbsolutePosition,
     kind: SyntaxKind
   ) -> Syntax? {
@@ -148,24 +172,13 @@ public struct IncrementalParseLookup {
       return true
     }
 
-    // Node can also not be reused if an edit has been made in the next token's
-    // text, e.g. because `private struct Foo {}` parses as a CodeBlockItem with
-    // a StructDecl inside and `private struc Foo {}` parses as two
-    // CodeBlockItems one for `private` and one for `struc Foo {}`
-    var nextLeafNodeLength: SourceLength = .zero
-    if let nextSibling = cursor.nextSibling {
-      // Fast path check: if next sibling is before all the edits then we can
-      // re-use the node.
-      if !edits.edits.isEmpty && edits.edits.first!.range.offset > nextSibling.endPosition.utf8Offset {
-        return true
-      }
-      if let nextToken = nextSibling.firstToken(viewMode: .sourceAccurate) {
-        nextLeafNodeLength = nextToken.leadingTriviaLength + nextToken.contentLength
-      }
+    guard let nodeAffectRangeLength = transition.lookaheadRanges.lookaheadRanges[node.raw.id] else {
+      return false
     }
+
     let nodeAffectRange = ByteSourceRange(
       offset: node.position.utf8Offset,
-      length: (node.totalLength + nextLeafNodeLength).utf8Length
+      length: nodeAffectRangeLength
     )
 
     for edit in edits.edits {
--- a/Sources/SwiftParser/Lexer/LexemeSequence.swift
+++ b/Sources/SwiftParser/Lexer/LexemeSequence.swift
@@ -32,10 +32,23 @@ extension Lexer {
     /// usually less than 0.1% of the memory allocated by the syntax arena.
     var lexerStateAllocator = BumpPtrAllocator(slabSize: 256)
 
-    fileprivate init(sourceBufferStart: Lexer.Cursor, cursor: Lexer.Cursor) {
+    /// The offset of the trailing trivia end of `nextToken` relative to the source buffer’s start.
+    var offsetToNextTokenEnd: Int {
+      self.getOffsetToStart(self.nextToken) + self.nextToken.byteLength
+    }
+
+    /// See doc comments in ``LookaheadTracker``
+    ///
+    /// This is an `UnsafeMutablePointer` for two reasons
+    ///  - When `LexemeSequence` gets copied (e.g. when a ``Lookahead`` gets created), it should still reference the same ``LookaheadTracker`` so that any lookahead performed in the ``Lookahead`` also affects the original ``Parser``. It thus needs to be a reference type
+    ///  - ``LookaheadTracker`` is not a class to avoid reference counting it. The ``Parser`` that creates the ``LexemeSequence`` will always outlive any ``Lookahead`` created for it.
+    let lookaheadTracker: UnsafeMutablePointer<LookaheadTracker>
+
+    fileprivate init(sourceBufferStart: Lexer.Cursor, cursor: Lexer.Cursor, lookaheadTracker: UnsafeMutablePointer<LookaheadTracker>) {
       self.sourceBufferStart = sourceBufferStart
       self.cursor = cursor
       self.nextToken = self.cursor.nextToken(sourceBufferStart: self.sourceBufferStart, stateAllocator: lexerStateAllocator)
+      self.lookaheadTracker = lookaheadTracker
     }
 
     @_spi(Testing)
@@ -43,11 +56,34 @@ extension Lexer {
       return self.advance()
     }
 
+    /// Record the offset of the end of `nextToken` as the furthest offset in ``LookaheadTracker``
+    private func recordNextTokenInLookaheadTracker() {
+      self.lookaheadTracker.pointee.recordFurthestOffset(self.offsetToNextTokenEnd)
+    }
+
     mutating func advance() -> Lexer.Lexeme {
       defer {
         self.nextToken = self.cursor.nextToken(sourceBufferStart: self.sourceBufferStart, stateAllocator: lexerStateAllocator)
       }
+      self.recordNextTokenInLookaheadTracker()
       return self.nextToken
+    }
+
+    /// Get the offset of the leading trivia start of `token` relative to `sourceBufferStart`.
+    func getOffsetToStart(_ token: Lexer.Lexeme) -> Int {
+      return self.sourceBufferStart.distance(to: token.cursor)
+    }
+
+    /// Advance the the cursor by `offset` and reset `currentToken`
+    ///
+    /// - Important: This should only be used for incremental parsing.
+    mutating func advance(by offset: Int, currentToken: inout Lexer.Lexeme) {
+      self.cursor = currentToken.cursor
+      self.cursor.position = self.cursor.position.advanced(by: offset)
+
+      self.nextToken = self.cursor.nextToken(sourceBufferStart: self.sourceBufferStart, stateAllocator: lexerStateAllocator)
+
+      currentToken = self.advance()
     }
 
     /// Reset the lexeme sequence to the state we were in when lexing `splitToken`
@@ -63,6 +99,7 @@ extension Lexer {
     }
 
     func peek() -> Lexer.Lexeme {
+      self.recordNextTokenInLookaheadTracker()
       return self.nextToken
     }
 
@@ -104,12 +141,13 @@ extension Lexer {
   @_spi(Testing)
   public static func tokenize(
     _ input: UnsafeBufferPointer<UInt8>,
-    from startIndex: Int = 0
+    from startIndex: Int = 0,
+    lookaheadTracker: UnsafeMutablePointer<LookaheadTracker>
   ) -> LexemeSequence {
     precondition(input.isEmpty || startIndex < input.endIndex)
     let startChar = startIndex == input.startIndex ? UInt8(ascii: "\0") : input[startIndex - 1]
     let start = Cursor(input: input, previous: UInt8(ascii: "\0"))
     let cursor = Cursor(input: UnsafeBufferPointer(rebasing: input[startIndex...]), previous: startChar)
-    return LexemeSequence(sourceBufferStart: start, cursor: cursor)
+    return LexemeSequence(sourceBufferStart: start, cursor: cursor, lookaheadTracker: lookaheadTracker)
   }
 }
--- a/Sources/SwiftParser/Parser.swift
+++ b/Sources/SwiftParser/Parser.swift
@@ -101,6 +101,15 @@ public struct Parser {
   /// When this nesting level is exceeded, the parser should stop parsing.
   let maximumNestingLevel: Int
 
+  /// See comments in ``IncrementalParseLookup``
+  var parseLookup: IncrementalParseLookup?
+
+  /// See comments in ``LookaheadRanges``
+  public internal(set) var lookaheadRanges = LookaheadRanges()
+
+  /// Parser should own a ``LookaheadTracker`` so that we can share one `furthestOffset` in a parse.
+  let lookaheadTrackerOwner = LookaheadTrackerOwner()
+
   /// A default maximum nesting level that is used if the client didn't
   /// explicitly specify one. Debug builds of the parser comume a lot more stack
   /// space and thus have a lower default maximum nesting level.
@@ -111,7 +120,11 @@ public struct Parser {
   #endif
 
   /// Initializes a ``Parser`` from the given string.
-  public init(_ input: String, maximumNestingLevel: Int? = nil) {
+  public init(
+    _ input: String,
+    maximumNestingLevel: Int? = nil,
+    parseTransition: IncrementalParseTransition? = nil
+  ) {
     self.maximumNestingLevel = maximumNestingLevel ?? Self.defaultMaximumNestingLevel
 
     self.arena = ParsingSyntaxArena(
@@ -124,8 +137,13 @@ public struct Parser {
       return arena.internSourceBuffer(buffer)
     }
 
-    self.lexemes = Lexer.tokenize(interned)
+    self.lexemes = Lexer.tokenize(interned, lookaheadTracker: lookaheadTrackerOwner.lookaheadTracker)
     self.currentToken = self.lexemes.advance()
+    if let parseTransition {
+      self.parseLookup = IncrementalParseLookup(transition: parseTransition)
+    } else {
+      self.parseLookup = nil
+    }
   }
 
   /// Initializes a ``Parser`` from the given input buffer.
@@ -142,7 +160,12 @@ public struct Parser {
   ///            arena is created automatically, and `input` copied into the
   ///            arena. If non-`nil`, `input` must be within its registered
   ///            source buffer or allocator.
-  public init(_ input: UnsafeBufferPointer<UInt8>, maximumNestingLevel: Int? = nil, arena: ParsingSyntaxArena? = nil) {
+  public init(
+    _ input: UnsafeBufferPointer<UInt8>,
+    maximumNestingLevel: Int? = nil,
+    parseTransition: IncrementalParseTransition? = nil,
+    arena: ParsingSyntaxArena? = nil
+  ) {
     self.maximumNestingLevel = maximumNestingLevel ?? Self.defaultMaximumNestingLevel
 
     var sourceBuffer: UnsafeBufferPointer<UInt8>
@@ -157,8 +180,13 @@ public struct Parser {
       sourceBuffer = self.arena.internSourceBuffer(input)
     }
 
-    self.lexemes = Lexer.tokenize(sourceBuffer)
+    self.lexemes = Lexer.tokenize(sourceBuffer, lookaheadTracker: lookaheadTrackerOwner.lookaheadTracker)
     self.currentToken = self.lexemes.advance()
+    if let parseTransition {
+      self.parseLookup = IncrementalParseLookup(transition: parseTransition)
+    } else {
+      self.parseLookup = nil
+    }
   }
 
   mutating func missingToken(_ kind: RawTokenKind, text: SyntaxText? = nil) -> RawTokenSyntax {
@@ -627,5 +655,47 @@ extension Parser {
       RawTokenSyntax(missing: .period, arena: arena),
       afterContainsAnyNewline
     )
+  }
+}
+
+/// Record the furthest offset to `sourceBufferStart` that is reached by  `LexemeSequence.advance()` or `LexemeSequence.peek()`.
+public struct LookaheadTracker {
+  private(set) var furthestOffset: Int = 0
+
+  public init() {}
+
+  mutating func recordFurthestOffset(_ furthestOffset: Int) {
+    /// We could lookahead multi-times to find different valid part of a node, so we should take the maximum of the lookahead offset as the possible affect range of a node.
+    self.furthestOffset = max(furthestOffset, self.furthestOffset)
+  }
+}
+
+/// Owns a ``LookaheadTracker``.
+///
+/// Once the `LookeaheadTrackerOwner` is deinitialized, the ``LookaheadTracker`` is also destroyed.
+class LookaheadTrackerOwner {
+  var lookaheadTracker: UnsafeMutablePointer<LookaheadTracker>
+
+  init() {
+    self.lookaheadTracker = .allocate(capacity: 1)
+    self.lookaheadTracker.initialize(to: LookaheadTracker())
+  }
+
+  deinit {
+    self.lookaheadTracker.deallocate()
+  }
+}
+
+/// Record the lookahead ranges for syntax nodes.
+public struct LookaheadRanges {
+  /// For each node that is recorded for re-use, the number of UTF-8 bytes that the parser looked ahead to parse the node, measured from the start of the node’s leading trivia.
+  ///
+  /// This information can be used to determine whether a node can be reused in incremental parse. A node can only be re-used if no byte in its looked range has changed.
+  var lookaheadRanges: [RawSyntax.ID: Int] = [:]
+
+  public init() {}
+
+  mutating func registerNodeForIncrementalParse(node: RawSyntax, lookaheadLength: Int) {
+    self.lookaheadRanges[node.id] = lookaheadLength
   }
 }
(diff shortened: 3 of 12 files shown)


StevenWong12:
To prevent incremental parse from missing the legal parts of re-used nodes ahead, I made following changes:
* Add a property `firstPotentialTokenKind` to `SyntaxKind` to represent the potential introducer for each `SyntaxKind`.
* Add a property `syntaxLayout` to `SyntaxKind` to represent the layout of nodes in the form of `SyntaxKind`.
* Add a property `nextPotentialTokenChoices` to `RawSyntax` to compute the next potential token of a certain node when it is initialized. This mechanism is supported by the `syntaxLayout` and `firstPotentialTokenKind ` above.
* Judge whether a node can be reused according to whether `nextPotentialTokenChoices` contains `lookahead.currentToken.rawTokenKind`.

This is still WIP but I am open to feedback.

StevenWong12:
I think this pr is ready for review. 

I will squash the commits if there aren't any big problems with the design.

CodeGeneration/Sources/SyntaxSupport/Node.swift
```diff
@@ -85,6 +85,70 @@ public class Node {
     }
   }
 
+  /// Get the first potential token of a node recursively
+  /// This information is used by incremental parse
+  public var firstPotentialToken: [TokenChoice] {
+    let getChildTokenChoice: (Child) -> [TokenChoice] = { child in
```
StevenWong12, line 91: Maybe we could move this recursion logic to `IncrementalParseUtilsFile.swift`. We only produce `[TokenChoice]` for child whose kind is `.token` and make `.collection` and `.nodeChoices` in the form of `SyntaxKind.someKind.firstPotentialToken + SyntaxKind.anotherKind.firstPotentialToken + ...` in `IncrementalParseUtils.swift`. 

This should reduce the code size. What do you think?

Sources/SwiftParser/IncrementalParseTransition.swift
```diff
@@ -46,9 +46,13 @@ public final class IncrementalParseReusedNodeCollector:
...
 /// occurred since it was created.
 public final class IncrementalParseTransition {
-  fileprivate let previousTree: SourceFileSyntax
-  fileprivate let edits: ConcurrentEdits
-  fileprivate let reusedDelegate: IncrementalParseReusedNodeDelegate?
+  fileprivate var previousTree: SourceFileSyntax?
+  fileprivate var edits: ConcurrentEdits?
+  fileprivate var reusedDelegate: IncrementalParseReusedNodeDelegate?
+
+  public var furthestOffset: Int?
```
ahoppen, line 53: Split `IncrementalParseTransition` into two types as discussed offline.

Sources/SwiftParser/IncrementalParseTransition.swift
```diff
@@ -57,36 +61,57 @@ public final class IncrementalParseTransition {
...
+  public func isValidTransition() -> Bool {
+    return previousTree != nil
+  }
 }
 
 /// Provides a mechanism for the parser to skip regions of an incrementally
 /// updated source that was already parsed during a previous parse invocation.
 public struct IncrementalParseLookup {
   fileprivate let transition: IncrementalParseTransition
-  fileprivate var cursor: SyntaxCursor
```
ahoppen, line 74: As discussed offline: Is it faster to have `cursor` as a member of `IncrementalParseLookup` or always generate it in `cursorLookup`. I would expect the former to be faster, but I might be wrong.

Sources/SwiftParser/IncrementalParseTransition.swift
```diff
@@ -109,7 +134,7 @@ public struct IncrementalParseLookup {
     let node = cursorLookup(prevPosition: prevPosition, kind: kind)
     if let delegate = reusedDelegate, let node {
       delegate.parserReusedNode(
-        range: ByteSourceRange(offset: newOffset, length: node.byteSize),
+        range: ByteSourceRange(offset: node.positionAfterSkippingLeadingTrivia.utf8Offset, length: node.byteSizeAfterTrimmingTrivia),
```
ahoppen, line 137: Maybe the reuse delegate can just receive the reused node instead of the reused range now that everything is implemented in Swift.

Sources/SwiftParser/Lexer/LexemeSequence.swift
```diff
@@ -50,6 +50,19 @@ extension Lexer {
       return self.nextToken
     }
 
+    func getOffsetToStart(_ token: Lexer.Lexeme) -> Int {
+      return self.sourceBufferStart.distance(to: token.cursor)
+    }
```
ahoppen, line 55: Not part of this PR: Can we use `getOffsetToStart` instead of `offset(of:)`.

Sources/SwiftParser/Lookahead.swift
```diff
@@ -27,22 +27,30 @@ extension Parser {
     /// i.e. how far it looked ahead.
     var tokensConsumed: Int = 0
 
+    var parseTransition: IncrementalParseTransition?
```
ahoppen, line 30: As discussed offline, we could store `furthestOffset` in a standalone `struct` and pass that over here as an `UnsafeMutablePointer`.

Sources/SwiftParser/Lookahead.swift
```diff
@@ -27,22 +27,30 @@ extension Parser {
...
+    /// Initialize the offset to the start of source
+    var currentOffset: Int
```
ahoppen, line 32: We can probably delete this.

Sources/SwiftParser/Parser.swift
```diff
@@ -142,7 +145,12 @@ public struct Parser {
   ///            arena is created automatically, and `input` copied into the
   ///            arena. If non-`nil`, `input` must be within its registered
   ///            source buffer or allocator.
-  public init(_ input: UnsafeBufferPointer<UInt8>, maximumNestingLevel: Int? = nil, arena: ParsingSyntaxArena? = nil) {
+  public init(
+    _ input: UnsafeBufferPointer<UInt8>,
+    maximumNestingLevel: Int? = nil,
+    arena: ParsingSyntaxArena? = nil,
+    parseTransition: IncrementalParseTransition? = nil
```
ahoppen, line 152: I would put `parseTransition` before `arena` because `arena` is always the last parameter.

Sources/swift-parser-cli/Commands/PerformanceTest.swift
```diff
@@ -39,10 +40,15 @@ struct PerformanceTest: ParsableCommand {
...
+    let parseTransition = IncrementalParseTransition()
+    var previousTreeDict: [Data:SourceFileSyntax] = [:]
+    for iter in 0..<self.iterations {
       for file in files {
+        if iter != 0 {
+          parseTransition.setupTransition(tree: previousTreeDict[file]!, edits: ConcurrentEdits(fromSequential: []))
+        }
         file.withUnsafeBytes { buf in
-          _ = Parser.parse(source: buf.bindMemory(to: UInt8.self))
+          previousTreeDict[file] = Parser.parse(source: buf.bindMemory(to: UInt8.self), parseTransition: parseTransition)
```
ahoppen, line 51: I would add a flag to decide whether we want to do incremental parsing on subsequent iterations.

Sources/SwiftParser/IncrementalParseTransition.swift
StevenWong12: I think making `cursor` a member of `IncrementalParseLookup` is a little bit faster than generating it in `cursorLookup`.

Sources/SwiftParser/IncrementalParseTransition.swift
StevenWong12: Yeah, I've made this change.

Sources/SwiftParser/IncrementalParseTransition.swift
StevenWong12: I add a `IncrementalParseNodeAffectRangeCollector` to collect this and make it as a trigger to enable incremental parse.

Sources/SwiftParser/Lookahead.swift
StevenWong12: Since we have to record the offset in both `Parser.peek()` and `Lookahead.consumeAnyToken()`, I make this pointer a member of `LexemeSequence`.

Sources/SwiftParser/Parser.swift
```diff
@@ -629,3 +649,49 @@ extension Parser {
...
+
+    return nil
+  }
+
+  func registerNodeForIncrementalParse(node: RawSyntax, startToken: Lexer.Lexeme) {
+    if let parseNodeAffectRange {
+      parseNodeAffectRange.registerNodeForIncrementalParse(
+        node: node,
+        length: max(lookaheadFurthestOffset - self.lexemes.getOffsetToStart(startToken), node.byteLength + currentToken.byteLength)
+      )
```
StevenWong12, line 676: I use `node.byteLength + currentToken.byteLength` here since we may encounter some cases that we lookahead only when we are at some specific tokens. i.e. we lookahead to find valid trailing closure only when we are at `leftBrace`.

Sources/SwiftParser/IncrementalParseTransition.swift
```diff
@@ -26,29 +26,43 @@ public protocol IncrementalParseReusedNodeDelegate {
...
 
 /// Keeps track of a previously parsed syntax tree and the source edits that
 /// occurred since it was created.
 public final class IncrementalParseTransition {
-  fileprivate let previousTree: SourceFileSyntax
-  fileprivate let edits: ConcurrentEdits
-  fileprivate let reusedDelegate: IncrementalParseReusedNodeDelegate?
+  fileprivate var previousTree: SourceFileSyntax
+  fileprivate var edits: ConcurrentEdits
+  fileprivate var reusedDelegate: IncrementalParseReusedNodeDelegate?
```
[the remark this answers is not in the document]
StevenWong12: Ah, these are redundant. I've changed back to `let`.

Sources/SwiftParser/Parser.swift
```diff
@@ -629,3 +649,49 @@ extension Parser {
...
+
+// MARK: Incremental Parsing
+extension Parser {
+  mutating func loadCurrentSyntaxNodeFromCache(for kind: SyntaxKind) -> Syntax? {
+    guard let parseTransition = self.parseTransition,
+      let parseNodeAffectRange = self.parseNodeAffectRange
+    else {
+      return nil
+    }
+    var lookUpHelper = IncrementalParseLookup(transition: parseTransition, nodeAffectRangeCollector: parseNodeAffectRange)
```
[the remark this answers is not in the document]
StevenWong12: Yeah, I've made `IncrementalParseLookup` a member and removed `IncrementalParseTransition`. That looks good.

Tests/SwiftParserTest/IncrementalParsingTests.swift
```diff
@@ -144,7 +143,10 @@ public class IncrementalParsingTests: XCTestCase {
     assertIncrementalParse(
       """
       self = ⏩️⏸️_                            _⏪️foo(1)[object1, object2] + o bar(1)
-      """
+      """,
+      reusedNodes: [
+        ReusedNodeSpec("bar(1)", kind: .codeBlockItem)
+      ]
```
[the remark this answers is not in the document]
StevenWong12: Ah, the `o ` seems very much like a typo here to me too. I've adjusted the test case, `bar(1)` is no longer re-usable though.

Sources/swift-parser-cli/Commands/PerformanceTest.swift
```diff
@@ -37,19 +41,39 @@ struct PerformanceTest: ParsableCommand {
...
       for file in files {
+        if incrementalParse == true {
+          incrementalParseAffectRangeCollector = previousTreeDict[file]?.1 ?? IncrementalParseNodeAffectRangeCollector()
+          if iter != 0 {
+            incrementalParseTransition = IncrementalParseTransition(
+              previousTree: previousTreeDict[file]!.0,
+              edits: ConcurrentEdits(fromSequential: []),
+              reusedNodeDelegate: nil
+            )
+          }
```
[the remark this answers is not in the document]
StevenWong12: Add an initial parse is a great idea! But I still store a `(SourceFileSyntax, IncrementalParseNodeAffectRangeCollector)` here since I think we could not get the collector from `IncrementalParseTransition`.

Sources/SwiftParser/Parser.swift
```diff
@@ -629,3 +649,49 @@ extension Parser {
...
+    let currentOffset = self.lexemes.getOffsetToStart(self.currentToken)
+    if let node = lookUpHelper.lookUp(currentOffset, kind: kind) {
+      self.lexemes.advance(by: node.byteSize, currentToken: &self.currentToken)
+      return node
+    }
+
+    return nil
+  }
+
+  func registerNodeForIncrementalParse(node: RawSyntax, startToken: Lexer.Lexeme) {
```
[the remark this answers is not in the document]
StevenWong12: It really gets better performance if we mark it as `mutating`.😮 It reduces the overhead from 4-5% to 2-3%.

Sources/swift-parser-cli/Commands/PerformanceTest.swift
```diff
@@ -37,19 +41,39 @@ struct PerformanceTest: ParsableCommand {
...
       .map { try Data(contentsOf: $0) }
 
-    let start = Date()
+    var incrementalParseTransition: IncrementalParseTransition? = nil
+    var incrementalParseAffectRangeCollector: IncrementalParseNodeAffectRangeCollector? = nil
+
+    var totalTime: TimeInterval = .zero
     let startInstructions = getInstructionsExecuted()
-    for _ in 0..<self.iterations {
+    var previousTreeDict: [Data: (SourceFileSyntax, IncrementalParseNodeAffectRangeCollector?)] = [:]
```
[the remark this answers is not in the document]
StevenWong12: I think we should still store a `(SourceFileSyntax, IncrementalParseNodeAffectRangeCollector)` here since we could not get the collector from `IncrementalParseTransition`.

Sources/swift-parser-cli/Commands/PerformanceTest.swift
```diff
@@ -37,12 +41,40 @@ struct PerformanceTest: ParsableCommand {
...
+          source: buf.bindMemory(to: UInt8.self),
+          parseNodeAffectRange: incrementalParseAffectRangeCollector
+        )
+
+        if self.incrementalParse == true {
+          fileTransition[file] = (tree, incrementalParseAffectRangeCollector!)
+        }
+      }
+    }
+
```
ahoppen, line 60: We only need to do the initial parse if `self.incrementalParse` is `true`, right? You could just put the entire loop in a `if self.incrementalParse` and then you don’t need to check that variable inside the loop.

Sources/swift-parser-cli/Commands/PerformanceTest.swift
```diff
@@ -37,19 +41,39 @@ struct PerformanceTest: ParsableCommand {
...
+            source: buf.bindMemory(to: UInt8.self),
+            parseNodeAffectRange: incrementalParseAffectRangeCollector,
+            parseTransition: incrementalParseTransition
+          )
+          totalTime += Date().timeIntervalSince(start)
+          previousTreeDict[file] = (tree, incrementalParseAffectRangeCollector)
         }
       }
     }
     let endInstructions = getInstructionsExecuted()
```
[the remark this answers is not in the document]
ahoppen: OK, interesting. But given that instructions can take different durations to execute based on cache hits, it is reasonable that they could diverge slightly.

Sources/swift-parser-cli/Commands/PerformanceTest.swift
```diff
@@ -37,12 +41,40 @@ struct PerformanceTest: ParsableCommand {
       .filter { $0.pathExtension == "swift" }
       .map { try Data(contentsOf: $0) }
 
+    var fileTransition: [Data: (SourceFileSyntax, IncrementalParseNodeAffectRangeCollector)] = [:]
```
ahoppen, line 44: Could you add labels to the tuple elements. It makes the access a lot easier to read IMO.

Sources/swift-parser-cli/Commands/PerformanceTest.swift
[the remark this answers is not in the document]
ahoppen: But we could just pass `nil` for the `parseNodeAffectRange` because we aren’t doing another incremental parse based on the tree produced in this loop, right?

Sources/swift-parser-cli/Commands/PerformanceTest.swift
```diff
@@ -37,12 +41,40 @@ struct PerformanceTest: ParsableCommand {
...
-          _ = Parser.parse(source: buf.bindMemory(to: UInt8.self))
+          let tree = Parser.parse(
+            source: buf.bindMemory(to: UInt8.self),
+            parseNodeAffectRange: incrementalParseAffectRangeCollector,
+            parseTransition: parseTransition
+          )
+
+          if self.incrementalParse == true {
+            fileTransition[file] = (tree, incrementalParseAffectRangeCollector!)
+          }
```
ahoppen, line 77: We don’t need this, right? We could just perform `self.iterations` incremental parses based on the tree generation in the initial parse loop.

Tests/SwiftParserTest/IncrementalParsingTests.swift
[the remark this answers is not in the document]
ahoppen: Yes, that makes sense.

Sources/swift-parser-cli/Commands/PerformanceTest.swift
StevenWong12: Ah, yeah.

Sources/swift-parser-cli/Commands/PerformanceTest.swift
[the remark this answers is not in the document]
StevenWong12: Yeah, but we still need to store a `IncrementalParseNodeAffectRangeCollector ` for every file if we are doing incremental parse, right? I store a `(IncrementalParseTransition, IncrementalParseNodeAffectRangeCollector)` here in the new commit and it should make the actual performance test loop simpler.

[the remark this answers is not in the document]
ahoppen: Why do we need to store an `IncrementalParseNodeAffetRangeCollector`? Can’t you just change the call in the performance measurement loop to pass `nil` for `parseNodeAffectRange` because we aren’t doing another incremental parse based on the tree that’s generated in the performance measurement loop?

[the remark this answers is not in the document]
StevenWong12: But if we pass `--incremental-parse` to performance test, we would need to use the `IncrementalParseNodeAffetRangeCollector` set up in the initial parse to enable incremental parse in the performance measurement loop. Do you mean we would not measure the performance of incremental parse when `self.iterations` > 1?

StevenWong12:
I add two `Parser.parse` methods which accept `IncrementalParseTransition` and return `(tree: SourceFileSyntax, nodeAffectRangeCollector: IncrementalParseNodeAffectRangeCollector)` as the incremental parse entries. Then we could use the `nodeAffectRangeCollector` returned by last parse to initialize `IncrementalParseTransition`. What do you think about it?

Sources/SwiftParser/Lexer/LexemeSequence.swift
```diff
@@ -32,24 +32,54 @@ extension Lexer {
     /// usually less than 0.1% of the memory allocated by the syntax arena.
     var lexerStateAllocator = BumpPtrAllocator(slabSize: 256)
 
+    /// Compute the offset of the end of next token
+    var offsetToNextTokenEnd: Int {
+      self.getOffsetToStart(self.nextToken) + self.nextToken.byteLength
+    }
+
+    /// See doc comments in ``LookaheadTracker``
+    var lookaheadTracker: UnsafeMutablePointer<LookaheadTracker>
```
ahoppen, line 41: Why does this need to be a `var`. Wouldn’t a `let` also work?

Sources/SwiftParser/Lexer/LexemeSequence.swift
```diff
@@ -32,24 +32,54 @@ extension Lexer {
...
+      self.lookaheadTracker.initialize(to: LookaheadTracker())
     }
 
     @_spi(Testing)
     public mutating func next() -> Lexer.Lexeme? {
       return self.advance()
     }
 
+    /// Record the offset of the end of next token as the furthest offset in ``LookaheadTracker``
+    func recordFurthestOffset() {
```
ahoppen, line 57: What do you think about naming this `recordNextTokenInLookaheadTracker`? I think that’s a little clearer, if only because it mentions `LookaheadTracker`. If you’ve got better naming ideas, I’m open to them.

Sources/SwiftParser/Parser.swift
```diff
@@ -629,3 +655,41 @@ extension Parser {
...
+
+// MARK: Incremental Parsing
+extension Parser {
+  mutating func loadCurrentSyntaxNodeFromCache(for kind: SyntaxKind) -> Syntax? {
+    guard var parseLookup else {
+      return nil
+    }
+
+    let currentOffset = self.lexemes.getOffsetToStart(self.currentToken)
+    if let node = parseLookup.lookUp(currentOffset, kind: kind) {
```
ahoppen, line 667: This won’t actually get any of the `SyntaxCursor` benefits because what you are modifying when calling `lookUp` below is the copy of `parseLookup` that you created above.

Could you check how the performance changes if you do the following? If there’s no performance win, we should consider removing `SyntaxCursor` entirely in a follow-up PR.

```suggestion
    guard parseLookup != nil else {
      return nil
    }

    let currentOffset = self.lexemes.getOffsetToStart(self.currentToken)
    if let node = self.parseLookup!.lookUp(currentOffset, kind: kind) {
```

Sources/SwiftParser/Parser.swift
```diff
@@ -629,3 +655,41 @@ extension Parser {
...
+  mutating func registerNodeForIncrementalParse(node: RawSyntax, startToken: Lexer.Lexeme) {
+    parseNodeAffectRange.registerNodeForIncrementalParse(
+      node: node,
+      length: max(lookaheadFurthestOffset - self.lexemes.getOffsetToStart(startToken), node.byteLength + currentToken.byteLength)
+    )
+  }
+
+  public var lookaheadFurthestOffset: Int {
+    return lexemes.lookaheadTracker.pointee.furthestOffset
+  }
```
ahoppen, line 684: You are only using this once so I don’t think we need it. Also: In any case, this shouldn’t be public.

Sources/SwiftParser/Parser.swift
```diff
@@ -629,3 +655,41 @@ extension Parser {
...
+}
+
+/// Record the furthest offset to `sourceBufferStart` that is reached by  ``Parser.Peek()`` or ``Lookahead`` in ``Lexer/LexemeSequence``
```
ahoppen, line 687: Typo in `peek. Also, since `peek` isn’t public, it can’t be a docc reference.

```suggestion
/// Record the furthest offset to `sourceBufferStart` that is reached by  `Parser.peek()` or ``Lookahead`` in ``Lexer/LexemeSequence``
```

Sources/swift-parser-cli/Commands/PerformanceTest.swift
```diff
@@ -37,12 +41,35 @@ struct PerformanceTest: ParsableCommand {
       .filter { $0.pathExtension == "swift" }
       .map { try Data(contentsOf: $0) }
 
+    var fileTransition: [Data: IncrementalParseTransition] = [:]
+
+    if self.incrementalParse {
+      /// The initial parse for incremental parse
```
ahoppen, line 47:
```suggestion
      /// The initial parse for incremental parsing
```

ahoppen (commented):
This looks very close now. I mostly have comments about naming and doc comments.

Sources/SwiftParser/IncrementalParseTransition.swift
```diff
@@ -20,26 +20,39 @@
...
 
-  public func parserReusedNode(range: ByteSourceRange, previousNode: Syntax) {
-    rangeAndNodes.append((range, previousNode))
+  public func parserReusedNode(previousNode: Syntax) {
+    nodes.append(previousNode)
+  }
+}
+
+/// Record the affect range for potential re-used nodes. When edits intersect the affect range, the node is not able to be re-used.
+public struct IncrementalParseNodeAffectRangeCollector {
```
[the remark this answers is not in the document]
StevenWong12: Ah, I think that's great. I also move this to `Parser.swift` considering its general purpose.

Sources/SwiftParser/Parser.swift
```diff
@@ -237,6 +262,7 @@ public struct Parser {
 extension Parser {
   /// Retrieves the token following the current token without consuming it.
   func peek() -> Lexer.Lexeme {
+    lexemes.recordFurthestOffset()
```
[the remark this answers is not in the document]
StevenWong12: Ah, I didn't do that because I was trap into some weird bugs and fail the tests then. And it turned out that it was because I called `recordNextTokenInLookaheadTracker` in the wrong time🥹.

> Or is there a performance regression associated with that?
About the performance, see my comments below.

Sources/SwiftParser/Parser.swift
StevenWong12: Oh, I forgot `parseLookup` is a struct. About the performance, see my comments below.

StevenWong12:
I re-run some performance tests at two points.

1. About the `SyntaxCursor`, we actually get many benefits from it (obvious in 100-iterations run) and I've been mis-using it before.
image


2. About calling `recordNextTokenInLookaheadTracker` in `LexemeSequence`. There aren't very much difference between calling in `LexemeSequence` and `Lookahead.comsumeAnyToken/Parser.peek` in terms of performance. And the difference is too small to be caught by instruments. I think calling them in `LexemeSequence` is more clear.
image

Sources/SwiftParser/generated/Parser+Entry.swift
```diff
@@ -18,23 +18,62 @@ extension Parser {
...
+    parseTransition: IncrementalParseTransition?
+  ) -> (tree: SourceFileSyntax, lookaheadRanges: LookaheadRanges) {
+    var parser = Parser(source, parseTransition: parseTransition)
+    return (SourceFileSyntax.parse(from: &parser), parser.lookaheadRanges)
+  }
+  
+  /// Parse the source code in the given buffer as Swift source file with support
+  /// for incremental parsing.
+  ///
+  /// See doc comments above
```
ahoppen, line 68: `above` doesn’t refer to anything when viewing the generated documentation, e.g. on https://swiftpackageindex.com/apple/swift-syntax/main/documentation/swiftsyntax. Can you change this to a docc-link? Something like `Parser/parseIncrementally(source:parseTransition)`. Xcode should offer you code completion to make the link work.

Sources/SwiftParser/Parser.swift
```diff
@@ -629,3 +658,69 @@ extension Parser {
...
+
+  deinit {
+    self.lookaheadTracker.deallocate()
+  }
+}
+
+/// Record the lookahead ranges for syntax nodes.
+public struct LookaheadRanges {
+  /// For each node that is recorded for re-use, the number of UTF-8 bytes that the parser looked ahead to parse the node, measured from the start of the node’s leading trivia.
+  /// This information can be used to determine whether a node can be reused in incremental parse. A node can only be re-used if no byte in its looked range has changed.
```
ahoppen, line 717: I forgot this in my suggestion as well but we should should have a newline between the summary and discussion for this property.

```suggestion
  /// For each node that is recorded for re-use, the number of UTF-8 bytes that the parser looked ahead to parse the node, measured from the start of the node’s leading trivia.
  ///
  /// This information can be used to determine whether a node can be reused in incremental parse. A node can only be re-used if no byte in its looked range has changed.
```

Sources/SwiftParser/Parser.swift
```diff
@@ -629,3 +658,69 @@ extension Parser {
...
+  var lookaheadRanges: [RawSyntax.ID: Int] = [:]
+
+  public init() {}
+
+  @_spi(RawSyntax)
+  public mutating func registerNodeForIncrementalParse(node: RawSyntax, lookaheadLength: Int) {
```
ahoppen, line 723: I think this method could be internal.

Sources/SwiftParser/Parser.swift
```diff
@@ -629,3 +658,69 @@ extension Parser {
...
+
+  mutating func registerNodeForIncrementalParse(node: RawSyntax, startToken: Lexer.Lexeme) {
+    lookaheadRanges.registerNodeForIncrementalParse(
+      node: node,
+      lookaheadLength: lexemes.lookaheadTracker.pointee.furthestOffset - self.lexemes.getOffsetToStart(startToken)
+    )
+  }
+}
+
+/// Record the furthest offset to `sourceBufferStart` that is reached by  `Parser.peek()` or `Parser.Lookahead` in ``Lexer/LexemeSequence``
```
ahoppen, line 686: Depending on whether `LexemeSequence` or `Parser`/`Lookahead` is responsible fro calling `recordFurthestOffset`

Sources/SwiftParser/Parser.swift
```diff
@@ -629,3 +658,69 @@ extension Parser {
...
+public struct LookaheadTracker {
+  private(set) var furthestOffset: Int = 0
+
+  public init() {}
+
+  public mutating func recordFurthestOffset(_ furthestOffset: Int) {
```
ahoppen, line 692: Does this method have to be public or could it be internal?

Sources/SwiftParser/Lexer/LexemeSequence.swift
```diff
@@ -32,24 +32,60 @@ extension Lexer {
     /// usually less than 0.1% of the memory allocated by the syntax arena.
     var lexerStateAllocator = BumpPtrAllocator(slabSize: 256)
 
-    fileprivate init(sourceBufferStart: Lexer.Cursor, cursor: Lexer.Cursor) {
+    /// The offset of the end trailing trivia of `nextToken` relative to the source buffer’s start.
```
ahoppen, line 35:
```suggestion
    /// The offset of the trailing trivia end of `nextToken` relative to the source buffer’s start.
```

Sources/SwiftParser/Parser.swift
```diff
@@ -101,6 +101,15 @@ public struct Parser {
   /// When this nesting level is exceeded, the parser should stop parsing.
   let maximumNestingLevel: Int
 
+  /// See commens in ``IncrementalParseLookup``
+  var parseLookup: IncrementalParseLookup?
+
+  /// See comments in ``LookaheadRanges``
+  public var lookaheadRanges = LookaheadRanges()
```
ahoppen, line 108: Ah I just realized that actually there’s no reason to make it publicly writable. Not sure if 

```suggestion
  public internal(set) var lookaheadRanges = LookaheadRanges()
```

Sources/SwiftParser/Lookahead.swift
```diff
@@ -90,6 +91,7 @@ extension Parser.Lookahead {
   mutating func consumeAnyToken() {
     tokensConsumed += 1
     self.currentToken = self.lexemes.advance()
+    self.lexemes.recordNextTokenInLookaheadTracker()
```
ahoppen, line 94: As the second part of my comment https://github.com/apple/swift-syntax/pull/1685#discussion_r1254507635:

Could we also call `recordNextTokenInLookaheadTracker` from `LexemeSequence.advance` so that we never need to call it from `Lookahead` or `Parser` (and could maybe even make that method `private`)? Or is there a performance regression associated with it because we’d now effectively also be calling `recordNextTokenInLookaheadTracker` from `Parser.consumeAnyToken`.

In either case, I think we should be consistent about who’s responsible for calling `recordNextTokenInLookaheadTracker`. Either `LexemeSequence` calls it from both `advance` and `peek` or `Lookahead` and `Parser` are responsible for calling it. I don’t think we should end up in a situation where `LexemeSequence` calls it on `peek` but `Lookeahead`/`Parser` need to call it on `advance` because that sort of inconsistency is a source of bugs.

Sources/SwiftParser/Parser.swift
```diff
@@ -101,6 +101,15 @@ public struct Parser {
   /// When this nesting level is exceeded, the parser should stop parsing.
   let maximumNestingLevel: Int
 
+  /// See commens in ``IncrementalParseLookup``
```
ahoppen, line 104:
```suggestion
  /// See comments in ``IncrementalParseLookup``
```

Sources/SwiftParser/Parser.swift
```diff
@@ -237,6 +265,7 @@ public struct Parser {
 extension Parser {
   /// Retrieves the token following the current token without consuming it.
   func peek() -> Lexer.Lexeme {
+    lexemes.recordNextTokenInLookaheadTracker()
```
ahoppen, line 268: I don’t think this call is necessary anymore if `LexemeSequence.peek` calls `recordNextTokenInLookaheadTracker`.

Sources/SwiftParser/Parser.swift
```diff
@@ -629,3 +658,69 @@ extension Parser {
     )
   }
 }
+
+// MARK: Incremental Parsing
```
ahoppen, line 662: What do you think about moving all of the incremental-parsing functions to its own file? It just keeps them nicely grouped and maybe we can even mark some functions as `fileprivate`.

Tests/SwiftParserTest/Assertions.swift
```diff
@@ -198,6 +200,7 @@ func assertLexemes(
     }
     assertTokens(lexemes, expectedLexemes, markerLocations: markerLocations, file: file, line: line)
   }
+  lookaheadTracker.deallocate()
```
ahoppen, line 203: Just a bit of Swift style: You can add this as a `defer`-block right after `allocate`. That way it’s even easier to verify that all allocated memory gets deallocated as well.

Same in `LexerTests.swift`

ahoppen (commented):
Even better performance, I love it. 

I think the only non-trivial question that remains is who’s responsible for calling `recordNextTokenInLookaheadTracker`, everything else is super minor now.

Sources/SwiftParser/Lookahead.swift
StevenWong12: Oops, I think that's something I forgot to delete after doing the performance test. `recordNextTokenInLookaheadTracker` is already added to `LexemeSequence.advance`.

Sources/SwiftParser/Parser.swift
StevenWong12: I think that's great. And I also make `IncrementalParseLookup.lookup` an `fileprivate` function.

Sources/SwiftParser/Lexer/LexemeSequence.swift
```diff
@@ -32,24 +32,60 @@ extension Lexer {
...
+      self.lookaheadTracker = lookaheadTracker
     }
 
     @_spi(Testing)
     public mutating func next() -> Lexer.Lexeme? {
       return self.advance()
     }
 
+    /// Record the offset of the end of `nextToken` as the furthest offset in ``LookaheadTracker``
+    func recordNextTokenInLookaheadTracker() {
```
ahoppen, line 60: Can this be `private` now if it’s only being called from `LexemeSequence`?

ahoppen (commented):
One last comment to make a method `private`, otherwise this is looking great now 🕺🏽💃🏽.

Sources/SwiftParser/Lexer/LexemeSequence.swift
StevenWong12: Sure.

ahoppen:
🚀 

@swift-ci Please test

ahoppen (approved):
🚀

ahoppen:
https://github.com/apple/swift-stress-tester/pull/242

@swift-ci Please test macOS

ahoppen:
https://github.com/apple/swift-stress-tester/pull/242

@swift-ci Please test Windows

StevenWong12:
Oops, the same problem as before. Maybe we need to re-trigger again🥲

PR #1690, opened 2023-05-22
Remove `requiresLeadingNewline` from basic format

kimdv:
As part of this comment I started by removing the first part

https://github.com/apple/swift-syntax/pull/1623#discussion_r1196921113

### DIFF
--- a/Sources/SwiftBasicFormat/BasicFormat.swift
+++ b/Sources/SwiftBasicFormat/BasicFormat.swift
@@ -155,10 +155,6 @@ open class BasicFormat: SyntaxRewriter {
     {
       return false
     } else if let second {
-      if second.requiresLeadingNewline {
-        return true
-      }
-
       var ancestor: Syntax = Syntax(second)
       while let parent = ancestor.parent {
         ancestor = parent
@@ -178,11 +174,16 @@ open class BasicFormat: SyntaxRewriter {
     }
 
     switch (first?.tokenKind, second?.tokenKind) {
+
     case (.multilineStringQuote, .backslash),  // string interpolation segment inside a multi-line string literal
       (.multilineStringQuote, .multilineStringQuote),  // empty multi-line string literal
       (.multilineStringQuote, .stringSegment),  // segment starting a multi-line string literal
       (.stringSegment, .multilineStringQuote),  // ending a multi-line string literal that has a string interpolation segment at its end
-      (.rightParen, .multilineStringQuote):  // ending a multi-line string literal that has a string interpolation segment at its end
+      (.rightParen, .multilineStringQuote),  // ending a multi-line string literal that has a string interpolation segment at its end
+      (_, .poundElseKeyword),
+      (_, .poundElseifKeyword),
+      (_, .poundEndifKeyword),
+      (_, .rightBrace):
       return true
     default:
       return false
--- a/CodeGeneration/Sources/SyntaxSupport/Child.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/Child.swift
@@ -62,7 +62,6 @@ public class Child {
   public let description: String?
   public let forceClassification: Bool
   public let isIndented: Bool
-  public let requiresLeadingNewline: Bool
   public let isOptional: Bool
   public let classification: SyntaxClassification?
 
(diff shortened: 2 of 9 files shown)


kimdv:
@swift-ci please test

ahoppen (approved):
Whooo 🎉

ahoppen:
@swift-ci Please test

kimdv:
@swift-ci please test Windows

kimdv:
@swift-ci please test macOS

kimdv:
@swift-ci please test macOS

ahoppen:
I prefer `precedencegroup MyPrecedence {}` over `precedencegroup MyPrecedence {\n}`, so updating the tests seems good to me.

To fix the test failures for `SourceKit/Macros/diags.swift`, you need to modify the following file https://github.com/apple/swift/blob/main/test/SourceKit/Macros/diags.swift.response. The test that is being run is in the compiler repo as well: https://github.com/apple/swift/blob/main/test/SourceKit/Macros/diags.swift

For `Macros/macro_expand.swift`, I’m not sure because I don’t see the failure anymore but you most likely need to modify some of the `CHECK` lines in https://github.com/apple/swift/blob/main/test/Macros/macro_expand.swift

kimdv:
@swift-ci please test

kimdv:
@ahoppen it looks like if we want `{}` there was more changes.

kimdv:
@swift-ci please test

Sources/SwiftBasicFormat/BasicFormat.swift
```diff
@@ -178,11 +174,15 @@ open class BasicFormat: SyntaxRewriter {
     }
 
     switch (first?.tokenKind, second?.tokenKind) {
+    case (.leftBrace, .rightBrace):  // empty declarations like `struct {}`.
+      return false
```
ahoppen, line 178: I would prefer this to be a whitelist that always returns `true` if something matches and only falls through to `return false`. In my experience those are a lot easier to reason about than if you interleave white and blacklists.

So, I’d prefer this to change the below to something like here

https://github.com/apple/swift-syntax/blob/3ed9dcf27c69b991d3cd8eadd64764638f3b0dec/Sources/SwiftBasicFormat/BasicFormat.swift#L260


---

When seeing the examples in the test case below, I think I changed my mind that I prefer to add a newline in between `{` and `}`. I prefer to write `func foo() {}` if I’m deliberately declaring an empty function but in the cases below, we are mostly adding `{` and `}` for Fix-Its and in that case you almost always want to add something between the braces.

kimdv: I will revert the change and update the two previously failing test cases

Sources/SwiftBasicFormat/BasicFormat.swift
```diff
@@ -178,11 +174,15 @@ open class BasicFormat: SyntaxRewriter {
...
     case (.multilineStringQuote, .backslash),  // string interpolation segment inside a multi-line string literal
       (.multilineStringQuote, .multilineStringQuote),  // empty multi-line string literal
       (.multilineStringQuote, .stringSegment),  // segment starting a multi-line string literal
       (.stringSegment, .multilineStringQuote),  // ending a multi-line string literal that has a string interpolation segment at its end
-      (.rightParen, .multilineStringQuote):  // ending a multi-line string literal that has a string interpolation segment at its end
+      (.rightParen, .multilineStringQuote),  // ending a multi-line string literal that has a string interpolation segment at its end
+      (_, .rightBrace),
+      (_, .poundEndifKeyword):
```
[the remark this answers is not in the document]
kimdv: You are totally right! 
I've added also test case for `#elseif`.

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test

ahoppen (approved):
Nice to see yet another use of CodeGeneration gone

kimdv:
https://github.com/apple/swift/pull/66106

@swift-ci plese test

kimdv:
https://github.com/apple/swift/pull/66106

@swift-ci please test

kimdv:
https://github.com/apple/swift/pull/66106

@swift-ci please test

kimdv:
@swift-ci please test windows

kimdv:
@swift-ci please test

kimdv:
https://github.com/apple/swift/pull/66106

@swift-ci please test

kimdv:
https://github.com/apple/swift/pull/66106

@swift-ci please test

kimdv:
@ahoppen after some tried I got it working. 

Will you also review: https://github.com/apple/swift/pull/66106

PR #1693, opened 2023-05-23
Add an initializer that accepts [ExprSyntax] for ArrayExprSyntax

### LINKED ISSUE #1674: `ArrayExprSyntax` initializer that takes an array of `ExprSyntax`
### Description

Constructing an `ArrayExprSyntax` from a Swift array of arbitrary `ExprSyntax` values is a bit of a pain. Right now, we do silly dances like:

```swift
let exprs: [ExprSyntax] = ...
let exprString = exprs.lazy
  .map(\.trimmedDescription)
  .joined(separator: ", ")
return ExprSyntax("[\(raw: exprString)]").as(ArrayExprSyntax.self)
```

Having a convenience initializer that inserts commas and necessary trailing trivia would make this much nicer.

gibachan:
Resolve #1674

I have added initializers for `ArrayExprSyntax` and `ArrayElementListSyntax` that allow codes like below

```swift
let expressions: [ExprSyntax] = ...
let arrayExprSyntax = ArrayExprSyntax(expressions: expressions)
```

( I'm not sure if I need to write a test for the `ArrayExprSyntax` initializer or where to place it. )

### DIFF
--- a/Sources/SwiftSyntaxBuilder/ConvenienceInitializers.swift
+++ b/Sources/SwiftSyntaxBuilder/ConvenienceInitializers.swift
@@ -13,6 +13,31 @@
 @_spi(RawSyntax) import SwiftParser
 @_spi(RawSyntax) import SwiftSyntax
 
+// MARK: - ArrayElementList
+
+extension ArrayElementListSyntax {
+  public init(expressions: [ExprSyntax]) {
+    let lastIndex = expressions.count - 1
+    let elements = expressions.enumerated().map { index, expression in
+      let element = ArrayElementSyntax(expression: expression)
+      if index < lastIndex {
+        return element.ensuringTrailingComma()
+      } else {
+        return element
+      }
+    }
+    self.init(elements)
+  }
+}
+
+// MARK: - ArrayExpr
+
+extension ArrayExprSyntax {
+  public init(expressions: [ExprSyntax]) {
+    self.init(elements: ArrayElementListSyntax(expressions: expressions))
+  }
+}
+
 // MARK: - CustomAttribute
 
 extension AttributeSyntax {
--- a/Tests/SwiftSyntaxBuilderTest/ArrayExprTests.swift
+++ b/Tests/SwiftSyntaxBuilderTest/ArrayExprTests.swift
@@ -47,4 +47,10 @@ final class ArrayExprTests: XCTestCase {
       """
     )
   }
+
+  func testInitializerWithExpressions() {
+    let expressions: [ExprSyntax] = ["0", "1", "2"]
+    let arrayExpr = ArrayExprSyntax(expressions: expressions)
+    XCTAssertEqual(arrayExpr.description, "[0,1,2]")
+  }
 }


kimdv:
Hi @gibachan 👋

The files you have changed are inside a generated folder. 
This means that running the code generation they will be reverted. You can read about it here: https://github.com/apple/swift-syntax/blob/main/CodeGeneration/README.md

Also I think the code would fit better inside the SwiftSyntaxBuilder, as this is about building the syntax code. 
I think you could add it to `ConvenienceInitializers.swift`.

You can then add test cases to `ArrayExprTests.swift`.

gibachan:
Hi @kimdv, thank you so much!

I moved the initializers to `ConvenienceInitializers.swift`!

kimdv (approved):

kimdv:
@swift-ci please test

kimdv:
Would you mind to squash the two commits @gibachan ? 😁

kimdv:
@swift-ci please test

Sources/SwiftSyntaxBuilder/ConvenienceInitializers.swift
```diff
@@ -13,6 +13,26 @@
 @_spi(RawSyntax) import SwiftParser
 @_spi(RawSyntax) import SwiftSyntax
 
+// MARK: - ArrayElementList
+
+extension ArrayElementListSyntax {
+  public init(expressions: [ExprSyntaxProtocol]) {
```
grynspan, line 19: You should explicitly write `any ExprSyntaxProtocol` here.

Sources/SwiftSyntaxBuilder/ConvenienceInitializers.swift
```diff
@@ -13,6 +13,26 @@
...
+    let elements = expressions.enumerated().map { index, expression in
```
grynspan, line 20: It's valid for the last element to have a trailing comma in an array literal, so there's no need for `enumerated()` or index-checking here.

Sources/SwiftSyntaxBuilder/ConvenienceInitializers.swift
ahoppen: We prefer to use `ExprSyntax` here. I’m adding some documentation about it in https://github.com/apple/swift-syntax/pull/1696

Sources/SwiftSyntaxBuilder/ConvenienceInitializers.swift
ahoppen: SwiftSyntaxBuilder doesn’t add a trailing comma for the last element, so I’d prefer to be consistent between the two initializers her and also not add a trailing comma here either. So: I’d prefer to keep it as you wrote it @gibachan.

Sources/SwiftSyntaxBuilder/ConvenienceInitializers.swift
```diff
@@ -13,6 +13,26 @@
...
+      let element = ArrayElementSyntax(expression: expression)
+      return index != (expressions.count - 1) ? element.with(\.trailingComma, .commaToken()) : element
```
ahoppen, line 22: This will override any potentially already existing trailing comma (and trivia associated with it). You should be able to use `element.ensuringTrailingComma()` instead.

Also, I’d prefer an `if` to the ternary expression. I just think it reads a lot more easily.

ahoppen (commented):
Thank you. I’ve got a couple of comments, but overall looks good

Sources/SwiftSyntaxBuilder/ConvenienceInitializers.swift
grynspan: Or that.

gibachan:
Thank you for all your comments 😄  I have made the fixes!

Tests/SwiftSyntaxBuilderTest/ArrayExprTests.swift
```diff
@@ -47,4 +47,10 @@ final class ArrayExprTests: XCTestCase {
       """
     )
   }
+
+  func testInitializerWithExpressions() {
+    let expressions: [ExprSyntax] = [.init(literal: 0), .init(literal: 1), .init(literal: 2)]
```
ahoppen, line 52: You could also make this even simpler as follows.

```suggestion
    let expressions: [ExprSyntax] = ["0", "1", "2"]
```

gibachan: nice!

kimdv:
@swift-ci please test

kimdv (approved):
LGTM

ahoppen (approved):

ahoppen:
@swift-ci Please test

PR #1694, opened 2023-05-23
Balance spacing around `=`

kimdv:
When the leading or trailing trivia was empty it was replacing the present nodes trivias

### DIFF
--- a/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
+++ b/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
@@ -169,8 +169,9 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
       changes.append(
         FixIt.MultiNodeChange.makePresent(
           correctToken,
-          leadingTrivia: misplacedToken.leadingTrivia,
-          trailingTrivia: misplacedToken.trailingTrivia
+          // Transfer any existing trivia. If there is no trivia in the misplaced token, pass `nil` so that `makePresent` can add required trivia, if necessary.
+          leadingTrivia: misplacedToken.leadingTrivia.isEmpty ? nil : misplacedToken.leadingTrivia,
+          trailingTrivia: misplacedToken.trailingTrivia.isEmpty ? nil : misplacedToken.trailingTrivia
         )
       )
     } else {
@@ -978,7 +979,7 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
         fixIts: [
           FixIt(
             message: ReplaceTokensFixIt(replaceTokens: [.binaryOperator("==")], replacements: [node.equal]),
-            changes: [.makeMissing(unexpected), .makePresent(node.equal, leadingTrivia: [])]
+            changes: [.makeMissing(unexpected), .makePresent(node.equal)]
           )
         ],
         handledNodes: [unexpected.id, node.equal.id]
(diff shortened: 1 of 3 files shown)


Tests/SwiftParserTest/translated/DiagnoseInitializerAsTypedPatternTests.swift
```diff
@@ -84,7 +84,7 @@ final class DiagnoseInitializerAsTypedPatternTests: XCTestCase {
       diagnostics: [
         DiagnosticSpec(message: "unexpected initializer in pattern; did you mean to use '='?", fixIts: ["replace ':' with '='"])
       ],
-      fixedSource: "let f=/*comment*/[X]()"
+      fixedSource: "let f = /*comment*/[X]()"
```
kimdv, line 87: This is currently not working because we are overriding the trivia where I added the `isEmpty`. 

We can accept it or we can do something like "if presentNode.trailingTrivia end with space and the replacement trivia don't start with with a space, merge the trivia"

Sources/SwiftParserDiagnostics/DiagnosticExtensions.swift
```diff
@@ -126,10 +126,10 @@ extension FixIt.MultiNodeChange {
     var presentNode = MissingNodesBasicFormatter(viewMode: .fixedUp).visit(Syntax(node))
     presentNode = PresentMaker().rewrite(presentNode)
 
-    if let leadingTrivia {
+    if let leadingTrivia, !leadingTrivia.isEmpty {
```
ahoppen, line 129: This doesn’t seem correct to me. I think you should be able to explicitly specify that you want empty trivia. That’s why we have the distinction between empty and `nil`. IMO, the solution here would be figure out what’s calling this and make it pass `nil` instead.

kimdv: Its happening here
https://github.com/apple/swift-syntax/blob/3ed9dcf27c69b991d3cd8eadd64764638f3b0dec/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift#L172-L173. 

We could change it to this `misplacedToken.leadingTrivia.isEmpty ? nil : misplacedToken.leadingTrivia` etc.

ahoppen: Yes, I think that would make sense with a comment like the following

```swift
// Transfer any existing trivia. If there is no trivia in the misplaced token, pass `nil` so that `makePresent` can add required trivia, if necessary.
```

Tests/SwiftParserTest/translated/DiagnoseInitializerAsTypedPatternTests.swift
ahoppen: The output I would expect for this is `let f =/*comment*/[X]()` (no space after `=`) because the comment already counts as trivia after the equal sign.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -978,7 +979,7 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
         fixIts: [
           FixIt(
             message: ReplaceTokensFixIt(replaceTokens: [.binaryOperator("==")], replacements: [node.equal]),
-            changes: [.makeMissing(unexpected), .makePresent(node.equal, leadingTrivia: [])]
```
kimdv, line 981: I found two places where we don't need to add trivia

kimdv:
@swift-ci please test

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
ahoppen: Nice. I suspect there might be a few more as well. Haven’t checked though.

ahoppen (approved):
Thanks

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
kimdv: Tried to do a quick look, didn't find any

PR #1695, opened 2023-05-23
Avoid adding diagnostic for array and dictionary when the closing bracket is on a new line

kimdv:
as mentioned in the comment: https://github.com/apple/swift-syntax/pull/1573#discussion_r1201066026

### DIFF
--- a/Sources/SwiftParser/Types.swift
+++ b/Sources/SwiftParser/Types.swift
@@ -1057,27 +1057,26 @@ extension Parser {
         return result
       }
 
-      if self.at(.rightSquareBracket) {
-        let (unexpectedBeforeRSquareBracket, rightSquareBracket) = self.expect(.rightSquareBracket)
+      // If the right square bracket is at a new line, we should just return the result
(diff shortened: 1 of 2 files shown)


Sources/SwiftParser/Types.swift
```diff
@@ -1057,20 +1057,23 @@ extension Parser {
...
-      if self.at(.rightSquareBracket) {
-        let (unexpectedBeforeRSquareBracket, rightSquareBracket) = self.expect(.rightSquareBracket)
+      var lookahead = self.lookahead()
+
+      // If we are at a new line, we cannot assume it's an unfinished dictionary or array.
+      if lookahead.currentToken.isAtStartOfLine {
+        return result
+      }
+
+      if let rightSquareBracket = self.consume(if: .rightSquareBracket) {
```
ahoppen, line 1067: This can just be something like the following, I think
```suggestion
      if let rightSquareBracket = self.consume(if: TokenSpec(.rightSquareBracket, allowStartOfLine: false)) {
```

kimdv:
@swift-ci please test

ahoppen (approved):

PR #1698, opened 2023-05-24
Add newline after consecutive declarations

### LINKED ISSUE #1612: Add newline after `}` to separate from declarations
### Description

https://github.com/apple/swift-syntax/pull/1549#discussion_r1179899781
Test case `testSwitch2`

https://github.com/apple/swift-syntax/pull/1549#discussion_r1179900425
Test case `TrailingSemiTests`
It would be nice to also offer a newline as that is properly what the developer wants. 

### Steps to Reproduce

_No response_

kimdv:
Fixes #1612

### DIFF
--- a/Sources/SwiftBasicFormat/Trivia+FormatExtensions.swift
+++ b/Sources/SwiftBasicFormat/Trivia+FormatExtensions.swift
@@ -45,7 +45,9 @@ extension Trivia {
 
   /// Returns the indentation of the last trivia piece in this trivia that is
   /// not a whitespace.
-  func indentation(isOnNewline: Bool) -> Trivia? {
+  /// - Parameter isOnNewline: Specifies if the character before this trivia is a newline character, i.e. if this trivia already starts on a new line.
+  /// - Returns: An optional ``Trivia`` with indentation of the last trivia piece.
+  public func indentation(isOnNewline: Bool) -> Trivia? {
     let lastNonWhitespaceTriviaPieceIndex = self.pieces.lastIndex(where: { !$0.isWhitespace }) ?? self.pieces.endIndex
     let piecesBeforeLastNonWhitespace = self.pieces[..<lastNonWhitespaceTriviaPieceIndex]
     let indentation: ArraySlice<TriviaPiece>
--- a/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
+++ b/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
@@ -51,6 +51,31 @@ fileprivate func getTokens(between first: TokenSyntax, and second: TokenSyntax) 
 }
 
 fileprivate extension TokenSyntax {
+  /// The indentation of this token
+  ///
+  /// In contrast to `indentation`, this does not walk to previous tokens to
+  /// find the indentation of the line this token occurs on.
+  private var indentation: Trivia? {
+    let previous = self.previousToken(viewMode: .sourceAccurate)
+    return ((previous?.trailingTrivia ?? []) + leadingTrivia).indentation(isOnNewline: false)
+  }
+
+  /// Returns the indentation of the line this token occurs on
+  var indentationOfLine: Trivia {
+    var token: TokenSyntax = self
+    if let indentation = token.indentation {
+      return indentation
+    }
+    while let previous = token.previousToken(viewMode: .sourceAccurate) {
+      token = previous
+      if let indentation = token.indentation {
+        return indentation
+      }
+    }
+
+    return []
+  }
+
   /// Assuming this token is a `poundAvailableKeyword` or `poundUnavailableKeyword`
   /// returns the opposite keyword.
   var negatedAvailabilityKeyword: TokenSyntax {
@@ -1155,13 +1180,28 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
         // Only diagnose the missing semicolon if the decl doesn't contain any errors.
         // If the decl contains errors, the root cause is most likely something different and not the missing semicolon.
         let position = semicolon.previousToken(viewMode: .sourceAccurate)?.endPositionBeforeTrailingTrivia
+        var fixIts: [FixIt] = [
+          FixIt(message: .insertSemicolon, changes: .makePresent(semicolon))
+        ]
+        if let firstToken = node.firstToken(viewMode: .sourceAccurate),
+          let lastToken = node.lastToken(viewMode: .sourceAccurate)
+        {
+          fixIts.insert(
+            FixIt(
+              message: .insertNewline,
+              changes: [
+                .replaceTrailingTrivia(token: lastToken, newTrivia: lastToken.trailingTrivia + .newlines(1) + firstToken.indentationOfLine)
+              ]
+            ),
+            at: 0
+          )
+        }
+
         addDiagnostic(
           semicolon,
           position: position,
           .consecutiveDeclarationsOnSameLine,
-          fixIts: [
-            FixIt(message: .insertSemicolon, changes: .makePresent(semicolon))
-          ],
+          fixIts: fixIts,
           handledNodes: [semicolon.id]
         )
       } else {
--- a/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
+++ b/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
@@ -727,13 +727,28 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
         // Only diagnose the missing semicolon if the item doesn't contain any errors.
         // If the item contains errors, the root cause is most likely something different and not the missing semicolon.
         let position = semicolon.previousToken(viewMode: .sourceAccurate)?.endPositionBeforeTrailingTrivia
+        var fixIts: [FixIt] = [
+          FixIt(message: .insertSemicolon, changes: .makePresent(semicolon))
+        ]
+        if let firstToken = node.firstToken(viewMode: .sourceAccurate),
+          let lastToken = node.lastToken(viewMode: .sourceAccurate)
+        {
+          fixIts.insert(
+            FixIt(
+              message: .insertNewline,
+              changes: [
+                .replaceTrailingTrivia(token: lastToken, newTrivia: lastToken.trailingTrivia + .newlines(1) + firstToken.indentationOfLine)
+              ]
+            ),
+            at: 0
+          )
+        }
+
         addDiagnostic(
           semicolon,
           position: position,
           .consecutiveStatementsOnSameLine,
-          fixIts: [
-            FixIt(message: .insertSemicolon, changes: .makePresent(semicolon))
-          ],
+          fixIts: fixIts,
           handledNodes: [semicolon.id]
         )
       } else {
--- a/Tests/SwiftParserTest/Assertions.swift
+++ b/Tests/SwiftParserTest/Assertions.swift
@@ -275,7 +275,7 @@ class FixItApplier: SyntaxRewriter {
   var changes: [FixIt.Change]
 
   init(diagnostics: [Diagnostic], withMessages messages: [String]?) {
-      let messages = messages ?? diagnostics.compactMap { $0.fixIts.first?.message.message }
+    let messages = messages ?? diagnostics.compactMap { $0.fixIts.first?.message.message }
 
     self.changes =
       diagnostics
--- a/Tests/SwiftParserTest/Assertions.swift
+++ b/Tests/SwiftParserTest/Assertions.swift
@@ -275,15 +275,13 @@ class FixItApplier: SyntaxRewriter {
   var changes: [FixIt.Change]
 
   init(diagnostics: [Diagnostic], withMessages messages: [String]?) {
+      let messages = messages ?? diagnostics.compactMap { $0.fixIts.first?.message.message }
+
     self.changes =
       diagnostics
       .flatMap { $0.fixIts }
       .filter {
-        if let messages {
-          return messages.contains($0.message.message)
-        } else {
-          return true
-        }
+        return messages.contains($0.message.message)
       }
       .flatMap { $0.changes }
   }
--- a/Tests/SwiftParserTest/ExpressionTests.swift
+++ b/Tests/SwiftParserTest/ExpressionTests.swift
@@ -857,8 +857,8 @@ final class ExpressionTests: XCTestCase {
       diagnostics: [
         DiagnosticSpec(
           locationMarker: "3️⃣",
-          message: "consecutive statements on a line must be separated by ';'",
-          fixIts: ["insert ';'"]
+          message: "consecutive statements on a line must be separated by newline or ';'",
+          fixIts: ["insert newline", "insert ';'"]
         ),
         DiagnosticSpec(
           locationMarker: "5️⃣",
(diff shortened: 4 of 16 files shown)


Tests/SwiftParserTest/translated/ConsecutiveStatementsTests.swift
```diff
@@ -102,8 +102,8 @@ final class ConsecutiveStatementsTests: XCTestCase {
       }
       """,
       diagnostics: [
-        DiagnosticSpec(locationMarker: "1️⃣", message: "consecutive declarations on a line must be separated by ';'", fixIts: ["insert ';'"]),
-        DiagnosticSpec(locationMarker: "2️⃣", message: "consecutive declarations on a line must be separated by ';'", fixIts: ["insert ';'"]),
+        DiagnosticSpec(locationMarker: "1️⃣", message: "consecutive declarations on a line must be separated by ';'", fixIts: ["insert newline", "insert ';'"]),
+        DiagnosticSpec(locationMarker: "2️⃣", message: "consecutive declarations on a line must be separated by ';'", fixIts: ["insert newline", "insert ';'"]),
```
kimdv, line 106: IMO we should offer newline before `;` as that is most common and swift style to do. 

What do you think?

kimdv: the fixed source below only adds newline if the `insert newline` is added as the last. 
That will insert newline before the `;`

ahoppen: > IMO we should offer newline before ; as that is most common and swift style to do.
> What do you think?

I agree

> the fixed source below only adds newline if the insert newline is added as the last.
> That will insert newline before the ;

I think what we should do is to change `FixItApplier.init(diagnostics: [Diagnostic], withMessages messages: [String]?)` to only apply the first Fix-It for each diagnostic, not all of them. If a diagnostic offers multiple Fix-Its they should be considered mutually exclusive.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -1046,8 +1046,16 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
           position: position,
           .consecutiveDeclarationsOnSameLine,
           fixIts: [
-            FixIt(message: .insertSemicolon, changes: .makePresent(semicolon))
-          ],
+            node.lastToken(viewMode: .sourceAccurate).map {
+              FixIt(
+                message: .insertNewline,
+                changes: [
+                  .replaceTrailingTrivia(token: $0, newTrivia: .newlines(1))
```
ahoppen, line 1053: We should be retaining the existing trivia as well because it could be a comment. If you want to trim trailing whitespace before the newline, you should make sure that you’re not removing comments.

kimdv: Fixed and added testcase

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -1047,8 +1047,16 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
           position: position,
           .consecutiveDeclarationsOnSameLine,
           fixIts: [
-            FixIt(message: .insertSemicolon, changes: .makePresent(semicolon))
-          ],
+            node.lastToken(viewMode: .sourceAccurate).map {
+              FixIt(
+                message: .insertNewline,
+                changes: [
+                    .replaceTrailingTrivia(token: $0, newTrivia: $0.trailingTrivia + .newlines(1))
```
ahoppen, line 1054:
```suggestion
                  .replaceTrailingTrivia(token: $0, newTrivia: $0.trailingTrivia + .newlines(1))
```

Didn’t swift-format catch this?

Tests/SwiftParserTest/Assertions.swift
```diff
@@ -275,15 +275,13 @@ class FixItApplier: SyntaxRewriter {
   var changes: [FixIt.Change]
 
   init(diagnostics: [Diagnostic], withMessages messages: [String]?) {
+      let messages = messages ?? diagnostics.compactMap { $0.fixIts.first?.message.message }
```
ahoppen, line 278:
```suggestion
    let messages = messages ?? diagnostics.compactMap { $0.fixIts.first?.message.message }
```

Again: Didn’t swift-format catch this?

Tests/SwiftParserTest/translated/ConsecutiveStatementsTests.swift
```diff
@@ -102,8 +102,8 @@ final class ConsecutiveStatementsTests: XCTestCase {
...
         DiagnosticSpec(locationMarker: "3️⃣", message: "consecutive statements on a line must be separated by ';'", fixIts: ["insert ';'"]),
         DiagnosticSpec(locationMarker: "4️⃣", message: "consecutive statements on a line must be separated by ';'", fixIts: ["insert ';'"]),
         DiagnosticSpec(locationMarker: "5️⃣", message: "consecutive statements on a line must be separated by ';'", fixIts: ["insert ';'"]),
```
ahoppen, line 109: Why don’t we offer to add a newline for these diagnostics?

Also, shouldn’t the Fix-It-Applier apply the `insert newline` Fix-It instead of `insert ';'` and thus the `fixedSource` should contain a newline at those positions instead of a `;`?

Tests/SwiftParserTest/translated/ConsecutiveStatementsTests.swift
```diff
@@ -142,12 +142,14 @@ final class ConsecutiveStatementsTests: XCTestCase {
...
-        DiagnosticSpec(message: "consecutive declarations on a line must be separated by ';'", fixIts: ["insert ';'"])
+        DiagnosticSpec(message: "consecutive declarations on a line must be separated by ';'", fixIts: ["insert newline", "insert ';'"])
       ],
+      applyFixIts: ["insert newline"],
       fixedSource: """
         class C {
           // In a sequence of declarations.
-          var a, b : Int; func d() -> Int {}
+          var a, b : Int
+        func d() -> Int {}
```
ahoppen, line 152: 😬 We should  pick up the indentation of the current line. I think you should be able to just walk the tokens backwards until you find a token that contains indentation. The functions in https://github.com/apple/swift-syntax/blob/main/Sources/SwiftBasicFormat/Trivia%2BFormatExtensions.swift might be useful for that. It seems like they might be generally useful, so you could make them public.

Tests/SwiftParserTest/translated/ConsecutiveStatementsTests.swift
```diff
@@ -157,25 +159,75 @@ final class ConsecutiveStatementsTests: XCTestCase {
     )
   }
 
-  func testConsecutiveStatements5() {
+    func testConsecutiveStatements4b() {
+        assertParse(
```
ahoppen, line 163: Indentation seems off

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
kimdv: Forgot to run it🫠

Tests/SwiftParserTest/translated/ConsecutiveStatementsTests.swift
```diff
@@ -15,12 +15,27 @@
 import XCTest
 
 final class ConsecutiveStatementsTests: XCTestCase {
-  func testSimple() {
+  func testSimple1s() {
```
ahoppen, line 18: Is the `1s` a typo?

Sources/SwiftBasicFormat/Trivia+FormatExtensions.swift
```diff
@@ -45,7 +45,7 @@ extension Trivia {
 
   /// Returns the indentation of the last trivia piece in this trivia that is
   /// not a whitespace.
```
ahoppen, line 47: If we make this public, could you explain the `isOnNewline` parameter as well?

Something like: `isOnNewline` specifies if the character before this trivia is a newline character, i.e. if this trivia already starts on a new line.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -50,6 +50,18 @@ fileprivate func getTokens(between first: TokenSyntax, and second: TokenSyntax)
...
 }
 
+fileprivate func getIndentation(for token: TokenSyntax) -> Trivia {
+  var ancestor: Syntax = Syntax(token)
+  while let parent = ancestor.parent {
+    ancestor = parent
+    if let indentation = ancestor.leadingTrivia.indentation(isOnNewline: true), !indentation.isEmpty {
+      return indentation
+    }
+  }
```
ahoppen, line 60: I think you want to walk `token.previousToken(viewMode: .sourceAccurate)` instead of walking the ancestors. 

For example consider

```swift
func
  test() {}
```

If you want to get the indentation of `(`, you wouldn’t get `  ` because `test` is not the first token in any of its ancestors.

Tests/SwiftParserTest/ExpressionTests.swift
```diff
@@ -2079,32 +2161,141 @@ final class StatementExpressionTests: XCTestCase {
...
+      applyFixIts: ["insert identifier", "insert ';'"],
       fixedSource: #"""
         #<#identifier#>
         "abc"; #<#identifier#>
         """#
     )
   }
 
-  func testUnterminatedString8() {
+  func testUnterminatedString8a() {
```
ahoppen, line 2265: I’m not sure if duplicating all these test cases for insertion of `;` and a newline provides a huge amount of value. But it doesn’t hurt either, so all good.

ahoppen (commented):
Thanks. The Fix-Its to add a newline result in much more swifty code.

Sources/SwiftBasicFormat/Trivia+FormatExtensions.swift
kimdv: Added more documentation

Sources/SwiftBasicFormat/Trivia+FormatExtensions.swift
```diff
@@ -59,6 +61,12 @@ extension Trivia {
     return Trivia(pieces: indentation)
   }
 
+  ///
+  ///
+  /// - parameter: isOnNewline
+  ///
+  ///
+
```
ahoppen, line 69: Committed accidentally?

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
ahoppen: Here’s the example I was thinking of

```swift
func 
  test() {} var other: Int
```

Based on how your code reads, I was expecting that we get the indentation of the missing `;` between `}` and `var`, which I would expect to be `  ` because that’s the indentation of the line that the `;` appears on. But I think your implementation would just go one level up and return empty indentation because `func` isn’t indented.
That’s why I was expecting `getIndentation` to walk tokens backwards using `previousToken` instead of walking up the tree hierarchy. 

Behavior-wise, I think your implementation is good. Not indenting `var` is definitely the superior choice to adding two spaces in front of it. To marry the two ideas, I think `getIndentation` should walk tokens backwards but base the indentation that we want on the first token in the `CodeBlockItem` that has the missing `;`. Does this make sense?

---

Unrelated, I would rename `getIndentation` -> `indentation`. Sounds more swifty.

kimdv: Okay, I think I got it. 
But we have a thing, that we can't find the right indentation for `var x = i3️⃣ x = i + x4️⃣ return Float(x)`

Because the return will get the `indentation` from `x = i + x`

We could of course check if the leading trivia is empty.

ahoppen: I was thinking more of something like the following. And then we could move the extension to SyntaxExtensions.swift.

```diff
diff --git a/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift b/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
index a21e346f..674c00b1 100644
--- a/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
+++ b/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
@@ -50,21 +50,31 @@ fileprivate func getTokens(between first: TokenSyntax, and second: TokenSyntax)
   return tokens
 }
 
-fileprivate func indentation(for token: TokenSyntax) -> Trivia {
-  var ancestor: Syntax = Syntax(token)
-  while let parent = ancestor.parent {
-    ancestor = parent
-
-    if let codeBlockItemSyntax = ancestor.as(CodeBlockItemSyntax.self) {
-      return codeBlockItemSyntax.leadingTrivia.indentation(isOnNewline: true) ?? []
+extension TokenSyntax {
+  /// The indentation of this token
+  ///
+  /// In contrast to `indentation`, this does not walk to previous tokens to
+  /// find the indentation of the line this token occurs on.
+  private var indentation: Trivia? {
+    let previous = self.previousToken(viewMode: .sourceAccurate)
+    return ((previous?.trailingTrivia ?? []) + leadingTrivia).indentation(isOnNewline: false)
+  }
+
+  /// Returns the indentation of the line this token occurs on
+  var indentationOfLine: Trivia {
+    var token: TokenSyntax = self
+    if let indentation = token.indentation {
+      return indentation
+    }
+    while let previous = token.previousToken(viewMode: .sourceAccurate) {
+      token = previous
+      if let indentation = token.indentation {
+        return indentation
+      }
     }
 
-    if let memberDeclListItemSyntax = ancestor.as(MemberDeclListItemSyntax.self) {
-      return memberDeclListItemSyntax.leadingTrivia.indentation(isOnNewline: true) ?? []
-    }
+    return []
   }
-
-  return []
 }
 
 fileprivate extension TokenSyntax {
@@ -719,21 +729,24 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
         // Only diagnose the missing semicolon if the item doesn't contain any errors.
         // If the item contains errors, the root cause is most likely something different and not the missing semicolon.
         let position = semicolon.previousToken(viewMode: .sourceAccurate)?.endPositionBeforeTrailingTrivia
+        var fixIts: [FixIt] = [
+          FixIt(message: .insertSemicolon, changes: .makePresent(semicolon)),
+        ]
+        if let firstToken = node.firstToken(viewMode: .sourceAccurate),
+          let lastToken = node.lastToken(viewMode: .sourceAccurate) {
+          fixIts.insert(FixIt(
+            message: .insertNewline,
+            changes: [
+              .replaceTrailingTrivia(token: lastToken, newTrivia: lastToken.trailingTrivia + .newlines(1) + firstToken.indentationOfLine)
+            ]
+          ), at: 0)
+        }
+
         addDiagnostic(
           semicolon,
           position: position,
           .consecutiveStatementsOnSameLine,
-          fixIts: [
-            node.lastToken(viewMode: .sourceAccurate).map {
-              FixIt(
-                message: .insertNewline,
-                changes: [
-                  .replaceTrailingTrivia(token: $0, newTrivia: $0.trailingTrivia + .newlines(1) + indentation(for: $0))
-                ]
-              )
-            },
-            FixIt(message: .insertSemicolon, changes: .makePresent(semicolon)),
-          ].compactMap { $0 },
+          fixIts: fixIts,
           handledNodes: [semicolon.id]
         )
       } else {
@@ -1180,21 +1193,23 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
         // Only diagnose the missing semicolon if the decl doesn't contain any errors.
         // If the decl contains errors, the root cause is most likely something different and not the missing semicolon.
         let position = semicolon.previousToken(viewMode: .sourceAccurate)?.endPositionBeforeTrailingTrivia
+        var fixIts: [FixIt] = [
+          FixIt(message: .insertSemicolon, changes: .makePresent(semicolon)),
+        ]
+        if let firstToken = node.firstToken(viewMode: .sourceAccurate),
+          let lastToken = node.lastToken(viewMode: .sourceAccurate) {
+          fixIts.insert(FixIt(
+            message: .insertNewline,
+            changes: [
+              .replaceTrailingTrivia(token: lastToken, newTrivia: lastToken.trailingTrivia + .newlines(1) + firstToken.indentationOfLine)
+            ]
+          ), at: 0)
+        }
         addDiagnostic(
           semicolon,
           position: position,
           .consecutiveDeclarationsOnSameLine,
-          fixIts: [
-            node.lastToken(viewMode: .sourceAccurate).map {
-              FixIt(
-                message: .insertNewline,
-                changes: [
-                  .replaceTrailingTrivia(token: $0, newTrivia: $0.trailingTrivia + .newlines(1) + indentation(for: $0))
-                ]
-              )
-            },
-            FixIt(message: .insertSemicolon, changes: .makePresent(semicolon)),
-          ].compactMap { $0 },
+          fixIts: fixIts,
           handledNodes: [semicolon.id]
         )
       } else {

```

kimdv:
@swift-ci please test

ahoppen (approved):

kimdv:
@swift-ci please test

PR #1651, opened 2023-05-10
In `nodesDescription` only merge tokens if they occur consecutively in the source code

ahoppen:
This was motiviated by the following discussion, where the issue becomes more apparent: https://github.com/apple/swift-syntax/pull/1643#discussion_r1187871863.

PR #1703, opened 2023-05-25
Fix it to emit a single diagnostic when both open and close quote are missing

### LINKED ISSUE #1691: Parser should emit a single diagnostic if both open and close quote are missing from string literal
### Issue Kind

Bad Diagnostic Produced

### Source Code

```swift
@_expose(Cxx, baz) func foo() {}
```

### Description

From https://github.com/apple/swift-syntax/pull/1688/files#r1201105564

In the above example the parser produces two diagnostics: One for the missing opening `"` before `baz` and one for the missing closing `"` after `baz`. It should emit a single diagnostic instead that offers to add both.

TTOzzi:
Resolve #1691 

When a string segment is located between missing quotes on both sides, looking ahead is interrupted, and each quote is diagnosed separately. 
I have modified the logic to continue looking ahead even if a standalone string segment is encountered during the search.
Furthermore, I added a condition to the `parentContextClause` in order to modify the diagnostic, but I am uncertain if it is the appropriate approach.
If there is a better solution available, please let me know. Thank you.

### DIFF
--- a/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
+++ b/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
@@ -1257,7 +1257,31 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
           node.closeQuote.id,
         ].compactMap { $0 }
       )
+    } else if node.openQuote.presence == .missing,
+      node.unexpectedBetweenOpenDelimiterAndOpenQuote == nil,
+      node.closeQuote.presence == .missing,
+      node.unexpectedBetweenCloseQuoteAndCloseDelimiter == nil,
+      !node.segments.isMissingAllTokens
+    {
+      addDiagnostic(
+        node,
+        MissingBothStringQuotesOfStringSegments(stringSegments: node.segments),
+        fixIts: [
+          FixIt(
+            message: InsertTokenFixIt(missingNodes: [Syntax(node.openQuote), Syntax(node.closeQuote)]),
+            changes: [
+              .makePresent(node.openQuote),
+              .makePresent(node.closeQuote),
+            ]
+          )
+        ],
+        handledNodes: [
+          node.openQuote.id,
+          node.closeQuote.id,
+        ]
+      )
     }
+
     for (diagnostic, handledNodes) in MultiLineStringLiteralIndentatinDiagnosticsGenerator.diagnose(node) {
       addDiagnostic(diagnostic, handledNodes: handledNodes)
     }
--- a/Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift
+++ b/Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift
@@ -386,6 +386,14 @@ public struct MissingAttributeArgument: ParserError {
   }
 }
 
+public struct MissingBothStringQuotesOfStringSegments: ParserError {
+  public let stringSegments: StringLiteralSegmentsSyntax
+
+  public var message: String {
+    return #"expected \#(stringSegments.shortSingleLineContentDescription) to be surrounded by '"'"#
+  }
+}
+
 public struct MissingConditionInStatement: ParserError {
   let node: SyntaxProtocol
 
--- a/Tests/SwiftParserTest/AttributeTests.swift
+++ b/Tests/SwiftParserTest/AttributeTests.swift
@@ -454,19 +454,14 @@ final class AttributeTests: XCTestCase {
 
     assertParse(
       """
-      @_expose(Cxx, 1️⃣baz2️⃣) func foo() {}
+      @_expose(Cxx, 1️⃣baz) func foo() {}
       """,
       diagnostics: [
         DiagnosticSpec(
           locationMarker: "1️⃣",
-          message: #"expected '"' in string literal"#,
-          fixIts: [#"insert '"'"#]
-        ),
-        DiagnosticSpec(
-          locationMarker: "2️⃣",
-          message: #"expected '"' to end string literal"#,
-          fixIts: [#"insert '"'"#]
-        ),
+          message: #"expected code 'baz' to be surrounded by '"'"#,
+          fixIts: [#"insert '"' and '"'"#]
+        )
       ],
       fixedSource: """
         @_expose(Cxx, "baz") func foo() {}
@@ -543,20 +538,15 @@ final class AttributeTests: XCTestCase {
 
     assertParse(
       """
-      @_unavailableFromAsync(message: 1️⃣abc2️⃣)
+      @_unavailableFromAsync(message: 1️⃣abc)
       func foo() {}
       """,
       diagnostics: [
         DiagnosticSpec(
           locationMarker: "1️⃣",
-          message: #"expected '"' in string literal"#,
-          fixIts: [#"insert '"'"#]
-        ),
-        DiagnosticSpec(
-          locationMarker: "2️⃣",
-          message: #"expected '"' to end string literal"#,
-          fixIts: [#"insert '"'"#]
-        ),
+          message: #"expected code 'abc' to be surrounded by '"'"#,
+          fixIts: [#"insert '"' and '"'"#]
+        )
       ],
       fixedSource: """
         @_unavailableFromAsync(message: "abc")
--- a/Tests/SwiftParserTest/translated/OriginalDefinedInAttrTests.swift
+++ b/Tests/SwiftParserTest/translated/OriginalDefinedInAttrTests.swift
@@ -86,13 +86,8 @@ final class OriginalDefinedInAttrTests: XCTestCase {
         ),
         DiagnosticSpec(
           locationMarker: "1️⃣",
-          message: #"expected '"' in string literal"#,
-          fixIts: [#"insert '"'"#]
-        ),
-        DiagnosticSpec(
-          locationMarker: "2️⃣",
-          message: #"expected '"' to end string literal"#,
-          fixIts: [#"insert '"'"#]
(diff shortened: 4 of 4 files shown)


Sources/SwiftParserDiagnostics/MissingNodesError.swift
```diff
@@ -364,7 +370,10 @@ extension ParseDiagnosticsGenerator {
           missingNodes += [sibling]
         } else if sibling.isMissingAllTokens && sibling.hasTokens {
           missingNodes += [sibling]
-        } else if sibling.kind.isSyntaxCollection && sibling.children(viewMode: .sourceAccurate).count == 0 {
+        } else if sibling.kind.isSyntaxCollection
+          && (sibling.children(viewMode: .sourceAccurate).count == 0
+            || sibling.children(viewMode: .sourceAccurate).only?.kind == .stringSegment)
```
TTOzzi, line 375: (If my approach is correct...) How about separating this part as a property named `isSingleToken`(or a more appropriate name) and accessing it as `sibling.isSingleToken`?

ahoppen (commented):
The way I would approach this is to add 

```swift
public override func visit(_ node: StringLiteralExprSyntax) -> SyntaxVisitorContinueKind {
```

to `ParseDiagnosticsGenerator`. If both `node.openQuote` and `node.closeQuote` are missing, you can generate a custom error message.

Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift
```diff
@@ -386,6 +386,14 @@ public struct MissingAttributeArgument: ParserError {
   }
 }
 
+public struct MissingBothStringQuotesOfStringSegments: ParserError {
+  public let stringSegments: StringLiteralSegmentsSyntax
+
+  public var message: String {
+    return "expected \(nodesDescription([stringSegments], format: false)) to be surrounded by '\"'"
```
ahoppen, line 393: If you just want to print the source test of `stringSegments` verbatim, I would use `. shortSingleLineContentDescription ` instead. The core functionality of `nodesDescription` is to paraphrase entire nodes by their node name but I don’t think that’s something we want here.

```suggestion
    return "expected '\(stringSegments.shortSingleLineContentDescription)' to be surrounded by '\"'"
```

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -1257,7 +1257,29 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
           node.closeQuote.id,
         ].compactMap { $0 }
       )
+    } else if (node.openQuote.tokenKind == .stringQuote && node.openQuote.presence == .missing),
+      (node.closeQuote.tokenKind == .stringQuote && node.closeQuote.presence == .missing),
+      !node.segments.isMissingAllTokens
```
ahoppen, line 1262: There’s no need to use `&&` or parentheses here. `,` in an `if` statement is a conjunction already. 

---

Also, I don’t think it’s necessary to check for the token kind of `openQuote` and `closeQuote`. The parser doesn’t generate any trees that hit this at the moment but in theory the diagnostic should also work for multi-line string literals where the `tokenKind` is `multilineStringQuote`. You’d just need to pass the kind of the quote to the error message.

Tests/SwiftParserTest/AttributeTests.swift
```diff
@@ -543,20 +538,15 @@ final class AttributeTests: XCTestCase {
...
-          message: #"expected '"' in string literal"#,
-          fixIts: [#"insert '"'"#]
-        ),
-        DiagnosticSpec(
-          locationMarker: "2️⃣",
-          message: #"expected '"' to end string literal"#,
-          fixIts: [#"insert '"'"#]
-        ),
+          message: #"expected 'abc' to be surrounded by '"'"#,
+          fixIts: [#"insert '""'"#]
```
ahoppen, line 548: I suspect that this will change to `insert '"' and '"'` after https://github.com/apple/swift-syntax/pull/1651 is merged.

ahoppen (commented):
Thank you. Looks very good. I only have a few minor comments to make it slightly more general

Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift
TTOzzi: Oh, I agree. Using shortSingleLineContentDescription instead of nodesDescription seems to be in line with the intention we want.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
TTOzzi: I just wanted to express the conditions for both the openQuote and closeQuote separately 😁
As there is no particular significance, I will make the change and differentiate them using a comma!

----

Hmm... If we don't check types of quotes here or check if a newline is included in the trivia of the quote token, this test case will fail.

```swift
  func testRawStringErrors6() {
    assertParse(
      ###"""
      let _ = ##"""1️⃣aa
        foobar
        aa2️⃣"""##
      """###,
      diagnostics: [
        DiagnosticSpec(locationMarker: "1️⃣", message: "multi-line string literal content must begin on a new line", fixIts: ["insert newline"]),
        DiagnosticSpec(locationMarker: "2️⃣", message: "multi-line string literal closing delimiter must begin on a new line", fixIts: ["insert newline"]),
      ],
      fixedSource: ###"""
        let _ = ##"""
          aa
          foobar
          aa
          """##
        """###
    )
  }
```

In the given case, both the openQuote and closeQuote tokens are marked as missing, but in reality, the quote exists with the missing only newline.

Are there cases in multiline string expressions where both quotes are missing and we need to emit a diagnostic similar to the one we just added, suggesting insert `"""` to surrounding them?

Or should we generate a single diagnostic here for the missing of a newline, instead of `handleInvalidMultilineStringQuote` called from `handleMissingToken`?

Tests/SwiftParserTest/AttributeTests.swift
TTOzzi: Thank you for letting me know! Once the PR is merged, I will rebase to apply the changes 🙂

Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift
```diff
@@ -386,6 +386,14 @@ public struct MissingAttributeArgument: ParserError {
   }
 }
 
+public struct MissingBothStringQuotesOfStringSegments: ParserError {
+  public let stringSegments: StringLiteralSegmentsSyntax
```
kimdv, line 390: The variable name is `stringSegments` but is not an array?
Is it?

Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift
kimdv, line 393: Instead of escaping `"` then you could use `#` at the start and end of the string.
I think Rae strings reads a little easier

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
ahoppen: Oh, I see. In that case maybe check that `unexpectedBeforeOpenQuote` and `unexpectedBeforeCloseQuote` are `nil` because they will contain the string quotes without a newline in the test case above.

TTOzzi: Ah, this works as intended and is more general than checking the token kind! Thank you for providing your feedback.

Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift
TTOzzi: This is not an array, but I declared it this way to align the node's property names and meanings passed when generating messages. Examples include `components` in `TrailingVersionAreIgnored` or `unexpectedNodes` in `UnexpectedNodesError`. 
Do you think it would be better to declare it simply as `stringSegment`?

Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift
TTOzzi: Oh, that's great! Using a raw string seems more readable. I will make the change. Thank you 🙂

Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift
kimdv: After diving in the code, it looks like `StringLiteralSegmentsSyntax` also is a `SyntaxCollection`. Just like `TrailingVersionAreIgnored ` and `UnexpectedNodesError `
So you can completely ignore this comment 🫠

kimdv (approved):

ahoppen (approved):
Thank you. This looks great now. Can you rebase it now that https://github.com/apple/swift-syntax/pull/1651 is merged? Afterwards, I’ll trigger CI.

Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift
TTOzzi: Ah, it seems that all of these have a common protocol called `SyntaxCollection`! Thank you for your feedback 🙇‍♂️

TTOzzi:
> Thank you. This looks great now. Can you rebase it now that #1651 is merged? Afterwards, I’ll trigger CI.

I'm done! Thank you @ahoppen 🙌

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test windows

PR #1705, opened 2023-05-25
build: add an option to control WMO

compnerd:
WMO requires the new swift-driver which requires an early swift-driver or a host toolchain that has Swift.  Since the only platform that guarantees this currently is Darwin, use a conditional value for the option, defaulting to `NO` for most other targets.  This should repair the build on Windows.

### DIFF
--- a/cmake/modules/AddSwiftHostLibrary.cmake
+++ b/cmake/modules/AddSwiftHostLibrary.cmake
@@ -56,11 +56,9 @@ function(add_swift_host_library name)
       -emit-module-interface-path;${module_interface_file}
       >)
 
-  if(NOT CMAKE_BUILD_TYPE STREQUAL "Debug")
+  if(CMAKE_VERSION VERSION_LESS 3.26.0 AND SWIFT_SYNTAX_ENABLE_WMO_PRE_3_26)
     target_compile_options(${name} PRIVATE
-        $<$<COMPILE_LANGUAGE:Swift>:
-        -wmo
-        >)
+        $<$<COMPILE_LANGUAGE:Swift>:-wmo>)
   endif()
 
   # NOTE: workaround for CMake not setting up include flags yet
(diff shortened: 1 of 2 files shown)


compnerd:
@swift-ci please test

cmake/modules/AddSwiftHostLibrary.cmake
```diff
@@ -56,7 +56,7 @@ function(add_swift_host_library name)
       -emit-module-interface-path;${module_interface_file}
       >)
 
-  if(NOT CMAKE_BUILD_TYPE STREQUAL "Debug")
+  if(SWIFT_DRIVER_ENABLE_WMO)
```
bnbarham, line 59: We don't want it enabled in debug, so seems like we'd still want the old check as well.

compnerd: How about sinking that into the default value check?

bnbarham: Sure, works for me 👍

bnbarham (approved):
Thanks @compnerd! I'll set this by default in https://github.com/apple/swift/pull/66043 as well (since we assume there's a host toolchain there).

cmake/modules/AddSwiftHostLibrary.cmake
compnerd: Discussed offline and settled on `SWIFT_SYNTAX_ENABLE_WMO_PRE_3_26`.  It was always meant to be `SWIFT_SYNTAX_ENABLE_WMO` but my brain wasn't braining very well.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:

compnerd:

PR #1706, opened 2023-05-25
Add a new test mutation strategy to flip the presence of token in the `assertParse` tests (fixes 5 parser bugs)

ahoppen:
The idea is simple: From the tree generated during parsing, iterate over all the tokens. For each token create a new test case that has the token missing, if it was present in the original source or present if it was missing in the original source. Then run the parser to check that it round-trips and check that generating diagnostics for the mutated tree doesn’t crash.

This testing strategy found three parser bugs:

1. A round-trip failure for unexpected modifiers in a tuple type
```swift
func test(a: (borrowing F o))
```
2. A round-trip failure for modifiers in front of an editor placeholder in declaration position
```swift
struct a {
  public<#declaration#>
}
```
3. A precondition failure when the primary associated type clause of a protocol is empty
```swift
protocol X<> {}
```
4. It also uncovered an issue where we weren’t able to retrieve the trivia pieces of a token after it had been modified using `with` because after the modification the token was a `parsedToken` that no longer resided in a `ParsingSyntaxArena`. The fix is to search through an arena’s `childRefs` to find the `ParsingSyntaxArena` that created it. @rintaro can you review this change?
5. It uncovered that we were making assumptions about the presence of tokens in `ParseDiagnosticGenerator`, mostly that tokens in `UnexpectedNodesSyntax` are present. While this assumption is valid for trees generated by `SwiftParser`, it doesn’t need to hold for manually generated trees. A couple of minor adjustments to filter for only present tokens fixes this.

rdar://109783066

### DIFF
--- a/Sources/SwiftParser/Nominals.swift
+++ b/Sources/SwiftParser/Nominals.swift
@@ -333,7 +333,7 @@ extension Parser {
   }
 
   mutating func parsePrimaryAssociatedTypes() -> RawPrimaryAssociatedTypeClauseSyntax {
-    let langle = self.consumeAnyToken(remapping: .leftAngle)
+    let langle = self.consumePrefix("<", as: .leftAngle)
     var associatedTypes = [RawPrimaryAssociatedTypeSyntax]()
     do {
       var keepGoing: RawTokenSyntax? = nil
--- a/Tests/SwiftParserTest/DeclarationTests.swift
+++ b/Tests/SwiftParserTest/DeclarationTests.swift
@@ -2222,4 +2222,14 @@ final class DeclarationTests: XCTestCase {
         """
     )
   }
+
+  func testEmptyPrimaryAssociatedType() {
+    assertParse(
+      "protocol X<1️⃣> {}",
+      diagnostics: [
+        DiagnosticSpec(message: "expected name in primary associated type clause", fixIts: ["insert name"])
+      ],
+      fixedSource: "protocol X<<#identifier#>> {}"
+    )
+  }
 }
--- a/Tests/SwiftParserTest/DeclarationTests.swift
+++ b/Tests/SwiftParserTest/DeclarationTests.swift
@@ -690,6 +690,19 @@ final class DeclarationTests: XCTestCase {
     )
   }
 
+  func testEditorPlaceholderWithModifier() {
+    assertParse(
+      """
+      struct a {
+        public1️⃣<#declaration#>
+      }
+      """,
+      diagnostics: [
+        DiagnosticSpec(message: "editor placeholder in source file")
+      ]
+    )
+  }
+
   func testMissingColonInFunctionSignature() {
     assertParse(
       "func test(first second 1️⃣Int)",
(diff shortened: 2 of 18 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test macOS

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

Sources/SwiftParser/Nominals.swift
```diff
@@ -333,7 +333,7 @@ extension Parser {
   }
 
   mutating func parsePrimaryAssociatedTypes() -> RawPrimaryAssociatedTypeClauseSyntax {
-    let langle = self.consumeAnyToken(remapping: .leftAngle)
+    let langle = self.consumePrefix("<", as: .leftAngle)
```
bnbarham, line 336: Did you search for `consumeAnyToken(.*leftAngle` and `rightAngle` to make sure we don't have any more of these 😅?

Tests/SwiftParserTest/DeclarationTests.swift
```diff
@@ -1920,7 +1933,7 @@ final class DeclarationTests: XCTestCase {
       }
       """,
       substructure: Syntax(
-        MemberDeclListItemSyntax(decl: EditorPlaceholderDeclSyntax(identifier: .identifier("<#code#>")))
+        MemberDeclListItemSyntax(decl: EditorPlaceholderDeclSyntax(placeholder: .identifier("<#code#>")))
```
bnbarham, line 1936: I really need to merge my refactoring PR. More conflicts :)!

Tests/SwiftParserTest/StatementTests.swift
```diff
@@ -96,7 +96,7 @@ final class StatementTests: XCTestCase {
   }
 
   func testNestedIfs() {
-    let nest = 20
+    let nest = 10
```
bnbarham, line 99: What was this one for? Do we hit the stack limit in debug at only 10 😓?

bnbarham (approved):

kimdv:
@swift-ci please test windows

Sources/SwiftParser/Nominals.swift
ahoppen: Yes, all other `consumeAnyToken(remapping:)` calls are fine.

Tests/SwiftParserTest/StatementTests.swift
ahoppen: I honestly don’t know what this test was meant to test. But I figured that there won’t really be any issues that we’d catch with 20 nested `if` statements that we wouldn’t catch if we nest 10.

And yes, this was hitting the maximum nesting level in debug builds. Not entirely sure why.

rintaro (approved):

PR #1708, opened 2023-05-26
Fix two memory corruption bugs when modifying SwiftSyntax nodes

ahoppen:
This PR fixes two different memory corruption bugs when modifying SwiftSyntax nodes

- `SyntaxData.forRoot` takes a `RawSyntax`, which holds a reference to its `SyntaxArena` but `RawSyntax` doesn’t retain its arena. We were already taking care of this in multiple places by adding a `withExtendedLifetime` for the arena around it but missed a case in `SyntaxData.replacingSelf`. Add an explicit `arena` parameter to that function to make sure the arena is always alive
- The `replacingChild` function took a `RawSyntax`, which again doesn’t retain its arena. When changing any syntax node’s child, we have a `value`, which is a `Syntax` node but all we did in the setter was to pass `value.raw` into `replacingChild`. This left an opportunity for the optimizer to deallocate the `Syntax` node (and thus it’s arena) before the call to `replacingChild` because all we needed from hat node was the `RawSyntax`. To fix this, add another version of `replacingChild` that takes a `SyntaxData` (which does retain the arena) and use that instead.

I measured performance of `swift-format` with this change applied and did not see a performance regression when linting swift-syntax.

rdar://109860357

### DIFF
--- a/Sources/SwiftSyntax/SyntaxData.swift
+++ b/Sources/SwiftSyntax/SyntaxData.swift
@@ -336,6 +336,15 @@ struct SyntaxData {
     return replacingSelf(newRaw, arena: arena)
   }
 
+  /// Identical to `replacingChild(at: Int, with: RawSyntax?, arena: SyntaxArena)`
+  /// that ensures that the arena of`newChild` doesn’t get de-allocated before
+  /// `newChild` has been addded to the result.
+  func replacingChild(at index: Int, with newChild: SyntaxData?, arena: SyntaxArena) -> SyntaxData {
+    return withExtendedLifetime(newChild) {
+      return replacingChild(at: index, with: newChild?.raw, arena: arena)
+    }
+  }
+
   func withLeadingTrivia(_ leadingTrivia: Trivia, arena: SyntaxArena) -> SyntaxData {
     if let raw = raw.withLeadingTrivia(leadingTrivia, arena: arena) {
       return replacingSelf(raw, arena: arena)
--- a/Sources/SwiftSyntax/SyntaxData.swift
+++ b/Sources/SwiftSyntax/SyntaxData.swift
@@ -273,11 +273,13 @@ struct SyntaxData {
     self.init(absoluteRaw.raw, parent: parent.data, absoluteInfo: absoluteRaw.info)
   }
 
-  /// Creates a `SyntaxData` for a root raw node.
+  /// Creates a ``SyntaxData`` for a root raw node.
   ///
-  /// `arena` must be the arena in which `raw` is allocated. It is passed to
-  /// make sure the arena doesn’t get de-allocated before the ``SyntaxData`` has
-  /// a chance to retain it.
+  /// - Parameters:
+  ///   - raw: The raw node that will be the root of the the tree
+  ///   - rawNodeArena: The arena in which `raw` is allocated. It is passed to
+  ///     make sure the arena doesn’t get de-allocated before the ``SyntaxData``
+  ///     has a chance to retain it.
   static func forRoot(_ raw: RawSyntax, rawNodeArena: SyntaxArena) -> SyntaxData {
     precondition(rawNodeArena === raw.arena)
     return SyntaxData(raw, info: .root(.init(arena: rawNodeArena)))
@@ -307,7 +309,7 @@ struct SyntaxData {
   /// - Parameters:
   ///   - newRaw: The node that should replace `self`
   ///   - rawNodeArena: The arena in which `newRaw` resides
-  ///   - arena: The arena in which  new nodes should be allocated
+  ///   - allocationArena: The arena in which  new nodes should be allocated
   /// - Returns: A syntax tree with all parents where this node has been
   ///            replaced by `newRaw`
   func replacingSelf(_ newRaw: RawSyntax, rawNodeArena: SyntaxArena, allocationArena: SyntaxArena) -> SyntaxData {
--- a/Sources/SwiftSyntax/SyntaxData.swift
+++ b/Sources/SwiftSyntax/SyntaxData.swift
@@ -274,8 +274,13 @@ struct SyntaxData {
   }
 
   /// Creates a `SyntaxData` for a root raw node.
-  static func forRoot(_ raw: RawSyntax) -> SyntaxData {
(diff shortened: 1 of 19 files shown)


ahoppen:
@swift-ci Please test

rintaro (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

Sources/SwiftSyntax/SyntaxData.swift
```diff
@@ -327,43 +331,52 @@ struct SyntaxData {
   ///   - index: The index pointing to where in the raw layout to place this
   ///            child.
   ///   - newChild: The raw syntax for the new child to replace.
+  ///   - newChildArena: The arena in which `newChild` resides.
   ///   - arena: The arena in which the new node will be allocated.
```
bnbarham: All the names no longer match up 😅 `newChild` is actually `with` in terms of the API, and `newChildArena == rawNodeArena`. `arena == allocationArena`

Sources/SwiftSyntax/SyntaxData.swift
```diff
@@ -327,43 +331,52 @@ struct SyntaxData {
...
   /// - Returns: The new root node created by this operation, and the new child
   ///            syntax data.
   /// - SeeAlso: replacingSelf(_:)
-  func replacingChild(at index: Int, with newChild: RawSyntax?, arena: SyntaxArena) -> SyntaxData {
-    let newRaw = raw.layoutView!.replacingChild(at: index, with: newChild, arena: arena)
-    return replacingSelf(newRaw, arena: arena)
+  func replacingChild(at index: Int, with newChild: RawSyntax?, rawNodeArena: SyntaxArena?, allocationArena: SyntaxArena) -> SyntaxData {
+    if let newChild {
+      precondition(newChild.arena === rawNodeArena)
```
bnbarham, line 341: Can this not just be `newChild?.arena == rawNodeArena`?

Sources/SwiftSyntax/SyntaxData.swift
```diff
@@ -327,43 +331,52 @@ struct SyntaxData {
...
+    let newRaw = withExtendedLifetime(rawNodeArena) {
+      raw.layoutView!.replacingChild(at: index, with: newChild, arena: allocationArena)
+    }
+    return replacingSelf(newRaw, rawNodeArena: allocationArena, allocationArena: allocationArena)
   }
 
   /// Identical to `replacingChild(at: Int, with: RawSyntax?, arena: SyntaxArena)`
   /// that ensures that the arena of`newChild` doesn’t get de-allocated before
   /// `newChild` has been addded to the result.
   func replacingChild(at index: Int, with newChild: SyntaxData?, arena: SyntaxArena) -> SyntaxData {
```
bnbarham, line 355: I still find this odd FWIW. I thought the idea behind `RawSyntax` was to avoid the retain, but using `SyntaxData` everywhere should remove that gain. Though I guess `replacingSelf` still just passes `RawSyntax`, so it'd only be a single retain vs one for each parent?

Sources/SwiftSyntax/SyntaxData.swift
ahoppen: If you generate this comment using Xcode (Cmd-Option-/), it will use the internal parameter names instead of the external ones 🤷🏽

Sources/SwiftSyntax/SyntaxData.swift
ahoppen: Changed to 
```
precondition(newChild?.arena === rawNodeArena || newChild == nil)
```

We don’t hit the `newChild == nil && rawNodeArena != nil` case at the moment AFAICT but there’s nothing really wrong with that, so I wanted to allow it.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

Sources/SwiftSyntax/SyntaxData.swift
bnbarham: Seems odd to me to allow it, they're supposed to be the same so if `newChild` is `nil` and `rawNodeArena` isn't that would suggest something strange is going on.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

Sources/SwiftSyntax/SyntaxData.swift
```diff
@@ -274,8 +274,13 @@ struct SyntaxData {
   }
 
   /// Creates a `SyntaxData` for a root raw node.
-  static func forRoot(_ raw: RawSyntax) -> SyntaxData {
-    SyntaxData(raw, info: .root(.init(arena: raw.arena)))
+  ///
+  /// `arena` must be the arena in which `raw` is allocated. It is passed to
```
bnbarham, line 278: `arena` is `rawNodeArena` now 😅

Sources/SwiftSyntax/SyntaxData.swift
```diff
@@ -298,21 +303,24 @@ struct SyntaxData {
...
+  ///
   /// - Parameters:
-  ///   - newRaw: The new RawSyntax that will back the new `Data`
-  ///   - arena: SyntaxArena to the result RawSyntax node data resides.
-  /// - Returns: A tuple of both the new root node and the new data with the raw
-  ///            layout replaced.
-  func replacingSelf(_ newRaw: RawSyntax, arena: SyntaxArena) -> SyntaxData {
+  ///   - newRaw: The node that should replace `self`
+  ///   - rawNodeArena: The arena in which `newRaw` resides
+  ///   - arena: The arena in which  new nodes should be allocated
```
bnbarham, line 310: And this `arena` is `allocationArena`

bnbarham (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

kimdv:
@swift-ci please test windows

PR #1710, opened 2023-05-28
Add @_used and @_section attributes to match #65901

kubamracek:
See <https://github.com/apple/swift/pull/65901>.

### DIFF
--- a/CodeGeneration/Sources/SyntaxSupport/AttributeKinds.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/AttributeKinds.swift
@@ -652,6 +652,16 @@ public let DECL_ATTR_KINDS: [Attribute] = [
     className: "RuntimeMetadata",
     swiftName: "runtimeMetadata"
   ),
+  SimpleDeclAttribute(
+    name: "_used",
+    className: "Used",
+    swiftName: "_used"
+  ),
+  DeclAttribute(
+    name: "_section",
+    className: "Section",
+    swiftName: "_section"
+  ),
 ]
 
 /// Schema for declaration modifiers:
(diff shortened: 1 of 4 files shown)


kubamracek:
@swift-ci please test

kimdv:
Hi @kubamracek 👋

You need to run the code generation to get attribute kinds updated with your changes.
You can read about it here: https://github.com/apple/swift-syntax/blob/main/CodeGeneration/README.md

kubamracek:
@kimdv Thanks for the info, I actually did run the regenerate step, but the SimpleDeclAttributes seem to simply have no effect (no need?) on the generated swift files?

kubamracek:
...which made me realize that `@_used` is supposed to be an SimpleDeclAttribute, but `@_section` isn't and it's supposed to have arguments, so I need to do some more work here :)

kubamracek:
@swift-ci please test

kubamracek:
@swift-ci please test

ahoppen (commented):
Hey Kuba,

Thanks for updating the parser. And it indeed looks like we are no longer using `DECL_ATTR_KINDS`. I’m removing it a follow-up PR.

Could you add a test case for `@_used` and `@_section`, e.g. in AttributeTests.swift?

ahoppen:
And I’m removing the code generation of decl modifiers and type attributes here: https://github.com/apple/swift-syntax/pull/1715

DougGregor:
Do we need this at all? They can be parsed as custom attributes, and it's fine, right?

ahoppen:
Ah, true. We would need if `@_used` and `@_section` were type attributes to disambiguate `@_used () -> Void` but that doesn’t appear to be the case.

kubamracek:
Ha! So I don't even need the parser changes (section attr takes a string argument), and I can just close this PR?

ahoppen:
Yes, looks like it.

PR #1712, opened 2023-05-28
Add utils for incremental parse test

StevenWong12:
This change should allow us to verify the correctness of incremental parse with a single function, containing round-trip, substructure, and reused nodes validation like `assertParse`.

### DIFF
--- a/Sources/_SwiftSyntaxTestSupport/IncrementalParseTestUtils.swift
+++ b/Sources/_SwiftSyntaxTestSupport/IncrementalParseTestUtils.swift
@@ -1,0 +1,137 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+import SwiftParser
+import XCTest
+
+/// This function is used to verify the correctness of incremental parsing
+/// containing three parts:
+/// 1. Round-trip on the incrementally parsed syntax tree.
+/// 2. verify that incrementally parsing the edited source base on the original source produces the same syntax tree as reparsing the post-edit file from scratch.
+/// 3. verify the reused nodes fall into expectations.
+public func assertIncrementalParse(
+  _ source: String,
+  reusedNodes expectedReusedNodes: [ReusedNodeSpec] = [],
+  file: StaticString = #file,
+  line: UInt = #line
+) {
+  let (concurrentEdits, originalSource, editedSource) = getEditsAndSources(source)
+
+  let originalString = String(originalSource)
+  let editedString = String(editedSource)
+
+  let originalTree = Parser.parse(source: originalString)
+
+  let reusedNodesCollector = IncrementalParseReusedNodeCollector()
+  let transition = IncrementalParseTransition(previousTree: originalTree, edits: concurrentEdits, reusedNodeDelegate: reusedNodesCollector)
+
+  let newTree = Parser.parse(source: editedString)
+  let incrementallyParsedNewTree = Parser.parse(source: editedString, parseTransition: transition)
+
+  // Round-trip
+  assertStringsEqualWithDiff(
+    editedString,
+    "\(incrementallyParsedNewTree)",
+    additionalInfo: """
+      Source failed to round-trip when parsing incrementally
+
+      Actual syntax tree:
+      \(incrementallyParsedNewTree.debugDescription)
+      """,
+    file: file,
+    line: line
+  )
+
+  // Substructure
+  let subtreeMatcher = SubtreeMatcher(Syntax(incrementallyParsedNewTree), markers: [:])
+  do {
+    try subtreeMatcher.assertSameStructure(Syntax(newTree), includeTrivia: true, file: file, line: line)
+  } catch {
+    XCTFail("Matching for a subtree failed with error: \(error)", file: file, line: line)
+  }
+
+  // Re-used nodes
+  if reusedNodesCollector.rangeAndNodes.count != expectedReusedNodes.count {
+    XCTFail(
+      """
+      Expected \(expectedReusedNodes.count) re-used nodes but received \(reusedNodesCollector.rangeAndNodes.count):
+      \(reusedNodesCollector.rangeAndNodes.map {$0.1.description}.joined(separator: "\n"))
+      """,
+      file: file,
+      line: line
+    )
+    return
+  }
+
+  for expectedReusedNode in expectedReusedNodes {
+    guard let range = getByteSourceRange(for: expectedReusedNode.source, in: originalString) else {
+      XCTFail("Fail to find string in original source,", file: expectedReusedNode.file, line: expectedReusedNode.line)
+      continue
+    }
+
+    guard let reusedNode = reusedNodesCollector.rangeAndNodes.first(where: { $0.0 == range })?.1 else {
+      XCTFail(
+        """
+        Fail to match the range of \(expectedReusedNode.source) in:
+        \(reusedNodesCollector.rangeAndNodes.map({"\($0.0): \($0.1.description)"}).joined(separator: "\n"))
+        """,
+        file: expectedReusedNode.file,
+        line: expectedReusedNode.line
+      )
+      continue
+    }
+
+    XCTAssertEqual(
+      expectedReusedNode.kind,
+      expectedReusedNode.kind,
+      """
+      Expected \(expectedReusedNode.kind) syntax kind but received \(reusedNode.kind)
+      """,
+      file: expectedReusedNode.file,
+      line: expectedReusedNode.line
+    )
+  }
+}
+
+fileprivate func getByteSourceRange(for substring: String, in sourceString: String) -> ByteSourceRange? {
+  if let range = sourceString.range(of: substring) {
+    return ByteSourceRange(
+      offset: sourceString.utf8.distance(from: sourceString.startIndex, to: range.lowerBound),
+      length: sourceString.utf8.distance(from: range.lowerBound, to: range.upperBound)
+    )
+  }
+  return nil
+}
+
+/// An abstract data structure to describe the how a re-used node produced by the incremental parse should look like.
+public struct ReusedNodeSpec {
+  /// The re-used string in original source without any ``Trivia``
+  let source: String
+  /// The ``SyntaxKind`` of re-used node
+  let kind: SyntaxKind
+  /// The file and line at which this ``ReusedNodeSpec`` was created, so that assertion failures can be reported at its location.
+  let file: StaticString
+  let line: UInt
+
+  public init(
+    _ source: String,
+    kind: SyntaxKind,
+    file: StaticString = #file,
+    line: UInt = #line
+  ) {
+    self.source = source
+    self.kind = kind
+    self.file = file
+    self.line = line
+  }
+}
--- a/Tests/SwiftParserTest/IncrementalParsingTests.swift
+++ b/Tests/SwiftParserTest/IncrementalParsingTests.swift
@@ -18,55 +18,23 @@ import _SwiftSyntaxTestSupport
 public class IncrementalParsingTests: XCTestCase {
 
   public func testIncrementalInvalid() {
-    let source = "struct A⏩️⏸️A⏪️ { func f() {"
-    let (concurrentEdits, originalSubString, editedSubString) = getEditsAndSources(source)
-
-    let originalSource = String(originalSubString)
-    let editedSource = String(editedSubString)
-
-    var tree = Parser.parse(source: originalSource)
-
-    let lookup = IncrementalParseTransition(previousTree: tree, edits: concurrentEdits)
-    tree = Parser.parse(source: editedSource, parseTransition: lookup)
-    XCTAssertEqual("\(tree)", editedSource)
+    assertIncrementalParse(
+      """
+      struct A⏩️⏸️A⏪️ { func f() {
+      """
+    )
   }
 
   public func testReusedNode() throws {
     try XCTSkipIf(true, "Swift parser does not handle node reuse yet")
-
-    let source =
+    assertIncrementalParse(
       """
       struct A⏩️⏸️A⏪️ {}
       struct B {}
-      """
-
-    let (concurrentEdits, originalSubString, editedSubString) = getEditsAndSources(source)
-
-    let originalSource = String(originalSubString)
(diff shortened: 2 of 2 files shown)


Sources/_SwiftSyntaxTestSupport/IncrementalParseTestUtils.swift
```diff
@@ -0,0 +1,127 @@
...
+import SwiftSyntax
+import SwiftParser
+import XCTest
+
+/// This function is used to verify the correctness of incremental parsing
+/// containing three parts:
+/// 1. Round-trip on the incrementally parsed syntax tree.
+/// 2. verify that incrementally parsing the edited source base on the original source produces the same syntax tree as reparsing the post-edit file from scratch.
+/// 3. verify the reused nodes fall into expectations.
+public func verifyIncrementalParse(
```
ahoppen, line 22: The naming convention for these assertion functions in swift-syntax would be `assertIncrementalParse`.

Sources/_SwiftSyntaxTestSupport/IncrementalParseTestUtils.swift
```diff
@@ -0,0 +1,127 @@
...
+  originalSource: String,
+  editedSource: String,
+  concurrentEdits: ConcurrentEdits,
+  expectedReusedRanges: [(ByteSourceRange, SyntaxKind)]
```
ahoppen, line 26: If you look at e.g. `assertParse`, it takes a defaulted `file` and `line` parameter that is passed to all `XCTAssert*` functions. That way, if an assertion fails, it is shown at the `assertIncrementalParse` call instead of inside this function, which generally makes it a lot easier to debug.

Sources/_SwiftSyntaxTestSupport/IncrementalParseTestUtils.swift
```diff
@@ -0,0 +1,127 @@
...
+
+/// This function is used to verify the correctness of incremental parsing
+/// containing three parts:
+/// 1. Round-trip on the incrementally parsed syntax tree.
+/// 2. verify that incrementally parsing the edited source base on the original source produces the same syntax tree as reparsing the post-edit file from scratch.
+/// 3. verify the reused nodes fall into expectations.
+public func verifyIncrementalParse(
+  originalSource: String,
+  editedSource: String,
+  concurrentEdits: ConcurrentEdits,
```
ahoppen, line 25: What do you think about passing the marked source to the `verifyIncrementalParse` function and running `extractMarkersForIncrementParseTest` from here. That should simplify the assert invocations.

Sources/_SwiftSyntaxTestSupport/IncrementalParseTestUtils.swift
```diff
@@ -0,0 +1,127 @@
...
+  let newTree = Parser.parse(source: editedSource)
+  let incrementallyParsedNewTree = Parser.parse(source: editedSource, parseTransition: transition)
+
+  // Round-trip
+  assertStringsEqualWithDiff(editedSource, "\(incrementallyParsedNewTree)")
+
+  // Substructure
+  let subtreeMatcher = SubtreeMatcher(Syntax(incrementallyParsedNewTree), markers: [:])
+  do {
+    try subtreeMatcher.assertSameStructure(Syntax(newTree))
```
ahoppen, line 43: Could you also set `includeTrivia: true` to verify that the trivia between of `newTree` and `incrementallyParsedNewTree` match?

Sources/_SwiftSyntaxTestSupport/IncrementalParseTestUtils.swift
```diff
@@ -0,0 +1,127 @@
...
+  } catch {
+    XCTFail("Matching for a subtree failed with error: \(error)")
+  }
+
+  // Re-used nodes
+  // The range of node should ignore Trivia
+  if expectedReusedRanges.isEmpty {
+    XCTAssert(reusedNodesCollector.rangeAndNodes.isEmpty)
```
ahoppen, line 51: Do you need this special case? To me it looks like this is covered by the `else` branch as well

Sources/_SwiftSyntaxTestSupport/IncrementalParseTestUtils.swift
```diff
@@ -0,0 +1,127 @@
...
+  // Round-trip
+  assertStringsEqualWithDiff(editedSource, "\(incrementallyParsedNewTree)")
+
+  // Substructure
+  let subtreeMatcher = SubtreeMatcher(Syntax(incrementallyParsedNewTree), markers: [:])
+  do {
+    try subtreeMatcher.assertSameStructure(Syntax(newTree))
+  } catch {
+    XCTFail("Matching for a subtree failed with error: \(error)")
+  }
```
ahoppen, line 46: You could just make this function `throws` and then you don’t need to deal with `do {} catch {}` blocks

Tests/SwiftSyntaxTestSupportTest/IncrementalParseTestUtilsTest.swift
```diff
@@ -0,0 +1,130 @@
...
+    )
+  }
+
+  public func testReuseRangeInsideMultipleEdits() {
+    let source =
+      """
+      ⏩️⏸️class foo {⏪️
+        ➡️let value1 = 👨‍👩‍👧‍👧⬅️
+      ⏩️⏸️}⏪️
+      """
```
ahoppen, line 96: I feel like we are getting into marker hell here with the two different kinds of markers. Just an idea: What do you think about adding a `reused: String...` parameter to `assertIncrementalParse`. In this case you would be passing `let value1 = 👨‍👩‍👧‍👧` to that parameter to assert that a node with that source text has been reused. We do something similar with the `highlight` parameter on `DiagnosticSpec`.

ahoppen (commented):
Thank you. I think we are going in the right direction by having a single `assert` function that performs all these checks because that did turn out super valuable for `assertParse`. I’ve got a few design comments inline.

Sources/_SwiftSyntaxTestSupport/IncrementalParseTestUtils.swift
StevenWong12: Well, I just want to keep this the same as `assertParse`. We do error handling here so that we don't have to mark test cases throwable or add some do-catch blocks there, which looks nicer IMO.

StevenWong12: Well, I just want to keep this the same as `assertParse`. We do error handling here so that we don't have to mark test cases throwable or add some do-catch blocks there, which looks nicer IMO.

Tests/SwiftSyntaxTestSupportTest/IncrementalParseTestUtilsTest.swift
StevenWong12: Yeah, I also added a `ReusedNodeSpec` in the new commit, which makes the usage of this function very much like `assertParse`.

Sources/_SwiftSyntaxTestSupport/IncrementalParseTestUtils.swift
StevenWong12: That's because in the previous design, we have to zip the re-used range and re-used nodes in the test case on our own. So I just separate this two functions.

I have changed this in the new commit.

StevenWong12:
What do you think about cherry-picking this commit to the functionality pr (https://github.com/apple/swift-syntax/pull/1685) and just close this one? Since we could just test this util with the incremental parse test case. (I don't know how `assertParse` was tested)

Sources/_SwiftSyntaxTestSupport/IncrementalParseTestUtils.swift
```diff
@@ -0,0 +1,111 @@
...
+
+public struct ReusedNodeSpec {
+  public let sourceString: String
+  
+  public let kind: SyntaxKind
+  
+  public init(
+    sourceString: String,
+    kind: SyntaxKind
+  ) {
```
ahoppen, line 107: Just an idea to make the tests a little bit more readable: I’d make the first parameter unlabeled. But that’s very much personal preference.

Sources/_SwiftSyntaxTestSupport/IncrementalParseTestUtils.swift
```diff
@@ -0,0 +1,111 @@
...
+  
+  let originalString = String(originalSource)
+  let editedString = String(editedSource)
+  
+  let originalTree = Parser.parse(source: originalString)
+
+  let reusedNodesCollector = IncrementalParseReusedNodeCollector()
+  let transition = IncrementalParseTransition(previousTree: originalTree, edits: concurrentEdits, reusedNodeDelegate: reusedNodesCollector)
+
+  let newTree = Parser.parse(source: String(editedSource))
```
ahoppen, line 38:
```suggestion
  let newTree = Parser.parse(source: editedString)
```

Tests/SwiftParserTest/IncrementalParsingTests.swift
```diff
@@ -18,55 +18,22 @@ import _SwiftSyntaxTestSupport
...
-    var tree = Parser.parse(source: originalSource)
-
-    let lookup = IncrementalParseTransition(previousTree: tree, edits: concurrentEdits)
-    tree = Parser.parse(source: editedSource, parseTransition: lookup)
-    XCTAssertEqual("\(tree)", editedSource)
+    assertIncrementalParse(
+      """
+      struct A⏩️⏸️A⏪️ { func f() {
+      """
+    )
```
ahoppen, line 25: These look so nice and short now 🤩

Tests/SwiftSyntaxTestSupportTest/IncrementalParseTestUtilsTest.swift
ahoppen: Thanks. It’s looking very good

Sources/_SwiftSyntaxTestSupport/IncrementalParseTestUtils.swift
ahoppen: Ah, yeah. Makes absolute sense. Thanks for clarifying.

ahoppen:
> What do you think about cherry-picking this commit to the functionality pr (#1685) and just close this one? Since we could just test this util with the incremental parse test case. (I don't know how `assertParse` was tested)

I think keeping this as a standalone PR makes sense. It provides value on its own and will allow us to keep the PR that actually implements incremental parsing focussed. Also: I think we’ve got this PR really close to being merged.

Sources/_SwiftSyntaxTestSupport/IncrementalParseTestUtils.swift
```diff
@@ -0,0 +1,111 @@
...
+  }
+}
+
+private func getByteSourceRange(for subString: String, in sourceString: String) -> ByteSourceRange? {
+  
+  if let range = sourceString.range(of: subString) {
+    
+    return ByteSourceRange(
+      offset: sourceString.utf8.distance(from: sourceString.startIndex, to: range.lowerBound),
+      length: sourceString.utf8.distance(from: range.lowerBound, to: range.upperBound))
```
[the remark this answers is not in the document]
StevenWong12: Oh, apparently I forgot to run `swift-format`. Sorry for that.

StevenWong12:
Great! let's just keep this.

Sources/_SwiftSyntaxTestSupport/IncrementalParseTestUtils.swift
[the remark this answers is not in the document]
ahoppen: No problem at all

Sources/_SwiftSyntaxTestSupport/IncrementalParseTestUtils.swift
```diff
@@ -0,0 +1,111 @@
...
+  }
+  return nil
+}
+
+public struct ReusedNodeSpec {
+  public let sourceString: String
```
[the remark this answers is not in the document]
ahoppen: OK, great!

Sources/_SwiftSyntaxTestSupport/IncrementalParseTestUtils.swift
```diff
@@ -0,0 +1,135 @@
...
+    XCTFail("Matching for a subtree failed with error: \(error)", file: file, line: line)
+  }
+
+  // Re-used nodes
+  XCTAssertEqual(
+    reusedNodesCollector.rangeAndNodes.count,
+    expectedReusedNodes.count,
+    """
+    Expected \(expectedReusedNodes.count) re-used nodes but received \(reusedNodesCollector.rangeAndNodes.count)
+    """,
```
ahoppen, line 69: In my experience, when the count of the diagnostics (and in this case reused ranges) doesn’t match, it doesn’t make sense to show how the specified ranges mismatched. So in that case, I would just not continue to the `for` loop below, similar to 

https://github.com/apple/swift-syntax/blob/873cbce8a3cde64807dba55e016ae2745cfcb5d8/Tests/SwiftParserTest/Assertions.swift#L589

To make this actionable, we should instead show the actually reused pieces of the source code as in

https://github.com/apple/swift-syntax/blob/873cbce8a3cde64807dba55e016ae2745cfcb5d8/Tests/SwiftParserTest/Assertions.swift#L593

Sources/_SwiftSyntaxTestSupport/IncrementalParseTestUtils.swift
```diff
@@ -0,0 +1,135 @@
...
+    file: file,
+    line: line
+  )
+
+  for targetNode in expectedReusedNodes {
```
ahoppen, line 74: These aren’t really nodes, right? Nitpick: `targetNode` is a little ambiguous. What do you think about `expectedReusedNode`?

ahoppen (commented):
Nice. I’ve got a few minor comments left, otherwise LGTM.

ahoppen:
Also, I think you might to need to rebase on top of main and run swift-format to pick up https://github.com/apple/swift-syntax/pull/1714.

Sources/_SwiftSyntaxTestSupport/IncrementalParseTestUtils.swift
StevenWong12: Yeah, these aren't actually nodes. I just can't come up with a more proper name for them😂. I've renamed `targetNode` to `expectedReusedNode`.

StevenWong12:
I've rebased this pr and re-run `swift-format`.

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test windows

ahoppen:
Looks like `XCTExpectFailure` is not available in XCTest on Linux. So we need to continue using

```swift
try XCTSkipIf(true, "Swift parser does not handle node reuse yet")

```

StevenWong12:
Ah, I see. Already fixed this.

ahoppen:
@swift-ci Please test

PR #1716, opened 2023-05-30
Rename `TupleExprSyntax.elementList` to  `TupleExprSyntax.elements`

### LINKED ISSUE #1713: Inconsistent naming scheme for the elements of tuple-like syntax nodes
### Description

`TupleExprSyntax.elementList` vs. `TupleTypeSyntax.elements`. I realize I am being pedantic but I also see no good reason to name them differently. No preference for a particular name, but perhaps worth noting that `ArraySyntaxExpr` uses `elements` and `FunctionTypeSyntax` uses `arguments`, i.e., without "list".

kimdv:
Resolves #1713

This is to be more consistent in naming

### DIFF
--- a/Sources/SwiftSyntax/SwiftSyntaxCompatibility.swift
+++ b/Sources/SwiftSyntax/SwiftSyntaxCompatibility.swift
@@ -116,3 +116,39 @@ public extension NamedOpaqueReturnTypeSyntax {
     )
   }
 }
+
+extension TupleExprSyntax {
+  @available(*, deprecated, renamed: "unexpectedBetweenLeftParenAndElements")
+  public var unexpectedBetweenLeftParenAndElementList: UnexpectedNodesSyntax? { unexpectedBetweenLeftParenAndElements }
+
+  @available(*, deprecated, renamed: "elements")
+  public var elementList: TupleExprElementListSyntax { elements }
+
+  @available(*, deprecated, renamed: "unexpectedBetweenElementsAndRightParen")
+  public var unexpectedBetweenElementListAndRightParen: UnexpectedNodesSyntax? { unexpectedBetweenElementsAndRightParen }
+
+  @available(*, deprecated, message: "Use an initializer with a elements argument")
+  public init(
+    leadingTrivia: Trivia? = nil,
+    _ unexpectedBeforeLeftParen: UnexpectedNodesSyntax? = nil,
+    leftParen: TokenSyntax = .leftParenToken(),
+    _ unexpectedBetweenLeftParenAndElementList: UnexpectedNodesSyntax? = nil,
+    elementList: TupleExprElementListSyntax,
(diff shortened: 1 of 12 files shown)


Sources/SwiftSyntaxBuilder/generated/BuildableNodes.swift
```diff
@@ -1350,20 +1350,20 @@ extension TupleExprSyntax {
       leadingTrivia: Trivia? = nil, 
       unexpectedBeforeLeftParen: UnexpectedNodesSyntax? = nil, 
       leftParen: TokenSyntax = .leftParenToken(), 
-      unexpectedBetweenLeftParenAndElementList: UnexpectedNodesSyntax? = nil, 
-      unexpectedBetweenElementListAndRightParen: UnexpectedNodesSyntax? = nil, 
+      unexpectedBetweenLeftParenAndElements: UnexpectedNodesSyntax? = nil,
```
kimdv, line 1353: While I was making I was thinking, that those `UnexpectedNodesSyntax` in syntax builder don't make that much sense? 
Because we are generating code here and AFAIK `UnexpectedNodesSyntax` is more related to parsing?

ahoppen: Since all the other initializers that you use to create nodes, which don’t take a result builder (e.g. `IntegerLiteralExprSyntax` also take `UnexpectedNodesSyntax`, I think it would be inconsistent if you couldn’t specify unexpected nodes for these initializers that take a result builder.

Sources/SwiftSyntax/SwiftSyntaxCompatibility.swift
```diff
@@ -116,3 +116,40 @@ public extension NamedOpaqueReturnTypeSyntax {
...
+    leadingTrivia: Trivia? = nil,
+    _ unexpectedBeforeLeftParen: UnexpectedNodesSyntax? = nil,
+    leftParen: TokenSyntax = .leftParenToken(),
+    _ unexpectedBetweenLeftParenAndElementList: UnexpectedNodesSyntax? = nil,
+    elementList: TupleExprElementListSyntax,
+    _ unexpectedBetweenElementListAndRightParen: UnexpectedNodesSyntax? = nil,
+    rightParen: TokenSyntax = .rightParenToken(),
+    _ unexpectedAfterRightParen: UnexpectedNodesSyntax? = nil,
+    trailingTrivia: Trivia? = nil
+
```
ahoppen, line 141: Superfluous newline

ahoppen (approved):

Sources/SwiftSyntaxBuilder/generated/BuildableNodes.swift
kimdv: I see ✌️

kimdv:
@swift-ci please test

Issue #1606, opened 2023-04-28
Wrong Fix-Its applied for `_unavailableFromAsync` attribute 


kimdv:
### Description

https://github.com/apple/swift-syntax/pull/1549#discussion_r1179886392
Test case `testUnavailableFromAsync`

ahoppen:
Tracked in Apple’s issue tracker as rdar://108655875

TiagoMaiaL:
@ahoppen, @kimdv I would like to start working on this issue. Can I get assigned to it?

It's not clear to me what the `fixedSource` should be, as I'm not acquainted with the `@_unavailableFromAsync` attribute.

kimdv:
You can read the proposal here: https://github.com/apple/swift-evolution/blob/main/proposals/0340-swift-noasync.md

The fixed source should be `@_unavailableFromAsync(message: "abc")`

Right now it is `@_unavailableFromAsync(message: "" = "abc") `

kimdv:
So I think if you just consume the `=` then you are good to go 😁

mininny:
It seems like this was now resolved 👍 https://github.com/apple/swift-syntax/pull/1733

PR #1717, opened 2023-05-31
Improve diagnostics for invalid version tuples

### LINKED ISSUE #1613: `10e10` is not emitted as an invalid version tuple
### Description

https://github.com/apple/swift-syntax/pull/1549#discussion_r1179886829

Test case `testVersionParsing`

### Steps to Reproduce

_No response_

TTOzzi:
Resolve #1613

### DIFF
--- a/Sources/SwiftParser/Lookahead.swift
+++ b/Sources/SwiftParser/Lookahead.swift
@@ -254,7 +254,7 @@ extension Parser.Lookahead {
     // If we have a 'didSet' or a 'willSet' label, disambiguate immediately as
     // an accessor block.
     let nextToken = self.peek()
-    if TokenSpec(.didSet) ~= nextToken || TokenSpec(.willSet) ~= nextToken {
+    if TokenSpec(.didSet) ~= nextToken || TokenSpec(.willSet) ~= nextToken || TokenSpec(.`init`) ~= nextToken {
       return true
     }
 
@@ -278,8 +278,8 @@ extension Parser.Lookahead {
       }
     }
 
-    // Check if we have 'didSet'/'willSet' after attributes.
-    return lookahead.at(.keyword(.didSet), .keyword(.willSet))
+    // Check if we have 'didSet'/'willSet' or 'init' after attributes.
+    return lookahead.at(.keyword(.didSet), .keyword(.willSet), .keyword(.`init`))
   }
 }
 
--- a/Tests/SwiftParserTest/DeclarationTests.swift
+++ b/Tests/SwiftParserTest/DeclarationTests.swift
@@ -2529,4 +2529,82 @@ final class DeclarationTests: XCTestCase {
       fixedSource: "let foo: [Int] = []"
     )
   }
+
+  func testInitAccessorsWithDefaultValues() {
+    assertParse(
+      """
+      struct Test {
+        var pair: (Int, Int) = (42, 0) {
+          init(initialValue) {}
+
+          get { (0, 42) }
+          set { }
+        }
+      }
+      """
+    )
+
+    assertParse(
+      """
+      struct Test {
+        var pair: (Int, Int) = (42, 0) {
+          init initializes(a) {}
+
+          get { (0, 42) }
+          set { }
+        }
+      }
+      """
+    )
+
+    assertParse(
+      """
+      struct Test {
+        var pair: (Int, Int) = (42, 0) {
+          get { (0, 42) }
+          set { }
+
+          init(initialValue1️⃣) {}
+        }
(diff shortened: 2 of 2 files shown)


Tests/SwiftParserTest/AttributeTests.swift
```diff
@@ -533,10 +533,10 @@ final class AttributeTests: XCTestCase {
...
         DiagnosticSpec(message: "expected ':' and string literal in @_unavailableFromAsync argument", fixIts: ["insert ':' and string literal"]),
-        DiagnosticSpec(message: #"unexpected code '= "abc"' in attribute"#),
+        DiagnosticSpec(message: #"unexpected code '= "abc"' in attribute"#, fixIts: [#"remove '= "abc"'"#]),
       ],
       fixedSource: """
-        @_unavailableFromAsync(message: ""= "abc")
+        @_unavailableFromAsync(message: "")
         func foo() {}
         """
     )
```
TTOzzi, line 542: My work have affected the `@_unavailableFromAsync` test case.
It seems that a diagnosis should be generated that replaces `=` with `:` instead of removing `abc` 🤔 
I have only modified the result to ensure that it does not fail, as this issue is being tracked in #1606.

kimdv: This is properly another bug in the parser.
I think it's not adding a missing `:` when it comes to the `= abc`

Tests/SwiftParserTest/AvailabilityTests.swift
```diff
@@ -149,10 +149,10 @@ final class AvailabilityTests: XCTestCase {
       """,
       diagnostics: [
         DiagnosticSpec(message: "expected version tuple in version restriction", fixIts: ["insert version tuple"]),
-        DiagnosticSpec(message: "unexpected code '10e10' in attribute"),
+        DiagnosticSpec(message: "unexpected code '10e10' in attribute", fixIts: ["remove '10e10'"]),
```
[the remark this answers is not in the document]
TTOzzi: Ah, I see. I will give it a try!

Tests/SwiftParserTest/AttributeTests.swift
TTOzzi: A-ha! Thank you for explaining.

Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift
```diff
@@ -287,6 +287,26 @@ public struct CannotParseVersionTuple: ParserError {
...
+        if expectedNode.position > unexpectedNodes.position {
+          return "expected \(expectedNodeTypeName), code '\(unexpectedNodes.description + expectedNode.description)' is not the valid \(expectedNodeTypeName)"
+        } else if expectedNode.position < unexpectedNodes.position {
+          return "expected \(expectedNodeTypeName), code '\(expectedNode.description + unexpectedNodes.description)' is not the valid \(expectedNodeTypeName)"
+        }
+      }
+    }
+    return "unexpected \(unexpectedNodes.shortSingleLineContentDescription)"
+  }
+}
```
TTOzzi, line 308: I can't think of a suitable error message and type name 🤔  
If you have any suggestions for a better message, I'll make the changes.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -459,6 +459,29 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
       )
       return .visitChildren
     }
+    if let unexpectedBetweenArgumentAndRightParen = node.unexpectedBetweenArgumentAndRightParen,
+      var missingArgumentNode = node.argument?.missingNodes.only
```
ahoppen, line 463: `missingNodes` seems a little magical to me and I get the feeling that it’s not the right design because you need to patch it up further below. I think a better design would be to

- Modify the parser and eat e.g. `0xff` as an unexpected token when parsing the `VersionTupleSyntax`.
- Generate the diagnostic in `visit(_ node: VersionTupleSyntax)`

TTOzzi: Aha I'll try again. Thanks for the feedback! 🙇

keith:
It looks like you have accidentally pulled in a few unrelated commits here

TTOzzi:
I apologize profusely 😓 It appears that there might be an issue with my branch.

keith:
No worries!

PR #1719, opened 2023-06-01
[5.9] Fix two memory corruption bugs in SwiftSyntax

ahoppen:
This is a slightly less risky version of https://github.com/apple/swift-syntax/pull/1708

--- 

* **Explanation**: This PR fixes three different potential memory corruption bugs when modifying SwiftSyntax nodes
  * `SyntaxData.forRoot` takes a `RawSyntax`, which holds a reference to its `SyntaxArena` but `RawSyntax` doesn’t retain its arena. We were already taking care of this in multiple places by adding a `withExtendedLifetime` for the arena around it but missed a case in `SyntaxData.replacingSelf`. Add `withExtendedLifetime` here as well.
  * The `replacingChild` function took a `RawSyntax`, which again doesn’t retain its arena. When changing any syntax node’s child, we have a `value`, which is a `Syntax` node but all we did in the setter was to pass `value.raw` into `replacingChild`. This left an opportunity for the optimizer to deallocate the `Syntax` node (and thus it’s arena) before the call to `replacingChild` because all we needed from hat node was the `RawSyntax`. To fix this, add another version of `replacingChild` that takes a `SyntaxData` (which does retain the arena) and use that instead.
  * `SyntaxArena.rewrite` passed a new `SyntaxArena` to `replacingSelf`, but `replacingSelf` expected to be passed the arena in which `rewritten` was allocated.
* **Scope**: Functions that modified syntax nodes
* **Risk**: Low, this effectively PR only adds `withExtendedLifetime` calls
* **Testing**: Ran unit tests and verified that `swift-format` doesn’t regress in performance
* **Issue**: rdar://109860357
* **Reviewer**:

### DIFF
--- a/CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodesFile.swift
+++ b/CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodesFile.swift
@@ -221,7 +221,7 @@ func syntaxNode(emitKind: String) -> SourceFileSyntax {
             AccessorDeclSyntax(
               """
               set(value) {
-                self = \(raw: node.name)(data.replacingChild(at: \(raw: index), with: value\(raw: child.isOptional ? "?" : "").raw, arena: SyntaxArena()))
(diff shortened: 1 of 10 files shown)


ahoppen:
@swift-ci Please test

bnbarham (approved):

rintaro (approved):

PR #1722, opened 2023-06-02
`Functiontypesyntax.arguments` should be named `parameters`

### LINKED ISSUE #1720: `FunctionTypeSyntax.arguments` should be named `parameters`
### Description

The non-generic parameters of a function declaration are *value* parameters, not type parameters. This means the input list of a function type is a parameter list, not an argument list. Contrast this with the *type* arguments of a bound generic type, which are substitutes for the generic type parameters of the corresponding type declaration.

kimdv:
Resolves #1720

### DIFF
--- a/Sources/SwiftSyntax/SwiftSyntaxCompatibility.swift
+++ b/Sources/SwiftSyntax/SwiftSyntaxCompatibility.swift
@@ -13,13 +13,6 @@
 // This file provides compatiblity aliases to keep dependents of SwiftSyntax building.
 // All users of the declarations in this file should transition away from them ASAP.
 
-public extension DeclGroupSyntax {
-  @available(*, deprecated, renamed: "memberBlock")
-  var members: MemberDeclBlockSyntax {
-    return self.memberBlock
-  }
-}
-
 @available(*, deprecated, renamed: "ImportPathSyntax")
 public typealias AccessPathSyntax = ImportPathSyntax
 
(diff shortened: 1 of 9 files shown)


kimdv:
@swift-ci please test

ahoppen (approved):
Thanks @AnthonyLatsis for reporting this and @kimdv for fixing it.

Sources/SwiftSyntax/SwiftSyntaxCompatibility.swift
```diff
@@ -85,6 +85,60 @@ public extension EnumDeclSyntax {
   }
 }
 
+public extension FunctionTypeSyntax {
+  @available(*, deprecated, renamed: "unexpectedBetweenLeftParenAndParameters")
+  var unexpectedBetweenLeftParenAndArguments: UnexpectedNodesSyntax? { unexpectedBetweenLeftParenAndParameters }
+
+  @available(*, deprecated, renamed: "parameters")
+  var arguments: TupleTypeElementListSyntax { parameters }
```
ahoppen, line 93: Ah, I totally forgot that these properties also need to be settable. Would you mind also changing all the other properties in this compatibility layer to be settable?

kimdv: Fixed

kimdv:
@swift-ci please test

PR #1727, opened 2023-06-03
Add diagnostic for wrong specialize label

### LINKED ISSUE #1605: Wrong diagnostics for `_specialize` attribute
### Description

https://github.com/apple/swift-syntax/pull/1549#discussion_r1179886228

https://github.com/apple/swift-syntax/blob/00b0ce7017707525b24d0d577277bdebbbc0c91a/CodeGeneration/Sources/SyntaxSupport/AttributeNodes.swift#L801

### Steps to Reproduce

_No response_

kimdv:
Resolves #1605

### DIFF
--- a/CodeGeneration/Sources/SyntaxSupport/AttributeNodes.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/AttributeNodes.swift
@@ -641,7 +641,15 @@ public let ATTRIBUTE_NODES: [Node] = [
     children: [
       Child(
         name: "Label",
-        kind: .node(kind: .token),
+        kind: .token(choices: [
+          .keyword(text: "target"),
+          .keyword(text: "availability"),
+          .keyword(text: "exported"),
+          .keyword(text: "kind"),
+          .keyword(text: "spi"),
+          .keyword(text: "spiModule"),
+          .keyword(text: "available"),
+        ]),
         nameForDiagnostics: "label",
         documentation: "The label of the argument"
       ),
--- a/CodeGeneration/Sources/generate-swiftsyntax/templates/swiftparser/ParserTokenSpecSetFile.swift
+++ b/CodeGeneration/Sources/generate-swiftsyntax/templates/swiftparser/ParserTokenSpecSetFile.swift
@@ -22,7 +22,12 @@ let parserTokenSpecSetFile = SourceFileSyntax(leadingTrivia: copyrightHeader) {
     for child in layoutNode.children {
       if case let .token(choices, _, _) = child.kind, choices.count > 1 {
         try! ExtensionDeclSyntax("extension \(raw: layoutNode.kind.syntaxType)") {
-          try EnumDeclSyntax("enum \(raw: child.name)Options: TokenSpecSet") {
+          try EnumDeclSyntax(
+            """
+            @_spi(Diagnostics)
+            public enum \(raw: child.name)Options: TokenSpecSet
+            """
+          ) {
             for choice in choices {
               switch choice {
               case .keyword(let keywordText):
@@ -66,6 +71,39 @@ let parserTokenSpecSetFile = SourceFileSyntax(leadingTrivia: copyrightHeader) {
                     SwitchCaseSyntax(
                       "case .\(token.varOrCaseName): return .\(token.varOrCaseName)"
                     )
+                  }
+                }
+              }
+            }
+
+            try VariableDeclSyntax(
+              """
+              /// Returns a token that satisfies the `TokenSpec` of this case.
+              ///
+              /// If the token kind of this spec has variable text, e.g. for an identifier, this returns a token with empty text.
+              @_spi(Diagnostics)
+              public var tokenSyntax: TokenSyntax
+              """
+            ) {
+              try SwitchExprSyntax("switch self") {
+                for choice in choices {
+                  switch choice {
+                  case .keyword(let keywordText):
+                    let keyword = KEYWORDS.first(where: { $0.name == keywordText })!
+                    SwitchCaseSyntax(
+                      "case .\(raw: keyword.escapedName): return .keyword(.\(raw: keyword.escapedName))"
+                    )
+                  case .token(let tokenText):
+                    let token = SYNTAX_TOKEN_MAP[tokenText]!
+                    if token.text != nil {
+                      SwitchCaseSyntax(
+                        "case .\(raw: token.varOrCaseName): return .\(raw: token.varOrCaseName)Token()"
+                      )
+                    } else {
+                      SwitchCaseSyntax(
+                        #"case .\#(raw: token.varOrCaseName): return .\#(raw: token.varOrCaseName)("")"#
+                      )
+                    }
                   }
                 }
               }
--- a/Sources/SwiftParser/Attributes.swift
+++ b/Sources/SwiftParser/Attributes.swift
@@ -638,55 +638,21 @@ extension Parser {
 }
 
 extension Parser {
-  enum SpecializeParameter: TokenSpecSet {
-    case target
-    case availability
-    case exported
-    case kind
-    case spi
-    case spiModule
-    case available
-
-    init?(lexeme: Lexer.Lexeme) {
-      switch PrepareForKeywordMatch(lexeme) {
-      case TokenSpec(.target): self = .target
-      case TokenSpec(.availability): self = .availability
-      case TokenSpec(.exported): self = .exported
-      case TokenSpec(.kind): self = .kind
-      case TokenSpec(.spi): self = .spi
-      case TokenSpec(.spiModule): self = .spiModule
-      case TokenSpec(.available): self = .available
-      default: return nil
-      }
-    }
-
-    var spec: TokenSpec {
-      switch self {
-      case .target: return .keyword(.target)
-      case .availability: return .keyword(.availability)
-      case .exported: return .keyword(.exported)
-      case .kind: return .keyword(.kind)
-      case .spi: return .keyword(.spi)
-      case .spiModule: return .keyword(.spiModule)
-      case .available: return .keyword(.available)
-      }
-    }
-  }
   mutating func parseSpecializeAttributeArgumentList() -> RawSpecializeAttributeArgumentListSyntax {
     var elements = [RawSpecializeAttributeArgumentListSyntax.Element]()
     // Parse optional "exported" and "kind" labeled parameters.
     var loopProgress = LoopProgressCondition()
-    while !self.at(.endOfFile, .rightParen, .keyword(.where)) && self.hasProgressed(&loopProgress) {
-      switch self.at(anyIn: SpecializeParameter.self) {
+    LOOP: while !self.at(.endOfFile, .rightParen, .keyword(.where)) && self.hasProgressed(&loopProgress) {
+      switch self.at(anyIn: LabeledSpecializeArgumentSyntax.LabelOptions.self) {
       case (.target, let handle)?:
-        let ident = self.eat(handle)
+        let label = self.eat(handle)
         let (unexpectedBeforeColon, colon) = self.expect(.colon)
         let declName = self.parseDeclReferenceExpr([.zeroArgCompoundNames, .keywordsUsingSpecialNames, .operators])
         let comma = self.consume(if: .comma)
         elements.append(
           .specializeTargetFunctionArgument(
             RawSpecializeTargetFunctionArgumentSyntax(
-              targetLabel: ident,
+              targetLabel: label,
               unexpectedBeforeColon,
               colon: colon,
               declName: declName,
@@ -696,14 +662,14 @@ extension Parser {
           )
         )
       case (.availability, let handle)?:
-        let ident = self.eat(handle)
+        let label = self.eat(handle)
         let (unexpectedBeforeColon, colon) = self.expect(.colon)
         let availability = self.parseAvailabilitySpecList()
         let (unexpectedBeforeSemi, semi) = self.expect(.semicolon)
         elements.append(
           .specializeAvailabilityArgument(
             RawSpecializeAvailabilityArgumentSyntax(
-              availabilityLabel: ident,
+              availabilityLabel: label,
               unexpectedBeforeColon,
               colon: colon,
               availabilityArguments: availability,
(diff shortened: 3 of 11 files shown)


Sources/SwiftParser/Attributes.swift
```diff
@@ -816,7 +816,8 @@ extension Parser {
         elements.append(
           .labeledSpecializeEntry(
             RawLabeledSpecializeEntrySyntax(
-              label: ident,
+              RawUnexpectedNodesSyntax([ident], arena: self.arena),
+              label: RawTokenSyntax(missing: .identifier, arena: self.arena),
```
kimdv, line 820: This is not correct. 
But I didn't know how we can add a missing token that is one of the correct choices.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -1120,6 +1120,33 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
+        UnknownParameterError(
+          parameter: unexpectedIdentifier,
+          validParameters: [
+            .keyword(.target),
+            .keyword(.availability),
+            .keyword(.exported),
+            .keyword(.kind),
+            .keyword(.spi),
+            .keyword(.spiModule),
+            .keyword(.available),
```
kimdv, line 1140: This is copied in. 

Should try to look into if we could use `enum SpecializeParameter: TokenSpecSet {}` for this? 

And second: Should we generate `SpecializeParameter` based on the label.choices in `AttributeNodes.swift`?

Tests/SwiftParserTest/AttributeTests.swift
```diff
@@ -73,41 +73,64 @@ final class AttributeTests: XCTestCase {
   func testMissingClosingParenToAttribute() {
     assertParse(
       """
-      @_specializeℹ️(e1️⃣
+      @_specializeℹ️(1️⃣e2️⃣
       """,
       diagnostics: [
         DiagnosticSpec(
+          locationMarker: "1️⃣",
+          message: "unknown parameter 'e' in attribute argument; valid parameters 'target', 'availability', 'exported', 'kind', 'spi', 'spiModule', and 'available'"
```
kimdv, line 81: Initial I tought that just telling `unknown parameter 'e' in attribute argument`. 
But we know the valid arguments, so why not help?

I was also thinking of adding a Fix-It for each valid parameter type. But it's seems a bit verbose to me. 
Let know what you think @ahoppen

CodeGeneration/Sources/SyntaxSupport/AttributeNodes.swift
```diff
@@ -656,7 +656,15 @@ public let ATTRIBUTE_NODES: [Node] = [
     children: [
       Child(
         name: "Label",
-        kind: .node(kind: .token),
```
ahoppen, line 659: Nice. Thanks for getting rid of the general `kind: .token` here.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
ahoppen: > Should try to look into if we could use `enum SpecializeParameter: TokenSpecSet {}` for this?

Yes, there is some precedence for that already. Look for the `TokenSpecSet` that are marked as `@_spi(RawSyntax)`. I think you could very reasonably move `SpecializeParameter` into TokenSpecSet.swift.



> And second: Should we generate `SpecializeParameter` based on the label.choices in `AttributeNodes.swift`?

That has been a dream of mine for a while now, to generate `TokenSpecSet` based on the node choices. If you would like to look into that, that would be great.

Sources/SwiftParser/generated/Parser+TokenSpecSet.swift
```diff
@@ -241,11 +318,28 @@ extension AvailabilityLabeledArgumentSyntax {
...
+      case .renamed:
+        return .keyword(.renamed)
+      case .introduced:
+        return .keyword(.introduced)
+      case .obsoleted:
+        return .keyword(.obsoleted)
+      case .deprecated:
+        return .keyword(.deprecated)
+      }
+    }
```
kimdv, line 336: @ahoppen I was thinking of adding something like this to the tokenSpec. 

It can then be used in the ParseDiagnosticsGenerator like this `LabeledSpecializeEntrySyntax.LabelOptions.allCases.map { $0.tokenSyntax }`

But the thing is, if it's an `identifier` etc. we don't have any value. Should we just pass an empty string or? 

Let me hear what you think

Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift
```diff
@@ -518,6 +518,23 @@ public struct UnknownDirectiveError: ParserError {
   }
 }
 
+public struct UnknownParameterError: ParserError {
+  public let parameter: TokenSyntax
+  public let validParameters: [TokenSyntax]
+
+  public var message: String {
+    var message = "unknown parameter '\(parameter.description)'"
```
ahoppen, line 526: Should this be `parameter.text` so that we don’t emit any trivia inside the single quotes?

Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift
```diff
@@ -518,6 +518,23 @@ public struct UnknownDirectiveError: ParserError {
...
+
+    if let parentTypeName = parameter.parent?.ancestorOrSelf(mapping: { $0.nodeTypeNameForDiagnostics(allowBlockNames: false) }) {
+      message += " in \(parentTypeName)"
+    }
+
+    message += "; valid parameters \(nodesDescription(validParameters, format: true))"
```
ahoppen, line 532: Let’s make this a proper english sentence.

```suggestion
    message += "; valid parameters are \(nodesDescription(validParameters, format: true))"
```

Tests/SwiftParserTest/AttributeTests.swift
```diff
@@ -73,41 +73,66 @@ final class AttributeTests: XCTestCase {
...
         ),
         DiagnosticSpec(
+          locationMarker: "2️⃣",
           message: "expected declaration after attribute",
           fixIts: ["insert declaration"]
         ),
       ],
       fixedSource: """
-        @_specialize(e:) <#declaration#>
+        @_specialize(e  <#identifier#>:) <#declaration#>
```
ahoppen, line 102: Which Fix-It inserts `<#identifier#>`? The fixed source looks quite incorrect.

kimdv: It is because of this error: https://github.com/apple/swift-syntax/pull/1727/files#r1221798316

We can handle it better by etc add a Fix-It for each label that is valid? 
What do you think of that?

Otherwise we should stop parsing the line at that point?

ahoppen: But which diagnostic is this attached to? No diagnostic mentions: “insert identifier” or anything of the kind. I would expect the “unknown parameter 'e'” diagnostic to remove `e` and substitute it with `<#identifier#>`.

kimdv: It's because we add a `RawTokenSyntax(missing: .identifier, arena: self.arena)`, which is wrong. 
But I could not decide how we can add a missing token, that can be one of x cases, like we can here.

ahoppen: But which diagnostic does insert it? In the test case, I see the following Fix-It messages? We have 
- `insert ':'`
- `insert ')'` -> doesn’t mention the insertion of an identifier
- `insert declaration`
None of these mention the insertion of an identifier, so I’m surprised that one gets added. Based on locations, I would suspect that `insert ':'` adds the identifier but then I’m surprised that it doesn’t say `insert identifier and ':'`.

ahoppen (approved):

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test

Tests/SwiftParserTest/AttributeTests.swift
```diff
@@ -99,12 +107,29 @@ final class AttributeTests: XCTestCase {
...
+          fixIts: ["insert ': false'"]
+        ),
+        DiagnosticSpec(
+          locationMarker: "4️⃣",
+          message: "expected declaration after attribute",
+          fixIts: ["insert declaration"]
+        ),
       ],
       fixedSource: """
         @_specialize(e:, exported: false) <#declaration#>
```
kimdv, line 135: @ahoppen  the CI fails with 

```
Test Case '-[SwiftParserTest.AttributeTests testMissingClosingParenToAttribute]' started.
/Users/ec2-user/jenkins/workspace/swift-syntax-PR-macOS/branch-main/swift-syntax/Sources/SwiftSyntax/generated/raw/RawSyntaxValidation.swift:1689: Fatal error: Error validating child at index 1 of labeledSpecializeEntry:
Expected token with one of [keyword('target'), keyword('availability'), keyword('exported'), keyword('kind'), keyword('spi'), keyword('spiModule'), keyword('available')] but received identifier with text ''
FAIL: Running tests failed
```

I'm a little unsure what to do here.

Sources/SwiftParser/Attributes.swift
```diff
@@ -798,7 +770,8 @@ extension Parser {
         elements.append(
           .labeledSpecializeEntry(
             RawLabeledSpecializeEntrySyntax(
-              label: ident,
+              RawUnexpectedNodesSyntax([ident], arena: self.arena),
```
ahoppen, line 773: The issue is that you’re specifying that in CodeGeneration you are specifying that `label` is one of the the keywords above but you are setting `label` to a missing identifier here. And identifier is not in the list of possible token kinds for the label. You can reproduce the failure by setting the `SWIFTSYNTAX_ENABLE_RAWSYNTAX_VALIDATION` environment variable when building the tests. This is what I mentioned in https://github.com/apple/swift-syntax/pull/1727#discussion_r1221798316 as well.

My suggestion in that comment would be to just break out of the outer loop in the case that we see an argument label that we don’t know, since the recover of eating a single token for the value also really isn’t great.

```swift
LOOP: while !self.at(.endOfFile, .rightParen, .keyword(.where)) && loopProgress.evaluate(currentToken) {
… 
case nil:
  break LOOP
```

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftparser/ParserTokenSpecSetFile.swift
```diff
@@ -70,6 +75,36 @@ let parserTokenSpecSetFile = SourceFileSyntax(leadingTrivia: copyrightHeader) {
                 }
               }
             }
+
+            try VariableDeclSyntax(
+              """
+              @_spi(Diagnostics)
+              public var tokenSyntax: TokenSyntax
```
ahoppen, line 82: Could you add a doc comment

```swift
/// Returns a token that satisfies the `TokenSpec` of this case. 
///
/// If the token kind of this spec has variable text, e.g. for an identifier, this returns a token with empty text.
```

ahoppen (approved):
I’d like one more doc comment after reading this with a few days break, otherwise LGTM.

Tests/SwiftParserTest/AttributeTests.swift
```diff
@@ -73,41 +73,55 @@ final class AttributeTests: XCTestCase {
...
-          message: "expected declaration after attribute",
-          fixIts: ["insert declaration"]
+          locationMarker: "1️⃣",
+          message: "expected declaration and ';' after attribute",
+          fixIts: ["insert declaration and ';'"]
         ),
       ],
       fixedSource: """
-        @_specialize(e:) <#declaration#>
+        @_specialize() <#declaration#>; e
```
kimdv, line 97: I did a bit of digging for this issue.
What I can see is, that the current token is `e`  we don't consume it, as `e`  is not a label in `@_spacialize(`
We then try to find the precedence of the current token. `e` is an identifier, thus there are none.

Thats why we get this result I think
```
testMissingClosingParenToAttribute(): failed - Actual output (+) differed from expected output (-):
–@_specialize(e:) <#declaration#>
+@_specialize() <#declaration#>; e
```

The parser tries to parse the label `e` and don't know it, so we break out if the loop as you suggested.
So I think it's doing its work correctly (for now)
But it would be more nice if we did consume `e` as unknown label, and then apply a fixit that suggest to make one if the ones defined in the token set.

ahoppen: Ah, what I missed here is that there is no closing `)` in the test case. If the input was `@_specialize(e)` (which is what I thought, I just didn’t check carefully), then `expect(.rightParen)` would eat the `e` as unexpected and recover to `)`. So, I agree that the current recovery is consistent and makes sense for now. 

Also, this test case is a little degenerate because it doesn’t have a declaration following the attribute. If you make it a little bit more realistic like

```swift
@_specialize(e
func foo() {}
```

then the fixed source is

```swift
@_specialize()e
func foo() {}
```

Putting the `e` outside the parentheses makes as much sense as putting it inside, so I think it’s fine and there’s nothing we need to do.

Tests/SwiftParserTest/ParserTests.swift
```diff
@@ -23,8 +23,8 @@ public class ParserTests: XCTestCase {
     let fileContents = try Data(contentsOf: fileURL)
     let parsed = fileContents.withUnsafeBytes({ buffer in
       // Release builds are fine with the default maximum nesting level of 256.
-      // Debug builds overflow with any stack size bigger than 25-ish.
-      Parser.parse(source: buffer.bindMemory(to: UInt8.self), maximumNestingLevel: 25)
+      // Debug builds overflow with any stack size bigger than 20-ish.
+      Parser.parse(source: buffer.bindMemory(to: UInt8.self), maximumNestingLevel: 20)
```
ahoppen, line 27: I forgot to mention last time. Could you also adjust `defaultNestingLevel` in Parser.swift:122?

ahoppen (approved):
One more comment that I forgot last time, otherwise LGTM. Let’s merge this after 2 months 😉

ahoppen (approved):

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test windows

PR #1729, opened 2023-06-03
Allow parsing for enum case element parameter on new line

### LINKED ISSUE #1709: Parsing fails if line break between enum case element name and parameter clause
### Issue Kind

Parse of Valid Source Failed

### Source Code

This errors out without good reason (with both the new and legacy parsers):
```swift
enum E {
  case a
    (Int)
}
```

Similarly formatted function/subscript parameter lists parse fine.

### Description

_No response_

kimdv:
Resolves #1709

### DIFF
--- a/Sources/SwiftParser/Declarations.swift
+++ b/Sources/SwiftParser/Declarations.swift
@@ -918,7 +918,7 @@ extension Parser {
         let (unexpectedBeforeName, name) = self.expectIdentifier(keywordRecovery: true)
 
         let associatedValue: RawEnumCaseParameterClauseSyntax?
-        if self.at(TokenSpec(.leftParen, allowAtStartOfLine: false)) {
+        if self.at(TokenSpec(.leftParen)) {
           associatedValue = self.parseParameterClause(RawEnumCaseParameterClauseSyntax.self) { parser in
             parser.parseEnumCaseParameter()
           }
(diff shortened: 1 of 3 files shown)


ahoppen:
CC @bnbarham @rintaro @DougGregor Any concerns with this PR? See https://github.com/apple/swift-syntax/issues/1709 for an explanation of the motivation.

rintaro:
Since func/subscript parameter clause allows newlines before `(`, it's good to be consistent. Call arguments enforces `(` be on the same line, but that is mainly for disambiguation.
TBH I personally don't want to allow newlines before `(` even for `func`, but I couldn't find a strong reason to disallow them.

DougGregor:
> Since func/subscript parameter clause allows newlines before `(`, it's good to be consistent.

This is the argument that wins me over. I agree with @rintaro that I'd prefer not to allow newlines before the `(` for any of func/decl/subscript/etc. However, since we already allow them, I'd rather be consistent and also allow them for cases. There's no chance for ambiguity here because there are no expressions within an enum scope.

DougGregor:
@swift-ci please test

DougGregor (approved):
This can't

ahoppen:
Thanks for your feedback @rintaro and @DougGregor. I agree with you as well, bug just wanted to double-check

ahoppen (approved):

PR #1730, opened 2023-06-04
Change children naming inconsistencies between similar/dual syntax nodes

### LINKED ISSUE #1721: Children naming inconsistencies between similar/dual syntax nodes
### Description

* `GenericParameterClauseSyntax.genericParameterList` vs. `GenericArgumentClauseSyntax.arguments`. Following the plural noun trend, I would expect the former to be named `parameters`.
* `AvailabilityConditionSyntax.availabilitySpec` vs. `AvailabilityEntrySyntax.availabilityList` (both have the same type).
* `OriginallyDefinedInArgumentsSyntax.platforms` vs.  `BackDeployedAttributeSpecListSyntax.versionList` (both have the same type).

TTOzzi:
Resolve #1721 

I have made the naming changes according to the comment I left on the issue.
If you have any feedback for better naming, please let me know and I will make the necessary updates 🙇‍♂️


CodeGeneration/Sources/SyntaxSupport/AttributeNodes.swift
```diff
@@ -168,8 +168,8 @@ public let ATTRIBUTE_NODES: [Node] = [
         documentation: "The colon separating the label and the value"
       ),
       Child(
-        name: "AvailabilityList",
-        kind: .collection(kind: .availabilitySpecList, collectionElementName: "Availability")
+        name: "AvailabilitySpecs",
+        kind: .collection(kind: .availabilitySpecList, collectionElementName: "AvailabilityArgument")
```
ahoppen, line 172: `name` and `collectionElementName` should match here. Also, I’m not a huge fan of the `Spec` part. IMO it would be nice to also rename `AvailabilitySpecListSyntax` -> `AvailabilityArgumentListSyntax`. https://github.com/apple/swift-syntax/issues/1737

---

So, I think we should use name: `AvailabilityArguments`.

--- 

And, could you also rename the `.availabilitySpecList` argument of `.availabilityCondition` to be consistent with this one?

Sources/SwiftSyntax/SwiftSyntaxCompatibility.swift
```diff
@@ -315,3 +315,228 @@ public extension EditorPlaceholderDeclSyntax {
     )
   }
 }
+
+public extension AvailabilityConditionSyntax {
```
ahoppen, line 319: Could you add the types in alphabetical order? And maybe add a comment at the end of the file similar to the following? When I started this file, I didn’t think it would grow as much that quickly but now that it does, I think some structure would be good.


https://github.com/apple/swift-syntax/blob/a90e1be0bb5171e6da973946e712ade67526c207/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift#L1783-L1786


Also, just an idle thought: If we continue to rename properties like this, it might be worth investing in Code-Generation of the file.

TTOzzi: I agree with you. I'll add comments and organize the code alphabetically.

ahoppen (approved):
Thanks

PR #1731, opened 2023-06-04
Parse `init` accessors.

hborla:
This is the SwiftSyntax-side change to https://github.com/apple/swift/pull/66283. Add parsing for [`init` accessors](https://github.com/hborla/swift-evolution/blob/init-accessors/proposals/NNNN-init-accessors.md):

```swift
struct Wrapper<T> {
  var wrappedValue: T
}

struct S {
  private var _value: Wrapper<Int>
  var value: Int {
    init(newValue) initializes(_value) {
      self._value = Wrapper(wrappedValue: newValue)
    }

    get { _value.wrappedValue }
    set { _value.wrappedValue = newValue }
  }
}
```


hborla:
@swift-ci please test

hborla:
@swift-ci please test

hborla:
@swift-ci please test

xedin:

Tests/SwiftParserTest/DeclarationTests.swift
```diff
@@ -869,6 +869,97 @@ final class DeclarationTests: XCTestCase {
...
+
+    assertParse(
+      """
+      struct S {
+        var value: Int {
+          init initializes(x, y, z) accesses(x, y, z) {}
+        }
+      }
+      """
+    )
```
ahoppen, line 960: Could you add a test case for the following? We should be offering a Fix-It to switch the positions of `initializes(x)` and `accesses(y)`

```swift
struct S {
  var value: Int {
    init accesses(y) initializes(x) {}
  }
}
```

--- 

Also I think the following tests would be useful

```swift
struct S {
  var value: Int {
    get accesses(y) {}
  }
}
```

```swift
struct S {
  var value: Int {
    init throws {}
  }
}
```

ahoppen (commented):
I’ve got a couple of post-merge comments for additional test cases / diagnostics.

The parsing pieces look good to me. I’ve got some opinions of how the node types and children should be named but I’ll try to put together a document around naming guidelines for these and will address my thoughts on the names once that’s finished.

PR #1733, opened 2023-06-06
Fix parsing `=` instead of `:`

kimdv:
I opened this PR as I think we can discuss how we maybe can resolve it in a more generic way.

### DIFF
--- a/Sources/SwiftParser/Attributes.swift
+++ b/Sources/SwiftParser/Attributes.swift
@@ -1065,12 +1065,21 @@ extension Parser {
   mutating func parseUnavailableFromAsyncArguments() -> RawUnavailableFromAsyncArgumentsSyntax {
     let (unexpectedBeforeLabel, label) = self.expect(.keyword(.message))
     let (unexpectedBeforeColon, colon) = self.expect(.colon)
+
+    let unexpectedBetweenColonAndMessage: RawUnexpectedNodesSyntax?
+    if let equalToken = self.consume(if: .equal) {
+      unexpectedBetweenColonAndMessage = RawUnexpectedNodesSyntax([equalToken], arena: self.arena)
+    } else {
+      unexpectedBetweenColonAndMessage = nil
+    }
+
     let message = self.parseStringLiteral()
     return RawUnavailableFromAsyncArgumentsSyntax(
       unexpectedBeforeLabel,
       messageLabel: label,
       unexpectedBeforeColon,
       colon: colon,
+      unexpectedBetweenColonAndMessage,
       message: message,
       arena: self.arena
     )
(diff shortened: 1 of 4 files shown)


Sources/SwiftParser/Attributes.swift
```diff
@@ -1065,12 +1065,21 @@ extension Parser {
   mutating func parseUnavailableFromAsyncArguments() -> RawUnavailableFromAsyncArgumentsSyntax {
     let (unexpectedBeforeLabel, label) = self.expect(.keyword(.message))
     let (unexpectedBeforeColon, colon) = self.expect(.colon)
+
+    let unexpectedBetweenColonAndMessage: RawUnexpectedNodesSyntax?
+    if let equalToken = self.consume(if: .equal) {
+      unexpectedBetweenColonAndMessage = RawUnexpectedNodesSyntax([equalToken], arena: self.arena)
+    } else {
+      unexpectedBetweenColonAndMessage = nil
+    }
```
kimdv, line 1074: The expect don't work, because the `.colon` and `.equal` have the same `weakPunctuator`.
This means it breaks here

https://github.com/apple/swift-syntax/blob/c28f9940ba2b89b25f1e754762df0253ca20ccdf/Sources/SwiftParser/Recovery.swift#L101-L#103

ahoppen: OK, I’ve thought about this a bit. `weakPunctuator` currently contains three kinds of tokens

Tokens that can be used within an expression (1)

- `backslash`
- `exclamationMark`
- `infixQuestionMark`
- `period`
- `postfixQuestionMark`
- `prefixAmpersand`

Tokens that are pretty strong structure markers between different expressions (2)

- `atSign`
- `colon`
- `comma`

Tokens that I don’t have a very strong opinion about (3)

- `backtick`
- `ellipsis`
- `equal` (OK, maybe I do have opinions about this one, but it would ruin my argument below)

If we split weakPunctuator into two precedence groups: `weakPunctuator` based on the list (1) and `mediumPunctuator` based on list (2) and decide to put all tokens in (3) to `weakPunctuator`, then expecting a colon would skip over `=`, which is what we want, I think. To maintain most of the current behavior `mediumPunctuator` would need to have a precedence between `weakPunctuator` and `weakBracketClose`. Do you want to give this a try?

ahoppen (approved):

Sources/SwiftParser/TokenPrecedence.swift
```diff
@@ -28,6 +28,7 @@ enum TokenPrecedence: Comparable {
   /// A punctuator that can occur inside a statement
   case weakPunctuator
   /// The closing delimiter of `weakBracketed`
+  case mediumPunctuator
```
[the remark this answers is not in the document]
kimdv: Added correct documentation. Does it make sense @ahoppen

kimdv:
@swift-ci please test

PR #1738, opened 2023-06-07
[Macros] Ability to opt-out from auto-propagation of attrs/modifiers

rintaro:
`DeclarationMacro` now have two optional requirements that is defaulted to 'true':
```
  static var propagateFreestandingMacroAttributes: Bool { get }
  static var propagateFreestandingMacroModifiers: Bool { get }
```
By implementing these and returning 'false', macro expansion system won't copy attributes and/or modifiers to the expanded declarations.

rdar://110364154

I will add a test case with actual plugins in `swift` repo.


rintaro:
@swift-ci Please test

ahoppen (approved):

DougGregor (approved):
Looks great, thank you!

PR #1741, opened 2023-06-07
Generate token spec sets

kimdv:
Context: https://github.com/apple/swift-syntax/pull/1727#discussion_r1221825652


CodeGeneration/Sources/generate-swiftsyntax/templates/swiftparser/ParserTokenSpecSetFile.swift
```diff
@@ -0,0 +1,80 @@
...
+import Utils
+
+let parserTokenSpecSetFile = SourceFileSyntax(leadingTrivia: copyrightHeader) {
+  DeclSyntax("@_spi(RawSyntax) import SwiftSyntax")
+
+  try! ExtensionDeclSyntax("extension Parser") {
+    for node in SYNTAX_NODES {
+      if let children = node.layoutNode?.children {
+        for child in children where !child.isUnexpectedNodes {
+          if case let .token(choices, _, _) = child.kind, choices.count > 1 {
```
kimdv, line 25: Not sure if it made sense to make TokenSets with one case in

Sources/SwiftParser/generated/Parser+TokenSpecSet.swift
```diff
@@ -0,0 +1,2027 @@
...
+      }
+    }
+  }
+  
+  enum AvailabilityLabeledArgumentLabel: TokenSpecSet {
+    case message
+    case renamed
+    case introduced
+    case obsoleted
+    case deprecated
```
kimdv, line 203: I found that this doesn't contain the same as in `AvailabilityArgumentKind`

It have also a special handling

https://github.com/apple/swift-syntax/blob/5474478ccf6bd3319c5c3564d2eed15dc3e45eac/Sources/SwiftParser/Availability.swift#L57-L96

kimdv (commented):
@ahoppen opened this PR to start a discussion if I'm on the right way. Added someinline comments. 

I would also like a first round of some initial toughts of yours

Sources/SwiftParser/generated/Parser+TokenSpecSet.swift
```diff
@@ -0,0 +1,2027 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+@_spi(RawSyntax) import SwiftSyntax
+
+extension Parser {
```
ahoppen, line 17: I don’t think there’s a need to namespace this to parser. The `TokenSpecSet` in TokenSpecSet.swift also aren’t namespaces.

Sources/SwiftParser/generated/Parser+TokenSpecSet.swift
ahoppen: These are exactly the interesting cases. A short-term solution would probably be to define the case from `AvailabilityArgumentKind` that aren’t in `AvailabilityLabeledArgumentLabel` in a separate `TokenSpecSet` and then use `self.at(anyIn:or:)` in Availability.swift:104. In general for these cases, we might need to consider if there is a bug in the parser or if we can model the syntax tree better so that the generated `TokenSpecSet` matches the parser.

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
```diff
@@ -534,7 +534,6 @@ public let DECL_NODES: [Node] = [
           .keyword(text: "public"),
           .keyword(text: "reasync"),
           .keyword(text: "required"),
-          .keyword(text: "setter_access"),
```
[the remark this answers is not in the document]
ahoppen: Yes, I noticed that in some other unmerged PR as well. Thanks for removing it.

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftparser/ParserTokenSpecSetFile.swift
```diff
@@ -0,0 +1,80 @@
...
+            try EnumDeclSyntax("enum \(raw: node.varOrCaseName.description.withFirstCharacterUppercased + child.name): TokenSpecSet") {
```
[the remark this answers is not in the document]
ahoppen: Just an idea: What do you think about generating something like

```swift
extension AccessorDecl {
  public enum AccessorKindOptions: TokenSpecSet {
  }
}
```

---

Instead of `node.varOrCaseName.description.withFirstCharacterUppercased`, you should use `self.kind.syntaxType`.

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftparser/ParserTokenSpecSetFile.swift
```diff
@@ -0,0 +1,80 @@
...
+import SwiftSyntaxBuilder
+import SyntaxSupport
+import Utils
+
+let parserTokenSpecSetFile = SourceFileSyntax(leadingTrivia: copyrightHeader) {
+  DeclSyntax("@_spi(RawSyntax) import SwiftSyntax")
+
+  try! ExtensionDeclSyntax("extension Parser") {
+    for node in SYNTAX_NODES {
+      if let children = node.layoutNode?.children {
```
ahoppen, line 23: Just to be consistent with the rest of CodeGeneration

```suggestion
    for node in SYNTAX_NODES.compatMap(\.layoutNode) {
```

And then I think you don’t need the `if` anymore

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftparser/ParserTokenSpecSetFile.swift
```diff
@@ -0,0 +1,80 @@
...
+        for child in children where !child.isUnexpectedNodes {
```
ahoppen, line 24: Do you need the check for `child.isUnexpectedNodes`. I think unexpected nodes haver have token choices.

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftparser/ParserTokenSpecSetFile.swift
ahoppen: I agree that creating a `TokenSpecSet` with one case doesn’t make sense.

ahoppen (commented):
Thank you. This is great. I wonder how many previously handwritten `TokenSpecSet` we can replace by the generated ones.

I also wonder whether we can use the generated `TokenSpecSet` in RawSyntaxValidation.swift.

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftparser/ParserTokenSpecSetFile.swift
```diff
@@ -0,0 +1,78 @@
...
+import SyntaxSupport
+import Utils
+
+let parserTokenSpecSetFile = SourceFileSyntax(leadingTrivia: copyrightHeader) {
+  DeclSyntax("@_spi(RawSyntax) import SwiftSyntax")
+
+    for layoutNode in SYNTAX_NODES.compactMap(\.layoutNode) {
+        for child in layoutNode.children {
+          if case let .token(choices, _, _) = child.kind, choices.count > 1 {
+              try! ExtensionDeclSyntax("extension \(raw: layoutNode.kind.rawType)") {
```
ahoppen, line 24: Is there any reason why you chose to generate the `TokenSpecSet` in the `Raw` types instead of the non-raw types? I’m thinking that if we define them on the public types, might make the `TokenSpecSet` public in there future as they could also be useful for clients to know which kind of tokens a child may contain.

ahoppen (commented):
Looks good to me. Only one comment inline.

CC @bnbarham because we talked about this at some point as well.

kimdv:
@ahoppen we get this `TokenSpecSet`

```swift
extension SimpleTypeIdentifierSyntax {
  enum NameOptions: TokenSpecSet {
    case identifier
    case `self`
    case `Self`
    case `Any`
    case keyword
    case wildcard
    
    init?(lexeme: Lexer.Lexeme) {
      switch PrepareForKeywordMatch(lexeme) {
      case TokenSpec(.identifier):
        self = .identifier
      case TokenSpec(.`self`):
        self = .`self`
      case TokenSpec(.`Self`):
        self = .`Self`
      case TokenSpec(.`Any`):
        self = .`Any`
      case TokenSpec(.keyword):
        self = .keyword
      case TokenSpec(.wildcard):
        self = .wildcard
      default:
        return nil
      }
    }
    
    var spec: TokenSpec {
      switch self {
      case .identifier:
        return .identifier
      case .`self`:
        return .keyword(.`self`)
      case .`Self`:
        return .keyword(.`Self`)
      case .`Any`:
        return .keyword(.`Any`)
      case .keyword:
        return .keyword
      case .wildcard:
        return .wildcard
      }
    }
  }
}
```

The `. keyword` is not valid code. 
What do you think we should do? 
Omit it?

This is the node 
```swift
Node(
    kind: .simpleTypeIdentifier,
    base: .type,
    nameForDiagnostics: "type",
    children: [
      Child(
        name: "Name",
        kind: .token(choices: [
          .token(tokenKind: "IdentifierToken"),
          .keyword(text: "self"),
          .keyword(text: "Self"),
          .keyword(text: "Any"),
          .token(tokenKind: "KeywordToken"),
          .token(tokenKind: "WildcardToken"),
        ]),
        classification: "TypeIdentifier"
      ),
      Child(
        name: "GenericArgumentClause",
        kind: .node(kind: .genericArgumentClause),
        isOptional: true
      ),
    ]
  )
```

Sources/SwiftParser/Declarations.swift
```diff
@@ -1968,7 +1893,7 @@ extension Parser {
         case (.assignment, let handle)?:
           let assignmentKeyword = self.eat(handle)
           let (unexpectedBeforeColon, colon) = self.expect(.colon)
-          let (unexpectedBeforeFlag, flag) = self.expect(.keyword(.true), .keyword(.false), default: .keyword(.true))
+          let (unexpectedBeforeFlag, flag) = self.expect(anyIn: PrecedenceGroupAssignmentSyntax.FlagOptions.self, default: .true)
```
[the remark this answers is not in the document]
kimdv: I will properly be able to find more of those

kimdv (commented):
I found very few usage of the new generated spec sets that matched 1-1. 
I can try to find more. 

And second, I found also some cases where there was a mismatch between the generated ones and the handwritten ones.
Should I try to apply your suggestion above, so we can see how many cases there are so we need what to do?

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxDeclNodes.swift
```diff
@@ -71,7 +71,7 @@ public struct AccessorDeclSyntax: DeclSyntaxProtocol, SyntaxHashable {
             unexpectedBetweenInitEffectsAndBody, 
             body, 
             unexpectedAfterBody
-          ))) { (arena, _) in
+          ))) {(arena, _) in
```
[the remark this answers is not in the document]
bnbarham: > Hmm... maybe my local CodeGen did not resolve the HEAD correctly. Will try tomorrow again

I run into this as well @kimdv but apparently @ahoppen does not 🤷. I just delete `.build` and `Package.resolved` each time.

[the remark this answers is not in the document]
kimdv: I normally also run `swift package update` and `swift package resolve` each time. 
But forgot it 😅

But thanks for the tip!

Should we file a bug in spm repo? 
It would be nice if it checked that automatically.

kimdv:
@swift-ci please test windows

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxDeclNodes.swift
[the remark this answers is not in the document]
ahoppen: Does `swift package update` and `swift package resolve` always fix the issue for you? Recently I didn’t have the feeling but I think I only ever ran one of the two commands, not both.

I think it’s expected that you need to run `swift package update`. That matches the SwiftPM behavior when you depend on branches where you also only want to update your dependencies (and thus Package.resolved) when you explicitly request it. We should add that to our documentation.

If `swift package resolve` is also necessary, I would say that’s a but in SwiftPM. If you could file an issue for that, that’d be great.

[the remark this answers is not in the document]
bnbarham: Hmmm. Just had a look at the docs, maybe `swift package resolve` is what we should be running? `swift package update` will update `Package.resolved` with the new hash and then `swift package resolve` would update the packages to match that version:
> The swift package resolve command resolves the dependencies, taking into account the current version restrictions in the Package.swift manifest and Package.resolved resolved versions file, and issuing an error if the graph cannot be resolved. For packages which have previously resolved versions recorded in the Package.resolved file, the resolve command will resolve to those versions as long as they are still eligible. If the resolved version's file changes (e.g., because a teammate pushed a new version of the file) the next resolve command will update packages to match that file. After a successful resolve command, the checked out versions of all dependencies and the versions recorded in the resolved versions file will match. In most cases the resolve command will perform no changes unless the Package.swift manifest or Package.resolved file have changed.

But having said that:
> Most SwiftPM commands will implicitly invoke the swift package resolve functionality before running, and will cancel with an error if dependencies cannot be resolved.

So maybe it's a bug if we have to run it, unsure.

CodeGeneration/Sources/SyntaxSupport/AttributeNodes.swift
```diff
@@ -657,7 +657,15 @@ public let ATTRIBUTE_NODES: [Node] = [
...
-        kind: .node(kind: .token),
+        kind: .token(choices: [
+          .keyword(text: "target"),
+          .keyword(text: "availability"),
+          .keyword(text: "exported"),
+          .keyword(text: "kind"),
+          .keyword(text: "spi"),
+          .keyword(text: "spiModule"),
+          .keyword(text: "available"),
+        ]),
```
kimdv, line 668: Will remove this @ahoppen  as the parser fails in some test cases. 
We are now allowed to provide `e` as a label. 

Will be added in #1727

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test windows

PR #1745, opened 2023-06-08
Rename detach() to detached

### LINKED ISSUE #1743: Rename `detach()` to `detached`
According to the Swift naming guidelines, `detach()` should be named `detached` because it’s not mutating. Also, I think it should be a computed property instead of a function.

gibachan:
Resolve #1743 😄

### DIFF
--- a/Sources/SwiftSyntax/Syntax.swift
+++ b/Sources/SwiftSyntax/Syntax.swift
@@ -225,7 +225,7 @@ extension SyntaxProtocol {
 
   /// Return this subtree with this node as the root, ie. detach this node
   /// from its parent.
-  public func detach() -> Self {
+  public var detached: Self {
     return Syntax(raw: self.raw).cast(Self.self)
   }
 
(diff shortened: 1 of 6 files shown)


Sources/SwiftSyntax/Syntax.swift
```diff
@@ -225,7 +225,7 @@ extension SyntaxProtocol {
 
   /// Return this subtree with this node as the root, ie. detach this node
   /// from its parent.
-  public func detach() -> Self {
+  public var detached: Self {
```
kimdv, line 228: I think we should keep the the function, but mark it as deprecated. 
Then next to this we add the new var.

Otherwise it might break for people using this method

gibachan: Thank you for your review! I made the function deprecated.

ahoppen (approved):
Thank you. Looks good to me.

ahoppen:
@swift-ci Please test

Issue #1747, opened 2023-06-08
`PrecedenceGroupRelationSyntax.higherThanOrLowerThan` should be named `higherThanOrLowerThanKeyword`


AnthonyLatsis:
### Description

The current name is inconsistent with `PrecedenceGroupAssignmentSyntax.assignmentKeyword` and `PrecedenceGroupAssociativitySyntax.associativityKeyword`.

ahoppen:
Tracked in Apple’s issue tracker as rdar://110482259

Issue #1723, opened 2023-06-02
Inconsistent name of `ImportDeclSyntax` introducer keyword


AnthonyLatsis:
### Description

`ImportDeclSyntax.importTok` vs. `*Keyword` (e.g. `structKeyword`) in all other decls that start with a fixed non-pound keyword.

ahoppen:
Tracked in Apple’s issue tracker as rdar://110170166

PR #1748, opened 2023-06-08
Add infrastructure for automated validation of syntax node structure

ahoppen:
The goal here is to generalize the API inconsistency reported by @AnthonyLatsis and automatically detect them.

There are valid reasons for nodes or children to violate these base rules. My idea is that we note these exceptions in expected failures, ideally with a sort explanation that motivates the deviation from standard naming rules.

We aren’t running these tests at the moment yet. I’d like to merge them as a starting point for API improvement ideas at the moment.

The current rules are
- All nodes with base kind e.g. `ExprSyntax` should end with `ExprSyntax`.
- If a child has a single keyword as its choice, it should be named `*Keyword` (e.g. `ImportKeyword`) https://github.com/apple/swift-syntax/issues/1723
- If a child has a single non-keyword token kind, name it the same as the the token choice (e.g. `LeftParen`)
- If a token only has keyword token choices, its name should end with `Keyword` https://github.com/apple/swift-syntax/issues/1747
- Children that have the same type also have the same child name https://github.com/apple/swift-syntax/issues/1721, https://github.com/apple/swift-syntax/issues/1720, https://github.com/apple/swift-syntax/issues/1713
- Every node that can conform to a trait does so

### DIFF
--- a/CodeGeneration/Package.swift
+++ b/CodeGeneration/Package.swift
@@ -45,6 +45,12 @@ let package = Package(
         "SyntaxSupport",
       ]
     ),
+    .testTarget(
+      name: "ValidateSyntaxNodes",
+      dependencies: [
+        "SyntaxSupport"
+      ]
+    ),
   ]
 )
 
--- a/CodeGeneration/Sources/SyntaxSupport/Child.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/Child.swift
@@ -12,7 +12,7 @@
 
 /// The kind of token a node can contain. Either a token of a specific kind or a
 /// keyword with the given text.
-public enum TokenChoice {
+public enum TokenChoice: Equatable {
   case keyword(text: String)
   case token(tokenKind: String)
 
--- a/CodeGeneration/Tests/ValidateSyntaxNodes/Utils.swift
+++ b/CodeGeneration/Tests/ValidateSyntaxNodes/Utils.swift
@@ -1,0 +1,32 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+extension Collection {
+  /// If the collection contains a single element, return it, otherwise `nil`.
+  var only: Element? {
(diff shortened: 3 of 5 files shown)


ahoppen:
@swift-ci Please test

CodeGeneration/Tests/ValidateSyntaxNodes/Validation.swift
```diff
@@ -0,0 +1,136 @@
...
+        continue
+      }
+      childrenByNodeKind[childKind, default: []].append((node, child))
+    }
+  }
+
+  for (kind, children) in childrenByNodeKind {
+    if kind.isBase || kind == .token {
+      continue
+    }
```
kimdv, line 118: Nit pick: But can this be added to the `for where` IMO it read a but nicer

CodeGeneration/Tests/ValidateSyntaxNodes/Validation.swift
```diff
@@ -0,0 +1,136 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import SyntaxSupport
+
+/// All nodes with base kind e.g. `ExprSyntax` should end with `ExprSyntax`.
+func validateBaseKindSuffix() -> [ValidationFailure] {
+  var failures: [ValidationFailure] = []
+  for node in SYNTAX_NODES where node.base != .syntaxCollection {
+    if !node.kind.syntaxType.description.hasSuffix(node.base.syntaxType.description) {
```
kimdv, line 19: I would also here add the if to the where? 
What do you think about that?

kimdv (approved):
This is looking good. 
Maybe when it's merged we could add issues for the failures we want to be fixed as they are good first issues for newcomers who want to join the community 🙌

CodeGeneration/Tests/ValidateSyntaxNodes/Validation.swift
ahoppen: Good idea 👍🏽

CodeGeneration/Tests/ValidateSyntaxNodes/Validation.swift
ahoppen: Here I think I’d like to keep it separate because this is less about the nodes that we would like to check but about doing the actual validation.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

bnbarham:
Nice, I'd say the rules generally make sense. And we can always have expected failures for when we *really* want to differ for some reason. Thanks Alex!

bnbarham (approved):

PR #1750, opened 2023-06-09
[CMake] Prevent rebuild of libraries

bnbarham:
`.swiftmodule` is listed as an output of swift libraries, but it isn't updated if there's been no changes (just like the library and its object files). Add it to the touch hack as well.

### DIFF
--- a/cmake/modules/AddSwiftHostLibrary.cmake
+++ b/cmake/modules/AddSwiftHostLibrary.cmake
@@ -37,7 +37,7 @@ function(add_swift_host_library name)
   add_custom_command(
       TARGET ${name}
       POST_BUILD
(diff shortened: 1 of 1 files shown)


bnbarham:
@swift-ci please test

kimdv:
@swift-ci please test windows

ahoppen:
@swift-ci Please test Windows

rintaro (approved):

etcwilde (approved):
Might be worth a comment saying that we can remove it once we require CMake 3.26.
That said, most of this function can be removed at that point so maybe not worth it?

bnbarham:
> Might be worth a comment saying that we can remove it once we require CMake 3.26.
That said, most of this function can be removed at that point so maybe not worth it?

Yeah, basically every "workaround" here and in swift-syntax can be removed when we require 3.26.

PR #1754, opened 2023-06-09
In `assertMacroExpansion` trim newlines from expanded source

ahoppen:
This removes the oddity that if you expand an attached macro on the first line, the expected expanded source needs to start with a newline.

Also, when the actual expanded source does not match the expected one, print the actual expanded source as it’s sometimes easier to read than the diff, especially if there are multiple differences.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #1757, opened 2023-06-09
[Macros] Pass macro role for freestanding macro expansion 

DougGregor:
Macros can only have a single freestanding macro role, so that their
expansion is unambiguous. However, that restriction is at the level of a
macro declaration, and is not reflected in a similar restriction on the
set of macro protocols to which a macro implementation can conform.
Therefore, have the compiler pass a freestanding macro role through to
macro expansion in the same way that we do for attached macros, and
use that to determine which macro protocol to use.

Part of fixing rdar://110418969.

### DIFF
--- a/Sources/SwiftCompilerPluginMessageHandling/Macros.swift
+++ b/Sources/SwiftCompilerPluginMessageHandling/Macros.swift
@@ -26,6 +26,7 @@ extension CompilerPluginMessageHandler {
   /// Expand `@freestainding(XXX)` macros.
   func expandFreestandingMacro(
     macro: PluginMessage.MacroReference,
+    macroRole pluginMacroRole: PluginMessage.MacroRole?,
     discriminator: String,
     expandingSyntax: PluginMessage.Syntax
   ) throws {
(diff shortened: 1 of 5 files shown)


DougGregor:
@swift-ci please test

Sources/SwiftCompilerPluginMessageHandling/Macros.swift
```diff
@@ -43,11 +44,26 @@ extension CompilerPluginMessageHandler {
...
+      let macroRole: MacroRole
+      if let pluginMacroRole {
+        switch pluginMacroRole {
+        case .expression: macroRole = .expression
+        case .declaration: macroRole = .declaration
+        case .codeItem: macroRole = .codeItem
+
+        case .accessor, .conformance, .member, .memberAttribute, .peer:
+          throw MacroExpansionError.invalidMacroRole(pluginMacroRole)
+        }
```
rintaro, line 59: (Not critical at all)
Could you use `MacroRole.init(messageMacroRole: PluginMessage.MacroRole)` here?
Since `expandFreestandingMacro` diagnoses unmatched macro role, diagnosing invalid is not necessary I think.

rintaro: But if we do diagnose invalid role here, we should do it in `expandAttachedMacro()` too

DougGregor: I'd rather be defensive here, because we're on one side of an IPC barrier

DougGregor: Nevermind, I see what you mean. It'll be checked and produce an error down in `expandFreestandingMacro`, so we don't need the redundant check here.

DougGregor: Cleanup is at https://github.com/apple/swift-syntax/pull/1762

PR #1759, opened 2023-06-09
BasicFormat should not format unexpected nodes

bnbarham:
`BasicFormat` depends on the parent layout nodes to determine formatting in some cases, so it doesn't make sense to visit unexpected nodes.

Also fix up `SyntaxRewriter` to take a `viewMode`, since as it was `BasicFormat` was actually visiting `missing` nodes, which also doesn't make much sense.

Resolves rdar://110463876.

### DIFF
--- a/CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxRewriterFile.swift
+++ b/CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxRewriterFile.swift
@@ -30,33 +30,28 @@ let syntaxRewriterFile = SourceFileSyntax(leadingTrivia: copyrightHeader) {
     open class SyntaxRewriter
     """
   ) {
-    DeclSyntax("public init() {}")
+    DeclSyntax("public let viewMode: SyntaxTreeViewMode")
 
-    for node in SYNTAX_NODES where !node.kind.isBase {
-      if (node.base == .syntax || node.base == .syntaxCollection) && node.kind != .missing {
-        DeclSyntax(
-          """
-          /// Visit a ``\(node.kind.syntaxType)``.
-          ///   - Parameter node: the node that is being visited
-          ///   - Returns: the rewritten node
-          open func visit(_ node: \(node.kind.syntaxType)) -> \(node.kind.syntaxType) {
-            return Syntax(visitChildren(node)).cast(\(node.kind.syntaxType).self)
-          }
-          """
-        )
-      } else {
-        DeclSyntax(
-          """
-          /// Visit a ``\(node.kind.syntaxType)``.
-          ///   - Parameter node: the node that is being visited
-          ///   - Returns: the rewritten node
-          open func visit(_ node: \(node.kind.syntaxType)) -> \(raw: node.baseType.syntaxBaseName) {
-            return \(raw: node.baseType.syntaxBaseName)(visitChildren(node))
-          }
-          """
-        )
+    DeclSyntax(
+      """
+      public init(viewMode: SyntaxTreeViewMode = .sourceAccurate) {
+        self.viewMode = viewMode
       }
-    }
+      """
+    )
+
+    DeclSyntax(
+      """
+      /// Rewrite `node` and anchor, making sure that the rewritten node also has
+      /// a parent if `node` had one.
+      public func rewrite(_ node: some SyntaxProtocol) -> Syntax {
+        let rewritten = self.visit(node.data)
+        return withExtendedLifetime(rewritten) {
+          return Syntax(node.data.replacingSelf(rewritten.raw, rawNodeArena: rewritten.raw.arena, allocationArena: SyntaxArena()))
+        }
+      }
+      """
+    )
 
     DeclSyntax(
       """
@@ -105,6 +100,7 @@ let syntaxRewriterFile = SourceFileSyntax(leadingTrivia: copyrightHeader) {
       /// Visit any Syntax node.
       ///   - Parameter node: the node that is being visited
       ///   - Returns: the rewritten node
+      @available(*, deprecated, renamed: "rewrite(_:)")
       public func visit(_ node: Syntax) -> Syntax {
         return visit(node.data)
       }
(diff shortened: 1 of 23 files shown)


bnbarham:
@swift-ci please test

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxRewriterFile.swift
```diff
@@ -258,42 +267,44 @@ let syntaxRewriterFile = SourceFileSyntax(leadingTrivia: copyrightHeader) {
...
-            // Node does not exist. If we are collecting rewritten nodes, we need to
-            // collect this one as well, otherwise we can ignore it.
+
+          guard let child = raw, viewMode.shouldTraverse(node: child) else {
+            // Node does not exist or should not be visited. If we are collecting
+            // rewritten nodes, we need to collect this one as well, otherwise we
+            // can ignore it.
             if newLayout != nil {
-              newLayout!.append(nil)
+              newLayout!.append(raw)
```
ahoppen, line 291: Did this manifest as a bug anywhere and we just didn’t notice it so far?

Sources/SwiftBasicFormat/BasicFormat.swift
```diff
@@ -12,6 +12,13 @@
 
 import SwiftSyntax
 
+/// A rewriter that performs a "basic" format of the passed tree. The base
+/// implementation is primarily aimed at adding whitespace where required such
+/// that re-parsing the tree's description results in the same tree. But it
+/// also makes an attempt at adding in formatting, eg. splitting lines where
+/// obvious and some basic indentation at nesting levels.
+///
+/// Any subclasses *must* return the same node type as given.
```
ahoppen, line 21: I know that I have been very bad at this as well but docc likes to have a short explanation on the first paragraph with a detailed explanation on the following paragraphs. So, in docc this will render more nicely if written as

```suggestion
/// A rewriter that performs a "basic" format of the passed tree. 

/// The base implementation is primarily aimed at adding whitespace where required such
/// that re-parsing the tree's description results in the same tree. But it
/// also makes an attempt at adding in formatting, eg. splitting lines where
/// obvious and some basic indentation at nesting levels.
///
/// Any subclasses *must* return the same node type as given.
```

Sources/_SwiftSyntaxTestSupport/SyntaxProtocol+Initializer.swift
```diff
@@ -23,7 +23,7 @@ private class InitializerExprFormat: BasicFormat {
   private func formatChildrenSeparatedByNewline<SyntaxType: SyntaxProtocol>(children: SyntaxChildren, elementType: SyntaxType.Type) -> [SyntaxType] {
     increaseIndentationLevel()
     var formattedChildren = children.map {
-      self.visit($0).as(SyntaxType.self)!
+      self.rewrite($0.cast(SyntaxType.self)).cast(SyntaxType.self)
```
ahoppen, line 26: Why do we need the first `.cast(SyntaxType.self)`?

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxRewriterFile.swift
bnbarham: It's not a bug previously. This used to only ever be `nil` (ie. we'd only be in this block when `let child = raw` failed)

Sources/SwiftBasicFormat/BasicFormat.swift
bnbarham: Ah, sure thing.

Sources/_SwiftSyntaxTestSupport/SyntaxProtocol+Initializer.swift
bnbarham: This is because of the removal of `visit(Syntax)`. Personally I still think it's worth it so that we don't accidentally use `visit(Syntax)` instead of `rewrite(some SyntaxProtocol)` (the latter keeps the parent, the former doesn't necessarily). Thoughts?

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxRewriterFile.swift
ahoppen: Ah, I see

bnbarham:
https://github.com/apple/swift/pull/66569

@swift-ci please test

bnbarham:
@swift-ci please test

Sources/SwiftOperators/OperatorTable+Folding.swift
```diff
@@ -475,6 +475,7 @@ extension OperatorTable {
     ) {
       self.opPrecedence = opPrecedence
       self.errorHandler = errorHandler
+      super.init(viewMode: .sourceAccurate)
```
ahoppen, line 478: Do we really only want to fold source accurate nodes here? I would think of sequence folding as a semantic operation and thus think that it might be better off folding the fixed up tree.

bnbarham: Ideally we wouldn't be folding cat all if there's any errors 😅. Let's say there's a bunch of unexpected nodes, do we expect folding to remove them all?

bnbarham:
swift-format also uses `visit(Syntax)`. I'm going to move dropping that function out of this PR. It's unrelated to not formatting unexpected nodes.

bnbarham:
@swift-ci please test

Sources/SwiftOperators/OperatorTable+Folding.swift
bnbarham: Ah, except we won't remove them (thanks @rintaro). So I suppose I don't really have a preference here. I can do fixed up 🤷

bnbarham:
@swift-ci please test

bnbarham:

ahoppen (approved):

bnbarham:
@swift-ci please test

bnbarham:
@swift-ci please test

bnbarham:

PR #1762, opened 2023-06-10
Eliminate redundant precondition checking

DougGregor:
Thanks to Rintaro for pointing this out.


DougGregor:
@swift-ci please test

PR #1763, opened 2023-06-10
[5.9] [Macros] Pass macro role for freestanding macro expansion

DougGregor:
* **Explanation**: Extend the protocol used by the macro server to include a freestanding macro role in the freestanding macro expansion request, and provide improved diagnostics when the role requested is not implemented by the macro implementation type. This is, on the whole, more informative than crashing the compiler, which is what happened before.
* **Scope**: Narrow; only affects validation of freestanding macro expansions.
* **Risk**: Low; narrow change to perform better validation when expanding freestanding macros.
* **Reviewed by**: @rintaro 
* **Issue**: rdar://110418969
* **Original pull request**: https://github.com/apple/swift-syntax/pull/1757, https://github.com/apple/swift-syntax/pull/1762


DougGregor:
@swift-ci please test

hborla (approved):

PR #1766, opened 2023-06-11
Add FixIts for `canImport` diagnose

### LINKED ISSUE #1765: Add Fix-Its for `canImport` diagnose
### Description

https://github.com/apple/swift-syntax/pull/1448#discussion_r1196915928

StevenWong12:
Resolves #1765.

### DIFF
--- a/Sources/SwiftParserDiagnostics/DiagnosticExtensions.swift
+++ b/Sources/SwiftParserDiagnostics/DiagnosticExtensions.swift
@@ -174,14 +174,15 @@ extension FixIt.MultiNodeChange {
     return FixIt.MultiNodeChange(primitiveChanges: changes)
   }
 
-  /// Make `oldNode` replaced by `newNode` but maintain `oldNode`'s trivia, where `oldNode` should be a missing node
-  static func makeReplacingPresentWithTrivia(oldNode: some SyntaxProtocol, newNode: some SyntaxProtocol) -> Self {
-    var formattedOldNode = MissingNodesBasicFormatter(viewMode: .fixedUp).visit(Syntax(oldNode))
+  /// Make `node` present while changing its content to `tokenKind`.
+  static func makePresent(_ node: TokenSyntax, with tokenKind: TokenKind) -> Self {
+    var formattedOldNode = MissingNodesBasicFormatter(viewMode: .fixedUp).visit(Syntax(node))
     formattedOldNode = PresentMaker().rewrite(formattedOldNode)
 
-    var presentNode = Syntax(newNode)
-    presentNode = presentNode.with(\.leadingTrivia, formattedOldNode.leadingTrivia)
-    presentNode = presentNode.with(\.trailingTrivia, formattedOldNode.trailingTrivia)
-    return FixIt.MultiNodeChange(primitiveChanges: [.replace(oldNode: Syntax(oldNode), newNode: presentNode)])
+    guard let formattedTokenSyntax = formattedOldNode.as(TokenSyntax.self)?.with(\.tokenKind, tokenKind) else {
+      return FixIt.MultiNodeChange(primitiveChanges: [.replace(oldNode: Syntax(node), newNode: formattedOldNode)])
+    }
+
+    return FixIt.MultiNodeChange(primitiveChanges: [.replace(oldNode: Syntax(node), newNode: Syntax(formattedTokenSyntax))])
   }
 }
--- a/Sources/SwiftParserDiagnostics/DiagnosticExtensions.swift
+++ b/Sources/SwiftParserDiagnostics/DiagnosticExtensions.swift
@@ -173,4 +173,15 @@ extension FixIt.MultiNodeChange {
 
     return FixIt.MultiNodeChange(primitiveChanges: changes)
   }
+
+  /// Make `oldNode` replaced by `newNode` but maintain `oldNode`'s trivia, where `oldNode` should be a missing node
+  static func makeReplacingPresentWithTrivia(oldNode: some SyntaxProtocol, newNode: some SyntaxProtocol) -> Self {
+    var formattedOldNode = MissingNodesBasicFormatter(viewMode: .fixedUp).visit(Syntax(oldNode))
+    formattedOldNode = PresentMaker().rewrite(formattedOldNode)
+
+    var presentNode = Syntax(newNode)
+    presentNode = presentNode.with(\.leadingTrivia, formattedOldNode.leadingTrivia)
+    presentNode = presentNode.with(\.trailingTrivia, formattedOldNode.trailingTrivia)
+    return FixIt.MultiNodeChange(primitiveChanges: [.replace(oldNode: Syntax(oldNode), newNode: presentNode)])
+  }
 }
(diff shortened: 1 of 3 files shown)


Tests/SwiftParserTest/translated/IfconfigExprTests.swift
```diff
@@ -330,8 +339,17 @@ final class IfconfigExprTests: XCTestCase {
       #endif
       """,
       diagnostics: [
-        DiagnosticSpec(message: "2nd parameter of canImport should be labeled as _version or _underlyingVersion")
-      ]
+        DiagnosticSpec(
+          message: "2nd parameter of canImport should be labeled as _version or _underlyingVersion",
+          fixIts: ["insert '_version'", "insert '_underlyingVersion'", "insert ':'"]
```
ahoppen, line 344: The way this would presents in an IDE is to offer three different options to fix it:
- `insert '_version'`
- `insert '_underlyingVersion'`
- `insert ':'`

But we only want to offer tow fixes, `insert '_version'` and `insert '_underlyingVersion'` that also insert the `:`. To fix this, instead of creating a separate Fix-It for the colon, you need to add the `.makePresent(node.colon)` change to the other two existing Fix-Its.

Tests/SwiftParserTest/translated/IfconfigExprTests.swift
```diff
@@ -317,8 +317,17 @@ final class IfconfigExprTests: XCTestCase {
       #endif
       """,
       diagnostics: [
-        DiagnosticSpec(message: "2nd parameter of canImport should be labeled as _version or _underlyingVersion")
-      ]
+        DiagnosticSpec(
+          message: "2nd parameter of canImport should be labeled as _version or _underlyingVersion",
+          fixIts: ["insert '_version'", "insert '_underlyingVersion'"]
```
ahoppen, line 322: Sorry for not noticing this in the first review. But the Fix-It actually replaces `unknown` by `_version`, so we shouldn’t be adding a Fix-It that says `insert`. `exchangeTokens` should be able to handle the generation of the diagnostic for you.

StevenWong12: I just added a different Fix-It message(`replace 'unknown' with '_version' and ':'`) for this case.

StevenWong12:
It seems there's some problems with GitHub (maybe?) that my former commit didn't sync in the pr, so I re-pushed it again, which may lead the `Compare` button not showing anything.

Here are the major changes for your convenience.

> https://github.com/StevenWong12/swift-syntax/blob/d93c5fb9f989a0c5503bb25d86fe0fbadb0a8e00/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift#L648-L659

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -645,9 +645,48 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
+      let versionMessage: ParserFixIt =
+        node.unexpectedBetweenLabelAndColon == nil
+        ? InsertFixIt(tokenToBeInserted: .identifier("_version"))
+        : ReplaceTokensFixIt(replaceTokens: replaceTokens, replacements: [.identifier("_version"), .colonToken()])
+
+      let underlyingVersionMessage: ParserFixIt =
+        node.unexpectedBetweenLabelAndColon == nil
+        ? InsertFixIt(tokenToBeInserted: .identifier("_underlyingVersion"))
+        : ReplaceTokensFixIt(replaceTokens: replaceTokens, replacements: [.identifier("_underlyingVersion"), .colonToken()])
+
```
StevenWong12, line 659: It seems there's some problems with GitHub (maybe?) that my former commit didn't sync in the pr, so I re-pushed it again, which may lead the Compare button not showing anything.

Here are the major changes for your convenience.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -645,13 +645,52 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
+      let fixIts = [
+        FixIt(
+          message: versionMessage,
+          changes: [
+            FixIt.MultiNodeChange(
+              .replace(
+                oldNode: Syntax(node.label),
+                newNode: Syntax(TokenSyntax(.identifier("_version"), presence: .present))
```
ahoppen, line 667: One more request: Can you run `MissingNodesBasicFormatter` on this node? That should fix the test case below where the fixed source is `A,_version` and doesn’t contain a space between `,` and `_version`. Same below for `_underlyingVersion`.

And while you’re at it, you could also add the following initializer to `MissingNodesBasicFormatter` because it never makes sense to run it on a source accurate tree.
```swift
init() {
  super.init(viewMode: .fixedUp)
}
```

ahoppen (commented):
I’ve got one more small comment, otherwise LGTM.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
StevenWong12: What do you think about adding a `makeReplaceWithTrivia` function (as my new commit) to do this?

Sources/SwiftParserDiagnostics/DiagnosticExtensions.swift
```diff
@@ -173,4 +173,15 @@ extension FixIt.MultiNodeChange {
 
     return FixIt.MultiNodeChange(primitiveChanges: changes)
   }
+
+  /// Make `oldNode` replaced by `newNode` but maintain `oldNode`'s trivia, where `oldNode` should be a missing node
+  static func makeReplacingPresentWithTrivia(oldNode: some SyntaxProtocol, newNode: some SyntaxProtocol) -> Self {
```
ahoppen, line 178: I think the comment is a little hard to understand and I couldn’t come up with a better explanation of what this function does, which I often take for a sign that it might not be the right API design. What do you think about the following?

```swift
/// Make `node` present while changing its content to `tokenKind`.
static func makePresent(_ node: some SyntaxProtocol, with tokenKind: TokenKind)
```

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -645,13 +645,51 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
     }
 
     if node.label.isMissing {
+      let replaceTokens: [TokenSyntax] = [node.unexpectedBetweenLabelAndColon].compactMap { $0?.tokens(viewMode: .sourceAccurate) }.flatMap { $0 }
```
ahoppen, line 648: Can’t this just be 

```suggestion
      let replaceTokens = node.unexpectedBetweenLabelAndColon.compactMap { $0?.tokens(viewMode: .sourceAccurate) }
```

StevenWong12: But we still need to cast `[TokenSequence]?` to `[TokenSyntax]`. I think mine would be simpler.

Sources/SwiftParserDiagnostics/DiagnosticExtensions.swift
StevenWong12: I think that's nice. I made `node` a `TokenSyntax` since I think this function should only be used by presenting a `TokenSyntax` and changing its token kind.

PR #1770, opened 2023-06-12
Suggest to add missing `equal` token to variable declarations

### LINKED ISSUE #1610: Bad diagnostic for let foo: [Int] []
### Issue Kind

Bad Diagnostic Produced

### Source Code

```swift
let foo: [Int] []
```

### Description

This currently produces the following. We should be suggesting to add a `=` if a variable declaration is followed by an expression on the same line without a `=` in between.

```
1 │ let foo: [Int] []
  │               ╰─ error: consecutive statements on a line must be separated by ';'
```

rdar://108623515

whiteio:
When a variable declaration is followed by an expression on the same line, it should be suggested to insert an `equal` token.

Resolves #1610

### DIFF
--- a/Sources/SwiftParser/Declarations.swift
+++ b/Sources/SwiftParser/Declarations.swift
@@ -1405,6 +1405,14 @@ extension Parser {
             value: initExpr,
             arena: self.arena
           )
+        } else if self.atStartOfExpression(), !self.at(.leftBrace), !self.currentToken.flags.contains(.isAtStartOfLine) {
+          let missingEqual = RawTokenSyntax(missing: .equal, arena: self.arena)
+          let expr = self.parseExpression()
+          initializer = RawInitializerClauseSyntax(
+            equal: missingEqual,
+            value: expr,
+            arena: self.arena
+          )
         } else {
           initializer = nil
         }
--- a/Tests/SwiftParserTest/translated/ConsecutiveStatementsTests.swift
+++ b/Tests/SwiftParserTest/translated/ConsecutiveStatementsTests.swift
@@ -70,7 +70,7 @@ final class ConsecutiveStatementsTests: XCTestCase {
         if i != j { i = j }
         // Errors
         i = j1️⃣ j = i
-        let r : Int2️⃣ i = j
+        let r : Int 2️⃣i = j
         let s : Int3️⃣ let t : Int
         _ = r; _ = s; _ = t
       }
@@ -83,8 +83,8 @@ final class ConsecutiveStatementsTests: XCTestCase {
         ),
         DiagnosticSpec(
           locationMarker: "2️⃣",
-          message: "consecutive statements on a line must be separated by newline or ';'",
-          fixIts: ["insert newline", "insert ';'"]
+          message: "expected '=' in variable",
+          fixIts: ["insert '='"]
         ),
         DiagnosticSpec(
           locationMarker: "3️⃣",
@@ -92,7 +92,7 @@ final class ConsecutiveStatementsTests: XCTestCase {
           fixIts: ["insert newline", "insert ';'"]
         ),
       ],
-      applyFixIts: ["insert newline"],
+      applyFixIts: ["insert newline", "insert '='"],
       fixedSource: """
         // Within a function
         func test(i: inout Int, j: inout Int) {
@@ -102,8 +102,7 @@ final class ConsecutiveStatementsTests: XCTestCase {
           // Errors
           i = j
           j = i
-          let r : Int
-          i = j
+          let r : Int = i = j
           let s : Int
           let t : Int
           _ = r; _ = s; _ = t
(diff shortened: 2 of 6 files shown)


Sources/SwiftParser/Declarations.swift
```diff
@@ -1405,6 +1407,14 @@ extension Parser {
             value: initExpr,
             arena: self.arena
           )
+        } else if self.atStartOfExpression() && !isProbablyVarDecl && !self.at(.leftBrace) && !self.currentToken.flags.contains(.isAtStartOfLine) {
```
ahoppen, line 1410: Can’t you just inline `isProbablyVarDecl` here? If you can, you could merge the `!self.at(.leftBrace)` check into it?

Also, instead of checking `self.currentToken.flags.contains(.isAtStartOfLine)` it’s nicer to check 
```swift
self.at(TokenSpec(.identifier, allowAtStartOfLine: false), TokenSpec(.wildcard, allowAtStartOfLine: false))`
```

Also: What do we need the `isProbablyVarDecl` check for?

ahoppen (commented):
Thanks @whiteio. I’ve got one question regarding the implementation. Also: Could you add a test case from the example in the issue?

Sources/SwiftParser/Declarations.swift
whiteio: @ahoppen Yes I've added in the test case from the example in the issue. `isProbablyVarDecl` can be inlined, I've inlined it and merged in `!self.at(.leftBrace)`.

> self.at(TokenSpec(.identifier, allowAtStartOfLine: false), TokenSpec(.wildcard, allowAtStartOfLine: false))`

Would this also need to include keywords? For example, if I change it to the above it would suggest inserting an '=' if the variable declaration is followed by an initializer on a new line, such as:
```swift
struct S {
  let _value: Int
                 ╰─ error: expected '=' in variable
  init() {
  }
}
```


> Also: What do we need the isProbablyVarDecl check for?

I added this in to make sure that the correct fixits are suggested when there are consecutive variable declarations on the same line.

For example, without the isProbablyVarDecl check it would suggest adding an '=' instead of a ';' before 'i' in the following case
```swift
let x: Int i = 10
```

ahoppen: > `self.at(TokenSpec(.identifier, allowAtStartOfLine: false), TokenSpec(.wildcard, allowAtStartOfLine: false))`

I am sorry, I missed that you had a negation in front of `self.at`. Forget what I said.

> I added this in to make sure that the correct fixits are suggested when there are consecutive variable declarations on the same line.

Hmm. But this also means that you don’t generate the diagnostic to add the missing `=` if you write the following, right?

```swift
let x: Int i
```

I’m not sure if it’s really beneficial to optimize for the case where someone incorrectly writes two variable declarations on the same line. Declaring two variables on the same line isn’t very common in Swift in my experience.

whiteio: > So: My take on this would be to just remove
>!self.at(.identifier, .wildcard, .leftBrace), !self.peek().rawTokenKind.is(.colon, .equal, .comma)


Yes, you're right. I've made your suggested change to remove this, I've added `!self.at(.leftBrace)` as it's needed for scenarios such as the one you've mentioned. 👍

Tests/SwiftParserTest/translated/ConsecutiveStatementsTests.swift
```diff
@@ -102,8 +102,7 @@ final class ConsecutiveStatementsTests: XCTestCase {
           // Errors
           i = j
           j = i
-          let r : Int
-          i = j
+          let r : Int i = j
```
ahoppen, line 105: Am I missing something or do you know why we aren’t interesting a `=` after `Int` as the Fix-It says?

whiteio: Looks like `"insert '='"` wasn't included in the `applyFixits` parameter. I've added it in now.

ahoppen: Oh, that makes sense.

Tests/SwiftParserTest/translated/RecoveryTests.swift
```diff
@@ -78,11 +78,10 @@ final class RecoveryTests: XCTestCase {
           message: "unexpected code in function"
         ),
       ],
-      applyFixIts: ["insert '>'", "insert newline", "insert expression"],
+      applyFixIts: ["insert '>'", "insert expression"],
```
ahoppen, line 81: The idea is that the `a` test inserts newline and the `b` test below adds the `;`. So, I think you should be able to remove `testRecovery7b`.

ahoppen (commented):
Just one more comment, then we’re ready to 🚢

Tests/SwiftParserTest/translated/RecoveryTests.swift
whiteio: Ok, that's it removed now.

ahoppen (approved):
Thank you @whiteio.

ahoppen:
@swift-ci Please test

PR #1771, opened 2023-06-12
Generate grammar doc comments for layout nodes

ahoppen:
Generate a `Grammar` and a `Children` section as doc-comments for layout nodes.

The `Grammar` section is particularly useful for fairly simple syntax nodes to convey their structure. The `Children` section allows you to see the children of the syntax node at a glance in the source order.

### DIFF
--- a/CodeGeneration/Sources/SyntaxSupport/GrammarGenerator.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/GrammarGenerator.swift
@@ -1,0 +1,63 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
+/// Generates grammar doc comments for syntax nodes.
+struct GrammarGenerator {
+  private func grammar(for tokenChoice: TokenChoice) -> String {
+    switch tokenChoice {
+    case .keyword(text: let text):
+      return "`'\(text)'`"
+    case .token(tokenKind: let tokenKind):
+      let token = SYNTAX_TOKEN_MAP[tokenKind]!
+      if let tokenText = token.text {
+        return "`'\(tokenText)'`"
+      } else {
+        return "`<\(token.swiftKind)>`"
+      }
+    }
+  }
+
+  private func grammar(for child: Child) -> String {
+    let optionality = child.isOptional ? "?" : ""
+    switch child.kind {
+    case .node(let kind):
+      return "``\(kind.syntaxType)``\(optionality)"
+    case .nodeChoices(let choices):
+      let choicesDescriptions = choices.map { grammar(for: $0) }
+      return "(\(choicesDescriptions.joined(separator: " | ")))\(optionality)"
+    case .collection(let kind, _, _):
+      return "``\(kind.syntaxType)``"
+    case .token(let choices, _, _):
+      if choices.count == 1 {
+        return "\(grammar(for: choices.first!))\(optionality)"
+      } else {
+        let choicesDescriptions = choices.map { grammar(for: $0) }
+        return "(\(choicesDescriptions.joined(separator: " | ")))\(optionality)"
+      }
+    }
+  }
+
+  /// Generates a markdown list containing the children’s names and their
+  /// grammar.
+  ///
+  /// - Parameter children: The children to show in the list
+  static func childrenList(for children: [Child]) -> String {
+    let generator = GrammarGenerator()
+    return
+      children
+      .filter { !$0.isUnexpectedNodes }
+      .map { " - `\($0.varName)`: \(generator.grammar(for: $0))" }
+      .joined(separator: "\n")
+  }
+}
--- a/CodeGeneration/Sources/SyntaxSupport/Node.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/Node.swift
@@ -220,6 +220,20 @@ public struct LayoutNode {
       preconditionFailure("NodeLayoutView must wrap a Node with data `.layout`")
     }
   }
+
+  public var grammar: SwiftSyntax.Trivia {
+    guard !children.isEmpty else {
+      return []
+    }
+
+    return docCommentTrivia(
+      from: """
+        ### Children
+
+        \(GrammarGenerator.childrenList(for: children))
+        """
+    )
+  }
 }
 
 /// Provides a view into a collection node that offers access to the
--- a/Sources/SwiftSyntax/generated/syntaxNodes/SyntaxDeclNodes.swift
+++ b/Sources/SwiftSyntax/generated/syntaxNodes/SyntaxDeclNodes.swift
@@ -15,6 +15,16 @@
 // MARK: - AccessorDeclSyntax
 
 
+
+/// ### Children
+/// 
+///  - `attributes`: ``AttributeListSyntax``
+///  - `modifier`: ``DeclModifierSyntax``?
+///  - `accessorKind`: (`'get'` | `'set'` | `'didSet'` | `'willSet'` | `'unsafeAddress'` | `'addressWithOwner'` | `'addressWithNativeOwner'` | `'unsafeMutableAddress'` | `'mutableAddressWithOwner'` | `'mutableAddressWithNativeOwner'` | `'_read'` | `'_modify'` | `'init'`)
+///  - `parameter`: ``AccessorParameterSyntax``?
+///  - `effectSpecifiers`: ``AccessorEffectSpecifiersSyntax``?
+///  - `initEffects`: ``AccessorInitEffectsSyntax``?
+///  - `body`: ``CodeBlockSyntax``?
 public struct AccessorDeclSyntax: DeclSyntaxProtocol, SyntaxHashable {
   public let _syntaxNode: Syntax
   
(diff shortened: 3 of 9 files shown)


ahoppen:
@swift-ci Please test

bnbarham:
I'm not convinced it's worth having both grammar + children here. In general children seems more useful to me. Do you have any examples where you'd consider the "grammar" as better than children?

CodeGeneration/Sources/SyntaxSupport/Node.swift
```diff
@@ -208,6 +208,24 @@ public struct LayoutNode {
...
+
+    return docCommentTrivia(
+      from: """
+        ### Grammar
+
+        \(GrammarGenerator.grammar(for: children))
+
+        ### Children
+
+        \(GrammarGenerator.childrenList(for: children))
```
WowbaggersLiquidLunch, line 225: The list of children should already be visible through DocC, under the type/protocol's _Instant Property_ section. Does it really need to be repeated in the main description too?

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxDeclNodes.swift
```diff
@@ -15,6 +15,26 @@
...
+
+/// ### Grammar
+/// 
+/// (``AttributeSyntax`` | ``IfConfigDeclSyntax``) `*`
+/// ``DeclModifierSyntax``?
+/// (`'get'` | `'set'` | `'didSet'` | `'willSet'` | `'unsafeAddress'` | `'addressWithOwner'` | `'addressWithNativeOwner'` | `'unsafeMutableAddress'` | `'mutableAddressWithOwner'` | `'mutableAddressWithNativeOwner'` | `'_read'` | `'_modify'` | `'init'`)
+/// ``AccessorParameterSyntax``?
+/// ``AccessorEffectSpecifiersSyntax``?
+/// ``AccessorInitEffectsSyntax``?
+/// ``CodeBlockSyntax``?
```
WowbaggersLiquidLunch, line 27: Is there a way to generate the translation between the type/protocol names and the actual grammar terms/components that they represent? I think having such a translation table can help a lot to contributors.

CodeGeneration/Sources/SyntaxSupport/GrammarGenerator.swift
```diff
@@ -0,0 +1,82 @@
...
+import SwiftSyntax
+
+/// Generates grammar doc comments for syntax nodes.
+struct GrammarGenerator {
+  let lookThroughCollections: Bool
+
+  private func grammar(for tokenChoice: TokenChoice) -> String {
+    switch tokenChoice {
+    case .keyword(text: let text):
+      return "`'\(text)'`"
```
WowbaggersLiquidLunch, line 22: I feel having quotation marks within the code voice makes it a bit ambiguous what the token actually is.

CodeGeneration/Sources/SyntaxSupport/GrammarGenerator.swift
```diff
@@ -0,0 +1,82 @@
...
+  static func grammar(for children: [Child]) -> String {
+    let generator = GrammarGenerator(lookThroughCollections: true)
+    return
+      children
+      .filter { !$0.isUnexpectedNodes }
+      .map { generator.grammar(for: $0) }
+      .joined(separator: "\n")
+  }
+
+  /// Generates a
```
WowbaggersLiquidLunch, line 73: Incomplete doc comment. Should be "/// Generates a list of child nodes"?

WowbaggersLiquidLunch (commented):
Thanks for this. I think this is a good approach in general. I have just some nitpicky-ish comments.

CodeGeneration/Sources/SyntaxSupport/Node.swift
ahoppen: I think it does because these are sorted in source order while the *Instance Properties* are sorted alphabetically.

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxDeclNodes.swift
ahoppen: What do you mean by “actual grammar terms/components that they represent”?

CodeGeneration/Sources/SyntaxSupport/GrammarGenerator.swift
ahoppen: My idea here was to optimize this for rendering in docc and I found that rendering the token name in code voice reads a little nicer in docc. Would you prefer to just use 

```swift
return "'\(text)'"
```

CodeGeneration/Sources/SyntaxSupport/GrammarGenerator.swift
ahoppen: Good catch. Thank you

ahoppen:
> I'm not convinced it's worth having both grammar + children though. In general children seems more useful to me. Do you have any examples where you'd consider the "grammar" as better than children? Right now they're basically duplicates, with children having the advantage of having the name as well.

My idea was that in particular for simple nodes, the grammar shows the structure of the node in a very nice and concise form. But looking at other nodes as well, I agree that it very much breaks down for more complex nodes. @WowbaggersLiquidLunch Do you have opinions on just showing the `Children` section?

Screenshot 2023-06-19 at 17 35 02

ahoppen:
OK, I removed the *Grammar* section and we’re just left with *Children* now. That seems to be cleaner.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

bnbarham (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Win

ahoppen:
@swift-ci Please test Windows

CodeGeneration/Sources/SyntaxSupport/Node.swift
WowbaggersLiquidLunch: OK I think this is a valid reason. Although, maybe in the long term we should have the ability to specify to DocC the ordering of properties.

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxDeclNodes.swift
WowbaggersLiquidLunch: I might be misunderstanding it. What I mean is: it seems to me that not all syntax nodes in the parser correlate one-to-one with the syntax nodes as documented in TSPL.

WowbaggersLiquidLunch:
> My idea was that in particular for simple nodes, the grammar shows the structure of the node in a very nice and concise form. But looking at other nodes as well, I agree that it very much breaks down for more complex nodes. @WowbaggersLiquidLunch Do you have opinions on just showing the `Children` section?
> 
> Screenshot 2023-06-19 at 17 35 02

> OK, I removed the _Grammar_ section and we’re just left with _Children_ now. That seems to be cleaner.

Sorry for the late response. I still think it's better to have the _Grammar_ section. It's probably a bit more clutter, but from my experience, having the grammar documented really helps new contributors find the relevant things to modify. Although, instead of "'await' `ExprSyntax`", I would prefer "await-operator" in the documentation, because that's probably what people will search for to find this syntax node.

Additionally something I just noticed: Currently the grammars are documented on the `parse` functions, not syntax node types. Are there plans to generated the grammar docs for those functions?

CodeGeneration/Sources/SyntaxSupport/Node.swift
ahoppen: Yes, if we could do that,  I agree that we might be able to remove the children list.

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxDeclNodes.swift
ahoppen: No, there isn’t always a 1:1 correlation between the SwiftSyntax nodes and the grammar rules in TSPL.

CodeGeneration/Sources/SyntaxSupport/GrammarGenerator.swift
ahoppen: The thing is that we use docc links for links to other syntactic categories, which are always monospaced. 

The documentation for [StructDeclSyntax](https://swiftpackageindex.com/apple/swift-syntax/main/documentation/swiftsyntax/structdeclsyntax#Children) shows all kinds of references. Which fonts would you use here?

ahoppen:
> Sorry for the late response. I still think it's better to have the _Grammar_ section. It's probably a bit more clutter, but from my experience, having the grammar documented really helps new contributors find the relevant things to modify. 

Which information do you think would be valuable in the *Grammar* section that’s not present in the *Children* section?

> Although, instead of "'await' `ExprSyntax`", I would prefer "await-operator" in the documentation, because that's probably what people will search for to find this syntax node.

To be clear: There is a difference between SwiftSyntax and TSPL, just by the way the trees are structured. 

For example, if we wanted to model the following production rules in SwiftSyntax, `prefix-operator` would need to be an enum with two cases, each with associated values and that’s just not how the SwiftSyntax tree is structured.

```
prefix-expression → prefix-operator? postfix-expression
prefix-expression → in-out-expression
```

So, there is not equivalent of the

```
expression → try-operator? await-operator? prefix-expression infix-expressions?
```

production rule or *await-operator* in SwiftSyntax, so I don’t think there’s a natural way to place it.

> Additionally something I just noticed: Currently the grammars are documented on the `parse` functions, not syntax node types. Are there plans to generated the grammar docs for those functions?

Since the grammar rules on the `parse` functions are mostly outdated, I want to remove them now that this PR is merged.

PR #1773, opened 2023-06-13
[Macros] Implement expansion of conformance macros

### LINKED ISSUE #1742: SyntaxProtocol.expand() does not expand conformance macros
### Description

Conformance macros used in the input source are discarded without expansion. A basic example:

```swift
struct Conformance: ConformanceMacro {
    public static func expansion(
        of node: AttributeSyntax,
        providingConformancesOf declaration: some DeclGroupSyntax,
    ) throws -> [(TypeSyntax, GenericWhereClauseSyntax?)] {
        fatalError()
    }
}
assertMacroExpansion("@Foo class Bar {}", expandedSource: "class Bar {\n}",
                     macros: ["Foo": Conformance.self])
```

This test passes despite the implementation being `fatalError()` due to `expansion(of:performanceConformanceOf:in:)` never being called. There appears to be two parallel implementations of macro expansion with ConformanceMacro only being supported in expandAttachedMacro(), which is not called by SyntaxProtocol.expand().

There may be other macro types not implemented as well; I did not try them all.

### Steps to Reproduce

_No response_

DougGregor:
Somehow, we forgot to implement this here, even though the compiler has supported it for a long time.

Fixes #1742 / rdar://110418242.

### DIFF
--- a/Sources/SwiftSyntaxMacros/MacroSystem.swift
+++ b/Sources/SwiftSyntaxMacros/MacroSystem.swift
@@ -407,11 +407,11 @@ extension MacroApplication {
   // refer to conformance macros, expand them and return the resulting
   // set of extension declarations.
   private func expandConformances(of decl: DeclGroupSyntax) -> [DeclSyntax] {
-    let identifier: String
+    let extendedType: Syntax
     if let identified = decl.asProtocol(IdentifiedDeclSyntax.self) {
-      identifier = identified.identifier.text
+      extendedType = Syntax(identified.identifier.trimmed)
     } else if let ext = decl.as(ExtensionDeclSyntax.self) {
-      identifier = ext.extendedType.trimmedDescription
+      extendedType = Syntax(ext.extendedType.trimmed)
     } else {
       return []
     }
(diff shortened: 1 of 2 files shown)


DougGregor:
@swift-ci please test

Sources/SwiftSyntaxMacros/MacroSystem.swift
```diff
@@ -394,6 +403,43 @@ extension MacroApplication {
     return peers
   }
 
+  // If any of the custom attributes associated with the given declaration
+  // refer to conformance macros, expand them and return the resulting
+  // set of extension declarations.
+  private func expandConformances(of decl: DeclGroupSyntax) -> [DeclSyntax] {
+    let identifier: String
```
ahoppen, line 410: Why not make `identifier: Syntax` and then you don’t need to convert to strings or  use the `raw` interpolation below. Also `identifier` isn’t correct because `ext.extendedType` is not an identifier.

Sources/SwiftSyntaxMacros/MacroSystem.swift
```diff
@@ -394,6 +403,43 @@ extension MacroApplication {
...
+    for (attribute, conformanceMacro) in macroAttributes {
+      do {
+        let newConformances = try conformanceMacro.expansion(of: attribute, providingConformancesOf: decl, in: context)
+
+        for (type, whereClause) in newConformances {
+          var ext: DeclSyntax = """
+            extension \(raw: identifier): \(type) { }
+            """
+          if let whereClause {
+            ext = DeclSyntax((ext.as(ExtensionDeclSyntax.self))!.with(\.genericWhereClause, whereClause))
```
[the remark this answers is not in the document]
DougGregor: Thank you, I've pulled this in!

Sources/SwiftSyntaxMacros/MacroSystem.swift
DougGregor: Great point, I've updated this code.

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test Windows

PR #1774, opened 2023-06-13
[Macros] Opaque params in expansion functions in macro protocols

rintaro:
Aligning with macro evolution proposals. Should be NFC for compilation.
But editor functionalities like code completion are affected by this, and `some` is preferred over explicit generic parameters.


rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test macOS

PR #1777, opened 2023-06-13
Fix warnings when generating documentation using docc

ahoppen:


ahoppen:
@swift-ci Please test

Sources/SwiftParser/Parser.swift
```diff
@@ -46,7 +46,7 @@
 ///     /*  */
 ///     let rbrace = self.expect(.rightBrace)
 ///
-/// Unlike ``TokenConsumer/eat(_:)``, ``Parser/expect(_:remapping:)`` returns
+/// Unlike `TokenConsumer.eat(_:)`, `Parser.expect(_:remapping:)` returns
```
bnbarham, line 49: Is there no way to link to these (and all the others that had similar treatment)?

ahoppen: No, because `TokenConsumer.eat` is not public, it doesn’t show up in the generated documentation and hence there’s no way to link to it.

bnbarham: Ah I see 👍

bnbarham (approved):

PR #1783, opened 2023-06-14
Move `IncrementalParseTransition` to `SwiftParser`

StevenWong12:
Also move `ConcurrentEdits` to `SwiftSyntax/Utils.swift`, since it might have its own value other than incremental parse in the future.

### DIFF
--- a/Sources/SwiftParser/IncrementalParseTransition.swift
+++ b/Sources/SwiftParser/IncrementalParseTransition.swift
@@ -10,6 +10,7 @@
 //
 //===----------------------------------------------------------------------===//
 
+@_spi(RawSyntax) import SwiftSyntax
 /// Accepts the re-used ``Syntax`` nodes that `IncrementalParseTransition`
 /// determined they should be re-used for a parse invocation.
 ///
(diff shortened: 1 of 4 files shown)


Sources/SwiftParser/IncrementalParseTransition.swift
```diff
@@ -285,20 +152,20 @@ public struct IncrementalParseLookup {
     // text, e.g. because `private struct Foo {}` parses as a CodeBlockItem with
     // a StructDecl inside and `private struc Foo {}` parses as two
     // CodeBlockItems one for `private` and one for `struc Foo {}`
-    var nextLeafNodeLength: SourceLength = .zero
+    var nextLeafNodeLength = 0
```
ahoppen, line 155: Is there any reason why we can’t use `SourceLength` anymore? It is pretty useful just to ensure everybody is talking about lengths in terms of UTF-8 bytes and we don’t accidentally sneak lengths in terms of Unicode grapheme clusters `String.count` in.

ahoppen (commented):
Thank you. `ConcurrentEdits` is really targeted for incremental parsing as well though because `IncrementalEdit` doesn’t even contain the text that a range is being replace with, so I don’t expect it to have any use cases outside the parser. Thus, I would just move it to the SwiftParser module as well.

StevenWong12:
Ah, I see. I've moved `ConcurrentEdits` back to `IncrementalParseTransition.swift`.

Sources/SwiftParser/IncrementalParseTransition.swift
StevenWong12: I think that's something left when I hacked around. I've changed this to `SourceLength`.

Tests/SwiftSyntaxTest/SequentialToConcurrentEditTranslationTests.swift
ahoppen, line 1: As discussed offline, can you move this test case to SwiftParserTest because the tested classes now live in the SwiftParser module?

StevenWong12: Yes, sure.

ahoppen (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

StevenWong12:
It seems that I forgot to move reference in `CmakeLists` as well, which causes CI failure. I've updated this.

kimdv:
@swift-ci please test

PR #1758, opened 2023-06-09
assertMacroExpansion should assert if there are errors

bnbarham:
Invalid macro expansions should generate an assertion in `assertMacroExpansion` rather than being ignored.

PR #1785, opened 2023-06-14
[5.9] assertMacroExpansion should assert if there are errors

bnbarham:
* Explanation: `assertMacroExpansion` is used for testing macro expansions. It should check that the expanded macro is itself valid.
* Scope: Macro unit testing
* Risk: Very low. It's possible some existing tests may now fail, but that means that the underlying expansion needs to be updated.
* Original PR: https://github.com/apple/swift-syntax/pull/1758


bnbarham:
@swift-ci please test

bnbarham:

Issue #1786, opened 2023-06-14
Macro expansion sometimes adds leading spaces inside string literals


tgoyne:
### Description

```swift
struct Names: MemberMacro {
    public static func expansion(
        of node: AttributeSyntax,
        providingMembersOf declaration: some DeclGroupSyntax,
        in context: some MacroExpansionContext
    ) throws -> [DeclSyntax] {
        guard let declaration = declaration.as(ClassDeclSyntax.self) else { return [] }
        let className = declaration.identifier.trimmed
        let names = declaration.memberBlock.members.compactMap { (member) -> String? in
            guard let property = member.decl.as(VariableDeclSyntax.self),
                  let binding = property.bindings.first,
                  let identifier = binding.pattern.as(IdentifierPatternSyntax.self)
            else {
                return nil
            }
            return identifier.identifier.text
        }
        let body = names.map { name in
            ExprSyntax("\(literal: name): \\\(className).\(raw: name)")
        }
        return ["""
        public static var names: [String: AnyKeyPath] = [\(raw: body.map(\.trimmedDescription).joined(separator: ", "))]
        public static var names2: [String: AnyKeyPath] = [\(raw: body.map(\.trimmedDescription).joined(separator: ", "))]
        """]
    }
}
assertMacroExpansion(
    #"""
    @Names class Foo {
        var value1: Int
        var value2: Int
        var value3: Int
    }
    """#,
    expandedSource: #"""
    @Names class Foo {
        var value1: Int
        var value2: Int
        var value3: Int
        public static var names: [String: AnyKeyPath] = ["value1": \Foo.value1, "value2": \Foo.value2, "value3": \Foo.value3]
        public static var names2: [String: AnyKeyPath] = ["value1": \Foo.value1, "value2": \Foo.value2, "value3": \Foo.value3]
    }
    """#,
    macros: [
        "Names": Names.self,
    ])
```

The actual result of this macro is:

```swift
@Names class Foo {
    var value1: Int
    var value2: Int
    var value3: Int
    public static var names: [String: AnyKeyPath] = ["value1": \Foo.value1, "value2": \Foo.value2, "value3": \Foo.value3]
    public static var names2: [String: AnyKeyPath] = [" value1": \Foo.value1, " value2": \Foo.value2, " value3": \Foo.value3]
}
```

Somehow each of the string literals in `names2` has add a space added to the beginning even though it is generated by the exact same code as `names`, which does not have the spaces.

ahoppen:
[...]

PR #1790, opened 2023-06-15
When creation of a syntax node using string interpolation failed, log the diagnostics

ahoppen:
From the first feedback of developers starting to build macros, the number one mistake was to build incorrect syntax nodes, e.g. build an `ExprSyntax` for a declaration. For example

- https://github.com/apple/swift-syntax/issues/1792
- https://github.com/apple/swift-syntax/issues/1786

Up until recently, these errors weren’t diagnosed at all. By now we at least issue an XCTest assertion if the expanded macro code contains syntax errors. But I think the issues are a lot easier to debug if we can emit an error at the location where the incorrect syntax node is being created.

Since we can’t throw an error from string interpolation, use OSLog to log the parsing error to Xcode or OS console. When OSLog is not available or we are building using CMake (and thus don’t want to introduce any dependency on the SDK), silently accept the error as we do right now. Logging to stderr might be too fragile in case some application is expecting to populate stderr itself.

### DIFF
--- a/Sources/SwiftSyntaxBuilder/SyntaxParsable+ExpressibleByStringInterpolation.swift
+++ b/Sources/SwiftSyntaxBuilder/SyntaxParsable+ExpressibleByStringInterpolation.swift
@@ -1,0 +1,61 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftDiagnostics
+import SwiftSyntax
+import SwiftParser
+import SwiftParserDiagnostics
+// Don't introduce a dependency on OSLog when building SwiftSyntax using CMake
+// for the compiler.
+#if canImport(OSLog) && !SWIFTSYNTAX_NO_OSLOG_DEPENDENCY
+import OSLog
+#endif
+
+extension SyntaxParseable {
+  public typealias StringInterpolation = SyntaxStringInterpolation
+
+  /// Assuming that this node contains a syntax error, log it using OSLog if we
+  /// are on a platform that supports OSLog, otherwise don't do anything.
+  private func logStringInterpolationParsingError() {
+    #if canImport(OSLog) && !SWIFTSYNTAX_NO_OSLOG_DEPENDENCY
+    if #available(macOS 11.0, *) {
+      let diagnostics = ParseDiagnosticsGenerator.diagnostics(for: self)
+      let formattedDiagnostics = DiagnosticsFormatter().annotatedSource(tree: self, diags: diagnostics)
+      Logger(subsystem: "SwiftSyntax", category: "ParseError").fault(
+        """
+        Parsing a `\(Self.self)` node from string interpolation produced the following parsing errors.
+        Set a breakpoint in `SyntaxParseable.logStringInterpolationParsingError()` to debug the failure.
+        \(formattedDiagnostics)
+        """
+      )
+    }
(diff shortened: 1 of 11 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

kimdv:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

Sources/SwiftSyntaxBuilder/SyntaxParsable+ExpressibleByStringInterpolation.swift
```diff
@@ -0,0 +1,56 @@
...
+  /// are on a platform that supports OSLog, otherwise don't do anything.
+  private func logStringInterpolationParsingError() {
+    #if canImport(OSLog) && !SWIFTSYNTAX_NO_OSLOG_DEPENDENCY
+    if #available(macOS 11.0, *) {
+      let diagnostics = ParseDiagnosticsGenerator.diagnostics(for: self)
+      let formattedDiagnostics = DiagnosticsFormatter().annotatedSource(tree: self, diags: diagnostics)
+      Logger(subsystem: "SwiftSyntax", category: "ParseError").fault(
+        """
+        Parsing a `\(Self.self)` node from string interpolation produced the following parsing errors.
+        Set a brakpoint in `SyntaxParseable.logStringInterpolationParsingError()` to debug the failure.
```
bnbarham, line 36:
```suggestion
        Parsing a `\(Self.self)` node from string interpolation produced the following parsing errors.
        Set a breakpoint in `SyntaxParseable.logStringInterpolationParsingError()` to debug the failure.
```

Tests/SwiftParserTest/LinkageTests.swift
```diff
@@ -69,9 +69,27 @@ final class LinkageTest: XCTestCase {
         .library("-lswiftCompatibilityConcurrency"),
         .library("-lswiftCompatibilityPacks", condition: .mayBeAbsent("Only in newer compilers")),
         .library("-lswiftCore"),
+        .library("-lswiftCoreFoundation", condition: .when(compilationCondition: .osLogDependency)),
```
bnbarham, line 72: We could just only run this check when we don't have the oslog dependency. We don't really care what's linked if we do.

ahoppen: You mean put the entire test behind `#if canImport(OSLog) && !SWIFTSYNTAX_NO_OSLOG_DEPENDENCY`?

This test is starting to get a little odd because in practice we only ever run this test using `swift test` on macOS where we don’t set `SWIFTSYNTAX_NO_OSLOG_DEPENDENCY`, so it does raise a question about its value.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

rintaro:
> But I think the issues are a lot easier to debug if we can emit an error at the location where the incorrect syntax node is being created.

It's not clear for me. How this OS log indicates the location? I think the information in the OS is the same as the parser diagnostics as https://github.com/apple/swift-syntax/pull/1758/files , no?

rintaro:
I feel it's `MacroSystem`'s responsibility to somehow propagate which expansion had errors.
Using OSLog for this just because string interpolation doesn't allow `throws` doesn't sound like a collect motivation. Also I believe what we want to validate is that "macro expansion should not have syntax errors", but not "string interpolation in general should not have errors".

Instead, I think `MacroSystem`'s each expansion should check the `hasError` flag and propagate it to the test system somehow with the information which expansion had errors.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test macOS

ahoppen:
@swift-ci Please test macOS

ahoppen:
@swift-ci Please test macOS

PR #1791, opened 2023-06-15
Fix the indentation for nodes that have a parent node

### LINKED ISSUE #1744: Extraneous indentation if the formatted node has parents
The following new test case fails

```swift
let decl: DeclSyntax = """
  func test() {
    print(1)
  }
"""

let body = decl.cast(FunctionDeclSyntax.self).body!

let expectedFormatted = """
  {
      print(1)
    }
  """

assertStringsEqualWithDiff(body.formatted().description, expectedFormatted) // fails: `print` is indented by 8 spaces
assertStringsEqualWithDiff(body.detach().formatted().description, expectedFormatted) // passes
```

For some reason we are adding 8 spaces in front of `print` if `body` is not detached.

gibachan:
Resolves #1744


Sources/SwiftBasicFormat/BasicFormat.swift
```diff
@@ -393,7 +394,7 @@ open class BasicFormat: SyntaxRewriter {
       }
     }
 
-    if leadingTrivia.indentation(isOnNewline: previousTokenWillEndWithNewline) == [] {
+    if leadingTrivia.indentation(isOnNewline: isInitialToken || previousTokenWillEndWithNewline) == [] {
```
gibachan, line 397: When the formatter takes a node that has a parent node, I think the node should be considered as a starting point for detecting indentation anchor points.

Tests/SwiftBasicFormatTest/BasicFormatTests.swift
```diff
@@ -341,4 +341,22 @@ final class BasicFormatTest: XCTestCase {
         """
     )
   }
+
+  func testNodeHavingParentNode() {
+    let decl: DeclSyntax = """
+      func test() {
+        print(1)
+      }
```
ahoppen, line 349: I just realized that there is a simpler and more intuitive example for this behavior:

```suggestion
    func test() {
        print(1)
    }
```

Sources/SwiftBasicFormat/BasicFormat.swift
ahoppen: Oh, very nice find. I was not expecting the issue in this area. 

This raises a more fundamental question: What does it even mean to format a `CodeBlockSyntax` item standalone while it still has a parent that’s a `FunctionDeclSyntax`? I’m not sure if there really is a correct answer so maybe calling `formatted` should detach the syntax node from the tree, in which case we can always assume that we are formatting an entire tree and not just a subtree. What do you think?

CC @bnbarham: Do you have any opinions on my thoughts above?

Tests/SwiftBasicFormatTest/BasicFormatTests.swift
gibachan: Fixed it 👍 [9d40557](https://github.com/apple/swift-syntax/pull/1791/commits/9d4055702bf696227c66c9bf64152a580cafce45)

Sources/SwiftBasicFormat/BasicFormat.swift
gibachan: Thank you for your consideration! I agree with you and have included your test case for now.

ahoppen: OK, thank you. Do you want to work on adding 4 spaces in front of `{`? If not, I’ll file an issue for it.

ahoppen (approved):

ahoppen:
@swift-ci Please test

Sources/SwiftBasicFormat/BasicFormat.swift
gibachan: Yes! I'd love to work on it 😄

ahoppen: That’s great. I’m looking forward to your PR.

ahoppen:
The indentation doesn’t seem to match `swift-format`. Could you run `./format.py`?

gibachan:
Ah, I did it and squashed the commits. Thanks!

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

kimdv:
@swift-ci please test windows

PR #1793, opened 2023-06-15
Don’t use `unsafeFlags` when building SwiftSyntax

ahoppen:
The group info file was only being used when building SwiftSyntax as part of the toolchain. But nobody should be using SwiftSyntax from the toolchain, so it wasn’t providing any value.

Exclusivity checking is no longer a performance problem, so there’s no reason to disable it anymore.


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

bnbarham (approved):

PR #1794, opened 2023-06-15
[5.9][bazel] Enable more strict dependencies

keith:
When using the more strict dependencies feature missing dependencies
cause build failures. This is now enabled in this project to catch these
sooner.


keith:
@swift-ci please test

PR #1796, opened 2023-06-15
[bazel] Fix missing dependencies

keith:
When using the more strict dependencies feature these missing dependencies caused build failures. This is now enabled in this project to catch these sooner.


keith:
@swift-ci please test

PR #1797, opened 2023-06-16
Allow user to depend on remote dependencies when running `build-script`

StevenWong12:
Do we have special reason that we need to have a local `swift-argument-parser` when building? @ahoppen 

If not, I think it would be nice to provide a flag to allow user to depend on remote dependencies when building or testing using `build-script`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,18 +7,22 @@ If you are seeing issues, it is generally a good idea to re-try with the latest 
 
 The easiest option to build SwiftSyntax it to open it in Xcode. 
 
-Alternatively you can also build it from the command line using `build-script.py`. To do this, perform the following steps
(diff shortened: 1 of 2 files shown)


ahoppen:
I think we used to have that option but removed it with the expectation that almost all local development uses `swift build` etc directly and don’t go through `build-script.py`, so we didn’t really see a need for it anymore. What’s your use case? Running the lit tests?

StevenWong12:
Yeah, I still invoke it when running lit tests.

ahoppen:
The proper solution here should be to just delete all the lit tests. I think we are on a good track for that. I suspect that you are the only person hitting this right now and I would suggest you just check out swift-argument-parser alongside swift-syntax. I would just prefer to not add another option to `build-script.py` that we then need to think carefully about removing again.

StevenWong12:
Ah, I see.

PR #1795, opened 2023-06-15
InitAccessors: Parse initializer exprs associated with computed prope…

xedin:
…rties that have `init` accessor

Initialization expressions are not allowed on computed properties but 
if a property has `init` accessor it should be allowed because it could
be used by a memberwise initializer.

PR #1803, opened 2023-06-16
[5.9] InitAccessors: Parse initializer exprs associated with computed prope…

xedin:
…rties that have `init` accessor

Cherry-pick of https://github.com/apple/swift-syntax/pull/1795

--- 

Initialization expressions are not allowed on computed properties 
but if a property has `init` accessor it should be allowed because
it could be used by a memberwise initializer.

(cherry picked from commit b05cd4e62a6fc7ae6c9fe2ae7773f5dcd478d493)


xedin:
@swift-ci please test

PR #1809, opened 2023-06-18
Improve diagnostics for invalid version tuples

### LINKED ISSUE #1613: `10e10` is not emitted as an invalid version tuple

TTOzzi:
Resolve #1613 

During the process of rebasing the changes from the main branch, an issue arose, leading to the closure and reopening of the previous PR(#1717).
I sincerely apologize for the inconvenience caused.

Based on the feedback provided in the comment here(https://github.com/apple/swift-syntax/pull/1717#discussion_r1221876190), I have made the necessary revisions and added several relevant test cases.

### DIFF
--- a/Sources/SwiftParser/Availability.swift
+++ b/Sources/SwiftParser/Availability.swift
@@ -253,6 +253,23 @@ extension Parser {
     return self.consumeAnyToken()
   }
 
+  /// Consume the unexpected version token(e.g. integerLiteral, floatingLiteral, identifier) until the period no longer appears.
+  private mutating func parseUnexpectedVersionTokens() -> RawUnexpectedNodesSyntax? {
+    var unexpectedTokens: [RawTokenSyntax] = []
+    var keepGoing: RawTokenSyntax? = nil
+    var loopProgress = LoopProgressCondition()
+    repeat {
+      if let keepGoing {
+        unexpectedTokens.append(keepGoing)
+      }
+      if let unexpectedVersion = self.consume(if: .integerLiteral, .floatingLiteral, .identifier) {
+        unexpectedTokens.append(unexpectedVersion)
+      }
+      keepGoing = self.consume(if: .period)
+    } while keepGoing != nil && loopProgress.evaluate(currentToken)
+    return RawUnexpectedNodesSyntax(unexpectedTokens, arena: self.arena)
+  }
+
   /// Parse a dot-separated list of version numbers.
   ///
   /// Grammar
@@ -286,24 +303,29 @@ extension Parser {
         } else {
           trailingComponents.append(versionComponent)
         }
+
+        if version.isMissing {
+          break
+        }
       }
 
-      var unexpectedTrailingComponents: RawUnexpectedNodesSyntax?
-
-      if !trailingComponents.isEmpty {
-        unexpectedTrailingComponents = RawUnexpectedNodesSyntax(elements: trailingComponents.compactMap { $0.as(RawSyntax.self) }, arena: self.arena)
-      }
-
+      let unexpectedTrailingComponents = RawUnexpectedNodesSyntax(trailingComponents, arena: self.arena)
+      let unexpectedAfterComponents = self.parseUnexpectedVersionTokens()
       return RawVersionTupleSyntax(
         major: major,
         components: RawVersionComponentListSyntax(elements: components, arena: self.arena),
-        unexpectedTrailingComponents,
+        RawUnexpectedNodesSyntax(combining: unexpectedTrailingComponents, unexpectedAfterComponents, arena: self.arena),
         arena: self.arena
       )
-
     } else {
       let major = self.expectDecimalIntegerWithoutRecovery()
-      return RawVersionTupleSyntax(major: major, components: nil, arena: self.arena)
+      let unexpectedAfterComponents = self.parseUnexpectedVersionTokens()
+      return RawVersionTupleSyntax(
+        major: major,
+        components: nil,
+        unexpectedAfterComponents,
+        arena: self.arena
+      )
     }
   }
 }
--- a/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
+++ b/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
@@ -1893,14 +1893,22 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
       return .skipChildren
     }
 
-    if let trailingComponents = node.unexpectedAfterComponents,
-      let components = node.components
-    {
-      addDiagnostic(
-        trailingComponents,
-        TrailingVersionAreIgnored(major: node.major, components: components),
-        handledNodes: [trailingComponents.id]
-      )
+    if let unexpectedAfterComponents = node.unexpectedAfterComponents {
+      if let components = node.components,
+        unexpectedAfterComponents.allSatisfy({ $0.is(VersionComponentSyntax.self) })
+      {
+        addDiagnostic(
+          unexpectedAfterComponents,
+          TrailingVersionAreIgnored(major: node.major, components: components),
+          handledNodes: [unexpectedAfterComponents.id]
+        )
+      } else {
+        addDiagnostic(
+          unexpectedAfterComponents,
+          CannotParseVersionTuple(versionTuple: unexpectedAfterComponents),
+          handledNodes: [node.major.id, node.components?.id, unexpectedAfterComponents.id].compactMap { $0 }
+        )
+      }
     }
 
     return .visitChildren
--- a/Tests/SwiftParserTest/AvailabilityTests.swift
+++ b/Tests/SwiftParserTest/AvailabilityTests.swift
@@ -144,17 +144,19 @@ final class AvailabilityTests: XCTestCase {
 
     assertParse(
       """
+      @available(OSX 10.0.1, *)
+      func test() {}
+      """
+    )
+
+    assertParse(
+      """
       @available(OSX 1️⃣10e10)
       func test() {}
       """,
       diagnostics: [
-        DiagnosticSpec(message: "expected version tuple in version restriction", fixIts: ["insert version tuple"]),
-        DiagnosticSpec(message: "unexpected code '10e10' in attribute"),
-      ],
-      fixedSource: """
-        @available(OSX <#integer literal#>10e10)
-        func test() {}
-        """
+        DiagnosticSpec(message: "cannot parse version component code '10e10'")
+      ]
     )
 
     assertParse(
@@ -163,13 +165,8 @@ final class AvailabilityTests: XCTestCase {
       func test() {}
       """,
       diagnostics: [
-        DiagnosticSpec(message: "expected integer literal in version tuple", fixIts: ["insert integer literal"]),
-        DiagnosticSpec(message: "unexpected code '0e10' in attribute"),
-      ],
-      fixedSource: """
-        @available(OSX 10.<#integer literal#>0e10)
-        func test() {}
-        """
+        DiagnosticSpec(message: "cannot parse version component code '0e10'")
+      ]
     )
 
     assertParse(
@@ -178,13 +175,8 @@ final class AvailabilityTests: XCTestCase {
       func test() {}
       """,
       diagnostics: [
-        DiagnosticSpec(message: "expected version tuple in version restriction", fixIts: ["insert version tuple"]),
-        DiagnosticSpec(message: "unexpected code '0xff' in attribute"),
-      ],
-      fixedSource: """
-        @available(OSX <#integer literal#>0xff)
-        func test() {}
-        """
+        DiagnosticSpec(message: "cannot parse version component code '0xff'")
+      ]
     )
 
     assertParse(
@@ -193,13 +185,28 @@ final class AvailabilityTests: XCTestCase {
       func test() {}
       """,
       diagnostics: [
-        DiagnosticSpec(message: "expected integer literal in version tuple", fixIts: ["insert integer literal"]),
-        DiagnosticSpec(message: "unexpected code '0xff' in attribute"),
-      ],
-      fixedSource: """
-        @available(OSX 1.0.<#integer literal#>0xff)
-        func test() {}
-        """
+        DiagnosticSpec(message: "cannot parse version component code '0xff'")
+      ]
+    )
+
+    assertParse(
+      """
+      @available(OSX 1.0.1️⃣0xff, *)
+      func test() {}
+      """,
+      diagnostics: [
+        DiagnosticSpec(message: "cannot parse version component code '0xff'")
+      ]
+    )
+
+    assertParse(
+      """
+      @available(OSX 1.0.1️⃣asdf)
+      func test() {}
+      """,
+      diagnostics: [
+        DiagnosticSpec(message: "cannot parse version component code 'asdf'")
+      ]
     )
   }
 }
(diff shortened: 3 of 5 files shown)


Sources/SwiftParser/Availability.swift
```diff
@@ -262,48 +262,97 @@ extension Parser {
...
+          if i < maxComponentCount {
+            components.append(versionComponent)
+          } else {
+            trailingComponents.append(versionComponent)
+          }
 
-      if !trailingComponents.isEmpty {
-        unexpectedTrailingComponents = RawUnexpectedNodesSyntax(elements: trailingComponents.compactMap { $0.as(RawSyntax.self) }, arena: self.arena)
-      }
+          if versionComponent.hasError {
```
TTOzzi, line 291: If the versionComponent has an error, the loop will be stopped, and both the major and the parsed components up to that point, as well as all tokens before the comma or right parenthesis, will be treated as unexpectedToken.

Sources/SwiftParser/Availability.swift
```diff
@@ -262,48 +262,96 @@ extension Parser {
...
+          RawUnexpectedNodesSyntax([unexpectedToken], arena: self.arena),
+          major: self.missingToken(.integerLiteral, text: nil),
+          components: nil,
+          arena: self.arena
+        )
+      }
     } else {
       let major = self.expectDecimalIntegerWithoutRecovery()
-      return RawVersionTupleSyntax(major: major, components: nil, arena: self.arena)
+      if major.isMissing {
```
TTOzzi, line 333: Similarly, in this context, if the integer literal major is missing, all tokens before the comma or right paren will be treated as unexpectedToken.
There are some duplicated codes. Would it be better to separate it into a function? 🤔

Tests/SwiftParserTest/AvailabilityTests.swift
```diff
@@ -142,34 +142,31 @@ final class AvailabilityTests: XCTestCase {
       )
     )
 
+    assertParse(
+      """
+      @available(OSX 10.0.1, *)
+      func test() {}
+      """
+    )
```
TTOzzi, line 150: I have added several test cases that include the comma.

Sources/SwiftParser/Availability.swift
ahoppen: I believe this incorrectly marks the entire version as invalid if you have eg. `1.2.0xff`. Ideally we should only be complaining about the `0xff` here.

Sources/SwiftParser/Availability.swift
```diff
@@ -262,48 +262,96 @@ extension Parser {
...
   ///     version-tuple-element -> '.' interger-literal
   mutating func parseVersionTuple(maxComponentCount: Int) -> RawVersionTupleSyntax {
-    if self.at(.floatingLiteral),
-      let periodIndex = self.currentToken.tokenText.firstIndex(of: UInt8(ascii: ".")),
-      self.currentToken.tokenText[0..<periodIndex].allSatisfy({ Unicode.Scalar($0).isDigit })
-    {
-      // The lexer generates a float literal '1.2' for the major and minor version.
-      // Split it into two integers if possible
-      let major = self.consumePrefix(SyntaxText(rebasing: self.currentToken.tokenText[0..<periodIndex]), as: .integerLiteral)
+    if self.at(.floatingLiteral) {
```
ahoppen, line 266: Superfluous newline?

Sources/SwiftParser/generated/Parser+TokenSpecSet.swift
```diff
@@ -2084,3 +2084,30 @@ extension VariableDeclSyntax {
     }
   }
 }
+
+extension VersionTupleSyntax {
```
ahoppen, line 2088: This file is automatically generated and if you can’t manually add content to it. If we need this, we should place it in TokenSpecSet.swift

Sources/SwiftParser/Availability.swift
```diff
@@ -262,48 +262,96 @@ extension Parser {
...
+          handle.unexpectedTokens > 0
+        {
+          var unexpectedTokens = [RawSyntax]()
+          for _ in 0..<handle.unexpectedTokens {
+            unexpectedTokens.append(RawSyntax(self.consumeAnyToken()))
+          }
+          unexpectedNodes = RawUnexpectedNodesSyntax(elements: unexpectedTokens, arena: self.arena)
+        } else {
+          unexpectedNodes = nil
+        }
```
ahoppen, line 345: I would prefer this to be a whitelist of characters that we want to consume instead of consuming until the end of a few characters that we currently assume to terminate a `VersionTupleSyntax` (think about how `VersionTupleSyntax` could be used in the future in some context that doesn’t terminate it using `)` or `,`. I.e. I would prefer to just continue eating all period, integer and float literals here.

Also, I think it would be good to share this code snippet between the two `if` branches.

Sources/SwiftParser/generated/Parser+TokenSpecSet.swift
TTOzzi: Oh, I'm sorry. I missed that it was a subfile of `generated` 😅

Sources/SwiftParser/Availability.swift
TTOzzi: I understand. As you mentioned, checking for right paren and commas seems too limited of an approach. 
Although I'm not sure if I fully understood your intention, I have made changes to eat integer and floating literals while checking for periods. 
Thank you for your detailed feedback 😃

Tests/SwiftParserTest/AvailabilityTests.swift
```diff
@@ -178,26 +175,41 @@ final class AvailabilityTests: XCTestCase {
...
         DiagnosticSpec(message: "expected integer literal in version tuple", fixIts: ["insert integer literal"]),
-        DiagnosticSpec(message: "unexpected code '0xff' in attribute"),
+        DiagnosticSpec(message: "unexpected code 'asdf' in attribute"),
       ],
       fixedSource: """
-        @available(OSX 1.0.<#integer literal#>0xff)
+        @available(OSX 1.0.<#integer literal#>asdf)
         func test() {}
         """
     )
```
TTOzzi, line 215: I have a question I would like to ask you. 
When checking for unexpected tokens, if only integers and floating literals that are included in the VersionTupleWhiteList are considered, all existing test cases generate the intended diagnosis. 
However, when there is a string literal like `asdf`, it generates a different diagnosis. 
Should we consider this as a separate case? 🤔

Sources/SwiftParser/Availability.swift
```diff
@@ -253,6 +253,28 @@ extension Parser {
...
+  /// Consume the unexpected version token(e.g. integerLiteral, floatingLiteral) until the period no longer appears.
+  private mutating func parseUnexpectedVersionTokens(unexpectedTokens: [RawSyntax] = []) -> RawUnexpectedNodesSyntax? {
+    var copiedUnexpectedTokens = unexpectedTokens
+    var keepGoing: RawTokenSyntax? = nil
+    var loopProgress = LoopProgressCondition()
+    repeat {
+      if let keepGoing {
+        copiedUnexpectedTokens.append(RawSyntax(keepGoing))
+      }
+      if let (_, handle) = self.at(anyIn: VersionTupleWhiteList.self) {
```
ahoppen, line 265: If you are only checking for integer and float without exhaustively switching over them you can just use `consume(if:)` and don’t need to define a `TokenSpecSubset`

```suggestion
      if versionToken = self.consume(if: .integerLiteral, .floatingLiteral)
```

Sources/SwiftParser/Availability.swift
```diff
@@ -253,6 +253,28 @@ extension Parser {
     return self.consumeAnyToken()
   }
 
+  /// Consume the unexpected version token(e.g. integerLiteral, floatingLiteral) until the period no longer appears.
+  private mutating func parseUnexpectedVersionTokens(unexpectedTokens: [RawSyntax] = []) -> RawUnexpectedNodesSyntax? {
```
ahoppen, line 257: I don’t think there’s a need to pass `unexpectedTokens` in here. This can just return `RawUnexpectedNodesSyntax?` and then you can combine the two unexpected lists in the function that calls this.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -1893,14 +1893,31 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
+      } else {
+        var handledNodes: [SyntaxIdentifier] = [unexpectedAfterComponents.id]
+        if node.major.hasError {
+          handledNodes.append(node.major.id)
+        }
+        if let components = node.components,
+          components.hasError
+        {
+          handledNodes.append(components.id)
+        }
```
ahoppen, line 1914: I think you should be able to just unconditionally add `node.major.id` and `components.id` to `handledNodes`

Tests/SwiftParserTest/AvailabilityTests.swift
```diff
@@ -142,19 +142,21 @@ final class AvailabilityTests: XCTestCase {
...
       """,
       diagnostics: [
-        DiagnosticSpec(message: "expected version tuple in version restriction", fixIts: ["insert version tuple"]),
-        DiagnosticSpec(message: "unexpected code '10e10' in attribute"),
-      ],
-      fixedSource: """
-        @available(OSX <#integer literal#>10e10)
-        func test() {}
-        """
+        DiagnosticSpec(message: "cannot parse version 10e10")
```
ahoppen, line 158: Ideally the error message here would be `cannot parse version component '10e10'`, i.e. add *component* and wrap 10e10 in single quotes. You should get the quote wrapping automatically if you use `shortSingleLineContentDescription`.

Tests/SwiftParserTest/AvailabilityTests.swift
ahoppen: If you want, you could also eat identifiers in addition to integer and float literals when consuming the unexpected version components.

TTOzzi: Oh, thank you! It's working just as I expected.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
TTOzzi: That's right, it's behaving just like adding always 😅

Sources/SwiftParser/Availability.swift
TTOzzi: Oh, there was a useful initializer that combines RawUnexpectedNodesSyntaxs! I've made the modification to use it.

Sources/SwiftParser/Availability.swift
```diff
@@ -286,24 +307,29 @@ extension Parser {
...
-      var unexpectedTrailingComponents: RawUnexpectedNodesSyntax?
-
-      if !trailingComponents.isEmpty {
-        unexpectedTrailingComponents = RawUnexpectedNodesSyntax(elements: trailingComponents.compactMap { $0.as(RawSyntax.self) }, arena: self.arena)
+        if version.isMissing {
+          break
+        }
       }
 
+      let unexpectedTrailingComponents = RawUnexpectedNodesSyntax(elements: trailingComponents.map(RawSyntax.init), arena: self.arena)
```
ahoppen, line 316: I would do the following

```suggestion
      let unexpectedTrailingComponents = RawUnexpectedNodesSyntax(elements: trailingComponents.map(RawSyntax.init), arena: self.arena)
```

This uses the following initializer

https://github.com/apple/swift-syntax/blob/1555299f9e1fac77cf92677dae4dc8b46e9aa3d0/Sources/SwiftParser/SyntaxUtils.swift#L39-L43

It has two advantages:
1. If `trailingComponents` is empty, it doesn’t allocate a `RawUnexpectedNodesSyntax`, reducing the memory footprint (remember that raw syntax nodes are allocated in a `SyntaxArena` and thus don’t get freed until the entire arena gets freed)
2. It’s also shorter 😉

Sources/SwiftParser/Availability.swift
```diff
@@ -253,6 +253,27 @@ extension Parser {
...
+      if let unexpectedVersion = self.consume(if: .integerLiteral, .floatingLiteral, .identifier) {
+        unexpectedTokens.append(RawSyntax(unexpectedVersion))
+      }
+      keepGoing = self.consume(if: .period)
+    } while keepGoing != nil && loopProgress.evaluate(currentToken)
+
+    guard !unexpectedTokens.isEmpty else {
+      return nil
+    }
+    return RawUnexpectedNodesSyntax(elements: unexpectedTokens, arena: self.arena)
```
ahoppen, line 274: You can simplify this slightly to the following. You can also change `unexpectedTokens` to `[RawTokenSyntax]`.
```suggestion
    return RawUnexpectedNodesSyntax(unexpectedTokens, arena: self.arena)
```

This uses the following convenience initializer.

https://github.com/apple/swift-syntax/blob/1555299f9e1fac77cf92677dae4dc8b46e9aa3d0/Sources/SwiftParser/SyntaxUtils.swift#L39-L43

ahoppen (approved):

Sources/SwiftParser/Availability.swift
TTOzzi: Oh, I understand. It's more efficient to return nil if the array is empty 😃

Sources/SwiftParser/Availability.swift
TTOzzi: I've changed both initializers, thank you for your kind feedback!

Sources/SwiftParser/Availability.swift
```diff
@@ -253,6 +253,23 @@ extension Parser {
...
+    var loopProgress = LoopProgressCondition()
+    repeat {
+      if let keepGoing {
+        unexpectedTokens.append(RawSyntax(keepGoing))
+      }
+      if let unexpectedVersion = self.consume(if: .integerLiteral, .floatingLiteral, .identifier) {
+        unexpectedTokens.append(RawSyntax(unexpectedVersion))
+      }
+      keepGoing = self.consume(if: .period)
+    } while keepGoing != nil && loopProgress.evaluate(currentToken)
```
ahoppen, line 269: You could just make `unexpectedTokens` a `[RawTokenSyntax]`, which simplifies the code a little

```suggestion
    var unexpectedTokens: [RawTokenSyntax] = []
    var keepGoing: RawTokenSyntax? = nil
    var loopProgress = LoopProgressCondition()
    repeat {
      if let keepGoing {
        unexpectedTokens.append(keepGoing)
      }
      if let unexpectedVersion = self.consume(if: .integerLiteral, .floatingLiteral, .identifier) {
        unexpectedTokens.append(unexpectedVersion)
      }
      keepGoing = self.consume(if: .period)
    } while keepGoing != nil && loopProgress.evaluate(currentToken)
```

ahoppen (commented):
One minor comment, otherwise LGTM.

Sources/SwiftParser/Availability.swift
TTOzzi: I have committed your suggestion, thank you!

ahoppen:
@swift-ci Please test

TTOzzi:
Hello @ahoppen 👋 It seems like the CI is completed. When can it be merged?

ahoppen:
Thanks for the ping. I forgot about this PR. I’m sorry.

PR #1812, opened 2023-06-19
[5.9] Fix SyntaxRewriterFile.swift to generate the expected SyntaxRewriter.swift

ahoppen:
We modified SyntaxRewriter.swift at some point without updating the generating SyntaxRewriterFile.swift. Fix that.


ahoppen:
@swift-ci Please test

kimdv (approved):

PR #1832, opened 2023-06-22
Add support for Musl libc

MaxDesiatov:
Since Musl is sufficiently different from Glibc (see https://wiki.musl-libc.org/functional-differences-from-glibc.html), it requires a different import, which now should be applied to files that have `import Glibc` in them.

### DIFF
--- a/Sources/SwiftSyntax/SyntaxText.swift
+++ b/Sources/SwiftSyntax/SyntaxText.swift
@@ -15,6 +15,8 @@
 @_implementationOnly import Darwin
 #elseif canImport(Glibc)
 @_implementationOnly import Glibc
+#elseif canImport(Musl)
+@_implementationOnly import Musl
 #endif
 #endif
 
--- a/Sources/swift-parser-cli/TerminalUtils.swift
+++ b/Sources/swift-parser-cli/TerminalUtils.swift
@@ -12,6 +12,8 @@
 
 #if canImport(Glibc)
 import Glibc
+#elseif canImport(Musl)
+import Musl
 #elseif os(Windows)
 import CRT
 #else


MaxDesiatov:
@swift-ci test

ahoppen (approved):
Out of curiosity: Which platforms use Musl?

MaxDesiatov:
There are multiple Linux distributions that enable Musl by default or as an option. Seems to me the most well-known is Alpine Linux. The main benefits are that binary size is reduced when compared to Glibc, and it also allows much easier full static linking.

ahoppen:
I see. So just to be clear: swift-syntax didn’t use to work on those platforms and with this change it does? And just out of curiosity: Do we have any CI jobs/manual tests/whatever that test this? If not, that’s also fine.

MaxDesiatov:
Thanks for the review!

As long as the Swift toolchain runs on those distributions, Swift Syntax will work too. I don't have PRs for the toolchain itself ready yet, but I have it working locally and there's a consensus with people I've talked to that Musl deserves its own self-titled platform module, using `import Glibc` would be quite confusing.

Merging this PR will simplify the toolchain development for me, as right now I have a dozen of patches I have to keep around and consistently apply when building locally.

MaxDesiatov:
To follow up on the CI question, I think we should have at least a single job that exercises Musl builds, but if that happens we have to land toolchain changes first.

al45tair:
>  there's a consensus with people I've talked to that Musl deserves its own self-titled platform module, using import Glibc would be quite confusing.

To be clear, the fact that importing the C library involves knowing which platform you're on and/or what the C library you're trying to use might be called is ridiculous. The divergent naming largely has to do with the fact that we have little control over what is presented by the various C library modules, so `import Darwin` will give you a different set of functionality to `import Glibc`. Even the details of some of the types can differ, which results in us having to conditionally cast arguments on the Swift side. None of this is good and we should fix it (I have a plan for how), but _for now_ it's fine to add a new `Musl` module, on the understanding that it'll probably go away at some point.

PR #1833, opened 2023-06-22
Add diagnostic if `#endif` is followed by any declarations

### LINKED ISSUE #1823: Additional declaration allowed on same line after `#endif`
### Issue Kind

Other

### Source Code

```swift
struct Foo {
  #if false
  var x: Int
#endif; var x = 1
}
```

### Description

The above example produces an error *Extra tokens following conditional compilation directive* in the old parser but is incorrectly accepted by the new parser.

kimdv:
Resolves #1823

### DIFF
--- a/Sources/SwiftBasicFormat/BasicFormat.swift
+++ b/Sources/SwiftBasicFormat/BasicFormat.swift
@@ -186,12 +186,12 @@ open class BasicFormat: SyntaxRewriter {
     }
 
     switch (first?.tokenKind, second?.tokenKind) {
-
     case (.multilineStringQuote, .backslash),  // string interpolation segment inside a multi-line string literal
       (.multilineStringQuote, .multilineStringQuote),  // empty multi-line string literal
       (.multilineStringQuote, .stringSegment),  // segment starting a multi-line string literal
       (.stringSegment, .multilineStringQuote),  // ending a multi-line string literal that has a string interpolation segment at its end
       (.rightParen, .multilineStringQuote),  // ending a multi-line string literal that has a string interpolation segment at its end
+      (.poundEndifKeyword, _),
       (_, .poundElseKeyword),
       (_, .poundElseifKeyword),
       (_, .poundEndifKeyword),
--- a/Sources/SwiftParser/Directives.swift
+++ b/Sources/SwiftParser/Directives.swift
@@ -163,16 +163,8 @@ extension Parser {
     }
 
     let (unexpectedBeforePoundEndIf, poundEndIf) = self.expect(.poundEndifKeyword)
-    var unexpectedAfterPoundEndif: RawUnexpectedNodesSyntax?
-    if !self.currentToken.isAtStartOfLine {
-      var unexpectedTokens = [RawTokenSyntax]()
-      var loopProgress = LoopProgressCondition()
-      while !self.at(.eof), !currentToken.isAtStartOfLine, loopProgress.evaluate(self.currentToken) {
-        unexpectedTokens += [self.consumeAnyToken()]
-      }
+    let unexpectedAfterPoundEndif = self.consumeRemainingTokenOnLine()
 
-      unexpectedAfterPoundEndif = RawUnexpectedNodesSyntax(unexpectedTokens, arena: self.arena)
-    }
     return RawIfConfigDeclSyntax(
       clauses: RawIfConfigClauseListSyntax(elements: clauses, arena: self.arena),
       unexpectedBeforePoundEndIf,
@@ -267,6 +259,8 @@ extension Parser {
       args = nil
     }
     let (unexpectedBeforeRParen, rparen) = self.expect(.rightParen)
+    let unexpectedAfterRightParen = self.consumeRemainingTokenOnLine()
+
     return RawPoundSourceLocationSyntax(
       poundSourceLocation: line,
       unexpectedBeforeLParen,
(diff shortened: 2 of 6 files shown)


Tests/SwiftParserTest/DirectiveTests.swift
```diff
@@ -184,4 +184,29 @@ final class DirectiveTests: XCTestCase {
...
+        )
+      ],
+      fixedSource: """
+      struct Foo {
+        #if false
+        var x: Int
+      #endif
+      ; var x = 1
+      }
+      """
```
kimdv, line 209: Not sure if we should add this Fix-It.
The current parser doesn't do it. It only add an error

Sources/SwiftParser/Directives.swift
```diff
@@ -163,10 +163,25 @@ extension Parser {
...
+    if !self.currentToken.isAtStartOfLine {
+      var unexpectedTokens = [RawTokenSyntax]()
+      var loopProgress = LoopProgressCondition()
+      while !self.at(.eof), loopProgress.evaluate(self.currentToken) {
+        if !currentToken.isAtStartOfLine {
+          unexpectedTokens += [self.consumeAnyToken()]
+        } else {
+          break
+        }
+      }
```
ahoppen, line 176: Can’t you simplify this?

```suggestion
      while !self.at(.eof), !currentToken.isAtStartOfLine, loopProgress.evaluate(self.currentToken) {
        unexpectedTokens.append(self.consumeAnyToken())
      }
```

Tests/SwiftParserTest/DirectiveTests.swift
ahoppen: I don’t think we should add the Fix-It because I don’t expect anyone to write a statement after `#endif` and expect it to work. The more realistic error case I would imagine is just some bogus left-over code like `#endif s` and inserting a newline isn’t the fix there. I’m not even sure if we need a custom diagnostic for it.

ahoppen (commented):
Could you do the same for `#sourceLocation`?

Sources/SwiftParser/Directives.swift
```diff
@@ -256,14 +259,34 @@ extension Parser {
...
+      var unexpectedTokens = [RawTokenSyntax]()
+      var loopProgress = LoopProgressCondition()
+      while !self.at(.eof), !currentToken.isAtStartOfLine, loopProgress.evaluate(self.currentToken) {
+        unexpectedTokens += [self.consumeAnyToken()]
+      }
+
+      rawUnexpectedNodesSyntax = RawUnexpectedNodesSyntax(unexpectedTokens, arena: self.arena)
+    }
+
+    return rawUnexpectedNodesSyntax
```
ahoppen, line 290: Just an idea: We could remove the nesting level by early-exiting if `self.currentToken.isAtStartOfLine`.

```suggestion
    guard !self.currentToken.isAtStartOfLine else {
      return nil
    }
    var unexpectedTokens = [RawTokenSyntax]()
    var loopProgress = LoopProgressCondition()
    while !self.at(.eof), !currentToken.isAtStartOfLine, loopProgress.evaluate(self.currentToken) {
      unexpectedTokens += [self.consumeAnyToken()]
    }

    return RawUnexpectedNodesSyntax(unexpectedTokens, arena: self.arena)
```

Sources/SwiftBasicFormat/BasicFormat.swift
```diff
@@ -186,12 +186,12 @@ open class BasicFormat: SyntaxRewriter {
     }
 
     switch (first?.tokenKind, second?.tokenKind) {
-
     case (.multilineStringQuote, .backslash),  // string interpolation segment inside a multi-line string literal
       (.multilineStringQuote, .multilineStringQuote),  // empty multi-line string literal
       (.multilineStringQuote, .stringSegment),  // segment starting a multi-line string literal
       (.stringSegment, .multilineStringQuote),  // ending a multi-line string literal that has a string interpolation segment at its end
       (.rightParen, .multilineStringQuote),  // ending a multi-line string literal that has a string interpolation segment at its end
+      (.poundEndifKeyword, _),
```
ahoppen, line 194: Do we also need `.poundSourceLocationKeyword` here?

kimdv: Don't this its needed. 

The reason I added that extra one is because 
```swift
#if hasAttribute(foo)
      @foo
      #endif
      struct MyStruct {}
```

becomes this without 
```swift
#if hasAttribute(foo)
@foo
#endif struct MyStruct {
}
```

ahoppen: Oh, and `#sourceLocation` always has arguments, so this doesn’t apply. Makes sense 👍🏽

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test windows

Sources/SwiftParser/Directives.swift
```diff
@@ -256,14 +259,33 @@ extension Parser {
...
       args: args,
       unexpectedBeforeRParen,
       rightParen: rparen,
+      unexpectedAfterRightParen,
       arena: self.arena
     )
   }
+
+  /// Consumes remaining token on the line and returns a ``RawUnexpectedNodesSyntax``
+  /// if there is any tokens conrumed.
```
ahoppen, line 277: Typo

```suggestion
  /// if there is any tokens consumed.
```

Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift
```diff
@@ -156,6 +156,12 @@ extension DiagnosticMessage where Self == StaticParserError {
   public static var expectedSequenceExpressionInForEachLoop: Self {
     .init("expected Sequence expression for for-each loop")
   }
+  public static var extraTokensAtTheEndOfSourceLocationDirective: Self {
+    .init("extra tokens at the end of #sourceLocation directive")
```
ahoppen, line 160: Should we make this consistent with the `#else` diagnostic?

```suggestion
    .init("extra tokens following the #sourceLocation directive")
```

ahoppen (commented):
Oh, and one thing that just crossed my mind: Could you also add test cases for statements on the same line as `#if`, `#elseif` and `#else`?

Sources/SwiftParser/Directives.swift
```diff
@@ -137,6 +141,7 @@ extension Parser {
         } else {
           condition = nil
         }
+        unexpectedBetweenConditionAndElements = self.consumeRemainingTokenOnLine()
       case .pound:
         if self.atElifTypo() {
```
kimdv, line 146: Is this some legacy in the codebase?

kimdv:
> Oh, and one thing that just crossed my mind: Could you also add test cases for statements on the same line as `#if`, `#elseif` and `#else`?

Good catch! 
Test cases added 🙌

kimdv:
@swift-ci please test

Sources/SwiftParser/Directives.swift
ahoppen: No, this was recently introduced by https://github.com/apple/swift-syntax/pull/1358

ahoppen (approved):
Thanks 👍🏽

PR #1843, opened 2023-06-25
Rename `Spec` syntax nodes to `Argument`

### LINKED ISSUE #1737: Rename `Spec` syntax nodes
I’m not a huge fan of the `Spec` part in `AvailabilitySpecList`. We should rename it to `AvailabilityArgumentListSyntax`. Similarly, we should rename all other nodes that contain `Spec`: 

- `AvailabilitySpecList`
- `BackDeployedAttributeSpecList`
- `SpecializeAttributeSpecList`

TTOzzi:
Resolve #1737


CodeGeneration/Sources/SyntaxSupport/AttributeNodes.swift
```diff
@@ -70,12 +70,12 @@ public let ATTRIBUTE_NODES: [Node] = [
             kind: .node(kind: .stringLiteralExpr)
           ),
           Child(
-            name: "Availability",
-            kind: .node(kind: .availabilitySpecList)
+            name: "AvailabilityArguments",
```
TTOzzi, line 73: I have renamed it to maintain consistency with other cases.

Sources/SwiftParser/Availability.swift
```diff
@@ -95,7 +95,7 @@ extension Parser {
     }
   }
 
-  mutating func parseAvailabilityArgumentSpecList() -> RawAvailabilitySpecListSyntax {
+  mutating func parseAvailabilityArguments() -> RawAvailabilityArgumentListSyntax {
```
TTOzzi, line 98: The functions `parseAvailabilitySpecList` and `parseAvailabilityArgumentSpecList` perform similar operations, but there are subtle differences. I tried to rename them to something other than `Spec`, but I couldn't think of a suitable name. Is there a better approach? 🤔

CodeGeneration/Sources/SyntaxSupport/AttributeNodes.swift
kimdv: We recently introduced `deprecatedName` on `Child`. 
This would generate the needed variables, initialisers etc.

Can you try to use that re-run the code generator?

TTOzzi: I tried using the `deprecatedName` like this, but there was no change. 
```swift
  Child(
    name: "ArgumentList",
    deprecatedName: "Availability",
    kind: .node(kind: .tupleExprElementList)
  )
```
The code generated from this section is `nodeChoices` (enum with associated values), so I thought it wasn't working 🤔 
Could there be something I missed?

kimdv: Have you committed the changes? 
Because it uses the HEAD of the branch. 

You can run swift package resolve to get the latest.

ahoppen: You might also need to try removing `CodeGeneration/Package.resolved` and `CodeGeneration/.build`. We are having some issues with the `HEAD` reference at the moment 😢

TTOzzi: I tried both
- committing the changes and resolving the package
- deleting `CodeGeneration/Package.resolved` and `CodeGeneration/.build`

but it still doesn't work 😭 
The strange thing is that modifying the `deprecatedName` in other parts, not this section, works fine. (e.g. deprecatedName at line 176)
Is there any way to debug this?

ahoppen (commented):
Can we hold this PR for a week or so? I’ve got a lot of renames in terms of consistency lined up locally and depending on how their reviews go, it might influence if I change my mind again how syntax collections should be named and I want to avoid renaming these types twice within a couple of weeks or so.

CodeGeneration/Sources/SyntaxSupport/AttributeNodes.swift
TTOzzi: Oh, I see! Thank you all for your help! 😊
May I ask one more question?
If I were to write a compatibility layer for an Enum case with associated values like this, how could I do it? 
```swift
    case `availability`(TupleExprElementListSyntax)
```
I know that I can write a `static var` for cases without associated values, but I'm unsure about cases with associated values as I couldn't find any examples.

TTOzzi: Aha, there is no perfect way to maintain compatibility with switch-case. Thank you for letting me know 🙇‍♂️

ahoppen:
@TTOzzi I did these renames during my huge rename walkthrough, so I’m closing this PR.

TTOzzi:
Ah! I see. Thanks for comment 🙂

PR #1835, opened 2023-06-23
Run ValidateSyntaxNodes tests in CI

ahoppen:
Also update ValidateSyntaxNodes to match current state of syntax node names so that tests pass.

I’ve got a local branch that fixes most of the naming inconsistencies found by these tests. I will put up PRs to fix those inconsistencies over the next couple of days.

rdar://110990221

PR #1844, opened 2023-06-25
Update the list(test cases) of API inconsistencies that need to be modified in CodeGeneration

TTOzzi:
These test cases, except for special cases, seem to be a list of issues to address for future API compatibility 😃
I have updated it by removing resolved issues and adding newly created ones.


kimdv:
I think it's a duplicate of https://github.com/apple/swift-syntax/pull/1835

TTOzzi:
Oh, I see.
I was curious why we didn't perform tests in the CI, but it turns out there was already a PR in progress!
Thank you for letting me know 🙇‍♂️

PR #1846, opened 2023-06-26
[5.9] When creation of a syntax node using string interpolation failed, log the diagnostics

ahoppen:
* **Explanation**: A very common mistake that clients of SwiftSyntax make, is to construct incorrect syntax nodes using string interpolation. E.g. creating an `ExprSyntax` node from source code that represents a declaration. Since construction of a type using string interpolation cannot throw an error and we don’t want to cause an assertion failure for invalid source code (since we don’t want to crash macros compiled using debug mode in case there is a syntax error), log a fault using `OSLog` on macOS informing the user about the API misuse. This log will show in Xcode’s console.
* **Risk**: Low, only adds logging and the logging is not present when building swift-syntax for the compiler.
* **Testing**: Manually tested that the log messages show up in Xcode’s console when running macro tests.
* **Issue**: n/a
* **Reviewer**:  @bnbarham and @rintaro on https://github.com/apple/swift-syntax/pull/1790


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #1849, opened 2023-06-26
Generate doc-comment for syntax node initializers

### LINKED ISSUE #1598: Generate doc-comment for syntax node initializers
For each initializer of a Syntax node, we should be generating a doc comment that documents each parameter with the first line of that parameter’s description.

Matejkob:
Resolves #1598. As discussed in #1598: 
* if the child's documentation is `nil` parameter is skipped, 
* only the first line of the child's documentation is shown,
* to all nodes `leadingTrivia `and `trailingTrivia` documentation (with static text) is added.

### DIFF
--- a/CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodesFile.swift
+++ b/CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodesFile.swift
@@ -78,7 +78,12 @@ func syntaxNode(emitKind: SyntaxNodeKind) -> SourceFileSyntax {
           """
         )
 
-        try! InitializerDeclSyntax("\(node.generateInitializerDeclHeader())") {
+        try! InitializerDeclSyntax(
+          """
+          \(raw: node.generateInitializerDocComment())
+          \(node.generateInitializerDeclHeader())
+          """
+        ) {
           let parameters = ClosureParameterListSyntax {
             for child in node.children {
               ClosureParameterSyntax(firstName: .identifier(child.varName.backtickedIfNeeded))
--- a/CodeGeneration/Sources/SyntaxSupport/Utils.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/Utils.swift
@@ -40,7 +40,7 @@ public func lowercaseFirstWord(name: String) -> String {
 
 /// Give a (possibly multi-line) string, prepends `///` to each line and creates
 /// a `.docLineComment` trivia piece for each line.
-func docCommentTrivia(from string: String?) -> SwiftSyntax.Trivia {
+public func docCommentTrivia(from string: String?) -> SwiftSyntax.Trivia {
   guard let string else {
     return []
   }
--- a/CodeGeneration/Sources/Utils/Utils.swift
+++ b/CodeGeneration/Sources/Utils/Utils.swift
@@ -29,6 +29,13 @@ public func flattened(indentedDocumentation: String) -> String {
     .trimmingCharacters(in: .whitespacesAndNewlines)
 }
 
+/// Removes all empty lines from a multi-line string.
+public func removedEmptyLines(string: String) -> String {
+  string.split(whereSeparator: \.isNewline)
+    .filter { !$0.allSatisfy(\.isWhitespace) }
+    .joined(separator: "\n")
+}
+
 public extension Collection {
   /// If the collection contains a single element, return it, otherwise `nil`.
   var only: Element? {
--- a/CodeGeneration/Sources/generate-swiftsyntax/LayoutNode+Extensions.swift
+++ b/CodeGeneration/Sources/generate-swiftsyntax/LayoutNode+Extensions.swift
@@ -13,6 +13,7 @@
 import SwiftSyntax
 import SwiftSyntaxBuilder
 import SyntaxSupport
+import Utils
 
 extension LayoutNode {
   func generateInitializerDeclHeader(useDeprecatedChildName: Bool = false) -> PartialSyntaxNodeString {
@@ -72,6 +73,31 @@ extension LayoutNode {
       \(params)
       )
       """
+  }
+
+  func generateInitializerDocComment() -> SwiftSyntax.Trivia {
+    func generateParamDocComment(for child: Child) -> String? {
+      guard let documentation = child.documentation,
+        let firstLine = documentation.split(whereSeparator: \.isNewline).first
+      else {
+        return nil
+      }
+
+      return "  - \(child.varName): \(firstLine)"
+    }
+
+    let formattedParams = removedEmptyLines(
+      string: """
+        - Parameters:
+          - leadingTrivia: Trivia to be prepended to the leading trivia of the node’s first token. \
+        If the node is empty, there is no token to attach the trivia to and the parameter is ignored.
+        \(children.compactMap(generateParamDocComment).joined(separator: "\n"))
+          - trailingTrivia: Trivia to be appended to the trailing trivia of the node’s last token. \
+        If the node is empty, there is no token to attach the trivia to and the parameter is ignored.
+        """
+    )
+
+    return docCommentTrivia(from: formattedParams)
   }
 
   /// Create a builder-based convenience initializer, if needed.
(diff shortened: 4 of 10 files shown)


Matejkob:
@swift-ci Please test

kimdv:
@swift-ci please test

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodesFile.swift
```diff
@@ -304,3 +309,27 @@ private func generateSyntaxChildChoices(for child: Child) throws -> EnumDeclSynt
...
+      " - Parameters:",
+      "   - leadingTrivia: The leading ``Trivia`` for this node.",
+    ]
+      + node.children.compactMap(generateParamDocComment)
+      + ["   - trailingTrivia: The trailing ``Trivia`` for this node."])
+      .joined(separator: "\n")
+  )
+
+  return paramsDoc
+}
```
ahoppen, line 335: My recent tendency has been to implement these kinds of things in the SyntaxSupport module like I did in https://github.com/apple/swift-syntax/pull/1771. There’s no real strict reason for it, I just thought that it made more sense.

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodesFile.swift
```diff
@@ -304,3 +309,27 @@ private func generateSyntaxChildChoices(for child: Child) throws -> EnumDeclSynt
...
+
+  let paramsDoc = docCommentTrivia(
+    from: ([
+      " - Parameters:",
+      "   - leadingTrivia: The leading ``Trivia`` for this node.",
+    ]
+      + node.children.compactMap(generateParamDocComment)
+      + ["   - trailingTrivia: The trailing ``Trivia`` for this node."])
+      .joined(separator: "\n")
+  )
```
ahoppen, line 332: I’m not sure what you think but maybe using a multi-line string literal reads nicer here instead of the array.

```suggestion
  let paramsDoc = docCommentTrivia(
    from: """
       - Parameters:
         - leadingTrivia: The leading ``Trivia`` for this node.
      \(node.children.compactMap(generateParamDocComment).joined(separator: "\n"))
         - trailingTrivia: The trailing ``Trivia`` for this node.
       """
  )
```

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodesFile.swift
```diff
@@ -304,3 +309,27 @@ private func generateSyntaxChildChoices(for child: Child) throws -> EnumDeclSynt
...
+      return nil
+    }
+
+    return "   - \(child.varName): \(firstLine)"
+  }
+
+  let paramsDoc = docCommentTrivia(
+    from: ([
+      " - Parameters:",
+      "   - leadingTrivia: The leading ``Trivia`` for this node.",
```
ahoppen, line 327: The semantics of `leadingTrivia` and `trailingTrivia` are a little subtle and I’m happy that we’re adding a doc comment for it now. Could you change the comment to

```
Trivia to be prepended to the leading trivia of the node’s first token. If the node is empty, there is no token to attach the trivia to and the parameter is ignored.
```

Similarly, the documentation for `trailingTrivia` should be 

```
Trivia to be appended to the trailing trivia of the node’s last token. If the node is empty, there is no token to attach the trivia to and the parameter is ignored.
```

ahoppen (commented):
Nice. Thank you 👍🏽 I’ve got a few minor comments inline but other than that, this looks really good to me.

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodesFile.swift
Matejkob: Oh, yeah, I've seen it. I think it's a really good approach, but I wonder if it's a good fit for this case. 

I've been thinking about it and I believe that if we would implement it within the `SyntaxSupport` module, like `GrammarGenerator` it would be reasonable to make it reusable. This specific function's implementation is highly contextual and assumes several conditions such as the manual addition of `leadingTrivia` and `trailingTrivia` in the `LayoutNode.generateInitializerDeclHeader(useDeprecatedChildName:)` method - it implies that there is always more than one parameter so we don't need to handle cases for a single parameter (instead of ` - Parameters: (...)` in this case, we should generate ` - Parameter {parameter name}: {parameter documentation}`) or no parameters. We also assume that the `LayoutNode` doesn't know about `leadingTrivia` and `trailingTrivia`. Given these assumptions, I don't think it would be a good idea to just move this function to `SyntaxSupport`. It might be more beneficial to create something more generic, like `ParamsDocCommentGenerator`, which could handle all cases and be useful in other contexts. To address the issue of `LayoutNode` not knowing about `leadingTrivia` and `trailingTrivia`, we might need to pass all parameters to `ParamsDocCommentGenerator` in an abstract way, using a tuple or something. 

If we agree this is the right course of action, I could develop something like the `ParamsDocCommentGenerator`. I'd appreciate your thoughts on this matter! 💙

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodesFile.swift
Matejkob: I love it! It's much more readable! But there is one issue with this code :/ If `node.children.compactMap(generateParamDocComment)` would return an empty array we would end with a comment like this:
```swift
///  - Parameters:
///    - leadingTrivia:  (...)
/// 
///    - trailingTrivia: (...)
```
the simplest way, I can think of, to address this issue in this approach is to add:
```swift
.split(whereSeparator: \.isNewline)
.filter { !$0.allSatisfy(\.isWhitespace) }
.joined(separator: "\n")
``` 
to this multi-line string. Or extract this code to `Utils` and make it reusable.

ahoppen: Ah, I didn’t think about that. Extracting that logic to a function `removingEmptyLines` and just calling it is probably the easiest option

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodesFile.swift
ahoppen: Oh, good point with `leadingTrivia` and `trailingTrivia`. I didn’t think about that. In that case I agree with you that this should be in the SyntaxNodesFile.swift file. Just one minor suggestion: I would make this a fileprivate extension on `LayoutNode` instead of a top-level function. It just reads nicer IMO.

Matejkob: Sounds good. So maybe if we wanna convert this function into a method in an extension of `LayoutNode` we could mark it as internal and move it to the `LayoutNode+Extensions.swift` file. Method to generate init header - `LayoutNode.generateInitializerDeclHeader(useDeprecatedChildName:)` is already in this file.

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/SyntaxNodesFile.swift
Matejkob: I've found `Utils.swift` file in `Utils` module and it already contains very similar functions like: `dedented(string:)`, `flattened(indentedDocumentation:)`. Maybe I should stick to the pattern of these functions and decelerate the new one in this way:
```swift
/// Removes all empty lines from a multi-line string.
public func removedEmptyLines(string: String) -> String {
  string.split(whereSeparator: \.isNewline)
    .filter { !$0.allSatisfy(\.isWhitespace) }
    .joined(separator: "\n")
}
```

ahoppen: That sounds like a good idea to me.

ahoppen (approved):
Thank you @Matejkob. This looks great. 👍🏽 I’ll trigger CI and once that passed, I’ll merge the PR.

ahoppen:
@swift-ci Please test

PR #1850, opened 2023-06-26
Support for https://github.com/apple/swift/pull/66540: @_silgen_name on globals, @_silgen_name(raw: ...)

kubamracek:


ahoppen (commented):
`@_silgen_name` can only be applied to types, right? If no disambiguation is necessary to decide whether the parentheses in eg. `@_silgen_name () -> Void` are attribute arguments or part of the closure type, I think you should be able to just remove all special handling for `@_silgen_name` and we’ll parse it like a user-defined attribute. ASTGen (which is still to be written) will then extract the arguments from the `AttributeSyntax`.

I.e. I think you can just delete all the code that you wrote in this PR and additionally remove `_silgen_name` from Keyword.swift because it will no longer used.

kubamracek:
> @_silgen_name can only be applied to types, right? 
No, it's the opposite, it only applies to function and (newly) global variable declarations. It never goes on a type.

> I think you can just delete all the code that you wrote in this PR and additionally remove _silgen_name from Keyword.swift because it will no longer used.

Let me try that!

kubamracek:
That works, thanks!

kubamracek:
@swift-ci please test

kubamracek:

ahoppen:
> No, it's the opposite, it only applies to function and (newly) global variable declarations. It never goes on a type.

Oh, sorry, in my original comment I meant to say *cannot* instead of *can*.

Issue #1847, opened 2023-06-26
#1759 cases 30% slowdown when generating diagnostics


ahoppen:
When changing `swift-parser-cli performance-test` to also fold operators and generate diagnostics (like the SourceKit stress tester does), then https://github.com/apple/swift-syntax/pull/1759 causes a 30% slowdown.

# Steps to Reproduce

1. Apply the following diff to `SwiftSyntax`
```diff
diff --git a/Sources/swift-parser-cli/Commands/PerformanceTest.swift b/Sources/swift-parser-cli/Commands/PerformanceTest.swift
index 5b3c0afa..844fdaaa 100644
--- a/Sources/swift-parser-cli/Commands/PerformanceTest.swift
+++ b/Sources/swift-parser-cli/Commands/PerformanceTest.swift
@@ -14,6 +14,9 @@ import _InstructionCounter
 import ArgumentParser
 import Foundation
 import SwiftParser
+import SwiftOperators
+import SwiftSyntax
+import SwiftParserDiagnostics
 
 struct PerformanceTest: ParsableCommand {
   static var configuration = CommandConfiguration(
@@ -42,7 +45,11 @@ struct PerformanceTest: ParsableCommand {
     for _ in 0..<self.iterations {
       for file in files {
         file.withUnsafeBytes { buf in
-          _ = Parser.parse(source: buf.bindMemory(to: UInt8.self))
+          var tree = Parser.parse(source: buf.bindMemory(to: UInt8.self))
+          if let foldedTree = OperatorTable.standardOperators.foldAll(tree, errorHandler: { _ in }).as(SourceFileSyntax.self) {
+            tree = foldedTree
+          }
+          _ = ParseDiagnosticsGenerator.diagnostics(for: tree)
         }
       }
     }
```
2. Save the following file to `/tmp/i/input.swift`
```
//  Copyright (c) 2015 Rob Rix. All rights reserved.

#if swift(>=4.2)
#if compiler(>=5)

// Use Swift.Result
extension Result {
	// ResultProtocol
	public

extension Result {
	/// The compatibility alias for the Swift 5's `Result` in the standard library.
	///
	/// See https://github.com/apple/swift-evolution/blob/master/proposals/0235-add-result.md
	/// and https://forums.swift.org/t/accepted-with-modifications-se-0235-add-result-to-the-standard-library/18603
	/// for the details.
	public typealias Success = Value
	/// The compatibility alias for the Swift 5's `Result` in the standard library.
	///
[...]

PR #1851, opened 2023-06-26
Allow rewrite to detach from the node's parent

bnbarham:
`MissingMaker` and `MissingNodesBasicFormatter` do not require the returned node to have a parent. Avoid re-creating all the parent nodes in this case by detaching the rewritten node (or rather, not re-attaching it).

Resolves rdar://111332992 (#1847).


bnbarham:
@swift-ci please test

bnbarham:
Ah, we can't detach because we're formatting... That's unfortunate 🤔

bnbarham:
@swift-ci please test

rintaro (approved):

bnbarham:
@swift-ci please test

PR #1855, opened 2023-06-27
Fix macro expansion order in compiler_plugin_basic.swift

ahoppen:
It looks like the extension expansion is recorded last now, it used to be the first reported extension.


ahoppen:
@swift-ci Please test

PR #1859, opened 2023-06-28
[Macros] Add attached `extension` macros.

hborla:
The SwiftSyntax-side change to https://github.com/apple/swift/pull/66967

### DIFF
--- a/Sources/SwiftCompilerPluginMessageHandling/PluginMessages.swift
+++ b/Sources/SwiftCompilerPluginMessageHandling/PluginMessages.swift
@@ -77,7 +77,7 @@ internal enum PluginToHostMessage: Codable {
 }
 
 /*namespace*/ internal enum PluginMessage {
-  static var PROTOCOL_VERSION_NUMBER: Int { 5 }  // Added 'expandMacroResult'.
+  static var PROTOCOL_VERSION_NUMBER: Int { 6 }  // Added 'expandMacroResult'.
 
   struct HostCapability: Codable {
     var protocolVersion: Int
--- a/Sources/SwiftCompilerPluginMessageHandling/PluginMessages.swift
+++ b/Sources/SwiftCompilerPluginMessageHandling/PluginMessages.swift
@@ -34,7 +34,8 @@ internal enum HostToPluginMessage: Codable {
     discriminator: String,
     attributeSyntax: PluginMessage.Syntax,
     declSyntax: PluginMessage.Syntax,
-    parentDeclSyntax: PluginMessage.Syntax?
+    parentDeclSyntax: PluginMessage.Syntax?,
+    extendedTypeSyntax: PluginMessage.Syntax?
   )
 
   /// Optionally implemented message to load a dynamic link library.
--- a/Sources/SwiftCompilerPluginMessageHandling/PluginMessages.swift
+++ b/Sources/SwiftCompilerPluginMessageHandling/PluginMessages.swift
@@ -108,6 +108,7 @@ internal enum PluginToHostMessage: Codable {
     case peer
     case conformance
     case codeItem
+    case `extension`
   }
 
   struct SourceLocation: Codable {
--- a/Sources/SwiftParser/Attributes.swift
+++ b/Sources/SwiftParser/Attributes.swift
@@ -332,11 +332,41 @@ extension Parser {
         )
       )
     case nil:
+      let isAttached = self.peek().isAttachedKeyword
       return parseAttribute(argumentMode: .customAttribute) { parser in
-        let arguments = parser.parseArgumentListElements(pattern: .none)
+        let arguments: [RawTupleExprElementSyntax]
+        if isAttached {
+          arguments = parser.parseAttachedArguments()
+        } else {
+          arguments = parser.parseArgumentListElements(pattern: .none)
+        }
+
         return .argumentList(RawTupleExprElementListSyntax(elements: arguments, arena: parser.arena))
       }
     }
+  }
+}
+
+extension Parser {
+  mutating func parseAttachedArguments() -> [RawTupleExprElementSyntax] {
+    let (unexpectedBeforeRole, role) = self.expect(.identifier, TokenSpec(.extension, remapping: .identifier), default: .identifier)
+    let roleTrailingComma = self.consume(if: .comma)
+    let roleElement = RawTupleExprElementSyntax(
+      label: nil,
+      colon: nil,
+      expression: RawExprSyntax(
+        RawIdentifierExprSyntax(
+          unexpectedBeforeRole,
+          identifier: role,
+          declNameArguments: nil,
+          arena: self.arena
+        )
+      ),
+      trailingComma: roleTrailingComma,
+      arena: self.arena
+    )
+    let additionalArgs = self.parseArgumentListElements(pattern: .none)
+    return [roleElement] + additionalArgs
   }
 }
 
--- a/Sources/SwiftSyntaxMacroExpansion/MacroExpansion.swift
+++ b/Sources/SwiftSyntaxMacroExpansion/MacroExpansion.swift
@@ -47,6 +47,7 @@ private enum MacroExpansionError: Error, CustomStringConvertible {
   case parentDeclGroupNil
   case declarationNotDeclGroup
   case declarationNotIdentified
+  case noExtendedTypeSyntax
   case noFreestandingMacroRoles(Macro.Type)
 
   var description: String {
@@ -62,6 +63,9 @@ private enum MacroExpansionError: Error, CustomStringConvertible {
 
     case .declarationNotIdentified:
       return "declaration is not a 'Identified' syntax"
+
+    case .noExtendedTypeSyntax:
+      return "no extended type for extension macro"
 
     case .noFreestandingMacroRoles(let type):
       return "macro implementation type '\(type)' does not conform to any freestanding macro protocol"
@@ -180,6 +184,7 @@ public func expandAttachedMacroWithoutCollapsing<Context: MacroExpansionContext>
   attributeNode: AttributeSyntax,
   declarationNode: DeclSyntax,
   parentDeclNode: DeclSyntax?,
+  extendedType: TypeSyntax?,
   in context: Context
 ) -> [String]? {
   do {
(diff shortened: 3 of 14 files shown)


hborla:
https://github.com/apple/swift/pull/66967

@swift-ci please test

hborla:
https://github.com/apple/swift/pull/66967

@swift-ci please test

Sources/SwiftCompilerPluginMessageHandling/PluginMessages.swift
```diff
@@ -34,7 +34,8 @@ internal enum HostToPluginMessage: Codable {
     discriminator: String,
     attributeSyntax: PluginMessage.Syntax,
     declSyntax: PluginMessage.Syntax,
-    parentDeclSyntax: PluginMessage.Syntax?
+    parentDeclSyntax: PluginMessage.Syntax?,
+    extendedTypeSyntax: PluginMessage.Syntax?
```
rintaro, line 38: Please increment the protocol version (line 80)

Sources/SwiftSyntaxMacroExpansion/MacroExpansion.swift
```diff
@@ -295,6 +302,39 @@ public func expandAttachedMacroWithoutCollapsing<Context: MacroExpansionContext>
...
+      ) throws -> [ExtensionDeclSyntax] {
+        return try attachedMacro.expansion(
+          of: attributeNode,
+          attachedTo: node,
+          providingExtensionsOf: extendedType,
+          in: context
+        )
+      }
+
+      let extensions = try _openExistential(
```
DougGregor, line 328: I suspect we don't need to do `_openExistential` tricks nowadays, but it's fine this way

DougGregor (approved):

Sources/SwiftSyntaxMacros/MacroProtocols/ExtensionMacro.swift
```diff
@@ -0,0 +1,34 @@
...
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
+/// Describes a macro that can add conformances to the declaration it's
+/// attached to.
+public protocol ExtensionMacro: AttachedMacro {
```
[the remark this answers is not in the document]
hborla: Oh! That is a really good idea

Tests/SwiftParserTest/AttributeTests.swift
```diff
@@ -629,4 +629,20 @@ final class AttributeTests: XCTestCase {
...
+      """
+    )
+
+    assertParse(
+      """
+      @attached(extension, names: named(test))
+      macro m()
+      """
+    )
+  }
```
bnbarham, line 647: `conformances:` as well? Shouldn't really matter since it's just parsed as a custom attribute anyway, but good to have the test case IMO.

Sources/SwiftSyntaxMacros/MacroSystem.swift
```diff
@@ -438,6 +439,23 @@ extension MacroApplication {
...
+          providingExtensionsOf: extendedTypeSyntax,
+          in: context
+        )
+
+        extensions.append(contentsOf: newExtensions.map(DeclSyntax.init))
+      } catch {
+        context.addDiagnostics(from: error, node: attribute)
+      }
+    }
+
```
Matejkob, line 458: I'm not entirely sure what the assumptions are about how we're structuring code in this project, but I wonder whether it might be better to extract this code into a separate method. The name and doc-comment of this method suggest that the responsibility of this method is solely to expand conformances - I guess expanding extensions could be seen as a bit different responsibility. Just a thought! 💜

Sources/SwiftSyntaxMacros/MacroSystem.swift
```diff
@@ -438,6 +439,23 @@ extension MacroApplication {
       }
     }
 
+    let extensionMacroAttrs = getMacroAttributes(attachedTo: decl.as(DeclSyntax.self)!, ofType: ExtensionMacro.Type.self)
+    let extendedTypeSyntax = TypeSyntax("\(extendedType.trimmed)")
```
Matejkob, line 443: In the [proposal](https://github.com/hborla/swift-evolution/blob/extension-macros/proposals/NNNN-extension-macros.md#implementing-extension-macros) it's mentioned: 

> Each `ExtensionDeclSyntax` in the resulting array must use the `providingExtensionsOf` parameter as the extended > type, which is a qualified type name. For example, for the following code:
> 
> ```swift
> struct Outer {
>   @MyProtocol
>   struct Inner {}
> }
> ```
> The type syntax passed to `ExtensionMacro.expansion` for `providingExtensionsOf` is `Outer.Inner`.

I'm not sure if I've understood this part of the proposal correctly. Does it mean that if I create this macro:
```swift
public struct SomeExtensionMacro: ExtensionMacro {
  public static func expansion(
    of node: AttributeSyntax,
    attachedTo declaration: some DeclGroupSyntax,
    providingExtensionsOf type: some TypeSyntaxProtocol,
  ) throws -> [ExtensionDeclSyntax] {
    [ExtensionDeclSyntax(extendedType: type, memberBlock: MemberDeclBlockSyntax {})]
  }
}
```
and I attach it like so
```swift
struct Outer {
  @SomeExtensionMacro
  struct Inner {}
}
```
are we expecting that `extendedType.description` here will look something like this: `Outer.Inner`? If I understand the assumptions correctly, I think this would happen because the code above:
```swift
let extendedType: Syntax
if let identified = decl.asProtocol(IdentifiedDeclSyntax.self) {
  extendedType = Syntax(identified.identifier.trimmed)
} else if let ext = decl.as(ExtensionDeclSyntax.self) {
  extendedType = Syntax(ext.extendedType.trimmed)
} else {
  return []
}
```
can only extract the identifier from the root type.

Here is a test case covering this example using `@AddSendableExtension` macro:
```swift
func testExtensionExpansionOfNestedType() {
  assertMacroExpansion(
    """
    struct Outer {
      @AddSendableExtension
      struct Inner {
      }
    }
    """,
    expandedSource:
      """

      struct Outer {
        struct Inner {
        }
      }
      extension Outer.Inner: Sendable {
      }
      """,
    macros: testMacros,
  )
}
```
right now it's failing with message: 
```
failed - Actual output (+) differed from expected output (-):
 struct Outer {
   struct Inner {
   }
 }
–extension Outer.Inner: Sendable {
–}

Actual expanded source:

struct Outer {
  struct Inner {
  }
}
extension Outer.Inner: Sendable {
}
```
Am I understanding it correctly that this shouldn't fail?

Matejkob (commented):
Hi there 👋,

As someone pretty fresh to this project, I've had a good look at this PR and wow - it's really cool! You've made some fantastic improvements here.

I've attempted to provide constructive feedback on some areas that I noticed while reviewing. Please bear in mind, my understanding of this project's conventions and code structure is still growing, so if any of my comments are off the mark or unrelated, I sincerely apologize. I deeply appreciate your understanding and patience.

Can't wait to get more involved with this project and learn from all of you. 💙

Sources/SwiftSyntaxMacroExpansion/MacroExpansion.swift
```diff
@@ -178,6 +184,7 @@ public func expandAttachedMacroWithoutCollapsing<Context: MacroExpansionContext>
   attributeNode: AttributeSyntax,
   declarationNode: DeclSyntax,
   parentDeclNode: DeclSyntax?,
+  extendedType: TypeSyntax?,
```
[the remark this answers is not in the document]
hborla: I actually had the same thought while I was implementing this! But @rintaro pointed out to me that we might also need to make the `extendedType` an associated value of `PluginMessage.MacroRole` which would complicate the `Codable` conformance, so it seemed more straightforward to pass `extendedType` as an argument to `expandAttachedMacro`. I'm still interested in exploring this approach further separately though, because it seems plausible that we might want more role-specific arguments, and those conceptually make sense as enum associated values.

Sources/SwiftSyntaxMacros/MacroProtocols/ExtensionMacro.swift
```diff
@@ -0,0 +1,34 @@
...
+  /// - Returns: the set of `(type, where-clause?)` pairs that each provide the
+  ///   protocol type to which the declared type conforms, along with
+  ///   an optional where clause.
+  static func expansion(
+    of node: AttributeSyntax,
+    attachedTo decl: some DeclGroupSyntax,
+    providingExtensionsOf type: some TypeSyntaxProtocol,
+    in context: some MacroExpansionContext
+  ) throws -> [ExtensionDeclSyntax]
+}
```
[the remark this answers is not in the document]
hborla: Fixed, thank you!

hborla:
https://github.com/apple/swift/pull/66967

@swift-ci please test

Sources/SwiftSyntaxMacros/MacroSystem.swift
hborla: This is a great question, and there's a little bit of nuance here because the expansion performed by the `MacroSystem` in SwiftSyntax is only used for unit testing macros, but the real expansion is integrated into the compiler (and implemented in https://github.com/apple/swift/pull/66967). It's not possible for `MacroSystem` to form the same fully qualified type that the compiler can form in the general case, e.g. the compiler can look through type aliases, while the macro system here cannot. I believe there are also some known issues with where the expansion is inserted in `MacroSystem` for conformance macros, because it doesn't insert them at the top-level like the compiler does. I definitely think we can improve what `MacroSystem` does in many cases (e.g. we could form a qualified name from the lexical nesting structure), but it can't perfectly match what the compiler does during extension macro expansion.

Sources/SwiftSyntaxMacros/MacroSystem.swift
hborla: Conformance macros are effectively a special case of extension macros. I'm going to take Doug's suggestion [here](https://github.com/apple/swift-syntax/pull/1859#discussion_r1245570504) in a follow-up PR and make `ConformanceMacro` refine `ExtensionMacro` so that they can be implemented in terms of extension macros. As part of that, I will clean up this code to only operate on `ExtensionMacro`, since `ConformanceMacro` will be a refinement of it!

hborla:
https://github.com/apple/swift/pull/66967

@swift-ci please test

hborla:
https://github.com/apple/swift/pull/66967

@swift-ci please test

hborla:
https://github.com/apple/swift/pull/66967

@swift-ci please test

hborla:
https://github.com/apple/swift/pull/66967

@swift-ci please test Windows

Sources/SwiftParser/Attributes.swift
```diff
@@ -332,14 +332,44 @@ extension Parser {
...
+      let isAttached = self.peek().isAttachedKeyword
       return parseAttribute(argumentMode: .customAttribute) { parser in
-        let arguments = parser.parseArgumentListElements(pattern: .none)
+        let arguments: [RawTupleExprElementSyntax]
+        if isAttached {
+          arguments = parser.parseAttachedArguments()
+        } else {
+          arguments = parser.parseArgumentListElements(pattern: .none)
+        }
+
```
ahoppen, line 343: Instead of matching the string contents of the current token here, it’s much clearer to add `attached` as a contextual keyword. I’m fixing that in https://github.com/apple/swift-syntax/pull/1881.

hborla: Thank you!

PR #1854, opened 2023-06-27
Fix child name validation failures in `testSingleTokenChoiceChildNaming`

ahoppen:
Fixes the naming validation found by the `testSingleTokenChoiceChildNaming` test, which is describes as follows:

Checks standardized naming of children with a single token choice

- If a child has a single keyword as its choice, it should be named `*Keyword` (e.g. `ImportKeyword`)
- If it’s another token kind, name it the same as the the token choice (e.g. `LeftParen`)

I will be opening more PRs like to fix naming inconsistencies over in the next couple of weeks.

PR #1860, opened 2023-06-28
Fix warnings & failing tests

kimdv:
Warnings and tests was introduced in #1854 from what I can see


kimdv:
@swift-ci please test

PR #1863, opened 2023-06-28
Add performance measurement test with instructions count

kimdv:
As we talked about, it would be nice to have performance tests as part of the CI. 

I did dig a bit into [`XCTCPUMetric`](https://developer.apple.com/documentation/xctest/xctcpumetric). 
But I didn't find a nice way to add baselines. 

So I tried to make my own measure method so we easily could start adding them in the code 

- [ ] Run CI 3 times

### DIFF
--- a/Tests/PerformanceTest/SyntaxClassifierPerformanceTests.swift
+++ b/Tests/PerformanceTest/SyntaxClassifierPerformanceTests.swift
@@ -26,17 +26,14 @@ public class SyntaxClassifierPerformanceTests: XCTestCase {
 
   func testClassifierPerformance() throws {
     try XCTSkipIf(ProcessInfo.processInfo.environment["SKIP_LONG_TESTS"] == "1")
-    XCTAssertNoThrow(
-      try {
-        let source = try String(contentsOf: inputFile)
-        let parsed = Parser.parse(source: source)
 
-        measure {
-          for _ in 0..<10 {
-            for _ in parsed.classifications {}
-          }
-        }
-      }()
-    )
+    let source = try String(contentsOf: inputFile)
+    let parsed = Parser.parse(source: source)
+
+    try measureInstructions {
+      for _ in 0..<10 {
+        for _ in parsed.classifications {}
+      }
+    }
   }
 }
--- a/Tests/PerformanceTest/VisitorPerformanceTests.swift
+++ b/Tests/PerformanceTest/VisitorPerformanceTests.swift
@@ -27,50 +27,38 @@ public class VisitorPerformanceTests: XCTestCase {
     try XCTSkipIf(ProcessInfo.processInfo.environment["SKIP_LONG_TESTS"] == "1")
     class EmptyVisitor: SyntaxVisitor {}
 
-    XCTAssertNoThrow(
-      try {
-        let parsed = Parser.parse(source: try String(contentsOf: inputFile))
+    let source = try String(contentsOf: inputFile)
+    let parsed = Parser.parse(source: source)
+    let emptyVisitor = EmptyVisitor(viewMode: .sourceAccurate)
 
-        let emptyVisitor = EmptyVisitor(viewMode: .sourceAccurate)
-
-        measure {
-          emptyVisitor.walk(parsed)
-        }
-      }()
-    )
+    try measureInstructions {
+      emptyVisitor.walk(parsed)
+    }
   }
 
   func testEmptyRewriterPerformance() throws {
     try XCTSkipIf(ProcessInfo.processInfo.environment["SKIP_LONG_TESTS"] == "1")
     class EmptyRewriter: SyntaxRewriter {}
 
-    XCTAssertNoThrow(
-      try {
-        let parsed = Parser.parse(source: try String(contentsOf: inputFile))
+    let source = try String(contentsOf: inputFile)
+    let parsed = Parser.parse(source: source)
+    let emptyRewriter = EmptyRewriter(viewMode: .sourceAccurate)
 
-        let emptyRewriter = EmptyRewriter(viewMode: .sourceAccurate)
-
-        measure {
-          _ = emptyRewriter.visit(parsed)
-        }
-      }()
-    )
+    try measureInstructions {
+      _ = emptyRewriter.visit(parsed)
+    }
   }
 
   func testEmptyAnyVisitorPerformance() throws {
     try XCTSkipIf(ProcessInfo.processInfo.environment["SKIP_LONG_TESTS"] == "1")
     class EmptyAnyVisitor: SyntaxAnyVisitor {}
 
-    XCTAssertNoThrow(
-      try {
-        let parsed = Parser.parse(source: try String(contentsOf: inputFile))
+    let source = try String(contentsOf: inputFile)
+    let parsed = Parser.parse(source: source)
+    let emptyVisitor = EmptyAnyVisitor(viewMode: .sourceAccurate)
 
-        let emptyVisitor = EmptyAnyVisitor(viewMode: .sourceAccurate)
-
-        measure {
-          emptyVisitor.walk(parsed)
-        }
-      }()
-    )
+    try measureInstructions {
+      emptyVisitor.walk(parsed)
+    }
   }
 }
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,5 @@ Package.resolved
 
 .DS_Store
 *.pyc
+
+Tests/PerformanceTest/baselines.json
--- a/Package.swift
+++ b/Package.swift
@@ -283,8 +283,8 @@ let package = Package(
 
     .testTarget(
       name: "PerformanceTest",
-      dependencies: ["SwiftIDEUtils", "SwiftParser", "SwiftSyntax"],
-      exclude: ["Inputs"]
+      dependencies: ["_InstructionCounter", "SwiftIDEUtils", "SwiftParser", "SwiftSyntax"],
+      exclude: ["Inputs", "ci-baselines.json"]
     ),
   ]
 )
--- a/Tests/PerformanceTest/InstructionsCountAssertion.swift
+++ b/Tests/PerformanceTest/InstructionsCountAssertion.swift
@@ -1,0 +1,67 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import XCTest
+import _InstructionCounter
+
+fileprivate var baselineURL: URL {
+  if let baselineFile = ProcessInfo.processInfo.environment["BASELINE_FILE"] {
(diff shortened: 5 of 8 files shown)


ahoppen:
> But I didn't find a nice way to add baselines.

AFAIK that’s a known issue that we can’t define baselines when running XCTest using `swift test`, so we need to define our own, like you did. I have one suggestion though: Define the baselines in a JSON file. By default that JSON file could be `Tests/PerformanceTest/baselines.json` and we can add the baseline to `.gitignore` so that everybody can define their own baselines locally. You could then be able to override the baseline using an environment variable and CI would have `Tests/PerformanceTest/ci-baselines.json`. If the key doesn’t exist in the baseline, the test would fail to make sure we don’t add new performance tests without adding a baseline to `ci-baselines.json`. 

The baseline file could then look as follows:

```json
{
  "testNativeParsingPerformance": 5850081534,
  ...
}
```

and `measureInstructions` could infer the test case name from the test function as follows: 

```swift
func measureInstructions(_ baselineName: StaticString = #function)
```

kimdv:
@ahoppen did you think of something like this? 

I'm not sure I understand this
> You could then be able to override the baseline using an environment variable ...

kimdv:
@swift-ci please test

Tests/PerformanceTest/XCTestCase+InstructionsCount.swift
```diff
@@ -0,0 +1,74 @@
...
+  private var baselineURL: URL {
+    if ProcessInfo.processInfo.environment["SWIFTCI_USE_LOCAL_DEPS"] == "1" {
+      return URL(fileURLWithPath: #file)
+        .deletingLastPathComponent()
+        .appendingPathComponent("ci-baselines.json")
+    } else {
+      return URL(fileURLWithPath: #file)
+        .deletingLastPathComponent()
+        .appendingPathComponent("baselines.json")
+    }
```
ahoppen, line 26: I meant that this could be

```suggestion
    if let baselineFile = ProcessInfo.processInfo.environment["BASELINE_FILE"] {
      return URL(fileURLWithPath: baselineFile)
    } else {
      return URL(fileURLWithPath: #file)
        .deletingLastPathComponent()
        .appendingPathComponent("baselines.json")
    }
```

And then `build-script.py` could set `BASELINE_FILE=whatever/ci-baselines.json`.

Tests/PerformanceTest/XCTestCase+InstructionsCount.swift
```diff
@@ -0,0 +1,74 @@
...
+      file: file,
+      line: line
+    )
+  }
+
+  private func baseline(for key: String) throws -> UInt64? {
+    let baselineData = try? Data(contentsOf: baselineURL)
+
+    guard let data = baselineData else {
+      fatalError("No baseline.json provided")
```
ahoppen, line 66: I’d prefer to `throw` or `return nil` here so we don’t take down the entire test execution if the baseline file is missing.

Tests/PerformanceTest/XCTestCase+InstructionsCount.swift
```diff
@@ -0,0 +1,74 @@
...
+  func measureInstructions(_ baselineName: StaticString = #function, block: () -> Void, file: StaticString = #file, line: UInt = #line) throws {
+    let strippedBaselineName = "\(baselineName)".replacingOccurrences(of: "()", with: "")
+    var baseline: UInt64? = try baseline(for: strippedBaselineName)
+
+    if let environmentBaseline = ProcessInfo.processInfo.environment[strippedBaselineName] {
+      baseline = UInt64(environmentBaseline)
+    }
+
+    guard let baseline = baseline else {
+      XCTFail("Missing baseline for \(strippedBaselineName)", file: file, line: line)
```
ahoppen, line 38: What do you think about still executing the test and then failing with an error message that tells you what to add to your baselines file, i.e. the key and the number of instructions this test case took to execute.

Tests/PerformanceTest/SyntaxClassifierPerformanceTests.swift
```diff
@@ -26,17 +26,16 @@ public class SyntaxClassifierPerformanceTests: XCTestCase {
 
   func testClassifierPerformance() throws {
     try XCTSkipIf(ProcessInfo.processInfo.environment["SKIP_LONG_TESTS"] == "1")
-    XCTAssertNoThrow(
-      try {
+    try measureInstructions {
+      do {
```
ahoppen, line 30: We deliberately only wanted to measure the time to run `classifications` here, not the time to `parse`. Parsing the file is just set-up. Could you make sure that `measureInstructions` covers the same lines of code? 

Same in some other test cases.

Tests/PerformanceTest/VisitorPerformanceTests.swift
```diff
@@ -27,50 +27,48 @@ public class VisitorPerformanceTests: XCTestCase {
     try XCTSkipIf(ProcessInfo.processInfo.environment["SKIP_LONG_TESTS"] == "1")
     class EmptyVisitor: SyntaxVisitor {}
 
-    XCTAssertNoThrow(
-      try {
```
ahoppen, line 31: If you are modifying the test cases: There’s no need for `XCTAssertNoThrow` anymore. This was only necessary at a time when test methods weren’t allowed to throw. You should be able to just remove the calls to `XCTAssertNoThrow`.

ahoppen (commented):
Very nice.

One thought that just crossed my mind: I think it’s a little bit frustrating if all performance tests fail if no baseline file is provided because this means that if you check out the repo and run tests initially, you’ll see test failures. Because of this, what do you think about changing it so that if `PerformanceTest/baselines.json` doesn’t exist, all performance tests pass. Once you do care about performance, you’ll create a `baselines.json` and all performance tests for which you don’t have baselines will start failing, which might make sense.

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test

Tests/PerformanceTest/XCTestCase+InstructionsCount.swift
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  func measureInstructions(_ baselineName: StaticString = #function, block: () -> Void, file: StaticString = #file, line: UInt = #line) throws {
+    let strippedBaselineName = "\(baselineName)".replacingOccurrences(of: "()", with: "")
+    var baseline: UInt64? = try baseline(for: strippedBaselineName)
+
+    if let environmentBaseline = ProcessInfo.processInfo.environment[strippedBaselineName] {
+      baseline = UInt64(environmentBaseline)
+    }
+
```
ahoppen, line 34: What do you think would be the use case for overriding the baseline of individual test cases with an environment variable?

Tests/PerformanceTest/XCTestCase+InstructionsCount.swift
```diff
@@ -0,0 +1,72 @@
...
+    } else {
+      return URL(fileURLWithPath: #file)
+        .deletingLastPathComponent()
+        .appendingPathComponent("baselines.json")
+    }
+  }
+
+  func measureInstructions(_ baselineName: StaticString = #function, block: () -> Void, file: StaticString = #file, line: UInt = #line) throws {
+    let strippedBaselineName = "\(baselineName)".replacingOccurrences(of: "()", with: "")
+    var baseline: UInt64? = try baseline(for: strippedBaselineName)
```
ahoppen, line 29: I think if 
- No baseline *file* exists, the test should execute `block` (to check that it works functionally) but not emit a failure about the missing baseline. This is so you don’t immediately get hit with test failures if you clone the repo and run the tests
- If a baseline *file* exists but no baseline for this test case, you have opted into performance testing. In that case, there should be a test failure about the missing baseline.

Tests/PerformanceTest/XCTestCase+InstructionsCount.swift
```diff
@@ -0,0 +1,72 @@
...
+    let endInstructions = getInstructionsExecuted()
+    let numberOfInstructions = endInstructions - startInstructions
+
+    guard let baseline = baseline else {
+      XCTFail("Missing baseline for \(strippedBaselineName) with number of instructions \(numberOfInstructions)", file: file, line: line)
+      return
+    }
+
+    let lowerBaseline = UInt64(Double(baseline) * 0.97)
+    let upperBaseline = UInt64(Double(baseline) * 1.03)
```
ahoppen, line 46: I suspect that we could tighten this a lot further. Maybe to 1% or even 0.5% because the instruction counts are very stable. I recently had a case where I could measure performance differences of <1% https://github.com/apple/swift-syntax/pull/1909/commits/f9982f0959fd4043122a00ebb2a03bb5b674281e

ahoppen: Sounds good. Once the PR is done,  I would like to run CI on it 3 times to check that the performance numbers are always within these bounds.

ahoppen (approved):
Nice 🤩

kimdv:
@swift-ci please test

1/3

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test windows

kimdv:
@swift-ci please test

Tests/PerformanceTest/XCTestCase+InstructionsCount.swift
```diff
@@ -0,0 +1,68 @@
...
+        """,
+        file: file,
+        line: line
+      )
+      return
+    }
+
+    let lowerBaseline = UInt64(Double(baseline) * 0.995)
+    let upperBaseline = UInt64(Double(baseline) * 1.005)
+    let baselineRange = lowerBaseline...upperBaseline
```
ahoppen, line 56: Looking at the failure from your second CI run, the first think that I did was to calculate the difference in percents. So, I’m thinking that maybe we should also phrase the error message in those terms.

```swift
let baseline = 2624475912
let numberOfInstructions = 2640510088

let relativeDifference = Double(numberOfInstructions)/Double(baseline) - 1
let allowedDeviation = 0.005

assert(
  (-allowedDeviation..<allowedDeviation).contains(relativeDifference),
  "Number of instructions (\(numberOfInstructions)) deviated from baseline by \(String(format: "%.4f", relativeDifference * 100))%. The maximum allowed deviation in either direction is \(allowedDeviation * 100)%"
)

```

ahoppen (commented):
Looks like we need to increase the allowed deviation to 1%.

Tests/PerformanceTest/XCTestCase+InstructionsCount.swift
kimdv: I slight changed the assert message. I also added the name of the baseline. Just so it's easier to find.

kimdv:
@swift-ci please test

Tests/PerformanceTest/XCTestCase+InstructionsCount.swift
ahoppen: OK, great 👍🏽

ahoppen:
Oh, we should only specify the baseline if we are running on macOS. `getInstructionCount` doesn’t work on Linux and also, even if it did, we would have different performance counters. You can check if we’re running on macOS using

```python
if platform.system() == "Darwin":
```

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test

ahoppen:
2/3

@swift-ci please test

kimdv:
3/3

@swift-ci please test

PR #1864, opened 2023-06-28
Add diagnostic for label with string segment

### LINKED ISSUE #1831: `#sourceLocation` should not allow string literals with string interpolation
### Issue Kind

None

### Source Code

```swift
#sourceLocation(file: "test\(1).swift", line: 1)
```

### Description

No string interpolation should be allowed in the `file` argument of `#sourceLocation`.

kimdv:
Resolves #1831


CodeGeneration/Sources/SyntaxSupport/ExprNodes.swift
```diff
@@ -1490,6 +1490,26 @@ public let EXPR_NODES: [Node] = [
...
+        name: "Segment",
+        kind: .node(kind: .stringSegment)
+      ),
+      Child(
+        name: "CloseQuote",
+        kind: .token(choices: [.token(tokenKind: "StringQuoteToken"), .token(tokenKind: "MultilineStringQuoteToken")])
+      ),
+    ]
+  ),
+
```
kimdv, line 1512: Was it something like this you had in mind?

CodeGeneration/Sources/SyntaxSupport/ExprNodes.swift
```diff
@@ -1497,6 +1497,38 @@ public let EXPR_NODES: [Node] = [
     elementChoices: [.stringSegment, .expressionSegment]
   ),
 
+  Node(
+    kind: .simpleStringLiteralExpr,
+    base: .expr,
+    nameForDiagnostics: "simple string literal",
+    documentation: "A simple string that can't be used for string interpolation or extended escaping.",
```
kimdv, line 1504: Also tried to add some documentation. Improvements are welcome!

ahoppen: Thank you. That’s very appreciated.

I would go for

```suggestion
    documentation: "A simple string that can’t contain string interpolation and cannot have raw string delimiters.",
```

CodeGeneration/Sources/SyntaxSupport/ExprNodes.swift
```diff
@@ -1497,6 +1497,38 @@ public let EXPR_NODES: [Node] = [
     elementChoices: [.stringSegment, .expressionSegment]
   ),
 
+  Node(
+    kind: .simpleStringLiteralExpr,
```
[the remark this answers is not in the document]
ahoppen: I think the name is good

Sources/SwiftParser/StringLiterals.swift
```diff
@@ -593,6 +593,37 @@ extension Parser {
...
+    var segments: [RawStringSegmentSyntax] = []
+    var loopProgress = LoopProgressCondition()
+    while loopProgress.evaluate(self.currentToken) {
+      // If we encounter a token with leading trivia, we're no longer in the
+      // string literal.
+      guard currentToken.leadingTriviaText.isEmpty else { break }
+
+      if let stringSegment = self.consume(if: .stringSegment, TokenSpec(.identifier, remapping: .stringSegment)) {
+        segments.append(RawStringSegmentSyntax(content: stringSegment, arena: self.arena))
+      }
```
ahoppen, line 611: How difficult, do you think, would it be to also consume string interpolation segments here and make them unexpected? Or do you know how the current diagnostics look like if a simple string literals contains interpolation segments?

Sources/SwiftParser/StringLiterals.swift
```diff
@@ -593,6 +593,37 @@ extension Parser {
...
+    }
+
+    let (unexpectedBetweenSegmentAndCloseQuote, closeQuote) = self.expect(anyIn: SimpleStringLiteralExprSyntax.OpenQuoteOptions.self, default: .stringQuote)
```
ahoppen, line 614: It doesn’t make a difference, but I think this should be `CloseQuoteOptions`. Also, we should default to `openDelimiter.tokenKind` so that a multi-line string literal without a closing quote produces a diagnostic that adds `"""`.

```suggestion
    let (unexpectedBetweenSegmentAndCloseQuote, closeQuote) = self.expect(anyIn: SimpleStringLiteralExprSyntax.CloseQuoteOptions.self, default: openDelimiter.tokenKind)
```

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -1465,6 +1492,33 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
+          node,
+          ForbiddenExtendedEscapingString(token: args.fileLabel),
+          handledNodes: rawDelimiters.map { $0.id }
+        )
+      }
+    }
+
+    return .visitChildren
+  }
+
```
ahoppen, line 1521: Can’t we unify this with `visit(_ node: AvailabilityLabeledArgumentSyntax)` by diagnosing it in `visit(_ node: SimpleStringLiteralSyntax)`?

Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift
```diff
@@ -327,6 +327,22 @@ public struct ExtraneousWhitespace: ParserError {
...
+}
+
+public struct ForbiddenInterpolatedString: ParserError {
+  public let token: TokenSyntax
+
+  public var message: String {
+    return "'\(token.text)' cannot be an interpolated string literal"
+  }
+}
+
```
ahoppen, line 345: Are you using `ForbiddenInterpolatedString` anywhere? I can’t see where you do.

Sources/SwiftSyntax/SourceLocation.swift
```diff
@@ -143,15 +140,10 @@ fileprivate struct SourceLocationDirectiveArguments {
   let line: Int
 
   init(_ args: PoundSourceLocationArgsSyntax) throws {
-    guard args.fileName.segments.count == 1,
-      case .stringSegment(let segment) = args.fileName.segments.first!
-    else {
-      throw Error.stringInterpolationInFileName(args.fileName)
-    }
-    self.file = segment.content.text
```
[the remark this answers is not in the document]
ahoppen: I agree that multi-line string literals don’t make any sense here. It’s on my personal list of things that we should consider changing for Swift 6 since this would technically be a source-breaking change.

Sources/SwiftSyntax/SourceLocation.swift
```diff
@@ -143,15 +140,10 @@ fileprivate struct SourceLocationDirectiveArguments {
...
     guard let line = Int(args.lineNumber.text) else {
       throw Error.nonDecimalLineNumber(args.lineNumber)
     }
+    self.file = args.fileName.segments.map {  $0.content.text}.joined()
```
ahoppen, line 146:
```suggestion
    self.file = args.fileName.segments.map {  $0.content.text }.joined()
```

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
kimdv: The reason was that we using `args.fileLabel` to create the message. 
We can unify it by making a more generic error message maybe?

ahoppen: Since this isn’t a super common error, I would prefer the generalized implementation at the expense of making the error message slightly more generic, e.g. just `argument cannot be an extended escaping string literal`

Sources/SwiftParser/StringLiterals.swift
```diff
@@ -595,10 +595,80 @@ extension Parser {
...
+    while loopProgress.evaluate(self.currentToken) {
+      // If we encounter a token with leading trivia, we're no longer in the
+      // string literal.
+      guard currentToken.leadingTriviaText.isEmpty else { break }
+
+      if let stringSegment = self.consume(if: .stringSegment, TokenSpec(.identifier, remapping: .stringSegment)) {
+        var unexpectedAfterContent: RawUnexpectedNodesSyntax?
+
+        if let backslash = self.consume(if: .backslash) {
+          var tokens: [RawTokenSyntax] = [backslash]
```
ahoppen, line 615:
```suggestion
          var unexpectedTokens: [RawTokenSyntax] = [backslash]
```

Sources/SwiftParser/StringLiterals.swift
```diff
@@ -595,10 +595,80 @@ extension Parser {
...
 
+fileprivate extension RawTokenSyntax {
+  var closeTokenKind: SimpleStringLiteralExprSyntax.CloseQuoteOptions {
+    switch self {
+    case .multilineStringQuote:
+      return .multilineStringQuote
+    case .stringQuote:
+      return .stringQuote
+    default:
+      fatalError("Unsupported type")
```
ahoppen, line 667: I hate seing `fatalError` in the parser unless it’s super obvious why it can’t happen. What do you think about defaulting this to `.stringQuote` here? That does seem like a reasonable default to me.

kimdv:
@swift-ci please test

ahoppen (approved):

kimdv:
@ahoppen `testDiagnoseAvailability18` fails.

It fails formatting the same code, this is the result

```swift
@available(*, unavailable, message: """
  foobar message.
  
""") func multilineMessage() {
}
multilineMessage()
```

A quick debug I think it's because we do some post processing of multiline string when parsing a `RawStringLiteralExprSyntax`.

We can use the same method, but that require we convert the `RawStringSegmentSyntax` to a `RawStringLiteralSegmentsSyntax.Element` and then back to `RawStringSegmentSyntax`.

Or we should try to extract the `postProcessMultilineStringLiteral` into multiple functions so we better can reuse it. 

What do you think?

ahoppen:
I think we should try just re-using `postProcessMultilineStringLiteral`. Mapping between `RawStringLiteralExprSyntax` and `RawStringLiteralSegmentsSyntax.Element` doesn’t result in an allocation on the bump allocator, so we shouldn’t be leaking any memory that way, so the mapping should be fine.

Sources/SwiftSyntax/SourceLocation.swift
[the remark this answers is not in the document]
ahoppen: Sorry for the slow reply. I just created a [Source Breaking](https://github.com/apple/swift-syntax/labels/Source%20Breaking) label that we can attach to bugs whose fix is source breaking.

Sources/SwiftParser/StringLiterals.swift
```diff
@@ -147,7 +147,7 @@ extension Parser {
     return RawTokenSyntax(
       kind: token.tokenKind,
       text: SyntaxText(rebasing: token.tokenText.dropFirst(reclassifyLeading.count).dropLast(reclassifyTrailing.count)),
-      leadingTriviaPieces: token.leadingTriviaPieces + TriviaParser.parseTrivia(reclassifyLeading, position: .trailing),
+      leadingTriviaPieces: token.leadingTriviaPieces + TriviaParser.parseTrivia(reclassifyLeading, position: .leading),
```
kimdv, line 150: @ahoppen I found this one. I tought it was a bug when reading

kimdv:
@ahoppen I was just debugging and trying to figure out why we get this error

I've tried to debug, but I can't find the reason

```
Trees do not match due to token in:
stringSegment("  foobar message.") [2:1...2:18]
when expecting:
stringSegment("foobar message.") [2:3...2:18]
```

ahoppen:
As the failure message says: What we do is remove all the trivia of the tree, run `BasicFormat` and check that we end up with the same tree. This basically tests `BasicFormat` to check that we indeed separate different tokens by space.

But when you remove the leading spaces from the closing `"""`, you are really changing the semantic meaning of the string literal because the two spaces in front of `foobar message.` now become part of the string literal.

That’s why in SwiftParserTest/Assertions.swift:735 (inside `TriviaRemover`, we check if the token is inside a `StringLiteralSyntax`, in which case we just don’t remove trivia from inside this subtree. You just need to add `SimpleStringLiteralExprSyntax` here.

And as I’m looking at the code, I’m realizing that we could actually simplify `TriviaRemover`

```swift
/// Removes trivia from all tokens that don’t occur inside multiline string
/// literals.
///
/// We keep trivia inside multiline string literals because the indentation of
/// the closing quote of a multi-line string literals has impact on how much
/// leading trivia is stripped from the literal’s content.
class TriviaRemover: SyntaxRewriter {
  override func visit(_ node: StringLiteralExprSyntax) -> ExprSyntax {
    if node.openingQuote == .multilineStringQuoteToken() {
      return ExprSyntax(node)
    } else {
      return super.visit(node)
    }
  }

  override func visit(_ node: SimpleStringLiteralExprSyntax) -> ExprSyntax {
    if node.openingQuote == .multilineStringQuoteToken() {
      return ExprSyntax(node)
    } else {
      return super.visit(node)
    }
  }

    return token.with(\.leadingTrivia, []).with(\.trailingTrivia, [])
  }
}
```

Sources/SwiftSyntax/SourceLocation.swift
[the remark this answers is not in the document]
kimdv: Added in https://github.com/apple/swift-syntax/issues/2002

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test

ahoppen:
@swift-ci Please test Windows

Sources/SwiftParser/StringLiterals.swift
ahoppen: Yes, seems like a bug to me. I don’t see a reason why this should be `trailing`.

Sources/SwiftParser/StringLiterals.swift
```diff
@@ -595,10 +595,105 @@ extension Parser {
...
+      return RawSimpleStringLiteralExprSyntax(
+        RawUnexpectedNodesSyntax(
+          combining: unexpectedBeforeOpenQuote,
+          postProcessed.unexpectedBeforeOpeningQuote,
+          openDelimiter,
+          arena: self.arena
+        ),
+        openingQuote: postProcessed.openingQuote,
+        segments: RawSimpleStringLiteralSegmentListSyntax(
+          elements: postProcessed.segments.compactMap { $0.as(RawStringSegmentSyntax.self) },
```
ahoppen, line 656: The reason why we can do the `compactMap` here is that we know we added only `RawStringSegmentSyntax` to the segments above, right? I think we should call this out in a short comment. At first it reads like we might be dropping syntax nodes here. And if we can guarantee that `postProcessed.segments` only contains `RawStringSegmentSyntax`, this can be the following, which makes the intent clearer

```suggestion
          elements: postProcessed.segments.map { $0.cast(RawStringSegmentSyntax.self) },
```

Sources/SwiftParser/StringLiterals.swift
```diff
@@ -595,10 +595,105 @@ extension Parser {
...
+          arena: self.arena
+        ),
+        unexpectedBetweenSegmentAndCloseQuote,
+        closingQuote: postProcessed.closingQuote,
+        RawUnexpectedNodesSyntax(
+          combining: closeDelimiter,
+          postProcessed.unexpectedBeforeClosingQuote,
```
ahoppen, line 663: You should be adding `postProcessed.unexpectedBeforeClosingQuote` to `unexpectedBetweenSegmentAndCloseQuote` (and after `unexpectedBetweenSegmentAndCloseQuote`) , otherwise this can fail to round-trip.

Sources/SwiftParser/StringLiterals.swift
```diff
@@ -595,10 +595,105 @@ extension Parser {
...
+        openQuote: openQuote,
+        segments: segments.compactMap { RawStringLiteralSegmentListSyntax.Element.stringSegment($0) },
+        closeQuote: closeQuote
+      )
+
+      return RawSimpleStringLiteralExprSyntax(
+        RawUnexpectedNodesSyntax(
+          combining: unexpectedBeforeOpenQuote,
+          postProcessed.unexpectedBeforeOpeningQuote,
+          openDelimiter,
```
ahoppen, line 651: I would revers this order because `unexpectedBeforeOpeningQuote` is directly before the opening quote and thus after the delimiter.

```suggestion
          combining: openDelimiter,
          unexpectedBeforeOpenQuote,
          postProcessed.unexpectedBeforeOpeningQuote,

```

ahoppen (commented):
A couple comments to prevent round-trip failures, otherwise LGTM.

Sources/SwiftParser/StringLiterals.swift
kimdv: We can guarantee that. It will not change the type of the segments.

kimdv: The method `.cast` is not available. 
`RawSyntaxNodeProtocol` only have `as` and `is`

Sources/SwiftParser/StringLiterals.swift
kimdv: Should we add a test case for this bug?

ahoppen: I you have an idea how to write one: Sure. If you don’t find a test case that exposes it 🤷🏽

kimdv:
@swift-ci Please test

kimdv:
@swift-ci Please test windows

ahoppen (approved):

PR #1867, opened 2023-06-29
Respect indentation in the line for sub-tree in BasicFormat

gibachan:
The pull request at https://github.com/apple/swift-syntax/pull/1791 emphasizes the need to indent sub-tree nodes while formatting.

I attempted to address another sub-tree indentation based on the comment in https://github.com/apple/swift-syntax/pull/1791#discussion_r1231314001.

### DIFF
--- a/Sources/SwiftBasicFormat/BasicFormat.swift
+++ b/Sources/SwiftBasicFormat/BasicFormat.swift
@@ -150,6 +150,18 @@ open class BasicFormat: SyntaxRewriter {
 
   // MARK: - Customization points
 
+  /// If we are formatting a subtree and the line that the initial token occurs on is indented,
+  /// use that line indentation for the first token in the subtree to format.
+  ///
+  /// For example, when formatting only the code block in the following,
+  /// then the opening `{` should be indented by four spaces.
+  /// ```
+  ///     func test() {
+  ///         print(1)
+  ///     }
+  /// ```
+  open var inferInitialTokenIndentaiton: Bool { true }
+
   /// Whether a leading newline on `token` should be added.
   open func requiresIndent(_ node: some SyntaxProtocol) -> Bool {
     return node.requiresIndent
@@ -411,6 +423,16 @@ open class BasicFormat: SyntaxRewriter {
       // If the token starts on a new line and does not have indentation, this
       // is the last non-indented token. Store its indentation level
       anchorPoints[token] = currentIndentationLevel
+    }
+
+    if inferInitialTokenIndentaiton
+      && isInitialToken
+      && token.presence == .present
+    {
+      let indentationOfLine = token.indentationOfLine
+      if token.leadingTrivia.pieces.suffix(indentationOfLine.pieces.count) != indentationOfLine.pieces {
+        leadingTrivia += indentationOfLine
+      }
     }
 
     // Add a trailing space to the token unless
--- a/Sources/SwiftBasicFormat/Syntax+Extensions.swift
+++ b/Sources/SwiftBasicFormat/Syntax+Extensions.swift
@@ -12,7 +12,7 @@
 
 import SwiftSyntax
 
-public extension TokenSyntax {
+extension TokenSyntax {
   /// The indentation of this token
   ///
   /// In contrast to `indentationOfLine`, this does not walk to previous tokens to
@@ -23,7 +23,7 @@ public extension TokenSyntax {
   }
 
   /// Returns the indentation of the line this token occurs on
-  var indentationOfLine: Trivia {
+  public var indentationOfLine: Trivia {
     var token: TokenSyntax = self
     if let indentation = token.indentation {
       return indentation
--- a/Sources/SwiftBasicFormat/Syntax+Extensions.swift
+++ b/Sources/SwiftBasicFormat/Syntax+Extensions.swift
@@ -1,0 +1,40 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
+public extension TokenSyntax {
+  /// The indentation of this token
+  ///
+  /// In contrast to `indentationOfLine`, this does not walk to previous tokens to
+  /// find the indentation of the line this token occurs on.
+  private var indentation: Trivia? {
+    let previous = self.previousToken(viewMode: .sourceAccurate)
+    return ((previous?.trailingTrivia ?? []) + leadingTrivia).indentation(isOnNewline: false)
+  }
+
+  /// Returns the indentation of the line this token occurs on
+  var indentationOfLine: Trivia {
+    var token: TokenSyntax = self
+    if let indentation = token.indentation {
+      return indentation
+    }
+    while let previous = token.previousToken(viewMode: .sourceAccurate) {
+      token = previous
+      if let indentation = token.indentation {
+        return indentation
+      }
+    }
+
+    return []
+  }
+}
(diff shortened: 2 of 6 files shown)


Sources/SwiftBasicFormat/BasicFormat.swift
```diff
@@ -413,6 +413,15 @@ open class BasicFormat: SyntaxRewriter {
       anchorPoints[token] = currentIndentationLevel
     }
 
+    // Add an indentation in the line for the initial token,
+    // if the token is present and its leadingTrivia does not contain the indentation.
```
ahoppen, line 417: I’m not sure if my formulation of the comment is more helpful than yours but I think some information like the following would have helped me.

````swift
// If we are formatting a subtree and the line that the initial token occurs on is indented, use that line indentation for the first token in the subtree to format. For example, when formatting only the code block in the following, then the opening `{` should be indented by four spaces.
// ```
//     func test() {
//         print(1)
//     }
// ```
````

File: Sources/SwiftParserDiagnostics/MissingNodesError.swift [hunk 2]
```diff
@@ -36,7 +36,11 @@ fileprivate enum NodesDescriptionPart {
     switch self {
     case .tokensWithDefaultText(var tokens):
       if format {
-        tokens = tokens.map({ NoNewlinesFormat(viewMode: .all).visit($0) })
+        tokens = tokens.enumerated()
+          .map { index, token in
+            // We detach the nodes from the second one onwards, considering the line indentation only for the first node
+            NoNewlinesFormat(viewMode: .all).visit(index == 0 ? token : token.detached)
+          }
```
ahoppen, line 43: Why is this needed? I would expect that we should be able to always do

```diff
- tokens = tokens.map({ NoNewlinesFormat(viewMode: .all).visit($0) })
+ tokens = tokens.map({ NoNewlinesFormat(viewMode: .all).visit($0.detached) })
```

ahoppen (commented):
Thanks a lot. I’ve got a couple comments/questions inline. It looks really good overall though.

Sources/SwiftBasicFormat/Syntax+Extensions.swift
```diff
@@ -0,0 +1,40 @@
...
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
+public extension TokenSyntax {
+  /// The indentation of this token
+  ///
+  /// In contrast to `indentation`, this does not walk to previous tokens to
```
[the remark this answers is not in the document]
gibachan: Fixed 👍 [90ab769](https://github.com/apple/swift-syntax/pull/1867/commits/90ab7699a30db1f7ce14067be2c2d833f8cf13c3)

Sources/SwiftBasicFormat/BasicFormat.swift
gibachan: Nice comment! Thank you https://github.com/apple/swift-syntax/pull/1867/commits/a27a8d3ff0da5f40ca6c104377d3472b6cc0f728

Sources/SwiftParserDiagnostics/MissingNodesError.swift
gibachan: I seem to have misunderstood, and now I agree with you.

However, [this]([cd3ad12](https://github.com/apple/swift-syntax/pull/1867/commits/cd3ad12ce53e826b6583e609528e7ac5b1994e7e)) will cause a test to fail for this test case.

https://github.com/apple/swift-syntax/blob/e7068e8eb76542591a189f703f8b62881a520588/Tests/SwiftParserTest/AttributeTests.swift#L99-L113


The error message appears in this line of the DiagnosticSpec, stating that it outputs `:false` instead of `: false`.
https://github.com/apple/swift-syntax/blob/e7068e8eb76542591a189f703f8b62881a520588/Tests/SwiftParserTest/AttributeTests.swift#L106

The whitespace between `:` and `false` should be inserted by BasicFormat, and I believe the code responsible for it is [here](https://github.com/apple/swift-syntax/blob/e7068e8eb76542591a189f703f8b62881a520588/Sources/SwiftBasicFormat/BasicFormat.swift#L416-L425).

But `requiresWhitespace(between:and:)` returns false because the detached `:` token will not find `false` as its next token. As a result, the whitespace is not inserted as expected in the test case.

I am currently investigating this issue.

ahoppen: Ah, I see what’s going wrong g here: We definitely shouldn’t `detach` because otherwise we don’t know that `:` is followed by `false` in `NoNewlinesFormat` and thus can’t decide whether to add a space after it. 

The issue is that inheriting the line’s indentation only makes sense if the formatted text should contain newlines. `NoNewlinesFormat` is a bit of an oddity here because it doesn’t do that. What I would do is to add the following property to `BasicFormat` and only perform your logic if `inferInitialTokenIndentaiton` is `true`. `NoNewlinesFormat` can then override it and set it to `false`.

```swift
/// If we are formatting a subtree and the line that the initial token occurs on is indented,
/// use that line indentation for the first token in the subtree to format.
/// 
/// For example, when formatting only the code block in the following,
/// then the opening `{` should be indented by four spaces.
/// ```
///     func test() {
///         print(1)
///     }
/// ```
open var inferInitialTokenIndentaiton: Bool { true }
```

And then, I think we also don’t need the inline comment in `visit(_: TokenSyntax)` anymore because the behavior is documented on `inferInitialTokenIndentaiton`.

gibachan: Wow, thank you for your great support!
I fixed it and squashed those commits into cc9f4496099a0596dcf99da6881551aa4e590c78

ahoppen (approved):
This looks good. Thanks 👍🏽

gibachan:
Oh, I forgot to run "./format.py" for my changes, sorry. I fixed it! https://github.com/apple/swift-syntax/pull/1867/commits/1622dc6c0fe21840cfe4245a186b20dec7ab1fe4

ahoppen:
@swift-ci Please test

Sources/SwiftBasicFormat/Syntax+Extensions.swift
ahoppen, line 1: When building swift-syntax for the compiler, we are building it using CMake and you need to add this file to `Sources/SwiftBasicFormat/CMakeLists.txt`.

Sources/SwiftBasicFormat/Syntax+Extensions.swift
```diff
@@ -0,0 +1,40 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
+public extension TokenSyntax {
```
ahoppen, line 15: Looking at this file again, is there a reason this extension needs to be public? If not, I’d prefer to keep it internal so we don’t increase our API surface area further.

Sources/SwiftBasicFormat/Syntax+Extensions.swift
gibachan: Added this file to Sources/SwiftBasicFormat/CMakeLists.txt. [73a21a8](https://github.com/apple/swift-syntax/pull/1867/commits/73a21a831be0b4cd8bccb11668c642e51f23bf0a)

Sources/SwiftBasicFormat/Syntax+Extensions.swift
gibachan: I did it to make `var indentationOfLine: Trivia` accessible from [ParseDiagnosticsGenerator.swift](https://github.com/apple/swift-syntax/blob/d7b149326f65d1721ad2210360ec7c8a2c18335f/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift#L737) which is in SwiftParserDiagnostics package.

I made this extension internal in [30bfccd](https://github.com/apple/swift-syntax/pull/1867/commits/30bfccdc9097a926e068bd839a4b79efc1e7d277).  Please let me know if this was not your intention 😄

ahoppen: Ah, I meant if we could make `indentationOfLine` internal but I forgot that we use it from `ParseDiagnosticsGenerator`. Makes sense to be public 👍🏽

ahoppen:
@swift-ci Please test

PR #1874, opened 2023-07-02
[Macros] Implement function body macros

DougGregor:
Introduce function body macros, which are comprised of two similar macro roles:

- Preamble macros introduce "preamble" code into a user-written function body, e.g., to perform tracing, logging, check additional preconditions, etc.
- Body macros: introduce a function body into a function that has none or replace the existing function body

### DIFF
--- a/Sources/SwiftSyntaxMacroExpansion/MacroExpansion.swift
+++ b/Sources/SwiftSyntaxMacroExpansion/MacroExpansion.swift
@@ -12,7 +12,7 @@
 
 import SwiftBasicFormat
 import SwiftSyntax
-@_spi(MacroExpansion) import SwiftSyntaxMacros
+@_spi(MacroExpansion) @_spi(ExperimentalLanguageFeature) import SwiftSyntaxMacros
 
 public enum MacroRole {
   case expression
@@ -24,8 +24,8 @@ public enum MacroRole {
   case conformance
   case codeItem
   case `extension`
-  case preamble
-  case body
+  @_spi(ExperimentalLanguageFeature) case preamble
+  @_spi(ExperimentalLanguageFeature) case body
 }
 
 extension MacroRole {
@@ -55,7 +55,7 @@ enum MacroExpansionError: Error, CustomStringConvertible {
   case declarationHasNoBody
   case noExtendedTypeSyntax
   case noFreestandingMacroRoles(Macro.Type)
-  case tooManyBodyMacros
+  case moreThanOneBodyMacro
   case preambleWithoutBody
 
   var description: String {
@@ -81,7 +81,7 @@ enum MacroExpansionError: Error, CustomStringConvertible {
     case .noFreestandingMacroRoles(let type):
       return "macro implementation type '\(type)' does not conform to any freestanding macro protocol"
 
-    case .tooManyBodyMacros:
+    case .moreThanOneBodyMacro:
       return "function can not have more than one body macro applied to it"
 
     case .preambleWithoutBody:
--- a/Sources/SwiftSyntaxMacroExpansion/MacroExpansion.swift
+++ b/Sources/SwiftSyntaxMacroExpansion/MacroExpansion.swift
@@ -24,6 +24,8 @@ public enum MacroRole {
   case conformance
   case codeItem
   case `extension`
+  case preamble
+  case body
 }
 
 extension MacroRole {
@@ -38,18 +40,23 @@ extension MacroRole {
     case .conformance: return "ConformanceMacro"
     case .codeItem: return "CodeItemMacro"
     case .extension: return "ExtensionMacro"
+    case .preamble: return "PreambleMacro"
+    case .body: return "BodyMacro"
     }
   }
 }
 
 /// Simple diagnostic message
-private enum MacroExpansionError: Error, CustomStringConvertible {
+enum MacroExpansionError: Error, CustomStringConvertible {
   case unmatchedMacroRole(Macro.Type, MacroRole)
   case parentDeclGroupNil
   case declarationNotDeclGroup
   case declarationNotIdentified
+  case declarationHasNoBody
   case noExtendedTypeSyntax
   case noFreestandingMacroRoles(Macro.Type)
+  case tooManyBodyMacros
+  case preambleWithoutBody
 
   var description: String {
     switch self {
@@ -65,12 +72,20 @@ private enum MacroExpansionError: Error, CustomStringConvertible {
     case .declarationNotIdentified:
       return "declaration is not a 'Identified' syntax"
 
+    case .declarationHasNoBody:
+      return "declaration is not a type with an optional code block"
+
     case .noExtendedTypeSyntax:
       return "no extended type for extension macro"
 
     case .noFreestandingMacroRoles(let type):
       return "macro implementation type '\(type)' does not conform to any freestanding macro protocol"
 
+    case .tooManyBodyMacros:
+      return "function can not have more than one body macro applied to it"
+
+    case .preambleWithoutBody:
+      return "preamble macro cannot be applied to a function with no body"
     }
   }
 }
@@ -125,7 +140,7 @@ public func expandFreestandingMacro(
       expandedSyntax = Syntax(CodeBlockItemListSyntax(rewritten))
 
     case (.accessor, _), (.memberAttribute, _), (.member, _), (.peer, _), (.conformance, _), (.extension, _), (.expression, _), (.declaration, _),
-      (.codeItem, _):
+      (.codeItem, _), (.preamble, _), (.body, _):
       throw MacroExpansionError.unmatchedMacroRole(definition, macroRole)
     }
     return expandedSyntax.formattedExpansion(definition.formatMode, indentationWidth: indentationWidth)
@@ -285,6 +300,38 @@ public func expandAttachedMacroWithoutCollapsing<Context: MacroExpansionContext>
 
       // Form a buffer of peer declarations to return to the caller.
       return extensions.map {
+        $0.formattedExpansion(definition.formatMode, indentationWidth: indentationWidth)
+      }
+
+    case (let attachedMacro as PreambleMacro.Type, .preamble):
+      guard let declToPass = Syntax(declarationNode).asProtocol(SyntaxProtocol.self) as? (DeclSyntaxProtocol & WithOptionalCodeBlockSyntax)
+      else {
+        // Compiler error: declaration must have a body.
+        throw MacroExpansionError.declarationHasNoBody
+      }
+
+      let preamble = try attachedMacro.expansion(
+        of: attributeNode,
+        providingPreambleFor: declToPass,
+        in: context
+      )
+      return preamble.map {
+        $0.formattedExpansion(definition.formatMode, indentationWidth: indentationWidth)
+      }
+
+    case (let attachedMacro as BodyMacro.Type, .body):
+      guard let declToPass = Syntax(declarationNode).asProtocol(SyntaxProtocol.self) as? (DeclSyntaxProtocol & WithOptionalCodeBlockSyntax)
+      else {
+        // Compiler error: declaration must have a body.
+        throw MacroExpansionError.declarationHasNoBody
+      }
+
+      let body = try attachedMacro.expansion(
+        of: attributeNode,
+        providingBodyFor: declToPass,
+        in: context
+      )
+      return body.map {
         $0.formattedExpansion(definition.formatMode, indentationWidth: indentationWidth)
       }
 
--- a/Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
+++ b/Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
@@ -349,6 +349,71 @@ private func expandExtensionMacro(
   return "\(raw: indentedSource)"
 }
 
+/// Expand a preamble macro into a list of code items.
+private func expandPreambleMacro(
+  definition: PreambleMacro.Type,
+  attributeNode: AttributeSyntax,
+  attachedTo decl: some DeclSyntaxProtocol & WithOptionalCodeBlockSyntax,
+  in context: some MacroExpansionContext,
+  indentationWidth: Trivia
+) -> CodeBlockItemListSyntax? {
+  guard
+    let expanded = expandAttachedMacro(
+      definition: definition,
+      macroRole: .preamble,
+      attributeNode: attributeNode.detach(
+        in: context,
+        foldingWith: .standardOperators
+      ),
+      declarationNode: DeclSyntax(decl.detach(in: context)),
+      parentDeclNode: nil,
+      extendedType: nil,
+      conformanceList: nil,
+      in: context,
+      indentationWidth: indentationWidth
+    )
+  else {
+    return []
+  }
+
+  // Match the indentation of the statements if we can, and put newlines around
+  // the preamble to separate it from the rest of the body.
+  let indentation = decl.body?.statements.indentationOfFirstLine ?? (decl.indentationOfFirstLine + indentationWidth)
+  let indentedSource = "\n" + expanded.indented(by: indentation) + "\n\n\n"
+  return "\(raw: indentedSource)"
+}
+
+private func expandBodyMacro(
+  definition: BodyMacro.Type,
+  attributeNode: AttributeSyntax,
+  attachedTo decl: some DeclSyntaxProtocol & WithOptionalCodeBlockSyntax,
+  in context: some MacroExpansionContext,
+  indentationWidth: Trivia
+) -> CodeBlockSyntax? {
+  guard
+    let expanded = expandAttachedMacro(
+      definition: definition,
+      macroRole: .body,
+      attributeNode: attributeNode.detach(
+        in: context,
+        foldingWith: .standardOperators
+      ),
+      declarationNode: DeclSyntax(decl.detach(in: context)),
+      parentDeclNode: nil,
+      extendedType: nil,
+      conformanceList: nil,
+      in: context,
+      indentationWidth: indentationWidth
+    )
+  else {
+    return nil
+  }
+
+  // Wrap the body in braces.
+  let indentedSource = " {\n" + expanded.indented(by: decl.indentationOfFirstLine + indentationWidth) + "\n}\n"
+  return "\(raw: indentedSource)" as CodeBlockSyntax
+}
+
 // MARK: - MacroSystem
 
 /// Describes the kinds of errors that can occur within a macro system.
@@ -567,14 +632,21 @@ private class MacroApplication<Context: MacroExpansionContext>: SyntaxRewriter {
     case .notAMacro:
       break
     }
-    if let declSyntax = node.as(DeclSyntax.self),
+    if var declSyntax = node.as(DeclSyntax.self),
       let attributedNode = node.asProtocol(WithAttributesSyntax.self),
       !attributedNode.attributes.isEmpty
     {
+      // Apply body and preamble macros.
+      if let nodeWithBody = node.asProtocol(WithOptionalCodeBlockSyntax.self),
+        let declNodeWithBody = nodeWithBody as? any DeclSyntaxProtocol & WithOptionalCodeBlockSyntax
+      {
+        declSyntax = DeclSyntax(visitBodyAndPreambleMacros(declNodeWithBody))
+      }
+
       // Visit the node, disabling the `visitAny` handling.
-      skipVisitAnyHandling.insert(node)
+      skipVisitAnyHandling.insert(Syntax(declSyntax))
       let visitedNode = self.visit(declSyntax)
-      skipVisitAnyHandling.remove(node)
+      skipVisitAnyHandling.remove(Syntax(declSyntax))
 
       let attributesToRemove = self.macroAttributes(attachedTo: visitedNode).map(\.attributeNode)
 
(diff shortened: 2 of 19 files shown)


DougGregor:
@swift-ci please test

DougGregor:
This is your fault, @ktoso

ktoso:
Sending much much love! 🥰🥰🥰

Sources/SwiftSyntaxMacroExpansion/MacroExpansion.swift
```diff
@@ -335,6 +343,38 @@ public func expandAttachedMacroWithoutCollapsing<Context: MacroExpansionContext>
...
+      let preamble = try attachedMacro.expansion(
+        of: attributeNode,
+        providingPreambleFor: declToPass,
+        in: context
+      )
+      return preamble.map {
+        $0.formattedExpansion(definition.formatMode)
+      }
+
+    case (let attachedMacro as BodyMacro.Type, .body):
```
ktoso, line 362: Hmmm, this doesn't trigger if the function we're attaching the macro to has a body, but that's the case we'd want -- am I confused about something here or are we talking about different kinds of macros perhaps?

ktoso (commented):
Really awesome, thank you! I played around with it and either I'm mis-expected this as a body "replacement" macro that this just isn't, or we miscomunicated how such a thing would be used? 🤔

DougGregor:
> Really awesome, thank you! I played around with it and either I'm mis-expected this as a body "replacement" macro that this just isn't, or we miscomunicated how such a thing would be used? 🤔

I'm fretting over this. Replacing the body wholesale, without it getting type-checked, seems likely to provide a poor user experience. Type-checking the entire body twice seems expensive. Hence, my attempt at "preamble" macros that let one introduce new statements into the beginning of an existing function body without changing how it is type checked. With `defer`, this lets one do prologue and epilogue stuff... but it's not quite good enough if what you really need is to wrap the whole body in a closure, as with the task local APIs.

I suppose we could allow body replacement with body macros, and tell people to only do that if they can't use preamble macros, because the user experience will be worse.

ktoso:
Yeah for better or worse, we have to wrap when using task local APIs. It would be nice if there was some general "`with x`"-like feature that would substitute wrapping, but we don't have that so wrapping it is 🤔 

Not type-checking seems like it would be a bit scary as default - would it be possible to be something on the passed `context` that could `context.typecheckBody = false` if one devloped the macro and is now "sure" that it'll work correctly? The transformation in question is pretty boring so once implemeted right once I don't think it'd cause much trouble... but it also seems a bit weird for only this macro kind to not typecheck "at all"? 

Unless maybe it would be some "`NotTypeCheckedBodyReplacementMacro`" to spell out that "this is the scary one, don't use if you can help it"? 🤔

Sources/SwiftSyntaxMacroExpansion/MacroExpansion.swift
```diff
@@ -12,6 +12,7 @@
 
 import SwiftSyntax
 @_spi(MacroExpansion) import SwiftSyntaxMacros
+import SwiftSyntaxBuilder  // FIXME: Why is HasTrailingOptionalCodeBlock there?
```
ahoppen, line 15: Would you mind defining a trait for `HasTrailingCodeBlock` and  `HasTrailingOptionalCodeBlock`?

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
```diff
@@ -382,6 +385,30 @@ class MacroApplication<Context: MacroExpansionContext>: SyntaxRewriter {
...
+    return visit(function: node)
+  }
+
+  override func visit(_ node: FunctionDeclSyntax) -> DeclSyntax {
+    return visit(function: node)
+  }
+
+  override func visit(_ node: InitializerDeclSyntax) -> DeclSyntax {
+    return visit(function: node)
+  }
```
ahoppen, line 411: It seems a little fragile that we’ll forget to add a new `visit` function here if there’s a new node that conforms to `HasTrailingOptionalCodeBlock`. Wouldn’t it be better to do the following in `visitAny` or am I missing something?

```swift
if let  node = node.as(SyntaxProtocol.self) as DeclSyntaxProtocol & HasTrailingOptionalCodeBlock {
  visit(node)
}
```

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
```diff
@@ -510,6 +537,56 @@ extension MacroApplication {
...
+        preamble.append(contentsOf: newPreamble)
+      } catch {
+        context.addDiagnostics(from: error, node: attribute)
+      }
+    }
+
+    return preamble
+  }
+
+  private func expandBody(
```
ahoppen, line 561: Since this expands both body and preamble, I would call the function `expandBodyAndPreamble`.

Sources/SwiftSyntaxMacros/MacroProtocols/BodyMacro.swift
```diff
@@ -0,0 +1,26 @@
...
+/// have one.
+public protocol BodyMacro: AttachedMacro {
+  /// Expand a macro described by the given custom attribute and
+  /// attached to the given declaration and evaluated within a
+  /// particular expansion context.
+  ///
+  /// The macro expansion can introduce a body for the given function.
+  static func expansion(
+    of node: AttributeSyntax,
+    providingBodyFor declaration: some DeclSyntaxProtocol & HasTrailingOptionalCodeBlock,
```
ahoppen, line 23: Why do we need `HasTrailingOptionalCodeBlock` here? If the body macro adds a body, it shouldn’t really care if `declaration` already contains a body.

Tests/SwiftSyntaxMacrosTest/MacroSystemTests.swift
```diff
@@ -726,6 +727,80 @@ public struct DeclsFromStringsMacroNoAttrs: DeclarationMacro {
...
+    providingPreambleFor declaration: some DeclSyntaxProtocol & HasTrailingOptionalCodeBlock,
+    in context: some MacroExpansionContext
+  ) throws -> [CodeBlockItemSyntax] {
+    // FIXME: Should be able to support (de-)initializers and accessors as
+    // well, but this is a lazy implementation.
+    guard let funcDecl = declaration.as(FunctionDeclSyntax.self) else {
+      return []
+    }
+
+    let funcBaseName = funcDecl.identifier.text
```
ahoppen, line 742: If you drop `.text` here, you don’t need to do `raw:` below.

Tests/SwiftSyntaxMacrosTest/MacroSystemTests.swift
```diff
@@ -726,6 +727,80 @@ public struct DeclsFromStringsMacroNoAttrs: DeclarationMacro {
...
+    let paramNames = funcDecl.signature.input.parameterList.map { param in
+      if let name = param.parameterName {
+        return name.text
+      } else {
+        return "_"
+      }
+    }
+
+    let passedArgs = paramNames.map { "\($0): \\(\($0))" }.joined()
```
ahoppen, line 751: I guess you want `.joined(separator: ", ")` here, right?

Tests/SwiftSyntaxMacrosTest/MacroSystemTests.swift
```diff
@@ -726,6 +727,80 @@ public struct DeclsFromStringsMacroNoAttrs: DeclarationMacro {
...
+
+    let entry: CodeBlockItemSyntax = """
+      log("Entering \(raw: funcBaseName)(\(raw: passedArgs))")
+      """
+
+    let argLabels = paramNames.map { "\($0):" }.joined()
```
ahoppen, line 757: Similarly here, should probably be `joined(separator: ", ")`

Tests/SwiftSyntaxMacrosTest/MacroSystemTests.swift
```diff
@@ -726,6 +727,80 @@ public struct DeclsFromStringsMacroNoAttrs: DeclarationMacro {
...
+
+    let exit: CodeBlockItemSyntax = """
+      log("Exiting \(raw: funcBaseName)(\(raw: argLabels))")
+      """
```
ahoppen, line 761: Any reason for not constructing the `defer` statement here directly?

```suggestion
    let exit: CodeBlockItemSyntax = """
      defer {
        log("Exiting \(raw: funcBaseName)(\(raw: argLabels))")
      }
      """
```

Tests/SwiftSyntaxMacrosTest/MacroSystemTests.swift
```diff
@@ -726,6 +727,80 @@ public struct DeclsFromStringsMacroNoAttrs: DeclarationMacro {
...
+    let funcBaseName = funcDecl.identifier.text
+    let paramNames = funcDecl.signature.input.parameterList.map { param in
+      if let name = param.parameterName {
+        return name.text
+      } else {
+        return "_"
+      }
+    }
+
+    let passedArgs = paramNames.map { "\"\($0)\": \($0)" }.joined(separator: ", ")
```
ahoppen, line 795: What do you think about constructing the dictionary using SwiftSyntaxBuilder? It has a little less escaping that way and is easier to read IMO. To do this, `paramNames` should be an array of `TokenSyntax` instead of array of strings.

```swift
let d = DictionaryExprSyntax(content: .elements(DictionaryElementListSyntax {
  for paramName in paramNames {
    DictionaryElementSyntax(
      keyExpression: ExprSyntax("\(literal: paramName.text)"),
      valueExpression: IdentifierExprSyntax(identifier: paramName)
    )
  }
}))
```

Tests/SwiftSyntaxMacrosTest/MacroSystemTests.swift
```diff
@@ -1236,4 +1313,44 @@ final class MacroSystemTests: XCTestCase {
...
+      }
+      """,
+      expandedSource: """
+
+        func doSomethingDangerous(operation: String) {
+          log("Entering doSomethingDangerous(operation: \\(operation))")
+          defer {
+            log("Exiting doSomethingDangerous(operation:)")
+          }
+            print("This is the operation: \\(operation)")
```
ahoppen, line 1332: Do you know why BasicFormat indents this `print` by four spaces here but everything else with two spaces?

ahoppen (commented):
Could you add test cases for the following
- Adding an body macro to a function that already has a body -> this should error IMO
- Add both a body and a preamble macro to a function body -> This should expand the body and add the preamble IMO.

Sources/SwiftSyntaxMacros/MacroProtocols/BodyMacro.swift
DougGregor: If the declaration already contains a body, the macro is likely going to want to incorporate that body into the new body it generates. For example, wrapping the body as written into a closure that's passed to some other function that does some additional work around the call to the closure.

mackoj:
Hi, Will this feature land in a future macro evolution ?

Sources/SwiftSyntaxMacroExpansion/MacroExpansion.swift
DougGregor: Turns out that we have `WithCodeBlock`, so I'll use that and add `WithOptionalCodeBlock`

DougGregor: https://github.com/apple/swift-syntax/pull/2359

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
DougGregor: Yes, I like that

Tests/SwiftSyntaxMacrosTest/MacroSystemTests.swift
DougGregor: Yes, good catch

Tests/SwiftSyntaxMacrosTest/MacroSystemTests.swift
DougGregor: Good idea

Tests/SwiftSyntaxMacrosTest/MacroSystemTests.swift
DougGregor: We're forming the full DeclName string here, so no

Tests/SwiftSyntaxMacrosTest/MacroSystemTests.swift
DougGregor: I liked having entry/exit called out separately, with `defer` as the way of introducing the exit.

Tests/SwiftSyntaxMacrosTest/MacroSystemTests.swift
DougGregor: I tend to like my string interpolations, but... sure. I can do this.

Tests/SwiftSyntaxMacrosTest/MacroSystemTests.swift
DougGregor: I think it wasn't getting formatted at the end. I've reworked this and formatting is more reasonable now.

DougGregor:
@swift-ci please test

Sources/SwiftSyntaxMacroExpansion/MacroExpansion.swift
```diff
@@ -38,18 +40,23 @@ extension MacroRole {
...
-private enum MacroExpansionError: Error, CustomStringConvertible {
+enum MacroExpansionError: Error, CustomStringConvertible {
   case unmatchedMacroRole(Macro.Type, MacroRole)
   case parentDeclGroupNil
   case declarationNotDeclGroup
   case declarationNotIdentified
+  case declarationHasNoBody
   case noExtendedTypeSyntax
   case noFreestandingMacroRoles(Macro.Type)
+  case tooManyBodyMacros
```
ahoppen, line 58: I would prefer `moreThanOneBodyMacro`. `tooManyBodyMacros` makes me wonder: How many is too many? 2, 3, 20?

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
```diff
@@ -584,6 +656,68 @@ private class MacroApplication<Context: MacroExpansionContext>: SyntaxRewriter {
...
+    }
+
+    // If there's no preamble, swap in the new body.
+    if preamble.isEmpty {
+      return node.with(\.body, body)
+    }
+
+    var statements = body.statements
+    statements.insert(contentsOf: preamble, at: statements.startIndex)
+    return node.with(\.body, body.with(\.statements, statements))
```
ahoppen, line 718: Can’t you simplify this to

```suggestion
    var node = node
    node.body.statements = preamble + body.statements
    return node
```

Sources/SwiftSyntaxMacros/MacroProtocols/BodyMacro.swift
```diff
@@ -0,0 +1,25 @@
...
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
+/// Describes a macro that can create the body for a function that does not
+/// have one.
+public protocol BodyMacro: AttachedMacro {
```
ahoppen, line 14: Can you mark this as `@_spi(ExperimentalLanguageFeature)`?

Sources/SwiftSyntaxMacros/MacroProtocols/PreambleMacro.swift
```diff
@@ -0,0 +1,27 @@
...
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
+/// Describes a macro that can introduce "preamble" code into an existing
+/// function body.
+public protocol PreambleMacro: AttachedMacro {
```
ahoppen, line 14: Can you mark this as `@_spi(ExperimentalLanguageFeature)`?

Tests/SwiftSyntaxMacroExpansionTest/BodyMacroTests.swift
```diff
@@ -0,0 +1,114 @@
...
+import SwiftSyntaxMacrosTestSupport
+import XCTest
+
+struct RemoteBodyMacro: BodyMacro {
+  static func expansion(
+    of node: AttributeSyntax,
+    providingBodyFor declaration: some DeclSyntaxProtocol & WithOptionalCodeBlockSyntax,
+    in context: some MacroExpansionContext
+  ) throws -> [CodeBlockItemSyntax] {
+    // FIXME: Should be able to support (de-)initializers and accessors as
```
ahoppen, line 34: How would a body macro on an accessor look like in source? Since AFAICT you can’t attach attributes to accessors.

(same for the comment in PreambleMacroTests.swift)

Sources/SwiftSyntaxMacros/MacroProtocols/BodyMacro.swift
```diff
@@ -0,0 +1,25 @@
...
+  /// Expand a macro described by the given custom attribute and
+  /// attached to the given declaration and evaluated within a
+  /// particular expansion context.
+  ///
+  /// The macro expansion can introduce a body for the given function.
+  static func expansion(
+    of node: AttributeSyntax,
+    providingBodyFor declaration: some DeclSyntaxProtocol & WithOptionalCodeBlockSyntax,
+    in context: some MacroExpansionContext
+  ) throws -> [CodeBlockItemSyntax]
```
ahoppen, line 24: Wouldn’t it be nicer to return a `CodeBlockItemListSyntax` here instead of `[CodeBlockItemSyntax]`?

(same for `PreambleMacro`)

Tests/SwiftSyntaxMacroExpansionTest/BodyMacroTests.swift
```diff
@@ -0,0 +1,114 @@
...
+
+    return [
+      """
+      return try await remoteCall(function: \(literal: funcBaseName), arguments: \(passedArgs))
+      """
+    ]
+  }
+}
+
+final class BodyMacroTests: XCTestCase {
```
ahoppen, line 70: Could you add a test case for adding an body macro to a function that already has a body?

Tests/SwiftSyntaxMacroExpansionTest/PreambleMacroTests.swift
```diff
@@ -0,0 +1,176 @@
...
+      return []
+    }
+
+    let funcBaseName = funcDecl.name
+    let paramNames = funcDecl.signature.parameterClause.parameters.map { param in
+      if let name = param.parameterName {
+        return name.text
+      } else {
+        return "_"
+      }
```
ahoppen, line 46: I think the entire `if let` ceremony isn’t really needed here.

```suggestion
      param.parameterName?.text ?? "_"
```

Tests/SwiftSyntaxMacroExpansionTest/PreambleMacroTests.swift
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    let passedArgs = paramNames.map { "\($0): \\(\($0))" }.joined(separator: ", ")
+
+    let entry: CodeBlockItemSyntax = """
+      log("Entering \(raw: funcBaseName)(\(raw: passedArgs))")
```
ahoppen, line 52:
```suggestion
      log("Entering \(funcBaseName)(\(raw: passedArgs))")
```

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
```diff
@@ -349,6 +349,71 @@ private func expandExtensionMacro(
...
+      indentationWidth: indentationWidth
+    )
+  else {
+    return []
+  }
+
+  // Match the indentation of the statements if we can, and put newlines around
+  // the preamble to separate it from the rest of the body.
+  let indentation = decl.body?.statements.indentationOfFirstLine ?? (decl.indentationOfFirstLine + indentationWidth)
+  let indentedSource = "\n" + expanded.indented(by: indentation) + "\n\n\n"
```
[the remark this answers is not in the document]
DougGregor: Eh, sure.

Tests/SwiftSyntaxMacroExpansionTest/BodyMacroTests.swift
DougGregor: Sure you can:

```swift
var something: Int {
  get { 5 }

  @available(*, deprecated, message: "Stop changing this value")
  set { }
}


print(something)
something = 17
```

DougGregor: I can improve this test macro to handle the other cases.

Sources/SwiftSyntaxMacros/MacroProtocols/BodyMacro.swift
DougGregor: The precent from other macros is to return an array rather than the `*ListSyntax` node.

Tests/SwiftSyntaxMacroExpansionTest/BodyMacroTests.swift
DougGregor: Sure

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
DougGregor: `body` is optional, which makes this not directly work. But I can simplify it with `preamble + body.statements`

Tests/SwiftSyntaxMacroExpansionTest/PreambleMacroTests.swift
DougGregor: Sure, that's cleaner

Tests/SwiftSyntaxMacroExpansionTest/BodyMacroTests.swift
ahoppen: Oh, I didn’t know that you could. Interesting.

Sources/SwiftSyntaxMacros/MacroProtocols/BodyMacro.swift
ahoppen: Is it, though? The only macro that I see that returns `[CodeBlockItemSyntax]` is the code item macro and that hasn’t been accepted, AFAICT, so it’s not really a precedence. And all the other macros return `[DeclSyntax]` or something more special that doesn’t have a corresponding list node in SwiftSyntax

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
ahoppen: Ah, I see. At least we can use `+` though 😉

Sources/SwiftSyntaxMacros/MacroProtocols/BodyMacro.swift
DougGregor: `MemberMacro` returns `[DeclSyntax]` instead of `MemberBlockItemSyntax`, for example.

ahoppen: Well, they aren’t quite the same since `MemberBlockItemSyntax` can also contain a semicolon. 

And yes, you could make an argument that those should in theory return `MemberBlockItemListSyntax` as well.

DougGregor:
https://github.com/apple/swift/pull/70034

@swift-ci please test

DougGregor:
https://github.com/apple/swift/pull/70034

@swift-ci please test Windows

Sources/SwiftSyntaxMacros/MacroProtocols/BodyMacro.swift
DougGregor: Let's chat either in the pitch thread or the review thread (when function body macros come up for a review).

PR #1881, opened 2023-07-06
Detect `attached` attributes using a contextual keyword instead of string matching

ahoppen:
We don’t want to do any string-matching based on token texts in SwiftParser. The correct design here is to add `attached` as a contextual keyword.


hborla (approved):
Thank you!

ahoppen:
@swift-ci Please test

Issue #1883, opened 2023-07-06
`logStringInterpolationParsingError()` fails to compile when building for swiftUI previews


zoltan-lippai:
### Description

The `SyntaxParseable+ExpressibleByStringInterpolation.swift` file's `logStringInterpolationParsingError()` method does not compile when building a project for swiftUI previews for iOS.

The error messages read:
```
... /SourcePackages/checkouts/swift-syntax/Sources/SwiftSyntaxBuilder/SyntaxParsable+ExpressibleByStringInterpolation.swift:33:7 'Logger' is only available in iOS 14.0 or newer
... /SourcePackages/checkouts/swift-syntax/Sources/SwiftSyntaxBuilder/SyntaxParsable+ExpressibleByStringInterpolation.swift:34:9 'OSLogMessage' is only available in iOS 14.0 or newer
... /SourcePackages/checkouts/swift-syntax/Sources/SwiftSyntaxBuilder/SyntaxParsable+ExpressibleByStringInterpolation.swift:35:21 'appendInterpolation(_:align:privacy:)' is only available in iOS 14.0 or newer
... /SourcePackages/checkouts/swift-syntax/Sources/SwiftSyntaxBuilder/SyntaxParsable+ExpressibleByStringInterpolation.swift:37:10 'appendInterpolation(_:align:privacy:)' is only available in iOS 14.0 or newer
```

I've noticed this error today when the package resolution for my project automatically updated the swift-syntax package to the version tagged `509.0.0-swift-DEVELOPMENT-SNAPSHOT-2023-07-04-a`.

### Steps to Reproduce

1. Start a new iOS app project with Xcode 15 beta 2 (15A5161b).

2. If necessary, replace the new `#Preview { ContentView() }` swiftUI preview template with the legacy preview provider solution.

3. Add the swift-syntax package to the project that includes the `SyntaxParseable+ExpressibleByStringInterpolation.swift` file - you can use any version that includes the commit `96557f7`, including today's main branch.

4. Try to launch the swiftUI preview of the `ContentView.swift` file. (Clean the project if necessary to trigger the preview build.)

Result: The preview build fails with the error messages above.

Building the project normally for simulator or device compiles without error. The error message about iOS version is only received when building for swiftUI previews.

ahoppen:
Tracked in Apple’s issue tracker as rdar://111865186

ahoppen:
Thanks for reporting this. I’m fixing it in https://github.com/apple/swift-syntax/pull/1884.

PR #1884, opened 2023-07-07
Guard `OSLog` calls on all platforms, not just macOS

ahoppen:
We only guarded the usage of `Logger` by an OS version on macOS, not on other Apple platforms. Add those platforms the `#available` clause as well.

Fixes https://github.com/apple/swift-syntax/issues/1883 
rdar://111865186


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #1887, opened 2023-07-07
Plugin should keep handling the conformance macro role

bnbarham:
New compilers will never send conformance as a role (only extension), but we should continue supporting conformance until support for those has dropped.

Resolves rdar://111911487.


bnbarham:
@swift-ci please test

bnbarham:
@swift-ci please test

rintaro (approved):

rintaro:
>  until support for those has dropped.

To clarify, "those" here means non-new compilers, not `conformance` macro, right?

bnbarham:
> To clarify, "those" here means non-new compilers, not conformance macro, right?

Yep

PR #1895, opened 2023-07-10
[Parser] Accept 'self' after 'each'

rintaro:
Also move 'repeat', 'each', and 'any' expression parsing to 'parseSequenceExpressionElement'.

rdar://107450487

### DIFF
--- a/Sources/SwiftParser/Expressions.swift
+++ b/Sources/SwiftParser/Expressions.swift
@@ -525,6 +525,46 @@ extension Parser {
           arena: self.arena
         )
       )
+
+    case (.repeat, let handle)?:
+      // 'repeat' is the start of a pack expansion expression.
+      return RawExprSyntax(parsePackExpansionExpr(repeatHandle: handle, flavor, pattern: pattern))
+
+    case (.each, let handle)?:
+      // `each` is only contextually a keyword, if it's followed by an
+      // identifier or 'self' on the same line. We do this to ensure that we do
+      // not break any 'each' functions defined by users. This is following with
+      // what we have done for the consume keyword.
+      switch self.peek() {
+      case TokenSpec(.identifier, allowAtStartOfLine: false),
+        TokenSpec(.dollarIdentifier, allowAtStartOfLine: false),
+        TokenSpec(.self, allowAtStartOfLine: false):
+        break
+      default:
+        // Break out of `outer switch` on failure.
+        break EXPR_PREFIX
+      }
+
+      let each = self.eat(handle)
+      let packRef = self.parseSequenceExpressionElement(flavor, pattern: pattern)
+      return RawExprSyntax(
+        RawPackElementExprSyntax(
+          eachKeyword: each,
+          packRefExpr: packRef,
+          arena: self.arena
+        )
+      )
+
+    case (.any, _)?:
+      // `any` is only contextually a keyword if it's followed by an identifier
+      // on the same line.
+      guard case TokenSpec(.identifier, allowAtStartOfLine: false) = self.peek() else {
+        break EXPR_PREFIX
+      }
+      // 'any' is parsed as a part of 'type'.
+      let type = self.parseType()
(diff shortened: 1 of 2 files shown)


rintaro:
https://github.com/apple/swift/pull/67206
@swift-ci Please test

Sources/SwiftParser/Expressions.swift
```diff
@@ -525,6 +525,46 @@ extension Parser {
...
         )
       )
+
+    case (.repeat, let handle)?:
+      // 'repeat' is the start of a pack expansion expression.
+      return RawExprSyntax(parsePackExpansionExpr(repeatHandle: handle, flavor, pattern: pattern))
+
+    case (.each, let handle)?:
+      // `each` is only contextually a keyword, if it's followed by an
+      // identifier or keyword on the same line. We do this to ensure that we do
```
ahoppen, line 535:
```suggestion
      // identifier or 'self' on the same line. We do this to ensure that we do
```

Sources/SwiftParser/Expressions.swift
```diff
@@ -525,6 +525,46 @@ extension Parser {
...
+      // not break any copy functions defined by users. This is following with
```
ahoppen, line 536:
```suggestion
      // not break any 'each' functions defined by users. This is following with
```

Sources/SwiftParser/Expressions.swift
```diff
@@ -525,6 +525,46 @@ extension Parser {
...
+          eachKeyword: each,
+          packRefExpr: packRef,
+          arena: self.arena
+        )
+      )
+
+    case (.any, _)?:
+      // `any` is only contextually a keyword if it's followed by an identifier
+      // on the same line.
+      if (self.peek().rawTokenKind != .identifier || self.peek().isAtStartOfLine) {
```
ahoppen, line 561: I would write this in terms of `TokenSpec`:

```suggestion
      if self.peek().rawTokenKind != .identifier || self.peek().isAtStartOfLine {
```

Sources/SwiftParser/Expressions.swift
```diff
@@ -525,6 +525,46 @@ extension Parser {
...
+        break EXPR_PREFIX
+      }
+      // 'any' is parsed as a part of 'ty'.
+      let ty = self.parseType()
```
ahoppen, line 565: Nitpick. Let’s name the variable `type`. No reason to save the two characters here.

ahoppen (approved):
This is much nicer. Thanks. Three minor nitpicks.

Sources/SwiftParser/Expressions.swift
```diff
@@ -525,6 +525,46 @@ extension Parser {
           arena: self.arena
         )
       )
+
+    case (.repeat, let handle)?:
+      // 'repeat' is the start of a pack expansion expression.
+      return RawExprSyntax(parsePackExpansionExpr(repeatHandle: handle, flavor, pattern: pattern))
```
ahoppen, line 531: I just noticed that we already never parse `repeat` as a function call in the following (not a regression introduced by this PR, this just made me realize it…

```swift
func repeat() {}
repeat()
```

Sources/SwiftParser/Expressions.swift
rintaro: Changed to 
```swift
      guard case TokenSpec(.identifier, allowAtStartOfLine: false) = self.peek() else {
        break EXPR_PREFIX
      }
```

rintaro:
https://github.com/apple/swift/pull/67206
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

Sources/SwiftParser/Expressions.swift
ahoppen: Ah, thanks for reminding me. Now that you mention it, I think I left the same comment somewhere else already 🤦🏽

PR #1896, opened 2023-07-11
Further naming improvements for syntax node children

ahoppen:
Add a few more syntax child name validation rules and update child names

- No child should end with `Token`
- All children that refer to a collection should have a plural name
- Children shouldn’t be named `Identifier` -> `Name` is usually the better name
- In fact, no child should even contain the substring `Identifier` -> often it doesn’t provide any additional value
- Don’t use abbreviations in child names like `Args`
- Children shouldn’t end with `Expr`, `Expression`, `Type` etc. -> usually we can just omit these words
- And a couple more hand-picked naming improvements


ahoppen:
https://github.com/apple/swift-format/pull/562

@swift-ci Please test

CodeGeneration/Sources/SyntaxSupport/ExprNodes.swift
```diff
@@ -1341,7 +1353,8 @@ public let EXPR_NODES: [Node] = [
         kind: .token(choices: [.keyword(text: "repeat")])
       ),
       Child(
-        name: "PatternExpr",
+        name: "Pack",
+        deprecatedName: "PatternExpr",
```
ahoppen, line 1357: @hborla I renamed the expression in `PackExpansionExprSyntax` from `patternExpr` to `pack`. Does this make sense to you. `PatternExpr` didn’t make sense to me because the expression after `repeat` isn’t a pattern in the Swift pattern matching sense.

CodeGeneration/Sources/SyntaxSupport/TypeNodes.swift
```diff
@@ -395,7 +400,8 @@ public let TYPE_NODES: [Node] = [
         kind: .token(choices: [.keyword(text: "repeat")])
       ),
       Child(
-        name: "PatternType",
+        name: "Pack",
+        deprecatedName: "PatternType",
```
ahoppen, line 404: @hborla I renamed the underlying type in `PackExpansionTypeSyntax` from `patternType` to `pack`. Does this make sense to you. `PatternType` didn’t make sense to me because the expression after `repeat` isn’t a pattern in the Swift pattern matching sense and it’s not really a pattern here.

ahoppen:
https://github.com/apple/swift-format/pull/562

@swift-ci Please test

ahoppen:
https://github.com/apple/swift-format/pull/562

@swift-ci Please test Windows

ahoppen:
https://github.com/apple/swift-format/pull/562

@swift-ci Please test

ahoppen:
https://github.com/apple/swift-format/pull/562

@swift-ci Please test Windows

kimdv (approved):

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
```diff
@@ -1460,7 +1471,8 @@ public let DECL_NODES: [Node] = [
         documentation: "The `#` sign."
       ),
       Child(
-        name: "Macro",
+        name: "MacroName",
```
bnbarham, line 1474: We use `Name` for most other nodes, any reason to prefer `MacroName` here?

CodeGeneration/Sources/SyntaxSupport/ExprNodes.swift
```diff
@@ -1154,7 +1163,8 @@ public let EXPR_NODES: [Node] = [
         documentation: "The `#` sign."
       ),
       Child(
-        name: "Macro",
+        name: "MacroName",
+        deprecatedName: "Macro",
```
bnbarham, line 1167: Same as the decl case, ie. `MacroName` vs `Name`

CodeGeneration/Sources/SyntaxSupport/ExprNodes.swift
```diff
@@ -1398,7 +1411,8 @@ public let EXPR_NODES: [Node] = [
         isOptional: true
       ),
       Child(
-        name: "PostfixExpression",
+        name: "BaseExpression",
```
bnbarham, line 1414: Keeping `Expression` here on purpose?

CodeGeneration/Tests/ValidateSyntaxNodes/ValidateSyntaxNodes.swift
```diff
@@ -497,16 +506,190 @@ class ValidateSyntaxNodes: XCTestCase {
...
+    assertFailuresMatchXFails(
+      failures,
+      expectedFailures: [
+        // The child is singular here, the path just consists of multiple components
+        ValidationFailure(node: .importDecl, message: "child 'Path' is a collection and should thus be named as a plural")
+      ]
+    )
+  }
+
+  /// Identifier is a wonderful ambiguous term. Almost always, 'name' or something similar is more expressive
```
bnbarham, line 598: FWIW I don't see that much difference between name and identifier. Other than name being a little simpler. Sure, in a general sense "identifier" is probably more ambiguous than "name". But in the parser sense not so much. Also all our names are `IdentifierToken` 🤷.

CodeGeneration/Tests/ValidateSyntaxNodes/ValidateSyntaxNodes.swift
```diff
@@ -497,16 +506,190 @@ class ValidateSyntaxNodes: XCTestCase {
...
+      }
+    }
+
+    assertFailuresMatchXFails(
+      failures,
+      expectedFailures: []
+    )
+  }
+
+  func testChildrenDontEndWithExprEtc() {
```
bnbarham, line 649: `testChildrenDontEndWithNodeKind`? Doesn't really matter though (and "Syntax" isn't a kind) 😅

bnbarham (approved):
Nice improvements in general IMO. I do much prefer "s" over "list". "returnClause.returnType" is amusing too, nice catch.

ahoppen:
@swift-ci Please test

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
ahoppen: We use `Name` for nodes that declare something (e.g. `StructDeclSyntax`), while this node refers to something. And I just thought that name alone here is a little ambiguous because it’s not entirely sure what the name is. I don’t have super strong opinions here though.

CodeGeneration/Sources/SyntaxSupport/ExprNodes.swift
ahoppen: Good catch, `Base` is better.

CodeGeneration/Tests/ValidateSyntaxNodes/ValidateSyntaxNodes.swift
ahoppen: For me `Identifier` in the SwiftSyntax context is an identifier token, ie. technically just a word. And that identifier token can server multiple purposes, refer to a value, name a type, be a function parameter label, … 

And wherever we have a clear usage of that identifier token (eg. when it is the name of a nominal type), I think we should name the child by the usage.

CodeGeneration/Tests/ValidateSyntaxNodes/ValidateSyntaxNodes.swift
ahoppen: Yes, `Syntax` isn‘t a kind, children still shouldn’t end with `Syntax` 😉

ahoppen:
https://github.com/apple/swift-format/pull/562

@swift-ci Please test

ahoppen:
https://github.com/apple/swift-format/pull/562

@swift-ci Please test Windows

ahoppen:
https://github.com/apple/swift-format/pull/562

@swift-ci Please test

ahoppen:
https://github.com/apple/swift-format/pull/562

@swift-ci Please test Windows

PR #1897, opened 2023-07-11
Reference swift-syntax from CodeGeneration using path instead of HEAD checkout

ahoppen:
The `HEAD` checkout has been causing more issues than it solved recently. I think it’s better to define `CodeGeneration` to directly reference the parent swift-syntax package and encourage developers of `CodeGeneration` to modify their local `Package.swift` to pull a version of swift-syntax from GitHub when making significant changes.

### DIFF
--- a/CodeGeneration/Package.swift
+++ b/CodeGeneration/Package.swift
@@ -12,7 +12,17 @@ let package = Package(
     .executable(name: "generate-swiftsyntax", targets: ["generate-swiftsyntax"])
   ],
   dependencies: [
-    .package(url: "..", revision: "HEAD")
+    // This directory is a standalone package that uses swift-syntax from the
+    // outer directory.
+    // If you are making significant changs to `CodeGeneration` locally and want
+    // to avoid breaking the build of `CodeGeneration` itself by generating new
+    // files in the parent swift-syntax package, it is encouraged to change the
+    // dependency to the following. That way `CodeGeneration` has its own
+    // checkout of swift-syntax that is unaffected by the newly generated files.
+    // Be sure to revert the change before committing your changes.
+    //
+    // .package(url: "https://github.com/apple/swift-syntax", branch: "main")
+    .package(path: "..")
   ],
   targets: [
     .executableTarget(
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

CodeGeneration/Package.swift
```diff
@@ -12,7 +12,7 @@ let package = Package(
     .executable(name: "generate-swiftsyntax", targets: ["generate-swiftsyntax"])
   ],
   dependencies: [
-    .package(url: "..", revision: "HEAD")
+    .package(path: "..")
```
bnbarham, line 15: My preference would be to move this out and then have an environment variable to control it. Then the README can just have something like "Run `SOME_ENV_VAR=1 swift ..." rather than the "change the dependency from...".

ahoppen: The problem with environment variables is that you can’t control them when opening the project in Xcode.

Matejkob: Maybe this is a stupid idea, but how about leaving a documentation comment in the `Package.swift` file that carries the same information as the README? We could leave commented code with a dependency to the remote repository, something like this:
```swift
  dependencies: [
    .package(path: "..")
//    .package(url: "https://github.com/apple/swift-syntax", branch: "main")
  ],
```
Yes, it's not the most elegant solution, but it's practical. Developers wouldn't have to repeatedly visit the README to copy and paste this code. Considering this is just for `CodeGeneration` SwiftPM, maybe we don't need anything fancy, just something that aids in accelerating development?"

ahoppen: Yes, I think leaving that information in CodeGeneration/Package.swift makes sense to me.

Matejkob: Or if we would prefer to avoid commenting code we could use an enum.
```swift
fileprivate enum SwiftSyntaxDependency {
  case local
  case remote

  fileprivate var package: Package.Dependency {
    switch self {
    case .local:
      return .package(url: "..", revision: "HEAD")
    case .remote:
      return .package(url: "https://github.com/apple/swift-syntax", branch: "main")
    }
  }
}

(...)
  products: [
    .executable(name: "generate-swiftsyntax", targets: ["generate-swiftsyntax"])
  ],
  dependencies: [
    SwiftSyntaxDependency.local.package
  ],
  targets: [
    .executableTarget(
      name: "generate-swiftsyntax",
(...)
```

bnbarham: > The problem with environment variables is that you can’t control them when opening the project in Xcode.

That's true, but also in that case you can still edit the package directly. So I'd still prefer it, even if we keep that in the README. I like @Matejkob suggestions, but I would still control it with an environment variable rather than hardcode to local/remote.

ahoppen:
@swift-ci Please test

PR #1899, opened 2023-07-12
Replace `offset(of: )` with `getOffsetToStart`

StevenWong12:
`LexemeSequence.offset(of: )` and `LexemeSequence.getOffsetToStart` perform the same function and this should make the usage more consistent.

### DIFF
--- a/Sources/SwiftParser/Parser.swift
+++ b/Sources/SwiftParser/Parser.swift
@@ -250,11 +250,7 @@ public struct Parser {
   public var alternativeTokenChoices: [Int: [TokenSpec]] = [:]
 
   mutating func recordAlternativeTokenChoice(for lexeme: Lexer.Lexeme, choices: [TokenSpec]) {
-    guard let lexemeBaseAddress = lexeme.tokenText.baseAddress,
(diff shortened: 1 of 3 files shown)


Sources/SwiftParser/Parser.swift
```diff
@@ -250,11 +250,7 @@ public struct Parser {
   public var alternativeTokenChoices: [Int: [TokenSpec]] = [:]
 
   mutating func recordAlternativeTokenChoice(for lexeme: Lexer.Lexeme, choices: [TokenSpec]) {
-    guard let lexemeBaseAddress = lexeme.tokenText.baseAddress,
-      let offset = lexemes.offset(of: lexemeBaseAddress)
-    else {
-      return
-    }
+    let offset = lexemes.getOffsetToStart(lexeme)
```
ahoppen, line 253: One nitpick since you’re changing this (I think I thought about this before but apparently never left a comment): Could you rename the function to `offsetToStart`. Swift generally doesn’t use the term `get` in function names.

ahoppen (approved):
Thanks

ahoppen:
@swift-ci Please test

PR #1900, opened 2023-07-12
Replace `IncrementalParseReusedNodeDelegate` with `ReusedNodeCallback`

StevenWong12:
Resolves comments in https://github.com/apple/swift-syntax/pull/1685/files#r1254465986

### DIFF
--- a/Sources/SwiftParser/IncrementalParseTransition.swift
+++ b/Sources/SwiftParser/IncrementalParseTransition.swift
@@ -35,7 +35,7 @@ extension Parser {
   }
 }
 
-/// Accepts the re-used ``Syntax`` nodes that `IncrementalParseTransition`
+/// Accepts a re-used ``Syntax`` node that `IncrementalParseTransition`
 /// determined they should be re-used for a parse invocation.
 ///
 /// The client can use this information to potentially avoid unnecessary work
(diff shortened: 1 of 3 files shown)


Sources/SwiftParser/IncrementalParseTransition.swift
```diff
@@ -43,36 +43,15 @@ extension Parser {
...
-{
-  public var nodes: [Syntax] = []
-
-  public init() {}
-
-  public func parserReusedNode(previousNode: Syntax) {
-    nodes.append(previousNode)
-  }
-}
+public typealias ReusedNodeCallback = (_ nodes: Syntax) -> ()
```
ahoppen, line 46: The callback only takes a single node.

```suggestion
public typealias ReusedNodeCallback = (_ node: Syntax) -> ()
```

Sources/SwiftParser/IncrementalParseTransition.swift
```diff
@@ -84,12 +63,12 @@ public final class IncrementalParseTransition {
     previousTree: SourceFileSyntax,
     edits: ConcurrentEdits,
     lookaheadRanges: LookaheadRanges,
-    reusedNodeDelegate: IncrementalParseReusedNodeDelegate? = nil
+    reusedNodeCallback: ReusedNodeCallback? = nil
```
ahoppen, line 66: Could you update the documentation as well? Something like

```
For each node that gets re-used `reusedNodeCallback` is called.
```

ahoppen (commented):
So much simpler

ahoppen (approved):

ahoppen:
@swift-ci Please test

PR #1905, opened 2023-07-14
Enable some skipped tests in `IncrementalParsingTests.swift`

StevenWong12:
Since we now enable re-using on `codeBlockItem`, I think we should not skip these test but modify the `reusedNodes` for these test cases.

To let this happen, I also made some tiny changes on `assertIncrementalParse` to make the `ByteSourceRange` correctly match in test cases have same substring in source. e.g.
```swift
func foo() {}
foo()
```

Also make some naming improvements:
* `getEditsAndSources()` -> `extractEditsAndSources(from: source)`
* `getByteSourceRange` -> `byteSourceRange`


Sources/_SwiftSyntaxTestSupport/IncrementalParseTestUtils.swift
```diff
@@ -19,13 +19,20 @@ import XCTest
 /// 1. Round-trip on the incrementally parsed syntax tree.
 /// 2. verify that incrementally parsing the edited source base on the original source produces the same syntax tree as reparsing the post-edit file from scratch.
 /// 3. verify the reused nodes fall into expectations.
+///
+/// - parameter reusedNodes: The element order should respect to the order of  `ReusedNodeSpec.source` in `source`.
+/// e.g. for `source` func foo() {}\n foo()
+///    The `reusedNodes` should be [
+///      ReusedNodeSpec("func foo() {}", kind: .codeBlockItem),
+///      ReusedNodeSpec("foo()", kind: .codeBlockItem)
+///    ]
```
ahoppen, line 28: Nit, just to maintain the indentation of `parameter` which makes it easier to read. Also the source should be in backticks to make it render nicely in docc


```suggestion
///
/// - parameter reusedNodes: The element order should respect to the order of  `ReusedNodeSpec.source` in `source`.
///   e.g. for `source` 
///   ```
///   func foo() {}
///    foo()
///    ```
///    The `reusedNodes` should be 
///    ```
///    [
///      ReusedNodeSpec("func foo() {}", kind: .codeBlockItem),
///      ReusedNodeSpec("foo()", kind: .codeBlockItem)
///    ]
///    ```
```

Sources/_SwiftSyntaxTestSupport/IncrementalParseTestUtils.swift
```diff
@@ -78,13 +85,22 @@ public func assertIncrementalParse(
...
-    guard let reusedNode = reusedNodes.first(where: { $0.byteRangeAfterTrimmingTrivia == range }) else {
+    guard
+      let reusedNode = reusedNodes.first(where: {
+        $0.byteRangeAfterTrimmingTrivia
+          == ByteSourceRange(
+            offset: range.offset + originalString.distance(from: originalString.startIndex, to: lastRangeUpperBound),
+            length: range.length
+          )
+      })
+    else {
```
ahoppen, line 103: Instead of doing all these index manipulations (which seem fairly brittle to me), I would keep everything as is and add a `after` parameter to `byteSourceRange`

```swift
fileprivate func byteSourceRange(for substring: String, in sourceString: String, after: String.Index) -> ByteSourceRange? {
  if let range = sourceString[after...].range(of: substring) {
    return ByteSourceRange(
      offset: sourceString.utf8.distance(from: sourceString.startIndex, to: range.lowerBound),
      length: sourceString.utf8.distance(from: range.lowerBound, to: range.upperBound)
    )
  }
  return nil
}
```

You can then call `byteSourceRange` above as

```swift
byteSourceRange(for: expectedReusedNode.source, in: originalString, after: lastRangeUpperBound)
```

ahoppen (commented):
Woohoo. Fewer disabled tests.

ahoppen (approved):

Issue #1901, opened 2023-07-13
Change `name` property of `Child` in `CodeGeneration` to be lowercase


ahoppen:
It’s a little counter-intuitive that the `name` property of `Child` needs to be uppercase because AFAICT we are always using it as lowercase. Change the argument to be lowercase. The generated code should not change.

ahoppen:
Tracked in Apple’s issue tracker as rdar://112178612

Matejkob:
I love this idea! May I try to develop it?

ahoppen:
Definitely!

PR #1908, opened 2023-07-14
Update code-gen tool for lowercase child name consistency

Matejkob:
This PR introduces changes to the code-gen tool to ensure consistency with already generated code. This is a preparatory step before we start changing the names of all child elements to be lowercase - #1901.

In the #1901 issue description @ahoppen mention that 

> It's a little counter-intuitive that the `name` property of `Child` needs to be uppercase because AFAICT we are always using it as lowercase

After research, I've learned that we're using the `name` property of `Child` as uppercased in the following contexts:
* as enum names e.g.: `enum accessorSpecifierOptions: TokenSpecSet` and them as types e.g.: `ElseBody` for `IfExprSyntax`,
* to build names of "unexpected" nodes.

I guess now with this knowledge the question is: should we still lowercased values of the `name` property of each `Child`?

### DIFF
--- a/CodeGeneration/Sources/SyntaxSupport/Node.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/Node.swift
@@ -114,16 +114,18 @@ public class Node {
       // any two defined children
       childrenWithUnexpected =
         children.enumerated().flatMap { (i, child) -> [Child] in
+          let childName = child.name.withFirstCharacterUppercased
+
           let unexpectedName: String
           let unexpectedDeprecatedName: String?
 
           if i == 0 {
-            unexpectedName = "UnexpectedBefore\(child.name)"
+            unexpectedName = "UnexpectedBefore\(childName)"
             unexpectedDeprecatedName = child.deprecatedName.map { "UnexpectedBefore\($0)" }
           } else {
-            unexpectedName = "UnexpectedBetween\(children[i - 1].name)And\(child.name)"
+            unexpectedName = "UnexpectedBetween\(children[i - 1].name.withFirstCharacterUppercased)And\(childName)"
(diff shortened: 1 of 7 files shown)


Matejkob:
@ahoppen I've drafted this PR to spark a discussion about what to do next with the issue

ahoppen:
I just found out about the usage of capitalized `name` for the unexpected nodes yesterday as well. The way I would go about this that:
- The child names as specified in SyntaxSupport should be lowercased because the name of the property is how we primarily think about the child
- `Child.name` should be private (and will probably store the lowercase name from the `Child` initializer)
- There’s `varName` (which I’m renaming to `varOrCaseName` in https://github.com/apple/swift-syntax/pull/1914 to be consistent with `Node.varOrCaseName`), which should return the lowercase name
- I’d introduce `typeName`, which returns a capitalized version of the child name (same as `Child.name` currently does). We should be able to use that in all the cases where we currently rely on `Child.name` being uppercase.

ahoppen:
Did you close this PR by accident @Matejkob?

Matejkob:
Nope, I didn't; just forgot to leave an explanation. 😅

So, here's why: PR #2011 introduces the last piece proposed by @ahoppen in [the above discussion](https://github.com/apple/swift-syntax/pull/1908#issuecomment-1638266939), allowing us to close #1901. In regard to this discussion, this draft became outdated and there isn't a single line of code in it that should be merged anymore.

After merging #2011, I'll publish a PR with lowercased child names.

ahoppen:
Oh, OK. I didn’t read the title properly and thought this was the PR to lowercase the child names themselves. Thanks for clarifying.

PR #1910, opened 2023-07-17
Simplify consumption of prefixes in a token

ahoppen:
In essence, introduce `at(prefix:)` and `consume(ifPrefix:)` and use them in the codebase.

I measured performance and there’s no performance regression associated with this change (based on parsing MovieSwiftUI).

### DIFF
--- a/Sources/SwiftParser/Nominals.swift
+++ b/Sources/SwiftParser/Nominals.swift
@@ -233,7 +233,7 @@ extension Parser {
     }
 
     let primaryOrGenerics: T.PrimaryOrGenerics?
-    if self.currentToken.starts(with: "<") {
+    if self.at(prefix: "<") {
       primaryOrGenerics = T.parsePrimaryOrGenerics(&self)
     } else {
       primaryOrGenerics = nil
@@ -352,18 +352,10 @@ extension Parser {
         )
       } while keepGoing != nil && loopProgress.evaluate(currentToken)
     }
-    let unexpectedBeforeRangle: RawUnexpectedNodesSyntax?
-    let rangle: RawTokenSyntax
-    if self.currentToken.starts(with: ">") {
-      unexpectedBeforeRangle = nil
-      rangle = self.consumePrefix(">", as: .rightAngle)
-    } else {
-      (unexpectedBeforeRangle, rangle) = self.expect(.rightAngle)
-    }
+    let rangle = self.expectWithoutRecovery(prefix: ">", as: .rightAngle)
     return RawPrimaryAssociatedTypeClauseSyntax(
       leftAngle: langle,
       primaryAssociatedTypeList: RawPrimaryAssociatedTypeListSyntax(elements: associatedTypes, arena: self.arena),
-      unexpectedBeforeRangle,
       rightAngle: rangle,
       arena: self.arena
     )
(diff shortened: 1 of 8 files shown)


ahoppen:
@swift-ci Please test

Sources/SwiftParser/Nominals.swift
```diff
@@ -352,18 +352,10 @@ extension Parser {
...
     }
-    let unexpectedBeforeRangle: RawUnexpectedNodesSyntax?
-    let rangle: RawTokenSyntax
-    if self.currentToken.starts(with: ">") {
-      unexpectedBeforeRangle = nil
-      rangle = self.consumePrefix(">", as: .rightAngle)
-    } else {
-      (unexpectedBeforeRangle, rangle) = self.expect(.rightAngle)
-    }
+    let rangle = self.expectWithoutRecovery(prefix: ">", as: .rightAngle)
```
bnbarham, line 355: Did you remove the recovery because you found it didn't actually work as expected?

I was expecting us to parse `protocol Test<Foo Bar> {}` mostly properly, ie. with just `Bar` as unexpected. But that doesn't happen as the `>` in recovery is parsed as a postfix operator (not a right angle) and thus what actually happens is that we skip up to the `{`. Did you decide it wasn't worth handling that case?

Sources/SwiftParser/TokenConsumer.swift
```diff
@@ -23,6 +23,13 @@ protocol TokenConsumer {
   /// Consume the current token and change its token kind to `remappedTokenKind`.
   mutating func consumeAnyToken(remapping remappedTokenKind: RawTokenKind) -> Token
```
bnbarham, line 25: Do we ever take a prefix where it's not just the default text of the token? Could we just pass the token in rather than the prefix as well?

ahoppen: Yes, there’s

https://github.com/apple/swift-syntax/blob/e990c2a4b6f1e00ff000ec06fd672d8d224a5753/Sources/SwiftParser/Availability.swift#L288

and

https://github.com/apple/swift-syntax/blob/e990c2a4b6f1e00ff000ec06fd672d8d224a5753/Sources/SwiftParser/Declarations.swift#L1115

Good idea though, I added a `precondition` to check that `prefix` is `tokenKind.defaultText` if that `tokenKind` has a `defaultText`.

Sources/SwiftParser/Nominals.swift
ahoppen: Yes, I couldn’t find any examples where the recovery kicked in. 

We certainly could have `expect(prefix:as:)` but it’s implementation is non-trivial because non of the current recovery infrastructure supports recovery a to somewhere inside a token. I filed https://github.com/apple/swift-syntax/issues/1923 and maybe we can implement it in a follow-up PR.

Sources/SwiftParser/TokenConsumer.swift
bnbarham: If they're the minority, could still be worth having a `consumePrefix` that only takes the token kind instead. Maybe this is more clear as to what's happening though 🤷

Sources/SwiftParser/Nominals.swift
ahoppen: That’s happening everywhere we currently expect a prefix. It would be nice to fix at some point.

Sources/SwiftParser/TokenConsumer.swift
ahoppen: The problem is that this method doesn’t have a clearly defined meaning if `defaultText == nil`. I prefer being explicit here and passing the prefix.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #1911, opened 2023-07-17
Simplify checking if a token can be an argument label 

ahoppen:
Mostly, change `self.currentToken.canBeArgumentLabel` to `self.atArgumentLabel`.

Also some other miscellaneous smaller improvements in the parser.

### DIFF
--- a/Sources/SwiftParser/Names.swift
+++ b/Sources/SwiftParser/Names.swift
@@ -24,7 +24,7 @@ extension Parser {
   }
 
   mutating func parseArgumentLabel() -> (RawUnexpectedNodesSyntax?, RawTokenSyntax) {
-    guard self.currentToken.canBeArgumentLabel(allowDollarIdentifier: true) else {
+    guard self.atArgumentLabel(allowDollarIdentifier: true) else {
       return (nil, missingToken(.identifier))
     }
     if let dollarIdent = self.consume(if: .dollarIdentifier) {
@@ -98,7 +98,7 @@ extension Parser {
     // A close parenthesis, if empty lists are allowed.
     let nextIsRParen = flags.contains(.zeroArgCompoundNames) && next.rawTokenKind == .rightParen
     // An argument label.
-    let nextIsArgLabel = next.canBeArgumentLabel() || next.rawTokenKind == .colon
+    let nextIsArgLabel = next.isArgumentLabel() || next.rawTokenKind == .colon
 
     guard nextIsRParen || nextIsArgLabel else {
       return nil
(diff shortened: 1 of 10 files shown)


Sources/SwiftParser/Names.swift
```diff
@@ -261,18 +261,16 @@ extension Parser.Lookahead {
...
-    switch self.rawTokenKind {
+  func isArgumentLabel(allowDollarIdentifier: Bool = false) -> Bool {
+    switch self {
     case .identifier, .wildcard:
       // Identifiers, escaped identifiers, and '_' can be argument labels.
       return true
     case .dollarIdentifier:
       return allowDollarIdentifier
+    case .keyword(.inout):
+      // `inout` is reserved as an argument label for historical reasons.
```
bnbarham, line 272: I realize this is copied from above, but wouldn't this make it *not* reserved, ie. `inout` cannot be an argument label?

Sources/SwiftParser/Types.swift
```diff
@@ -235,19 +235,50 @@ extension Parser {
   mutating func parseSimpleType(
     stopAtFirstPeriod: Bool = false
   ) -> RawTypeSyntax {
+    enum TypeBaseStart: TokenSpecSet {
```
bnbarham, line 238: So much boilerplate :(

bnbarham (approved):

Sources/SwiftParser/Names.swift
ahoppen: Yes. Updated

Sources/SwiftParser/Types.swift
```diff
@@ -235,19 +235,50 @@ extension Parser {
...
+        switch self {
+        case .Self: return .keyword(.Self)
+        case .Any: return .keyword(.Any)
+        case .identifier: return .identifier
+        case .leftParen: return .leftParen
+        case .leftSquare: return .leftSquare
+        case .wildcard: return .wildcard
+        }
+      }
+
```
kimdv, line 268: Nitpick: newline 😬

Sources/SwiftParser/Types.swift
kimdv: Maybe a stupid question: is it possible to use one of the generated ones?

kimdv (approved):

Sources/SwiftParser/Types.swift
ahoppen: I don’t think there’s any one that has these cases because we do walk into different parsing functions based on the token, similar to `ExpressionModifierKeyword`.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

kimdv (approved):

PR #1912, opened 2023-07-17
Audit public API on `SyntaxProtocol`

ahoppen:
First, re-arrange the properties in `SyntaxProtocol` to group them logically.

Then, I found two areas of improvement:
- Deprecated `isImplicit`
  - This property seems to be pretty useless and probably dates back to a completely different design of the syntax tree where implicit nodes were a thing.
- Clean up position/range API on `SyntaxProtocol`
  - The position and range API on `SyntaxProtocol` was a little inconsistent and misleading. Clean it up so that we have `totalLength` (including trivia), `contentLength` (excluding trivia) and a matching pair for `byteRange`.

### DIFF
--- a/Sources/SwiftSyntax/Syntax.swift
+++ b/Sources/SwiftSyntax/Syntax.swift
@@ -477,14 +477,9 @@ public extension SyntaxProtocol {
     return data.endPosition
   }
 
-  /// The textual byte length of this node including leading and trailing trivia.
-  var byteSize: Int {
-    return totalLength.utf8Length
-  }
-
-  /// The byte source range of this node including leading and trailing trivia.
-  var byteRange: ByteSourceRange {
-    return ByteSourceRange(offset: position.utf8Offset, length: byteSize)
+  /// The length of this node including all of its trivia.
+  var totalLength: SourceLength {
+    return raw.totalLength
   }
 
   /// The length this node takes up spelled out in the source, excluding its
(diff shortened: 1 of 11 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/SwiftSyntax/Syntax.swift
```diff
@@ -454,6 +488,40 @@ public extension SyntaxProtocol {
...
+  }
+
+  @available(*, deprecated, renamed: "totalByteRange")
+  var byteRange: ByteSourceRange {
+    return ByteSourceRange(offset: position.utf8Offset, length: totalLength.utf8Length)
+  }
+
+  /// The textual byte length of this node including leading and trailing trivia.
+  @available(*, deprecated, message: "Use totalLength.utf8Length")
+  var byteSize: Int {
```
bnbarham, line 511: FWIW I've always found `SourceLength` and `utf8Length` very confusing. It's really `numUTF8CodeUnits` or something similar. For some reason `utf8Length` makes my brain think "length of the string when the given bytes are decoded as utf8". Maybe it's just the `length` that does that to me. If no one else has this problem I'm happy to drop it. I figured I'd bring it up since you're renaming here anyway.

Sources/SwiftSyntax/Syntax.swift
```diff
@@ -454,6 +488,40 @@ public extension SyntaxProtocol {
...
+    return totalLength.utf8Length
+  }
+
+  /// The textual byte length of this node exluding leading and trailing trivia.
```
bnbarham, line 515:
```suggestion
  /// The textual byte length of this node excluding leading and trailing trivia.
```

Sources/SwiftSyntax/Syntax.swift
```diff
@@ -210,29 +228,21 @@ public extension SyntaxProtocol {
...
-  /// Access the raw syntax assuming the node is a Syntax.
-  @_spi(RawSyntax)
-  public var raw: RawSyntax {
-    return _syntaxNode.data.raw
-  }
 
   /// Return this subtree with this node as the root, ie. detach this node
   /// from its parent.
-  public var detached: Self {
+  var detached: Self {
```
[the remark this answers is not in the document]
ahoppen: Let’s enable the swift-format rule that we don’t have an access level on an extension in a follow-up PR.

bnbarham (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #1915, opened 2023-07-17
Don’t generate BasicFormat+Extensions.swift and SyntaxClassifcation.swift

ahoppen:
Instead of generating these files, we can just define them directly.

Also: Remove a couple of unused properties in `TokenSpec`.


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/SwiftBasicFormat/BasicFormat.swift
```diff
@@ -164,7 +164,43 @@ open class BasicFormat: SyntaxRewriter {
 
   /// Whether a leading newline on `token` should be added.
   open func requiresIndent(_ node: some SyntaxProtocol) -> Bool {
-    return node.requiresIndent
```
bnbarham, line 167: The nice thing about this being generated is that when a node is added, `isIndented` is right in the constructor and would *hopefully* have some thought put into it. Where as now you have to know to also update `BasicFormat`. Maybe that's fine (it's a similar case for the whitespace now), but seems worth mentioning.

bnbarham (approved):

Sources/SwiftBasicFormat/BasicFormat.swift
ahoppen: That’s a price I’m willing to pay for the reduced complexity. And I don’t really expect many new nodes that require indentation anyway 😉

PR #1916, opened 2023-07-17
Add `assertClassification` for `ClassificationTests`

StevenWong12:
I made use of some infrastructure we have in `IncrementalParseTestUtils.swift` to add a `assertClassification` function to make `ClassificationTests` look nicer.

This could also be a start effort to translate the `coloring_*` `lit`-based test cases into XCTest.

(This pr is based on https://github.com/apple/swift-syntax/pull/1905)

### DIFF
--- a/Tests/SwiftIDEUtilsTest/Assertions.swift
+++ b/Tests/SwiftIDEUtilsTest/Assertions.swift
@@ -1,0 +1,99 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import XCTest
+import SwiftIDEUtils
+import SwiftParser
+import SwiftSyntax
+import _SwiftSyntaxTestSupport
+
+/// Parse `source` and checks its `classifications` is the same as `expected`.
+///
+/// The `expected` classifications should only contain classifications that are not `none`. All uncovered ranges are expected to have no classification.
+///
+/// - Parameters:
+///   - range: An optional parameter to specify the ``ByteSourceRange`` in `source` that we want to test the `classifications` in.
+///   - expected: The element order should respect to the order of  `ClassificationSpec.source` in `source`.
+func assertClassification(
+  _ source: String,
+  in range: ByteSourceRange? = nil,
+  expected: [ClassificationSpec],
+  file: StaticString = #file,
+  line: UInt = #line
+) {
+  let tree = Parser.parse(source: source)
+
+  var classifications: Array<SyntaxClassifiedRange>
+  if let range {
+    classifications = Array(tree.classifications(in: range))
+  } else {
+    classifications = Array(tree.classifications)
+  }
+  classifications = classifications.filter { $0.kind != .none }
+
+  if expected.count != classifications.count {
+    XCTFail("Expected \(expected.count) re-used nodes but received \(classifications.count)", file: file, line: line)
+  }
+
+  var lastRangeUpperBound = source.startIndex
+  for (classification, spec) in zip(classifications, expected) {
+    guard let range = byteSourceRange(for: spec.source, in: source, after: lastRangeUpperBound) else {
+      XCTFail("Fail to find string in original source,", file: spec.file, line: spec.line)
+      continue
+    }
+
+    XCTAssertEqual(
+      range,
+      classification.range,
+      """
+      Expected \(range) but received \(classification.range)
+      """,
+      file: spec.file,
+      line: spec.line
+    )
+
+    XCTAssertEqual(
+      spec.kind,
+      classification.kind,
+      """
+      Expected \(spec.kind) syntax classification kind but received \(classification.kind)
+      """,
+      file: spec.file,
+      line: spec.line
+    )
+
+    lastRangeUpperBound = source.index(source.startIndex, offsetBy: range.endOffset)
+  }
+}
+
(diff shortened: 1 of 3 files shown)


Tests/SwiftIDEUtilsTest/Assertions.swift
```diff
@@ -0,0 +1,103 @@
...
+
+import XCTest
+import SwiftIDEUtils
+import SwiftParser
+import SwiftSyntax
+import _SwiftSyntaxTestSupport
+
+func assertClassification(
+  _ source: String,
+  expectedClassificationSpec: [ClassificationSpec] = []
```
ahoppen, line 21: I don’t think the default value of `[]` makes sense for any of the `assertClassification` functions. If you use them, you almost certainly want to check that some classifications are produced.

---

What you think of naming this parameter `expected`. I think that would clean up the calls a little.

Tests/SwiftIDEUtilsTest/Assertions.swift
```diff
@@ -0,0 +1,103 @@
...
+}
+
+
+func assertTokensClassification(
+  _ source: String,
+  expectedClassificationSpec: [ClassificationSpec]
+) {
+  let tree = Parser.parse(source: source)
+  assertClassificationImpl(source: source, classifications: tree.tokens(viewMode: .sourceAccurate).map { $0.tokenClassification }.filter { $0.kind != .none }, expectedClassificationSpec: expectedClassificationSpec)
+}
```
ahoppen, line 43: I don’t think this function should exist. We should be using `assertClassification` for all test cases.

Tests/SwiftIDEUtilsTest/Assertions.swift
```diff
@@ -0,0 +1,103 @@
...
+
+func assertClassificationImpl(
+  source: String,
+  classifications: [SyntaxClassifiedRange],
+  expectedClassificationSpec: [ClassificationSpec] = []
+) {
+  let classifications = Array(classifications)
```
ahoppen, line 50: Is this needed?

Tests/SwiftIDEUtilsTest/Assertions.swift
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func assertClassification(
+  _ source: String,
+  in range: ByteSourceRange,
+  expectedClassificationSpec: [ClassificationSpec] = []
+) {
+  let tree = Parser.parse(source: source)
+  assertClassificationImpl(source: source, classifications: tree.classifications.filter { $0.kind != .none && ($0.endOffset > range.offset) && ($0.offset < range.endOffset) }, expectedClassificationSpec: expectedClassificationSpec)
+}
```
ahoppen, line 34: Could we unify this with `assertClassification` above by having `range` be optional and defaulted to `nil`? Then we also wouldn’t need `assertClassificationImpl` anymore if we get rid of `assertTokensClassification`.

Tests/SwiftIDEUtilsTest/ClassificationTests.swift
```diff
@@ -23,177 +23,50 @@ public class ClassificationTests: XCTestCase {
...
-    do {
-      let tok = tree.lastToken(viewMode: .sourceAccurate)!.previousToken(viewMode: .sourceAccurate)!
-      XCTAssertEqual("\(tok)", "Int")
-      let classif = Array(tok.classifications).first!
-      XCTAssertEqual(classif.kind, .typeIdentifier)
-    }
+    
+    assertClassification(
+    source, expectedClassificationSpec: [
+      .init(source: "// blah.", kind: .lineComment),
```
ahoppen, line 29: I generally prefer to spell this as `ClassificaitonSpec`. Then you’re not left wondering what object you’re constructing here while reading the code.

Tests/SwiftIDEUtilsTest/Assertions.swift
StevenWong12: The reason why I have an empty array as default value for classification spec is there are some test cases that gives a certain `ByteSourceRange` and there are only trivia in that range. And our `assertClassification` function filter out trivia.

StevenWong12:
I make the `range` in `assertClassification` is essentially the range we want to test in instead of the range of all tokens that touch this range, which I think more intuitive. What do you think?

ahoppen:
The key point of the classification tests that had a range was to test the `range` parameter of `classification(in:)` so we should continue to pass the range to the `classification` method.

StevenWong12:
Ah, I see. Since that, do you think we should also test the `classification(at: )` function? (which may lead to splitting the `assertClassification` into three functions like before to make it look nicer.)

ahoppen:
I think having a test case that tests `classification(at:)` does make sense. But since I don’t think we need a lot of them (one or two are sufficient), and the test cases are even smaller, I don’t think we’ll need a separate `assert` function for it. We should be able to just check `kind` and `range` manually like we do today.

Tests/SwiftIDEUtilsTest/Assertions.swift
```diff
@@ -0,0 +1,97 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import XCTest
+import SwiftIDEUtils
+import SwiftParser
+import SwiftSyntax
+import _SwiftSyntaxTestSupport
+
+/// Parse `source` and checks its `classifications` is the same as `expected`. The `classification` filter out all the ``Trivia``s in `source` i.e. ``SyntaxClassification`` is `none` to make test cases look nicer.
```
ahoppen, line 19: We aren’t actually filtering out the trivia. We do have classifications for comments. We are just not including any `none` classifications.
```suggestion
/// Parse `source` and checks its `classifications` is the same as `expected`. 
/// 
/// The `expected` classifications should only contain classifications that are not `none`. All uncovered ranges are expected to have no classification.
```

ahoppen (commented):
Two minor comments, otherwise LGTM.

ahoppen (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #1917, opened 2023-07-17
Display actual expanded source, not expected source

JosephDuffy:
While using this function I was confused by the error message. I think it should include the actual expansion, not what's expected?

I also renamed the variable to try and make it a little clearer and (assuming this is correct) prevent a mistake like this again. This was done in 3c46da3712dec11e5d485c96b6b7d9d40b7c8f25 so can be easily reverted.

### DIFF
--- a/Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
+++ b/Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
@@ -297,7 +297,7 @@ public func assertMacroExpansion(
     expandedSource.trimmingCharacters(in: .newlines),
     additionalInfo: """
       Actual expanded source:
-      \(expandedSource)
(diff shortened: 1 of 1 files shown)


ahoppen (approved):
Thank you. What a stupid mistake. Could you also open a second PR targeting the `release/5.9` branch so that we can also get it fixed for the Swift 5.9 release?

ahoppen:
@swift-ci Please test

JosephDuffy:
> Thank you. What a stupid mistake. Could you also open a second PR targeting the `release/5.9` branch so that we can also get it fixed for the Swift 5.9 release?

It was a bit confusing when I saw the message 😅 I'm glad I could fix it 😄

I've opened #1918.

ahoppen:
Thank you @JosephDuffy!

PR #1927, opened 2023-07-19
Don’t add two spaces in front of `{` if a property already has an accessor

ahoppen:
If a property already had an accessor, the type annotation contained a trailing whitespace. We were explicitly adding a leading whitespace to the `{` for the accessors, resulting in two whitespace.

To fix it, don’t explicitly add the leading whitespace to the opening `{` and rely on `BasicFormat` to add it, if needed.

rdar://111215192


ahoppen:
@swift-ci Please test

ahoppen:
Superseded by https://github.com/apple/swift-syntax/pull/1946

PR #1930, opened 2023-07-19
[bazel] Add missing dependency

keith:


keith:
@swift-ci please test

PR #1939, opened 2023-07-23
Use rewrite func instead of visit for SyntaxRewriter

kitasuke:
In my opinion, using `rewrite` func is a bit easier to understand for `SyntaxRewriter`. `visit` func can literally be used to visit for `SyntaxVisitor` instead.


ahoppen (approved):
Thanks, `rewrite` should indeed be used instead of `visit`.

ahoppen:
@swift-ci Please test

PR #1942, opened 2023-07-24
Introduce `peek(isAt:)`

ahoppen:
This makes it a lot easier to check if the next token is of a specific kind.

### DIFF
--- a/Sources/SwiftParser/TokenConsumer.swift
+++ b/Sources/SwiftParser/TokenConsumer.swift
@@ -172,6 +172,57 @@ extension TokenConsumer {
   }
 }
 
+extension TokenConsumer {
+  /// Returns whether the next (peeked) token matches `spec`
+  @inline(__always)
+  mutating func peek(isAt spec: TokenSpec) -> Bool {
+    #if SWIFTPARSER_ENABLE_ALTERNATE_TOKEN_INTROSPECTION
+    if shouldRecordAlternativeTokenChoices {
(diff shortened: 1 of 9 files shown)


Sources/SwiftParser/TokenConsumer.swift
```diff
@@ -172,6 +172,57 @@ extension TokenConsumer {
   }
 }
 
+extension TokenConsumer {
+  /// Returns whether the next (peeked) token matches `spec`
+  @inline(__always)
+  mutating func peek(isAt spec: TokenSpec) -> Bool {
```
bnbarham, line 178: Thoughts on just `at` instead of `isAt`? `next` could work as well. Or maybe `nextAt(_:)` instead of `peek` to match `at(_:)`?

ahoppen: I deliberately decided against `peek(at:)` because we already have `peek(at offset: Int)` in the lexer, which peeks `offset` bytes ahead, which has completely different semantics.

We could consider renaming `peek` to `next` altogether. I’m not sure how much of a term of art `peek` is in parsers in general and whether we should stick to it because of that reason though.

bnbarham: Could also change `peek(at offset: Int)` to something else. You don't need to block the PR on this though, easy enough to rename `peek(isAt:)` if we want to after.

PR #1944, opened 2023-07-24
Fix string interpolation parsing errors in CodeGeneration

ahoppen:
Running CodeGeneration causes syntax errors to be logged using `os_log`. Fix those.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #1947, opened 2023-07-25
Only enable performance testing on Intel machines that run macOS

ahoppen:
Apple Silicon machines produce different instruction counts and thus obviously shouldn’t use the Intel baseline.


ahoppen:
@swift-ci Please test

kimdv (approved):

PR #1948, opened 2023-07-25
Disable performance testing in CI

ahoppen:
It turns out that regular changes to the compiler can cause fluctuations to these test cases that are sufficient to make them trip over so we can’t run them in CI. The test should still be valuable to bisect a performance regression after we find it or to verify that a change doesn’t cause a performance regression.


ahoppen:
@swift-ci Please test

kimdv (approved):

PR #1953, opened 2023-07-27
Convert `lit`-based syntax classification test to XCTest

StevenWong12:
We should use the `assertClassification` to make these test cases look nicer and get rid of the dependency on `FileCheck` eventually.

I converted a subset of the original `lit`-tests since I think most of them are mutually covered. There are some tests not producing the expected results and I wrote down the reasons in `XCTExpectFailure`. Maybe we should determine whether changing the test cases or marking these as bugs to fix.

### DIFF
--- a/Sources/SwiftIDEUtils/SyntaxClassification.swift
+++ b/Sources/SwiftIDEUtils/SyntaxClassification.swift
@@ -39,8 +39,6 @@ public enum SyntaxClassification {
   case lineComment
   /// The token should not receive syntax coloring.
   case none
-  /// An image, color, etc. literal.
-  case objectLiteral
   /// An identifier referring to an operator.
   case operatorIdentifier
   /// A `#` token like `#warning`.
@@ -66,7 +64,7 @@ extension SyntaxClassification {
   internal static func classify(_ keyPath: AnyKeyPath) -> (SyntaxClassification, Bool)? {
     switch keyPath {
     case \AttributeSyntax.attributeName:
-      return (.attribute, false)
+      return (.attribute, true)
     case \PlatformVersionItemSyntax.availabilityVersionRestriction:
       return (.keyword, false)
     case \AvailabilityVersionRestrictionSyntax.platform:
--- a/Sources/SwiftIDEUtils/SyntaxClassifier.swift
+++ b/Sources/SwiftIDEUtils/SyntaxClassifier.swift
@@ -160,17 +160,24 @@ private struct ClassificationVisitor {
     classifications.append(range)
   }
 
+  /// Classifies `triviaPieces` starting from `offset` and returns the number of bytes the trivia took up in the source
+  private mutating func classify(triviaPieces: [RawTriviaPiece], at offset: Int) -> Int {
+    var classifiedBytes = 0
+    for triviaPiece in triviaPieces {
+      let range = triviaPiece.classify(offset: offset + classifiedBytes)
+      report(range: range)
+      classifiedBytes += triviaPiece.byteLength
+    }
+    return classifiedBytes
+  }
+
   // Report classification ranges in `descriptor.node` that is a token.
   private mutating func handleToken(_ descriptor: Descriptor) -> VisitResult {
     let tokenView = descriptor.node.tokenView!
     var byteOffset = descriptor.byteOffset
 
     // Leading trivia.
-    for piece in tokenView.leadingRawTriviaPieces {
-      let range = piece.classify(offset: byteOffset)
-      report(range: range)
-      byteOffset += piece.byteLength
-    }
+    byteOffset += classify(triviaPieces: tokenView.leadingRawTriviaPieces, at: byteOffset)
     // Token text.
     do {
       let range = TokenKindAndText(kind: tokenView.rawKind, text: tokenView.rawText)
@@ -179,11 +186,7 @@ private struct ClassificationVisitor {
       byteOffset += tokenView.rawText.count
     }
     // Trailing trivia.
-    for piece in tokenView.trailingRawTriviaPieces {
-      let range = piece.classify(offset: byteOffset)
-      report(range: range)
-      byteOffset += piece.byteLength
-    }
+    byteOffset += classify(triviaPieces: tokenView.trailingRawTriviaPieces, at: byteOffset)
 
     precondition(byteOffset == descriptor.byteOffset + descriptor.node.byteLength)
     return .continue
@@ -196,12 +199,37 @@ private struct ClassificationVisitor {
 
     for case (let index, let child?) in children.enumerated() {
 
-      let classification: (SyntaxClassification, Bool)?
+      let classification: (classification: SyntaxClassification, force: Bool)?
       if case .layout(let layout) = descriptor.node.kind.syntaxNodeType.structure {
         classification = SyntaxClassification.classify(layout[index])
       } else {
         classification = nil
       }
+
+      if let classification, classification.force {
+        // Leading trivia.
+        if let leadingTriviaPieces = child.leadingTriviaPieces {
+          byteOffset += classify(triviaPieces: leadingTriviaPieces, at: byteOffset)
+        }
+        // Layout node text.
+        let layoutNodeTextLength = child.byteLength - child.leadingTriviaByteLength - child.trailingTriviaByteLength
+        let range = SyntaxClassifiedRange(
+          kind: classification.classification,
+          range: ByteSourceRange(
+            offset: byteOffset,
+            length: layoutNodeTextLength
+          )
+        )
+        report(range: range)
+        byteOffset += layoutNodeTextLength
+
+        // Trailing trivia.
+        if let trailingTriviaPieces = child.trailingTriviaPieces {
+          byteOffset += classify(triviaPieces: trailingTriviaPieces, at: byteOffset)
+        }
+        continue
+      }
+
       let result = visit(
         .init(
           node: child,
--- a/Tests/SwiftIDEUtilsTest/ClassificationTests.swift
+++ b/Tests/SwiftIDEUtilsTest/ClassificationTests.swift
@@ -121,4 +121,462 @@ public class ClassificationTests: XCTestCase {
       ]
     )
   }
+
+  public func testFuncDeclaration() {
+    assertClassification(
+      """
+      func foo(x: Int, y: Int) -> Int { return x + y }
+      """,
+      expected: [
+        ClassificationSpec(source: "func", kind: .keyword),
+        ClassificationSpec(source: "foo", kind: .identifier),
+        ClassificationSpec(source: "x", kind: .identifier),
+        ClassificationSpec(source: "Int", kind: .typeIdentifier),
+        ClassificationSpec(source: "y", kind: .identifier),
+        ClassificationSpec(source: "Int", kind: .typeIdentifier),
+        ClassificationSpec(source: "Int", kind: .typeIdentifier),
+        ClassificationSpec(source: "return", kind: .keyword),
+        ClassificationSpec(source: "x", kind: .identifier),
+        ClassificationSpec(source: "+", kind: .operatorIdentifier),
+        ClassificationSpec(source: "y", kind: .identifier),
+      ]
+    )
+  }
+
+  public func testDocCommentLine() {
+    assertClassification(
+      """
+      /// Brief.
+      ///
+      /// Simple case.
+      ///
+      /// - parameter x: A number
+      /// - parameter y: Another number
+      /// - returns: single-line, 2 spaces
+      """,
+      expected: [
+        ClassificationSpec(source: "/// Brief.", kind: .docLineComment),
+        ClassificationSpec(source: "///", kind: .docLineComment),
+        ClassificationSpec(source: "/// Simple case.", kind: .docLineComment),
+        ClassificationSpec(source: "///", kind: .docLineComment),
+        ClassificationSpec(source: "/// - parameter x: A number", kind: .docLineComment),
+        ClassificationSpec(source: "/// - parameter y: Another number", kind: .docLineComment),
+        ClassificationSpec(source: "/// - returns: single-line, 2 spaces", kind: .docLineComment),
+      ]
+    )
+  }
+
+  public func testDocCommentBlock() {
+    let docCommentBlock =
+      """
+      /**
+        Does pretty much nothing.
+
+        Not a parameter list: improper indentation.
+          - Parameters: sdfadsf
+
+        - WARNING: - WARNING: Should only have one field
+
+        - $$$: Not a field.
+
+        Empty field, OK:
+      */
+      """
+    assertClassification(
+      docCommentBlock,
+      expected: [
+        ClassificationSpec(source: docCommentBlock, kind: .docBlockComment)
+      ]
+    )
+  }
+
+  public func testEmptyDocBlockComment() {
+    assertClassification(
+      "/***/",
+      expected: [
+        ClassificationSpec(source: "/***/", kind: .docBlockComment)
+      ]
+    )
+  }
+
+  public func testPlaygoundCommentLine() {
+    assertClassification(
+      "//: playground doc comment line",
+      expected: [
+        ClassificationSpec(source: "//: playground doc comment line", kind: .lineComment)
+      ]
+    )
+  }
+
+  public func testIfConfig() {
+    assertClassification(
+      """
+      #if os(macOS)
+        var x : Int
+      #elseif
+        var x : Float
+      #else
+      #endif
+      """,
+      expected: [
+        ClassificationSpec(source: "#if", kind: .poundDirective),
+        ClassificationSpec(source: "os", kind: .buildConfigId),
+        ClassificationSpec(source: "macOS", kind: .buildConfigId),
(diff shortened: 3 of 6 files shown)


Tests/SwiftIDEUtilsTest/ClassificationTests.swift
```diff
@@ -121,4 +121,379 @@ public class ClassificationTests: XCTestCase {
...
+  }
+  
+  public func testAttribute() {
+    XCTExpectFailure(
+      """
+      '@available' is classified as @ and a typeIdentifier
+      'iOS' is classified as a typeIdentifier
+      '8.0' and '10.10' are classified as integerLiteral
+      """
+    )
```
ahoppen, line 303: Hmm, this is expected by the way the syntax tree gets represented because by now we always model the name of an attribute as a type. Also the lines between builtin attributes (like `@availablable`) with completely custom syntax, user-defined attributes that are types (like result builders and property wrappers) and user-defined types that aren’t types (attached macros) are blurring. So, what I would expect is `available` to be classified as an `attribute`. 

You should be able to achieve that by changing `\AttributeSyntax.attributeName` to return `(.attribute, false)` in `SyntaxClassification.classify(_ keyPath:)`

Same for `testAttribute2` where `objc` and `IBOutlet` should be classified as `attribute`.

Tests/SwiftIDEUtilsTest/ClassificationTests.swift
```diff
@@ -121,4 +121,379 @@ public class ClassificationTests: XCTestCase {
...
+  }
+  
+  public func testIndirectCase() throws {
+    XCTExpectFailure(
+      """
+      'indirect' is classified as an identifier
+      'T' is classified as an identifier
+      'List' is classified as an identifier
+      """
+    )
```
ahoppen, line 346: This should get correct classification if you wrap it inside an `enum`.

```swift
enum List {
  indirect case cons(T, List)
}
```

And while you’re at it, you can change the case to be lowercase to reflect the naming guidelines which changed around Swift 3.

Tests/SwiftIDEUtilsTest/ClassificationTests.swift
```diff
@@ -121,4 +121,379 @@ public class ClassificationTests: XCTestCase {
...
+        ClassificationSpec(source: "T", kind: .typeIdentifier),
+        ClassificationSpec(source: "List", kind: .typeIdentifier)
+      ]
+    )
+  }
+  
+  public func testMutatingKeyword() {
+    assertClassification(
+      "mutating func func_mutating_1() {}"
+      , expected: [
```
ahoppen, line 362: I think you need to run `format.py` to format this correctly.

Tests/SwiftIDEUtilsTest/ClassificationTests.swift
```diff
@@ -121,4 +121,379 @@ public class ClassificationTests: XCTestCase {
...
+        ClassificationSpec(source: "infix", kind: .keyword),
+        ClassificationSpec(source: "operator", kind: .keyword),
+        ClassificationSpec(source: "*-*", kind: .operatorIdentifier),
+        ClassificationSpec(source: "FunnyPrecedence", kind: .identifier)
+      ]
+    )
+  }
+  
+  public func testPrecedenceGroup() {
+    XCTExpectFailure("'left' is classified as a keyword in the test")
```
ahoppen, line 407: That’s OK. You can change the test

Tests/SwiftIDEUtilsTest/ClassificationTests.swift
```diff
@@ -121,4 +121,379 @@ public class ClassificationTests: XCTestCase {
...
+      "protocol Prot2 : Prot {}", expected: [
+        ClassificationSpec(source: "protocol", kind: .keyword),
+        ClassificationSpec(source: "Prot2", kind: .identifier),
+        ClassificationSpec(source: "Prot", kind: .typeIdentifier)
+      ]
+    )
+  }
+  
+  public func testGenericType() throws {
+    XCTExpectFailure(" '_' is not classified")
```
ahoppen, line 437: That’s also OK. Just remove the classification. I don’t really care about the classification of `_` and classifying it as a keyword never really made sense anyway.

Tests/SwiftIDEUtilsTest/ClassificationTests.swift
```diff
@@ -121,4 +121,379 @@ public class ClassificationTests: XCTestCase {
...
+        ClassificationSpec(source: "where", kind: .keyword),
+        ClassificationSpec(source: "T", kind: .typeIdentifier),
+        ClassificationSpec(source: "Blarg", kind: .typeIdentifier),
+        ClassificationSpec(source: "Prot2", kind: .typeIdentifier)
+      ])
+  }
+  
+  public func testStringInterpolationAnchor() {
+    assertClassification(
+      "\"This is string \\(1) interpolation\""
```
ahoppen, line 457: I think if you write this as a raw string `#"` or a multiline string, it becomes easier to read because you don’t need to do any escaping. 

Same in the `ClassificationSpec` below

Tests/SwiftIDEUtilsTest/ClassificationTests.swift
```diff
@@ -121,4 +121,379 @@ public class ClassificationTests: XCTestCase {
...
+        ClassificationSpec(source: "unowned", kind: .keyword),
+        ClassificationSpec(source: "unsafe", kind: .identifier),
+        ClassificationSpec(source: "var", kind: .keyword),
+        ClassificationSpec(source: "uu", kind: .identifier)
+      ]
+    )
+  }
+  
+  public func testClosure() {
+    XCTExpectFailure("There is a FIXME in lit-test, not quite sure why")
```
ahoppen, line 491: I think we weren’t handling closure capture lists properly for a long time (but that’s years in the past). What classification does it produce now?

Tests/SwiftIDEUtilsTest/ClassificationTests.swift
```diff
@@ -121,4 +121,379 @@ public class ClassificationTests: XCTestCase {
...
+        ClassificationSpec(source: "!", kind: .operatorIdentifier),
+        ClassificationSpec(source: "CONF", kind: .buildConfigId),
+        ClassificationSpec(source: "#else", kind: .poundDirective),
+      ]
+    )
+  }
+  
+  public func testBuiltinDirectives() {
+    // https://github.com/apple/swift-syntax/issues/1044
+    XCTExpectFailure("the directive keywords are classied as # and identifier")
```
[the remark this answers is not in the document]
StevenWong12: Since that, do we still need `the objectLiteral` type? `#fileLiteral`, `#imageLiteral`, etc.. are also classified as identifier now.

Tests/SwiftIDEUtilsTest/ClassificationTests.swift
StevenWong12: > You should be able to achieve that by changing \AttributeSyntax.attributeName to return (.attribute, false)

Do you mean we changing this to `(.attribute, true)`? Since `\AttributeSyntax.attributeName` already returns `(.attribute, false)` now. 

If so, we may need to make some changes on `handleLayout` (as my new commits) since `\AttributeSyntax.attributeName` is not a `TokenSyntax` and the second return value of `SyntaxClassification.classify(_ keyPath:)` only has effects in `handleToken`.

But I can't find a proper workaround for classifying the trivia pieces in a layout node for now though.

Tests/SwiftIDEUtilsTest/ClassificationTests.swift
StevenWong12: It just produces normal `SyntaxClassification` i.e. `weak` as a keyword, `x` as an identifier, etc..

Tests/SwiftIDEUtilsTest/ClassificationTests.swift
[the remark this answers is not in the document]
ahoppen: I believe we can delete it, yes.

Tests/SwiftIDEUtilsTest/ClassificationTests.swift
ahoppen: > Do you mean we changing this to `(.attribute, true)`? Since `\AttributeSyntax.attributeName` already returns `(.attribute, false)` now.
>
> If so, we may need to make some changes on `handleLayout` (as my new commits) since `\AttributeSyntax.attributeName` is not a `TokenSyntax` and the second return value of `SyntaxClassification.classify(_ keyPath:)` only has effects in `handleToken`.

Yes, that’s what I meant. But I didn’t realize it only had an effect on tokens 😢 

> On a second thought, what do you think about refactor `ClassificationVisitor` as a subclass of `SyntaxAnyVisitor` like `ParseDiagnosticsGenerator`. That should make this case handled easier. And the `8.0` and `10.10` can be specified as `floatingLiteral` rather that some `integerLiteral`s in a version tuple.

I think that would be a good idea and should make the classifier easier to maintain. If the `SyntaxAnyVisitor`-based implementation is significantly slower than the current one, we should look into why that is and see if we can make `SyntaxAnyVisitor` faster.

In any case, I would prefer to do this in a follow-up PR. We can just keep the test case disabled for now. One note though: `XCTExpectFailure` is not supported on Linux IIRC, so you would need to do `try XCTSkipIf(true, <message>)`

Tests/SwiftIDEUtilsTest/ClassificationTests.swift
ahoppen: That sounds good to me 👍🏽

Sources/SwiftIDEUtils/SyntaxClassification.swift
```diff
@@ -66,7 +66,7 @@ extension SyntaxClassification {
   internal static func classify(_ keyPath: AnyKeyPath) -> (SyntaxClassification, Bool)? {
     switch keyPath {
     case \AttributeSyntax.attributeName:
-      return (.attribute, false)
+      return (.attribute, true)
```
ahoppen, line 69: If this has no effect, I’d prefer to revert it.

Sources/SwiftIDEUtils/SyntaxClassifier.swift
```diff
@@ -196,12 +196,23 @@ private struct ClassificationVisitor {
...
+      if let classification, classification.force == true {
+        let range = SyntaxClassifiedRange(
+          kind: classification.classification,
+          range: ByteSourceRange(offset: byteOffset, length: child.byteLength - child.leadingTriviaByteLength - child.trailingTriviaByteLength)
+        )
+        report(range: range)
+        byteOffset += child.byteLength
+        continue
+      }
+
```
ahoppen, line 215: Oh, did you find a solution to https://github.com/apple/swift-syntax/pull/1953#discussion_r1276769666 after all? Or doesn’t this work?

Sources/SwiftIDEUtils/SyntaxClassifier.swift
```diff
@@ -196,12 +196,23 @@ private struct ClassificationVisitor {
...
 
-      let classification: (SyntaxClassification, Bool)?
+      let classification: (classification: SyntaxClassification, force: Bool)?
       if case .layout(let layout) = descriptor.node.kind.syntaxNodeType.structure {
         classification = SyntaxClassification.classify(layout[index])
       } else {
         classification = nil
       }
+
+      if let classification, classification.force == true {
```
ahoppen, line 206: `== true` seems redundant to me.

Sources/SwiftIDEUtils/SyntaxClassifier.swift
```diff
@@ -196,12 +196,23 @@ private struct ClassificationVisitor {
...
+        let range = SyntaxClassifiedRange(
+          kind: classification.classification,
+          range: ByteSourceRange(offset: byteOffset, length: child.byteLength - child.leadingTriviaByteLength - child.trailingTriviaByteLength)
```
ahoppen, line 209: If you are subtracting `child.leadingTriviaByteLength` from `length` you need to add it to `offset`

Tests/SwiftIDEUtilsTest/ClassificationTests.swift
```diff
@@ -121,4 +121,393 @@ public class ClassificationTests: XCTestCase {
...
+      "#fileLiteral(resourceName: \"cloud.png\")",
+      expected: [
+        ClassificationSpec(source: "fileLiteral", kind: .identifier),
+        ClassificationSpec(source: "resourceName", kind: .identifier),
+        ClassificationSpec(source: "\"cloud.png\"", kind: .stringLiteral),
+      ]
+    )
+  }
+
+  // TODO@hzy discuz about this and below
```
ahoppen, line 320: LGTM. I would just prefer to also classify the `@` as `attribute`. AFAICT you just need revert one of the changes you did above.

Sources/SwiftIDEUtils/SyntaxClassifier.swift
StevenWong12: That could be a solution if we don't refactoring `ClassificationVisitor`,  and that works.😉

Sources/SwiftIDEUtils/SyntaxClassifier.swift
```diff
@@ -196,12 +196,37 @@ private struct ClassificationVisitor {
...
+      let classification: (classification: SyntaxClassification, force: Bool)?
       if case .layout(let layout) = descriptor.node.kind.syntaxNodeType.structure {
         classification = SyntaxClassification.classify(layout[index])
       } else {
         classification = nil
       }
+
+      if let classification, classification.force {
+        // Leading trivia.
+        report(range: SyntaxClassifiedRange(kind: .none, range: ByteSourceRange(offset: byteOffset, length: child.leadingTriviaByteLength)))
```
ahoppen, line 208: We should also classify the trivia for doc comments (both leading and trailing). A test case to test that would be.

```swift
@MyAttribute // some comment
func foo() {}
```

To do that, I think you should factor out the following and call it to classify the trivia.

https://github.com/apple/swift-syntax/blob/e04c5c117bdf5a92ddc309ac4ce8e33fe302db41/Sources/SwiftIDEUtils/SyntaxClassifier.swift#L168-L173

Sources/SwiftIDEUtils/SyntaxClassifier.swift
ahoppen: I think that’s a fine solution. Good idea 👍🏽

ahoppen (commented):
Thank you. Your idea with forcing the classification of an entire subtree sounds like a good solution for now. We would still be classifying the comment in the following as an attribute but that’s something we can fix in a rewrite of `SyntaxClassifier`.

```swift
@MyType /* abc */ . NestedResultBuilder
func foo() {}
```

Sources/SwiftIDEUtils/SyntaxClassifier.swift
StevenWong12: Ah, yeah. But we also have to add two public properties `leadingTriviaPieces` and `trailingTriviaPieces` in `RawSyntax` to get the trivia pieces as in my new commit.

ahoppen: Oh, I thought we could have a function to classify the trivia pieces so we don’t need to repeat this loop everywhere. 

Something like

```swift
/// Classifies `trivia` and returns the number of bytes the trivia took up in the source
func classify(trivia: [RawTriviaPiece]) -> Int {
  var classifiedBytes = 0
  for triviaPiece in triviaPieces {
    let range = triviaPiece.classify(offset: byteOffset)
    report(range: range)
    classifiedBytes += triviaPiece.byteLength
  }
  return classifiedBytes
}
```

And then here you can just do

```swift
if let triviaPieces = child.leadingTriviaPieces {
  byteOffset += classify(trivia: triviaPieces)
}
```

And you can also use the function in `handleToken`.

My thinking is that repeating an implementation twice is still OK but as soon as we reach 3 or more copies, we should really factor it out.

Sources/SwiftIDEUtils/SyntaxClassifier.swift
```diff
@@ -196,12 +196,49 @@ private struct ClassificationVisitor {
...
+          for triviaPiece in triviaPieces {
+            let range = triviaPiece.classify(offset: byteOffset)
+            report(range: range)
+            byteOffset += triviaPiece.byteLength
+          }
+        }
+        child.leadingTriviaPieces?.compactMap { $0 }.forEach {
+          let range = $0.classify(offset: byteOffset)
+          report(range: range)
+        }
```
ahoppen, line 218: I think you’re classifying `leadingTriviaPieces` twice: Once in the `for` loop above and then once here.

Sources/SwiftIDEUtils/SyntaxClassifier.swift
StevenWong12: Ah, I think that looks nice. 👍

ahoppen (approved):
Looks good 👍🏽 Thanks

ahoppen:
@swift-ci Please test

ahoppen:
This breaks `sourcekit-lsp` because you are removing `SyntaxClassification.objectLiteral`. Could you create a corresponding sourcekit-lsp PR?

PR #1954, opened 2023-07-27
Perform a couple of renames on differentiability attribute nodes

ahoppen:
Mostly change parameter to argument where the parameter is actually an argument.


Tests/SwiftParserTest/AttributeTests.swift
```diff
@@ -51,8 +51,8 @@ final class AttributeTests: XCTestCase {
       diagnostics: [
         DiagnosticSpec(
           locationMarker: "1️⃣",
-          message: "expected ':' and parameters in '@differentiable' argument",
-          fixIts: ["insert ':' and parameters"]
+          message: "expected ':' and arguments in '@differentiable' argument",
+          fixIts: ["insert ':' and arguments"]
```
bnbarham, line 55: I guess this doesn't read any worse than before. But... it is a bit odd 😅. The problem is that `Clause` (containing wrt and `Arguments`), `Arguments` (which contains the parentheses), and `ArgumentList` (between the parentheses) all have basically the same name (+- an s).

bnbarham (approved):

ahoppen:
https://github.com/apple/swift-format/pull/574

@swift-ci Please test

ahoppen:
https://github.com/apple/swift-format/pull/574

@swift-ci Please test Windows

Tests/SwiftParserTest/AttributeTests.swift
ahoppen: Yes. If there was more use of the `@differentiable` attribute, I would probably agree but for now, I’d suggest we keep it as-is.

PR #1956, opened 2023-07-27
Generate documentation for the SwiftSyntaxMacroExpansion module on swiftpackageindex.com

ahoppen:
This module was missing from .spi.yml

This will be caught by the check from https://github.com/apple/swift-syntax/pull/1635 once the CI nodes have been updated so that we can use Swift regex on them.


ahoppen:
@swift-ci Please test

bnbarham (approved):

kimdv (approved):

PR #1958, opened 2023-07-27
Add methods from `RangeReplaceableCollection` to `SyntaxCollection` and remove `add*` methods on syntax nodes that have a collection as child

ahoppen:
Especially the `+` operators make manipulation of `SyntaxCollection` a lot easier.

`SyntaxCollection` doesn’t conform to `RangeReplaceableCollection` because of two reasons:
 - `RangeReplaceableCollection` assumes that creating a new colleciton and adding all elements from another collection to it results in the same collection. This is not true for `SyntaxCollection` because the new collection wouldn’t have a parent. This causes eg. the default implementation of `filter` to misbehave.
 - It can’t make the complexity guarantees that `RangeReplaceableCollection` requires, e.g. `append` is `O(tree depth)` instead of `O(1)`.

For the same reason, also remove the conformance to `MutableCollection`.

With those operators and methods, we no longer need the `add*` methods on syntax nodes that have a syntax collection as their child.

### DIFF
--- a/Sources/SwiftSyntax/SyntaxCollection.swift
+++ b/Sources/SwiftSyntax/SyntaxCollection.swift
@@ -10,7 +10,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-public protocol SyntaxCollection: SyntaxProtocol, BidirectionalCollection, MutableCollection where Element: SyntaxProtocol {
+public protocol SyntaxCollection: SyntaxProtocol, BidirectionalCollection where Element: SyntaxProtocol, Index == SyntaxChildrenIndex {
   associatedtype Iterator = SyntaxCollectionIterator<Element>
 
   /// The ``SyntaxKind`` of the syntax node that conforms to ``SyntaxCollection``.
@@ -31,6 +31,17 @@ extension SyntaxCollection {
   /// is undefined.
   internal init(_ data: SyntaxData) {
     self.init(Syntax(data))!
+  }
+
+  /// Initialize the collection from a collection of the same type.
+  ///
+  /// - Note: This initializer is needed to improve source compatibility after
+  ///   the `+` operators have been added. Previously, they created an array
+  ///   that was converted to a ``SyntaxCollection`` and now they create a
+  ///   ``SyntaxCollection``, which makes the initializer a no-op.
+  @available(*, deprecated, message: "Call to initializer is not necessary.")
+  public init(_ collection: Self) {
+    self = collection
   }
 
   public init<Children: Sequence>(_ children: Children) where Children.Element == Element {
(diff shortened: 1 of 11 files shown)


ahoppen:
@swift-ci Please test

kimdv (approved):

ahoppen:
https://github.com/apple/swift-format/pull/575

@swift-ci Please test

ahoppen:
https://github.com/apple/swift-format/pull/575

@swift-ci Please test

ahoppen:
https://github.com/apple/swift-format/pull/575

@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

bnbarham (approved):
My main concern here is that many of the methods list a Big O that isn't true for our implementation. Eg. even something like `append` ends up doing much more work than you'd otherwise expect (since it is constructing all parents + has to copy the elements into a new raw syntax node).

That's what all the `add*` are doing today though, so while I don't think this is blocking, perhaps we should add some extra documentation in. There's also the `filter` case, which we should probably make a custom implementation for.

ahoppen:
I added an overload of `filter` and two more to override the methods in `RangeReplacableCollection` that were calling `Self()` and thus dropped the parent. Could you take another quick look at them @bnbarham?

ahoppen:
https://github.com/apple/swift-format/pull/575

pt

ahoppen:
https://github.com/apple/swift-format/pull/575

@swift-ci Please test

Sources/SwiftSyntax/SyntaxCollection.swift
```diff
@@ -33,6 +33,11 @@ extension SyntaxCollection {
     self.init(Syntax(data))!
   }
 
+  /// Create an empty ``SyntaxCollection`` with no children.
+  public init() {
+    self.init([])
```
bnbarham, line 38: I guess this is the other downside of using `RangeReplaceableCollection`. This constructor should really never be used and makes it super easy to drop the parent accidentally.

kimdv:
https://github.com/apple/swift-format/pull/575

@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

Tests/SwiftSyntaxTest/SyntaxCollectionsTests.swift
```diff
@@ -146,4 +146,32 @@ public class SyntaxCollectionsTests: XCTestCase {
...
+      integerLiteralElement(1),
+      integerLiteralElement(2),
+    ])
+    let array = ArrayExprSyntax(elements: elements)
+
+    let addedElements: ArrayElementListSyntax = Array([integerLiteralElement(4)]) + array.elements
+    XCTAssertEqual(addedElements.count, 4)
+    // If the first collection of + doesn’t have a parent, it doesn’t really
+    // make sense to inherit the parent from the second collection.
+    XCTAssert(addedElements.parent?.is(ArrayExprSyntax.self) ?? false)
```
bnbarham, line 175: I wouldn't expect this. This is basically prepending and I think I'd want to keep the parent in that case?

bnbarham: `testPrependingElement` tests this case FWIW

Tests/SwiftSyntaxTest/SyntaxCollectionsTests.swift
```diff
@@ -146,4 +146,32 @@ public class SyntaxCollectionsTests: XCTestCase {
...
+  }
```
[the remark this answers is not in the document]
ahoppen: Added tests

Tests/SwiftSyntaxTest/SyntaxCollectionsTests.swift
bnbarham: That's... true 🤔 The comment seems to be suggesting the opposite doesn't it? And then I read the assert as an `XCTAssertFalse`.

ahoppen: Oh, the comment was outdated…

ahoppen:
@swift-ci Please test Windows

Tests/SwiftSyntaxTest/SyntaxCollectionsTests.swift
```diff
@@ -14,136 +14,253 @@ import XCTest
...
-      integerLiteralElement(0)
-    ])
-
-    let newArrayElementList = ArrayElementListSyntax(arrayElementList + [integerLiteralElement(1)])
-    XCTAssert(newArrayElementList.kind.isSyntaxCollection)
-    XCTAssertEqual(newArrayElementList.count, 2)
-    XCTAssertNotNil(newArrayElementList.child(at: 1))
-    XCTAssert(!newArrayElementList.child(at: 1)!.kind.isSyntaxCollection)
-    XCTAssertEqual("\(newArrayElementList.child(at: 1)!)", "1")
+fileprivate func assertSyntaxCollectionManipulation(
```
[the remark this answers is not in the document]
ahoppen: That’s why I am performing all the transformations on an `ArrayExprSyntax` instead of the actual collection type `ArrayElementListSyntax`. That way the substructure test checks that we still have the collection anchored in a parent.

Issue #1959, opened 2023-07-28
BasicFormat outputs incomplete indentation when formatting multiline string literals containing blank lines


kishikawakatsumi:
### Description

The following code is formatted using BasicFormat (spaces are replaced with `␣` for clarity),

```swift
assertionFailure("""
␣␣First␣line

␣␣Second␣line

␣␣""")
```

It produces the following code

```swift
assertionFailure("""
␣␣First␣line
␣␣␣␣
␣␣Second␣line
␣␣␣␣
␣␣␣␣␣␣""")
```

This is an incorrectly indented literal closing quote.
This causes a compilation error `Insufficient indentation of line in multi-line string literal`.

This problem can easily occur if both the "Automatically trim trailing whitespace" and "Including whitespace-only lines" settings in Xcode are enabled.

Also, macros now automatically format the code they generate by default, which can easily cause problems if the above settings are enabled.

For example, passing the following string to the `#stringify` macro created by the Swift Macro Project template will result in a compilation error due to the same issue.

```swift
#stringify(
␣␣"""
␣␣First line

␣␣Second line

␣␣"""
)
```

^ The above macro will generate the following code, which will result in a compiation error

```swift
(
␣␣"""
␣␣First␣line
␣␣
␣␣Second␣line
␣␣
␣␣␣␣""",␣#"\#n␣␣"""\#n␣␣First␣line\#n\#n␣␣Second␣line\#n\#n␣␣""""#)
```

### Steps to Reproduce

To reproduce this problem, run the following code

```swift
let source = #"""
  assertionFailure("""
    First line

    Second line

    """)
  """#
print(Parser.parse(source: source).formatted().description)
```

The result of the execution will be as follows This code results in a compilation error "Insufficient indentation of line in multi-line string literal".

```swift
assertionFailure("""
  First line
    
  Second line
    
      """)
```

ahoppen:
Tracked in Apple’s issue tracker as rdar://113021776

PR #1960, opened 2023-07-28
Do not format blank lines in multiline string literals

kishikawakatsumi:
Fixes https://github.com/apple/swift-syntax/issues/1959

For an example, see the following code

```swift
assertionFailure("""
  First line

  Second line

  """)
```

There are blank lines in line 3 and 5.

The conditions for a blank line are that StringSegments is only a newline character with no trivia (line 3), or that StringSegments is a zero-length string and the trailing trivia is a newline (line 5).

However, this condition would also match empty StringSegments immediately after string interpolation in the following code. (This case is covered by [`testStringInterpolationAtStartOfMultiLineStringLiteral`](https://github.com/apple/swift-syntax/blob/c2f6b9d18ed8c49b6e515b5e1f34b0fb5594eec7/Tests/SwiftBasicFormatTest/BasicFormatTests.swift#L211-L234).)

```swift
assertionFailure("""
  ABC
  \(myVar)
  """)
```

Therefore, I also added the condition that the previous token must be ended by a newline.

### DIFF
--- a/Sources/SwiftBasicFormat/BasicFormat.swift
+++ b/Sources/SwiftBasicFormat/BasicFormat.swift
@@ -433,6 +433,19 @@ open class BasicFormat: SyntaxRewriter {
       return trailingTrivia + Trivia(pieces: nextTokenLeadingWhitespace)
     }()
 
+    /// Whether the leading trivia of the token is folloed by a newline.
+    let leadingTriviaIsFollowedByNewline: Bool = {
+      if token.text.isEmpty && token.trailingTrivia.startsWithNewline {
+        return true
+      } else if token.text.first?.isNewline ?? false {
+        return true
+      } else if token.text.isEmpty && token.trailingTrivia.isEmpty && nextTokenWillStartWithNewline {
+        return true
+      } else {
+        return false
+      }
+    }()
+
     var leadingTrivia = token.leadingTrivia
     var trailingTrivia = token.trailingTrivia
 
@@ -455,9 +468,11 @@ open class BasicFormat: SyntaxRewriter {
       }
     }
 
-    if leadingTrivia.indentation(isOnNewline: isInitialToken || previousTokenWillEndWithNewline) == [] {
+    let isEmptyLine = token.leadingTrivia.isEmpty && leadingTriviaIsFollowedByNewline
+    if leadingTrivia.indentation(isOnNewline: isInitialToken || previousTokenWillEndWithNewline) == [] && !isEmptyLine {
       // If the token starts on a new line and does not have indentation, this
-      // is the last non-indented token. Store its indentation level
+      // is the last non-indented token. Store its indentation level.
+      // Do not store the indentation level if the line is empty.
       anchorPoints[token] = currentIndentationLevel
     }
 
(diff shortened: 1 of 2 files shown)


Sources/SwiftBasicFormat/BasicFormat.swift
```diff
@@ -508,6 +508,17 @@ open class BasicFormat: SyntaxRewriter {
       return token
     }
 
+    if token.text == "\n" && token.leadingTrivia.isEmpty && token.trailingTrivia.isEmpty {
```
Matejkob, line 511: I'm wondering whether it would be more beneficial to use [`isNewline`](https://developer.apple.com/documentation/swift/character/isnewline) instead of `"\n"`. What do you think about this?

This can be achieved in a variety of ways. For example, we could use `text.allSatisfy(\.isNewline)`. However, if we are prioritizing performance, we could aim for O(1) complexity with an approach like the following:
```swift
extension String {
  func isNewline() -> Bool {
    (first?.isNewline ?? false) && dropFirst().isEmpty
  }
}
```

Sources/SwiftBasicFormat/BasicFormat.swift
```diff
@@ -508,6 +508,17 @@ open class BasicFormat: SyntaxRewriter {
...
+      return token
+    }
+    if token.text.isEmpty
+        && token.leadingTrivia.isEmpty
+        && token.trailingTrivia.startsWithNewline
+        && token.trailingTrivia.endsWithNewline
+        && previousTokenWillEndWithNewline {
+      return token
+    }
```
Matejkob, line 520: I noticed some minor formatting differences in this code compared to other code in the repo. Would you mind [runing `swift-format`](https://github.com/apple/swift-syntax/blob/main/CONTRIBUTING.md#formatting)?

Matejkob (commented):
I've been reviewing your PR to learn more about the swift-syntax project, and I've come across a few points of interest. I've left a few in-line comments with some minor suggestions. 💙

ahoppen (commented):
Thanks for the PR @kishikawakatsumi.

I have just investigated and it looks like the root causes of the issue are that 
1. We are setting an `anchorPoint` for empty lines. Anchor points are meant to be starting points at which the user didn’t provide any manual indentation relative to which we should format the rest of the tree. But for empty lines it’s not like the user didn’t provide any indentation. There simply wasn’t anything to indent.
2. `leadingTrivia.trimmingTrailingWhitespaceBeforeNewline` was not considering whether the token text itself contained a newline in `isBeforeNewline`.

To fix these root causes, what I would suggest is to 

- Add a new variable behind `combinedTrailingTrivia`

```swift
/// Whether the leading trivia of the token is folloed by a newline.
let leadingTriviaIsFollowedByNewline: Bool = {
  if token.text.isEmpty && token.trailingTrivia.startsWithNewline {
    return true
  } else if token.text.first?.isNewline ?? false {
    return true
  } else if token.text.isEmpty && token.trailingTrivia.isEmpty && nextTokenWillStartWithNewline {
    return true
  } else {
    return false
  }
}()
```

- Change the anchor point computation to the following

```swift
let isEmptyLine = token.leadingTrivia.isEmpty && leadingTriviaIsFollowedByNewline
if leadingTrivia.indentation(isOnNewline: isInitialToken || previousTokenWillEndWithNewline) == [] && !isEmptyLine {
  // If the token starts on a new line and does not have indentation, this
  // is the last non-indented token. Store its indentation level.
  // Do not
}
```

- And finally, also pass `leadingTriviaIsFollowedByNewline` to `trimmingTrailingWhitepsaceBeforeNewline`

```swift
leadingTrivia = leadingTrivia.trimmingTrailingWhitespaceBeforeNewline(isBeforeNewline: leadingTriviaIsFollowedByNewline)
```

kishikawakatsumi:
@ahoppen Thanks for the guide. The formatting problem for blank lines in multi-line string literals has been fixed. However, it seems to affect other test cases. The following three test-cases fail.

- `DeclarationTests.testProtocolParsing()`
- `EnumTests.testEnum21a()`
- `OperatorDeclTests.testOperatorDecl15()`

The diff for each is as follows. Perhaps a space that should be inserted has not been inserted?

```
// DeclarationTests.testProtocolParsing()
–protocol P{{}case <#identifier#>
+protocol P{{}case<#identifier#>
 }
```

```
// EnumTests.testEnum21a()
 enum Recovery1 {
–  case <#identifier#>:
+  case<#identifier#>:
 }
```

```
// OperatorDeclTests.testOperatorDecl15()
 precedencegroup E {
–  higherThan: <#identifier#>
+  higherThan:<#identifier#>
 }
```

ahoppen:
Hi @kishikawakatsumi,

Oh, I should have totally checked whether my changes still formatted all the remaining diagnostics correctly, I’m sorry. The issue is that in the parsed tree, we represent the missing identifiers by identifiers with empty text as well and they only get transformed to the `<#identifier#>` placeholders in a later `PresentMaker` pass. Thus, all my rules of checking whether the text is empty also applied to these identifiers, which was causing the issues.

Looking into it further, I realized that the proper fix required some more, slightly more invasive changes that I, myself, would like to get a review on. I have opened https://github.com/apple/swift-syntax/pull/1964 with these changes and hope you don’t mind me closing this PR in favor of that one.

Sources/SwiftBasicFormat/BasicFormat.swift
ahoppen: I just wanted to say thanks for your review, @Matejkob. Those are valuable comments, they only got overridden by the fact that I found bigger design-ish improvements in the PR.

kishikawakatsumi:
Thanks!

PR #1962, opened 2023-07-28
Change `lazy var` in BasicFormat that are always evaluated to `let`

ahoppen:
Using `lazy var` only is beneficial if the variable is not always accessed and if we can thus skip the computation. But quite a few of these were always accessed in BasicFormat now. I didn’t measure but I remember that there’s a tiny performance penalty associated with `lazy var`, which doesn’t give us any benefits if the variable is always accessed. Thus, change them to `let`.


ahoppen:
@swift-ci Please test

kimdv (approved):

bnbarham (approved):

ahoppen:
@swift-ci Please test Windows

PR #1964, opened 2023-07-29
Unify `PresentMaker` and `MissingNodesBasicFormatter`, fixing a formatting bug in string literals

### LINKED ISSUE #1959: BasicFormat outputs incomplete indentation when formatting multiline string literals containing blank lines

ahoppen:
The underlying issues of the malformatted multi-line string literal are that

1. We are setting an `anchorPoint` for empty lines. Anchor points are meant to be starting points at which the user didn’t provide any manual indentation relative to which we should format the rest of the tree. But for empty lines it’s not like the user didn’t provide any indentation. There simply wasn’t anything to indent.
2. `leadingTrivia.trimmingTrailingWhitespaceBeforeNewline` was not considering whether the token text itself contained a newline in `isBeforeNewline`.

Now, to detect if the token is followed by a newline, we should check if its text is empty, which is the case for empty string literals. Unfortunately, this is also the case for missing identifiers, which also have an empty text but whose empty text will be replace by a placeholder by `PresentMaker`. To distinguish between the two, I merged `PresentMaker` and `MissingNodesBasicFormatter` so that the `BasicFormat` also performs the text replacement to placeholders.

Fixes #1959

### DIFF
--- a/Sources/SwiftBasicFormat/BasicFormat.swift
+++ b/Sources/SwiftBasicFormat/BasicFormat.swift
@@ -48,6 +48,14 @@ open class BasicFormat: SyntaxRewriter {
   /// been visited yet.
   private var previousToken: TokenSyntax? = nil
 
+  /// The number of ancestors that are `StringLiteralExprSyntax`.
+  private var stringLiteralNestingLevel = 0
+
+  /// Whether we are currently visiting the subtree of a `StringLiteralExprSyntax`.
+  private var isInsideStringLiteral: Bool {
+    return stringLiteralNestingLevel > 0
+  }
+
   public init(
     indentationWidth: Trivia = .spaces(4),
     initialIndentation: Trivia = [],
@@ -83,6 +91,9 @@ open class BasicFormat: SyntaxRewriter {
   }
 
   open override func visitPre(_ node: Syntax) {
+    if node.is(StringLiteralExprSyntax.self) {
+      stringLiteralNestingLevel += 1
+    }
     if requiresIndent(node) {
       if let firstToken = node.firstToken(viewMode: viewMode),
         let tokenIndentation = firstToken.leadingTrivia.indentation(isOnNewline: false),
@@ -98,6 +109,9 @@ open class BasicFormat: SyntaxRewriter {
   }
 
   open override func visitPost(_ node: Syntax) {
+    if node.is(StringLiteralExprSyntax.self) {
+      stringLiteralNestingLevel -= 1
+    }
     if requiresIndent(node) {
       decreaseIndentationLevel()
     }
@@ -498,10 +512,12 @@ open class BasicFormat: SyntaxRewriter {
       }
     }
 
-    let isEmptyLine = token.leadingTrivia.isEmpty && leadingTriviaIsFollowedByNewline
-    if leadingTrivia.indentation(isOnNewline: isInitialToken || previousTokenWillEndWithNewline) == [] && !isEmptyLine {
+    if leadingTrivia.indentation(isOnNewline: isInitialToken || previousTokenWillEndWithNewline) == [] && !token.isStringSegment {
       // If the token starts on a new line and does not have indentation, this
-      // is the last non-indented token. Store its indentation level
+      // is the last non-indented token. Store its indentation level.
+      // But never consider string segments as anchor points since you can’t
+      // indent individual lines of a multi-line string literals without breaking
+      // their integrity.
       anchorPoints[token] = currentIndentationLevel
     }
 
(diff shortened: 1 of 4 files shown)


ahoppen:
@swift-ci Please test

Tests/SwiftBasicFormatTest/BasicFormatTests.swift
```diff
@@ -250,6 +250,23 @@ final class BasicFormatTest: XCTestCase {
     )
   }
 
+  func testMultilineStringLiteralWithBlackLines() {
```
kishikawakatsumi, line 253:
```suggestion
  func testMultilineStringLiteralWithBlankLines() {
```

I made a typo in the test method name.

ahoppen: Haha, thanks for pointing it out.

Tests/SwiftBasicFormatTest/BasicFormatTests.swift
```diff
@@ -250,6 +250,23 @@ final class BasicFormatTest: XCTestCase {
     )
   }
 
+  func testMultilineStringLiteralWithBlankLines() {
+    assertFormatted(
+      source: #"""
+        assertionFailure("""
+          First line
```
kishikawakatsumi, line 257: The blank line is gone. The code is the same as the test code above `testMultilineStringLiteral()`.

kishikawakatsumi:
In the following test case (no spaces in the empty lines), an indent is inserted in the first blank line. This is not identical to the original string.

```swift
assertFormatted(
  source: #"""
    assertionFailure("""


      First line
      Second line
      """)
    """#,
  expected: #"""
    assertionFailure("""


      First line
      Second line
      """)
    """#
)
```

Actual Result:
```
assertionFailure("""
␣␣␣␣

  First line
  Second line
  """)
```

Sources/SwiftBasicFormat/BasicFormat.swift
```diff
@@ -436,6 +458,19 @@ open class BasicFormat: SyntaxRewriter {
       return trailingTrivia + Trivia(pieces: nextTokenLeadingWhitespace)
     }()
 
+    /// Whether the leading trivia of the token is folloed by a newline.
```
kimdv, line 461: Typo in `followed`

kimdv (approved):

bnbarham (approved):

Tests/SwiftBasicFormatTest/BasicFormatTests.swift
ahoppen: Oh, copying code from GitHub apparently drops empty lines 🤯 Thanks for catching it. I updated the test case

ahoppen:
> In the following test case (no spaces in the empty lines), an indent is inserted in the first blank line. This is not identical to the original string.

Fixed

Sources/SwiftBasicFormat/BasicFormat.swift
ahoppen: Thanks. Found another two typos in the file as well.

ahoppen:
@swift-ci Please test

kishikawakatsumi:
Running the following test case with the latest commit 3961b9c3bf3019e38c17db190de774d32b22ff61 produces incorrectly indented results.


```
assertFormatted(
  source: #"""
    assertionFailure("""

      \("""
        First Line
      """)
    """)
    """#,
  expected: ...
)
```

Actual result:

```
assertionFailure("""

      \("""
    First Line
  """)
    """)
```

ahoppen:
@swift-ci Please test

ahoppen:
Fixed as well. I can’t think of any more ways to break formatting of multi-line string literals but if you do find more, please let me know @kishikawakatsumi. And thanks for the thorough testing.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

PR #1966, opened 2023-07-31
Remove syntax classification `lit`-based tests

StevenWong12:
The follow-up PR for https://github.com/apple/swift-syntax/pull/1953.

Also remove the `lit-test-helper` in codebase. 🥳


lit_tests/coloring.swift
```diff
@@ -1,404 +0,0 @@
...
-func foo2(O1 : Int?, O2: Int?, O3: Int?) {
-  guard let _ = O1, var _ = O2, let _ = O3 else { }
-// CHECK:  <kw>guard</kw> <kw>let</kw> <kw>_</kw> = <id>O1</id>, <kw>var</kw> <kw>_</kw> = <id>O2</id>, <kw>let</kw> <kw>_</kw> = <id>O3</id> <kw>else</kw> { }
-  if let _ = O1, var _ = O2, let _ = O3 {}
-// CHECK: <kw>if</kw> <kw>let</kw> <kw>_</kw> = <id>O1</id>, <kw>var</kw> <kw>_</kw> = <id>O2</id>, <kw>let</kw> <kw>_</kw> = <id>O3</id> {}
-}
-
-func keywordInCaseAndLocalArgLabel(_ for: Int, for in: Int, class _: Int) {
-// CHECK:  <kw>func</kw> <id>keywordInCaseAndLocalArgLabel</id>(<kw>_</kw> <id>for</id>: <type>Int</type>, <id>for</id> <id>in</id>: <type>Int</type>, <id>class</id> <kw>_</kw>: <type>Int</type>) {
-  switch(`for`, `in`) {
```
ahoppen, line 378: I think some test case with escaped identifiers would be good to keep.

ahoppen (commented):
I just looked through the test cases that you didn’t translate to XCTest and I agree that most of them are redundant or pointless with the new SwiftParser implementation. The only exception is that I’d like to have a test case where an escaped keyword is used as an identifier.

ahoppen:
Otherwise: Thanks a lot for getting rid of `lit-test-helper` 🚀

lit_tests/coloring.swift
StevenWong12: I think that's good👍. I added two in my new commit in https://github.com/apple/swift-syntax/pull/1953.

ahoppen:
@swift-ci Please test

ahoppen (approved):

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test macOS

Issue #1825, opened 2023-06-21
Parse shebang in source file as an child of 'SourceFile'


ahoppen:
Instead of parsing a `#!` shebang in a source file as trivia, we should represent it in the source tree as a list of tokens.

rdar://98309158

PR #1979, opened 2023-08-02
Parse hashbang as a child of `SourceFileSyntax`

StevenWong12:
Resolves https://github.com/apple/swift-syntax/issues/1825.

Instead of parsing hashbang into a token list, I think the easier way is to parse the whole as an identifier-like token since we don't care about the content of it.  

Also, we could make use of `Lexer.Cursor.advanceToEndOfLine` to lex the first line as a hashbang token instead of doing this in the parser. 

What do you think about it?

### DIFF
--- a/CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
@@ -1998,6 +1998,12 @@ public let DECL_NODES: [Node] = [
     traits: ["WithStatements"],
     children: [
       Child(
+        name: "Shebang",
+        kind: .token(choices: [.token(.shebang)]),
+        documentation: "A shebang can specify the path of the compiler when using Swift source file as a script.",
+        isOptional: true
+      ),
+      Child(
         name: "Statements",
         kind: .collection(kind: .codeBlockItemList, collectionElementName: "Statement")
       ),
--- a/Sources/SwiftIDEUtils/SyntaxClassification.swift
+++ b/Sources/SwiftIDEUtils/SyntaxClassification.swift
@@ -193,6 +193,8 @@ extension RawTokenKind {
       return .none
     case .semicolon:
       return .none
+    case .shebang:
+      return .none
     case .singleQuote:
       return .stringLiteral
     case .stringQuote:
--- a/Sources/SwiftParser/Lexer/Cursor.swift
+++ b/Sources/SwiftParser/Lexer/Cursor.swift
@@ -905,6 +905,11 @@ extension Lexer.Cursor {
     case UInt8(ascii: "\\"): _ = self.advance(); return Lexer.Result(.backslash)
 
     case UInt8(ascii: "#"):
+      // Try lex shebang.
+      if self.isAtStartOfFile, self.peek(at: 1) == UInt8(ascii: "!") {
+        self.advanceToEndOfLine()
+        return Lexer.Result(.shebang)
+      }
       // Try lex a raw string literal.
       if let delimiterLength = self.advanceIfOpeningRawStringDelimiter() {
         return Lexer.Result(.rawStringPoundDelimiter, stateTransition: .push(newState: .afterRawStringDelimiter(delimiterLength: delimiterLength)))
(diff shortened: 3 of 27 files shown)


CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
```diff
@@ -1996,6 +1996,11 @@ public let DECL_NODES: [Node] = [
     parserFunction: "parseSourceFile",
     traits: ["WithStatements"],
     children: [
+      Child(
+        name: "Hashbang",
```
ahoppen, line 2000: Could you add some documentation for this child? Just briefly describe in which cases the shebang can exist and what it means in terms of Swift.

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
```diff
@@ -1996,6 +1996,11 @@ public let DECL_NODES: [Node] = [
...
+        kind: .token(choices: [.token(tokenKind: "IdentifierToken")]),
```
ahoppen, line 2001: This would need to be `HashbangToken`, right? Because you’re not parsing this as an `IdentifierToken`. This should fail if you compile swift-syntax with `SWIFTSYNTAX_ENABLE_RAWSYNTAX_VALIDATION`.

Sources/SwiftIDEUtils/SyntaxClassification.swift
```diff
@@ -128,6 +128,8 @@ extension RawTokenKind {
       return .none
     case .floatingLiteral:
       return .floatingLiteral
+    case .hashbang:
+      return .poundDirective
```
ahoppen, line 132: I’m not sure that makes sense because it’s not a pound directive in the Swift sense. I would just classify it as `none`.

Sources/SwiftParser/Lexer/Cursor.swift
```diff
@@ -905,6 +905,11 @@ extension Lexer.Cursor {
     case UInt8(ascii: "\\"): _ = self.advance(); return Lexer.Result(.backslash)
 
     case UInt8(ascii: "#"):
+      // Try lex hashbang
+      if self.isAtStartOfFile, self.peek(at: 1) == UInt8(ascii: "!") {
+        self.advanceToEndOfLine()
+        return Lexer.Result(.hashbang)
+      }
```
ahoppen, line 912: Oh, nice. That’s so simple.

ahoppen (commented):
Looks great to me. One general note: I think the standard term here is [Shebang](https://en.wikipedia.org/wiki/Shebang_(Unix)) instead of hashbang.

Sources/SwiftIDEUtils/SyntaxClassification.swift
StevenWong12: Oh, that's because I found hashbang is colored the same as other macros in Xcode so I think we could classify it as `.poundDirective` and make it highlighted the same as other macros in sourcekit-lsp.

ahoppen: Hmm, I think we can deviate here. `poundDirective` might make sense now but it makes less sense if we rename `poundDirective` to `ifConfigDirective` as discussed offline. 

If you do want to color it, I would color it as a line comment, because that’s kind-of what it is: The `#` indicates a comment in bash, I think that’s what the `#!` has evolved from and Swift just supports it as well.

StevenWong12: Ah, that makes sense. I'll just leave it as `none` for now.

Sources/SwiftParser/Lexer/Cursor.swift
```diff
@@ -1172,11 +1177,7 @@ extension Lexer.Cursor {
           break
         }
       case UInt8(ascii: "#"):
-        guard start.isAtStartOfFile, self.advance(if: { $0 == "!" }) else {
-          break
-        }
-        self.advanceToEndOfLine()
-        continue
+        break
```
ahoppen, line 1180: Actually, now that I read this again, we can just add `UInt8(ascii: "#")` to the `Punctuation` section a few lines further below. It doesn’t need its own case anymore.

ahoppen (commented):
Oh, one more general comment: Could you add an `assertParse` text for the shebang? You might need to adjust `BasicFormat.requiresNewline` as well to make sure it adds a newline after the shebang.

StevenWong12:
I think we already have an `assertParse` test in `testHashbangLibrary1`.😉

ahoppen:
> I think we already have an assertParse test in testHashbangLibrary1.😉

Oh, I didn’t see that. Thanks for pointing it out

ahoppen:
@swift-ci Please test

ahoppen (approved):

StevenWong12:
I created a corresponding pr in https://github.com/apple/swift-format/pull/590. Could you trigger CI again?

kimdv:
@swift-ci please test

allevato:
https://github.com/apple/swift-format/pull/590

@swift-ci please test

PR #1982, opened 2023-08-02
[reference-binding] Add _consuming/_mutating/_borrowing as var decl bindings

gottesmm:
Title says it all.

### DIFF
--- a/CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
@@ -1163,6 +1163,9 @@ public let DECL_NODES: [Node] = [
           .keyword(.let),
           .keyword(.func),
           .keyword(.inout),
+          .keyword(._mutating),
+          .keyword(._consuming),
+          .keyword(._borrowing),
         ]),
         documentation: """
           The kind of declaration being imported. 
@@ -2330,7 +2333,10 @@ public let DECL_NODES: [Node] = [
       Child(
         name: "bindingSpecifier",
         deprecatedName: "bindingKeyword",
-        kind: .token(choices: [.keyword(.let), .keyword(.var), .keyword(.inout)]),
+        kind: .token(choices: [
+          .keyword(.let), .keyword(.var), .keyword(.inout),
+          .keyword(._mutating), .keyword(._borrowing), .keyword(._consuming),
+        ]),
         documentation: """
           The specifier that defines the type of the variables declared (`let` or `var`).
           """
--- a/CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
@@ -1163,9 +1163,6 @@ public let DECL_NODES: [Node] = [
           .keyword(.let),
           .keyword(.func),
           .keyword(.inout),
-          .keyword(._mutating),
-          .keyword(._consuming),
-          .keyword(._borrowing),
         ]),
         documentation: """
           The kind of declaration being imported. 
--- a/CodeGeneration/Sources/SyntaxSupport/StmtNodes.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/StmtNodes.swift
@@ -435,7 +435,10 @@ public let STMT_NODES: [Node] = [
       Child(
         name: "bindingSpecifier",
         deprecatedName: "bindingKeyword",
-        kind: .token(choices: [.keyword(.let), .keyword(.var), .keyword(.inout)])
+        kind: .token(choices: [
+          .keyword(.let), .keyword(.var), .keyword(.inout), .keyword(._mutating), .keyword(._borrowing),
+          .keyword(._consuming),
+        ])
       ),
       Child(
         name: "pattern",
--- a/Sources/SwiftParser/Declarations.swift
+++ b/Sources/SwiftParser/Declarations.swift
@@ -234,7 +234,10 @@ extension Parser {
     case (.subscript, let handle)?:
       return RawDeclSyntax(self.parseSubscriptDeclaration(attrs, handle))
     case (.let, let handle)?, (.var, let handle)?,
-      (.inout, let handle)?:
+      (.inout, let handle)?,
+      (._borrowing, let handle)?,
+      (._mutating, let handle)?,
+      (._consuming, let handle)?:
       return RawDeclSyntax(self.parseBindingDeclaration(attrs, handle, inMemberDeclList: inMemberDeclList))
     case (.`init`, let handle)?:
       return RawDeclSyntax(self.parseInitializerDeclaration(attrs, handle))
--- a/Sources/SwiftParser/Declarations.swift
+++ b/Sources/SwiftParser/Declarations.swift
@@ -106,7 +106,7 @@ extension TokenConsumer {
       declStartKeyword = subparser.at(anyIn: DeclarationKeyword.self)?.0
     }
     switch declStartKeyword {
-    case .actor:
+    case .lhs(.actor):
       // actor Foo {}
       if subparser.peek().rawTokenKind == .identifier {
         return true
@@ -119,16 +119,16 @@ extension TokenConsumer {
         lookahead.consumeAnyToken()
       } while lookahead.atStartOfDeclaration(isAtTopLevel: isAtTopLevel, allowInitDecl: allowInitDecl)
       return lookahead.at(.identifier)
-    case .case:
+    case .lhs(.case):
       // When 'case' appears inside a function, it's probably a switch
       // case, not an enum case declaration.
       return false
-    case .`init`:
+    case .lhs(.`init`):
       return allowInitDecl
-    case .macro:
+    case .lhs(.macro):
       // macro Foo ...
       return subparser.peek().rawTokenKind == .identifier
-    case .pound:
+    case .lhs(.pound):
       // Force parsing '#<identifier>' after attributes as a macro expansion decl.
       if hasAttribute || hasModifier {
         return true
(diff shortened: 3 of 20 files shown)


gottesmm:
@swift-ci test

gottesmm:
@swift-ci test

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
```diff
@@ -1179,6 +1179,9 @@ public let DECL_NODES: [Node] = [
           .keyword(text: "let"),
           .keyword(text: "func"),
           .keyword(text: "inout"),
+          .keyword(text: "mutating"),
+          .keyword(text: "consuming"),
+          .keyword(text: "borrowing"),
```
ahoppen, line 1184: Is `import mutating Module.var` something we want to support? Same for `consuming`. Could you also add test cases for it?

CodeGeneration/Sources/SyntaxSupport/StmtNodes.swift
```diff
@@ -440,7 +440,10 @@ public let STMT_NODES: [Node] = [
       Child(
         name: "BindingSpecifier",
         deprecatedName: "BindingKeyword",
-        kind: .token(choices: [.keyword(text: "let"), .keyword(text: "var"), .keyword(text: "inout")])
+        kind: .token(choices: [
+          .keyword(text: "let"), .keyword(text: "var"), .keyword(text: "inout"),
+          .keyword(text: "mutating"), .keyword(text: "borrowing"), .keyword(text: "consuming"),
+        ])
```
ahoppen, line 446: AFAICT we don’t have tests for these. Could you add some? Should be `if mutating x = x {}`

Sources/SwiftParser/TokenSpecSet.swift
```diff
@@ -405,7 +405,7 @@ enum DeclarationModifier: TokenSpecSet {
     case .internal: return .keyword(.internal)
     case .isolated: return TokenSpec(.isolated, recoveryPrecedence: .declKeyword)
     case .lazy: return .keyword(.lazy)
-    case .mutating: return .keyword(.mutating)
+    case .mutating: return TokenSpec(.mutating, recoveryPrecedence: .declKeyword)
```
ahoppen, line 408: `mutating` already has a `TokenPrecedence` of `declKeyword`, so I don’t think this change is needed.

https://github.com/apple/swift-syntax/blob/250b431726c7b05309770caaa9989a4166ea6ba7/Sources/SwiftParser/TokenPrecedence.swift#L229-L233

Sources/SwiftParser/Patterns.swift
```diff
@@ -93,7 +102,10 @@ extension Parser {
       )
     case (.let, let handle)?,
       (.var, let handle)?,
-      (.inout, let handle)?:
+      (.inout, let handle)?,
+      (.mutating, let handle)?,
+      (.borrowing, let handle)?,
+      (.consuming, let handle)?:
```
ahoppen, line 108: You don’t need to add all the cases to `ExpectedTokens` here, if you 
- Remove `let`, `var`, `inout` from `ExpectedTokens`
- Add a new `switch self.at(anyIn: ValueBindingPatternSyntax.BindingSpecifierOptions.self) {` here.

`ValueBindingPatternSyntax.BindingSpecifierOptions` is generated from the options you specify in `CodeGeneration`. It’s a recent-ish feature, that’s why it isn’t used yet. But a great opportunity to clean up the codebase a little.

Sources/SwiftParser/Patterns.swift
```diff
@@ -217,7 +229,10 @@ extension Parser {
     switch self.at(anyIn: MatchingPatternStart.self) {
     case (.var, let handle)?,
       (.let, let handle)?,
-      (.inout, let handle)?:
+      (.inout, let handle)?,
+      (.mutating, let handle)?,
+      (.borrowing, let handle)?,
+      (.consuming, let handle)?:
```
ahoppen, line 235: Same comment of using `ValueBindingPatternSyntax.BindingSpecifierOptions` also applies here.

Sources/SwiftParser/Patterns.swift
```diff
@@ -274,6 +289,9 @@ extension Parser.Lookahead {
       case `var`
       case leftParen
       case `inout`
+      case `mutating`
+      case borrowing
+      case consuming
```
ahoppen, line 294: Same comment of using `ValueBindingPatternSyntax.BindingSpecifierOptions` applies here.

ahoppen (commented):
Mostly looks good to me. I’ve got one suggestion to simplify the `TokenSpecSets` by using the auto-generated ones.

Also, since you are touching the test cases, would you mind splitting them so that each statement/case/variable its own `assertParse` call? It just makes it easier to debug the tests in case one of them failed because you start off with less source and you don’t actually need to reduce the input.

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
gottesmm: @ahoppen are there any tests like this that exist already? I did a git grep for import var/let/etc.

ahoppen: I don’t think so. DeclarationTests.swift seems like a good place to add them though

gottesmm:
@swift-ci test

Sources/SwiftParser/TokenSpecSet.swift
```diff
@@ -21,6 +21,38 @@ protocol TokenSpecSet: CaseIterable {
...
+
+  init?(lexeme: Lexer.Lexeme) {
+    if let x = LHS.init(lexeme: lexeme) {
+      self = .lhs(x)
+      return
+    }
+    if let y = RHS.init(lexeme: lexeme) {
+      self = .rhs(y)
+      return
+    }
```
ahoppen, line 37: Nit: You don’t need `.init`

```suggestion
    if let x = LHS(lexeme: lexeme) {
      self = .lhs(x)
      return
    }
    if let y = RHS(lexeme: lexeme) {
      self = .rhs(y)
      return
    }
```

Sources/SwiftParser/TokenSpecSet.swift
```diff
@@ -227,7 +259,10 @@ enum ContextualDeclKeyword: TokenSpecSet {
   }
 }
 
-enum DeclarationKeyword: TokenSpecSet {
+/// DeclarationKeywords that are only semantic DeclarationKeywords.
+///
+/// Excludes things like ValueBindingPatterns. They are injected into DeclarationKeyword via an either.
```
ahoppen, line 264:
```suggestion
/// Excludes things like ValueBindingPatterns. They are injected into `DeclarationKeyword` via an `EitherTokenSpecSet`
```

Sources/SwiftParser/TokenSpecSet.swift
```diff
@@ -719,32 +750,31 @@ enum ExpressionPrefixOperator: TokenSpecSet {
   }
 }
 
-enum MatchingPatternStart: TokenSpecSet {
+/// MatchingPatternStart without ValueBindingPatterns.
+///
+/// We use an EitherTokenSpecSet to inject ValueBindingPatterns.
```
ahoppen, line 755:
```suggestion
/// `MatchingPatternStart` without `ValueBindingPatterns`.
///
/// We use an `EitherTokenSpecSet` to inject `ValueBindingPatterns`.
```

ahoppen (commented):
Thanks for adding `EitherTokenSpecSet`. I think it will start to be useful in other cases as well.

Looks good to me, I just think that we should be guarding the feature behind `perimentalFeatures.contains(.whateverYouWantToCallThisFeature)`.

Sources/SwiftParser/TokenSpecSet.swift
```diff
@@ -227,7 +259,10 @@ enum ContextualDeclKeyword: TokenSpecSet {
   }
 }
 
-enum DeclarationKeyword: TokenSpecSet {
+/// DeclarationKeywords that are only semantic DeclarationKeywords.
```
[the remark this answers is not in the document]
gottesmm: What I am trying to say here is that they are not ValueBinding ones. I will fix that.

gottesmm:
@swift-ci test

Sources/SwiftParser/Declarations.swift
```diff
@@ -210,81 +210,99 @@ extension Parser {
     // while recovering to the declaration start.
     let recoveryPrecedence = inMemberDeclList ? TokenPrecedence.closingBrace : nil
 
-    switch self.canRecoverTo(anyIn: DeclarationKeyword.self, overrideRecoveryPrecedence: recoveryPrecedence) {
-    case (.import, let handle)?:
+    CanRecoverToSwitch: switch self.canRecoverTo(anyIn: DeclarationKeyword.self, overrideRecoveryPrecedence: recoveryPrecedence) {
```
ahoppen, line 213: I don’t actually think we need the label. If we `break` in the nested `switch`, then it just executes to the end of the outer `switch`, which is the same as `break CanRecoverToSwitch`.

---

If we do need it: We currently all caps our labels, so for consistency, this should be `CAN_REVOVER_TO_SWITCH`.

Sources/SwiftParser/Declarations.swift
```diff
@@ -210,81 +210,99 @@ extension Parser {
...
-    case (.pound, let handle)?:
+    case (.lhs(.pound), let handle)?:
       return RawDeclSyntax(self.parseMacroExpansionDeclaration(attrs, handle))
+    case (.rhs(let binding), let handle)?:
+      switch binding {
+      case ._mutating:
+        fallthrough
+      case ._borrowing:
+        fallthrough
+      case ._consuming:
```
ahoppen, line 256: Slightly shorter version

```suggestion
      case ._mutating, ._borrowing, ._consuming:
```

Sources/SwiftParser/Declarations.swift
```diff
@@ -210,81 +210,99 @@ extension Parser {
...
+        guard experimentalFeatures.contains(.referenceBindings) else {
+          break CanRecoverToSwitch
+        }
+        fallthrough
+      case .let:
+        fallthrough
+      case .var:
+        fallthrough
+      case .inout:
```
ahoppen, line 265: Slightly shorter version

```suggestion
      case .let, .var, .inout:
```

Sources/SwiftParser/Patterns.swift
```diff
@@ -91,32 +86,48 @@ extension Parser {
...
-      if self.currentToken.isLexerClassifiedKeyword, !self.atStartOfLine {
-        // Recover if a keyword was used instead of an identifier
-        let keyword = self.consumeAnyToken()
+    case (.rhs(let binding), let handle)?:
+      switch binding {
+      case ._mutating:
+        fallthrough
+      case ._borrowing:
+        fallthrough
+      case ._consuming:
```
ahoppen, line 95:
```suggestion
      case ._mutating, ._borrowing, ._consuming:
```

Sources/SwiftParser/Patterns.swift
```diff
@@ -91,32 +86,48 @@ extension Parser {
...
+        guard experimentalFeatures.contains(.referenceBindings) else {
+          break
+        }
+        fallthrough
+      case .let:
+        fallthrough
+      case .var:
+        fallthrough
+      case .inout:
```
ahoppen, line 104:
```suggestion
      case .let, .var, .inout:
```

Sources/SwiftParser/Patterns.swift
```diff
@@ -214,20 +225,8 @@ extension Parser {
...
-      let value = self.parseMatchingPattern(context: .bindingIntroducer)
-      return RawPatternSyntax(
-        RawValueBindingPatternSyntax(
-          bindingSpecifier: bindingSpecifier,
-          pattern: value,
-          arena: self.arena
-        )
-      )
-    case (.is, let handle)?:
+    OuterSwitch: switch self.at(anyIn: MatchingPatternStart.self) {
```
ahoppen, line 228: I don’t actually think we need the label. If we `break` in the nested `switch`, then it just executes to the end of the outer `switch`, which is the same as `break OuterSwitch`.

---

If we do need it: We currently all caps our labels, so for consistency, this should be `OUTER_SWITCH`.

Sources/SwiftParser/Patterns.swift
```diff
@@ -237,23 +236,52 @@ extension Parser {
...
-        // FIXME: This is pretty gross. Let's find a way to disambiguate let
-        // binding patterns much earlier.
-        return RawPatternSyntax(pat.pattern)
+    case (.rhs(let binding), let handle)?:
+      switch binding {
+      case ._mutating:
+        fallthrough
+      case ._borrowing:
+        fallthrough
+      case ._consuming:
```
ahoppen, line 245:
```suggestion
      case ._mutating, ._borrowing, ._consuming:
```

Sources/SwiftParser/Patterns.swift
```diff
@@ -237,23 +236,52 @@ extension Parser {
...
+        guard experimentalFeatures.contains(.referenceBindings) else {
+          break OuterSwitch
+        }
+        fallthrough
+      case .let:
+        fallthrough
+      case .var:
+        fallthrough
+      case .inout:
```
ahoppen, line 254:
```suggestion
      case .let, .var, .inout:
```

Sources/SwiftParser/Declarations.swift
```diff
@@ -210,81 +210,99 @@ extension Parser {
...
+        // Handle all cases of `ValueBindingPatternSyntax.BindingSpecifierOptions` in the same way.
```
ahoppen, line 266: I don’t think we need this comment anymore if we exhaustively switch over `rhs`.

Sources/SwiftParser/Patterns.swift
```diff
@@ -91,32 +86,48 @@ extension Parser {
...
+        guard experimentalFeatures.contains(.referenceBindings) else {
+          break
+        }
+        fallthrough
+      case .let:
+        fallthrough
+      case .var:
+        fallthrough
+      case .inout:
+        // Handle all cases of `ValueBindingPatternSyntax.BindingSpecifierOptions` in the same way.
```
ahoppen, line 105: I don’t think we need this comment anymore if we exhaustively switch over `rhs`.

Sources/SwiftParser/Patterns.swift
```diff
@@ -291,25 +313,40 @@ extension Parser.Lookahead {
...
+      case ._consuming:
+        guard experimentalFeatures.contains(.referenceBindings) else {
+          return false
+        }
+        fallthrough
+      case .let:
+        fallthrough
+      case .var:
+        fallthrough
+      case .inout:
```
ahoppen, line 345:
```suggestion
      case .let, .var, .inout:
```

Sources/SwiftParser/Patterns.swift
```diff
@@ -291,25 +313,40 @@ extension Parser.Lookahead {
...
-      self.eat(handle)
-      return self.canParsePattern()
-    case (.leftParen, _)?:
+    case (.rhs(let binding), let handle)?:
+      switch binding {
+      case ._mutating:
+        fallthrough
+      case ._borrowing:
+        fallthrough
+      case ._consuming:
```
ahoppen, line 336:
```suggestion
      case ._mutating, ._borrowing, ._consuming:
```

ahoppen (approved):
Looks great to me. Just a couple syntactic improvement suggestions. If you want to land this now, you can do them in a follow-up PR.

ahoppen:
@swift-ci Please test Windows

PR #1986, opened 2023-08-02
Disallow the `self`/`Self` keyword where it’s not allowed in types

ahoppen:
We weren’t matching the C++ parser’s behavior here. The `self` keyword is not allowed as a type name. `Self` in a nested type refers to a type named `Self` instead of the `Self` keyword.


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Tests/SwiftParserTest/TypeTests.swift
```diff
@@ -192,4 +192,42 @@ final class TypeTests: ParserTestCase {
...
+
+  func testLowercaseSelf() {
+    assertParse(
+      "1️⃣self2️⃣",
+      { TypeSyntax.parse(from: &$0) },
+      diagnostics: [
+        DiagnosticSpec(locationMarker: "1️⃣", message: "unexpected 'self' keyword before type"),
+        DiagnosticSpec(locationMarker: "2️⃣", message: "expected type", fixIts: ["insert type"]),
+      ],
+      fixedSource: "<#type#>self"
```
bnbarham, line 204: `<#type#>.self`?

Tests/SwiftParserTest/AttributeTests.swift
```diff
@@ -220,6 +220,24 @@ final class AttributeTests: ParserTestCase {
       ) {}
       """
     )
+
+    assertParse(
+      """
+      @derivitave(of: 1️⃣Self.other)
```
bnbarham, line 226:
```suggestion
      @derivative(of: 1️⃣Self.other)
```

Tests/SwiftParserTest/TypeTests.swift
ahoppen: We consume `self` as unexpected here. I’m not sure if `<#type#>.self` is really any more useful than what we currently produce, so not sure if it’s worth adding a special case for this recovery. 

To put it into context, if you write `let a: self`, I don’t think that `let a: <#type#>.self` is a meaningful recovery strategy. We could suggest wrapping `self` in backticks here, but I’m 100% sure that’s not what the developer intended either.

ahoppen:
@swift-ci Please test

bnbarham (approved):

ahoppen:
@swift-ci Please test Windows

PR #2004, opened 2023-08-04
Add conformsTo argument to member macro expansion operation

DougGregor:
Stage in an entrypoint for member macros that allows them to learn about which conformances that they've asked about are "missing", meaning that they are not present on the type (ignoring those that would be generated by an extension macro). This information is equivalent to the information provided to extension macros, although the member macro itself cannot create the conformance.

### DIFF
--- a/Sources/SwiftSyntaxMacroExpansion/MacroExpansion.swift
+++ b/Sources/SwiftSyntaxMacroExpansion/MacroExpansion.swift
@@ -236,6 +236,7 @@ public func expandAttachedMacroWithoutCollapsing<Context: MacroExpansionContext>
       let members = try attachedMacro.expansion(
         of: attributeNode,
         providingMembersOf: declGroup,
+        conformingTo: conformanceList?.map(\.type) ?? [],
         in: context
       )
 
--- a/Sources/SwiftSyntaxMacros/MacroProtocols/MemberMacro.swift
+++ b/Sources/SwiftSyntaxMacros/MacroProtocols/MemberMacro.swift
@@ -23,9 +23,54 @@ public protocol MemberMacro: AttachedMacro {
   ///
   /// - Returns: the set of member declarations introduced by this macro, which
   /// are nested inside the `attachedTo` declaration.
+  @available(*, deprecated, message: "Use expansion(of:providingMembersOf:conformingTo:in:")
   static func expansion(
     of node: AttributeSyntax,
     providingMembersOf declaration: some DeclGroupSyntax,
     in context: some MacroExpansionContext
   ) throws -> [DeclSyntax]
+
+  /// Expand an attached declaration macro to produce a set of members.
+  ///
+  /// - Parameters:
+  ///   - node: The custom attribute describing the attached macro.
+  ///   - declaration: The declaration the macro attribute is attached to.
+  ///   - conformingTo: The set of protocols that were declared
+  ///     in the set of conformances for the macro and to which the declaration
+  ///     does not explicitly conform. The member macro itself cannot declare
+  ///     conformances to these protocols (only an extension macro can do that),
+  ///     but can provide supporting declarations, such as a required
+  ///     initializer or stored property, that cannot be written in an
+  ///     extension.
+  ///   - context: The context in which to perform the macro expansion.
+  ///
+  /// - Returns: the set of member declarations introduced by this macro, which
+  /// are nested inside the `attachedTo` declaration.
+  static func expansion(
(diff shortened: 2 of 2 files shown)


DougGregor:
@swift-ci please test

Sources/SwiftSyntaxMacros/MacroProtocols/MemberMacro.swift
```diff
@@ -28,4 +28,39 @@ public protocol MemberMacro: AttachedMacro {
...
+  /// Default implementation that ignores the unhandled conformances.
+  static func expansion(
+    of node: AttributeSyntax,
+    providingMembersOf declaration: some DeclGroupSyntax,
+    missingConformancesTo protocols: [TypeSyntax],
+    in context: some MacroExpansionContext
+  ) throws -> [DeclSyntax] {
+    return try expansion(of: node, providingMembersOf: declaration, in: context)
+  }
 }
```
ahoppen, line 66: AFAICT this means that you always need to implement `expansion(of:providingMembersOf:in:)`, even if you implement `expansion(of:providingMembersOf:missingConformancesTo:in:)`. I don’t have a better idea but just want to put this fact out there.

Sources/SwiftSyntaxMacroExpansion/MacroExpansion.swift
```diff
@@ -236,6 +236,7 @@ public func expandAttachedMacroWithoutCollapsing<Context: MacroExpansionContext>
       let members = try attachedMacro.expansion(
         of: attributeNode,
         providingMembersOf: declGroup,
+        missingConformancesTo: conformanceList?.map(\.type) ?? [],
```
ahoppen, line 239: Why is this parameter called `missingConformancesTo:`? AFAICT you are passing in the existing conformances, not some list of missing conformances.

Sources/SwiftSyntaxMacros/MacroProtocols/MemberMacro.swift
DougGregor: Yeah, I can create a default implementation in the other direction and deprecate one.

Sources/SwiftSyntaxMacroExpansion/MacroExpansion.swift
DougGregor: I'm struggling mightily to name this. Let's discussion naming over at https://forums.swift.org/t/pitch-member-macros-that-know-what-conformances-are-missing/66590

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please clean test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test Windows

ahoppen:
Can you add this change to the release notes?

Sources/SwiftSyntaxMacros/MacroProtocols/MemberMacro.swift
```diff
@@ -23,9 +23,54 @@ public protocol MemberMacro: AttachedMacro {
...
+
+  /// Default implementation that ignores the unhandled conformances.
+  static func expansion(
+    of node: AttributeSyntax,
+    providingMembersOf declaration: some DeclGroupSyntax,
+    conformingTo protocols: [TypeSyntax],
+    in context: some MacroExpansionContext
+  ) throws -> [DeclSyntax] {
+    return try expansion(of: node, providingMembersOf: declaration, in: context)
+  }
```
ahoppen, line 75: I don’t think adding default implementations that forward in both directions is a good way. If I write the following,

```swift
public struct X: MemberMacro {}
```

it compiles just fine without any error that I’m missing any protocol requirement and the call to the expansion function will just infinitely recurse.

ahoppen (commented):
Could you add the API change to the release notes and add a test case for it?

Sources/SwiftSyntaxMacros/MacroProtocols/MemberMacro.swift
DougGregor: I can stage out one of them, but for now I want forward/backward compatibility.

DougGregor:
> Could you add the API change to the release notes and add a test case for it?

I can do this in a follow-up

PR #2006, opened 2023-08-05
Replace string-based token references with `Token` enum in CodeGeneration

### LINKED ISSUE #1978: Remove string-based reference of tokens in CodeGeneration
CodeGeneration currently refers to tokens e.g. as `IdentifierToken`. That string-based reference isn’t very idiomatic Swift. What we should have instead is a `Token` enum of all the possible tokens in `TokenSpec.swift`, a computed property on that enum that returns the corresponding `TokenSpec` of that token. We could then use members of the `Token` enum to refer to the token kinds, instead of the string-based reference.

Matejkob:
The CodeGeneration module was previously referring to tokens using string-based references like `IdentifierToken`, which is not idiomatic Swift. To make the code more concise and in line with Swift's best practices, this commit introduces a `Token` enum in the `TokenSpec.swift` file. 
Now, instead of string-based references, the new `Token` enum contains all the possible token kinds. It also includes a computed property that returns the corresponding `TokenSpec` for a given token, allowing for more concise and idiomatic references. 
The code has been refactored to replace all string-based token references with members of the newly introduced `Token` enum. Associated tests have also been updated to reflect these changes.

Resolves #1978

### DIFF
--- a/CodeGeneration/Sources/SyntaxSupport/AttributeNodes.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/AttributeNodes.swift
@@ -39,7 +39,7 @@ public let ATTRIBUTE_NODES: [Node] = [
       Child(
         name: "AtSign",
         deprecatedName: "AtSignToken",
-        kind: .token(choices: [.token(tokenKind: "AtSignToken")]),
+        kind: .token(choices: [.token(.atSign)]),
         documentation: "The `@` sign."
       ),
       Child(
@@ -50,7 +50,7 @@ public let ATTRIBUTE_NODES: [Node] = [
       ),
       Child(
         name: "LeftParen",
-        kind: .token(choices: [.token(tokenKind: "LeftParenToken")]),
+        kind: .token(choices: [.token(.leftParen)]),
         documentation: "If the attribute takes arguments, the opening parenthesis.",
         isOptional: true
       ),
@@ -144,7 +144,7 @@ public let ATTRIBUTE_NODES: [Node] = [
       ),
       Child(
         name: "RightParen",
-        kind: .token(choices: [.token(tokenKind: "RightParenToken")]),
+        kind: .token(choices: [.token(.rightParen)]),
         documentation: "If the attribute takes arguments, the closing parenthesis.",
         isOptional: true
       ),
@@ -166,7 +166,7 @@ public let ATTRIBUTE_NODES: [Node] = [
       ),
       Child(
         name: "Colon",
-        kind: .token(choices: [.token(tokenKind: "ColonToken")]),
+        kind: .token(choices: [.token(.colon)]),
         documentation: "The colon separating the label and the value"
       ),
       Child(
@@ -176,7 +176,7 @@ public let ATTRIBUTE_NODES: [Node] = [
       ),
       Child(
         name: "Semicolon",
-        kind: .token(choices: [.token(tokenKind: "SemicolonToken")])
+        kind: .token(choices: [.token(.semicolon)])
       ),
     ]
   ),
@@ -196,7 +196,7 @@ public let ATTRIBUTE_NODES: [Node] = [
       ),
       Child(
         name: "TrailingComma",
-        kind: .token(choices: [.token(tokenKind: "CommaToken")]),
+        kind: .token(choices: [.token(.comma)]),
         documentation: "A trailing comma if the argument is followed by another argument",
         isOptional: true
       ),
@@ -227,7 +227,7 @@ public let ATTRIBUTE_NODES: [Node] = [
       ),
       Child(
         name: "Colon",
-        kind: .token(choices: [.token(tokenKind: "ColonToken")]),
+        kind: .token(choices: [.token(.colon)]),
         documentation: "The colon separating \"before\" and the parameter list."
       ),
       Child(
@@ -248,12 +248,12 @@ public let ATTRIBUTE_NODES: [Node] = [
     children: [
       Child(
         name: "ConventionLabel",
-        kind: .token(choices: [.token(tokenKind: "IdentifierToken")]),
+        kind: .token(choices: [.token(.identifier)]),
         documentation: "The convention label."
       ),
       Child(
         name: "Comma",
-        kind: .token(choices: [.token(tokenKind: "CommaToken")]),
+        kind: .token(choices: [.token(.comma)]),
         isOptional: true
       ),
       Child(
@@ -263,7 +263,7 @@ public let ATTRIBUTE_NODES: [Node] = [
       ),
       Child(
         name: "Colon",
-        kind: .token(choices: [.token(tokenKind: "ColonToken")]),
+        kind: .token(choices: [.token(.colon)]),
         isOptional: true
       ),
       Child(
@@ -287,11 +287,11 @@ public let ATTRIBUTE_NODES: [Node] = [
       ),
       Child(
         name: "Colon",
-        kind: .token(choices: [.token(tokenKind: "ColonToken")])
+        kind: .token(choices: [.token(.colon)])
       ),
       Child(
         name: "ProtocolName",
-        kind: .token(choices: [.token(tokenKind: "IdentifierToken")])
+        kind: .token(choices: [.token(.identifier)])
       ),
     ]
   ),
@@ -316,7 +316,7 @@ public let ATTRIBUTE_NODES: [Node] = [
       ),
       Child(
         name: "Colon",
-        kind: .token(choices: [.token(tokenKind: "ColonToken")]),
+        kind: .token(choices: [.token(.colon)]),
         documentation: "The colon separating the \"of\" label and the original declaration name."
       ),
       Child(
@@ -326,7 +326,7 @@ public let ATTRIBUTE_NODES: [Node] = [
       ),
       Child(
         name: "Period",
-        kind: .token(choices: [.token(tokenKind: "PeriodToken")]),
+        kind: .token(choices: [.token(.period)]),
         documentation: "The period separating the original declaration name and the accessor name.",
         isOptional: true
       ),
@@ -339,7 +339,7 @@ public let ATTRIBUTE_NODES: [Node] = [
       ),
       Child(
         name: "Comma",
-        kind: .token(choices: [.token(tokenKind: "CommaToken")]),
+        kind: .token(choices: [.token(.comma)]),
         isOptional: true
       ),
       Child(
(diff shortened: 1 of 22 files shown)


CodeGeneration/Sources/SyntaxSupport/Child.swift
```diff
@@ -113,20 +113,18 @@ public class Child {
...
         }
+      } else if choices.allSatisfy(\.isKeyword) {
+        return .keyword
       } else {
-        if choices.allSatisfy({ $0.isKeyword }) {
-          return "KeywordToken"
-        } else {
-          return "Token"
-        }
+        return .unknown
```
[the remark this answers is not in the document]
ahoppen: I agree that this is super confusing. My hope is that we can just remove it altogether (I would do it in a follow-up PR, not this one). 

The only non-trivial place that needs to be updated is `Child.type` in `SyntaxBuildableChild` and, IMHO, the confusion stems from the fact that we try to collapse multiple token kinds into a single one without giving any indication of doing so in the variable names.

I think the way to go about fixing this is to change the associated value of `SyntaxOrTokenNodeKind.token` to `[TokenChoice]`. Then we won’t need to call `Child.tokenKind` in `Child.type` anymore and could shift the computation that’s currently done in this property down to uses where there’s a lot more context. Potentially, the computation might also just become a lot simpler once we see that context.

I think the implementation is good enough for now. Maybe just add a comment along the lines

```swift
// FIXME: Technically, returning `.unknown` is not correct here. The old string-based implementation returned "Token", which no longer exists if we model the token kinds as an enum in code generation. We should try to remove this computed property altogether <link to this discussion>
```

CodeGeneration/Sources/SyntaxSupport/TokenSpec.swift
```diff
@@ -66,58 +66,159 @@ public struct TokenSpec {
...
-  .other(name: "stringSegment", nameForDiagnostics: "string segment"),
-  .other(name: "unknown", nameForDiagnostics: "token"),
-  .other(name: "wildcard", nameForDiagnostics: "wildcard", text: "_"),
-]
+public enum Token: CaseIterable {
+  // Please keep this list sorted alphabetically
 
-public let SYNTAX_TOKEN_MAP = Dictionary(
-  uniqueKeysWithValues: SYNTAX_TOKENS.map { ("\($0.varOrCaseName.description.withFirstCharacterUppercased)Token", $0) }
-)
```
ahoppen, line 123: 🙌🏽   Nice to get rid of this one.

CodeGeneration/Sources/Utils/SyntaxBuildableChild.swift
```diff
@@ -71,18 +71,16 @@ public extension Child {
...
-    var textChoice: String
-    switch choices[0] {
-    case .keyword(let text), .token(let text):
-      textChoice = text
-    }
-    if textChoice == "init" {
-      textChoice = "`init`"
+    if case .token(let choices, _, _) = kind,
+      choices.count == 1,
+      case .keyword(var text) = choices[0]
```
ahoppen, line 76: Since you are touching this, you should be able to clean it up by slightly `.only.

```suggestion
      case .keyword(var text) = choices.only
```


--- 

Also: Interesting, don’t we need the `case .token(let text)` at all?

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftparser/ParserTokenSpecSetFile.swift
```diff
@@ -93,15 +92,15 @@ let parserTokenSpecSetFile = SourceFileSyntax(leadingTrivia: copyrightHeader) {
...
                     )
-                  case .token(let tokenText):
-                    let token = SYNTAX_TOKEN_MAP[tokenText]!
-                    if token.text != nil {
+                  case .token(let tokenKind):
+                    let caseName = tokenKind.tokenSpec.varOrCaseName
+                    if tokenKind.tokenSpec.text != nil {
                       SwitchCaseSyntax(
-                        "case .\(raw: token.varOrCaseName): return .\(raw: token.varOrCaseName)Token()"
+                        "case .\(raw: caseName): return .\(raw: caseName)Token()"
```
ahoppen, line 99: Since you’re touching this, I think you can remove the `raw:` here. Looks like I forgot it when I added `varOrCaseName`.

Same 4 lines below

CodeGeneration/Tests/ValidateSyntaxNodes/ValidateSyntaxNodes.swift
```diff
@@ -178,16 +178,16 @@ class ValidateSyntaxNodes: XCTestCase {
...
       break
-    case .token(tokenKind: "CommaToken"):
+    case .token(tokenKind: .comma):
       if child.name != "TrailingComma" && child.name != "Comma" {
         return "child '\(child.name)' has a comma keyword as its only token choice and should thus be named 'Comma' or 'TrailingComma'"
       }
     case .token(tokenKind: let tokenKind):
       let expectedChildName =
-        tokenKind
+        tokenKind.tokenSpec.varOrCaseName.text.withFirstCharacterUppercased
```
ahoppen, line 190: You can remove the `withFirstCharacterUppercased` here and just lowercase the child names in the expected failures of the `assertFailuresMatchXFails` calls. That cleans them up as well.

ahoppen (commented):
Wohooo. Looks very good, I love it. I’ve got a couple of minor comments but otherwise this looks ready to ship.

CodeGeneration/Sources/SyntaxSupport/AttributeNodes.swift
```diff
@@ -373,11 +373,11 @@ public let ATTRIBUTE_NODES: [Node] = [
       Child(
         name: "Argument",
         deprecatedName: "Parameter",
-        kind: .token(choices: [.token(tokenKind: "IdentifierToken"), .token(tokenKind: "IntegerLiteralToken"), .keyword(text: "self")])
+        kind: .token(choices: [.token(tokenKind: .identifier), .token(tokenKind: .integerLiteral), .keyword(text: "self")])
```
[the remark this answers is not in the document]
Matejkob: I totally agree with you. I'll add a new issue in this matter. 🚀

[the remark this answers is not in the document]
ahoppen: Thank you!

CodeGeneration/Sources/SyntaxSupport/Child.swift
```diff
@@ -16,7 +16,7 @@ import SwiftSyntax
 /// keyword with the given text.
 public enum TokenChoice: Equatable {
   case keyword(text: String)
-  case token(tokenKind: String)
+  case token(tokenKind: Token)
```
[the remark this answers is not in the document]
Matejkob: I like it! I'm also considering the following changes:
* Renaming all occurrences of `tokenKind` to `token`
```diff
switch choice {
case .keyword(let keywordText):
  let keyword = KEYWORDS.first(where: { $0.name == keywordText })!
  DeclSyntax("case \(raw: keyword.escapedName)")
-case .token(let tokenKind):
-DeclSyntax("case \(tokenKind.tokenSpec.varOrCaseName)")
+case .token(let token):
+DeclSyntax("case \(token.tokenSpec.varOrCaseName)")
}
```
* Isn't renaming `tokenSpec` to just `spec` or `specification` better? Repeating the `token` keyword seems redundant.

What do u think?

CodeGeneration/Sources/SyntaxSupport/Child.swift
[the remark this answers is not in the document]
Matejkob: Thanks. 

I've added the following comment to the code:
```swift
/*
 FIXME: Technically, returning `.unknown` is not correct here.
 The old string-based implementation returned "Token" to ensure that `tokenKind` is not nil
 and that `isToken` computed-property will return true, but the value "Token" had never been
 used in other cases. We should try to remove this computed property altogether in the issue:
 https://github.com/apple/swift-syntax/issues/2010
 */
```
I made some modifications to your original proposal to provide additional context regarding the implementation - I hope this would be useful.

I also created issue #2010 to address this. I had some difficulties in crafting the title and description, so feel free to make any necessary corrections or adjustments to clarify the issue."

CodeGeneration/Sources/SyntaxSupport/TokenSpec.swift
Matejkob: 🙌

CodeGeneration/Sources/Utils/SyntaxBuildableChild.swift
Matejkob: Ooo, that's cool, thanks! - consider it done ;)

> Also: Interesting, don’t we need the `case .token(let text)` at all?

Actually, I have to confess that I don't know if we need `case .token(let text)` at all because it's quite hard to understand the requirements only from this code. What I'm pretty sure is that in the previews version of this code:
```swift
 guard case .token(let choices, _, _) = kind, choices.count == 1, token.kind == .keyword else {
   return nil
 }
 var textChoice: String
 switch choices[0] {
 case .keyword(let text), .token(let text):
   textChoice = text
 }
 if textChoice == "init" {
   textChoice = "`init`"
 }
 return ExprSyntax(".\(token.varOrCaseName)(.\(raw: textChoice))")
```
the branch in which we would achieve `case .token(let text)` didn't exist. Namely in the `guard` above we had a condition `token.kind == .keyword` which it prevents

CodeGeneration/Tests/ValidateSyntaxNodes/ValidateSyntaxNodes.swift
Matejkob: I'm not entirely sure if I completely understand you, but I believe that your suggestion may not work. Consider an example like `binaryOperator`: we drop the `binary` prefix and expect `Operator`. Until we can lowercase the first letters of children's names, the current solution appears to be the most reliable

CodeGeneration/Sources/SyntaxSupport/Child.swift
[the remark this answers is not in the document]
ahoppen: Both seem like good ideas to me 👍🏽

CodeGeneration/Sources/Utils/SyntaxBuildableChild.swift
ahoppen: Oh, you’re absolutely right. Makes sense that we don’t need `case .token`. Your version is definitely a cleanup – thanks!

CodeGeneration/Tests/ValidateSyntaxNodes/ValidateSyntaxNodes.swift
ahoppen: Oh, sorry, you’re right. Also, I thought that we had already lowercased the child names, which isn’t true. So: Nothing to be done here.

Matejkob: Speaking about lowercasing the child names check out DM's on swift forums... XD That's a pretty funny situation... XD

CodeGeneration/Sources/SyntaxSupport/Child.swift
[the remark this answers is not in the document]
Matejkob: Done

ahoppen (approved):
Looks good. Thank yo

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

PR #2011, opened 2023-08-06
Introduce `tokenSpecSetType` and `syntaxChoicesType` for child nodes in CodeGeneration

Matejkob:
This PR ties back to our discussion in #1908. Specifically, [@ahoppen proposed](https://github.com/apple/swift-syntax/pull/1908#issuecomment-1638266939) this approach:

>I just found out about the usage of capitalized name for the unexpected nodes yesterday as well. The way I would go about this that:
>
> * The child names as specified in SyntaxSupport should be lowercased because the name of the property is how we primarily think about the child
> * `Child.name` should be private (and will probably store the lowercase name from the `Child` initializer)
> * There’s `varName` (which I’m renaming to `varOrCaseName` in https://github.com/apple/swift-syntax/pull/1914 to be consistent with `Node.varOrCaseName`), which should return the lowercase name
> * I’d introduce `typeName`, which returns a capitalized version of the child name (same as `Child.name` currently does). We should be able to use that in all the cases where we currently rely on `Child.name` being uppercase.

Here are the tweaks I made:
1. Added a `type` property to `Child`. This returns a `TypeSyntax` built from the capitalized child name. @ahoppen initially suggested using `typeName`, but I felt that `type` of `TypeSyntax` type fit better. I'm open to feedback though – should we stick to `typeName` of type `TokenSyntax`?
2. I replaced references from `name` to `type` in some scenarios:
   - When creating enum names, like `enum accessorSpecifierOptions: TokenSpecSet`, and then when used as types, e.g., `ElseBody` for `IfExprSyntax`.
   - When constructing "unexpected" node names.
3. Added `deprecatedType` and `hasDeprecatedName` properties. This aligns with other parts of `Child`, and it means we can make `deprecatedName` private, improving encapsulation.

### DIFF
--- a/CodeGeneration/Sources/SyntaxSupport/Child.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/Child.swift
@@ -46,6 +46,14 @@ public enum ChildKind {
     }
   }
 
+  public var isToken: Bool {
+    if case .token = self {
+      return true
+    } else {
+      return false
+    }
+  }
+
   public var isNodeChoicesEmpty: Bool {
     if case .nodeChoices(let nodeChoices) = self {
       return nodeChoices.isEmpty
@@ -103,12 +111,33 @@ public class Child {
     return .identifier(lowercaseFirstWord(name: name))
   }
 
+  /// If this child has node choices, the type that the nested `SyntaxChildChoices` type should get.
+  ///
+  /// For any other kind of child nodes, accessing this property crashes.
+  public var syntaxChoicesType: TypeSyntax {
+    precondition(kind.isNodeChoices, "Cannot get `syntaxChoicesType` for node that doesn’t have nodeChoices")
+    return "\(raw: name.withFirstCharacterUppercased)"
+  }
+
+  /// If this child only has tokens, the type that the generated `TokenSpecSet` should get.
+  ///
+  /// For any other kind of child nodes, accessing this property crashes.
+  public var tokenSpecSetType: TypeSyntax {
+    precondition(kind.isToken, "Cannot get `tokenSpecSetType` for node that isn’t a token")
+    return "\(raw: name.withFirstCharacterUppercased)Options"
+  }
+
   /// The deprecated name of this child that's suitable to be used for variable or enum case names.
   public var deprecatedVarName: TokenSyntax? {
     guard let deprecatedName = deprecatedName else {
       return nil
     }
     return .identifier(lowercaseFirstWord(name: deprecatedName))
+  }
+
+  /// Determines if this child has a deprecated name
+  public var hasDeprecatedName: Bool {
+    return deprecatedName != nil
   }
 
   /// If the child ends with "token" in the kind, it's considered a token node.
--- a/CodeGeneration/Sources/SyntaxSupport/Node.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/Node.swift
@@ -114,18 +114,20 @@ public class Node {
       // any two defined children
       childrenWithUnexpected =
         children.enumerated().flatMap { (i, child) -> [Child] in
+          let childName = child.name.withFirstCharacterUppercased
+
           let unexpectedName: String
           let unexpectedDeprecatedName: String?
 
           if i == 0 {
-            unexpectedName = "UnexpectedBefore\(child.name)"
-            unexpectedDeprecatedName = child.deprecatedName.map { "UnexpectedBefore\($0)" }
+            unexpectedName = "UnexpectedBefore\(childName)"
+            unexpectedDeprecatedName = child.deprecatedName.map { "UnexpectedBefore\($0.withFirstCharacterUppercased)" }
           } else {
-            unexpectedName = "UnexpectedBetween\(children[i - 1].name)And\(child.name)"
-            if let deprecatedName = children[i - 1].deprecatedName {
-              unexpectedDeprecatedName = "UnexpectedBetween\(deprecatedName)And\(child.deprecatedName ?? child.name)"
-            } else if let deprecatedName = child.deprecatedName {
-              unexpectedDeprecatedName = "UnexpectedBetween\(children[i - 1].name)And\(deprecatedName)"
+            unexpectedName = "UnexpectedBetween\(children[i - 1].name.withFirstCharacterUppercased)And\(childName)"
+            if let deprecatedName = children[i - 1].deprecatedName?.withFirstCharacterUppercased {
+              unexpectedDeprecatedName = "UnexpectedBetween\(deprecatedName)And\(child.deprecatedName?.withFirstCharacterUppercased ?? childName)"
+            } else if let deprecatedName = child.deprecatedName?.withFirstCharacterUppercased {
+              unexpectedDeprecatedName = "UnexpectedBetween\(children[i - 1].name.withFirstCharacterUppercased)And\(deprecatedName)"
             } else {
               unexpectedDeprecatedName = nil
             }
@@ -139,9 +141,9 @@ public class Node {
           return [unexpectedBefore, child]
         } + [
           Child(
-            name: "UnexpectedAfter\(children.last!.name)",
-            deprecatedName: children.last!.deprecatedName.map { "UnexpectedAfter\($0)" },
-            kind: .collection(kind: .unexpectedNodes, collectionElementName: "UnexpectedAfter\(children.last!.name)"),
+            name: "UnexpectedAfter\(children.last!.name.withFirstCharacterUppercased)",
+            deprecatedName: children.last!.deprecatedName.map { "UnexpectedAfter\($0.withFirstCharacterUppercased)" },
+            kind: .collection(kind: .unexpectedNodes, collectionElementName: "UnexpectedAfter\(children.last!.name.withFirstCharacterUppercased)"),
             isOptional: true
           )
         ]
--- a/CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/RenamedChildrenCompatibilityFile.swift
+++ b/CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/RenamedChildrenCompatibilityFile.swift
@@ -20,8 +20,8 @@ let renamedChildrenCompatibilityFile = try! SourceFileSyntax(leadingTrivia: copy
     try ExtensionDeclSyntax("extension \(raw: layoutNode.type.syntaxBaseName)") {
       for child in layoutNode.children {
         if let deprecatedVarName = child.deprecatedVarName {
-          let childType: TypeSyntax = child.kind.isNodeChoicesEmpty ? child.syntaxNodeKind.syntaxType : "\(raw: child.name)"
-          let type = child.isOptional ? TypeSyntax("\(raw: childType)?") : TypeSyntax("\(raw: childType)")
+          let childType: TypeSyntax = child.kind.isNodeChoicesEmpty ? child.syntaxNodeKind.syntaxType : child.syntaxChoicesType
+          let type = child.isOptional ? TypeSyntax("\(childType)?") : childType
 
           DeclSyntax(
             """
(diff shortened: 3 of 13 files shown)


CodeGeneration/Sources/SyntaxSupport/Child.swift
```diff
@@ -58,16 +58,14 @@ public enum ChildKind {
 /// A child of a node, that may be declared optional or a token with a
 /// restricted subset of acceptable kinds or texts.
 public class Child {
+  /// If the child has been renamed, its old, now deprecated, name.
+  private let deprecatedName: String?
+
   /// The name of the child.
   ///
   /// The first character of the name is always uppercase.
   public let name: String
```
Matejkob, line 67: I've been thinking about privatizing the `name` property as proposed in the original discussion but we are using it in the role of `id` to identify children in tests `ValidateSyntaxNodes.swift` for example: https://github.com/apple/swift-syntax/blob/e340af46cd6725e1567c42adcbb1ec979bf7f7b4/CodeGeneration/Tests/ValidateSyntaxNodes/ValidateSyntaxNodes.swift#L83-L85

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/RenamedChildrenCompatibilityFile.swift
```diff
@@ -20,8 +20,8 @@ let renamedChildrenCompatibilityFile = try! SourceFileSyntax(leadingTrivia: copy
     try ExtensionDeclSyntax("extension \(raw: layoutNode.type.syntaxBaseName)") {
       for child in layoutNode.children {
         if let deprecatedVarName = child.deprecatedVarName {
-          let childType: TypeSyntax = child.kind.isNodeChoicesEmpty ? child.syntaxNodeKind.syntaxType : "\(raw: child.name)"
-          let type = child.isOptional ? TypeSyntax("\(raw: childType)?") : TypeSyntax("\(raw: childType)")
+          let childType: TypeSyntax = child.kind.isNodeChoicesEmpty ? child.syntaxNodeKind.syntaxType : "\(child.type)"
+          let type = child.isOptional ? TypeSyntax("\(childType)?") : TypeSyntax("\(childType)")
```
ahoppen, line 24: Can be shortened a little bit more to

```suggestion
          let type = child.isOptional ? TypeSyntax("\(childType)?") : childType
```

CodeGeneration/Sources/SyntaxSupport/Node.swift
```diff
@@ -118,14 +118,14 @@ public class Node {
           let unexpectedDeprecatedName: String?
 
           if i == 0 {
-            unexpectedName = "UnexpectedBefore\(child.name)"
-            unexpectedDeprecatedName = child.deprecatedName.map { "UnexpectedBefore\($0)" }
+            unexpectedName = "UnexpectedBefore\(child.type)"
```
ahoppen, line 121: I would prefer to do `child.name.withFirstLetterUppercased` when building the unexpected name. I know that the result is the same as if we used the type name, but that’s more of a coincidence right now. What we’re intending to do here is to stitch the names together and the code should reflect that in my opinion. Otherwise every time I read this, I would wonder: What has the type name to do with the property name here.

CodeGeneration/Sources/SyntaxSupport/Child.swift
```diff
@@ -58,16 +58,14 @@ public enum ChildKind {
 /// A child of a node, that may be declared optional or a token with a
 /// restricted subset of acceptable kinds or texts.
 public class Child {
+  /// If the child has been renamed, its old, now deprecated, name.
+  private let deprecatedName: String?
+
```
ahoppen, line 63: Any reason why you changed the order of `deprecatedName` and `name`?

CodeGeneration/Sources/SyntaxSupport/Child.swift
ahoppen: You should be able to do `type.description == other.type.description` instead of `name == other.name` here.

CodeGeneration/Sources/SyntaxSupport/Child.swift
Matejkob: Oh yeah :/ - habit from all my projects where we are trying to separate `public` properties from `private` ones. Should I revert the changes?

CodeGeneration/Sources/SyntaxSupport/Node.swift
Matejkob: I fully agree with you. I guess I was trying too hard to change the code base in order to make `child.name` private.

CodeGeneration/Sources/SyntaxSupport/Child.swift
Matejkob: Done

CodeGeneration/Sources/generate-swiftsyntax/templates/swiftsyntax/RenamedChildrenCompatibilityFile.swift
Matejkob: Nice catch! Thanks. Done.

CodeGeneration/Sources/SyntaxSupport/Node.swift
Matejkob: Done

CodeGeneration/Sources/SyntaxSupport/Child.swift
Matejkob: Done

CodeGeneration/Sources/SyntaxSupport/Node.swift
```diff
@@ -114,18 +114,20 @@ public class Node {
...
+          let childName = child.name.withFirstCharacterUppercased
+
           let unexpectedName: String
           let unexpectedDeprecatedName: String?
 
           if i == 0 {
-            unexpectedName = "UnexpectedBefore\(child.name)"
-            unexpectedDeprecatedName = child.deprecatedName.map { "UnexpectedBefore\($0)" }
+            unexpectedName = "UnexpectedBefore\(childName)"
+            unexpectedDeprecatedName = child.deprecatedType.map { "UnexpectedBefore\($0)" }
```
ahoppen, line 124: Just like we do `child.name.withFirstCharacterUppercased`, we should also do `child.deprecatedName.withFirstCharacterUppercased`.

And I think below are couple usages of `deprecatedName` that need to be uppercased once we lowercase `deprecatedName`.

CodeGeneration/Sources/SyntaxSupport/Child.swift
```diff
@@ -103,6 +103,11 @@ public class Child {
     return .identifier(lowercaseFirstWord(name: name))
   }
 
+  /// A type of this child.
+  public var type: TypeSyntax {
+    return "\(raw: name.withFirstCharacterUppercased)"
+  }
```
ahoppen, line 109: Sorry, after reading this PR again, I noticed that this isn’t actually the type of the child, unless it has multiple node choices. What this technically is, is the name that a nested `SyntaxChoices` type gets, if it gets created. With this in mind, a better name would be `syntaxChoicesType` and it should fail if the child doesn’t have `kind == .nodeChoices` to prevent misuse.

```suggestion
  /// If this child has node choices, the type that the nested `SyntaxChildChoices` type should get.
  /// 
  /// For any other kind of child nodes, accessing this property crashes.
  public var syntaxChoicesType: TypeSyntax {
    precondition(kind.isNodeChoices, "Cannot get syntaxChoicesType for node that doesn’t have nodeChoices")
    return "\(raw: name.withFirstCharacterUppercased)"
  }
```

And I think most of the uses of `type` in `ValidateSyntaxNodes` should be `varOrCaseName` or, where necessary `varOrCaseName.description`.

CodeGeneration/Sources/SyntaxSupport/Node.swift
Matejkob: At least one `Child` has a deprecated name that begins with a lowercase letter, which is inconsistent with the majority that starts with uppercase. This has led to the generation of an unexpected node name like `UnexpectedBeforedeprecatedName`. If we want to avoid changing the API, it seems we must adhere to this implementation.

> And I think below are couple usages of `deprecatedName` that need to be uppercased once we lowercase deprecatedName.

Actually, I've not considered changing `deprecatedName` before. Do you think that we should do that?
I guess it's necessary if we are changing `name`. But what would be the best course of action regarding the aforementioned edge case?

ahoppen: > At least one `Child` has a deprecated name that begins with a lowercase

Oh, that’s a bug. The deprecated name of `FunctionType.paramters` should be uppercased. Fixing in https://github.com/apple/swift-syntax/pull/2023.

> Actually, I've not considered changing `deprecatedName` before. 

Yes, absolutely, we should also lowercase `deprecatedName`.

CodeGeneration/Sources/SyntaxSupport/Child.swift
Matejkob: It seems very reasonable. I just wondering what about places like this one: https://github.com/apple/swift-syntax/blob/83348e7eb826b978896421b2a233a4290596eb58/CodeGeneration/Sources/generate-swiftsyntax/templates/swiftparser/ParserTokenSpecSetFile.swift#L23C4-L29 ? Do we wanna use `name.withFirstCharacterUppercased` instead of `syntaxChoicesType`?

Matejkob: @ahoppen  I'm sorry, but I got lost. Could you please summarize once more what we want to achieve?

Matejkob: Did you mean introducing two new properties on `Child`: one for `kind.isNodeChoices` and the other for `kind.isToken`?

ahoppen: Yes, exactly.

Matejkob: Done

CodeGeneration/Sources/SyntaxSupport/Child.swift
```diff
@@ -103,6 +111,22 @@ public class Child {
...
+  /// For any other kind of child nodes, accessing this property crashes.
+  public var syntaxChoicesType: TypeSyntax {
+    precondition(kind.isNodeChoices, "Cannot get `syntaxChoicesType` for node that doesn’t have nodeChoices")
+    return "\(raw: name.withFirstCharacterUppercased)"
+  }
+
+  /// If this child only has tokens, the type that the generated `TokenSpecSet` should get.
+  ///
+  /// For any other kind of child nodes, accessing this property crashes.
+  public var syntaxTokenType: TypeSyntax {
```
ahoppen, line 125: I’m not sure if I suggested `syntaxTokenType` but in light of `syntaxChoicesType`, I think `tokenSpecSetType` would be a better name.

Also remember to update the `precondition` below.

ahoppen (commented):
Wow, this is a lot nicer. I have one naming suggestion (and I was probably the one who suggested the old name that I now dislike, so sorry for that), otherwise this looks great.

CodeGeneration/Sources/SyntaxSupport/Child.swift
Matejkob: Thanks. Done

ahoppen (approved):
Thanks for doing this @Matejkob!

ahoppen:
@swift-ci Please test

ahoppen:
@Matejkob Looks like there are some casing issues in the CodeGeneration package. You should be able to reproduce them locally by opening the CodeGeneration package and running the tests in there

Matejkob:
> @Matejkob Looks like there are some casing issues in the CodeGeneration package. You should be able to reproduce them locally by opening the CodeGeneration package and running the tests in there

Oh, something went wrong during the merge. Now everything should work fine.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

PR #2012, opened 2023-08-06
Add correct diagnostic for missing parentheses

### LINKED ISSUE #1925: Missing parentheses in closure signature produces bad diagnostic
### Issue Kind

Bad Diagnostic Produced

### Source Code

```swift
_ = { a: Int, b: Int in
}
```

### Description

The diagnostic produced here is:
> error: unexpected code ': Int, b: Int in' in closure

But it would be nice to instead have an error about missing parentheses (and a fix-it to add them). Apparently we're also not recovering to the `in` here properly, which isn't great.

rdar://112407661

kimdv:
Resolves #1925

### DIFF
--- a/Sources/SwiftParser/Expressions.swift
+++ b/Sources/SwiftParser/Expressions.swift
@@ -1746,7 +1746,9 @@ extension Parser {
     var effectSpecifiers: RawTypeEffectSpecifiersSyntax?
     var returnClause: RawReturnClauseSyntax? = nil
     if !self.at(.keyword(.in)) {
-      if self.at(.leftParen) {
+      // If the next token is ':', then it looks like the code contained a non-shorthand closure parameter with a type annotation.
+      // These need to be wrapped in parentheses.
+      if self.at(.leftParen) || self.peek(isAt: .colon) {
         // Parse the closure arguments.
         let params = self.parseParameterClause(RawClosureParameterClauseSyntax.self) { parser in
           parser.parseClosureParameter()
@@ -2430,34 +2432,28 @@ extension Parser.Lookahead {
     }
 
     // Parse pattern-tuple func-signature-result? 'in'.
-    if lookahead.consume(if: .leftParen) != nil {  // Consume the ')'.
-
+    if lookahead.at(.leftParen) {  // Consume the '('.
       // While we don't have '->' or ')', eat balanced tokens.
       var skipProgress = LoopProgressCondition()
-      while !lookahead.at(.endOfFile, .rightParen) && lookahead.hasProgressed(&skipProgress) {
+      while !lookahead.at(.endOfFile, .rightBrace, .keyword(.in)) && !lookahead.at(.arrow) && lookahead.hasProgressed(&skipProgress) {
         lookahead.skipSingle()
       }
-
-      // Consume the ')', if it's there.
-      if lookahead.consume(if: .rightParen) != nil {
-        lookahead.consumeEffectsSpecifiers()
-
-        // Parse the func-signature-result, if present.
-        if lookahead.consume(if: .arrow) != nil {
-          guard lookahead.canParseType() else {
-            return false
-          }
-
-          lookahead.consumeEffectsSpecifiers()
-        }
-      }
-      // Okay, we have a closure signature.
     } else if lookahead.at(.identifier) || lookahead.at(.wildcard) {
       // Parse identifier (',' identifier)*
       lookahead.consumeAnyToken()
 
+      /// If the next token is a colon, interpret is as a type annotation and consume a type after it.
+      /// While type annotations aren’t allowed in shorthand closure parameters, we consume them to improve recovery.
+      func consumeOptionalTypeAnnotation() -> Bool {
+        if lookahead.consume(if: .colon) != nil {
+          return lookahead.canParseType()
+        } else {
+          return true
+        }
+      }
+
       var parametersProgress = LoopProgressCondition()
-      while lookahead.consume(if: .comma) != nil && lookahead.hasProgressed(&parametersProgress) {
+      while consumeOptionalTypeAnnotation() && lookahead.consume(if: .comma) != nil && lookahead.hasProgressed(&parametersProgress) {
         if lookahead.at(.identifier) || lookahead.at(.wildcard) {
           lookahead.consumeAnyToken()
           continue
@@ -2465,17 +2461,20 @@ extension Parser.Lookahead {
 
         return false
       }
+    }
+
+    // Consume the ')', if it's there.
+    lookahead.consume(if: .rightParen)
+
+    lookahead.consumeEffectsSpecifiers()
+
+    // Parse the func-signature-result, if present.
+    if lookahead.consume(if: .arrow) != nil {
+      guard lookahead.canParseType() else {
+        return false
+      }
 
       lookahead.consumeEffectsSpecifiers()
-
-      // Parse the func-signature-result, if present.
-      if lookahead.consume(if: .arrow) != nil {
-        guard lookahead.canParseType() else {
-          return false
-        }
-
-        lookahead.consumeEffectsSpecifiers()
-      }
     }
 
     // Parse the 'in' at the end.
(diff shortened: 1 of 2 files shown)


Sources/SwiftParser/Expressions.swift
```diff
@@ -2427,52 +2429,51 @@ extension Parser.Lookahead {
...
 
+      func consumeTypesIfAvailable() -> Bool {
+        // If identifiers have types, we need to have parens around.
+        // If they are missing but colon an type is there, it might be an error.
+        if lookahead.consume(if: .colon) != nil {
+          return lookahead.canParseType() && lookahead.consume(if: .comma) != nil
+        } else {
+          return lookahead.consume(if: .comma) != nil
+        }
+      }
```
ahoppen, line 2450: I wonder if all of this would be easier to understand if we had

```suggestion
      /// If the next token is a colon, interpret is as a type annotation and consume a type after it.
      /// While type annotations aren’t allowed in shorthand closure parameters, we consume them to improve recovery.
      func consumeOptionalTypeAnnotation() -> Bool {
        if lookahead.consume(if: .colon) != nil {
          return lookahead.canParseType()
        } else {
          return true
        }
      }
```

and then below `while consumeOptionalTypeAnnotation() && lookahead.consume(if: .comma) != nil `

Sources/SwiftParser/Expressions.swift
```diff
@@ -2427,52 +2429,51 @@ extension Parser.Lookahead {
...
 
         return false
       }
+    }
 
-      lookahead.consumeEffectsSpecifiers()
+    // Consume the ')', if it's there.
+    if lookahead.at(.rightParen) {
+      lookahead.skipSingle()
+    }
```
ahoppen, line 2466: Shouldn’t this be

```suggestion
    // Consume the ')', if it's there.
    lookahead.consume(if: .rightParen)
```

Tests/SwiftParserTest/ExpressionTests.swift
```diff
@@ -2689,4 +2689,49 @@ final class StatementExpressionTests: XCTestCase {
...
+          message: "expected '(' to start parameter clause",
+          fixIts: ["insert '('"]
+        )
+      ],
+      fixedSource: """
+        _ = { (a: Int, b: Int) in
+        }
+        """
+    )
+  }
```
ahoppen, line 2736: Could you also add a test case with a return type?

```swift
_ = { a: Int, b: Int -> Int in
}
```

Tests/SwiftParserTest/ExpressionTests.swift
```diff
@@ -2689,4 +2689,74 @@ final class StatementExpressionTests: XCTestCase {
...
+        ),
+        DiagnosticSpec(
+          locationMarker: "2️⃣",
+          message: "expected ')' to end parameter clause",
+          fixIts: ["insert ')'"]
+        ),
+      ],
+      fixedSource: """
+        _ = { (a: Int, b: Int) in
+        }
```
ahoppen, line 2738: I just checked and it looks like this test is failing. The fixed source should be 

```swift
        _ = { (a: Int, b: Int) -> Int in
        }
```

kimdv: Was looking into it.
Haven't found a fix yet 😄

kimdv: While debugging it hit me that `_ = { (a: Int, b: (Int)  -> Int) in }` is a valid actually valid swift code. 

So what do you think?
Should we let it be as is?

Otherwise we should stop parsing the type until we hit the arrow?

ahoppen: Oh, I forgot that. I don’t think there’s an easy way to tell closure parameter parsing to stop at an arrow to diagnose this better. So, it’s fine for now, I would say.

kimdv:
@swift-ci please test

Tests/SwiftParserTest/ExpressionTests.swift
kimdv: Added the test above to ensure that is correct.

ahoppen (approved):

Sources/SwiftParser/Expressions.swift
```diff
@@ -1746,7 +1746,9 @@ extension Parser {
     var effectSpecifiers: RawTypeEffectSpecifiersSyntax?
     var returnClause: RawReturnClauseSyntax? = nil
     if !self.at(.keyword(.in)) {
-      if self.at(.leftParen) {
+      // If the next token is ':', then it looks like the code contained a non-shorthand closure parameter with a type annotation.
+      // These need to be wrapped in parentheses.
+      if self.at(.leftParen) || self.peek(isAt: .colon) {
```
kimdv, line 1751: I added this test and it fails because we are peeking here. 
When at this place current token is `comparisonPredicate`.

```swift
    assertParse(
      """
      withInvalidOrderings { (comparisonPredicate: @escaping (Int, Int) -> Bool) in
      }
      """
    )
```

So the peek seems not the correct solutions here. 
Maybe we should do something like `self.peek(isAt: .colon) && previousToken != .leftParen`

What do you think @ahoppen

kimdv: @ahoppen 😬

ahoppen: Sorry, it’s been a busy week and I missed the comment. Thanks for the ping.

The issue happens because in `canPareClosureSignature`, we now stop at the arrow. Since we start doing `skipSingle` at `comparisonPredicate`. If you make the following change, then we’re calling `skipSingle` on the opening `(`, which consumes all the contents until the matching `)` before the `in`, including the `->`, which fixes the issue.

```diff
- if lookahead.consume(if: .leftParen) != nil {  // Consume the ')'.
+ if lookahead.at(.leftParen) {  // Consume the '('.
```

kimdv: Oh okay nice! 
Thanks a lot. 

And no worries! Take the time to reply when you can 😁

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test macOS

kimdv:
@ahoppen the job fails here: https://ci.swift.org/job/swift-syntax-PR-macOS/3177/console

Not sure why. 
When I add the tests in the test suite they succeeds correctly.

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test

ahoppen:
It’s failing at the following test case. The `in` in the `for` statement is critical to reproduce the failure. Could you look into why it’s failing? If you can’t find it, I can try debugging as well.

```swift
state.withCriticalRegion { (1 + 2) }
for action in tracking {
  action()
}
```

kimdv:
@ahoppen found it. 
We had inside `canParseClosureSignature` this `!lookahead.at(.endOfFile, .rightParen, .keyword(.in))` and it didn't stop before hitting the `in` keyword.

Changed `rightParen` -> `rightBrace`

kimdv:
@swift-ci please test

ahoppen:
@swift-ci Please test Windows

Sources/SwiftParser/Expressions.swift
```diff
@@ -2430,52 +2432,49 @@ extension Parser.Lookahead {
     }
 
     // Parse pattern-tuple func-signature-result? 'in'.
-    if lookahead.consume(if: .leftParen) != nil {  // Consume the ')'.
-
+    if lookahead.at(.leftParen) {  // Consume the '('.
       // While we don't have '->' or ')', eat balanced tokens.
       var skipProgress = LoopProgressCondition()
-      while !lookahead.at(.endOfFile, .rightParen) && lookahead.hasProgressed(&skipProgress) {
+      while !lookahead.at(.endOfFile, .rightBrace, .keyword(.in)) && !lookahead.at(.arrow) && lookahead.hasProgressed(&skipProgress) {
```
ahoppen, line 2438: Oh, yes, that does make sense. `skipSingle` is looking for the `rightParen` now that we previously consumed manually in `lookahead.consume(if: .leftParen)`.

PR #2013, opened 2023-08-07
Compare `Token.tokenKind` instead of `Token.text` in a couple of places in SwiftOperators

ahoppen:
Comparing `tokenKind` is the cleaner (and possibly also slightly more performant way) of checking the content of a token.


ahoppen:
@swift-ci Please test

kimdv (approved):

bnbarham (approved):

PR #2014, opened 2023-08-07
Add a function to get the `SyntaxChildrenIndex` of the n-th element in a `SyntaxCollection`

ahoppen:
We have had a similar function that was only available in `_SwiftSyntaxTestSupport`. I think it would be useful in general.


ahoppen:
@swift-ci Please test

Sources/SwiftSyntax/SyntaxCollection.swift
```diff
@@ -100,6 +100,28 @@ extension SyntaxCollection {
...
+  ///
+  /// - Complexity: O(`offset`) because all previous element need to be iterated
+  ///   to find the byte offset of the `offset`-th node within the source file.
+  ///
+  /// - Note: Because getting the `n`-th element in a ``SyntaxCollection`` is
+  ///   not O(1), ``SyntaxCollection`` doesn’t provide a subscript to retrieve
+  ///   the `n`-th element. Such a subscript would mask the complexity of
+  ///   getting the `n`-th element.
+  public func index(at offset: Int) -> SyntaxChildrenIndex {
+    if offset > self.count / 2 {
```
Matejkob, line 116: Hmmm, that's a really interesting approach!

I just wonder, `SyntaxCollection` doesn't conform to the `RandomAccessCollection`, so wouldn't the usage of the `count` there add `O(n)` complexity to this method?

ahoppen: `count` is O(1) because we overwrite it

https://github.com/apple/swift-syntax/blob/64e7475cc00313a36688b38fc7133319d90defd1/Sources/SwiftSyntax/SyntaxCollection.swift#L62-L64

Matejkob: Thanks for clearing that up. Much appreciated!

bnbarham (approved):

PR #2020, opened 2023-08-07
Allow trivia to be added to `SyntaxStringInterpolation` without adding a `raw:` label

ahoppen:
`Trivia` is a first-class citizen in SwiftSyntax and you shouldn’t need to add `raw` in front of it to add it to `SyntaxStringInterpolation`.


ahoppen:
@swift-ci Please test

kimdv (approved):

bnbarham (approved):

PR #2021, opened 2023-08-07
Add fixed source to `assertMacroExpansion`

### LINKED ISSUE #2015: Add a `fixedSource` parameter to `assertMacroExpansion`
The `assertParse` function has a `fixedSource` parameter to check that the application of Fix-Its produces valid results and it has been a valuable tool so far. `assertMacroExpansion` could gain the same parameter to check that the macro’s Fix-Its work as expected.

kimdv:
Resolves #2015

### DIFF
--- a/Release Notes/511.md
+++ b/Release Notes/511.md
@@ -2,6 +2,11 @@
 
 ## New APIs
 
+- `assertMacroExpansion` now have new parameters named `applyFixIts` and `fixedSource`
+  - Description: `applyFixIts` and `fixedSource` are used to assert so ensure that the source code after applying Fix-Its matches this string.
+  - Issue: https://github.com/apple/swift-syntax/issues/2015
+  - Pull Request: https://github.com/apple/swift-syntax/pull/2021
+
 ## API Behavior Changes
 
 ## Deprecations
@@ -21,9 +26,11 @@
 - Effect specifiers:
   - Description: The `unexpectedAfterThrowsSpecifier` node of the various effect specifiers has been removed.
   - Pull request: https://github.com/apple/swift-syntax/pull/2219
+
 - `SyntaxKind` removed conformance to `CaseIterable`
   - Description: `SyntaxKind` no longer conforms to `CaseIterable` since there is no good use case to iterate over all syntax kinds. 
   - Pull request: https://github.com/apple/swift-syntax/pull/2292
+
 - `IntegerLiteralExprSyntax.Radix` removed conformance to `CaseIterable`
   - Description: `IntegerLiteralExprSyntax.Radix` no longer conforms to `CaseIterable` since there is no good use case to iterate over all radix kinds. 
   - Pull request: https://github.com/apple/swift-syntax/pull/2292
--- a/Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
+++ b/Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
@@ -254,13 +254,18 @@ func assertDiagnostic(
 ///   - macros: The macros that should be expanded, provided as a dictionary
 ///     mapping macro names (e.g., `"stringify"`) to implementation types
 ///     (e.g., `StringifyMacro.self`).
+///   - applyFixIts: If specified, filters the Fix-Its that are applied to generate `fixedSource` to only those whose message occurs in this array. If `nil`, all Fix-Its from the diagnostics are applied.
+///   - fixedSource: If specified, asserts that the source code after applying Fix-Its matches this string.
 ///   - testModuleName: The name of the test module to use.
 ///   - testFileName: The name of the test file name to use.
+///   - indentationWidth: The indentation width used in the expansion.
 public func assertMacroExpansion(
   _ originalSource: String,
   expandedSource expectedExpandedSource: String,
   diagnostics: [DiagnosticSpec] = [],
   macros: [String: Macro.Type],
+  applyFixIts: [String]? = nil,
+  fixedSource expectedFixedSource: String? = nil,
   testModuleName: String = "TestModule",
   testFileName: String = "test.swift",
   indentationWidth: Trivia = .spaces(4),
@@ -316,5 +321,17 @@ public func assertMacroExpansion(
     for (actualDiag, expectedDiag) in zip(context.diagnostics, diagnostics) {
       assertDiagnostic(actualDiag, in: context, expected: expectedDiag)
     }
+  }
+
+  // Applying Fix-Its
+  if let expectedFixedSource = expectedFixedSource {
+    let fixedTree = FixItApplier.applyFixes(from: context.diagnostics, filterByMessages: applyFixIts, to: origSourceFile)
+    let fixedTreeDescription = fixedTree.description
+    assertStringsEqualWithDiff(
+      fixedTreeDescription.trimmingTrailingWhitespace(),
+      expectedFixedSource.trimmingTrailingWhitespace(),
+      file: file,
+      line: line
+    )
   }
 }
--- a/Tests/SwiftSyntaxMacroExpansionTest/PeerMacroTests.swift
+++ b/Tests/SwiftSyntaxMacroExpansionTest/PeerMacroTests.swift
@@ -28,89 +28,117 @@ import XCTest
 final class PeerMacroTests: XCTestCase {
   private let indentationWidth: Trivia = .spaces(2)
 
-  func testAddCompletionHandler() {
-    struct AddCompletionHandler: PeerMacro {
-      static func expansion(
-        of node: AttributeSyntax,
-        providingPeersOf declaration: some DeclSyntaxProtocol,
-        in context: some MacroExpansionContext
-      ) throws -> [DeclSyntax] {
-        // Only on functions at the moment. We could handle initializers as well
-        // with a bit of work.
-        guard let funcDecl = declaration.as(FunctionDeclSyntax.self) else {
-          throw MacroExpansionErrorMessage("@addCompletionHandler only works on functions")
-        }
-
-        // This only makes sense for async functions.
-        if funcDecl.signature.effectSpecifiers?.asyncSpecifier == nil {
-          throw MacroExpansionErrorMessage(
-            "@addCompletionHandler requires an async function"
-          )
-        }
-
-        // Form the completion handler parameter.
-        let resultType: TypeSyntax? = funcDecl.signature.returnClause?.type.trimmed
-
-        let completionHandlerParam =
-          FunctionParameterSyntax(
-            firstName: .identifier("completionHandler"),
-            colon: .colonToken(trailingTrivia: .space),
-            type: TypeSyntax("(\(resultType ?? "")) -> Void")
-          )
-
-        // Add the completion handler parameter to the parameter list.
-        let parameterList = funcDecl.signature.parameterClause.parameters
-        var newParameterList = parameterList
-        if !parameterList.isEmpty {
-          // We need to add a trailing comma to the preceding list.
-          newParameterList[newParameterList.index(before: newParameterList.endIndex)].trailingComma = .commaToken(trailingTrivia: .space)
-        }
-        newParameterList.append(completionHandlerParam)
-
-        let callArguments: [String] = parameterList.map { param in
-          let argName = param.secondName ?? param.firstName
-
-          if param.firstName.text != "_" {
-            return "\(param.firstName.text): \(argName.text)"
-          }
-
-          return "\(argName.text)"
-        }
-
-        let call: ExprSyntax =
-          "\(funcDecl.name)(\(raw: callArguments.joined(separator: ", ")))"
-
-        // FIXME: We should make CodeBlockSyntax ExpressibleByStringInterpolation,
-        // so that the full body could go here.
-        let newBody: ExprSyntax =
-          """
-
-            Task {
-              completionHandler(await \(call))
-            }
-
-          """
-
-        // Drop the @addCompletionHandler attribute from the new declaration.
-        let newAttributeList = funcDecl.attributes.filter {
-          guard case let .attribute(attribute) = $0 else {
-            return true
-          }
-          return attribute.attributeName.as(IdentifierTypeSyntax.self)?.name == "addCompletionHandler"
-        }
-
-        var newFunc = funcDecl
-        newFunc.signature.effectSpecifiers?.asyncSpecifier = nil  // drop async
-        newFunc.signature.returnClause = nil  // drop result type
-        newFunc.signature.parameterClause.parameters = newParameterList
-        newFunc.signature.parameterClause.trailingTrivia = []
-        newFunc.body = CodeBlockSyntax { newBody }
-        newFunc.attributes = newAttributeList
-
-        return [DeclSyntax(newFunc)]
+  fileprivate struct AddCompletionHandler: PeerMacro {
+    static func expansion(
+      of node: AttributeSyntax,
+      providingPeersOf declaration: some DeclSyntaxProtocol,
+      in context: some MacroExpansionContext
+    ) throws -> [DeclSyntax] {
+      // Only on functions at the moment. We could handle initializers as well
+      // with a bit of work.
+      guard let funcDecl = declaration.as(FunctionDeclSyntax.self) else {
+        throw MacroExpansionErrorMessage("@addCompletionHandler only works on functions")
       }
-    }
 
+      // This only makes sense for async functions.
+      if funcDecl.signature.effectSpecifiers?.asyncSpecifier == nil {
+        let newEffects: FunctionEffectSpecifiersSyntax
+        if let existingEffects = funcDecl.signature.effectSpecifiers {
+          newEffects = existingEffects.with(\.asyncSpecifier, .keyword(.async))
+        } else {
+          newEffects = FunctionEffectSpecifiersSyntax(asyncSpecifier: .keyword(.async))
+        }
+
+        let newSignature = funcDecl.signature.with(\.effectSpecifiers, newEffects)
+
+        let diag = Diagnostic(
+          node: Syntax(funcDecl.funcKeyword),
+          message: MacroExpansionErrorMessage(
+            "can only add a completion-handler variant to an 'async' function"
+          ),
+          fixIts: [
+            FixIt(
+              message: MacroExpansionFixItMessage(
+                "add 'async'"
+              ),
+              changes: [
+                FixIt.Change.replace(
+                  oldNode: Syntax(funcDecl.signature),
+                  newNode: Syntax(newSignature)
+                )
+              ]
+            )
+          ]
+        )
+
+        context.diagnose(diag)
+        return []
+      }
+
+      // Form the completion handler parameter.
+      let resultType: TypeSyntax? = funcDecl.signature.returnClause?.type.trimmed
+
+      let completionHandlerParam =
+        FunctionParameterSyntax(
+          firstName: .identifier("completionHandler"),
+          colon: .colonToken(trailingTrivia: .space),
+          type: TypeSyntax("(\(resultType ?? "")) -> Void")
+        )
+
+      // Add the completion handler parameter to the parameter list.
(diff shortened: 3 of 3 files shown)


Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
```diff
@@ -254,16 +254,24 @@ func assertDiagnostic(
...
   _ originalSource: String,
   expandedSource expectedExpandedSource: String,
   diagnostics: [DiagnosticSpec] = [],
   macros: [String: Macro.Type],
+  applyFixIts: [String]? = nil,
+  fixedSource expectedFixedSource: String? = nil,
   testModuleName: String = "TestModule",
   testFileName: String = "test.swift",
   indentationWidth: Trivia = .spaces(4),
+  options: AssertParseOptions = [],
```
ahoppen, line 274: I would prefer not to expose `AssertParseOptions` for now. It’s always easier to add public API in the future and I don’t think it’s newline normalization is needed to test macros in 99% of all cases. And since we don’t have substructure testing, `substructureCheckTrivia` is irrelevant here as well.

ahoppen:
CC @bnbarham I would appreciate your review on this as well.

Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
```diff
@@ -254,16 +254,24 @@ func assertDiagnostic(
 ///   - macros: The macros that should be expanded, provided as a dictionary
 ///     mapping macro names (e.g., `"stringify"`) to implementation types
 ///     (e.g., `StringifyMacro.self`).
+///   - applyFixIts: Applies only the fix-its with these messages.
+///     Ohterwise all Fix-Its from diagnostics will be applied.
```
bnbarham, line 258:
```suggestion
///     Otherwise all Fix-Its from diagnostics will be applied.
```

Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
```diff
@@ -254,16 +254,24 @@ func assertDiagnostic(
...
+///   - fixedSource: Asserts that the source after applying fix-its matches
```
bnbarham, line 259:
```suggestion
///   - fixedSource: The source code to expect after applying the Fix-Its matching `applyFixIts`.
```

bnbarham:
Thanks for doing this @kimdv 🙇 !

Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
kimdv: After added this comment, I tought that `Ohterwise all Fix-Its from diagnostics will be applied.` didn't make sense. 
So tried to make it a bit more explicit. 

Can you prof read again?

Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
```diff
@@ -254,13 +254,19 @@ func assertDiagnostic(
 ///   - macros: The macros that should be expanded, provided as a dictionary
 ///     mapping macro names (e.g., `"stringify"`) to implementation types
 ///     (e.g., `StringifyMacro.self`).
+///   - applyFixIts: The source code to expect after applying the Fix-Its matching `applyFixIts`.
+///     If none Fix-Its is provided all Fix-Its from the diagnostics will be applied.
```
ahoppen, line 258:
```suggestion
///   - applyFixIts: If specified, filters the Fix-Its that are applied to generate `fixedSource` to only those whose message occurs in this array. If `nil`, all Fix-Its from the diagnostics are applied.
```

Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
```diff
@@ -254,13 +254,19 @@ func assertDiagnostic(
...
+///   - fixedSource: Asserts that the source after applying fix-its matches
```
ahoppen, line 259:
```suggestion
///   - fixedSource: If specified, asserts that the source code after applying Fix-Its matches this string.
```

ahoppen (commented):
As always, I have some comments on documentation 😉

ahoppen (commented):
Oh, sorry for not mentioning this earlier. But it would be good to have a test case that exercises `fixedSource`. I don’t think any macro in `MacroSystemTests.swift` is currently a good candidate for Fix-Its, so it’s probably easiest if you write a new macro in there.

Tests/SwiftSyntaxMacroExpansionTest/MacroSystemTests.swift
```diff
@@ -284,9 +284,37 @@ public struct AddCompletionHandler: PeerMacro {
...
+          "can only add a completion-handler variant to an 'async' function"
+        ),
+        fixIts: [
+          FixIt(
+            message: MacroExpansionFixItMessage(
+              "add 'async'"
+            ),
+            changes: [
+              FixIt.Change.replace(
+                oldNode: Syntax(funcDecl.signature),
```
kimdv, line 308: I don't understand why. 
But when doing this the node.id changes. 

When debugging it first get 
```
▿ SyntaxIdentifier
  - rootId : 4320226936
  ▿ indexInTree : SyntaxIndexInTree
    - indexInTree : 8
```

And when printing inside the FixItApplier I get this
```
▿ SyntaxIdentifier
  - rootId : 105553162409640
  ▿ indexInTree : SyntaxIndexInTree
    - indexInTree : 7
```

ahoppen (commented):
One of the issues why you needed to add `transferTrivia: false` was that we did `nodes.map { .makeMissing(…) }`. This caused us to transfer the trivia for each token individually. Conceptually, what we want to do is to make a list of node missing and only transfer the leading trivia at the first node in the array and the trailing trivia of the last element of the array. To fix this, we should have a `makeMissing` function that takes an array of `Syntax`.

If we choose to not transfer trivia of one token, we should generally set it as the leading and trailing trivia of the newly inserted token. That way, we don’t drop it.

The reason why you needed to do `.with(\.trailingTrivia, [])` in `visit(AvailabilityConditionSyntax)` is that without it we truly have overlapping edits in `testAvailabilityQueryUnavailability34b`. `FixItApplier` checks if any of the edits intersect. If they do, we can only keep one of the edits. I would suggest that we choose to keep the larger edit. Currently we only have the `Ensure edits are not overlapping:` comment but don’t actually check that the edits are not overlapping.

I would like to split the change that updates `FixItApplier` from the change that adds `fixedSource` to `assertMacroExpansion`. That way we can merge the larger refactoring of rewriting `FixItApplier` independently of changing the public API. When you do change `assertMacroExpansion`, could you add it to the release notes?

ahoppen:
Oh, and here’s a suggested diff for the first two points in my review

<details><summary>Diff</summary>
<p>

```diff
diff --git a/Sources/SwiftParserDiagnostics/DiagnosticExtensions.swift b/Sources/SwiftParserDiagnostics/DiagnosticExtensions.swift
index 27fb25cc..2982e9bb 100644
--- a/Sources/SwiftParserDiagnostics/DiagnosticExtensions.swift
+++ b/Sources/SwiftParserDiagnostics/DiagnosticExtensions.swift
@@ -48,6 +48,7 @@ extension FixIt {
 
 extension FixIt.MultiNodeChange {
   /// Replaced a present token with a missing node.
+  ///
   /// If `transferTrivia` is `true`, the leading and trailing trivia of the
   /// removed node will be transferred to the trailing trivia of the previous token.
   static func makeMissing(_ token: TokenSyntax, transferTrivia: Bool = true) -> Self {
@@ -55,21 +56,13 @@ extension FixIt.MultiNodeChange {
   }
 
   /// Replace present tokens with missing tokens.
-  /// If `transferTrivia` is `true`, the leading and trailing trivia of the
-  /// removed node will be transferred to the trailing trivia of the previous token.
+  ///
+  /// If `transferTrivia` is `true`, the leading trivia of the first token and
+  /// the trailing trivia of the last token will be transferred to their adjecent
+  /// tokens.
   static func makeMissing(_ tokens: [TokenSyntax], transferTrivia: Bool = true) -> Self {
-    precondition(!tokens.isEmpty)
     precondition(tokens.allSatisfy({ $0.isPresent }))
-    var changes = tokens.map {
-      FixIt.Change.replace(
-        oldNode: Syntax($0),
-        newNode: Syntax($0.with(\.presence, .missing))
-      )
-    }
-    if transferTrivia {
-      changes += FixIt.MultiNodeChange.transferTriviaAtSides(from: tokens).primitiveChanges
-    }
-    return FixIt.MultiNodeChange(primitiveChanges: changes)
+    return .makeMissing(tokens.map(Syntax.init), transferTrivia: transferTrivia)
   }
 
   /// If `transferTrivia` is `true`, the leading and trailing trivia of the
@@ -104,6 +97,25 @@ extension FixIt.MultiNodeChange {
       return FixIt.MultiNodeChange()
     }
   }
+
+  /// Replace present nodes with their missing equivalents.
+  ///
+  /// If `transferTrivia` is `true`, the leading trivia of the first node and
+  /// the trailing trivia of the last node will be transferred to their adjecent
+  /// tokens.
+  static func makeMissing(_ nodes: [Syntax], transferTrivia: Bool = true) -> Self {
+    precondition(!nodes.isEmpty)
+    var changes = nodes.map {
+      FixIt.Change.replace(
+        oldNode: $0,
+        newNode: MissingMaker().rewrite($0, detach: true)
+      )
+    }
+    if transferTrivia {
+      changes += FixIt.MultiNodeChange.transferTriviaAtSides(from: nodes).primitiveChanges
+    }
+    return FixIt.MultiNodeChange(primitiveChanges: changes)
+  }
 }
 
 // MARK: - Make present
diff --git a/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift b/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
index 629d91a0..574c800a 100644
--- a/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
+++ b/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
@@ -182,7 +182,7 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
       correctToken.isMissing
     {
       // We are exchanging two adjacent tokens, transfer the trivia from the incorrect token to the corrected token.
-      changes += misplacedTokens.map { FixIt.MultiNodeChange.makeMissing($0, transferTrivia: false) }
+      changes += [FixIt.MultiNodeChange.makeMissing(misplacedTokens, transferTrivia: false)]
       changes.append(
         FixIt.MultiNodeChange.makePresent(
           correctToken,
@@ -772,7 +772,7 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
             message: ReplaceTokensFixIt(replaceTokens: [token], replacements: [.commaToken()]),
             changes: [
               .makeMissing(token, transferTrivia: false),
-              .makePresent(trailingComma),
+              .makePresent(trailingComma, leadingTrivia: token.leadingTrivia, trailingTrivia: token.trailingTrivia),
             ]
           )
         ],
@@ -813,7 +813,7 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
         fixIts: [
           FixIt(
             message: RemoveNodesFixIt(nodes),
-            changes: nodes.map { .makeMissing($0, transferTrivia: !node.deinitKeyword.trailingTrivia.contains(where: { $0.isWhitespace })) }
+            changes: .makeMissing(nodes)
           )
         ],
         handledNodes: nodes.map { $0.id }
@@ -1537,7 +1537,7 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
         fixIts: [
           FixIt(
             message: RemoveNodesFixIt(rawDelimiters),
-            changes: rawDelimiters.map { .makeMissing($0) }
+            changes: .makeMissing(rawDelimiters)
           )
         ],
         handledNodes: rawDelimiters.map { $0.id }
@@ -1858,7 +1858,7 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
             ),
             changes: [
               .makeMissing(equalToken, transferTrivia: false),
-              .makePresent(node.colon),
+              .makePresent(node.colon, leadingTrivia: equalToken.leadingTrivia, trailingTrivia: equalToken.trailingTrivia),
             ]
           )
         ],
@@ -1966,8 +1966,9 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
             FixIt(
               message: fixItMessage,
               changes: [
-                FixIt.MultiNodeChange.makePresent(detail.detail)
-              ] + unexpectedTokens.map { .makeMissing($0, transferTrivia: false) }
+                FixIt.MultiNodeChange.makePresent(detail.detail),
+                .makeMissing(unexpectedTokens)
+              ]
             )
           ],
           handledNodes: [detail.id] + unexpectedTokens.map(\.id)
```

</p>
</details>

Release Notes/511.md
```diff
@@ -2,6 +2,11 @@
 
 ## New APIs
 
+- `assertMacroExpansion` now have new parameters named `fixedSource` and `fixedSource`
+  - Description: `fixedSource` and `fixedSource` are used to assert so ensure that the source code after applying Fix-Its matches this string.
```
bnbarham, line 6: I'm not crazy and these do both say `fixedSource` right 😅? I assume this is meant to be `applyFixIts`?

kimdv: You are right! 🤦‍♂️

kimdv:
@swift-ci please test

kimdv:
@bnbarham sorry for the delay! 
I think it should be ready now

ahoppen (approved):

Tests/SwiftSyntaxMacroExpansionTest/PeerMacroTests.swift
```diff
@@ -193,4 +221,30 @@ final class PeerMacroTests: XCTestCase {
...
+          message: "can only add a completion-handler variant to an 'async' function",
+          line: 2,
+          column: 1,
+          fixIts: [FixItSpec(message: "add 'async'")]
+        )
+      ],
+      macros: ["addCompletionHandler": AddCompletionHandler.self],
+      fixedSource: """
+        @addCompletionHandler
+        func f(a: Int, for b: String, _ value: Double) async-> String { }
```
bnbarham, line 245: I assume we don't care about adding the proper trivia just in this test case? Though FWIW the completion handler macro seems like a fairly beefy macro to have in our tests in general.

bnbarham (approved):

Tests/SwiftSyntaxMacroExpansionTest/PeerMacroTests.swift
kimdv: Should I make a smaller macro? 

The idea is to have a more real-show case in a test case, so developers can see an example and what the expected result is

bnbarham: Examples seem like they would be better off living in the examples. The test should be the minimum of whatever we want to test. I don't really mind if you want to leave it for now (it was already there as far as I can tell), this was more of an off hand comment.

Having said that, if the idea is that this is an example, then that would presumably imply that we do care about the missing trivia 😅

kimdv: @bnbarham @ahoppen I've tried to reduce the test case. 
Should I try to reduce it more. It's properly not a valid real scenario, but it does the job in test I think

ahoppen: Am I missing something or are you no longer adding a `fixedSource` parameter anymore? As it stands right now, I’m pretty sure the test will fail because we are no longer adding the completion handler.

What I would suggest is that we keep the completion handler test case for now because I’m not sure if it tests anything that’s not covered by the other peer macro test cases and you add the test that you wrote as a new test case that is intended to test the `fixedSource`.

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test windows

Matejkob (commented):
Hi there!

I've been integrating the new functionality of the `assertMacroExpansion` function into our macro examples and I've run into a few issues:

1. It's not a major concern, but the `AddCompletionHandler` macro from the unit tests fails to compile when used in regular code. Although it's only a test case, I think it would be beneficial for it to produce valid Swift code. The issue arises here: https://github.com/apple/swift-syntax/blob/098b177830c0d2e614b3757241b2572d758f9143/Tests/SwiftSyntaxMacroExpansionTest/PeerMacroTests.swift#L85, where the closure would be called from `Task`, resulting in the error `Escaping closure captures non-escaping parameter 'completionHandler'`.

2. We have the following test case:

```swift
func testExpansionOnNonAsyncFunctionAppliesFixIt() {
  assertMacroExpansion(
    """
    struct Test {
      @AddCompletionHandler
      func fetchData() -> String {
        return "Hello, World!"
      }
    }
    """,
    expandedSource: """
      struct Test {
        func fetchData() -> String {
          return "Hello, World!"
        }
      }
      """,
    diagnostics: [
      DiagnosticSpec(
        message: "can only add a completion-handler variant to an 'async' function",
        line: 3,
        column: 3,
        severity: .error,
        fixIts: [FixItSpec(message: "add 'async'")]
      )
    ],
    macros: macros,
    applyFixIts: ["add 'async'"],
    fixedSource: """
    struct Test {
      @AddCompletionHandler
      func fetchData() async -> String {
        return "Hello, World!"
      }
    }
    """,
    indentationWidth: .spaces(2)
  )
}
```

The expected fixed source is:

```swift
struct Test {
  @AddCompletionHandler
  func fetchData() async -> String {
    return "Hello, World!"
  }
}
```

However, the actual result is:
Screenshot 2023-11-05 at 4 42 29 PM

This is clearly an issue on our end because the compiler is correctly applying the fix-it.

3. I'm considering whether we should still assert the `expandedSource` against the expansion before applying any fix-its, even after using the `applyFixIts` parameter in the `assertMacroExpansion` function. A comprehensive test case for clients could look like this: the user provides the `originalSource` containing an error, then the assert function applies the fix-it and attempts to expand the corrected source code. @mbrandonw and @stephencelis have implemented a similar approach in their `swift-macro-testing` tool, which can be found here: https://github.com/pointfreeco/swift-macro-testing/blob/35acd9468d40ae87e75991a18af6271e8124c261/README.md?plain=1#L148-L188. Reflecting on this further, I believe `expandedSource` should remain unchanged and we might consider adding a new parameter—`expandedSourceAfterApplyingFixIt`—to allow users to provide the expanded code post-fix-it application. This is important since a macro can offer a fix-it associated with a diagnostic of a severity other than `error`, and the expansion process can proceed regardless of the fix-it.

ahoppen:
> 1. It's not a major concern, but the AddCompletionHandler macro from the unit tests fails to compile when used in regular code.

I don’t have strong opinions on this one. It is only a test case to test that peer macros work on a syntactical level and for me it’s OK if the generated code has semantic issues. As the comment on the top of the file says

> The macros defined in this file are intended to test the behavior of MacroSystem. Many of them do not serve as good examples of how macros should be written. In particular, they often lack error handling because it is not needed in the few test cases in which these macros are invoked.

---

> 2. We have the following test case. […] The expected fixed source is […]However, the actual result is […] This is clearly an issue on our end because the compiler is correctly applying the fix-it.

Oh, could you file an issue?

---

> 3. I'm considering whether we should still assert the `expandedSource` against the expansion before applying any fix-its, even after using the `applyFixIts` parameter in the `assertMacroExpansion` function.

That does sound reasonable to me. Would you like to open a PR with the change? We can discuss the details on that PR.

Matejkob:
> > 2. We have the following test case. […] The expected fixed source is […]However, the actual result is […] This is clearly an issue on our end because the compiler is correctly applying the fix-it.
> 
> Oh, could you file an issue?
> 

I'm on it. 

> > 3. I'm considering whether we should still assert the `expandedSource` against the expansion before applying any fix-its, even after using the `applyFixIts` parameter in the `assertMacroExpansion` function.
> 
> That does sound reasonable to me. Would you like to open a PR with the change? We can discuss the details on that PR.

I'm on it.

PR #2022, opened 2023-08-07
Re-architect format.py in swift

ahoppen:
Make a couple of changes to `format.py` and use the opportunity to re-write it in Swift.

1. Always build a local copy of swift-format in `swift-syntax/.swift-format-build` in release mode. This eliminates issues that occur if you had `swift-format` installed on your system and `format.py` would pick it up
2. Since the local swift-format build is persistent (doesn’t live in `/tmp`), we can build it in release. This increases the first run but decreases the time for any subsequent runs from ~15s to ~1s, which should pay usually pay off for frequent contributors because of faster turnaround times.
3. Add a little tip to add `swift-format` as a git hook to Contributing.md

And as an added bonus, the rewrite reduced the runtime of `format.(py|swift)` from 1.2s to 800ms.


ahoppen:
@swift-ci Please test

ahoppen:
https://github.com/apple/swift/pull/67777

@swift-ci Please test

CONTRIBUTING.md
```diff
@@ -19,12 +19,22 @@ swift-syntax is a SwiftPM package, so you can build and test it using anything t
...
-
-If you are seeing surprising formatting results, you most likely have a `swift-format` installed on your system that’s not the most recent version built from the `main` branch. To fix this, clone [swift-format](http://github.com/apple/swift-format), build it using `swift build` and pass the freshly built executable to `format.py` as `--swift-format path/to/swift-format/.build/debug/swift-format`. Alternatively, you can uninstall `swift-format` on your system and `format.py` will build it from scratch.
+To format your changes run `format.swift` at the root of this repository from the terminal as `./format.swift`. It will build a local copy of swift-format from the `main` branch and format the repository. Since it is building a release version of `swift-format`, the first run will take about 5 minutes. Subsequent runs take less than 2 seconds.
 
 Generated source code is not formatted to make it easier to spot changes when re-running code generation.
 
+> 💡 Tip: You can add a git hook to ensure all commits to the swift-syntax repository are correctly formatted. 
+> 1. Save the following contentes to `swift-syntax/.git/hooks/pre-commit`
```
kimdv, line 27: Is there a typo in `contentes`?

format.swift
```diff
@@ -0,0 +1,503 @@
...
+  static var swiftExec: URL {
+    get throws {
+      return try lookupExecutable(for: "swift")
+    }
+  }
+
+  /// The root of the swift-syntax package
+  static var packageDir: URL {
+    URL(fileURLWithPath: #file)
+      .deletingLastPathComponent()
```
kimdv, line 56: Some places we use `return` in `var`s like above or a bit below. 

Do we have a specific style guide for that in SwiftSyntax. Just to keep consistency?

format.swift
```diff
@@ -0,0 +1,503 @@
...
+  if let arguments = arguments {
+    message += " \(arguments.joined(separator: " "))"
+  }
+
+  print(message)
+}
+
+// MARK: - Paths
+
+enum Paths {
```
kimdv, line 40: The extension below this is `fileprivate` and this isn't 

I don't think that is a big issue. But what do you think of also making this fileprivate or removing below. 
Just for the sake of consistency? 

Same for the usage of `public` a bit further down?

kimdv (approved):

format.swift
Matejkob: I've been using [`SwiftFormat`](https://github.com/nicklockwood/SwiftFormat) for years and it has the ability to keep/add a `return` statement in such cases or remove (based on config file). I'm new to `swift-format` so I'm unsure if this tool has similar abilities - but it would be really cool to have something like it here. 

Oh, and as I was sneaking through the code, my nitpick-dar (that's nitpick radar) also did a little beep-beep about this `return`s. 📡😂

ahoppen: We don’t have a consistent style here. I personally dislike the implicit return, especially if the returned expression spans multiple lines (which is very subjective and not based on facts) but acknowledge that the general style seems to be to omit them. I have removed the `return`s.

AFAIK there’s no rule for that in swift-format. It would be nice to add one though.

format.swift
ahoppen: Good catch. It just doesn’t matter in a single file script but I made all types `internal` now for consistency.

format.swift
kimdv: Opened an issue here: https://github.com/apple/swift-format/issues/582

format.swift
```diff
@@ -0,0 +1,548 @@
...
+      } else {
+        throw Error.swiftFormatNotFound
+      }
+    }
+  }
+}
+
+SigIntListener.registerSigIntSubprocessTerminationHandler()
+do {
+  try SwiftSyntaxFormatter().main()
```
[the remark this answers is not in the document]
ahoppen: To be honest: I like that it’s so quiet. It terminated with an exit code of 0, so there wasn’t a failure…

ahoppen:
https://github.com/apple/swift/pull/67777

@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
https://github.com/apple/swift/pull/67777

@swift-ci Please test Windows

ahoppen:
https://github.com/apple/swift/pull/67777

@swift-ci Please test Windows

ahoppen:
https://github.com/apple/swift/pull/67777

@swift-ci Please test macOS

ahoppen:
https://github.com/apple/swift/pull/67777

@swift-ci Please test

ahoppen:
https://github.com/apple/swift/pull/67777

@swift-ci Please test

ahoppen:
https://github.com/apple/swift/pull/67777

@swift-ci Please test

ahoppen:
https://github.com/apple/swift/pull/67777

@swift-ci Please test macOS

ahoppen:
https://github.com/apple/swift/pull/67777

@swift-ci Please test macOS

ahoppen:
https://github.com/apple/swift/pull/67777

@swift-ci Please test macOS

ahoppen:
https://github.com/apple/swift/pull/67777

@swift-ci Please test Linux

ahoppen:
https://github.com/apple/swift/pull/67777

@swift-ci Please test Windows

ahoppen:
https://github.com/apple/swift/pull/67777

@swift-ci Please test Linux

ahoppen:
https://github.com/apple/swift/pull/67777

@swift-ci Please test

CONTRIBUTING.md
```diff
@@ -19,12 +19,22 @@ swift-syntax is a SwiftPM package, so you can build and test it using anything t
...
 
-To format your changes run `format.py` at the root of this repository. If you have a `swift-format` executable ready, you can pass it to `format.py`. If you do not, `format.py` will build its own copy of `swift-format` in `/tmp/swift-format`.
-
-If you are seeing surprising formatting results, you most likely have a `swift-format` installed on your system that’s not the most recent version built from the `main` branch. To fix this, clone [swift-format](http://github.com/apple/swift-format), build it using `swift build` and pass the freshly built executable to `format.py` as `--swift-format path/to/swift-format/.build/debug/swift-format`. Alternatively, you can uninstall `swift-format` on your system and `format.py` will build it from scratch.
+To format your changes run `format.swift` at the root of this repository from the terminal as `./format.swift`. It will build a local copy of swift-format from the `main` branch and format the repository. Since it is building a release version of `swift-format`, the first run will take few minutes. Subsequent runs take less than 2 seconds.
```
bnbarham, line 22:
```suggestion
To format your changes, run `format.swift` at the root of this repository from the terminal as eg. `./format.swift`. It will build a local copy of swift-format from the `main` branch and format all files. Since it is building a release version of `swift-format`, the first run will take few minutes. Subsequent runs are much faster (a couple seconds).
```

CONTRIBUTING.md
```diff
@@ -19,12 +19,22 @@ swift-syntax is a SwiftPM package, so you can build and test it using anything t
...
 
 Generated source code is not formatted to make it easier to spot changes when re-running code generation.
 
+> 💡 Tip: You can add a git hook to ensure all commits to the swift-syntax repository are correctly formatted.
```
bnbarham, line 26:
```suggestion
> [!NOTE]
> You can add a git hook to ensure all commits to the swift-syntax repository are correctly formatted. 
```

CONTRIBUTING.md
```diff
@@ -19,12 +19,22 @@ swift-syntax is a SwiftPM package, so you can build and test it using anything t
...
-
-If you are seeing surprising formatting results, you most likely have a `swift-format` installed on your system that’s not the most recent version built from the `main` branch. To fix this, clone [swift-format](http://github.com/apple/swift-format), build it using `swift build` and pass the freshly built executable to `format.py` as `--swift-format path/to/swift-format/.build/debug/swift-format`. Alternatively, you can uninstall `swift-format` on your system and `format.py` will build it from scratch.
+To format your changes run `format.swift` at the root of this repository from the terminal as `./format.swift`. It will build a local copy of swift-format from the `main` branch and format the repository. Since it is building a release version of `swift-format`, the first run will take few minutes. Subsequent runs take less than 2 seconds.
 
 Generated source code is not formatted to make it easier to spot changes when re-running code generation.
 
+> 💡 Tip: You can add a git hook to ensure all commits to the swift-syntax repository are correctly formatted. 
+> 1. Save the following contents to `swift-syntax/.git/hooks/pre-commit`
+> ```bash
+> #!/bin/bash
```
bnbarham, line 29:
```suggestion
> #!/usr/bin/env bash
```

CONTRIBUTING.md
```diff
@@ -19,12 +19,22 @@ swift-syntax is a SwiftPM package, so you can build and test it using anything t
...
+> set -e
+> SOURCE_DIR=$(realpath "$(dirname $0)/../..")
+> "$SOURCE_DIR/format.swift" --lint
+> ```
+> 2. Mark the file as executable: `chmod a+x swift-syntax/.git/hooks/pre-commit`
+> 3. If you have any global git hooks installed, be sure to call them at the end of the script: `path/to/global/hooks/pre-commit "$@"
```
bnbarham, line 35:
```suggestion
> 3. If you have any global git hooks installed, be sure to call them at the end of the script with `/path/to/global/hooks/pre-commit "$@"`
```

format.swift
```diff
@@ -0,0 +1,548 @@
...
+#elseif canImport(Glibc)
+import Glibc
+#endif
+
+// MARK: - Logging
+
+// It’s a little unfortunate that we need to duplicate all these functions from
+// swift-syntax-dev-utils. But executing format.swift as a script is significantly
+// faster than running it from a package using `swift run` (~1s vs ~15s), which
+// makes a big difference when running the script multiple times.
```
bnbarham, line 27: Little unfortunate indeed :(

format.swift
```diff
@@ -0,0 +1,548 @@
...
+    get throws {
+      try lookupExecutable(for: "git")
+    }
+  }
+
+  static var swiftExec: URL {
+    get throws {
+      try lookupExecutable(for: "swift")
+    }
+  }
```
bnbarham, line 56: Could compute these up front and just fatal error rather than throw? Same with most of these computed properties really.

format.swift
```diff
@@ -0,0 +1,548 @@
...
+    guard let executable else {
+      throw ExecutableLookupError.notFound(executableName: filename)
+    }
+    return executable
+  }
+}
+
+extension URL {
+  var isExecutableFile: Bool {
+    return (self.isFile(path) || self.isSymlink(path)) && FileManager.default.isExecutableFile(atPath: path)
```
bnbarham, line 118: Do we need the file/symlink checks here? Does `isExecutableFile` return true for directories?

format.swift
```diff
@@ -0,0 +1,548 @@
...
+      SigIntListener.hasReceivedSigInt = true
+      for process in SigIntListener.runningSubprocesses {
+        kill(process.processIdentifier, SIGINT)
+      }
+    }
+    #endif
+  }
+}
+
+/// Provides convenience APIs for launching and gathering output from a subprocess
```
bnbarham, line 157: I'm assuming just a copy paste, so haven't actually checked this. FWIW this makes me very sad :(

format.swift
```diff
@@ -0,0 +1,548 @@
...
+  let stderr: String
+  let exitCode: Int
+}
+
+/// Error thrown if `stdout` or `stderr` could not be decoded as UTF-8.
+struct FailedToDecodeUTF8Error: Error {
+  let data: Data
+}
+
+// MARK: Actual implementation
```
bnbarham, line 269: Could you flip this? ie. implementation at the start of the file, copied junk that we don't care about at the end?

format.swift
```diff
@@ -0,0 +1,548 @@
...
+
+          Run the following command to format swift-syntax
+
+            swift-syntax/format.swift
+
+          If the issue persists, try updating swift-format by running
+
+            swift-syntax/format.swift --update
+
+          Should that still fail, fix any remaining issues manually and verify they match the swift-format style using
```
bnbarham, line 334: Do we expect this to fail at all?

format.swift
```diff
@@ -0,0 +1,548 @@
...
+      executableURL: Paths.gitExec,
+      arguments: ["-C", Paths.swiftFormatBuildDir.path] + arguments
+    ).run(
+      verbose: CommandLineArguments.verbose,
+      captureStdout: false,
+      captureStderr: false
+    )
+  }
+
+  /// Run `swift` with for the `.swift-format-build` package with the provided arguments.
```
bnbarham, line 360:
```suggestion
  /// Run `swift` for the `.swift-format-build` package with the provided arguments.
```

format.swift
```diff
@@ -0,0 +1,548 @@
...
+  }
+
+  /// Ensure that we have an up-to-date checkout of swift-format in `.swift-format-build`.
+  func cloneOrUpdateSwiftFormat() throws {
+    if FileManager.default.fileExists(atPath: Paths.swiftFormatBuildDir.appendingPathComponent(".git").path) {
+      try runGitCommand("checkout", branch)
+      try runGitCommand("pull")
+    } else {
+      try FileManager.default.createDirectory(atPath: Paths.swiftFormatBuildDir.path, withIntermediateDirectories: true)
+      try runGitCommand("clone", "https://github.com/apple/swift-format.git", ".")
```
bnbarham, line 379:
```suggestion
      try runGitCommand("clone", "https://github.com/apple/swift-format.git", Paths.swiftFormatBuildDir.path)
```

Or am I missing something here? Wouldn't need the create directory above either.

format.swift
```diff
@@ -0,0 +1,548 @@
...
+    if !swiftFormatExec.isExecutableFile {
+      throw Error.swiftFormatNotFound
+    }
+    return swiftFormatExec
+  }
+
+  /// Get the list of files that should be formatted using swift-format.
+  ///
+  /// This excludes some files like generated files or test inputs.
+  private func filesToFormat() -> [URL] {
```
bnbarham, line 411: Could we make this more top level? eg. a global with the directories to skip?

format.swift
```diff
@@ -0,0 +1,548 @@
...
+        This script automatically builds the '\(branch)' branch of swift-format \
+        in the '\(Paths.swiftFormatBuildDir.lastPathComponent)' directory of this \
+        repository.
+
+        Since this is performing a release build, the first run of this script \
+        will take about 5 minutes. Subsequent calls only take 1-2 seconds.
+
+        Optional Arguments:
+          --help      Show this help message
+          --lint      Instead of formatting in-place verify that the files are correctly formatted.
```
bnbarham, line 498:
```suggestion
          --lint      Instead of formatting in-place, verify that the files are correctly formatted.
```

format.swift
```diff
@@ -0,0 +1,548 @@
...
+      print(
+        """
+        Format files in SwiftSyntax using swift-format. 
+
+        This script automatically builds the '\(branch)' branch of swift-format \
+        in the '\(Paths.swiftFormatBuildDir.lastPathComponent)' directory of this \
+        repository.
+
+        Since this is performing a release build, the first run of this script \
+        will take about 5 minutes. Subsequent calls only take 1-2 seconds.
```
bnbarham, line 494: I'd drop this, since we output a message below about it building anyway

format.swift
ahoppen: I can’t find the exact PR but @neonichu made some changes to improve `swift run` performance drastically by not having to re-build the build graph. With that change, we can make `format.py` part of swift-syntax-dev-utils`. 🎉

format.swift
ahoppen: I don’t really see any benefit in doing that. For example, there’s no reason to stop script execution if `diff` cannot be found (for verification that code-gen generated the right files) but all you want to do is format swift-syntax using swift-format.

format.swift
ahoppen: Yes, exactly. FWIW, this was copied from

https://github.com/apple/swift-tools-support-core/blob/3b13e439a341bbbfe0f710c7d1be37221745ef1a/Sources/TSCBasic/FileSystem.swift#L359-L362

format.swift
ahoppen: During the initial introduction of swift-format, I had two or three cases where swift-format complained but wasn’t able to fix the issue by itself. I don’t fee super strongly about this hint.

format.swift
bnbarham: Just seems weird to compute them every time 🤷 Reasonable point with `diff` though. I don't really mind in any case, it's just for formatting.

format.swift
ahoppen: `runGitCommand` adds an implicit `-C Paths.swiftFormatBuildDir.path` and the `.` is relative to that.

format.swift
bnbarham: Huh, I didn't know that was a thing. Feel free to keep, is just an extra sentence 👍

format.swift
ahoppen: Why would this make more sense as a global function? Also, it might be different now that formatting is part of swift-syntax-dev-utils.

format.swift
bnbarham: Ah I see 👍

format.swift
bnbarham: I didn't mean this function specifically, more the list of filenames. It would be nice to have it at the top of the file where it's nice and obvious.

ahoppen: I just thought about it and I couldn’t find a really nice way to put it at the top given that I
- Don’t want to introduce globals and
- Want to maintain list members before functions

ahoppen:
https://github.com/apple/swift/pull/67777

@swift-ci Please test

ahoppen:
https://github.com/apple/swift/pull/67777

@swift-ci Please test macOS

ahoppen:
https://github.com/apple/swift/pull/67777

@swift-ci Please test Linux

ahoppen:
https://github.com/apple/swift/pull/67777

@swift-ci Please test Linux

ahoppen:
https://github.com/apple/swift/pull/67777

@swift-ci Please test

ahoppen:
https://github.com/apple/swift/pull/67777

@swift-ci Please test macOS

ahoppen:
apple/swift#67777

@swift-ci Please test macOS

Issue #1972, opened 2023-08-01
Remove `CanImportExprSyntax` and provide API to interpret an `ExprSyntax` as a `VersionTupleSyntax`


ahoppen:
`canImport` inside of `#if` should get parsed the same as in a normal expression position, eliminating the need for `CanImportExprSyntax` (sparked by [link](https://github.com/apple/swift-syntax/pull/1816#discussion_r1235775605)). This means that the parser won’t be able to generate diagnostics for invalid version numbers inside `canImport` anymore. 

Instead, we should have a function on `ExprSyntax` that interprets the expression (which would be a `FloatingLiteralExprSyntax` for eg. `1.2`) as a `VersionTupleSyntax`. If the `ExprSyntax` doesn’t represent a version, that function should return a list of diagnostics. `SwiftIfConfig` would then call that getter and would be responsible for reporting any errors that occur from interpreting the `ExprSyntax` of the `_version` argument as `VersionTupleSyntax`.

ahoppen:
Tracked in Apple’s issue tracker as rdar://113236781

PR #2025, opened 2023-08-08
Remove CanImportExprSyntax and provide API to interpret an ExprSyntax as a VersionTupleSyntax

StevenWong12:
Resolves https://github.com/apple/swift-syntax/issues/1972


Tests/SwiftParserTest/ExpressionInterpretedAsVersionTupleTests.swift
```diff
@@ -0,0 +1,27 @@
...
+
+final class ExpressionInterpretedAsVersionTupleTest: XCTestCase {
+  func testExpressionInterpretedAsVersionTuple() throws {
+    let expression: ExprSyntax = "2.2.2.2"
+    let versionTuple = try XCTUnwrap(expression.asVersionTuple(with: 4))
+    let versionTupleComponents = try XCTUnwrap(versionTuple.components)
+    XCTAssertTrue(!versionTuple.hasError)
+    XCTAssertEqual(versionTuple.description, "2.2.2.2")
+    XCTAssertEqual(versionTuple.major.description, "2")
+    XCTAssertEqual(versionTupleComponents.description, ".2.2.2")
```
ahoppen, line 25: You could just check the substructure instead

```swift
let subtreeMatcher = SubtreeMatcher(versionTuple, markers: [:])
try subtreeMatcher.assertSameStructure(VersionTupleSyntax(
  major: .integerLiteral("2"),
  components: VersionComponentListSyntax([
    VersionComponentSyntax(
      period: .periodToken(),
      number: .integerLiteral("2")
    ),
    VersionComponentSyntax(
      period: .periodToken(),
      number: .integerLiteral("2")
    ),
    VersionComponentSyntax(
      period: .periodToken(),
      number: .integerLiteral("2")
    )
  ])
))
```

Sources/SwiftParser/ExpressionInterpretedAsVersionTuple.swift
```diff
@@ -0,0 +1,21 @@
...
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+@_spi(RawSyntax) import SwiftSyntax
+
+extension ExprSyntax {
+  public func asVersionTuple(with maxComponentCount: Int) -> VersionTupleSyntax? {
```
ahoppen, line 16: Could you add a doc comment explaining what this does and why it’s necessary?

---

`asVersionTuple` sound like a cast to me. I would prefer something like `interpretAsVersionTuple`.

Sources/SwiftParser/ExpressionInterpretedAsVersionTuple.swift
```diff
@@ -0,0 +1,21 @@
...
+    var parser = Parser(self.description)
+    let raw = parser.parseVersionTuple(maxComponentCount: maxComponentCount)
```
ahoppen, line 18: You should make sure that all tokens in the `ExprSyntax` are represented idk the `VersionTupleSyntax`, similar to what we do when creating syntax nodes from string literals 

https://github.com/apple/swift-syntax/blob/e42bece52df2496d60b1b2a762fee9ffde7fc205/Sources/SwiftParser/generated/LayoutNodes%2BParsable.swift#L34

And now that I think about it, I think the best thing to do, would be add a parser entry function for `VersionTupleSyntax` so that all that code gets auto-generated, including the following, which is necessary for memory safety

https://github.com/apple/swift-syntax/blob/e42bece52df2496d60b1b2a762fee9ffde7fc205/Sources/SwiftParser/generated/LayoutNodes%2BParsable.swift#L23-L32


---

Could you add a test case for `2.0 + 1`. `+ 1` should be unexpected nodes in that case.

StevenWong12: I think making this a parser entry function is a good idea! And we have to pass the `maxComponentCount` parameter to that function so I add a helper function below.

Sources/SwiftParser/generated/LayoutNodes+Parsable.swift
```diff
@@ -345,4 +363,8 @@ fileprivate extension Parser {
     }
     return node
   }
+  
+  mutating func parseVersionTupleWithFourComponents() -> RawVersionTupleSyntax {
+    return self.parseVersionTuple(maxComponentCount: 4)
```
ahoppen, line 368: I’m not a huge fan of this. 4 seems as arbitrary a value as any other, so we shouldn’t default to it. And actually, since we are moving more and more of the version handling logic out of the parser and into higher-level libraries, I’m thinking that we should remove the `maxComponentCount` parameter altogether. If, e.g. the IfConfig library can only handle 4 components, it should be responsible for diagnosing this, instead of the parser.

So, my suggestion is to remove the `maxComponentCount` parameter, even if it means that some of the current diagnostics will regress. We have the expectation that these diagnostics will be emitted by the libraries that interpret the version tuple.

StevenWong12: That makes sense to me👍.

Tests/SwiftParserTest/ExpressionInterpretedAsVersionTupleTests.swift
```diff
@@ -0,0 +1,73 @@
...
+              TokenSyntax.integerLiteral("1"),
+              TokenSyntax.periodToken(),
+              TokenSyntax.binaryOperator("+"),
+              TokenSyntax.integerLiteral("1"),
+            ]
+          )
+        )
+      )
+    )
+  }
```
ahoppen, line 72: Two interesting cases I would like to test are `.1.2` and `1.2.`.

Tests/SwiftParserTest/ExpressionInterpretedAsVersionTupleTests.swift
```diff
@@ -0,0 +1,73 @@
...
+import SwiftParser
+import SwiftSyntax
+import _SwiftSyntaxTestSupport
+import XCTest
+
+final class ExpressionInterpretedAsVersionTupleTest: XCTestCase {
+  func testExpressionInterpretedAsVersionTuple() throws {
+    let expression: ExprSyntax = "2.2.2.2"
+    let versionTuple = try XCTUnwrap(expression.interpretAsVersionTuple())
+    let subtreeMatcher = SubtreeMatcher(Syntax(versionTuple), markers: [:])
```
ahoppen, line 22: As of https://github.com/apple/swift-syntax/pull/2027, you don’t need the `Syntax` upcast here anymore. Same for a couple more `Syntax` upcasts in this file

Sources/SwiftParser/ExpressionInterpretedAsVersionTuple.swift
```diff
@@ -0,0 +1,21 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+@_spi(RawSyntax) import SwiftSyntax
+
+/// When we have `SwiftIfConfig` library, we will move the diagnose for `canImport` expression there and this function should help determine whether an `ExprSyntax` is a legal `VersionTupleSyntax`, which is the second parameter of a `canImport` expression
```
ahoppen, line 15: I don’t think this comment provides any value, especially if you consider that every client may use `interpretAsVersionTuple`

Sources/SwiftParser/ExpressionInterpretedAsVersionTuple.swift
```diff
@@ -0,0 +1,21 @@
...
+extension ExprSyntax {
+  public func interpretAsVersionTuple() -> VersionTupleSyntax? {
```
ahoppen, line 17: 1. According to the [naming guidelines](https://www.swift.org/documentation/api-design-guidelines/#name-according-to-side-effects), this should be `interpretedAsVersionTuple`. 
2. This can be a computed property since it doesn’t take any parameters
3. This should return a non-optional.

Sources/SwiftParser/ExpressionInterpretedAsVersionTuple.swift
StevenWong12: Would it make more sense to comment like below?
> This property can be used to help determine whether an `ExprSyntax` is a legal `VersionTupleSyntax`.

ahoppen: I would prefer to add a comment that explains why this is necessary. Something like

```swift
Parse the source code of this node as a `VersionTupleSyntax`.

In some situations, like the `_version` argument of the `canImport` condition in `#if`, the version gets parsed as a normal expression, which results in 
- an integer literal for single-component versions 
- a floating point number for two-component versions
- a floating point number with a member accesses for versions with more than three components

This is done so the parser can parse `canImport` like any other function call, reducing the need for special handling.

This property re-interprets such an expression as a version tuple in cases where the client know that it should semantically represent a version. 
```

Sources/SwiftParser/ExpressionInterpretedAsVersionTuple.swift
```diff
@@ -0,0 +1,30 @@
...
+/// Parse the source code of this node as a `VersionTupleSyntax`.
+///
+/// In some situations, like the `_version` argument of the `canImport` condition in `#if`, the version gets parsed as a normal expression, which results in
+/// - an integer literal for single-component versions
+/// - a floating point number for two-component versions
+/// - a floating point number with a member accesses for versions with more than three components
+///
+/// This is done so the parser can parse `canImport` like any other function call, reducing the need for special handling.
+///
+/// This property re-interprets such an expression as a version tuple in cases where the client know that it should semantically represent a version.
```
ahoppen, line 24: Could you add the comment on the `interpretedAsVersionTuple` member? Otherwise it won’t show up in docc or if you invoke Quick Help on the property.

ahoppen (approved):

ahoppen:
@swift-ci Please test

ahoppen:
Picking this PR up.

ahoppen:
@swift-ci Please test

bnbarham (approved):

rintaro (approved):

PR #2030, opened 2023-08-08
Remove SyntaxTransformVisitor from the package release of swift-syntax

ahoppen:
`SyntaxTransformVisitor` was introduced for ASTGen but it wasn’t a great API and we don’t want to use it there either. So, let’s remove it.


ahoppen:
@swift-ci Please test

kimdv (approved):

SimplyDanny:
What's the recommended alternative? I found this API quite helpful for things like collecting information about the syntax tree, e.g. deciding if a boolean expression tree might evaluated to a constant value.

ahoppen:
What I usually do is write a `SyntaxVisitor` that contains a member, like the following. This also allows you to skip the visitation of subtrees that are not needed.

https://github.com/apple/swift-syntax/blob/b4749403093de49b11418d7bd80f7ee2b4a926dd/Sources/SwiftParserDiagnostics/PresenceUtils.swift#L17-L35

I agree that there is some room for a well-designed API that allow the computation of a single, aggregated value from a syntax tree. For the case I linked above, the requirements would be that this API has three actions:
- Visit the subtree
- Skip visitation of the subtree since it’s not relevant
- Return a value

I am just very certain that `SyntaxTransformVisitor` is not the API that we want.

---

Alternatively, you could write your own type with all the visit functions for the syntax nodes that you can handle and then you would need to have your own dispatch function like the following. That also has the advantage that you’re explicit about the nodes that you do handle.

```swift
func visit(_ node: Syntax) {
  switch node.as(SyntaxEnum.self) {
  case .integerLiteralExpr(let integerLiteral): self.visit(integerLiteral)
  ...
  }
}
```

SimplyDanny:
Got that. The problem with the member in the visitor class is its unhandy call side, though. For example in a `ConstantExprVisitor` there might be the method
```swift
func visit(_ node: InfixOperatorExprSyntax) -> SyntaxVisitorContinueKind {
    // ...
    let leftVisitor = Self()
    leftVisitor.visit(node.leftOperand)
    let leftState = leftVisitor.state
    // ...
    let rightVisitor = Self()
    rightVisitor.visit(node.rightOperand)
    let rightState = rightVisitor.state
    // ...
}
```
that requires a new visitor instance and a few lines of code for each sub-tree analysis. Performance-wise this might not be an issue, but the code unfortunately is a bit awkward.

I agree, however, that an early stop of traversal is a good reason to replace `SyntaxTransformVisitor` with something more advanced.

ahoppen:
In the case of operator evaluation, I think the best thing to do for now, is keep your current type that conformed to `SyntaxTransformVisitor` and add to copy `SyntaxTransformVisitor.visit(_:Syntax)` with all the cases that you handle.

 https://github.com/apple/swift-syntax/blob/acf23c80f58e10292fd5e58ca24302f11a81ee81/Sources/SwiftSyntax/generated/SyntaxTransform.swift#L3329-L3884

Another upside of this approach is that you can e.g. propagate errors using `throws` if the expression contains a term that you don’t know how to handle.

PR #2034, opened 2023-08-08
Format doc comments so they render better in docc

ahoppen:
Add a couple of line breaks to doc comments so they render more nicely in docc.

For example, `DO NOT CONFORM TO THIS PROTOCOL YOURSELF!` should not show up in the abstract of the base syntax nodes

### DIFF
--- a/CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
@@ -242,7 +242,9 @@ public let DECL_NODES: [Node] = [
     base: .decl,
     nameForDiagnostics: "associatedtype declaration",
     documentation: """
-      An associated type declaration like the following.
+      An `associatedtype` declaration
+
+      An example of an associatedtype declaration is
 
       ```swift
       associatedtype Item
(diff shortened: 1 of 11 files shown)


ahoppen:
@swift-ci Please test

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
```diff
@@ -242,7 +242,9 @@ public let DECL_NODES: [Node] = [
     base: .decl,
     nameForDiagnostics: "associatedtype declaration",
     documentation: """
-      An associated type declaration like the following.
+      An `associatedtype` declaration
+
+      For exmaple, an associatedtype declaration can look like the following.
```
bnbarham, line 247: Typo in `example` but I think for all these I would just remove everything after `For example`

bnbarham (approved):

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
ahoppen: Removed `For example` from all these comments

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

kimdv:
@swift-ci Please test windows

PR #2035, opened 2023-08-08
Add documentation on pattern bindings

ahoppen:
Another area that’s not super obvious. Add some documentation that hopefully sheds some light into why pattern bindings exist and what they do.

And also improve documentation of `AttributeSyntax.arguments`.

### DIFF
--- a/CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
@@ -2326,7 +2326,7 @@ public let DECL_NODES: [Node] = [
     documentation: """
       Declaration of one or more variables
 
-      The core of a variable declaration consists of a binding specifier (`let` or `var),
+      The core of a variable declaration consists of a binding specifier (`let` or `var`),
       followed by any number of pattern bindings, which define the variables.
       """,
     traits: [
@@ -2364,7 +2364,7 @@ public let DECL_NODES: [Node] = [
           initializers and accessors.
 
           A variable declaration can contain multiple pattern bindings, because it’s possible
-          to define multiple variables after a single `let` keyword as follows.
+          to define multiple variables after a single `let` keyword, for example
 
           ```swift
           let x: Int = 1, y: Int = 2
--- a/CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
@@ -1619,29 +1619,56 @@ public let DECL_NODES: [Node] = [
     kind: .patternBinding,
     base: .syntax,
     nameForDiagnostics: nil,
+    documentation: """
+      Defines variables inside a variable declaration.
+      """,
     traits: [
       "WithTrailingComma"
     ],
     children: [
       Child(
         name: "Pattern",
-        kind: .node(kind: .pattern)
+        kind: .node(kind: .pattern),
+        documentation: """
+          The pattern that defines the variables.
+
+          In simple variable declarations this is an ``IdentifierPatternSyntax``, which defines
+          the name of a single variable.
+
+          In more complex variable declaration, this can, for example, be a ``TuplePatternSyntax``
+          that destructures a tuple.
+
+          ```swift
+          let (x, y) = (1, 2)
+          ```
+          """
       ),
       Child(
         name: "TypeAnnotation",
         kind: .node(kind: .typeAnnotation),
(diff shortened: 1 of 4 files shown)


ahoppen:
@swift-ci Please test

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
```diff
@@ -2296,6 +2323,12 @@ public let DECL_NODES: [Node] = [
     kind: .variableDecl,
     base: .decl,
     nameForDiagnostics: "variable",
+    documentation: """
+      Declaration of one or more variables
+
+      The core of a variable declaration consists of a binding specifier (`let` or `var),
```
bnbarham, line 2329:
```suggestion
      The core of a variable declaration consists of a binding specifier (`let` or `var`),
```

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
```diff
@@ -2316,13 +2349,28 @@ public let DECL_NODES: [Node] = [
...
-        kind: .collection(kind: .patternBindingList, collectionElementName: "Binding")
+        kind: .collection(kind: .patternBindingList, collectionElementName: "Binding"),
+        documentation: """
+          The pattern bindings that define the actual variables.
+
+          The pattern bindings contain the declared variables’ names, their types,
+          initializers and accessors.
+
+          A variable declaration can contain multiple pattern bindings, because it’s possible
+          to define multiple variables after a single `let` keyword as follows.
```
bnbarham, line 2367: I'd drop the `as follows`. Or use `eg`/`, for example`.

bnbarham (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
```diff
@@ -2316,13 +2349,28 @@ public let DECL_NODES: [Node] = [
...
+
+          The pattern bindings contain the declared variables’ names, their types,
+          initializers and accessors.
+
+          A variable declaration can contain multiple pattern bindings, because it’s possible
+          to define multiple variables after a single `let` keyword, for example
+
+          ```swift
+          let x: Int = 1, y: Int = 2
+          ```
```
kimdv, line 2371: I've seen that some SwiftUI [documentation](https://developer.apple.com/documentation/swiftui/view/labelstyle(_:)) also have code examples on how to use. 

What do you think of also adding it here to make it more explicit for consumers on how to make the example you provided?

kimdv (approved):

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
ahoppen: I'm not entirely sure what you mean.

kimdv: I mean added the example that constructs `let x: Int = 1, y: Int = 2`.
I think it would make it easier to consume for the user to see actual examples in the documentation. 

Maybe something like

```
A variable declaration can contain multiple pattern bindings, because it’s possible
          to define multiple variables after a single `let` keyword, for example
          
```swift
VariableDeclSyntax(bindingSpecifier: .keyword(.let)) {
            PatternBindingSyntax(pattern: IdentifierPatternSyntax(identifier: .identifier("x")),
                                 typeAnnotation: TypeAnnotationSyntax(type: IdentifierTypeSyntax(name: .identifier("Int"))),
                                 initializer: InitializerClauseSyntax(value: IntegerLiteralExprSyntax(1))
            )

            PatternBindingSyntax(pattern: IdentifierPatternSyntax(identifier: .identifier("y")),
                                 typeAnnotation: TypeAnnotationSyntax(type: IdentifierTypeSyntax(name: .identifier("Int"))),
                                 initializer: InitializerClauseSyntax(value: IntegerLiteralExprSyntax(2))
            )
        }
```
          
generates          

          ```swift
          let x: Int = 1, y: Int = 2
          ```
```

ahoppen: Oh, I don’t think that would be helpful in this context. The purpose of this comment is to educate the reader why a single variable declaration might have multiple pattern bindings and thus explain why the API seems so complicated. I very much doubt that anyone wants to generate such a declaration.

PR #2038, opened 2023-08-09
Simplify `SyntaxRewriter` and `SyntaxVisitor`

ahoppen:
Use generics instead of code generation to simplify `SyntaxVisitor` and `SyntaxRewriter` and improve their compile times. 

Debug performance is the same between the old and the new version, release performance regressed by ~3%. I’m putting it out here to decide whether the runtime performance regression is worth the clean up in code – if you would have proposed it the other way round, I don’t think I would have agreed to generate all these functions for a 3% performance increase.


allevato:
Out of curiosity, what were the compile time improvements that you saw?

allevato:
> But the compile time of SyntaxVisitor.swift, TokenKind.swift, Tokens.swift, TriviaPieces.swift went down from 8.74s to 1.8s.

😮 

IMO that's absolutely worth the minor runtime regression.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

bnbarham (approved):
Performance hit seems worth it for the simplification to me 👍

ahoppen:
@swift-ci Please test macOS

PR #2046, opened 2023-08-09
Add a `?` in the `Children` section for optional collections

ahoppen:
Potentially, this is why I didn’t think about the fact that syntax collections can be optional. 
 
It’s basically a no-op change now that we don’t have optional collections anymore but we should still fix the bug.


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

bnbarham (approved):

kimdv:
@swift-ci please test Windows

kimdv (approved):

PR #2052, opened 2023-08-11
Improve compatibility layer for `KeyPathPropertyComponentSyntax`

ahoppen:
The rename of `identifier` -> `declName` was not correct in `KeyPathPropertyComponentSyntax` because we also chagned its type from `TokenSyntax` to `DeclReferenceExprSyntax`.

Instead, provide two hand-written compatibility properties that transparently access the `DeclReferenceExprSyntax`.


ahoppen:
@swift-ci Please test

bnbarham (approved):
IIUC there were a few that we changed to `declReferenceExpr` - does that apply to any of the others?

PR #2055, opened 2023-08-12
Make the first character of both `name` and `deprecatedName` in every `Child` lowercase

### LINKED ISSUE #1901: Change `name` property of `Child` in `CodeGeneration` to be lowercase

Matejkob:
Resolves #1901

### DIFF
--- a/CodeGeneration/Sources/SyntaxSupport/Child.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/Child.swift
@@ -214,7 +214,7 @@ public class Child {
     documentation: String? = nil,
     isOptional: Bool = false
   ) {
-    precondition(name.first?.isUppercase ?? true, "The first letter of a child’s name should be uppercase")
+    precondition(name.first?.isLowercase ?? true, "The first letter of a child’s name should be lowercase")
     precondition(deprecatedName?.first?.isUppercase ?? true, "The first letter of a child’s name should be uppercase")
     self.name = name
     self.deprecatedName = deprecatedName
--- a/CodeGeneration/Sources/SyntaxSupport/Child.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/Child.swift
@@ -215,7 +215,7 @@ public class Child {
     isOptional: Bool = false
   ) {
     precondition(name.first?.isLowercase ?? true, "The first letter of a child’s name should be lowercase")
-    precondition(deprecatedName?.first?.isUppercase ?? true, "The first letter of a child’s name should be uppercase")
+    precondition(deprecatedName?.first?.isLowercase ?? true, "The first letter of a child’s deprecatedName should be lowercase")
     self.name = name
     self.deprecatedName = deprecatedName
     self.kind = kind
--- a/CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
@@ -17,7 +17,7 @@ public let DECL_NODES: [Node] = [
     nameForDiagnostics: nil,
     children: [
       Child(
-        name: "Name",
+        name: "name",
         kind: .token(choices: [
           .token(.identifier),
           .token(.binaryOperator),
@@ -27,7 +27,7 @@ public let DECL_NODES: [Node] = [
         nameForDiagnostics: "name"
       ),
       Child(
-        name: "TrailingPeriod",
+        name: "trailingPeriod",
         deprecatedName: "TrailingDot",
         kind: .token(choices: [.token(.period)]),
         isOptional: true
@@ -52,24 +52,24 @@ public let DECL_NODES: [Node] = [
     ],
     children: [
       Child(
-        name: "LeftBrace",
+        name: "leftBrace",
         kind: .token(choices: [.token(.leftBrace)])
       ),
       Child(
-        name: "Accessors",
+        name: "accessors",
         kind: .nodeChoices(choices: [
           Child(
-            name: "Accessors",
+            name: "accessors",
             kind: .collection(kind: .accessorDeclList, collectionElementName: "Accessor")
           ),
           Child(
-            name: "Getter",
+            name: "getter",
             kind: .node(kind: .codeBlockItemList)
           ),
         ])
       ),
       Child(
-        name: "RightBrace",
+        name: "rightBrace",
         kind: .token(choices: [.token(.rightBrace)])
       ),
     ]
@@ -83,18 +83,18 @@ public let DECL_NODES: [Node] = [
     traits: ["WithAttributes"],
     children: [
       Child(
-        name: "Attributes",
+        name: "attributes",
         kind: .collection(kind: .attributeList, collectionElementName: "Attribute", defaultsToEmpty: true),
         nameForDiagnostics: "attributes"
       ),
       Child(
-        name: "Modifier",
+        name: "modifier",
         kind: .node(kind: .declModifier),
         nameForDiagnostics: "modifiers",
         isOptional: true
       ),
       Child(
-        name: "AccessorSpecifier",
+        name: "accessorSpecifier",
         deprecatedName: "AccessorKind",
         kind: .token(choices: [
           .keyword(text: "get"),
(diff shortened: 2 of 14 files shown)


CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
```diff
@@ -2006,11 +2006,11 @@ public let DECL_NODES: [Node] = [
...
     children: [
       Child(
-        name: "Statements",
+        name: "statements",
         kind: .collection(kind: .codeBlockItemList, collectionElementName: "Statement")
       ),
       Child(
-        name: "EndOfFileToken",
+        name: "endOfFileToken",
         deprecatedName: "EOFToken",
```
Matejkob, line 2014: Of course! Here's a revised version of your message for clarity and grammar:

I'm uncertain about the optimal way to handle this edge case.

If we want to make this `deprecatedName` lowercase as well, I'm not sure there's a better solution than simply hardcoding the condition `if deprecatedName == "eofToken" { ... ` somewhere 😕

CodeGeneration/Sources/SyntaxSupport/Traits.swift
Matejkob, line 1: I'm also wondering if we should change `Child`s names for `Traits`, but I actually don't see any reasons why not.

CodeGeneration/Sources/SyntaxSupport/Child.swift
```diff
@@ -214,8 +214,7 @@ public class Child {
     documentation: String? = nil,
     isOptional: Bool = false
   ) {
-    precondition(name.first?.isUppercase ?? true, "The first letter of a child’s name should be uppercase")
-    precondition(deprecatedName?.first?.isUppercase ?? true, "The first letter of a child’s name should be uppercase")
+    precondition(name.first?.isLowercase ?? true, "The first letter of a child’s name should be lowercase")
```
ahoppen, line 217: I think we should also precondition that `deprecatedName` starts with a lowercase letter.

ahoppen (commented):
Thank you @Matejkob . This is a lot more intuitive.

CodeGeneration/Sources/SyntaxSupport/Child.swift
Matejkob: I agree with you, but before selecting the exact implementation, I would like to discuss the best option. This is why I've added [this comment](https://github.com/apple/swift-syntax/pull/2055#discussion_r1292200947):

> I'm uncertain about the optimal way to handle this edge case.
>
> If we want to make `deprecatedName` lowercase as well, I'm not sure there's a better solution than simply hardcoding the condition `if deprecatedName == "eofToken" { ... ` somewhere 😕

Sorry if it wasn't understandable in the first place.

So, we are dealing with an edge case where `deprecatedName` is set to `"EOFToken"`, and I'm wondering what would be the best approach to address this situation. The best solution that comes to my mind is hardcoding the condition `if deprecatedName == "eofToken" { ... ` somewhere... :/

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
Matejkob: I wasn't suggesting this approach because I was thinking that we don't want to break the capability layer. However, if you're saying that we can do such changes, then I'm on board.

Here is the exact diff in the generated files:
```diff
diff --git a/Sources/SwiftSyntax/generated/RenamedChildrenCompatibility.swift b/Sources/SwiftSyntax/generated/RenamedChildrenCompatibility.swift
index 2df170c1..972b1377 100644
--- a/Sources/SwiftSyntax/generated/RenamedChildrenCompatibility.swift
+++ b/Sources/SwiftSyntax/generated/RenamedChildrenCompatibility.swift
@@ -7027,7 +7027,7 @@ extension SomeOrAnyTypeSyntax {

 extension SourceFileSyntax {
   @available(*, deprecated, renamed: "unexpectedBetweenStatementsAndEndOfFileToken")
-  public var unexpectedBetweenStatementsAndEOFToken: UnexpectedNodesSyntax? {
+  public var unexpectedBetweenStatementsAndEofToken: UnexpectedNodesSyntax? {
     get {
       return unexpectedBetweenStatementsAndEndOfFileToken
     }
@@ -7047,7 +7047,7 @@ extension SourceFileSyntax {
   }

   @available(*, deprecated, renamed: "unexpectedAfterEndOfFileToken")
-  public var unexpectedAfterEOFToken: UnexpectedNodesSyntax? {
+  public var unexpectedAfterEofToken: UnexpectedNodesSyntax? {
     get {
       return unexpectedAfterEndOfFileToken
     }
@@ -7062,9 +7062,9 @@ extension SourceFileSyntax {
       leadingTrivia: Trivia? = nil,
       _ unexpectedBeforeStatements: UnexpectedNodesSyntax? = nil,
       statements: CodeBlockItemListSyntax,
-      _ unexpectedBetweenStatementsAndEOFToken: UnexpectedNodesSyntax? = nil,
+      _ unexpectedBetweenStatementsAndEofToken: UnexpectedNodesSyntax? = nil,
       eofToken: TokenSyntax = .endOfFileToken(),
-      _ unexpectedAfterEOFToken: UnexpectedNodesSyntax? = nil,
+      _ unexpectedAfterEofToken: UnexpectedNodesSyntax? = nil,
       trailingTrivia: Trivia? = nil

   ) {
@@ -7072,9 +7072,9 @@ extension SourceFileSyntax {
         leadingTrivia: leadingTrivia,
         leadingTrivia: leadingTrivia,
         unexpectedBeforeStatements,
         statements: statements,
-        unexpectedBetweenStatementsAndEOFToken,
+        unexpectedBetweenStatementsAndEofToken,
         endOfFileToken: eofToken,
-        unexpectedAfterEOFToken,
+        unexpectedAfterEofToken,
         trailingTrivia: trailingTrivia
       )
   }
diff --git a/Sources/SwiftSyntaxBuilder/generated/RenamedChildrenBuilderCompatibility.swift b/Sources/SwiftSyntaxBuilder/generated/RenamedChildrenBuil
derCompatibility.swift
index ecaee9bc..53132ddb 100644
--- a/Sources/SwiftSyntaxBuilder/generated/RenamedChildrenBuilderCompatibility.swift
+++ b/Sources/SwiftSyntaxBuilder/generated/RenamedChildrenBuilderCompatibility.swift
@@ -751,9 +751,9 @@ extension SourceFileSyntax {
   public init(
       leadingTrivia: Trivia? = nil,
       unexpectedBeforeStatements: UnexpectedNodesSyntax? = nil,
-      unexpectedBetweenStatementsAndEOFToken: UnexpectedNodesSyntax? = nil,
+      unexpectedBetweenStatementsAndEofToken: UnexpectedNodesSyntax? = nil,
       eofToken: TokenSyntax = .endOfFileToken(),
-      unexpectedAfterEOFToken: UnexpectedNodesSyntax? = nil,
+      unexpectedAfterEofToken: UnexpectedNodesSyntax? = nil,
       @CodeBlockItemListBuilder statementsBuilder: () throws -> CodeBlockItemListSyntax,
       trailingTrivia: Trivia? = nil
     ) rethrows {
@@ -761,9 +761,9 @@ extension SourceFileSyntax {
         leadingTrivia: leadingTrivia,
         unexpectedBeforeStatements,
         statements: statementsBuilder(),
-        unexpectedBetweenStatementsAndEOFToken,
+        unexpectedBetweenStatementsAndEofToken,
         endOfFileToken: eofToken,
-        unexpectedAfterEOFToken,
+        unexpectedAfterEofToken,
         trailingTrivia: trailingTrivia
       )
   }
```
Is `EofToken` acceptable?

ahoppen: Yes, that’s acceptable.

ahoppen (approved):

ahoppen:
@Matejkob There is a merge conflict. Could you rebase your PR?

Matejkob:
@ahoppen I've just solved the merge conflict. Thank you.

ahoppen:
Let’s try to quickly get it in before there’s another conflict.

@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #2062, opened 2023-08-15
Update unit tests after the removal of the `QualifiedDeclNameSyntax`

Matejkob:
After removing the `QualifiedDeclNameSyntax` #2045, we should also remove this case from unit tests. 

The unit tests for the `CodeGeneration` package are not building on the main branch.


kimdv (approved):

kimdv:
@swift-ci please test

kimdv:
Thanks @Matejkob 💪

PR #2064, opened 2023-08-15
Remove the remaining file in `lit-test-helper`

StevenWong12:
That's definitely something we forgot before.


kimdv:
@swift-ci please test

ahoppen (approved):

PR #2069, opened 2023-08-15
Improve swift-syntax-dev-utils 

ahoppen:
- Improve error message if a subprocess finishes with a non-zero exit code
- Don’t capture stdout and stderr for `swift build` and `swift test` commands
  - We want the output of those processes to show up in console, so we shouldn’t capture it


kimdv (approved):
Nice!

PR #2071, opened 2023-08-15
Add messages for some `assertStringsEqualWithDiff`

ahoppen:
The motivation was that it’s ambiguous what’s going wrong `assertParse` fails because applying all Fix-Its does not produce the fixed source.

CC @gottesmm


ahoppen:
@swift-ci Please test

bnbarham (approved):

kimdv (approved):

Issue #1977, opened 2023-08-02
Reduce number of `raw:` interpolations in CodeGeneration


ahoppen:
CodeGeneration still uses many `raw:` interpolations, which should be avoided unless it’s really necessary. Some of these `raw:` interpolations don’t need to the `raw:` label anymore. For some others, we could restructure how nodes are represented in SyntaxSupport to use `TokenSyntax` instead of `String`, similar to what I’ve done in https://github.com/apple/swift-syntax/pull/1914.

ahoppen:
Tracked in Apple’s issue tracker as rdar://113237456

Matejkob:
I've grabbed the lowest-hanging fruit and and removed all the no-longer-needed `raw:` interpolations. Further reductions in the use of `raw:` will require changing types.

Matejkob:
So actually, if I've already started doing things related to this issue, I can do the whole thing.

Matejkob:
We've already eliminated most of the `raw:` interpolations in CodeGeneration. I have a few more potential places in mind where we could eliminate them, but some of them seem to be incorrect. I would love to hear your feedback @ahoppen @kimdv 

- `Node.getter:parserFunction` - I'm thinking of changing the type to TokenSyntax. If I'm not mistaken, the function identifier is of this type, right?
- `let optionalMark = child.isOptional ? "?" : ""` into `let optionalMark: TokenSyntax = child.isOptional ? TokenSyntax.postfixQuestionMarkToken() : TokenSyntax("")`. The second branch of this condition seems to be incorrect. `TokenSyntax = ""` doesn't feel right, does it?
- `let iuoMark = child.isOptional ? "" : "!"` into `let iuoMark: TokenSyntax = child.isOptional ? TokenSyntax("") : TokenSyntax.exclamationMarkToken()`.

Let me know if there are any issues with these changes or if you can identify other potential usages of `raw:` that could be eliminated but I've missed them.

ahoppen:
> * `Node.getter:parserFunction` - I'm thinking of changing the type to TokenSyntax. If I'm not mistaken, the function identifier is of this type, right?

Yes, that seems like a good change to me

> * `let optionalMark = child.isOptional ? "?" : ""` into `let optionalMark: TokenSyntax = child.isOptional ? TokenSyntax.postfixQuestionMarkToken() : TokenSyntax("")`. The second branch of this condition seems to be incorrect. `TokenSyntax = ""` doesn't feel right, does it?
[...]

PR #2076, opened 2023-08-16
Improve modeling of `KeywordSpec` in CodeGeneration

Matejkob:
Resolves part of #1977

### DIFF
--- a/CodeGeneration/Sources/SyntaxSupport/KeywordSpec.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/KeywordSpec.swift
@@ -20,8 +20,6 @@ public struct KeywordSpec {
   public var isExperimental: Bool
 
   public var isLexerClassified: Bool
-  public var requiresLeadingSpace: Bool
-  public var requiresTrailingSpace: Bool
 
   public var escapedName: String {
     if isLexerClassified || name == "Type" || name == "Protocol" {
@@ -40,12 +38,10 @@ public struct KeywordSpec {
 
   /// `isLexerClassified` determines whether the token kind is switched from being an identifier to a keyword in the lexer.
   /// This is true for keywords that used to be considered non-contextual.
-  init(_ name: String, isExperimental: Bool = false, isLexerClassified: Bool = false, requiresLeadingSpace: Bool = false, requiresTrailingSpace: Bool = false) {
+  init(_ name: String, isExperimental: Bool = false, isLexerClassified: Bool = false) {
     self.name = name
     self.isExperimental = isExperimental
     self.isLexerClassified = isLexerClassified
-    self.requiresLeadingSpace = requiresLeadingSpace
-    self.requiresTrailingSpace = requiresTrailingSpace
   }
 }
 
@@ -375,17 +371,17 @@ public enum Keyword: CaseIterable {
     case .any:
       return KeywordSpec("any")
     case .Any:
-      return KeywordSpec("Any", isLexerClassified: true, requiresTrailingSpace: true)
+      return KeywordSpec("Any", isLexerClassified: true)
     case .as:
-      return KeywordSpec("as", isLexerClassified: true, requiresTrailingSpace: true)
+      return KeywordSpec("as", isLexerClassified: true)
     case .assignment:
       return KeywordSpec("assignment")
     case .associatedtype:
-      return KeywordSpec("associatedtype", isLexerClassified: true, requiresTrailingSpace: true)
+      return KeywordSpec("associatedtype", isLexerClassified: true)
     case .associativity:
       return KeywordSpec("associativity")
     case .async:
-      return KeywordSpec("async", requiresTrailingSpace: true)
+      return KeywordSpec("async")
     case .attached:
       return KeywordSpec("attached")
     case .autoclosure:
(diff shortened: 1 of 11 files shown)


CodeGeneration/Sources/SyntaxSupport/KeywordSpec.swift
```diff
@@ -13,34 +13,69 @@
...
-  public var escapedName: String {
+  /// Indicates if the keyword requires a leading space.
+  ///
+  /// If `true`, the keyword should have a space character before it when used in Swift code.
+  public let requiresLeadingSpace: Bool
+
+  /// Indicates if the keyword requires a trailing space.
+  ///
+  /// If `true`, the keyword should have a space character after it when used in Swift code.
+  public let requiresTrailingSpace: Bool
```
ahoppen, line 36: I think we no longer use `requiresLeadingSpace` and `requiresTrailingSpace` to generate any code. You should be able to just remove them. You can also do it in a follow-up PR if you want.

CodeGeneration/Sources/SyntaxSupport/KeywordSpec.swift
```diff
@@ -13,34 +13,69 @@
...
   }
 
-  /// Retrieve the attributes that should be printed on any API for the
-  /// generated keyword.
+  /// The name of the keyword enclosed in double quotes.
+  ///
+  /// This property is used in code generation where a string representation of the keyword is needed.
+  public var quotedName: TokenSyntax {
+    "\"\(raw: name)\""
+  }
```
ahoppen, line 54: This isn’t correct because you’re returning a string literal here, which, in SwiftSyntax, is represented by a [StringLiteralExprSyntax](https://swiftpackageindex.com/apple/swift-syntax/main/documentation/swiftsyntax/stringliteralexprsyntax) node, and not a single token. What I would suggest is that in the code that access is, you do `\(literal: keyword.name)`, which would also add pounds in case the name would contain `"` or anything else that needed escaping (not that I would ever expect that from a keyword name but it’s good to be safe).

CodeGeneration/Sources/SyntaxSupport/KeywordSpec.swift
```diff
@@ -13,34 +13,69 @@
...
+  ///
+  /// This is useful when the keyword is also an identifier or has special characters that need escaping.
+  public var escapedName: TokenSyntax {
     if isLexerClassified || name == "Type" || name == "Protocol" {
-      return "`\(name)`"
+      return "`\(raw: name)`"
     } else {
-      return name
+      return "\(raw: name)"
     }
```
ahoppen, line 46: I think in a lot of cases where we currently add backticks in the generated code, they aren’t actually needed. For example ``self = .`inout` `` doesn’t require backticks. What do you think about adding the following and then using it everywhere after a `.`. And where you are declaring the cases, you use `varOrCaseName.backtickedIfNeeded`.

```swift
public var varOrCaseName: TokenSynax {
  if name == "init" {
    return "`init`"
  } else {
    return TokenSyntax.identifier(name)
  }
}
```

ahoppen (commented):
Thank you for improving this area of CodeGeneration as well. I’ve got a couple suggestions of additional improvements we can make.

CodeGeneration/Sources/SyntaxSupport/KeywordSpec.swift
Matejkob: Oh right! My bad.

> What I would suggest is that in the code that access is, you do `\(literal: keyword.name)`, which would also add pounds in case the name would contain `"` or anything else that needed escaping (not that I would ever expect that from a keyword name but it’s good to be safe).

But it would change generated source code, right? Do we mind such a change?

CodeGeneration/Sources/SyntaxSupport/KeywordSpec.swift
Matejkob: I've removed both properties in one of the commits in this PR. Thanks for catching that!

CodeGeneration/Sources/SyntaxSupport/KeywordSpec.swift
Matejkob: Oh right! My bad! Sorry for that. I've removed this property and I've used `literal:` as you suggested. Great idea, thanks so much!

CodeGeneration/Sources/SyntaxSupport/KeywordSpec.swift
Matejkob: I love it! I've already introduced the changes. I've had to extend the condition in `backtickedIfNeeded` to be able to cover `Type` and `Protocol` cases.

CodeGeneration/Sources/SyntaxSupport/KeywordSpec.swift
Matejkob: And by the way @ahoppen do you prefer splitting commits like I've done there or all changes should be in one commit?

ahoppen (approved):
Thanks 👍🏽

CodeGeneration/Sources/SyntaxSupport/KeywordSpec.swift
ahoppen: I think if they are more or less independent changes, having them in separate commits is good 👍🏽

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

kimdv:
@swift-ci Please test Windows

PR #2078, opened 2023-08-16
Add arena to syntax nodes

### LINKED ISSUE #2036: Improve SyntaxRewriter performance for operator folding
Running operator folding on MovieSwiftUI takes as long as parsing. That really shouldn’t be the case. Running with Instruments shows that we spend the majority of time in memory-related operations.

The first steps to take would be:
- Add an `arena` parameter to all the syntax node initializers 
- Add an `arena` parameter to `SyntaxRewriter`
- Construct all nodes in `SyntaxRewriter` and operator folding in that arena
- Investigate where else `SyntaxRewriter` is causing memory allocations and try to eliminate them

Attached is a WIP diff where I started to eliminate some of the memory allocations by `SyntaxRewriter` but I stopped once I realized that we’d need to modify the initializers of all syntax nodes. 

[reduce-arena-memory-allocations.txt](https://github.com/apple/swift-syntax/files/12297749/reduce-arena-memory-allocations.txt)

kimdv:
Resolves #2036

Open this PR to start discussion and getting help when needed.

### DIFF
--- a/CodeGeneration/Sources/generate-swift-syntax/LayoutNode+Extensions.swift
+++ b/CodeGeneration/Sources/generate-swift-syntax/LayoutNode+Extensions.swift
@@ -66,6 +66,9 @@ extension LayoutNode {
 
       FunctionParameterSyntax("trailingTrivia: Trivia? = nil")
         .with(\.leadingTrivia, .newline)
+
+      FunctionParameterSyntax("arena: __shared SyntaxArena = SyntaxArena()")
+        .with(\.leadingTrivia, .newline)
     }
 
     return """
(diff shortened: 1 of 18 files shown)


ahoppen (commented):
This looks good to me but I’d like to see a whole PR that improves the performance of `SyntaxRewriter` before merging any of the piece PRs, just to make sure we’re going in the right direction.

kimdv:
> This looks good to me but I’d like to see a whole PR that improves the performance of `SyntaxRewriter` before merging any of the piece PRs, just to make sure we’re going in the right direction.

This not done yet.
But I wanted to open the PR so I can get help if needed.
It's easier to explain here on GitHub 😁

ahoppen:
Yeah, I totally agree. Just wanted to comment that we shouldn’t merge this as-is.

PR #2085, opened 2023-08-20
Add `appendInterpolation` with optional syntax nodes

Matejkob:
This PR introduces a new method, `appendInterpolation`, to the `SyntaxStringInterpolation` struct, which takes an optional syntax node parameter. If the provided node is nil, no interpolation is added. Otherwise, interpolation is performed as usual.

Additionally, this new method has been integrated into the CodeGeneration package to eliminate the usage of interpolation.

Resolves part of #1977

### DIFF
--- a/Sources/SwiftSyntaxBuilder/Syntax+StringInterpolation.swift
+++ b/Sources/SwiftSyntaxBuilder/Syntax+StringInterpolation.swift
@@ -63,6 +63,12 @@ extension SyntaxStringInterpolation: StringInterpolationProtocol {
   }
 
   /// Append a syntax node to the interpolation.
+  ///
+  /// This method accepts a syntax node and appends it to the interpolation.
+  /// If there was a previous indentation value, the method will indent the
+  /// syntax node with that value. If not, it will use the syntax node as-is.
+  ///
+  /// - Parameter node: A syntax node that conforms to `SyntaxProtocol`.
   public mutating func appendInterpolation<Node: SyntaxProtocol>(
     _ node: Node
   ) {
@@ -82,6 +88,20 @@ extension SyntaxStringInterpolation: StringInterpolationProtocol {
       )
     )
     self.lastIndentation = nil
+  }
+
+  /// Append an optional syntax node to the interpolation.
+  ///
+  /// This method accepts an optional syntax node and appends it to the interpolation
+  /// if it exists. If the syntax node is nil, this method does nothing.
+  ///
+  /// - Parameter node: An optional syntax node that conforms to `SyntaxProtocol`.
+  public mutating func appendInterpolation<Node: SyntaxProtocol>(
+    _ node: Node?
+  ) {
+    if let node {
+      appendInterpolation(node)
+    }
   }
 
   public mutating func appendInterpolation<T>(raw value: T) {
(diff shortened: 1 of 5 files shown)


Sources/SwiftSyntaxBuilder/Syntax+StringInterpolation.swift
```diff
@@ -84,6 +90,20 @@ extension SyntaxStringInterpolation: StringInterpolationProtocol {
...
+  /// Append an optional syntax node to the interpolation.
+  ///
+  /// This method accepts an optional syntax node and appends it to the interpolation
+  /// if it exists. If the syntax node is nil, this method does nothing.
+  ///
+  /// - Parameter node: An optional syntax node that conforms to `SyntaxProtocol`.
+  public mutating func appendInterpolation<Node: SyntaxProtocol>(
+    _ node: Node?
+  ) {
+    if let node {
```
kimdv, line 102: As we need to support older versions of Swift, this shorthand syntax is not allowed.

kimdv (changes requested):

Sources/SwiftSyntaxBuilder/Syntax+StringInterpolation.swift
Matejkob: Yeah, at first I thought the same, but then I noticed that such syntax has been used in the method above: https://github.com/apple/swift-syntax/blob/39a42395dfa8a2890374bf4734d8f0068919832f/Sources/SwiftSyntaxBuilder/Syntax%2BStringInterpolation.swift#L71. 
How is that? O.o

kimdv: Oh that must have been a mis! Nice catch.

Would you mind also change that? 😁

kimdv: @ahoppen do we have SwiftSyntax included in some compatibility test suite?

Matejkob: > Oh that must have been a mis! Nice catch.
> 
> Would you mind also change that? 😁

Such syntax is used in 49 other places in the SwiftSyntax repo 😅

Matejkob: > @ahoppen do we have SwiftSyntax included in some compatibility test suite?

The Swift Package Index is functioning as a sort of compatibility test suite. Based on the [swift-syntax profile](https://swiftpackageindex.com/apple/swift-syntax), we no longer support 5.6 starting from version 5.9.

> We require that swift-syntax builds with the latest released compiler and the previous major version (e.g. with Swift 5.8 and Swift 5.7).

I believe that the lowest version we should support on the `main` branch now is 5.7, in which the shorthand syntax is already available.

kimdv: Hmm.. Strange 🤔
I'm pretty sure we use current version - 2. So that means 5.8, 5.7 and 5.6.

We need Alex to give an answer here :D

kimdv: > > @ahoppen do we have SwiftSyntax included in some compatibility test suite?
> 
> The Swift Package Index is functioning as a sort of compatibility test suite. Based on the [swift-syntax profile](https://swiftpackageindex.com/apple/swift-syntax), we no longer support 5.6 starting from version 5.9.
> 
> > [We require that swift-syntax builds with the latest released compiler and the previous major version (e.g. with Swift 5.8 and Swift 5.7).](https://github.com/apple/swift-syntax/blob/main/CONTRIBUTING.md#swift-version)
> 
> I believe that the lowest version we should support on the `main` branch now is 5.7, in which the shorthand syntax is already available.

Oh yes, that is true, 
Look away from this comment and sorry for the confusion. Will resolve.

kimdv (approved):

Sources/SwiftSyntaxBuilder/Syntax+StringInterpolation.swift
Matejkob: No worries! As I've mention I've also been a little confused at first 😅

ahoppen (approved):
Thanks 👍🏽

PR #2086, opened 2023-08-21
Remove all remaining use of `raw:` interpolations

### LINKED ISSUE #1977: Reduce number of `raw:` interpolations in CodeGeneration

Matejkob:
Finally resolves #1977


kimdv:
@swift-ci please test

kimdv (approved):

ahoppen (approved):
🙌🏽

PR #2087, opened 2023-08-22
Refactor `SyntaxClassifier` with `SyntaxVisitor`

StevenWong12:
A very early stage of the refactoring of `SyntaxClassifier`. Refactoring `SyntaxClassifier` with `SyntaxVisitor` makes it easier to maintain.

We do not pass all the tests in `ClassificationTests` yet and we also have to improve the performance of `SyntaxVisitor` to make this pr ready to be merged.

- [ ] functionality
- [ ] performance


ahoppen:
Thanks for setting up the PR. Let’s use this as a starting point to investigate the performance of `SyntaxVisitor`. I filed  https://github.com/apple/swift-syntax/issues/2091 for that.

PR #2088, opened 2023-08-22
Fix a formatting issue if a member attribute macro is applied to properties that already have an attribute

ahoppen:
If a variable already had a macro already had an attribute and we add a new attribute to it via a member macro, the produced code was invalid. For example,

```swift
@wrapStoredProperties
struct Foo {
  @available(*, deprecated) var x: Int
}
```

produced

```swift
struct Foo {
  @available(*, deprecated)
  @Wrappervar x: Int
}
```

which is invalid because there is no space between `Wrapper` and `var`.

To fix this, transfer the trailing trivia from the last attribute to the last newly inserted attribute.

This way, we essentially insert the new attributes right after the last attribute in source, but before its trailing trivia, keeping the trivia that separates the attribute block from the variable itself.

```swift
struct Foo {
  @available(*, deprecated) 
  @Wrapper var x: Int
}
```

rdar://114275860

### DIFF
--- a/Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
+++ b/Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
@@ -573,14 +573,21 @@ private class MacroApplication<Context: MacroExpansionContext>: SyntaxRewriter {
       // Expand member attribute members attached to the declaration context.
       // Note that MemberAttribute macros are _not_ applied to generated members
       if let parentDeclGroup, let decl = item.decl.asProtocol(WithAttributesSyntax.self) {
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

bnbarham (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Linux

ahoppen:
@swift-ci Please test Linux

ahoppen:
@swift-ci Please test Linux

PR #2090, opened 2023-08-23
[CMake] Add padding to RUNPATH in linux

rintaro:
Shared libraries RUNPATH will be updated when they are copied to swift build directory. CMake's 'file(SET_RPATH ...)' requires RUNPATH field have enough lengh before updating.

### DIFF
--- a/cmake/modules/AddSwiftHostLibrary.cmake
+++ b/cmake/modules/AddSwiftHostLibrary.cmake
@@ -71,10 +71,16 @@ function(add_swift_host_library name)
   )
 
   get_target_property(lib_type ${name} TYPE)
-  if(lib_type STREQUAL SHARED_LIBRARY AND CMAKE_SYSTEM_NAME STREQUAL Darwin)
-    # Allow install_name_tool to update paths (for rdar://109473564)
-    set_property(TARGET ${name} APPEND_STRING PROPERTY
-                 LINK_FLAGS " -Xlinker -headerpad_max_install_names")
+  if(lib_type STREQUAL SHARED_LIBRARY)
+    if (CMAKE_SYSTEM_NAME STREQUAL Darwin)
+      # Allow install_name_tool to update paths (for rdar://109473564)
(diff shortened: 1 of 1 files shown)


rintaro:
@swift-ci Please test

cmake/modules/AddSwiftHostLibrary.cmake
```diff
@@ -71,10 +71,16 @@ function(add_swift_host_library name)
...
-                 LINK_FLAGS " -Xlinker -headerpad_max_install_names")
+  if(lib_type STREQUAL SHARED_LIBRARY)
+    if (CMAKE_SYSTEM_NAME STREQUAL Darwin)
+      # Allow install_name_tool to update paths (for rdar://109473564)
+      set_property(TARGET ${name} APPEND_STRING PROPERTY
+                   LINK_FLAGS " -Xlinker -headerpad_max_install_names")
+    elseif (CMAKE_SYSTEM_NAME STREQUAL Linux)
+      # Make some room to update paths.
+      set_property(TARGET ${name} APPEND PROPERTY
+                   INSTALL_RPATH ":::::::::::::::::::::::::::::::::::::::::::::::::::::::")
```
bnbarham, line 82: Is this enough to cover all the platforms we know of?

bnbarham (approved):

cmake/modules/AddSwiftHostLibrary.cmake
bnbarham: And also, could this be at least `$ORIGIN::::::::::...` so that a swift-syntax build works by itself?

rintaro: It's will be changed to `$ORIGIN:$ORIGIN/../{platform name}`. Yes, I don't believe any platform name would be longer than 45 characters.

rintaro: 🤔 I thought I made it `:` x 64 but this is actually 55. Anyway 36 characters for platform name should be enough :)

PR #2099, opened 2023-08-24
Set `-derivedDataPath` when building the EditorExtension Xcode project

ahoppen:
Previously, we were sometimes seeing a `build` directory being created in the `swift-syntax` sources directory, which contained dummy input files that didn’t have a trailing newline, which caused swift-format to complain about their formatting.

And while at it, also make sure that we clean the temporary directory that we create to verify the generated files.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #2100, opened 2023-08-24
Don’t mutate line endings in `assertParseWithAllNewlineEndings`

ahoppen:
My current suspicion is that this is causing swift-syntax CI failures, probably caused by https://github.com/apple/swift/pull/67729. While I investigate what the root cause is, just run a normal `assertParse` here to fix CI.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #2103, opened 2023-08-24
Small refactoring improvements around `ExprFlavor`

ahoppen:
- Unify `forDirective` with `ExprFlavor`, getting rid of one boolean parameter in expression parsing
- Rename `ExprFlavor.basic` -> `stmtCondition` because this case is used primarily in statement conditions that need to disambiguate trailing closures from the statement’s body
- Rename `ExprFlavor.trailingClosure` -> `basic` because this is the base case that we use inside normal expression parsing
- Add argument labels to `flavor` to match the `pattern` parameter that also has a label and make `parseExpression` parameters required so that each caller needs to make a conscious choice about the values and so it’s easier to search for where each of the flavors is being used.


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

kimdv (approved):

bnbarham (approved):

PR #2104, opened 2023-08-24
Add test cases for parsing closures inside `if` statement conditions

### LINKED ISSUE #795: SwiftParser allows closures in condition lists where the compiler doesn't
### Issue Kind

Bad Diagnostic Produced

### Source Code

```swift
if {x}() {}
if x.map{$0} {}
```

### Description

While I would _love_ to have these examples permitted, Swift forbids them but they currently succeed in the new Swift parser without diagnostics and produce the following valid parse tree:

```
SourceFileSyntax children=2
  0: CodeBlockItemListSyntax children=2        // if {x}() {}
      0: IfStmtSyntax children=3
        0: ifKeyword
        1: ConditionElementListSyntax children=1
          0: ConditionElementSyntax children=1
            0: FunctionCallExprSyntax children=4
              0: ClosureExprSyntax children=3
                0: leftBrace
                1: CodeBlockItemListSyntax children=1
                      0: identifier("x")
                2: rightBrace
              1: leftParen
              2: TupleExprElementListSyntax
              3: rightParen
        2: CodeBlockSyntax children=3
          0: leftBrace
          1: CodeBlockItemListSyntax
          2: rightBrace
    1: CodeBlockItemSyntax children=1        // if x.map{$0} {}
      0: IfStmtSyntax children=3
        0: ifKeyword
        1: ConditionElementListSyntax children=1
          0: ConditionElementSyntax children=1
            0: FunctionCallExprSyntax children=3
              0: MemberAccessExprSyntax children=3
                  0: identifier("x")
                1: period
                2: identifier("map")
              1: TupleExprElementListSyntax
              2: ClosureExprSyntax children=3
                0: leftBrace
                1: CodeBlockItemListSyntax children=1
                      0: dollarIdentifier("$0")
                2: rightBrace
        2: CodeBlockSyntax children=3
          0: leftBrace
          1: CodeBlockItemListSyntax
          2: rightBrace
  1: eof
```

ahoppen:
Add a bunch of test cases that verify that SwiftParser accepts closures inside `if` conditions if and only if the C++ parser does.

The only difference is that the C++ parser emits a warning in cases like `if test { $0 } {}`: `Trailing closure in this context is confusable with the body of the statement; pass as a parenthesized argument to silence this warning`. We currently don’t have any infrastructure to emit warnings from SwiftParser and I’m not sure if this warning is really necessary. NFC on whether we need it.

While writing the test cases, I found an issue in `BasicFormat` that adds a newline after the opening `{` of the closure inside these conditions, which causes the closure to get parsed as the statement’s body. Fixed that as well.

Fixes #795
rdar://99948919

### DIFF
--- a/Sources/SwiftBasicFormat/BasicFormat.swift
+++ b/Sources/SwiftBasicFormat/BasicFormat.swift
@@ -123,14 +123,7 @@ open class BasicFormat: SyntaxRewriter {
   // MARK: - Helper functions
 
   private func isInsideStringInterpolation(_ token: TokenSyntax) -> Bool {
-    var ancestor: Syntax = Syntax(token)
-    while let parent = ancestor.parent {
-      ancestor = parent
-      if ancestor.is(ExpressionSegmentSyntax.self) {
-        return true
-      }
-    }
-    return false
+    return token.ancestorOrSelf { $0.as(ExpressionSegmentSyntax.self) } != nil
   }
 
   private func childrenSeparatedByNewline(_ node: Syntax) -> Bool {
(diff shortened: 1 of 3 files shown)


ahoppen:
@swift-ci Please test

kimdv:
@swift-ci Please test

Sources/SwiftBasicFormat/BasicFormat.swift
```diff
@@ -607,3 +627,17 @@ fileprivate extension TokenSyntax {
     }
   }
 }
+
+fileprivate extension SyntaxProtocol {
+  /// Returns this node or the first ancestor that satisfies `condition`.
```
bnbarham, line 632: We have in SwiftParserDiagnostics and here now. Worth just adding in general?

Tests/SwiftParserTest/StatementTests.swift
```diff
@@ -728,4 +728,171 @@ final class StatementTests: ParserTestCase {
...
+      if true, { a in
+        x + a
+      }(1) {}
+      """
+    )
+  }
+
+  func testTrailingClosureInGuard() {
+    assertParse(
+      "guard test 1️⃣{ $0 } 2️⃣else {}",
```
bnbarham, line 839: We differ for `guard`?

ahoppen: Oh, yes we do. the trailing closure is not really ambiguous in `guard` statements so we … disallow it there. 🤷🏽 

I’m happy to open a conversation about this behavior independently of this issue.

Sources/SwiftBasicFormat/BasicFormat.swift
ahoppen: I don’t feel happy enough with this API to make it public yet. So I prefer to just have two copies of it for now.

bnbarham (approved):

PR #2106, opened 2023-08-25
Fix newline detection logic for carriage return line feeds

ahoppen:
`\r\n` is considered a single character in Unicode and we thus need to check for it specifically and checking for `\n` on its own is not sufficient.

Also reverts #2100 since the tests are passing again now.

rdar://114406758


ahoppen:
@swift-ci Please test

bnbarham (approved):

Issue #2092, opened 2023-08-23
Prevent unrelated casts using `as`


ahoppen:
Currently, `SyntaxProtocol` has an `as` method that allows casting to any other syntax node type. Because of this, it’s possible to e.g. cast a `FunctionDeclSyntax` to an `IdentifierExprSyntax` without any compiler errors or warnings, even though we know that the cast will always fail.

What we should do instead, is to only have a `as(_: ExprSyntaxProtocol.Type)` function on `ExprSyntax` (analogous for `StmtSyntax` and the other base nodes).

The `as` function on `SyntaxProtocol` can then be marked as deprecated and produce a warning that the cast will always fail.

`is` and `cast` will need to be updated accordingly.

ahoppen:
Tracked in Apple’s issue tracker as rdar://114330180

Matejkob:
What does the term "as well as `as(_: SyntaxProtocol.Type)`" mean? Does it suggest that every base node should have this method? Would this be equivalent to having the `as` method directly on `SyntaxProtocol`?

ahoppen:
Oh, that was an artifact of a sentence I started writing but then didn’t finish. Ignore it – and I removed it from the description

grynspan:
If I have a `Syntax`, how do I then cast it to something else?

Matejkob:
> If I have a `Syntax`, how do I then cast it to something else?

I believe we should keep this part of the casting API. If you find a while you could review my pull request #2108, which contains a draft proposal on casting constraints?

ahoppen:
> If I have a `Syntax`, how do I then cast it to something else?

The proposal here is to remove the `as` methods on all types that are *not* base nodes, e.g. `FunctionDeclSyntax.as`. `Syntax.as` will continue to exist.

grynspan:
Okay, so a `Syntax` can still be cast, but an arbitrary `some SyntaxProtocol` can't be? And if I find myself needing to cast in that case, I can do `Syntax(foo).as(...)`?

ahoppen:
@Matejkob also noted that in https://github.com/apple/swift-syntax/pull/2108. The only case where you won’t be able to cast is on anything that’s statically known to be a leaf syntax node type. Casting will still be possible on `some SyntaxProtocol`, `some DeclSyntaxProtocol`, `any SyntaxProtocol` etc.

grynspan:
Oh, I get it. Sorry, I misunderstood what was being described here. I'll be quiet. :)

[...]

PR #2108, opened 2023-08-26
Introduce leaf protocols to prevent leaf nodes from being casted 

Matejkob:
To-Do after concept approval:
- [x] Fix all warnings related to API changes across the entire repo
- [x] Add documentation for Code Generation and the new API

I'm opening this PR as a draft to initiate a discussion about the proposed changes.

In issue #2092, @ahoppen mentioned:
> What we should do instead, is to only have a `as(_: ExprSyntaxProtocol.Type)` function on `ExprSyntax` (analogous for `StmtSyntax` and the other base nodes).
>
> The `as` function on `SyntaxProtocol` can then be marked as deprecated and produce a warning that the cast will always fail.

Best to my understanding, we cannot remove the `as` method from `SyntaxProtocol`. This function is necessary due to entities like `Syntax` or `ExprSyntax`. Without these entities, we could use the `is`/`as` Swift keywords. To illustrate, consider the example below:
```swift
func visit(node: some SyntaxProtocol) {
  if node is ClassExprSyntax { ... }
}
```
The following argument can be passed to this function:
```swift
let node = Syntax(ClassExprSyntax("class Foo {}"))

visit(node)
```
In this scenario, using the `is` Swift keyword won't correctly identify that the argument is actually of type `ClassExprSyntax`.

Consequently, I've tried to enhance the downcasting API in a slightly more intricate manner by adding casting methods for lower types.

If any decisions are unclear, please let me know. I'm also eager to hear your general feedback!

This PR resolves part of #2092.

### DIFF
--- a/CodeGeneration/Sources/SyntaxSupport/SyntaxNodeKind.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/SyntaxNodeKind.swift
@@ -368,6 +368,19 @@ public enum SyntaxNodeKind: String, CaseIterable {
     }
   }
 
+  /// For base node types, generates the name of the protocol to which all
+  /// concrete leaf nodes that derive from this base kind should conform.
+  ///
+  /// - Warning: This property can only be accessed for base node kinds; attempting to
+  /// access it for a non-base kind will result in a runtime error.
+  public var leafProtocolType: TypeSyntax {
+    if isBase {
+      return "_Leaf\(syntaxType)NodeProtocol"
+    } else {
+      fatalError("Only base kind can define leaf protocol")
+    }
+  }
+
   /// If the syntax kind has been renamed, the previous raw value that is now
   /// deprecated.
   public var deprecatedRawValue: String? {
--- a/Release Notes/510.md
+++ b/Release Notes/510.md
@@ -25,6 +25,11 @@
 
 ## Deprecations
 
+- Leaf Node Casts
+  - Description: Syntax nodes that do not act as base nodes for other syntax types have the casting methods marked as deprecated. This prevents unsafe type-casting by issuing deprecation warnings for methods that will always result in failed casts.
+  - Issue: https://github.com/apple/swift-syntax/issues/2092
+  - Pull Request: https://github.com/apple/swift-syntax/pull/2108
+  
 ## API-Incompatible Changes
 
 
--- a/Release Notes/510.md
+++ b/Release Notes/510.md
@@ -30,6 +30,11 @@
   - Issue: https://github.com/apple/swift-syntax/issues/2092
   - Pull Request: https://github.com/apple/swift-syntax/pull/2108
   
+- Same-Type Casts 
+  - Description: `is`, `as`, and `cast` overloads on `SyntaxProtocol` with same-type conversions are marked as deprecated. The deprecated methods emit a warning indicating the cast will always succeed.
+  - Issue: https://github.com/apple/swift-syntax/issues/2092
+  - Pull Request: https://github.com/apple/swift-syntax/pull/2108
+  
 ## API-Incompatible Changes
 
 
--- a/Release Notes/510.md
+++ b/Release Notes/510.md
@@ -29,12 +29,17 @@
   - Description: Syntax nodes that do not act as base nodes for other syntax types have the casting methods marked as deprecated. This prevents unsafe type-casting by issuing deprecation warnings for methods that will always result in failed casts.
   - Issue: https://github.com/apple/swift-syntax/issues/2092
   - Pull Request: https://github.com/apple/swift-syntax/pull/2108
-  
+
 - Same-Type Casts 
   - Description: `is`, `as`, and `cast` overloads on `SyntaxProtocol` with same-type conversions are marked as deprecated. The deprecated methods emit a warning indicating the cast will always succeed.
   - Issue: https://github.com/apple/swift-syntax/issues/2092
   - Pull Request: https://github.com/apple/swift-syntax/pull/2108
-  
+
+- Base Node Casts
+  - Description: `is`, `as`, and `cast` methods on base node protocols with base-type conversions are marked as deprecated. The deprecated methods emit a warning that informs the developer that the cast will always succeed and should be done using the base node's initializer.
+  - Issue: https://github.com/apple/swift-syntax/issues/2092
+  - Pull Request: https://github.com/apple/swift-syntax/pull/2108
+
 ## API-Incompatible Changes
 
 
--- a/Sources/SwiftSyntax/generated/syntaxNodes/SyntaxNodesAB.swift
+++ b/Sources/SwiftSyntax/generated/syntaxNodes/SyntaxNodesAB.swift
@@ -24,7 +24,7 @@
 /// 
 ///  - ``PatternBindingSyntax``.``PatternBindingSyntax/accessorBlock``
 ///  - ``SubscriptDeclSyntax``.``SubscriptDeclSyntax/accessorBlock``
-public struct AccessorBlockSyntax: SyntaxProtocol, SyntaxHashable {
+public struct AccessorBlockSyntax: SyntaxProtocol, SyntaxHashable, _LeafSyntaxNodeProtocol {
   public enum Accessors: SyntaxChildChoices, SyntaxHashable {
     case `accessors`(AccessorDeclListSyntax)
     case `getter`(CodeBlockItemListSyntax)
@@ -223,7 +223,7 @@ public struct AccessorBlockSyntax: SyntaxProtocol, SyntaxHashable {
 /// ### Contained in
 /// 
 ///  - ``AccessorDeclListSyntax``
-public struct AccessorDeclSyntax: DeclSyntaxProtocol, SyntaxHashable {
+public struct AccessorDeclSyntax: DeclSyntaxProtocol, SyntaxHashable, _LeafDeclSyntaxNodeProtocol {
   public let _syntaxNode: Syntax
   
   public init?(_ node: some SyntaxProtocol) {
@@ -480,7 +480,7 @@ public struct AccessorDeclSyntax: DeclSyntaxProtocol, SyntaxHashable {
 /// ### Contained in
 /// 
 ///  - ``AccessorDeclSyntax``.``AccessorDeclSyntax/effectSpecifiers``
-public struct AccessorEffectSpecifiersSyntax: SyntaxProtocol, SyntaxHashable {
+public struct AccessorEffectSpecifiersSyntax: SyntaxProtocol, SyntaxHashable, _LeafSyntaxNodeProtocol {
   public let _syntaxNode: Syntax
   
   public init?(_ node: some SyntaxProtocol) {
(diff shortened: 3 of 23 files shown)


Tests/SwiftRefactorTest/ReformatIntegerLiteral.swift
Matejkob, line 1: I've temporary commented this file because it is problematic for compiler ;(

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxNodesAB.swift
```diff
@@ -207,6 +207,21 @@ public struct AccessorBlockSyntax: SyntaxProtocol, SyntaxHashable {
...
+  
+  @available(*, deprecated, message: "This cast will always fail")
+  public func `as`<S: SyntaxProtocol>(_ syntaxType: S.Type) -> S? {
+    return S.init(self)
+  }
+  
+  @available(*, deprecated, message: "This cast will always fail")
+  public func cast<S: SyntaxProtocol>(_ syntaxType: S.Type) -> S {
+    return self.as(S.self)!
+  }
```
ahoppen, line 224: I think we should be able to avoid generating all these methods by creating the following protocol and then conforming all the non-base nodes to it.

```swift
/// Protocol that syntax nodes conform to which don’t have any semantic subtypes, i.e. syntax nodes that aren't base nodes.
public protocol LeafSyntaxNodeProtocol: SyntaxProtocol {}

public extension LeafSyntaxNodeProtocol {
  @available(*, deprecated, message: "This cast will always fail")
  public func `is`<S: SyntaxProtocol>(_ syntaxType: S.Type) -> Bool {
    return self.as(syntaxType) != nil
  }

  @available(*, deprecated, message: "This cast will always fail")
  public func `as`<S: SyntaxProtocol>(_ syntaxType: S.Type) -> S? {
    return S.init(self)
  }

  @available(*, deprecated, message: "This cast will always fail")
  public func cast<S: SyntaxProtocol>(_ syntaxType: S.Type) -> S {
    return self.as(S.self)!
  }
}
```

ahoppen (commented):
The approach looks good to me. I have one suggestion of how to avoid all the generated code. 

Out of curiosity: How many warnings does this create in swift-syntax?

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxNodesAB.swift
Matejkob: Oh yeah, I believe we should be able to do that. Thanks!

bnbarham: > This cast will always fail

Isn't *necessarily* true right 😅? They could be casting to the same type. I suspect that's usually not the case though, so probably makes sense still.

Matejkob: In our codebase, there are a few instances where we're casting leaf nodes to their own types. I've assumed this is redundant and unnecessary. Have I overlooked something?

Matejkob: Oh, I've just realized you probably meant the content of the message, not the mechanism. 😄 I need coffee 🙈

bnbarham: Sorry, to be clear I was referring to:
> "This cast will always fail"

In that case it won't fail, it's just pointless. The message is fine, I was just being pedantic :)

Matejkob: Introduced 🎸

Matejkob:
> Out of curiosity: How many warnings does this create in swift-syntax?

With changes to also choice nodes something around 25 warnings. Speaking about choice nodes I'll introduce changes to those nodes in separate PR.

CodeGeneration/Sources/SyntaxSupport/SyntaxNodeKind.swift
```diff
@@ -367,6 +367,19 @@ public enum SyntaxNodeKind: String, CaseIterable {
...
   }
 
+  /// For base node types, generates the name of the protocol to which all
+  /// concrete leaf nodes that derive from this base kind should conform.
+  ///
+  /// - Warning: This property can only be accessed for base node kinds; attempting to
+  /// access it for a non-base kind will result in a runtime error.
+  public var leafProtocolType: TypeSyntax {
+    if isBase {
+      return "Leaf\(syntaxType)NodeProtocol"
```
ahoppen, line 377: Sorry, I should have noted this earlier. Since this protocol needs to be public so that that methods defined within it are visible publicly but since we don’t want to offer any guarantees about supporting it in the future, we should prefix it with an underscore.

```suggestion
      return "_Leaf\(syntaxType)NodeProtocol"
```

Release Notes/510.md
```diff
@@ -11,6 +11,11 @@
 
 ## API Behavior Changes
 
+- Leaf Node Protocols
+  - Description: Syntax nodes that do not act as base nodes for other syntax types now conform to a new protocols that deprecates inherited casting methods (e.g. `AccessorBlockSyntax` to `LeafSyntaxNodeProtocol` or `AccessorDeclSyntax` to `LeafDeclSyntaxNodeProtocol`). This change aims to prevent unsafe type-casting by issuing deprecation warnings for methods that will result in failed casts.
```
ahoppen, line 15: I would try to avoid the introduction of the protocols since it’s just a technical detail.

```suggestion
  - Description: Syntax nodes that do not act as base nodes for other syntax types have the casting methods marked as deprecated. This prevents unsafe type-casting by issuing deprecation warnings for methods that will always result in failed casts.
```

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxNodesAB.swift
ahoppen: Good comment, @bnbarham. 

I think we should introduce the `as`/`is`/`cast` methods as follows on `SyntaxProtocol` that will catch same-type conversions an emit a warning the the cast will always succeed, rather than fail.

```swift
  @available(*, deprecated, message: "This cast will always succeed")
  func `as`(_ syntaxType: Self.Type) -> Self? {
    return self
  }
```

Tests/SwiftSyntaxBuilderTest/BreakStmtSyntaxTests.swift
```diff
@@ -17,7 +17,7 @@ import SwiftSyntaxBuilder
 final class BreakStmtSyntaxTests: XCTestCase {
   func testBreakStmtSyntax() {
     let testCases: [UInt: (StmtSyntax, String)] = [
-      #line: (BreakStmtSyntax().as(StmtSyntax.self)!, "break"),
+      #line: (StmtSyntax(BreakStmtSyntax()), "break"),
```
ahoppen, line 20: Hmm, we should support `BreakStmtSyntax().as(StmtSyntax.self)`. It definitely shouldn’t emit a warning that the cast will always fail because that’s not true, it will always succeed.

I think in `LeafDeclSyntaxNodeProtocol`, we need the following (same for `is` and `cast`).

```swift
  func `as`(_ syntaxType: DeclSyntax.Type) -> Self? {
    return self
  }
```

ahoppen (commented):
Nice. Looks very good to me. Just a couple comments.

Tests/SwiftSyntaxBuilderTest/BreakStmtSyntaxTests.swift
Matejkob: I've been thinking more about producing a warning when the user is trying to case to the base node with a message like:
```
 `as` method should be used to downcasting for upcasting to base node use `BaseNode.init`
```
What do you think?

ahoppen: What about making it slightly more action-focused

```
Use `BaseNode.init` for upcasting.
```

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxNodesAB.swift
Matejkob: Really good idea! Implemented in https://github.com/apple/swift-syntax/pull/2108/commits/9986d57dfd80800c30b686ae05b55f8739c168e9

CodeGeneration/Sources/SyntaxSupport/SyntaxNodeKind.swift
Matejkob: Done

Release Notes/510.md
Matejkob: Thanks, done

Tests/SwiftSyntaxBuilderTest/BreakStmtSyntaxTests.swift
Matejkob: I've added such a feature for all base nodes and `Syntax` node here: https://github.com/apple/swift-syntax/pull/2108/commits/9986d57dfd80800c30b686ae05b55f8739c168e9

Matejkob:
In a separate PR, I'm going to address casting issues related to syntax traits and syntax choice nodes.
<hr>
I'm thinking about changing the approach here a bit: Rather than scattering bits of casting logic across multiple files, it might be more efficient to consolidate all the casting functionality for each node and case into a single file. This could make it easier to grasp the overall logic and any edge cases.

How does that sound?

Release Notes/510.md
```diff
@@ -19,6 +19,21 @@
...
+- Same-Type Casts 
+  - Description: `is`, `as`, and `cast` overloads on `SyntaxProtocol` with same-type conversions are marked as deprecated. The deprecated methods emit a warning indicating the cast will always succeed.
+  - Issue: https://github.com/apple/swift-syntax/issues/2092
+  - Pull Request: https://github.com/apple/swift-syntax/pull/2108
+
+- Base Node Casts
+  - Description: `is`, `as`, and `cast` methods on base node protocols with base-type conversions are marked as deprecated. The deprecated methods emit a warning that informs the developer that the cast will always succeed and should be done using the base node's initializer.
+  - Issue: https://github.com/apple/swift-syntax/issues/2092
+  - Pull Request: https://github.com/apple/swift-syntax/pull/2108
+
```
ahoppen, line 36: I think this should go in the `Deprecations` sections since the runtime behavior of the API doesn’t change.

ahoppen (commented):
One minor comment on the release notes, otherwise LGTM.

> I'm thinking about changing the approach here a bit: Rather than scattering bits of casting logic across multiple files, it might be more efficient to consolidate all the casting functionality for each node and case into a single file. This could make it easier to grasp the overall logic and any edge cases.
How does that sound?

I think that could be a good idea. Let’s do it in a follow-up PR. That way we can get this behavior change merged and the discussion on the follow-up PR can be purely around the implementation structuring.

ahoppen:
@swift-ci Please test

Release Notes/510.md
Matejkob: Done

ahoppen (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #2110, opened 2023-08-28
Add whitespace if needed when returning syntaxTextBytes

### LINKED ISSUE #2107: Required spacing is not added in `SwiftSyntaxBuilder` for interpolated syntax node
### Description

For like syntax nodes like `FunctionEffectSpecifiersSyntax` and `FunctionParameterSyntax`, some spacings are required internally. However, if a handcraft node is interpolated into another expression string, `SwiftSyntaxBuilder` fails to add these spacing, causing the parser to fail and the builder result being wrong.

### Steps to Reproduce

```swift
let functionPrameterClause = FunctionParameterClauseSyntax {
  FunctionParameterSyntax(firstName: "_", secondName: "name", type: TypeSyntax("String"))
  FunctionParameterSyntax(firstName: "description", type: TypeSyntax("String"))
}
let functionSignature = FunctionSignatureSyntax(
  parameterClause: functionPrameterClause,
  effectSpecifiers: .init(asyncSpecifier: .keyword(.async), throwsSpecifier: .keyword(.throws))
)
assertBuildResult(functionSignature, "(_ name: String, description: String) async throws") // OK
let initializerDecl = InitializerDeclSyntax(signature: functionSignature) {
  "self.name = name"
  "self.description = description"
}
assertBuildResult(
  initializerDecl,
  """
  init(_ name: String, description: String) async throws {
      self.name = name
      self.description = description
  }
  """
) // OK
let initializerDeclFromString = try InitializerDeclSyntax("init\(functionSignature)") {
  "self.name = name"
  "self.description = description"
}
assertBuildResult(
  initializerDeclFromString,
  """
  init(_ name: String, description: String) async throws {
      self.name = name
      self.description = description
  }
  """
) // Error
```
Outputs:
```
xctest [ParseError] Parsing a `DeclSyntax` node from string interpolation produced the following parsing errors.
Set a breakpoint in `SyntaxParseable.logStringInterpolationParsingError()` to debug the failure.
1 │ init(_name:String,description:String)asyncthrows {}
  │                                      ╰─ error: unexpected code 'asyncthrows' in initializer
error: Test failed - Actual output (+) differed from expected output (-):
–init(_ name: String, description: String) async throws {
+init(_name: String, description: String) {
     self.name = name
     self.description = description
 }
```

TTOzzi:
Resolve #2107 

While it's not a problem if different kinds of tokens that can be distinguished by their kinds appear consecutively without whitespace; however, if tokens of the same kind are positioned without a delimiter, there arises an issue when converting syntaxTextBytes into a SyntaxNode.

When two indistinguishable(of the same kind) tokens are positioned consecutively without whitespace or punctuation, I have added whitespace for differentiation.

### DIFF
--- a/Sources/SwiftSyntax/Raw/RawSyntax.swift
+++ b/Sources/SwiftSyntax/Raw/RawSyntax.swift
@@ -41,6 +41,21 @@ internal struct RawSyntaxData {
     case parsedToken(ParsedToken)
     case materializedToken(MaterializedToken)
     case layout(Layout)
+
+    var tokenKind: RawTokenKind? {
+      switch self {
+      case .parsedToken(let dat):
+        return dat.tokenKind
+      case .materializedToken(let dat):
+        return dat.tokenKind
+      case .layout(let dat):
+        guard dat.descendantCount == 1 else {
+          return nil
+        }
+        let singleChild = dat.layout.compactMap({ $0 }).first
+        return singleChild?.payload.tokenKind
+      }
+    }
   }
 
   /// Token with lazy trivia parsing.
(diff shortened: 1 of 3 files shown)


ahoppen (changes requested):
This is not correct because it breaks round-tripness of the syntax tree. If you generate source code, parse it and print the syntax tree, it must exactly correspond to the original source tree. 

With this change a space that wasn’t present in the original source is added during the printing, which breaks this property.

As I commented in #2107, I think the fix for that issue is a lot more involved.

TTOzzi:
From the comments you left on the issue, I understand that it's much more difficult than I realized to fix the issue without modifications to the original source.
Thanks for the feedback! I'll close this PR 🙇

Issue #1984, opened 2023-08-02
Add initializer to `LabeledExprSyntax` that automatically adds a colon if the label exists


ahoppen:
Currently, if you do the following

```swift
let node = EnumCaseParameterSyntax(firstName: "label", type: TypeSyntax("MyType"))
print(node.formatted())
```
you get `label MyType`, which is surprising because the colon is missing and you would expect the colon to be present if there is a first name of the enum case parameter.

To fix this, we should add a hand-written initializer of `EnumCaseParameterSyntax`, which takes a non-optional `firstName` and automatically adds a colon, because it knows that the `firstName` exists. The above call would pick that overload and thus add a colon.

This probably also applies to other nodes that have an optional colon. 

And the same strategy can also be applied to automatically add parentheses for the following nodes if arguments are present
- `AttributeSyntax`
- `MacroExpansionDeclSyntax`
- `ClosureCaptureModifierSyntax`
- `FunctionCallExprSyntax`
- `FreestandingMacroExpansionSyntax`

rdar://107794232

natikgadzhi:
I would love to tackle this after the `isAt` to `at` change. This looks a bit more involved, and I would love to dive in and learn a bit more about how things work.

ahoppen:
I assigned the issue to you. Let me know if you have any questions.

natikgadzhi:
I started looking into it. The first two PRs I was playing on easy, but for this one, I'll stop and read the documentation on the basic building blocks of swift-syntax — so that one will take a few days. 

I hope to put together the first pull request with just the `EnumCaseParameterSyntax` and ask you for a review, so we check and make sure I got it right. I'll expand to the rest of the list then. 

How does that sound, @ahoppen?

ahoppen:
Sounds good to me 👍🏽

natikgadzhi:
Found a little bug somewhere in swift-markdown or swift-docc: https://github.com/apple/swift-docc/issues/685 

I'm continuing to work on this — hope to have some focus time on Sunday to figure out where the definitions are for the generated code, and how to edit them. Taking it slow to not cut too many corners.

natikgadzhi:
@ahoppen, by handwritten, do you suggest that we add an extension for `EnumCaseParameterSyntax` with the initializer, and put it somewhere like `Sources/SwiftSyntax/SyntaxNodes-overrides.swift`? Is there an existing convention on how we add handwritten functions / properties to the generated structs?

ahoppen:
[...]

PR #2112, opened 2023-08-29
Add automatic colons or paranthesis to nodes when needed

natikgadzhi:
This pull request adds a convenience init to `EnumCaseParameterSyntax` so that when you're initializing it with a `firstName`, it adds `colon` automatically.

It's the first step in #1984.

## Changes
- `SwiftSyntax` already had a `Convenience.swift`, so I'm adding an extension there for now. The file is small and readable. 
- I've added a unit test to `SwiftSyntaxTest/SyntaxTest.swift`, however as we progress in #1984, I'll perhaps extract them into a `ConvenienceInitTests.swift`.

## How to review this
- I love tinkering with this, but I do feel like Alice in Wonderland. @ahoppen, did I capture your idea well, does this approach seem right? 
- I've used `let firstName: TokenSyntax? = firstName` to trick Swift into using the right overload with an optional `firstName` so that the underlying generated `init` will be called, instead of copying it over.
- While this PR might be self-sufficient, we should probably add the rest of the todos to this as well? 

## TODO
Copying over from the issue: 
And the same strategy can also be applied
- [ ] Other nodes that have an optional colon
- [ ] `ClosureCaptureSyntax.equal`
- Automatically add parentheses for the following nodes if arguments are present:
   - [ ] `AttributeSyntax`
   - [ ] `MacroExpansionDeclSyntax`
   - [ ] `ClosureCaptureModifierSyntax`
   - [ ] `FunctionCallExprSyntax`
   - [ ] `FreestandingMacroExpansionSyntax`

### DIFF
--- a/Sources/SwiftSyntax/Convenience.swift
+++ b/Sources/SwiftSyntax/Convenience.swift
@@ -10,6 +10,37 @@
 //
 //===----------------------------------------------------------------------===//
 
+extension EnumCaseParameterSyntax {
+
+  /// Creates an `EnumCaseParameterSyntax` with a `firstName`, and automatically adds a `colon` to it.
+  ///
+  ///  - SeeAlso: For more information on the arguments, see ``EnumCaseParameterSyntax/init(leadingTrivia:_:modifiers:_:firstName:_:secondName:_:colon:_:type:_:defaultArgument:_:trailingComma:_:trailingTrivia:)``
+  ///
+  public init(
+    leadingTrivia: Trivia? = nil,
+    modifiers: DeclModifierListSyntax = [],
+    firstName: TokenSyntax,
+    secondName: TokenSyntax? = nil,
+    colon: TokenSyntax = TokenSyntax.colonToken(),
+    type: some TypeSyntaxProtocol,
+    defaultValue: InitializerClauseSyntax? = nil,
+    trailingComma: TokenSyntax? = nil,
+    trailingTrivia: Trivia? = nil
+  ) {
+    self.init(
+      leadingTrivia: leadingTrivia,
+      modifiers: modifiers,
+      firstName: firstName as TokenSyntax?,
+      secondName: secondName,
+      colon: colon,
+      type: type,
+      defaultValue: defaultValue,
+      trailingComma: trailingComma,
+      trailingTrivia: trailingTrivia
+    )
+  }
+}
+
 extension MemberAccessExprSyntax {
   /// Creates a new ``MemberAccessExprSyntax`` where the accessed member is represented by
   /// an identifier without specifying argument labels.
@@ -45,3 +76,8 @@ extension MemberAccessExprSyntax {
     )
   }
 }
+
+//==========================================================================//
+// IMPORTANT: If you are tempted to add an extension here, please insert    //
+// it in alphabetical order above                                           //
+//==========================================================================//
--- a/Tests/SwiftSyntaxTest/SyntaxTests.swift
+++ b/Tests/SwiftSyntaxTest/SyntaxTests.swift
@@ -125,4 +125,12 @@ public class SyntaxTests: XCTestCase {
     XCTAssertEqual(funcKW.endPosition, AbsolutePosition(utf8Offset: 7))
     XCTAssertEqual(funcKW.trimmedLength, SourceLength(utf8Length: 4))
   }
+
+  public func testEnumCaseParameterSyntaxConvenienceInit() {
+    let noFirstName = EnumCaseParameterSyntax(type: TypeSyntax("MyType"))
+    XCTAssertEqual(noFirstName.formatted().description, "MyType")
+
+    let node = EnumCaseParameterSyntax(firstName: "label", type: TypeSyntax("MyType"))
+    XCTAssertEqual(node.formatted().description, "label: MyType")
+  }
 }


Sources/SwiftSyntax/Convenience.swift
```diff
@@ -45,3 +45,56 @@ extension MemberAccessExprSyntax {
...
+  /// - Parameters:
+  ///   - leadingTrivia: Trivia to be prepended to the leading trivia of the node’s first token. If the node is empty, there is no token to attach the trivia to and the parameter is ignored.
+  ///   - colon: If the parameter has a label, the colon separating the label from the type. SInce `firstName` is not nil, the default `colon` will be set to `TokenSyntax.colonToken()`.
+  ///   - type: The parameter's type.
+  ///   - defaultValue: If the parameter has a default value, the initializer clause describing the default value.
+  ///   - trailingComma: If the parameter is followed by another parameter, the comma separating them.
+  ///   - trailingTrivia: Trivia to be appended to the trailing trivia of the node’s last token. If the node is empty, there is no token to attach the trivia to and the parameter is ignored.
+  public init(
+    leadingTrivia: Trivia? = nil,
+    _ unexpectedBeforeModifiers: UnexpectedNodesSyntax? = nil,
```
Matejkob, line 62: I believe this handwritten initializer is specifically designed for manual use by developers. Given that, I'm wondering if there is the need to include unexpected nodes in it. It doesn't seem likely that anyone would want to manually create an unexpected node. What are your thoughts?

Sources/SwiftSyntax/Convenience.swift
```diff
@@ -45,3 +45,56 @@ extension MemberAccessExprSyntax {
...
+
+  /// Creates an `EnumCaseParameterSyntax` with a `firstName`, and automatically adds a `colon` to it.
+  ///
+  /// - Parameters:
+  ///   - leadingTrivia: Trivia to be prepended to the leading trivia of the node’s first token. If the node is empty, there is no token to attach the trivia to and the parameter is ignored.
+  ///   - colon: If the parameter has a label, the colon separating the label from the type. SInce `firstName` is not nil, the default `colon` will be set to `TokenSyntax.colonToken()`.
+  ///   - type: The parameter's type.
+  ///   - defaultValue: If the parameter has a default value, the initializer clause describing the default value.
+  ///   - trailingComma: If the parameter is followed by another parameter, the comma separating them.
+  ///   - trailingTrivia: Trivia to be appended to the trailing trivia of the node’s last token. If the node is empty, there is no token to attach the trivia to and the parameter is ignored.
```
Matejkob, line 59: Awesome that you've added a doc comment to this `init`! I noticed it's the same as the one in the original `init`. Given that the original comment is auto-generated by the [`CodeGeneration` package](https://github.com/apple/swift-syntax/tree/main/CodeGeneration), it could be updated fairly easily, making this one obsolete pretty quickly. I'm wondering if there's a way to address it. Though, to be honest, coming up with an advanced solution might just be overengineering the whole thing. What do you think?

Sources/SwiftSyntax/Convenience.swift
```diff
@@ -45,3 +45,56 @@ extension MemberAccessExprSyntax {
...
+    defaultValue: InitializerClauseSyntax? = nil,
+    _ unexpectedBetweenDefaultValueAndTrailingComma: UnexpectedNodesSyntax? = nil,
+    trailingComma: TokenSyntax? = nil,
+    _ unexpectedAfterTrailingComma: UnexpectedNodesSyntax? = nil,
+    trailingTrivia: Trivia? = nil
+
+  ) {
+      // Wrapping firstName as an Optional so that we can use the generated overload.
+    let firstNameAsOptional: TokenSyntax? = firstName
+    self = .init(leadingTrivia: leadingTrivia,
```
Matejkob, line 82: Nitpick: I think using `self.init(` in that spot would achieve the same result while also making the code more consistent with other areas in the codebase. 😅
```suggestion
    self.init(leadingTrivia: leadingTrivia,
```

Sources/SwiftSyntax/Convenience.swift
```diff
@@ -45,3 +45,56 @@ extension MemberAccessExprSyntax {
...
+    colon: TokenSyntax = TokenSyntax.colonToken(),
+    _ unexpectedBetweenColonAndType: UnexpectedNodesSyntax? = nil,
+    type: some TypeSyntaxProtocol,
+    _ unexpectedBetweenTypeAndDefaultValue: UnexpectedNodesSyntax? = nil,
+    defaultValue: InitializerClauseSyntax? = nil,
+    _ unexpectedBetweenDefaultValueAndTrailingComma: UnexpectedNodesSyntax? = nil,
+    trailingComma: TokenSyntax? = nil,
+    _ unexpectedAfterTrailingComma: UnexpectedNodesSyntax? = nil,
+    trailingTrivia: Trivia? = nil
+
```
Matejkob, line 78: I believe you can go ahead and delete this empty line. The original `init` probably contains it because the `CodeGeneration` tool is not perfect.

Matejkob (commented):
Great work on this! I took the liberty of adding a few questions and suggestions—I hope that's alright with you.

I saw in the description that you're still considering whether this will be the final PR or if additional initializers are coming. If you're leaning towards making this the final one, don't forget to [run swift-format after your latest changes](https://github.com/apple/swift-syntax/blob/main/CONTRIBUTING.md#formatting).

Sources/SwiftSyntax/Convenience.swift
natikgadzhi: That's a _great_ question. I'm actually clueless yet, and just learning. As I work with swift-syntax some more, I hope to understand it better, and actually have an opinion. But alas, not yet.

So I appreciate your feedback and direction on this. Your call. Should I clean them out?

Sources/SwiftSyntax/Convenience.swift
natikgadzhi: You're absolutely right re: getting out of sync. 

> Though, to be honest, coming up with an advanced solution might just be overengineering the whole thing.

When I looked at that whole issue, I thought that if we really wanted to, there must be a way to tweak code generation to generate multiple convenience initializers as well. That would probably take some time, but could be quite elegant.

The most simple way out of this is to make a shorter comment and link to the generated method docs for details on other parameters.

natikgadzhi:
@Matejkob HUGE thank you for the review!
- Committed your suggestions
- Formatted with swift-format from brew. 
- I _think_ swift-format@main is broken, filing an issue and will do some yak shaving tomorrow.

Sources/SwiftSyntax/Convenience.swift
ahoppen: For now, my suggestion would be to keep the part that you wrote by hand and add a `- SeeAlso:` section that links to the underlying initializer. 

As we write more of these initializers, maybe a pattern will emerge and we decide to generate the initializers, in which case we can also generate the section. But for now, that would just be over-engineering IMO.

ahoppen:
> * I _think_ swift-format@main is broken, filing an issue and will do some yak shaving tomorrow.

Which error are you seeing? Running `./format.py` should work.

natikgadzhi:
> Which error are you seeing? Running ./format.py should work.

I cooked SwiftPM wrong, and it couldn't reset `Package.resolved`, so it tried to build a fresh `swift-format` with older `swift-syntax`. Fixed now.

@Matejkob @ahoppen thank you both for reviewing, and for your feedback! I'm learning the pieces, and it starts to make more sense with every hour 🙃 I worked through the feedback and suggestions.

Since we're not squashing commits, I've amended the original one, so it looks clean in the repo history. I will add the next items on the list as separate commits.

I have a suggestion: @ahoppen, @Matejkob, how do you feel about splitting the work into three pull requests: 
1. This one, that will include the other nodes that have an optional colon (um, I will search through the codebase and try to actually use my brain to come up with a list, but may need help).
2. Another small pull request that will add an `init` for `ClosureCaptureSyntax` that adds `equal` token. It should be small and easy.
3. The third pull request that will include parentheses for a bunch of nodes.

That way, I can make progress on them and ship them one by one faster than holding a bigger PR. Hopefully, it will be easier to review as well.

Sources/SwiftSyntax/Convenience.swift
```diff
@@ -45,3 +45,57 @@ extension MemberAccessExprSyntax {
     )
   }
 }
+
+extension EnumCaseParameterSyntax {
```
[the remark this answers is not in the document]
natikgadzhi: I _think_ that writing a script that does this automatically for an arbitrary file would be a nice exercise and an example of how to use SwiftSyntax!

How would you feel about having an additional script that `./format.py` would invoke? It could live as an executable target in `swift-format` as well, perhaps?

Or.. maybe that could be a part of `swift-format` that would only trigger on files with paths matching a setting? I.e. 

```
{
  // ...
  "rules" : {
    "AllPublicDeclarationsHaveDocumentation" : false,
    "AlwaysUseLowerCamelCase" : true,
    
    // What if a rule would support not only a Boolean value, but an array of paths for which the value is true (or different from a default value)
    "AlphabetizeDeclarations": [ "Sources/SwiftSyntax/generated/Convenience.swift", "**/Convenience.swift"]
    // ...
  },
  // ... 
}
```

I could start with a little experiment as an example to learn to use SwiftSyntax, and if I can get a prototype, I could write a proposal on Swift Forums to make the settings support path arrays?

Sources/SwiftSyntax/Convenience.swift
natikgadzhi: Done!

Sources/SwiftSyntax/Convenience.swift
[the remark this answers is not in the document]
Matejkob: We discussed a similar idea [here](https://github.com/apple/swift-syntax/pull/1903#discussion_r1262293841). At the time, we concluded that it wouldn't be a trivial task, so the subject was put on the back burner. However, the fact that the idea has resurfaced suggests there might be a real need for such a tool, and it could be beneficial to revisit the discussion.

How about we move this discussion to a separate issue? @natikgadzhi, would you like to do the honors and open one? ☺️

ahoppen (approved):
Nice. Thanks a lot. Do you also want to cover the other cases mentioned in the issue in follow-up PRs?

ahoppen:
@swift-ci Please test

Sources/SwiftSyntax/Convenience.swift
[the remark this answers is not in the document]
ahoppen: If you have an idea of how to implement it, I’d be happy to take it. But as I mentioned in the other discussion, I think it’s far from a trivial task, if not even (nearly) impossible. 

And so far, we have been good with these comments so at least as far as I’m concerned there’s no immediate need for action.

ahoppen:
> I have a suggestion: @ahoppen, @Matejkob, how do you feel about splitting the work into three pull requests:
> 
> 1. This one, that will include the other nodes that have an optional colon (um, I will search through the codebase and try to actually use my brain to come up with a list, but may need help).
> 2. Another small pull request that will add an `init` for `ClosureCaptureSyntax` that adds `equal` token. It should be small and easy.
> 3. The third pull request that will include parentheses for a bunch of nodes.
> 
> That way, I can make progress on them and ship them one by one faster than holding a bigger PR. Hopefully, it will be easier to review as well.

I’m also happy to merge this one as-is as a proof of concept and then make any similar changes to other nodes in follow-up PRs. But you can also keep amending this PR, whatever you prefer.

ahoppen:
@swift-ci Please test Windows

natikgadzhi:
> Nice. Thanks a lot. Do you also want to cover the other cases mentioned in the issue in follow-up PRs?

Yep, makes sense — happy to work on the rest of it.

ahoppen:
Great. Thanks

PR #2113, opened 2023-08-29
Improvements to MacroSystemTests

ahoppen:
- Refactor MacroSystemTests to make most macro definitions local to their test case: Most of the macro definitions were only used in a single test case. Achieve a higher locality of functionality by moving the macro definitions inside these test cases.
- Some general improvements to the usage of SwiftSyntax in the macros in MacroSystemTests
- Removal of trivia handling in these macros because trivia is being added by MacroSystem now.

### DIFF
--- a/Tests/SwiftSyntaxMacroExpansionTest/MacroSystemTests.swift
+++ b/Tests/SwiftSyntaxMacroExpansionTest/MacroSystemTests.swift
@@ -20,174 +20,10 @@ import SwiftSyntaxMacroExpansion
 import SwiftSyntaxMacrosTestSupport
 import XCTest
 
-// MARK: Example macros
-public struct StringifyMacro: ExpressionMacro {
-  public static func expansion(
-    of macro: some FreestandingMacroExpansionSyntax,
-    in context: some MacroExpansionContext
-  ) throws -> ExprSyntax {
-    guard let argument = macro.argumentList.first?.expression else {
-      throw MacroExpansionErrorMessage("missing argument")
-    }
+// MARK: Macros shared between multiple test cases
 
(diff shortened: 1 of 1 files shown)


Tests/SwiftSyntaxMacroExpansionTest/MacroSystemTests.swift
Matejkob, line 1: I'm wondering if it would be beneficial to split these tests into smaller files. This file already has over 1,600 lines, and with perspective on future macros (like the body macro, for instance), it's going to keep growing. I personally prefer to break code into smaller chunks—it makes it easier to read and maintain, I'm not sure yet what is the chosen aproceh there. 

A reasonable approach might be to divide the tests by macro type. However, we do have a few test cases that involve multiple macros of different types simultaneously. So, I'm not quite sure how to handle that situation. 

I'm mainly thinking out loud here. Let me know what you think.

Matejkob (commented):
Great idea to move the macro definitions closer to where they're actually used!

Tests/SwiftSyntaxMacroExpansionTest/MacroSystemTests.swift
ahoppen: I was thinking about that as well but then just stopped after moving the macro definitions. 

I think having one test file for each macro type and maybe one file that tests the integration of multiple macro types would make sense.

Matejkob: > I think having one test file for each macro type and maybe one file that tests the integration of multiple macro types would make sense.

Oh yeah, I agree

bnbarham (approved):

PR #2115, opened 2023-08-29
Fix issue that caused errors thrown from macro expansion to show up twice in MacroSystem

### LINKED ISSUE #2111: `assertMacroExpansion` reports duplicate diagnostic
### Description

Expression macros that throw errors report duplicate diagnostics to `assertMacroExpansion`, and these duplicate diagnostics must be asserted against.

### Steps to Reproduce

I have a slightly modified version of the default template `#stringify` macro that throws an error when the input is too long:

```diff
 public struct StringifyMacro: ExpressionMacro {
   public static func expansion<
     F: FreestandingMacroExpansionSyntax, C: MacroExpansionContext
   >(
     of node: F,
     in context: C
-  ) -> ExprSyntax {
+  ) throws -> ExprSyntax {
     guard let argument = node.argumentList.first?.expression else {
       fatalError("compiler bug: the macro does not have any arguments")
     }

+    guard argument.description.count <= 20
+    else {
+      throw SomeError()
+    }
+
     return "(\(argument), \(literal: argument.description))"
   }
 }
+
+struct SomeError: Error {}
```

If I write an expansion test for this:

```swift
assertMacroExpansion(
  """
  #stringify(a + b + c + d + e + f)
  """,
  expandedSource: """
    #stringify(a + b + c + d + e + f)
    """,
  macros: testMacros
)
```

I get the following failure:

> 🛑 failed - Expected 0 diagnostics but received 2:
> 1:1: SomeError()
> 1:1: SomeError()

Updating the assertion requires including _two_ diagnostics even though there should be just one.

ahoppen:
If the macro expansion of a freestanding expression macro throws an error, `expandCodeBlockItem` returned `nil` while adding the thrown error to the macro expansion context. `visit(_:CodeBlockItemListSyntax).addResult` took the `nil` return value as an indicator that the macro wasn’t expanded because its macro definition wasn’t found and ended up calling the expansion again in

```swift
// Recurse on the child node
newItems.append(visit(node))
```

Change the return value of `expandCodeBlockItem` to an enum that indicates whether the macro was not found or if the expansion failed. If the macro was found but the expansion threw an error, we just just retain the macro as-is without calling into `visit` again.

Fixes #2111
rdar://114592410

### DIFF
--- a/Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
+++ b/Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
@@ -485,10 +485,14 @@ private class MacroApplication<Context: MacroExpansionContext>: SyntaxRewriter {
     // Note that 'MacroExpansionExpr'/'MacroExpansionExprDecl' at code item
     // position are handled by 'visit(_:CodeBlockItemListSyntax)'.
     // Only expression expansions inside other syntax nodes is handled here.
-    if let expanded = expandExpr(node: node) {
+    switch expandExpr(node: node) {
+    case .success(let expanded):
       return Syntax(visit(expanded))
+    case .failure:
+      return Syntax(node)
+    case .notAMacro:
+      break
     }
-
     if let declSyntax = node.as(DeclSyntax.self),
       let attributedNode = node.asProtocol(WithAttributesSyntax.self),
       !attributedNode.attributes.isEmpty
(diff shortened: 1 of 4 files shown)


ahoppen:
@swift-ci Please test

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
```diff
@@ -842,20 +854,36 @@ extension MacroApplication {
...
+  enum MacroExpansionResult<ResultType> {
+    /// Expansion of the macro succeeded.
+    case success(ResultType)
+
+    /// Macro system found the macro to expand but running the expansion threw
+    /// an error and thus no expansion result exists.
+    case failure
+
+    /// The node that should be expanded was not a macro known to the macro system.
+    case notAMacro
```
bnbarham, line 866: `unknownMacro`

bnbarham (approved):

PR #2120, opened 2023-08-30
Remove the `parent` parameter from `SyntaxData.child(at:parent:)`

ahoppen:
This parameter isn’t used anymore, so we can remove it.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #2121, opened 2023-08-30
Use two-step dispatch to `SyntaxVisitor.vistImpl`

ahoppen:
`SyntaxRewriter` has had a problem with stack allocation space for a while because cases in a switch statement don’t share stack space in debug builds (rdar://55929175). We started seeing this issue with `SyntaxVisitor` as well now. Use the same approach to fix the problem:

To circumvent this problem, make calling the specific visitation function a two-step process: First determine the function to call in this function and return a reference to it, then call it. This way, the stack frame that determines the correct visitation function will be popped of the stack before the function is being called, making the switch's stack space transient instead of having it linger in the call stack.

And to make the entire approach look sane, allow 4 arguments on the same line for SyntaxVisitor instead of the 3 that we allow in all other files.


ahoppen:
@swift-ci Please test

CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxVisitorFile.swift
```diff
@@ -113,27 +113,108 @@ let syntaxVisitorFile = SourceFileSyntax(leadingTrivia: copyrightHeader) {
...
+                """
+                /// Implementation detail of visit(_:). Do not call directly.
+                ///
+                /// Returns the function that shall be called to visit a specific syntax node.
+                ///
+                /// To determine the correct specific visitation function for a syntax node,
+                /// we need to switch through a huge switch statement that covers all syntax
+                /// types. In debug builds, the cases of this switch statement do not share
+                /// stack space (rdar://55929175). Because of this, the switch statement
+                /// requires allocates about 15KB of stack space. In scenarios with reduced
```
bnbarham, line 142:
```suggestion
                /// stack space (rdar://55929175). Because of this, the switch statement
                /// requires about 15KB of stack space. In scenarios with reduced
```

bnbarham (approved):

ahoppen:
@swift-ci Please test

kimdv:
@swift-ci please test Windows

PR #2123, opened 2023-08-30
Add a release notes document for swift-syntax 510

ahoppen:
My idea is that we can build up the list of changes during development. That will make it more transparent for adopters of the snapshot tags to see what has changed and, in case of deprecations or API-incompatible changes, how they need to update.

It will also make it easier to compose the list of changes for the proper (non-snapshot) release of swift-syntax.

### DIFF
--- a/Release Notes/510.md
+++ b/Release Notes/510.md
@@ -1,0 +1,27 @@
+# Swift Syntax 510 Release Notes
+
+## New APIs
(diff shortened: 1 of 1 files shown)


bnbarham (approved):

ahoppen:
@swift-ci Please test

PR #2124, opened 2023-08-30
Set the privacy level of string interpolation errors in os_log to private

ahoppen:
If string interpolation results in parsing errors, we create an os_log fault to notify the user about a potential bug. This log message contains the source code with the syntax error, which could be sensitive.

Set the os_log privacy level to `private` to make sure the source code does not get persisted in logs. It will still show up in the Xcode console.


ahoppen:
@swift-ci Please test

kimdv (approved):

bnbarham (approved):

PR #2132, opened 2023-09-01
Fold operators using the standard operator table in `MacroSystem`

### LINKED ISSUE #2128: No way to describe operator precedence to `assertMacroExpansion`
### Description

It doesn't seem possible to write tests for macros that visit `InfixOperatorExprSyntax`. This includes the "add blocker" example from swift-macro-examples.

### Steps to Reproduce

1. Try to write a macro test for: https://github.com/DougGregor/swift-macro-examples/blob/66ded9b2774c06c5a51254b0640a456ef1055a68/MacroExamplesPlugin/AddBlocker.swift
2. The `AddVisitor` never visits the node because the node doesn't exist in the test run.

ahoppen:
The compiler folds operators in attributes and freestanding macro nodes but `MacroSystem` wasn’t doing that. But it should to match the compiler behavior.

rdar://114786803
Fixes #2128

### DIFF
--- a/Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
+++ b/Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
@@ -11,6 +11,7 @@
 //===----------------------------------------------------------------------===//
 
 import SwiftDiagnostics
+import SwiftOperators
 @_spi(MacroExpansion) import SwiftParser
 import SwiftSyntax
 import SwiftSyntaxBuilder
@@ -55,7 +56,7 @@ private func expandFreestandingMemberDeclList(
     let expanded = try expandFreestandingMacro(
       definition: definition,
       macroRole: inferFreestandingMacroRole(definition: definition),
-      node: node.detach(in: context),
+      node: node.detach(in: context, foldingWith: .standardOperators),
       in: context,
       indentationWidth: indentationWidth
     )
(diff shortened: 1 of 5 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
```diff
@@ -979,4 +980,47 @@ private extension SyntaxProtocol {
...
+
+private extension FreestandingMacroExpansionSyntax {
+  /// Same as `SyntaxProtocol.detach(in:foldingWith:)` but returns a node of type
+  /// `Self` since we know that operator folding doesn't change the type of any
+  /// `FreestandingMacroExpansionSyntax`.
+  func detach(
+    in context: MacroExpansionContext,
+    foldingWith operatorTable: OperatorTable?
+  ) -> Self {
+    return (detach(in: context, foldingWith: operatorTable) as Syntax).cast(Self.self)
```
bnbarham, line 1012:
```suggestion
    return (detach(in: context, foldingWith: operatorTable) as Syntax).cast(Self.self)
```

What's the `as` needed for here? Doesn't this already return a `Syntax`?

bnbarham (approved):

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
ahoppen: It’s to disambiguate the version of `detach` that is being called. Without `as Syntax` we would have an infinite recursion here.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #2127, opened 2023-08-31
Add a convenience init to ClosureCaptureSyntax

natikgadzhi:
## Summary

This adds a convenience `init` to `ClosureCaptureSyntax` that adds an `equal` token if the provided `name` is not nil.

It's supposed to be used by developers, so it omits the unexpected tokens.

Related: #1984

PR #2142, opened 2023-09-01
WIP: Make convenience initializers with CodeGeneration

### LINKED ISSUE #1984: Add initializer to `LabeledExprSyntax` that automatically adds a colon if the label exists

natikgadzhi:
## Summary

**Work in progress, not ready for a full-on review, ideas and feedback welcome!**

This pull request closes #1984. Instead of doing one-off initializers by hand, like in #2127 and #2112, we add a code generation clause that makes the initializers based on `NodeInitRule`s. 

## Approach overview

- `Node` gets a new property: `rules: [NodeInitRule]`. 
- Each `NodeInitRule` has `nonOptionalChildName` and `childDefaultValues: [String: Token]`. 
- The generator makes initializers that take non-optional parameter `nonOptionalChildName` and set defaults to `childDefaultValues.keys`. 
- There's an example of a rule in `AttributeSyntax`:
   ```swift
    rules: [
      NodeInitRule(
        nonOptionalChildName: "arguments",
        childDefaultValues: [
          "leftParen": .leftParen,
          "rightParen": .rightParen
        ])
    ],
   ```

## Next steps

- [x] Generate the init declaration
- [x] Generate the init documentation
- [ ] Generate the init body: call `self.init` with the parameters list.
- [ ] Figure out how to test the initializers
- [ ] Add `NodeInitRules` to cover everything in #1984, and other `SyntaxNodes` where `colon` is optional where it's applicable.
- [ ] Clean up the code
- [ ] Rebase and split into commits for easier review


CodeGeneration/Sources/SyntaxSupport/Node.swift
```diff
@@ -380,3 +391,8 @@ fileprivate extension Child {
 fileprivate extension Node {
 
 }
+
```
natikgadzhi, line 394: TODO: This needs a comment
TODO: Move `NodeInitRule` into it's own file.

Validation rules can live within NodeInitRule — perhaps in it's `init()` or in a separate function, i.e. `validate(for node: Node)`.

CodeGeneration/Sources/Utils/SyntaxBuildableChild.swift
```diff
@@ -84,7 +84,7 @@ public extension Child {
 
   /// If the child node has a default value, return an expression of the form
   /// ` = default_value` that can be used as the default value to for a
-  /// function parameter. Otherwise, return `nil`.
+  /// function parameter. Otherwise, return `nil`.]
```
natikgadzhi, line 87: Typo

```suggestion
  /// function parameter. Otherwise, return `nil`.
```

CodeGeneration/Sources/generate-swift-syntax/LayoutNode+Extensions.swift
```diff
@@ -16,12 +16,65 @@ import SyntaxSupport
...
+  /// - parameters:
+  ///  - rule: The ``NodeInitRule`` to use for generating the initializer. Applying a rule will make some children non-optional, and set default values for other children.
+  ///  - useDeprecatedChildName: Whether to use the deprecated child name for the initializer parameter.
+  func generateInitializerDeclHeader(for rule: NodeInitRule? = nil, useDeprecatedChildName: Bool = false) -> SyntaxNodeString {
     if children.isEmpty {
       return "public init()"
     }
 
-    func createFunctionParameterSyntax(for child: Child) -> FunctionParameterSyntax {
+    func childParameterName(for child: Child) -> TokenSyntax {
```
natikgadzhi, line 29: This is monstrously large — extract these helper functions up one level, make them private?

CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxNodesFile.swift
```diff
@@ -78,6 +78,35 @@ func syntaxNode(nodesStartingWith: [Character]) -> SourceFileSyntax {
...
+          ) {
+            // Convenience initializer just calls the full initializer
+            // with certain child parameters specified as optional types
+            // and providing the rule-based default value for the affected
+            // parameters.
+            FunctionCallExprSyntax(
+              calledExpression: ExprSyntax("self.init"),
+              leftParen: .leftParenToken(),
+              arguments: LabeledExprListSyntax {
+                // generate the list of children for the call site.
```
natikgadzhi, line 99: That's where the list of parameters will go to pass them to the upstream init.

ahoppen:
I only glanced over it and the direction looks good. Before going into full-steam review on this, I would like to see a list of all the types that we want to apply rules to. Just to make sure that the rules are general enough to fit the needs and that there are sufficiently many that it’s easier to generate the code than to hand-write it.

natikgadzhi:
@ahoppen, thank you! Yeah, I got excited, and first trying to make the end to end prototype and not care about how clean the code is. 

Then, I'll post the list — I remember you're interested in looking at that first. I doubt I know all of the instances of where we'll want convenience initializers, so it's a possibility that the rules won't be flexible enough. 

I think once I have the list and the prototype, I'll groom this for the first review. I expect about 10 initializers, but perhaps I will think of more on my own, too.

ahoppen:
I think you’re doing a great job here and I’m all in favor of generating code instead of handwriting it.

natikgadzhi:
@ahoppen, the proof of concept works! The code in this PR successfully generates one initializer for `AttributeSyntax`. Not ready for a review yet, just celebrating a little win 🥳 

I've updated the initial comments with the list of next steps to emphasize getting the list of convenience init methods to generate first, then I'll clean up the code a little bit and split it into two commits — the infrastructure of the rules, and the list of actual nodes to process, so it's easier to review.

I'll take a day away from this to let this sit, I'll re-reading the code after will give me a fresh lens, I'll spot more mistakes.

natikgadzhi:
@ahoppen, I've got some interesting findings! 

- First off, I cleaned up the code a bit, and it's more tolerable now.
- I've added `ConvenienceInitRules` to replace my previous two PRs — so `EnumCaseParameterSyntax` and `ClosureCaptureSyntax` are now covered in generators instead of hand-written initializers, and those two work.
- Some others don't: `MacroExpansionDeclSyntax` and `FunctionCallExprSyntax` actually require their arguments to be non-optional, so while sure, we can add an initializer that also automatically sets the parentheses, we should not pass the arguments as optional to the default initializer.
   - Two possible paths forward: ignore them for now and do in a follow-up PR, or work in this one to improve the rule engine. I think that the initializer generator can check if the target `child.isOptional` originally to make things work.
   - Or we can get this to a state where we merge it, and improve later. I would recommend not doing that because seeing more edge cases will allow us to evaluate how good the rule engine is, and get it right from the start. But, my experience is limited 🙃 

Next steps:
- The code needs a rebase
- Let's do a review of what I have. I am not _entirely_ sure about naming, but I think it makes sense. Let's align on it.
- Once the code generally feels good, I will also add more rules for situations where `colon` is optional, and do that optionality check for the trigger argument. That's easy. 
- If we can get this done in 1-2 weeks, I think that would be a good state to merge this to not spend too much time on rebasing often, and then just add rules as we see more situations where they are useful.

ahoppen:
> Some others don't: `MacroExpansionDeclSyntax` and `FunctionCallExprSyntax` actually require their arguments to be non-optional, so while sure, we can add an initializer that also automatically sets the parentheses, we should not pass the arguments as optional to the default initializer.
> * Two possible paths forward: ignore them for now and do in a follow-up PR, or work in this one to improve the rule engine. I think that the initializer generator can check if the target `child.isOptional` originally to make things work.
> * Or we can get this to a state where we merge it, and improve later. I would recommend not doing that because seeing more edge cases will allow us to evaluate how good the rule engine is, and get it right from the start. But, my experience is limited 🙃

I think it really depends how many cases like this we have. If `MacroExpansionDeclSyntax` and `FunctionCallExprSyntax` are the only two nodes where we want to add parentheses if a collection is empty, I think it’s totally fine to just hand-write the convenience initializers instead of setting up the infrastructure to generate them. If we find more cases like this and it’s, say,  5 or more, I think introducing a rule for that would make sense.

PR #2143, opened 2023-09-02
Diagnostic format improvements

DougGregor:
Introduce a few improvements to make the "grouped" diagnostics more similar to the diagnostics provided by the existing Swift compiler, as a step toward enabling these diagnostics by default:

* Emit the "primary diagnostic" on the first line, matching the style of the existing compiler diagnostics and making it easier for humans tools to find the primary issue. 
* When the "primary diagnostic" is in a generated buffer, such as a macro expansion buffer, also show the outermost source file that triggered the generation of that buffer, so it's easy to find that.
* Only colorize the severity (`error`, `warning`, etc.) and not the whole diagnostic message, to match the Swift compiler's current style.
* Add `remark` diagnostic severity, which is used by the Swift compiler and needs to be representable.

Example now:
```
t.swift:5:3: error: no exact matches in call to global function 'f'
1 │ func f(a: Int) { }
  │      ╰─ note: incorrect labels for candidate (have: '(_:)', expected: '(a:)')
2 │ func f(b: Int) { }
  │      ╰─ note: incorrect labels for candidate (have: '(_:)', expected: '(b:)')
3 │ 
4 │ func g() {
5 │   f(1)
  │   ╰─ error: no exact matches in call to global function 'f'
6 │   f(a: 2.5)
7 │ }
```

### DIFF
--- a/Sources/SwiftCompilerPluginMessageHandling/PluginMessages.swift
+++ b/Sources/SwiftCompilerPluginMessageHandling/PluginMessages.swift
@@ -155,6 +155,7 @@ public enum PluginMessage {
       case error
       case warning
       case note
+      case remark
     }
     public struct Position: Codable {
       public var fileName: String
--- a/Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
+++ b/Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
@@ -321,26 +321,31 @@ public struct DiagnosticsFormatter {
   /// Annotates the given ``DiagnosticMessage`` with an appropriate ANSI color code (if the value of the `colorize`
   /// property is `true`) and returns the result as a printable string.
   func colorizeIfRequested(_ message: DiagnosticMessage) -> String {
-    switch message.severity {
-    case .error:
-      let annotation = ANSIAnnotation(color: .red, trait: .bold)
-      return colorizeIfRequested("error: \(message.message)", annotation: annotation)
-
-    case .warning:
-      let color = ANSIAnnotation(color: .yellow)
-      let prefix = colorizeIfRequested("warning: ", annotation: color.withTrait(.bold))
-
-      return prefix + colorizeIfRequested(message.message, annotation: color);
-    case .note:
-      return colorizeNoteIfRequested(message.message)
-    }
+    colorizeIfRequested(severity: message.severity, message: message.message)
   }
 
-  /// Annotate a note with an appropriate ANSI color code (if requested).
-  func colorizeNoteIfRequested(_ message: String) -> String {
-    let color = ANSIAnnotation(color: .default, trait: .bold)
-    let prefix = colorizeIfRequested("note: ", annotation: color)
-    return prefix + message
+  /// Annotates a diagnostic message with the given severity and text with an appropriate ANSI color code.
+  func colorizeIfRequested(severity: DiagnosticSeverity, message: String) -> String {
+    let severityText: String
+    let severityAnnotation: ANSIAnnotation
+
+    switch severity {
+    case .error:
+      severityText = "error"
+      severityAnnotation = .errorText
+
+    case .warning:
+      severityText = "warning"
+      severityAnnotation = .warningText
+
+    case .note:
+      severityText = "note"
+      severityAnnotation = .noteText
+    }
+
+    let prefix = colorizeIfRequested("\(severityText): ", annotation: severityAnnotation)
+
+    return prefix + colorizeIfRequested(message, annotation: .diagnosticText);
   }
 
   /// Apply the given color and trait to the specified text, when we are
@@ -417,5 +422,21 @@ struct ANSIAnnotation {
   /// Annotation used for highlighting source text.
   static var sourceHighlight: ANSIAnnotation {
     ANSIAnnotation(color: .default, trait: .underline)
+  }
+
+  static var diagnosticText: ANSIAnnotation {
+    ANSIAnnotation(color: .default, trait: .bold)
+  }
+
+  static var errorText: ANSIAnnotation {
+    ANSIAnnotation(color: .red, trait: .bold)
+  }
+
+  static var warningText: ANSIAnnotation {
+    ANSIAnnotation(color: .yellow, trait: .bold)
+  }
+
+  static var noteText: ANSIAnnotation {
+    ANSIAnnotation(color: .default, trait: .bold)
   }
 }
--- a/Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
+++ b/Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
@@ -176,20 +176,24 @@ public struct DiagnosticsFormatter {
     return resultSourceString
   }
 
+  struct PrimaryDiagnostic {
+    var location: SourceLocation
+    var diagnostic: Diagnostic
+  }
+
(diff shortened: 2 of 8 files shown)


DougGregor:
https://github.com/apple/swift/pull/68289

@swift-ci please test

DougGregor:
https://github.com/apple/swift/pull/68289
https://github.com/apple/swift-format/pull/605

@swift-ci please test

Sources/SwiftCompilerPluginMessageHandling/PluginMessages.swift
```diff
@@ -155,6 +155,7 @@ public enum PluginMessage {
       case error
       case warning
       case note
+      case remark
```
ahoppen, line 158: Could you add the addition of `PluginMessage.Diagnostic.Severity` to the [release notes](https://github.com/apple/swift-syntax/blob/main/Release%20Notes/510.md)?

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
```diff
@@ -176,20 +176,24 @@ public struct DiagnosticsFormatter {
     return resultSourceString
   }
 
+  struct PrimaryDiagnostic {
+    var location: SourceLocation
+    var diagnostic: Diagnostic
+  }
+
```
ahoppen, line 183: I don’t think you use `PrimaryDiagnostic`.

Sources/SwiftDiagnostics/GroupedDiagnostics.swift
```diff
@@ -133,6 +133,32 @@ extension GroupedDiagnostics {
     }
   }
 
+  // Find the "primary" diagnostic that will be shown at the top of the diagnostic
+  // message. This is typically the error, warning, or remark.
+  private func findPrimaryDiagnostic(in sourceFile: SourceFile) -> (SourceFile, Diagnostic)? {
```
ahoppen, line 138: Can a source file only have a single primary diagnostic? What if it has one error on line 10 and one on line 200?

Also: If there is a warning in line 5 and an error in line 7, I think the error should be the primary diagnostic, not the warning.

Sources/SwiftDiagnostics/GroupedDiagnostics.swift
```diff
@@ -133,6 +133,32 @@ extension GroupedDiagnostics {
...
+    // If there is a non-note diagnostic, it's the primary diagnostic.
+    if let primaryDiag = sourceFile.diagnostics.first(where: { $0.diagMessage.severity != .note }) {
```
ahoppen, line 140: Should we also check for remark here? Would it make sense that `Severity` is `Comparable` for that?

Sources/SwiftDiagnostics/GroupedDiagnostics.swift
```diff
@@ -164,11 +190,41 @@ extension GroupedDiagnostics {
...
+        let primaryDiagSLC = SourceLocationConverter(fileName: primaryDiagSourceFile.displayName, tree: primaryDiagSourceFile.tree)
+        let location = primaryDiag.location(converter: primaryDiagSLC)
+
+        // Display file/line/column and diagnostic text for the primary diagnostic.
+        prefixString = "\(location.file):\(location.line):\(location.column): \(formatter.colorizeIfRequested(primaryDiag.diagMessage))\n"
+
+        // If the primary diagnostic source file is not the same as the root source file, we're pointing into a generated buffer.
+        // Provide a link back to the original source file where this generated buffer occurred, so it's easy to find if
+        // (for example) the generated buffer is no longer available.
+        if sourceFile.id != primaryDiagSourceFile.id,
```
ahoppen, line 208: You can drop `.id` here. Equality of syntax nodes is defined in terms of `id`.

Sources/SwiftDiagnostics/GroupedDiagnostics.swift
```diff
@@ -164,11 +190,41 @@ extension GroupedDiagnostics {
...
+
+          if rootSourceID == sourceFileID {
+            let bufferLoc = slc.location(for: rootPosition)
+            let coloredMessage = formatter.colorizeIfRequested(severity: .note, message: "expanded code originates here")
+            prefixString += "╰─ \(bufferLoc.file):\(bufferLoc.line):\(bufferLoc.column): \(coloredMessage)\n"
+          }
+        }
+      } else {
+        let firstLine = sourceFile.diagnostics.first.map { $0.location(converter: slc).line } ?? 0
+        prefixString = "\(sourceFile.displayName): \(firstLine):"
```
ahoppen, line 225: In which case would we have non-empty `sourceFile.diagnostics` but wouldn’t find a primary diagnostic?

Sources/SwiftDiagnostics/Message.swift
```diff
@@ -30,6 +30,7 @@ public enum DiagnosticSeverity {
   case error
   case warning
   case note
+  case remark
```
ahoppen, line 33: Same here, could you add this to the release notes?

Sources/SwiftCompilerPluginMessageHandling/PluginMessages.swift
DougGregor: Sure, done

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
DougGregor: You're right, thanks! Removed.

Sources/SwiftDiagnostics/GroupedDiagnostics.swift
DougGregor: The paired change in the Swift compiler ensures that there is at most one non-note "primary diagnostic" per call to the diagnostic formatter. We can decide what to do for any clients that don't ensure this, which could be to take the most severe diagnostic or the first non-note diagnostic or something else.

Sources/SwiftDiagnostics/GroupedDiagnostics.swift
DougGregor: Remarks are treated like warnings or errors, and can have notes attached.

Sources/SwiftDiagnostics/GroupedDiagnostics.swift
DougGregor: Ah... this isn't `SourceFileSyntax`, but the `SourceFile` type used within grouped diagnostics.

Sources/SwiftDiagnostics/GroupedDiagnostics.swift
DougGregor: A file only containing notes, for example because the diagnostic shows up in one file but the notes point into another.

Sources/SwiftDiagnostics/Message.swift
DougGregor: Sure

DougGregor:
https://github.com/apple/swift/pull/68289
https://github.com/apple/swift-format/pull/605

@swift-ci please test

DougGregor:
https://github.com/apple/swift/pull/68289
https://github.com/apple/swift-format/pull/605

@swift-ci please test Windows

DougGregor:
https://github.com/apple/swift/pull/68289
https://github.com/apple/swift-format/pull/605

@swift-ci please test

PR #2145, opened 2023-09-02
Order all imports lexicographically

natikgadzhi:
## Motivation
- Part of #2136

## Changes
- Adds `OrderedImports` rule to `.swift-format`
- Removes a trailing space in `format.py`
- Formats the codebase to comply with `OrderedImports`

Issue #2136, opened 2023-09-01
Enable the ordered imports rule in swift-format


ahoppen:
AFAIK swift-format has a rule to force import statements to be ordered lexicographically. But that rule doesn’t appear to be enabled in this repository. W should enable it.

natikgadzhi:
Mind if I use this one as a reason to read more about swift-format rules and set it up?

ahoppen:
Sure, go ahead.

Matejkob:
I recall that in one of the PRs with @kimdv  and @ahoppen , we discussed how sometimes we skip the `return` keyword in the codebase, and sometimes we don't. At that time, `swift-format` didn't have a rule for this, but that's changed. There's now a rule to omit the `return` keyword whenever possible. 

So the question is, do we want to adopt this rule?

ahoppen:
Tracked in Apple’s issue tracker as rdar://114841762

ahoppen:
It would definitely be interesting to see what the code looks like with the implicit return rule enabled. I think I commented somewhere that I‘m personally not a fan of omitting the return but consistency is more important than my personal preference.

natikgadzhi:
#2145 adds `OrderedImports`, but I'll push up another commit to it with implicit returns so we can take a look at the code and see if we like it.

Interesting finds: 
- `swift-format` will not complain if you make a typo in a rule name. I made a typo `OrderedImpora` and ran `./format.py` and it didnt' output anything, and didn't change any files, of course — would be nice to have a warning form `swift-format`. I'll file an issue and make a PR.
- `swift-format dump-configuration` dumps the default json config, but does not explain what each rule does. It would be nice to have a comment with each of the config settings that summarizes how it works. Again, easy PR probably.

natikgadzhi:
Follow-up swift-format issues:
- Comments in dump-configuration: https://github.com/apple/swift-format/issues/606
- Detect mistakes in the Rules configuration: https://github.com/apple/swift-format/issues/607

natikgadzhi:
So, the ordered imports rule PR is merged in (#2145). Since omitting returns is unlikely to get merged, I think we should close this issue, @ahoppen.

Both follow-up issues in `swift-format` have pending pull requests, and I think those are going to get merged in later this week, too. 🥳

kimdv:
Thanks @natikgadzhi 😁

Resolved by #2145

PR #2146, opened 2023-09-02
Omit explicit returns where possible

natikgadzhi:
## Motivation
A separate formatting PR to make space for @Matejkob and @ahoppen and @kimdv to review how our codebase would look with omitting `return` statements where it's possible, a follow-up to #2136. Since this topic might spawn a longer discussion, I've put that rule in a separate PR.

## Changes
- Adds `OmitExplicitReturns` `swift-format` rule to `.swift-format`
- Formats the codebase
- Follow-up to #2145, but in a separate commit for easier review.

### DIFF
--- a/.swift-format
+++ b/.swift-format
@@ -10,6 +10,7 @@
         "AlwaysUseLowerCamelCase": false,
         "AmbiguousTrailingClosureOverload": false,
         "NoBlockComments": false,
+        "OmitExplicitReturns": true,
         "OrderedImports": true,
         "UseLetInEveryBoundCaseVariable": false,
         "UseSynthesizedInitializer": false
--- a/CodeGeneration/Sources/SyntaxSupport/Child.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/Child.swift
@@ -108,7 +108,7 @@ public class Child {
 
   /// A name of this child that's suitable to be used for variable or enum case names.
   public var varOrCaseName: TokenSyntax {
-    return .identifier(lowercaseFirstWord(name: name))
+    .identifier(lowercaseFirstWord(name: name))
   }
 
   /// If this child has node choices, the type that the nested `SyntaxChildChoices` type should get.
@@ -137,7 +137,7 @@ public class Child {
 
   /// Determines if this child has a deprecated name
   public var hasDeprecatedName: Bool {
-    return deprecatedName != nil
+    deprecatedName != nil
   }
 
   /// If the child ends with "token" in the kind, it's considered a token node.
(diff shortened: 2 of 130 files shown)


kimdv:
Personally I'm not in favor of omitting returns.
I think it makes it harder to read code.
I known that the SwiftSyntaxBuilder in code gen don't use, just like SwiftUI. But for me it makes sense and is fine.

But a normal variable or function need a return IMO

natikgadzhi:
I'm only a beginner in Swift, but so far omitting returns _feels_ good in:
- one-line closures
- result builders
- one-line / one-statement dynamic property bodies
- perhaps functions that don't have any control flow in them and are very short (one statement)

But, I agree that making this a mandated format might not work well in situations where explicit return keywords make the code much more readable. 

While a lot of swift-syntax code are small functions, and they look pretty nice without `return` to a newcomer like myself, I don't have a strong opinion, and don't have the experience to back up my opinions even if I had one. 

So, that PR is for demonstration purposes. If we chose to enable the rule, we can do so at any point in the future.

xedin:
Just a general note about `return` omission. As per proposal, `return` statements could only be omitted if body constitutes of a single expression, anything more context always requires an explicit `return`.

natikgadzhi:
@ahoppen, I _think_ we should probably close this for now? We know this happened, and if we want to return to that — the PR will be here. @Matejkob, good with you too?

Matejkob:
I would personally advocate for merging. I might be a bit biased since I'm accustomed to omitting the `return` keyword in my projects. I also appreciate the consistency it brings. We already have numerous instances where `return` is omitted, so implementing this rule would contribute to overall codebase consistency. However, I'm just a random individual and don't really feel in a position to strongly push for my approach.

natikgadzhi:
> However, I'm just a random individual and don't really feel in a position to strongly push for my approach.

Big same here.

natikgadzhi:
I've rebased this PR so we can review this with ordered imports and other improvements merged in. 

It's growing on me, to be honest. Should I perhaps start a topic in swift forums to pitch this and get more feedback from the community, @ahoppen?

ahoppen:
I’ll gather some feedback on this internally and come back to you in the next week-ish. I’m also very hesitant to merge this as-is.

ahoppen:
I just had another conversation about this and we don’t want this change for now. If there was an official Swift API design guideline to follow regarding when to omit `return`, we should definitely adopt it in swift-syntax but for now we believe that enforcing the omission of `return` in general makes the code less readable.

Issue #2109, opened 2023-08-28
Any word inside a class, struct or enum without the keyword func is considered as FunctionDecl


HassanTaleb90:
Why when we type **any word** inside a class, struct or enum without the keyword `func` is considered as **FunctionDecl**?

```
///Bug
class MyClass {
  word1 //FunctionDecl
}

enum MyEnum {
  word2 //FunctionDecl
}

struct MyStruct {
  word3 //FunctionDecl
}

///Normal:
class MyClass {
  func word1 //FunctionDecl
}

enum MyEnum {
  func word2 //FunctionDecl
}

struct MyStruct {
  func word3 //FunctionDecl
}
```

allevato:
I transferred this to swift-syntax, since it's a parser question.

ahoppen:
Tracked in Apple’s issue tracker as rdar://114559579

ahoppen:
This is the parser recovery kicking in. We want to represent the code as something and we judge that a function declaration is the most likely, since functions are the most comment members of a type. What would you have expected?

HassanTaleb90:
> This is the parser recovery kicking in. We want to represent the code as something and we judge that a function declaration is the most likely, since functions are the most comment members of a type. What would you have expected?

I think that just writing the word should not give priority to any possibility, until there are some conditions that support this possibility, such as the presence of one parenthese after the word or the presence of the func keyword before it. Until these conditions exist, it can be considered as an UnknownExpr.

From here let me ask the same question to the word we write outside of class, structure or enumeration. Why is it considered an IdentifierExpr and not a FunctionDecl?

ahoppen:
> I think that just writing the word should not give priority to any possibility, until there are some conditions that support this possibility, such as the presence of one parenthese after the word or the presence of the func keyword before it. Until these conditions exist, it can be considered as an UnknownExpr.

OK, I see a point that we should parse this as `MissingDeclSyntax`, making `word1` an unexpected token within that node.

> From here let me ask the same question to the word we write outside of class, structure or enumeration. Why is it considered an IdentifierExpr and not a FunctionDecl?

That’s because a word at the top level of a function is actually syntactically valid if you file allows top-level code. For example

```swift
let someWord = 1
someWord
```

is a valid Swift program.

mininny:
@ahoppen 
I thought this was interesting and tried to resolve some of the concerns in this issue here: https://github.com/apple/swift-syntax/pull/2149

[...]

PR #2149, opened 2023-09-04
Limit conditions for inferencing a possible FuncDecl

mininny:
Follow-up from https://github.com/apple/swift-syntax/issues/2109

```
class MyClass {
  word1 // Parsed as FunctionDecl
}
```

Currently, the parser consumes the `word1` token as an incomplete function declaration since most declarations within a MemberDeclList are functions. 

However, without more information around the token like `func` or `()` which would indicate that it is a possible FuncDecl, it could also be other declarations, like `let`, `var`, or even something like `typealias`.

This PR peeks the tokens after the identifier and checks if it has `(` or `<` which would indicate that it's a function. 

Any freestanding identifiers as shown in the example above would be parsed as `MissingDeclSyntax`, which spits out `unexpected 'word1' in MyClass`.

This would affect existing test cases that rely on the previous behavior and change the resulting diagnostics.

---

I've validated that this change is compatible with most test cases, but I'm not sure how(if I'm allowed to) to fix existing test cases, and add test cases(because many of them are translations from swift repo)

Also, I think a better diagnostic in the example above is something like: ``expected keyword before `word1`. did you mean to add `func` or `var`?``, but I'm not yet sure how I can feed the parsed information from Parser to Diagnostics.

### DIFF
--- a/Sources/SwiftParser/Declarations.swift
+++ b/Sources/SwiftParser/Declarations.swift
@@ -281,9 +281,10 @@ extension Parser {
         )
       }
 
-      let isProbablyFuncDecl = self.at(.identifier, .wildcard) || self.at(anyIn: Operator.self) != nil
-
-      if isProbablyFuncDecl {
+      let isPossibleFuncIdentifier = self.at(.identifier, .wildcard)
+      let isPossibleFuncParen = self.peek(isAt: .leftParen, .binaryOperator)
+      // Treat operators specially because they're likely to be functions.
+      if (isPossibleFuncIdentifier && isPossibleFuncParen) || self.at(anyIn: Operator.self) != nil {
         return RawDeclSyntax(self.parseFuncDeclaration(attrs, .missing(.keyword(.func))))
       }
     }
--- a/Tests/SwiftParserTest/DeclarationTests.swift
+++ b/Tests/SwiftParserTest/DeclarationTests.swift
@@ -91,6 +91,61 @@ final class DeclarationTests: ParserTestCase {
         type: IdentifierTypeSyntax(name: .identifier("Int"))
       )
     )
+      
+    assertParse(
+      """
+      class MyClass {
+        1️⃣foo()
+      }
+      """,
+      diagnostics: [
+        DiagnosticSpec(
+          locationMarker: "1️⃣",
+          message: "expected 'func' in function",
+          fixIts: ["insert 'func'"]
+        )
+      ],
+      fixedSource: """
+      class MyClass {
+        func foo()
+      }
+      """
+    )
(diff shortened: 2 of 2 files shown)


Sources/SwiftParser/Declarations.swift
```diff
@@ -281,9 +281,11 @@ extension Parser {
         )
       }
 
-      let isProbablyFuncDecl = self.at(.identifier, .wildcard) || self.at(anyIn: Operator.self) != nil
+      let isPossibleFuncIdentifier = self.at(.identifier, .wildcard)
+      let isPossibleFuncParen = self.peek(isAt: .identifier, .leftParen) || self.peek(isAt: .identifier, .leftAngle)
```
ahoppen, line 285: You should be able to shorten this to

```suggestion
      let isPossibleFuncParen = self.peek(isAt: .identifier, .leftParen, .leftAngle)
```

ahoppen (commented):
Thanks for the PR @mininny. The changes look good to me and I quite like the `unexpected 'word1'` diagnostic because at this point you can just view `word1` as garbage text inside the member decl list.

And you can definitely change the existing tests. If the diagnostics improve, it’s totally fine to change their expected behavior.

Sources/SwiftParser/Declarations.swift
mininny: Thanks! I shortened it. However, while testing it out, I found that in
```
class MyClass { 
    foo<Int>
}
```
`<` is parsed as a binaryOperator rather than a left angle. I think this might need some follow-up item that should be looked at.

mininny:
> Thanks for the PR @mininny. The changes look good to me and I quite like the `unexpected 'word1'` diagnostic because at this point you can just view `word1` as garbage text inside the member decl list.
> 
> And you can definitely change the existing tests. If the diagnostics improve, it’s totally fine to change their expected behavior.

Thanks for the review and comment! I've added some tests and marked the PR as ready for review. 

I have another question; when adding tests, is it okay to add tests to `SwiftParserTests/translated`? I see that it's translated from the original swift repo.

ahoppen (approved):
Thanks. LGTM

ahoppen:
@swift-ci Please test

ahoppen:
It looks like your changes aren’t formatted. Could you run `format.py`?

mininny:
> It looks like your changes aren’t formatted. Could you run `format.py`?

Yep, I ran the the format.py :+1:

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #2150, opened 2023-09-04
Handle pattern as editor placeholder

### LINKED ISSUE #2147: Sometimes the placeholder is identified as IdentifierPattern
### Description

Sometimes the placeholder is identified as IdentifierPattern and not as EditorPlaceholderExpr.

### Steps to Reproduce

```
let <#name#> = <#value#>
```
<#name#> //Bug:  is identified as `IdentifierPattern`
<#value#> //Normal: is identified as `EditorPlaceholderExpr`

### LINKED ISSUE #2156: Support parsing of editor placeholders in type positions
### Issue Kind

Other

### Source Code

```swift
let a: <#T#> = x
```

### Description

`<#T#>` currently gets parsed as a `NamedOpaqueReturnTypeSyntax`. We should create a `EditorPlaceholderTypeSyntax` node and parse `<#T#>` as an editor placeholder.

kimdv:
Fixes #2147
Fixes #2156

### DIFF
--- a/CodeGeneration/Sources/SyntaxSupport/PatternNodes.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/PatternNodes.swift
@@ -182,4 +182,22 @@ public let PATTERN_NODES: [Node] = [
     ]
   ),
 
+  Node(
+    kind: .editorPlaceholderPattern,
+    base: .pattern,
+    nameForDiagnostics: "editor placeholder",
+    documentation: """
+      An editor placeholder, e.g. `<#pattern#>` that is used in a position that expects a pattern.
+      """,
+    children: [
+      Child(
+        name: "placeholder",
+        kind: .token(choices: [.token(.identifier)]),
+        documentation: """
+          The actual editor placeholder that starts with `<#` and ends with `#>`.
+          """
+      )
+    ]
+  ),
+
 ]
--- a/Sources/SwiftParser/Patterns.swift
+++ b/Sources/SwiftParser/Patterns.swift
@@ -70,12 +70,21 @@ extension Parser {
       )
     case (.lhs(.identifier), let handle)?:
       let identifier = self.eat(handle)
-      return RawPatternSyntax(
-        RawIdentifierPatternSyntax(
-          identifier: identifier,
-          arena: self.arena
+      if identifier.tokenText.isEditorPlaceholder {
+        return RawPatternSyntax(
+          RawEditorPlaceholderPatternSyntax(
+            placeholder: identifier,
+            arena: self.arena
+          )
         )
-      )
+      } else {
+        return RawPatternSyntax(
+          RawIdentifierPatternSyntax(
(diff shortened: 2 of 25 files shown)


Sources/SwiftParser/Patterns.swift
```diff
@@ -70,12 +70,26 @@ extension Parser {
...
-      return RawPatternSyntax(
-        RawIdentifierPatternSyntax(
-          identifier: identifier,
-          arena: self.arena
+      if identifier.tokenText.isEditorPlaceholder {
+        return RawPatternSyntax(
+          RawExpressionPatternSyntax(
+            expression: RawExprSyntax(
+              RawEditorPlaceholderExprSyntax(
+                placeholder: identifier,
```
kimdv, line 78: The thing here is that the `identifier` includes `<#` and `#>`. 
Not sure how to come around that

ahoppen: For now, that’s intentional. I don’t think that splitting the token into `<#`, the contents and `#>` provides too much value for now.

ahoppen (commented):
Instead of creating an `ExpressionPatternSyntax`  that contains a `EditorPlaceholderExprSyntax`, I would prefer to introduce a `EditorPlaceholderPatternSyntax` type that we can use here.

CodeGeneration/Sources/SyntaxSupport/PatternNodes.swift
```diff
@@ -182,4 +182,23 @@ public let PATTERN_NODES: [Node] = [
...
+    kind: .editorPlaceholderPattern,
+    base: .pattern,
+    nameForDiagnostics: "editor placeholder",
+    documentation: """
+      An editor placeholder, e.g. `<#pattern#>` that is used in a position that expects a pattern.
+      """,
+    children: [
+      Child(
+        name: "placeholder",
+        deprecatedName: "identifier",
```
ahoppen, line 195: This is a new node, so it doesn’t have to have a `deprecatedName`. Same for `EditorPlaceholderTypeSyntax`.

Tests/SwiftParserTest/TypeTests.swift
```diff
@@ -216,4 +216,31 @@ final class TypeTests: ParserTestCase {
...
+              )
+            )
+          )
+        ]
+      ),
+      diagnostics: [
+        DiagnosticSpec(message: "editor placeholder in source file")
+      ]
+    )
+  }
```
ahoppen, line 245: Could you add a test case for an editor placeholder with generic arguments?

CodeGeneration/Sources/SyntaxSupport/PatternNodes.swift
kimdv: 🤦‍♂️

kimdv:
@swift-ci please test

ahoppen (approved):

PR #2151, opened 2023-09-04
Make keyword  precedence exhaustive

### LINKED ISSUE #2072: Make the switch over Keywords in TokenPrecedence.swift exhaustive
Fix this To Do by making the switch exhaustive. We should consider all the keywords that we add cases for and decide which token precedence they should get.

https://github.com/apple/swift-syntax/blob/2e3c42cf38defd998c87ecfe8df138f925d22736/Sources/SwiftParser/TokenPrecedence.swift#L236

kimdv:
Fixes #2072


Sources/SwiftParser/TokenPrecedence.swift
```diff
@@ -192,47 +192,266 @@ enum TokenPrecedence: Comparable {
...
+      .__setter_access, .indirect, .isolated, .nonisolated, .distributed, .unchecked, ._local,
+      .inout, ._mutating, ._borrow, ._borrowing, .borrowing, ._consuming, .consuming, .consume,
+      ._backDeploy, .backDeployed, .inline, ._specialize, .availability, .unavailable, .deprecated, ._spi,
       // Misc
       .import:
       self = .declKeyword
-    default:
-      // Treat all keywords that weren't handled above as expression keywords as a fallback option.
-      // FIXME: We should assign a token precedence to all keywords
+
```
kimdv, line 239: Just a general question. 
Are all keyword still used or are there leftovers from previous major Swift versions? 

If yes, can some of those be removed?

kimdv:
Just a general question. 
Are all keyword still used or are there leftovers from previous major Swift versions? 

If yes, can some of those be removed?

Sources/SwiftParser/TokenPrecedence.swift
```diff
@@ -192,47 +192,266 @@ enum TokenPrecedence: Comparable {
     case  // Literals
     .Self, .false, .nil, .`self`, .super, .true:
       self = .identifierLike
+
     // MARK: Expr keyword
     case  // Keywords
-    .as, .is, .try,
+    .as, .is, .some, .try,
+      .await, .reasync, .async,
```
ahoppen, line 199: Could you add `reasync` and `async` on the same line as `regrows` and `throws`?

Sources/SwiftParser/TokenPrecedence.swift
```diff
@@ -192,47 +192,266 @@ enum TokenPrecedence: Comparable {
...
       // We don't know much about which contextual keyword it is, be conservative an allow considering it as unexpected.
       // Keywords in function types (we should be allowed to skip them inside parenthesis)
       .rethrows, .throws,
       // Consider 'any' a prefix operator to a type and a type is expression-like.
       .Any,
       // 'where' can only occur in the signature of declarations. Consider the signature expression-like.
       .where,
+      .each,
```
ahoppen, line 207: I would add `each` right after `try` and `await` because it’s a similar kind of trivia expression modifier.

Sources/SwiftParser/TokenPrecedence.swift
```diff
@@ -192,47 +192,266 @@ enum TokenPrecedence: Comparable {
...
       // Secondary parts of control-flow constructs
       .case, .catch, .default, .else,
       // Return-like statements
       .break, .continue, .fallthrough, .return, .throw, .then, .yield:
       self = .stmtKeyword
+
     // MARK: Decl keywords
     case  // Types
-    .associatedtype, .class, .enum, .extension, .protocol, .struct, .typealias,
+    .associatedtype, .class, .enum, .extension, .protocol, .struct, .typealias, .actor, .Sendable,
```
ahoppen, line 222: `Sendable` is an attribute here and not part of a type declaration and thus doesn’t really fit here. I would just place it in the other section.

Sources/SwiftParser/TokenPrecedence.swift
```diff
@@ -192,47 +192,266 @@ enum TokenPrecedence: Comparable {
...
       // Access modifiers
       .fileprivate, .internal, .private, .public, .static,
       // Functions
       .deinit, .func, .`init`, .subscript,
       // Variables
-      .let, .var,
+      .let, .var, .weak,
```
ahoppen, line 228: `weak` should be an `exprKeyword` for now since we can use it in closure capture items and in that context we are more in an expression-style context than a context that introduces new declarations.

Sources/SwiftParser/TokenPrecedence.swift
```diff
@@ -192,47 +192,266 @@ enum TokenPrecedence: Comparable {
...
       // Operator stuff
       .operator, .precedencegroup,
       // Declaration Modifiers
       .__consuming, .final, .required, .optional, .lazy, .dynamic, .infix, .postfix, .prefix, .mutating, .nonmutating, .convenience, .override, .package, .open,
-      .__setter_access, .indirect, .nonisolated, .distributed, ._local,
-      .inout, ._mutating, ._borrowing, ._consuming,
+      .__setter_access, .indirect, .isolated, .nonisolated, .distributed, .unchecked, ._local,
+      .inout, ._mutating, ._borrow, ._borrowing, .borrowing, ._consuming, .consuming, .consume,
+      ._backDeploy, .backDeployed, .inline, ._specialize, .availability, .unavailable, .deprecated, ._spi,
```
ahoppen, line 235: What was the reasoning to change these to `declKeyword`? For example `_spi` should be an `exprKeyword` since it’s a type attribute name and when parsing a type attribute we are usually looking for an identifier that specifies the variable name, which is not an introducer of a new declaration.

That being said, I think it might make sense to open a section for the cases in `TypeAttribute` and one for the cases in `DeclarationAttributeWithSpecialSyntax` just to group them together.

Sources/SwiftParser/TokenPrecedence.swift
```diff
@@ -192,47 +192,266 @@ enum TokenPrecedence: Comparable {
...
+      self = .exprKeyword
+    case .visibility:
+      self = .exprKeyword
+    case .willSet:
+      self = .exprKeyword
+    case .witness_method:
+      self = .exprKeyword
+    case .wrt:
+      self = .exprKeyword
+    case .unsafe:
```
ahoppen, line 454: I would prefer to have a single case for all of these that basically says: 

```swift
/// Treat all other keywords as expression keywords in the absence of any better information.
```

Sources/SwiftParser/TokenPrecedence.swift
```diff
@@ -192,47 +192,266 @@ enum TokenPrecedence: Comparable {
...
+    case .before:
+      self = .exprKeyword
+    case .block:
+      self = .exprKeyword
+    case .canImport:
+      self = .exprKeyword
+    case .compiler:
+      self = .exprKeyword
+    case .copy:
+      self = .exprKeyword
```
ahoppen, line 347: `copy` one could also belong in the same section as `try` and `await`.

Sources/SwiftParser/TokenPrecedence.swift
```diff
@@ -192,47 +192,266 @@ enum TokenPrecedence: Comparable {
...
+    case .convention:
+      self = .exprKeyword
+    case .cType:
+      self = .exprKeyword
+    case .derivative:
+      self = .exprKeyword
+    case .didSet:
+      self = .exprKeyword
```
ahoppen, line 355: I think all the accessors (`AccessorDeclSyntax`) should have a `declKeyword` precedence since they start an `AccessorDeclSyntax`, which is a declaration. Doesn’t have to be part of this PR though.

Sources/SwiftParser/TokenPrecedence.swift
```diff
@@ -192,47 +192,266 @@ enum TokenPrecedence: Comparable {
...
+    case .left:
+      self = .exprKeyword
+    case .line:
+      self = .exprKeyword
+    case .linear:
+      self = .exprKeyword
+    case .lowerThan:
+      self = .exprKeyword
+    case .macro:
+      self = .exprKeyword
```
ahoppen, line 389: `macro` should definitely be a `declKeyword`

ahoppen (commented):
Thanks for doing this @kimdv. I have some opinions on the grouping and classification 😉

Sources/SwiftParser/TokenPrecedence.swift
kimdv: Added in 3e14cfe3f9575fc8c1e92af5c1436d1635fdedd5

Sources/SwiftParser/TokenPrecedence.swift
kimdv: Added a new group for `DeclarationAttributeWithSpecialSyntax`.
Not sure tough if it have the right precedence. 
Name can also be discussed 😬

Question: Why is `.rethrows` in here? Does that have special syntax?

b4c0a3d7dada89146f2f8f615db9fc96e2371c42

kimdv: Added new `typeKeyword` group in 59d1c9383174c4bb2e4873b30c80ed3d16d0efd9

Again here some keyword are already used. 
Not sure where to place them. 


Same if they have the right precedence.

ahoppen: Oh, I think you misunderstood me. I didn’t mean to introduce a new precedence level, just to have a separate `case` clause in the `switch` for all the `TypeAttribute`s and cases in `DeclarationAttributeWithSpecialSyntax`.

> Question: Why is .rethrows in here? Does that have special syntax?

There’s the `@rethrows` and I think we would be rejecting that otherwise because the parser expects an identifier after `@` and `rethrows` is a laxer-classified keyword.

kimdv: Oh sorry about that. 
I've added them as decl. keyword now. 
But I'm not sure where the redundant keyword should be placed tough? 

1. `.async` and `._local` belongs both to `declKeyword`. 
    But what group to you think it's the best fit?
2. `rethrows` and `differentiable` is also in two groups.
    Again, what do you think is the best fit?

ahoppen: If in doubt, I would put them in the non-attribute section. IMO the `TypeAttribute` and `DeclarationAttributeWithSpecialSyntax` sections are just to keep the `Other` group slightly smaller.

Also, I would change `TypeAttribute` and `DeclarationAttributeWithSpecialSyntax` to return `exprKeyword` instead of `declKeyword`. That’s what I meant to say by

> For example `_spi` should be an `exprKeyword` since it’s a type attribute name and when parsing a type attribute we are usually looking for an identifier that specifies the variable name, which is not an introducer of a new declaration.

kimdv: Ah okay, think it starts to make sense now. 
Thanks, and sorry for the confusion 😁

ahoppen: Not a problem at all. Thanks for doing this.

ahoppen (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

PR #2159, opened 2023-09-06
Fix a string interpolation error in CodeGeneration

ahoppen:
`leafProtocolDecl` placed two declarations within the same `DeclSyntax`, which didn’t parse correctly, triggering `logStringInterpolationParsingError`. Split the function so that one returns the protocol declaration and the other returns the extension.


ahoppen:
@swift-ci Please test

Matejkob:
Hmmm, that's interesting. So if we wanna create two functions do need two different `DeclSyntax`?
```swift
DeclSyntax(
"""
func foo() { }
func bar() { }
"""
)
```
vs.
```swift
DeclSyntax(
"""
func foo() { }
"""
DeclSyntax(
"""
func bar() { }
"""
)
```
?

ahoppen:
Yes, a `DeclSyntax` only represents a single declaration. To represent a list of declarations, we should use [`CodeBlockItemListSyntax`](https://swiftpackageindex.com/apple/swift-syntax/main/documentation/swiftsyntax/codeblockitemlistsyntax) or [`MemberBlockItemListSyntax`](https://swiftpackageindex.com/apple/swift-syntax/main/documentation/swiftsyntax/memberblockitemlistsyntax).

PR #2160, opened 2023-09-06
Remove the last lit test

ahoppen:
We only had a single lit test left, which was testing the integration of the macro plugin with the compiler. That testing should be covered by us building the macro examples in the `Examples` package, so it’s not needed anymore.

rdar://114469622


ahoppen:
https://github.com/apple/swift/pull/68364

@swift-ci Please test

bnbarham (approved):

kimdv (approved):

Matejkob (commented):
I'm wondering if we still need the [`ExamplePlugin` target in the `Examples` package](https://github.com/apple/swift-syntax/tree/main/Examples/Sources/ExamplePlugin) - was it mainly used for lit tests, and is it now a bit redundant with `MacroExamples`?

ahoppen:
Yes, I think merging the two targets would make sense. Could you be interested in doing that, @Matejkob?

Matejkob:
Sure, I've added an issue for that.

PR #2162, opened 2023-09-07
build: set the `CMAKE_RUNTIME_OUTPUT_DIRECTORY`

compnerd:
Add a `CMAKE_RUNTIME_OUTPUT_DIRECTORY` to co-locate the runtime components which is important for resolving the interdependencies. Without this the libraries are not really usable.


compnerd:
@swift-ci please test

compnerd:
5.9 support for macros is complete and this should not be needed.

PR #2163, opened 2023-09-07
Enhancements to Macro Examples: Unit Tests and Code Organization

Matejkob:
This PR introduces a set of improvements to the macro examples, including the addition of unit tests, the categorization of macro interfaces, and the restructuring of the directory to make it more accessible and navigable.

Changes:
* **Add Unit Tests to Macro Examples.**
* Categorize Macro Interfaces and Standardize Playground: I've categorized macro interfaces and added `MARK` comments. This standardization not only cleans up the code but also makes it easier to find your way around.
* Organize Macro Examples into Type-Specific Directories: To streamline the search process for newcomers, I've restructured macro examples into type-specific directories.


Examples/Sources/MacroExamples/Implementation/Expression/FontLiteralMacro.swift
```diff
@@ -40,6 +40,9 @@ private func replaceFirstLabel(
   }
 
   var tuple = tuple
-  tuple[tuple.startIndex] = firstElement.with(\.label, .identifier(newLabel))
+  tuple[tuple.startIndex] =
+    firstElement
+    .with(\.label, .identifier(newLabel))
+    .with(\.colon, .colonToken())
```
ahoppen, line 46: My preferred way of spelling this is to actually modify the node instead of using `with`. In my experience it shortens the code and makes it easier to read.

```swift
tuple[tuple.startIndex].label = .identifier(newLabel)
tuple[tuple.startIndex].colon = .colonToken()
```

And yes, I know that the current examples use `with` everywhere. If you’d be interested in migrating many of those cases to variable setters, that would be amazing. I wouldn’t make it part of this PR though.

Examples/Sources/MacroExamples/Playground/main.swift
ahoppen, line 1: What do you think about splitting this file into multiple files as well along the same lines as the files in the *Interface* directory?

Examples/Tests/MacroExamples/Implementation/ComplexMacros/DictionaryIndirectionMacroTests.swift
```diff
@@ -0,0 +1,108 @@
...
+              get {
+                      _storage["y", default: 2] as! Int
+                }
+              set {
+                          _storage["y"] = newValue
+                }
+          }
+
+            var _storage: [String: Any] = [:]
+        }
```
ahoppen, line 53: Oh, this formatting is disgusting. You should be able to fix it by changing `DictionaryStoragePropertyMacro` to return

```swift
return [
  """
  get {
    _storage[\(literal: identifier.text), default: \(defaultValue)] as! \(type)
  }
  """,
  """
  set {
    _storage[\(literal: identifier.text)] = newValue
  }
  """,
]
```

Examples/Tests/MacroExamples/Implementation/ComplexMacros/DictionaryIndirectionMacroTests.swift
```diff
@@ -0,0 +1,108 @@
...
+        let x: Int
+        let y: Int
+      }
+      """,
+      expandedSource: """
+        class Point {
+          let x: Int
+          let y: Int
+
+            var _storage: [String: Any] = [:]
```
ahoppen, line 73: Since the input to the test files use 2 spaces for indentation, we should pass `indentationWidth: .spaces(2)` to `assertMacroExpansion` so that the expanded source looks nice as well. This applies to all the new test cases.

Examples/Tests/MacroExamples/Implementation/Member/CustomCodableTests.swift
```diff
@@ -0,0 +1,72 @@
...
+      """,
+      expandedSource: """
+        struct Person {
+          let name: String
+          let age: Int
+
+            enum CodingKeys: String, CodingKey {
+                    case name
+                case age
+              }
```
ahoppen, line 38: You should be able to fix this formatting by changing `CustomCodable` to return

```swift
let codingKeys: DeclSyntax = """
  enum CodingKeys: String, CodingKey {
  \(raw: cases.joined(separator: "\n"))
  }
  """
```

Examples/Tests/MacroExamples/Implementation/Member/MetaEnumMacroTests.swift
```diff
@@ -0,0 +1,120 @@
...
+                        case .integer:
+                            self = .integer
+                        case .text:
+                            self = .text
+                        case .boolean:
+                            self = .boolean
+                        case .null:
+                            self = .null
+                        }
+                    }
```
ahoppen, line 55: You should be able to fix this formatting by changing MetaEnumMacro to

```swift
  func makeMetaEnum() -> DeclSyntax {
    // FIXME: Why does this need to be a string to make trailing trivia work properly?
    let caseDecls = childCases.map { childCase in
      "  case \(childCase.name)"
    }.joined(separator: "\n")

    return """
      \(access)enum Meta {
      \(raw: caseDecls)
      \(makeMetaInit())
      }
      """
  }

  func makeMetaInit() -> DeclSyntax {
    // FIXME: Why does this need to be a string to make trailing trivia work properly?
    let caseStatements = childCases.map { childCase in
      """
        case .\(childCase.name):
          self = .\(childCase.name)
      """
    }.joined(separator: "\n")

    return """
      \(access)init(_ \(parentParamName): \(parentTypeName)) {
        switch \(parentParamName) {
      \(raw: caseStatements)
        }
      }
      """
  }
```

Examples/Tests/MacroExamples/Implementation/Peer/AddAsyncMacroTests.swift
```diff
@@ -0,0 +1,131 @@
...
+
+              switch returnValue {
+          case .success(let value):
+              continuation.resume(returning: value)
+          case .failure(let error):
+              continuation.resume(throwing: error)
+          }
+
+            }
+          }
```
ahoppen, line 46: You should be able to fix this formatting by changing the implementation of `AddAsyncMacro` to 

```swift
    let switchBody: ExprSyntax =
      """
            switch returnValue {
            case .success(let value):
              continuation.resume(returning: value)
            case .failure(let error):
              continuation.resume(throwing: error)
            }
      """

    let newBody: ExprSyntax =
      """

        \(raw: isResultReturn ? "try await withCheckedThrowingContinuation { continuation in" : "await withCheckedContinuation { continuation in")
          \(raw: funcDecl.name)(\(raw: callArguments.joined(separator: ", "))) { \(raw: returnType != nil ? "returnValue in" : "")

      \(raw: isResultReturn ? switchBody : "continuation.resume(returning: \(raw: returnType != nil ? "returnValue" : "()"))")

          }
        }

      """
```

ahoppen (commented):
Thanks a lot @Matejkob, this makes the examples a lot easier to understand. I’ve got a couple of suggestions, mostly to fix formatting, inline. 

@DougGregor I would appreciate if you could take a look at the new test cases to make sure they are actually testing the intended behavior of the macros.

Examples/Sources/MacroExamples/Implementation/Expression/FontLiteralMacro.swift
```diff
@@ -16,13 +16,13 @@ import SwiftSyntaxMacros
 /// Implementation of the `#fontLiteral` macro, which is similar in spirit
 /// to the built-in expressions `#colorLiteral`, `#imageLiteral`, etc., but in
 /// a small macro.
-public struct FontLiteralMacro: ExpressionMacro {
+public enum FontLiteralMacro: ExpressionMacro {
```
kimdv, line 19: Not dived that much into macros. 
But is `enum` the way to use macros ?

ahoppen: It can be either, really. The enum type just needs to provide static methods.

kimdv: Okay, so it's same when developers use an `enum` for name spaces?

What is then the "preferred" way to do it? 
What does documentation and WWDC videos do?

It's more to do what they do, so people reading macro for the first time and comparing to the videos, and get a bit confused.

Matejkob: Since the declaration needs to provide static methods, an `enum` appears to be the best choice. I assume there won't be a scenario where a user would want to create an instance of a declaration implementing macro protocols.

I haven't yet caught up on the WWDC videos (they're still on my to-watch list 🙈), but from what I've seen in the documentation, there's only one instance where a `struct` is used. Moreover, the available documentation on macros seems somewhat sparse; it doesn't appear to be very thorough.

By the way, if I find some time, I'm gonna to write DocC tutorials about macros. That way, we could make it the primary entry point for those interested in learning about macros.

Examples/Sources/MacroExamples/Implementation/Expression/FontLiteralMacro.swift
Matejkob: Sure thing. I've added an issue to track all enhancements to macro examples including this remake: #2169. Thx.

Examples/Sources/MacroExamples/Playground/main.swift
Matejkob: Great idea! Done!

Examples/Tests/MacroExamples/Implementation/ComplexMacros/DictionaryIndirectionMacroTests.swift
Matejkob: Done, thanks

Examples/Sources/MacroExamples/Playground/main.swift
Matejkob: In this commit https://github.com/apple/swift-syntax/pull/2163/commits/94decb8f1c930a8013762dbd96ed25d84ebbc5f9

Examples/Tests/MacroExamples/Implementation/ComplexMacros/DictionaryIndirectionMacroTests.swift
Matejkob: I've been thinking about doing it as another part of macros enhancements - I didn't want to change code without writing unit tests first but "this formatting is disgusting" is really good point so I've added all formatting related suggestion in a new commit after adding unit test: https://github.com/apple/swift-syntax/pull/2163/commits/39032ab805de40a7a4c7416a71a3f7704662a8ce

Examples/Tests/MacroExamples/Implementation/Member/CustomCodableTests.swift
Matejkob: Done in https://github.com/apple/swift-syntax/pull/2163/commits/39032ab805de40a7a4c7416a71a3f7704662a8ce. Thanks!

Examples/Tests/MacroExamples/Implementation/Member/MetaEnumMacroTests.swift
Matejkob: Done in https://github.com/apple/swift-syntax/pull/2163/commits/39032ab805de40a7a4c7416a71a3f7704662a8ce. Thanks!

Examples/Tests/MacroExamples/Implementation/Peer/AddAsyncMacroTests.swift
Matejkob: Done in https://github.com/apple/swift-syntax/pull/2163/commits/39032ab805de40a7a4c7416a71a3f7704662a8ce. Thanks!

Examples/Tests/MacroExamples/Implementation/ComplexMacros/DictionaryIndirectionMacroTests.swift
ahoppen: Yeah, I thought the same about fixing it in a follow-up PR but the formatting was just too ugly for me. And since the fix is fairly straightforward, I thought we can fix it in this PR.

Examples/Sources/MacroExamples/Implementation/Expression/FontLiteralMacro.swift
ahoppen: @DougGregor Do you have any opinions here?

DougGregor: Only static methods are used to implement macros, so in essence it doesn't matter what kind of type you use. I'm fine with `enum`, as a caseless enum makes it clear that there will never be an instance.

DougGregor (approved):
This is great, thank you!

ahoppen (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

Matejkob:
I believe something went wrong with Swift Test Windows

bnbarham:

PR #2164, opened 2023-09-07
[SE-0407] Add missingConformancesTo argument to MemberMacro expansion operation

DougGregor:
Stage in an entrypoint for member macros that allows them to learn about which conformances that they've asked about are "missing", meaning that they are not present on the type (ignoring those that would be generated by an extension macro). This information is equivalent to the information provided to extension macros, although the member macro itself cannot create the conformance.


DougGregor:
@swift-ci please test

PR #2167, opened 2023-09-07
Fix compilation error in Macro Versioning.md

mbrandonw:
When specifying a range in `.package` you do not provide the `from` argument name.


ahoppen (approved):
Thanks!

ahoppen:
@swift-ci Please test

PR #2168, opened 2023-09-08
List the token kinds in child documentation in SyntaxNodes

### LINKED ISSUE #1987: For children of syntax nodes that are `TokenSyntax`, list the kinds of tokens it can contain
For example `DeclReferenceExprSyntax.baseName` should have a comment saying the following (token choices can be generated using `GrammarGenerator`): 

```
For syntax trees generated by the parser, this is guaranteed to be one of the following token kinds:
- `<identifier>`
- 'self'
- 'Self'
- 'init'
- <dollarIdentifier>
- <binaryOperator>
- <integerLiteral>
```

For children that only have one token kind, like `StructDeclSyntax.structKeyword` it should say something like the following

```
For syntax trees generated by the parser, this is guaranteed to be the following kind: 'struct'
```

natikgadzhi:
## Motivation
- Fixes #1987. 
- Provides improved documentation and lists possible token choices in node's children.

## Changes
- Changes `Child.documentation` to be a computed property that wraps documentation provided when declaring the child, with the list of possible token choices. This should be extensible in the future to add more documentation for other child kinds as well.
- Provides additional func in `GrammarGenerator` to generate token choices in a list.
- Includes the generated code in the same commit.


CodeGeneration/Sources/SyntaxSupport/Child.swift
```diff
@@ -87,8 +87,27 @@ public class Child {
   /// This is used to e.g. describe the child if all of its tokens are missing in the source file.
   public let nameForDiagnostics: String?
 
-  /// A doc comment describing the child.
-  public let documentation: SwiftSyntax.Trivia
+  /// A summary of a doc comment describing the child. Full docc comment for the
+  /// child is available in ``Child/documentation``, and includes detailed list
+  /// of possible choices for the child if it's a token kind.
+  public let documentationSummary: SwiftSyntax.Trivia
```
natikgadzhi, line 93: The good old `documentation` is now `documentationSummary`, but the spots that used it would benefit from having the list in them. That's why this commit does not need to change `SyntaxNodesFile.swift` at all — we just change the underlying `child.documentation`.

CodeGeneration/Sources/SyntaxSupport/Child.swift
```diff
@@ -87,8 +87,27 @@ public class Child {
...
+
+  /// A docc comment describing the child, including the trivia provided when
+  /// initializing the Child, and the list of possible token choices inferred automatically.
+  public var documentation: SwiftSyntax.Trivia {
+    if case .token(let choices, _, _) = kind {
+      var tokenKindsComment = choices.count == 1 ?
+        "For syntax trees generated by the parser, this is guaranteed to be the following kind:" :
```
natikgadzhi, line 100: I _think_ we could shave off the extra line and make the pluralization easier, but that's probably fine.

CodeGeneration/Sources/SyntaxSupport/GrammarGenerator.swift
```diff
@@ -60,4 +60,20 @@ struct GrammarGenerator {
...
+    let generator = GrammarGenerator()
+
+    if case .token(let choices, _, _) = child.kind {
+      if choices.count == 1 {
+        return " \(generator.grammar(for: choices.first!))"
+      } else {
+        return choices.map { " - \(generator.grammar(for: $0))" }.joined(separator: "\n")
+      }
+    } else {
+      return ""
```
natikgadzhi, line 76: This is sus: if the `child` is not a `.token`, then its token choices are not an empty string, they are technically `nil`. No practical harm out of that, perhaps that's just a technicality. Also mentioned in the docc comment of the function.

@ahoppen, want me to make this func return an optional string, and unwrap upstream?

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxNodesD.swift
```diff
@@ -636,6 +678,14 @@ public struct DeclReferenceExprSyntax: ExprSyntaxProtocol, SyntaxHashable, _Leaf
     }
   }
   
+  /// For syntax trees generated by the parser, this is guaranteed to be one of the following token kinds:
+  ///  - `<identifier>`
```
natikgadzhi, line 682: Mmmm, backticks are not needed — I'll remove them in the `GrammarGenerator` for this.

natikgadzhi (commented):
@ahoppen, @kimdv, here's the first attempt!

This pull request handles `.token` children, but not `.keyword` yet. I would prefer to keep this small and get this one merged, then immediately follow up with `.keyword` (to cover https://github.com/apple/swift-syntax/issues/1987#issuecomment-1707663923) for easier review. 

How do you feel about the approach? What can I improve?

CodeGeneration/Sources/SyntaxSupport/GrammarGenerator.swift
```diff
@@ -60,4 +69,20 @@ struct GrammarGenerator {
...
+  /// The child must be of kind ``ChildKind/token``. Otherwise, an empty string will be returned.
+  static func childTokenChoices(for child: Child) -> String {
+    let generator = GrammarGenerator()
+
+    if case .token(let choices, _, _) = child.kind {
+      if choices.count == 1 {
+        return " \(generator.grammar(for: choices.first!, backticked: false))"
+      } else {
+        return choices.map { " - \(generator.grammar(for: $0, backticked: false))" }.joined(separator: "\n")
+      }
```
ahoppen, line 83: Instead of switching on `choices.count == 1` here again, what do you think about making the ternary operator in `Child.documentation` a full-grown `if` and then directly calling into `generator.grammar` from there? That would simplify `childTokenChoices` to always return a list.

natikgadzhi: Makes sense to only have the if/else in any form only in one place instead of two — I'll take a look.

CodeGeneration/Sources/SyntaxSupport/GrammarGenerator.swift
natikgadzhi: I actually even started with `precondition` at first, and then thought that if I'm carefully unwrapping, it's not necesary. 

But, 💯, it's better to have an explicit failure than an unexpected empty output.

Sources/SwiftSyntax/generated/SyntaxTraits.swift
```diff
@@ -16,11 +16,13 @@
 
 
 public protocol BracedSyntax: SyntaxProtocol {
+  /// For syntax trees generated by the parser, this is guaranteed to be the following kind: '{'
```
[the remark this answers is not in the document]
natikgadzhi: That's actually nice, I started with that and than had to add `backticked: Bool = true` argument to the function above — it'd be good to clean it back out.

Sources/SwiftSyntax/generated/SyntaxTraits.swift
```diff
@@ -284,7 +297,7 @@ public extension SyntaxProtocol {
 /// 
 /// See the types conforming to this protocol for examples of where missing nodes can occur.
 public protocol MissingNodeSyntax: SyntaxProtocol {
-  /// A placeholder, i.e. `<#placeholder#>`, that can be inserted into the source code to represent the missing node.
+  /// A placeholder, i.e. `<#placeholder#>`, that can be inserted into the source code to represent the missing node./// For syntax trees generated by the parser, this is guaranteed to be the following kind: <identifier>
```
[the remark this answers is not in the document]
natikgadzhi: OUCH! I thought it's smart enough to do it when I'm doing `trivia.appending(otherTrivia)` and missed it. Thank you! I'll clean it up.

CodeGeneration/Sources/SyntaxSupport/Child.swift
```diff
@@ -87,8 +87,45 @@ public class Child {
...
-  /// A doc comment describing the child.
-  public let documentation: SwiftSyntax.Trivia
+  /// A summary of a doc comment describing the child. Full docc comment for the
+  /// child is available in ``Child/documentation``, and includes detailed list
+  /// of possible choices for the child if it's a token kind.
+  public let documentationSummary: SwiftSyntax.Trivia
+
+  /// A docc comment describing the child, including the trivia provided when
+  /// initializing the ``Child``, and the list of possible token choices inferred automatically.
+  public var documentation: SwiftSyntax.Trivia {
```
natikgadzhi, line 97: The part that I don't quite like about this is that we're essentially templating the doc comment for `Child` in here. But for `Node`, it's in `templates/SyntaxNodeFile.swift`. So this feels like we have a way of storing templates — in `templates`, but this sort of dilutes that distinction. "Don't put logic in your views", as we say in the web apps world.

CodeGeneration/Sources/SyntaxSupport/Child.swift
```diff
@@ -87,8 +87,45 @@ public class Child {
...
+        ? "kind: \(grammar.grammar(for: choices.first!))"
+        : """
+        kinds:
+        \(choices.map { " - \(grammar.grammar(for: $0))" }.joined(separator: "\n"))
+        """
+
+      // trivia's last piece does not end in \n.
+      // So if the existing documentation is not emtpy, we need to append an
+      // additional TriviaPiece.newlines(1) before appending the token choices.
+      if documentationSummary.isEmpty {
```
natikgadzhi, line 117: Might be a good idea to provide `Trivia.appending(trivia, newLines: Int = 0)` that inserts new lines between trivia pieces automatically if we see this pattern frequently, but I wanted to keep this change as small as possible.

natikgadzhi (commented):
@ahoppen ready for a quick review. 

I'm not 100% sold on putting the templating code in `Child`, but maybe it's good for now — perhaps we can merge this, and then add `.kind` comment to, and then I could move the template logic out. 

The initial reason I've put this code in `Child` is because in `SyntaxNodeFile`, we can do `let documentation = ...` and then render it in the builder. But, we're iterating over children already in the result builder, so not like we can do some free-standing code in there.

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxNodesAB.swift
```diff
@@ -908,6 +945,9 @@ public struct ActorDeclSyntax: DeclSyntaxProtocol, SyntaxHashable, _LeafDeclSynt
     }
   }
   
+  /// ### Tokens
+  /// 
+  /// For syntax trees generated by the parser, this is guaranteed to be the following kind: `'actor'`
```
natikgadzhi, line 950: @kimdv, the actor keyword doc showed up!

ahoppen (commented):
I think we could merge this as is but if you want to clean the code a little bit more, I’ve got a couple suggestions inline.

CodeGeneration/Sources/SyntaxSupport/Child.swift
natikgadzhi: I think perhaps this is a little bit bigger than one occurrence. The pattern is as follows:
- A file in `templates/`, let's say `SyntaxNodesFile.swift` has a result builder that renders the code. Suddenly, you need to generate something that is not easily available as an object property or a function on a `Node` or a `Child`.
- You can't comfortably put that code in the result builder, so you shove that code _somwehere_.
- So far, I've seen extensions of `LayoutNode` (in the other PR of mine) and now I'm doing a function on `Child`. 

That feels slightly unorthodox, but not too bad: 
- templates should be as easy to read as possible. Sure, we're not SwuftUI here, but the closer we can get in terms of readability, the better.
- Node and Child and other definitions should ideally contain the information you need to generate the code, but not the generation code itself — it's like the view layer and data layer to me.
- Seems like Utils / extensions solve that problem partially, but since we're extending Child / Node, it feels like we're injecting view code in the model.


I'm uncertain if my opinion is valid — I might not have enough exposure and experience. But so far, my idea is something like:
- Add `templates/helpers/*.Swift`
- Make `enums` or `structs` that you can't instantiate with static functions that are clean. They take the object in, and render syntax out. They can only depend on their arguments, without side effects.

@ahoppen want me to start a discussion / post on forums on this?

CodeGeneration/Sources/SyntaxSupport/Child.swift
```diff
@@ -87,8 +87,46 @@ public class Child {
...
+      )
+
+      let separator = SwiftSyntax.Trivia(pieces: [TriviaPiece.newlines(1), TriviaPiece.docLineComment("///"), TriviaPiece.newlines(1)])
+
+      // Grab the documentation secions, filter out empty ones, and join them
+      // with an empty documentation line inbetween.
+      return [documentationSummary, tokenChoicesTrivia]
+        .filter { !$0.isEmpty }
+        .joined(separator: separator)
+        .reduce(SwiftSyntax.Trivia(), { $0.appending($1) })
```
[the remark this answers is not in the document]
natikgadzhi: Exactly! Triggered me, too: https://github.com/apple/swift-syntax/pull/2168#discussion_r1320284758 

I'll try to choose the right spot for it. There's a helper that generates `Trivia` from `String` in `Utils` — I think I will take that helper too, and move it + this new helper in `CodeGeneration/SwiftSupport/Helpers` or `CodeGeneration/Utils/Helpers`, with docc comment on helpers enum that outlines why they are here. 

Should be a nice proof of concept. Let me think for a bit whether that should be a separate PR though — I think the discussion on it might take a little bit, and I want to scope my attention 🙃

Sources/SwiftSyntax/generated/SyntaxTraits.swift
```diff
@@ -121,6 +127,11 @@ public protocol EffectSpecifiersSyntax: SyntaxProtocol {
     set
   }
   
+  /// ### Tokens
+  /// 
+  /// For syntax trees generated by the parser, this is guaranteed to be the following kinds:
```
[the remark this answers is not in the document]
natikgadzhi: Let me factor this in, I agree. Then, I would ask to merge this, with a condition that I start an issue and follow up on extracting repetitive helper / template code into helpers.

CodeGeneration/Sources/SyntaxSupport/Child.swift
ahoppen: I agree that we aren’t super strict on the distinction between the pure data layer and the view layer in CodeGeneration and you probably notice this a lot more than I do because I haven’t been doing much view-based / UI development in a few years, where you think a lot more about that separation.

But then also, the distinction is a little bit muddy IMO. Just taking the documentation as an example. I think we both agree that the documentation should be a property of a node/child but should that documentation already contain the generated *Token* section? I think you can argue in either direction. 

That being said, if you have ideas how to make that distinction clearer, I’m all in favor for them. One thing that’s important to me, though, is that we don’t introduce too many new abstraction layers that make the code less readable. Since CodeGeneration is still fairly reasonable in size and I don’t expect it to grow too much faster in the future, I would prefer an unorthodox decision here and there to an overengineered system of abstraction layers.

Sources/SwiftSyntax/generated/SyntaxTraits.swift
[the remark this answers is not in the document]
ahoppen: Sounds good to me

CodeGeneration/Sources/SyntaxSupport/Child.swift
```diff
@@ -87,8 +87,39 @@ public class Child {
...
+
+      let tokenChoicesTrivia = SwiftSyntax.Trivia.docCommentTrivia(
+        from: """
+          ### Tokens
+
+          For syntax trees generated by the parser, this is guaranteed to be one of the following \(choices)
+          """
+      )
+
+      return SwiftSyntax.Trivia(joining: [documentationSummary, tokenChoicesTrivia])
```
natikgadzhi, line 117: @ahoppen, refactored joining trivia pieces — here and in other spots.

CodeGeneration/Sources/SyntaxSupport/Utils.swift
```diff
@@ -38,21 +38,39 @@ public func lowercaseFirstWord(name: String) -> String {
...
-  guard let string else {
-    return []
-  }
-  let lines = string.split(separator: "\n", omittingEmptySubsequences: false)
-  let pieces = lines.enumerated().map { (index, line) in
-    var line = line
-    if index != lines.count - 1 {
-      line = "\(line)\n"
+// Helpers to create trivia pieces
+extension SwiftSyntax.Trivia {
```
natikgadzhi, line 42: The reason I moved `docCommentTrivia` is just to be consistent, and to hint at where it is located.

CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxNodesFile.swift
```diff
@@ -22,24 +22,12 @@ import Utils
 func syntaxNode(nodesStartingWith: [Character]) -> SourceFileSyntax {
   SourceFileSyntax(leadingTrivia: copyrightHeader) {
     for node in SYNTAX_NODES.compactMap(\.layoutNode) where nodesStartingWith.contains(node.kind.syntaxType.description.first!) {
-      let documentationSections = [
```
natikgadzhi, line 25: While this is clean, there's a way to make this even nicer — we can rename `Node.documentation` to `Node.documenationSummary` and make a computed property `documentation` that works like `Child.documentation` and internally joins them together.

I'll push that change in a bit to this PR.

natikgadzhi (commented):
Almost ready! One more cleanup for consistency in `node.documentation` and I'll try to convince you to merge it ;)

CodeGeneration/Sources/SyntaxSupport/Child.swift
```diff
@@ -87,8 +87,39 @@ public class Child {
...
+        : """
+        kinds:
+        \(choices.map { " - \(grammar.grammar(for: $0))" }.joined(separator: "\n"))
+        """
+
+      let tokenChoicesTrivia = SwiftSyntax.Trivia.docCommentTrivia(
+        from: """
+          ### Tokens
+
+          For syntax trees generated by the parser, this is guaranteed to be one of the following \(choices)
```
ahoppen, line 113: Oh, sorry, maybe I should have been cleaner. In the singular, it should be “this is guaranteed to be the following kind” (“one of” should only be there in the plural form). Overall I think it might be easiest to just have two completely separate branches for the singular and plural case.

CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxNodesFile.swift
ahoppen: I agree. In particular, it would be nice to be consistent with how we stitch together the comment sections.

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxNodesAB.swift
kimdv: I would make it `actor` and remove single quote

Sources/SwiftSyntax/generated/SyntaxTraits.swift
```diff
@@ -16,11 +16,17 @@
 
 
 public protocol BracedSyntax: SyntaxProtocol {
+  /// ### Tokens
+  /// 
+  /// For syntax trees generated by the parser, this is guaranteed to be one of the following kind: `'{'`
```
kimdv, line 21: Not sure if I think the `### Tokens` is needed? 
@ahoppen what is the reason to use this?

CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxNodesFile.swift
natikgadzhi: If we agree on this, then perhaps I should include that cleanup in this PR. 

I'm hesitant to bring new abstractions, or even move things into `Grammar` yet — I want to understand it's place better. But what we can do, is we can absolutely make `Node` and `Child` consistent with each other. I'm clear on how to do that, and have everything to prepare a commit.

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxNodesAB.swift
natikgadzhi: We discussed that in here: https://github.com/apple/swift-syntax/pull/2168#discussion_r1319977750

But what I missed is I've put backticks back, but left single quotes in. 

To confirm, what I want to do here is: 

`/// For syntax trees generated by the parser, this is guaranteed to be of kind \`actor\`.`

Sounds good?

Sources/SwiftSyntax/generated/SyntaxTraits.swift
natikgadzhi: My gut feel is there are two reasons for it:
- It's consistent with how we structure documentation for `Node`
- Using these headings works nice in DocC when it's rendered.

I, personally, would like to keep them, but I see how it might feel like an overkill for a situation where there's one possible token kind.

kimdv: Nice with DocC. 
I will need to see how it renders then 😁

natikgadzhi: So in `SyntaxNodesFile.swift` we construct node documentation of three pieces. Here's the `node.grammaer` piece that generates the Children subheading pictured below: https://github.com/apple/swift-syntax/blob/ddd05efdf15574413c70645803488f2c4311014c/CodeGeneration/Sources/SyntaxSupport/Node.swift#L357

Screenshot 2023-09-12 at 11 57 02 AM

Ahem... should we put them in a table? Should we also backtich the children names? Oh gawd why have I looked at it.

kimdv: I just saw the children. Looks nice. 

But I think they are a bit duplicate of "Instance Properties". 
But that is not related to this PR

natikgadzhi: One of the follow-up items on this PR is to make _how_ we render those pieces (children) in the code consistent between Node and Children. I might do that in this PR, or in another one. 

Since I'm already elbow deep in this, I'm happy to help to explore how we can improve the autogenerated docs, here and otherwise. Feel free to write / assign issues to me.

ahoppen: > But I think they are a bit duplicate of "Instance Properties".
> But that is not related to this PR

It kind of is but the *Children* section has a few properties that the *Instance Properties* section doesn’t have
- It’s in the order that the children appears within the syntax node, compared to being alphabetical, which is a huge point
- It shows the token kinds for all token children (which might be less important with this PR) but it’s nice to see that in a concise way.
- It omits the `unexpected` children, which is kind of nice.

kimdv: When I search for documentation I would actually argue it's easier for users to find it when ordered alphabetical, than in the order they appear in the code, as I don't see the code right next to the docs (I use SPI). 
We know the order, as we look into the code a lot. But I'm not sure that people are thinking on that order in the code, when reading the docs. 

To point two we could add that to the instance properties. 
Auto generate that and add it as part of the provided documentation? 

The `unexpected` I agree that's nice, for me it's actually a bigger point that it's omitted that they ordered alphabetical.

ahoppen: > When I search for documentation I would actually argue it's easier for users to find it when ordered alphabetical, than in the order they appear in the code, as I don't see the code right next to the docs (I use SPI).
> We know the order, as we look into the code a lot. But I'm not sure that people are thinking on that order in the code, when reading the docs.

At least for me, it makes a huge difference to have them ordered in the same order that the children will occur in the parsed source code. It might be a silly example but if I’m looking at `FunctionDeclSyntax` and want to know which node represents `public`, I know that `public` is spelled before `func`. Thus, I look for the child that represents `func` and then just look at the nodes before it: `attributes` and `modifiers`. I know that attributes are spelled with an `@`, so I know that `modifiers` need to contain `public`. And I didn’t need to wonder if `public` might be part of `signature`.

> To point two we could add that to the instance properties. And if there is added some documentation inside the Node or Child in SyntaxSupport, I would just append it to the end. So the handwritten is first, and then the generated documentation.

We could, but at least with the current rendering of docc, I would argue that the `Children` section is denser and more readable. Also, just for some background, the *Children* section got introduced as a replacement for the grammar comments that we had scattered throughout the source code and I think it does fulfill that role.

natikgadzhi:
I've just pushed up a commit that addresses a couple of review items: 
- @kimdv, I dropped single quotes from token choices list. 
- @ahoppen, hopefully I've made the wording cleaner. I like it more this way. 
- I moved some of the string generation into `GrammarGenerator` for now. 

TODOs:
- I'm deliberately not doing further refactoring / extraction of generation code, or extra documentation on what should go into `GrammarGenerator` in the future — I would love to work more on related problems to learn and listen to users first. Are we okay with that for now? 
- I noticed that if the child already has documentation screen passed from the declaration, it might look awkward, like this: 
   ```
  /// The `@` sign.
  ///
  /// ### Tokens
  /// 
  /// For syntax trees generated by the parser, this is guaranteed to be `@` 
  public var atSign: TokenSyntax {
   ```
   @ahoppen, are we okay with these, and will remove manual comment? We could make some logit that omits the generated comment but only if it's a single choice and if `child.documentation` is not empty, but that's hacky. 
- This needs a rebase now. 

Looking to get your thumbs up on the above, but will do a rebase in a sec to make the PR cleaner.

CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxNodesFile.swift
```diff
@@ -22,24 +22,12 @@ import Utils
...
-        .joined(separator: [Trivia.newline, Trivia.docLineComment("///"), Trivia.newline])
-        .reduce(Trivia(), +)
-
       // We are actually handling this node now
       try! StructDeclSyntax(
         """
         // MARK: - \(node.kind.syntaxType)
 
-        \(documentation)
+        \(SwiftSyntax.Trivia(joining: [node.documentation, node.grammar, node.containedIn]))
```
natikgadzhi, line 30: @ahoppen I think you told me there's this code in _three_ spots, but I only caught two: here, and in `Child.swift`. Where's the third one?

ahoppen: There’s another one in SyntaxCollectionsFile.swift:21

CodeGeneration/Sources/SyntaxSupport/GrammarGenerator.swift
```diff
@@ -60,4 +65,25 @@ struct GrammarGenerator {
...
+  /// Generates a markdown string describing possible choices for the given child
+  /// token.
+  static func childTokenChoices(for choices: [TokenChoice]) -> String {
+    let grammar = GrammarGenerator()
+    
+    if choices.count == 1 {
+      return """
+        ### Tokens
+
+        For syntax trees generated by the parser, this is guaranteed to be \(grammar.grammar(for: choices.first!))
```
ahoppen, line 78: And a very minor nitpick.

```suggestion
        For syntax trees generated by the parser, this is guaranteed to be \(grammar.grammar(for: choices.first!)).
```

ahoppen (approved):
One minor nitpick, otherwise looks good to merge to me.

ahoppen:
> * I'm deliberately not doing further refactoring / extraction of generation code, or extra documentation on what should go into `GrammarGenerator` in the future — I would love to work more on related problems to learn and listen to users first. Are we okay with that for now?

Yes, let’s leave any further refactorings to follow-up PRs.

> * I noticed that if the child already has documentation screen passed from the declaration, it might look awkward, like this:
>   ```
>   /// The `@` sign.
>   ///
>   /// ### Tokens
>   /// 
>   /// For syntax trees generated by the parser, this is guaranteed to be `@` 
>   public var atSign: TokenSyntax {
>   ```
>     
>   @ahoppen, are we okay with these, and will remove manual comment? We could make some logit that omits the generated comment but only if it's a single choice and if `child.documentation` is not empty, but that's hacky.

Yes, that’s OK

natikgadzhi:
@ahoppen, thank you for showing up and reviewing on a Saturday! Cleaned up the nit, and cleaned up `SwiftCollectionsFile` to align. Then ran the generator to ensure the generated sources stay intact.

ahoppen (approved):

ahoppen:
@swift-ci Please test

ahoppen:
Looks like the code isn’t formatted correctly. Could you run `format.py`?

natikgadzhi:
Done. I need to set this up on my precommit hook 👀

ahoppen:
@swift-ci Please test

kimdv:
@swift-ci please test windows

PR #2173, opened 2023-09-08
[CMake] Updates to allow inclusion using FetchContent

rintaro:
Update for https://github.com/apple/swift/pull/68408


CMakeLists.txt
```diff
@@ -15,12 +15,28 @@ project(SwiftSyntax LANGUAGES C Swift)
 set(SWIFT_VERSION 5)
 set(CMAKE_Swift_LANGUAGE_VERSION ${SWIFT_VERSION})
 
+if(CMAKE_VERSION VERSION_LESS 3.21)
+  get_property(parent_dir DIRECTORY PROPERTY PARENT_DIRECTORY)
+  if(NOT parent_dir)
+    set(PROJECT_IS_TOP_LEVEL TRUE)
+  endif()
+endif()
```
compnerd, line 23: That is horrible!

CMakeLists.txt
```diff
@@ -15,12 +15,28 @@ project(SwiftSyntax LANGUAGES C Swift)
...
+if(SWIFT_HOST_LIBRARIES_DEST_DIR)
+  set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY "${SWIFT_HOST_LIBRARIES_DEST_DIR}")
+  set(CMAKE_LIBRARY_OUTPUT_DIRECTORY "${SWIFT_HOST_LIBRARIES_DEST_DIR}")
+else()
+  set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/lib/${SWIFT_HOST_LIBRARIES_SUBDIRECTORY}")
+  set(CMAKE_LIBRARY_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/lib/${SWIFT_HOST_LIBRARIES_SUBDIRECTORY}")
+endif()
+if(SWIFT_HOST_RUNTIME_DEST_DIR)
+  set(CMAKE_RUNTIME_OUTPUT_DIRECTORY "${SWIFT_HOST_RUNTIME_DEST_DIR}")
+else()
```
compnerd, line 37: Thoughts on doing something like this?

```cmake
if(NOT CMAKE_RUNTIME_OUTPUT_DIRECTORY)
  set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
endif()
```

This would allow us to control the behaviour from the `FetchContent` with the standard CMake options.

CMakeLists.txt
```diff
@@ -49,22 +65,25 @@ if (NOT SWIFT_SUPPORTS_DISABLE_IMPLICIT_STRING_PROCESSING_MODULE_IMPORT)
...
-if(CMAKE_Swift_COMPILER_TARGET)
-  string(REGEX REPLACE "macosx[0-9]+([.][0-9]+)?" "macos" SWIFT_MODULE_TRIPLE
-    ${CMAKE_Swift_COMPILER_TARGET})
-elseif("${CMAKE_SYSTEM_NAME}" STREQUAL "Windows")
-  if(CMAKE_SYSTEM_PROCESSOR MATCHES "x86_64|AMD64")
-    set(SWIFT_MODULE_TRIPLE "x86_64-unknown-windows-msvc")
-  elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64|ARM64|arm64")
-    set(SWIFT_MODULE_TRIPLE "aarch64-unknown-windows-msvc")
+if("${SWIFT_HOST_MODULE_TRIPLE}" STREQUAL "")
+  # FIXME: This is a hack. It's all a hack. Windows isn't setting
```
compnerd, line 69: Not sure I follow this.  This is meant to be a user controlled option, what were you expecting?  Or is this is a bug CMake that we should also try to get fixed?

cmake/modules/AddSwiftHostLibrary.cmake
```diff
@@ -6,22 +6,44 @@
...
+#
+# For now workaround this by touching a file whenever B is rebuilt and then compiling that file as
+# part of A. Ideally this file would be generated by each of the targets, but that dependency didn't
+# seem to be being tracked.
+#
+# Remove once rdar://102202478 is fixed.
+function(target_link_swift_syntax_libraries TARGET)
+  cmake_parse_arguments(ARGS "" "" "PUBLIC" ${ARGN})
+
+  target_link_libraries(${TARGET} PUBLIC ${ARGS_PUBLIC})
```
compnerd, line 22: Should `PUBLIC` not be conditional on `${ARGS_PUBLIC}`?

CMakeLists.txt
bnbarham: Such is pre 3.21 life 😭

CMakeLists.txt
bnbarham: Ooo that's a very good point! That would be much nicer, assuming it works.

CMakeLists.txt
bnbarham: The comment is just what was already there, I don't think this is a CMake bug. For non-Windows it's all being set via build-script.

bnbarham: Just to expand on this... The reason we need the module triple at all is to output the module in the subfolder format rather than the flat format. Ideally we'll get that fixed at some point 😅. `CMAKE_Swift_COMPILER_TARGET` is being set in build-script, but I assume that if this was added, it wasn't set for Windows somewhere (in PR testing?). `SWIFT_HOST_MODULE_TRIPLE` is being passed from the Swift build, so this path would only be taken in the standalone. Not ideal, but also not really anything to do with the fetchcontent change itself.

cmake/modules/AddSwiftHostLibrary.cmake
rintaro: It was a hack. It used to parse dependencies as `PUBLIC` multi-value arguments.
I updated it because it's was certainly confusing.

bnbarham (approved):

bnbarham:
Thanks! This should make our lives much simpler :)

rintaro:
https://github.com/apple/llvm-project/pull/7424
https://github.com/apple/swift/pull/68408
@swift-ci Please test

PR #2177, opened 2023-09-11
Add experimental-skip-synthesized-symbols parameter for doc generation on SPI

finestructure:
As just discussed on the Documentation WG, setting this parameter shrinks the generated docs down dramatically:

```
normal:
mb size:    708.0
file count: 84,619


skip synthesized symbols:
mb size:    155.0
file count: 17,537
```

cc @ethan-kusters

### DIFF
--- a/.spi.yml
+++ b/.spi.yml
@@ -22,4 +22,3 @@ builder:
       - SwiftSyntaxMacroExpansion
       - SwiftSyntaxMacrosTestSupport
       custom_documentation_parameters: [--experimental-skip-synthesized-symbols]
-      swift_version: '5.9'
--- a/.spi.yml
+++ b/.spi.yml
@@ -21,3 +21,5 @@ builder:
       - SwiftSyntaxMacros
       - SwiftSyntaxMacroExpansion
       - SwiftSyntaxMacrosTestSupport
+      custom_documentation_parameters: [--experimental-skip-synthesized-symbols]
+      swift_version: '5.9'


.spi.yml
```diff
@@ -21,3 +21,5 @@ builder:
       - SwiftSyntaxMacros
       - SwiftSyntaxMacroExpansion
       - SwiftSyntaxMacrosTestSupport
+      custom_documentation_parameters: [--experimental-skip-synthesized-symbols]
+      swift_version: '5.9'
```
finestructure, line 25: Ethan mentioned the flag should be supported on 5.8 as well, so this additional key is perhaps not necessary. I can run a test before merging.

.spi.yml
```diff
@@ -21,3 +21,5 @@ builder:
       - SwiftSyntaxMacros
       - SwiftSyntaxMacroExpansion
       - SwiftSyntaxMacrosTestSupport
+      custom_documentation_parameters: [--experimental-skip-synthesized-symbols]
```
ahoppen, line 24: Should this use a single dash? When I set it as `OTHER_DOCC_FLAGS`, I need to use a single dash, not two.

```suggestion
      custom_documentation_parameters: [-experimental-skip-synthesized-symbols]
```

ahoppen (approved):

.spi.yml
finestructure: In my testing this took two dashes, because it's a parameter on `swift package generate-docs` (we generate the docs with SwiftPM). This might be the case if you generate with docs with `xcodebuild` on iOS?

Either way, I'll need to verify it runs with 5.8 so I'll double check tomorrow and will report back!

ahoppen: OK, if it works with two dashes, I’m all in favor for using that version. 👍🏽

.spi.yml
finestructure: I've confirmed this works on 5.8 and with the double dash command.

```
Size (/MB): 146.0
File count: 16557
```

ahoppen:
@swift-ci Please test

PR #2184, opened 2023-09-13
Prevent Unrelated Casts on Child Choice Node

Matejkob:
Resolves part of #2092

### DIFF
--- a/CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxNodeCasting.swift
+++ b/CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxNodeCasting.swift
@@ -1,0 +1,88 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+import SwiftSyntaxBuilder
+import SyntaxSupport
+
+@MemberBlockItemListBuilder
+func choiceNodeCastingMethods(for syntaxNodeKind: SyntaxNodeKind) -> MemberBlockItemListSyntax {
+  if syntaxNodeKind.isBase {
+    DeclSyntax(
+      """
+      /// Checks if the current syntax node can be cast to the type conforming to the ``\(syntaxNodeKind.protocolType)`` protocol.
+      ///
+      /// - Returns: `true` if the node can be cast, `false` otherwise.
+      public func `is`<S: \(syntaxNodeKind.protocolType)>(_ syntaxType: S.Type) -> Bool {
+        return self.as(syntaxType) != nil
+      }
+      """
+    )
+
+    DeclSyntax(
+      """
+      /// Attempts to cast the current syntax node to the type conforming to the ``\(syntaxNodeKind.protocolType)`` protocol.
+      ///
+      /// - Returns: An instance of the specialized type, or `nil` if the cast fails.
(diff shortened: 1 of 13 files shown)


ahoppen (commented):
Hmm, this does significantly reduce the readability in many cases. I didn’t consider that you couldn’t easily chain `if case`…

After reading this, I think I changed my mind and we should offer non-deprecated casting methods to any option that the `SyntaxChildChoices` can have. And if one of that option is a base kind like `DeclSyntax`, we should offer a casting method to any `DeclSyntaxProtocol`.

Matejkob:
> Hmm, this does significantly reduce the readability in many cases. I didn’t consider that you couldn’t easily chain if case…

I completely agree. That was my main concern with value binding in the first place. I haven't yet found the time to go through this pitch: [https://forums.swift.org/t/pitch-is-case-expressions/64185](https://forums.swift.org/t/pitch-is-case-expressions/64185), so I'm not sure why it hasn't gone through yet. The issue definitely should be addressed—it's a day-to-day problem for many folks.

> After reading this, I think I changed my mind and we should offer non-deprecated casting methods to any option that the SyntaxChildChoices can have. And if one of that option is a base kind like DeclSyntax, we should offer a casting method to any DeclSyntaxProtocol.

Sounds good to me. I'll try cook something

ahoppen:
> so I'm not sure why it hasn't gone through yet. The issue definitely should be addressed—it's a day-to-day problem for many folks.

I haven’t read the pitch completely either but I would imagine it’s just a matter of someone writing the official proposal and creating an implementation for it.

Matejkob:
Base on last John's post https://forums.swift.org/t/pitch-is-case-expressions/64185/69 the whole pitch is abandoned 😢

CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxNodeCasting.swift
```diff
@@ -0,0 +1,68 @@
...
+import SwiftSyntax
+import SwiftSyntaxBuilder
+import SyntaxSupport
+
+@MemberBlockItemListBuilder
+func choiceNodeCastingMethods(for syntaxNodeKind: SyntaxNodeKind) -> MemberBlockItemListSyntax {
+  if syntaxNodeKind.isBase {
+    DeclSyntax(
+      """
+      public func `is`<S: \(syntaxNodeKind.protocolType)>(_ syntaxType: S.Type) -> Bool {
```
Matejkob, line 22: How about adding to those methods documentation in the same manner as to others e.g. https://github.com/apple/swift-syntax/blob/00dc4d966c25459a95cc65d4cab11013616a2d63/Sources/SwiftSyntax/SyntaxProtocol.swift#L133-L143 ?

Release Notes/510.md
```diff
@@ -49,6 +49,11 @@
   - Description: `is`, `as`, and `cast` methods on base node protocols with base-type conversions are marked as deprecated. The deprecated methods emit a warning that informs the developer that the cast will always succeed and should be done using the base node's initializer.
   - Issue: https://github.com/apple/swift-syntax/issues/2092
   - Pull Request: https://github.com/apple/swift-syntax/pull/2108
+  
+- Child Choice Node Casts
+  - Description: TODO: Add!
```
Matejkob, line 54: Gonna add at the end

Matejkob:
@ahoppen how about now? Usage is cleaner, but there is plenty of code more on the implementation part :/

CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxNodeCasting.swift
kimdv: Sounds good for me

ahoppen: Yes, those doc comments would be useful.

ahoppen (commented):
Looks good to me in general

ahoppen (approved):

ahoppen:
@swift-ci Please test

kimdv:
@swift-ci please test windows

ahoppen:
@swift-ci Please test Windows

PR #2185, opened 2023-09-13
Handle `DeinitializerDeclSyntax` errors in a single diagnostic

kimdv:
Part of #2021

I wanted to open this PR so we can discuss it isolated

### DIFF
--- a/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
+++ b/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
@@ -788,40 +788,36 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
     if shouldSkip(node) {
       return .skipChildren
     }
-    if let unexpected = node.unexpectedBetweenDeinitKeywordAndEffectSpecifiers,
-      let name = unexpected.presentTokens(satisfying: { $0.tokenKind.isIdentifier == true }).only
-    {
+
+    let name: TokenSyntax? = node
+      .unexpectedBetweenDeinitKeywordAndEffectSpecifiers?
+      .presentTokens(satisfying: \.tokenKind.isIdentifier)
+      .only
+
+    let params: FunctionParameterClauseSyntax? = node
+      .unexpectedBetweenDeinitKeywordAndEffectSpecifiers?
+      .compactMap({ $0.as(FunctionParameterClauseSyntax.self) })
+      .only
+
+    let returnType: ReturnClauseSyntax? = node
+      .unexpectedBetweenEffectSpecifiersAndBody?
+      .compactMap({ $0.as(ReturnClauseSyntax.self) })
+      .only
+
+    let nodes = [Syntax(name), Syntax(params), Syntax(returnType)].compactMap { $0 }
+
+    if let firstNode = nodes.first {
       addDiagnostic(
-        name,
-        .deinitCannotHaveName,
+        firstNode,
+        DeinitializerSignatureError(name: name, params: params, returnClause: returnType),
+        highlights: nodes,
         fixIts: [
-          FixIt(message: RemoveNodesFixIt(name), changes: .makeMissing(name))
+          FixIt(
+            message: RemoveNodesFixIt(nodes),
+            changes: nodes.map { .makeMissing($0) }
+          )
         ],
-        handledNodes: [name.id]
-      )
-    }
-    if let unexpected = node.unexpectedBetweenDeinitKeywordAndEffectSpecifiers,
-      let params = unexpected.compactMap({ $0.as(FunctionParameterClauseSyntax.self) }).only
-    {
-      addDiagnostic(
-        params,
-        .deinitCannotHaveParameters,
-        fixIts: [
-          FixIt(message: RemoveNodesFixIt(params), changes: .makeMissing(params))
-        ],
-        handledNodes: [params.id]
-      )
-    }
-    if let unexpected = node.unexpectedBetweenEffectSpecifiersAndBody,
-      let returnType = unexpected.compactMap({ $0.as(ReturnClauseSyntax.self) }).only
-    {
-      addDiagnostic(
-        returnType,
-        .deinitCannotHaveReturnType,
-        fixIts: [
-          FixIt(message: RemoveNodesFixIt(returnType), changes: .makeMissing(returnType))
-        ],
-        handledNodes: [returnType.id]
+        handledNodes: nodes.map { $0.id }
       )
     }
 
--- a/Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift
+++ b/Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift
@@ -120,17 +120,8 @@ extension DiagnosticMessage where Self == StaticParserError {
   public static var defaultOutsideOfSwitch: Self {
     .init("'default' label can only appear inside a 'switch' statement")
   }
-  public static var deinitCannotHaveName: Self {
-    .init("deinitializers cannot have a name")
-  }
-  public static var deinitCannotHaveParameters: Self {
-    .init("deinitializers cannot have parameters")
-  }
   public static var deinitCannotThrow: Self {
     .init("deinitializers cannot throw")
-  }
-  public static var deinitCannotHaveReturnType: Self {
-    .init("deinitializers cannot have return type")
   }
   public static var editorPlaceholderInSourceFile: Self {
     .init("editor placeholder in source file")
@@ -302,6 +293,36 @@ public struct CannotParseVersionTuple: ParserError {
 
   public var message: String {
     return "cannot parse version component \(versionTuple.shortSingleLineContentDescription)"
+  }
+}
+
+public struct DeinitializerSignatureError: ParserError {
+  public let name: TokenSyntax?
+  public let params: FunctionParameterClauseSyntax?
+  public let returnClause: ReturnClauseSyntax?
+
+  public var message: String {
+    var descriptions: [String] = []
+
+    if name != nil {
+      descriptions.append("a name")
+    }
+
+    if params != nil {
+      descriptions.append("parameters")
+    }
+
+    if returnClause != nil {
+      var message = "return clause"
+
+      if descriptions.isEmpty {
+        message = "a " + message
+      }
+
+      descriptions.append(message)
+    }
+
+    return "deinitializers cannot have \(formatDescriptions(descriptions))"
   }
 }
 
--- a/Tests/SwiftParserTest/translated/InitDeinitTests.swift
+++ b/Tests/SwiftParserTest/translated/InitDeinitTests.swift
@@ -143,12 +143,15 @@ final class InitDeinitTests: ParserTestCase {
     assertParse(
       """
       struct FooStructDeinitializerA {
-        deinit 1️⃣x2️⃣()
+        deinit 1️⃣x()
       }
       """,
       diagnostics: [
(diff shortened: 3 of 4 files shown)


Tests/SwiftParserTest/translated/InitDeinitTests.swift
```diff
@@ -129,7 +129,7 @@ final class InitDeinitTests: ParserTestCase {
       }
       """,
       diagnostics: [
-        DiagnosticSpec(message: "deinitializers cannot have a name", fixIts: ["remove 'x'"])
+        DiagnosticSpec(message: "deinitializers cannot have token", fixIts: ["remove 'x'"])
```
kimdv, line 132: This is a down side, as we only have a TokenSyntax here. 

Could not quickly think of a solution that would be more informative. 
Maybe we could add an enum with error types, to the new error. 
And then give it as params instead of the nodes. But that is a bit of anti pattern of what we do other places? 

What do you think @ahoppen

Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift
```diff
@@ -305,6 +296,14 @@ public struct CannotParseVersionTuple: ParserError {
   }
 }
 
+public struct DeinitializerSignatureError: ParserError {
+  public let nodes: [SyntaxProtocol]
+
+  public var message: String {
+    return "deinitializers cannot have \(nodes.map { $0.nodeTypeNameForDiagnostics(allowBlockNames: false) ?? "syntax" }.joined(separator: " or "))"
```
ahoppen, line 303: Could you just call `nodesDescription(nodes)` instead here? That should call into `childNameForDiagnostics`. And if you set the `nameForDiagnostics` of `DeinitializerSyntax.name`, I would imagine that this should give you a decent error message.

kimdv: The thing is `DeinitializerSyntax` don't have a name 😅
Updated all the test cases so you can see the result

Tests/SwiftParserTest/translated/InvalidTests.swift
```diff
@@ -283,7 +283,7 @@ final class InvalidTests: ParserTestCase {
           message: "expected ')' in function type",
           fixIts: ["insert ')'"]
         ),
-        DiagnosticSpec(locationMarker: "3️⃣", message: "expected return type in function type", fixIts: ["insert return type"]),
+        DiagnosticSpec(locationMarker: "3️⃣", message: "expected return type in return clause", fixIts: ["insert return type"]),
```
kimdv, line 286: Not sure if all those cases are effects of that I added `return clause` as `nameForDiagnostics`

Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift
ahoppen: Ah, sorry, I forgot that. 

What about making `DeinitializerSignatureError` have three children `name: TokenSyntax?`, `params: FunctionParameterClauseSyntax?` and `returnClause: ReturnClauseSyntax?`. And then `DeinitializerSignatureError` would stitch together the works `name`, `parameters` and `return clause`.

Tests/SwiftParserTest/DeclarationTests.swift
```diff
@@ -913,7 +913,7 @@ final class DeclarationTests: ParserTestCase {
       }
       """,
       diagnostics: [
-        DiagnosticSpec(message: "expected '->' and return type in subscript", fixIts: ["insert '->' and return type"])
+        DiagnosticSpec(message: "expected '->' and return type in return clause", fixIts: ["insert '->' and return type"])
```
ahoppen, line 916: I think I like the old diagnostic better and would thus prefer `returnClause` to not have a `nameForDiagnostic`. “return type in return clause” just sounds weird.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -788,40 +788,44 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
+      nodes.append(name)
+    }
+
+    if let params {
+      nodes.append(params)
+    }
+
+    if let returnType {
+      nodes.append(returnType)
+    }
```
kimdv, line 816: I wasn't able to get it compile by using `let nodes: [any Syntaxprotocol] = [name, params, returnType].compactMap { $0 }`

Matejkob: It appears that you need to cast the default array to make it work. You can do this in one of the following ways:

Option 1:
```swift
let nodes: [AnySyntaxProtocol] = ([name, params, returnType] as [AnySyntaxProtocol?]).compactMap { $0 }
```

Option 2:
```swift
let optionalNodes: [AnySyntaxProtocol?] = [name, params, returnType]
let nodes: [AnySyntaxProtocol] = optionalNodes.compactMap { $0 }
```

kimdv: Hmm, I can't find the `AnySyntaxProtocol` type?
Is that something that isn't on main yet?

Matejkob: Sorry, I'm using Grammarly to correct my message and sometimes it changes code blocks... I've edited my comment

kimdv: Thanks!

ahoppen: What about 

```swift
    let name: TokenSyntax? = node
      .unexpectedBetweenDeinitKeywordAndEffectSpecifiers?
      .presentTokens(satisfying: { $0.tokenKind.isIdentifier == true })
      .only

    let params: FunctionParameterClauseSyntax? = node
      .unexpectedBetweenDeinitKeywordAndEffectSpecifiers?
      .compactMap({ $0.as(FunctionParameterClauseSyntax.self) })
      .only

    let returnType: ReturnClauseSyntax? = node
      .unexpectedBetweenEffectSpecifiersAndBody?
      .compactMap({ $0.as(ReturnClauseSyntax.self) })
      .only

    let nodes = [Syntax(name), Syntax(params), Syntax(returnType)].compactMap { $0 }
```

kimdv:
@swift-ci please test

Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift
```diff
@@ -305,6 +296,30 @@ public struct CannotParseVersionTuple: ParserError {
...
+      descriptions += ["name"]
+    }
+
+    if params != nil {
+      descriptions += ["parameters"]
+    }
+
+    if returnClause != nil {
+      descriptions += ["return clause"]
+    }
```
Matejkob, line 317: Another minor suggestion: How about using the `append` method instead of creating new arrays with single elements? The code works just fine as is, but using `append` could make it feel a bit more like Swift in its natural habitat. For example:
```suggestion
    if name != nil {
      descriptions.append("name")
    }

    if params != nil {
      descriptions.append("parameters")
    }

    if returnClause != nil {
      descriptions.append("return clause")
    }
```

Matejkob (commented):
First off, apologies for not diving deep into the changes earlier. I initially responded to one of your comments and didn't go through the other changes. Now that you've requested a re-review, I've looked at all of them. So, my comments might be coming in a tad late. Feel free to take them or leave them; they're mostly minor suggestions. Overall, the changes look solid to me. That said, I don't feel fully equipped to weigh in on the underlying business logic just yet.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
Matejkob: Nice 😍

kimdv:
> First off, apologies for not diving deep into the changes earlier. I initially responded to one of your comments and didn't go through the other changes. Now that you've requested a re-review, I've looked at all of them. So, my comments might be coming in a tad late. Feel free to take them or leave them; they're mostly minor suggestions. Overall, the changes look solid to me. That said, I don't feel fully equipped to weigh in on the underlying business logic just yet.

No need to apologies. No worries! 
Thanks for the reviews, I've applied them

kimdv:
@swift-ci please test

Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift
```diff
@@ -706,7 +721,7 @@ public struct RemoveRedundantFixIt: ParserFixIt {
 public struct RemoveNodesFixIt: ParserFixIt {
   public let nodesToRemove: [Syntax]
 
-  init(_ nodesToRemove: [some SyntaxProtocol]) {
+  init(_ nodesToRemove: [any SyntaxProtocol]) {
```
ahoppen, line 724: `any SyntaxProtocol` requires us to create an existential, which is quite costly. I would prefer to keep it as-is and instead pass `Syntax` for any kind of heterogeneous list.

Tests/SwiftParserTest/translated/InitDeinitTests.swift
```diff
@@ -125,11 +125,11 @@ final class InitDeinitTests: ParserTestCase {
     assertParse(
       """
       struct FooStructDeinitializerA {
-        deinit 1️⃣x
+        1️⃣deinit x
       }
       """,
       diagnostics: [
-        DiagnosticSpec(message: "deinitializers cannot have a name", fixIts: ["remove 'x'"])
+        DiagnosticSpec(message: "deinitializers cannot have name", fixIts: ["remove 'x'"])
```
ahoppen, line 132: Hmm, this isn’t grammatically correct anymore. I think we need the `a` in case the first element in the list is `name` or `return clause`.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -788,40 +788,35 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
+      .compactMap({ $0.as(ReturnClauseSyntax.self) })
+      .only
+
+    let nodes = [Syntax(name), Syntax(params), Syntax(returnType)].compactMap { $0 }
+
+    if !nodes.isEmpty {
       addDiagnostic(
-        returnType,
-        .deinitCannotHaveReturnType,
+        node,
```
ahoppen, line 811: Could we anchor this at `nodes.first!`? That way the diagnostic actually shows up at the error unexpected nodes instead of the `deinit` keyword. And also, I would add a `highlight` for all the unexpected nodes (would be worth checking in at least one test case as well).

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -788,40 +788,36 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
+      .compactMap({ $0.as(ReturnClauseSyntax.self) })
+      .only
+
+    let nodes = [Syntax(name), Syntax(params), Syntax(returnType)].compactMap { $0 }
+
+    if !nodes.isEmpty {
       addDiagnostic(
-        returnType,
-        .deinitCannotHaveReturnType,
+        nodes.first!,
```
ahoppen, line 811: Ah, one more thought that just crossed my mind.

```swift
    if let firstNode = nodes.first {
      addDiagnostic(
        returnType,
        .deinitCannotHaveReturnType,
        firstNode,
```

Allows us to get rid of the force-unwrap.

ahoppen (approved):

kimdv:
@swift-ci please test windows

PR #2187, opened 2023-09-13
Don’t add any accessors if accessor macro returned an empty array

ahoppen:
If an accessor macro returns an empty array, MacroSystem was failing with internal error messages because it tried ot parse an `AccessorBlockSyntax` from an empty string.

To fix this, check if the expanded source is empty before trying to parse the `AccessorBlockSyntax`.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #2189, opened 2023-09-13
Delete the linkage tests

ahoppen:
The linkage tests have been causing more issues than they caught issues. And really what we want to test, is that the libraries built for the compiler using CMake don’t have any unexpected dependencies – it doesn’t really matter if the swift-syntax package itself gains a new link dependency.

Furthermore, the test was only checking `SwiftSyntax`, `SwiftParser` and `SwiftSyntaxBuilder` and was missing e.g. `SwiftSyntaxMacros`.

We should properly test that no dylibs in the toolchain have unintended linkages as part of integration testing in the apple/swift repo.


bnbarham (approved):

kimdv (approved):

ahoppen:
@swift-ci Please test

PR #2190, opened 2023-09-13
Print environment variables for commands executed by `swift-syntax-dev-utils`

ahoppen:
Environment variables are as much input to the executed subprocess as the parameters and we should be logging them as well in verbose mode.


bnbarham (approved):

kimdv (approved):

ahoppen:
@swift-ci Please test

PR #2192, opened 2023-09-14
WIP: Make sure that we always build swift-syntax consistently in CI

ahoppen:
We were currently setting the `SWIFTSYNTAX_ENABLE_RAWSYNTAX_VALIDATION` and `SWIFTPARSER_ENABLE_ALTERNATE_TOKEN_INTROSPECTION` environment variables when building swift-syntax in CI. But when building e.g. sourcekit-lsp from the same unified workspace, we weren’t setting these environment variables. I’m not actually sure what the behavior here was: Either we re-built swift-syntax (which we don’t want) or we used the already built swift-syntax that had had `RawSyntax` validation enabled, which didn’t actually match the build of sourcekit-lsp that was requested.

In either case, the solution is to have a single environment variable `SWIFT_BUILD_SCRIPT_ENVIRONMENT` that gets set by all projects building in the unfied SwiftPM workspace. That environment variable then enables all the things that we want to have enabled in CI.


ahoppen:
@swift-ci Please test macOS

https://github.com/apple/swift/pull/68506

ahoppen:
https://github.com/apple/swift/pull/68506

@swift-ci Please test macOS

PR #2193, opened 2023-09-14
Fix ExperimentalFeatures raw values

hamishknight:
Looks like we had a race between PRs and ended up assigning the first bit for both 'then' statements and reference bindings. Correct the assignment here.


hamishknight:
@swift-ci please test

PR #2209, opened 2023-09-16
Handle missing `func` keyword

kimdv:
Part of #1373

### DIFF
--- a/Sources/SwiftParser/Declarations.swift
+++ b/Sources/SwiftParser/Declarations.swift
@@ -206,11 +206,24 @@ extension Parser {
       modifiers: self.parseDeclModifierList()
     )
 
-    // If we are inside a memberDecl list, we don't want to eat closing braces (which most likely close the outer context)
-    // while recovering to the declaration start.
-    let recoveryPrecedence = inMemberDeclList ? TokenPrecedence.closingBrace : nil
+    let recoveryResult: (match: DeclarationKeyword, handle: RecoveryConsumptionHandle)?
+    if let atResult = self.at(anyIn: DeclarationKeyword.self) {
+      // We are at a keyword that starts a declaration. Parse that declaration.
+      recoveryResult = (atResult.spec, .noRecovery(atResult.handle))
+    } else if atFunctionDeclarationWithoutFuncKeyword() {
+      // We aren't at a declaration keyword and it looks like we are at a function
+      // declaration. Parse a function declaration.
+      recoveryResult = (.lhs(.func), .missing(.keyword(.func)))
+    } else {
+      // In all other cases, use standard token recovery to find the declaration
+      // to parse.
+      // If we are inside a memberDecl list, we don't want to eat closing braces (which most likely close the outer context)
+      // while recovering to the declaration start.
+      let recoveryPrecedence = inMemberDeclList ? TokenPrecedence.closingBrace : nil
+      recoveryResult = self.canRecoverTo(anyIn: DeclarationKeyword.self, overrideRecoveryPrecedence: recoveryPrecedence)
+    }
 
-    switch self.canRecoverTo(anyIn: DeclarationKeyword.self, overrideRecoveryPrecedence: recoveryPrecedence) {
+    switch recoveryResult {
     case (.lhs(.import), let handle)?:
       return RawDeclSyntax(self.parseImportDeclaration(attrs, handle))
     case (.lhs(.class), let handle)?:
@@ -273,10 +286,7 @@ extension Parser {
         )
       }
 
-      let isPossibleFuncIdentifier = self.at(.identifier, .wildcard)
-      let isPossibleFuncParen = self.peek(isAt: .leftParen, .binaryOperator)
-      // Treat operators specially because they're likely to be functions.
-      if (isPossibleFuncIdentifier && isPossibleFuncParen) || self.at(anyIn: Operator.self) != nil {
+      if atFunctionDeclarationWithoutFuncKeyword() {
         return RawDeclSyntax(self.parseFuncDeclaration(attrs, .missing(.keyword(.func))))
       }
     }
@@ -287,6 +297,24 @@ extension Parser {
         arena: self.arena
       )
     )
+  }
+
+  /// Returns `true` if it looks like the parser is positioned at a function declaration that’s missing the `func` keyword.
+  fileprivate mutating func atFunctionDeclarationWithoutFuncKeyword() -> Bool {
+    var nextTokenIsLeftParenOrLeftAngle: Bool {
+      self.peek(isAt: .leftParen) || self.peek().tokenText.hasPrefix("<")
+    }
+
+    if self.at(.identifier) {
+      return nextTokenIsLeftParenOrLeftAngle
+    } else if self.at(anyIn: Operator.self) != nil {
+      if self.currentToken.tokenText.hasSuffix("<") && self.peek(isAt: .identifier) {
+        return true
+      }
+      return nextTokenIsLeftParenOrLeftAngle
+    } else {
+      return false
+    }
   }
 }
 
@@ -1044,7 +1072,7 @@ extension Parser {
     let identifier: RawTokenSyntax
     if self.at(anyIn: Operator.self) != nil || self.at(.exclamationMark, .prefixAmpersand) {
       var name = self.currentToken.tokenText
-      if !currentToken.isEditorPlaceholder && name.count > 1 && name.hasSuffix("<") && self.peek(isAt: .identifier) {
+      if !currentToken.isEditorPlaceholder && name.hasSuffix("<") && self.peek(isAt: .identifier) {
         name = SyntaxText(rebasing: name.dropLast())
       }
       unexpectedBeforeIdentifier = nil
--- a/Tests/SwiftParserTest/DeclarationTests.swift
+++ b/Tests/SwiftParserTest/DeclarationTests.swift
@@ -2015,7 +2015,7 @@ final class DeclarationTests: ParserTestCase {
     assertParse(
       """
       class A {
-        1️⃣^2️⃣
+        1️⃣^
       }
       class B {
       }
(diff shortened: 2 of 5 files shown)


Sources/SwiftParser/Declarations.swift
```diff
@@ -298,6 +295,13 @@ extension Parser {
       )
     )
   }
+
+  fileprivate mutating func isMissingFuncKeyword() -> Bool {
+    let isPossibleFuncIdentifier = self.at(.identifier, .wildcard)
+    let isPossibleFuncParen = self.peek(isAt: .leftParen, .binaryOperator)
+    // Treat operators specially because they're likely to be functions.
+    return (isPossibleFuncIdentifier && isPossibleFuncParen) || self.at(anyIn: Operator.self) != nil
+  }
```
ahoppen, line 304: 1. I would prefix this function with `at` so that it matches all the other functions that check the current token the parser is at, like `atFunctionDeclarationWithoutFuncKeyword`.
2. I would also add a doc comment.
3. And a couple suggestions to improve its performance since this becomes part of a hot loop now (the optimization is to only call `peek` if `isPossibleFuncIdentifier` is `true`
4. And finally, why do we need to check for `_` as the func identifier? I don’t think that `_` is a valid function name.
5. And we shouldn’t just be checking for a binary operator. It makes more sense to check for the `<` that starts a generic parameter clause here.

```suggestion
  /// Returns `true` if it looks like the parser is positioned at a function declaration that’s missing the `func` keyword.
  fileprivate mutating func atFunctionDeclarationWithoutFuncKeyword() -> Bool {
    if self.at(.identifier) && (self.peek(isAt: .leftParen) || self.peek().tokenText.hasPrefix("<")) {
      return true
    }
    if self.at(anyIn: Operator.self) != nil {
      return true
    }
    return false
  }
```

kimdv: Not sure about number 4. It was already there.

Tests/SwiftParserTest/DeclarationTests.swift
```diff
@@ -452,18 +452,20 @@ final class DeclarationTests: ParserTestCase {
...
+        DiagnosticSpec(locationMarker: "3️⃣", message: "expected ':' and type in parameter", fixIts: ["insert ':' and type"]),
       ],
       fixedSource: """
-        private(set) +
-          set
+        private(set) func + (
+          set: <#type#>
         ) var a = 0
         """
     )
```
kimdv, line 471: The changes affected this test case. 
Not sure if it got worse or better. 

I think it would be nice if the `+` would be unexpected code, and we could parse the rest correctly

ahoppen: Ah, I hoped that we could get around checking whether the token after the operator is `(` or `<` because it also requires us to do the check if the operator itself ends with `<` (because if you have eg. `%%<` then `<` is part of the operator).

I did investigation and figured that also we should only employ the recovery if we are not at a `DeclarationKeyword`, otherwise we incorrectly parsed `_mutating (a, b)` as a function declaration while it should be parsed as a variable declaration if the experimental bindings flag is enabled.


Anyway, here’s the diff I came up with

<details><summary>Diff</summary>
<p>

```diff
diff --git a/Sources/SwiftParser/Declarations.swift b/Sources/SwiftParser/Declarations.swift
index 801243da..8e36abab 100644
--- a/Sources/SwiftParser/Declarations.swift
+++ b/Sources/SwiftParser/Declarations.swift
@@ -208,15 +208,25 @@ extension Parser {
       modifiers: self.parseDeclModifierList()
     )
 
-    if atFunctionDeclarationWithoutFuncKeyword() {
-      return RawDeclSyntax(self.parseFuncDeclaration(attrs, .missing(.keyword(.func))))
-    }
+    let recoveryResult: (match: DeclarationKeyword, handle: RecoveryConsumptionHandle)?
+    if let atResult = self.at(anyIn: DeclarationKeyword.self) {
+      // We are at a keyword that starts a declaration. Parse that declaration.
+      recoveryResult = (atResult.spec, .noRecovery(atResult.handle))
+    } else if atFunctionDeclarationWithoutFuncKeyword() {
+      // We aren't at a declaration keyword and it looks like we are at a function
+      // declaration. Parse a function declaration.
+      recoveryResult = (.lhs(.func), .missing(.keyword(.func)))
+    } else {
+      // In all other cases, use standard token recovery to find the declaration
+      // to parse.
 
-    // If we are inside a memberDecl list, we don't want to eat closing braces (which most likely close the outer context)
-    // while recovering to the declaration start.
-    let recoveryPrecedence = inMemberDeclList ? TokenPrecedence.closingBrace : nil
+      // If we are inside a memberDecl list, we don't want to eat closing braces (which most likely close the outer context)
+      // while recovering to the declaration start.
+      let recoveryPrecedence = inMemberDeclList ? TokenPrecedence.closingBrace : nil
+      recoveryResult = self.canRecoverTo(anyIn: DeclarationKeyword.self, overrideRecoveryPrecedence: recoveryPrecedence)
+    }
 
-    switch self.canRecoverTo(anyIn: DeclarationKeyword.self, overrideRecoveryPrecedence: recoveryPrecedence) {
+    switch recoveryResult {
     case (.lhs(.import), let handle)?:
       return RawDeclSyntax(self.parseImportDeclaration(attrs, handle))
     case (.lhs(.class), let handle)?:
@@ -302,13 +312,19 @@ extension Parser {
 
   /// Returns `true` if it looks like the parser is positioned at a function declaration that’s missing the `func` keyword.
   fileprivate mutating func atFunctionDeclarationWithoutFuncKeyword() -> Bool {
-    if self.at(.identifier) && (self.peek(isAt: .leftParen) || self.peek().tokenText.hasPrefix("<")) {
-      return true
+    var nextTokenIsLeftParenOrLeftAngle: Bool {
+      self.peek(isAt: .leftParen) || self.peek().tokenText.hasPrefix("<")
     }
-    if self.at(anyIn: Operator.self) != nil {
-      return true
+    if self.at(.identifier) {
+      return nextTokenIsLeftParenOrLeftAngle
+    } else if self.at(anyIn: Operator.self) != nil {
+      if self.currentToken.tokenText.hasSuffix("<") && self.peek(isAt: .identifier) {
+        return true
+      }
+      return nextTokenIsLeftParenOrLeftAngle
+    } else {
+      return false
     }
-    return false
   }
 }
 
@@ -1065,7 +1081,7 @@ extension Parser {
     let identifier: RawTokenSyntax
     if self.at(anyIn: Operator.self) != nil || self.at(.exclamationMark, .prefixAmpersand) {
       var name = self.currentToken.tokenText
-      if name.count > 1 && name.hasSuffix("<") && self.peek(isAt: .identifier) {
+      if name.hasSuffix("<") && self.peek(isAt: .identifier) {
         name = SyntaxText(rebasing: name.dropLast())
       }
       unexpectedBeforeIdentifier = nil
diff --git a/Sources/SwiftParser/Parser.swift b/Sources/SwiftParser/Parser.swift
index 9cf9aaad..164c02e6 100644
--- a/Sources/SwiftParser/Parser.swift
+++ b/Sources/SwiftParser/Parser.swift
@@ -443,12 +443,9 @@ extension Parser {
   /// Checks if it can reach a token of the given `spec` by skipping unexpected
   /// tokens that have lower ``TokenPrecedence`` than specified by `spec`.
   @inline(__always)
-  mutating func canRecoverTo(_ spec: TokenSpec) -> RecoveryConsumptionHandle? {
+  mutating func gcanRecoverTo(_ spec: TokenSpec) -> RecoveryConsumptionHandle? {
     if self.at(spec) {
-      return RecoveryConsumptionHandle(
-        unexpectedTokens: 0,
-        tokenConsumptionHandle: TokenConsumptionHandle(spec: spec)
-      )
+      return .constant(spec)
     }
     var lookahead = self.lookahead()
     return lookahead.canRecoverTo(spec)
@@ -465,7 +462,7 @@ extension Parser {
     overrideRecoveryPrecedence: TokenPrecedence? = nil
   ) -> (match: SpecSet, handle: RecoveryConsumptionHandle)? {
     if let (kind, handle) = self.at(anyIn: specSet) {
-      return (kind, RecoveryConsumptionHandle(unexpectedTokens: 0, tokenConsumptionHandle: handle))
+      return (kind, .noRecovery(handle))
     }
     var lookahead = self.lookahead()
     return lookahead.canRecoverTo(anyIn: specSet, overrideRecoveryPrecedence: overrideRecoveryPrecedence)
diff --git a/Sources/SwiftParser/Recovery.swift b/Sources/SwiftParser/Recovery.swift
index 58b91a5d..9ef52653 100644
--- a/Sources/SwiftParser/Recovery.swift
+++ b/Sources/SwiftParser/Recovery.swift
@@ -14,13 +14,20 @@
 
 // MARK: Lookahead
 
-/// After calling `consume(ifAnyFrom:)` we know which token we are positioned
+/// After calling `canRecoverTo(anyIn:)` we know which token we are positioned
 /// at based on that function's return type. This handle allows consuming that
 /// token.
 struct RecoveryConsumptionHandle {
   var unexpectedTokens: Int
   var tokenConsumptionHandle: TokenConsumptionHandle
 
+  fileprivate init(unexpectedTokens: Int, tokenConsumptionHandle: TokenConsumptionHandle) {
+    self.unexpectedTokens = unexpectedTokens
+    self.tokenConsumptionHandle = tokenConsumptionHandle
+  }
+
+  /// A `RecoveryConsumptionHandle` that doesn't skip over any unexpected tokens
+  /// and consumes a token matching `spec`.
   static func constant(_ spec: TokenSpec) -> RecoveryConsumptionHandle {
     return RecoveryConsumptionHandle(
       unexpectedTokens: 0,
@@ -28,6 +35,15 @@ struct RecoveryConsumptionHandle {
     )
   }
 
+  /// A `RecoveryConsumptionHandle` that doesn't skip over any unexpected tokens
+  /// and consumes `handle`.
+  static func noRecovery(_ handle: TokenConsumptionHandle) -> RecoveryConsumptionHandle {
+    return RecoveryConsumptionHandle(
+      unexpectedTokens: 0,
+      tokenConsumptionHandle: handle
+    )
+  }
+
   /// A `RecoveryConsumptionHandle` that will not eat any tokens but instead
   /// synthesize a missing token of kind `token`.
   static func missing(_ spec: TokenSpec) -> RecoveryConsumptionHandle {
diff --git a/Tests/SwiftParserTest/DeclarationTests.swift b/Tests/SwiftParserTest/DeclarationTests.swift
index 129bf46e..6a0aec1c 100644
--- a/Tests/SwiftParserTest/DeclarationTests.swift
+++ b/Tests/SwiftParserTest/DeclarationTests.swift
@@ -452,20 +452,18 @@ final class DeclarationTests: ParserTestCase {
 
     assertParse(
       """
-      private(1️⃣+2️⃣
-        set3️⃣
+      private(1️⃣+
+        set
       ) var a = 0
       """,
       diagnostics: [
-        DiagnosticSpec(locationMarker: "1️⃣", message: "expected 'set)' to end modifier", fixIts: ["insert 'set)'"]),
-        DiagnosticSpec(locationMarker: "1️⃣", message: "expected 'func' in function", fixIts: ["insert 'func'"]),
-        DiagnosticSpec(locationMarker: "2️⃣", message: "expected '(' to start parameter clause", fixIts: ["insert '('"]),
+        DiagnosticSpec(message: "expected 'set)' to end modifier", fixIts: ["insert 'set)'"]),
         // FIXME: It should print `+` as detail of text.
-        DiagnosticSpec(locationMarker: "3️⃣", message: "expected ':' and type in parameter", fixIts: ["insert ':' and type"]),
+        DiagnosticSpec(message: "unexpected code in variable"),
       ],
       fixedSource: """
-        private(set) func + (
-          set: <#type#>
+        private(set) +
+          set
         ) var a = 0
         """
     )
```

</p>
</details>

kimdv:
@swift-ci please test

kimdv:
@ahoppen it should be in a state where I think it's ready for review.

ahoppen (approved):
LGTM

PR #2211, opened 2023-09-18
Teach TokenSpecSet about experimental features

hamishknight:
Pass down the current set of experimental features to TokenSpecSets, and have generated TokenSpecSets automatically account for experimental features by returning `nil` if a given feature isn't active. Should be NFC.


hamishknight:
@swift-ci please test

hamishknight:
I can say I _love_ having to pass down experimental features to each TokenSpecSet, but I think it's ultimately better having them automatically handle experimental features instead of needing ad-hoc checks in the parser which, as demonstrated recently, can be error prone

hamishknight:
@swift-ci please test

hamishknight:
@swift-ci please test

Sources/SwiftParser/TokenConsumer.swift
```diff
@@ -148,6 +148,17 @@ extension TokenConsumer {
     return nil
   }
 
+  /// Checks whether the parser's next token is any token in `Subset`.
+  /// If this is the case, return the corresponding `Subset` case.
+  @inline(__always)
+  func nextTokenAt<SpecSet: TokenSpecSet>(anyIn specSet: SpecSet.Type) -> SpecSet? {
```
ahoppen, line 154: The consistent naming of this method would be `peek(isAtAnyIn:)`.

Sources/SwiftParser/Statements.swift
```diff
@@ -943,14 +943,13 @@ extension Parser.Lookahead {
...
-    let next = peek()
-
     // If 'then' is followed by a binary or postfix operator, prefer to parse as
     // an expr.
-    if BinaryOperatorLike(lexeme: next) != nil || PostfixOperatorLike(lexeme: next) != nil {
+    if nextTokenAt(anyIn: BinaryOperatorLike.self) != nil || nextTokenAt(anyIn: PostfixOperatorLike.self) != nil {
       return false
     }
 
+    let next = peek()
```
ahoppen, line 952: If you’re already touching this, you can just remove `next` completely and replace both usages by `peek()`. Calling `peek` is cheap enough that we don’t need to store it in a variable.

Sources/SwiftParser/TokenSpecSet.swift
```diff
@@ -17,21 +17,22 @@
 protocol TokenSpecSet: CaseIterable {
   var spec: TokenSpec { get }
 
-  /// Creates an instance if `lexeme` satisfy the condition of this subset.
-  init?(lexeme: Lexer.Lexeme)
+  /// Creates an instance if `lexeme` satisfy the condition of this subset,
```
ahoppen, line 20:
```suggestion
  /// Creates an instance if `lexeme` satisfies the condition of this subset,
```

ahoppen (commented):
Thanks for doing this. I agree that passing the experimental feature list to `TokenSpecSet` isn’t great but it’s definitely less error-prone.

I measured performance and couldn’t see any regression when parsing NetNewsWire.

hamishknight:
@swift-ci please test Windows

hamishknight:
@swift-ci please test Windows

ahoppen (approved):

hamishknight:
@swift-ci please test macOS

hamishknight:
@swift-ci please test macOS

PR #2213, opened 2023-09-18
Improve assertMacroExpansion highlight verification

rauhul:
Adds a new initializer to DiagnosticSpec which takes an optional array of highlights instead of a single optional highlight string. Updates existing initializer to map the string into an array of one.

Updates assertMacroExpansion to check each highlight in the diagnostic spec against the actual produced diagnostic individually. Additionally changes the behavior to drop leading and trailing trivial when performing the highlight comparison to better match the Swift compiler's behavior.

### DIFF
--- a/Tests/SwiftSyntaxMacrosTestSupportTests/AssertionsTests.swift
+++ b/Tests/SwiftSyntaxMacrosTestSupportTests/AssertionsTests.swift
@@ -56,7 +56,7 @@ final class AssertionsTests: XCTestCase {
             message: OnlyStruct(),
             highlights: [
               Syntax(structDecl.structKeyword),
-              Syntax(structDecl.name)
+              Syntax(structDecl.name),
             ]
           )
         )


rauhul:
This is behavior breaking iff:
- you use explicit highlight checking (it seems like most users don't) and
  - your explicit highlights had trailing space to match the misreported highlights from swift-syntax or
  - your diagnostics emit multiple highlights, which now need to be passed as an array instead of string with odd concatenation
  
IMO this should be considered a bug fix because the current highlight checking doesn't actually allow the user to be confident in their highlight assertion.

ahoppen (commented):
Looks good to me. Could you add a test case that tests passing multiple `highlights` to `DiagnosticSpec`?

Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
```diff
@@ -129,7 +129,7 @@ public struct DiagnosticSpec {
   public let severity: DiagnosticSeverity
 
   /// If not `nil`, the text the diagnostic is expected to highlight
```
ahoppen, line 131:
```suggestion
  /// If not `nil`, the text fragments the diagnostic is expected to highlight
```

rauhul: fixed in `cd73ad23cc9502efcb654c2d3da4c3de95942dfd`

rauhul:
@swift-ci please test

ahoppen:
@swift-ci Please test Windows

rauhul:
The linux error seems unrelated to me:
```
error: multiple targets named '_InstructionCounter' in: 'swift-format', 'swift-syntax'; consider using the `moduleAliases` parameter in manifest to provide unique names
```

ahoppen:
Yes, fixed by https://github.com/apple/swift-format/pull/632.

ahoppen:
@swift-ci Please test

ahoppen (commented):
I have two comments about API compatibility. If you could address those and add the change to `Release Notes/510.md`, we can get this merged.

rauhul:
@swift-ci please test

rauhul:
@ahoppen could you take another look when you have some time?

rauhul:
@swift-ci please test

ahoppen (approved):
Good to merge 👍🏽

rauhul:
@swift-ci please test

Issue #2166, opened 2023-09-07
GroupedDiagnostics does not handle notes on its diagnostics


DougGregor:
### Description

The `GroupedDiagnostics` formatter takes an array of `Diagnostic`s that it formats into a buffer. However, it only handles the top-level diagnostic messages, not any [attached notes](https://github.com/apple/swift-syntax/blob/main/Sources/SwiftDiagnostics/Diagnostic.swift#L30C3-L31C1)---which also have locations and messages and should be treated the same way as top-level diagnostics.

### Steps to Reproduce

To reproduce, pass a `Diagnostic` with a non-empty `notes` member into the grouped diagnostics formatter, and note that the note doesn't show up at all.

The following patch updates the test infrastructure for `GroupedDiagnostics` to make it easier to create notes associated with the diagnostics, demonstrating this issue. 

```
diff --git a/Tests/SwiftDiagnosticsTest/GroupDiagnosticsFormatterTests.swift b/T
ests/SwiftDiagnosticsTest/GroupDiagnosticsFormatterTests.swift
index eccd50dc1..461bc2333 100644
--- a/Tests/SwiftDiagnosticsTest/GroupDiagnosticsFormatterTests.swift
+++ b/Tests/SwiftDiagnosticsTest/GroupDiagnosticsFormatterTests.swift
@@ -27,6 +27,32 @@ extension SimpleDiagnosticMessage: FixItMessage {
   var fixItID: MessageID { diagnosticID }
 }
 
+struct SimpleNoteMessage: NoteMessage {
+  let message: String
+  let fixItID: MessageID
+}
+
+/// The severity of a particular marked diagnostic. This differs from
+/// DiagnosticSeverity in that notes are associated within another marked
+/// diagnostic.
+enum MarkedSeverity {
+  case error
+  case warning
+  case remark
+
+  /// Note attached to the diagnostic at a particular location
+  case note(String)
+
+  var asSeverity: DiagnosticSeverity {
+    switch self {
+      case .error: return .error
+      case .warning: return .warning
+      case .remark: return .remark
+      case .note: fatalError("Notes should never be mapped this way")
+    }
+  }
+}
+
 extension GroupedDiagnostics {
   /// Add a new test file to the group, starting with marked source and using
   /// the markers to add any suggested extra diagnostics at the marker
@@ -35,29 +61,58 @@ extension GroupedDiagnostics {
     _ markedSource: String,
     displayName: String,
[...]

PR #2214, opened 2023-09-18
[WIP] Improve `DiagnosticsFormatter`

Matejkob:
In my side-project, I've encountered the same issue that @DougGregor mentioned in #2166, so I've taken the liberty to address it.

At this stage, the PR is purely for discussing the changes; it's far from being the final version.


Tests/SwiftDiagnosticsTest/NewDiagnosticsFormatterTests.swift
```diff
@@ -0,0 +1,222 @@
...
+import SwiftParser
+
+struct DiagnosticDescriptor {
+  let locationMarker: String
+  let id: MessageID
+  let message: String
+  let severity: DiagnosticSeverity
+  let highlight: [Syntax] // TODO: How to create an abstract model for this?
+  let noteDescriptors: [NoteDescriptor]
+  let fixIts: [FixIt] // TODO: How to create an abstract model for this?
```
Matejkob, line 27: @DougGregor in #2166 mentioned:
> Note: a similar issue exists with Fix-Its, where we'd need a way to render what they actually do.

Do we already have a way to render this somewhere?

Tests/SwiftDiagnosticsTest/NewDiagnosticsFormatterTests.swift
```diff
@@ -0,0 +1,222 @@
...
+      ],
+      matches: """
+        1 │ func foo() -> Int {
+        2 │   if 1 != 0 {
+          │      │      ╰─ note: Other message
+          │      ├─ error: My message goes here!
+          │      ├─ note: First message
+          │      ╰─ note: Second message
+        3 │     return 0
+        4 │   }
```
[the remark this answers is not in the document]
ahoppen: Notes are usually used for things like “Previously declared” here if there’s a duplicate declaration. For example, you can trigger a note in the compiler using 

```bash
$ cat /tmp/a.swift
func test() {}

func test() {}
$ swift /tmp/a.swift
/tmp/a.swift:3:6: error: invalid redeclaration of 'test()'
func test() {}
     ^
/tmp/a.swift:1:6: note: 'test()' previously declared here
func test() {}
     ^
```

Rendering where the note came from is crucial here.

Tests/SwiftDiagnosticsTest/NewDiagnosticsFormatterTests.swift
```diff
@@ -0,0 +1,222 @@
...
+        ),
+      ],
+      matches: """
+        1 │ func foo() -> Int {
+          |      ╰─ note: Second message // This note comes from from diagnostic, how to present it?
+        2 │   if 1 != 0 {
+          │      ├─ error: My message goes here!
+          │      ╰─ note: First message
+        3 │     return 0
+        4 │   }
```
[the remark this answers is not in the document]
ahoppen: I think this looks good if we use `GroupedDiagnosticFormatter`. The idea would be that each error basically gets its own printout of the source (we wouldn’t be merging multiple diagnostics into a single source printout anymore) and thus it’s clear which diagnostic the note belongs to.

Matejkob:
@ahoppen @DougGregor @bnbarham @kimdv 
I've introduced a substantial number of changes and am now opening this PR for an initial round of reviews. While the code may have minor issues, I'd appreciate any general feedback before diving into the finer details.

ahoppen:
Could you split this PR into multiple smaller PRs? 1800 is just quite hard to review. I think commits 1 and 2 would make a good PR by themselves.

Matejkob:
> Could you split this PR into multiple smaller PRs? 1800 is just quite hard to review. I think commits 1 and 2 would make a good PR by themselves.

I've split commit 1 and 2 into separate PR: #2238

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
```diff
@@ -12,311 +12,811 @@
...
+          return nil
+        }
+
+        return startColumn..<endColumn
+      }
+      .sorted { (lhs, rhs) in
+        lhs.lowerBound < rhs.lowerBound
       }
+      .mergingOverlappingRanges()
+  }
```
Matejkob, line 547: This method will be further refactored in the #2242

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
```diff
@@ -12,311 +12,811 @@
...
+    Self.annotatedSource(
+      inSyntaxTree: tree,
+      withDiagnostics: diags,
+      limitedToContextRange: .limited(contextSize),
+      withDiagnosticDecorator: colorize ? .ANSI : .basic
     )
   }
 }
+
+extension Sequence where Element == Range<Int> {
```
ahoppen, line 787: Since you’re refactoring this already, I think this can be shortened to 

```suggestion
extension Sequence<Range<Int>> {
```

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
```diff
@@ -12,311 +12,811 @@
...
+/// This formatter enables better development practices by:
+///
+/// - **Improving Debugging**: Diagnostic messages, including errors, warnings, notes, fix-its, highlights, and remarks,
+/// are aligned in context to the source code, making it easier to understand what needs to be corrected or considered.
+///
+/// - **Enhanced Readability**: The formatter enriches the source code with line numbers,
+/// annotations, and optional text suffixes, providing a full picture to the developer.
+///
+/// - **Flexible Customization**: Through the `DiagnosticDecorator` protocol, you can customize
+/// the appearance of diagnostic output.
```
ahoppen, line 71: Nitpick

```suggestion
/// - **Improving Debugging**: Diagnostic messages, including errors, warnings, notes, fix-its, highlights, and remarks,
///   are aligned in context to the source code, making it easier to understand what needs to be corrected or considered.
///
/// - **Enhanced Readability**: The formatter enriches the source code with line numbers,
///   annotations, and optional text suffixes, providing a full picture to the developer.
///
/// - **Flexible Customization**: Through the `DiagnosticDecorator` protocol, you can customize
///   the appearance of diagnostic output.
```

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
```diff
@@ -12,311 +12,811 @@
...
-    /// that occurs on that line.
-    var suffixText: String
-
-    /// Whether this line is free of annotations.
-    var isFreeOfAnnotations: Bool {
-      return diagnostics.isEmpty && suffixText.isEmpty
-    }
+  /// An enumeration that describes the amount of contextual lines to display around each diagnostic message.
+  ///
+  /// - `limited(Int)`: Display a limited number of lines around each diagnostic. The integer value specifies the number of lines before and after the diagnostic message.
```
ahoppen, line 75: I know that swift-format doesn’t complain about overly long comment lines but would be good to manually break these lines so they are < 160 columns

---

Also, I think the doc comments for the enum cases should be put on those cases instead of on the `ContextRange` enum itself.

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
```diff
@@ -12,311 +12,811 @@
...
+  /// - `full`: Display the entire source code along with the diagnostics. Useful for getting a full overview, but could result in a large output.
+  enum ContextRange {
+    case limited(Int)
+    case full
```
ahoppen, line 79: When do you think the full context range could be useful? I think recently, we have been aiming to only display a single error/warning + its inline buffer. If we display multiple errors inside the same context, it’s also not clear which notes belong to which errors.

CC @DougGregor

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
```diff
@@ -12,311 +12,811 @@
...
-    diags: [Diagnostic],
-    contextSize: Int = 2,
-    colorize: Bool = false
-  ) -> String {
-    let formatter = DiagnosticsFormatter(contextSize: contextSize, colorize: colorize)
-    return formatter.annotatedSource(tree: tree, diags: diags)
+  /// Initializes a new `DiagnosticsFormatter` instance.
+  ///
+  /// - Parameters:
+  ///   - contextSize: Specifies the number of contextual lines around each diagnostic message. Default is `.limited(2)`.
```
ahoppen, line 101:
```suggestion
  ///   - contextRange: Specifies the number of contextual lines around each diagnostic message. Default is `.limited(2)`.
```

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
```diff
@@ -12,311 +12,811 @@
...
+    switch contextRange {
+    case .limited(let contextSizeLimit):
+      return
+        annotatedSourceLines
+        .compactMap { annotatedLine -> Range<Int>? in
+          let lineNumber = annotatedLine.sourceLineNumber
+          if annotatedLine.isFreeOfAnnotations {
+            return nil
+          } else {
+            return Range<Int>(uncheckedBounds: (lower: lineNumber - contextSizeLimit, upper: lineNumber + contextSizeLimit + 1))
```
ahoppen, line 332: I don’t think that constructing the range is a performance bottleneck here, so I would prefer to not use the `unecheckedBounds` initializers. In case something goes wrong (for example `contextSizeLimit` could be negative), it’s much better to crash here than somewhere further downstream.


```suggestion
            return (lineNumber - contextSizeLimit)..<(lineNumber + contextSizeLimit + 1)
```

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
```diff
@@ -12,311 +12,811 @@
...
+          }
         }
+    case .full:
+      return [Range<Int>(uncheckedBounds: (lower: 0, upper: annotatedSourceLines.count))]
```
ahoppen, line 336: Same here. Let’s not use the unchecked initializer. It’s also easier to read.

```suggestion
      return [0..<annotatedSourceLines.count]
```

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
```diff
@@ -12,311 +12,811 @@
...
-    }.mergingOverlappingRanges()
+      // Append a decorated current source line
+      appendDecoratedSourceLine(
+        to: &annotatedSource,
+        from: annotatedLine,
+        within: tree,
+        via: sourceLocationConverter,
+        indentingWith: indentString,
+        upToMaxDigits: maxNumberOfDigits
+      )
```
ahoppen, line 218: I think it would be cleaner to make `appendDecoratedSourceLine` return a `String` and just append that to `annotatedSource` here.

```suggestion
      // Append a decorated current source line
      annotatedSource += decoratedSourceLine(
        from: annotatedLine,
        within: tree,
        via: sourceLocationConverter,
        indentingWith: indentString,
        upToMaxDigits: maxNumberOfDigits
      )
```

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
```diff
@@ -12,311 +12,811 @@
...
-      if startColumn == endColumn {
-        return nil
+      /// Add a skipped line marker to indicate that lines have been omitted, but only under two conditions:
+      /// 1. Lines have actually been skipped ('hasLineBeenSkipped' is true).
+      /// 2. There's already annotated content in 'annotatedSource', meaning this isn't the start of the annotated lines.
+      /// This ensures that skipped line markers only appear between annotated lines.
+      if hasLineBeenSkipped && !annotatedSource.isEmpty {
+        let skippedLineMarker = diagnosticDecorator.decorateBufferOutline("┆")
+        let skippedLine = indentString + String(repeating: " ", count: maxNumberOfDigits) + " " + skippedLineMarker + "\n"
+        annotatedSource.append(skippedLine)
```
ahoppen, line 206: Nit: I generally prefer `+=` to `append`. But feel free to leave as-is if you disagree. Same in a couple more places.

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
```diff
@@ -12,311 +12,811 @@
...
-    // Form the annotated string by copying in text from the original source,
-    // highlighting the column ranges.
-    var resultSourceString: String = ""
-    let sourceString = annotatedLine.sourceString
-    var sourceIndex = sourceString.startIndex
-    for highlightRange in highlightIndexRanges {
-      // Text before the highlight range
-      resultSourceString += sourceString[sourceIndex..<highlightRange.lowerBound]
+      // Append all diagnostic messages for the current line
+      appendFormattedDiagnostics(
```
ahoppen, line 226: Same here, I think would be cleaner as 

```suggestion
      annotatedSource += formattedDiagnostics(
```

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
```diff
@@ -12,311 +12,811 @@
...
 
-    // Map the column ranges into index ranges within the source string itself.
-    let sourceStringUTF8 = annotatedLine.sourceString.utf8
-    let highlightIndexRanges: [Range<String.Index>] = highlightRanges.map { highlightRange in
-      let startIndex = sourceStringUTF8.index(sourceStringUTF8.startIndex, offsetBy: highlightRange.lowerBound - 1)
-      let endIndex = sourceStringUTF8.index(startIndex, offsetBy: highlightRange.count)
-      return startIndex..<endIndex
-    }
+      // If the line did not end with new line (e.g. the last line), append it manually
+      if annotatedSource.last?.isNewline == false {
```
ahoppen, line 221: I generally find that my brain takes a lot more cycles to parse these `== true/false` comparisons on optionals than if using `??` to provide a default value for the `nil` case.

```suggestion
      if !(annotatedSource.last?.isNewline ?? false) {
```

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
```diff
@@ -12,311 +12,811 @@
...
+      )
 
-      // Highlighted source text
-      let highlightString = String(sourceString[highlightRange])
-      resultSourceString += diagnosticDecorator.decorateHighlight(highlightString).highlightedSourceCode
+      // Add suffix text.
+      annotatedSource.append(annotatedLine.suffixText)
 
-      sourceIndex = highlightRange.upperBound
+      if annotatedSource.last?.isNewline == false {
```
ahoppen, line 237:
```suggestion
      if !(annotatedSource.last?.isNewline ?? false) {
```

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
```diff
@@ -12,311 +12,811 @@
...
+  /// ```
+  private func appendDecoratedSourceLine(
+    to annotatedSource: inout String,
+    from annotatedLine: AnnotatedSourceLine,
+    within tree: SyntaxProtocol,
+    via sourceLocationConverter: SourceLocationConverter,
+    indentingWith indentString: String,
+    upToMaxDigits maxNumberOfDigits: Int
+  ) {
+    // 1. Decorate source line with highlights
```
ahoppen, line 375: Nit: I wouldn’t do `1.` etc. Just makes it harder to change if we ever need to introduce anything in between 1. and 2.

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
```diff
@@ -12,311 +12,811 @@
...
+    var highlightedSourceCode = ""
+    var additionalHighlightedLine = ""
+    var lastIndex = sourceString.startIndex
+
+    // Iterate over each range and decorate it.
+    for highlightIndexRange in highlightIndexRanges {
+      // Append the non-highlighted part to the result string.
+      let nonHighlighted = sourceString[lastIndex..<highlightIndexRange.lowerBound]
+      highlightedSourceCode += nonHighlighted
+      additionalHighlightedLine += Array(repeating: " ", count: nonHighlighted.count).joined()
```
ahoppen, line 450: No need to use `Array` here

```suggestion
      additionalHighlightedLine += String(repeating: " ", count: nonHighlighted.count)
```

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
```diff
@@ -12,311 +12,811 @@
...
+      let (decoratedHighlight, decoratedAdditionalLine) = diagnosticDecorator.decorateHighlight(highlight)
+
+      // Append the decorated highlight to the result string.
+      highlightedSourceCode += decoratedHighlight
+
+      // If there's an additional line, append it.
+      if let decoratedAdditionalLine {
+        additionalHighlightedLine += decoratedAdditionalLine
+      } else {
+        additionalHighlightedLine += Array(repeating: " ", count: decoratedHighlight.count).joined()
```
ahoppen, line 463:
```suggestion
        additionalHighlightedLine += String(repeating: " ", count: decoratedHighlight.count)
```

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
```diff
@@ -12,311 +12,811 @@
...
+    }
 
-      // If necessary, print a line that indicates that there was lines skipped in the source code
-      if hasLineBeenSkipped && !annotatedSource.isEmpty {
-        let lineMissingInfoLine = indentString + String(repeating: " ", count: maxNumberOfDigits) + " \(diagnosticDecorator.decorateBufferOutline("┆"))"
-        annotatedSource.append("\(lineMissingInfoLine)\n")
+    // Append any remaining part of the source string.
+    let remaining = sourceString[lastIndex...]
+    highlightedSourceCode += remaining
+    additionalHighlightedLine += Array(repeating: " ", count: remaining.count).joined()
```
ahoppen, line 472:
```suggestion
    additionalHighlightedLine += String(repeating: " ", count: remaining.count)
```

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
```diff
@@ -12,311 +12,811 @@
...
+  ) -> [Range<Int>] {
+    annotatedSourceLine.diagnostics
+      .compactMap {
+        switch $0 {
+        case let .diagnostic(diagnostic):
+          return diagnostic.highlights
+        case .note:
+          return nil
+        }
       }
```
ahoppen, line 506: Instead of `compactMap {}.joined()` I think you can just do

```suggestion
      .flatMap {
        switch $0 {
        case let .diagnostic(diagnostic):
          return diagnostic.highlights
        case .note:
          return []
        }
      }
```

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
```diff
@@ -12,311 +12,811 @@
...
+      withDiagnostics: diagnostics,
+      usingIndentString: indentString,
+      appendingSuffixTexts: suffixTexts,
+      employingSourceLocationConverter: sourceLocationConverter
+    )
+  }
+}
+
+/// Extends `DiagnosticsFormatter` with internal helper types for working with diagnostics.
+extension DiagnosticsFormatter {
```
ahoppen, line 651: Since all of these are `private` and thus not visible outside of this file, I think we can also make them top-level types, reducing the nesting level by one.

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
```diff
@@ -12,311 +12,811 @@
...
+  /// Consolidates different types of diagnostic information, such as messages and notes, under a unified structure.
+  ///
+  /// `FlattenDiagnostic` offers a unified approach to managing and processing various forms of diagnostic information.
+  /// It abstracts away the differences between diagnostic messages and informational notes, allowing for seamless integration
+  /// and handling within the diagnostics formatting process.
+  private enum FlattenDiagnostic {
+    /// Represents a diagnostic message.
+    case diagnostic(Diagnostic)
+
+    /// Represents an informational note from a diagnostic(
```
ahoppen, line 661:
```suggestion
    /// Represents an informational note from a diagnostic
```

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
```diff
@@ -12,311 +12,811 @@
...
+  /// A representation of a single source line, potentially enriched with diagnostics or annotations.
+  ///
+  /// `AnnotatedSourceLine` encapsulates a line of source code alongside any associated diagnostics (errors, notes, etc.)
+  /// or annotations for enhanced code readability and insight. The diagnostics are represented by the `FlattenDiagnostic`
+  /// enum, which generalizes the handling of diagnostic messages and informational notes, providing a unified
+  /// approach to represent and process various types of diagnostic information.
+  ///
+  /// - Note: This struct is designed for internal use by the `DiagnosticsFormatter` to organize and manage
+  ///         source lines with their related diagnostics and annotations, facilitating a clearer understanding
+  ///         of code issues and contextual notes.
```
ahoppen, line 704: Nit

```suggestion
  /// - Note: This struct is designed for internal use by the `DiagnosticsFormatter` to organize and manage
  ///  source lines with their related diagnostics and annotations, facilitating a clearer understanding
  ///  of code issues and contextual notes.
```

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
```diff
@@ -12,311 +12,811 @@
...
     }
-    return annotatedSource
+  }
+}
+
+// Extension providing all necessary APIs to maintain backward compatibility. In the future,
+// upon committing to the new API of `DiagnosticsFormatter` and diagnostic decorators,
+// all members will be marked as deprecated.
+extension DiagnosticsFormatter {
+  public var contextSize: Int {
```
ahoppen, line 747: I think we should deprecate this and make `contextRange` public instead. 

```suggestion
  public var contextSize: Int {
```

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
```diff
@@ -12,311 +12,811 @@
...
+      return -1
+    }
+  }
+
+  @available(*, deprecated, message: "Store the `colorize` property passed to the initializer instead")
+  public var colorize: Bool {
+    diagnosticDecorator is ANSIDiagnosticDecorator
+  }
+
+  public init(contextSize: Int = 2, colorize: Bool = false) {
```
ahoppen, line 761: Let’s deprecate this and make the `ContextRange` initializer public instead.

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
```diff
@@ -12,311 +12,811 @@
...
   ) -> String {
-    return annotatedSource(
-      tree: tree,
-      diags: diags,
-      indentString: "",
-      suffixTexts: [:]
+    annotatedSource(inSyntaxTree: tree, withDiagnostics: diags)
+  }
+
+  public static func annotatedSource(
```
ahoppen, line 772: Same here, let’s deprecate it in favor of `contextRange`.

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
```diff
@@ -12,311 +12,811 @@
...
+    tree: some SyntaxProtocol,
+    diags: [Diagnostic],
+    contextSize: Int = 2,
+    colorize: Bool = false
+  ) -> String {
+    Self.annotatedSource(
+      inSyntaxTree: tree,
+      withDiagnostics: diags,
+      limitedToContextRange: .limited(contextSize),
+      withDiagnosticDecorator: colorize ? .ANSI : .basic
```
ahoppen, line 782: We also need a compatibility layer for https://github.com/apple/swift-syntax/blob/bff0b282360083f9aebbb2a1c5ed05fa666e3757/Sources/SwiftDiagnostics/DiagnosticsFormatter.swift#L311-L321

ahoppen (commented):
Very nice with all the comments explaining what the code is doing. I mostly have nitpicky comments.

ahoppen:
CC @DougGregor I’d like to hear your thoughts on the note printing as well.

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
Matejkob: My thinking behind offering the full context range was to provide end-developers with extensive customization options. For example, [the pointfree project uses `contextSize` set to max number of lines to achieve a full context view](https://github.com/pointfreeco/swift-macro-testing/blob/15916c0c328339f54c15d616465d79700e3f7de8/Sources/MacroTesting/AssertMacro.swift#L205).

ahoppen: Ah, I see. I still have concerns about how this interacts with the notes but maybe it’s good enough even if the connection between notes and errors/warnings is not obvious.

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
Matejkob: I've moved documentation to each enum case + I've ensured that every line of documentation is shorter than 120 columns.

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
Matejkob: Great point! I've changed it to a checked version. Thanks for pointing that out!

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
Matejkob: That's a good idea. Thanks. I've changed it.

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
Matejkob: I've replaced all occurrence of `append` with `+=`. Thanks.

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
Matejkob: Yup, it's cleaner. Thanks. Done.

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
Matejkob: Removed.

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
Matejkob: Haha, I don't know why I've used an array in the first place. Thanks for pointing out. Done.

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
Matejkob: Oh yeah, it will do the job. Thx.

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
Matejkob: Sure, why not; done.

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
Matejkob: New lines should be aligned to `Note`? 
Like so:
Screenshot 2024-02-21 at 2 42 20 PM

And for `Returns:` like so:
Screenshot 2024-02-21 at 2 44 17 PM

?

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
Matejkob: I understand the appeal of using `??` for clarity. However, in this case, I find `== true/false` clearer due to the double negation in `if !(annotatedSource.last?.isNewline ?? false) {`, which seems more complex. It feels simpler to directly compare the optional, avoiding the extra mental gymnastics.

Matejkob: I'm quite convinced that using `== true/false` with optionals might seem a bit uncommon at first, but once you get accustomed to it, it becomes much simpler.

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
ahoppen: Yes, that’s how I align them because that’s how I would align any markdown list.

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
ahoppen: I don’t think you deprecated this one.

Also, I just remembered that we should add release notes for the deprecations.

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
ahoppen: Oh, sorry, I missed that.

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
ahoppen: If you think that your version is more readable, that’s a tie. Let’s leave your version then.

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
Matejkob: So if we decide to discard the idea with `full` context we would need to deprecate this property. In this case, contextSize can still be an Int. So I guess we waiting for Doug's respond here: https://github.com/apple/swift-syntax/pull/2214#discussion_r1495229284

ahoppen:
Almost looks good to me. I’d like to 
- [ ] Deprecate the `contextSize` methods mentioned in my last review
- [ ] Add release notes for the deprecations
- [ ] Get @DougGregor’s opinion on the change, in particular the way notes are printed and `ContextRange.full`

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
DougGregor: I'm not opposed to having the full context range. From a user perspective, I do wonder if I want to have a different context range for the top-level diagnostic vs. diagnostics in nested buffers (e.g., for macro expansions), because the nested buffers generally have content that's been synthesized.

Tests/SwiftDiagnosticsTest/NewDiagnosticsFormatterTests.swift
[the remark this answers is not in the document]
DougGregor: Yes, this looks good to me, too. Right now, the Swift compiler is intentionally grouping things so that there is only a single top-level diagnostic, because that made sure all of the information about a single issue (with all of its notes) would occur in the one printout.

Notes in different files are still a little tricky---they look like top-level things right now, but @ahoppen has reasonably suggested that they should be in a "box" as if they were nested, because notes are never standalone---they're always associated with some other error/warning/remark.

Tests/SwiftDiagnosticsTest/NewDiagnosticsFormatterTests.swift
DougGregor: I don't think we should try to render Fix-Its into the diagnostic output. We tried with the diagnostic formatter that we've been using in the compiler for years now, and it's hard to provide that much information via a console interface without obscuring the rest of the error. Fix-Its are best handled by editors.

ahoppen:
I chatted with @DougGregor about this the other day and he doesn’t have objections. If you could address the remaining open points from my previous comment, I would like to get this in.

Matejkob:
> I chatted with @DougGregor about this the other day and he doesn’t have objections. If you could address the remaining open points from my previous comment, I would like to get this in.

I'll try to carve out some time this weekend to tackle that. Thanks for your patience.

I do have one question, though:

Are we ready to finalize the design of `DiagnosticDecorator` and make the following initializer public?

```swift
  /// Initializes a new `DiagnosticsFormatter` instance.
  ///
  /// - Parameters:
  ///   - contextRange: Specifies the number of contextual lines around each diagnostic message. Default is `.limited(2)`.
  ///   - diagnosticDecorator: The decorator used for formatting diagnostic output. Default is `BasicDiagnosticDecorator`.
  init(
    contextRange: ContextRange = .limited(2),
    diagnosticDecorator: DiagnosticDecorator = BasicDiagnosticDecorator()
  ) {
    self.contextRange = contextRange
    self.diagnosticDecorator = diagnosticDecorator
  }
```

DPrakashhh:
@ahoppen Can I work on this problem?

ahoppen:
@DPrakashhh You already have quite a few open PRs. I suggest that you focus on those first, before starting any new tasks.

filip-sakel:
@ahoppen  Is anyone else working on this? I see #2238 was merged, but #2557 and this this PR both seem stale. Briefly looking at the GroupedDiagnostics code, it seems like we're still not presenting the `notes` field of a diagnostic (as [this issue](https://github.com/swiftlang/swift-syntax/issues/2166) reports). I could take a crack at this, especially since a lot more diagnostic infrastructure is now in place.

ahoppen:
@Matejkob Are you planning to pick this up again? If not, that’s totally fine, just don’t want to make sure we don’t have two contributors working on the same issue. 

@filip-sakel, thanks for your interest in fixing this. Let’s see what @Matejkob has to say. Feel free to pick this up but if @Matejkob responds in, say, 1 week that he wants to continue working on this, this PR has precedence because we’ve already gone through a few iterations here.

filip-sakel:
@ahoppen I opened a small PR that contains targeted changes for including notes in the output of `GroupedDiagnostics`. Feel free to review if you get chance to.

Of course, @Matejkob if you want to finish this up since you originally started working on it, feel free to. If you find it useful you can also just fork my [branch](https://github.com/filip-sakel/swift-syntax/tree/notes-diagnostic-group).

PR #2215, opened 2023-09-18
Fix trivia handling in `AttributeRemover` of `MacroSystem`

### LINKED ISSUE #2203: Unexpected trivia when removing attributes in `MacroSystem`
### Description

On `509.0.0` and `main` ([`8966ec7`](https://github.com/apple/swift-syntax/commit/8966ec7157f9774dd3b03b53520e75bc570a284a)), I'm running into the following unexpected trivia handling in a peer macro:

```swift
fileprivate struct NoOpPeerMacro: PeerMacro {
  static func expansion(
    of node: AttributeSyntax,
    providingPeersOf declaration: some DeclSyntaxProtocol,
  ) throws -> [DeclSyntax] {
    return []
  }
}

final class PeerMacroTests: XCTestCase {
  // …
  func testEmptyBeforeAttributeOnSameLineAsMemberVariable() {
    assertMacroExpansion(
      """
      struct Foo {
        @Test @State var x: Int
      }
      """,
      expandedSource: """
        struct Foo {
          @State var x: Int
        }
        """,
      macros: ["Test": NoOpPeerMacro.self]
    )
  }
  // …
}
```

Output:
```diff
–struct Foo {
–  @State var x: Int
+struct Foo {@State 
+  var x: Int
 }
```

_Quick reference_: [`AttributeRemover`]( https://github.com/apple/swift-syntax/blob/8966ec7157f9774dd3b03b53520e75bc570a284a/Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift#L387C15-L387C31).

To work around, I can tailor the whitespace in my test in a specific manner as follows:

```swift
struct Foo {
  @Test
  @State 
  var x: Int
}
```

What is the expected trivia treatment in this case? Should trivia in test cases be deliberately crafted, or should the `MacroSystem` strive to normalize all syntactically valid trivia in some predictable way?

_Aside: Here are a couple other tests and outputs that I found surprising:_

<details>
  <summary>Click to expand!</summary>

```swift
  func testAttributeOnOwnLineThenEmptyBeforeVariable() {
    assertMacroExpansion(
      """
      @State
      @Test var x: Int
      """,
      expandedSource: """
        @State
        var x: Int
        """,
      macros: [
        "Test": NoOpPeerMacro.self
      ]
    )
  }
```

Output:
```diff
–@State
–var x: Int
+@Statevar x: Int
```
---

```swift
  func testEmptyAfterAttributeOnSameLineAsMemberVariable() {
    assertMacroExpansion(
      """
      struct Foo {
        @State @Test var x: Int
      }
      """,
      expandedSource: """
        struct Foo {
          @State var x: Int
        }
        """,
      macros: ["Test": NoOpPeerMacro.self]
    )
  }
```

Output:
```diff
 struct Foo {
–  @State var x: Int
+  @State 
+  var x: Int
 }
```

---

```swift
  func testEmptyOnOwnLineThenAttributedMemberVariable() {
    assertMacroExpansion(
      """
      struct Foo {
        @Test
        @State var x: Int
      }
      """,
      expandedSource: """
        struct Foo {
          @State var x: Int
        }
        """,
      macros: ["Test": NoOpPeerMacro.self]
    )
  }
```

Output:
```diff
 struct Foo {
–  @State var x: Int
+  @State 
+  var x: Int
 }
```
</details>

### Steps to Reproduce

Add the tests included in the description to "Tests/SwiftMacroExpansionTest/PeerMacroTests.swift" and run them.

gohanlon:
Initially opened for discussion of #2203, and now a fix for it.
    
Also, adds `AttributeRemoverTests` covering a variety of attribute removal cases involving surrounding trivia.

Fixes #2203
rdar://115585919

### DIFF
--- a/Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
+++ b/Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
@@ -397,36 +397,97 @@ private class AttributeRemover: SyntaxRewriter {
     var filteredAttributes: [AttributeListSyntax.Element] = []
     for case .attribute(let attribute) in node {
       if attributesToRemove.contains(attribute) {
-        var leadingTrivia = node.leadingTrivia
-        if let lastNewline = leadingTrivia.pieces.lastIndex(where: { $0.isNewline }),
+        var leadingTrivia = attribute.leadingTrivia
+
+        // Don't leave behind an empty line when the attribute being removed is on its own line,
+        // based on the following conditions:
+        //  - Leading trivia ends with a newline followed by arbitrary number of spaces or tabs
+        //  - All leading trivia pieces after the last newline are just whitespace, ensuring
+        //    there are no comments or other non-whitespace characters on the same line
+        //    preceding the attribute.
+        //  - There is no trailing trivia and the next token has leading trivia.
+        if let lastNewline = leadingTrivia.pieces.lastIndex(where: \.isNewline),
           leadingTrivia.pieces[lastNewline...].allSatisfy(\.isWhitespace),
-          node.trailingTrivia.isEmpty,
-          node.nextToken(viewMode: .sourceAccurate)?.leadingTrivia.first?.isNewline ?? false
+          attribute.trailingTrivia.isEmpty,
+          let nextToken = attribute.nextToken(viewMode: .sourceAccurate),
+          !nextToken.leadingTrivia.isEmpty
         {
-          // If the attribute is on its own line based on the following conditions,
-          // remove the newline from it so we don’t end up with an empty line
-          //  - Trailing trivia ends with a newline followed by arbitrary number of spaces or tabs
-          //  - There is no trailing trivia and the next token starts on a new line
           leadingTrivia = Trivia(pieces: leadingTrivia.pieces[..<lastNewline])
         }
+
         // Drop any spaces or tabs from the trailing trivia because there’s no
         // more attribute they need to separate.
-        let trailingTrivia = Trivia(pieces: attribute.trailingTrivia.drop(while: { $0.isSpaceOrTab }))
+        let trailingTrivia = attribute.trailingTrivia.trimmingPrefix(while: \.isSpaceOrTab)
         triviaToAttachToNextToken += leadingTrivia + trailingTrivia
+
+        // If the attribute is not separated from the previous attribute by trivia, as in
+        // `@First@Second var x: Int` (yes, that's valid Swift), removing the `@Second`
+        // attribute and dropping all its trivia would cause `@First` and `var` to join
+        // without any trivia in between, which is invalid. In such cases, the trailing trivia
+        // of the attribute is significant and must be retained.
+        if triviaToAttachToNextToken.isEmpty,
+          let previousToken = attribute.previousToken(viewMode: .sourceAccurate),
+          previousToken.trailingTrivia.isEmpty
+        {
+          triviaToAttachToNextToken = attribute.trailingTrivia
+        }
       } else {
-        filteredAttributes.append(.attribute(attribute))
+        filteredAttributes.append(.attribute(prependAndClearAccumulatedTrivia(to: attribute)))
+      }
+    }
+
+    // Ensure that any horizontal whitespace trailing the attributes list is trimmed if the next
+    // token starts a new line.
+    if let nextToken = node.nextToken(viewMode: .sourceAccurate),
+      nextToken.leadingTrivia.startsWithNewline
+    {
+      if !triviaToAttachToNextToken.isEmpty {
+        triviaToAttachToNextToken = triviaToAttachToNextToken.trimmingSuffix(while: \.isSpaceOrTab)
+      } else if let lastAttribute = filteredAttributes.last {
+        filteredAttributes[filteredAttributes.count - 1].trailingTrivia = lastAttribute.trailingTrivia.trimmingSuffix(while: \.isSpaceOrTab)
       }
     }
     return AttributeListSyntax(filteredAttributes)
   }
 
   override func visit(_ token: TokenSyntax) -> TokenSyntax {
-    if !triviaToAttachToNextToken.isEmpty {
-      defer { triviaToAttachToNextToken = Trivia() }
-      return token.with(\.leadingTrivia, triviaToAttachToNextToken + token.leadingTrivia)
-    } else {
-      return token
-    }
+    return prependAndClearAccumulatedTrivia(to: token)
+  }
+
+  /// Prepends the accumulated trivia to the given node's leading trivia.
+  ///
+  /// To preserve correct formatting after attribute removal, this function reassigns
+  /// significant trivia accumulated from removed attributes to the provided subsequent node.
+  /// Once attached, the accumulated trivia is cleared.
+  ///
+  /// - Parameter node: The syntax node receiving the accumulated trivia.
+  /// - Returns: The modified syntax node with the prepended trivia.
+  private func prependAndClearAccumulatedTrivia<T: SyntaxProtocol>(to syntaxNode: T) -> T {
+    defer { triviaToAttachToNextToken = Trivia() }
+    return syntaxNode.with(\.leadingTrivia, triviaToAttachToNextToken + syntaxNode.leadingTrivia)
+  }
+}
+
+private extension Trivia {
+  func trimmingPrefix(
+    while predicate: (TriviaPiece) -> Bool
+  ) -> Trivia {
+    Trivia(pieces: self.drop(while: predicate))
+  }
+
+  func trimmingSuffix(
+    while predicate: (TriviaPiece) -> Bool
+  ) -> Trivia {
+    Trivia(
+      pieces: self[...]
+        .reversed()
+        .drop(while: predicate)
+        .reversed()
+    )
+  }
+
+  var startsWithNewline: Bool {
+    self.first?.isNewline ?? false
   }
 }
 
--- a/Tests/SwiftSyntaxMacroExpansionTest/AttributeRemoverTests.swift
+++ b/Tests/SwiftSyntaxMacroExpansionTest/AttributeRemoverTests.swift
@@ -1,0 +1,514 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+//==========================================================================//
+// IMPORTANT: The macros defined in this file are intended to test the      //
+// behavior of MacroSystem. Many of them do not serve as good examples of   //
+// how macros should be written. In particular, they often lack error       //
+// handling because it is not needed in the few test cases in which these   //
+// macros are invoked.                                                      //
+//==========================================================================//
+
+import SwiftDiagnostics
+import SwiftSyntax
+import SwiftSyntaxMacroExpansion
+import SwiftSyntaxMacros
+import SwiftSyntaxMacrosTestSupport
+import XCTest
+
+fileprivate struct NoOpPeerMacro: PeerMacro {
+  static func expansion(
+    of node: AttributeSyntax,
+    providingPeersOf declaration: some DeclSyntaxProtocol,
+    in context: some MacroExpansionContext
+  ) throws -> [DeclSyntax] {
+    return []
+  }
+}
+
+final class AttributeRemoverTests: XCTestCase {
+  func testEmptyOnSameLineAsVariable() {
+    assertMacroExpansion(
+      "@Test var x: Int",
+      expandedSource: "var x: Int",
+      macros: [
+        "Test": NoOpPeerMacro.self
+      ]
+    )
+  }
+
+  func testEmptyTwiceOnSameLineAsVariable() {
+    assertMacroExpansion(
+      "@Test @Test var x: Int",
+      expandedSource: "var x: Int",
+      macros: [
+        "Test": NoOpPeerMacro.self
+      ]
+    )
+  }
+
+  func testEmptyOnOwnLineBeforeVariable() {
+    assertMacroExpansion(
+      """
+      @Test
+      var x: Int
+      """,
+      expandedSource: "var x: Int",
+      macros: [
+        "Test": NoOpPeerMacro.self
+      ]
+    )
+  }
+
+  func testEmptyTwiceOnOwnLineBeforeVariable() {
+    assertMacroExpansion(
+      """
+      @Test @Test
+      var x: Int
+      """,
+      expandedSource: "var x: Int",
+      macros: [
+        "Test": NoOpPeerMacro.self
+      ]
+    )
+  }
+
+  func testEmptyAndAttributeOnOwnLineBeforeVariable() {
+    assertMacroExpansion(
+      """
+      @Test @State
+      var x: Int
+      """,
+      expandedSource: """
+        @State
+        var x: Int
+        """,
+      macros: [
+        "Test": NoOpPeerMacro.self
+      ]
+    )
+  }
+
+  func testAttributeAndEmptyOnOwnLineBeforeVariable() {
+    assertMacroExpansion(
+      """
+      @State @Test
+      var x: Int
+      """,
+      expandedSource: """
+        @State
+        var x: Int
+        """,
+      macros: [
+        "Test": NoOpPeerMacro.self
+      ]
+    )
+  }
+
+  func testAttributeAndEmptyAndCommentOnOwnLineBeforeVariable() {
+    assertMacroExpansion(
+      """
+      @State @Test// comment
+      var x: Int
+      """,
+      expandedSource: """
+        @State // comment
+        var x: Int
+        """,
+      macros: [
+        "Test": NoOpPeerMacro.self
+      ]
+    )
+  }
+
+  func testAttributeAndEmptyAndCommentOnOwnLineBeforeVariable2() {
+    assertMacroExpansion(
+      """
+      @State @Test // comment
+      var x: Int
+      """,
+      expandedSource: """
+        @State // comment
+        var x: Int
+        """,
+      macros: [
+        "Test": NoOpPeerMacro.self
+      ]
+    )
+  }
+
+  func testCommentsAroundEmpty() {
+    assertMacroExpansion(
+      """
+      struct S {
+        /// Some doc comment
+        @Test /* trailing */
+        var value: Int
+      }
+      """,
+      expandedSource: """
+        struct S {
+          /// Some doc comment
+          /* trailing */
+          var value: Int
+        }
+        """,
+      macros: ["Test": NoOpPeerMacro.self]
+    )
+  }
+
+  func testBlockCommentNewlineEmpty() {
+    assertMacroExpansion(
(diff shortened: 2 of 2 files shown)


gohanlon:
@ahoppen I think I see a simple fix that will get all these tests passing. But first, do these test cases and expected outputs look good to you?

gohanlon:
The tests I added exercise `MacroSystem.AttributeRemover`, but I've added the tests to `PeerMacroTests`. Should I move the tests to a new test file, e.g. `AttributeRemoverTests`?

I did a cursory search and turned up these tests that seem to test `MacroSystem.AttributeRemover`:

- `AccessorMacroTests.testEmpty`
- `MemberAttributeMacroTests.testEmpty`
- `MemberMacroTests`
  - `testCommentAroundeAttachedMacro` *[sic]*
  - `testEmpty`
- `PeerMacroTests` (includes this PR)
  - `testEmptyOnSameLineAsVariable`
  - `testEmptyTwiceOnSameLineAsVariable`
  - `testEmptyOnOwnLineBeforeVariable`
  - `testEmptyTwiceOnOwnLineBeforeVariable`
  - `testEmptyAndAttributeOnOwnLineBeforeVariable`
  - `testAttributeAndEmptyOnOwnLineBeforeVariable`
  - `testAttributeAndEmptyAndCommentOnOwnLineBeforeVariable`
  - `testAttributeAndEmptyAndCommentOnOwnLineBeforeVariable2`
  - `testCommentAroundAttachedMacro`
  - `testAttributeAndEmptyOnOwnLinesBeforeVariable`
  - `testEmptyAndAttributeOnOwnLinesBeforeVariable`
  - `testAttributeOnOwnLineThenEmptyBeforeVariable`
  - `testEmptyOnOwnLineThenEmptyBeforeVariable`
  - `testEmptyOnMemberVariable`
  - `testEmptyBeforeAttributeOnSameLineAsMemberVariable`
  - `testEmptyAfterAttributeOnSameLineAsMemberVariable`
  - `testEmptyAfterAttributeOnSameLineAsMemberVariable_AwkwardWhitespace`
  - `testEmptyOnOwnLineThenAttributedMemberVariable`
  - `testAttributeOnOwnLineThenEmptyOnMemberVariable`

gohanlon:
Getting a little ahead, I sketched out the fix I have in mind. With the following change, the test suite passes:

```diff
diff --git a/Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift b/Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
index f97f02b5..1198c292 100644
--- a/Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
+++ b/Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
@@ -397,11 +397,12 @@ private class AttributeRemover: SyntaxRewriter {
     var filteredAttributes: [AttributeListSyntax.Element] = []
     for case .attribute(let attribute) in node {
       if attributesToRemove.contains(attribute) {
-        var leadingTrivia = node.leadingTrivia
+        var leadingTrivia = attribute.leadingTrivia
         if let lastNewline = leadingTrivia.pieces.lastIndex(where: { $0.isNewline }),
           leadingTrivia.pieces[lastNewline...].allSatisfy(\.isWhitespace),
-          node.trailingTrivia.isEmpty,
-          node.nextToken(viewMode: .sourceAccurate)?.leadingTrivia.first?.isNewline ?? false
+          attribute.trailingTrivia.isEmpty//,
+        // Both before this PR and in this PR, removing  `node.nextToken`/`attribute.nextToken` condition has no effect on the test suite
+//          attribute.nextToken(viewMode: .sourceAccurate)?.leadingTrivia.first?.isNewline ?? false
         {
           // If the attribute is on its own line based on the following conditions,
           // remove the newline from it so we don’t end up with an empty line
@@ -414,22 +415,25 @@ private class AttributeRemover: SyntaxRewriter {
         let trailingTrivia = Trivia(pieces: attribute.trailingTrivia.drop(while: { $0.isSpaceOrTab }))
         triviaToAttachToNextToken += leadingTrivia + trailingTrivia
       } else {
-        filteredAttributes.append(.attribute(attribute))
+        filteredAttributes.append(.attribute(prependAccumulatedTrivia(attribute)))
       }
     }
     return AttributeListSyntax(filteredAttributes)
   }
 
   override func visit(_ token: TokenSyntax) -> TokenSyntax {
-    if !triviaToAttachToNextToken.isEmpty {
-      defer { triviaToAttachToNextToken = Trivia() }
-      return token.with(\.leadingTrivia, triviaToAttachToNextToken + token.leadingTrivia)
-    } else {
-      return token
-    }
+    return prependAccumulatedTrivia(token)
+  }
+
+  private func prependAccumulatedTrivia<T: SyntaxProtocol>(_ node: T) -> T {
+    guard !triviaToAttachToNextToken.isEmpty else { return node }
+
+    defer { triviaToAttachToNextToken = Trivia() }
+    return node.with(\.leadingTrivia, triviaToAttachToNextToken + node.leadingTrivia)
   }
 }
 
+
 let diagnosticDomain: String = "SwiftSyntaxMacroExpansion"
 
 private enum MacroApplicationError: DiagnosticMessage, Error {
```

ahoppen:
The tests look good to me and the source changes also make sense on first sight. Can you push them and then we can work from there.

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
```diff
@@ -397,36 +397,38 @@ private class AttributeRemover: SyntaxRewriter {
     var filteredAttributes: [AttributeListSyntax.Element] = []
     for case .attribute(let attribute) in node {
       if attributesToRemove.contains(attribute) {
-        var leadingTrivia = node.leadingTrivia
+        var leadingTrivia = attribute.leadingTrivia
         if let lastNewline = leadingTrivia.pieces.lastIndex(where: { $0.isNewline }),
           leadingTrivia.pieces[lastNewline...].allSatisfy(\.isWhitespace),
-          node.trailingTrivia.isEmpty,
-          node.nextToken(viewMode: .sourceAccurate)?.leadingTrivia.first?.isNewline ?? false
```
ahoppen, line 404: If we remove this check, then the following test case fails (yes, you don’t need to put spaces between attributes 🤯)

```swift
assertMacroExpansion(
  """
  struct Foo {
    @Test@State var x: Int
  }
  """,
  expandedSource: """
    struct Foo {
      @State var x: Int
    }
    """,
  macros: ["Test": NoOpPeerMacro.self]
)
```

And while writing this, I also realized that we don’t handle `@State@Test` correctly either. Here’s a test case for that

```swift
assertMacroExpansion(
  """
  struct Foo {
    @State@Test
    var x: Int
  }
  """,
  expandedSource: """
    struct Foo {
      @State
      var x: Int
    }
    """,
  macros: ["Test": NoOpPeerMacro.self]
)
```

And a fix suggestion to add this after `triviaToAttachToNextToken += leadingTrivia + trailingTrivia`

```swift
/// If the attribute is not separated by the previous attribute using trivia, eg.
/// `@First@Second var x: Int` (yes, taht's valid Swift), and removing
/// the `@Second` attribute then dropping all trivia of this attribute
/// would join `First` and `var` without any trivia in between, which is
/// invalid. In this case the trailing trivia of the attribute is
/// significant and we need to keep it.
if triviaToAttachToNextToken.isEmpty &&
    attribute.previousToken(viewMode: .sourceAccurate)?.trailingTrivia.isEmpty ?? false {
  triviaToAttachToNextToken = attribute.trailingTrivia
}
```

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
```diff
@@ -397,36 +397,38 @@ private class AttributeRemover: SyntaxRewriter {
...
-    if !triviaToAttachToNextToken.isEmpty {
-      defer { triviaToAttachToNextToken = Trivia() }
-      return token.with(\.leadingTrivia, triviaToAttachToNextToken + token.leadingTrivia)
-    } else {
-      return token
-    }
+    return prependAccumulatedTrivia(token)
+  }
+
+  private func prependAccumulatedTrivia<T: SyntaxProtocol>(_ node: T) -> T {
```
ahoppen, line 427: Could you add a short doc comment for this function?

ahoppen (commented):
Thanks. Your removal of the newline check made me realize that attributes don’t need to be separated by spaces to be valid and that we have a bug surrounding that as well. I added a suggestion of how to fix that one as well, since you’re at it.

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
gohanlon: Wow, thanks! I'll add and cover both those test cases.

Tests/SwiftSyntaxMacroExpansionTest/PeerMacroTests.swift
```diff
@@ -174,23 +184,318 @@ final class PeerMacroTests: XCTestCase {
...
+        @State
+        var x: Int
+        """,
+      macros: [
+        "Test": NoOpPeerMacro.self
+      ]
+    )
+  }
+
+  // todo: This test passes, but the behavior of leaving "dangling" whitespace at the EOL is not great
```
ahoppen, line 249: I agree that it’s not amazing but I think it’s good enough for now.

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
gohanlon: Done.

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
```diff
@@ -397,36 +397,38 @@ private class AttributeRemover: SyntaxRewriter {
...
+        var leadingTrivia = attribute.leadingTrivia
         if let lastNewline = leadingTrivia.pieces.lastIndex(where: { $0.isNewline }),
           leadingTrivia.pieces[lastNewline...].allSatisfy(\.isWhitespace),
-          node.trailingTrivia.isEmpty,
-          node.nextToken(viewMode: .sourceAccurate)?.leadingTrivia.first?.isNewline ?? false
+          attribute.trailingTrivia.isEmpty
         {
           // If the attribute is on its own line based on the following conditions,
           // remove the newline from it so we don’t end up with an empty line
           //  - Trailing trivia ends with a newline followed by arbitrary number of spaces or tabs
```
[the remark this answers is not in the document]
gohanlon: Done.

Tests/SwiftSyntaxMacroExpansionTest/PeerMacroTests.swift
gohanlon: This PR now ensures that any horizontal whitespace trailing the attributes list is trimmed if the next token starts a new line.

gohanlon:
@ahoppen Pending your feedback, I think this PR is in a pretty good spot.

There are a few improvements I'd like to make/discuss:
- The test coverage is encouraging. You can comment out specific blocks or conditions and get a failing test or tests that reveal the purpose of that code. For me, this pretty much negates the need for the multiline comments—they could be replaced with single-line summarizing comments, with any remaining necessary details moved to the outer doc comment. (And removing the multiline comments would mean we wouldn't have to worry about keeping them updated).
- Change `if attributesToRemove.contains(attribute) { … } else { … }` to `if !attributesToRemove.contains(attribute) { … continue }` to remove the distant `else` and decrease nesting.
- Now, almost all the tests in `PeerMacroTests` actually test `MacroSystem.AttributeRemover`. Should we create a new test file, e.g. `AttributeRemoverTests`?
- `MacroSystem.AttributeRemover` now has a *lot* of conditional logic. It would be worthwhile to attempt to refactor it to be more succint/readable and ensure there isn't any unnecessary logic.

Tests/SwiftSyntaxMacroExpansionTest/PeerMacroTests.swift
Matejkob, line 1: Fantastic job on the unit tests you've added! 🎉 It's really refreshing to see such comprehensive testing, especially in a codebase as complex as ours. 

That said, I noticed there's a hefty chunk of tests focused specifically on attribute removal. Have you considered breaking these out into a separate test file? It might make it easier to navigate and maintain down the line.

On another note, free thought here: what do you think about running those attribute removal test cases directly against `AttributeRemover` instead of going through the `PeerMacro`? It could make the tests a bit more targeted, and might even expose some edge cases we haven't thought of.

Looking forward to your thoughts! Keep up the awesome work. 👏👏

Matejkob: Ah, I see you've already been thinking along the same lines:

> * Most tests in `PeerMacroTests` are actually focused on `MacroSystem.AttributeRemover`. Should we move these to a new test file, perhaps called `AttributeRemoverTests`?
> * `MacroSystem.AttributeRemover` has become quite complex with lots of conditional logic. It might be a good idea to refactor it for better readability and to eliminate any unnecessary logic.

I'm in full agreement with moving these tests. What is more, I don't see a compelling reason to keep `MacroSystem.AttributeRemover` as `private`. Making it visible for testing could allow us to cover it more directly, don't you think?

As for the refactoring, it's definitely a good idea but probably in follow-up PR?

bnbarham: I'm personally fine with either direction here - targeted testing or testing just on `MacroSystem` itself. But if we go for the former, `AttributeRemover` should be SPI as it shouldn't be part of the public API. We should also still keep one or two tests on `MacroSystem` to ensure it's using `AttributeRemover` correctly.

ahoppen:
> The test coverage is encouraging. You can comment out specific blocks or conditions and get a failing test or tests that reveal the purpose of that code. For me, this pretty much negates the need for the multiline comments—they could be replaced with single-line summarizing comments, with any remaining necessary details moved to the outer doc comment. (And removing the multiline comments would mean we wouldn't have to worry about keeping them updated).

I disagree here. For example when reviewing code on GitHub, I don’t have the option of uncommenting the line and re-running the tests. Also, while reading the code, IMO it’s good practice to capture the intended purpose of these lines since it explains the *Why* in addition to the *What*. The purpose of the test cases is, IMO, to verify that the code works on specific examples, which is yet another level of specifying what the code does and why it does it. 


> * Change `if attributesToRemove.contains(attribute) { … } else { … }` to `if !attributesToRemove.contains(attribute) { … continue }` to remove the distant `else` and decrease nesting. I've held off on making this change because it'd make the diffs much harder to read.

Sounds good to me. Do you want to create a follow-up PR once this one is merged?

> * Now, almost all the tests in `PeerMacroTests` actually test `MacroSystem.AttributeRemover`. Should we create a new test file, e.g. `AttributeRemoverTests`?

Splitting seems like a good idea to me. Like @bnbarham I don’t have a strong opinion of whether we should test `AttributeRemover` directly or do it as part of `MacrosSystem` testing.

> * `MacroSystem.AttributeRemover` now has a _lot_ of conditional logic. It would be worthwhile to attempt to refactor it to be more succint/readable and ensure there isn't any unnecessary logic.

I’m always in favor of refactorings that make code easier to read. I personally don’t think there’s to be much/any unnecessary code that can be deleted in `AttributeRemove` right now and I don’t have any ideas off the top of my head of how to improve it. But if you have ideas, I’m happy to see them.

Tests/SwiftSyntaxMacroExpansionTest/PeerMacroTests.swift
ahoppen: 🎉 Nice

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
```diff
@@ -397,36 +397,100 @@ private class AttributeRemover: SyntaxRewriter {
...
     for case .attribute(let attribute) in node {
       if attributesToRemove.contains(attribute) {
-        var leadingTrivia = node.leadingTrivia
-        if let lastNewline = leadingTrivia.pieces.lastIndex(where: { $0.isNewline }),
-          leadingTrivia.pieces[lastNewline...].allSatisfy(\.isWhitespace),
-          node.trailingTrivia.isEmpty,
-          node.nextToken(viewMode: .sourceAccurate)?.leadingTrivia.first?.isNewline ?? false
+        var leadingTrivia = attribute.leadingTrivia
+        if let lastNewline = leadingTrivia.pieces.lastIndex(where: \.isNewline),
+          attribute.trailingTrivia.isEmpty,
```
ahoppen, line 402: Since you removed the `leadingTrivia.pieces[lastNewline...].allSatisfy(\.isWhitespace)` check, how do we handle the following now? I think we’ll remove the block comment but we shouldn’t.

```swift
@State
/*doc comment*/@Test
var x: Int
```

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
```diff
@@ -397,36 +397,100 @@ private class AttributeRemover: SyntaxRewriter {
...
+    Trivia(
+      pieces: self[...]
+        .reversed()
+        .drop(while: predicate)
+        .reversed()
+    )
+  }
+
+  func startsWithNewline() -> Bool {
+    self.first?.isNewline ?? false
```
ahoppen, line 493: This could be a computed property

```suggestion
  var startsWithNewline: Bool {
    self.first?.isNewline ?? false
```

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
```diff
@@ -397,36 +397,100 @@ private class AttributeRemover: SyntaxRewriter {
...
+        if let last = filteredAttributes.last {
+          filteredAttributes =
+            filteredAttributes.dropLast()
+            + [
+              last.with(
+                \.trailingTrivia,
+                last.trailingTrivia.trimmingSuffix(while: \.isSpaceOrTab)
+              )
+            ]
+        }
```
ahoppen, line 450: Not sure if you agree but I think this would read nicer as

```swift
} else if let lastAttribute = filteredAttributes.last {
  filteredAttributes[filteredAttributes.count - 1].trailingTrivia = lastAttribute.trailingTrivia.trimmingSuffix(while: \.isSpaceOrTab)
}
```

ahoppen (commented):
Who would have guessed that removing an attribute could be this difficult? Thanks for your work on fixing it. I have one more functional comment. Otherwise it looks good to me.

gohanlon:
@ahoppen I believe I've addressed your feedback. Care to take a (perhaps!) final look?

Now that we're close, would you like me to squash this PR into a single commit, and annotate it with references to the GitHub and rdar issues?

ahoppen (approved):
Yes, this looks great. If you could squash the commits, that would be great and I’ll trigger CI.

gohanlon:
> Yes, this looks great. If you could squash the commits, that would be great and I’ll trigger CI.

@ahoppen Squashed. Let me know if there's anything else!

ahoppen:
@swift-ci Please test

ahoppen:
Let’s get this in 🚢

ahoppen:
@swift-ci Please test Windowss

gohanlon:
@ahoppen Is there something in this PR that needs to be done to satisfy the Windows CI, or perhaps Windows CI is/was generally broken?

bnbarham:
Nothing on your side. I'll run again, though could be there's a failure that needs fixing.

bnbarham:

gohanlon:
@ahoppen Quick question: Do you know the release plan for this PR's changes?

ahoppen:
This was merged into `main` after we cut the `release/5.10` branch in early September, so this will be part of swift-syntax 600, aligned with Swift 6.

PR #2200, opened 2023-09-14
Clean ups to the swift-syntax-dev-utils and Package.swift

ahoppen:
- Move the environment variables that we check for in Package.swift to the top and add documentation to them
- Perform some clean-ups in Package.swift
  - Don’t set `SWIFTSYNTAX_NO_OSLOG_DEPENDENCY` because it’s not needed
  - Use the Swift settings of a target also for the corresponding test target. That’s what we already did for SwiftParser and SwiftParserTests. We should be consistent here.
- Don’t set `--verbose` twice when invoking SwiftPM from `swift-syntax-dev-utils`
  - `invokeSwiftPM` already sets `--verbose`. No need to pass it again in the `additionalArguments`.
- Remove `--disable-sandbox`. This flag wasn’t used anywhere.

PR #2217, opened 2023-09-19
Fix incorrect `if` condition in the main package manifest 

Matejkob:
This PR corrects the conditional logic for loading local versus remote dependencies in the Package.swift file.

I believe in PR #2200 we have merged the incorrect `if` condition in the `Package.swift` file.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -332,12 +332,12 @@ package.targets.append(
 )
 
 if useLocalDependencies {
+  package.dependencies += [
+    .package(path: "../swift-argument-parser")
+  ]
+} else {
   // Building standalone.
   package.dependencies += [
     .package(url: "https://github.com/apple/swift-argument-parser.git", from: "1.2.2")
-  ]
-} else {
-  package.dependencies += [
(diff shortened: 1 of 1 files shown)


kimdv (approved):

kimdv:
@swift-ci please test

ahoppen (approved):
I wonder how that change made it through CI. 🤔

Matejkob:
Isn't this flag turned on on the CI?

ahoppen:
It should be, that’s why I’m wondering…

compnerd:
This seems to have caused some issues on Windows - `SWIFTCI_USE_LOCAL_DEPS` is not defined, and we no longer can build the package: https://dev.azure.com/compnerd/swift-build/_build/results?buildId=61881&view=logs&j=186393e6-7c47-52d0-cbab-0f64f605ebf2&t=55352f38-40c8-5fcb-13f5-1be913c41525

This was previously working properly.

Matejkob:
This PR should fix this problem - are you sure this build contains changes from this PR?

PR #2218, opened 2023-09-19
Add HasDeclModifierList trait

rauhul:
Adds a new protocol to enable accessing the modifiers of a decl without needing to switch over concrete syntaxes which contain modifiers. This allows users to extract AccessLevels and other modifiers from decls in a generic manner.


allevato:
Isn't this already supported by `WithModifiersSyntax`?

rauhul:
> Isn't this already supported by `WithModifiersSyntax`?

yes 🤦 I completely missed this...

PR #2222, opened 2023-09-20
Merge macro examples from `ExamplePlugin` target into `MacroExamples`

### LINKED ISSUE #2165: Incorporate the `ExamplePlugin` target  into the `MacroExamples` target
I'm wondering if we still need the [`ExamplePlugin` target in the `Examples` package](https://github.com/apple/swift-syntax/tree/main/Examples/Sources/ExamplePlugin) - was it mainly used for lit tests, and is it now a bit redundant with `MacroExamples`?

_Originally posted by @Matejkob in https://github.com/apple/swift-syntax/pull/2160#pullrequestreview-1614800921_
            
Since [the last lit test](https://github.com/apple/swift-syntax/pull/2160) has been removed we don't need a separated `ExamplePlugin` target so we can merge macros into `MacroExamples`.

Matejkob:
Resolves #2165


Examples/Sources/ExamplePlugin/Macros.swift
```diff
@@ -1,145 +0,0 @@
...
-  ) throws -> [AccessorDeclSyntax] {
-    guard
-      let identifier = declaration.as(VariableDeclSyntax.self)?.bindings.first?.pattern.as(IdentifierPatternSyntax.self)?.identifier
-    else {
-      return []
-    }
-
-    return ["didSet { print(\(identifier)) }"]
-  }
-}
```
Matejkob, line 132: Accessor macro cannot add an observable accessor so this example doesn't work so I removed it.

Examples/Sources/ExamplePlugin/Macros.swift
```diff
@@ -1,145 +0,0 @@
...
-  static func expansion(
-    of node: some FreestandingMacroExpansionSyntax,
-    in context: some MacroExpansionContext
-  ) throws -> [CodeBlockItemSyntax] {
-    guard let expr = node.arguments.first?.expression else {
-      return []
-    }
-    return ["print(\(expr))"]
-  }
-}
```
Matejkob, line 145: The code item macro is still behind a feature flag, so I've decided not to merge it into the official macro examples yet to avoid confusing users.

Examples/Sources/MacroExamples/Playground/PeerMacrosPlayground.swift
```diff
@@ -46,4 +46,15 @@ func runPeerMacrosPlayground() {
...
+  // MARK: - Peer Value With Suffix Name
+
+  @PeerValueWithSuffixName
+  actor Counter {
+    var value = 0
+  }
+
+  let counter = Counter()
+
+  //  print("Peer value with suffix name for \(Counter.self): \(String(describing: Counter_peer))")
```
Matejkob, line 59: Do you have any idea why `Counter_peer` con not be accessed here?

Matejkob (commented):
I've merged almost every macro example. I have some doubts about the quality of certain merged macros and whether they serve as good examples, but I'll leave that to your assessment.

ahoppen (commented):
I’m not sure if all of these examples are all that valuable. I think the ones that do provide additional value are:
- `FuncUniqueMacro` since we don’t have any declaration macro examples right now
- `SendableExtensionMacro` since we don’t have any extension macro examples right now
- `MemberDeprecatedMacro` seems like a nice example and is a lot simpler than our existing MemberAttribute macro
- `PeerValueWithSuffixNameMacro` also seems useful since it’s a very short example of a peer macro and all our current examples are a lot more complicated 

I think the following aren’t really useful and I would just delete them:
- `EchoExpressionMacro` since it doesn’t do anything and we have plenty expression macro examples already
- `EquatableConformanceMacro` since conformance macros aren’t really a thing anymore
- `MetadataMacro`: I think `CaseDetectionMacro` shows similar ideas and is a more realistic example

@rintaro @bnbarham Do you have opinions here as well?

Examples/Sources/MacroExamples/Playground/PeerMacrosPlayground.swift
ahoppen: OK, looks like it might be intentional that peer macro can’t introduce declarations on a local scope. If so, the error message should be improved. Filed https://github.com/apple/swift/issues/68661

bnbarham:
> I’m not sure if all of these examples are all that valuable. I think the ones that do provide additional value are:
> 
> * `FuncUniqueMacro` since we don’t have any declaration macro examples right now
> * `SendableExtensionMacro` since we don’t have any extension macro examples right now
> * `MemberDeprecatedMacro` seems like a nice example and is a lot simpler than our existing MemberAttribute macro
> * `PeerValueWithSuffixNameMacro` also seems useful since it’s a very short example of a peer macro and all our current examples are a lot more complicated
> 
> I think the following aren’t really useful and I would just delete them:
> 
> * `EchoExpressionMacro` since it doesn’t do anything and we have plenty expression macro examples already
> * `EquatableConformanceMacro` since conformance macros aren’t really a thing anymore
> * `MetadataMacro`: I think `CaseDetectionMacro` shows similar ideas and is a more realistic example
> 
> @rintaro @bnbarham Do you have opinions here as well?

That list sounds mostly reasonable to me, though the reasoning for not including `EquatableConformanceMacro` is more because `SendableExtensionMacro` is basically the exact same but with a better implementation.

Matejkob:
Speak of `SendableExtensionMacro`:
image
Do you know what is goin' on?

Examples/Sources/MacroExamples/Playground/PeerMacrosPlayground.swift
Matejkob: Thanks Alex

Examples/Sources/MacroExamples/Implementation/Plugin.swift
```diff
@@ -35,6 +35,13 @@ struct MyPlugin: CompilerPlugin {
     OptionSetMacro.self,
     NewTypeMacro.self,
     URLMacro.self,
+    EchoExpressionMacro.self,
+    FuncUniqueMacro.self,
+    MetadataMacro.self,
+    PeerValueWithSuffixNameMacro.self,
+    MemberDeprecatedMacro.self,
+    EquatableConformanceMacro.self,
+    SendableExtensionMacro.self,
```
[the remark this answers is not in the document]
Matejkob: Done

ahoppen:
> Do you know what is goin' on?

Oh, the macro expansion is considered a different source file in the compiler. Could you file a bug at https://github.com/apple/swift/issues? 

I would suggest we just remove that example as well. We should add an example for an extension macro but I we could do that in a follow-up PR.

bnbarham:
> I would suggest we just remove that example as well. We should add an example for an extension macro but I we could do that in a follow-up PR.

Could just rename to to `Equatable` 😅

Matejkob:
> > I would suggest we just remove that example as well. We should add an example for an extension macro but I we could do that in a follow-up PR.
> 
> Could just rename to to `Equatable` 😅

That could work! 😅

Examples/Sources/MacroExamples/Implementation/Extension/SendableExtensionMacro.swift
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    let sendableExtension: DeclSyntax =
+      """
+      extension \(type.trimmed): Sendable {}
+      """
+
+    guard let extensionDecl = sendableExtension.as(ExtensionDeclSyntax.self) else {
+      return []
+    }
```
bnbarham, line 35: @ahoppen do we have any guidelines on the canonical way to do this? Seems like:
```
let sendableExtension = try ExtensionDeclSyntax("extension \(type.trimmed): Sendable {}")
```
would be nicer to me, since we don't expect that to ever actually fail. We could also actually use `protocols` here rather than assuming `Sendable`.

Matejkob: I'd suggest sticking with `let sendableExtension = try ExtensionDeclSyntax("extension \(type.trimmed): Sendable {}")` for now. We currently lack the capability to pass `protocols` in our unit tests, so keeping this implementation will enable us to write tests.

bnbarham: > We currently lack the capability to pass protocols in our unit tests, so keeping this implementation will enable us to write tests.

Indeed , though that's something else we need to fix :)

Matejkob: Done. Tests are passing ✅

bnbarham: Thanks! 🥳

Matejkob: The pleasure is all mine 😍

bnbarham:
@swift-ci please test

ahoppen:
@swift-ci Please test Windows

ahoppen (approved):

PR #2224, opened 2023-09-21
Do not add whitespace between right angle and period

bnbarham:
`Foo<T>.bar` should not have whitespace added between `>` and `.`.


bnbarham:
@swift-ci please test

ahoppen (approved):
Thanks 👍🏽

PR #2225, opened 2023-09-21
Replace text `"async"` with `keyword(.async)` in macro examples

Matejkob:


Matejkob:
When running tests from the command line, I'm encountering these unusual messages:
```
--------------- throwing @NewType can only be applied to a struct declarations.
--------------- throwing @NewType requires the raw type as an argument, in the form "RawType.self".
```
This is the first time I've seen something like this. Any ideas what's going on?

Screenshot 2023-09-21 at 9 14 11 AM

kimdv:
@swift-ci please test

kimdv (approved):

ahoppen (approved):

PR #2226, opened 2023-09-21
Rewrite FixItApplier to be string based

kimdv:
Part of https://github.com/apple/swift-syntax/pull/2021

### DIFF
--- a/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
+++ b/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
@@ -182,7 +182,6 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
       correctToken.isMissing
     {
       // We are exchanging two adjacent tokens, transfer the trivia from the incorrect token to the corrected token.
-      changes += misplacedTokens.map { FixIt.MultiNodeChange.makeMissing($0, transferTrivia: false) }
       changes.append(
         FixIt.MultiNodeChange.makePresent(
           correctToken,
@@ -191,6 +190,7 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
           trailingTrivia: misplacedToken.trailingTrivia.isEmpty ? nil : misplacedToken.trailingTrivia
         )
       )
+      changes.append(FixIt.MultiNodeChange.makeMissing(misplacedTokens, transferTrivia: false))
     } else {
       changes += misplacedTokens.map { FixIt.MultiNodeChange.makeMissing($0) }
       changes += correctAndMissingTokens.map { FixIt.MultiNodeChange.makePresent($0) }
@@ -236,7 +236,7 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
     exchangeTokens(
       unexpected: misplacedSpecifiers,
       unexpectedTokenCondition: { EffectSpecifier(token: $0) != nil },
-      correctTokens: [effectSpecifiers?.throwsSpecifier, effectSpecifiers?.asyncSpecifier],
+      correctTokens: [effectSpecifiers?.asyncSpecifier, effectSpecifiers?.throwsSpecifier],
       message: { EffectsSpecifierAfterArrow(effectsSpecifiersAfterArrow: $0) },
       moveFixIt: { MoveTokensInFrontOfFixIt(movedTokens: $0, inFrontOf: .arrow) },
       removeRedundantFixIt: { RemoveRedundantFixIt(removeTokens: $0) }
@@ -764,10 +764,8 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
     if let unexpected = node.unexpectedBetweenRequirementAndTrailingComma,
       let token = unexpected.presentTokens(satisfying: { $0.tokenKind == .binaryOperator("&&") }).first,
       let trailingComma = node.trailingComma,
-      trailingComma.isMissing,
-      let previous = node.unexpectedBetweenRequirementAndTrailingComma?.previousToken(viewMode: .sourceAccurate)
+      trailingComma.isMissing
     {
-
       addDiagnostic(
         unexpected,
         .expectedCommaInWhereClause,
@@ -775,9 +773,8 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
           FixIt(
             message: ReplaceTokensFixIt(replaceTokens: [token], replacements: [.commaToken()]),
             changes: [
-              .makeMissing(token),
-              .makePresent(trailingComma),
-              FixIt.MultiNodeChange(.replaceTrailingTrivia(token: previous, newTrivia: [])),
+              .makeMissing(token, transferTrivia: false),
+              .makePresent(trailingComma, leadingTrivia: token.leadingTrivia, trailingTrivia: token.trailingTrivia),
             ]
           )
         ],
@@ -818,7 +815,7 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
         fixIts: [
           FixIt(
             message: RemoveNodesFixIt(nodes),
-            changes: nodes.map { .makeMissing($0) }
+            changes: .makeMissing(nodes)
           )
         ],
         handledNodes: nodes.map { $0.id }
@@ -1542,7 +1539,7 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
         fixIts: [
           FixIt(
             message: RemoveNodesFixIt(rawDelimiters),
-            changes: rawDelimiters.map { .makeMissing($0) }
+            changes: .makeMissing(rawDelimiters)
           )
         ],
         handledNodes: rawDelimiters.map { $0.id }
@@ -1862,8 +1859,8 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
               replacements: [node.colon]
             ),
             changes: [
-              FixIt.MultiNodeChange.makeMissing(equalToken),
-              FixIt.MultiNodeChange.makePresent(node.colon),
+              .makeMissing(equalToken, transferTrivia: false),
+              .makePresent(node.colon, leadingTrivia: equalToken.leadingTrivia, trailingTrivia: equalToken.trailingTrivia),
             ]
           )
         ],
@@ -1971,8 +1968,9 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
             FixIt(
               message: fixItMessage,
               changes: [
-                FixIt.MultiNodeChange.makePresent(detail.detail)
-              ] + unexpectedTokens.map { FixIt.MultiNodeChange.makeMissing($0) }
+                .makePresent(detail.detail),
+                .makeMissing(unexpectedTokens),
+              ]
             )
           ],
           handledNodes: [detail.id] + unexpectedTokens.map(\.id)
--- a/Sources/_SwiftSyntaxTestSupport/FixItApplier.swift
+++ b/Sources/_SwiftSyntaxTestSupport/FixItApplier.swift
@@ -1,0 +1,114 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftDiagnostics
+import SwiftSyntax
+
+public enum FixItApplier {
+  struct Edit: Equatable {
+    var startUtf8Offset: Int
+    var endUtf8Offset: Int
+    let replacement: String
+
+    var replacementLength: Int {
+      return replacement.utf8.count
+    }
+
+    var replacementRange: Range<Int> {
+      return startUtf8Offset..<endUtf8Offset
+    }
+  }
+
+  /// Applies selected or all Fix-Its from the provided diagnostics to a given syntax tree.
+  ///
+  /// - Parameters:
+  ///   - diagnostics: An array of `Diagnostic` objects, each containing one or more Fix-Its.
+  ///   - filterByMessages: An optional array of message strings to filter which Fix-Its to apply.
+  ///     If `nil`, the first Fix-It from each diagnostic is applied.
+  ///   - tree: The syntax tree to which the Fix-Its will be applied.
+  ///
+  /// - Returns: A ``String`` representation of the modified syntax tree after applying the Fix-Its.
+  public static func applyFixes(
+    from diagnostics: [Diagnostic],
+    filterByMessages messages: [String]?,
+    to tree: any SyntaxProtocol
+  ) -> String {
+    let messages = messages ?? diagnostics.compactMap { $0.fixIts.first?.message.message }
+
+    let changes =
+      diagnostics
+      .flatMap(\.fixIts)
+      .filter { messages.contains($0.message.message) }
+      .flatMap(\.changes)
+
+    var edits: [Edit] = changes.map(\.edit)
+    var source = tree.description
+
+    while let edit = edits.first {
+      edits = Array(edits.dropFirst())
+
+      let startIndex = source.utf8.index(source.utf8.startIndex, offsetBy: edit.startUtf8Offset)
+      let endIndex = source.utf8.index(source.utf8.startIndex, offsetBy: edit.endUtf8Offset)
+
+      source.replaceSubrange(startIndex..<endIndex, with: edit.replacement)
+
+      edits = edits.compactMap { remainingEdit -> FixItApplier.Edit? in
+        var remainingEdit = remainingEdit
+
+        if remainingEdit.replacementRange.overlaps(edit.replacementRange) {
+          // The edit overlaps with the previous edit. We can't apply both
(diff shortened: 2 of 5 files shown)


Sources/_SwiftSyntaxTestSupport/FixItApplier.swift
```diff
@@ -0,0 +1,107 @@
...
+    let endUtf8Offset: Int
+    let replacement: String
+  }
+
+  private let changes: [FixIt.Change]
+  private let tree: any SyntaxProtocol
+
+  private lazy var locationConverter: SourceLocationConverter = {
+    return SourceLocationConverter(fileName: "", tree: tree)
+  }()
```
Matejkob, line 28: I belive you are not using this property 
```suggestion
```

Sources/_SwiftSyntaxTestSupport/FixItApplier.swift
```diff
@@ -0,0 +1,107 @@
...
+
+  init(diagnostics: [Diagnostic], withMessages messages: [String]?, tree: any SyntaxProtocol) {
+    let messages = messages ?? diagnostics.compactMap { $0.fixIts.first?.message.message }
```
Matejkob, line 31: Am I understanding this correctly? When `messages`` is nil, we want to include all the Fix-Its from the `diagnostics``, right? However, doesn't this code `diagnostics.compactMap { $0.fixIts.first?.message.message }`` only add the message from the first Fix-It of each diagnostic?

If my understanding is correct, perhaps we should have something like this instead:

```swift
self.changes =
      diagnostics
      .flatMap(\.fixIts)
      .filter { messages?.isEmpty() == true ? messages?.contains($0.message.message) : false } // Is nil representing the same state as empty array? If not instead of `.isEmpty()` we can use `== nil ? false : messages?.contains($0.message.message)`
      .flatMap(\.changes) 
```
instead.

Sources/_SwiftSyntaxTestSupport/FixItApplier.swift
Matejkob, line 1: Any chance we could add some unit tests specifically for FixItApplier? It'd really help boost our confidence in the code.

Matejkob (commented):
It's always a joy to dive into a PR that's pushing our codebase forward, and this one is no exception.

I've left some comments and suggestions to consider. 

Looking forward to the next iterations and can't wait to get these improvements merged in. 💙

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -1968,8 +1965,9 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
             FixIt(
               message: fixItMessage,
               changes: [
-                FixIt.MultiNodeChange.makePresent(detail.detail)
-              ] + unexpectedTokens.map { FixIt.MultiNodeChange.makeMissing($0) }
+                FixIt.MultiNodeChange.makePresent(detail.detail),
```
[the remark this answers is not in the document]
kimdv: Yes I think we should make is consistent. So good catch here

Sources/_SwiftSyntaxTestSupport/FixItApplier.swift
kimdv: We are already running more than 2500+ tests 😅
So I think it should be good if they all are green IMO

Sources/_SwiftSyntaxTestSupport/FixItApplier.swift
kimdv: The reason is because some diagnostics give us two Fix-Its.
So we default to the first one.

Sources/_SwiftSyntaxTestSupport/FixItApplier.swift
```diff
@@ -0,0 +1,107 @@
...
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftDiagnostics
+import SwiftSyntax
+
+public class FixItApplier {
```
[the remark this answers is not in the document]
kimdv: Nice idea! Applied

Sources/_SwiftSyntaxTestSupport/FixItApplier.swift
Matejkob: Wow, 2500+ tests is impressive 😅. I totally get where you're coming from with the "if they're green, we're good" philosophy. I usually lean more towards unit testing individual components like `FixItApplier` for granular control and easier debugging. But I guess in the context of our project, with such extensive test coverage already, integration tests might indeed suffice. 😊

Sources/_SwiftSyntaxTestSupport/FixItApplier.swift
Matejkob: I'm thinking... we might need to revisit this based on what we've stated in PR #2021. According to the documentation for the `assert` method there, if `applyFixIts` is set to `nil`, it should apply all Fix-Its from the diagnostics. If we continue to only map the first Fix-It, we might end up with a discrepancy between the code behavior and the documentation, which could be confusing for end-users. 

I'm not sure if you are going to address it in the #2021 PR so I've added this comment here 🫶

kimdv: We should document it yes.
The reason is because when we have a Fix-It that offer two solutions, what one should we apply?
We cannot apply both of them.

You can take a loot at this example etc
https://github.com/apple/swift-syntax/blob/7c6d4f729f37f5e5a206b39ffa7bc1fac9479c24/Tests/SwiftParserTest/translated/RecoveryTests.swift#L993-L1017

Sources/_SwiftSyntaxTestSupport/FixItApplier.swift
kimdv: Yeah, with all what there is I think it's sufficient

Sources/_SwiftSyntaxTestSupport/FixItApplier.swift
Matejkob: Ah, I get it now! Thanks for clarifying that the `fixIts` collection in `Diagnostic` is designed to offer different solutions for resolving the same diagnostic issue. I was under the impression that it's a collection where each Fix-It is a part of a multi-step solution for the problem at hand. For example, one changing a particular node and another modifies something else to fully resolve the issue. Now it all makes sense! Appreciate the explanation. 🙈

kimdv:
@swift-ci please test

Sources/_SwiftSyntaxTestSupport/FixItApplier.swift
```diff
@@ -0,0 +1,149 @@
...
+      return startUtf8Offset..<endUtf8Offset
+    }
+  }
+
+  /// Applies selected or all Fix-Its from the provided diagnostics to a given syntax tree.
+  ///
+  /// - Parameters:
+  ///   - diagnostics: An array of `Diagnostic` objects, each containing one or more Fix-Its.
+  ///   - filterByMessages: An optional array of message strings to filter which Fix-Its to apply.
+  ///                      If `nil`, all Fix-Its in `diagnostics` are applied.
```
ahoppen, line 40: AFAICT this comment isn’t actually correct.

```suggestion
  ///     If `nil`, the first Fix-It from each diagnostic is applied.
```

Sources/_SwiftSyntaxTestSupport/FixItApplier.swift
```diff
@@ -0,0 +1,149 @@
...
+
+      case .replaceTrailingTrivia(let token, let newTrivia):
+        edits.append(
+          Edit(
+            startUtf8Offset: token.endPositionBeforeTrailingTrivia.utf8Offset,
+            endUtf8Offset: token.endPosition.utf8Offset,
+            replacement: newTrivia.description
+          )
+        )
+      }
```
ahoppen, line 87: What do you think about making this `switch` an extension on `Change`?

I could very selfishly use that for https://github.com/apple/sourcekit-lsp/pull/902/files#diff-6147b806788b196876d46885643488f44c48b14791d7adc3ef5d2d91b361bd11R74-R76 😉

Sources/_SwiftSyntaxTestSupport/FixItApplier.swift
```diff
@@ -0,0 +1,149 @@
...
+    // As we need to start apply the edits at the end of a source, start by reversing edit
+    // and then sort edits by decrementing start offset. If they are equal then descrementing end offset.
+    //      edits = edits.reversed().sorted(by: { edit1, edit2 in
+    //      if edit1.startUtf8Offset == edit2.startUtf8Offset {
+    //        return edit1.endUtf8Offset > edit2.endUtf8Offset
+    //      } else {
+    //        return edit1.startUtf8Offset > edit2.startUtf8Offset
+    //      }
+    //    })
+
```
ahoppen, line 101: Leftover?

Sources/_SwiftSyntaxTestSupport/FixItApplier.swift
```diff
@@ -0,0 +1,149 @@
...
+        var remainingEdit = remainingEdit
+
+        if remainingEdit.replacementRange.overlaps(edit.replacementRange) {
+          // The edit overlaps with the previous edit. We can't apply both
+          // without conflicts. Apply the one that's listed first and drop the
+          // later edit.
+          return nil
+        }
+
+        // Keep the origianl length as we start to edit the offset.
```
ahoppen, line 120:
```suggestion
        // Keep the original length as we start to edit the offset.
```

Sources/_SwiftSyntaxTestSupport/FixItApplier.swift
```diff
@@ -0,0 +1,149 @@
...
+        let originalLength = remainingEdit.originalLength
+
+        // If the current is after the edit we want to update, don't change the offset.
+        // If the remaining edit have same start and end position, don't change the start offset.
```
ahoppen, line 124: I think the second line is no longer correct because you are checking for `<=` below (or I’m missing something). And I think it’s easier to understand if you phrase it based on the remaining edit than the current edit.

```suggestion
        // If the remaining edit starts after or at the end of the edit that we just applied, shift it by the current edit's difference in length.
```

Sources/_SwiftSyntaxTestSupport/FixItApplier.swift
```diff
@@ -0,0 +1,149 @@
...
+        if edit.endUtf8Offset <= remainingEdit.startUtf8Offset {
+          // Take the largest value as we in some cases can end up with a negative offset.
+          remainingEdit.startUtf8Offset = Swift.max(remainingEdit.startUtf8Offset - edit.originalLength + edit.replacementLength, 0)
```
ahoppen, line 127: Why do we need `Swift.max` here? I think `remainingEdit.startUtf8Offset - edit.originalLength + edit` should always be ≥ 0.

And since I haven’t done a proof in a while and I like them, here’s the proof😉 

```
    remainingEdit.startUtf8Offset - edit.originalLength + edit.replacementLength ≥ 0
<=> remainingEdit.startUtf8Offset - (edit.endOffset - edit.startOffset) + edit.replacementLength ≥ 0
<=> remainingEdit.startUtf8Offset - edit.endOffset + edit.startOffset + edit.replacementLength ≥ 0
<=  remainingEdit.startUtf8Offset - edit.endOffset ≥ 0
<=> remainingEdit.startUtf8Offset ≥ edit.endOffset
(true by the check if the if-condition)
```

I would be fine with adding a precondition for ≥ 0 if you want to.

Sources/_SwiftSyntaxTestSupport/FixItApplier.swift
```diff
@@ -0,0 +1,149 @@
...
+        }
+
+        // If we need to replace a string at the end, we need to ensure we don't get an out-of-bounds error.
+        remainingEdit.endUtf8Offset = Swift.min(remainingEdit.startUtf8Offset + originalLength, source.utf8.count)
```
ahoppen, line 131: Honestly, I would be less surprised by this if you didn’t store `originalLength` and instead did the same adjustment that you do to `startUtf8Offset` also to `endUtf8Offset` inside the `if`.

Sources/_SwiftSyntaxTestSupport/FixItApplier.swift
```diff
@@ -0,0 +1,149 @@
...
+}
+
+//extension Array where Element == FixItApplier.Edit {
+//  fileprivate func canInsert(editToApply: Element) -> Bool {
+//    return self.contains { edit in
+//      return
+//        !(editToApply.startUtf8Offset >= edit.startUtf8Offset
+//        && editToApply.endUtf8Offset < edit.endUtf8Offset)
+//    }
+//  }
```
ahoppen, line 148: Leftover?

Tests/SwiftParserTest/ExpressionTests.swift
```diff
@@ -963,23 +963,23 @@ final class ExpressionTests: ParserTestCase {
...
+    //
+    //    assertParse(
+    //      "let _ = [Int throws 1️⃣Int]()",
+    //      diagnostics: [
+    //        DiagnosticSpec(message: "expected '->' in array element", fixIts: ["insert '->'"])
+    //      ],
+    //      fixedSource: """
+    //        let _ = [Int throws -> Int]()
+    //        """
+    //    )
```
ahoppen, line 982: Any reason why you disabled these? AFAICT they pass.

Tests/SwiftSyntaxMacroExpansionTest/PeerMacroTests.swift
ahoppen, line 1: What’s the reason that you refactored this file? It seems pretty unrelated to this PR (but I might be missing something).

ahoppen (commented):
Wohoo. Looks good to me. Just a few localized comments.

Tests/SwiftSyntaxMacroExpansionTest/PeerMacroTests.swift
kimdv: It was to ensure it worked correctly. I will commit it with #2021

Sources/_SwiftSyntaxTestSupport/FixItApplier.swift
```diff
@@ -0,0 +1,123 @@
...
+      return startUtf8Offset..<endUtf8Offset
+    }
+  }
+
+  /// Applies selected or all Fix-Its from the provided diagnostics to a given syntax tree.
+  ///
+  /// - Parameters:
+  ///   - diagnostics: An array of `Diagnostic` objects, each containing one or more Fix-Its.
+  ///   - filterByMessages: An optional array of message strings to filter which Fix-Its to apply.
+  ///   If `nil`, the first Fix-It from each diagnostic is applied..
```
ahoppen, line 36:
```suggestion
  ///     If `nil`, the first Fix-It from each diagnostic is applied.
```

Sources/_SwiftSyntaxTestSupport/FixItApplier.swift
```diff
@@ -0,0 +1,123 @@
...
+      )
+    }
+  }
+}
+
+fileprivate extension Array where Element == FixIt.Change {
+  var edits: [FixItApplier.Edit] {
+    return self.map { $0.edit }
+  }
+}
```
ahoppen, line 123: Do we really need this? Wouldn’t it be simpler if we just do `changes.map { $0.edit }` inline above?

Sources/_SwiftSyntaxTestSupport/FixItApplier.swift
```diff
@@ -0,0 +1,123 @@
...
+          // The edit overlaps with the previous edit. We can't apply both
+          // without conflicts. Apply the one that's listed first and drop the
+          // later edit.
+          return nil
+        }
+
+        // If the remaining edit starts after or at the end of the edit that we just applied,
+        // shift it by the current edit's difference in length.
+        if edit.endUtf8Offset <= remainingEdit.startUtf8Offset {
+          // Take the largest value as we in some cases can end up with a negative offset.
```
ahoppen, line 77: I think this comment is out-of-date now

Sources/_SwiftSyntaxTestSupport/FixItApplier.swift
```diff
@@ -0,0 +1,123 @@
...
+          remainingEdit.startUtf8Offset = remainingEdit.startUtf8Offset - edit.replacementRange.count + edit.replacementLength
+
+          // If we need to replace a string at the end, we need to ensure we don't get an out-of-bounds error.
```
ahoppen, line 80: I think this comment is also out-of-date and can be removed.

Sources/_SwiftSyntaxTestSupport/FixItApplier.swift
```diff
@@ -0,0 +1,114 @@
...
+      return startUtf8Offset..<endUtf8Offset
+    }
+  }
+
+  /// Applies selected or all Fix-Its from the provided diagnostics to a given syntax tree.
+  ///
+  /// - Parameters:
+  ///   - diagnostics: An array of `Diagnostic` objects, each containing one or more Fix-Its.
+  ///   - filterByMessages: An optional array of message strings to filter which Fix-Its to apply.
+  ///     If `nil`, the first Fix-It from each diagnostic is applied..
```
ahoppen, line 36: Nitpick: Still has two periods

```suggestion
  ///     If `nil`, the first Fix-It from each diagnostic is applied.
```

ahoppen (approved):
Just one nitpick

kimdv:
@swift-ci please test

kimdv:
@swift-ci please windows

kimdv:
@swift-ci please windows

ahoppen:
@swift-ci Please test Windows

PR #2227, opened 2023-09-21
Format `expectedExpandedSource` in `assertStringsEqualWithDiff` before comparing

BrianHenryIE:
I was encountering mis-matches between my `expectedExpandedSource` and the source returned from the `expansion` of my macro, as tested by `assertStringsEqualWithDiff`.

[Discussed on swift-open-source Slack](https://swift-open-source.slack.com/archives/C04N0ET1K3Q/p1695275043763609)

As I understand what is happening, when I return a string in the `MemberMacro::expansion() -> [SwiftSyntax.DeclSyntax]`, it will be parsed and then its AST will be printed in `assertMacroExpansion()` using `expandedSourceFile.description` which will reflect the standard way of printing it, not the way it was written as a string in the macro code.

e.g. my macro code has:

```
return [
                """
                func aUselessFunction() {}
                """
            ]
```

and `expandedSourceFile.description` opens the brackets:

```
struct MyStruct{

    func aUselessFunction() {
    }
}
```

my test expectation is written as:

```
struct MyStruct{

    func aUselessFunction() {}
}
```

As I write the macro, I'm copying and pasting between the macro code and the test code, assuming most lines will remain the same.

Similarly, switch cases were expanded from one line to two

```
return [
    """
    func aUselessFunction() -> Bool? {
        let foo = "bar"
        switch "foo" {
            case "bar": return true
            default: return nil
        }
    }
    """
]
```

```
Actual expanded source:

struct MyStruct{

    func aUselessFunction() -> Bool? {
        let foo = "bar"
        switch "foo" {
            case "bar":
            return true
            default:
            return nil
        }
    }
}
```

By running the `expectedExpandedSource` through `Parser.parse()` and comparing it `.formatted()` with `expandedSourceFile.formatted()` we are comparing like with like.

The downside to this is that in the failure reports' "Macro expansion did not produce the expected expanded source", the "expected output" is now not _exactly_ what the user has in their test case.

I think that's a reasonable trade-off, especially since the code that really matters is the macro output and that is already formatted, so this is a time-saver for developers.

Let me know if that's ok and I'll add some tests to the PR.

### DIFF
--- a/Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
+++ b/Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
@@ -292,8 +292,8 @@ public func assertMacroExpansion(
   }
 
   assertStringsEqualWithDiff(
-    expandedSourceFile.description.trimmingCharacters(in: .newlines),
-    Parser.parse(source: expectedExpandedSource).description.trimmingCharacters(in: .newlines),
+    expandedSourceFile.formatted().description.trimmingCharacters(in: .newlines),
+    Parser.parse(source: expectedExpandedSource).formatted().description.trimmingCharacters(in: .newlines),
     "Macro expansion did not produce the expected expanded source",
     additionalInfo: """
       Actual expanded source:
--- a/Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
+++ b/Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
@@ -292,8 +292,8 @@ public func assertMacroExpansion(
   }
 
   assertStringsEqualWithDiff(
+    expandedSourceFile.description.trimmingCharacters(in: .newlines),
     Parser.parse(source: expectedExpandedSource).description.trimmingCharacters(in: .newlines),
-    expectedExpandedSource.trimmingCharacters(in: .newlines),
     "Macro expansion did not produce the expected expanded source",
     additionalInfo: """
       Actual expanded source:
--- a/Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
+++ b/Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
@@ -292,7 +292,7 @@ public func assertMacroExpansion(
   }
 
   assertStringsEqualWithDiff(
-    expandedSourceFile.description.trimmingCharacters(in: .newlines),
+    Parser.parse(source: expectedExpandedSource).description.trimmingCharacters(in: .newlines),
     expectedExpandedSource.trimmingCharacters(in: .newlines),
     "Macro expansion did not produce the expected expanded source",
     additionalInfo: """


ahoppen:
I don’t think this is change something we want to do. As you already mentioned, the major downside is that you are no longer comparing the generated source to the expected source that you specified in the test case, which is counterintuitive.

Also, in the case of the macro that you describe in the PR’s description, the macro expansion in the compiler will actually insert the formatted source code 
```swift
func aUselessFunction() {
}
```
instead of 
```swift
func aUselessFunction() {}
```

So `assertMacroExpansion` matches what the compiler is doing.

If you do want to control the formatting of your macro manually, you can set the [`formatMode`](https://swiftpackageindex.com/apple/swift-syntax/main/documentation/swiftsyntaxmacros/macro/formatmode-10dhi) to `.disabled`. In that case no formatting will be applied in `assertMacroExpansion` or the compiler expansion.

BrianHenryIE:
OK. I don't think I was entirely clear above, so here's some examples.

Given a macro:

```swift
public struct MyTestFuncMacro: MemberMacro {

    public static func expansion(
        of node: SwiftSyntax.AttributeSyntax, 
        providingMembersOf declaration: some SwiftSyntax.DeclGroupSyntax, 
        in context: some SwiftSyntaxMacros.MacroExpansionContext
        ) throws -> [SwiftSyntax.DeclSyntax] {
        return [
            """
            func aUselessFunction() {}
            """
        ]
    }
}
```

Here are some tests that unintuitively fail:

```swift
    func testFuncMacro1() throws {
        assertMacroExpansion(
            """
            @MyTestFuncMacro
            struct MyStruct{
            }
            """,
            expandedSource: """
            struct MyStruct{

                func aUselessFunction() { }
            }
            """,
            macros: testMacros
        )
    }
```

We know this first one fails because of the expanded function.

```swift
    func testFuncMacro2() throws {
        assertMacroExpansion(
            """
            @MyTestFuncMacro
            struct MyStruct{
            }
            """,
            expandedSource: """
            struct MyStruct {

                func aUselessFunction() {
                }
            }
            """,
            macros: testMacros
        )
    }
```

Whitespace.

```swift
    func testFuncMacro3() throws {
        assertMacroExpansion(
            """
            @MyTestFuncMacro
            struct MyStruct{
            }
            """,
            expandedSource: """
            struct MyStruct{

                func aUselessFunction() { 
                }
            }
            """,
            macros: testMacros
        )
    }
```

Different whitespace.

```swift
    func testFuncMacro4() throws {
        assertMacroExpansion(
            """
            @MyTestFuncMacro
            struct MyStruct{
            }
            """,
            expandedSource: """
            struct MyStruct{
                func aUselessFunction() {
                }
            }
            """,
            macros: testMacros
        )
    }
```

The blank line (not seen here – it is added by the macro expansion).

---

I would not characterize the downside as "major" because the test feedback is already giving a different output than what was input in the `MemberMacro::expansion()` function. 

I wasn't aware of `formatMode` but since the default is formatted, it's feels reasonable to format the `expectedExpandedSource` too.

The `formatMode` value could be used in `assertMacroExpansion`, either pulling directly from the `Macro` value, or as a new parameter to `assertMacroExpansion`, to toggle the parsing of the `expectedExpandedSource` too.

This is a tedious issue to resolve right now. I've started wrapping my `expectedExpandedSource` input with `Parser.parse(source: ...).formatted().description` but it doesn't feel the best way to address it.

ahoppen:
> Here are some tests that unintuitively fail.

The failure here actually teaches you something about how the macro is expanded. Your syntax tree *will* get formatted, and the expanded macro *will* have a newline in between the braces unless you specify `formatMode = .disabled`. This is how macro expansion works and it is needed eg. if you built the syntax tree yourself using
```swift
FunctionDeclSyntax(
  attributes: [],
  modifiers: [],
  funcKeyword: .keyword(.func),
  name: .identifier("foo"),
  signature: FunctionSignatureSyntax(parameterClause: FunctionParameterClauseSyntax(parameters: [])),
  body: CodeBlockSyntax(statements: [])
)
```

and we wouldn’t format the source code, you would get `funcfoo(){}`, which isn’t valid since the space between `func` and `foo` is missing. And that’s totally intentional and by design since no trailing space was specified for the `func` keyword.

> I would not characterize the downside as "major" because the test feedback is already giving a different output than what was input in the `MemberMacro::expansion()` function.

With that in mind, yes, the output is different than `MemberMacro.expansion` but there’s more going on than just calling `expansion`, namely formatting, inserting the expanded macro at the correct location, removing the attribute, …

> I wasn't aware of formatMode but since the default is formatted, it's feels reasonable to format the `expectedExpandedSource` too.
> 
> The `formatMode` value could be used in `assertMacroExpansion`, either pulling directly from the `Macro` value, or as a new parameter to `assertMacroExpansion`, to toggle the parsing of the `expectedExpandedSource` too.

That will be counter-intuitive as well. While expanding the macro, only the text that the macro expands to is being formatted, not the entire source file. If we formatted the entire expected source file, then you will run into issue if your test input’s formatting doesn’t agree with swift-syntax’s formatting.

ahoppen:
Closing this because this isn’t something we want to take as mentioned in my previous comments.

PR #2231, opened 2023-09-22
Update assertions so that the mutation is more clear

bnbarham:
Move the text for test failure reason to the message itself, rather than the more generic "actual didn't match expected".

Also fixes `failStringsEqualWithDiff` on non-macOS, where we weren't actually printing the actual or expected output at all.

### DIFF
--- a/Tests/SwiftParserTest/Assertions.swift
+++ b/Tests/SwiftParserTest/Assertions.swift
@@ -531,7 +531,7 @@ public struct AssertParseOptions: OptionSet {
 extension ParserTestCase {
   /// After a test case has been mutated, assert that the mutated source
   /// round-trips and doesn’t hit any assertion failures in the parser.
-  fileprivate func assertRoundTrip<S: SyntaxProtocol>(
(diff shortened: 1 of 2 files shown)


bnbarham:
@swift-ci please test

bnbarham:
@kavon this would now look like:
```
[SwiftParserTest.DeclarationTests testSuppressedImplicitConformance] : failed - A mutation of the original test case failed to round-trip
–struct Hello: ~{}
+struct Hello: {}

Parsed syntax tree of mutation:
SourceFileSyntax
├─statements: CodeBlockItemListSyntax
...
```

Thoughts?

Tests/SwiftParserTest/Assertions.swift
```diff
@@ -549,13 +549,9 @@ extension ParserTestCase {
       assertStringsEqualWithDiff(
         "\(mutatedTree)",
         mutatedSource,
+        "A mutation of the original test case failed to round-trip",
```
kavon, line 552: Much clearer!

kavon (approved):

ahoppen (approved):

PR #2233, opened 2023-09-24
Fixes a small threading issue in code generation and a TODO

benpious:
`errors` was potentially mutated on multiple threads. `previouslyGeneratedFiles`, which is mutated on every loop, is locked just above it, so clearly we need to lock on _all_ state that's mutated here. 

I also cleaned up the TODO.

### DIFF
--- a/CodeGeneration/Sources/generate-swift-syntax/GenerateSwiftSyntax.swift
+++ b/CodeGeneration/Sources/generate-swift-syntax/GenerateSwiftSyntax.swift
@@ -154,6 +154,7 @@ struct GenerateSwiftSyntax: ParsableCommand {
       }
     )
 
+    let errorsLock = NSLock()
     var errors: [Error] = []
     DispatchQueue.concurrentPerform(iterations: fileSpecs.count) { index in
       let fileSpec = fileSpecs[index]
@@ -173,13 +174,23 @@ struct GenerateSwiftSyntax: ParsableCommand {
           verbose: verbose
         )
       } catch {
+        errorsLock.lock()
         errors.append(error)
+        errorsLock.unlock()
       }
     }
 
-    if let firstError = errors.first {
-      // TODO: It would be nice if we could emit all errors
-      throw firstError
+    if errors.count > 0 {
+      struct ComposititeError: Error, CustomStringConvertible {
+        var errors: [Error]
+        var description: String {
+          "Multiple errors emitted: \n" +
+          errors
+            .map(String.init(describing: ))
+            .joined(separator: "\n")
+        }
+      }
+      throw ComposititeError(errors: errors)
     }
 
     for file in previouslyGeneratedFiles {
--- a/CodeGeneration/Sources/generate-swift-syntax/GenerateSwiftSyntax.swift
+++ b/CodeGeneration/Sources/generate-swift-syntax/GenerateSwiftSyntax.swift
@@ -216,9 +216,9 @@ struct GenerateSwiftSyntax: ParsableCommand {
 struct ComposititeError: Error, CustomStringConvertible {
   var errors: [Error]
   var description: String {
-    "Multiple errors emitted:\n" +
-    errors
-      .map(String.init(describing: ))
+    "Multiple errors emitted:\n"
+      + errors
+      .map(String.init(describing:))
       .joined(separator: "\n")
   }
 }
--- a/CodeGeneration/Sources/generate-swift-syntax/GenerateSwiftSyntax.swift
+++ b/CodeGeneration/Sources/generate-swift-syntax/GenerateSwiftSyntax.swift
@@ -139,7 +139,7 @@ struct GenerateSwiftSyntax: ParsableCommand {
 
     let modules = Set(fileSpecs.compactMap { $0.pathComponents.first })
 
-    let previouslyGeneratedFilesLock = NSLock()
+    let filesAndErrorsLock = NSLock()
     var previouslyGeneratedFiles = Set(
       modules.flatMap { (module) -> [URL] in
         let generatedDir =
@@ -154,7 +154,6 @@ struct GenerateSwiftSyntax: ParsableCommand {
       }
     )
 
-    let errorsLock = NSLock()
     var errors: [Error] = []
     DispatchQueue.concurrentPerform(iterations: fileSpecs.count) { index in
       let fileSpec = fileSpecs[index]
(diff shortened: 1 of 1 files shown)


kimdv:
@swift-ci please test

CodeGeneration/Sources/generate-swift-syntax/GenerateSwiftSyntax.swift
```diff
@@ -154,6 +154,7 @@ struct GenerateSwiftSyntax: ParsableCommand {
       }
     )
 
+    let errorsLock = NSLock()
```
bnbarham, line 157: Seems like we should use async/await with multiple tasks that we reduce after here, but that's a larger change. For this IMO just rename `previouslyGenerateFilesLock` and use it for both.

CodeGeneration/Sources/generate-swift-syntax/GenerateSwiftSyntax.swift
```diff
@@ -173,13 +174,23 @@ struct GenerateSwiftSyntax: ParsableCommand {
...
         errors.append(error)
+        errorsLock.unlock()
       }
     }
 
-    if let firstError = errors.first {
-      // TODO: It would be nice if we could emit all errors
-      throw firstError
+    if errors.count > 0 {
+      struct ComposititeError: Error, CustomStringConvertible {
```
bnbarham, line 184:
```suggestion
      struct CompositeError: Error, CustomStringConvertible {
```

Personally I'd define this top level, probably at the bottom of this file.

CodeGeneration/Sources/generate-swift-syntax/GenerateSwiftSyntax.swift
```diff
@@ -173,13 +174,23 @@ struct GenerateSwiftSyntax: ParsableCommand {
...
+        var errors: [Error]
+        var description: String {
+          "Multiple errors emitted: \n" +
```
bnbarham, line 187:
```suggestion
          "Multiple errors emitted:\n" +
```

CodeGeneration/Sources/generate-swift-syntax/GenerateSwiftSyntax.swift
```diff
@@ -173,13 +174,23 @@ struct GenerateSwiftSyntax: ParsableCommand {
...
+          errors
+            .map(String.init(describing: ))
+            .joined(separator: "\n")
```
bnbarham, line 190: Do they already have a newline, or would it be worth adding an extra newline here? Do you happen to have an example of the output here for multiple errors?

bnbarham (commented):
Thanks @benpious! The failures are because `format.py` needs to be run.

CodeGeneration/Sources/generate-swift-syntax/GenerateSwiftSyntax.swift
benpious: Agreed that that should be the long term direction. Is this allowed in this project? I don't think that change should take long to make, provided the CI can verify that it works on Linux and windows.

CodeGeneration/Sources/generate-swift-syntax/GenerateSwiftSyntax.swift
benpious: I can make you some real output in a few hours, but in my original testing `NSError` does not include a newline, and that's the error that we're most likely to encounter in this code path today due to the use of `FileManager`. 

I think in general that it will depend on how the error implements its `description` or `debugDescription` var and that the safest option is to add a newline even if we can find some errors that happen to do this for us.

CodeGeneration/Sources/generate-swift-syntax/GenerateSwiftSyntax.swift
bnbarham: We require being able to compile with >= 5.7 currently and this is just a generation tool, so unless there's some bugs preventing it, I don't see any reason it would be an issue.

CodeGeneration/Sources/generate-swift-syntax/GenerateSwiftSyntax.swift
bnbarham: Sorry, to be clear, we definitely want at least one newline here. My question was about whether we might want *two* so that there's a clearer cut between the errors.

benpious: Ah, yeah that makes a lot more sense. Here's the sample output of an `NSError` in this context: 
```
Error: Multiple errors emitted: 
Error Domain=test error Code=0 "(null)"
Error Domain=test error Code=0 "(null)"
Error Domain=test error Code=0 "(null)"
Error Domain=test error Code=0 "(null)"
Error Domain=test error Code=0 "(null)"
Error Domain=test error Code=0 "(null)"
```
I think in an ideal world it would look something like 
```
Multiple errors emitted: {

    Error Domain=test error Code=0 "(null)"

    Error Domain=test error Code=0 "(null)"

   ...

}
```
because they're separate, but grouped. But that would require adding a means of indenting text (in case of multi-line errors) which might be overkill. But I'm happy to add the indentation and newline, or just the second newline, or leave it out, whichever you prefer.

CodeGeneration/Sources/generate-swift-syntax/GenerateSwiftSyntax.swift
benpious: [Here is an async version of this pr](https://github.com/apple/swift-syntax/compare/main...benpious:swift-syntax:switch-to-async). If this looks good to you I can abandon this pr and pr the Structured Concurrency version instead.

bnbarham: LGTM, happy for you to put that up instead. Regarding the grouping below, I wouldn't worry about indenting personally - this is just for generation which typically shouldn't fail. What you could do is just separate them by file path, eg. ```
```
Errors occurred during generation.

'<file>':
<error here>
<empty newline>
'<file>':
<error here>
```

benpious: Created https://github.com/apple/swift-syntax/pull/2245. It also incorporates this feedback about printing which file produced which error.

ahoppen: Nice, using structured concurrency is so much cleaner. Thanks for putting it up. This means we can close this PR, right?

ahoppen:
Closing because it’s been superseded by https://github.com/apple/swift-syntax/pull/2245 AFAICT.

PR #2234, opened 2023-09-24
Require Swift 5.8 for SPI experimental feature syntax nodes

hamishknight:
Swift versions prior to 5.8 did not correctly handle SPI enum cases, so avoid marking experimental feature nodes SPI if building with such a compiler.


hamishknight:
@swift-ci please test

ahoppen (approved):

PR #2236, opened 2023-09-25
Add some more documentaiton on `assertParse`

ahoppen:


Tests/SwiftParserTest/Assertions.swift
```diff
@@ -233,7 +233,7 @@ struct NoteSpec {
 
 /// An abstract data structure to describe how a diagnostic produced by the parser should look like.
 struct DiagnosticSpec {
-  /// The name of a maker (of the form `1️⃣`) in the source code that marks the location where the diagnostis should be produced.
+  /// The name of a maker (of the form `1️⃣`) in the source code that marks the location where the diagnostic should be produced.
```
bnbarham, line 236:
```suggestion
  /// The name of a marker (of the form `1️⃣`) in the source code that marks the location where the diagnostic should be produced.
```

Tests/SwiftParserTest/Assertions.swift
```diff
@@ -549,7 +549,11 @@ extension ParserTestCase {
       assertStringsEqualWithDiff(
         "\(mutatedTree)",
         mutatedSource,
-        "A mutation of the original test case failed to round-trip",
+        """
+        A mutation of the original test case failed to round-trip. 
+
+        To debug the failure, add a new `assertParse` test with the mutated source as input.
```
bnbarham, line 555: You could add this part to `additionalInfo` maybe? Or do you think it could be missed if it was after the diff?

Tests/SwiftParserTest/Assertions.swift
```diff
@@ -560,23 +564,42 @@ extension ParserTestCase {
...
+  ///    parsed tree is the same as the input.
+  /// 2. Checks that parsing produces the expected list of diagnostics. If no
+  ///    diagnostics are passed, asserts that the input parses without any errors.
+  /// 3. Checks that applying all Fix-Its of the source code results in the
+  ///    expected fixed source, effectively testing the Fix-Its.
+  /// 4. If a substructure is passed, asserts that the parsed tree contains a
+  ///    subtree of that structure.
+  /// 5. Mutates the test input by flipping each token's presence (ie. for every
+  ///    token, remove it from the input if it is present in the parsed tree and
+  ///    synthesize a missing tokens into the input) and verifies that this
```
bnbarham, line 580:
```suggestion
  ///    synthesize any missing tokens into the input) and verifies that this
```

kimdv (approved):

Tests/SwiftParserTest/Assertions.swift
ahoppen: Good idea 👍🏽

Tests/SwiftParserTest/Assertions.swift
ahoppen: This is intentionally `a` because only a single to ken will be synthesized not all of them.

bnbarham: How about:
```
  /// 5. Mutates the test input by flipping each token's presence (ie. for every
  ///    token, either remove it from the input if it is present in the parsed tree or
  ///    synthesize it if it was missing) and verifies that this
```

ahoppen: Sounds good

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

bnbarham (approved):

PR #2237, opened 2023-09-25
Deprecate placeholder nodes

kimdv:


CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
```diff
@@ -548,39 +548,6 @@ public let DECL_NODES: [Node] = [
     elementChoices: [.designatedType]
   ),
 
-  Node(
-    kind: .editorPlaceholderDecl,
-    base: .decl,
-    nameForDiagnostics: "editor placeholder",
```
kimdv, line 554: @ahoppen we don't have nay functionally for setting an complete node as deprecated.
Should we add it?

ahoppen: Sorry for the delayed response. We needed to discuss how to handle this. 

We should deprecate the nodes. The parser should just stop generating them but that way they will still exist for anyone using swift-syntax for code generation and it will make transitioning to a new swift-syntax version easier because the deprecation warnings can guide you in the right direction.

kimdv: Would the best approach to to just copy the needed code and place it in the deprecated files or should we introduce a new field on the `Node` where we can deprecate it?

Because we also need to ensure the parser have the `open func visit(_ node: EditorPlaceholderDeclSyntax) -> DeclSyntax` that is deprecated. 
Otherwise it will break

ahoppen: I think we need to properly generate all the generated cases/functions using CodeGeneration. Maintaining them manually won’t scale, I think.

kimdv: As I tought. Just needed to confirm

kimdv:
It will introduce a lot of warnings inside the project. 
Not sure if we should try to reduce some of them?

CodeGeneration/Sources/SyntaxSupport/Node.swift
```diff
@@ -106,10 +109,16 @@ public class Node {
...
         experimentalSPI.with(\.trailingTrivia, .newline)
       }
+
       if forRaw {
         "@_spi(RawSyntax)"
       }
+
+      if isDeprecated, !ignoreAvailableAttribute {
+        #"@available(*, deprecated, message: "Will be removed in a future version.")"#
+      }
```
kimdv, line 119: A single place there is a enum that conforms to `CaseItaretable` and then we cannot apply this attribute.

Tests/SwiftParserTest/DeclarationTests.swift
```diff
@@ -2272,7 +2272,7 @@ final class DeclarationTests: ParserTestCase {
         1️⃣<#code#>
       }
       """,
-      substructure: MemberBlockItemSyntax(decl: EditorPlaceholderDeclSyntax(placeholder: .identifier("<#code#>"))),
+      // substructure: MemberBlockItemSyntax(decl: DeclReferenceExprSyntax(baseName: .identifier("<#code#>"))),
```
ahoppen, line 2275: Should this really be a comment?

Tests/SwiftParserTest/translated/TrailingClosuresTests.swift
```diff
@@ -185,14 +185,14 @@ final class TrailingClosuresTests: ParserTestCase {
...
-              item: .init(EditorPlaceholderExprSyntax(placeholder: .identifier("<#T##() -> Void#>")))
+              item: .init(DeclReferenceExprSyntax(baseName: .identifier("<#T##() -> Void#>")))
             )
           ]),
           rightBrace: .rightBraceToken(presence: .missing)
         )
       ),
       diagnostics: [
-        DiagnosticSpec(message: "editor placeholder in source file")
+        DiagnosticSpec(locationMarker: "1️⃣", message: "editor placeholder in source file")
```
ahoppen, line 195: Is this change needed?

ahoppen:
> It will introduce a lot of warnings inside the project.
> Not sure if we should try to reduce some of them?

Sorry, only read this comment now by chance.

swift-syntax should definitely build without warnings after this change. It’s OK for dependent projects to have deprecation warnings but swift-syntax should be warning-free.

kimdv:
Okay @ahoppen I've now updated the parser, so it don't parse the placeholder anymore. 
As we talked about, they will not be marked as deprecated because they will make warnings in the project

ahoppen:
As we discussed offline, could you add a warning doc comment to the editor placeholder nodes? Something along the lines of:
```swift
/// - Warning: Theses `EditorPlaceholderDeclSyntax` node is not generated by the parser anymore. Placeholders are represented by a `MissingDeclSyntax`.
```

kimdv:
> As we discussed offline, could you add a warning doc comment to the editor placeholder nodes? Something along the lines of:
> 
> ```swift
> /// - Warning: Theses `EditorPlaceholderDeclSyntax` node is not generated by the parser anymore. Placeholders are represented by a `MissingDeclSyntax`.
> ```

Something like this?

Release Notes/511.md
```diff
@@ -30,6 +30,10 @@
   - Issue: https://github.com/apple/swift-syntax/issues/2267
   - Pull request: https://github.com/apple/swift-syntax/pull/2272
 
+- `EditorPlaceholderExprSyntax` & `EditorPlaceholderExprSyntax`:
+  - Description: `EditorPlaceholderExprSyntax` & `EditorPlaceholderExprSyntax` are now deprecated and not parsed anymore. They will now be parsed as identifiers.
```
bnbarham, line 34: `Expr` twice, presumably wanted `Decl` as well? Super nitpick, but I'd also just use `and` rather than `&`.

```suggestion
- `EditorPlaceholderDeclSyntax` and `EditorPlaceholderExprSyntax` deprecated:
  - Description: `EditorPlaceholderExprSyntax` and `EditorPlaceholderExprSyntax` are now deprecated and placeholders are instead parsed as identifiers within a `MissingDeclSyntax` or `DeclReferenceExprSyntax`.
```

kimdv: I tell you. This time I didn't want any copy/paste error. But it happened again.. 🤦‍♂️

but it's fixed and nothing is to nitpick

ahoppen (approved):

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test Windows

PR #2238, opened 2023-09-26
Introduce Diagnostic Decorator and Testing Utilities

Matejkob:
This PR is part of the efforts related to #2214.

#### Commit 1: Add `DiagnosticDecorator` Protocol, `ANSIDiagnosticDecorator` and `BacisDiagnosticDecorator` Implementations
- Introduce a new protocol, `DiagnosticDecorator`, to standardize the decoration of diagnostic output.
- Implement `ANSIDiagnosticDecorator`, which provides ANSI colorization and severity-based prefixes for enhanced visibility.

#### Commit 2: Introduce Diagnostic Testing Utilities
- Add `DiagnosticTestingUtils.swift` to facilitate more effective unit testing of diagnostics.

### DIFF
--- a/Sources/SwiftDiagnostics/DiagnosticDecorators/ANSIDiagnosticDecorator.swift
+++ b/Sources/SwiftDiagnostics/DiagnosticDecorators/ANSIDiagnosticDecorator.swift
@@ -1,0 +1,211 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+extension DiagnosticDecorator where Self == ANSIDiagnosticDecorator {
+  /// - SeeAlso: ``ANSIDiagnosticDecorator``
+  static var ANSI: Self {
+    Self()
+  }
+}
+
+/// An implementation of the `DiagnosticDecorator` protocol that enhances various diagnostic elements—including messages,
+/// buffer outlines, and code highlights—by applying severity-based prefixes and ANSI color codes.
+///
+/// This decorator uses ANSI codes—control characters specialized for text formatting in terminals—to provide visual cues.
+@_spi(Testing) public struct ANSIDiagnosticDecorator: DiagnosticDecorator {
+
+  @_spi(Testing) public init() {}
+
+  /// Decorates a diagnostic message by appending a severity-based prefix and applying ANSI color codes.
+  ///
+  /// - Parameters:
+  ///   - message: The diagnostic message that needs to be decorated.
+  ///   - severity: The severity level associated with the diagnostic message.
+  ///
+  /// - Returns: A string that combines the severity-specific prefix and the original diagnostic message, with ANSI colorization.
+  ///
+  /// ## Example
+  ///
+  /// ```swift
+  /// let decorator = ANSIDiagnosticDecorator()
+  /// let decoratedMessage = decorator.decorateMessage("File not found", basedOnSeverity: .error)
+  /// // Output would be: "[1;31merror: [1;39mFile not found[0;0m"
+  /// ```
+  /// In this example, the "error: " prefix is colorized, likely appearing in red, while the message retains its default text color.
+  ///
+  /// For a similar colorized output in the console, you can use `printf` in Bash:
+  /// ```bash
+  /// printf "\e[1;31merror: \e[1;39mFile not found\e[0;0m\n"
+  /// ```
+  @_spi(Testing) public func decorateMessage(_ message: String, basedOnSeverity severity: DiagnosticSeverity) -> String {
+    let severityText: String
+    let severityAnnotation: ANSIAnnotation
+
+    switch severity {
+    case .error:
+      severityText = "error"
+      severityAnnotation = .errorText
+
+    case .warning:
+      severityText = "warning"
+      severityAnnotation = .warningText
+
+    case .note:
+      severityText = "note"
+      severityAnnotation = .noteText
+
+    case .remark:
+      severityText = "remark"
+      severityAnnotation = .remarkText
+    }
+
+    let prefix = colorizeIfNotEmpty("\(severityText): ", usingAnnotation: severityAnnotation, resetAfterApplication: false)
+
+    return prefix + colorizeIfNotEmpty(message, usingAnnotation: .diagnosticText)
+  }
+
+  /// Decorates a source code buffer outline using ANSI cyan color codes.
+  ///
+  /// - Parameter bufferOutline: The string representation of the source code buffer outline.
+  ///
+  /// - Returns: A string featuring ANSI cyan color codes applied to the source code buffer outline.
+  @_spi(Testing) public func decorateBufferOutline(_ bufferOutline: String) -> String {
+    colorizeIfNotEmpty(bufferOutline, usingAnnotation: .bufferOutline)
+  }
+
+  /// Emphasizes a specific text segment within a source code snippet using ANSI color codes.
+  ///
+  /// - Parameter highlight: The text segment within the source code snippet that should be emphasized.
+  ///
+  /// - Returns: A tuple containing:
+  ///   - `highlightedSourceCode`: The underlined version of the original source code snippet.
+  ///   - `additionalHighlightedLine`: Always nil.
+  ///
+  /// ## Example
+  ///
+  /// ```swift
+  /// let decorator = ANSIDiagnosticDecorator()
+  /// let decoratedHighlight = decorator.decorateHighlight("let x = 10")
+  /// // Output would be: ["\u{1B}[4;39mlet x = 10\u{1B}[0;0m"]
+  /// ```
+  ///
+  /// To reproduce a similar colorized output manually in the console, you can use `printf` in Bash:
+  /// ```bash
+  /// printf "\e[4;39mlet x = 10\e[0;0m\n"
+  /// ```
+  @_spi(Testing) public func decorateHighlight(_ highlight: String) -> (highlightedSourceCode: String, additionalHighlightedLine: String?) {
+    (highlightedSourceCode: colorizeIfNotEmpty(highlight, usingAnnotation: .sourceHighlight), additionalHighlightedLine: nil)
+  }
+
+  /// Applies ANSI annotation to a given text segment, if the text is not empty.
+  ///
+  /// - Parameters:
+  ///   - text: The text segment to which the annotation should be applied.
+  ///   - annotation: The ANSI annotation to apply.
+  ///   - resetAfter: A flag indicating whether to reset ANSI settings after applying them. Defaults to true.
+  ///
+  /// - Returns: A potentially colorized version of the input text.
+  private func colorizeIfNotEmpty(
+    _ text: String,
+    usingAnnotation annotation: ANSIAnnotation,
+    resetAfterApplication resetAfter: Bool = true
+  ) -> String {
+    if text.isEmpty {
+      return text
+    } else {
+      return annotation.applied(to: text, resetAfter: resetAfter)
+    }
+  }
+}
+
+/// Defines text attributes to be applied to console output.
+private struct ANSIAnnotation {
+  /// Represents ANSI color codes.
+  enum Color: UInt8 {
+    case normal = 0
+    case black = 30
+    case red = 31
+    case green = 32
+    case yellow = 33
+    case blue = 34
+    case magenta = 35
+    case cyan = 36
+    case white = 37
+    case `default` = 39
+  }
+
+  /// Represents ANSI text traits.
+  enum Trait: UInt8 {
+    case normal = 0
+    case bold = 1
+    case underline = 4
+  }
+
+  /// The ANSI color to be used.
+  let color: Color
+
+  /// The ANSI text trait to be used.
+  let trait: Trait
+
+  /// Returns ANSI code as a string, including both trait and color.
+  var code: String {
+    "\u{001B}[\(trait.rawValue);\(color.rawValue)m"
+  }
+
+  /// Applies the ANSI code to a message string. Optionally resets the code after the message.
+  func applied(to message: String, resetAfter: Bool = true) -> String {
+    guard resetAfter else {
+      return "\(code)\(message)"
+    }
+    return "\(code)\(message)\(ANSIAnnotation.normal.code)"
+  }
+
+  /// The default 'normal' ANSIAnnotation used to reset styles.
+  static var normal: Self {
+    Self(color: .normal, trait: .normal)
+  }
+
+  /// Annotation used for the outline and line numbers of a buffer.
+  static var bufferOutline: Self {
+    Self(color: .cyan, trait: .normal)
+  }
+
+  /// Annotation used for highlighting source text.
+  static var sourceHighlight: Self {
+    Self(color: .default, trait: .underline)
+  }
+
+  /// Annotation used for making text bold, commonly used in diagnostic messages.
+  static var diagnosticText: Self {
+    Self(color: .default, trait: .bold)
+  }
+
+  /// Annotation used for error text.
+  static var errorText: Self {
+    Self(color: .red, trait: .bold)
+  }
+
+  /// Annotation used for warning text.
+  static var warningText: Self {
+    Self(color: .yellow, trait: .bold)
+  }
+
+  /// Annotation used for note text.
+  static var noteText: Self {
+    Self(color: .default, trait: .bold)
+  }
+
+  /// Annotation used for remarks or less critical text.
+  static var remarkText: Self {
+    Self(color: .blue, trait: .bold)
+  }
+}
--- a/Sources/SwiftDiagnostics/DiagnosticDecorators/DiagnosticDecorator.swift
+++ b/Sources/SwiftDiagnostics/DiagnosticDecorators/DiagnosticDecorator.swift
@@ -1,0 +1,74 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// Protocol that defines a standard interface for decorating diagnostic output in source code.
+///
+/// This protocol is intended to be used by entities such as ``DiagnosticsFormatter`` and ``GroupedDiagnostics``
+/// to apply custom decorations to diagnostic messages, buffer outlines, and code highlights.
+///
+/// ## Conforming to `DiagnosticDecorator`:
+///
+/// To conform to the `DiagnosticDecorator` protocol, you must implement three required methods:
+///
+/// 1. `decorateMessage(_:basedOnSeverity:)`: For decorating diagnostic messages.
+/// 2. `decorateBufferOutline(_:)`: For decorating the outlines of source code buffers.
+/// 3. `decorateHighlight(_:)`: For decorating individual highlights within a source code snippet.
+///
+/// ## Customization:
+///
+/// The protocol is designed to be easily customizable. Developers can create their own entities that conform
+/// to `DiagnosticDecorator` to implement custom decorating logic. This allows for different visual representations,
+/// such as using ANSI colors, underscores, emoji-based or other markers, for diagnostics in source code.
+protocol DiagnosticDecorator {
+  /// Decorates a diagnostic message based on its severity level.
+  ///
+  /// Implementations are expected to prepend a severity-specific prefix (e.g., "error: ", "warning: ") to the diagnostic message.
+  ///
+  /// - Parameters:
+  ///   - message: The diagnostic message that needs to be decorated.
+  ///   - severity: The severity level associated with the diagnostic message.
+  ///
+  /// - Returns: A decorated version of the diagnostic message, enhanced by visual cues like color, text styles, or other markers,
+  ///            as well as a severity-specific prefix, based on its severity level.
+  func decorateMessage(_ message: String, basedOnSeverity severity: DiagnosticSeverity) -> String
+
+  /// Decorates the outline of a source code buffer to visually enhance its structure.
+  ///
+  /// - Parameter bufferOutline: The string representation of the source code buffer outline.
+  ///
+  /// - Returns: A decorated version of the buffer outline, improved with visual cues like color, text styles, or other markers.
+  func decorateBufferOutline(_ bufferOutline: String) -> String
+
+  /// Decorates a highlight within a source code snippet to emphasize it.
+  ///
+  /// - Parameter highlight: The text segment within the source code snippet that should be emphasized.
+  ///
+  /// - Returns: A tuple containing:
+  ///   - `highlightedSourceCode`: A string that represents the decorated version of the original source code snippet.
+  ///   - `additionalHighlightedLine`: An optional string containing additional lines of highlighting, if applicable.
+  ///
+  /// - Note: The method returns a tuple to offer more flexibility in decorating highlights.
+  ///         This allows for a variety of techniques to be used, such as ANSI codes for color
+  ///         and additional lines for contextual emphasis, which will be combined during the rendering process.
+  func decorateHighlight(_ highlight: String) -> (highlightedSourceCode: String, additionalHighlightedLine: String?)
+}
+
+extension DiagnosticDecorator {
+  /// Decorates a ``DiagnosticMessage`` instance by delegating to the `decorateMessage(_:basedOnSeverity:)` method.
+  ///
+  /// - Parameter diagnosticMessage: The ``DiagnosticMessage`` instance that encapsulates both the message and its severity level.
+  ///
+  /// - Returns: A decorated version of the diagnostic message, determined by its severity level.
+  func decorateDiagnosticMessage(_ diagnosticMessage: DiagnosticMessage) -> String {
+    decorateMessage(diagnosticMessage.message, basedOnSeverity: diagnosticMessage.severity)
+  }
+}
--- a/Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
+++ b/Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
@@ -73,12 +73,20 @@ public struct DiagnosticsFormatter {
   /// Number of lines which should be printed before and after the diagnostic message
   public let contextSize: Int
 
-  /// Whether to colorize formatted diagnostics.
-  public let colorize: Bool
+  /// An instance that conforms to the ``DiagnosticDecorator`` protocol, responsible for formatting diagnostic messages.
+  ///
+  /// This property allows for the flexible customization of diagnostic messages, buffer outlines, and code highlighting.
+  /// Different implementations can be swapped in to tailor the output according to user preferences or specific environmental conditions.
+  let diagnosticDecorator: DiagnosticDecorator
+
+  @available(*, deprecated, message: "Store the `colorize` property passed to the initializer instead")
+  public var colorize: Bool {
+    return diagnosticDecorator is ANSIDiagnosticDecorator
+  }
 
   public init(contextSize: Int = 2, colorize: Bool = false) {
     self.contextSize = contextSize
-    self.colorize = colorize
+    self.diagnosticDecorator = colorize ? .ANSI : .basic
   }
 
   public static func annotatedSource(
@@ -98,7 +106,7 @@ public struct DiagnosticsFormatter {
     tree: some SyntaxProtocol,
     sourceLocationConverter slc: SourceLocationConverter
   ) -> String {
-    guard colorize, !annotatedLine.diagnostics.isEmpty else {
+    if annotatedLine.diagnostics.isEmpty {
       return annotatedLine.sourceString
     }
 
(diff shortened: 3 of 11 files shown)


Sources/SwiftDiagnostics/DiagnosticDecorators/ANSIDiagnosticDecorator.swift
```diff
@@ -0,0 +1,248 @@
...
+  }
+
+  /// Decorates a text segment intended for highlighting within a source code snippet. The method can also optionally
+  /// apply ANSI colorization for enhanced visual cues.
+  ///
+  /// - Parameter highlight: The text segment to be emphasized.
+  ///
+  /// - Returns: A tuple containing:
+  ///   - `highlightedSourceCode`: A string that represents the decorated version of the original source code snippet.
+  ///   - `additionalHighlightedLine`: always with nil.
```
ahoppen, line 120: Why do we need to return `additionalHighlightedLine` if it’s always `nil`?

Sources/SwiftDiagnostics/DiagnosticDecorators/DiagnosticDecorator.swift
```diff
@@ -0,0 +1,71 @@
...
+// Copyright (c) 2014 - 2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// Protocol that defines a standard interface for decorating diagnostic output in source code.
+/// This protocol is intended to be used by entities such as ``DiagnosticsFormatter`` and ``GroupedDiagnostics``
```
ahoppen, line 14: The abstract and discussion of a docc comment should be separated by a newline

```suggestion
/// Protocol that defines a standard interface for decorating diagnostic output in source code.
///
/// This protocol is intended to be used by entities such as ``DiagnosticsFormatter`` and ``GroupedDiagnostics``
```

Tests/SwiftDiagnosticsTest/DiagnosticTestingUtils.swift
```diff
@@ -0,0 +1,221 @@
...
+import SwiftDiagnostics
+import SwiftParser
+import SwiftParserDiagnostics
+import SwiftSyntax
+import XCTest
+import _SwiftSyntaxTestSupport
+
+/// A typealias representing a location marker.
+///
+/// This string serves to pinpoint the exact location of a particular token in the Abstract Syntax Tree (AST).
```
ahoppen, line 22: Some terminology: I try to use AST for the C++ AST tree that’s used in the compiler, which isn’t source-accurate (that’s the abstract part to me). swift-syntax works on SwiftSyntax trees or syntax trees.

Tests/SwiftDiagnosticsTest/DiagnosticTestingUtils.swift
```diff
@@ -0,0 +1,221 @@
...
+///
+/// ```swift
+/// func foo() -> Int {
+///   if 1️⃣1 != 0 2️⃣{
+///     return 0
+///   }
+///   return 1
+/// }
+/// ```
+typealias LocationMarker = String
```
ahoppen, line 40: What do you think about the following? That way you can really specify which node you want to refer to at a specific position and you could also use `NodeLocator` to specify the highlighted nodes.

```swift
/// Refers to the first node in a syntax tree after `marker` that satisfies `match`.
///
/// For example, if you have
/// ```swift
/// var x: Int
/// 1️⃣func foo() {}
/// ```
///
/// Then `NodeLocator(marker: "1️⃣")` refers to the `CodeBlockItemSyntax` that
/// contains the function declaration and
/// `NodeLocator(marker: "1️⃣", matches: { $0.is(FunctionDeclSyntax.self) })`
/// refers to the function within.
struct NodeLocator {
  let marker: String
  let match: (Syntax) -> Bool = { _ in true }

  func findNode(in tree: Syntax) -> Syntax? {
    // TODO: Implement
  }
}
```

Tests/SwiftDiagnosticsTest/DiagnosticTestingUtils.swift
```diff
@@ -0,0 +1,221 @@
...
+  ///   - locationMarker: The marker pointing to location in source code.
+  ///   - id: The message ID of the diagnostic.
+  ///   - message: The textual message to display for the diagnostic.
+  ///   - severity: The severity level of the diagnostic. Default is `.error`.
+  ///   - highlight: The syntax elements to be highlighted. Default is an empty array.
+  ///   - noteDescriptors: An array of note descriptors for additional context. Default is an empty array.
+  ///   - fixIts: An array of Fix-It descriptors for quick fixes. Default is an empty array.
+  init(
+    locationMarker: LocationMarker,
+    id: MessageID,
```
ahoppen, line 77: How about defaulting this to `.test`? We don’t really care about the ID in these tests.

ahoppen (commented):
Thanks for splitting the PR. It makes the review a lot easier. And also thanks for the long comments, they also really help understand what’s going on.

I left a couple comments inline. On a high level, I’ve got a couple of questions:
- What do you imagine the use case of implementing a `DiagnosticDecorator` to be? AFAICT `ANSIDiagnosticGenerator` is the only one that we currently have and that you’ll introduce in #2214 and I would prefer to not have the `DiagnosticDecorator` protocol abstraction unless we actually use it.
- I would prefer to not make `DiagnosticDecorator`, `ANSIDiagnosticDecorator` public. If we do, it becomes part of the public API surface and it’s a lot harder to change in the future if we don’t want to break it.
- AFAICT nothing is using the `DiagnosticDecorator` yet. I understand that you will be using it in #2214 but I think it would be good to start using it in the current implementation already unless it’s a huge amount of work. I understand that it’s more work that might seem unnecessary but it would really split #2214 into two self-contained pieces that are easier to review.

Matejkob:
Thank you for taking the time to review my PR @ahoppen ; your feedback is invaluable, and I really appreciate the time you've invested.

Addressing your concerns:

> * What do you imagine the use case of implementing a `DiagnosticDecorator` to be? AFAICT `ANSIDiagnosticGenerator` is the only one that we currently have and that you’ll introduce in [Enhancements to diagnostic formatters and testing utilities #2214](https://github.com/apple/swift-syntax/pull/2214) and I would prefer to not have the `DiagnosticDecorator` protocol abstraction unless we actually use it.

The overarching vision for the `DiagnosticDecorator`, as well as other key components like `DiagnosticsFormatter` and `GroupedDiagnostics` in the `SwiftDiagnostics` package, which I have in mind, is to serve as highly customizable and modular tools. The long-term goal would be to empower end-users either by allowing them to build upon these components or by using them directly. To give you a concrete example, @pointfreeco [swift-macro-testing](https://github.com/pointfreeco/swift-macro-testing) library has already adopted the `DiagnosticsFormatter`. Unfortunately, they had to copy-paste the code into their project and manually adjust its implementation, even though their project depends on swift-syntax. This is hardly an ideal scenario and also illustrates the need for a more modular approach.

But it's not just about third-party implementations; we also have potential use-cases within our own ecosystem. For instance, consider the area of macro diagnostics testing. Currently, we have an assert method for macros, which, while functional, lacks advanced diagnostic capabilities. I believe there's a real opportunity here to provide the developer community with a simple yet effective tool for visualizing and debugging macro diagnostics. As we know, debugging macros can be challenging due to the compiler expansions involved, and this feature could significantly alleviate that pain.

Lastly, let's talk about the immediate benefits. We are already planning enhancements to `DiagnosticsFormatter` by adding missing diagnostic elements, which will require robust unit testing. Currently, those unit tests containing diagnostic's highlights are hard to read and maintain, especially when they include complicated diagnostic output expectations. Take this example:

```swift
let expectedOutput = """
  \u{001B}[0;36m1 │\u{001B}[0;0m for \u{001B}[4;39m(i\u{001B}[0;0m \u{001B}[4;39m= 🐮; i != 👩‍👩‍👦‍👦; i += 1)\u{001B}[0;0m { }
    \u{001B}[0;36m│\u{001B}[0;0m │      ╰─ \u{001B}[1;31merror: \u{001B}[1;39mexpected ')' to end tuple pattern\u{001B}[0;0m
    \u{001B}[0;36m│\u{001B}[0;0m ╰─ \u{001B}[1;31merror: \u{001B}[1;39mC-style for statement has been removed in Swift 3\u{001B}[0;0m
  """
```

With the decorator abstraction in place, we can create more readable and developer-friendly unit tests, as shown below:

```swift
let expectedOutput = """
  1 │ for (i = 🐮; i != 👩‍👩‍👦‍👦; i += 1) { }
    │     ~~ ~~~~~~~~~~~~~~~~~~~~~~~
    │ │      ╰─ error: expected ')' to end tuple pattern
    │ ╰─ error: C-style for statement has been removed in Swift 3

  """
```

> * I would prefer to not make `DiagnosticDecorator`, `ANSIDiagnosticDecorator` public. If we do, it becomes part of the public API surface and it’s a lot harder to change in the future if we don’t want to break it.

You bring up a valid concern about making `DiagnosticDecorator` and `ANSIDiagnosticDecorator` public, given that it would lock us into maintaining those APIs and limit our ability to make future changes. I completely agree with you on this. In fact, with the understanding that these APIs are likely to evolve, I've already marked `DiagnosticDecorator` and `ANSIDiagnosticDecorator` with `@_spi(Testing)`. This should help us maintain the flexibility to adapt and refine these components as we continue to build upon them.

> * AFAICT nothing is using the `DiagnosticDecorator` yet. I understand that you will be using it in [Enhancements to diagnostic formatters and testing utilities #2214](https://github.com/apple/swift-syntax/pull/2214) but I think it would be good to start using it in the current implementation already unless it’s a huge amount of work. I understand that it’s more work that might seem unnecessary but it would really split [Enhancements to diagnostic formatters and testing utilities #2214](https://github.com/apple/swift-syntax/pull/2214) into two self-contained pieces that are easier to review.

Great point about immediate use. I've already integrated `DiagnosticDecorator` into `DiagnosticsFormatter` to make this PR more self-contained and easier to review.

> [!note]
> **Disclaimer**:  I want to clarify that my proposed changes stem from research, requirement gathering, and learning about the existing landscape—rather than imposing my own singular vision. I recognize the challenges in fully capturing the current and future needs for diagnostic formatters. The PRs are meant as a collaborative stepping stone, combining what I've learned, observed in other projects, and a few of my own ideas. My aim is for us to build a unified vision together, using the code I've already written as a foundation.

Sources/SwiftDiagnostics/DiagnosticDecorators/ANSIDiagnosticDecorator.swift
Matejkob: It is `nil` in this implementation because the ANSI codes are able to underline lines directly. The assumption is that in future implementations it will contain characters like `~~~` or `---`.

Tests/SwiftDiagnosticsTest/DiagnosticTestingUtils.swift
Matejkob: I have a feeling this is going to be a bigger discussion. How about we move it out of the scope of this PR for now and discuss it in the issue: [Introduce Efficient Node Localization in Syntax Tree for Unit Tests](https://github.com/apple/swift-syntax/issues/2251)?

ahoppen: Yes, sure!

Sources/SwiftDiagnostics/DiagnosticDecorators/ANSIDiagnosticDecorator.swift
ahoppen: I would have a slight preference to only add `additionalHighlightedLine` once we actually add a diagnostic annotator that uses it, but I’m fine with keeping it as well.

Sources/SwiftDiagnostics/DiagnosticDecorators/ANSIDiagnosticDecorator.swift
```diff
@@ -0,0 +1,248 @@
...
+/// or another type of diagnostic output.
+///
+/// In essence, this decorator serves two primary functions:
+/// 1. It appends a categorization prefix (e.g., "error: ", "warning: ", "note: ", "remark: ") to the diagnostic message
+/// to indicate its level of urgency or importance.
+/// 2. It has the capability to colorize this prefix and other segments of the message for enhanced visibility
+/// and quick interpretation. This colorization feature can be toggled on or off, offering flexibility in its usage.
+@_spi(Testing) public struct ANSIDiagnosticDecorator: DiagnosticDecorator {
+  /// A flag indicating whether or not to apply colorization.
+  private let colorize: Bool
```
ahoppen, line 34: Instead of having a `colorize` flag here, shouldn’t we just have two decorators? One that does add color and one that does not?

Sources/SwiftDiagnostics/DiagnosticDecorators/DiagnosticDecorator.swift
```diff
@@ -0,0 +1,72 @@
...
+/// such as using ANSI colors, underscores, emoji-based or other markers, for diagnostics in source code.
+@_spi(Testing) public protocol DiagnosticDecorator {
+  /// Decorates a diagnostic message based on its severity level.
+  ///
+  /// - Parameters:
+  ///   - message: The diagnostic message that needs to be decorated.
+  ///   - severity: The severity level associated with the diagnostic message.
+  ///
+  /// - Returns: A decorated version of the diagnostic message, enhanced by visual cues like color, text styles, or other markers
+  ///            based on its severity level.
```
ahoppen, line 39: I think this doc comment should also contain that `decorateMessage` should add any prefix like `error:` or `warning:`. It’s not obvious from the protocol definition.

Sources/SwiftDiagnostics/DiagnosticDecorators/DiagnosticDecorator.swift
```diff
@@ -0,0 +1,72 @@
...
+/// 1. `decorateMessage(_:basedOnSeverity:)`: For decorating diagnostic messages.
+/// 2. `decorateBufferOutline(_:)`: For decorating the outlines of source code buffers.
+/// 3. `decorateHighlight(_:)`: For decorating individual highlights within a source code snippet.
+///
+/// ## Customization:
+///
+/// The protocol is designed to be easily customizable. Developers can create their own entities that conform
+/// to `DiagnosticDecorator` to implement custom decorating logic. This allows for different visual representations,
+/// such as using ANSI colors, underscores, emoji-based or other markers, for diagnostics in source code.
+@_spi(Testing) public protocol DiagnosticDecorator {
```
ahoppen, line 31: I think we don’t need to make `DiagnosticDecorator` `@_spi(Testing) public` yet, we can just keep it internal (`ANSIDiagnosticDecorator` needs to say `@_spi(Testing)`).

And just to clarify: I think that making `DiagnosticDecorator` might be a very good idea in the future. I would just like to get some experience with the API by implementing a few more diagnostic decorators internally before deciding that we are happy enough with the API to make it public.

Sources/SwiftDiagnostics/GroupedDiagnostics.swift
```diff
@@ -172,7 +172,7 @@ extension GroupedDiagnostics {
       tree: sourceFile.tree
     )
 
-    let colorizeBufferOutline = formatter.colorizeBufferOutline
+    let decorateBufferOutline = formatter.diagnosticDecorator.decorateBufferOutline
```
ahoppen, line 175: I would prefer to call `formatter.diagnosticDecorator.decorateBufferOutline` below multiple times instead of storing the function as a closure. If you do want to abbreviate it, I would be fine with storing the diagnostic decorator in a temporary variable. 

Storing functions in variables just makes it a lot harder to understand what’s going on IMO because you can’t do Quick Help on the call and jump to definition requires two jumps instead of one.

```swift
let decorator = formatter.diagnosticDecorator
…
… = decorator.decorateBufferOutline(…)
```

Tests/SwiftDiagnosticsTest/DiagnosticTestingUtils.swift
```diff
@@ -0,0 +1,221 @@
...
+
+  /// Creates a ``Diagnostic`` instance from a given ``DiagnosticDescriptor``, syntax tree, and location markers.
+  ///
+  /// - Parameters:
+  ///   - tree: The syntax tree where the diagnostic is rooted.
+  ///   - markers: A dictionary mapping location markers to their respective offsets in the source code.
+  ///   - file: The file where the test is located, used for reporting failures.
+  ///   - line: The line where the test is located, used for reporting failures.
+  ///
+  /// - Returns: A ``Diagnostic`` instance populated with details from the ``DiagnosticDescriptor``, or `nil` if it fails.
```
ahoppen, line 101: Instead of silently failing and not generating a `Diagnostic` for a `DiagnosticDescriptor` if it fails, what do you think about throwing an error instead? I think that should make it a lot clearer what’s going wrong if you write eg. a `DiagnosticDescriptor` with a marker that doesn’t exist in the source code.

Tests/SwiftDiagnosticsTest/DiagnosticTestingUtils.swift
```diff
@@ -0,0 +1,221 @@
...
+    var notes = [Note]()
+    for noteDescriptor in self.noteDescriptors {
+      guard let noteNode = node(at: noteDescriptor.locationMarker) else { continue }
+
+      let note = Note(
+        node: noteNode,
+        message: SimpleNoteMessage(message: noteDescriptor.message, fixItID: noteDescriptor.id)
+      )
+      notes.append(note)
+    }
```
ahoppen, line 132: I think using `map` here might make it a little bit more swifty + easier to understand, especially if `node(at:)` throws on failure.

Tests/SwiftDiagnosticsTest/DiagnosticTestingUtils.swift
```diff
@@ -0,0 +1,221 @@
...
+
+  let (markers, source) = extractMarkers(markedSource)
+  let tree = Parser.parse(source: source)
+
+  // Check for any unexpected diagnostics generated by parser.
+  let generatedDiagnostics = ParseDiagnosticsGenerator.diagnostics(for: tree)
+  if !generatedDiagnostics.isEmpty {
+    let diagnosticMessages = generatedDiagnostics.map { $0.message }.joined(separator: ", ")
+    XCTFail("Unexpected diagnostics found in the source: \(diagnosticMessages)", file: file, line: line)
+  }
```
ahoppen, line 207: Why do we need to check this? I think it would be totally fine to add our own diagnostics based on the `DiagnosticDescriptor`s even if the syntax tree contains errors.

Tests/SwiftDiagnosticsTest/GroupDiagnosticsFormatterTests.swift
```diff
@@ -35,34 +25,22 @@ extension GroupedDiagnostics {
...
-        node: node,
-        message: SimpleDiagnosticMessage(
-          message: message,
-          diagnosticID: MessageID(domain: "test", id: "conjured"),
-          severity: severity
-        )
-      )
-      diagnostics.append(diag)
+
+    let parserDiagnostics = ParseDiagnosticsGenerator.diagnostics(for: tree)
```
ahoppen, line 33: Regarding my comment above about why we need to check that the tree didn’t contain any errors, I think if we removed this line as well and only displayed errors based on `diagnosticDescriptors`, we would be completely independent of `SwiftParserDiagnostics` here.

ahoppen (commented):
Thanks for providing the detailed response. I think the overall direction is very good. I would just like to keep `DiagnosticDecorator` internal until we have gained some more experience with it. I think all the other comments are fairly minor.

Tests/SwiftDiagnosticsTest/DiagnosticTestingUtils.swift
Matejkob: I believe that implementing the `assert` function in this way would indeed make the code clearer, but I'm not entirely convinced that the end result would be more transparent to the users of this high-level API. 

My primary motivation behind this implementation approach was to avoid interrupting the assertion process unnecessarily. The `assertAnnotated` function takes a collection of `diagnosticDescriptors` as an argument. I chose to return `nil` while also invoking `XCTFail` so that we could iterate over all elements, inform the user only about the missing ones, and apply the existing ones to provide as much assertion validation as possible. For instance, if only one location marker out of four is missing, the user will end up with the following result:

Screenshot 2023-10-11 at 7 18 50 PM

I believe this approach benefits end clients. If I'm not mistaken, it wouldn't work in the same way if we chose to throw an error, would it?

Tests/SwiftDiagnosticsTest/DiagnosticTestingUtils.swift
Matejkob: My idea was to avoid stoping the assertion process if just one of the nodes couldn't be found. That's why I've used a `for` loop instead of `map`, as I'm employing `continue` to skip to the next iteration. However, I believe we could also use `compactMap` as an alternative.

Tests/SwiftDiagnosticsTest/DiagnosticTestingUtils.swift
Matejkob: Sounds good to me!

Tests/SwiftDiagnosticsTest/GroupDiagnosticsFormatterTests.swift
Matejkob: I love the idea, but I'd prefer to hold off on that for this PR. `GroupedDiagnostics` has its own set of issues, and if we were to remove this part now, we'd likely find ourselves fixing additional problems. After I've improved `DiagnosticsFormatter`, I plan on revisiting `GroupedDiagnostics`. I'll definitely tackle it at that point.

Tests/SwiftDiagnosticsTest/DiagnosticTestingUtils.swift
Matejkob: Removed

Sources/SwiftDiagnostics/GroupedDiagnostics.swift
Matejkob: I've added `let diagnosticDecorator = formatter.diagnosticDecorator` instead of storing func inn closure.

Sources/SwiftDiagnostics/DiagnosticDecorators/DiagnosticDecorator.swift
Matejkob: Done

Sources/SwiftDiagnostics/DiagnosticDecorators/DiagnosticDecorator.swift
Matejkob: Is now better? I've added more information to the documentation.

Sources/SwiftDiagnostics/DiagnosticDecorators/ANSIDiagnosticDecorator.swift
Matejkob: Great idea! Thanks! Implemented!

Matejkob: 🫶

Sources/SwiftDiagnostics/DiagnosticDecorators/ANSIDiagnosticDecorator.swift
Matejkob: In the [Enhance diagnostic formatter to include trivia between highlight nodes](#2242) thread, you mentioned the idea of changing `assertSyntaxHighlightRanges` to actually underline the highlighted ranges, either via `DiagnosticDecorator` or a custom function. I think that's a really neat idea. Given that we plan to use `DiagnosticDecorator` in the next PR, I'd opt for keeping it as is for now.

Sources/SwiftDiagnostics/DiagnosticDecorators/DiagnosticDecorator.swift
ahoppen: Sounds good 👍🏽

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
```diff
@@ -73,12 +73,15 @@ public struct DiagnosticsFormatter {
   /// Number of lines which should be printed before and after the diagnostic message
   public let contextSize: Int
 
-  /// Whether to colorize formatted diagnostics.
-  public let colorize: Bool
```
ahoppen, line 77: The removal of the `colorize` property is an API-breaking change. 

To maintain source compatibility, could you add

```swift
@available(*, deprecated, message: "Store the `colorize` property passed to the initializer instead")
public var  colorize: Bool {
  return diagnosticDecorator is ANSIDiagnosticDecorator
}

File: Tests/SwiftDiagnosticsTest/DiagnosticTestingUtils.swift (hunk 10 above)
Comment (ahoppen) [reply Tests/SwiftDiagnosticsTest/DiagnosticTestingUtils.swift:101]: Oh, I missed the fact that we always emitted an `XCTFail` if we return `nil`.

And I think me missing that is a reason why I would prefer to `throw`. My thinking is that if you are missing one of the diagnostic markers in your test input, there’s something wrong with the test case itself (not just a runtime failure) and thus early exiting is quite alright.

As it stands right now, every time I would read this code, I would wonder the same thing that I originally commented: 
- Are we silently dropping one of our test case assertions here? That seems dangerous
- Looks at implementation/documentation: Oh, no, it’s fine we emit an `XCTFail` in that case.
- Now, what was I doing originally? Needs to restore mental context after side-stepping.

File: Tests/SwiftDiagnosticsTest/GroupDiagnosticsFormatterTests.swift (hunk 13 above)
Comment (ahoppen) [reply Tests/SwiftDiagnosticsTest/GroupDiagnosticsFormatterTests.swift:33]: OK sounds good to me.

File: Tests/SwiftDiagnosticsTest/DiagnosticTestingUtils.swift (hunk 10 above)
Comment (Matejkob) [reply Tests/SwiftDiagnosticsTest/DiagnosticTestingUtils.swift:101]: You make some solid points. You got me—I'm convinced.

I'll swap out the `nil` and `XCTFail` combo for a good ol' error throw. Thanks for the nudge!

File: Sources/SwiftDiagnostics/DiagnosticDecorators/DiagnosticDecorator.swift (hunk 8 above)
Comment (Matejkob) [reply Sources/SwiftDiagnostics/DiagnosticDecorators/DiagnosticDecorator.swift]: I've modified the implementation to enable error throwing. Let me know what you think.

File: Sources/SwiftDiagnostics/DiagnosticsFormatter.swift (hunk 14 above)
Comment (Matejkob) [reply Sources/SwiftDiagnostics/DiagnosticsFormatter.swift:77]: Done

File: Tests/SwiftDiagnosticsTest/DiagnosticTestingUtils.swift [hunk 15]
```diff
@@ -0,0 +1,224 @@
...
+
+    let diagnosticNode = try node(at: self.locationMarker)
+
+    let notes = try self.noteDescriptors.map { noteDescriptor in
+      let noteNode = try node(at: noteDescriptor.locationMarker)
+
+      return Note(
+        node: noteNode,
+        message: SimpleNoteMessage(message: noteDescriptor.message, fixItID: noteDescriptor.id)
+      )
```
ahoppen, line 136: Nit: `noteNode` is a weird name and we don’t actually need it.

```suggestion
      return Note(
        node: try node(at: noteDescriptor.locationMarker),
        message: SimpleNoteMessage(message: noteDescriptor.message, fixItID: noteDescriptor.id)
      )
```

ahoppen (commented):
One really small nitpick, otherwise looks good to me 👍🏽

Tests/SwiftDiagnosticsTest/DiagnosticTestingUtils.swift
Matejkob: Great idea! Done

Tests/SwiftDiagnosticsTest/DiagnosticTestingUtils.swift
Matejkob: Changed to `map` – done.

ahoppen:
@swift-ci Please test

Matejkob:
CI failed with the following message:

> DiagnosticsFormatter.swift:80:28: error: cannot find type 'DiagnosticDecorator' in scope

From a Swift perspective, this doesn't seem to be failing, since it's in the same module and the type is internal. Should I modify `CMakeLists.txt`? Is there any documentation where I can learn more about what's going on here?

bnbarham:
> Should I modify CMakeLists.txt?

Yeah, you'll need to add all the new files to cmake as well.

> Is there any documentation where I can learn more about what's going on here?

Not really. The compiler uses cmake rather than SPM to build swift-syntax (for the libraries included in the toolchain), so any new files have to be added to the relevant CMakeLists.txt.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

ahoppen:
```
Test Case '-[SwiftDiagnosticsTest.GroupedDiagnosticsFormatterTests testGroupingForMacroExpansion]' failed (0.487 seconds).
```
@Matejkob Could you look at the failure?

Matejkob:
Oops! I mistakenly committed a modified test and disabled the `GroupedDiagnosticsFormatterTests` test case in Xcode. Consequently, when I ran all the tests, it didn't fail. My apologies for the oversight. 🙈

ahoppen:
No problem. Let’s try again.

@swift-ci Please test

Matejkob:
Is this the main problem `/Users/ec2-user/jenkins/workspace/swift-syntax-PR-macOS/branch-main/swift/lib/LLVMPasses/ARCEntryPointBuilder.h:21:10: fatal error: 'llvm/TargetParser/Triple.h' file not found
`?

ahoppen:
Seems like an unrelated issue. Trying again.

@swift-ci Please test macOS

ahoppen:
@swift-ci Please test Windows

PR #2241, opened 2023-09-27
Remove the self parse performance tests

ahoppen:
These weren’t run anywhere and they don’t give menaingful results anyway since the source code that is being parsed changes as swift-syntax changes.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #2242, opened 2023-09-27
Enhance diagnostic formatter to include trivia between highlight nodes

### LINKED ISSUE #2208: Diagnostic formatter should highlight trivia in between consecutive highlight nodes
Currently, if a diagnostic has two nodes to highlight, it will highlight them individually, leaving the trivia in between them non-highlighted. The easiest way to see this is by running

```bash
swift-parser-cli print-diags -s 'for x = 1; x < 10; x++ {}'
```

Screenshot 2023-09-16 at 09 14 02

Note that the space between `x` and `=` should also be highlighted.

Matejkob:
- Separate logic for calculating highlight ranges from `DiagnosticsFormatter` into `SyntaxHighlightRangeCalculator`.
- Introduce an `ExtendedHighlight` private struct to manage trivia settings for each highlight.
- Extend logic to consider leading and trailing trivia when highlighting consecutive nodes.
- The `computeHighlightRanges` method now takes into account special cases for trivia between consecutive highlight nodes.
- Add unit tests to cover all paths in `SyntaxHighlightRangeCalculator`.
- Add documentation about trivia handling, explaining the conditions under which trivia should be included or excluded.

Screenshot 2023-09-27 at 4 25 27 PM

Resolves #2208

### DIFF
--- a/Sources/SwiftDiagnostics/SyntaxHighlightRangeCalculator.swift
+++ b/Sources/SwiftDiagnostics/SyntaxHighlightRangeCalculator.swift
@@ -1,0 +1,225 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
+/// Serves as a namespace for methods related to computing
+/// highlight ranges within syntax trees, considering trivia and other syntax nuances.
+@_spi(Testing) public enum SyntaxHighlightRangeCalculator {
+  /// Captures details about a specific highlight region within a line of code.
+  ///
+  /// This struct is used to manage how leading and trailing trivia are treated for each highlight.
+  /// It stores the actual syntax highlight and two Boolean flags for trivia settings.
+  ///
+  /// - Important:
+  ///   - `afterLeadingTrivia`: When `true`, the starting column will be computed after considering leading trivia.
+  ///   - `afterTrailingTrivia`: When `false`, the ending column will be computed before considering trailing trivia.
+  private struct ExtendedHighlight {
+    /// The syntax element to be highlighted.
+    let highlight: Syntax
+    /// Flag to indicate whether leading trivia should be included in the highlight range.
+    var afterLeadingTrivia = true
+    /// Flag to indicate whether trailing trivia should be included in the highlight range.
+    var afterTrailingTrivia = false
+  }
+
+  /// Computes the ranges of highlightable regions on a specific line of source code.
+  ///
+  /// This method computes the start and end columns for each highlight in the line,
+  /// taking into account whether the leading or trailing trivia should be included for each highlight.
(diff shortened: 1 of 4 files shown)


Sources/SwiftDiagnostics/SyntaxHighlightRangeCalculator.swift
```diff
@@ -0,0 +1,225 @@
...
+    var extendedHighlights = annotatedSourceLine.highlights.map { ExtendedHighlight(highlight: $0) }
+
+    // Loop through each extended highlight to fine-tune its trivia settings.
+    for extendedHighlightsIndex in extendedHighlights.indices {
+      adjustTriviaSettingsForHighlight(
+        atIndex: extendedHighlightsIndex,
+        in: &extendedHighlights,
+        usingSourceLocationConverter: sourceLocationConverter
+      )
+    }
```
ahoppen, line 70: I am not a fan of creating the `extendedHighlights` array and then fixing it up later. What I would do instead is:
Iterate over the highlighted nodes. If the node’s `positionBeforeTrailingTrivia` is equal to the last node’s `positionAfterTrailingTrivia`, extend the last highlight range in the array. If not, add a new entry to it.

Sources/SwiftDiagnostics/SyntaxHighlightRangeCalculator.swift
```diff
@@ -0,0 +1,225 @@
...
+  /// - Note:
+  ///   1. If the highlight is the first in the line, its leading trivia is always excluded.
+  ///   2. If a highlight's end aligns with the start of the next highlight, their shared trivia settings are adjusted.
+  ///   3. The last highlight in the line never includes trailing trivia.
+  @_spi(Testing) public static func computeHighlightRanges(
+    forLine annotatedSourceLine: (highlights: [Syntax], sourceString: String),
+    at lineNumber: Int,
+    fromTree tree: some SyntaxProtocol,
+    usingSourceLocationConverter sourceLocationConverter: SourceLocationConverter
+  ) -> [Range<Int>] {
```
ahoppen, line 59: I’m not really a fan of how this takes a line number and returns an array of `Int`s because it’s ambiguous whether these integers are offsets within the source string or the line and whether we are counting in UTF-8 (`str.utf8.count`), UTF-16 (`str.utf16.count`) or grapheme clusters (`str.count`).

I would prefer if this method didn’t take any line number and instead returned `[Range<AbsolutePosition>]`. IMO that’s the most generic way to write this function and converting the positions to `[Range<SourceLocation>]` and then filtering out the highlights that aren’t relevant for the current line.

Or am I missing something?

Tests/SwiftDiagnosticsTest/SyntaxHighlightRangeCalculatorTests.swift
```diff
@@ -0,0 +1,297 @@
...
+      line: line
+    )
+  }
+}
+
+final class SyntaxHighlightRangeCalculatorTests: XCTestCase {
+  func testComputeHighlightRangesFirstHighlightExcludesLeadingTrivia() {
+    let decl: DeclSyntax =
+      "@PropertyWrapper private(set) var variable: Decimal = 123 * 456"
+    // 123456789012345678901234567890
```
ahoppen, line 79: Instead of adding this comment, what do you think about changing `assertSyntaxHighlightRanges` to actually underline the highlighted ranges? Either using a `DiagnosticDecorator` or using a hand-rolled function? I just think that it would make reading the tests a bit easier.

PR #2245, opened 2023-09-28
Fix a minor threading issue and TODO by switching to structured concurrency

benpious:
In #2233, it was suggested that a better long term direction was to adopt Structured Concurrency. This PR does exactly that. 

The issue this pr is fixing is that the `errors` var had no synchronization and could cause crashes if it was written to.

### DIFF
--- a/CodeGeneration/Sources/generate-swift-syntax/GenerateSwiftSyntax.swift
+++ b/CodeGeneration/Sources/generate-swift-syntax/GenerateSwiftSyntax.swift
@@ -68,7 +68,7 @@ struct TemplateSpec {
 }
 
 @main
-struct GenerateSwiftSyntax: ParsableCommand {
+struct GenerateSwiftSyntax: AsyncParsableCommand {
   @Argument(help: "The path to the source directory (i.e. 'swift-syntax/Sources') where the source files are to be generated")
   var destination: String = {
     let sourcesURL = URL(fileURLWithPath: #filePath)
@@ -83,7 +83,7 @@ struct GenerateSwiftSyntax: ParsableCommand {
   @Flag(help: "Enable verbose output")
   var verbose: Bool = false
 
-  func run() throws {
+  func run() async throws {
     let destination = URL(fileURLWithPath: self.destination).standardizedFileURL
 
     var fileSpecs: [GeneratedFileSpec] = [
@@ -139,7 +139,6 @@ struct GenerateSwiftSyntax: ParsableCommand {
 
     let modules = Set(fileSpecs.compactMap { $0.pathComponents.first })
 
-    let previouslyGeneratedFilesLock = NSLock()
     var previouslyGeneratedFiles = Set(
       modules.flatMap { (module) -> [URL] in
         let generatedDir =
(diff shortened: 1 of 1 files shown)


bnbarham:
@swift-ci please test

bnbarham:
Thanks @benpious! Out of interest, did you see any increase in generation time with this change?

benpious:
> Out of interest, did you see any increase in generation time with this change?

I didn't notice any when I originally wrote it, and in my tests there doesn't seem to be any difference. I ran `time swift run --package-path CodeGeneration generate-swift-syntax Sources` and subtracted the reported build time from the total; it seems to be consistently between 118 and 135 seconds on my computer regardless of whether I'm on main or this branch.

CodeGeneration/Sources/generate-swift-syntax/GenerateSwiftSyntax.swift
```diff
@@ -154,32 +153,37 @@ struct GenerateSwiftSyntax: ParsableCommand {
...
     )
 
-    var errors: [Error] = []
-    DispatchQueue.concurrentPerform(iterations: fileSpecs.count) { index in
-      let fileSpec = fileSpecs[index]
-      do {
-        var destination = destination
-        for component in fileSpec.pathComponents {
-          destination = destination.appendingPathComponent(component)
+    await withTaskGroup(of: (URL, Error?).self) { group in
```
ahoppen, line 156: I would add labels to the tuple here. That way you don’t need to do the pretty magical and ambiguous `.0` and `.1` below.

```suggestion
    await withTaskGroup(of: (url: URL, error: Error?).self) { group in
```

CodeGeneration/Sources/generate-swift-syntax/GenerateSwiftSyntax.swift
```diff
@@ -189,24 +193,25 @@ struct GenerateSwiftSyntax: ParsableCommand {
...
-    let start = Date()
-    try contents().write(to: destination, atomically: true, encoding: .utf8)
-    if verbose {
-      print("Generated \(destination.path) in \((Date().timeIntervalSince(start) * 1000).rounded() / 1000)s")
-    }
+fileprivate func generateFile(
+  contents: @autoclosure () -> String,
+  destination: URL,
+  verbose: Bool
+) throws {
```
ahoppen, line 202: I’m happy with making this a top-level function but just FYI, you could keep this a member of `GenerateSwiftSyntax` if you mark the function as `nonisolated`.

ahoppen (commented):
Very nice. 🤩  I only have one minor comment to make the code slightly more readable, otherwise looks very good to me.

CodeGeneration/Sources/generate-swift-syntax/GenerateSwiftSyntax.swift
bnbarham: It doesn't actually use any members, so IMO a top level fileprivate function makes sense anyway.

bnbarham:
> it seems to be consistently between 118 and 135 seconds on my computer regardless of whether I'm on main or this branch

Thanks for checking!

CodeGeneration/Sources/generate-swift-syntax/GenerateSwiftSyntax.swift
benpious: It turns out you don't have to even mark it as `nonisolated` in the current setup of `run`, it still works. 

I originally made this change because when I tried to mark `run` as async without changing the conformance to the async version of `ParseableCommand`, the compiler error kind of implied this was the problem. That was not the case. 

So I don't have a strong opinion on this, but I'm partial to the top level free function.

CodeGeneration/Sources/generate-swift-syntax/GenerateSwiftSyntax.swift
benpious: Made this change locally and pushed. It's definitely more readable now.

bnbarham (approved):
Could you squash your commits and add a commit message, see https://github.com/apple/swift-syntax/blob/main/CONTRIBUTING.md#authoring-commits? I'll start testing after that so we don't have to run it twice.

ahoppen (approved):
Thank you for the PR. Looks very good to me and great to see Swift Concurrency making its way into swift-syntax 👍🏽

benpious:
> Thanks for the updates! LGTM, could you squash your commits and add a commit message, see https://github.com/apple/swift-syntax/blob/main/CONTRIBUTING.md#authoring-commits? I'll start testing after that so we don't have to run it twice.

Done!

bnbarham:
> Done!

FWIW it's generally good practice to have a max 72 length title (to avoid truncation with the rest added to description). No need to fix for this commit, I've started testing. Thanks again!

ahoppen:
@swift-ci Please test macOS

PR #2247, opened 2023-09-29
Unit testing of `AttributeRemover`

gohanlon:
Following discussions in #2215, this PR isolates `AttributeRemoverTests` to `AttributeRemover` (`SwiftSyntaxMacroExpansion` module).

**Test Changes**:
- Added a `fileprivate` assert helper: `assertSyntaxRemovingTestAttributes`.

**Core Changes**:
- Rather than depending on `SyntaxNode`'s id-only `Equatable` conformance, refactor `AttributeRemover` to remove attributes matching a predicate: `AttributeRemover.init(removingWhere:)` of type `(AttributeSyntax) -> Bool)`. This change allows `AttributeRemover` to be used without first plucking the `AttributeSyntax` values to be removed from the live tree.
- Marks the `AttributeRemover` class with `@_spi(Testing)` to make the class visible from the test module without including `AttributeRemover` in the public API.

### DIFF
--- a/Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
+++ b/Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
@@ -384,19 +384,24 @@ struct MacroSystem {
 // MARK: - MacroApplication
 
 /// Removes attributes from a syntax tree while maintaining their surrounding trivia.
-private class AttributeRemover: SyntaxRewriter {
-  var attributesToRemove: [AttributeSyntax]
+@_spi(Testing)
+public class AttributeRemover: SyntaxRewriter {
+  let predicate: (AttributeSyntax) -> Bool
 
   var triviaToAttachToNextToken: Trivia = Trivia()
 
-  init(attributesToRemove: [AttributeSyntax]) {
-    self.attributesToRemove = attributesToRemove
+  /// Initializes an attribute remover with a given predicate to determine which attributes to remove.
+  ///
+  /// - Parameter predicate: A closure that determines whether a given `AttributeSyntax` should be removed.
+  ///   If this closure returns `true` for an attribute, that attribute will be removed.
+  public init(removingWhere predicate: @escaping (AttributeSyntax) -> Bool) {
+    self.predicate = predicate
   }
 
-  override func visit(_ node: AttributeListSyntax) -> AttributeListSyntax {
+  public override func visit(_ node: AttributeListSyntax) -> AttributeListSyntax {
     var filteredAttributes: [AttributeListSyntax.Element] = []
     for case .attribute(let attribute) in node {
-      if attributesToRemove.contains(attribute) {
+      if self.predicate(attribute) {
         var leadingTrivia = attribute.leadingTrivia
 
         // Don't leave behind an empty line when the attribute being removed is on its own line,
@@ -450,7 +455,7 @@ private class AttributeRemover: SyntaxRewriter {
     return AttributeListSyntax(filteredAttributes)
   }
 
-  override func visit(_ token: TokenSyntax) -> TokenSyntax {
+  public override func visit(_ token: TokenSyntax) -> TokenSyntax {
     return prependAndClearAccumulatedTrivia(to: token)
   }
 
@@ -573,7 +578,7 @@ private class MacroApplication<Context: MacroExpansionContext>: SyntaxRewriter {
 
       let attributesToRemove = self.macroAttributes(attachedTo: visitedNode).map(\.attributeNode)
 
-      return AttributeRemover(attributesToRemove: attributesToRemove).rewrite(visitedNode)
+      return AttributeRemover(removingWhere: { attributesToRemove.contains($0) }).rewrite(visitedNode)
     }
 
     return nil
--- a/Tests/SwiftSyntaxMacroExpansionTest/AttributeRemoverTests.swift
+++ b/Tests/SwiftSyntaxMacroExpansionTest/AttributeRemoverTests.swift
@@ -10,144 +10,131 @@
 //
 //===----------------------------------------------------------------------===//
 
-//==========================================================================//
-// IMPORTANT: The macros defined in this file are intended to test the      //
-// behavior of MacroSystem. Many of them do not serve as good examples of   //
-// how macros should be written. In particular, they often lack error       //
-// handling because it is not needed in the few test cases in which these   //
-// macros are invoked.                                                      //
-//==========================================================================//
-
-import SwiftDiagnostics
+import SwiftParser
 import SwiftSyntax
-import SwiftSyntaxMacroExpansion
-import SwiftSyntaxMacros
-import SwiftSyntaxMacrosTestSupport
+@_spi(Testing) import SwiftSyntaxMacroExpansion
 import XCTest
+import _SwiftSyntaxTestSupport
 
-fileprivate struct NoOpPeerMacro: PeerMacro {
-  static func expansion(
-    of node: AttributeSyntax,
-    providingPeersOf declaration: some DeclSyntaxProtocol,
-    in context: some MacroExpansionContext
-  ) throws -> [DeclSyntax] {
-    return []
-  }
+fileprivate func assertSyntaxRemovingTestAttributes(
+  _ originalSource: String,
+  reduction expectedReducedSource: String,
+  file: StaticString = #filePath,
+  line: UInt = #line
+) {
+  let attributeToRemove = AttributeSyntax(stringLiteral: "@Test")
+
+  let reducedSource = AttributeRemover(
+    removingWhere: { $0.trimmedDescription == attributeToRemove.trimmedDescription }
+  )
+  .rewrite(
+    Parser.parse(source: originalSource)
+  )
+
+  assertStringsEqualWithDiff(
+    reducedSource.description,
+    expectedReducedSource,
+    "Attribute removal did not produce the expected reduced source",
+    additionalInfo: """
+      Actual reduced source:
+      \(reducedSource)
+      """,
+    file: file,
+    line: line
+  )
 }
 
 final class AttributeRemoverTests: XCTestCase {
   func testEmptyOnSameLineAsVariable() {
-    assertMacroExpansion(
+    assertSyntaxRemovingTestAttributes(
       "@Test var x: Int",
-      expandedSource: "var x: Int",
-      macros: [
-        "Test": NoOpPeerMacro.self
-      ]
+      reduction: "var x: Int"
     )
   }
 
   func testEmptyTwiceOnSameLineAsVariable() {
-    assertMacroExpansion(
+    assertSyntaxRemovingTestAttributes(
       "@Test @Test var x: Int",
-      expandedSource: "var x: Int",
-      macros: [
-        "Test": NoOpPeerMacro.self
-      ]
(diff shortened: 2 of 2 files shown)


gohanlon:
@ahoppen @bnbarham @Matejkob Do you think unit testing of `AttributeRemover` is worth pursuing?

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
```diff
@@ -384,19 +384,25 @@ struct MacroSystem {
...
+public class AttributeRemover: SyntaxRewriter {
   var attributesToRemove: [AttributeSyntax]
+  let attributeEquals: (AttributeSyntax, AttributeSyntax) -> Bool
 
   var triviaToAttachToNextToken: Trivia = Trivia()
 
-  init(attributesToRemove: [AttributeSyntax]) {
+  public init(
+    attributesToRemove: [AttributeSyntax],
+    attributeEquals: @escaping (AttributeSyntax, AttributeSyntax) -> Bool = { $0 == $1 }
```
ahoppen, line 396: Instead of having `attributesToRemove` and `attributeEquals` what do you think about just passing a closure that returns `true` for all the attributes that should be removed and `false` for the attributes that should stick around?

Tests/SwiftSyntaxMacroExpansionTest/AttributeRemoverTests.swift
```diff
@@ -10,505 +10,484 @@
...
-    providingPeersOf declaration: some DeclSyntaxProtocol,
-    in context: some MacroExpansionContext
-  ) throws -> [DeclSyntax] {
-    return []
+import _SwiftSyntaxTestSupport
+@_spi(AttributeRemover) import SwiftSyntaxMacroExpansion
+
+fileprivate func assertSyntaxRemovingTestAttributes(
+  of originalSource: () throws -> String,
+  reduction expectedReducedSource: () -> String,
```
ahoppen, line 21: Any reason why you made theses closures instead of normal arguments?

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
```diff
@@ -384,19 +384,25 @@ struct MacroSystem {
 // MARK: - MacroApplication
 
 /// Removes attributes from a syntax tree while maintaining their surrounding trivia.
-private class AttributeRemover: SyntaxRewriter {
+@_spi(AttributeRemover)
```
ahoppen, line 387: I would just name this `@_spi(Testing)` because it describes the intent that it’s only exposed publicly for testing.

Tests/SwiftSyntaxMacroExpansionTest/AttributeRemoverTests.swift
```diff
@@ -10,505 +10,484 @@
...
+      "Attribute removal did not produce the expected reduced source",
+      additionalInfo: """
+      Actual reduced source:
+      \(reducedSource)
+      """,
+      file: file,
+      line: line
+    )
+  } catch {
+    XCTFail("Threw error: \(error)", file: file, line: line)
```
ahoppen, line 51: If you made the inputs normal parameters instead of closures, I think there’s no need for the `do {} catch {}` anymore.

Tests/SwiftSyntaxMacroExpansionTest/AttributeRemoverTests.swift
```diff
@@ -10,505 +10,484 @@
...
+    )
+      .rewrite(
+        Parser.parse(
+          source: try originalSource()
+        )
+      )
+
+    assertStringsEqualWithDiff(
+      reducedSource.description.trimmingCharacters(in: .newlines),
+      expectedReducedSource().trimmingCharacters(in: .newlines),
```
ahoppen, line 41: Do we need the `.trimmingCharacters(in: .newlines)` here?

ahoppen (commented):
I like this. A couple of comments inline

Tests/SwiftSyntaxMacroExpansionTest/AttributeRemoverTests.swift
gohanlon: Trimming is needed on `reducedSource` for the tests to pass. IMO, that means `AttributeRemover` needs to be fixed, as this at least undesirable if not a bug. (This behavior was previously masked by `assertMacroExpansion`.)

I originally included trimming newlines in order to mimic `assertMacroExpansion`&rsquo;s behavior: https://github.com/apple/swift-syntax/blob/f9744529b24e6a118ca29c12dfdc1103a3738819/Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift#L295-L296

Without trimming newlines on the reduced source, 5 tests fail due to an extra leading newline when the removed attribute has no proceeding tokens: e.g.:

```swift
assertSyntaxRemovingTestAttributes {
  """
  @Test
  var x: Int
  """
} reduction: {
  "var x: Int"
}
//failed - Attribute removal did not produce the expected reduced source
//+
// var x: Int
//
//Actual reduced source:
//
//var x: Int
```

and:

```swift
assertSyntaxRemovingTestAttributes {
  """
  @Test
  /* comment */
  var value: Int
  """
} reduction: {
  """
  /* comment */
  var value: Int
  """
}
//failed - Attribute removal did not produce the expected reduced source
//+
// /* comment */
// var value: Int
//
//Actual reduced source:
//
///* comment */
//var value: Int
```

I'm leaning towards getting rid of the newline trimming in the assert helper for now, and then fixing the tests that fail because of that by including the unwanted leading newlines in the expected output. I'll include some comments to draw attention to these not-so-great test outputs. Once this PR lands, I'll open an issue so this can be fixed.

My reasoning is that, before this PR, it's hard to replicate the bad behavior, and I'm not sure that the fix will be trivial.

Thoughts?

Tests/SwiftSyntaxMacroExpansionTest/AttributeRemoverTests.swift
gohanlon: Yes, my intent is to make the resulting tests easier to read:

```swift
// string parameters
assertSyntaxRemovingTestAttributes(
  """
  struct Foo {
    @Test @State var x: Int
  }
  """,
  reduction: """
    struct Foo {
      @State var x: Int
    }
    """
)
```

vs.

```swift
// closure parameters
assertSyntaxRemovingTestAttributes {
  """
  struct Foo {
    @Test @State var x: Int
  }
  """
} reduction: {
  """
  struct Foo {
    @State var x: Int
  }
  """
}
```

I think the use of closure syntax improves readability and better supports visual comparison. There are already 30 test cases using the helper, and the benefits would increase over time as more tests are added. (I've already come up with some missing tests to add, some of which are failing.) However, I recognize this is subjective, and this style may well be out of place in this project. With that considered, let me know how you want me to proceed.

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
gohanlon: I like this. My original intent was to isolate the changes to `AttributeRemover`, but I'm happy to try changing `AttributeRemover` to take single closure and updating `MacroSystem`&rsquo;s usage.

Tests/SwiftSyntaxMacroExpansionTest/AttributeRemoverTests.swift
ahoppen: > I'm leaning towards getting rid of the newline trimming in the assert helper for now, and then fixing the tests that fail because of that by including the unwanted leading newlines in the expected output. I'll include some comments to draw attention to these not-so-great test outputs. Once this PR lands, I'll open an issue so this can be fixed.

That sounds good to me 👍🏽 Could you also file an issue that we should fix `AttributeRemover` to remove the empty newline in this case?

Tests/SwiftSyntaxMacroExpansionTest/AttributeRemoverTests.swift
ahoppen: I disagree here that the trailing closures are easier to read. I personally prefer normal arguments, which would also be more consistent with eg. `assertParse`, which also takes several multi-line string literals.

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
ahoppen: Also, just something to consider: If we do this change, it might be straightforward to generalize `AttributeRemover` to a general `NodeRemover`, which we can also use to determine how to shift trivia if we want to remove a token in `ParseDiagnosticGenerator`. Definitely not part of this PR though, but it could be a nice follow-up if you’re interested.

gohanlon: I changed `AttributeRemover` to use a predicate, and updated `MacroApplication`&rsquo;s and `assertSyntaxRemovingTestAttributes`&rsquo;s usage accordingly.

Tests/SwiftSyntaxMacroExpansionTest/AttributeRemoverTests.swift
gohanlon: I refactored the `assertSyntaxRemovingTestAttributes` helper to use String arguments.

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
gohanlon: Done.

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
```diff
@@ -384,19 +384,20 @@ struct MacroSystem {
...
+@_spi(Testing)
+public class AttributeRemover: SyntaxRewriter {
+  let predicate: (AttributeSyntax) -> Bool
 
   var triviaToAttachToNextToken: Trivia = Trivia()
 
-  init(attributesToRemove: [AttributeSyntax]) {
-    self.attributesToRemove = attributesToRemove
+  public init(where predicate: @escaping (AttributeSyntax) -> Bool) {
+    self.predicate = predicate
```
ahoppen, line 394: Could you add a doc comment here that say that all the attributes for which the predicate returns `true` will be removed just to clarify that it’s not other way round.

Tests/SwiftSyntaxMacroExpansionTest/AttributeRemoverTests.swift
ahoppen: Thank you!

ahoppen (commented):
Thanks. Looks good to me. Just one small comment for adding documentation.

gohanlon:
@ahoppen I think this PR is just about ready for final review, pending your review of the latest changes. (Particularly the `init(where:)` to `init(removingWhere:)` rename and the doc comment I added.)

And, when you say to, I'll squash this PR into a single commit and mark the PR as "Ready" instead of "Draft".

ahoppen (approved):
Great. Let's get this in.

gohanlon:
@ahoppen Squashed. Let me know if there's anything else!

kimdv:
@swift-ci please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test macOS

PR #2249, opened 2023-09-29
mark the BumpPtrAllocator class as @_spi(BumpPtrAllocator).

ChiduAnush:
mark the BumpPtrAllocator class as @_spi(BumpPtrAllocator) instead of @_spi(RawSyntax).
resolves the issue #68351 (https://github.com/apple/swift/issues/68351)

### DIFF
--- a/Sources/SwiftParser/Lexer/LexemeSequence.swift
+++ b/Sources/SwiftParser/Lexer/LexemeSequence.swift
@@ -10,7 +10,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-@_spi(RawSyntax) import SwiftSyntax
+@_spi(RawSyntax) @_spi(BumpPtrAllocator) import SwiftSyntax
 
 extension Lexer {
   /// A sequence of ``Lexer/Lexeme`` tokens starting from a ``Lexer/Cursor``
--- a/Sources/SwiftParser/Lexer/Cursor.swift
+++ b/Sources/SwiftParser/Lexer/Cursor.swift
@@ -10,7 +10,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-@_spi(RawSyntax) import SwiftSyntax
+@_spi(RawSyntax) @_spi(BumpPtrAllocator) import SwiftSyntax
 
 extension SyntaxText {
   fileprivate func containsPlaceholderEnd() -> Bool {
--- a/Sources/SwiftParser/Lexer/RegexLiteralLexer.swift
+++ b/Sources/SwiftParser/Lexer/RegexLiteralLexer.swift
@@ -10,7 +10,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-@_spi(RawSyntax) import SwiftSyntax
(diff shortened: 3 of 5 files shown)


ahoppen (commented):
Could you make sure that swift-syntax builds with this change? There are a couple of references to `BumpPtrAllocator` within the swift-syntax package that now no longer work. You will need to add `@_spi(BumpPtrAllocator)` to the imports in those files as well.

ChiduAnush:
just to confirm, in the swift-syntax repo, inside the sources folder, inside the swiftSyntax folder, I need to include the `@_spi(BumpPtrAllocator)` in the required files?

which files exactly should I add this import statement?

when I do a search in vscode for `BumpPtrAllocator`, it returns about 16 results. most of then pass BumpPtrAllocator as StateAllocator

bnbarham:
> just to confirm, in the swift-syntax repo, inside the sources folder, inside the swiftSyntax folder, I need to include the @_spi(BumpPtrAllocator) in the required files?

If you run a build (either in your editor or just `swift build`) you'll see there will be many `error: cannot find 'BumpPtrAllocator' in scope` errors. They're caused because those files import `SwiftSyntax` with `@_spi(RawSyntax)` instead of `BumpPtrAllocator` as you've just changed it to.

You'll need to update those to use `@_spi(BumpPtrAllocator)`, though It might be that they need *both* `@_spi(RawSyntax)` and `@_spi(BumpPtrAllocator)` depending on what's used - you'll need to check.

ChiduAnush:
whenever I try running `swift build` in the cloned repo, I get the following error, 
    xcrun: error: unable to lookup item 'PlatformPath' from command line tools installation
    xcrun: error: unable to lookup item 'PlatformPath' in SDK '/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk'
    
I have checked that `xcode-select -p` returns a valid path. so it is installed properly.

ahoppen:
Oh, that's odd. I have never seen this issue before. A couple of questions:
- Which version of Xcode do you have installed. 
- What's the output of `xcode-select -p`
- What's the output of `swift --version`?

ChiduAnush:
okay, I got it. had to reinstall my Xcode.
worked fine. I saw the errors and put in the import statements wherever required.

ChiduAnush:
I squashed the commits.

Sources/SwiftParser/Lexer/LexemeSequence.swift
```diff
@@ -11,6 +11,7 @@
 //===----------------------------------------------------------------------===//
 
 @_spi(RawSyntax) import SwiftSyntax
+@_spi(BumpPtrAllocator) import SwiftSyntax
```
AnthonyLatsis, line 14: I think we prefer to [pile up the attributes](https://github.com/apple/swift-syntax/blob/main/Sources/SwiftParser/TokenSpecSet.swift) on a single import.

ahoppen (approved):
LGTM 👍🏽

ahoppen:
@swift-ci Please test Windows

ahoppen:
https://github.com/apple/swift/pull/68860

@swift-ci Please test

PR #2259, opened 2023-10-03
Fix StringifyMacro's interpolation test

stephencelis:
While bringing in the latest macro tests into our [MacroTesting](https://github.com/pointfreeco/swift-macro-testing) project [here](https://github.com/pointfreeco/swift-macro-testing/pull/5), I noticed a regression in the stringify macro's test. The original test in the macro template asserts that raw interpolation is preserved by the macro, but it seems that this test accidentally is interpolating the XCTestCase's `name` property into the assertion instead.


kimdv:
@swift-ci please test

Matejkob (approved):
Thank you for doin' this! ❤️

PR #2260, opened 2023-10-04
Delete SyntaxTransformVisitor

ahoppen:
Since ASTGen is no longer using it, we can delete it.


ahoppen:
@swift-ci Please test

bnbarham (commented):
Will need to remove from CMake as well

ahoppen:
@swift-ci Please test

ahoppen:
> Heh, I was experimenting with using this in a swift-format branch to rewrite parts of the pretty-printer 😭 What I'm in need of is a visitor where each `visit` method is truly recursive instead of splitting `visit` and `visitPost` into separate operations.

Why doesn’t `SyntaxVisitor` work for you? You could just ignore `visitPost`, right? I feel like I’m missing something.

> I could coax `SyntaxRewriter` to do what I want by just always returning the input node; what would the overhead of that be? If the `visit` methods always return the same node, would there be any excessive overhead there?

The overhead shouldn’t be big. `SyntaxRewriter` has a special check to see if the syntax node has changed and if no node is changed, it doesn’t actually do any rewriting.

allevato:
> Why doesn’t `SyntaxVisitor` work for you? You could just ignore `visitPost`, right? I feel like I’m missing something.

One of the challenges of today's implementation is that because visitation of children is handled by returning `.visitChildren/.skipChildren`, we're not doing any direct recursion within the `visit` methods. That makes it error-prone because we can't use recursion and scoping to require that certain formatting tokens are correctly positioned before and after syntax nodes/tokens. For example, every `open` grouping command should have a corresponding `close`, and they need to be nested correctly.

Here's the structure I'd like to have instead, using true recursion and closure-based helper methods to define scopes that ensure everything lines up: https://github.com/allevato/swift-format/blob/new-pretty-printer/Sources/SwiftFormat/PrettyPrint/NewTokenStreamCreator.swift#L340-L371

But maybe I'm overthinking this. I guess I could stick with `SyntaxVisitor` but manually call `visit` on a node's children and always return `.skipChildren` from the `visit` methods to short-circuit the built-in traversal? Do you see any problems with that approach?

ahoppen:
@swift-ci Please test

PR #2262, opened 2023-10-04
[509] Do not add whitespace between right angle and period

ahoppen:
Cherry-pick https://github.com/apple/swift-syntax/pull/2224 to package-release/509.

---

`Foo<T>.bar` should not have whitespace added between `>` and `.`.


ahoppen:
✅ Local testing passed for this change.

bnbarham (approved):

PR #2264, opened 2023-10-05
Rename `NoteMessage.fixItID` to `noteID`

### LINKED ISSUE #2261: Rename `NoteMessage.fixItID` to `noteID`
`NoteMessage` having a property named `fixItID` was a stupid copy-paste error. I should be named `noteID`.

We need to consider how to mitigate API-breakage impact from this change, but AFAICT we will need to have an API-breaking change to rename the property. But maybe there is some clever way to avoid the API breakage.

kimdv:
Fixes #2261

### DIFF
--- a/Release Notes/510.md
+++ b/Release Notes/510.md
@@ -42,6 +42,10 @@
 
 ## API-Incompatible Changes
 
+- `NoteMessage.fixItID` renamed to `noteID`
+  - Description: This was an error that it was named `fixItID` and should have been named `noteID` instead. Accesses to `fixItID` are deprecated and forward to `noteID`. Any types that conform `NoteMessage` it will need to be updated to provide a `noteID` instead of a `fixItID`.
+  - Issue: https://github.com/apple/swift-syntax/issues/2261
+  - Pull Request: https://github.com/apple/swift-syntax/pull/2264
 
 ## Template
 
--- a/Sources/SwiftDiagnostics/Note.swift
+++ b/Sources/SwiftDiagnostics/Note.swift
@@ -20,7 +20,14 @@ public protocol NoteMessage {
   var message: String { get }
 
   /// See ``MessageID``.
-  var fixItID: MessageID { get }
+  var noteID: MessageID { get }
+}
+
+extension NoteMessage {
+  @available(*, deprecated, message: "Use noteID instead.", renamed: "noteID")
+  public var fixItID: MessageID {
+    return noteID
+  }
 }
 
 /// A note that points to another node that's relevant for a Diagnostic.
--- a/Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift
+++ b/Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift
@@ -36,16 +36,21 @@ public extension ParserError {
 }
 
 public protocol ParserNote: NoteMessage {
-  var fixItID: MessageID { get }
+  var noteID: MessageID { get }
 }
 
 public extension ParserNote {
+  @available(*, deprecated, message: "Use noteID instead.", renamed: "noteID")
   static var fixItID: MessageID {
+    return Self.noteID
+  }
+
+  static var noteID: MessageID {
     return MessageID(domain: diagnosticDomain, id: "\(self)")
   }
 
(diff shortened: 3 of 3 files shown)


Sources/SwiftDiagnostics/Note.swift
```diff
@@ -20,7 +20,14 @@ public protocol NoteMessage {
...
   /// See ``MessageID``.
-  var fixItID: MessageID { get }
+  var noteID: MessageID { get }
+}
+
+extension NoteMessage {
+  @available(*, deprecated, message: "Use noteID instead.", renamed: "noteID")
+  public var fixItID: MessageID {
+    return noteID
+  }
```
kimdv, line 30: Not sure if this is what you meant.

ahoppen (commented):
Could you add an entry to the release notes? 

If you previously had a type that’s conforming to `NoteMessage`, there’s a source breaking change because you need to now implement `noteID` instead of `fixItID`. The same that you found if you didn’t update `ParserNote`.

kimdv:
> Could you add an entry to the release notes?

Yes. To the 5.11 version or 5.10?

> If you previously had a type that’s conforming to `NoteMessage`, there’s a source breaking change because you need to now implement `noteID` instead of `fixItID`. The same that you found if you didn’t update `ParserNote`.

I'm not sure we can avoid that as we cannot make a default implementation.
So should we just accept it?

Matejkob (commented):
I was quite confused when I saw this for the first time; now everything makes sense! 

The only solution that comes to mind, without causing code to break, is to create an entirely new protocol in place of `NoteMessage` and mark the whole `NoteMessage` protocol as deprecated. This approach has its own drawbacks, and to be honest, I'm not a fan. But it seems like the only viable way to avoid API breakage.

ahoppen:
> Yes. To the 5.11 version or 5.10?

Oh, whoops. I never split the release notes after we branched `release/5.10`. I’m splitting them here: https://github.com/apple/swift-syntax/pull/2270

kimdv:
> > Yes. To the 5.11 version or 5.10?
> 
> 
> 
> Oh, whoops. I never split the release notes after we branched `release/5.10`. I’m splitting them here: https://github.com/apple/swift-syntax/pull/2270. After that’s merged, please add the release notes to `510.md`.

Should I also cherry pick this to 5.10?

ahoppen:
> Should I also cherry pick this to 5.10?

Opening a PR would definitely be good.

kimdv:
> > Should I also cherry pick this to 5.10?
> 
> Opening a PR would definitely be good.

Will open when this is merged then. 
@ahoppen placed under breaking change, as you mention it will not compile if you conform to the protocol.

Release Notes/510.md
```diff
@@ -42,6 +42,11 @@
 
 ## API-Incompatible Changes
 
+- Note message Fix-It ID
+  - Description: This was an error that it was named `fixItID` and should have been named `noteID` instead. The deprecation methods
+      emit a warning that informs the developer that it have been renamed. If you previously conformed to `fixItID` it will not compile.
+  - Issue: https://github.com/apple/swift-syntax/issues/2261
+  - Pull Request: https://github.com/apple/swift-syntax/pull/2264
```
ahoppen, line 49: Minor suggestsions

```suggestion
- `NoteMessage.fixItID` renamed to `noteID`
  - Description: This was an error that it was named `fixItID` and should have been named `noteID` instead. Accesses to `fixItID` are deprecated and forward to `noteID`. Any types that conform `NoteMessage` it will need to be updated to provide a `noteID` instead of a `fixItID`.
  - Issue: https://github.com/apple/swift-syntax/issues/2261
  - Pull Request: https://github.com/apple/swift-syntax/pull/2264
```

ahoppen (commented):
Minor comment on the release notes, otherwise LGTM.

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test windows

kimdv:
@swift-ci please test windows

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test windows

PR #2265, opened 2023-10-06
Casts to children of `SyntaxProtocol` are not always invalid

bnbarham:
Update these overrides so they don't always fail.

### DIFF
--- a/CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxBaseNodesFile.swift
+++ b/CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxBaseNodesFile.swift
@@ -71,6 +71,7 @@ let syntaxBaseNodesFile = SourceFileSyntax(leadingTrivia: copyrightHeader) {
         /// Force-casts the current syntax node to a given specialized syntax type.
         ///
         /// - Returns: An instance of the specialized type.
+        ///
         /// - Warning: This function will crash if the cast is not possible. Use `as` to safely attempt a cast.
         public func cast<S: \#(node.kind.protocolType)>(_ syntaxType: S.Type) -> S {
           return self.as(S.self)!
@@ -79,9 +80,6 @@ let syntaxBaseNodesFile = SourceFileSyntax(leadingTrivia: copyrightHeader) {
         /// Checks if the current syntax node can be upcast to its base node type (``\#(node.kind.syntaxType)``).
         ///
         /// - Returns: `true` since the node can always be upcast to its base node.
-        ///
-        /// - Note: This method overloads the general `is` method and is marked deprecated to produce a warning
-        ///         informing the user that the upcast will always succeed.
         @available(*, deprecated, message: "This cast will always succeed")
         public func `is`(_ syntaxType: \#(node.kind.syntaxType).Type) -> Bool {
           return true
@@ -90,9 +88,6 @@ let syntaxBaseNodesFile = SourceFileSyntax(leadingTrivia: copyrightHeader) {
         /// Attempts to upcast the current syntax node to its base node type (``\#(node.kind.syntaxType)``).
         ///
         /// - Returns: The base node created from the current syntax node, as the node can always be upcast to its base type.
-        ///
-        /// - Note: This method overloads the general `as` method and is marked deprecated to produce a warning
-        ///         informing the user the upcast should be performed using the target base node's initializer.
         @available(*, deprecated, message: "Use `\#(node.kind.syntaxType).init` for upcasting")
         public func `as`(_ syntaxType: \#(node.kind.syntaxType).Type) -> \#(node.kind.syntaxType)? {
           return \#(node.kind.syntaxType)(self)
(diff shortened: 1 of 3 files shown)


bnbarham:
@swift-ci please test

CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxBaseNodesFile.swift
```diff
@@ -319,7 +214,85 @@ let syntaxBaseNodesFile = SourceFileSyntax(leadingTrivia: copyrightHeader) {
...
+      }
+
+      /// Force-upcast the current syntax node to its base node type (``\#(node.kind.syntaxType)``).
+      ///
+      /// - Returns: The base node created from the current syntax node, as the node can
+      ///            always be upcast to its base type.
+      @available(*, deprecated, message: "Use `\#(node.kind.syntaxType).init` for upcasting")
+      public func cast(_ syntaxType: \#(node.kind.syntaxType).Type) -> \#(node.kind.syntaxType) {
+        return \#(node.kind.syntaxType)(self)
+      }
```
bnbarham, line 250: FWIW I'm not sure we should prefer the constructor here. If it wasn't for the performance, we'd use existential's instead and then upcasting with the regular `as` would be the way to do this. `.as` is analogous so preferring the init is a little odd. Have left it in for now though.

CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxBaseNodesFile.swift
```diff
@@ -35,117 +35,6 @@ let syntaxBaseNodesFile = SourceFileSyntax(leadingTrivia: copyrightHeader) {
...
-        public func cast<S: \#(node.kind.protocolType)>(_ syntaxType: S.Type) -> S {
-          return self.as(S.self)!
-        }
-
-        /// Checks if the current syntax node can be upcast to its base node type (``\#(node.kind.syntaxType)``).
-        ///
-        /// - Returns: `true` since the node can always be upcast to its base node.
-        ///
-        /// - Note: This method overloads the general `is` method and is marked deprecated to produce a warning
-        ///         informing the user that the upcast will always succeed.
```
bnbarham, line 84: I left the note's out. To me they read as notes to *us*, rather than developers using swift-syntax.

ahoppen (commented):
Hmm, AFAICT, this will no longer warn when casting to types that are a `SyntaxProtocol` but not an `ExprSyntaxProtocol` etc.

For example, I think the following should warn but doesn’t with this change:
```swift
func test(node: FunctionDeclSyntax) {
  _ = node.as(CodeBlockItemListSyntax.self)
}

**Review by Matejkob — COMMENTED** [review]
Hey @bnbarham, thanks for the PR! I'm trying to wrap my head around the reasoning behind these changes. Could you shed some light?

> Casts to children of `SyntaxProtocol` are not always invalid

Got any examples? It'd help me get a clearer picture of what you're getting at.

> It is weird to cast to `SyntaxProtocol` rather than the hierarchy that's expected, but it isn't invalid.

So, are you talking about using a custom `as` method for casting to `SyntaxProtocol` directly, like this: `DeclSyntax("").as(SyntaxProtocol.self)`? If I'm not mistaken, that would be invalid code, right? What am I missing?

File: CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxBaseNodesFile.swift (hunk 1 above)
Comment (bnbarham) [reply CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxBaseNodesFile.swift:250]: > In my understating it distinguishes between using constructors and as methods for upcasting to base node types

Thanks for pointing that out! If we didn't need `ExprSyntax` then the cast wouldn't be needed anyway, so it probably doesn't matter what we choose. Given we have docs suggesting the constructor already, I'll probably just leave it.

**Comment by ahoppen** [comment]
> The alternative here is to leave the deprecation, but not have it fail unconditionally.

That’s what I would prefer

File: Sources/SwiftSyntax/SyntaxProtocol.swift [hunk 3]
```diff
@@ -89,7 +89,7 @@ public extension SyntaxProtocol {
     return S.init(self)
   }
 
-  /// Attempts to upcast the current syntax node to ``Syntax`` node..
+  /// Attempts to upcast the current syntax node to ``Syntax`` node.
```
kimdv, line 92: 🦅-eyes 😅

Matejkob:
Thanks, @bnbarham, for the explanation ❤️

> > The alternative here is to leave the deprecation, but not have it fail unconditionally.
> 
> That’s what I would prefer

Changing it to not fail unconditionally is a great idea. Additionally, I'm wondering whether we should apply this change to every overload. When I initially wrote the code, I was pretty confident that I had covered all the cases. However, as we can see, I was mistaken. Now, I'm questioning if there might be other edge cases that exist elsewhere.

bnbarham:
Updated to just not unconditionally fail rather than the refactor I had previously. Open to ideas for wording the deprecation warning 😅 

> Changing it to not fail unconditionally is a great idea. Additionally, I'm wondering whether we should apply this change to every overload. When I initially wrote the code, I was pretty confident that I had covered all the cases. However, as we can see, I was mistaken. Now, I'm questioning if there might be other edge cases that exist elsewhere.

I'm fairly confident with the others, so should be fine :). If not, we can always fix later!

bnbarham:
@swift-ci please test

Sources/SwiftSyntax/generated/SyntaxBaseNodes.swift
```diff
@@ -90,49 +84,55 @@ extension DeclSyntaxProtocol {
...
-    return false
+    return self.as(syntaxType) != nil
   }
   
 
-  /// Attempts to cast the current syntax node to a given node type from the different base node protocol hierarchy than ``DeclSyntaxProtocol``.
+  /// Attempts to cast the current syntax node to a given node type from the a base node protocol hierarchy other than
+  /// ``DeclSyntaxProtocol``. This is generally does not make sense to call - if in a generic ``SyntaxProtocol``
+  /// context, use ``DeclSyntaxProtocol`` instead. Otherwise this must be a base node other than
+  /// ``DeclSyntaxProtocol``, in which case this cast will always return `nil`.
```
ahoppen, line 110: I think this is duplicating the `- Note:`

Sources/SwiftSyntax/generated/SyntaxBaseNodes.swift
```diff
@@ -90,49 +84,55 @@ extension DeclSyntaxProtocol {
...
   ///
-  /// - Returns: `false` since the node can not be cast to the node type from different base node protocol hierarchy than ``DeclSyntaxProtocol``.
+  /// - Returns: `true` if the node can be cast, `false` otherwise.
   ///
-  /// - Note: This method overloads the general `is` method and is marked as deprecated to produce a warning,
-  ///         informing the user that the cast will always fail.
-  @available(*, deprecated, message: "This cast will always fail")
+  /// - Note: This generally does not make sense to call - if in a generic ``SyntaxProtocol`` context, use
+  /// ``DeclSyntaxProtocol`` instead. Otherwise this must be a node outside of ``DeclSyntaxProtocol``,
+  /// in which case this call will always fail.
```
ahoppen, line 100: This is a little confusing to me. I’m trying to come up with a better phrasing.

Sources/SwiftSyntax/generated/SyntaxBaseNodes.swift
```diff
@@ -90,49 +84,55 @@ extension DeclSyntaxProtocol {
...
+    return S.init(self)
   }
   
 
-  /// Force-casts the current syntax node to a given node type from the different base node protocol hierarchy than ``DeclSyntaxProtocol``.
+  /// Force-casts the current syntax node to a given node type from a base node protocol hierarchy other than
+  /// ``DeclSyntaxProtocol``.
   ///
   /// - Returns: This method will always trigger a runtime crash and never return.
   ///
```
ahoppen, line 127: This comment is no longer correct

Sources/SwiftSyntax/generated/SyntaxBaseNodes.swift
ahoppen: What about

```swift
  /// - Note: In most cases, this is casting a ``DeclSyntaxProtocol`` to a node that is not a ``DeclSyntaxProtocol``, which will always fail.
  ///   If the `syntaxType` argument is a generic type, constrain it to ``DeclSyntaxProtocol`` instead of ``SyntaxProtocol``.
```

Sources/SwiftSyntax/generated/SyntaxBaseNodes.swift
bnbarham: I swear I removed this 🤔. Thanks

Sources/SwiftSyntax/generated/SyntaxBaseNodes.swift
bnbarham: Thanks, sounds better 🙇

Sources/SwiftSyntax/generated/SyntaxBaseNodes.swift
bnbarham: Thanks. I reverted my old PR and apparently didn't copy across all the comment changes 😅

bnbarham:
@swift-ci please test

ahoppen:
@swift-ci Please test Windows

ahoppen (approved):

PR #2268, opened 2023-10-06
[Macros] Granular diagnostics when macro type is not found in a plugin

rintaro:
rdar://115571427

### DIFF
--- a/Sources/SwiftCompilerPlugin/CompilerPlugin.swift
+++ b/Sources/SwiftCompilerPlugin/CompilerPlugin.swift
@@ -63,7 +63,8 @@ public protocol CompilerPlugin {
 }
 
 extension CompilerPlugin {
-  func resolveMacro(moduleName: String, typeName: String) throws -> Macro.Type {
+  @_spi(Testing)
+  public func resolveMacro(moduleName: String, typeName: String) throws -> Macro.Type {
     let qualifedName = "\(moduleName).\(typeName)"
 
     for type in providingMacros {
@@ -77,11 +78,6 @@ extension CompilerPlugin {
 
     let pluginPath = CommandLine.arguments.first ?? Bundle.main.executablePath ?? ProcessInfo.processInfo.processName
     throw CompilerPluginError(message: "macro implementation type '\(moduleName).\(typeName)' could not be found in executable plugin '\(pluginPath)'")
-  }
-
-  // @testable
-  public func _resolveMacro(moduleName: String, typeName: String) -> Macro.Type? {
-    try? resolveMacro(moduleName: moduleName, typeName: typeName)
   }
 }
 
--- a/Sources/SwiftCompilerPlugin/CompilerPlugin.swift
+++ b/Sources/SwiftCompilerPlugin/CompilerPlugin.swift
@@ -63,7 +63,7 @@ public protocol CompilerPlugin {
 }
 
 extension CompilerPlugin {
-  func resolveMacro(moduleName: String, typeName: String) -> Macro.Type? {
+  func resolveMacro(moduleName: String, typeName: String) throws -> Macro.Type {
     let qualifedName = "\(moduleName).\(typeName)"
 
     for type in providingMacros {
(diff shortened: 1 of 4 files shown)


rintaro:
https://github.com/apple/swift/pull/69027
@swift-ci Please test

Sources/SwiftCompilerPlugin/CompilerPlugin.swift
```diff
@@ -74,12 +74,14 @@ extension CompilerPlugin {
...
-    return nil
+
+    let pluginPath = CommandLine.arguments.first ?? Bundle.main.executablePath ?? ProcessInfo.processInfo.processName
+    throw CompilerPluginError(message: "macro implementation type '\(moduleName).\(typeName)' could not be found in executable plugin '\(pluginPath)'")
   }
 
   // @testable
   public func _resolveMacro(moduleName: String, typeName: String) -> Macro.Type? {
-    resolveMacro(moduleName: moduleName, typeName: typeName)
+    try? resolveMacro(moduleName: moduleName, typeName: typeName)
```
ahoppen, line 84: Any reason why we don’t want to make `_resoleMacro` `throws` as well. Just because it’s not needed for test now doesn’t mean it won’t be needed in tests later.

Also: Why do we need `_resolveMacro` at all. Couldn’t `resolveMacro` just be `@_spi(Testing)`?

Sources/SwiftCompilerPlugin/CompilerPlugin.swift
```diff
@@ -242,3 +244,10 @@ private extension FileHandle {
     }
   }
 }
+
+struct CompilerPluginError: Error, CustomStringConvertible {
+  var description: String
+  init(message: String) {
+    self.description = message
+  }
+}
```
ahoppen, line 253: Do you think there would be value in structuring the errors and making `CompilerPluginError` an `enum` with a computed description property instead of shoving an entire error message in there?

Sources/SwiftCompilerPlugin/CompilerPlugin.swift
rintaro: I don't think it's worth it. If we want it later, we can do it later.

> Why do we need _resolveMacro at all

I took your advice https://github.com/apple/swift-syntax/pull/1359#discussion_r1118823628 If you recommend `@_spi(Testing)` on `resolveMacro`, I'd do it now and update the test.

Sources/SwiftCompilerPlugin/CompilerPlugin.swift
rintaro: Not at this point. As long as we only need the message, I don't think make it an enum benefits anything.

bnbarham: I was also considering commenting on this. It would be nice if rather than passing a random string (which we've duplicated in many places) into the eventual diagnostic, that we instead passed the paths/macro type/plugin kind along. I'm not sure how difficult that would be or if it's even feasible though so maybe it's not worth it.

As an example, one of the messages in this PR is `macro implementation type '\(moduleName).\(typeName)' could not be found in executable plugin '\(pluginPath)'` and another has just `type` instead. And IMO I think it would be better with neither.

rintaro: I have a plan to consolidate logics of plugin-server and in-process in the follow-ups (because they are both library plugins). In that change, I think we can consolidate this error message too between all three of executable, server, and in-process.

Sources/SwiftCompilerPlugin/CompilerPlugin.swift
ahoppen: Hmm, interesting. I didn’t know that I suggested it. I suppose this was before we introduced `@_spi(Testing)`, which I hadn’t thought of before. I like `@_spi(Testing)` better now.

Sources/SwiftCompilerPlugin/CompilerPlugin.swift
ahoppen: It’s not a blocker for me. It was just something that crossed my mind but I’ll leave it up to you to decide how to model the error.

rintaro:
https://github.com/apple/swift/pull/69027
@swift-ci Please test

ahoppen (approved):

Sources/SwiftCompilerPlugin/CompilerPlugin.swift
bnbarham: Okay, that would be nice 👍

bnbarham (approved):

rintaro:
https://github.com/apple/swift/pull/69027
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

rintaro:
https://github.com/apple/swift/pull/69027
@swift-ci Please test Windows

PR #2278, opened 2023-10-12
Refactor syntax node manipulation to use variable setters

Matejkob:
This commit replaces the use of `with` method for syntax node manipulation in SwiftSyntax examples with direct variable setters for better clarity and readability.

### DIFF
--- a/Examples/Sources/MacroExamples/Implementation/ComplexMacros/DictionaryIndirectionMacro.swift
+++ b/Examples/Sources/MacroExamples/Implementation/ComplexMacros/DictionaryIndirectionMacro.swift
@@ -21,10 +21,7 @@ extension DictionaryStorageMacro: MemberMacro {
     providingMembersOf declaration: some DeclGroupSyntax,
     in context: some MacroExpansionContext
   ) throws -> [DeclSyntax] {
-    let storage: DeclSyntax = "var _storage: [String: Any] = [:]"
-    return [
-      storage.with(\.leadingTrivia, [.newlines(1), .spaces(2)])
-    ]
+    return ["\n  var _storage: [String: Any] = [:]"]
   }
 }
 
@@ -43,11 +40,11 @@ extension DictionaryStorageMacro: MemberAttributeMacro {
 
     return [
       AttributeSyntax(
+        leadingTrivia: [.newlines(1), .spaces(2)],
         attributeName: IdentifierTypeSyntax(
           name: .identifier("DictionaryStorageProperty")
         )
       )
-      .with(\.leadingTrivia, [.newlines(1), .spaces(2)])
     ]
   }
 }
--- a/Examples/Sources/MacroExamples/Implementation/Expression/AddBlocker.swift
+++ b/Examples/Sources/MacroExamples/Implementation/Expression/AddBlocker.swift
@@ -27,7 +27,7 @@ public struct AddBlocker: ExpressionMacro {
       _ node: InfixOperatorExprSyntax
     ) -> ExprSyntax {
       // Identify any infix operator + in the tree.
-      if let binOp = node.operator.as(BinaryOperatorExprSyntax.self) {
+      if var binOp = node.operator.as(BinaryOperatorExprSyntax.self) {
         if binOp.operator.text == "+" {
           // Form the warning
           let messageID = MessageID(domain: "silly", id: "addblock")
@@ -72,17 +72,9 @@ public struct AddBlocker: ExpressionMacro {
             )
           )
 
-          return ExprSyntax(
-            node.with(
-              \.operator,
-              ExprSyntax(
-                binOp.with(
-                  \.operator,
-                  binOp.operator.with(\.tokenKind, .binaryOperator("-"))
-                )
-              )
-            )
-          )
+          binOp.operator.tokenKind = .binaryOperator("-")
+
+          return ExprSyntax(node.with(\.operator, ExprSyntax(binOp)))
         }
       }
 
--- a/Examples/Sources/MacroExamples/Implementation/Expression/FontLiteralMacro.swift
+++ b/Examples/Sources/MacroExamples/Implementation/Expression/FontLiteralMacro.swift
@@ -35,14 +35,12 @@ private func replaceFirstLabel(
   of tuple: LabeledExprListSyntax,
   with newLabel: String
 ) -> LabeledExprListSyntax {
-  guard let firstElement = tuple.first else {
+  if tuple.isEmpty {
     return tuple
   }
 
   var tuple = tuple
-  tuple[tuple.startIndex] =
-    firstElement
-    .with(\.label, .identifier(newLabel))
-    .with(\.colon, .colonToken())
+  tuple[tuple.startIndex].label = .identifier(newLabel)
+  tuple[tuple.startIndex].colon = .colonToken()
   return tuple
 }
--- a/Examples/Sources/MacroExamples/Implementation/MemberAttribute/WrapStoredPropertiesMacro.swift
+++ b/Examples/Sources/MacroExamples/Implementation/MemberAttribute/WrapStoredPropertiesMacro.swift
@@ -48,11 +48,11 @@ public struct WrapStoredPropertiesMacro: MemberAttributeMacro {
 
     return [
       AttributeSyntax(
+        leadingTrivia: [.newlines(1), .spaces(2)],
         attributeName: IdentifierTypeSyntax(
           name: .identifier(wrapperName.content.text)
         )
       )
-      .with(\.leadingTrivia, [.newlines(1), .spaces(2)])
     ]
   }
 }
--- a/Examples/Sources/MacroExamples/Implementation/Peer/AddAsyncMacro.swift
+++ b/Examples/Sources/MacroExamples/Implementation/Peer/AddAsyncMacro.swift
@@ -30,7 +30,7 @@ public struct AddAsyncMacro: PeerMacro {
   ) throws -> [DeclSyntax] {
 
     // Only on functions at the moment.
-    guard let funcDecl = declaration.as(FunctionDeclSyntax.self) else {
+    guard var funcDecl = declaration.as(FunctionDeclSyntax.self) else {
       throw CustomError.message("@addAsync only works on functions")
     }
 
@@ -42,7 +42,7 @@ public struct AddAsyncMacro: PeerMacro {
     }
 
     // This only makes sense void functions
-    if funcDecl.signature.returnClause?.type.with(\.leadingTrivia, []).with(\.trailingTrivia, []).description != "Void" {
+    if funcDecl.signature.returnClause?.type.as(IdentifierTypeSyntax.self)?.name.text != "Void" {
       throw CustomError.message(
         "@addAsync requires an function that returns void"
       )
@@ -58,7 +58,7 @@ public struct AddAsyncMacro: PeerMacro {
     }
 
     // Completion handler needs to return Void
-    if completionHandlerParameter.returnClause.type.with(\.leadingTrivia, []).with(\.trailingTrivia, []).description != "Void" {
+    if completionHandlerParameter.returnClause.type.as(IdentifierTypeSyntax.self)?.name.text != "Void" {
       throw CustomError.message(
         "@addAsync requires an function that has a completion handler that returns Void"
       )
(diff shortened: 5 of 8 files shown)


Examples/Sources/MacroExamples/Implementation/ComplexMacros/DictionaryIndirectionMacro.swift
```diff
@@ -41,14 +40,14 @@ extension DictionaryStorageMacro: MemberAttributeMacro {
...
-          name: .identifier("DictionaryStorageProperty")
-        )
+    var attribute = AttributeSyntax(
+      attributeName: IdentifierTypeSyntax(
+        name: .identifier("DictionaryStorageProperty")
       )
-      .with(\.leadingTrivia, [.newlines(1), .spaces(2)])
-    ]
+    )
+    attribute.leadingTrivia = [.newlines(1), .spaces(2)]
```
ahoppen, line 48: I think this can be simplified further to 
```swift
    return [AttributeSyntax(
      leadingTrivia: [.newlines(1), .spaces(2)],
      attributeName: IdentifierTypeSyntax(
        name: .identifier("DictionaryStorageProperty")
      )
    )]
```

Matejkob:
@ahoppen In one of my macro-related PRs you mentioned replacing node modification using `with` with direct variable setters. This PR is addressing this comment. Let me know what you think. I've changed every occurrence of `with` but I'm not convinced if in every case that is the way to go.

Examples/Sources/MacroExamples/Implementation/Expression/FontLiteralMacro.swift
```diff
@@ -35,14 +35,14 @@ private func replaceFirstLabel(
...
+  firstElement.label = .identifier(newLabel)
+  firstElement.colon = .colonToken()
+
   var tuple = tuple
-  tuple[tuple.startIndex] =
-    firstElement
-    .with(\.label, .identifier(newLabel))
-    .with(\.colon, .colonToken())
+  tuple[tuple.startIndex] = firstElement
   return tuple
```
ahoppen, line 47: Can we simplify this further to

```swift
guard !tuple.isEmpty else {
  return tuple
}
tuple[tuple.startIndex].label = .identifier(newLabel)
tuple[tuple.startIndex].colon = .colonToken()
return tuple
```

Examples/Sources/MacroExamples/Implementation/ComplexMacros/DictionaryIndirectionMacro.swift
```diff
@@ -21,10 +21,9 @@ extension DictionaryStorageMacro: MemberMacro {
     providingMembersOf declaration: some DeclGroupSyntax,
     in context: some MacroExpansionContext
   ) throws -> [DeclSyntax] {
-    let storage: DeclSyntax = "var _storage: [String: Any] = [:]"
-    return [
-      storage.with(\.leadingTrivia, [.newlines(1), .spaces(2)])
-    ]
+    var storage: DeclSyntax = "var _storage: [String: Any] = [:]"
+    storage.leadingTrivia = [.newlines(1), .spaces(2)]
+    return [storage]
```
ahoppen, line 26: What about simplifying this further to

```suggestion
    return ["\n  var _storage: [String: Any] = [:]"]
```

Examples/Sources/MacroExamples/Implementation/Expression/AddBlocker.swift
```diff
@@ -72,17 +72,12 @@ public struct AddBlocker: ExpressionMacro {
...
-                )
-              )
-            )
-          )
+          binOp.operator.tokenKind = .binaryOperator("-")
+
+          var node = node
+          node.operator = ExprSyntax(binOp)
+
+          return ExprSyntax(node)
```
ahoppen, line 80: I think in this case where we really are creating a copy with only one property modified, the `with` syntax reads nicer.

```suggestion
          return ExprSyntax(node.with(\.operator,  ExprSyntax(binOp)))
```

Examples/Sources/MacroExamples/Implementation/Member/MetaEnumMacro.swift
```diff
@@ -22,18 +22,22 @@ public struct MetaEnumMacro {
...
+
+    parentTypeName = enumDecl.name
 
     access = enumDecl.modifiers.first(where: \.isNeededAccessLevelModifier)
 
     childCases = enumDecl.caseElements.map { parentCase in
-      parentCase.with(\.parameterClause, nil)
+      var parentCase = parentCase
+      parentCase.parameterClause = nil
+      return parentCase
```
ahoppen, line 40: Similarly here where we are creating a copy with only one property modified, I think `with` is actually nicer. 
I just think that `with` doesn’t scale if you are modifying nested properties or multiple.

Examples/Sources/MacroExamples/Implementation/Member/MetaEnumMacro.swift
```diff
@@ -22,18 +22,22 @@ public struct MetaEnumMacro {
...
+    guard var enumDecl = declaration.as(EnumDeclSyntax.self) else {
       throw DiagnosticsError(diagnostics: [
         CaseMacroDiagnostic.notAnEnum(declaration).diagnose(at: Syntax(node))
       ])
     }
 
-    parentTypeName = enumDecl.name.with(\.trailingTrivia, [])
+    enumDecl.name.trailingTrivia = []
+
+    parentTypeName = enumDecl.name
```
ahoppen, line 33: Since we don’t actually want to modify the `enumDecl` here, I think `with` reads nicer. Because that’s really what we are doing. Setting the `parentTypeName` to the `enumDecl`’s name but without trailing trivia.

Examples/Sources/MacroExamples/Implementation/MemberAttribute/WrapStoredPropertiesMacro.swift
```diff
@@ -46,14 +46,13 @@ public struct WrapStoredPropertiesMacro: MemberAttributeMacro {
...
-        )
+    var attribute = AttributeSyntax(
+      attributeName: IdentifierTypeSyntax(
+        name: .identifier(wrapperName.content.text)
       )
-      .with(\.leadingTrivia, [.newlines(1), .spaces(2)])
-    ]
+    )
+    attribute.leadingTrivia = [.newlines(1), .spaces(2)]
+    return [attribute]
```
ahoppen, line 55: Similar to above, I think `leadingTrivia` can be part of the constructor call.

Examples/Sources/MacroExamples/Implementation/Peer/AddAsyncMacro.swift
```diff
@@ -42,7 +42,9 @@ public struct AddAsyncMacro: PeerMacro {
     }
 
     // This only makes sense void functions
-    if funcDecl.signature.returnClause?.type.with(\.leadingTrivia, []).with(\.trailingTrivia, []).description != "Void" {
+    guard let returnIdentifier = funcDecl.signature.returnClause?.type.as(IdentifierTypeSyntax.self),
+      returnIdentifier.name.text == "Void"
+    else {
```
ahoppen, line 47: Much nicer. Is there any particular reason why you chose to store the `returnIdentifier` instead of doing

```suggestion
    guard funcDecl.signature.returnClause?.type.as(IdentifierTypeSyntax.self).name.text == "Void" else {
```

Examples/Sources/MacroExamples/Implementation/Peer/AddAsyncMacro.swift
```diff
@@ -58,7 +60,9 @@ public struct AddAsyncMacro: PeerMacro {
     }
 
     // Completion handler needs to return Void
-    if completionHandlerParameter.returnClause.type.with(\.leadingTrivia, []).with(\.trailingTrivia, []).description != "Void" {
+    guard let returnIdentifier = completionHandlerParameter.returnClause.type.as(IdentifierTypeSyntax.self),
+      returnIdentifier.name.text == "Void"
+    else {
```
ahoppen, line 65: Similar here, any reason why you chose to store the `returnIdentifier` in a variable?

Examples/Sources/MacroExamples/Implementation/Peer/AddAsyncMacro.swift
```diff
@@ -121,42 +127,37 @@ public struct AddAsyncMacro: PeerMacro {
...
+    funcDecl.signature.effectSpecifiers = FunctionEffectSpecifiersSyntax(
+      leadingTrivia: .space,
+      asyncSpecifier: .keyword(.async),
+      throwsSpecifier: isResultReturn ? .keyword(.throws) : nil
+    )
+
+    // add result type
+    if var successReturnType {
+      successReturnType.leadingTrivia = .space
+      funcDecl.signature.returnClause = ReturnClauseSyntax(leadingTrivia: .space, type: successReturnType)
```
ahoppen, line 140: Similar argumentation about the create-a-copy-with-one-property-changed paradigm. I would prefer

```suggestion
    if let successReturnType {
      funcDecl.signature.returnClause = ReturnClauseSyntax(leadingTrivia: .space, type: successReturnType.with(\.leadingTrivia, .space))
```

Examples/Sources/MacroExamples/Implementation/Peer/AddCompletionHandlerMacro.swift
```diff
@@ -25,20 +25,22 @@ public struct AddCompletionHandlerMacro: PeerMacro {
...
     }
 
     // This only makes sense for async functions.
     if funcDecl.signature.effectSpecifiers?.asyncSpecifier == nil {
       let newEffects: FunctionEffectSpecifiersSyntax
-      if let existingEffects = funcDecl.signature.effectSpecifiers {
-        newEffects = existingEffects.with(\.asyncSpecifier, .keyword(.async))
+      if var existingEffects = funcDecl.signature.effectSpecifiers {
+        existingEffects.asyncSpecifier = .keyword(.async)
+        newEffects = existingEffects
```
ahoppen, line 37: I would flip this, so we are modifying the new effects because technically we can’t really modify the existing effects (they are the old ones after all).

```suggestion
        newEffects = existingEffects
        newEffects.asyncSpecifier = .keyword(.async)
```

ahoppen (commented):
Thank you for continuing to clean up the macro examples. As was to be expected, I have a couple of opinions on style 😉

Examples/Sources/MacroExamples/Implementation/Expression/FontLiteralMacro.swift
Matejkob: Great idea! One suggestion: what are your thoughts on using `if` instead of `guard` for future considerations as well? I find that avoiding negation makes the code easier to read, rather than using it solely to justify a `guard` statement.

```swift
  if tuple.isEmpty {
    return tuple
  }

  var tuple = tuple
  tuple[tuple.startIndex].label = .identifier(newLabel)
  tuple[tuple.startIndex].colon = .colonToken()
  return tuple
```

Examples/Sources/MacroExamples/Implementation/Expression/AddBlocker.swift
Matejkob: Totally agree with you! Initially, I wasn't sure if you were suggesting changing all occurrences of `with` or just specific ones. But this one is definitely a keeper! :D

Examples/Sources/MacroExamples/Implementation/Peer/AddAsyncMacro.swift
Matejkob: Oops, my brain took a coffee break there! No real reason, just one of those moments. 😄

Examples/Sources/MacroExamples/Implementation/Peer/AddCompletionHandlerMacro.swift
Matejkob: Mea culpa 😅 that definitely makes more sense!

Matejkob:
> Thank you for continuing to clean up the macro examples. As was to be expected, I have a couple of opinions on style 😉

Thank you for the review!

Examples/Sources/MacroExamples/Implementation/Expression/FontLiteralMacro.swift
ahoppen: I don’t have strong opinions here. Pick whichever you like better.

Matejkob: I feel you. I've picked `if`.

ahoppen (approved):

ahoppen:
@swift-ci Please test

PR #2286, opened 2023-10-15
Handle generic parameter on enum case

### LINKED ISSUE #2273: Bad diagnostic for generic parameter list on enum case
### Issue Kind

Bad Diagnostic Produced

### Source Code

```swift
enum Foo {
    case bar<T>(baz: T)
}
```
```
/Users/mac/Desktop/test.swift:106:13: error: consecutive declarations on a line must be separated by newline or ';' [bridged_error]
    case bar<T>(baz: T)
            ^
/Users/mac/Desktop/test.swift:106:13: note: insert newline [bridged_note]
    case bar<T>(baz: T)
            ^
/Users/mac/Desktop/test.swift:106:13: note: insert ';' [bridged_note]
    case bar<T>(baz: T)
            ^
            ;
/Users/mac/Desktop/test.swift:106:13: error: expected 'func' in function [bridged_error]
    case bar<T>(baz: T)
            ^
/Users/mac/Desktop/test.swift:106:13: note: insert 'func' [bridged_note]
    case bar<T>(baz: T)
            ^
            func 
/Users/mac/Desktop/test.swift:106:14: error: unexpected code 'T>' before parameter clause [bridged_error]
    case bar<T>(baz: T)
```

### Description

Twin of https://github.com/apple/swift/issues/69036.

kimdv:
Fixes #2273


Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -1077,6 +1077,24 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
+    }
+
+    if let unexpectedBetweenNameAndParameterClause = node.unexpectedBetweenNameAndParameterClause,
+      let genericParameter = unexpectedBetweenNameAndParameterClause.compactMap({ $0.as(GenericParameterClauseSyntax.self) }).only
+    {
+      addDiagnostic(
+        genericParameter,
+        .genericParamCantBeUsedInEnumCaseDecl,
+        handledNodes: [unexpectedBetweenNameAndParameterClause.id]
+      )
```
kimdv, line 1092: The C++ version have a Fix-Its. 
Not sure if we can do the same as we need to modify the enum decl.

ahoppen: Hmm, it’s not easily possible in the current design where we assume all nodes for the Fix-It to already be present in the syntax tree so that we just need to flip the token presence. 

I would say that it’s good enough for now but adding support for these kinds of Fix-Its might be useful in general. What we need for that is, IMO, a new `Fixit.Change` that takes a syntax node, a key path within that node and a replacement node, i.e. something like `case replace<NodeType, ChildType>(node: NodeType, path: KeyPath<NodeType, ChildType>, child: ChildType`) – and yes I see the irony of declaring a generic enum case here, you would need to come up with some design that actually compiles, I haven’t thought about it too much yet.

I think this kind of Fix-It change might also be useful for macro authors. IIRC it came up once or twice but was never super important.

Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift
```diff
@@ -162,6 +162,9 @@ extension DiagnosticMessage where Self == StaticParserError {
   public static var forbiddenInterpolatedString: Self {
     return .init("argument cannot be an interpolated string literal")
   }
+  public static var genericParamCantBeUsedInEnumCaseDecl: Self {
+    return .init("generic signature cannot be declared in enum 'case'. did you mean to attach it to enum declaration?")
```
ahoppen, line 166: Since we don’t offer a Fix-It

```suggestion
    return .init("generic signature cannot be declared in enum 'case'")
```

Tests/SwiftParserTest/DeclarationTests.swift
```diff
@@ -2934,4 +2934,117 @@ final class DeclarationTests: ParserTestCase {
...
     )
   }
+
+  // https://github.com/apple/swift-syntax/issues/2273
+  func testEnumCaseWithGenericParameter() {
+    assertParse(
+      """
+      enum Foo {
+          case foo1️⃣<T>(T)
+      }
```
ahoppen, line 2944: Nitpick: Could you indent by 2 space here to make it look consistent?

Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift
ahoppen: You still need to update the tests

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
kimdv: Let's do that in another PR!

Opened an issue: https://github.com/apple/swift-syntax/issues/2298

Tests/SwiftParserTest/DeclarationTests.swift
```diff
@@ -2934,4 +2934,117 @@ final class DeclarationTests: ParserTestCase {
...
+    assertParse(
+      """
+      enum Foo {
+        case foo1️⃣<T>(T)
+      }
+      """,
+      diagnostics: [
+        DiagnosticSpec(
+          locationMarker: "1️⃣",
+          message: "generic signature cannot be declared in enum 'case'."
```
ahoppen, line 2949: AFAIK, none of the other error messages end with a period and neither should this.

ahoppen (approved):

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test windows

kimdv:
@ahoppen I totally forgot to ask: Should I add this to release notes?
Same gores for the Fit-It PR?

PR #2290, opened 2023-10-16
Update the interpolation failure message

ahoppen:
When a user is getting the interpolation `os_log` failure message but actually supports parsing invalid source code, offer them an alternative parses the source code without checking for errors.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #2291, opened 2023-10-16
Fix missing space in LabeledExprSyntax init

rauhul:
Updates LabeledExprSyntax convenience initializer to include a trailing space trivia after the colon if an argument label is specified. This ensures the rendered description is 'arg: value' instead of 'arg:value'.

### DIFF
--- a/Tests/SwiftSyntaxBuilderTest/Assertions.swift
+++ b/Tests/SwiftSyntaxBuilderTest/Assertions.swift
@@ -23,7 +23,8 @@ func assertBuildResult<T: SyntaxProtocol>(
   file: StaticString = #file,
   line: UInt = #line
 ) {
-  var buildableDescription = format
+  var buildableDescription =
+    format
     ? buildable.formatted().description
     : buildable.description
   var expectedResult = expectedResult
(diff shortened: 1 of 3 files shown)


bnbarham:
@swift-ci please test

Tests/SwiftSyntaxBuilderTest/Assertions.swift
```diff
@@ -19,10 +19,13 @@ func assertBuildResult<T: SyntaxProtocol>(
...
   _ expectedResult: String,
   trimTrailingWhitespace: Bool = true,
+  format: Bool = true,
   file: StaticString = #file,
   line: UInt = #line
 ) {
-  var buildableDescription = buildable.formatted().description
+  var buildableDescription = format
+    ? buildable.formatted().description
+    : buildable.description
```
bnbarham, line 28: @ahoppen do you remember why you changed this to format by default? Makes all these basic format tests too, which maybe we want, but it does end up meaning we miss things like this.

ahoppen: I don’t remember. But I believe it was something along the lines of: You are using `SwiftSyntaxBuilder` to build your syntax tree and if you are doing that, it’s expected that you call `.formatted` in the end go get something that actually parses again, so that’s what we’re testing here.

rauhul:
@swift-ci please test

rauhul:
@swift-ci please test windows

ahoppen (approved):

bnbarham (approved):

PR #2297, opened 2023-10-16
[5.10][CMake] Remove set(CACHE) line for SWIFT_MODULE_ABI_NAME_PREFIX

rintaro:
Cherry-pick https://github.com/apple/swift-syntax/pull/2296 into release/5.10

Due to https://cmake.org/cmake/help/latest/policy/CMP0126.html set(CACHE) overwrites the existing value if the cache has not been set to any value. Since we don't need any default value, just remove it.


rintaro:
@swift-ci Please test

PR #2299, opened 2023-10-17
Add incremental parse function that return a new IncrementalParseResult

### LINKED ISSUE #2267: Change `parseIncrementally` to return a struct instead of a tuple
From https://github.com/apple/sourcekit-lsp/pull/857#discussion_r1349304038

> Is there a reason `Parser.parseIncrementally` returns a tuple instead of a dedicated struct? Seems like a single type would be easier to manage, and would help ensure you don't pass a mismatching syntax tree + lookahead ranges to `IncrementalParseTransition` (and would make it trivial to add any additional state later down the road if needed).

Like with all API changes, we need to see if we can do this in an API-compatible way. I think breaking the API here might be acceptable because the user base of this function should be fairly small, but an API-compatible way would definitely be preferred.

kimdv:
Fixes #2267

### DIFF
--- a/Sources/SwiftParser/ParseSourceFile.swift
+++ b/Sources/SwiftParser/ParseSourceFile.swift
@@ -13,6 +13,14 @@
 @_spi(RawSyntax) import SwiftSyntax
 
 extension Parser {
+  public struct IncrementalParseResult {
+    /// The parsed tree.
+    public let tree: SourceFileSyntax
+    /// ``LookaheadRanges`` that describe how far the parser looked ahead while parsing a node,
+    /// which is necessary to construct an ``IncrementalParseTransition`` for a subsequent incremental parse.
+    public let lookaheadRanges: LookaheadRanges
+  }
+
   /// Parse the source code in the given string as Swift source file. See
   /// `Parser.init` for more details.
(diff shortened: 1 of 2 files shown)


Sources/SwiftParser/ParseSourceFile.swift
```diff
@@ -62,12 +70,42 @@ extension Parser {
...
+  /// the post-edit source and that parse transition, the parser will re-use
+  /// nodes that haven’t changed.
+  ///
+  /// - Parameters:
+  ///   - source: The source code to parse
+  ///   - parseTransition: If a similar source file has already been parsed, the
+  ///     ``IncrementalParseTransition`` that contains the previous tree as well
+  ///     as the edits that were performed to it.
+  /// - Returns: The ``IncrementalParseResult``.
+  public static func parseIncrementally(
```
kimdv, line 99: okay this will break as is, as the compiler cannot figure out what function to use. 

We could, for now, say that it should use `@_disfavoredOverload` on the new one if we want.

Then people can transition

ahoppen:
@divalue is already working on this here: https://github.com/apple/swift-syntax/pull/2299

PR #2305, opened 2023-10-18
Add extension macro example

Matejkob:
This PR introduces a new macro called `ProtocolDefaultImplMacro` that automates the generation of default method implementations within protocol extensions. By attaching this macro to any given protocol, it will automatically generate extension code blocks filled with default implementations.

The inspiration for adding this new macro example originates from this discussion on the Swift forums: [Workaround for macros not allowed to add extensions?](https://forums.swift.org/t/workaround-for-macros-not-allowed-to-add-extensions/67916).

### DIFF
--- a/Examples/Sources/MacroExamples/Interface/ExtensionMacros.swift
+++ b/Examples/Sources/MacroExamples/Interface/ExtensionMacros.swift
@@ -14,3 +14,43 @@
 
 @attached(extension, conformances: Equatable)
 public macro equatable() = #externalMacro(module: "MacroExamplesImplementation", type: "EquatableExtensionMacro")
+
+// MARK: - Default Fatal Error Implementation
+
+/// A macro that provides default `fatalError` implementations for protocol methods.
+///
+/// This macro generates Swift extensions for the attached protocol,
+/// adding a default `fatalError` implementation for each method defined within it.
+///
+/// ## Example usage:
+/// ```swift
+/// @defaultFatalErrorImplementation
+/// protocol MyProtocol {
+///   func someMethod()
+/// }
+/// ```
+///
+/// The generated code would look like:
+/// ```swift
+/// protocol MyProtocol {
+///   func someMethod()
+/// }
+///
+/// extension MyProtocol {
+///   func someMethod() {
+///     fatalError("whoops 😅")
+///   }
+/// }
+/// ```
+///
+/// ## Edge Cases
+/// - **No Methods in Protocol**: If the protocol does not contain any methods,
+///   the macro will not generate an empty extension.
+/// - **Incorrect Attachment**: If the macro is attached to a non-protocol declaration,
+///   it will produce an error diagnostic stating that it can only be applied to a protocol.
+@attached(extension, names: arbitrary)
+public macro defaultFatalErrorImplementation() =
+  #externalMacro(
+    module: "MacroExamplesImplementation",
+    type: "DefaultFatalErrorImplementationMacro"
+  )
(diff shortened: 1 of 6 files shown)


Examples/Sources/MacroExamples/Implementation/Extension/ProtocolDefaultImplMacro.swift
```diff
@@ -0,0 +1,50 @@
...
+  public static func expansion(
+    of node: AttributeSyntax,
+    attachedTo declaration: some DeclGroupSyntax,
+    providingExtensionsOf type: some TypeSyntaxProtocol,
+    conformingTo protocols: [TypeSyntax],
+    in context: some MacroExpansionContext
+  ) throws -> [ExtensionDeclSyntax] {
+    // Ensure that the declaration is a protocol declaration
+    guard let protocolDecl = declaration.as(ProtocolDeclSyntax.self) else {
+      return []
```
ahoppen, line 27: We should emit an error saying that the macro can only be applied to protocols instead of silently doing nothing.

Examples/Sources/MacroExamples/Implementation/Extension/ProtocolDefaultImplMacro.swift
```diff
@@ -0,0 +1,50 @@
...
+    var methods = protocolDecl.memberBlock.members
+      .map(\.decl)
+      .compactMap { $0.as(FunctionDeclSyntax.self) }
+
+    // Add implementation to all methods from the protocol
+    methods.indices.forEach { index in
+      methods[index].body = CodeBlockSyntax {
+        ExprSyntax(#"fatalError("whoops 😅")"#)
+      }
+    }
```
ahoppen, line 40: You should be able to simplify this further. Also, usage of `forEach` is generally considered non-idiomatic Swift.

```suggestion
    var methods = protocolDecl.memberBlock.members
      .map(\.decl)
      .compactMap { 
        guard var function = $0.as(FunctionDeclSyntax.self) else {
          return nil
        }
        function.body = CodeBlockSyntax {
          ExprSyntax(#"fatalError("whoops 😅")"#)
        }
        return function
      }
```

Examples/Sources/MacroExamples/Interface/ExtensionMacros.swift
```diff
@@ -14,3 +14,11 @@
 
 @attached(extension, conformances: Equatable)
 public macro equatable() = #externalMacro(module: "MacroExamplesImplementation", type: "EquatableExtensionMacro")
+
+// MARK: - Protocol Default Impl
+
+@attached(extension, names: arbitrary)
+public macro defaultImpl() = #externalMacro(
```
ahoppen, line 21: I would like to avoid `Impl` here since it’s an abbreviation. What about `defaultFatalErrorImplementation`?

ahoppen (commented):
I think this is a good example of a slightly more complicated extension macro and would make sense to add.

Examples/Sources/MacroExamples/Implementation/Extension/ProtocolDefaultImplMacro.swift
Matejkob: done

Examples/Sources/MacroExamples/Implementation/Extension/ProtocolDefaultImplMacro.swift
Matejkob: Oh, that's much simpler—done 😅

Examples/Sources/MacroExamples/Interface/ExtensionMacros.swift
Matejkob: I always struggle with names; thank you for the suggestion. I've renamed macro to `DefaultFatalErrorImplementationMacro` and changed the attribute name to `defaultFatalErrorImplementation`.

ahoppen (approved):
Thank you. Let’s get it in.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #2306, opened 2023-10-20
nonisolated(unsafe) to opt out of strict concurrency static checking for global variables

sophiapoirier:


sophiapoirier:
https://github.com/apple/swift/pull/69280

@swift-ci please test

ahoppen (commented):
Implementation looks good to me. A test case for something like `nonisolated(safe)` would be nice as well to make sure we don’t parse arbitrary detail text.

On a general level: Should this feature be behind an experimental flag?

sophiapoirier:
> Implementation looks good to me. A test case for something like `nonisolated(safe)` would be nice as well to make sure we don’t parse arbitrary detail text.

Thanks, that's a good point, and unfortunately any argument is accepted with this current implementation. Which is particularly interesting to me because I noticed that in `Modifiers.swift`, the `unowned` attribute is implemented to expect `set` as a parenthetical argument even though, as far as I can tell from language documentation, `unsafe` is its potential optional argument. Given that I refactored to use the same optional argument parsing logic that `unowned` is using, it would seem there is an existing problem there. Would it be okay with you if I addressed this in a follow-up PR? This pair of PRs is currently blocking an evolution proposal review that there is pressure to begin soon.

> On a general level: Should this feature be behind an experimental flag?

The functionality is behind the feature flag `GlobalConcurrency` in the compiler's type checker. Is this sufficient, or should some guard be added to the parser here as well?

Sources/SwiftParser/Modifiers.swift
```diff
@@ -98,9 +99,9 @@ extension Parser {
 }
 
 extension Parser {
-  mutating func parseModifierDetail() -> RawDeclModifierDetailSyntax {
+  mutating func parseModifierDetail(_ keyword: Keyword) -> RawDeclModifierDetailSyntax {
     let (unexpectedBeforeLeftParen, leftParen) = self.expect(.leftParen)
-    let (unexpectedBeforeDetailToken, detailToken) = self.expect(.identifier, TokenSpec(.set, remapping: .identifier), default: .identifier)
+    let (unexpectedBeforeDetailToken, detailToken) = self.expect(.identifier, TokenSpec(keyword, remapping: .identifier), default: .identifier)
```
ahoppen, line 104: > Thanks, that's a good point, and unfortunately any argument is accepted with this current implementation. Which is particularly interesting to me because I noticed that in `Modifiers.swift`, the `unowned` attribute is implemented to expect `set` as a parenthetical argument even though, as far as I can tell from language documentation, `unsafe` is its potential optional argument. Given that I refactored to use the same optional argument parsing logic that `unowned` is using, it would seem there is an existing problem there. Would it be okay with you if I addressed this in a follow-up PR? This pair of PRs is currently blocking an evolution proposal review that there is pressure to begin soon.

Oh, I missed that. In that case you don’t need to change `parseModifierDetail` at all. We only need to check for `set` specifically because it’s a lexer-classified keyword and thus not an identifier. `unsafe` is an identifier at this point still and will only be changed to be a keyword once we expect it.

https://github.com/apple/swift-syntax/blob/f8be751eaefe0d7c72c021f12024a8328a5f3568/Sources/SwiftParser/generated/IsLexerClassified.swift#L18-L20

ahoppen (approved):
> The functionality is behind the feature flag `GlobalConcurrency` in the compiler's type checker. Is this sufficient, or should some guard be added to the parser here as well?

It should be guarded by a feature flag as well because you can use `SwiftParser` as a standalone module, which means that you don’t get any of the checks from the type checker.

> Would it be okay with you if I addressed this in a follow-up PR? This pair of PRs is currently blocking an evolution proposal review that there is pressure to begin soon.

I’m fine with merging this now and addressing the comments in a follow-up PR.

PR #2314, opened 2023-10-27
Add edits to FixIt for sourcekit-lsp to access

joehsieh:
sourcekit-lsp would use the info provided by swift-syntax to get text edits for the FixIt.Change

https://github.com/apple/sourcekit-lsp/issues/909#issuecomment-1779622942


ahoppen (commented):
I’ve got a couple of high-level comments:
1. Instead of making `FixIt.Change.Edit` public, I think it would be good to re-use `SourceEdit` in the `SwiftRefactor` module. The way I would do this, is to 
  - Move `SourceEdit` into the `SwiftDiagnostics` module
  - Make `SwiftRefactor` depend on `SwiftDiagnostics`
  - Define a deprecated typealias in `SwiftRefactor` for `SourceEdit` so we don’t break clients that use `SwiftRefactor.SourceEdit`
```swift
@available(*, deprecated, message: "SourceEdit has been moved to the SwiftDiagnostics module")
public typealias SourceEdit = SwiftDiagnsotics.SourceEdit
```
2. The API I would like to offer is to have a member `edits` on `FixIt` that exposes all the edits. That way clients don’t need to worry about the `FixIt.Change` details.
3. I think it would also be good if `FixIt.edits` would detect if any of the changes have overlapping edits. I don’t remember the case that hits this but you should be able to find an example by setting a breakpoint [here](https://github.com/apple/swift-syntax/blob/e5962c445dcbbb80f0cce49abee6c8a126875ba7/Sources/_SwiftSyntaxTestSupport/FixItApplier.swift#L71) and running the `SwiftParserTests`.

joehsieh:
Thanks for these details! It helped me a lot.

I updated the code per this comment.

Sources/SwiftDiagnostics/FixIt.swift
```diff
@@ -46,3 +46,48 @@ public struct FixIt {
...
+  var edits: [SourceEdit] {
+    var existingEdits = [SourceEdit]()
+    for change in changes {
+      let edit = edit(from: change)
+      var isOverlapping = false
+      for existingEdit in existingEdits {
+        // The edit overlaps with the previous edit. We can't apply both
+        // without conflicts. Apply the one that's listed first and drop the
+        // later edit.
+        if edit.range.overlaps(existingEdit.range) {
```
joehsieh, line 60: [Question] the ranges of edits in a FixIt wouldn't overlap with each other, but the range of edits in multiple FixIts still possibly overlap each other. i was wondering if it's expected in our case.

Sources/SwiftDiagnostics/FixIt.swift
```diff
@@ -46,3 +46,48 @@ public struct FixIt {
...
+          isOverlapping = true
+          break
+        }
+      }
+      if !isOverlapping {
+        existingEdits.append(edit)
+      }
```
ahoppen, line 67: I think you should be able to write this as 

```suggestion
      let isOverlapping = existingEdits.contains { edit.range.overlaps($0.range) }
      if !isOverlapping {
        // The edit overlaps with the previous edit. We can't apply both
        // without conflicts. Apply the one that's listed first and drop the
        // later edit.
        existingEdits.append(edit)
      }
```

Sources/SwiftDiagnostics/FixIt.swift
```diff
@@ -46,3 +46,48 @@ public struct FixIt {
...
+    }
+    return existingEdits
+  }
+
+  private func edit(from change: Change) -> SourceEdit {
```
ahoppen, line 72: You could make this a `private` extension on `SourceEdit`. I think it reads nicer that way.

Sources/SwiftDiagnostics/SourceEdit.swift
ahoppen, line 1: @bnbarham and I just discussed about the placement of this file and agree that it would fit better in the `SwiftSyntax` module. Could you move it there?

Sources/SwiftDiagnostics/SourceEdit.swift
```diff
@@ -0,0 +1,62 @@
+import SwiftSyntax
```
ahoppen, line 1: Could you add the copyright header?

Tests/SwiftDiagnosticsTest/FixItTests.swift
```diff
@@ -0,0 +1,37 @@
+import SwiftParser
```
ahoppen, line 1: Could you add a copyright header here as well?

Tests/SwiftDiagnosticsTest/FixItTests.swift
```diff
@@ -0,0 +1,37 @@
...
+        [(9..<15, "Multiident "), (15..<21, "")],
+        [(9..<15, "MultiIdent "), (15..<21, "")]
+      ]
+
+    let markedSource = "protocol Multi 1️⃣ident {}"
+    let (_, source) = extractMarkers(markedSource)
+    let tree = Parser.parse(source: source)
+    let diags = ParseDiagnosticsGenerator.diagnostics(for: tree)
+    XCTAssertEqual(diags.count, 1)
+    let diag = diags.first!
```
ahoppen, line 19: Let’s just `XCTUnwrap`. That way the test execution doesn’t crash if `diags` is empty.

```suggestion
    let diag = try XCTUnwrap(diags.first)
```

Tests/SwiftDiagnosticsTest/FixItTests.swift
```diff
@@ -0,0 +1,37 @@
...
+    XCTAssertEqual(diag.fixIts.count, 2)
+
+    for i in 0..<diag.fixIts.count {
+      let fixIt = diag.fixIts[i]
+      let changes = fixIt.changes
+      let edits = fixIt.edits
+      XCTAssertNotEqual(changes.count, edits.count)
+      let expctedEdits = testCases[i]
```
ahoppen, line 27: This will crash if we end up having more than 2 Fix-Its, which makes it harder to debug the tests. Since both changes are pretty similar, I think it would be OK to just test the changes of the first Fix-It here. I can’t imagine any scenario where testing the second would catch an issue that we didn’t find with the first Fix-It.

Tests/SwiftDiagnosticsTest/FixItTests.swift
```diff
@@ -0,0 +1,37 @@
+import SwiftParser
+import SwiftParserDiagnostics
+import SwiftSyntax
+import XCTest
+import _SwiftSyntaxTestSupport
+
+final class FixItTests: XCTestCase {
+  func testEditsForFixIt() {
+    let testCases: [[(range: Range<Int>, replacement: String)]] = [
+        [(9..<15, "Multiident "), (15..<21, "")],
```
ahoppen, line 10: Instead of using hard offsets here, I would prefer to use location markers in the `markedSource`. To get `9`, you can add the location marker `1️⃣` at after `protocol ` and then do 

```swift
let (markers, source) = extractMarkers(markedSource)
let something = markers["1️⃣"] 
```

The advantage of using location markers is that
1. You can easily tell where in the input source code they correspond to without counting characters
2. You don’t need to fix up a bunch of offsets when eg. changing `protocol` to `struct` at some point. It’s just more stable that way.

Tests/SwiftDiagnosticsTest/FixItTests.swift
```diff
@@ -0,0 +1,37 @@
...
+      let edits = fixIt.edits
+      XCTAssertNotEqual(changes.count, edits.count)
+      let expctedEdits = testCases[i]
+      XCTAssertEqual(expctedEdits.count, edits.count)
+      for j in 0..<edits.count {
+        let edit = edits[j]
+        let expectedEdit = expctedEdits[j]
+        XCTAssertEqual(edit.range, AbsolutePosition(utf8Offset: expectedEdit.range.lowerBound)..<AbsolutePosition(utf8Offset: expectedEdit.range.upperBound))
+        XCTAssertEqual(edit.replacement, expectedEdit.replacement)
+      }
```
ahoppen, line 34: Shouldn’t you be able to just compare the actual edits with the expected edits using `XCTAssertEqual`?

Tests/SwiftDiagnosticsTest/FixItTests.swift
joehsieh: my first time to use XCTUnwrap, thanks for sharing.

Tests/SwiftDiagnosticsTest/FixItTests.swift
joehsieh: true, let me update the code to avoid the crash if test is failed.

Tests/SwiftDiagnosticsTest/FixItTests.swift
joehsieh: my bad. let me fix it.

Tests/SwiftDiagnosticsTest/FixItTests.swift
joehsieh: wow, that's a great design. i understand how it works. thanks!

Sources/SwiftDiagnostics/FixIt.swift
ahoppen: Sorry, I meant to make it `private` extension on `FixIt.Change` so you can write `change.edit` instead of `SourceEdit.edit(from: change)`. Or, alternatively, we could have an initializer on `SourceEdit` that takes a `FixIt.Change`. I don’t have strong opinions so whatever you prefer.

Tests/SwiftDiagnosticsTest/FixItTests.swift
```diff
@@ -0,0 +1,42 @@
...
+import SwiftParserDiagnostics
+import SwiftSyntax
+import XCTest
+import _SwiftSyntaxTestSupport
+
+final class FixItTests: XCTestCase {
+  func testEditsForFixIt() throws {
+    let markedSource = "protocol 0️⃣Multi 1️⃣ident 2️⃣{}"
+    let (markers, source) = extractMarkers(markedSource)
+    let positions = ["0️⃣", "1️⃣", "2️⃣"].compactMap { markers[$0] }
```
ahoppen, line 23: A few nitpicks:
- We start counting at `1` for the location markers in the other test cases, would be nice to be consistent here as well.
- Instead of using the `compactMap` here and then doing `positions[<integer>]` below, I would prefer to use `marker["1️⃣"]` below. Again just to be consistent with the other test cases.

Sources/SwiftDiagnostics/FixIt.swift
```diff
@@ -46,3 +46,44 @@ public struct FixIt {
     self.changes = changes
   }
 }
+
+public extension FixIt {
+  var edits: [SourceEdit] {
```
ahoppen, line 51: Could you add a doc comment for `edits`?

--- 

Also a nitpick: I would prefer `extension` and `public var` over `public extension` and `var` because it makes it obvious that the variable is part of the public API. With `public extension` it’s quite easy to add a new member without a type annotation to the extension which then becomes part of the public API without notice.

ahoppen (commented):
Thanks for addressing my feedback, @joehsieh. I’ve got a couple more comments inline but they are pretty narrow in scope. The overall shape looks great

ahoppen:
Oh, and could you also add two entries in `Release Notes/511.md`? One for the addition of `FixIt.edits` and one for move of `SourceEdit` to SwiftSyntax?

Tests/SwiftDiagnosticsTest/FixItTests.swift
joehsieh: compactMap here is to avoid force-unwrapping marker value. i.e. marker["1️⃣"]!, but you're right, marker["1️⃣"] is more elegant. Let me force-unwrap the value just like what we did [here](https://github.com/apple/swift-syntax/blob/06b417950cdbe7f5591bb6849bf36c40fcfe69ad/Tests/SwiftDiagnosticsTest/GroupDiagnosticsFormatterTests.swift#L85).

Sources/SwiftDiagnostics/FixIt.swift
joehsieh: yeah, i'll add a doc comment and update the code due to above reason.

ahoppen (approved):
Thank you! Looks good to me. Let’s get this in 🚢

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
You need to add FixIt.swift to https://github.com/apple/swift-syntax/blob/main/Sources/SwiftSyntax/CMakeLists.txt.

joehsieh:
> You need to add FixIt.swift to https://github.com/apple/swift-syntax/blob/main/Sources/SwiftSyntax/CMakeLists.txt.

Sorry, i just noticed this error in Jenkins. let me add the missing file name to the CMakeLists.txt
```
Shell Script - (12 min in self) | export PATH=/usr/local/bin/:/usr/bin:/bin:/usr/sbin:/sbin && /Users/ec2-user/jenkins/workspace/swift-syntax-PR-macOS/branch-main/swift/utils/build-script --preset=buildbot_swiftsyntax_macos --cmake-c-launcher /usr/local/bin/sccache --cmake-cxx-launcher /usr/local/bin/sccache
-- | --


[Shell Script - (12 min in self)](https://ci.swift.org/job/swift-syntax-PR-macOS/3579/execution/node/48/)	export PATH=/usr/local/bin/:/usr/bin:/bin:/usr/sbin:/sbin && /Users/ec2-user/jenkins/workspace/swift-syntax-PR-macOS/branch-main/swift/utils/build-script --preset=buildbot_swiftsyntax_macos --cmake-c-launcher /usr/local/bin/sccache --cmake-cxx-launcher /usr/local/bin/sccache
```

ahoppen:
I didn’t notice it before CI ran either. Let’s try again.

@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
One more thing: Looks like the code isn’t formatted according to the style guide. Could you format it as described here: https://github.com/apple/swift-syntax/blob/main/CONTRIBUTING.md#formatting?

joehsieh:
Thanks for your comments! I learned a lot from them.

joehsieh:
to confirm, could i run `@swift-ci Please test`  by myself?

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

joehsieh:
let me know if there is a way to fix the test for windows platform.
i don't have a Windows laptop for testing tho.

ahoppen:
Windows failure is unrelated. We are trying to figure out what’s causing the issue.

ahoppen:
OK, Windows CI should work again. Let’s try again

@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

PR #2316, opened 2023-10-29
[SwiftParser] Support function body and type member skipping

rintaro:
* Introduce `ParsingOptions` option-set type with `bodySkipping` member, and add `options: ParsingOptions` to all `Parser.parse(...)` functions
* Skipped body is represented as a newly introduced `SkippedDeclSyntax`
  * `SkippedDeclSyntax` contains only one `.unknown` token that covers the whole skipped characters including trivia.
* Function body can be skipped as long as:
  * it does _not_ contain nested type decls
* Type member block can be skipped as long as:
  * it does _not_ contain nested class decls
  * it does _not_ contain operator functions
  * it does _not_ contain `#if` directives


Sources/SwiftSyntax/generated/ChildNameForKeyPath.swift
```diff
@@ -2791,6 +2791,12 @@ public func childName(_ keyPath: AnyKeyPath) -> String? {
     return "closingQuote"
   case \SimpleStringLiteralExprSyntax.unexpectedAfterClosingQuote:
     return "unexpectedAfterClosingQuote"
+  case \SkippedDeclSyntax.unexpectedBeforeText:
+    return "unexpectedBeforeText"
```
rintaro, line 2795: Can we remove these `unexpected` items?

Sources/SwiftParser/Lexer/Cursor.swift
```diff
@@ -1236,10 +1236,8 @@ extension Lexer.Cursor {
 
         fallthrough
       default:
-        if let peekedScalar = start.peekScalar(), peekedScalar.isValidIdentifierStartCodePoint {
-          break
-        }
-        if let peekedScalar = start.peekScalar(), peekedScalar.isOperatorStartCodePoint {
+        if let peekedScalar = start.peekScalar(),
+           peekedScalar.isValidIdentifierStartCodePoint || peekedScalar.isOperatorStartCodePoint {
```
rintaro, line 1240: unrelated?

Sources/SwiftParser/DelayedParsing.swift
```diff
@@ -0,0 +1,116 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+@_spi(RawSyntax) import SwiftSyntax
+
+extension Parser {
+  mutating func skippedBraceBody(unless: Lookahead.SkipBodyResult) -> RawSkippedDeclSyntax? {
+    return self.withLookahead { lookahead in
+      let result = lookahead.advanceUntilMatchingRightBrace()
+      guard result.isDisjoint(with: unless) else {
```
rintaro, line 19: Instead of checking the result on the skipped area, the skipping function should bail out and return `nil` as soon as it finds any `unless` condition.

Sources/SwiftParser/DelayedParsing.swift
```diff
@@ -0,0 +1,117 @@
...
+
+      // '#' directives.
+      if tokenKind == .poundIf || tokenKind == .poundSourceLocation {
+        return false
+      }
+
+      // Nested class. Because?
+      if tokenKind == .keyword && tokenText == "class" {
+        return false
+      }
```
hamishknight, line 43: It's because of AnyObject lookup :shakesfist:

PR #2325, opened 2023-11-02
build: emit additional content even in sub-build mode

compnerd:
Enable installing the additional content of the swiftmodules and cmake when built under the Swift compiler. This allows us to build the LSP against the same copy of Swift Syntax as the compiler, shaving ~6MiB off of SourceKit-LSP.

### DIFF
--- a/cmake/modules/CMakeLists.txt
+++ b/cmake/modules/CMakeLists.txt
@@ -1,5 +1,3 @@
-if(PROJECT_IS_TOP_LEVEL)
-  export(EXPORT SwiftSyntaxTargets
-         FILE ${CMAKE_CURRENT_BINARY_DIR}/SwiftSyntaxConfig.cmake
-         NAMESPACE SwiftSyntax::)
-endif()
+export(EXPORT SwiftSyntaxTargets
+       FILE ${CMAKE_CURRENT_BINARY_DIR}/SwiftSyntaxConfig.cmake
+       NAMESPACE SwiftSyntax::)
--- a/cmake/modules/AddSwiftHostLibrary.cmake
+++ b/cmake/modules/AddSwiftHostLibrary.cmake
@@ -124,22 +124,20 @@ function(add_swift_syntax_library name)
     endif()
   endif()
 
-  if(PROJECT_IS_TOP_LEVEL)
-    # Install this target
-    install(TARGETS ${name}
-      EXPORT SwiftSyntaxTargets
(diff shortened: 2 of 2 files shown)


cmake/modules/CMakeLists.txt
```diff
@@ -1,5 +1,3 @@
-if(PROJECT_IS_TOP_LEVEL)
-  export(EXPORT SwiftSyntaxTargets
-         FILE ${CMAKE_CURRENT_BINARY_DIR}/SwiftSyntaxConfig.cmake
-         NAMESPACE SwiftSyntax::)
-endif()
+export(EXPORT SwiftSyntaxTargets
+       FILE ${CMAKE_CURRENT_BINARY_DIR}/SwiftSyntaxConfig.cmake
+       NAMESPACE SwiftSyntax::)
```
rintaro, line 3: I feel `SwiftSyntaxConfig.cmake` should be exported from swift repo's CMakeLists.txt.
Could you add `EXPORT` to 
https://github.com/apple/swift/blob/3094e661f51c9190ba9da2c3ef60cff80e28fa8d/lib/CMakeLists.txt#L32-L35
https://github.com/apple/swift/blob/3094e661f51c9190ba9da2c3ef60cff80e28fa8d/lib/CMakeLists.txt#L60-L66
and put `export(EXPORT)` there?

compnerd: I don't think that works. We are using `FetchContent` rather than `add_subdirectory` and as a result, this is a fully isolated build and thus the `install` rules are not evaluated as part of the Swift repository.  I don't think that trying to reproduce something like this by hand makes sense.

rintaro: I tried this patch https://github.com/rintaro/swift/commit/2a6d5b3a2b10ebca9a572486f59beb4e6f33cf72 and `${SWIFT_BUILD_DIR}/cmake/modules/SwiftSyntaxConfig.cmake` seems to be contain the correct information.

compnerd: Ah, that is a clever idea!  I think that I can make this work, and I assume that you are okay with installing the swiftmodules and import libraries as well from the Swift toolchain?

compnerd: Seems that I was wrong! I was mixing up the `FetchContent_MakeAvailable` and `FetchContent_Provide`.  So, I think that you are right we can just hoist the export.  Thanks for the idea!

compnerd: Yeap, this works to build!  Just verified locally.

PR #2329, opened 2023-11-05
Add fix-it handling tests for macro examples

Matejkob:
It introduces an enhancement to our macro expansion functionality with the addition of fix-it tests.

In `AddBlockerTests`, a new test case has been added to validate the scenario where an addition operation should be a subtraction, including a diagnostic message and the application of the suggested fix-it.

Similarly, `AddCompletionHandlerMacroTests` is now equipped with a test case that applies a fix-it for adding the 'async' keyword to functions that are missing it but have the `@AddCompletionHandler` attribute.

### DIFF
--- a/Examples/Sources/MacroExamples/Implementation/Peer/AddCompletionHandlerMacro.swift
+++ b/Examples/Sources/MacroExamples/Implementation/Peer/AddCompletionHandlerMacro.swift
@@ -34,9 +34,9 @@ public struct AddCompletionHandlerMacro: PeerMacro {
       var newEffects: FunctionEffectSpecifiersSyntax
       if let existingEffects = funcDecl.signature.effectSpecifiers {
         newEffects = existingEffects
-        newEffects.asyncSpecifier = .keyword(.async)
+        newEffects.asyncSpecifier = .keyword(.async, trailingTrivia: .space)
       } else {
-        newEffects = FunctionEffectSpecifiersSyntax(asyncSpecifier: .keyword(.async))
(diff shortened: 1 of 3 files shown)


Matejkob:
Blocked by #2337

Examples/Sources/MacroExamples/Implementation/Peer/AddCompletionHandlerMacro.swift
```diff
@@ -38,6 +38,7 @@ public struct AddCompletionHandlerMacro: PeerMacro {
       } else {
         newEffects = FunctionEffectSpecifiersSyntax(asyncSpecifier: .keyword(.async))
       }
+      newEffects.asyncSpecifier?.trailingTrivia = [.spaces(1)]
```
ahoppen, line 41: I think it would be cleaner if you set the trailing trivia directly when creating the `async` keyword.

```swift
        newEffects.asyncSpecifier = .keyword(.async, trailingTrivia: .space)
      } else {
        newEffects = FunctionEffectSpecifiersSyntax(asyncSpecifier: .keyword(.async, trailingTrivia: .space))
      }
```

ahoppen (commented):
Thank you!

Examples/Sources/MacroExamples/Implementation/Peer/AddCompletionHandlerMacro.swift
Matejkob: Done

ahoppen (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #2337, opened 2023-11-08
Adjust locations for edits based on detached nodes when computing `fixedSource`

### LINKED ISSUE #2332: Fix-it Application Failure in `assertMacroExpansion` Function
### Description

The fix-it provided by a peer macro is correctly applied by the compiler, yet the `assertMacroExpansion` function fails to apply it appropriately.

### Steps to Reproduce

1. Reference the example Macro from SwiftSyntax repository: https://github.com/apple/swift-syntax/blob/098b177830c0d2e614b3757241b2572d758f9143/Examples/Sources/MacroExamples/Implementation/Peer/AddCompletionHandlerMacro.swift#L18-L164
2. Utilize the macro in the code as shown:
   ```swift
   struct Test {
     @AddCompletionHandler
     func fetchData() -> String {
       return "Hello, World!"
     }
   }
   ```
   The macro generates a diagnostic with a fix-it `add 'async'`.
3. When the fix-it is applied using the compiler, the `async` keyword is correctly inserted before `-> String`.
4. Construct a unit test to validate this scenario:
   ```swift
   func testExpansionOnNonAsyncFunctionAppliesFixIt() {
     assertMacroExpansion(
       """
       struct Test {
         @AddCompletionHandler
         func fetchData() -> String {
           return "Hello, World!"
         }
       }
       """,
       expandedSource: """
         struct Test {
           func fetchData() -> String {
             return "Hello, World!"
           }
         }
         """,
       diagnostics: [
         DiagnosticSpec(
           message: "can only add a completion-handler variant to an 'async' function",
           line: 3,
           column: 3,
           severity: .error,
           fixIts: [FixItSpec(message: "add 'async'")]
         )
       ],
       macros: macros,
       applyFixIts: ["add 'async'"],
       fixedSource: """
       struct Test {
         @AddCompletionHandler
         func fetchData() async -> String {
           return "Hello, World!"
         }
       }
       """,
       indentationWidth: .spaces(2)
     )
   }
   ```
   The expectation is for the test to pass, but it fails. The `fixedSource` string produced by the `assertMacroExpansion` function erroneously manifests as:
   ```swift
   struct Test {
     @AddCompletionHandler
     f() async -> String () -> String {
       return "Hello, World!"
     }
   }
   ```
   Notice the incorrect placement and duplication of the function signature, suggesting that the fix-it application is not handling the source code transformation as intended.

ahoppen:
The Fix-Its returned by the macro are defined in terms of the nodes that it got passed and thus their positions are relative to the start of the detached nodes that get passed to the macro. We need to translate these offsets back to the the offsets in the original source file before computing edits and apply the offset-based edits.

Fixes #2332
rdar://118012820

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -93,7 +93,7 @@ let package = Package(
 
     .target(
       name: "_SwiftSyntaxTestSupport",
-      dependencies: ["SwiftBasicFormat", "SwiftSyntax", "SwiftSyntaxBuilder"]
+      dependencies: ["SwiftBasicFormat", "SwiftSyntax", "SwiftSyntaxBuilder", "SwiftSyntaxMacroExpansion"]
     ),
 
     .testTarget(
--- a/Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
+++ b/Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
@@ -325,7 +325,16 @@ public func assertMacroExpansion(
 
   // Applying Fix-Its
   if let expectedFixedSource = expectedFixedSource {
-    let fixedTree = FixItApplier.applyFixes(from: context.diagnostics, filterByMessages: applyFixIts, to: origSourceFile)
+    let messages = applyFixIts ?? context.diagnostics.compactMap { $0.fixIts.first?.message.message }
+
+    let edits =
+      context.diagnostics
+      .flatMap(\.fixIts)
+      .filter { messages.contains($0.message.message) }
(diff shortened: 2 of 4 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

bnbarham (approved):

PR #2345, opened 2023-11-09
Don't add a space between literals and period in BasicFormat

ahoppen:
Fixes https://github.com/apple/swift-syntax/issues/2344 
rdar://118176218

Issue #2344, opened 2023-11-09
Possible bug with inserting extra space between string literal and dot symbol


mkowalski87:
### Description

Not sure if this is bug or I'm doing some not allowed operation but please take a look.

let's consider such small macro 
```swift
import SwiftCompilerPlugin
import SwiftSyntax
import SwiftSyntaxBuilder
import SwiftSyntaxMacros
import SwiftDiagnostics

public struct TestMacro: PeerMacro {
    public static func expansion(
        of node: SwiftSyntax.AttributeSyntax,
        providingPeersOf declaration: some SwiftSyntax.DeclSyntaxProtocol,
        in context: some SwiftSyntaxMacros.MacroExpansionContext
    ) throws -> [SwiftSyntax.DeclSyntax] {

        guard let declaration = declaration.as(FunctionDeclSyntax.self) else {
            fatalError()
        }
        let body = declaration.body!.statements
        return ["""
                func test1() {
                    \(raw: body)
                }
                """]
    }
}

@main
struct MyMacro: CompilerPlugin {
    let providingMacros: [Macro.Type] = [
        TestMacro.self,
    ]
}
```
and when using in this example
```swift
import MyMacro

struct Struct {
    @TestMacro
    func method() {
        let x = 100
        "Swift Macro".lowercased()
        let y = 123.byteSwapped
    }
}
```
macro is expanded to this code
```swift
func test1() {

        let x = 100
        "Swift Macro" .lowercased()
        let y = 123.byteSwapped
}
```
There is extra space after string literal `Swift Macro` and before `.lowercased()`. As you can notice for int literal in next line  it is working fine.

swift-syntax 509.0.2

Here is link to full macro project.
https://github.com/mkowalski87/swift-macro-issue

ahoppen:
Tracked in Apple’s issue tracker as rdar://118176218

ahoppen:
That was indeed an issue with `BasicFormat`. I’m fixing it in https://github.com/apple/swift-syntax/pull/2345.

PR #2348, opened 2023-11-09
[509] Don't add a space between literals and period in BasicFormat

ahoppen:
* **Explanation**: `SwiftBasicFormat` was adding a space between string/regex literals and period. This caused macros to generate code like `"hello" .lowercased()`, which didn’t compile because the spaced on the left- and right-hand side of the period must be balanced
* **Risk**: Low
* **Testing**: Added new test cases
* **Issue**:  rdar://118176218 (#2344)
* **Reviewer**:  @bnbarham on https://github.com/apple/swift-syntax/pull/2345


ahoppen:
✅ Local testing passed

bnbarham (approved):

PR #2351, opened 2023-11-11
Add SyntaxEnum per base syntax kind

rintaro:
Add `DeclSyntaxEnum`, `StmtSyntaxEnum`, `ExprSyntaxEnum`, `TypeSyntaxEnum` and `PatternSyntaxEnum`. These are useful for switch over all nodes of specific syntax kind.

### DIFF
--- a/CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxEnumFile.swift
+++ b/CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxEnumFile.swift
@@ -57,4 +57,50 @@ let syntaxEnumFile = SourceFileSyntax(leadingTrivia: copyrightHeader) {
       }
     }
   }
+
+  for base in SYNTAX_NODES where base.kind.isBase {
+    let baseKind = base.kind
+    let baseName = baseKind.rawValue.withFirstCharacterUppercased;
+    let enumType: TypeSyntax = "\(raw: baseName)SyntaxEnum"
+
+    try! EnumDeclSyntax(
+      """
+      /// Enum to exhaustively switch over all different \(raw: baseName) syntax nodes.
+      public enum \(enumType)
+      """
+    ) {
+      for node in NON_BASE_SYNTAX_NODES where node.base == baseKind {
+        DeclSyntax(
+          """
+          \(node.apiAttributes())\
+          case \(node.varOrCaseName)(\(node.kind.syntaxType))
+          """
+        )
+      }
+    }
+
+    try! ExtensionDeclSyntax(
+      """
+      public extension \(baseKind.syntaxType)
+      """
+    ) {
+      try FunctionDeclSyntax(
+        """
+        /// Get an enum that can be used to exhaustively switch over all \(raw: baseName) syntax nodes.
+        func `as`(_: \(enumType).Type) -> \(enumType)
+        """
+      ) {
+        try SwitchExprSyntax("switch raw.kind") {
+          for node in NON_BASE_SYNTAX_NODES where node.base == baseKind {
+            SwitchCaseSyntax("case .\(node.varOrCaseName):") {
+              StmtSyntax("return .\(node.varOrCaseName)(\(node.kind.syntaxType)(self)!)")
+            }
+          }
+          SwitchCaseSyntax("default:") {
+            ExprSyntax(#"preconditionFailure("unknown \#(raw: baseName) syntax kind")"#)
+          }
+        }
+      }
+    }
+  }
 }
--- a/Sources/SwiftSyntax/generated/SyntaxEnum.swift
+++ b/Sources/SwiftSyntax/generated/SyntaxEnum.swift
@@ -866,3 +866,436 @@ public extension Syntax {
     }
   }
 }
+
+/// Enum to exhaustively switch over all different Decl syntax nodes.
+public enum DeclSyntaxEnum {
+  case accessorDecl(AccessorDeclSyntax)
+  case actorDecl(ActorDeclSyntax)
+  case associatedTypeDecl(AssociatedTypeDeclSyntax)
+  case classDecl(ClassDeclSyntax)
+  case deinitializerDecl(DeinitializerDeclSyntax)
+  case editorPlaceholderDecl(EditorPlaceholderDeclSyntax)
+  case enumCaseDecl(EnumCaseDeclSyntax)
+  case enumDecl(EnumDeclSyntax)
+  case extensionDecl(ExtensionDeclSyntax)
+  case functionDecl(FunctionDeclSyntax)
+  case ifConfigDecl(IfConfigDeclSyntax)
+  case importDecl(ImportDeclSyntax)
+  case initializerDecl(InitializerDeclSyntax)
+  case macroDecl(MacroDeclSyntax)
+  case macroExpansionDecl(MacroExpansionDeclSyntax)
+  case missingDecl(MissingDeclSyntax)
+  case operatorDecl(OperatorDeclSyntax)
+  case poundSourceLocation(PoundSourceLocationSyntax)
+  case precedenceGroupDecl(PrecedenceGroupDeclSyntax)
+  case protocolDecl(ProtocolDeclSyntax)
+  case structDecl(StructDeclSyntax)
+  case subscriptDecl(SubscriptDeclSyntax)
+  case typeAliasDecl(TypeAliasDeclSyntax)
+  case variableDecl(VariableDeclSyntax)
+}
+
+public extension DeclSyntax {
+  /// Get an enum that can be used to exhaustively switch over all Decl syntax nodes.
+  func `as`(_: DeclSyntaxEnum.Type) -> DeclSyntaxEnum {
+    switch raw.kind {
+    case .accessorDecl:
+      return .accessorDecl(AccessorDeclSyntax(self)!)
+    case .actorDecl:
+      return .actorDecl(ActorDeclSyntax(self)!)
+    case .associatedTypeDecl:
+      return .associatedTypeDecl(AssociatedTypeDeclSyntax(self)!)
+    case .classDecl:
+      return .classDecl(ClassDeclSyntax(self)!)
+    case .deinitializerDecl:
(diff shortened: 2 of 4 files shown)


rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxEnumFile.swift
```diff
@@ -57,4 +57,47 @@ let syntaxEnumFile = SourceFileSyntax(leadingTrivia: copyrightHeader) {
       }
     }
   }
+
+  for base: SyntaxNodeKind in [.decl, .stmt, .expr, .type, .pattern] {
+    let baseName = base.rawValue.withFirstCharacterUppercased;
+    try! EnumDeclSyntax(
+      """
+      /// Enum to exhaustively switch over all different \(raw: baseName) syntax nodes.
+      public enum \(raw: baseName)SyntaxEnum
```
Matejkob, line 66: What do you think about adding the name creation directly to the `SyntaxNodeKind`? This would centralize the source for all types generated from `SyntaxNodeKind`, ensuring consistency and reducing manual repetition if one would like to use the name in different part of code generation package. The code could be refactored as follows:

```swift
public var baseTypeEnum: TypeSyntax {
  if isBase {
    return "\(syntaxType)Enum"
  } else {
    fatalError("Only base kinds can define base enum types.")
  }
}
```

CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxEnumFile.swift
```diff
@@ -57,4 +57,47 @@ let syntaxEnumFile = SourceFileSyntax(leadingTrivia: copyrightHeader) {
...
+          \(node.apiAttributes())\
+          case \(node.varOrCaseName)(\(node.kind.syntaxType))
+          """
+        )
+      }
+    }
+
+    try! ExtensionDeclSyntax(
+      """
+      public extension \(raw: baseName)Syntax
```
Matejkob, line 81: I believe you will be able to use the `syntaxType` type property from `SyntaxNodeKind` here. 
```suggestion
      public extension \(base.syntaxType)
```

CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxEnumFile.swift
```diff
@@ -57,4 +57,47 @@ let syntaxEnumFile = SourceFileSyntax(leadingTrivia: copyrightHeader) {
...
+      """
+    ) {
+      try FunctionDeclSyntax(
+        """
+        /// Get an enum that can be used to exhaustively switch over all \(raw: baseName) syntax nodes.
+        func `as`(_: \(raw: baseName)SyntaxEnum.Type) -> \(raw: baseName)SyntaxEnum
```
Matejkob, line 87: How about adding an `is` method here as well? I believe it would be beneficial and consistent with the other casting methods we have.

Matejkob (commented):
I really like the implementation of SyntaxEnum for each base syntax kind, particularly the casting aspect of this feature. Great work!

Sources/SwiftSyntax/generated/SyntaxEnum.swift
```diff
@@ -866,3 +866,436 @@ public extension Syntax {
...
+  case protocolDecl(ProtocolDeclSyntax)
+  case structDecl(StructDeclSyntax)
+  case subscriptDecl(SubscriptDeclSyntax)
+  case typeAliasDecl(TypeAliasDeclSyntax)
+  case variableDecl(VariableDeclSyntax)
+}
+
+public extension DeclSyntax {
+  /// Get an enum that can be used to exhaustively switch over all Decl syntax nodes.
+  func `as`(_: DeclSyntaxEnum.Type) -> DeclSyntaxEnum {
```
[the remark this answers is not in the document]
rintaro: This is consistent with other casting function, including the original `SyntaxEnum`.

CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxEnumFile.swift
```diff
@@ -57,4 +57,47 @@ let syntaxEnumFile = SourceFileSyntax(leadingTrivia: copyrightHeader) {
       }
     }
   }
+
+  for base: SyntaxNodeKind in [.decl, .stmt, .expr, .type, .pattern] {
```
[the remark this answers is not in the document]
rintaro: That would include `.syntax`, `.syntaxCollection`, which we don't want.

CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxEnumFile.swift
rintaro: If anyone finds the actual use-case of it, they can propose it later.
Personally, I don't think a good idea to make it a method/property (unless it's used in _many_ places) because `enum \(raw: baseName)SyntaEnum` is clearer than `enum \(raw: base.baseTypeEnum)`.

CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxEnumFile.swift
rintaro: I don't think it's really valuable.  Is there any difference from e.g. `.is(ExprSyntax.self)`?

CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxEnumFile.swift
rintaro: Nice, thanks for pointing out!

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

Sources/SwiftSyntax/generated/SyntaxEnum.swift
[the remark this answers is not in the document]
Matejkob: Oh, I see. That's a bit odd API but if it is consistent that's good

CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxEnumFile.swift
Matejkob: I see where you're coming from. However, considering that the name is used in multiple places within this code, I propose at least moving the name to a local variable. For instance:

```swift
let enumName: TypeSyntax = "\(syntaxType)Enum"
```

CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxEnumFile.swift
Matejkob: Oh yeah, of course in this form it doesn't make any sense. What I had in mind – and didn't articulate well – is an `is` method that accepts a syntax enum type as an argument and returns a boolean indicating whether the node is of that type. It would look something like this:

```swift
extension ExprSyntax {
  func `is`(_ nodeType: ExprSyntaxEnum) -> Bool { ... }
}
```

What are your thoughts on this approach? While one could use the regular `is` method, I believe using an enum could make the API more user-friendly.

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxEnumFile.swift
rintaro: You mean for exapple, like this?
```swift
func test(node1: ExprSyntax, node2: ExprSyntax) {
  let someExpr = ExprSyntaxEnum(node1)
  if node2.is(someExpr) {
    // node2 is the same type as node1
  }
}
```
Sorry, I personally still don't see a useful use-case of it. Maybe I'm still missing something.
Anyway, all the current `is(_:)` functions accept a meta type, not the value. Adding such method is outside the scope this PR.

Matejkob: Since we already have `SyntaxEnum` with only an `as` method, I agree that expanding its scope might not fit this PR. However, while we're on the topic of these enums, I'd like to delve a bit deeper into whether this enhancement could be beneficial. Now that I've had my first coffee of the day, I hope to explain my idea more clearly 😅.

Currently, to check if a base node is of a specific leaf node type, we use the `is` method like this:
```swift
let someExpNode: ExprSyntax = "await completion?(argument)"

if someExpNode.is(AwaitExprSyntax.self) {
    // ...
}
```
With the new enums, we could potentially streamline this API:
```swift
let someExpNode: ExprSyntax = "await completion?(argument)"

if someExpNode.is(.awaitExpr) {
    // ...
}
```
The real question is, should we implement this?

I find it quite convenient to have an enumerated list of all possible node types to which a base node can be cast. It seems to make the API more intuitive and user-friendly. What do you think?

hamishknight (approved):

CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxEnumFile.swift
rintaro: I see. but `SyntaxEnum` and `XXXSyntaxEnum` cases have the associated value. So `is(.awaitExpr)` is not possible using these `*SyntaxEnum` types.
The question would be, should we make `{Decl|Stmt|Expr|Type|Pattern}SyntaxKind`  and ``func `is`(_ kind: ExprSyntaxKind) -> Bool`` is valuable and not confusing while we already have `node.kind == .awaitExpr`. I recommend to open an issue, and continue discussion there with wider audience.

Issue #2333, opened 2023-11-06
Add a possibility to validate `fixedSource` post macro expansion in the `assertMacroExpansion`


Matejkob:
### Description

In light of the discussion [here](https://github.com/apple/swift-syntax/pull/2021#pullrequestreview-1714013108), there's a need to re-evaluate the testing strategy for macro expansions, particularly concerning the application of fix-its. The current methodology asserts the `expandedSource` before the application of any fix-its, which may not fully validate the macro's behavior in the presence of fixable code errors.

A more robust testing process would involve the `assertMacroExpansion` function not only applying fix-its to the `originalSource` but also asserting the result of the macro expansion on the corrected source code. This approach has been effectively utilized in `swift-macro-testing` by mbrandonw and stephencelis, showcased [here](https://github.com/pointfreeco/swift-macro-testing/blob/35acd9468d40ae87e75991a18af6271e8124c261/README.md?plain=1#L148-L188).

To align with this improved testing paradigm, I propose the addition of a new parameter—perhaps `expandedSourceAfterApplyingFixIt`—which would enable users to assert against the post-fix-it expanded source. This addition is crucial as macros may propose fix-its for diagnostics that do not halt the expansion process, such as those with a severity less than `error`.

By incorporating this feature, `assertMacroExpansion` would gain the capability to confirm the fixed source after the macro expansion, ensuring a more comprehensive verification process.

Matejkob:
I'm on it

ahoppen:
Tracked in Apple’s issue tracker as rdar://118012844

PR #2355, opened 2023-11-13
Enhance assertMacroExpansion to Validate Post-Fix-It Macro Expansions

Matejkob:
This PR introduces an enhancement to the `assertMacroExpansion` function in SwiftSyntax's testing suite. The key addition is the capability to validate macro expansions on source code after the application of Fix-Its. This change ensures a more robust and comprehensive testing process, particularly for macros that propose fix-its for diagnostics.

Modifications include:
- Addition of a new parameter `expandedFixedSource` in test assertions. This parameter allows users to assert against the expanded source code after applying Fix-Its.
- Update of `assertMacroExpansion` to handle the new testing workflow. It now performs macro expansion tests on both the original and the fixed source code, ensuring that both the pre-fix and post-fix states are correctly validated.
- Refinement of the internal logic for applying Fix-Its and conducting macro expansion tests, ensuring accuracy and efficiency in the testing process.
- Update of test cases in `AddBlockerTests.swift` and `AddCompletionHandlerMacroTests.swift` to demonstrate and validate the new functionality.

Resolves: https://github.com/apple/swift-syntax/issues/2333

### DIFF
--- a/Examples/Tests/MacroExamples/Implementation/Expression/AddBlockerTests.swift
+++ b/Examples/Tests/MacroExamples/Implementation/Expression/AddBlockerTests.swift
@@ -58,9 +58,11 @@ final class AddBlockerTests: XCTestCase {
       ],
       macros: macros,
       applyFixIts: ["use '-'"],
-      fixedSource:
-        """
+      fixedSource: """
         #addBlocker(x * y - z)
+        """,
+      expandedFixedSource: """
+        x * y - z
         """,
       indentationWidth: .spaces(2)
     )
--- a/Examples/Tests/MacroExamples/Implementation/Peer/AddCompletionHandlerMacroTests.swift
+++ b/Examples/Tests/MacroExamples/Implementation/Peer/AddCompletionHandlerMacroTests.swift
@@ -135,6 +135,19 @@ final class AddCompletionHandlerMacroTests: XCTestCase {
           }
         }
         """,
+      expandedFixedSource: """
+        struct Test {
+          func fetchData() async -> String {
+            return "Hello, World!"
+          }
+
+          func fetchData(completionHandler: @escaping (String) -> Void) {
+            Task {
+              completionHandler(await fetchData())
+            }
+          }
+        }
+        """,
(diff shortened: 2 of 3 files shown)


Matejkob:
Blocked by #2329

ahoppen (commented):
I’m not sure how useful this is since the compiler would never automatically apply the Fix-Its and then expand the macro. IMO you should write a test case that checks the error case (and `fixedSource`) and then test the actual expansion in a different test case.

Matejkob:
> I’m not sure how useful this is since the compiler would never automatically apply the Fix-Its and then expand the macro. IMO you should write a test case that checks the error case (and `fixedSource`) and then test the actual expansion in a different test case.

Thank you for your input. I understand your point about the compiler's typical behavior. However, the enhancement to `assertMacroExpansion` is intended to streamline testing by simulating a developer's real-world workflow. In practice, developers often apply Fix-Its and then expect macros to function correctly. This update allows for testing this entire process within a single test case, reflecting how users actually interact with macros.

By incorporating `expandedFixedSource`, we're not just following the compiler's process, but providing a more efficient and comprehensive testing approach. This helps in validating the macro's behavior in scenarios that developers are likely to encounter, making our testing suite more practical and robust.

bnbarham:
I understand the motivation here, but `assertMacroExpansion` is already relatively complex and adding yet another parameter to it only adds to that complexity. IMO it's generally more understandable for each test to have a focus - in this case, tests for the fix-its and other tests for the expansion of valid code. Mixing them together ends up creating giant tests that are difficult to process and likely duplicates already tested functionality.

Matejkob:
In that case I'm closing the PR and associated issue

ahoppen:
> It would be beneficial if the application of fix-its were covered in those test cases. However, if I'm not mistaken, the current official version of `swift-syntax` does not include the feature to assert fix-it applications in the assert macro function. 😟

FWIW @kimdv added the capability to test the fixed source in https://github.com/apple/swift-syntax/pull/2021 but it’s not part of a swift-syntax release yet.

PR #2357, opened 2023-11-15
[NFC] Minor annotations to minimize diagnostics under strict concurrency checking.

hborla:
I tried building SwiftSyntax with `.enableExperimentalFeature("StrictConcurrency")`, and there were only a handful of warnings. Most of them were about accessing mutable global/static variables, which I've changed to `let`s because they weren't mutated anyway. For global variables to be concurrency-safe without isolation, they need to be immutable and `Sendable`, so I also annotated `SyntaxMemoryLayout.Value` as `Sendable` (which is trivial, but conformance synthesis is suppressed because the type is `public`).

The only remaining strict concurrency warning in SwiftSyntax itself after this change is here:

```swift
fileprivate struct SourceLocationDirectiveArguments {
  enum Error: Swift.Error, CustomStringConvertible {
    case nonDecimalLineNumber(TokenSyntax)
    case stringInterpolationInFileName(SimpleStringLiteralExprSyntax)

    var description: String {
      switch self {
      case .nonDecimalLineNumber(let token):
        return "'\(token.text)' is not a decimal integer"
      case .stringInterpolationInFileName(let stringLiteral):
        return "The string literal '\(stringLiteral)' contains string interpolation, which is not allowed"
      }
    }
  }

...
```

because `Error`-conforming types must conform to `Sendable`, but `TokenSyntax` and `SimpleStringLiteralExprSyntax` are not `Sendable`. There are also a handful of warnings in `MultithreadingTests.swift` for similar reasons.

### DIFF
--- a/Sources/SwiftSyntax/BumpPtrAllocator.swift
+++ b/Sources/SwiftSyntax/BumpPtrAllocator.swift
@@ -17,8 +17,8 @@
 public class BumpPtrAllocator {
   typealias Slab = UnsafeMutableRawBufferPointer
 
-  static private var GROWTH_DELAY: Int = 128
-  static private var SLAB_ALIGNMENT: Int = 8
+  static private let GROWTH_DELAY: Int = 128
+  static private let SLAB_ALIGNMENT: Int = 8
 
   /// Initial slab size.
   private var slabSize: Int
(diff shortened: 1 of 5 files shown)


hborla:
@swift-ci please test

hborla:
@swift-ci please test Linux

PR #2292, opened 2023-10-16
Remove `SyntaxKind` conformance to `CaseIterable`

ahoppen:
There’s no good reason to iterate over all cases in `SyntaxKind` and if `SyntaxKind` conforms to `CaseIterable`, we can’t mark members of it as deprecated (https://github.com/apple/swift-syntax/pull/2237/files#r1359917461)

While doing this I noticed that `IntegerLiteralExprSyntax.Radix` also conformed to `CaseIterable`. Similarly, there’s no good reason to iterate over all radix kinds, so I removed that conformance as well.

PR #2363, opened 2023-11-16
Added conformance to UInt16 to SyntaxKind and made it CaseIterable

joehinkle11:


ahoppen:
We explicitly removed the conformance of `SyntaxKind` to `CaseIterable` in https://github.com/apple/swift-syntax/pull/2292. 

Out of curiosity: Why do you need to iterate over all elements in `SyntaxKind`? There shouldn’t be any good reason to iterate over all the nodes, AFAICT.

joehinkle11:
No I don't need it. Not sure what I was thinking TBH

PR #2366, opened 2023-11-18
Update DeclGroupSyntax to refine DeclSyntaxProtocol

### LINKED ISSUE #2365: Change DeclGroupSyntax to refine DeclSyntaxProtocol
### Description

Changing `DeclGroupSyntax` to refine `DeclSyntaxProtocol` would allow users to write generic functions over `DeclSyntaxProtocol` which also work for `DeclGroupSyntax`. One use case in swift-mmio is a function that attempts to cast an A as a B or emits an error diagnostic.

```swift
// from swift-mmio
extension DeclSyntaxProtocol {
 func requireAs<Other>(
    _ other: Other.Type,
    _ context: MacroContext<some ParsableMacro, some MacroExpansionContext>
  ) throws -> Other where Other: DiagnosableDeclSyntaxProtocol {
    if let decl = self.as(Other.self) { return decl }

    let node: any SyntaxProtocol =
      (self as? DiagnosableDeclSyntaxProtocol)?.introducerKeyword ?? self

    throw context.error(
      at: node,
      message: .expectedDecl(Other.self))
  }
}
```

rauhul:
Updates the swift-syntax code generator to support traits refining a syntax base type. Uses this new code generator feature to change DeclGroupSyntax to refine DeclSyntaxProtocol.

Fixes #2365

### DIFF
--- a/CodeGeneration/Sources/SyntaxSupport/Traits.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/Traits.swift
@@ -14,14 +14,19 @@ import SwiftSyntax
 
 public class Trait {
   public let traitName: String
-  public let syntaxBaseName: TokenSyntax?
+  /// The kind of syntax the trait refines.
+  ///
+  /// Base kind _must_ be a base syntax node, e.g. `.decl`, `.expr` , or
+  /// others. See ``SyntaxNodeKind/isBase`` for more details.
+  public let baseKind: SyntaxNodeKind?
   public let protocolName: TokenSyntax
   public let documentation: SwiftSyntax.Trivia
   public let children: [Child]
 
-  init(traitName: String, syntaxBaseName: String?, documentation: String? = nil, children: [Child]) {
+  init(traitName: String, baseKind: SyntaxNodeKind? = nil, documentation: String? = nil, children: [Child]) {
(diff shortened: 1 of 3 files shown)


CodeGeneration/Sources/SyntaxSupport/Traits.swift
```diff
@@ -14,12 +14,14 @@ import SwiftSyntax
 
 public class Trait {
   public let traitName: String
+  public let syntaxBaseName: TokenSyntax?
```
ahoppen, line 17: I would change this to `baseKind: SyntaxNodeKind?` and then use `baseKind.protocolType` to get the type that you want to get the protocol’s type name. 

Also: A doc comment would be appreciated.

rauhul:
@swift-ci please test

CodeGeneration/Sources/SyntaxSupport/Traits.swift
rauhul: Should be fixed in [1166176](https://github.com/apple/swift-syntax/pull/2366/commits/11661761d2693e2a80bdec8b12f089e511a54a3f)

ahoppen (approved):
Looks good to me 👍🏽

PR #2373, opened 2023-11-29
[5.10] improve assert macro expansion highlight verification

rauhul:
Adds a new initializer to DiagnosticSpec which takes an optional array of highlights instead of a single optional highlight string. Updates existing initializer to map the string into an array of one.

Updates assertMacroExpansion to check each highlight in the diagnostic spec against the actual produced diagnostic individually. Additionally changes the behavior to drop leading and trailing trivial when performing the highlight comparison to better match the Swift compiler's behavior.


rauhul:
@swift-ci please test

rauhul:
@swift-ci please test

PR #2379, opened 2023-12-01
Replace throwsSpecifier/thrownTypeClause with a unified throwsClause 

DougGregor:
The current modeling of typed throws kept the `throwSpecifier` separate from the thrown error type. Pull them together in a unified "throws clause" .


DougGregor:
https://github.com/apple/swift/pull/70169

@swift-ci please test

Sources/SwiftParser/Specifiers.swift
```diff
@@ -615,8 +602,8 @@ extension Parser {
     var asyncKeyword: RawTokenSyntax? = nil
     var unexpectedBeforeThrows: [RawSyntax] = []
     var throwsKeyword: RawTokenSyntax?
-    var thrownError: RawThrownTypeClauseSyntax?
-    var unexpectedAfterThrownError: [RawSyntax] = []
+    var throwsClause: RawThrowsClauseSyntax?
+    var unexpectedAfterThrowsClause: [RawSyntax] = []
```
ahoppen, line 606: I don’t like having both a `throwsKeyword` and a `throwsClause` here because it’s not clear when we need to include the `throwsKeyword` in the source tree and when the `throwsClause` and thus it’s easy to get source-accuracy wrong.

I would prefer to only have `throwsClause` here and for a `throwsClause` whenever we parse a `throws` keyword.

Sources/SwiftParser/Specifiers.swift
```diff
@@ -774,7 +766,7 @@ extension Parser {
       } else if let (spec, handle, matchedSubset) = self.at(anyIn: S.MisspelledThrowsTokenKinds.self, or: S.CorrectThrowsTokenKinds.self) {
         let misspelledThrows = self.eat(handle)
         unexpected.append(misspelledThrows)
-        if effectSpecifiers?.throwsSpecifier == nil {
+        if effectSpecifiers?.throwsClause == nil {
           if matchedSubset == S.CorrectThrowsTokenKinds.self {
             synthesizedThrows = missingToken(spec)
```
ahoppen, line 771: We should synthesize an entire throws clause including the thrown type here instead of just synthesizing the `throws` keyword. This should be reflected in the Fix-It for something like

```swift
func foo() -> throws(MyError) Int {}
```

(it would be good to add a test case for this)

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -236,7 +236,7 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
     exchangeTokens(
       unexpected: misplacedSpecifiers,
       unexpectedTokenCondition: { EffectSpecifier(token: $0) != nil },
-      correctTokens: [effectSpecifiers?.asyncSpecifier, effectSpecifiers?.throwsSpecifier],
+      correctTokens: [effectSpecifiers?.asyncSpecifier, effectSpecifiers?.throwsClause?.throwsSpecifier],
```
ahoppen, line 239: This won’t produce a correct Fix-It for the following because it will only move `throws` in front of `->` but not `MyError`.

```swift
func foo() -> throws(MyError) Int
```

Sources/SwiftSyntax/SwiftSyntaxCompatibility.swift
```diff
@@ -50,6 +50,32 @@ public extension DeclGroupSyntax {
   }
 }
 
+
+public extension EffectSpecifiersSyntax {
+  @available(*, deprecated, message: "use throwsClause.throwsSpecifier")
+  var throwsSpecifier: TokenSyntax? {
+    throwsClause?.throwsSpecifier
+  }
```
ahoppen, line 58: This should also provide a setter that sets the `throwsSpecifier` inside the `throwsClause` and creates a `throwsClause` if none exists.

Sources/SwiftSyntax/generated/SyntaxEnum.swift
```diff
@@ -267,7 +267,7 @@ public enum SyntaxEnum {
   #endif
   case thenStmt(ThenStmtSyntax)
   case throwStmt(ThrowStmtSyntax)
-  case thrownTypeClause(ThrownTypeClauseSyntax)
+  case throwsClause(ThrowsClauseSyntax)
```
ahoppen, line 270: This is what makes this change technically API-incompatible in case you exhaustively switched over `SyntaxEnum`. The release notes should point that out. (same for `SyntaxKind`)

Sources/SwiftSyntax/generated/SyntaxTraits.swift
```diff
@@ -137,34 +137,17 @@ public protocol EffectSpecifiersSyntax: SyntaxProtocol {
     set
   }
   
-  var unexpectedBetweenAsyncSpecifierAndThrowsSpecifier: UnexpectedNodesSyntax? {
+  var unexpectedBetweenAsyncSpecifierAndThrowsClause: UnexpectedNodesSyntax? {
```
ahoppen, line 140: We should also provide compatibility layers for `unexpectedBetweenAsyncSpecifierAndThrowsSpecifier` and `unexpectedAfterThrownError`.

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxNodesAB.swift
```diff
@@ -543,18 +542,15 @@ public struct AccessorEffectSpecifiersSyntax: SyntaxProtocol, SyntaxHashable, _L
...
       _ unexpectedBeforeAsyncSpecifier: UnexpectedNodesSyntax? = nil,
       asyncSpecifier: TokenSyntax? = nil,
-      _ unexpectedBetweenAsyncSpecifierAndThrowsSpecifier: UnexpectedNodesSyntax? = nil,
-      throwsSpecifier: TokenSyntax? = nil,
-      _ unexpectedBetweenThrowsSpecifierAndThrownError: UnexpectedNodesSyntax? = nil,
-      thrownError: ThrownTypeClauseSyntax? = nil,
-      _ unexpectedAfterThrownError: UnexpectedNodesSyntax? = nil,
+      _ unexpectedBetweenAsyncSpecifierAndThrowsClause: UnexpectedNodesSyntax? = nil,
+      throwsClause: ThrowsClauseSyntax? = nil,
+      _ unexpectedAfterThrowsClause: UnexpectedNodesSyntax? = nil,
```
ahoppen, line 553: We need a compatibility layer for this initializer.

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxNodesAB.swift
```diff
@@ -613,7 +605,7 @@ public struct AccessorEffectSpecifiersSyntax: SyntaxProtocol, SyntaxHashable, _L
     }
   }
   
-  public var unexpectedBetweenAsyncSpecifierAndThrowsSpecifier: UnexpectedNodesSyntax? {
+  public var unexpectedBetweenAsyncSpecifierAndThrowsClause: UnexpectedNodesSyntax? {
```
ahoppen, line 608: This needs a compatibility layer.

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxNodesAB.swift
```diff
@@ -622,21 +614,17 @@ public struct AccessorEffectSpecifiersSyntax: SyntaxProtocol, SyntaxHashable, _L
...
   }
   
-  /// The `throws` keyword.
-  ///
-  /// ### Tokens
-  /// 
-  /// For syntax trees generated by the parser, this is guaranteed to be `throws`.
-  public var throwsSpecifier: TokenSyntax? {
+  /// The clause specifying thrown errors
+  public var throwsClause: ThrowsClauseSyntax? {
```
ahoppen, line 618: This needs a compatibility layer.

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxNodesAB.swift
```diff
@@ -622,21 +614,17 @@ public struct AccessorEffectSpecifiersSyntax: SyntaxProtocol, SyntaxHashable, _L
...
-      return Syntax(self).child(at: 3)?.cast(TokenSyntax.self)
+      return Syntax(self).child(at: 3)?.cast(ThrowsClauseSyntax.self)
     }
     set(value) {
       self = Syntax(self).replacingChild(at: 3, with: Syntax(value), arena: SyntaxArena()).cast(AccessorEffectSpecifiersSyntax.self)
     }
   }
   
-  public var unexpectedBetweenThrowsSpecifierAndThrownError: UnexpectedNodesSyntax? {
+  public var unexpectedAfterThrowsClause: UnexpectedNodesSyntax? {
```
ahoppen, line 627: This needs a compatibility layer.

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxNodesEF.swift
```diff
@@ -3797,7 +3789,7 @@ public struct FunctionEffectSpecifiersSyntax: SyntaxProtocol, SyntaxHashable, _L
     }
   }
   
-  public var unexpectedBetweenAsyncSpecifierAndThrowsSpecifier: UnexpectedNodesSyntax? {
+  public var unexpectedBetweenAsyncSpecifierAndThrowsClause: UnexpectedNodesSyntax? {
```
ahoppen, line 3792: This needs a compatibility layer.

Sources/SwiftSyntax/SwiftSyntaxCompatibility.swift
```diff
@@ -50,6 +50,32 @@ public extension DeclGroupSyntax {
...
+}
+
+
+public extension FunctionEffectSpecifiersSyntax {
+  @available(*, deprecated, message: "use throwsClause instead of throwsSpecifier")
+  init(
+      leadingTrivia: Trivia? = nil,
+      asyncSpecifier: TokenSyntax? = nil,
+      throwsSpecifier: TokenSyntax? = nil,
+      trailingTrivia: Trivia? = nil
```
ahoppen, line 68: You should also be able to specify unexpected nodes here. This function should have the same signature as the original `FunctionEffectSpecifiersSyntax` initializer that took a `throwsSpecifier`.

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxNodesEF.swift
```diff
@@ -3806,23 +3798,17 @@ public struct FunctionEffectSpecifiersSyntax: SyntaxProtocol, SyntaxHashable, _L
...
   }
   
-  /// The `throws` or `rethrows` keyword.
-  ///
-  /// ### Tokens
-  /// 
-  /// For syntax trees generated by the parser, this is guaranteed to be one of the following kinds:
-  ///  - `throws`
-  ///  - `rethrows`
-  public var throwsSpecifier: TokenSyntax? {
```
ahoppen, line 3816: This needs a compatibility layer.

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxNodesEF.swift
```diff
@@ -3806,23 +3798,17 @@ public struct FunctionEffectSpecifiersSyntax: SyntaxProtocol, SyntaxHashable, _L
...
-      return Syntax(self).child(at: 3)?.cast(TokenSyntax.self)
+      return Syntax(self).child(at: 3)?.cast(ThrowsClauseSyntax.self)
     }
     set(value) {
       self = Syntax(self).replacingChild(at: 3, with: Syntax(value), arena: SyntaxArena()).cast(FunctionEffectSpecifiersSyntax.self)
     }
   }
   
-  public var unexpectedBetweenThrowsSpecifierAndThrownError: UnexpectedNodesSyntax? {
+  public var unexpectedAfterThrowsClause: UnexpectedNodesSyntax? {
```
ahoppen, line 3811: This needs a compatibility layer.

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxNodesTUVWXYZ.swift
```diff
@@ -2261,17 +2296,15 @@ public struct TypeEffectSpecifiersSyntax: SyntaxProtocol, SyntaxHashable, _LeafS
...
       _ unexpectedBeforeAsyncSpecifier: UnexpectedNodesSyntax? = nil,
       asyncSpecifier: TokenSyntax? = nil,
-      _ unexpectedBetweenAsyncSpecifierAndThrowsSpecifier: UnexpectedNodesSyntax? = nil,
-      throwsSpecifier: TokenSyntax? = nil,
-      _ unexpectedBetweenThrowsSpecifierAndThrownError: UnexpectedNodesSyntax? = nil,
-      thrownError: ThrownTypeClauseSyntax? = nil,
-      _ unexpectedAfterThrownError: UnexpectedNodesSyntax? = nil,
+      _ unexpectedBetweenAsyncSpecifierAndThrowsClause: UnexpectedNodesSyntax? = nil,
+      throwsClause: ThrowsClauseSyntax? = nil,
+      _ unexpectedAfterThrowsClause: UnexpectedNodesSyntax? = nil,
```
ahoppen, line 2307: This needs a compatibility layer.

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxNodesTUVWXYZ.swift
```diff
@@ -2328,7 +2357,7 @@ public struct TypeEffectSpecifiersSyntax: SyntaxProtocol, SyntaxHashable, _LeafS
     }
   }
   
-  public var unexpectedBetweenAsyncSpecifierAndThrowsSpecifier: UnexpectedNodesSyntax? {
+  public var unexpectedBetweenAsyncSpecifierAndThrowsClause: UnexpectedNodesSyntax? {
```
ahoppen, line 2360: This needs a compatibility layer.

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxNodesTUVWXYZ.swift
```diff
@@ -2337,19 +2366,17 @@ public struct TypeEffectSpecifiersSyntax: SyntaxProtocol, SyntaxHashable, _LeafS
     }
   }
   
-  /// ### Tokens
-  /// 
-  /// For syntax trees generated by the parser, this is guaranteed to be `throws`.
-  public var throwsSpecifier: TokenSyntax? {
```
ahoppen, line 2343: This needs a compatibility layer.

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxNodesTUVWXYZ.swift
```diff
@@ -2337,19 +2366,17 @@ public struct TypeEffectSpecifiersSyntax: SyntaxProtocol, SyntaxHashable, _LeafS
...
-      return Syntax(self).child(at: 3)?.cast(TokenSyntax.self)
+      return Syntax(self).child(at: 3)?.cast(ThrowsClauseSyntax.self)
     }
     set(value) {
       self = Syntax(self).replacingChild(at: 3, with: Syntax(value), arena: SyntaxArena()).cast(TypeEffectSpecifiersSyntax.self)
     }
   }
   
-  public var unexpectedBetweenThrowsSpecifierAndThrownError: UnexpectedNodesSyntax? {
+  public var unexpectedAfterThrowsClause: UnexpectedNodesSyntax? {
```
ahoppen, line 2379: This needs a compatibility layer.

Sources/SwiftSyntax/generated/SyntaxEnum.swift
DougGregor: `thrownTypeClause` doesn't exist in any release; it was added for typed throws, which are still experiment.

Sources/SwiftSyntax/SwiftSyntaxCompatibility.swift
DougGregor: Per my other comment, API compatibility with unexpected notes is unlikely to be useful in practice. I do not think we should do this.

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxNodesEF.swift
DougGregor: It's handled via the extension of the `EffectSpecifiersSyntax` protocol.

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxNodesTUVWXYZ.swift
DougGregor: It's handled via the extension of the `EffectSpecifiersSyntax` protocol.

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxNodesAB.swift
DougGregor: It's handled via the extension of the EffectSpecifiersSyntax protocol.

Sources/SwiftSyntax/SwiftSyntaxCompatibility.swift
DougGregor: That's a good point

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxNodesTUVWXYZ.swift
DougGregor: Sure

Sources/SwiftParser/Specifiers.swift
DougGregor: That's fair; I've found a cleaner way to implement it without `throwsKeyword`

Sources/SwiftParser/Specifiers.swift
DougGregor: There's a bit more to do in this area. We're not doing a good job in general of moving the whole throws clause around.

DougGregor:
https://github.com/apple/swift/pull/70169

@swift-ci please test

DougGregor:
https://github.com/apple/swift/pull/70169

@swift-ci please test

Sources/SwiftParser/Specifiers.swift
```diff
@@ -596,12 +582,24 @@ extension TokenConsumer {
...
+      return RawThrowsClauseSyntax(
+        throwsSpecifier: throwsKeyword,
+        leftParen: nil,
+        type: nil,
+        rightParen: nil,
+        arena: self.arena
+      )
+    }
+
+    let (unexpectedBetweenThrowsSpecifierAndLeftParen, leftParen) = self.expect(.leftParen)
```
ahoppen, line 597: Instead of doing `self.at(.leftParen)` above and then `self.expect`, you can just do the following above.

```swift
guard experimentalFeatures.contains(.typedThrows), let leftParenn = self.consume(if: .leftParen)
```

Sources/SwiftParser/Specifiers.swift
ahoppen: Could you file issues so that we don’t forget to do it after the proposal is accepted?

Sources/SwiftSyntax/generated/SyntaxEnum.swift
ahoppen: I mean that it’s API incompatible because of the *addition* of the `throwsClause` case. Since packages aren’t built without library evolution enabled by SwiftPM, the addition of cases to enums is API-incompatible if any client exhaustively switches over them.

Sources/SwiftSyntax/SwiftSyntaxCompatibility.swift
```diff
@@ -13,6 +13,23 @@
...
 // All users of the declarations in this file should transition away from them ASAP.
 
+public extension AccessorEffectSpecifiersSyntax {
+  @available(*, deprecated, message: "use throwsClause instead of throwsSpecifier")
+  init(
+    leadingTrivia: Trivia? = nil,
+    asyncSpecifier: TokenSyntax? = nil,
+    throwsSpecifier: TokenSyntax? = nil,
+    trailingTrivia: Trivia? = nil
+  ) {
```
ahoppen, line 23: These convenience initializers need to also take unexpected nodes. Their API signature should match the initializers that we previously had.

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxNodesAB.swift
ahoppen: Ah, I missed that `EffectSpecifiersSyntax` was a protocol. 👍🏽

Sources/SwiftSyntax/SwiftSyntaxCompatibility.swift
ahoppen: I think we should do our best to maintain API compatibility wherever possible. Adding these parameters and the `unexpected` properties is very little work on our side and we shouldn’t make any assumptions on how clients are using swift-syntax. 

If you *really* disagree, the API incompatibility needs to be listed in the release notes.

ahoppen:
Summarizing the open review comments:

- File issues to move throws clause around
  - https://github.com/apple/swift-syntax/pull/2379#discussion_r1414388723
  - https://github.com/apple/swift-syntax/pull/2379#discussion_r1412679547
- Addition of case to enum is API incompatible
  - https://github.com/apple/swift-syntax/pull/2379#discussion_r1414393379
- API compatibility layer for unexpected nodes
  - https://github.com/apple/swift-syntax/pull/2379#discussion_r1414399964
- Simplify implementation
  - https://github.com/apple/swift-syntax/pull/2379#discussion_r1414385077

Sources/SwiftSyntax/SwiftSyntaxCompatibility.swift
DougGregor: I *really* disagree. It's wasted work, and that time is better spent on more significant improvements.

Sources/SwiftParser/Specifiers.swift
DougGregor: Filed an issue at https://github.com/apple/swift-syntax/issues/2391 to cover this

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxNodesEF.swift
ahoppen: The rule that we are currently following is that two swift-syntax versions should be API-compatible except for the incompatibilities noted in the release notes. And as the unexpected fields are part of the public API, I think they should follow the same rules – even if we don’t expect them to be used much

ahoppen: I opened https://github.com/apple/swift-syntax/pull/2393 to add the compatibility  layers.

PR #2380, opened 2023-12-02
Let `PrecedenceGroupDeclSyntax` conform to `BracedSyntax`

SimplyDanny:


SimplyDanny:
> You also need to remove the XFail here
> 
> https://github.com/apple/swift-syntax/blob/6e708c4fa643057f41e3af7e6c0a2c405a487e56/CodeGeneration/Tests/ValidateSyntaxNodes/ValidateSyntaxNodes.swift#L520

Oh, that means this conformance was intentionally left out? Is it okay then to add it now?

ahoppen:
I created these XFails at some point but it looks like we just never went through this one and re-considered it. I don’t see a reason why `PrecedenceGroupDeclSyntax` should’t conform to `BracedSyntax`.

ahoppen (approved):

ahoppen:
@swift-ci Please test Windows

Issue #2206, opened 2023-09-16
`assertMacroExpansion` should emit an error if member macro is applied to declaration that can’t have members


ahoppen:
When applying an attached member macro to e.g. a variable, `assertMacroExpansion` will currently happily swallow the attribute. It should, however, emit an error that member macros can’t be applied to variables (specifically, if the declaration is not a `DeclGroupSyntax`). Ie. the following test case should emit an error.

```swift
func testMemberMacroOnVar() {
  struct TestMacro: MemberMacro {
    static func expansion(
      of node: AttributeSyntax,
      providingMembersOf declaration: some DeclGroupSyntax,
      conformingTo protocols: [TypeSyntax],
      in context: some MacroExpansionContext
    ) throws -> [DeclSyntax] {
      return []
    }
  }

  assertMacroExpansion(
    """
    @Test var x: Int
    """,
    expandedSource: """
      var x: Int
      """,
    macros: [
      "Test": TestMacro.self
    ]
  )
}
```

Rajveer100:
@ahoppen 
Any specific prerequisites or setup that's needed to get me going for `swift-syntax`? (I just need a quick head start to get the flow.)

I also saw another issue which is clearly `implementation` based [shell-script](https://github.com/apple/swift-syntax/issues/2341), which I think is interesting and is doable with some grinding.

PS: Having a great time contributing towards `Swift` and `LLVM`! :)

Rajveer100:
If I remember correctly, `macros` were introduced this year (WWDC 2023) from `Swift 5.9` and earlier it was always hidden behind the implementation of many Swift features? 

I am digging down to learn more about it, let me know if there any specific `resources` as well for this issue.

Rajveer100:
@ahoppen 
I am facing an issue with the following command:

```Shell
> swift run --package-path CodeGeneration
```

Error:
```Shell
Fetching https://github.com/apple/swift-argument-parser.git
Fetched https://github.com/apple/swift-argument-parser.git (2.24s)
Computing version for https://github.com/apple/swift-argument-parser.git
Computed https://github.com/apple/swift-argument-parser.git at 1.2.3 (0.61s)
Creating working copy for https://github.com/apple/swift-argument-parser.git
Working copy of https://github.com/apple/swift-argument-parser.git resolved at 1.2.3
Building for debugging...
[272/272] Linking generate-swift-syntax
Build complete! (17.20s)
[...]

PR #2382, opened 2023-12-02
Emit error for assertMacroExpansion when applying an attached member macro to declaration that can’t have members

Rajveer100:
Resolves Issue #2206

When we apply an attached `member macro` to a declaration that cannot have members, `assertMacroExpansion` fails to emit appropriate `diagnostic` message.


Tests/SwiftSyntaxMacroExpansionTest/MemberMacroTests.swift
```diff
@@ -299,4 +299,29 @@ final class MemberMacroTests: XCTestCase {
...
+      """
+      @Test var x: Int
+      """,
+      expandedSource: """
+        var x: Int
+        """,
+      macros: [
+        "Test": TestMacro.self
+      ]
+    )
```
ahoppen, line 325: I think you have it the wrong way round. From the issue description

> the following test case should emit an error. 

Also: Did you run all the other tests? I suspect all member tests will fail with your current changes.

Rajveer100: > I think you have it the wrong way round. From the issue description

You mean the test itself is wrong or it's in the wrong place?

> Also: Did you run all the other tests? I suspect all member tests will fail with your current changes.

Well yes, as I said other member tests are failing.

ahoppen: > You mean the test itself is wrong or it's in the wrong place?

Yes, having an attached member macro on a variable should emit an error.

Rajveer100: Ok so I had mistakenly put a `guard` instead of `if`, although I am yet to find the `trigger` for `VariableDeclSyntax` so I can see the test failing and then add the error message, the previous errors and tests are good which weren't supposed to fail.

Rajveer100: What `diagnostic` do you recommend adding here? Is there an existing one for such case or we need a custom addition to the `enum` which we then `throw` during detection?

Rajveer100: What property should be check under `MacroExpansion` in the below case (directed here from `MacroSystem` -> `expandAttachedMacro` -> `expandAttachedMacroWithoutCollapsing`) if I am under the right place:

```Swift
case (let attachedMacro as MemberMacro.Type, .member):
// ...
```

ahoppen: In the issue I wrote 
> specifically, if the declaration is not a `DeclGroupSyntax`

So that’s the syntax node type I would check.

Rajveer100: This check is for the `declGroup` I suppose, not for the actual check if it is a `variable`?

Also, could you briefly differentiate between `attributeNode`, `declarationNode` and `declGroup`?

Rajveer100: @ahoppen 
According to the issue, as you mentioned:

> specifically, if the declaration is not a DeclGroupSyntax

But, the following check exists:

```Swift
guard let declGroup = declarationNode.asProtocol(DeclGroupSyntax.self)
```

How does it not throw an error here if the node isn't a `DeclGroupSyntax`?

PR #2386, opened 2023-12-06
Add initial support for _resultDependsOn

meg-gupta:
Add them under a new NonEscapableTypes experimental feature

### DIFF
--- a/Sources/SwiftParser/Patterns.swift
+++ b/Sources/SwiftParser/Patterns.swift
@@ -354,7 +354,7 @@ extension Parser.Lookahead {
         && !self.at(.keyword(.__owned))
         && !self.at(.keyword(.borrowing))
         && !self.at(.keyword(.consuming))
-        && (!experimentalFeatures.contains(.nonEscapableTypes) || !self.at(.keyword(._resultDependsOn)))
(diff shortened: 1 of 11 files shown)


meg-gupta:
https://github.com/apple/swift/pull/70263

@swift-ci Please test

Sources/SwiftParser/Patterns.swift
```diff
@@ -354,6 +354,7 @@ extension Parser.Lookahead {
         && !self.at(.keyword(.__owned))
         && !self.at(.keyword(.borrowing))
         && !self.at(.keyword(.consuming))
+        && (!experimentalFeatures.contains(.nonEscapableTypes) || !self.at(.keyword(._resultDependsOn)))
```
ahoppen, line 357: FWIW I think this would be more readable as follows because that matches the pattern of having a list of and-not expressions.

```suggestion
        && !(experimentalFeatures.contains(.nonEscapableTypes) && self.at(.keyword(._resultDependsOn)))
```

meg-gupta:
https://github.com/apple/swift/pull/70263

@swift-ci Please test

ahoppen (approved):

PR #2395, opened 2023-12-09
Remove `WildcardPattern.typeAnnotation`

rintaro:
Type annotation has never been a part of `wildcard-pattern`.

### DIFF
--- a/Sources/SwiftSyntax/SwiftSyntaxCompatibility.swift
+++ b/Sources/SwiftSyntax/SwiftSyntaxCompatibility.swift
@@ -541,6 +541,45 @@ public extension TypeEffectSpecifiersSyntax {
   }
 }
 
+public extension WildcardPatternSyntax {
+  @available(*, deprecated, message: "remove 'typeAnnotation'")
+  init(
+    leadingTrivia: Trivia? = nil,
+    _ unexpectedBeforeWildcard: UnexpectedNodesSyntax? = nil,
+    wildcard: TokenSyntax = .wildcardToken(),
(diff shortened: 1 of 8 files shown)


rintaro:
@swift-ci Please test

Sources/SwiftSyntax/SwiftSyntaxCompatibility.swift
```diff
@@ -509,6 +509,30 @@ public extension TypeEffectSpecifiersSyntax {
...
+      nil,
+      trailingTrivia: trailingTrivia
+    );
+  }
+
+  @available(*, deprecated, message: "'typeAnnotation' was removed")
+  var typeAnnotation: TypeAnnotationSyntax? {
+    get { nil }
+    set {}
+  }
```
ahoppen, line 533: I think we should also cover the unexpected children in the compatibility layer. See https://github.com/apple/swift-syntax/pull/2379#discussion_r1421197690

ahoppen (commented):
I think it would be good to add a release note for this change + add a test case so we don’t make the same mistake again 😉

rintaro:
@ahoppen  I'm going to get this into 5.10. What's the manner for wring the release notes? Is it OK to write it directly in 5.10 section before the approval?

ahoppen:
Yes, that’s OK

ahoppen (approved):

rintaro:
@swift-ci Please test macOS

rintaro:
@swift-ci Please test Linux

PR #2397, opened 2023-12-10
build: clean up `Package.swift` slightly

compnerd:
Now that macros are enabled on all the platforms we currently support (Windows, Linux, Apple), the optional addition of the macro targets was removed.  Clean up the structure of the file as the initial changes were done quickly to enable testing on all platforms.


compnerd:
@swift-ci please test

PR #2398, opened 2023-12-11
Don’t use switch expressions

ahoppen:
I forgot that we need to support Swift 5.8 when I introduced the switch expression. Write the test without relying on Swift 5.9 features.


ahoppen:
@swift-ci Please test

bnbarham (approved):

ahoppen:
@swift-ci Please test Windows

PR #2401, opened 2023-12-12
Address remaining `FIXME`s for the new `NameMatcher`

ahoppen:
rdar://118996561

### DIFF
--- a/Sources/SwiftIDEUtils/NameMatcher.swift
+++ b/Sources/SwiftIDEUtils/NameMatcher.swift
@@ -55,9 +55,9 @@ public class NameMatcher: SyntaxAnyVisitor {
 
   // MARK: - Public entry
 
-  public static func resolve(baseNamePositions: some Sequence<AbsolutePosition>, in sourceFile: SourceFileSyntax) -> [DeclNameLocation] {
+  public static func resolve(baseNamePositions: some Sequence<AbsolutePosition>, in tree: some SyntaxProtocol) -> [DeclNameLocation] {
     let matcher = NameMatcher(baseNamePositions: baseNamePositions)
-    matcher.walk(sourceFile)
+    matcher.walk(tree)
     return matcher.resolvedLocs
   }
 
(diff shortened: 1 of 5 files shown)


ahoppen:
@swift-ci Please test

Sources/SwiftIDEUtils/NameMatcher.swift
```diff
@@ -193,16 +195,28 @@ public class NameMatcher: SyntaxAnyVisitor {
 
   public override func visit(_ token: TokenSyntax) -> SyntaxVisitorContinueKind {
     while let baseNamePosition = firstPositionToResolve(in: token.leadingTriviaRange) ?? firstPositionToResolve(in: token.trailingTriviaRange) {
```
bnbarham, line 197: Why is this a loop 🤔?

Also, IIUC this will cause issues with the DocC disambiguation mechanism. Consider:
```
func foo(arg: Int) -> Int { return 1 }
func foo(arg: Int) -> String { return "" }
```

If there was a comment with `foo(arg:)-9x9e2` and then `arg` (or `foo`) was renamed, the hash no longer matches. It looks like that hash is a FNV1 radix-36 hash of the USR, so we can't generate it here either.

ahoppen: > Why is this a loop 🤔?

Because there might be multiple locations to resolve in the comment. And we need to start parsing the comment’s text at those specific offsets so that previous tokens can’t mess up the part of the comment that we’re interested in. E.g. consider the following

```swift
// 1️⃣foo(arg:) // 2️⃣foo(arg:)
```

We need to start parsing twice, once after 1️⃣ (parsing `foo(arg:) // foo(arg:)`) and once after 2️⃣ (parsing `foo(arg:)`). With a single parse, we would never parse the second `foo` reference as a compound decl name.

> Also, IIUC this will cause issues with the DocC disambiguation mechanism

That’s already true today, if you are renaming the base name, right? So, it’s no better or worse than what we had before AFAICT.

bnbarham (approved):

ahoppen:
https://github.com/apple/swift/pull/70403

@swift-ci Please test

ahoppen:
https://github.com/apple/swift/pull/70403

@swift-ci Please test

ahoppen:
https://github.com/apple/swift/pull/70403

@swift-ci Please test

ahoppen:
https://github.com/apple/swift/pull/70403

@swift-ci Please test Windows

ahoppen:
https://github.com/apple/swift/pull/70403

@swift-ci Please test

ahoppen:
https://github.com/apple/swift/pull/70403

@swift-ci Please test Windows

ahoppen:
https://github.com/apple/swift/pull/70403

@swift-ci Please test Windows

kimdv:
https://github.com/apple/swift/pull/70403

@swift-ci Please test Windows

PR #2403, opened 2023-12-13
Modify `assertMacroExpansion()` so that it only conditionally calls `XCTFail()`.

### LINKED ISSUE #2400: Needed: variant of `assertMacroExpansion()` that does not use XCTest
### Description

Test targets that use swift-testing instead of XCTest need some way to work with this function; let's add a throwing equivalent that can be used without any dependency on a specific testing library.

grynspan:
This PR changes the signature of `assertMacroExpansion()` such that an alternative failure handler can be passed for use when e.g. swift-testing is being used.

Right now, the function calls `XCTAssertEqual()`, `XCTFail()`, etc. on the assumption that the caller is using XCTest. This assumption does not always hold and the dependency on XCTest may prevent a developer from writing effective tests leveraging `assertMacroExpansion()` while also using swift-testing or other testing libraries.

It is not possible for swift-syntax to include swift-testing as a dependency at this time, so the function cannot be amended to directly call the equivalent of `XCTFail()` from that library. Nor can the function check if the current process is running an XCTest-based test because XCTest does not expose API to determine if a test is running.

Until such time as swift-syntax can directly call into swift-testing, developers can do so when calling `assertMacroExpansion()` like so:

```swift
assertMacroExpansion(
  "@foo bar",
  "expanded_foo bar",
  macros: ...,
  ...
) { message, fileID, filePath, line, column in
  Issue.record(Comment(rawValue: message), fileID: fileID, filePath: String(describing: filePath), Int(line), Int(column))
}
```

An improved interface is a future direction.

Resolves #2400.


grynspan:
@swift-ci please test

Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
```diff
@@ -34,41 +34,76 @@ public struct NoteSpec {
...
+  if note.message != spec.message {
+    let message = describeDifferenceBetweenStrings(note.message, spec.message, "message of note does not match")
+    failureHandler(
+      message,
+      spec.originatorFileID,
+      spec.originatorFilePath,
+      spec.originatorLine,
+      spec.originatorColumn
+    )
+  }
```
ahoppen, line 87: What do you think about having a `assertStringsEqualWithDiff` version that takes a `failureHandler`. I think that would make that this comparison is essentially doing an assertion.

(same in the other locations where you replaced `assertStringsEqualWithDiff `)

Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
```diff
@@ -152,8 +208,10 @@ public struct DiagnosticSpec {
   ///   - highlights: If not empty, the text fragments the diagnostic is expected to highlight
   ///   - notes: The notes that are expected to be attached to the diagnostic
   ///   - fixIts: The messages of the Fix-Its the diagnostic is expected to produce
-  ///   - originatorFile: The file at which this ``NoteSpec`` was created, so that assertion failures can be reported at its location.
+  ///   - originatorFileID: The file ID at which this ``NoteSpec`` was created, so that assertion failures can be reported at its location.
+  ///   - originatorFilePath: The file path at which this ``NoteSpec`` was created, so that assertion failures can be reported at its location.
   ///   - originatorLine: The line at which this ``NoteSpec`` was created, so that assertion failures can be reported at its location.
+  ///   - originatorColumn: The column at which this ``NoteSpec`` was created, so that assertion failures can be reported at its location.
```
ahoppen, line 214:
```suggestion
  ///   - originatorFileID: The file ID at which this ``DiagnosticSpec`` was created, so that assertion failures can be reported at its location.
  ///   - originatorFilePath: The file path at which this ``DiagnosticSpec `` was created, so that assertion failures can be reported at its location.
  ///   - originatorLine: The line at which this ``DiagnosticSpec `` was created, so that assertion failures can be reported at its location.
  ///   - originatorColumn: The column at which this ``DiagnosticSpec `` was created, so that assertion failures can be reported at its location.
```

Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
```diff
@@ -34,41 +34,76 @@ public struct NoteSpec {
...
-  XCTAssertEqual(location.column, spec.column, "column of note does not match", file: spec.originatorFile, line: spec.originatorLine)
+  if location.line != spec.line {
+    failureHandler(
+      "line \(location.line) of note does not match \(spec.line)",
+      spec.originatorFileID,
+      spec.originatorFilePath,
+      spec.originatorLine,
+      spec.originatorColumn
+    )
+  }
```
ahoppen, line 97: Similar here, what about having `assertEqual` that takes a `failureHandler`?

(same in the other locations where you replaced `XCTAssertEqual`)

Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
```diff
@@ -34,41 +34,76 @@ public struct NoteSpec {
   public let column: Int
 
   /// The file and line at which this ``NoteSpec`` was created, so that assertion failures can be reported at its location.
-  internal let originatorFile: StaticString
+  internal let originatorFileID: String
+  internal let originatorFilePath: StaticString
   internal let originatorLine: UInt
+  internal let originatorColumn: UInt
```
ahoppen, line 40: Instead of storing/passing all these 4 members every single time, what do you think about introducing a struct that holds them? That should also simplify the signature of `failureHandler` and make `failureHandler` extensible if we choose to add one of the parameters / deprecate one of them like `filePath`.

Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
```diff
@@ -299,6 +423,7 @@ func assertDiagnostic(
 ///   - testModuleName: The name of the test module to use.
 ///   - testFileName: The name of the test file name to use.
 ///   - indentationWidth: The indentation width used in the expansion.
+///   - failureHandler: A closure to call if the assertion fails. This closure may be called more than once. If `nil`, `XCTFail()` is called.
```
ahoppen, line 426:
```suggestion
///   - failureHandler: A closure to call if the assertion fails. This closure may be called more than once. If `nil` defaults to `XCTFail`.
```

Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
```diff
@@ -218,70 +280,132 @@ extension DiagnosticSpec {
...
 
+private func defaultFailureHandler(
+  _ message: String,
+  fileID: String = #fileID,
+  filePath: StaticString = #filePath,
+  line: UInt = #line,
+  column: UInt = #column
+) {
+  // FIXME: detect if XCTest is currently running a test and only call XCTFail() if it is.
+  // FIXME: support swift-testing.
```
bnbarham, line 405: I'm not a huge fan of FIXME's that don't have corresponding issues. I'd just leave this out and file an issue to support swift-testing by default when we can.

grynspan:
I don't think we're pursuing the right solution here.

PR #2408, opened 2023-12-23
Add an article version of my WWDC talk on how to write macros

ahoppen:
Might need some more polishing but it’s something to start with for everybody who prefers reading to watching videos.

### DIFF
--- a/Sources/SwiftSyntaxMacros/SwiftSyntaxMacros.docc/SwiftSyntaxMacros.md
+++ b/Sources/SwiftSyntaxMacros/SwiftSyntaxMacros.docc/SwiftSyntaxMacros.md
@@ -1,37 +1,299 @@
-# ``SwiftSyntaxMacros``
+# Swift Macros
+
+Learn how to write Swift Macros.
+
+Swift Macros allow you to extend the Swift language by generating code at compile time based on annotations in the source code itself. Conceptually, Macros are compiler extensions. The macro’s generation happens at build time and communicates directly with the compiler, which will incorporate the generated code into the compilation process.
+
+With their tight integration into the compiler and the Swift language, macros use some compiler concepts like syntax trees that are able to represent the entire Swift language. Due to their complexity, macros are an advanced concept and are intended for library authors, who can use them to provide extensions to Swift that would be impossible without macros. Wherever possible, language features like generics, protocols and extensions should be preferred over macros. 
+
+> Note: The core concepts of this article are also available as a video at <https://developer.apple.com/wwdc23/10166>
+
+## Table of Contents
+
+- [Macro Overview](#Macro-Overview)
+  - [Macro Declaration](#Macro-Declaration)
+  - [Macro Implementation](#Macro-Implementation)
+- [Testing Macros](#Testing-Macros)
+- [Macro Roles](#Macro-Roles)
+- [Implement an `EnumSubset` macro](#Implement-an-EnumSubset-macro)
+  - [Motivation](#Motivation)
+  - [Declare EnumSubset](#Declare-EnumSubset)
+  - [Implement EnumSubset](#Implement-EnumSubset)
+- [Inspect the SwiftSyntax Tree](#Inspect-the-SwiftSyntax-Tree)
+- [Building Syntax Nodes](#Building-Syntax-Nodes)
+- [Emitting Errors](#Emitting-Errors)
 
 
+## Macro Overview
 
-An experimental library to provide syntactic macro support for Swift. New macros can be introduced by creating new types that conform to the various `*Macro` protocols, e.g., `ExpressionMacro` for an expression-like macro. The `MacroSystem` can then rewrite uses of that those macros in the source code. This library is a prototype for a potential Swift macro feature, but is not tied in with the compiler like such a feature would be.
+A new Swift Macro template can be created from Xcode or the command line. 
 
-## Trying out macro expansion
+- In Xcode, click File -> New -> Package… and select the *Swift Macro* template
+- On the command line, run `swift package init --type macro`
 
-The `swift-parser-cli` program has an `expand-macros` command that will expand any all of the example macros (see below) in the program's source. For example, providing it with the source code:
+Within the template, the `main.swift` file contains the following code.
 
 ```swift
-let b = #stringify(x + y)
+let (result, code) = #stringify(a + b)
 ```
 
-will produce
+It uses an *expression macro*, indicated by the `#` character. At compile time, the compiler expands `#stringify(a + b)` to a tuple that contains the value of the macro’s argument as well as a string with the argument’s: `(a + b, "a + b")`. If `a = 17` and `b = 25` this evaluates to `(42, "a + b")` at runtime. In Xcode, the expanded code can be viewed by right-clicking on `stringify` and selecting *Expand Macro*.
+
+To expand the `stringify` macro, the compiler requires two ingredients: The macro declaration and the macro implementation.
+
+### Macro Declaration
+
+The macro’s declaration declares the macro’s signature, including its input and output type. The declaration of `stringify` looks as follows
 
 ```swift
-let b = (x + y, #"x + y"#)
+@freestanding(expression)
+public macro stringify<T>(_ value: T) -> (T, String) = #externalMacro(module: "MyMacroMacros", type: "StringifyMacro")
 ```
 
-Use `swift-parser-cli` to test out your macros and see what kind of syntactic transformations are possible.
+`@freestanding(expression)` declares the macro with the freestanding expression macro role. This means that the macro can be used wherever an expression can be used and that its use is indicated by the hash character, like `#stringify`. More macro roles are discussed below in [Macro Roles](#Macro-Roles).
 
-## Example macros
+The next part of the macro declaration looks like a function signature. It takes a generic argument of type `T` and outputs a tuple containing the result, again of type `T`, as well as a string containing the source code of the macro’s argument. All arguments need to type check and the macro’s result type is inferred from the return type. If the arguments don’t type check or the result type doesn’t match its surrounding context, the compiler will emit an error without applying the macro expansion. This is different to, for example, C macros, which are evaluated at the pre-processor stage before type-checking.
 
-A number of example macros have been implemented in the file `MacroSytem+Examples.swift`. They include:
+The last part of the macro links the macro declaration to the macro implementation. `#externalMacro` tells the compiler that to expand the macro, it needs to use the `StringifyMacro` type in the `MyMacroMacros` module.
 
-* `#line`: Implements the `#line` expression.
-* `#column`: Implements the `#column` expression.
-* `#function`: Implements the `#function` expression.
-* `#stringify(x)`: returns both the value `x` and the string form of its expression, as a tuple.
-* `#colorLiteral(red: <value>, green: <value>, blue: <value>, alpha: <value>)`: Implements the `#colorLiteral` expression.
+### Macro Implementation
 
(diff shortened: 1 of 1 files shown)


Sources/SwiftSyntaxMacros/SwiftSyntaxMacros.docc/SwiftSyntaxMacros.md
```diff
@@ -1,37 +1,308 @@
...
+    providingMembersOf declaration: some DeclGroupSyntax,
+    in context: some MacroExpansionContext
+  ) throws -> [DeclSyntax] {
+    guard let enumDecl = declaration.as(EnumDeclSyntax.self) else {
+      // Emit an error here, see section *Emitting Errors*
+      return []
+    }
+
+    // Extract the name of the generic parameter. 
+    // See section *Inspect the SwiftSyntax Tree* how to build this expresion.
```
bnbarham, line 227:
```suggestion
    // See section *Inspect the SwiftSyntax Tree* for more details on building this expression.
```

Sources/SwiftSyntaxMacros/SwiftSyntaxMacros.docc/SwiftSyntaxMacros.md
```diff
@@ -1,37 +1,308 @@
...
+```
+
+Compared to the `stringify` macro, `EnumSubset` differs in two ways:
+1. It is declared as an attached member macro and defines the names of the members it introduces. Declaring the introduced names improves the compiler’s performance: When accessing a member that is not declared in the macro’s names, the compiler doesn’t need to try expanding the macro to find it.
+2. `EnumSubset` doesn’t take any argument. Instead, it defines a generic parameter that is used to customize the superset – in the example this is `Keyword`. Using a generic parameter ensures that the superset type exists.
+
+### Implement EnumSubset
+
+A possible implementation of `EnumSubsetMacro` can look as follows. 
+In the following sections, we explore techniques for how to write this code.
```
bnbarham, line 212:
```suggestion
A possible implementation of `EnumSubsetMacro` is as follows.
```

The latter part isn't needed IMO, you explicitly mention the sections in comments anyway.

Sources/SwiftSyntaxMacros/SwiftSyntaxMacros.docc/SwiftSyntaxMacros.md
```diff
@@ -1,37 +1,308 @@
...
+    }
+
+    // Extract the name of the generic parameter. 
+    // See section *Inspect the SwiftSyntax Tree* how to build this expresion.
+    guard let supersetType = attribute,
+      .attributeName.as(SimpleTypeIdentifierSyntax.self)?
+      .genericArgumentClause?
+      .arguments.first?
+      .argumentType else {
+      // Emit an error here.
```
bnbarham, line 233: I think it would be good to include emitting the error here to serve as a complete example.

Sources/SwiftSyntaxMacros/SwiftSyntaxMacros.docc/SwiftSyntaxMacros.md
```diff
@@ -1,37 +1,308 @@
...
+  }
+}
+```
+
+## Inspect the SwiftSyntax Tree
+
+Navigating the syntax tree is not trivial. When the syntax tree seems overly complicated, this is usually because it needs to represent the entire Swift language. For example, each enum case declaration can declare multiple elements because it is possible to declare multiple enum elements after a single case keyword like:
+```swift
+case `actor`, `class`
+```
```
bnbarham, line 267: I know why you have this here, but it's not needed IMO. I'd just start with:
> One of the best ways to explore the syntax tree is by example....

Sources/SwiftSyntaxMacros/SwiftSyntaxMacros.docc/SwiftSyntaxMacros.md
```diff
@@ -1,37 +1,308 @@
...
+
+Usually, the best way to explore the syntax trees is by example. <https://swift-ast-explorer.com> is a great webpage that allows the shows the syntax tree of entered Swift code.  
+
+> Experiment: Try pasting the following code to <https://swift-ast-explorer.com> and follow the syntax tree to understand the above code that extracts the name of the generic parameter in `EnumSubsetMacro.expansion`.
```
bnbarham, line 271:
```suggestion
> Experiment: Try pasting the following code into <https://swift-ast-explorer.com> and compare the generated syntax tree to the above code that extracts the name of the generic parameter in `EnumSubsetMacro.expansion`.
```

Sources/SwiftSyntaxMacros/SwiftSyntaxMacros.docc/SwiftSyntaxMacros.md
```diff
@@ -1,37 +1,308 @@
...
+> @EnumSubset<Keyword>
+> ```
+
+Alternatively, the syntax tree can be printed in the debugger. When the debugger is stopped and `node` is a syntax node, the syntax tree can be printed using `po node`. This produces a syntax tree as shown in the [Macro Implementation](#Macro-Implementation) section.
+
+## Building Syntax Nodes
+
+There are three core approaches to build syntax nodes: Parsing from string literals, result builder initializers and memberwise initializers. 
+
+The `stringify` macro created its result type, an `ExprSyntax` from a string literal. As described in that section, this invokes the Swift parser to parse the contents of the string literal into a syntax tree. This technique is great for statically known trees or trees with a fixed number of parameters.
```
bnbarham, line 282:
```suggestion
The `stringify` macro created its result type, an `ExprSyntax`, from a string literal. As described in that section, this invokes the Swift parser to parse the contents of the string literal into a syntax tree. This technique works well for statically known trees, or trees with a fixed number of parameters.
```

Sources/SwiftSyntaxMacros/SwiftSyntaxMacros.docc/SwiftSyntaxMacros.md
```diff
@@ -1,37 +1,308 @@
...
+
+The `EnumSubset` macro uses a result builder initializer to generate the `InitializerDeclSyntax` and `SwitchExprSyntax`. The initializer takes a header – the `swiftch` keyword and subject – and a trailing closure, which is a result builder that adds a `SwitchCaseSyntax` for every `case` item in the `switch` statement. Result builder initializers are a great tool for repetitive syntax constructs, like the `switch` in `EnumSubset`. Also note how `EnumSubsetMacro` combines result builders and string literals. It uses the result builder to generate the `SwitchExprSyntax` but the cases are constructed using string interpolation.
```
bnbarham, line 284:
```suggestion
The `EnumSubset` macro uses a result builder initializer to generate the `InitializerDeclSyntax` and `SwitchExprSyntax`. The initializer takes a header – the `switch` keyword and subject – and a trailing closure, which is a result builder that adds a `SwitchCaseSyntax` for every `case` item in the `switch` statement. Result builder initializers are a great tool for repetitive syntax constructs, like the `switch` in `EnumSubset`. Also note how `EnumSubsetMacro` combines result builders and string literals. It uses the result builder to generate the `SwitchExprSyntax` but the cases are constructed using string interpolation.
```

Sources/SwiftSyntaxMacros/SwiftSyntaxMacros.docc/SwiftSyntaxMacros.md
```diff
@@ -1,37 +1,308 @@
...
+
+Finally, every syntax type has a memberwise initializer with which a syntax node can be created by specifying all its children. This is the most low-level way of creating syntax nodes and has the most flexibility. It is also a good option for nodes that are not representable by string literals and don’t have result builder initializers.
```
bnbarham, line 286:
```suggestion
Finally, every syntax type has a memberwise initializer with which a syntax node can be created by specifying all its children.
```

Or reword the last two sentences - "good option" is a bit odd since it's actually the *only* option. Maybe:
> Finally, every syntax type has a memberwise initializer with which a syntax node can be created by specifying all its children. This initializer gives you full control when creating a syntax node, but is also the most verbose.

Sources/SwiftSyntaxMacros/SwiftSyntaxMacros.docc/SwiftSyntaxMacros.md
```diff
@@ -1,37 +1,308 @@
...
 
-Well, that depends. The `SwiftSyntaxMacros` module is part of a potential vision for a macros feature in Swift, where the effect of macro evaluation is as it works here---executable Swift code that transforms a given syntax tree into another syntax tree. That design also integrates macros into the type checker, so that macros are only evaluated on well-typed inputs and are expected to either produce well-typed syntax trees as output or emit diagnostics along the way. If that design is accepted as the design for Swift, much of this module will remain as it is---but the examples themselves will be divided into a builtin set of macros that are always available to the compiler (e.g., to provide the implementation for things like `#line`) or moved into a separate examples module to provide documentation on how to use the feature.
+The ``MacroExpansionContext/addDiagnostics(from:node:)`` method on the ``MacroExpansionContext`` allows further customization of the error message, like showing the error message at a different location than the attribute, generating warnings, or even providing Fix-Its in Xcode.
```
bnbarham, line 308: > There's two ways to emit this error:
>   1. Throw the error from the expansion function. This will generate a compilation error on the line of the attribute.
>   2. Add it as a diagnostic with  ``MacroExpansionContext/addDiagnostics(from:node:)``. This allows further customization, eg. changing the location, emitting a warning instead of an error, or even providing Fix-Its.

bnbarham (commented):
> An experimental library

Heh, thanks for updating!

Sources/SwiftSyntaxMacros/SwiftSyntaxMacros.docc/SwiftSyntaxMacros.md
```diff
@@ -1,37 +1,308 @@
...
+
+The last part of the macro links the macro declaration to the macro implementation. `#externalMacro` tells the compiler that to expand the macro, it needs to use the `StringifyMacro` type in the `MyMacroMacros` module.
+
+### Macro Implementation
+
+The `StringifyMacro` type is declared in a separate target, which is a compiler plug-in. The compiler will send the source code of the entire macro expression to that plug-in. The first thing that the macro plug-in does, is to parse the source code of the macro into a SwiftSyntax tree. This tree is a source-accurate, structural representation of the macro, and it will be the basis on which the macro operates.
```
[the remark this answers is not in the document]
ahoppen: Changed to 

> The compiler will send the macro expression to that plug-in. The plugin processes the macro expression as a SwiftSyntax tree, which is a source-accurate, structural representation of the macro, and it will be the basis on which the macro operates.

Sources/SwiftSyntaxMacros/SwiftSyntaxMacros.docc/SwiftSyntaxMacros.md
ahoppen: Added the error handling to the example.

Sources/SwiftSyntaxMacros/SwiftSyntaxMacros.docc/SwiftSyntaxMacros.md
ahoppen: Changed to

> There are two ways to emit this error:
>
> 1. Throw an error from the `expansion` function. This generates a compilation error on the line of the attribute. This is the technique that the above example uses. 
> 2. Add it as a diagnostic with  ``MacroExpansionContext/diagnose(_:)``. This allows further customization, like changing the location, emitting a warning instead of an error, or even providing Fix-Its.

ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #2416, opened 2024-01-05
Create swift.yml

ScorpiusDraconis83:


bnbarham:
Apple has GitHub actions disabled at the organization level, so running for PRs won't have any effect - individuals can enable GitHub actions on their own forks if they are so inclined. `@swift-ci test` is used for testing in this repo (and most other swift-family projects).

PR #2418, opened 2024-01-06
Fix lookahead for effects specifiers in function types

DougGregor:
We were failing to properly look past effect specifiers like `async throws(X)` in a function type starting with an attribute.


DougGregor:
@swift-ci please test

rintaro (approved):

PR #2423, opened 2024-01-09
Remove ability to parse `available` in `@_specialize`

ahoppen:
The `available` argument to `@_specialize` is only supported in SIL, which SwiftSyntax doesn’t parse.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #2425, opened 2024-01-09
Add Sendable annotations to swift-syntax

ahoppen:
Audit all modules in swift-syntax, add `Sendable` annotations to the types that are sendable.

The audit uncovered that `SyntaxArena` was sent across actor boundaries even though `SyntaxArena` is not thread safe. I fixed these issues by:
- Changing the semantics of `ParsingSyntaxArena` slightly: If you modify a parsed token and the new arena is a different one than the old arena (and thus might not be a `ParsingSyntaxArena`), we materialize it and its trivia. That way we no longer need to search through all child arenas to find a `ParsingSyntaxArena` and their `parseTrivia` function. It also fixes a potential correctness issue if the modified node is allocated in a `ParsingSyntaxArena` that has a different `parseTrivia` function than the original `ParsingSyntaxArena`.
- Introducing `RetainedSyntaxArena`: In many cases where we were passing around a `SyntaxArena`, we just needed to ensure that the syntax arena stayed alive but never accessed any properties of it. `RetainedSyntaxArena` is an opaque wrapper around `SyntaxArena` and even though `SyntaxArena` is not sendable, we can send this type across actor boundaries because it doesn’t allow access to the  internal state of `SyntaxArena`.
- Introducing `SyntaxArenaAllocated(Buffer)Pointer`: This is a wrapper around `Unsafe(Buffer)Pointer` that indicates that the buffer is allocated in a syntax arena. Because the syntax arena will always outlive any syntax nodes that reference its contents, we know that the buffer's contents won't get deallocated while being accessed and thus we can add an unchecked `Sendable` conformance to it. `Unsafe(Buffer)Pointer` is not sendable by default.
- There is one concurrency issue remaining when setting/accessing `SyntaxArena.hasParent`, which already had a `FIXME` associated with it. This PR won’t make the situation any worse than it is today, so I just moved the `FIXME` but we should fix it in a follow-up PR.

rdar://117813694

### DIFF
--- a/Sources/SwiftSyntax/Raw/RawSyntaxTokenView.swift
+++ b/Sources/SwiftSyntax/Raw/RawSyntaxTokenView.swift
@@ -95,7 +95,7 @@ public struct RawSyntaxTokenView: Sendable {
   public var leadingRawTriviaPieces: [RawTriviaPiece] {
     switch raw.rawData.payload {
     case .parsedToken(let dat):
-      let arena = raw.arena.parsingArena!
+      let arena = raw.arena as! ParsingSyntaxArena
       return arena.parseTrivia(source: dat.leadingTriviaText, position: .leading)
     case .materializedToken(let dat):
       return Array(dat.leadingTrivia)
@@ -108,7 +108,7 @@ public struct RawSyntaxTokenView: Sendable {
   public var trailingRawTriviaPieces: [RawTriviaPiece] {
     switch raw.rawData.payload {
     case .parsedToken(let dat):
-      let arena = raw.arena.parsingArena!
+      let arena = raw.arena as! ParsingSyntaxArena
       return arena.parseTrivia(source: dat.trailingTriviaText, position: .trailing)
     case .materializedToken(let dat):
       return Array(dat.trailingTrivia)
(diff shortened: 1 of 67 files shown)


ahoppen:
@swift-ci Please test

Sources/SwiftSyntax/SyntaxArena.swift
```diff
@@ -284,11 +277,48 @@ struct SyntaxArenaRef: Hashable {
...
 
+  /// Get an opaque wrapper that keeps the syntax arena alive.
+  var retained: RetainedSyntaxArena {
+    return RetainedSyntaxArena(value)
+  }
+
+  #if DEBUG || SWIFTSYNTAX_ENABLE_ASSERTIONS
+  /// Accessor for ther underlying's `SyntaxArena.hasParent`
+  var hasParent: Bool {
+    // FIXME: This accesses mutable state across actor boundaries and is thus not currency-safe.
```
bnbarham, line 288:
```suggestion
    // FIXME: This accesses mutable state across actor boundaries and is thus not concurrent-safe.
```
😅 same below

bnbarham (approved):
Other than the silly nitpick, LGTM. Would be good to have @rintaro take a look too though.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/SwiftSyntax/Raw/RawSyntaxTokenView.swift
```diff
@@ -204,8 +202,21 @@ public struct RawSyntaxTokenView {
...
+      // a different or no `parseTrivia` function. We thus cannot use a
+      // `parsedToken` anymore.
+      return .makeMaterializedToken(
+        kind: formKind(),
+        leadingTrivia: formLeadingTrivia(),
+        trailingTrivia: formTrailingTrivia(),
+        presence: newValue,
+        tokenDiagnostic: tokenDiagnostic,
+        arena: arena
+      )
```
rintaro, line 219: Just to make sure I understand correctly, this should be reverted when `SwiftParser` is merged, right?

rintaro (approved):
I didn't check what types should be `Sendable`, but other than that, `SyntaxArena`/`RawSyntax` related changes looks good to me.

Sources/SwiftSyntax/Raw/RawSyntaxTokenView.swift
ahoppen: I’d prefer to keep it even if/when we merge SwiftSyntax + SwiftParser. Technically, the old implementation is not correct because the new arena could have a different `parseTriviaFunction` than the arena the node is allocated in and thus the trivia pieces could change here if we don’t materialize them.

PR #2426, opened 2024-01-10
[Parser] Correct the start byte range for UTF8 characters.

pinkjuice66:
This PR is nearly identical to [the fix in the compiler's lexer](https://github.com/apple/swift/pull/70763), addressing the correction of the UTF-8 start byte range.
- '0x80' is not a valid byte for the start of a UTF8 character; it's a continuation byte.(RFC 3629)

But while examining the SwiftSyntax version, I found an additional semantic error in the start byte validation. 
Existing computed property __isStartOfUTF8Character__ is extented on Unicode.Scalar type, which is very confusing because a value of Unicode.Scalar represents a valid decoded unicode value, not a single UTF-8 byte. 
For example, Unicode.Scalar(0x80) consists of 2 UTF-8 bytes, and it doesn't make sense semantically to check whether it's within the range of the UTF-8 start byte.
So let's check this instead with UInt8 type.

### DIFF
--- a/Sources/SwiftParser/Lexer/UnicodeScalarExtensions.swift
+++ b/Sources/SwiftParser/Lexer/UnicodeScalarExtensions.swift
@@ -156,12 +156,6 @@ extension Unicode.Scalar {
     // including and above the DEL character U+7F.
     return self.value >= 0x20 && self.value < 0x7F
   }
-
-  var isStartOfUTF8Character: Bool {
-    // RFC 2279: The octet values FE and FF never appear.
-    // RFC 3629: The octet values C0, C1, F5 to FF never appear.
-    return self.value <= 0x80 || (self.value >= 0xC2 && self.value < 0xF5)
-  }
 }
 
 extension Unicode.Scalar {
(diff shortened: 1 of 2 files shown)


Sources/SwiftParser/Lexer/UnicodeScalarExtensions.swift
```diff
@@ -179,20 +173,22 @@ extension Unicode.Scalar {
...
+      while let peeked = peek(), !peeked.isStartOfUTF8Character {
         _ = advance()
       }
       return nil
     }
 
+    // Read the number of high bits set, which indicates the number of bytes in
+    // the character.
+    let encodedBytes = (~curByte).leadingZeroBitCount
+    precondition(encodedBytes >= 2 && encodedBytes <= 4)
```
ahoppen, line 190: Could you add a doc comment why this precondition holds? Something like

```swift
// We have a multi-byte UTF-8 scalar.
// Single-byte UTF-8 scalars are handled at the start of the function by checking `curByte < 0x80`.
// `isStartOfUTF8Character` guaranteed that the `curByte` has 2 to 4 leading ones.
```

Just to clarify that we should never hit this precondition if running the lexer on invalid UTF-8.

ahoppen:
Thanks for also opening a PR to this repo 👍🏽

Sources/SwiftParser/Lexer/UnicodeScalarExtensions.swift
pinkjuice66: Added! 
Your words accurately articulate the rationale for the precondition, so no modifications are needed.

ahoppen (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #2320, opened 2023-10-31
Fix missing newline in member macro

kimdv:
Fixes #2307

PR #2428, opened 2024-01-13
Fix indentation issue if trailing trivia contains a newline at the end of an indentation scope

ahoppen:
We were always adding indentation to the trailing trivia of a token. This could cause issues if the token at the end of an indentation scope (in most cases a `}` token) contained a trailing newline.

In that case we would add indentation after that final newline based on the indentation scope of `}`, which effectively indented the next token to the indentation scope inside the `}` and thus indenting it one level to far.

To fix this, don’t add indentation for trailing newlines of tokens. Instead, add the indentation as the leading trivia of the next token, which will pick up the correct indentation scope.

Fixes the issue discussed in https://github.com/apple/swift-syntax/pull/2320#discussion_r1401257138

### DIFF
--- a/Sources/SwiftBasicFormat/Trivia+FormatExtensions.swift
+++ b/Sources/SwiftBasicFormat/Trivia+FormatExtensions.swift
@@ -62,7 +62,20 @@ extension Trivia {
   }
 
   /// Adds `indentation` after every newline in this trivia.
-  func indented(indentation: Trivia, isOnNewline: Bool) -> Trivia {
+  ///
+  /// - Parameters:
+  ///   - indentation: The amount of indentation to add.
+  ///   - isOnNewline: Whether this token starts on a new line.
+  ///     This causes the indentation to get added at the start of the trivia.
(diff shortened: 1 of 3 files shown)


Sources/SwiftBasicFormat/Trivia+FormatExtensions.swift
```diff
@@ -62,7 +62,20 @@ extension Trivia {
   }
 
   /// Adds `indentation` after every newline in this trivia.
-  func indented(indentation: Trivia, isOnNewline: Bool) -> Trivia {
+  ///
+  /// - Parameters:
+  ///   - indentation: The amount of indentation to add
```
kimdv, line 67: Missing a period after`add`

kimdv (approved):

bnbarham (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Issue #2228, opened 2023-09-22
Getting warnings about `@_implementationOnly` when building a package dependent on SwiftSyntax


grynspan:
### Description

When building swift-testing, I get a bunch of warnings along these lines while SwiftSyntax builds:

```
/Volumes/Dev/Source/swift-testing/.build/checkouts/swift-syntax/Sources/SwiftSyntax/SyntaxText.swift:15:22: warning: using '@_implementationOnly' without enabling library evolution for 'SwiftSyntax' may lead to instability during execution
@_implementationOnly import Darwin
                     ^
/Volumes/Dev/Source/swift-testing/.build/checkouts/swift-syntax/Sources/SwiftCompilerPlugin/CompilerPlugin.swift:16:22: warning: using '@_implementationOnly' without enabling library evolution for 'SwiftCompilerPlugin' may lead to instability during execution
@_implementationOnly import SwiftCompilerPluginMessageHandling
                     ^
```

This then pollutes the output of the downstream package (swift-testing) that's using SwiftSyntax. :(

### Steps to Reproduce

Using toolchain swift-DEVELOPMENT-SNAPSHOT-2023-09-20-a and SwiftSyntax 509.0.0.

To repro:

1. Check out https://github.com/apple/swift-testing
2. `export TOOLCHAINS=swift` (if needed)
3. `swift test`

ahoppen:
Tracked in Apple’s issue tracker as rdar://115910675

ahoppen:
We need to replace `@_implementaitonOnly` by the following when https://github.com/apple/swift-evolution/blob/main/proposals/0409-access-level-on-imports.md is accepted.

```swift
#if compiler(>= 5.10)
private import XXX
#else
@_implementationOnly import XXX
#endif
```

keith:
@ahoppen do you know if there a ticket tracking that difference? We're experiencing the same with other libraries that haven't made a similar change like this

ahoppen:
AFAICT it’s expected behavior that Swift 5.10 will warn about `@_implementationOnly` imports in non-library-evolution-enabled modules and scoped imports will only be introduced in Swift 5.11. I don’t know the exact specifics but apparently `@_implementationOnly` was never intended to work without library evolution and had some quirks when library evolution wasn’t enabled, so the warning is expected.

PR #2429, opened 2024-01-20
Remove `@_implementationOnly` annotations

ahoppen:
These annotations produce warnings when compiling swift-syntax without library evolution using Swift ≥5.10.

Replace them by `private import` when compiling using Swift ≥5.11.

Also mark the import of XCTest from `SwiftSyntaxMacrosTestSupport` as `private`, fixing rdar://119517162.

Fixes https://github.com/apple/swift-syntax/issues/2228
rdar://119517162
rdar://115910675

### DIFF
--- a/Sources/SwiftCompilerPlugin/CompilerPlugin.swift
+++ b/Sources/SwiftCompilerPlugin/CompilerPlugin.swift
@@ -12,12 +12,22 @@
 // NOTE: This basic plugin mechanism is mostly copied from
 // https://github.com/apple/swift-package-manager/blob/main/Sources/PackagePlugin/Plugin.swift
 
-@_implementationOnly import Foundation
-@_implementationOnly import SwiftCompilerPluginMessageHandling
 import SwiftSyntaxMacros
 
+#if swift(>=5.11)
+private import Foundation
+private import SwiftCompilerPluginMessageHandling
+#else
+import Foundation
+import SwiftCompilerPluginMessageHandling
+#endif
+
 #if os(Windows)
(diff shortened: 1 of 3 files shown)


ahoppen:
@swift-ci Please test

rintaro (dismissed):

Sources/SwiftCompilerPlugin/CompilerPlugin.swift
```diff
@@ -12,12 +12,22 @@
 // NOTE: This basic plugin mechanism is mostly copied from
 // https://github.com/apple/swift-package-manager/blob/main/Sources/PackagePlugin/Plugin.swift
 
-@_implementationOnly import Foundation
-@_implementationOnly import SwiftCompilerPluginMessageHandling
 import SwiftSyntaxMacros
 
+#if swift(>=5.11)
```
rintaro, line 17: How about
```suggestion
#if hasFeature(AccessLevelOnImport)
```

rintaro (approved):

Sources/SwiftCompilerPlugin/CompilerPlugin.swift
rintaro: It turned out `hasFeature(AccessLevelOnImport)` doesn't work.

ahoppen: Discussed offline that `AccessLevelOnImport` is an experimental feature that’s always enabled, thus the `hasFeature` check doesn’t work. So we just check for Swift 5.11.

PR #2434, opened 2024-01-22
Add a function to check if a name can be used as an identifier in a given context

ahoppen:
rdar://120721971

### DIFF
--- a/Sources/SwiftParser/IsValidIdentifier.swift
+++ b/Sources/SwiftParser/IsValidIdentifier.swift
@@ -1,0 +1,120 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+@_spi(RawSyntax) import SwiftSyntax
+
+/// Context in which to check if a name can be used as an identifier.
+///
+/// - SeeAlso: `Swift.isValidSwiftIdentifier(for:)` extension added by SwiftParser.
+public enum IdentifierCheckContext {
+  /// Check if a name can be used to declare a variable, ie. if it can be used after a `let` or `var` keyword.
+  ///
+  /// ### Examples
+  /// - `test` is a valid variable name and `let test: Int` is valid Swift code
+  /// - `class` is not a valid variable and `let class: Int` is invalid Swift code
+  case variableName
+
+  /// Check if a name can be used as a member access, ie. if it can be used after a `.`.
+  ///
+  /// ### Examples
+  /// - `test` is a valid identifier for member access because `myStruct.test` is valid
+  /// - `class` is a valid identifier for member access because `myStruct.class` is valid, even though `class`
+  ///   needs to be wrapped in backticks when used to declare a variable.
+  /// - `self` is not a valid identifier for member access because `myStruct.self` does not access a member named
+  ///   `self` on `myStruct` and instead returns `myStruct` itself.
+  case memberAccess
+}
+
+extension String {
+  /// Checks whether `name` can be used as an identifier in a certain context.
+  ///
+  /// If the name cannot be used as an identifier in this context, it needs to be escaped.
+  ///
+  /// For example, `class` is not a valid identifier for a variable name and needs to be be wrapped in backticks
+  /// to be valid Swift code, like the following.
+  ///
+  /// ```swift
+  /// let `class`: String
+  /// ```
+  ///
+  /// The context is important here – some names can be used as identifiers in some contexts but not others.
+  /// For example, `myStruct.class` is valid without adding backticks `class`, but as mentioned above,
+  /// backticks need to be added when `class` is used as a variable name.
+  ///
+  /// - SeeAlso: ``SwiftParser/IdentifierCheckContext``
+  public func isValidSwiftIdentifier(for context: IdentifierCheckContext) -> Bool {
+    switch context {
+    case .variableName:
+      return isValidVariableName(self)
+    case .memberAccess:
+      return isValidMemberAccess(self)
+    }
+  }
+}
+
+private func isValidVariableName(_ name: String) -> Bool {
+  var parser = Parser("var \(name)")
+  let decl = DeclSyntax.parse(from: &parser)
+  guard parser.at(.endOfFile) else {
+    // We didn't parse the entire name. Probably some garbage left in the name, so not an identifier.
+    return false
+  }
+  guard !decl.hasError && !decl.hasWarning else {
+    // There were syntax errors in the source code. So not valid.
+    return false
+  }
+  guard let variable = decl.as(VariableDeclSyntax.self) else {
+    return false
+  }
+  guard let identifier = variable.bindings.first?.pattern.as(IdentifierPatternSyntax.self)?.identifier else {
+    return false
+  }
+  guard identifier.rawTokenKind == .identifier else {
+    // We parsed the name as a keyword, eg. `self`, so not a valid identifier.
+    return false
+  }
+  guard identifier.rawText.count == name.utf8.count else {
+    // The identifier doesn't cover all the characters in `name`, so we parsed
+    // some of these characters into trivia or another token.
+    // Thus, `name` is not a valid identifier.
+    return false
+  }
+  return true
+}
+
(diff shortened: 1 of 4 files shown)


ahoppen:
@swift-ci Please test

Sources/SwiftParser/IsValidIdentifier.swift
```diff
@@ -0,0 +1,70 @@
...
+  /// - `class` is not a valid variable and `let class: Int` is invalid Swift code
+  case variableName
+}
+
+/// Checks whether `name` can be used as an identifier in a certain context.
+///
+/// If the name cannot be used as an identifier in this context, it needs to be escaped.
+/// For example, `class` is not a valid identifier for a variable name and needs to be spelled as ```let `class`: String``` to be valid Swift code.
+///
+/// The context is important here some names can be used as identifiers in some contexts but not others. For example, `myStruct.class` is valid without adding backticks `class`, but as mentioned above, backticks need to be added when `class` is used as a variable name.
```
bnbarham, line 32:
```suggestion
/// The context is important here some names - can be used as identifiers in some contexts but not others. For example, `myStruct.class` is valid without adding backticks `class`, but as mentioned above, backticks need to be added when `class` is used as a variable name.
```

It's a little odd that we mention the `myStruct.class` case but don't have a corresponding case for it in `IdentifierCheckContext`.

Sources/SwiftParser/IsValidIdentifier.swift
```diff
@@ -0,0 +1,70 @@
...
+  guard parser.at(.endOfFile) else {
+    // We didn't parse the entire name. Probably some garbage left in the name, so not an identifier.
+    return false
+  }
+  guard let variable = decl.as(VariableDeclSyntax.self) else {
+    return false
+  }
+  guard Array(variable.tokens(viewMode: .sourceAccurate)).count == 2 else {
+    // We expect to parse exactly two tokens ('var' and the name).
+    // If we parsed more, the name go parsed as two tokens and is thus not a
```
bnbarham, line 54:
```suggestion
    // If we parsed more, the name got parsed as two tokens and is thus not a
```

Sources/SwiftParser/IsValidIdentifier.swift
```diff
@@ -0,0 +1,70 @@
...
+    return false
+  }
+  guard identifier.tokenKind == .identifier(name) else {
+    // We parsed the name as a keyword, eg. `self`, so not a valid identifier.
+    return false
+  }
+  guard identifier.leadingTrivia.isEmpty && identifier.trailingTrivia.isEmpty else {
+    // The name contained whitespace, which isn't valid in an identifier.
+    return false
+  }
```
bnbarham, line 68: Ooo nice catch, I forgot about this case.

bnbarham (approved):

Sources/SwiftParser/IsValidIdentifier.swift
ahoppen: OK, added a new case to also check if a name is valid identifier for member access. Seems reasonable to have that as well anyway.

ahoppen:
@swift-ci Please test

Sources/SwiftParser/IsValidIdentifier.swift
```diff
@@ -0,0 +1,126 @@
...
+  /// ### Examples
+  /// - `test` is a valid variable name and `let test: Int` is valid Swift code
+  /// - `class` is not a valid variable and `let class: Int` is invalid Swift code
+  case variableName
+
+  /// Check if a name can be used as a member access, ie. if it can be used after a `.`.
+  ///
+  /// ### Examples
+  /// - `test` is a valid identifier for member access because `myStruct.test` is valid
+  /// - `class` is a valid identifier for member access becuase `myStruct.class` is valid, even though `class`
```
bnbarham, line 30:
```suggestion
  /// - `class` is a valid identifier for member access because `myStruct.class` is valid, even though `class`
```

Sources/SwiftParser/IsValidIdentifier.swift
```diff
@@ -0,0 +1,126 @@
...
+  ///   needs to be wrapped in backticks when used to declare a variable.
+  /// - `self` is not a valid identifiers for member access because `myStruct.self` does not access a member named
```
bnbarham, line 32:
```suggestion
  /// - `self` is not a valid identifier for member access because `myStruct.self` does not access a member named
```

Tests/SwiftParserTest/IsValidIdentifierTests.swift
```diff
@@ -0,0 +1,45 @@
...
+    XCTAssertFalse(isValidIdentifier("test\0test", for: .variableName))
+  }
+
+  func testIsValidMemberAccessName() {
+    XCTAssertTrue(isValidIdentifier("test", for: .memberAccess))
+    XCTAssertTrue(isValidIdentifier("class", for: .memberAccess))
+    XCTAssertTrue(isValidIdentifier("`class`", for: .memberAccess))
+    XCTAssertFalse(isValidIdentifier("self", for: .memberAccess))
+    XCTAssertTrue(isValidIdentifier("`self`", for: .memberAccess))
+    XCTAssertTrue(isValidIdentifier("`let`", for: .memberAccess))
```
bnbarham, line 37: `let` as well? Same above

bnbarham (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Tests/SwiftParserTest/IsValidIdentifierTests.swift
bnbarham: Ah, nice. If you're going that far you could have the init just take a map rather than each case, but it's just a test so I what you have is also fine 👍

Sources/SwiftParser/IsValidIdentifier.swift
```diff
@@ -0,0 +1,126 @@
...
+    return isValidVariableName(name)
+  case .memberAccess:
+    return isValidMemberAccess(name)
+  }
+}
+
+private func isValidVariableName(_ name: String) -> Bool {
+  var parser = Parser("var \(name)")
+  let decl = DeclSyntax.parse(from: &parser)
+  guard parser.at(.endOfFile) else {
```
rintaro, line 65: This should also check the current lexeme's `leadingTriviaByteLength` == 0.

Sources/SwiftParser/IsValidIdentifier.swift
```diff
@@ -0,0 +1,126 @@
...
+    return false
+  }
+  guard let identifier = variable.bindings.first?.pattern.as(IdentifierPatternSyntax.self)?.identifier else {
+    return false
+  }
+  guard identifier.tokenKind == .identifier(name) else {
+    // We parsed the name as a keyword, eg. `self`, so not a valid identifier.
+    return false
+  }
+  guard identifier.leadingTrivia.isEmpty && identifier.trailingTrivia.isEmpty else {
```
rintaro, line 89: `variable.bindingSpecifier.trailingTriviaLength == 1` too. Consider `name = " foo"` case.

rintaro: Or maybe `identifier.rawText.count == name.utf8.count` so we can remove all the trivia check including `.endOfFile` check above.

Sources/SwiftParser/IsValidIdentifier.swift
```diff
@@ -0,0 +1,126 @@
...
+  guard Array(variable.tokens(viewMode: .sourceAccurate)).count == 2 else {
+    // We expect to parse exactly two tokens ('var' and the name).
+    // If we parsed more, the name got parsed as two tokens and is thus not a
+    // valid name.
+    return false
+  }
+  guard let identifier = variable.bindings.first?.pattern.as(IdentifierPatternSyntax.self)?.identifier else {
+    return false
+  }
+  guard identifier.tokenKind == .identifier(name) else {
```
rintaro, line 85:
```suggestion
  guard identifier.rawTokenKind == .identfiier && identifier.rawTokenText.count == name.utf8.count else {
```
should give us a _tiny_ performance increase.

Sources/SwiftParser/IsValidIdentifier.swift
ahoppen: Yes, you’re right. The trivia checks aren’t actually needed. Nice 👍🏽

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #2436, opened 2024-01-23
[509] Add visionOS to the list of platforms that can't start a new process from swift-parser-cli

ahoppen:
* **Explanation**: `Process` is not available on visionOS, causing a build failure of swift-syntax 509.x.x for that platform when building the `swift-parser-cli` command line test utility. We should use the same `#if` check that we have for iOS, tvOS and watchOS to also guard the process spawning code on visionOS.
* **Scope**: Building the swift-parser-cli target on visionOS. Building commonly used products like SwiftSyntax and SwiftParser is fine
* **Risk**: Very low, treat visionOS the same as iOS
* **Testing**: Tested that the swift-syntax package now builds for visionOS
* **Issue**: n/a
* **Reviewer**:  @bnbarham


bnbarham (approved):

ahoppen:
✅ Local testing succeeded

PR #2438, opened 2024-01-23
[Parser] Remove redundant UTF-16 surrogates check logic

pinkjuice66:
The initializer of Unicode.Scalar is intelligent enough to handle the case where a UTF-16 surrogates value is provided as its parameter. Therefore, there is no need for manual checking. Let Unicode.Scalar perform its job here.

### DIFF
--- a/Tests/SwiftParserTest/LexerTests.swift
+++ b/Tests/SwiftParserTest/LexerTests.swift
@@ -1083,6 +1083,41 @@ public class LexerTests: ParserTestCase {
     }
   }
 
+  func testUTF16Surrogates1() {
+    // U+D800 <= (UTF16 surrogates code point) <= U+DFFF
+    let sourceBytes: [UInt8] = [0xED, 0xA0, 0x80]  // The bytes represent the code point U+D800
+
+    lex(sourceBytes) { lexemes in
+      guard lexemes.count == 1 else {
+        return XCTFail("Expected 1 lexemes, got \(lexemes.count)")
+      }
+      assertRawBytesLexeme(
+        lexemes[0],
+        kind: .endOfFile,
+        leadingTrivia: [0xED, 0xA0, 0x80],
+        text: [],
+        error: TokenDiagnostic(.invalidUtf8, byteOffset: 0)
+      )
+    }
+  }
+
+  func testUTF16Surrogates2() {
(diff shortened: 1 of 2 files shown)


Tests/SwiftParserTest/LexerTests.swift
```diff
@@ -1080,6 +1080,41 @@ public class LexerTests: ParserTestCase {
     }
   }
 
+  func testUTF16Surrogates_1() {
```
ahoppen, line 1083: Nitpick: In other placed of the codebase where we have two similar test cases that are just numbered, we don’t use underscores.

```suggestion
  func testUTF16Surrogates1() {
```

Same for the `_2` test case below.

Tests/SwiftParserTest/LexerTests.swift
```diff
@@ -1080,6 +1080,41 @@ public class LexerTests: ParserTestCase {
...
+    // 0xD800 <= (UTF16 Surrogates value) <= 0xDFFF
+    let sourceBytes: [UInt8] = [0xED, 0xA0, 0x80]  // 0xD800
```
ahoppen, line 1085: Also a little nitpicky but I think we could improve clarity here by changing `// 0xD800` to `The bytes represent the code point 0xD800` (I think that code point is the correct term here, but I might be wrong).

Same below

ahoppen (commented):
Nice. Thank you

Tests/SwiftParserTest/LexerTests.swift
pinkjuice66: `Code point` is correct here, and I found that we normally suffix it with `U+`, for example, `U+D800`, when referring to code points.

Tests/SwiftParserTest/LexerTests.swift
pinkjuice66: I might have been confused by the test case named `testInvalidUtf8_3`, but I believe it's the only exception.
I changed its name to not use an underscore.
@ahoppen

ahoppen (approved):
Thank you. Looks good 👍🏽

pinkjuice66:
Would you mind running the CI?
@ahoppen

ahoppen:
@swift-ci Please test

pinkjuice66:
I believe we're ready to merge this.
@ahoppen

ahoppen:
Oh, sorry. I forgot to merge this.

PR #2440, opened 2024-01-24
Make `SyntaxArena.hasParent` an atomic boolean

ahoppen:
This fixes a concurrency issue in `SyntaxArena`.

Building `_AtomicBool` using CMake is more tricky because the CMake infrastructure in swift-syntax doesn’t know how to build C modules. Since the compiler is only single-threaded race conditions are impossible, so we fake `AtomicBool` for now in CMake builds.

rdar://121324116

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,5 +88,9 @@ if (SWIFTSYNTAX_ENABLE_ASSERTIONS)
   )
 endif()
 
+add_compile_definitions(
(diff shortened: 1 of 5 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test macOS

ahoppen:
@swift-ci Please test Windows

rintaro (approved):

PR #2441, opened 2024-01-24
[SwiftSyntax] Refine several aspects of BumpPtrAllocator.

pinkjuice66:
While skimming the `BumpPtrAllocator` class, I identified areas for improvement.
- Rename `slapSize` to `initialSlapSize` :  The variable name `slabSize` is quite confusing because it accidently implies that the allocated amount at a time is fixed, which is not the case. `BumpPtrAllocator` calculates the next size to be allocated based on the current index of the slabs array. So, I think we should clarify that it's only used initially.
- Remove unnecessary cleanup : Removed some unnecessary cleanup in the deinitializer of `BumpPtrAllocator`.
- Clarify the conditions for allocating a custom slab : We should check it based on the `slabSize`, not the `initialSlabSize`.

### DIFF
--- a/Sources/SwiftSyntax/BumpPtrAllocator.swift
+++ b/Sources/SwiftSyntax/BumpPtrAllocator.swift
@@ -11,6 +11,7 @@
 //===----------------------------------------------------------------------===//
 
 /// A ``BumpPtrAllocator`` that allocates `slabSize` at a time.
+/// `slabSize` initiates with `initialSlabSize` and doubles periodically as allocations occur.
 /// Once all memory in a slab has been used, it allocates a new slab and no
 /// memory allocations are necessary until that slab is completely filled up.
 @_spi(BumpPtrAllocator) @_spi(Testing)
@@ -20,8 +21,7 @@ public class BumpPtrAllocator {
   static private let GROWTH_DELAY: Int = 128
   static private let SLAB_ALIGNMENT: Int = 8
 
-  /// Initial slab size.
-  private var slabSize: Int
+  private let initialSlabSize: Int
 
   private var slabs: [Slab]
   /// Pair of pointers in the current slab.
@@ -38,8 +38,8 @@ public class BumpPtrAllocator {
   private var _totalBytesAllocated: Int
 
   /// Construct a new ``BumpPtrAllocator``.
-  public init(slabSize: Int) {
-    self.slabSize = slabSize
+  public init(initialSlabSize: Int) {
+    self.initialSlabSize = initialSlabSize
     slabs = []
     current = nil
     customSizeSlabs = []
@@ -61,7 +61,7 @@ public class BumpPtrAllocator {
   /// Calculate the size of the slab at the index.
   private func slabSize(at index: Int) -> Int {
     // Double the slab size every 'GROWTH_DELAY' slabs.
-    return self.slabSize * (1 << min(30, index / Self.GROWTH_DELAY))
+    return self.initialSlabSize * (1 << min(30, index / Self.GROWTH_DELAY))
   }
 
   private func startNewSlab() {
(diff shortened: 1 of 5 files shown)


Sources/SwiftSyntax/BumpPtrAllocator.swift
```diff
@@ -113,8 +111,9 @@ public class BumpPtrAllocator {
       return allocated
     }
 
-    // If the size is too big, allocate a dedicated slab for it.
-    if byteCount >= self.slabSize {
+    // If the size is too large to be contained in the new slab,
+    // allocate a dedicated slab for it.
+    if byteCount > self.slabSize(at: slabs.count) {
```
rintaro, line 116: I'm not sure about this, the current semantics of "custom size slab" for _reasonably_ big allocations.

For example, say the init slab size is `1024`, the current slab size is `4096`, and we already used `128` bytes of the current slab. If the next allocation requested `4000` bytes, the current slab cannot hold it, the next can, but if we do that, we'd waste `3968` bytes of the current slab. Custom slab avoid that kind of waste.

ahoppen (commented):
Looks good to me but I agree with @rintaro’s comment.

Sources/SwiftSyntax/BumpPtrAllocator.swift
pinkjuice66: That makes sense. 
But in the case of your example, if the allocator continues to be requested `4000` bytes, a memory allocation occurs as per the request, and I think this should be avoided. I'm not sure if that could be problematic in our use cases because I'm quite new to this project. If so, I suppose we need a more sophisticated mechanism to ensure that we minimize memory waste and avoid frequent memory allocations.
If that's not be problem in our use cases, do you think it would be better to clarify that we do our best not to waste memory here? If so, I'll add some comments.
@rintaro

rintaro: Each object allocation SwiftSyntax is below 64 bytes. Layout/collection data can be long but it's just (pointer size * number-of-elements), so several hundreds won't be an issue in ParsingArena (slabSize == 4096). Another one is the source buffer which totally makes sense to use a custom size slab. For non-parsing arena (sizeSize == 128) which usually used for editing, memory allocation is not a dominant factor in performance, afaik.
We also use BumpPtrAllocator in ASTGen, but it's similar.

We could make an option to change the behavior, but I don't think it's worth it at this point.

(FWIW, the current behavior resembles [the default behavior of `llvm::BumpPtrAllocator`'s `SizeThreshold` which defaults to `SlabSize`](https://github.com/apple/llvm-project/blob/1fbdb6c1d020d27c18546f422c7b7f5c781aed4c/llvm/include/llvm/Support/Allocator.h#L61-L62) )

rintaro:
@swift-ci Please test

rintaro:
```
/home/build-user/swift/lib/ASTGen/Sources/ASTGen/ASTGen.swift:79:66: error: incorrect argument label in call (have 'slabSize:', expected 'initialSlabSize:')
  fileprivate let allocator: SwiftSyntax.BumpPtrAllocator = .init(slabSize: 256)
                                                                 ^~~~~~~~~
                                                                  initialSlabSize
```
@pinkjuice66  Please update ASTGen too.

pinkjuice66:
I suppose ASTGen.swift resides in the [swift repository](https://github.com/apple/swift).
Should I open a pull request on the swift repository first and then merge this pull request on our side?
I believe their CI might also fail. What would you suggest I do in this situation?
@rintaro

rintaro:
`@swift-ci` can do cross repo PR testing. Could you just open a PR for ASTGen changes in swift repo?

rintaro:
https://github.com/apple/swift/pull/71210
@swift-ci Please test

PR #2442, opened 2024-01-24
Emit diagnostic when missing label in function call

kimdv:


Sources/SwiftParser/Expressions.swift
```diff
@@ -1861,6 +1861,10 @@ extension Parser {
       if self.atArgumentLabel(allowDollarIdentifier: true) && self.peek(isAt: .colon) {
         (unexpectedBeforeLabel, label) = parseArgumentLabel()
         colon = consumeAnyToken()
+      } else if let _colon = self.consume(if: .colon) {
```
kimdv, line 1864: I added `_colon` here otherwise it will clash with name with on 1860

ahoppen (approved):
Thanks. LGTM

ahoppen:
@swift-ci Please test

PR #2447, opened 2024-01-29
[Macros] Checking expression macro as caller-side default argument

ApolloZhu:
Companion for https://github.com/apple/swift/pull/70602 to allow only literals as arguments in a macro expression used as a default argument

## Note

This also impact accepted literals as mentioned in SE-0382 that

- Negative numbers are now accepted
- String interpolations are no longer accepted, even if all the interpolated expressions are literals or parameter references

### DIFF
--- a/Sources/SwiftSyntaxMacroExpansion/MacroArgument.swift
+++ b/Sources/SwiftSyntaxMacroExpansion/MacroArgument.swift
@@ -1,0 +1,138 @@
+import SwiftDiagnostics
+import SwiftSyntax
+
+enum DeclReferenceError: DiagnosticMessage {
+  case nonLiteral
+
+  var message: String {
+    switch self {
+    case .nonLiteral:
+      return "only literals are permitted"
+    }
+  }
+
+  var diagnosticID: MessageID {
+    .init(domain: "SwiftMacros", id: "\(self)")
+  }
+
+  var severity: DiagnosticSeverity {
+    .error
+  }
+}
+
+class CheckDeclReferenceVisitor: SyntaxAnyVisitor {
+  var diagnostics: [Diagnostic] = []
+
+  init() {
+    super.init(viewMode: .fixedUp)
+  }
+
+  // Integer literals
+  override func visit(_ node: IntegerLiteralExprSyntax) -> SyntaxVisitorContinueKind {
+    .visitChildren
+  }
+
+  // Floating point literals
+  override func visit(_ node: FloatLiteralExprSyntax) -> SyntaxVisitorContinueKind {
+    .visitChildren
+  }
+
+  // Negative numbers
+  override func visit(_ node: PrefixOperatorExprSyntax) -> SyntaxVisitorContinueKind {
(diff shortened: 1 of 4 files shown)


Sources/SwiftSyntaxMacroExpansion/MacroArgument.swift
```diff
@@ -0,0 +1,138 @@
+import SwiftDiagnostics
```
ahoppen, line 1: Missing copyright header.

Sources/SwiftSyntaxMacroExpansion/MacroArgument.swift
```diff
@@ -0,0 +1,138 @@
...
+        message: DeclReferenceError.nonLiteral
+      )
+    )
+
+    return .skipChildren
+  }
+}
+
+extension MacroExpansionExprSyntax {
+  public func checkDefaultArgumentMacroExpression() throws {
```
ahoppen, line 130: Does this need to be public? If it only needs to be public for testing or the compiler could you add `@_spi(Compiler)` and/or `@_spi(Testing)`?

Sources/SwiftSyntaxMacroExpansion/MacroArgument.swift
```diff
@@ -0,0 +1,138 @@
...
+  var diagnosticID: MessageID {
+    .init(domain: "SwiftMacros", id: "\(self)")
+  }
+
+  var severity: DiagnosticSeverity {
+    .error
+  }
+}
+
+class CheckDeclReferenceVisitor: SyntaxAnyVisitor {
```
ahoppen, line 23: I don’t think this name is quite accurate. I would prefer something like `OnlyLiteralExprChecker` or something of the kind because it also rejects eg. `{ return 1 }()`, which doesn’t contain any reference. 

Also a short doc comment would be appreciated.

Tests/SwiftSyntaxMacroExpansionTest/MacroArgumentTests.swift
```diff
@@ -0,0 +1,55 @@
...
+
+    let diags: [Diagnostic]
+    do {
+      try macro.as(MacroExpansionExprSyntax.self)!
+        .checkDefaultArgumentMacroExpression()
+      XCTFail("should have failed with an error")
+      fatalError()
+    } catch let diagError as DiagnosticsError {
+      diags = diagError.diagnostics
+    }
```
ahoppen, line 45: You should be able to use `XCTAssertError` here. That way you don’t need to `fatalError` and abort test execution.

Sources/SwiftSyntaxMacroExpansion/MacroArgument.swift
ApolloZhu: Changed to `@_spi(Compiler)`

ahoppen (approved):

ApolloZhu:
@swift-ci please test

ahoppen:
@swift-ci please test Windows

PR #2449, opened 2024-01-30
[Parser] Handle `deinit` and `subscript` as declaration name references

DougGregor:
Treat `deinit` and `subscript` the same way we treat `init` when parsing an expression, treating it as a declaration reference expression.

Fixes rdar://121687968.

### DIFF
--- a/Sources/SwiftParser/Names.swift
+++ b/Sources/SwiftParser/Names.swift
@@ -66,7 +66,7 @@ extension Parser {
   mutating func parseDeclReferenceExpr(_ flags: DeclNameOptions = []) -> RawDeclReferenceExprSyntax {
     // Consume the base name.
     let base: RawTokenSyntax
-    if let identOrSelf = self.consume(if: .identifier, .keyword(.self), .keyword(.Self)) ?? self.consume(if: .keyword(.`init`)) {
+    if let identOrSelf = self.consume(if: .identifier, .keyword(.self), .keyword(.Self)) ?? self.consume(if: .keyword(.`init`)) ?? self.consume(if: .keyword(.`deinit`)) ?? self.consume(if: .keyword(.`subscript`)) {
       base = identOrSelf
     } else if flags.contains(.operators), let (_, _) = self.at(anyIn: Operator.self) {
       base = self.consumeAnyToken(remapping: .binaryOperator)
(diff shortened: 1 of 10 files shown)


DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test Windows

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test Windows

DougGregor:
@swift-ci please test Windows

DougGregor:
@swift-ci please test macOS

Sources/SwiftParser/Names.swift
```diff
@@ -66,7 +66,9 @@ extension Parser {
   mutating func parseDeclReferenceExpr(_ flags: DeclNameOptions = []) -> RawDeclReferenceExprSyntax {
     // Consume the base name.
     let base: RawTokenSyntax
-    if let identOrSelf = self.consume(if: .identifier, .keyword(.self), .keyword(.Self)) ?? self.consume(if: .keyword(.`init`)) {
+    if let identOrSelf = self.consume(if: .identifier, .keyword(.self), .keyword(.Self)) ?? self.consume(if: .keyword(.`init`)) ?? self.consume(
+      if: .keyword(.`deinit`)
+    ) ?? self.consume(if: .keyword(.`subscript`)) {
```
ahoppen, line 71: Would be better written as 

```swift
   if let identOrSelf = self.consume(if: .identifier, .keyword(.self), .keyword(.Self)) ?? 
self.consume(if: .keyword(.`init`), .keyword(.`deinit`), .keyword(.`subscript`)) {
```

ahoppen (commented):
This changes existing parsing behavior in an incompatible way. 

For example in the following, the `subscript` after `self` used to get parsed as an identifier and it’s now parsed as a keyword, which would imply that it referred to a subscript, not a function named `subscript`.

```swift
struck Foo {
  func `subscript`() {}
  func test() {
    self.subscript()
  }
}
```

What we should do instead is:
- Introduce a new `ExprFlavor` mode for attribute parsing
- Inside that `ExprFlavor` mode, parse allow `subscript` and `deinit` as top-level base names of `DeclReferenceExprSyntax`. `something.subscript` should still parse `subscript` as an identifier.

DougGregor:
> What we should do instead is:
> 
> * Introduce a new `ExprFlavor` mode for attribute parsing
> * Inside that `ExprFlavor` mode, parse allow `subscript` and `deinit` as top-level base names of `DeclReferenceExprSyntax`. `something.subscript` should still parse `subscript` as an identifier.

This is a great approach. I've adopted it in https://github.com/apple/swift-syntax/pull/2453

PR #2450, opened 2024-01-31
[macros] prepare generic arguments for replacement in macros

ktoso:
Swift-syntax changes necessary of allowing the following code:

```swift
@freestanding(expression)
public macro checkGeneric_root<DAS>() = #externalMacro(module: "MacroDefinition", type: "GenericToVoidMacro")

@freestanding(expression)
public macro checkGeneric<DAS>() = #checkGeneric_root<DAS>()
```

I didn't quite implement it entirely right and need to add a test here as well.

Resolves rdar://122004157

### DIFF
--- a/Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
+++ b/Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
@@ -299,17 +299,21 @@ extension MacroDeclSyntax {
 /// uses of macro parameters with their corresponding arguments.
 private final class MacroExpansionRewriter: SyntaxRewriter {
   let parameterReplacements: [DeclReferenceExprSyntax: Int]
-  let genericParameterReplacements: [DeclReferenceExprSyntax: Int]
   let arguments: [ExprSyntax]
+  // let genericParameterReplacements: [DeclReferenceExprSyntax: Int]
+  let genericParameterReplacements: [GenericArgumentSyntax: Int]
+  let genericArguments: [TypeSyntax]
 
   init(
     parameterReplacements: [DeclReferenceExprSyntax: Int],
-    genericReplacements: [DeclReferenceExprSyntax: Int],
-    arguments: [ExprSyntax]
+    arguments: [ExprSyntax],
+    genericReplacements: [GenericArgumentSyntax: Int],
+    genericArguments: [TypeSyntax]
   ) {
     self.parameterReplacements = parameterReplacements
-    self.genericParameterReplacements = genericReplacements
     self.arguments = arguments
+    self.genericParameterReplacements = genericReplacements
+    self.genericArguments = genericArguments
     super.init(viewMode: .sourceAccurate)
   }
 
@@ -321,6 +325,25 @@ private final class MacroExpansionRewriter: SyntaxRewriter {
     // Swap in the argument for this parameter
     return arguments[parameterIndex].trimmed
   }
+
+  override func visit(_ node: GenericArgumentSyntax) -> GenericArgumentSyntax {
+    guard let parameterIndex = genericParameterReplacements[node] else {
+      return super.visit(node)
+    }
+
+    // Swap in the argument for type parameter
+    return GenericArgumentSyntax(
+      leadingTrivia: node.leadingTrivia,
+      node.unexpectedBeforeArgument,
+      argument: genericArguments[parameterIndex].trimmed,
+      node.unexpectedBetweenArgumentAndTrailingComma,
+      trailingComma: node.trailingComma,
+      node.unexpectedAfterTrailingComma
+      // TODO: seems we're getting spurious trailing " " here,
+      //  skipping trailing trivia for now
+      // trailingTrivia: node.trailingTrivia
+    )
+  }
 }
 
 extension MacroDeclSyntax {
@@ -328,30 +351,42 @@ extension MacroDeclSyntax {
   /// argument list.
   private func expand(
     argumentList: LabeledExprListSyntax?,
+    genericArgumentList: GenericArgumentClauseSyntax?,
     definition: MacroExpansionExprSyntax,
     replacements: [MacroDefinition.Replacement],
-    genericReplacements: [MacroDefinition.GenericArgumentReplacement]
+    genericReplacements: [MacroDefinition.GenericArgumentReplacement] = []
   ) -> ExprSyntax {
     // FIXME: Do real call-argument matching between the argument list and the
     // macro parameter list, porting over from the compiler.
+    let parameterReplacements = Dictionary(
+        uniqueKeysWithValues: replacements.map { replacement in
+          (replacement.reference, replacement.parameterIndex)
+        }
+      )
     let arguments: [ExprSyntax] =
       argumentList?.map { element in
         element.expression
       } ?? []
 
-    return MacroExpansionRewriter(
-      parameterReplacements: Dictionary(
-        uniqueKeysWithValues: replacements.map { replacement in
-          (replacement.reference, replacement.parameterIndex)
-        }
-      ),
-      genericReplacements: Dictionary(
-        uniqueKeysWithValues: replacements.map { replacement in
-          (replacement.reference, replacement.parameterIndex)
-        }
-      ),
-      arguments: arguments
-    ).visit(definition)
+
+    let genericReplacements = Dictionary(
+      uniqueKeysWithValues: genericReplacements.map { replacement in
+        (replacement.reference, replacement.parameterIndex)
+      }
+    )
+    let genericArguments: [TypeSyntax] =
+      genericArgumentList?.arguments.map { element in 
+        element.argument
+      } ?? []
+
+
+    let rewriter: MacroExpansionRewriter = MacroExpansionRewriter(
+      parameterReplacements: parameterReplacements,
+      arguments: arguments,
+      genericReplacements: genericReplacements,
+      genericArguments: genericArguments
+    )
+    return rewriter.visit(definition)
   }
 
   /// Given a freestanding macro expansion syntax node that references this
@@ -365,6 +400,7 @@ extension MacroDeclSyntax {
   ) -> ExprSyntax {
     return expand(
       argumentList: node.arguments,
+      genericArgumentList: node.genericArgumentClause,
       definition: definition,
       replacements: replacements,
       genericReplacements: genericReplacements
@@ -390,6 +426,7 @@ extension MacroDeclSyntax {
 
     return expand(
       argumentList: argumentList,
+      genericArgumentList: .init(arguments: []),
       definition: definition,
       replacements: replacements,
       genericReplacements: genericReplacements
--- a/Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
+++ b/Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
@@ -331,6 +331,10 @@ private final class MacroExpansionRewriter: SyntaxRewriter {
       return super.visit(node)
     }
 
+    guard parameterIndex < genericArguments.count else {
+      return super.visit(node)
+    }
+
     // Swap in the argument for type parameter
     return GenericArgumentSyntax(
       leadingTrivia: node.leadingTrivia,
@@ -339,9 +343,9 @@ private final class MacroExpansionRewriter: SyntaxRewriter {
       node.unexpectedBetweenArgumentAndTrailingComma,
       trailingComma: node.trailingComma,
       node.unexpectedAfterTrailingComma
-      // TODO: seems we're getting spurious trailing " " here,
-      //  skipping trailing trivia for now
-      // trailingTrivia: node.trailingTrivia
+        // TODO: seems we're getting spurious trailing " " here,
+        //  skipping trailing trivia for now
+        // trailingTrivia: node.trailingTrivia
     )
   }
 }
@@ -359,15 +363,14 @@ extension MacroDeclSyntax {
     // FIXME: Do real call-argument matching between the argument list and the
     // macro parameter list, porting over from the compiler.
     let parameterReplacements = Dictionary(
-        uniqueKeysWithValues: replacements.map { replacement in
-          (replacement.reference, replacement.parameterIndex)
-        }
-      )
+      uniqueKeysWithValues: replacements.map { replacement in
+        (replacement.reference, replacement.parameterIndex)
+      }
+    )
     let arguments: [ExprSyntax] =
       argumentList?.map { element in
         element.expression
       } ?? []
-
 
     let genericReplacements = Dictionary(
       uniqueKeysWithValues: genericReplacements.map { replacement in
@@ -375,10 +378,9 @@ extension MacroDeclSyntax {
       }
     )
     let genericArguments: [TypeSyntax] =
-      genericArgumentList?.arguments.map { element in 
+      genericArgumentList?.arguments.map { element in
         element.argument
       } ?? []
-
 
     let rewriter: MacroExpansionRewriter = MacroExpansionRewriter(
       parameterReplacements: parameterReplacements,
--- a/Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
+++ b/Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
@@ -18,7 +18,7 @@ enum MacroExpanderError: DiagnosticMessage {
   case undefined
   case definitionNotMacroExpansion
   case nonParameterReference(TokenSyntax)
-  case nonTypeReference(TypeSyntax)
+  case nonTypeReference(TokenSyntax)
   case nonLiteralOrParameter(ExprSyntax)
 
   var message: String {
@@ -187,13 +187,17 @@ fileprivate class ParameterReplacementVisitor: SyntaxAnyVisitor {
   }
 
   override func visit(_ node: GenericArgumentSyntax) -> SyntaxVisitorContinueKind {
-    let baseName = node.argument
+    guard let baseName = node.argument.as(IdentifierTypeSyntax.self)?.name else {
+      // Handle error
+      return .visitChildren
+    }
+
     guard let genericParameterClause = macro.genericParameterClause else {
       return .skipChildren
     }
 
-    let parameterIndex = genericParameterClause.parameters.firstIndex { (index, parameter) in
-      return parameter.name.text == "\(baseName)"
+    let matchedParameter = genericParameterClause.parameters.enumerated().first { (index, parameter) in
+      return parameter.name.text == baseName.text
     }
 
     guard let (parameterIndex, _) = matchedParameter else {
--- a/Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
+++ b/Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
@@ -18,6 +18,7 @@ enum MacroExpanderError: DiagnosticMessage {
   case undefined
   case definitionNotMacroExpansion
   case nonParameterReference(TokenSyntax)
+  case nonTypeReference(TypeSyntax)
   case nonLiteralOrParameter(ExprSyntax)
 
   var message: String {
@@ -30,6 +31,9 @@ enum MacroExpanderError: DiagnosticMessage {
 
     case .nonParameterReference(let name):
       return "reference to value '\(name.text)' that is not a macro parameter in expansion"
+
+    case .nonTypeReference(let name):
+      return "reference to type '\(name)' that is not a macro type parameter in expansion"
 
     case .nonLiteralOrParameter:
       return "only literals and macro parameters are permitted in expansion"
@@ -58,7 +62,14 @@ public enum MacroDefinition {
   /// defining macro. These subtrees will need to be replaced with the text of
   /// the corresponding argument to the macro, which can be accomplished with
   /// `MacroDeclSyntax.expandDefinition`.
-  case expansion(MacroExpansionExprSyntax, replacements: [Replacement], genericReplacements: [GenericArgumentReplacement]) // FIXME: do we need to evolve this without breaking? I assume so?
+  case expansion(MacroExpansionExprSyntax, replacements: [Replacement], genericReplacements: [GenericArgumentReplacement])
+}
+
+extension MacroDefinition {
+  /// Best effort compatibility shim, the case has gained additional parameters.
+  public func expansion(_ node: MacroExpansionExprSyntax, replacements: [Replacement]) -> Self {
+    .expansion(node, replacements: replacements, genericReplacements: [])
+  }
 }
 
 extension MacroDefinition {
@@ -175,27 +186,23 @@ fileprivate class ParameterReplacementVisitor: SyntaxAnyVisitor {
   }
 
   override func visit(_ node: GenericArgumentSyntax) -> SyntaxVisitorContinueKind {
-    let baseName = "\(node)" // FIXME: where to get the name from properly
+    let baseName = node.argument
     guard let genericParameterClause = macro.genericParameterClause else {
       return .skipChildren
     }
 
     let matchedParameter = genericParameterClause.parameters.enumerated().first { (index, parameter) in
-//      guard let parameterName = parameter.parameterName else {
-//        return false
-//      }
-
-      return true; // FIXME: what do we need to match here
+      return parameter.name.text == "\(baseName)"
     }
 
     guard let (parameterIndex, _) = matchedParameter else {
       // We have a reference to something that isn't a parameter of the macro.
-//      diagnostics.append(
-//        Diagnostic(
-//          node: Syntax(baseName), // FIXME: error should be about type argument
-//          message: MacroExpanderError.nonParameterReference(baseName)
-//        )
-//      )
+     diagnostics.append(
+       Diagnostic(
+         node: Syntax(baseName),
+         message: MacroExpanderError.nonTypeReference(baseName)
+       )
+     )
 
       return .visitChildren
     }
@@ -287,10 +294,14 @@ extension MacroDeclSyntax {
 /// uses of macro parameters with their corresponding arguments.
 private final class MacroExpansionRewriter: SyntaxRewriter {
   let parameterReplacements: [DeclReferenceExprSyntax: Int]
+  let genericParameterReplacements: [DeclReferenceExprSyntax: Int]
   let arguments: [ExprSyntax]
 
-  init(parameterReplacements: [DeclReferenceExprSyntax: Int], arguments: [ExprSyntax]) {
+  init(parameterReplacements: [DeclReferenceExprSyntax: Int],
+       genericReplacements: [DeclReferenceExprSyntax: Int],
+       arguments: [ExprSyntax]) {
     self.parameterReplacements = parameterReplacements
+    self.genericParameterReplacements = genericReplacements
     self.arguments = arguments
     super.init(viewMode: .sourceAccurate)
   }
@@ -311,7 +322,8 @@ extension MacroDeclSyntax {
   private func expand(
     argumentList: LabeledExprListSyntax?,
     definition: MacroExpansionExprSyntax,
-    replacements: [MacroDefinition.Replacement]
+    replacements: [MacroDefinition.Replacement],
+    genericReplacements: [MacroDefinition.GenericArgumentReplacement]
   ) -> ExprSyntax {
     // FIXME: Do real call-argument matching between the argument list and the
     // macro parameter list, porting over from the compiler.
@@ -326,6 +338,11 @@ extension MacroDeclSyntax {
           (replacement.reference, replacement.parameterIndex)
         }
       ),
+      genericReplacements: Dictionary(
+        uniqueKeysWithValues: replacements.map { replacement in
+          (replacement.reference, replacement.parameterIndex)
+        }
+      ),
       arguments: arguments
     ).visit(definition)
   }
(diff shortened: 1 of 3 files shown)


ktoso:
@swift-ci please smoke test

Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
```diff
@@ -58,7 +58,7 @@ public enum MacroDefinition {
   /// defining macro. These subtrees will need to be replaced with the text of
   /// the corresponding argument to the macro, which can be accomplished with
   /// `MacroDeclSyntax.expandDefinition`.
-  case expansion(MacroExpansionExprSyntax, replacements: [Replacement])
+  case expansion(MacroExpansionExprSyntax, replacements: [Replacement], genericReplacements: [GenericArgumentReplacement]) // FIXME: do we need to evolve this without breaking? I assume so?
```
ktoso, line 61: FIXME: I assume we probably can't change this and need to add a new case since we need to be compatible with various versions of compiler?

(I have no idea if we give such guarantees here)

Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
```diff
@@ -156,6 +166,45 @@ fileprivate class ParameterReplacementVisitor: SyntaxAnyVisitor {
...
+    }
+
+    guard let (parameterIndex, _) = matchedParameter else {
+      // We have a reference to something that isn't a parameter of the macro.
+//      diagnostics.append(
+//        Diagnostic(
+//          node: Syntax(baseName), // FIXME: error should be about type argument
+//          message: MacroExpanderError.nonParameterReference(baseName)
+//        )
+//      )
```
ktoso, line 198: TODO: cause a warning

Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
rintaro: `SwiftSyntaxMacroExpansion` module is intended to be only used by the compiler that's built at the same time. We don't need to worry about the version compatibility as long as `swift` and `swift-syntax` repo are synced when building.

ahoppen: It would definitely need to be release-noted though because it’s changing public API and it might make sense to define a compatibility layer to minimize the breakage.

```swift
static func expansion(MacroExpansionExprSyntax) -> MacroDefinition
```

ktoso: Ok, I'll add a func like that -- it'll still break code exhaustively switching over the enum, but if the compiler is expected to revlock that's good.

ahoppen: Yes, the static func is just to reduce the number cases where it breaks clients to a minimum.

ktoso:
Thanks folks! I added the notes, and a test covering the new functionality as well.

Please have a look 👀

ktoso:
@swift-ci please smoke test

ktoso:
@swift-ci Please test

Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
```diff
@@ -58,7 +62,14 @@ public enum MacroDefinition {
   /// defining macro. These subtrees will need to be replaced with the text of
   /// the corresponding argument to the macro, which can be accomplished with
   /// `MacroDeclSyntax.expandDefinition`.
-  case expansion(MacroExpansionExprSyntax, replacements: [Replacement])
+  case expansion(MacroExpansionExprSyntax, replacements: [Replacement], genericReplacements: [GenericArgumentReplacement])
+}
+
+extension MacroDefinition {
+  /// Best effort compatibility shim, the case has gained additional parameters.
+  public func expansion(_ node: MacroExpansionExprSyntax, replacements: [Replacement]) -> Self {
```
ahoppen, line 70: Let’s add a deprecation message to the compatibility shim.

```suggestion
  @available(*, deprecated, message: "Use the expansion case with three associated values instead")
  public func expansion(_ node: MacroExpansionExprSyntax, replacements: [Replacement]) -> Self {
```

Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
```diff
@@ -156,6 +177,41 @@ fileprivate class ParameterReplacementVisitor: SyntaxAnyVisitor {
...
+
+  override func visit(_ node: GenericArgumentSyntax) -> SyntaxVisitorContinueKind {
+    let baseName = node.argument
+    guard let genericParameterClause = macro.genericParameterClause else {
+      return .skipChildren
+    }
+
+    let matchedParameter = genericParameterClause.parameters.enumerated().first { (index, parameter) in
+      return parameter.name.text == "\(baseName)"
+    }
```
ahoppen, line 196: You should be able to use `firstIndex(where:)`

```suggestion
    let parameterIndex = genericParameterClause.parameters.firstIndex { (index, parameter) in
      return parameter.name.text == "\(baseName)"
    }
```

Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
```diff
@@ -156,6 +177,41 @@ fileprivate class ParameterReplacementVisitor: SyntaxAnyVisitor {
...
+  }
+
+  override func visit(_ node: GenericArgumentSyntax) -> SyntaxVisitorContinueKind {
+    let baseName = node.argument
+    guard let genericParameterClause = macro.genericParameterClause else {
+      return .skipChildren
+    }
+
+    let matchedParameter = genericParameterClause.parameters.enumerated().first { (index, parameter) in
+      return parameter.name.text == "\(baseName)"
```
ahoppen, line 195: I think this will fail if you add spaces around `node.argument` because `"\(baseName)"` contains trivia while `parameter.name.text` does not.

You should probably do something like

```swift
guard let baseName node.argument.as(IdentifierTypeSyntax.self)?.name else {
  // Handle error
}
```

and then down here `baseName.text`.

Ie. I would expect a failing test case to be the following (both the space after `A` and the comment after `B` should trigger the failure.

```swift
macro gen<A , B /* some comment */>(a: A, b: B) = #otherMacro<A, B>(first: a, second: b)
```

Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
```diff
@@ -156,6 +177,41 @@ fileprivate class ParameterReplacementVisitor: SyntaxAnyVisitor {
     return .visitChildren
   }
 
+  override func visit(_ node: GenericArgumentClauseSyntax) -> SyntaxVisitorContinueKind {
+    return .visitChildren
+  }
+
```
ahoppen, line 183: IIUC you only want to visit the `GenericArgumentClauseSyntax` of the `MacroExpansion{Expr, Decl}Syntax`, right? I think we should check that the parent is that type. Otherwise you also walk into the `GenericArgumentClauseSyntax` of eg. a type. 

A test case I would expect to hit this is

```swift
macro gen(a: Array<Int>) = …
```

We shouldn’t walk the `GenericArgumentClauseSyntax` of `Array`.

Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
```diff
@@ -230,18 +286,22 @@ extension MacroDeclSyntax {
...
 /// uses of macro parameters with their corresponding arguments.
 private final class MacroExpansionRewriter: SyntaxRewriter {
   let parameterReplacements: [DeclReferenceExprSyntax: Int]
+  let genericParameterReplacements: [DeclReferenceExprSyntax: Int]
   let arguments: [ExprSyntax]
 
-  init(parameterReplacements: [DeclReferenceExprSyntax: Int], arguments: [ExprSyntax]) {
+  init(parameterReplacements: [DeclReferenceExprSyntax: Int],
+       genericReplacements: [DeclReferenceExprSyntax: Int],
+       arguments: [ExprSyntax]) {
```
ahoppen, line 302: Could you format this using swift-format? https://github.com/apple/swift-syntax/blob/main/CONTRIBUTING.md#formatting

CI will reject it otherwise

Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
```diff
@@ -277,6 +338,11 @@ extension MacroDeclSyntax {
           (replacement.reference, replacement.parameterIndex)
         }
       ),
+      genericReplacements: Dictionary(
+        uniqueKeysWithValues: replacements.map { replacement in
+          (replacement.reference, replacement.parameterIndex)
+        }
+      ),
```
ahoppen, line 345: A user might specify `genericReplacements` when calling `expand`. So we can’t assume that the keys are indeed unique. We should fail somewhat gracefully if they are not unique.

Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
```diff
@@ -287,12 +353,14 @@ extension MacroDeclSyntax {
   public func expand(
     _ node: some FreestandingMacroExpansionSyntax,
     definition: MacroExpansionExprSyntax,
-    replacements: [MacroDefinition.Replacement]
+    replacements: [MacroDefinition.Replacement],
+    genericReplacements: [MacroDefinition.GenericArgumentReplacement]
```
ahoppen, line 357: We should default `genericReplacements` to `[]`. Otherwise this is a source-breaking change.

Also, I think we should have a release note entry for the API addition of `genericReplacements`.

Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
```diff
@@ -302,7 +370,8 @@ extension MacroDeclSyntax {
   public func expand(
     _ node: AttributeSyntax,
     definition: MacroExpansionExprSyntax,
-    replacements: [MacroDefinition.Replacement]
+    replacements: [MacroDefinition.Replacement],
+    genericReplacements: [MacroDefinition.GenericArgumentReplacement]
```
ahoppen, line 374:
```suggestion
    genericReplacements: [MacroDefinition.GenericArgumentReplacement] = []
```

Tests/SwiftSyntaxMacroExpansionTest/MacroReplacementTests.swift
```diff
@@ -113,4 +114,52 @@ final class MacroReplacementTests: XCTestCase {
...
+    let definition = try macroDecl.checkDefinition()
+    guard case let .expansion(expansion, replacements, genericReplacements) = definition else {
+      XCTFail("not a normal expansion")
+      fatalError()
+    }
+
+    guard let replacementA = genericReplacements.first else {
+      XCTFail("Expected generic replacement for A")
+      fatalError()
+    }
```
ahoppen, line 139: We should not `fatalError` in test cases because that takes down the entire test suite when it fails. A `return` would be better. Or, even better use `XCTUnwrap` where possible.

```suggestion
    let replacementA = try XCTUnwrap(genericReplacements.first)
```

Same below and above.

Tests/SwiftSyntaxMacroExpansionTest/MacroReplacementTests.swift
```diff
@@ -113,4 +114,52 @@ final class MacroReplacementTests: XCTestCase {
...
+    let expandedSyntax = macroDecl.expand(
+      use.as(MacroExpansionExprSyntax.self)!,
+      definition: expansion,
+      replacements: replacements,
+      genericReplacements: genericReplacements
+    )
+    assertStringsEqualWithDiff(
+      expandedSyntax.description,
+      """
+      #otherMacro<A, B>(first: 5, second: "Hello")
```
ahoppen, line 161: Am I missing something or should `A` and `B` be replaced by `Int` and `String` so this would be

```swift
#otherMacro<Int, String>(first: 5, second: "Hello")
```

ktoso: I think the actual replacing only happens in Swift -- at least that's where the code to handle replacing the text is. I was confused by this actually, but this seems to work in Swift itself then 🤔

ktoso:
Thanks a lot for the review! I'll follow up shortly

Tests/SwiftSyntaxMacroExpansionTest/MacroReplacementTests.swift
ktoso: Aha Ben explained this to me -- it seems we have two macro expansion mechanisms huh... I'll implement it here too then

Tests/SwiftSyntaxMacroExpansionTest/MacroReplacementTests.swift
ktoso: Ok, sure - i was copying the style this file seems to be doing. I can return instead.

Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
ktoso: Oh thank you, I'll handle that

Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
ktoso: Weirdly enough this seems to just work already, but I'll adjust to use the `as ` as you suggested

ktoso: Done, thanks  -- added tests too

Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
ktoso: done thanks -- nice to see a swift repo actually use formatting, nice surprise :) 
(We do on all server libs, big fan of forcing it in CI 🙏)

Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
ktoso: ok done

Tests/SwiftSyntaxMacroExpansionTest/MacroReplacementTests.swift
ktoso: Implemented it over here as well

I was a bit confused about the trailing trivia attached... I left a TODO and would appreciate someone helping out on that bit.

@bnbarham

ktoso:
@swift-ci please smoke test

ktoso:
@swift-ci please smoke test

ktoso:
@swift-ci please smoke test

ktoso:
@swift-ci Please test

ktoso:
@swift-ci Please test

Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
ahoppen: I think you’re still using `first` and then only using the index. Is there any reason why `firstIndex` doesn’t work?

Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
ahoppen: AFAICT we don’t handle this yet. Or am I missing something?

Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
ahoppen: Still open AFAICT

ktoso: Whoops, thank you

Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
ktoso: I did check the snippet and worked fine, I'll add more tests but I guess I'm not sure how to cause a bad case?

Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
ktoso: Using firstIndex I run into: 

```
/Users/ktoso/code/swift-project/swift-syntax/Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift:215:71: error: cannot convert value of type 'SyntaxChildrenIndex' to expected argument type 'Int'
    genericReplacements.append(.init(reference: node, parameterIndex: parameterIndex))
                                                                      ^
```

so kept first and getting the index integer; is there a better trick here?

Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
ktoso: Ok defaulted everywhere

Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
ktoso: For what it's worth other code uses the same pattern as we have here.

ktoso:
https://github.com/apple/swift/pull/71271
@swift-ci Please test

Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
ahoppen: Ah, yes. Seems to work. I didn’t realize we were only walking the definition and that it can’t contain complex types because it can only contain literals.

Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
```diff
@@ -156,6 +178,45 @@ fileprivate class ParameterReplacementVisitor: SyntaxAnyVisitor {
...
+    return .visitChildren
+  }
+
+  override func visit(_ node: GenericArgumentListSyntax) -> SyntaxVisitorContinueKind {
+    return .visitChildren
+  }
+
+  override func visit(_ node: GenericArgumentSyntax) -> SyntaxVisitorContinueKind {
+    guard let baseName = node.argument.as(IdentifierTypeSyntax.self)?.name else {
+      // Handle error
```
ahoppen, line 191: Is this a leftover todo?

Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
```diff
@@ -239,10 +300,20 @@ extension MacroDeclSyntax {
 private final class MacroExpansionRewriter: SyntaxRewriter {
   let parameterReplacements: [DeclReferenceExprSyntax: Int]
   let arguments: [ExprSyntax]
-
-  init(parameterReplacements: [DeclReferenceExprSyntax: Int], arguments: [ExprSyntax]) {
+  // let genericParameterReplacements: [DeclReferenceExprSyntax: Int]
```
ahoppen, line 303: Leftover comment?

Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
```diff
@@ -254,31 +325,72 @@ private final class MacroExpansionRewriter: SyntaxRewriter {
...
+    return GenericArgumentSyntax(
+      leadingTrivia: node.leadingTrivia,
+      node.unexpectedBeforeArgument,
+      argument: genericArguments[parameterIndex].trimmed,
+      node.unexpectedBetweenArgumentAndTrailingComma,
+      trailingComma: node.trailingComma,
+      node.unexpectedAfterTrailingComma
+        // TODO: seems we're getting spurious trailing " " here,
+        //  skipping trailing trivia for now
+        // trailingTrivia: node.trailingTrivia
```
ahoppen, line 348: Do you have a failing test case for this? I tried adding it again and none of the tests failed.

Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
```diff
@@ -254,31 +325,72 @@ private final class MacroExpansionRewriter: SyntaxRewriter {
...
+    )
```
ahoppen, line 349: You should be able to simplify this slightly to

```suggestion
var node = node
    node.argument = genericArguments[parameterIndex].trimmed
    return node
```

Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
```diff
@@ -254,31 +325,72 @@ private final class MacroExpansionRewriter: SyntaxRewriter {
...
+        (replacement.reference, replacement.parameterIndex)
+      },
+      uniquingKeysWith: { l, r in l }
+    )
+    let genericArguments: [TypeSyntax] =
+      genericArgumentList?.arguments.map { element in
+        element.argument
+      } ?? []
+
+    let rewriter: MacroExpansionRewriter = MacroExpansionRewriter(
```
ahoppen, line 387: Type annotation seems superfluous to me.

Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
```diff
@@ -254,31 +325,72 @@ private final class MacroExpansionRewriter: SyntaxRewriter {
...
+    let genericReplacements = Dictionary(
+      genericReplacements.map { replacement in
+        (replacement.reference, replacement.parameterIndex)
+      },
+      uniquingKeysWith: { l, r in l }
+    )
+    let genericArguments: [TypeSyntax] =
+      genericArgumentList?.arguments.map { element in
+        element.argument
+      } ?? []
```
ahoppen, line 385: Could be shortened slightly to

```suggestion
    let genericArguments: [TypeSyntax] = genericArgumentList?.arguments.map { $0.argument } ?? []
```

Tests/SwiftSyntaxMacroExpansionTest/MacroReplacementTests.swift
```diff
@@ -113,4 +114,230 @@ final class MacroReplacementTests: XCTestCase {
...
+      """
+      macro gen<A, B>(a: A, b: B) = #otherMacro<A, B>(first: a, second: b)
+      """
+
+    let use: ExprSyntax =
+      """
+      #gen<Int, String>(a: 5, b: "Hello")
+      """
+
+    let macroDecl = macro.as(MacroDeclSyntax.self)!
```
ahoppen, line 129:
```suggestion
    let macroDecl = macro.cast(MacroDeclSyntax.self)
```

Avoids the force unwrap

Tests/SwiftSyntaxMacroExpansionTest/MacroReplacementTests.swift
```diff
@@ -113,4 +114,230 @@ final class MacroReplacementTests: XCTestCase {
...
+    XCTAssertEqual(genericReplacements.count, 2)
+
+    XCTAssertEqual(replacementA.parameterIndex, 0)
+    XCTAssertEqual("\(replacementA.reference.argument)", "A")
+
+    XCTAssertEqual(replacementB.parameterIndex, 1)
+    XCTAssertEqual("\(replacementB.reference.argument)", "B")
+
+    let expandedSyntax = macroDecl.expand(
+      use.as(MacroExpansionExprSyntax.self)!,
```
ahoppen, line 148:
```suggestion
      use.cast(MacroExpansionExprSyntax.self),
```

Tests/SwiftSyntaxMacroExpansionTest/MacroReplacementTests.swift
```diff
@@ -113,4 +114,230 @@ final class MacroReplacementTests: XCTestCase {
...
+      """
+      macro gen<A, B /* some comment */>(a: A, b: B) = #otherMacro<A, B>(first: a, second: b)
+      """
+
+    let use: ExprSyntax =
+      """
+      #gen<Int, String>(a: 5, b: "Hello")
+      """
+
+    let macroDecl = macro.as(MacroDeclSyntax.self)!
```
ahoppen, line 172:
```suggestion
    let macroDecl = macro.cast(MacroDeclSyntax.self)
```

Tests/SwiftSyntaxMacroExpansionTest/MacroReplacementTests.swift
```diff
@@ -113,4 +114,230 @@ final class MacroReplacementTests: XCTestCase {
...
+    XCTAssertEqual(genericReplacements.count, 2)
+
+    XCTAssertEqual(replacementA.parameterIndex, 0)
+    XCTAssertEqual("\(replacementA.reference.argument)", "A")
+
+    XCTAssertEqual(replacementB.parameterIndex, 1)
+    XCTAssertEqual("\(replacementB.reference.argument)", "B")
+
+    let expandedSyntax = macroDecl.expand(
+      use.as(MacroExpansionExprSyntax.self)!,
```
ahoppen, line 190:
```suggestion
      use.cast(MacroExpansionExprSyntax.self),
```

Same a couple more times below

Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
ktoso: Cool, glad I didn't misunderstand then :)

Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
ktoso: Ah, good call

Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
ktoso: Oh, weird... yes I would get a weird extra space after `Int, [ ]String` like this. Let me check again hmm

Tests/SwiftSyntaxMacroExpansionTest/MacroReplacementTests.swift
```diff
@@ -113,4 +114,230 @@ final class MacroReplacementTests: XCTestCase {
       """
     )
   }
+
+  func testMacroGenericArgumentExpansion_base() throws {
```
ahoppen, line 118: We generally don’t use `_` in test names. I would prefer to name this `testMacroGenericArgumentExpansionBase`. Same for the other tests.

Tests/SwiftSyntaxMacroExpansionTest/MacroReplacementTests.swift
```diff
@@ -113,4 +114,230 @@ final class MacroReplacementTests: XCTestCase {
...
+
+  func testMacroExpansion_dontCrashOnDuplicates() throws {
+    let macro: DeclSyntax =
+      """
+      macro gen(a: Array<Int>) = #other<A>(first: a)
+      """
+
+    let use: ExprSyntax =
+      """
+      #otheren<Int>(a: [1, 2, 3])
```
ahoppen, line 313: Is this a typo and should this be the following?

```suggestion
      #other<Int>(a: [1, 2, 3])
```

Tests/SwiftSyntaxMacroExpansionTest/MacroReplacementTests.swift
ahoppen, line 1: Just an open question: Would it make sense to specialize macros with types that aren’t part of the second-level macro? Ie. would the following make sense?

```swift
macro gen<T>(a: T) = #externalMacro …
macro genString(a: String) = #gen<String>(a: a)
```

And then `#genString(a: "x")` would have the intermediate expansion step `#gen<String>(a: "x")`

---

Similarly, would it make sense to allow generic nesting in a second-level macro definition? Ie. would something like the following make sense?

```swift
macro gen<T>(a: T) = #externalMacro …
macro genArray<T>(a: [T]) = #gen<Array<T>>(a: a)
```

And then `#genArray<Int>(a: [1])` would have the intermediate expansion step `#gen<Array<Int>>(a: [1])`

AFAICT neither of those are currently supported.

Tests/SwiftSyntaxMacroExpansionTest/MacroReplacementTests.swift
ktoso: Ok I'll follow the repo style on it, thanks

Tests/SwiftSyntaxMacroExpansionTest/MacroReplacementTests.swift
ktoso: typo oh no, guess the test run is going to fail on this now 😮‍💨 

Gotta fix other things anyway, thanks

Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
ktoso: wasn't sure if to error or just return, I think we can skip children here and call it a day

Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
ktoso: Seems to pass, you're right -- I wonder what I messed up back then

Sources/SwiftSyntaxMacroExpansion/MacroReplacement.swift
ktoso: Wasn't sure what to do here, but actually best might be to just .skipChildren and call it a day

Tests/SwiftSyntaxMacroExpansionTest/MacroReplacementTests.swift
ktoso: Yeah these may be worth supporting as well -- I made a ticket for it https://github.com/apple/swift-syntax/issues/2476

Thanks for the thorough review Alex. I think we're good now, if CI gods are willing 😇

ktoso:
https://github.com/apple/swift/pull/71271
@swift-ci Please smoke test

ktoso:
https://github.com/apple/swift/pull/71271
@swift-ci Please test

PR #2451, opened 2024-01-31
protocol ListBuilder

RoBo-Inc:
Introducing the `ListBuilder` protocol with a default function implementation can significantly reduce the amount of generated code in the `Sources/SwiftSyntaxBuilder/generated/ResultBuilders.swift` file. Additionally, it makes the codebase more comprehensible by emphasizing similarities among all the result builders and highlighting differences between them.

Apart from this, the PR addresses another issue that requires detailed explanation. Two generated result builders (`ExprListBuilder` and `UnexpectedNodesBuilder`) stood out from the others. They were "special" because their `Expression` was not a concrete type; instead, it was a protocol (`ExprSyntaxProtocol` and `SyntaxProtocol` respectively). So we were dealing with existentials here, which are not very good. Furthermore, we had to create additional extensions dedicated to these two result builders, located at the bottom of the `Sources/SwiftSyntaxBuilder/ResultBuilderExtensions.swift` file, accompanied by a lengthy comment.

The key point is that these two result builders don't need to be treated as special cases, and they don't have to involve existentials. They can conform to the `ListBuilder` protocol, just like all the other builders do. Despite this, we can still make them work with protocols just as before, maintaining the same approach used for some other result builders in the same file (`Sources/SwiftSyntaxBuilder/ResultBuilderExtensions.swift`). This eliminates the need for special treatment and enhances codebase consistency.

In summary, this PR represents a refactoring. All the functionality remains unchanged. However, the refactoring is significant, making the code more structured, consistent, and efficient.

### DIFF
--- a/Sources/SwiftSyntaxBuilder/ConvenienceInitializers.swift
+++ b/Sources/SwiftSyntaxBuilder/ConvenienceInitializers.swift
@@ -13,7 +13,7 @@
 @_spi(RawSyntax) import SwiftParser
 @_spi(RawSyntax) import SwiftSyntax
 
-// MARK: - ArrayElementList
+// MARK: - ArrayElementListSyntax
 
 extension ArrayElementListSyntax {
   public init(expressions: [ExprSyntax]) {
@@ -30,7 +30,7 @@ extension ArrayElementListSyntax {
   }
 }
 
-// MARK: - ArrayExpr
+// MARK: - ArrayExprSyntax
 
 extension ArrayExprSyntax {
   public init(expressions: [ExprSyntax]) {
@@ -38,7 +38,7 @@ extension ArrayExprSyntax {
   }
 }
 
-// MARK: - CustomAttribute
+// MARK: - AttributeSyntax
 
 extension AttributeSyntax {
   /// A convenience initializer that allows passing in arguments using a result builder
@@ -58,7 +58,7 @@ extension AttributeSyntax {
   }
 }
 
-// MARK: - BinaryOperatorExpr
+// MARK: - BinaryOperatorExprSyntax
 
 extension BinaryOperatorExprSyntax {
   public init(text: String) {
@@ -66,7 +66,7 @@ extension BinaryOperatorExprSyntax {
   }
 }
 
-// MARK: - BooleanLiteralExpr
+// MARK: - BooleanLiteralExprSyntax
 
 extension BooleanLiteralExprSyntax: ExpressibleByBooleanLiteral {
   public init(_ value: Bool) {
@@ -78,7 +78,7 @@ extension BooleanLiteralExprSyntax: ExpressibleByBooleanLiteral {
   }
 }
 
-// MARK: - CatchClause
+// MARK: - CatchClauseSyntax
 
 extension CatchClauseSyntax {
   /// A convenience initializer that calculates spacing around the `catch` keyword.
@@ -96,7 +96,7 @@ extension CatchClauseSyntax {
   }
 }
 
-// MARK: - DictionaryExpr
+// MARK: - DictionaryExprSyntax
 
 extension DictionaryExprSyntax {
   /// A convenience initializer that allows passing in members using a result builder
@@ -115,7 +115,15 @@ extension DictionaryExprSyntax {
   }
 }
 
-// MARK: - Expr
+// MARK: - ExprListSyntax
+
+extension ExprListSyntax {
+  public init(_ elements: [ExprSyntaxProtocol]) {
+    self.init(elements.map { ExprSyntax(fromProtocol: $0) } as [ExprSyntax])
+  }
+}
+
+// MARK: - ExprSyntax
 
 extension ExprSyntax {
   /// Returns a syntax tree for an expression that represents the value of the
@@ -163,7 +171,7 @@ extension FloatLiteralExprSyntax: ExpressibleByFloatLiteral {
   }
 }
 
-// MARK: - FunctionCallExpr
+// MARK: - FunctionCallExprSyntax
 
 extension FunctionCallExprSyntax {
   /// A convenience initializer that allows passing in arguments using a result builder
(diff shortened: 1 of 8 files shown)


Sources/SwiftSyntaxBuilder/ResultBuilderExtensions.swift
```diff
@@ -50,20 +50,14 @@ extension MemberBlockItemListBuilder {
...
 
-// MARK: Initializing collections from protocols
-// These initializers allow the creation of syntax collections that have a base
-// node as their element from the corresponding protocol type.
-// These are used by the result builders.
-// Since we only have two of these, it doesn’t make sense to generate them.
-
-extension ExprListSyntax {
-  init(_ elements: [ExprSyntaxProtocol]) {
-    self = ExprListSyntax(elements.map { ExprSyntax(fromProtocol: $0) } as [ExprSyntax])
```
ahoppen, line 61: I think we still need this so that you can write eg. 

```swift
ExprListSyntax([DeclReferenceExprSyntax(…), ForceUnwrapExprSyntax(…)])`
```

Or any heterogeneous array of expressions, which is what we do in the tests you changed, I think.
Otherwise this is an API-breaking change.

Same for the initializer on `UnexpectedNodesSyntax` below.

Sources/SwiftSyntaxBuilder/generated/ResultBuilders.swift
```diff
@@ -15,82 +15,10 @@
...
-  /// which defaults to Component if buildExpression() is not provided.
-  public typealias Expression = AccessorDeclSyntax
-  
-  /// The type of a partial result, which will be carried through all of the
-  /// build methods.
-  public typealias Component = [Expression]
-  
+public struct AccessorDeclListBuilder: ListBuilder {
   /// The type of the final returned result, which defaults to Component if
   /// buildFinalResult() is not provided.
```
ahoppen, line 20: I don’t think theses comments provide much value anymore. I would just remove them.

Sources/SwiftSyntaxBuilder/generated/ResultBuilders.swift
```diff
@@ -188,93 +41,21 @@ public extension ArrayElementListSyntax {
...
-  /// Required by every result builder to build combined results from
-  /// statement blocks.
-  public static func buildBlock(_ components: Self.Component...) -> Self.Component {
-    return components.flatMap {
-      $0
-    }
-  }
-  
   /// If declared, provides contextual type information for statement
   /// expressions to translate them into partial results.
```
ahoppen, line 50: Similarly here, I don’t think the comments provide much value and I would just remove them.

ahoppen (commented):
Removing >3000 LOC. I love it 😍

Sources/SwiftSyntaxBuilder/ResultBuilderExtensions.swift
RoBo-Inc: Yeah, no problem. I will put it back 🙂

RoBo-Inc:
Updated ✅

1. `generated/ResultBuilders` file:
-    Removed the unnecessary comments as requested.
-    Added "MARK: -" separators instead, for better readability.

2. Put the extensions back, but:
- Moved them from `ResultBuilderExtensions` to `ConvenienceInitializers` file. Because these extensions are not specifically related to the result builders anymore. At the same time they indeed are convenience initializers.
- Improved its bodies a bit.
- Made the both initializers to be public for consistency (before only one of them was public).

Note:  The `ConvenienceInitializers` file has warning to insert new extensions in alphabetical order. But I just append the two extensions (which I moved) to the end of the file. Because the extensions are related and has the common comment. Besides the extensions in the file actually are not in alphabetical order anyways... 😳

Sources/SwiftSyntaxBuilder/ConvenienceInitializers.swift
```diff
@@ -380,6 +380,22 @@ extension VariableDeclSyntax {
...
+    self.init(elements.map { ExprSyntax(fromProtocol: $0) } as [ExprSyntax])
+  }
+}
+
+extension UnexpectedNodesSyntax {
+  public init(_ elements: [SyntaxProtocol]) {
+    self.init(elements.map { Syntax(fromProtocol: $0) } as [Syntax])
+  }
+}
+
```
ahoppen, line 398: Let’s keep them in alphabetical order. AFAICT `TupleExprElement` is the only extension that’s not currently in alphabetical order (I think because it was renamed from `TupleExprElement`). If you want, you can move that as well to restore alphabetical order.

ahoppen (commented):
Thank you. Let’s keep the ordering in ConvenienceInitializers.swift alphabetical, though. Or at least not break it any more than it’s already broken.

RoBo-Inc:
Done ✅
All alphabetical now 😃

ahoppen (approved):

ahoppen:
@swift-ci Please test

Sources/SwiftSyntaxBuilder/ListBuilder.swift
ahoppen, line 1: Ah, sorry that I didn’t catch this as well. You need to add `ListBuilder.swift` to [SwiftSyntaxBuilder/CMakeLists.txt](https://github.com/apple/swift-syntax/blob/main/Sources/SwiftSyntaxBuilder/CMakeLists.txt). We are building swift-syntax using CMake when it gets linked into the compiler.

RoBo-Inc: Done ✅
Maybe we should add this info into `Review and CI Testing` section (which already exists) of the `CONTRIBUTING.md` file 🙂

ahoppen: Yeah, maybe that would be a good idea 👍🏽

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

KidiIT (commented):
All checked and approved🫶❤️

KidiIT:
All checked out. Thank you and you're welcome🤓❤️

ahoppen:
Thank *you* for the PR.

KidiIT (commented):
All reviewed and approved for the job and the application for approval for the new position in September will be completed in the next few days🫶🫶

PR #2453, opened 2024-02-01
Only parse subscript and deinit as declaration references within macro role attributes

DougGregor:
The macro role attributes require us to parse `subscript` and `deinit` is
normal member references. Thread an `ExprFlavor` through to make this happen,
and do so for both `@attached` and `@freestanding`.

### DIFF
--- a/Sources/SwiftParser/Expressions.swift
+++ b/Sources/SwiftParser/Expressions.swift
@@ -79,9 +79,9 @@ extension Parser {
     /// We don't allow allow newlines here.
     case poundIfDirective
 
-    /// Parsing an attribute argument, which can contain declaration references
-    /// like `subscript` or `deinit`.
-    case attributeArgument
+    /// Parsing an attribute's arguments, which can contain declaration
+    /// references like `subscript` or `deinit`.
+    case attributeArguments
   }
 
   enum PatternContext {
@@ -1245,7 +1245,7 @@ extension Parser {
     var options: DeclNameOptions = .compoundNames
     switch flavor {
     case .basic, .poundIfDirective, .stmtCondition: break
-    case .attributeArgument: options.insert(.keywords)
+    case .attributeArguments: options.insert(.keywordsUsingSpecialNames)
     }
 
     let declName = self.parseDeclReferenceExpr(options)
@@ -2565,7 +2565,7 @@ private extension Parser.ExprFlavor {
   var callArgumentFlavor: Parser.ExprFlavor {
     switch self {
     case .basic, .poundIfDirective, .stmtCondition: return .basic
-    case .attributeArgument: return .attributeArgument
+    case .attributeArguments: return .attributeArguments
     }
   }
 }
--- a/Sources/SwiftParser/Expressions.swift
+++ b/Sources/SwiftParser/Expressions.swift
@@ -78,6 +78,10 @@ extension Parser {
     ///
     /// We don't allow allow newlines here.
     case poundIfDirective
+
+    /// Parsing an attribute argument, which can contain declaration references
+    /// like `subscript` or `deinit`.
+    case attributeArgument
   }
 
   enum PatternContext {
@@ -725,7 +729,7 @@ extension Parser {
 
       // If there is an expr-call-suffix, parse it and form a call.
       if let lparen = self.consume(if: TokenSpec(.leftParen, allowAtStartOfLine: false)) {
-        let args = self.parseArgumentListElements(pattern: pattern)
+        let args = self.parseArgumentListElements(pattern: pattern, flavor: flavor.callArgumentFlavor)
         let (unexpectedBeforeRParen, rparen) = self.expect(.rightParen)
 
         // If we can parse trailing closures, do so.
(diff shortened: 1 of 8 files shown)


DougGregor:
@swift-ci please test

Sources/SwiftParser/Expressions.swift
```diff
@@ -78,6 +78,10 @@ extension Parser {
     ///
     /// We don't allow allow newlines here.
     case poundIfDirective
+
+    /// Parsing an attribute argument, which can contain declaration references
+    /// like `subscript` or `deinit`.
+    case attributeArgument
```
ahoppen, line 84: Nit, this is parsing multiple arguments, so I would make it plural

```suggestion
    case attributeArguments
```

Sources/SwiftParser/Expressions.swift
```diff
@@ -1237,8 +1241,14 @@ extension Parser {
 
 extension Parser {
   /// Parse an identifier as an expression.
-  mutating func parseIdentifierExpression() -> RawExprSyntax {
-    let declName = self.parseDeclReferenceExpr(.compoundNames)
+  mutating func parseIdentifierExpression(flavor: ExprFlavor) -> RawExprSyntax {
+    var options: DeclNameOptions = .compoundNames
+    switch flavor {
+    case .basic, .poundIfDirective, .stmtCondition: break
+    case .attributeArgument: options.insert(.keywords)
```
ahoppen, line 1248: This does change the behavior a little bit more and allows eg. `names: named(class)`. I’m not sure if we want this or if we want to require backticks around `class`.

Sources/SwiftParser/Names.swift
```diff
@@ -66,12 +66,13 @@ extension Parser {
...
     // Consume the base name.
     let base: RawTokenSyntax
-    if let identOrSelf = self.consume(if: .identifier, .keyword(.self), .keyword(.Self)) ?? self.consume(if: .keyword(.`init`)) ?? self.consume(
-      if: .keyword(.`deinit`)
-    ) ?? self.consume(if: .keyword(.`subscript`)) {
+    if let identOrSelf = self.consume(if: .identifier, .keyword(.self), .keyword(.Self)) ?? self.consume(if: .keyword(.`init`)) {
       base = identOrSelf
     } else if flags.contains(.operators), let (_, _) = self.at(anyIn: Operator.self) {
       base = self.consumeAnyToken(remapping: .binaryOperator)
+    } else if flags.contains(.keywordsUsingSpecialNames),
```
ahoppen, line 73: Haha, the flag finally gets a use 😄

Sources/SwiftParser/Names.swift
```diff
@@ -66,12 +66,13 @@ extension Parser {
...
+              let special = self.consume(if: .keyword(.`deinit`), .keyword(.`subscript`)) {
```
ahoppen, line 74: I don’t think this is correct `swift-format`ing.

Tests/SwiftParserTest/AttributeTests.swift
```diff
@@ -720,7 +720,42 @@ final class AttributeTests: ParserTestCase {
...
+    assertParse(
+      """
+      @freestanding(member, names: named(subscript))
+      macro m()
+      """
+    )
+
+    assertParse(
+      """
+      @freestanding(member, names: named(subscript(a:b:)))
```
ahoppen, line 758: I think it would make sense to add substructure tests here and check whether `subscript` etc. get parsed as a keyword or not?

Sources/SwiftParser/Names.swift
DougGregor: Argh, I forgot to run the formatter

Sources/SwiftParser/Expressions.swift
DougGregor: Sure

Sources/SwiftParser/Expressions.swift
DougGregor: Ah, so I actually wanted `keywordsUsingSpecialNames` here.

Tests/SwiftParserTest/AttributeTests.swift
DougGregor: Alright

DougGregor:
@swift-ci please test

Sources/SwiftParser/Expressions.swift
ahoppen: That too but AFAICT `keywordsUsingSpecialNames` will still parse of `named(class)` with `class` as an identifier. Currently `class` needs to be backticked. We might want this behavior or not. I’m not sure.

DougGregor: No, `keywordsUsingSpecialNames` is only for `subscript` and `deinit`. It was made just for this carve-out

DougGregor:
@swift-ci please test Windows

Sources/SwiftParser/Expressions.swift
ahoppen: Oh, looks like I was wrong because the case in `parsePriamryExpression` doesn’t have a case for `class` and thus doesn’t enter `parseIdentifierExpression`. Sorry for the confusion.

ahoppen (approved):
Thanks!

ahoppen:
@swift-ci Please test Windows

PR #2454, opened 2024-02-02
Ask the swift compiler for SWIFT_HOST_MODULE_TRIPLE if not supplied

dabrahams:


CMakeLists.txt
```diff
@@ -77,7 +77,8 @@ if("${SWIFT_HOST_MODULE_TRIPLE}" STREQUAL "")
       message(FATAL_ERROR "Unrecognized architecture for Windows host")
     endif()
   else()
```
bnbarham, line 79: I wonder if the current `if` contents is even worth keeping. `SWIFT_HOST_MODULE_TRIPLE` would be set when this is part of the compiler build. So we'd only be in here if standalone, in which case falling back to the module triple from `-print-target-info` sounds the most reasonable to me anyway. Thoughts @etcwilde/@compnerd?

(And thanks @dabrahams, this LGTM even if we don't rip out the rest of the if).

bnbarham (approved):

CMakeLists.txt
etcwilde: Yeah, something like this should work for all cases, I believe.
```
set(module_triple_command "${CMAKE_Swift_COMPILER}" -print-target-info)
  if(CMAKE_Swift_COMPILER_TARGET)
    list(APPEND module_triple_command -target ${CMAKE_Swift_COMPILER_TARGET})
  endif()
  execute_process(COMMAND ${module_triple_command} OUTPUT_VARIABLE target_info_json)
  string(JSON SWIFT_HOST_MODULE_TRIPLE GET "${target_info_json}" "target" "moduleTriple")
```

dabrahams: I had the same thought but didn't want to presume.

dabrahams: @etcwilde You probably know CMake much better than me but your version seems like a lot of code for what I did in two lines.  Is there a reason to make it so elaborate?

etcwilde: The extra bits I added are for cross-compiling. Running `-print-target-info` alone will only give you the module triple for the triple that the compiler was configured to output to by default. If someone has specified the triple they want to output to, we want to ensure that the module triple reflects that. It's not always set, so we need to ensure that it is before adding it to the command.

bnbarham: His is aimed at removing the entire contents of `if("${SWIFT_HOST_MODULE_TRIPLE}" STREQUAL ""))`. Where as yours is only in the else, which assumes `CMAKE_Swift_COMPILER_TARGET ` isn't set because otherwise we wouldn't be in the else.

etcwilde: That is correct.

compnerd: I definitely prefer @etcwilde's approach - cross-compilation is going to be most likely the path we take to get the Windows ARM64 toolchain release setup.

dabrahams: So are y'all suggesting I replace lines 65-80 with @etcwilde's code?

dabrahams: I ask because @bnbarham approved these changes as-is, but I can't run the CI or merge myself, and I'd like to move the PR forward.  If that means making the edit I just described, please LMK.

bnbarham:
@swift-ci please test

CMakeLists.txt
bnbarham: It'd be good to do, I am fine to merge this as is though - it's an improvement regardless. If you'd like to put up another (or you @etcwilde), feel free :). I'll run tests and merge. Thanks @dabrahams!

dabrahams: I for one encourage @etcwilde to submit the improved PR once this lands.

bnbarham:

PR #2467, opened 2024-02-03
Move `FixItApplier` to `SwiftIDEUtils` so that SourceKit-LSP can use it

ahoppen:
SourceKit-LSP currently contains a copy of FixItApplier. We should be able to use the version from SwiftSyntax.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -107,7 +107,7 @@ let package = Package(
 
     .target(
       name: "SwiftIDEUtils",
-      dependencies: ["SwiftSyntax", "SwiftParser"],
+      dependencies: ["SwiftSyntax", "SwiftDiagnostics", "SwiftParser"],
       exclude: ["CMakeLists.txt"]
     ),
 
@@ -191,7 +191,7 @@ let package = Package(
 
     .target(
       name: "SwiftSyntaxMacrosTestSupport",
-      dependencies: ["_SwiftSyntaxTestSupport", "SwiftDiagnostics", "SwiftParser", "SwiftSyntaxMacros", "SwiftSyntaxMacroExpansion"]
+      dependencies: ["_SwiftSyntaxTestSupport", "SwiftDiagnostics", "SwiftIDEUtils", "SwiftParser", "SwiftSyntaxMacros", "SwiftSyntaxMacroExpansion"]
     ),
 
     .testTarget(
@@ -211,7 +211,7 @@ let package = Package(
 
     .testTarget(
       name: "SwiftParserTest",
-      dependencies: ["_SwiftSyntaxTestSupport", "SwiftDiagnostics", "SwiftOperators", "SwiftParser", "SwiftSyntaxBuilder"],
+      dependencies: ["_SwiftSyntaxTestSupport", "SwiftDiagnostics", "SwiftIDEUtils", "SwiftOperators", "SwiftParser", "SwiftSyntaxBuilder"],
       swiftSettings: swiftParserSwiftSettings
     ),
 
--- a/Sources/SwiftIDEUtils/FixItApplier.swift
+++ b/Sources/SwiftIDEUtils/FixItApplier.swift
@@ -12,8 +12,8 @@
 
 import SwiftDiagnostics
 import SwiftSyntax
-import SwiftSyntaxMacroExpansion
 
+@_spi(FixItApplier)
 public enum FixItApplier {
   /// Applies selected or all Fix-Its from the provided diagnostics to a given syntax tree.
   ///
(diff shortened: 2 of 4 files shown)


ahoppen:
@swift-ci Please test

bnbarham:
I have no great suggestions other than copying the code, but... I'd prefer *something* other than typealiasing the type in the underscored module. Otherwise it's non-obvious that a change there impacts actually public (non-underscored) API.

ahoppen:
What do you think about something like

```swift

public enum FixItApplier {
  /// Applies selected or all Fix-Its from the provided diagnostics to a given syntax tree.
  ///
  /// - Parameters:
  ///   - diagnostics: An array of `Diagnostic` objects, each containing one or more Fix-Its.
  ///   - filterByMessages: An optional array of message strings to filter which Fix-Its to apply.
  ///   - tree: The syntax tree to which the Fix-Its will be applied.
  ///
  /// - Returns: A `String` representation of the modified syntax tree after applying the Fix-Its.
  public static func applyFixes(
    from diagnostics: [Diagnostic],
    filterByMessages messages: [String]? = nil,
    to tree: any SyntaxProtocol
  ) -> String {
    return _SwiftSyntaxTestSupport.FixItApplier.applyFixes(from: diagnostics, filterByMessages: messages, to: tree)
  }

  /// Apply the given edits to the syntax tree.
  ///
  /// - Parameters:
  ///   - edits: The edits to apply to the syntax tree
  ///   - tree: he syntax tree to which the edits should be applied.
  /// - Returns: A `String` representation of the modified syntax tree after applying the edits.
  public static func apply(
    edits: [SourceEdit],
    to tree: any SyntaxProtocol
  ) -> String {
    return _SwiftSyntaxTestSupport.FixItApplier.apply(edits: edits, to: tree)
  }
}
```

bnbarham:
> What do you think about something like

Hmmm, better I think? My next question would be if `SwiftSyntaxMacrosTestSupport` is the best place for it, or if in `SwiftDiagnostics`/`SwiftIDEUtils`/`SwiftRefactor` is a better place 😅 (but that then means we can't just grab the one from _...TestSupport).

ahoppen:
Hmmm, true, we could move it `FixItApplier` to `SwiftIDEUtils`. That way I could also use it from SourceKit-LSP’s `Diagnose` module (which currently has a copy of the applier).

Release Notes/511.md
```diff
@@ -35,6 +35,10 @@
 - `String.isValidIdentifier(for:)`
   - Description: `SwiftParser` adds an extension on `String` to check if it can be used as an identifier in a given context.
   - Pull Request: https://github.com/apple/swift-syntax/pull/2434
+  
+- `FixItApplier`
+  - `FixItApplier` can apply edits or Fix-Its to a syntax tree and return the fixed-up source
+  - Pull Request: xxx
```
Matejkob, line 41: I'm leaving this suggestion here in case you overlook it before merging:
```suggestion
  - Pull Request: https://github.com/apple/swift-syntax/pull/2467
```

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/SwiftIDEUtils/FixItApplier.swift
```diff
@@ -14,6 +14,7 @@ import SwiftDiagnostics
 import SwiftSyntax
 import SwiftSyntaxMacroExpansion
 
+@_spi(FixItApplier)
```
bnbarham, line 17: I assume this is "public but we're not sure about the API yet"?

Sources/SwiftIDEUtils/FixItApplier.swift
```diff
@@ -14,6 +14,7 @@ import SwiftDiagnostics
 import SwiftSyntax
 import SwiftSyntaxMacroExpansion
```
bnbarham, line 15: Same question here.

Package.swift
```diff
@@ -107,7 +107,7 @@ let package = Package(
 
     .target(
       name: "SwiftIDEUtils",
-      dependencies: ["SwiftSyntax", "SwiftParser"],
+      dependencies: ["SwiftSyntax", "SwiftSyntaxMacroExpansion", "SwiftDiagnostics", "SwiftParser"],
```
bnbarham, line 110: Is `SwiftSyntaxMacroExpansion` left over from when it was moved there?

ahoppen: Good catch. Removed the dependency.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #2468, opened 2024-02-04
Fix "escaped newline symbol" bug in `CodeGeneration` module

RoBo-Inc:
I stumbled upon a small bug by accident, and upon investigation, it appeared to have deep roots. Eventually, it warranted its own PR. The bug involved duplicating empty lines in the generated `SyntaxTraits` file after every `// MARK: -` without documentation text.

The file is generated based on `SyntaxTraitsFile` file of `CodeGeneration` module:
```
      // MARK: - \(trait.protocolName)

      \(trait.documentation)
      public protocol \(trait.protocolName): SyntaxProtocol\(raw:
```

In other parts of the module, similar situations are addressed using the "escaped newline" symbol: "\". For instance, in the `ResultBuilderFile` file:
```
      // MARK: - \(type.resultBuilderType)

      @resultBuilder
      \(node.node.apiAttributes())\
      public struct \(type.resultBuilderType): ListBuilder
```

So in this PR, I resolve the bug by employing the same approach. This not only enhances the consistency of the codebase but also serves as an elegant solution to the problem itself. 
All the changes are confined to the `CodeGeneration` module and have no impact on the rest of the `SwiftSyntax` library, making them safe to merge.

### DIFF
--- a/CodeGeneration/Sources/SyntaxSupport/Utils.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/Utils.swift
@@ -42,35 +42,19 @@ public func lowercaseFirstWord(name: String) -> String {
 extension SwiftSyntax.Trivia {
   /// Make a new trivia from a (possibly multi-line) string, prepending `///`
   /// to each line and creating a `.docLineComment` trivia piece for each line.
-  public static func docCommentTrivia(from string: String?) -> SwiftSyntax.Trivia {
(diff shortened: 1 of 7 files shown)


ahoppen (approved):
Nice. Much simpler

ahoppen:
@swift-ci Please test

PR #2472, opened 2024-02-05
Add representation and parser support for the @isolated type attribute

rjmccall:


CodeGeneration/Sources/SyntaxSupport/AttributeNodes.swift
```diff
@@ -141,6 +141,10 @@ public let ATTRIBUTE_NODES: [Node] = [
             name: "documentationArguments",
             kind: .node(kind: .documentationAttributeArgumentList)
           ),
+          Child(
+            name: "isolatedArguments",
+            kind: .node(kind: .isolatedAttributeArguments)
+          ),
```
ahoppen, line 147: Instead of adding a new child choice you should be able to reuse the `token` child choice.

Sources/SwiftParser/Attributes.swift
```diff
@@ -848,6 +848,17 @@ extension Parser {
   }
 }
 
+extension Parser {
+  mutating func parseIsolatedAttributeArguments() -> RawIsolatedAttributeArgumentsSyntax {
+    let (unexpectedBeforeIsolationKind, label) = self.expect(.identifier)
+    return RawIsolatedAttributeArgumentsSyntax(
+      unexpectedBeforeIsolationKind,
+      isolationKind: label,
+      /*unexpectedAfterIsolationKind*/ nil,
```
ahoppen, line 857: This is defaulted and you should be able to omit it.

Sources/SwiftParser/Attributes.swift
```diff
@@ -848,6 +848,17 @@ extension Parser {
...
+      arena: self.arena)
```
ahoppen, line 858: I think this is not correctly `swift-format`ed? Could you run `./swift-syntax-dev-utils format`?

Sources/SwiftParser/Lookahead.swift
```diff
@@ -171,6 +171,15 @@ extension Parser.Lookahead {
...
       }
+      if case .isolated = attr {
+        guard
+          self.consume(if: .leftParen) != nil,
+          self.consume(if: .identifier) != nil,
+          self.consume(if: .rightParen) != nil
+        else {
+          return
+        }
+      }
```
ahoppen, line 182: You should be able to merge this with `if case .convention = attr` by changing it to `switch attr { case .convention, .isolated`

Tests/SwiftParserTest/AttributeTests.swift
```diff
@@ -886,4 +886,19 @@ final class AttributeTests: ParserTestCase {
...
+      var fn: @isolated(any) () -> ()
+      """
+    )
+
+    // We don't validate the kind in the parser
+    assertParse(
+      """
+      var fn: @isolated(sdfhsdfi) () -> ()
+      """
+    )
```
ahoppen, line 902: Can you combine `@isolated` with other attributes? Might be worth adding a test case that has two attributes eg. `@isolated` and `@convention`? Thinking of this because you added the `@isoalted` check after `@convention` in `skipTypeAttribute`.

CodeGeneration/Sources/SyntaxSupport/AttributeNodes.swift
rjmccall: If you don't mind, I would rather leave room for more structure here.

Sources/SwiftParser/Lookahead.swift
rjmccall: I could, but the `@convention` handling is wrong and will need to be split out when it's fixed.

Tests/SwiftParserTest/AttributeTests.swift
rjmccall: Added a test specifically testing the combination with `@convention`.

rjmccall:
@swift-ci Please test

CodeGeneration/Sources/SyntaxSupport/AttributeNodes.swift
ahoppen: What would that structure look like. We generally prefer to parse attributes like a normal function call now instead of adding dedicated syntax for every possible built-in attribute. So, unless you have concrete ideas for a syntax that can’t be represented by a function call, I’d prefer to stay in that realm.

rjmccall: I'm not imagining anything grammatically that wouldn't look somewhat like a function call, but a single token is not exactly the function call grammar.  I just don't want to have to refactor all the code consuming this if we add new forms of the attribute that want more arguments.  We do expect this to grow over time: we're already talking about two different forms (the second being `@isolated(caller)`), so I'm pretty comfortable adding an experimental node in the short term.

I believe this first form (`@isolated(any)`) already technically cannot be parsed as a function call, since `any` is a keyword.  Any variation is likely to be "call-inspired", but not strictly limited to the call grammar.

ahoppen: I would prefer to use the call grammar and change `parsePrimaryExpression` to allow keywords when the expression parsing flavor is `attributeArguments`. Something like the following (this currently essentially copy-pastes the identifier parsing code for keywords.

```swift
case nil:
  if flavor == .attributeArguments {
    if pattern.admitsBinding && self.lookahead().isInBindingPatternPosition() {
      let identifier = self.consumeAnyToken(remapping: .identifier)
      let pattern = RawPatternSyntax(
        RawIdentifierPatternSyntax(
          identifier: identifier,
          arena: self.arena
        )
      )
    }

    return RawExprSyntax(self.parseIdentifierExpression(flavor: flavor))
  }
  return RawExprSyntax(RawMissingExprSyntax(arena: self.arena))
}
```

@rintaro Do you have thoughts. I would like to hear your opinion on this.

Sources/SwiftParser/Lookahead.swift
ahoppen: Oh, good catch. Filed https://github.com/apple/swift-syntax/issues/2474 and fixing it in https://github.com/apple/swift-syntax/pull/2475.

You should be able to tag along with that PR and just call `skipSingle` for `@isolated` parsing. That should also cover any future additions to the `@isolated` grammar.

rjmccall:
@swift-ci Please test

rjmccall:
@swift-ci Please test

rjmccall:
@swift-ci Please test

Sources/SwiftParser/Lookahead.swift
rjmccall: Rebased onto your commit.

rjmccall:

CodeGeneration/Sources/SyntaxSupport/AttributeNodes.swift
rintaro: We're certainly leaning toward to reducing special `AttributeSyntax.Arguments` choices, representing them with generic `LabeledExprListSyntax` syntax, and - as Alex mentioned - making an expression parsing "flavor" for attribute arguments.

>  I'm going to work with it this way in the short term, though.

I think I'm fine with that, but could you make it `@_spi(ExperimentalLanguageFeatures)`? Maybe the current `CodeGeneration` doesn't recognize `Child.experimentalFeature` for `.nodeChoices([Child])` nodes, but implementing it shouldn't be that hard.

rintaro: Probably, prepending `\(choice.apiAttributes)` to here
https://github.com/apple/swift-syntax/blob/e77b169ba11273be5c2f4ea6a22150e5d082ffb6/CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxNodesFile.swift#L220

rjmccall: I've changed this to use `LabeledExprListSyntax`.

rjmccall:
@swift-ci Please test

ahoppen (approved):
Thank you! Much simpler as well 👍🏽

rjmccall:
@swift-ci Please test Windows

rjmccall:
@swift-ci Please test

rjmccall:
@swift-ci Please test Windows

PR #2475, opened 2024-02-06
Correctly skip arguments to `@convention` attribute

### LINKED ISSUE #2474: New parser fails to parse `@convention` attribute with `cType` argument inside an array type
### Issue Kind

Parse of Valid Source Failed

### Source Code

```swift
_ = [@convention(c, cType: "int (*)(int)") (Int32) -> Int32]()
```

### Description

The issue is that the implementation of `skipTypeAttribute` doesn’t skip over the `cType` argument here. We should probably be able to just `skipSingle` to skip over all the arguments.

https://github.com/apple/swift-syntax/blob/89dd2bde9404950dede113b7898c97f6e2bc2fc4/Sources/SwiftParser/Lookahead.swift#L165-L173

ahoppen:
Fixes #2474
rdar://122358055


ahoppen:
@swift-ci Please test

PR #2479, opened 2024-02-06
Fix sendable warnings

Matejkob:
Because the `MacroExpansionError` implements the `DiagnosticMessage` protocol, which requires the type to be sendable, we are getting sendable warnings for cases with associated values in the `MacroExpansionError` enum.


ahoppen (approved):

ahoppen:
@swift-ci Please test

PR #2480, opened 2024-02-06
[bazel] Update bazel version and BUILD for new modules

keith:


keith:
@swift-ci please test

PR #2484, opened 2024-02-07
[bazel] Use globs for _AtomicBool sources

keith:
These were renamed


keith:
@swift-ci please test

PR #2486, opened 2024-02-07
Don't declare attribute names as keywords

rintaro:
They were declared as `Keyword` only for `TokenSpecs` (DeclarationAttributeWithSpecialSyntax and TypeAttribute), the attribute names are always parsed as `TypeSyntax` so those keywords weren't even appear in the parsed syntax tree.

Instead, make those `TokenSpecSet` to accept `.identifier` lexemes.

### DIFF
--- a/Sources/SwiftParser/Attributes.swift
+++ b/Sources/SwiftParser/Attributes.swift
@@ -67,41 +67,52 @@ extension Parser {
     case transpose
 
     init?(lexeme: Lexer.Lexeme, experimentalFeatures: Parser.ExperimentalFeatures) {
-      switch PrepareForKeywordMatch(lexeme) {
-      case TokenSpec(._alignment): self = ._alignment
-      case TokenSpec(._backDeploy): self = ._backDeploy
-      case TokenSpec(._cdecl): self = ._cdecl
-      case TokenSpec(._documentation): self = ._documentation
-      case TokenSpec(._dynamicReplacement): self = ._dynamicReplacement
-      case TokenSpec(._effects): self = ._effects
-      case TokenSpec(._expose): self = ._expose
-      case TokenSpec(._implements): self = ._implements
-      case TokenSpec(._nonSendable): self = ._nonSendable
-      case TokenSpec(._objcImplementation): self = ._objcImplementation
-      case TokenSpec(._objcRuntimeName): self = ._objcRuntimeName
-      case TokenSpec(._optimize): self = ._optimize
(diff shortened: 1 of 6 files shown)


rintaro:
@swift-ci Please test

Sources/SwiftParser/Attributes.swift
```diff
@@ -67,82 +67,61 @@ extension Parser {
...
+      switch lexeme.rawTokenKind {
+      case .keyword:
+        switch Keyword(lexeme.tokenText) {
+        case .rethrows: self = .rethrows
+        default:
+          return nil
+        }
+      case .identifier:
+        switch lexeme.tokenText {
+        case "_alignment": self = ._alignment
```
rintaro, line 79: I'm leaning toward making `TokenSpec` for fixed text. I.e. `TokenSpec("_alignment")`.

rintaro:
The last commit implements `TokenSpec` for a fixed text.

rintaro:
@swift-ci Please test

Sources/SwiftParser/TokenSpec.swift
```diff
@@ -46,18 +49,24 @@ struct PrepareForKeywordMatch {
...
+    /// The keyword we are expecting.
+    case keyword(Keyword)
+
+    /// The kind we expect the token that we want to consume to have.
+    /// This can be a keyword, in which case the ``TokenSpec`` will also match an
+    /// identifier with the same text as the keyword and remap it to that keyword
+    /// when consumed.
+    ///
+    /// `fileprivate` because only functions in this file should access it since
+    /// they know how to handle the identifier -> keyword remapping.
```
ahoppen, line 65: This comment doesn’t really make sense on an enum case.

Sources/SwiftParser/TokenSpec.swift
```diff
@@ -169,29 +191,50 @@ public struct TokenSpec {
...
+      default: return TokenKind.fromRaw(kind: kind, text: "")
+      }
     }
   }
+
+  static func keyword(_ keyword: Keyword) -> TokenSpec {
+    return TokenSpec(keyword)
+  }
+
+  static func fixedText(_ text: SyntaxText) -> TokenSpec {
```
ahoppen, line 221: I think it would be good to add a short comment that explains when somebody should use `fixedText` vs adding a keyword. The chance of misuse seems quite high to me.

ahoppen (commented):
Nice. Could you / did you check that this doesn’t have a performance impact?

PR #2487, opened 2024-02-08
Parse `_borrowing x` contextually as a pattern binding when `.borrowingSwitch` feature is enabled.

jckarter:
The existing parsing under `.referenceBindings` parses this (along with `_mutating` and `_consuming`) as unconditional keywords, which would be a source break we can't ship. Narrow the behavior so that `_borrowing` is only parsed immediately before an identifier in a pattern.

### DIFF
--- a/CodeGeneration/Sources/SyntaxSupport/KeywordSpec.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/KeywordSpec.swift
@@ -19,6 +19,7 @@ public struct KeywordSpec {
   /// The experimental feature the keyword is part of, or `nil` if this isn't
   /// for an experimental feature.
   public let experimentalFeature: ExperimentalFeature?
+  public let experimentalFeature2: ExperimentalFeature?
 
   /// Indicates if the token kind is switched from being an identifier to a keyword in the lexer.
   public let isLexerClassified: Bool
@@ -59,6 +60,26 @@ public struct KeywordSpec {
   ) {
     self.name = name
     self.experimentalFeature = experimentalFeature
+    self.experimentalFeature2 = nil
+    self.isLexerClassified = isLexerClassified
+  }
+
+  /// Initializes a new `KeywordSpec` instance.
+  ///
+  /// - Parameters:
+  ///   - name: A name of the keyword.
+  ///   - experimentalFeature: The experimental feature the keyword is part of, or `nil` if this isn't for an experimental feature.
+  ///   - or: A second experimental feature the keyword is also part of, or `nil` if this isn't for an experimental feature.
+  ///   - isLexerClassified: Indicates if the token kind is switched from being an identifier to a keyword in the lexer.
+  init(
+    _ name: String,
+    experimentalFeature: ExperimentalFeature,
+    or experimentalFeature2: ExperimentalFeature,
+    isLexerClassified: Bool = false
+  ) {
+    self.name = name
+    self.experimentalFeature = experimentalFeature
+    self.experimentalFeature2 = experimentalFeature2
     self.isLexerClassified = isLexerClassified
   }
 }
@@ -720,7 +741,7 @@ public enum Keyword: CaseIterable {
     case .yield:
       return KeywordSpec("yield")
     case ._borrowing:
-      return KeywordSpec("_borrowing", experimentalFeature: .referenceBindings)
+      return KeywordSpec("_borrowing", experimentalFeature: .referenceBindings, or: .borrowingSwitch)
     case ._consuming:
       return KeywordSpec("_consuming", experimentalFeature: .referenceBindings)
     case ._mutating:
--- a/Sources/SwiftParser/Expressions.swift
+++ b/Sources/SwiftParser/Expressions.swift
@@ -118,9 +118,21 @@ extension Parser {
     //
     // Only do this if we're parsing a pattern, to improve QoI on malformed
     // expressions followed by (e.g.) let/var decls.
-    if pattern != .none, self.at(anyIn: MatchingPatternStart.self) != nil {
-      let pattern = self.parseMatchingPattern(context: .matching)
-      return RawExprSyntax(RawPatternExprSyntax(pattern: pattern, arena: self.arena))
+    if pattern != .none {
+      switch self.at(anyIn: MatchingPatternStart.self) {
+      case (spec: .rhs(let bindingIntroducer), handle: _)? where self.withLookahead { $0.shouldParsePatternBinding(introducer: bindingIntroducer) }:
+        fallthrough
+      case (spec: .lhs(_), handle: _)?:
+        let pattern = self.parseMatchingPattern(context: .matching)
+        return RawExprSyntax(RawPatternExprSyntax(pattern: pattern, arena: self.arena))
+
+      // If the token can't start a pattern, or contextually isn't parsed as a
+      // binding introducer, handle as the start of a normal expression.
+      case (spec: .rhs(_), handle: _)?,
+        nil:
+        // Not a pattern. Break out to parse as a normal expression below.
+        break
+      }
     }
     return RawExprSyntax(self.parseSequenceExpression(flavor: flavor, pattern: pattern))
   }
--- a/Sources/SwiftParser/Patterns.swift
+++ b/Sources/SwiftParser/Patterns.swift
@@ -67,7 +67,19 @@ extension Parser {
           arena: self.arena
         )
       )
-    case (.lhs(.identifier), let handle)?:
+    case (.rhs(let introducer), let handle)? where self.withLookahead { $0.shouldParsePatternBinding(introducer: introducer) }:
+      let bindingSpecifier = self.eat(handle)
+      let value = self.parsePattern()
+      return RawPatternSyntax(
+        RawValueBindingPatternSyntax(
+          bindingSpecifier: bindingSpecifier,
+          pattern: value,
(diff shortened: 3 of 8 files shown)


jckarter:
@swift-ci Please test

jckarter:
@swift-ci Please test

CodeGeneration/Sources/SyntaxSupport/KeywordSpec.swift
```diff
@@ -19,6 +19,7 @@ public struct KeywordSpec {
   /// The experimental feature the keyword is part of, or `nil` if this isn't
   /// for an experimental feature.
   public let experimentalFeature: ExperimentalFeature?
+  public let experimentalFeature2: ExperimentalFeature?
```
ahoppen, line 22: I think it would be nicer to make this an array of experimental features + update the doc comment to say that the keyword is enabled if any of the experimental features are enabled.

CodeGeneration/Sources/SyntaxSupport/KeywordSpec.swift
```diff
@@ -59,6 +60,26 @@ public struct KeywordSpec {
...
+  ///
+  /// - Parameters:
+  ///   - name: A name of the keyword.
+  ///   - experimentalFeature: The experimental feature the keyword is part of, or `nil` if this isn't for an experimental feature.
+  ///   - or: A second experimental feature the keyword is also part of, or `nil` if this isn't for an experimental feature.
+  ///   - isLexerClassified: Indicates if the token kind is switched from being an identifier to a keyword in the lexer.
+  init(
+    _ name: String,
+    experimentalFeature: ExperimentalFeature,
+    or experimentalFeature2: ExperimentalFeature,
```
ahoppen, line 77: And then make the initializer take a variadic parameter: `experimentalFeature: ExperimentalFeature…`,

jckarter:
@swift-ci Please test

Tests/SwiftParserTest/translated/MatchingPatternsTests.swift
```diff
@@ -606,4 +606,102 @@ final class MatchingPatternsTests: ParserTestCase {
...
+
+    assertParse(
+      """
+      switch bar {
+      case _borrowing x: // parses as binding
+        break
+      }
+      """,
+      experimentalFeatures: .borrowingSwitch
+    )
```
ahoppen, line 639: You can check that it actually parses as a binding (and in the previous test case that it parses as a pattern) by adding a `substructure` parameter. The easiest way to generate them is to run a single `assertParse` function, set a breakpoint in `assertParse`, run `po tree.debugInitCall(includeTrivia: false)`, trim it down and then add it as the `substructure` parameter to `assertParse`.

Sources/SwiftParser/Patterns.swift
```diff
@@ -294,9 +325,14 @@ extension Parser.Lookahead {
...
-    case (.rhs, let handle)?:
-      self.eat(handle)
-      return self.canParsePattern()
+    case (.rhs(let introducer), let handle)?:
+      if shouldParsePatternBinding(introducer: introducer) {
+        self.eat(handle)
+        return self.canParsePattern()
+      } else {
+        self.eat(handle)
+        return true
```
ahoppen, line 334: I was confused about this for a little bit until I read your comment in `parsePattern`. I think adding a comment like the following would make it clearer.

```swift
// The introducer doesn’t start a pattern binding. We should parse it as if it was an identifier.
```

Sources/SwiftParser/Patterns.swift
```diff
@@ -218,16 +231,19 @@ extension Parser {
...
-      let bindingSpecifier = self.eat(handle)
-      let value = self.parseMatchingPattern(context: .bindingIntroducer)
-      return RawPatternSyntax(
-        RawValueBindingPatternSyntax(
-          bindingSpecifier: bindingSpecifier,
-          pattern: value,
-          arena: self.arena
+    case (.rhs(let introducer), let handle)?:
+      var backtrack = lookahead()
+      if backtrack.shouldParsePatternBinding(introducer: introducer) {
```
ahoppen, line 236: We call it `lookahead` not `backtrack` in the new parser because the parser doesn’t actually backtrack. It clones itself on `lookahead()` and the clone looks ahead while the parser itself stays at the same position.

Sources/SwiftParser/Patterns.swift
```diff
@@ -85,16 +85,29 @@ extension Parser {
...
-      let bindingSpecifier = self.eat(handle)
-      let value = self.parsePattern()
-      return RawPatternSyntax(
-        RawValueBindingPatternSyntax(
-          bindingSpecifier: bindingSpecifier,
-          pattern: value,
-          arena: self.arena
+    case (.rhs(let introducer), let handle)?:
+      var backtrack = self.lookahead()
+      if backtrack.shouldParsePatternBinding(introducer: introducer) {
```
ahoppen, line 90: Same here, should be named `lookahead`, not `backtrack`.

Sources/SwiftParser/Expressions.swift
```diff
@@ -118,9 +118,19 @@ extension Parser {
...
-    if pattern != .none, self.at(anyIn: MatchingPatternStart.self) != nil {
-      let pattern = self.parseMatchingPattern(context: .matching)
-      return RawExprSyntax(RawPatternExprSyntax(pattern: pattern, arena: self.arena))
+    if pattern != .none,
+      let matchStart = self.at(anyIn: MatchingPatternStart.self)
+    {
+      switch matchStart.spec {
+      case .rhs(let bindingIntroducer):
+        var backtrack = lookahead()
+        if backtrack.shouldParsePatternBinding(introducer: bindingIntroducer) {
```
ahoppen, line 127: Same here `backtrack` -> `lookahead`.

Sources/SwiftParser/Expressions.swift
```diff
@@ -118,9 +118,19 @@ extension Parser {
...
+          fallthrough
+        }
+      default:
```
ahoppen, line 130: Just a slight suggestion to make it easier to read IMO:

```swift
case .rhs(let bindingIntroducer) where self.withLookahead { $0.shouldParsePatternBinding(bindingIntroducer) }:
  fallthrough
case .lhs
  let pattern = self.parseMatchingPattern(context: .matching)
default:
  break
```

CodeGeneration/Sources/SyntaxSupport/KeywordSpec.swift
jckarter: I know this stuff is performance-sensitive so I wasn't sure about introducing a refcounted field into the structure just for a one-off case where we happen to need two features for the same keyword. I'm also hoping I can just take this out again once the `.borrowingSwitch` feature is enabled, which will hopefully occur in the nearish term.

Sources/SwiftParser/Patterns.swift
jckarter: I think all of these uses can be replaced by `withLookahead`, which I didn't know about. I used `backtrack` because it is used in several places to name a lookahead variable elsewhere in the codebase; are those other places needing updating?

```
% git grep 'backtrack =' 
Sources/SwiftParser/Expressions.swift:      var backtrack = self.lookahead()
Sources/SwiftParser/Expressions.swift:      var backtrack = self.lookahead()
Sources/SwiftParser/Expressions.swift:      let backtrack = self.lookahead()
Sources/SwiftParser/Expressions.swift:        var backtrack = lookahead()
Sources/SwiftParser/Expressions.swift:      var backtrack = self.lookahead()
Sources/SwiftParser/Expressions.swift:    var backtrack = self.lookahead()
Sources/SwiftParser/Expressions.swift:    var backtrack = self.lookahead()
Sources/SwiftParser/Lookahead.swift:        var backtrack = self.lookahead()
Sources/SwiftParser/Names.swift:    var backtrack = self.lookahead()
Sources/SwiftParser/Names.swift:      var backtrack = self.lookahead()
Sources/SwiftParser/Patterns.swift:      var backtrack = self.lookahead()
Sources/SwiftParser/Patterns.swift:      var backtrack = lookahead()
Sources/SwiftParser/Types.swift:        var backtrack = self.lookahead()
Sources/SwiftParser/Types.swift:        var backtrack = self.lookahead()
```

jckarter:
@swift-ci Please test

jckarter:
@ahoppen Thanks for the tip about `withLookahead`; using that I was able to integrate the contextual `shouldParsePatternBinding` check directly into the existing switches/if lets and simplify the control flow quite a bit. I also added `substructure` checks to the new tests. How's this look now?

jckarter:
@swift-ci Please test

jckarter:
@swift-ci Please test

jckarter:
@swift-ci Please test

jckarter:
@swift-ci Please test Windows

CodeGeneration/Sources/SyntaxSupport/KeywordSpec.swift
ahoppen: This is just part of the code generation, so it’s not performance critical at all.

Sources/SwiftParser/Patterns.swift
ahoppen: Huh, yeah, they should probably all be updated. Probably still a leftover from the migration from C++. Filed https://github.com/apple/swift-syntax/issues/2494

PR #2492, opened 2024-02-10
Fix: Incorrect insertion location in fix-it for `)`

### LINKED ISSUE #2431: Wrong insertion location in fix-it for missing parentheses around parameter type list
### Description

It seems we’re forgetting about trailing trivia when computing the insertion location for the closing paren.

### Steps to Reproduce

```swift
let _ : Float -> Int
```
```
error: expected '(' to start function type [bridged_error]
let _ : Float -> Int
        ^
note: insert '(' [bridged_note]
let _ : Float -> Int
        ^
        (
error: expected ')' in function type [bridged_error]
let _ : Float -> Int
              ^
note: insert ')' [bridged_note]
let _ : Float -> Int
             ~^
              )
```

vinu-vanjari:
Resolves #2431 

---

##### 1. Added `shouldBeInsertedBeforePreviousTokenTrivia`
Extracted this from existing code from `DiagnosticExtensions.swift`
Inserted this just after existing property `shouldBeInsertedAfterNextTokenTrivia`
This extraction is needed as we want to reuse this while deciding correct insertion location of fix-it

##### 2. Correct insertion location for fix-it in `MissingNodesError.swift`
This should correct `)`, `:`, `,` syntax positions which should be inserted before previous token trivia. i.e. should not have whitespace before them. (check unit tests corrections for examples) 

##### 3. Correction of Unit Tests
Corrected the existing unit tests where `missing ')' token` error was expected.
Notice that we only change location markers and not other things in the unit tests.
None of the `fixedSource` in the examples in unit tests are changed

##### 4. Added couple of Unit Tests 
- ParserDiagnosticsFormatterIntegrationTests.swift->`testRighParenLocation()`
- TypeExprTests.swift->`testTypeExpr31()`

### DIFF
--- a/Sources/SwiftParserDiagnostics/DiagnosticExtensions.swift
+++ b/Sources/SwiftParserDiagnostics/DiagnosticExtensions.swift
@@ -198,12 +198,8 @@ extension FixIt.MultiNodeChange {
       }
     }
 
-    if let previousToken = node.previousToken(viewMode: .fixedUp),
-      previousToken.isPresent,
-      let firstToken = node.firstToken(viewMode: .all),
-      previousToken.trailingTrivia.allSatisfy({ $0.isWhitespace }),
-      !BasicFormat().requiresWhitespace(between: previousToken, and: firstToken),
-      !BasicFormat().requiresNewline(between: previousToken, and: firstToken)
+    if node.shouldBeInsertedBeforePreviousTokenTrivia,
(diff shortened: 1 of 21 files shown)


ahoppen (approved):
Very nice and simple fix 😍

ahoppen:
@swift-ci Please test

PR #2496, opened 2024-02-13
Never bulid the `Examples` package with `-warnings-as-errors`

ahoppen:
Some macros in the Examples package are expected to produce errors, so we never want to build it with `-warnings-as-errors`


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #2498, opened 2024-02-14
Provide a SourceLocationConverter when adding a source file to a diagnostics group

DougGregor:
The creation of SourceLocationConverters is expensive, so allow (but do not require) clients to provide one when adding a source file to a diagnostics group.


DougGregor:
@swift-ci please test

PR #2501, opened 2024-02-19
Use presumed file/line number when printing grouped diagnostics

DougGregor:
For the primary diagnostic, print the presumed file/line number in the same manner as the Swift compiler does. In the actual code listing, keep the line numbers and code from the actual source file that was parsed, because there's no sensible way to map the source from the `#sourceLocation`-named file (which might not exist).

### DIFF
--- a/Tests/SwiftDiagnosticsTest/GroupDiagnosticsFormatterTests.swift
+++ b/Tests/SwiftDiagnosticsTest/GroupDiagnosticsFormatterTests.swift
@@ -60,6 +60,31 @@ extension GroupedDiagnostics {
 }
 
 final class GroupedDiagnosticsFormatterTests: XCTestCase {
+  func testSourceLocations() {
+    var group = GroupedDiagnostics()
+
+    // Main source file.
+    _ = group.addTestFile(
+      """
+      #sourceLocation(file: "other.swift", line: 123)
+      let pi = 3.14159 x
+      """,
+      displayName: "main.swift",
+      diagnosticDescriptors: []
+    )
+    let annotated = DiagnosticsFormatter.annotateSources(in: group)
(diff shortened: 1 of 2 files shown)


DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test Linux

DougGregor:
@swift-ci please test macOS

DougGregor:
@swift-ci please clean test macOS

DougGregor:
@swift-ci please clean test Linux

DougGregor:
https://github.com/apple/swift-package-manager/pull/7349

@swift-ci please test

Tests/SwiftDiagnosticsTest/GroupDiagnosticsFormatterTests.swift
```diff
@@ -60,6 +60,31 @@ extension GroupedDiagnostics {
 }
 
 final class GroupedDiagnosticsFormatterTests: XCTestCase {
+  func testSourceLocations() {
```
bnbarham, line 63: It'd be nice to put the actual file's name somewhere as well, but I have no great suggestions as to where to put it.

bnbarham: With `#sourceLocation` we have two "files" - the real file on disk (usually generated in this case) and the "presumed" file (the one noted in `#sourceLocation`). After this change, the file noted in the diagnostic is the presumed file, which matches the Swift compiler. That makes sense since `#sourceLocation` was added for a reason, but I have found that knowing the underlying file is useful too.

That would have been more clear if I put this comment on the actual `#sourceLocation` line :)

PR #2502, opened 2024-02-19
Change release notes and version indicator module from 511 to 600

ahoppen:
The next Swift version is called Swift 6.0. We should adjust swift-syntax to call the next version 600 instead of 511.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -120,7 +120,7 @@ let package = Package(
 
     .target(
       name: "SwiftSyntax",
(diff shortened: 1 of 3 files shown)


ahoppen:
@swift-ci Please test

Matejkob:
I believe we also need to change the directory name for the `SwiftSyntax600` target.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #2505, opened 2024-02-21
[Macros] Add a default implementation of `MacroExpansionContext.lexicalContext`.

hborla:
Existing macro adopters have conformances to `MacroExpansionContext`, so add a default implementation of `lexicalContext` to preserve source compatibility. We can consider deprecating the default implementation to prompt conformances to implement the requirement.


hborla:
@swift-ci please test

PR #2508, opened 2024-02-22
Add a few integration tests

Matejkob:
Is it a good idea to add these unit tests?


Matejkob:
@ahoppen What do you think about adding these unit tests?

ahoppen:
I don’t think adding these is beneficial. I actually think it’s more harmful because it makes it more or less impossible to change the way the diagnostic formatter prints diagnostics if you don’t want to spend a day adjusting test cases.

If anything, I think we could get benefit out of these tests by auditing them all and checking if the printed diagnostic is reasonable. For any case where it is not (I’m not sure if there aren’t any, I haven’t looked at them), and filing issues for those. We can then add these test cases because they cover behavior that wasn’t covered anywhere else.

Matejkob:
Thanks for the feedback. I totally get where you're coming from on the potential downsides of locking in the diagnostic formatter's output too tightly. But, I've got a bit of a solution that might just do the trick. I whipped up a quick proof of concept that could help us. Here is a commit with quick implementation: https://github.com/apple/swift-syntax/pull/2508/commits/9df6534603c98ff727a53acd68f1a06f6958a202 

Would love to hear your thoughts on this approach!

ahoppen:
Thanks for the suggestion but I’m still not a huge fan. My biggest concern is that if you change something in the diagnostics formatter, many of these test cases will get updated. In practice, it’s unlikely that a reviewer will look through, say, 1000 nearly identical test changes with full attention, and it runs the risk of creating so much noise that any changes hand-crafted test cases get drowned, reducing effective test coverage.

ahoppen:
Closing this because we are not planning to merge this PR as discussed in my previous comments.

PR #2509, opened 2024-02-22
[CMake] Add cross-compilation flags for when built as part of the compiler

finagolfin:
The compiler build defines `BOOTSTRAPPING_MODE` and these other flags, so this will do nothing when they aren't set. Android requires using its patched clang for linking, so use that.

I made sure this won't break a standalone CMake build by running
```
cmake -GNinja . -DSWIFT_HOST_MODULE_TRIPLE=aarch64-unknown-linux-android
ninja
```
and everything built fine with this pull natively on Android too. This implements 2. of apple/swift#71507, and I just tested it by [cross-compiling a recent trunk compiler for Android](https://github.com/finagolfin/termux-packages/actions/runs/8007466151) and running the resulting Swift 5.11 toolchain on my Android phone.

### DIFF
--- a/cmake/modules/AddSwiftHostLibrary.cmake
+++ b/cmake/modules/AddSwiftHostLibrary.cmake
@@ -97,6 +97,16 @@ function(add_swift_syntax_library name)
     $<$<COMPILE_LANGUAGE:Swift>:-color-diagnostics>
   )
 
+  if(BOOTSTRAPPING_MODE STREQUAL "CROSSCOMPILE")
+    target_compile_options(${name} PRIVATE
+      $<$<COMPILE_LANGUAGE:Swift>:-sdk;${SWIFT_SDK_${SWIFT_HOST_VARIANT_SDK}_ARCH_${SWIFT_HOST_VARIANT_ARCH}_PATH};>
+      $<$<COMPILE_LANGUAGE:Swift>:-resource-dir;${SWIFTLIB_DIR};>)
+    if(SWIFT_HOST_VARIANT_SDK STREQUAL "ANDROID" AND NOT "${SWIFT_ANDROID_NDK_PATH}" STREQUAL "")
+      swift_android_tools_path(${SWIFT_HOST_VARIANT_ARCH} tools_path)
+      target_compile_options(${name} PRIVATE $<$<COMPILE_LANGUAGE:Swift>:-tools-directory;${tools_path};>)
+    endif()
+  endif()
+
   if(LLVM_USE_LINKER)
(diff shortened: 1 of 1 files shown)


cmake/modules/AddSwiftHostLibrary.cmake
```diff
@@ -97,6 +97,16 @@ function(add_swift_syntax_library name)
     $<$<COMPILE_LANGUAGE:Swift>:-color-diagnostics>
   )
 
+  if(BOOTSTRAPPING_MODE STREQUAL "CROSSCOMPILE")
```
bnbarham, line 100: As I've mentioned in the other PRs - the goal is to require a builder toolchain, fully remove `BOOTSTRAPPING_MODE`, and to properly run multiple CMake invocations. We are moving towards that goal, but adding more references bootstrapping everywhere is opposite to the direction we want to go.

bnbarham (changes requested):

cmake/modules/AddSwiftHostLibrary.cmake
finagolfin: You're going to have to clarify what you mean here, as I did on that other pull of yours.

> fully remove BOOTSTRAPPING_MODE

You said earlier that you want to remove [two of the options for `BOOTSTRAPPING_MODE`, `BOOTSTRAPPING` and  `BOOTSTRAPPING-WITH-HOSTLIBS`](https://github.com/apple/swift/blob/1ce3935eb77ab564fb700c09aed4ca45d60c854b/CMakeLists.txt#L346). When I responded that that was fine but that we'd still need to differentiate between the other three options listed there, [you didn't respond](https://github.com/apple/swift/pull/70343#discussion_r1483920819).

> properly run multiple CMake invocations

No idea what this refers to, as building a cross-compiled toolchain like this already involves dozens of CMake invocations, a couple times for each Swift repo.

> adding more references bootstrapping everywhere is opposite to the direction we want to go.

This is a single reference, because this is the build variable that currently exists to do it. Whenever you replace this build variable in the future, we can easily update this to use your new scheme.

finagolfin:
Alright, I asked Doug directly what the plan is on the forum and [he told me that once the bootstrap modes are dropped, he wants to switch to using CMake's new built-in Swift support](https://forums.swift.org/t/implementing-parts-of-the-swift-compiler-in-swift/59524/46), rather than rolling our own flags as I'm doing here. I was just following the existing method in the compiler config, but if you guys plan on rolling that new cross-compilation method out, I'm happy to wait and see how that works, then chip in my own changes as needed.

PR #2511, opened 2024-02-25
Fix trivia handling issues in editor placeholder expansion

ahoppen:
rdar://123287930

### DIFF
--- a/Sources/SwiftBasicFormat/BasicFormat.swift
+++ b/Sources/SwiftBasicFormat/BasicFormat.swift
@@ -69,6 +69,16 @@ open class BasicFormat: SyntaxRewriter {
     super.init(viewMode: viewMode)
   }
 
+  /// Clears all stateful data from this `BasicFormat`.
+  ///
+  /// This needs to be called between multiple `rewrite` calls to a syntax tree.
+  func reset() {
+    indentationStack = [indentationStack.first!]
+    anchorPoints = [:]
+    previousToken = nil
+    stringLiteralNestingLevel = 0
+  }
+
   // MARK: - Updating indentation level
 
   public func increaseIndentationLevel(to userDefinedIndentation: Trivia? = nil) {
@@ -145,7 +155,7 @@ open class BasicFormat: SyntaxRewriter {
 
   /// Find the indentation of the nearest ancestor whose first token is an
   /// anchor point (see `anchorPoints`).
-  private func anchorPointIndentation(for token: TokenSyntax) -> Trivia? {
+  private func anchorPointIndentation(for token: TokenSyntax) -> Trivia {
     var ancestor: Syntax = Syntax(token)
     while let parent = ancestor.parent {
       ancestor = parent
@@ -155,7 +165,7 @@ open class BasicFormat: SyntaxRewriter {
         return anchorPointIndentation
       }
     }
-    return nil
+    return Trivia()
   }
 
   // MARK: - Customization points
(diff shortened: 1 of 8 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
https://github.com/apple/swift/pull/71877

@swift-ci Please test

ahoppen:
https://github.com/apple/swift/pull/71877

@swift-ci Please test Windows

Sources/SwiftBasicFormat/BasicFormat.swift
```diff
@@ -575,23 +585,24 @@ open class BasicFormat: SyntaxRewriter {
       trailingTrivia += .space
     }
 
-    var leadingTriviaIndentation = self.currentIndentationLevel
-    var trailingTriviaIndentation = self.currentIndentationLevel
+    var leadingTriviaIndentation: Trivia
+    var trailingTriviaIndentation: Trivia
```
Matejkob, line 589: I believe since you removed default values in these properties we can change them to `let`'s since they are not mutating anymore.

```suggestion
    let leadingTriviaIndentation: Trivia
    let trailingTriviaIndentation: Trivia
```

Sources/SwiftBasicFormat/BasicFormat.swift
```diff
@@ -69,6 +69,16 @@ open class BasicFormat: SyntaxRewriter {
...
   }
 
+  /// Clears all stateful data from this `BasicFormat`.
+  ///
+  /// This needs to be called between multiple `rewrite` calls to a syntax tree.
+  public func reset() {
+    self.indentationStack = [self.indentationStack.first!]
+    self.anchorPoints = [:]
+    self.previousToken = nil
+    self.stringLiteralNestingLevel = 0
```
Matejkob, line 79: NIT-PICK ALERT!

From what I can see we are not using `self` in other places in this class. 
```suggestion
    indentationStack = [indentationStack.first!]
    anchorPoints = [:]
    previousToken = nil
    stringLiteralNestingLevel = 0
```

ahoppen: Oh, if only we were consistent about using `self` or not using `self`. But good catch, I removed it.

Matejkob: Doesn't swift-format has a rule for that? 🤔

ahoppen: No, because knowing whether `self` requires semantic knowledge. For example you can’t omit the `self` if there’s also a local variable with the same name.

ahoppen:
@swift-ci Please test

ahoppen:
https://github.com/apple/swift/pull/71877

@swift-ci Please test

ahoppen:
https://github.com/apple/swift/pull/71877

@swift-ci Please test macOS

Sources/SwiftBasicFormat/BasicFormat.swift
```diff
@@ -69,6 +69,16 @@ open class BasicFormat: SyntaxRewriter {
     super.init(viewMode: viewMode)
   }
 
+  /// Clears all stateful data from this `BasicFormat`.
+  ///
+  /// This needs to be called between multiple `rewrite` calls to a syntax tree.
```
bnbarham, line 74: Could we just override `rewrite` and do this in there instead?

ahoppen: `rewrite` currently isn’t `open`. And I don’t really want to make it `open` just for that reason… Thinking about this, I thought about introducing `BasicFormat.format(_ node: some SyntaxProtocol)` but that would be the same as `SyntaxProtocol.formatted(using: BasicFormat)`. So, maybe we should just make `reset` internal and tell people to use `SyntaxProtocol.formatted`?

ahoppen: Updated it to make `reset` internal again with the intention that you format code by calling `SyntaxProtocol.formatted`.

ahoppen:
@swift-ci Please test

ahoppen:
https://github.com/apple/swift/pull/71877

@swift-ci Please test

bnbarham (approved):

PR #2512, opened 2024-02-27
Use character columns instead of UTF-8 columns in the diagnostics printer

### LINKED ISSUE #2507: `DiagnosticsFormatter` crashes when source code contains an emoji
### Description

`DiagnosticsFormatter` crashes with a  `Fatal error: Range requires lowerBound <= upperBound` message when the source code contains an emoji. 

Screenshot 2024-02-21 at 10 51 44 PM

### Steps to Reproduce

Run: 

```bash
swift run swift-parser-cli print-diags -s 'func o() {
  _ = {
    0
    /x}}1️⃣} /
    2
  }
}'
```

This version:
```bash
swift run swift-parser-cli print-diags -s 'func o() {
  _ = {
    0
    /x}}$} /
    2
  }
}'
```
works.

Tested on `main`.

ahoppen:
Fixes a crash and improves the printing of errors on lines that contain multi-byte UTF-8 characters.

Fixes #2507
rdar://123409748

### DIFF
--- a/Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
+++ b/Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
@@ -68,6 +68,22 @@ public struct DiagnosticsFormatter {
     var isFreeOfAnnotations: Bool {
       return diagnostics.isEmpty && suffixText.isEmpty
     }
+
+    /// Converts a UTF-8 column index into an index that considers each character as a single column, not each UTF-8
+    /// byte.
+    ///
+    /// For example the 👨‍👩‍👧‍👦 character is considered as a single character, not 25 bytes.
+    ///
+    /// Both the input and the output column are 1-based.
+    func characterColumn(ofUtf8Column utf8Column: Int) -> Int {
+      let index =
+        sourceString.utf8.index(
+          sourceString.utf8.startIndex,
+          offsetBy: utf8Column - 1,
+          limitedBy: sourceString.utf8.endIndex
+        ) ?? sourceString.utf8.endIndex
+      return sourceString.distance(from: sourceString.startIndex, to: index) + 1
+    }
   }
 
   /// Number of lines which should be printed before and after the diagnostic message
@@ -139,7 +155,7 @@ public struct DiagnosticsFormatter {
 
       let endColumn: Int
       if endLine > lineNumber {
-        endColumn = annotatedLine.sourceString.count
+        endColumn = annotatedLine.sourceString.utf8.count
       } else if endLine == lineNumber {
         endColumn = endLoc.column
       } else {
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
```diff
@@ -68,6 +68,22 @@ public struct DiagnosticsFormatter {
...
       return diagnostics.isEmpty && suffixText.isEmpty
     }
+
+    /// Converts a UTF-8 column index into an index that considers each character as a single column, not each UTF-8
+    /// byte.
+    ///
+    /// For example the 👨‍👩‍👧‍👦 character is considered as a single character, not 25 bytes.
+    ///
+    /// Both the input and the output column are 1-based.
+    func characterColumn(ofUtf8Column utf8Column: Int) -> Int {
```
bnbarham, line 78: I assume you don't think this should be part of the location converter?

bnbarham (approved):

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
ahoppen: No, I think it’s quite specific to this use case.

ahoppen:
@swift-ci Please test Windows

Matejkob (commented):
I'm not so sure if considering emojis as one character is correct. Are they two-character symbols? 

I believe in this test:
```swift
func testEmojiInSourceCode() {
    let source = """
      let 👨‍👩‍👧‍👦 = ;
      """

    let expectedOutput = """
      1 │ let 👨‍👩‍👧‍👦 = ;
        │         ╰─ error: expected expression in variable
      """

    assertStringsEqualWithDiff(annotate(source: source), expectedOutput)
}
```
the diagnostic should point to `;` not the space between `=` and `;`. I can not test it right now, but I think if we change the variable name to regular characters like so:
```swift
func testEmojiInSourceCode() {
    let source = """
      let e = ;
      """

    let expectedOutput = """
      1 │ let e = ;
        │         ╰─ error: expected expression in variable
      """

    assertStringsEqualWithDiff(annotate(source: source), expectedOutput)
}
```
the annotation will be pointing to semi column.

ahoppen:
The problem is that, as far as I know, it depends on the font that renders the font. A font can render an emoji as a single-width character, a double width character, anything in between or however it likes, really. And as far as I know there’s no Unicode specification for it and my thought that assuming that 👨‍👩‍👧‍👦 is rendered as a single character is closer to the truth than assuming that it gets rendered as 25 characters.

Matejkob:
Ran the emoji test, and indeed, the annotation shifts on difficult Xcode font (also in the terminal), highlighting the complexity of emoji rendering in code. 
Screenshot 2024-02-27 at 9 09 49 PM
Your explanation about the unpredictable rendering of emojis due to font differences is super insightful, thanks!

Out of curiosity, I've built this C code:
```c
#include <stdio.h>

int main() {
    printf("Hello, World!🤯")
    return 0;
}
```
and the annotation in the diagnostic method produced by gcc is shifted as well:
 Screenshot 2024-02-27 at 9 11 48 PM

al45tair:
> The problem is that, as far as I know, it depends on the font that renders the font. A font can render an emoji as a single-width character, a double width character, anything in between or however it likes, really. And as far as I know there’s no Unicode specification for it and my thought that assuming that 👨‍👩‍👧‍👦 is rendered as a single character is closer to the truth than assuming that it gets rendered as 25 characters.

This is mainly to do with terminal emulator behaviour rather than the font (though the font can certainly make things worse if it so desires); generally speaking terminal emulators try to match the behaviour of the `wcwidth()` function on their respective platform, and _if_ that supports Unicode, it tends to just use the narrow or wide property from the East Asian Width table without doing anything much more complicated. That usually goes wrong for cases where the base character is narrow but some Emoji sequence is in play (so that the result is wide).

PR #2513, opened 2024-02-27
Generalize computing SWIFT_HOST_MODULE_TRIPLE

etcwilde:
This patch updates the logic for figuring out the Swift module triple from manually trying to patch something together to just asking the Swift driver to tell us what it is based on what we're compiling for.

I'm a little confused on the terminology/intent though. The name is `SWIFT_HOST_MODULE_TRIPLE`, which in CMake parlance would be the build machine, while it's looking at variables like `CMAKE_SYSTEM_PROCESSOR`, which are the machine it's building for. I'm going with the machine it's building for in this though.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,23 +63,12 @@ endif()
 
 # Determine the module triple.
 if("${SWIFT_HOST_MODULE_TRIPLE}" STREQUAL "")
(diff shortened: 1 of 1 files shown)


bnbarham (approved):

etcwilde:
@swift-ci please test

PR #2514, opened 2024-02-27
Add API to infer the indentation of a syntax tree

ahoppen:
Part of rdar://123600798

### DIFF
--- a/Sources/SwiftBasicFormat/InferIndentation.swift
+++ b/Sources/SwiftBasicFormat/InferIndentation.swift
@@ -1,0 +1,127 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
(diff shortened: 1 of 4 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test macOS

ahoppen:
@swift-ci Please test macOS

Sources/SwiftBasicFormat/InferIndentation.swift
```diff
@@ -0,0 +1,127 @@
...
+    visitor.walk(tree)
+    if visitor.linesProcessed < 3 {
+      // We don't have enough lines to infer indentation reliably
+      return nil
+    }
+
+    // Pick biggest indentation that encompasses at least 90% of the source lines.
+    let threshold = Int(Double(visitor.linesProcessed) * 0.9)
+
+    for spaceIndentation in [8, 4, 2] {
```
bnbarham, line 98: Poor 3 space users 😅. Is 8 actually a thing that people use?

I think one issue we're likely to run into here is that for highly nested snippets, this is often going to return 8 even if it's eg. 2 space. Is the idea that this would only be run on entire files/functions/something else?

ahoppen: That shouldn’t be an issue. If we have 4 lines indented with 8 spaces and 1 line indented with 10 spaces, we’ll infer the indentation to 2 spaces because only 80% of the lines have an indentation that’s a multiple of 8 but all lines have an indentation that’s a multiple of 2. 

And I think I’ve seen codebases with 8 spaces of indentation. Can’t remember where and when, though.

PR #2516, opened 2024-02-29
[Parser] Cleanup UnicodeScalar handling

rintaro:
Mostly just removing  redundant `UnicodeScalar()` callings.

Change associated type of `EscapedCharacterLex.success(UInt32)` and `CharacterLex.validatedEscapeSequence(Character)` to both `Unicode.Scalar` because there's no reason not to use it, and to avoid unnecessary conversions.

### DIFF
--- a/Sources/SwiftParser/StringLiteralRepresentedLiteralValue.swift
+++ b/Sources/SwiftParser/StringLiteralRepresentedLiteralValue.swift
@@ -71,8 +71,9 @@ extension StringSegmentSyntax {
   ) {
     precondition(!hasError, "appendUnescapedLiteralValue relies on properly parsed literals")
 
-    var text = content.text
-    text.withUTF8 { buffer in
+    let rawText = content.rawText
+
+    rawText.withBuffer { buffer in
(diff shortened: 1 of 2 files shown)


Sources/SwiftParser/StringLiteralRepresentedLiteralValue.swift
```diff
@@ -71,8 +71,14 @@ extension StringSegmentSyntax {
...
     precondition(!hasError, "appendUnescapedLiteralValue relies on properly parsed literals")
 
-    var text = content.text
-    text.withUTF8 { buffer in
+    let rawText = content.rawText
+    if !rawText.contains("\\") {
+      // Fast path. No escape sequence.
+      output.append(String(syntaxText: rawText))
+      return
+    }
```
rintaro, line 79: This first path is an addition. It's _probably_ just faster than always lexing the UTF8 sequence.

ahoppen (approved):

rintaro:
@swift-ci Please smoke test

rintaro:
@swift-ci Please test

PR #2517, opened 2024-03-01
(fix): Bug where length return is negative

mustiikhalil:
Fixes a bug where the SourceEdit will return a negative length


ahoppen (approved):
Thanks. That bug was kind of embarrassing.

ahoppen:
@swift-ci Please test

PR #2519, opened 2024-03-02
Offer `EditorPlaceholderData` as SPI to sourcekit-lsp

ahoppen:
This way, we don’t have to duplicate the logic between swift-syntax and sourcekit-lsp.

Companion of https://github.com/apple/sourcekit-lsp/pull/1114


bnbarham (approved):
You could change the SPI of the one we moved to IDEUtils to LSP too.

ahoppen:
@swift-ci Please test

PR #2520, opened 2024-03-02
Group version marker modules in one directory

Matejkob:
Very small change but I believe it reduces the noise in the main, `Sources` directory in the SwfitPM. With more and more Swift versions it should be helpful.


ahoppen (approved):
Seems reasonable to me. Thanks

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test macOS

PR #2522, opened 2024-03-02
[bazel] Update BUILD.bazel for new deps

keith:


keith:
@swift-ci please test

PR #2525, opened 2024-03-05
include freestanding macros in lexical context

elizablock:
This makes it possible for a macro to know if it appears within an argument to a freestanding macro.

[Updated: now includes the full freestanding macro syntax in the lexical context. However, this required a new way of unit testing, because the previous approach, which emitted all the decls from the lexical context as the expansion of a freestanding macro `#allLexicalContexts`, produced a cycle when one of the decls in the lexical context itself contained the macro `#allLexicalContexts`.]

rdar://122330706

### DIFF
--- a/Sources/SwiftSyntaxMacros/Syntax+LexicalContext.swift
+++ b/Sources/SwiftSyntaxMacros/Syntax+LexicalContext.swift
@@ -57,6 +57,11 @@ extension SyntaxProtocol {
       patternBinding.initializer = nil
       return Syntax(patternBinding)
 
+    // Freestanding macros are fine as-is because if any arguments change
+    // the whole macro would have to be re-evaluated.
+    case let freestandingMacro as FreestandingMacroExpansionSyntax:
+      return Syntax(freestandingMacro.detached) as Syntax
+
     default:
       return nil
     }
--- a/Tests/SwiftSyntaxMacroExpansionTest/LexicalContextTests.swift
+++ b/Tests/SwiftSyntaxMacroExpansionTest/LexicalContextTests.swift
@@ -200,6 +200,21 @@ public struct AllLexicalContextsMacro: DeclarationMacro {
   }
 }
 
+public struct LexicalContextDescriptionMacro: ExpressionMacro {
+  public static func expansion(
+    of node: some FreestandingMacroExpansionSyntax,
+    in context: some MacroExpansionContext
+  ) throws -> ExprSyntax {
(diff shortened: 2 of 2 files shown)


Sources/SwiftSyntaxMacros/Syntax+LexicalContext.swift
```diff
@@ -56,6 +56,14 @@ extension SyntaxProtocol {
...
       patternBinding.initializer = nil
       return Syntax(patternBinding)
+        
+    // Freestanding macros have their arguments and trailing closures removed.
+    case var freestandingMacro as FreestandingMacroExpansionSyntax:
+      freestandingMacro = freestandingMacro.detached
+      freestandingMacro.arguments = LabeledExprListSyntax()
+      freestandingMacro.trailingClosure = nil
+      freestandingMacro.additionalTrailingClosures = MultipleTrailingClosureElementListSyntax()
+      return Syntax(freestandingMacro) as Syntax
```
ahoppen, line 66: Shouldn’t we also remove `leftParen` and `rightParen`? What about the `genericArgumentClause`? Should that stay or go? And lastly `attributes` and `modifiers` for `MacroExpansionDeclSyntax`? I would think that all of them should be removed. 

@DougGregor what do you think?

elizablock: I was following the lead of the other cases -- functions and types keep their parens and their attributes and modifiers.

elizablock: I wasn't sure about the generic argument clause, since a freestanding macro is a bit more like a function call than a declaration. Happy to remove it.

DougGregor: So I, personally, would say that we can just return `freestandingMacro.detached` here

elizablock: Ah -- I didn't understand the purpose of removing that information, but that makes sense. I'll fix.

ahoppen (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

DougGregor:
@swift-ci please test macOS

DougGregor (approved):

DougGregor:
@swift-ci please test macOS

DougGregor:
@swift-ci please clean test macOS

PR #2527, opened 2024-03-05
Store the replacement string in `IncrementalEdit`

ahoppen:
IIRC the main reason that `IncrementalEdit` didn’t contain the replacement string was that the edits needed to be transferred over XPC at some point and transferring strings over XPC is always a little tricky. Since the actual contents weren’t actually needed, the actual contents weren’t actually needed.

I would like to use `ConcurrentEdits(fromSequential:)` in SourceKit-LSP and for that I need the actual replacement contents. Instead of rolling our own sequential to concurrent replacement (which is non-trivial), let’s re-use the one in SwiftSyntax.

### DIFF
--- a/Release Notes/600.md
+++ b/Release Notes/600.md
@@ -1,13 +1,13 @@
 # Swift Syntax 600 Release Notes
 
 ## New APIs
-- FixIt now has a new computed propery named edits
-  - Description: the edits represent the non-overlapping textual edits that need to be performed when the Fix-It is applied.
+- FixIt now has a new computed property named `edits`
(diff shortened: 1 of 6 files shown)


ahoppen:
@swift-ci Please test

bnbarham (approved):

ahoppen:
@swift-ci Please test macOS

Matejkob (commented):
Should we add a note to the changelog about this change?

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test w

ahoppen:
@swift-ci Please test Windows

PR #2528, opened 2024-03-05
Check compiler version to determine if scoped imports are supported

ahoppen:
We were checking for the language version, which is 5.10 for a Swift 6 compiler in Swift 5 mode (which is the default). But we don’t actually care about the selected language mode, but about the compiler’s version to check if scoped imports are supported.


bnbarham (approved):

ahoppen:
Superseded by https://github.com/apple/swift-syntax/pull/2531

PR #2537, opened 2024-03-08
[WIP][5.10] Significantly improve performance of SyntaxVisitor

### LINKED ISSUE #2091: Improve performance of `SyntaxVisitor`
When @StevenWong12 authored https://github.com/apple/swift-syntax/pull/2087, we noticed that the new `SyntaxVisitor`-based implementation is 50% slower than the old implementation. It looks `SyntaxVisitor` is performing quite a few more retain and release calls that might not be necessary. We should investigate if there’s something we can do about that.

For newcomers to the project: Note that this is a fairly advanced issue. It will require debugging code at the SIL level and I’m not sure if there’s a road to success.

ahoppen:
There are two core ideas here:
- Before this change, we were allocating `Syntax.Info` for every visited node. We need a heap-allocated object here because `Syntax` can access its parent. In most cases, however, the syntax node was not stored by the visitor and we would just destroy the `Syntax.Info` after finishing the `visit` call. Instead, check if `Syntax.Info` is uniquely referenced, which means that the the node was not stored. In that case, don’t deallocate the memory but place it into a reuse pool. I have seen reductions in `malloc` calls from ~50,000 to ~100 (500x) in `testEmptyVisitorPerformance`.
- Now retain and release calls made up a significant portion of the visitor’s performance. Add some annotations to eliminate reference counting. I added comments for the remaining retain calls.

I measured that this improves the performance of the `EmptyParametersRule` in SwiftLint by ~35% compared to `509.0.0` (and by ~50% compared to `510.0.0`).

Fixes #2091
rdar://114330163

### DIFF
--- a/CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxVisitorFile.swift
+++ b/CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxVisitorFile.swift
@@ -34,6 +34,24 @@ let syntaxVisitorFile = SourceFileSyntax(leadingTrivia: copyrightHeader) {
 
     DeclSyntax(
       """
+      /// `Syntax.Info` objects created in `visitChildren` but whose `Syntax` nodes were not retained by the `visit`
+      /// functions implemented by a subclass of `SyntaxVisitor`.
+      ///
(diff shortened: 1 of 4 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
Assessed that the risk of taking this in a patch release of 510 is too big.

PR #2538, opened 2024-03-11
Exclude SwiftCompilerPlugin from `all` target

etcwilde:
The SwiftCompilerPlugin library is only needed when writing Swift macros, but not by the rest of the Swift compiler. The Swift build-script only knows how to build the `all` target, resulting in this target also being built as part of the compiler when it should not be.

rdar://124395736


etcwilde:
@swift-ci please test

bnbarham (approved):

PR #2539, opened 2024-03-12
Add an introducer property to DeclGroupSyntax

### LINKED ISSUE #2535: DeclGroupSyntax should have introducerKeyword
### Description

It should be possible to `.as(DeclGroupSyntax.self)?.introducerKeyword` when writing macros in order to easily write diagnostics like "cannot be applied to **struct**"  etc.

Currently one has to do manual if/else of all the kinds we support

### Steps to Reproduce

_No response_

bnbarham:
It can be useful to grab the introducer of a particular declaration, add a property for this purpose to `DeclGroupSyntax`.

Fixes #2535.
Resolves rdar://124247480.


bnbarham:
@swift-ci please test

Sources/SwiftSyntax/generated/SyntaxTraits.swift
```diff
@@ -72,6 +72,22 @@ public protocol DeclGroupSyntax: SyntaxProtocol, DeclSyntaxProtocol {
     set
   }
```
bnbarham, line 74: Also open to renaming `DeclGroupSyntax` if anyone has a better name for it.

ktoso (approved):
Nice, thank you!

Sources/SwiftSyntax/CustomTraits.swift
```diff
@@ -0,0 +1,80 @@
...
+  var introducer: TokenSyntax {
+    get {
+      return extensionKeyword
+    }
+    set {
+      extensionKeyword = newValue
+    }
+  }
+}
+extension ProtocolDeclSyntax {
```
ahoppen, line 56: Very nitpicky: Above you have a newline between the extensions, here you don’t.

Sources/SwiftSyntax/generated/SyntaxTraits.swift
ahoppen: I’m not a huge fan of my suggestion but what this conceptually is, is a `DeclWithMembersSyntax`, I think. At least that describes the trait IMO.

Sources/SwiftSyntax/CustomTraits.swift
bnbarham: Ah thanks, copy paste strikes again :)

Sources/SwiftSyntax/generated/SyntaxTraits.swift
bnbarham: That matches the other "With" at least 😅

bnbarham:
@swift-ci please test

ahoppen:
@swift-ci Please test Windows

ahoppen (approved):

bnbarham:
@swift-ci please test

bnbarham:
@swift-ci please test

PR #2540, opened 2024-03-14
Resolve `switch covers known cases, but 'Enum' may have additional unknown values` warnings

tshortli:
When building with CMake during the Swift compiler build, the libraries in this package are built with library evolution enabled. That causes the compiler to emit diagnostics that encourage use of `@unknown default` to handle unknown cases that might be added in the future. Silence these warnings by adding `@unknown default` cases where they are missing. Guard these with the `RESILIENT_LIBRARIES` compilation condition to avoid triggering a `Default will never be executed` diagnostic during the package build.


tshortli:
@swift-ci please test

tshortli:
@swift-ci please smoke test

ahoppen (approved):
Thanks!

tshortli:
@swift-ci please test macOS

tshortli:
@swift-ci please test Windows

PR #2541, opened 2024-03-14
Allow return types on initializers

ahoppen:
SE-2305 allows returning `Self` and `Self?` from initializers so that the return type can be annotated with `_borrow(a)` etc.

So we need to allow return clauses on initializers.

rdar://123905900

### DIFF
--- a/Sources/SwiftParser/Declarations.swift
+++ b/Sources/SwiftParser/Declarations.swift
@@ -968,7 +968,7 @@ extension Parser {
     }
 
     // Parse the signature.
-    let signature = self.parseFunctionSignature(allowOutput: false)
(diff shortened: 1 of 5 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

bnbarham (approved):

bnbarham:
Note that this should probably go behind experimental

ahoppen:
I don’t think it needs to be behind an experimental flag because the syntax tree has always allowed return types on initializers.

meg-gupta:
Thanks!

Issue #2533, opened 2024-03-06
Parser enters infinite loop in string parsing


ahoppen:
### Issue Kind

Parser Crash

### Source Code

```swift
"\("\(
```

ahoppen:
Tracked in Apple’s issue tracker as rdar://124168557

PR #2545, opened 2024-03-16
[6.0] Fix a false negative assertion failure in `LoopProgressCondition` for…

ahoppen:
* **Explanation**: If we have two nested, unterminated string interpolation segments, the lexer generates two empty `stringLiteral` tokens (one after each interpolation segment). When consuming the first empty string segment, we did actually make progress in the lexer by popping one nested string interpolation off the state stack. However, `LoopProgressCondition` did not consider this progress because it only looked at the top state in the state stack.

To fix this, consider the state stack size in `LoopProgressCondition` as well.
* **Scope**: Relaxes an assertion failure
* **Risk**: Minimal, previously this causes an assertion failure in assert builds and was terminating parsing in a slightly undefined way in non-assert builds
* **Issue**: #2533, rdar://124168557
* **Reviewer**:  @bnbarham on https://github.com/apple/swift-syntax/pull/2543

### DIFF
--- a/Sources/SwiftParser/Lexer/Cursor.swift
+++ b/Sources/SwiftParser/Lexer/Cursor.swift
@@ -52,7 +52,7 @@ extension Lexer.Cursor {
   ///  - A string interpolation inside is entered
   ///  - A regex literal is being lexed
   ///  - A narrow case for 'try?' and 'try!' to ensure correct regex lexing
-  enum State {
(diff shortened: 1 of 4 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test macOS

bnbarham (approved):

PR #2547, opened 2024-03-19
Wrap `<SyntaxNode>.parse` calls in a closure inside tests

ahoppen:
Because SE-0418 has not been implemented in Swift 5.10, we get a warning “Converting non-sendable function value to '@Sendable (inout Parser) -> TypeSyntax' may introduce data races” when building SwiftSyntax tests using Swift 5.10. Wrapping the calls to `parse` in closures allows the type system to infer the actor isolation for the closure, fixing the warning.

rdar://124993444

### DIFF
--- a/Tests/SwiftParserTest/AttributeTests.swift
+++ b/Tests/SwiftParserTest/AttributeTests.swift
@@ -361,7 +361,7 @@ final class AttributeTests: ParserTestCase {
 
       assertParse(
         "@_implements(1️⃣\(baseType), f())",
-        AttributeSyntax.parse,
+        { AttributeSyntax.parse(from: &$0) },
         substructure: TypeSyntax.parse(from: &parser),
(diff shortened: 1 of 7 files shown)


ahoppen:
@swift-ci Please test

Tests/SwiftParserTest/AttributeTests.swift
```diff
@@ -361,7 +361,7 @@ final class AttributeTests: ParserTestCase {
 
       assertParse(
         "@_implements(1️⃣\(baseType), f())",
-        AttributeSyntax.parse,
+        { AttributeSyntax.parse(from: &$0) },
```
bnbarham, line 364: Does this actually fix things, or does each closure have to be annotated with `@Sendable`?

ahoppen: It does fix it. Sendable is automatically inferred.

PR #2550, opened 2024-03-19
[6.0] Parse `#if canImport` as a function call instead of a specialized expression node

ahoppen:
* **Explanation**: The parser had special knowledge about how to parse `canImport` inside `#if` conditions. This was inconsistent with eg. the parsing of `#if swift(<6.0)`, which gets parsed as a function call. To make it easier to interpret the version passed to `canImport(MyModule, _version: <#version#>)`, add `ExprSyntax.interpretedAsVersionTuple` that takes eg. `1.2.3`, which got parsed as a member access `3` to the `1.2` float literal and re-parses it as a `VersionTupleSyntax`. This relaxation also allows string literals as version parameters to `#if canImport`.
* **Scope**: Parsing of `canImport` directives in `#if` conditions.
* **Risk**: Low, we’ve to quite a few tests for this, function calls are more permissive than `CanImportExprSyntax` and it is unlikely that anyone is relying on the interpretation of the version number
* **Testing**: Updated the existing test cases
* **Issue**: rdar://121070235, #1972
* **Reviewer**:  @bnbarham and @rintaro on https://github.com/apple/swift-syntax/pull/2025


ahoppen:
@swift-ci Please test

bnbarham (approved):

ahoppen:
@swift-ci Please test Windows

PR #2552, opened 2024-03-20
[6.0] Significantly improve performance of SyntaxVisitor

ahoppen:
* **Explanation**: This improves the performance of `SyntaxVisitor` by minimizing the number of memory allocations we need to perform, resulting in a ~35% performance improvement in `EmptyParametersRule` of SwiftLint compared to `509.0.0` (and by ~50% compared to `510.0.0`).
* **Scope**: And client that uses `SyntaxVisitor`
* **Risk**: There’s always an inherent risk when doing some kind of manual memory management but I think we should be safe because we didn’t find any issues in the swift-syntax tests or when running SwiftLint
* **Testing**: 
  * Correctness: Ran swift-syntax tests and SwiftLint rules
  * Performance: Measured performance of `SwiftLint` and `testEmptyVisitorPerformance` and verified on the SIL level that we do a minimal amount of ref-counting
* **Issue**: rdar://114330163
* **Reviewer**:  @bnbarham on https://github.com/apple/swift-syntax/pull/2536


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #2553, opened 2024-03-20
[6.0] Allow return types on initializers

ahoppen:
* **Explanation**: SE-2305 allows returning `Self` and `Self?` from initializers so that the return type can be annotated with `_borrow(a)` etc. So we need to allow return clauses on initializers.
* **Scope**: Parsing of initializers that have a return clause
* **Risk**: I don’t see a risk in this change
* **Testing**: Added a regression test
* **Issue**: rdar://123905900
* **Reviewer**:  @bnbarham on https://github.com/apple/swift-syntax/pull/2541

### DIFF
--- a/Sources/SwiftParser/Declarations.swift
+++ b/Sources/SwiftParser/Declarations.swift
@@ -968,7 +968,7 @@ extension Parser {
     }
 
     // Parse the signature.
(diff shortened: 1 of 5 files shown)


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #2548, opened 2024-03-19
Specify attributes inside `#if` before any unconditional attributes

ahoppen:
Looks like the Swift 5.7 compiler errors if unconditional attributes are before attributes inside `#if`.

Fixes #2546
rdar://124979624

Issue #2546, opened 2024-03-19
Possible compiler backward compatibility problem


dabrahams:
See https://github.com/apple/swift-syntax/commit/ea71e1eb2373b059322db8ae4092f8051f650511#r139940839

ahoppen:
Tracked in Apple’s issue tracker as rdar://124979624

ahoppen:
Should be fixed by https://github.com/apple/swift-syntax/pull/2548.

PR #2556, opened 2024-03-21
[6.0] Specify attributes inside `#if` before any unconditional attributes

ahoppen:
* **Explanation**: Specifying the `#if` conditionalized attribute after an unconditional attribute results in a compilation error when using a Swift 5.7 compiler. Reversing the order makes swift-syntax build with a Swift 5.7 compiler.
* **Scope**: Fixes a compilation error when building swift-syntax with a 5.7 compiler
* **Risk**: Very low, just flips two declaration attributes
* **Testing**: Tested that that the package builds with 5.7 with this change
* **Issue**: rdar://124979624, #2546 
* **Reviewer**:  @bnbarham on https://github.com/apple/swift-syntax/pull/2548


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #2557, opened 2024-03-21
Make `DiagnosticDecorator` semi-public

Matejkob:
The refactored implementation of `DiagnosticFormatter`, as seen in [this pull request](https://github.com/apple/swift-syntax/pull/2214), is now closely integrated with `DiagnosticDecorator`. In order to expose the new initializers and methods of `DiagnosticFormatter` as public, it's necessary to also make `DiagnosticDecorator` public.

Given that `decorateHighlight` is still experimental, I've renamed it to start with an underscore (`_`) and added a warning to it's documentation to emphasize that this API is not yet intended for production use.


ahoppen:
What’s your motivation for making the type public? Do you expect clients to provide their own diagnostic formatter?

Matejkob:
Yes, indeed. The original intension was to allow clients to customize diagnostic formatters.

ahoppen:
Sorry for the delay: Do you have a concrete example of a client that might want to customize the diagnostic formatter? I would prefer to not increase the API surface area unnecessarily

PR #2558, opened 2024-03-21
Make swift-syntax build without warnings

ahoppen:
The main problem here was that we can’t just deprecate syntax nodes since swift-syntax needs to still be able to use them internally (eg. when creating `SyntaxEnum` cases for `as(SyntaxEnum.self)`). What we do instead, is to have an underscored node that is not deprecated and deprecate a public typealias from the non-underscored version to the underscored version of the node.

While doing this, I also realized that the `SyntaxKind` compatibility layer didn’t add deprecation warnings and that we were still using some of them.

And finally, we need to annotate `Syntax.Info.info` as `nonisolated(unsafe)`. It already had a comment that detailed when it might be modified.


ahoppen:
@swift-ci Please test

ahoppen:
https://github.com/apple/swift/pull/72557

@swift-ci Please test

bnbarham (approved):

ahoppen:
https://github.com/apple/swift/pull/72557

@swift-ci Please test

ahoppen:
https://github.com/apple/swift/pull/72557

@swift-ci Please test Windows

rintaro (approved):

ahoppen:
https://github.com/apple/swift/pull/72557

@swift-ci Please test

ahoppen:
https://github.com/apple/swift/pull/72557

@swift-ci Please test Windows

PR #2564, opened 2024-03-24
Update lifetime dependence syntax

meg-gupta:
Pitch - https://github.com/apple/swift-evolution/pull/2305

Update spellings to:
dependsOn(paramName...) and dependsOn(scoped paramName...) syntax


meg-gupta:
@swift-ci test

meg-gupta:
https://github.com/apple/swift/pull/72506

@swift-ci Please test

meg-gupta:
https://github.com/apple/swift/pull/72506

@swift-ci Please test

meg-gupta:
https://github.com/apple/swift/pull/72506

@swift-ci Please test

meg-gupta:
@swift-ci smoke test windows platform

CodeGeneration/Sources/SyntaxSupport/TypeNodes.swift
```diff
@@ -581,21 +570,32 @@ public let TYPE_NODES: [Node] = [
     experimentalFeature: .nonescapableTypes,
     nameForDiagnostics: "lifetime specifier",
     documentation: "A specifier that specifies function parameter on whose lifetime a type depends",
+    traits: [
+      "Parenthesized"
+    ],
```
ahoppen, line 575: The node shouldn’t have the `Parenthesized` trait because the left parenthesis is not the first token. You need to add an XFail for this in the CodeGeneration tests.

CodeGeneration/Sources/SyntaxSupport/TypeNodes.swift
```diff
@@ -581,21 +570,32 @@ public let TYPE_NODES: [Node] = [
...
-        name: "specifier",
-        kind: .token(choices: [
-          .keyword(._copy),
-          .keyword(._consume),
-          .keyword(._borrow),
-          .keyword(._mutate),
-        ]),
+        name: "dependsOnKeyword",
+        kind: .token(choices: [.keyword(.dependsOn)]),
         documentation: "The specifier token that's attached to the type."
```
ahoppen, line 580: I think the documentation is no longer up-to-date.

CodeGeneration/Sources/SyntaxSupport/TypeNodes.swift
```diff
@@ -581,21 +570,32 @@ public let TYPE_NODES: [Node] = [
...
       ),
+      Child(
+        name: "leftParen",
+        kind: .token(choices: [.token(.leftParen)])
+      ),
+      Child(
+        name: "scopedKeyword",
+        kind: .token(choices: [.keyword(.scoped)]),
```
ahoppen, line 588: Could you add a line of documentation what the `scoped` keyword does?

Sources/SwiftParser/Types.swift
```diff
@@ -900,33 +900,33 @@ extension Parser.Lookahead {
 }
 
 extension Parser {
-  private mutating func parseLifetimeTypeSpecifier(
-    specifierHandle: TokenConsumptionHandle
-  ) -> RawTypeSpecifierListSyntax.Element {
-    let specifier = self.eat(specifierHandle)
+  private mutating func parseLifetimeTypeSpecifier() -> RawTypeSpecifierListSyntax.Element {
+    let specifier = self.eat(TokenSpec(.dependsOn))
```
ahoppen, line 904: Could you use `self.expect` instead of `self.eat`. That’s generally safer because it won’t hit an assertion failure if we currently aren’t positioned at `dependsOn` (which might become the case in the future if we should decide to change the `self.at(.dependsOn)` to `self.canRecoverTo(.dependsOn)` in `parseTypeAttributeList ` in the future). We just generally try to be as error-tolerant as possible in the parser.

---

Also, I would call the variable `depensOnKeyword` because it no longer is an arbitrary specifier.

---

To be consistent with the rest of the codebase I would use `.keyword(.dependsOn)` instead of `TokenSpec(.dependsOn)`.

Sources/SwiftParser/Types.swift
```diff
@@ -976,20 +977,18 @@ extension Parser {
...
-    SPECIFIER_PARSING: while canHaveParameterSpecifier,
-      let (specifierSpec, specifierHandle) = self.at(anyIn: SimpleOrLifetimeSpecifier.self)
-    {
-      switch specifierSpec {
-      case .lhs: specifiers.append(parseSimpleTypeSpecifier(specifierHandle: specifierHandle))
-      case .rhs:
+    SPECIFIER_PARSING: while canHaveParameterSpecifier {
+      if let (_, specifierHandle) = self.at(anyIn: SimpleTypeSpecifierSyntax.SpecifierOptions.self) {
+        specifiers.append(parseSimpleTypeSpecifier(specifierHandle: specifierHandle))
+      } else if self.at(TokenSpec(.dependsOn)) {
```
ahoppen, line 984: Can be simplified slightly to

```swift
} else if self.at(.keyword(.dependsOn)) {
```

PR #2569, opened 2024-03-25
Fix two diagnostics when building swift-syntax using pre Swift-6 compilers

ahoppen:
- Only mark keyword as experimental when using a Swift ≥5.8 compiler
  - In most other parts of the code base we only use the SPI attribute for Swift ≥5.8 compilers. If we unconditionally mark a keywords as SPI we run into issues if a syntax node (which is only experimental in Swift 5.8+) uses that keyword as a default value in its initializers. The concrete case here is the usage of `.dependsOn`.
- Mark `Syntax.Info` as `@unchecked Sendable`
  - We were relying on the `nonisolated(unsafe)` annotation of `Info.info` to make `Sendable` checking pass. But that feature is only available in Swift 6.0. To be able to build swift-syntax without warnings using Swift <6.0, we need to use `@unchecked Sendable` here.


ahoppen:
@swift-ci Please test

PR #2570, opened 2024-03-26
Add @implementation to parser

beccadax:
Matches apple/swift#72596.

### DIFF
--- a/CodeGeneration/Sources/SyntaxSupport/KeywordSpec.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/KeywordSpec.swift
@@ -221,6 +221,7 @@ public enum Keyword: CaseIterable {
   case `guard`
   case higherThan
   case `if`
+  case implementation
   case `import`
   case `in`
   case indirect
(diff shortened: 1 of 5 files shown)


beccadax:
With apple/swift#72596

@swift-ci please test

CodeGeneration/Sources/SyntaxSupport/KeywordSpec.swift
```diff
@@ -550,6 +551,8 @@ public enum Keyword: CaseIterable {
       return KeywordSpec("higherThan")
     case .if:
       return KeywordSpec("if", isLexerClassified: true)
+    case .implementation:
+      return KeywordSpec("implementation")
```
ahoppen, line 555: Could you mark the keyword as experimental, eg. like `_borrowing`. For that you’ll need to plumb the experimental feature through.
- Define it in https://github.com/apple/swift-syntax/blob/main/CodeGeneration/Sources/SyntaxSupport/ExperimentalFeatures.swift
- Map it in https://github.com/apple/swift/blob/f81f7f9c544abb362f148589bc18609591c0952d/lib/ASTGen/Sources/ASTGen/SourceFile.swift#L53

beccadax: Sure, will do.

beccadax:
In further discussions with @ahoppen, it became clear that I don't really need to modify SwiftSyntax for this—we can parse it as a custom attribute and convert it to our own attribute in ASTGen. Gonna close this PR and do that instead.

PR #2572, opened 2024-03-27
Add `EffectfulExprSyntax` syntax trait

### LINKED ISSUE #2549: "`EffectfulExprSyntax`" protocol covering `TryExprSyntax` and `AwaitExprSyntax`
### Description

It'd be very helpful to have a protocol that covers all effectful expressions (currently `TryExprSyntax` and `AwaitExprSyntax`.) swift-testing needs to walk syntax trees and diagnose when either of these expression types appear in certain contexts, and because the types are unrelated there is a fair bit of duplicated logic between them.

Matejkob:
Introduces the `EffectfulExprSyntax` protocol, designed to provide a common interface for effectful expressions such as `try` and `await`.

Resolves #2549


CodeGeneration/Sources/SyntaxSupport/Traits.swift
```diff
@@ -66,6 +66,13 @@ public let TRAITS: [Trait] = [
       Child(name: "memberBlock", kind: .node(kind: .memberBlock)),
     ]
   ),
+  Trait(
+    traitName: "EffectfulExpr",
+    documentation: "Syntax trait for effectful expressions, such as `try` and `await`.",
+    children: [
+      Child(name: "expression", kind: .node(kind: .expr))
+    ]
+  ),
```
ahoppen, line 75: We should probably also expose the `try` or `await` keyword, similar to https://github.com/apple/swift-syntax/pull/2539

Matejkob: I've incorporated a `keyword` property into the trait.

However, I have some reservations regarding this API, particularly because using the `try` keyword without accompanying it with either a question mark or an exclamation mark might lead to confusion.

grynspan:
If you do land this change, I'd like to get it cherry-picked to 600 so we can adopt it in swift-testing sooner.

Matejkob:
> If you do land this change, I'd like to get it cherry-picked to 600 so we can adopt it in swift-testing sooner.

I don't see any problems with that on my end.

Do you have an issue in swift-testing regarding adaptation?

grynspan:
swift-testing should generally be able to adopt swift-syntax-600 features once the release is tagged; swift-syntax-610, speculatively, would be quite a ways off.

Matejkob:
Oh, I meant an issue about adopting `EffectfulExprSyntax` syntax trait, so I could look at it. 😅

Release Notes/610.md
Matejkob, line 1: @ahoppen, please let me know if you want to land this PR in version 600. I will then move this release note to the `600.md` file.

CodeGeneration/Sources/SyntaxSupport/Traits.swift
grynspan: Perhaps you could expose an `effectKeyword` property (mirroring `tryKeyword` and `awaitKeyword`) as well as a `questionOrExclamationMark` property that's just always `nil` for `AwaitExprSyntax`?

(In the future, we could conceivably add some sort of `await?` or `await!` syntax and then the property would be automagically available. And no, I don't know what such syntax would _do_ since it's purely hypothetical.)

ahoppen:
I did find the time to actually think about it now and wasn’t able to come up with a concise definition that differentiates `TryExprSyntax` and `AwaitExprSyntax` from any other syntax nodes, which is usually a very good indicator that we’re not on the right track.
At first I thought that try and await are keywords/modifiers/whatever you want to call them that don’t actually change semantic meaning but just indicate something to anyone reading the code. But `TryExprSyntax` also represents `try!` and `try?`, which do change how the expression behaves. With those in mind, I don’t think there’s anything fundamental that’s separating eg. `try?` from eg. optional chaining or force-unwrapping and thus it doesn’t make sense to add a syntax trait for them.

Issue #1936, opened 2023-07-21
Backticks on identifiers should be separate tokens


grynspan:
### Description

If I have an expression like so:

```swift
let x = y.`z`()
```

Then the syntax node representing `z` contains the surrounding backticks. I then need to strip these off manually to get at the actual symbol name so I can use it in other contexts like `context.makeUniqueName()` or diagnostic messages. These backticks should probably be represented by separate tokens, no?

ahoppen:
Tracked in Apple’s issue tracker as rdar://112672035

grynspan:
Or perhaps they should be excluded from `.text`. Not sure the ideal approach here, but having to manually filter them is slowly driving me to madness. :)

ahoppen:
It makes sense that the backticks are part of the identifier token because the `` `z` `` token is formed in the lexer including the backticks and they thus are part of the identifier token. And we used to represent backticks as trivia at some point years ago but moved away from that for reasons that I can’t recall right now, but there were reasons.

I absolutely agree that we need to have an API on `TokenSyntax` that removes the trivia. We briefly touched upon this here: https://github.com/apple/swift-syntax/pull/1816#discussion_r1238428108

The suggestion is that we introduce something like the following and then you could use `token.identifier`.

```swift
struct Identifier {
  var name: String

  init(_ token: TokenSyntax) {
     // FIXME: Handle escaping
    self.name = token.text
  }
}

extension TokenSyntax {
  var identifier: Identifier {
    Identifier(self)
  }
}
```

grynspan:
Oh, sure—I am maybe fixating on a specific technical solution here. Feel free to come up with a better idea! :)

adamcarter93:
This looks interesting! 

I'm currently looking at macros and getting in to swift syntax and would love to contribute my first issue/fix and this one seems a nice isolated one to kick things off

Aware this is 8 months ago so things might have changed but sounds like this is all still viable as above if I were to look in to it? 👨🏻‍💻

ahoppen:
Yes, the issue is still open and still applies.

adamcarter93:
I'm just looking in to this now, from what I can tell from this issue and the one linked by @ahoppen here: https://github.com/apple/swift-syntax/pull/1816#discussion_r1238428108 ...


It looks like the criteria here is:
1. Avoid using backticks as their own trivia
> we used to represent backticks as trivia at some point years ago but moved away from that for reasons that I can’t recall right now, but there were reasons.
[...]

PR #2576, opened 2024-03-28
Add Identifier wrapper that strips backticks from token text

adammcarter:
## Motivation

Following the issue opened in https://github.com/apple/swift-syntax/issues/1936, take the code below as an example:

```swift
let x = y.`z`()
```

When wanting to pull out the character
```swift
`z`
```
you will have to manually strip the backslashes to get the `z` character.

The backslashes are completely valid in this case but often using the Swift syntax you might want to remove the backslashes to get the sanitised name of the token for your own purposes which can introduce unnecessary boilerplate.

## Solution

Thinking forward, this new `Identifier` type allows an abstraction over tokens to further sanitise/strip data out, for example, backticks (as in this PR) or future updates like:
- combined characters, e.g. `é`
- `#` symbols such as in `#if`
- comment blocks like `/*`, `///` or `//`

With the changes in this PR we can take the same example code:
```swift
let x = y.`z`()
```

And by calling `[token].identifier.name` we can get `z` without the backticks.

## Alternatives considered

Stripping the backslashes when returning the `.text` is one solution, but the backticks are a valid part of the text and so wouldn't be an accurate way of returning the source code

Trimming the backticks as part of `trimmedDescription` is a valid and simpler solution IMO, but doesn't allow for future implementations like the above under the `Identifier` abstraction


Tests/SwiftSyntaxTest/IdentifierTests.swift
```diff
@@ -0,0 +1,24 @@
...
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+import XCTest
+
+class IdentifierTests: XCTestCase {
```
adamcarter93, line 16: I got a bit lost on where to put tests exactly so created two new files as the existing places didn't seem right, more than happy to move the tests to somewhere else if there's a better place though

Sources/SwiftSyntax/Identifier.swift
```diff
@@ -0,0 +1,31 @@
...
+  }
+}
+
+private func sanitizing(_ name: String) -> String {
+  guard name.contains("`") else {
+    return name
+  }
+
+  return name.trimmingCharacters(in: CharacterSet(charactersIn: "`"))
+}
```
ahoppen, line 31: It’s only called once so maybe it would be easier to just put the code directly into the initializer.

ahoppen (commented):
Thanks for putting up the PR 👍🏽

Tests/SwiftSyntaxTest/IdentifierTests.swift
adammcarter: Does this include moving [Tests/SwiftSyntaxTest/TokenSyntaxTests.swift](https://github.com/apple/swift-syntax/pull/2576/files#diff-d30dc020c6661e25b14357d63b64979f3452868c9c3917fc60ed9d40ee429c50) to here too? Or keep them separate?

Sources/SwiftSyntax/Identifier.swift
```diff
@@ -0,0 +1,31 @@
...
+
+import Foundation
+
+/// An abstraction for sanitized values on a token.
+public struct Identifier: Equatable, Sendable {
+  /// The sanitized `text` of a token.
+  public let name: String
+
+  public init(_ token: TokenSyntax) {
+    self.name = sanitizing(token.text)
```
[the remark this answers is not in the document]
adammcarter: Sure! 

In what situation is a `TokenSyntax` not an identifier?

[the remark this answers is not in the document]
Matejkob: Here you can find all kinds of tokens:
https://github.com/apple/swift-syntax/blob/cfd0487a9c70bd91935d7ea13ad4410cc4569cf1/Sources/SwiftSyntax/generated/TokenKind.swift#L17-L66

Tests/SwiftSyntaxTest/IdentifierTests.swift
ahoppen: Yes, I would merge the two files you created.

Sources/SwiftSyntax/Identifier.swift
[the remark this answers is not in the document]
Matejkob: Constructing `TokenSyntax` directly from a string literal is not ideal; as far as I understand, it invariably parses the input as an identifier kind. To manually create a token of a different kind, it's recommended to utilize one of the static functions available in `TokenSyntax`: https://github.com/apple/swift-syntax/blob/cfd0487a9c70bd91935d7ea13ad4410cc4569cf1/Sources/SwiftSyntax/generated/Tokens.swift#L16-L20

If you aim to examine tokens within live code, you'll first need to create a syntax node. After that, you can retrieve the tokens using the https://github.com/apple/swift-syntax/blob/cfd0487a9c70bd91935d7ea13ad4410cc4569cf1/Sources/SwiftSyntax/SyntaxProtocol.swift#L329 method.

Here's a snippet for illustration:
```swift
DeclSyntax("let a, b: Int!").tokens(viewMode: .all)
```

Below is an example demonstrating how the token sequence appears:
```
▿ TokenSequence
  ▿ 0 : keyword(SwiftSyntax.Keyword.let)
    - text : "let"
    ▿ leadingTrivia : []
      - pieces : 0 elements
    ▿ trailingTrivia : spaces(1)
      ▿ pieces : 1 element
        ▿ 0 : spaces(1)
          - spaces : 1
    ▿ tokenKind : TokenKind
      - keyword : SwiftSyntax.Keyword.let
  ▿ 1 : identifier("a")
    - text : "a"
    ▿ leadingTrivia : []
      - pieces : 0 elements
    ▿ trailingTrivia : []
      - pieces : 0 elements
    ▿ tokenKind : TokenKind
      - identifier : "a"
  ▿ 2 : comma
    - text : ","
    ▿ leadingTrivia : []
      - pieces : 0 elements
    ▿ trailingTrivia : spaces(1)
      ▿ pieces : 1 element
        ▿ 0 : spaces(1)
          - spaces : 1
    - tokenKind : SwiftSyntax.TokenKind.comma
  ▿ 3 : identifier("b")
    - text : "b"
    ▿ leadingTrivia : []
      - pieces : 0 elements
    ▿ trailingTrivia : []
      - pieces : 0 elements
    ▿ tokenKind : TokenKind
      - identifier : "b"
  ▿ 4 : colon
    - text : ":"
    ▿ leadingTrivia : []
      - pieces : 0 elements
    ▿ trailingTrivia : spaces(1)
      ▿ pieces : 1 element
        ▿ 0 : spaces(1)
          - spaces : 1
    - tokenKind : SwiftSyntax.TokenKind.colon
  ▿ 5 : identifier("Int")
    - text : "Int"
    ▿ leadingTrivia : []
      - pieces : 0 elements
    ▿ trailingTrivia : []
      - pieces : 0 elements
    ▿ tokenKind : TokenKind
      - identifier : "Int"
  ▿ 6 : exclamationMark
    - text : "!"
    ▿ leadingTrivia : []
      - pieces : 0 elements
    ▿ trailingTrivia : []
      - pieces : 0 elements
    - tokenKind : SwiftSyntax.TokenKind.exclamationMark
```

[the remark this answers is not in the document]
adammcarter: Thanks for the explanation that makes a lot of sense!

Sources/SwiftSyntax/Identifier.swift
```diff
@@ -0,0 +1,31 @@
...
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+
+/// An abstraction for sanitized values on a token.
+public struct Identifier: Equatable, Sendable {
+  /// The sanitized `text` of a token.
+  public let name: String
```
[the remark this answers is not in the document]
ahoppen: I just chatted to @rintaro about this.

What we want is two types:
- An `@_spi(RawSyntax) public struct RawIdentifier` that stores the name of the identifier without keeping the `SyntaxArena` alive. This is intended to be used in performance-critical situations that can guarantee that the `SyntaxArena` stays alive and don’t want to pay the ref-counting overhead for it.
- A `public struct Identifier` that wraps `RawIdentifier` and also keeps the syntax arena alive using `RetainedSyntaxArena`. Identifier should then have a computed property `name: String` that returns a Unicode-normalized version of the identifier. `String` is probably what most clients choose to use but for uses in the compiler, we need to be byte accurate because the compiler considers `U+00E0` (à LATIN SMALL LETTER A WITH GRAVE) to be different than `U+0061 U+0300` (a LATIN SMALL LETTER A followed by ̀ COMBINING GRAVE ACCENT) while Swift’s `String` performs Unicode normalization and considers them the same (`print("\u{e0}" == "\u{61}\u{300}")` prints `true`)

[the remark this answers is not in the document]
adammcarter: Hey @ahoppen thanks for following up on this!

I'm looking in to the above and wanted to clarify whether the `RawIdentifier.name` needs to be a `SyntaxText` or a `RawSyntax`

I've pushed a WIP commit which covers the above and uses a `SyntaxText` which mostly works (except for the test `testRawIdentifier()` which I'll look in to after the conclusion of this conversation)

However on attempting to use `RawSyntax` type I'm getting a bit lost in the code from me being so new to this codebase.

Are you able to:
- clarify if we want `RawIdentifier.name` to be a `SyntaxText` or a `RawSyntax`
- look at my latest WIP commit and provide some input on my attempts so far?

Thanks!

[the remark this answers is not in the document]
ahoppen: `RawIdentifier.name` should be a `SyntaxText`. Sorry if that wasn’t clear from my last comment.

Sources/SwiftSyntax/Identifier.swift
ahoppen, line 1: Could you indent the file using 2 spaces and run swift-format on it? https://github.com/apple/swift-syntax/blob/main/CONTRIBUTING.md#formatting

Sources/SwiftSyntax/Identifier.swift
```diff
@@ -0,0 +1,56 @@
...
+        self.arena = RetainedSyntaxArena(syntaxArena)
+    }
+    
+    public static func == (lhs: Identifier, rhs: Identifier) -> Bool {
+        lhs.rawIdentifier == rhs.rawIdentifier
+    }
+    
+    public func hash(into hasher: inout Hasher) {
+        hasher.combine(rawIdentifier)
+    }
```
ahoppen, line 50: Don’t these implementations get automatically synthesized?

Sources/SwiftSyntax/Identifier.swift
```diff
@@ -0,0 +1,56 @@
...
+}
+
+@_spi(RawSyntax) 
+public struct RawIdentifier: Equatable, Hashable, Sendable {
+    public let name: SyntaxText
```
ahoppen, line 55: I think we should do the trimming of backticks in `RawIdentifier`. That way it’s possible to construct a `RawIdentifier` from a `RawSyntaxTokenView` that contains a token. Ie. I think it should have an initializer with the signature `init(_ raw: RawSyntaxTokenView)`.

Sources/SwiftSyntax/Identifier.swift
```diff
@@ -0,0 +1,56 @@
...
+        
+        var rawText = text.contains("`") ? text.trimmingCharacters(in: "`") : Substring(text)
+        
+        let syntaxArena = SyntaxArena()
+        
+        let name = rawText.withUTF8 {
+            syntaxArena.intern(
+                SyntaxText(buffer: SyntaxArenaAllocatedBufferPointer<UInt8>($0))
+            )
+        }
```
ahoppen, line 38: I don’t think we need to create a new arena and intern the string at all. If you pick up my suggestion of doing the trimming in `RawIdentifier`, you should be able to
- Get the `RawSyntaxTokenView.rawText`
- If the text has a leading and trailing backtick, slice off the first and last byte of the `SyntaxText` using a subscript
- Constructing a `SyntaxText` from the slice again using `SyntaxText.init(rebasing:)`

What’s conceptually happening then, is that the whole text including backticks is allocated memory in the `SyntaxArena` and `RawIdentifier` just references the slice of that text without the backticks without doing any more memory allocations.

---

Also, I think we don’t want to trim an arbitrary number of backticks at the front and back, but a single backtick at the front and a single backtick at the back if both exist.

Sources/SwiftSyntax/Identifier.swift
```diff
@@ -0,0 +1,56 @@
...
+public struct Identifier: Equatable, Hashable, Sendable {
+    /// The sanitized `text` of a token.
+    public var name: String {
+        String(syntaxText: rawIdentifier.name)
+    }
+    
+    @_spi(RawSyntax)
+    public let rawIdentifier: RawIdentifier
+    
+    let arena: RetainedSyntaxArena
```
ahoppen, line 23: I think the arena can even be private.

Tests/SwiftSyntaxTest/IdentifierTests.swift
```diff
@@ -0,0 +1,51 @@
...
+
+@_spi(RawSyntax) import SwiftSyntax
+import XCTest
+
+class IdentifierTests: XCTestCase {
+    public func testIdentifierInit() {
+        let someToken = TokenSyntax(stringLiteral: "someToken")
+        XCTAssertNotNil(Identifier(someToken))
+        
+        let nonIdentifierToken = DeclSyntax("let a = 1").firstToken(viewMode: .all)!
```
ahoppen, line 21: Could you use `XCTUnwrap` instead of the force unwrapping? That way test execution continues and doesn’t crash if `firstToken` should be `nil`.

Sources/SwiftSyntax/Identifier.swift
```diff
@@ -0,0 +1,56 @@
...
+
+/// An abstraction for sanitized values on a token.
+public struct Identifier: Equatable, Hashable, Sendable {
+    /// The sanitized `text` of a token.
+    public var name: String {
+        String(syntaxText: rawIdentifier.name)
+    }
+    
+    @_spi(RawSyntax)
+    public let rawIdentifier: RawIdentifier
```
ahoppen, line 21: I would just call this `public let raw: RawIdentifier` to match the naming scheme that `Syntax` have a `raw: RawSyntax` property.

ahoppen (commented):
Thank you! This is going in a great direction. It’s about the design that I had in mind.

plemarquand:
@adammcarter Are you planning on coming back to this? Do you mind if I put up a patch that addresses these comments?

plemarquand:
@adammcarter @ahoppen I took a stab at addressing the review comments.

Sources/SwiftSyntax/Identifier.swift
plemarquand: the `arena` doesn't conform so Hashable/Equatable.

Sources/SwiftSyntax/Identifier.swift
plemarquand: I think these comments have now been resolved, let me know what you think.

Sources/SwiftSyntax/Identifier.swift
```diff
@@ -0,0 +1,60 @@
...
+
+@_spi(RawSyntax)
+public struct RawIdentifier: Equatable, Hashable, Sendable {
+  public let name: SyntaxText
+
+  private let arena: RetainedSyntaxArena
+
+  @_spi(RawSyntax)
+  fileprivate init(_ raw: RawSyntaxTokenView) {
+    let backtick = SyntaxText(StaticString(stringLiteral: "`"))
```
ahoppen, line 40: Is the `StaticString` needed here? Couldn’t this just be the following? I would expect the type checker to infer the string literal to be a static string.

```suggestion
    let backtick = SyntaxText("`")
```

Sources/SwiftSyntax/Identifier.swift
```diff
@@ -0,0 +1,60 @@
...
+    let arena = SyntaxArena()
```
ahoppen, line 41: I don’t think we need to create an arena at all. The arena that contains the `RawSyntaxTokenView` already contains the name of the identifier, so there are no allocations to be made, retaining `raw.raw.arenaReference.retained` should be sufficient. Also, then you don’t need to intern any text into the arena.

Sources/SwiftSyntax/Identifier.swift
```diff
@@ -0,0 +1,60 @@
...
+
+    self.raw = RawIdentifier(token.tokenView)
+  }
+}
+
+@_spi(RawSyntax)
+public struct RawIdentifier: Equatable, Hashable, Sendable {
+  public let name: SyntaxText
+
+  private let arena: RetainedSyntaxArena
```
ahoppen, line 36: The `Raw*` syntax nodes in swift-syntax shouldn’t retain their arenas, `RawSyntaxData` for example, also just has a `SyntaxArenaRef` instead of a `RetainedSyntaxArena`. The idea is that the user of the raw syntax nodes guarantees that the arena stays alive and thus doesn’t need to pay reference counting traffic.

In this concrete example, I think `RawIdentifier` should just store a `SyntaxText` and not have any arena reference (since it’s not needed) and `Identifier` is the safe type on top that makes sure the arena doesn’t get deallocated by retaining it through a `RetainedSyntaxArena` reference.

Sources/SwiftSyntax/TokenSyntax.swift
```diff
@@ -153,6 +153,11 @@ public struct TokenSyntax: SyntaxProtocol, SyntaxHashable {
     }
   }
 
+  /// An identifier created from `self`.
+  public var identifier: Identifier? {
+    Identifier(self)
```
ahoppen, line 158: This should return `nil` if `self.tokenKind` is not `.identifier` or `.dollarIdentifier`.

Sources/SwiftSyntax/Identifier.swift
```diff
@@ -0,0 +1,60 @@
...
+//
+// Copyright (c) 2014 - 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// An abstraction for sanitized values on a token.
```
ahoppen, line 13: How about

```suggestion
/// A canonicalized representation of an identifier that strips away backticks.
```

ahoppen (commented):
Thanks for picking this up @plemarquand! I’ve got a few comments inline. Could you also add an entry to the release notes?

Sources/SwiftSyntax/Identifier.swift
plemarquand: The `testRawIdentifier` test references the `RawIdentifier`, and by moving the arena reference up to the Identifier its no longer retained and the test fails. Does testRawIdentifier need refactoring as well? I'm not sure what RawIdentifier can be used for without retaining the string that backs it?

plemarquand: OK, as long as thats expected with SPI types then this works. I've updated the PR.

bnbarham: It's expected for the raw syntax nodes, not really to do with SPI or not (it's just that they are SPI as they aren't the intended user-facing API).

bnbarham:
@swift-ci please test

ahoppen (approved):
Nice. Looks great!

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/SwiftSyntax/Identifier.swift
ahoppen, line 1: Could you add Identifier.swift to [CMakeLists.txt](https://github.com/apple/swift-syntax/blob/main/Sources/SwiftSyntax/CMakeLists.txt). swift-syntax is built using CMake when it’s linked into the compiler.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

PR #2578, opened 2024-03-29
DRAFT - Add `Trivia.commentValues` for sanitised comments

adammcarter:


Tests/SwiftSyntaxTest/TriviaTests.swift
```diff
@@ -69,4 +69,325 @@ class TriviaTests: XCTestCase {
...
+    XCTAssertEqual(
+      Trivia(pieces: [
+        .docBlockComment(
+          """
+          /**
+          *  Some doc block comment
+          *  // spread on many lines
+          *  with a line comment
+          */
+          """
```
ahoppen, line 347: I think it’s quite common for the `*` to be indented by a single space to align with the first `*` of `/*`. Could you also add a test case for something like

```swift
          /**
           *  Some doc block comment
           *  with a line comment
          */
          """
```

Sources/SwiftSyntax/Trivia.swift
```diff
@@ -42,6 +42,68 @@ public struct Trivia: Sendable {
...
+      case .blockComment(let text), .docBlockComment(let text):
+        let text = text.trimmingCharacters(in: "\n")
+
+        foundStartOfCodeBlock = text.hasPrefix("/*")
+        foundEndOfCodeBlock = text.hasSuffix("*/")
+
+        let sanitized =
+          text
+          .split(separator: "\n")
+          .map { $0.trimmingAnyCharacters(in: "/*").trimmingAnyCharacters(in: " ") }
```
ahoppen, line 67: I think this would also trim any slashes at the start of a line. And I don’t think we want that. Eg. in

```swift
/* 
/abc
*/
```

I would expect the comment value to be `/`.

I would just trim the following:
- At the start of the comment: `/*` or `/**`
- Inside the block comment: Any number of spaces followed by an optional `*`
- At the end of the comment: `*/`

Sources/SwiftSyntax/Trivia.swift
```diff
@@ -42,6 +42,68 @@ public struct Trivia: Sendable {
...
+          .filter { !$0.isEmpty }
```
ahoppen, line 68: Does this drop empty lines in the comment? Eg. in the following? Because I don’t think we want to. For example if you interpret the comment as markdown, empty lines have meaning.

```swift
/*
Paragraph 1

Paragraph 2
*/
```

Sources/SwiftSyntax/Trivia.swift
```diff
@@ -42,6 +42,68 @@ public struct Trivia: Sendable {
...
+          .joined(separator: " ")
```
ahoppen, line 69: Similarly here, I don’t think we should concatenate lines using a space. Again, if interpreting a doc comment as markdown, the newlines have meaning.

Tests/SwiftSyntaxTest/TriviaTests.swift
```diff
@@ -69,4 +69,325 @@ class TriviaTests: XCTestCase {
     XCTAssertNotEqual(TriviaPiece.unexpectedText("e"), .unexpectedText("f"))
     XCTAssertNotEqual(TriviaPiece.unexpectedText("e"), .lineComment("e"))
   }
+
+  func testTriviaCommentValues() {
```
ahoppen, line 73: I think it’s fine if we tailor `commentValue` to the kind of trivia that the parser produces. If we do that we can construct the `Trivia` from a string literal and have something like the following:

```swift
func assertCommentValue(
  _ trivia: Trivia,
  commentValue: String,
  file: StaticString = #filePath,
  line: UInt = #line
) {
  XCTAssertEqual(trivia.commentValue, commentValue, file: file, line: line)
}

func testExample() {
  assertCommentValue(
    "// line comment",
    commentValue: "line comment"
  )
}

```

ahoppen (commented):
Thanks for setting up the PR @adammcarter. I have some initial review comments inline.

PR #2579, opened 2024-03-30
Parse types with inverted constraints in expressions

DougGregor:
Lookahead to determine when we are parsing a type within an expression context was failing to consider inverted constraints. Make sure we look past the `~` to an identifier.


DougGregor:
@swift-ci please test

PR #2581, opened 2024-04-01
[Diagnostics] Stop using unicode line characters

rintaro:
ASCII characters are more reliable.

```
┆    → :
│    → |
├─   → |-
╰─   → `-
╭───   +---
│    → |
╰───   +---
```

### DIFF
--- a/Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
+++ b/Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
@@ -268,13 +268,13 @@ public struct DiagnosticsFormatter {
       // line numbers should be right aligned
       let lineNumberString = String(lineNumber)
       let leadingSpaces = String(repeating: " ", count: maxNumberOfDigits - lineNumberString.count)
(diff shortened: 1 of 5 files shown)


rintaro:
@swift-ci Please test

DougGregor (approved):

bnbarham:
This makes me a little sad, but given the issues we've had across various platforms... IMO it seems like the most reasonable route to take for now.

bnbarham (approved):

rintaro:
https://github.com/apple/swift/pull/72754
@swift-ci Please test

Matejkob:
> This makes me a little sad, but given the issues we've had across various platforms... IMO it seems like the most reasonable route to take for now.

Why not use different annotation characters based on the context?

DougGregor:
> Why not use different annotation characters based on the context?

Unfortunately, we don't really have enough context. We could try to detect when we're in a terminal, but some terminals handle Unicode properly and some do not. And when we're emitting diagnostics to a file, we have no notion of whether the tool reading that file will handle Unicode. So at best, we'd be able to introduce the Unicode in a few places where we know it would work... but then folks would be surprised at the output differing in different environments.

PR #2583, opened 2024-04-02
[WIP] Symbol name

### LINKED ISSUE #2488: FunctionDecl and similar should have method to get name as "hello(name:surname:)"
### Description

_No response_

Matejkob:
WIP

Resolves #2488


Tests/SwiftSyntaxTest/SymbolNameTests.swift
```diff
@@ -0,0 +1,210 @@
...
+  }
+
+  func testFunctionNameWithParameterNameInBackticks() {
+    assertSymbolName(ofFunction: "func name(`one`: Int) {}", expecting: "name(one:)")
+  }
+
+  // FIXME: Does this make sense?
+  func testOperatorFunctionName() {
+    assertSymbolName(ofFunction: "func == (one: Int, two: Int)", expecting: "==(one:two:)")
+  }
```
Matejkob, line 75: Should the operator function have a `symbolName`?

Tests/SwiftSyntaxTest/SymbolNameTests.swift
```diff
@@ -0,0 +1,210 @@
...
+    assertSymbolName(ofFunction: "func name(_ two: Int, _ four: Int) {}", expecting: "name(_:_:)")
+  }
+
+  func testFunctionNameWithBackticks() {
+    assertSymbolName(ofFunction: "func `name`() {}", expecting: "name()")
+  }
+
+  func testFunctionNameWithParameterNameInBackticks() {
+    assertSymbolName(ofFunction: "func name(`one`: Int) {}", expecting: "name(one:)")
+  }
```
Matejkob, line 70: This test is correct, isn't it?

Tests/SwiftSyntaxTest/SymbolNameTests.swift
ktoso: Yeah I think so.

Tests/SwiftSyntaxTest/SymbolNameTests.swift
ktoso: I think that's right but @ahoppen may want to confirm

ahoppen: I think we need to include the backticks in the function name. Otherwise the function name of

```swift
struct Foo {
  func foo(`inout`: Int) {}
}
```

is `Foo.foo(inout:)` and that doesn’t compile.

Also, in practice there is very little reason for anyone to add backticks to argument labels and I think it’s easier to just leave the backticks as the user spelled them in the function name.

Tests/SwiftSyntaxTest/SymbolNameTests.swift
ahoppen: Yes, and I think the test makes sense.

Tests/SwiftSyntaxTest/SymbolNameTests.swift
Matejkob: Given that Xcode's "Copy Symbol Name" functionality removes the backticks, it used to seem reasonable to me to align with that behavior for consistency. However, I hadn't considered that specific scenario you have put above.

Matejkob (commented):
Which other nodes should have a symbolName property?

1. enum, protocol, struct, class, actor?
2. typealias?
3. property - this one for sure.
4. deinit ?
5. macro ?
6. operator ?

ahoppen:
I think it only makes sense to offer it for nodes that have a `FunctionParameterClauseSyntax` or `EnumCaseParameterClauseSyntax` and thus need handling of argument labels. That is
- `FunctionDeclSyntax`
- `InitializerDeclSyntax`
- `MacroDeclSyntax`
- `SubscriptDeclSyntax`
- `EnumCaseDeclSyntax`

Sources/SwiftSyntax/SymbolName.swift
```diff
@@ -0,0 +1,142 @@
...
+  /// The symbol name is a string representation of the enum case name along with its associated value labels (if any).
+  /// For example, an enum case `case foo(bar: Int)` will have the symbol name `foo(bar:)`.
+  public var symbolName: String {
+    let caseName = name.text
+
+    guard let associatedValue = parameterClause else {
+      return caseName
+    }
+
+    let argumentNames = associatedValue.parameters
```
ahoppen, line 65: Let’s call them `argumentLabels` or `parameterLabels`. A short distinction between arguments and parameters: Parameters are on the function declaration side and declare the things that can be passed to the functions. Arguments are on the call side and specify the values that can be passed to those parameters. The labels are equivalent between arguments and parameters, so we can use either here.

Sources/SwiftSyntax/SymbolName.swift
```diff
@@ -0,0 +1,142 @@
...
+}
+
+extension MacroDeclSyntax {
+  /// The symbol name of the macro declaration.
+  ///
+  /// The symbol name is a string representation of the macro name along with its parameter labels.
+  /// For example, a macro `macro greet(name: String)` will have the symbol name `greet(name:)`.
+  public var symbolName: String {
+    SwiftSyntax.symbolName(
+      fromBaseName: name.trimmedDescription,
```
ahoppen, line 86: Same here: I’d use `name.text`.

Sources/SwiftSyntax/SymbolName.swift
```diff
@@ -0,0 +1,142 @@
...
+  /// If the parameter has two names (e.g., `external internal: Int`), the first name is considered the argument label.
+  /// If the parameter has only one name and it's not a subscript parameter, it is considered the argument label.
+  fileprivate var argumentName: TokenSyntax? {
+    // If we have two names, the first one is the argument label
+    if secondName != nil {
+      return firstName.asIdentifierToken
+    }
+
+    // If we have only one name, it might be an argument label.
+    if let superParent = parent?.parent?.parent, superParent.is(SubscriptDeclSyntax.self) {
```
ahoppen, line 124: Instead of walking the tree up again (which might be fragile if for some reasons the `FunctionParameterSyntax` would occur eg. in the subscript body and just happens to be 3 levels down from the `SubscriptDeclSyntax` node – not that this might happen today but could theoretically in the future), I’d just add a parameter `isForSubscript`.

And then I’d probably flip the implementation around and have a `if isForSubscript` at the top level. I think that’s easier to read.

Sources/SwiftSyntax/SymbolName.swift
```diff
@@ -0,0 +1,142 @@
...
+  }
+}
+
+extension TokenSyntax {
+  /// Converts the token to an identifier token if it represents an identifier.
+  ///
+  /// - Returns: The trimmed token if it is an identifier or dollar identifier, otherwise `nil`.
+  fileprivate var asIdentifierToken: TokenSyntax? {
+    switch tokenKind {
+    case .identifier, .dollarIdentifier: return self.trimmed
```
ahoppen, line 138: I don’t think dollar identifiers can be parameter labels and so I don’t think we need to consider them.

---

I would use `self.text` again. It’s more performant because `self.trimmed` needs to create a copy of the tokens without trivia – and also clearer.

Tests/SwiftSyntaxTest/SymbolNameTests.swift
```diff
@@ -0,0 +1,358 @@
...
+
+  func testFunctionNameWithBackticks() {
+    assertSymbolName(ofFunction: "func `name`() {}", expecting: "`name`()")
+  }
+
+  func testFunctionNameWithParameterNameInBackticks() {
+    assertSymbolName(ofFunction: "func name(`one`: Int) {}", expecting: "name(`one`:)")
+  }
+
+  // FIXME: Does this make sense?
```
ahoppen, line 72: I think you can remove this comment.

Tests/SwiftSyntaxTest/SymbolNameTests.swift
```diff
@@ -0,0 +1,358 @@
...
+    file: StaticString = #filePath,
+    line: UInt = #line
+  ) {
+    guard let functionDecl = function.as(FunctionDeclSyntax.self) else {
+      XCTFail("Expected function declaration not found.", file: file, line: line)
+      return
+    }
+
+    XCTAssertEqual(functionDecl.symbolName, expectedSymbolName, file: file, line: line)
+  }
```
ahoppen, line 89: I’d prefer to not interleave these `assert` functions with the test cases. I’d make them top-level functions at the start or the end of the file.

Tests/SwiftSyntaxTest/SymbolNameTests.swift
```diff
@@ -0,0 +1,358 @@
...
+  func testEnumCaseElementNameWithBackticks() {
+    assertSymbolName(ofEnumCaseElementFromEnumCase: "case `i`", expecting: "`i`")
+  }
+
+  func testEnumCaseElementWithAssociatedValueNameInBackticks() {
+    assertSymbolName(ofEnumCaseElementFromEnumCase: "case j(`k`: Int)", expecting: "j(`k`:)")
+  }
+
+  private func assertSymbolName(
+    ofEnumCaseElementFromEnumCase enumCaseDeclaration: DeclSyntax,
```
ahoppen, line 245: Let’s just call it `ofEnumCase` to make it a little shorter.

Tests/SwiftSyntaxTest/SymbolNameTests.swift
```diff
@@ -0,0 +1,358 @@
...
+      return
+    }
+    XCTAssertEqual(enumCaseElement.symbolName, expectedSymbolName, file: file, line: line)
+  }
+
+  // MARK: - Macro
+
+  func testMacroDeclWithoutParameters() {
+    assertSymbolName(
+      ofMacroDecl: ##"macro myMacro() = #externalMacro"##,
```
ahoppen, line 263: I don’t think we need the `= #externalMacro` part for the test. 

And I think normal string literals without `##` should be sufficient.

ahoppen (commented):
Thank you. Looks great, I’m just being nitpicky again.

Sources/SwiftSyntax/SymbolName.swift
Matejkob: `argumentLabels` seems to be the best fit – done.

Sources/SwiftSyntax/SymbolName.swift
Matejkob: Good point; done

Sources/SwiftSyntax/SymbolName.swift
Matejkob: I believe it would make sense to consider `dollarIdentifier` if we added `symbolName` for variables. But since it's to complex to provide such an API we can remove it – done.

Tests/SwiftSyntaxTest/SymbolNameTests.swift
Matejkob: Yup, thanks. I've missed it 😅

Tests/SwiftSyntaxTest/SymbolNameTests.swift
Matejkob: Works for me, moved.

Matejkob:
Thanks for the review Alex 🫶

ahoppen (approved):

bnbarham (approved):

ktoso:
@swift-ci please smoke test

Release Notes/601.md
```diff
@@ -2,6 +2,11 @@
 
 ## New APIs
 
+- `FunctionDecl`, `InitializerDecl`, `SubscriptDecl`, `EnumCaseElement`, and `MacroDecl` gained a new computed property: `symbolName`
+  - Description: The `symbolName` property provides a string representation of the declaration's name along with its parameter labels. For example, a function `func greet(name: String)` will have the symbol name `greet(name:)`.
```
hamishknight, line 6: Nit: These are called argument labels in [TSPL](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/declarations#Parameter-Names), and we have e.g `SyntaxClassification.argumentLabel`, I think we probably ought to be consistent.

```suggestion
  - Description: The `symbolName` property provides a string representation of the declaration's name along with its argument labels. For example, a function `func greet(name: String)` will have the symbol name `greet(name:)`.
```

Sources/SwiftSyntax/SymbolName.swift
```diff
@@ -0,0 +1,139 @@
...
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+extension FunctionDeclSyntax {
+  /// The symbol name of the function declaration.
+  ///
+  /// The symbol name is a string representation of the function name along with its parameter labels.
```
hamishknight, line 16:
```suggestion
  /// The symbol name is a string representation of the function name along with its argument labels.
```

Sources/SwiftSyntax/SymbolName.swift
```diff
@@ -0,0 +1,139 @@
...
+      fromBaseName: name.text,
+      parameters: signature.parameterClause
+    )
+  }
+}
+
+extension InitializerDeclSyntax {
+  /// The symbol name of the initializer declaration.
+  ///
+  /// The symbol name is a string representation of the initializer along with its parameter labels.
```
hamishknight, line 29:
```suggestion
  /// The symbol name is a string representation of the initializer along with its argument labels.
```

Sources/SwiftSyntax/SymbolName.swift
```diff
@@ -0,0 +1,139 @@
...
+      fromBaseName: initKeyword.text,
+      parameters: signature.parameterClause
+    )
+  }
+}
+
+extension SubscriptDeclSyntax {
+  /// The symbol name of the subscript declaration.
+  ///
+  /// The symbol name is a string representation of the subscript along with its parameter labels.
```
hamishknight, line 42:
```suggestion
  /// The symbol name is a string representation of the subscript along with its argument labels.
```

Sources/SwiftSyntax/SymbolName.swift
```diff
@@ -0,0 +1,139 @@
...
+      .joined()
+
+    return "\(caseName)(\(argumentLabels))"
+  }
+}
+
+extension MacroDeclSyntax {
+  /// The symbol name of the macro declaration.
+  ///
+  /// The symbol name is a string representation of the macro name along with its parameter labels.
```
hamishknight, line 82:
```suggestion
  /// The symbol name is a string representation of the macro name along with its argument labels.
```

Sources/SwiftSyntax/SymbolName.swift
```diff
@@ -0,0 +1,139 @@
...
+  /// For example, a macro `macro greet(name: String)` will have the symbol name `greet(name:)`.
+  public var symbolName: String {
+    SwiftSyntax.symbolName(
+      fromBaseName: name.text,
+      parameters: signature.parameterClause
+    )
+  }
+}
+
+/// Generates the symbol name by combining the base name and parameter labels.
```
hamishknight, line 92:
```suggestion
/// Generates the symbol name by combining the base name and argument labels.
```

hamishknight (approved):
A small nit over the naming in the documentation, but otherwise this LGTM

rintaro:
This resembles [`DeclName` in C++ compiler](https://github.com/apple/swift/blob/7b8f76e3c404638d9e606b4906baf00b316ce74b/include/swift/AST/Identifier.h#L434-L435). 
I _think_ we should also make it a struct, like:
```swift
struct SymbolBaseName {
  enum Kind { case normal, subscript, initializer, deinitializer }
  var asString: String {get}
  var kind: Bool {get}
  var isOperator: Bool {get}
}
struct SymbolName {  
  var asString: String {get}
  var baseName: SymbolBaseName {get}
  var parameterNames: some Collection<String> {get}
}
```
WDYT?

ahoppen:
That is a great idea, @rintaro.

PR #2594, opened 2024-04-10
Make `SyntaxIndexInTree` public and serializable into a `UInt64`

ahoppen:
If you have two copies of the same syntax tree within two processes, this allows us to communicate the identifier of a node in one process the the equivalent node in the other process.

rdar://126221355

PR #2603, opened 2024-04-12
[6.0] Make `SyntaxIndexInTree` public and serializable into a `UInt64`

ahoppen:
* **Explanation**: If you have two copies of the same syntax tree within two processes, this allows us to communicate the identifier of a node in one process the the equivalent node in the other process.
* **Scope**: Adds new API / makes existing API public
* **Risk**: Low, almost purely additive
* **Issue**: rdar://126221355
* **Reviewer**:  @rintaro and @bnbarham on https://github.com/apple/swift-syntax/pull/2594


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #2588, opened 2024-04-03
Deprecate `ByteSourceRange` in favor of `Range<AbsolutePosition>`

ahoppen:
While being a little more verbose, this has a few advantages:
- We only have a single type to represent ranges in SwiftSyntax instead of `Range<AbsolutePosition>` and `ByteSourceRange`, both of which we needed to use in sourcekit-lsp
- Unifying the convenience functions on the two results in a single type that has all the convenience functions, instead of spreading them to two distinct sets
- The use of `AbsolutePosition` and `SourceLength` makes type-system guarantees that these are UTF-8 byte positions / length, making it harder to accidentally add eg. UTF-16 lengths with UTF-8 lengths.

rdar://125624626

PR #2604, opened 2024-04-13
Merge `IncrementalEdit` and`SourceEdit`

### LINKED ISSUE #2532: Merge `IncrementalEdit` and `SourceEdit`
After https://github.com/apple/swift-syntax/pull/2527, `IncrementalEdit` and `SourceEdit` are more or less equivalent types. Both store a replacement range and a replacement. The only meaningful difference is that `SourceEdit` has `String` as a replacement and `IncrementalEdit` has an array of bytes as a replacement and can thus also represent invalid UTF-8. We should merge them, ideally into a type that is byte-based (because it’s more versatile) but that offers `String`-based accessors.

The best outcome would be if `SourceEdit` could be the type that’s used to represent the edits, `IncrementalEdit` becomes a deprecated `typealias` to `SourceEdit` and `SourceEdit` gains (possibly deprecated) members for all the members that `IncrementalEdit` had, for API compatibility.

kimdv:
Resolves #2532

I'm not sure if this what you envisioned.

### DIFF
--- a/Sources/SwiftParser/IncrementalParseTransition.swift
+++ b/Sources/SwiftParser/IncrementalParseTransition.swift
@@ -299,11 +299,11 @@ public struct ConcurrentEdits: Sendable {
 
   /// The raw concurrent edits. Are guaranteed to satisfy the requirements
   /// stated above.
-  public let edits: [IncrementalEdit]
+  public let edits: [SourceEdit]
 
   /// Initialize this struct from edits that are already in a concurrent form
   /// and are guaranteed to satisfy the requirements posed above.
-  public init(concurrent: [IncrementalEdit]) throws {
+  public init(concurrent: [SourceEdit]) throws {
     if !Self.isValidConcurrentEditArray(concurrent) {
       throw ConcurrentEditsError.editsNotConcurrent
     }
@@ -319,7 +319,7 @@ public struct ConcurrentEdits: Sendable {
   ///  - insert 'z' at offset 2
   ///  to '012345' results in 'xyz012345'.
 
-  public init(fromSequential sequentialEdits: [IncrementalEdit]) {
+  public init(fromSequential sequentialEdits: [SourceEdit]) {
     do {
       try self.init(concurrent: Self.translateSequentialEditsToConcurrentEdits(sequentialEdits))
     } catch {
@@ -332,7 +332,7 @@ public struct ConcurrentEdits: Sendable {
   /// Construct a concurrent edits struct from a single edit. For a single edit,
   /// there is no differentiation between being it being applied concurrently
   /// or sequentially.
-  public init(_ single: IncrementalEdit) {
+  public init(_ single: SourceEdit) {
     do {
       try self.init(concurrent: [single])
     } catch {
@@ -341,9 +341,9 @@ public struct ConcurrentEdits: Sendable {
   }
 
   private static func translateSequentialEditsToConcurrentEdits(
-    _ edits: [IncrementalEdit]
-  ) -> [IncrementalEdit] {
-    var concurrentEdits: [IncrementalEdit] = []
+    _ edits: [SourceEdit]
+  ) -> [SourceEdit] {
+    var concurrentEdits: [SourceEdit] = []
     for editToAdd in edits {
       var editToAdd = editToAdd
       var editIndicesMergedWithNewEdit: [Int] = []
@@ -353,14 +353,14 @@ public struct ConcurrentEdits: Sendable {
             existingEdit.replacementRange.clamped(to: editToAdd.range).length
           let replacement: [UInt8]
           replacement =
-            existingEdit.replacement.prefix(
+            existingEdit.replacementBytes.prefix(
               max(0, editToAdd.range.lowerBound.utf8Offset - existingEdit.replacementRange.lowerBound.utf8Offset)
             )
-            + editToAdd.replacement
-            + existingEdit.replacement.suffix(
+            + editToAdd.replacementBytes
+            + existingEdit.replacementBytes.suffix(
               max(0, existingEdit.replacementRange.upperBound.utf8Offset - editToAdd.range.upperBound.utf8Offset)
             )
-          editToAdd = IncrementalEdit(
+          editToAdd = SourceEdit(
             range: Range(
               position: Swift.min(existingEdit.range.lowerBound, editToAdd.range.lowerBound),
               length: existingEdit.range.length + editToAdd.range.length - intersectionLength
@@ -369,7 +369,7 @@ public struct ConcurrentEdits: Sendable {
           )
           editIndicesMergedWithNewEdit.append(index)
         } else if existingEdit.range.lowerBound < editToAdd.range.upperBound {
-          editToAdd = IncrementalEdit(
+          editToAdd = SourceEdit(
             range: Range(
               position: editToAdd.range.lowerBound + existingEdit.range.length - existingEdit.replacementLength,
               length: editToAdd.range.length
@@ -392,7 +392,7 @@ public struct ConcurrentEdits: Sendable {
     return concurrentEdits
   }
 
-  private static func isValidConcurrentEditArray(_ edits: [IncrementalEdit]) -> Bool {
+  private static func isValidConcurrentEditArray(_ edits: [SourceEdit]) -> Bool {
     // Not quite sure if we should disallow creating an `IncrementalParseTransition`
     // object without edits but there doesn't seem to be much benefit if we do,
     // and there are 'lit' tests that want to test incremental re-parsing without edits.
@@ -412,7 +412,7 @@ public struct ConcurrentEdits: Sendable {
   }
 
   /// **Public for testing purposes only**
-  public static func _isValidConcurrentEditArray(_ edits: [IncrementalEdit]) -> Bool {
+  public static func _isValidConcurrentEditArray(_ edits: [SourceEdit]) -> Bool {
     return isValidConcurrentEditArray(edits)
   }
 }
--- a/Sources/SwiftSyntax/SourceEdit.swift
+++ b/Sources/SwiftSyntax/SourceEdit.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the Swift.org open source project
 //
-// Copyright (c) 2014 - 2023 Apple Inc. and the Swift project authors
+// Copyright (c) 2014 - 2024 Apple Inc. and the Swift project authors
 // Licensed under Apache License v2.0 with Runtime Library Exception
 //
 // See https://swift.org/LICENSE.txt for license information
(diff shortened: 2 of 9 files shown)


Sources/SwiftSyntax/ByteSourceRange.swift
ahoppen, line 1: Let’s leave `ByteSourceRange` in its current location because I’m deprecating it anyway in favor of `Range<AbsolutePosition>` in https://github.com/apple/swift-syntax/pull/2588

Sources/SwiftSyntax/SourceEdit.swift
```diff
@@ -13,24 +13,68 @@
...
-  public let range: Range<AbsolutePosition>
-  /// The text to replace the original range with. Empty for a deletion.
-  public let replacement: String
-
-  /// Length of the original source range that this edit applies to. Zero if
-  /// this is an addition.
-  public var length: SourceLength {
-    return SourceLength(utf8Length: range.upperBound.utf8Offset - range.lowerBound.utf8Offset)
+  /// The byte range of the original source buffer that the edit applies to.
+  public let range: ByteSourceRange
```
ahoppen, line 17: I’d prefer to use `Range<AbsolutePosition>` here because I’m deprecating `ByteSourceRange` in https://github.com/apple/swift-syntax/pull/2588

Sources/SwiftSyntax/SourceEdit.swift
```diff
@@ -13,24 +13,68 @@
...
+
+  /// The UTF-8 bytes that should be inserted as part of the edit
+  public let replacement: [UInt8]
+
+  /// A string representation of the replacement
+  public var replacementText: String { return String(decoding: replacement, as: UTF8.self) }
```
ahoppen, line 23: To not make this a source-breaking change, this would need to continue being called `replacement`. I’d call the `[UInt8]` version `replacementBytes`.

We don’t need to be API-compatible with `IncrementalEdit.replacement` because that property only gets introduced in swift-syntax 600.

Sources/SwiftSyntax/SourceEdit.swift
```diff
@@ -13,24 +13,68 @@
...
+
+  /// The length of the edit replacement in UTF8 bytes.
+  public var replacementLength: Int { replacement.count }
+
+  public var offset: Int { return range.offset }
+
+  public var length: Int { return range.length }
+
+  public var endOffset: Int { return range.endOffset }
```
ahoppen, line 32: Can we mark these properties as deprecated?

Sources/SwiftSyntax/SourceEdit.swift
```diff
@@ -13,24 +13,68 @@
...
+  }
+
+  public func intersectsOrTouchesRange(_ other: ByteSourceRange) -> Bool {
+    return self.range.intersectsOrTouches(other)
+  }
+
+  public func intersectsRange(_ other: ByteSourceRange) -> Bool {
+    return self.range.intersects(other)
+  }
+
```
ahoppen, line 77: I would also deprecate these.

Sources/SwiftSyntax/SourceEdit.swift
```diff
@@ -13,24 +13,68 @@
...
+  public init(offset: Int, length: Int, replacement: [UInt8]) {
+    self.range = ByteSourceRange(offset: offset, length: length)
     self.replacement = replacement
   }
 
+  public init(offset: Int, length: Int, replacement: String) {
+    self.init(offset: offset, length: length, replacement: Array(replacement.utf8))
+  }
```
ahoppen, line 68: I would like to deprecate the initializers that take an `Int` because it’s very easy to mix up UTF-8, UTF-16 and character offsets into a source file. That’s why I prefer to use `AbsolutePosition` and `SourceLength`.

ahoppen (commented):
Thank you @kimdv!

A few comments about deprecations I think we could perform. If those are too wide-reaching, I think it would make sense to do them in a follow-up PR.

Sources/SwiftSyntax/SourceEdit.swift
kimdv: What text could we use?

kimdv:
@ahoppen do you mind do a quick review? 
Just to ensure I'm on the right path before I start fixes deprecation warnings. There is a few 😅

Sources/SwiftParser/IncrementalParseTransition.swift
```diff
@@ -172,15 +172,15 @@ struct IncrementalParseLookup {
...
     }
 
-    let nodeAffectRange = ByteSourceRange(
-      offset: node.position.utf8Offset,
-      length: nodeAffectRangeLength
-    )
+    let nodeAffectRange =
+      AbsolutePosition(
+        utf8Offset: node.position.utf8Offset
+      )..<AbsolutePosition(utf8Offset: node.position.utf8Offset + nodeAffectRangeLength)
```
ahoppen, line 178: I think we can use `Range<AbsolutePosition>(start: node.position.utf8Offset, length: nodeAffectRangeLength)` here (or it’s called something like that).

Sources/SwiftParser/IncrementalParseTransition.swift
```diff
@@ -345,29 +345,31 @@ public struct ConcurrentEdits: Sendable {
...
+  ) -> [SourceEdit] {
+    var concurrentEdits: [SourceEdit] = []
     for editToAdd in edits {
       var editToAdd = editToAdd
       var editIndicesMergedWithNewEdit: [Int] = []
       for (index, existingEdit) in concurrentEdits.enumerated() {
-        if existingEdit.replacementRange.intersectsOrTouches(editToAdd.range) {
-          let intersectionLength =
-            existingEdit.replacementRange.intersected(editToAdd.range).length
+        if existingEdit.range.intersectsOrTouches(editToAdd.range),
```
ahoppen, line 355: I think `range` is not the same as `replacementRange`, so this is a change in behavior, right? 

Same in the line below.

Sources/SwiftParser/IncrementalParseTransition.swift
```diff
@@ -345,29 +345,31 @@ public struct ConcurrentEdits: Sendable {
...
-          editToAdd = IncrementalEdit(
+            existingEdit.replacementBytes.prefix(max(0, editToAdd.offset - existingEdit.replacementRange.offset))
+            + editToAdd.replacementBytes
+            + existingEdit.replacementBytes.suffix(
+              max(0, existingEdit.replacementRange.endOffset - editToAdd.endOffset)
+            )
+          editToAdd = SourceEdit(
             offset: Swift.min(existingEdit.offset, editToAdd.offset),
-            length: existingEdit.length + editToAdd.length - intersectionLength,
+            length: existingEdit.length + editToAdd.length - intersectionLength.utf8Length,
```
ahoppen, line 367: Isn’t `intersectionLength` a `SourceLength` like the reset? So why do we need to go `.utf8Length` here?

Sources/SwiftSyntax/SourceEdit.swift
```diff
@@ -13,22 +13,74 @@
...
-  public var length: SourceLength {
-    return SourceLength(utf8Length: range.upperBound.utf8Offset - range.lowerBound.utf8Offset)
+  /// The UTF-8 bytes that should be inserted as part of the edit
+  public let replacementBytes: [UInt8]
+
+  /// A string representation of the replacement
+  public var replacement: String { return String(decoding: replacementBytes, as: UTF8.self) }
+
+  /// The length of the edit replacement in UTF8 bytes.
+  public var replacementLength: Int { replacement.utf8.count }
```
ahoppen, line 26: I would prefer to make this a `SourceLength` as well, even if it is API-breaking.

Sources/SwiftSyntax/SourceEdit.swift
```diff
@@ -13,22 +13,74 @@
...
+
+  @available(*, deprecated, message: "Use range instead")
```
ahoppen, line 28:
```suggestion
  @available(*, deprecated, renamed: "range.lowerBound.utf8Offset")
```

And please double check that my suggestion is actually correct.

Sources/SwiftSyntax/SourceEdit.swift
```diff
@@ -13,22 +13,74 @@
...
+  public var offset: Int { return range.lowerBound.utf8Offset }
+
+  @available(*, deprecated, message: "Use replacementLength instead")
+  public var length: Int { return range.upperBound.utf8Offset - range.lowerBound.utf8Offset }
+
+  @available(*, deprecated, message: "Use range instead")
```
ahoppen, line 34:
```suggestion
  @available(*, deprecated, renamed: "range.upperBound.utf8Offset")
```

Again, please check that the suggestion actually works.

Sources/SwiftSyntax/SourceEdit.swift
```diff
@@ -13,22 +13,74 @@
...
+  public var endOffset: Int { return range.upperBound.utf8Offset }
+
+  /// After the edit has been applied the range of the replacement text.
+  @available(*, deprecated, message: "Use range instead")
+  public var replacementRange: ByteSourceRange {
```
ahoppen, line 39: `range` and `replacementRange` are not the same. Once has the length of the replaced text and one has the length of the text that is being inserted. So we need both.

ahoppen (commented):
Looks good to me. I think it would be good to gather a list of all the deprecations + API-incompatible changes as well. And once this is finalized, I’d like to do another double-check about which API is broken and which is deprecated to make sure it matches the release notes.

ahoppen (dismissed):
Thank you. Looks good 👍🏽

bnbarham (approved):

ahoppen:
Sorry for changing my mind, @kimdv but I think it would be safer to not include this in 6.0. We are approaching the release and maybe changing/deprecating API that’s not strictly necessary is too big a risk for its value.

Sources/SwiftSyntax/SourceEdit.swift
```diff
@@ -13,22 +13,75 @@
...
-  public var length: SourceLength {
-    return SourceLength(utf8Length: range.upperBound.utf8Offset - range.lowerBound.utf8Offset)
+  /// The UTF-8 bytes that should be inserted as part of the edit
+  public let replacementBytes: [UInt8]
+
+  /// A string representation of the replacement
+  public var replacement: String { return String(decoding: replacementBytes, as: UTF8.self) }
+
+  /// The length of the edit replacement in UTF8 bytes.
+  public var replacementLength: SourceLength { return SourceLength(utf8Length: replacement.utf8.count) }
```
hamishknight, line 26: Should this return `replacementBytes.count`? I guess the question is for an invalid replacement such as `[0xFF]`, should this return 1 or 3 (the number of code units for `�`)? cc @ahoppen

hamishknight (approved):
LGTM, just one question about `replacementLength`

kimdv:
> Sorry for changing my mind, @kimdv but I think it would be safer to not include this in 6.0. Changing/deprecating API that’s not strictly necessary is too big a risk for its value.

@ahoppen I changed to 601 now

kimdv:
@swift-ci please test

Sources/SwiftSyntax/SourceEdit.swift
ahoppen: `replacementBytes.count` seems like the correct choice to me

kimdv: Fixed :D

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test

kimdv:
https://github.com/apple/sourcekit-lsp/pull/1250
@swift-ci please test

kimdv:
https://github.com/apple/sourcekit-lsp/pull/1250
@swift-ci please test Windows

Issue #405, opened 2020-09-22
[SR-13584] Make convenience methods returning an the value of an integer literal as Int and of floating literal as Float return an optional


swift-ci:
|                  |                 |
   |------------------|-----------------|
   |Previous ID       | SR-13584      |
   |Radar             | None         |
   |Original Reporter | vermont42 (JIRA User)      |
   |Type              | Bug    |

   <details>
  <summary>Additional Detail from JIRA</summary>

   |                  |                 |
   |------------------|-----------------|
   |Votes             | 0         |
   |Component/s       | SwiftSyntax    |
   |Labels            | Bug        |
   |Assignee          | vermont42 (JIRA)      |
   |Priority          | Medium      |

   

   md5: 14ad515b7c276d59a1dd762b567c3a4b

  </details>

**Issue Description:**

After the [PR](https://github.com/apple/swift-syntax/pull/239) for SR-11580 merged, there was consensus, based on feedback, that the new convenience methods `floatingValue` and `integerValue`, should be optional in order to prevent potential crashes. This story is for implementation of this change. There was additional feedback involving naming and genericity. Depending on PR-review feedback, this story may encompass implementation of these changes as well.

ahoppen:
rdar://95634857

ahoppen:
If anyone is interested in doing this: We would need to port the following code to Swift for the integer literal:

https://github.com/apple/swift/blob/15f7a28ff3fb6427438d3dcc9d2111fb82f85023/lib/AST/Expr.cpp#L1078-L1167

kimdv:
@ahoppen I could be interested in this, if it's still needed.

ahoppen:
Yes, I think this would still be great to have.

PR #2605, opened 2024-04-17
Add `integerValue` to `IntegerLiteralExprSyntax` and `floatingValue` to `FloatLiteralExprSyntax`

kimdv:
Resolves https://github.com/apple/swift-syntax/issues/405

### DIFF
--- a/Release Notes/601.md
+++ b/Release Notes/601.md
@@ -2,11 +2,21 @@
 
 ## New APIs
 
+- `IntegerLiteralExprSyntax` and `FloatLiteralExprSyntax` now have a computed `representedLiteralValue` property.
+  - Description: Allows retrieving the represented literal value when valid.
+  - Issue: https://github.com/apple/swift-syntax/issues/405
+  - Pull Request: https://github.com/apple/swift-syntax/pull/2605
+
 ## API Behavior Changes
 
 ## Deprecations
 
 ## API-Incompatible Changes
+
+- Moved `Radix` and `IntegerLiteralExprSyntax.radix` from `SwiftRefactor` to `SwiftSyntax`.
+  - Description: Allows retrieving the radix value from the `literal.text`.
+  - Issue: https://github.com/apple/swift-syntax/issues/405
+  - Pull Request: https://github.com/apple/swift-syntax/pull/2605
 
 ## Template
 
--- a/Sources/SwiftSyntax/Convenience.swift
+++ b/Sources/SwiftSyntax/Convenience.swift
@@ -68,6 +68,21 @@ extension EnumCaseParameterSyntax {
   }
 }
 
+extension FloatLiteralExprSyntax {
+  ///  A computed property representing the floating-point value parsed from the associated `literal.text` property.
+  ///
+  /// - Returns: A double value parsed from the associated`literal.text`, or `nil` if the text cannot be parsed as a double.
+  public var representedLiteralValue: Double? {
+    guard !hasError else { return nil }
+
+    let floatingDigitsWithoutUnderscores = literal.text.filter {
+      $0 != "_"
+    }
+
+    return Double(floatingDigitsWithoutUnderscores)
+  }
+}
+
 extension IntegerLiteralExprSyntax {
   public enum Radix {
     case binary
--- a/Sources/SwiftSyntax/Convenience.swift
+++ b/Sources/SwiftSyntax/Convenience.swift
@@ -68,6 +68,75 @@ extension EnumCaseParameterSyntax {
   }
 }
 
+extension IntegerLiteralExprSyntax {
+  public enum Radix {
+    case binary
+    case octal
+    case decimal
+    case hex
+
+    public var size: Int {
+      switch self {
+      case .binary: return 2
+      case .octal: return 8
+      case .decimal: return 10
+      case .hex: return 16
+      }
+    }
+
+    /// The prefix that is used to express an integer literal with this
+    /// radix in Swift source code, e.g., "0x" for hexadecimal.
+    public var literalPrefix: String {
+      switch self {
+      case .binary: return "0b"
+      case .octal: return "0o"
+      case .hex: return "0x"
+      case .decimal: return ""
+      }
+    }
+
+    fileprivate var offset: Int {
+      switch self {
+      case .binary, .hex, .octal:
+        return 2
+      case .decimal:
+        return 0
+      }
+    }
+  }
+
+  public var radix: Radix {
+    let text = self.literal.text
+    if text.starts(with: "0b") {
+      return .binary
+    } else if text.starts(with: "0o") {
+      return .octal
+    } else if text.starts(with: "0x") {
+      return .hex
+    } else {
+      return .decimal
+    }
+  }
+
+  /// A computed property representing the integer value parsed from the associated `literal.text` property, considering the specified radix.
+  ///
+  /// - Returns: An integer value parsed from the associated `literal.text`, or `nil` if the text cannot be parsed as an integer.
+  public var representedLiteralValue: Int? {
+    guard !hasError else { return nil }
+
+    let text = literal.text
+    let radix = self.radix
+    let digitsStartIndex = text.index(text.startIndex, offsetBy: radix.offset)
+    let textWithoutPrefix = text.suffix(from: digitsStartIndex)
+
+    let textWithoutPrefixOrUnderscores = textWithoutPrefix.filter {
+      $0 != "_"
+    }
+
+    return Int(textWithoutPrefixOrUnderscores, radix: radix.size)
+  }
+}
+
 extension MemberAccessExprSyntax {
   /// Creates a new ``MemberAccessExprSyntax`` where the accessed member is represented by
   /// an identifier without specifying argument labels.
--- a/Tests/SwiftSyntaxTest/SyntaxTests.swift
+++ b/Tests/SwiftSyntaxTest/SyntaxTests.swift
@@ -198,4 +198,25 @@ class SyntaxTests: XCTestCase {
       XCTAssertEqual(expr.representedLiteralValue, expected, line: line)
     }
   }
+
+  func testFloatLiteralExprSyntax() {
+    let testCases: [UInt: (String, Double?)] = [
+      #line: ("2", 2),
+      #line: ("2_00_00.001", 2_00_00.001),
+      #line: ("5.3_8", 5.3_8),
+      #line: ("12e3", 12000.0),
+      #line: ("32E1", 320.0),
+      #line: ("0xdEFACE.C0FFEEp+1", 0xdEFACE.C0FFEEp+1),
+      #line: ("0xaffab1e.e1fP-2", 0xaffab1e.e1fP-2),
+      #line: ("🥥", nil),
+      #line: ("12e+3", 12000.0),
+      #line: ("12e-3", 0.012),
+    ]
+
+    for (line, testCase) in testCases {
+      let (value, expected) = testCase
+      let expr = FloatLiteralExprSyntax(literal: .floatLiteral(value))
+      XCTAssertEqual(expr.representedLiteralValue, expected, line: line)
+    }
+  }
 }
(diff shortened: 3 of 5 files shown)


Tests/SwiftSyntaxTest/SyntaxTests.swift
```diff
@@ -168,4 +168,39 @@ class SyntaxTests: XCTestCase {
...
+      #line: ("02", 2),
+      #line: ("020", 20),
+      #line: ("-02", -2),
+      #line: ("2_00_0000", 2000000),
+      #line: ("-2_00_0000", -2000000),
+      #line: ("foo", nil),
+      #line: ("999999999999999999999999999999999999999999999999999999999999999999999999999999", nil),
+      #line: ("0b1010101", 85),
+      #line: ("0xFF", 255),
+      #line: ("0o777", 511),
```
ahoppen, line 184: What do you think about making the right-hand-side of this check be exactly the same as the left-hand side, just without the string literals? That way you can easily tell that `integerValue` and `floatingValue` getters parse the value in the same way as the parser during compilation time.

Sources/SwiftSyntax/Convenience.swift
```diff
@@ -68,6 +68,73 @@ extension EnumCaseParameterSyntax {
...
+    } else if text.starts(with: "0o") {
+      return .octal
+    } else if text.starts(with: "0x") {
+      return .hex
+    } else {
+      return .decimal
+    }
+  }
+
+  public var integerValue: Int? {
```
ahoppen, line 124: A doc comment would be good here, especially detailing why this can be `nil`.

Same for `floatingValue`.

ahoppen (commented):
Thanks. Looks like the `Float` and `Int` initializers do indeed do the job here. Could you also add an entry to the release notes?

ahoppen:
> [StringLiteralExprSyntax has representedLiteralValue property](https://github.com/apple/swift-syntax/blob/728e2f6d5f9fd4a8e45eab6e52e86ad1519a7e2a/Sources/SwiftParser/StringLiteralRepresentedLiteralValue.swift#L21-L26). IMO, we should use the same name. One question is whether we should put them in SwiftParser, WDYT @ahoppen ?

I agree with the naming. Unless there is a reason to put them in the parser (which I don’t think there is, `StringLIteralExprSyntax.representedLiteralValue` only needs it because it hooks back into the lexer for unicode parsing (eg. `\u{1F36D}`). But since we don’t need it here, I’d just put it in the `SwiftSyntax` module.

Sources/SwiftSyntax/Convenience.swift
kimdv: I've tried to add something, could you please verify if that was something you had in mind?

Tests/SwiftSyntaxTest/SyntaxTests.swift
kimdv: Do you mean changing `#line: ("2_00_0000", 2_00_0000),` etc?

kimdv:
> representedLiteralValue

Renamed to `representedLiteralValue`

Tests/SwiftSyntaxTest/SyntaxTests.swift
ahoppen: Yes, exactly.

ahoppen (approved):
Looks good to me.

kimdv:
@ahoppen should this also have a release note?

kimdv:
> Yes, please. I thought I amended my review with that but looks like I never hit the submit button.

Could this do it?

ahoppen (approved):

Release Notes/601.md
```diff
@@ -2,6 +2,11 @@
 
 ## New APIs
 
+- `IntegerLiteralExprSyntax` now has a computed property named `representedLiteralValue` and `FloatLiteralExprSyntax` have one named `representedLiteralValue`.
+  - Description: Provides a convenience property to get the represented literval value.
```
hamishknight, line 6:
```suggestion
  - Description: Provides a convenience property to get the represented literal value.
```

Sources/SwiftSyntax/Convenience.swift
```diff
@@ -68,6 +68,83 @@ extension EnumCaseParameterSyntax {
   }
 }
 
+extension FloatLiteralExprSyntax {
+  ///  A computed property representing the  floating-point value parsed from the associated `literal.text` property.
```
hamishknight, line 72:
```suggestion
  ///  A computed property representing the floating-point value parsed from the associated `literal.text` property.
```

Sources/SwiftSyntax/Convenience.swift
```diff
@@ -68,6 +68,83 @@ extension EnumCaseParameterSyntax {
...
+    }
+
+    fileprivate var offset: Int {
+      switch self {
+      case .binary:
+        return 2
+      case .octal:
+        return 2
+      case .hex:
+        return 2
```
hamishknight, line 109: I would merge these

```suggestion
      case .binary, .octal, .hex:
        return 2
```

Sources/SwiftSyntax/Convenience.swift
```diff
@@ -68,6 +68,83 @@ extension EnumCaseParameterSyntax {
...
+    } else if text.starts(with: "0x") {
+      return .hex
+    } else {
+      return .decimal
+    }
+  }
+
+  /// A computed property representing the integer value parsed from the associated `literal.text` property, considering the specified radix.
+  ///
+  /// - Returns: An  integer value parsed from the associated `literal.text`, or `nil` if the text cannot be parsed as an integer.
```
hamishknight, line 131:
```suggestion
  /// - Returns: An integer value parsed from the associated `literal.text`, or `nil` if the text cannot be parsed as an integer.
```

Tests/SwiftSyntaxTest/SyntaxTests.swift
```diff
@@ -168,4 +168,50 @@ class SyntaxTests: XCTestCase {
     XCTAssertEqual(funcKeywordInTree2?.as(TokenSyntax.self)?.tokenKind, .keyword(.func))
     XCTAssertNotEqual(funcKeywordInTree1.id, funcKeywordInTree2?.id)
   }
+
+  func testIntegerLiteralExprSyntax() {
+    let testCases: [UInt: (String, Int?)] = [
+      #line: ("2", 2),
+      #line: ("02", 2),
+      #line: ("020", 20),
+      #line: ("-02", -2),
```
hamishknight, line 177: Should cases like `-0xA`, `-0o7`, and `-0b1` also be handled? I should note though that we parse the `-` as an operator, so these cases at least won't come for parsed source code.

Tests/SwiftSyntaxTest/SyntaxTests.swift
```diff
@@ -168,4 +168,50 @@ class SyntaxTests: XCTestCase {
...
+      XCTAssertEqual(expr.representedLiteralValue, expected, line: line)
+    }
+  }
+
+  func testFloatLiteralExprSyntax() {
+    let testCases: [UInt: (String, Double?)] = [
+      #line: ("2", 2),
+      #line: ("2_00_00.001", 2_00_00.001),
+      #line: ("5.3_8", 5.3_8),
+      #line: ("12e3", 12000.0),
```
hamishknight, line 204: Could also check for signed exponents, e.g `12e+3` and `12e-3`

kimdv:
@swift-ci please test

hamishknight (approved):

Tests/SwiftSyntaxTest/SyntaxTests.swift
kimdv: @hamishknight I've added the test cases. 
Are the results what you expect `nil`?

Sources/SwiftSyntax/Convenience.swift
```diff
@@ -68,6 +68,79 @@ extension EnumCaseParameterSyntax {
...
+      switch self {
+      case .binary, .hex, .octal:
+        return 2
+      case .decimal:
+        return 0
+      }
+    }
+  }
+
+  public var radix: Radix {
```
rintaro, line 112: Is this supposed to be `public`? If this is not necessary, I'd suggest not to expose this, or we'd need to update the release notes. Same for `enum Radix`.

kimdv: It was already public as part of `Sources/SwiftRefactor/IntegerLiteralUtilities.swift`?
So I'm not sure if it would be a breaking change to make it private

rintaro: Oooh, I see you just moved the code. No worries. Sorry for confusion.

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test windows

Tests/SwiftSyntaxTest/SyntaxTests.swift
ahoppen: I don’t think we need to care about negative cases. Negative numbers should be represented by a number literal with a  prefix operator in the syntax tree.

Release Notes/601.md
```diff
@@ -2,6 +2,11 @@
 
 ## New APIs
 
+- `IntegerLiteralExprSyntax` now has a computed property named `representedLiteralValue` and `FloatLiteralExprSyntax` have one named `representedLiteralValue`.
+  - Description: Provides a convenience property to get the represented literal value.
```
bnbarham, line 6:
```suggestion
  - Description: Allows retrieving the represented literal value when valid.
```

Or something similar?

Sources/SwiftSyntax/Convenience.swift
bnbarham: This would have needed importing `SwiftRefactor` before, so... maybe also worth release noting then?

Tests/SwiftSyntaxTest/SyntaxTests.swift
hamishknight: In that case, yeah LGTM

Sources/SwiftSyntax/Convenience.swift
kimdv: import `SwiftRefactor`?

Not sure I understand. 

It was moved from `SwiftRefactor` to `SwiftSyntax`.

Where it's used `IntegerLitralUtilities` is already `import SwiftSyntax`

bnbarham: Yes, my point is that to use it previously you would have needed to import `SwiftRefactor`, that is no longer the case.

kimdv: Ah okay, but then the "maybe also worth release noting then?" do you mean adding a notice in the release note? 
Just so I understand. Sorry for confusing 😅

ahoppen: Yes, Ben is suggesting to add an entry to the release notes.

Release Notes/601.md
```diff
@@ -2,8 +2,18 @@
...
 ## New APIs
 
+- `IntegerLiteralExprSyntax` and `FloatLiteralExprSyntax` now have a computed `representedLiteralValue` property.
+  - Description: Allows retrieving the represented literal value when valid.
+  - Issue: https://github.com/apple/swift-syntax/issues/405
+  - Pull Request: https://github.com/apple/swift-syntax/pull/2605
+
 ## API Behavior Changes
 
+- Moved `Radix` and `IntegerLiteralExprSyntax.radix` from `SwiftRefactor` to `SwiftSyntax`.
```
kimdv, line 12: @ahoppen @bnbarham not sure if this is a change or a new api?

ahoppen: I think *API-Incompatible Changes* would be the correct section because technically somebody could be importing just `SwiftRefactor` but not `SwiftSyntax`, in which case their code would break. But I think that’s very unlikely since you can’t really do anything without also importing `SwiftSyntax`

kimdv:
@bnbarham @rintaro can you please review :D 
Rebased on latest main

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test windows

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test

rintaro (approved):

PR #2616, opened 2024-04-22
[cmake] Build version marker modules.

jeffdav:
This change adds the version marker modules to the CMake build, which were previously missing.

This replicates the behavior one gets when building w/SPM and allows consumers built with CMake to follow the `#if canImport(SwiftSyntax###)` convention for versioning macros.


ahoppen:
The version marker modules are intentionally missing from the CMake build. The CMake build is primarily designed to be used when swift-syntax gets built to be linked into the compiler, which is always going to be in sync with the swift-syntax repository and thus doesn’t need them.

What’s your motivation for adding the version marker modules to the CMake build?

jeffdav:
> What’s your motivation for adding the version marker modules to the CMake build?

I'm building `swift-testing` with CMake on Windows, and it does the version checks.

jeffdav:
Is there any downside to including it?

ahoppen (approved):
Ah, I see. Just wanted to understand what you wanted to use them for.

ahoppen:
@swift-ci Please test

PR #2621, opened 2024-04-24
Enable Swift 6 mode unless we are in a self-hosted Swift CI job

ahoppen:
We disable Swift 6 mode when the `SWIFTCI_IS_SELF_HOSTED` environment variable is set to work around the following issue: The self-hosted SwiftPM job has Xcode 15.3 (Swift 5.10) installed and builds a Swift 6 SwiftPM from source. It then tries to build itself as a fat binary using the just-built Swift 6 SwiftPM, which uses xcbuild from Xcode as the build system. But the xcbuild in the installed Xcode is too old and doesn't know about Swift 6 mode, so it fails with: SWIFT_VERSION '6' is unsupported, supported versions are: 4.0, 4.2, 5.0 (rdar://126952308)

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -298,7 +298,12 @@ let package = Package(
       exclude: ["Inputs"]
     ),
   ],
-  swiftLanguageVersions: [.v5]
+  // Disable Swift 6 mode when the `SWIFTSYNTAX_DISABLE_SWIFT_6_MODE` environment variable is set. This works around the following
+  // issue: The self-hosted SwiftPM job has Xcode 15.3 (Swift 5.10) installed and builds a Swift 6 SwiftPM from source.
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

bnbarham (approved):

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

ahoppen:
We’re seeing the same kind of issue in the source compatibility suite job, which doesn’t set the `SWIFTCI_IS_SELF_HOSTED` environment variable and thus wouldn’t be covered by this PR.

ahoppen:
https://github.com/apple/swift-package-manager/pull/7495

@swift-ci Please test

ahoppen:
https://github.com/apple/swift-package-manager/pull/7495

@swift-ci Please test

PR #2622, opened 2024-04-24
Work around a diagnostic using `CommandLine.arguments` on older toolchains

grynspan:
`CommandLine.arguments` is declared read/write in older toolchains and this causes an error diagnostic when it's used in Swift 6 mode. This PR uses `CommandLine.argc` and `CommandLine.unsafeArgv` directly instead so the diagnostic does not occur.


grynspan:
@swift-ci please test

PR #2623, opened 2024-04-24
Add support for WebAssembly Macros

kabiroberai:
This PR allows `CompilerPlugin`s to be built with a wasm32-unknown-wasi target, enabling them to be invoked by the Wasm Plugin runtime in https://github.com/apple/swift/pull/73031.

I've chosen to create a `swift_wasm_macro_pump` export to allow the caller to "drive the event loop" since issuing a `read` would by default be blocking. We need nonblocking IO because some runtimes (eg JavaScriptCore) run Wasm on the same thread as the rest of the interpreted code (JavaScript).

To test this, one can build an example macro with
```bash
Examples$ swift build \
  --experimental-swift-sdk wasm32-unknown-wasi \
  --product MacroExamplesImplementation \
  -c release
```

And then, with the changes from the swift and swift-driver PRs, a client can be compiled with
```bash
Examples$ swiftc Client.swift \
  -load-plugin-executable .build/release/MacroExamplesImplementation.wasm#MacroExamplesImplementation
```


Sources/SwiftCompilerPlugin/CompilerPlugin.swift
```diff
@@ -281,3 +303,24 @@ private func describe(errno: CInt) -> String {
...
+
+#if compiler(>=6.0)
+
+@_expose(wasm, "swift_wasm_macro_pump")
+@_cdecl("swift_wasm_macro_pump")
+func wasmPump() {
+  readabilityHandler()
+}
+
+#endif
```
ahoppen, line 324: Does this code require Swift 6 to build for wasm? What happens if you build using a Swift 5 compiler? If it doesn’t build with an older compiler, I think we should put a `#error` in the `#else` branch telling you that you need a Swift 6 compiler to build the plugin server for WASM.

ahoppen (commented):
Looks good from my side but I  would like to get a review from @rintaro as well.

Sources/SwiftCompilerPlugin/CompilerPlugin.swift
kabiroberai: Turns out this was causing pre-6.0 builds (even those not targeting wasm) to fail — somehow Swift ignores the `#if compiler` check and complains that it doesn't recognize `@_expose(wasm)`. I moved the export over to C, so this should now work fine with older compilers. That said I haven't tested a pre-6.0 Wasm-compatible compiler yet, since afaik Swift didn't officially support Wasm until 6.0 — but I can give it a spin with a SwiftWasm toolchain.

kabiroberai:
A couple of things worth mentioning:

1. I didn't add `wasm_support.c` to the CMake source set for _SwiftSyntaxCShims since I don't think we ever build SwiftSyntax for WASI with CMake — and I'm not sure how I would even test that. Is this okay or should I try to allow cross-compiling with CMake?
2. Can we add a smoke test to ensure that SwiftSyntax continues to build for WASI? Unsure of the best way to go about doing this.

kateinoigakukun:
@kabiroberai 
> 2. Can we add a smoke test to ensure that SwiftSyntax continues to build for WASI? Unsure of the best way to go about doing this.

I'm not sure we have a way to run smoke test with Wasm target in CI right now. But we are trying to build Wasm SDK in CI here https://github.com/apple/swift/pull/72728 and after the integration, we will be able to use Wasm target in CI.

Sources/SwiftCompilerPluginMessageHandling/StandardIOMessageConnection.swift
```diff
@@ -146,6 +152,15 @@ public struct StandardIOMessageConnection: MessageConnection {
   }
 }
 
+#if os(WASI)
+// fatalError writes to stdout, which is the message
+// output stream under WASI
+public func internalError(_ message: String) -> Never {
+  fputs("Internal Error: \(message)\n", _stderr)
```
rintaro, line 159: > fatalError writes to stdout

As far as I can see it's `stderr`. Am I missing anything?  https://github.com/apple/swift/blob/77f53a5e50f0cf964ae9dcc42f78ec43227a3db2/stdlib/public/runtime/Errors.cpp#L323-L332

But anyway, I think I'm going to use `fputs("Internal Error: \(message)\n", _stderr)` in all platforms, (we don't want `#file` etc, and other magic in `fatalError`). So I will bring `internalError()` back to `CompilerPlugin.swift`

Sources/SwiftCompilerPlugin/CompilerPlugin.swift
```diff
@@ -112,12 +112,45 @@ extension CompilerPlugin {
...
+    #if os(WASI)
+    // Rather than blocking on read(), let the host tell us when there's data.
+    readabilityHandler = {
+      do {
+        _ = try impl.handleNextMessage()
+      } catch {
+        internalError("\(error)")
+      }
+    }
+    #else
```
rintaro, line 124: I _feel_ this logic should be sinked to `impl.main()`, so `handleNextMessage()` can be `internal`.

Sources/SwiftCompilerPluginMessageHandling/CompilerPluginMessageHandler.swift
```diff
@@ -86,10 +86,20 @@ public class CompilerPluginMessageListener<Connection: MessageConnection, Provid
...
+  ///
+  /// - Returns: `true` if there was a message to read, `false`
+  /// if the end-of-file was reached.
+  public func handleNextMessage() throws -> Bool {
+    guard let message = try connection.waitForNextMessage(HostToPluginMessage.self) else {
+      return false
     }
+    let result = handler.handleMessage(message)
+    try connection.sendMessage(result)
+    return true
```
rintaro, line 102: I opened a PR https://github.com/apple/swift-syntax/pull/2631

My intention is, in this PR, you'd make it like:
```swift
  public func main() {
    #if os(WASI)
    readabilityHandler = { _ = self.handleNextMessage() }
    #else
    while handleNextMessage() {}
    #endif
  }
  
  func handleNextMessage() {
    do {
      guard let message = try connection.waitForNextMessage(HostToPluginMessage.self) else {
        return false
      }
      let result = handler.handleMessage(message)
      try connection.sendMessage(result)
      return true
      }
    } catch {
      fputs("Internal Error: \(message)\n", _stderr)
      exit(1)
    }
  }
```
I think this should minimize `#if os(WASI)` branch length and improves the readability.

Sources/SwiftCompilerPluginMessageHandling/StandardIOMessageConnection.swift
kabiroberai: Indeed, looks like you're right. Not sure why I thought it wrote to stdout — might've forgotten to attach stderr to the WASI bridge at some point.

Sources/_SwiftSyntaxCShims/wasm_support.c
```diff
@@ -0,0 +1,35 @@
...
+#define SWIFT_WASM_MACRO_ABI 1
+
+#define _STR(X) #X
+#define STR(X) _STR(X)
+
+// LLVM has special-cased handling to map .custom_section.foo to
+// Wasm Custom Section "foo". this must be a metadata section rather
+// than a data section so we can't use __attribute__((section)) for it.
+// See: https://reviews.llvm.org/D43097
+__asm__("\t.section .custom_section.swift_wasm_macro_abi,\"\",@\n\t.4byte " STR(SWIFT_WASM_MACRO_ABI) "\n");
```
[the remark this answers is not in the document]
kabiroberai: This adds a custom `swift_wasm_macro_abi` section to the binary with a little-endian uint32 value. We check the ABI version in the wasm executor in order to future-proof. See:

https://github.com/apple/swift/pull/73031/files#diff-2261ec558c289a4bd568f211772e906caa997f8c26b38e4c6d5d37b7b1554379R47

Will add a brief explanation to the comment.

Sources/SwiftCompilerPlugin/CompilerPlugin.swift
kabiroberai: hmm I'm worried that moving `readabilityHandler` into `SwiftCompilerPluginMessageHandling` would be leaky, since the `swift_wasm_macro_pump` export (which invokes `readabilityHandler`) is an impl detail of the `SwiftCompilerPlugin` target. Also worth mentioning that `CompilerPluginMessageListener` is `@_spi(PluginMessage)` so I don't think we're creating any new API contracts even if `handleNextMessage` is technically public.

kabiroberai: Actually I just realized that `_SwiftSyntaxCShims` is a dependency of `SwiftCompilerPluginMessageHandling` so the latter is already strongly coupled to our wasm ABI. Moved `readabilityHandler` into that module.

Sources/_SwiftSyntaxCShims/wasm_support.c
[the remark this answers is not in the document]
kabiroberai: hmm actually @kateinoigakukun what do you think about using the export name to indicate the ABI instead? E.g. we could export the pump function with the name `swift_wasm_macro_pump_v1` and the runtime could check for the presence of this to indicate the v1 ABI. The custom section support feels shaky to me — I wouldn't be surprised if there are tools that don't know how to handle custom wasm sections (or, alternatively, strip them out.)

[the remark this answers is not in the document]
kabiroberai: Pushed this as a tentative change, happy to revert if the Custom Sections approach is preferable but I do like this atm because it's a bit more lightweight.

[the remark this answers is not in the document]
kabiroberai: For posterity: discussed this with @kateinoigakukun [here](https://swift-open-source.slack.com/archives/C06MUFU7ZCZ/p1715394639346069?thread_ts=1711575681.096789&cid=C06MUFU7ZCZ); we aligned on using the exported function name rather than a custom section, since some Wasm tooling (_ahem_ wasm-strip) can be too trigger-happy about stripping custom sections.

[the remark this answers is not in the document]
grynspan: I think Swift tries to be conservative with custom sections because it's a terribly non-portable solution. Although portability isn't at issue here, it might maybe make sense to use an exported function name. That'd be the typical solution here if we were talking about a C function on another platform, I think.

kabiroberai:
Hey @rintaro, could you please take another look at this when you have a minute? Thanks!

Sources/SwiftCompilerPlugin/CompilerPlugin.swift
DougGregor: FWIW, if you're having trouble with an `#if compiler` check, use something like `#if compiler(>=5.3) && hasAttribute(_expose)`, and older compilers won't look into the `#if` at all.

kabiroberai: So this didn't quite work but it got me poking around the `Parser::parseIfConfig` code and I discovered a bug:

The following fails to compile
```swift
#if os(WASI)
#if compiler(>=100)
foo
#endif
#endif
```

whereas it compiles fine if we swap the conditions
```swift
#if compiler(>=100)
#if os(WASI)
foo
#endif
#endif
```

Seems like if we're already in an inactive `#if` clause, the parser doesn't respect the "disable parsing entirely" functionality of nested `#if compiler` decls. I'll see if I can make a followup in apple/swift to fix this but in the meantime the fix is thankfully as easy as swapping the statements around.

Sources/SwiftCompilerPluginMessageHandling/PluginMessages.swift
```diff
@@ -229,7 +229,8 @@ public enum PluginMessage {
     public var notes: [Note]
     public var fixIts: [FixIt]
 
-    internal init(
+    @_spi(PluginMessage)
+    public init(
```
rintaro, line 233:
```suggestion
    public init(
```
All symbols in `@_spi` implicitly inherits it. Could you tell me what's using this?

Sources/SwiftCompilerPluginMessageHandling/CompilerPluginMessageHandler.swift
```diff
@@ -105,10 +126,18 @@ public class CompilerPluginMessageListener<Connection: MessageConnection, Provid
...
-/// 'CompilerPluginMessageHandler' is a type that handle a message and do the
-/// corresponding operation.
+/// A type that handles a plugin message and returns a response.
+///
+/// - SeeAlso: ``PluginProviderMessageHandler``
+@_spi(PluginMessage)
+public protocol PluginMessageHandler {
+  /// Handles a single message received from the plugin host.
+  func handleMessage(_ message: HostToPluginMessage) -> PluginToHostMessage
+}
```
rintaro, line 136: I guess this abstraction is needed anyway, so this is okay. But  TBH, I'm not a fan of `WasmInterceptingMessageHandler` in https://github.com/apple/swift/pull/73031.

Sources/SwiftCompilerPluginMessageHandling/CompilerPluginMessageHandler.swift
```diff
@@ -195,6 +224,10 @@ public class CompilerPluginMessageHandler<Provider: PluginProvider> {
   }
 }
 
+@_spi(PluginMessage)
+@available(*, deprecated, renamed: "PluginProviderMessageHandler")
+public typealias CompilerPluginMessageHandler<Provider: PluginProvider> = PluginProviderMessageHandler<Provider>
```
rintaro, line 229: Is anything in this repo or the compiler repo using this? If not, `deprecated` is not needed. Let's just remove it.

rintaro (commented):
I like the new approach of the versioning!

Sources/SwiftCompilerPluginMessageHandling/PluginMessages.swift
kabiroberai: We're using this to emit diagnostics from `WasmInterceptingMessageHandler`:

https://github.com/apple/swift/blob/caa1a8ccd01eb568bc1b65007882bf8840e08ad6/tools/swift-plugin-server/Sources/swift-plugin-server/WasmMessageHandler.swift#L86-L100

Sources/SwiftCompilerPluginMessageHandling/CompilerPluginMessageHandler.swift
kabiroberai: Yeah I'm not the biggest fan either :/ but that's the only approach I could think of which allows us to send _some_ messages from `swift-plugin-server` to `LibraryPluginProvider` while intercepting others. The alternative is to have an entirely separate `swift-wasm-plugin-server` (so that the primary plugin server can use `LibraryPluginProvider` directly) but both Doug and Yuta feel that having a unified plugin server is preferable (see https://github.com/apple/swift/pull/73031#discussion_r1580502591), and I agree that the benefits outweigh the downsides atm.

Curious to hear if you have any alternative suggestions though.

rintaro:
@swift-ci Please test

kabiroberai:
@rintaro I updated spi.yml which should fix the test failure — could you please request a re-run?

kateinoigakukun:
@swift-ci Please test

kabiroberai:
would appreciate another run, third time's the charm hopefully 🤞

kateinoigakukun:
@swift-ci test

DougGregor:
@swift-ci please test Windows

Sources/SwiftCompilerPluginMessageHandling/CompilerPluginMessageHandler.swift
```diff
@@ -212,3 +241,31 @@ extension PluginProvider {
...
+/// This is safe to access without serialization as Wasm plugins are single-threaded.
+nonisolated(unsafe) private var readabilityHandler: () -> Void = {
+  fatalError(
+    """
+    CompilerPlugin.main wasn't called. Did you annotate your plugin with '@main'?
+    """
+  )
+}
+
+@_expose(wasm,"swift_wasm_macro_v1_pump")
```
kateinoigakukun, line 258: Incorrect formatting on the attribute arguments will be fixed by https://github.com/apple/swift-format/pull/750

Sources/SwiftCompilerPlugin/CompilerPlugin.swift
kabiroberai: https://github.com/apple/swift/pull/74415

kabiroberai:
Hi folks, seems like the one remaining request is to adjust the indentation here once the corresponding fix lands in swift-format: https://github.com/apple/swift-syntax/pull/2623#discussion_r1639172780

Seeing as the swift-format PR is in turn blocked by some CI issues, would it make sense to merge this PR without that one whitespace change and follow up once the swift-format PR lands?

kateinoigakukun:
@swift-ci test

kabiroberai:
Linux and macOS are green! Can we please run the Windows suite too? cc @kateinoigakukun

kateinoigakukun:
@swift-ci test Windows

kabiroberai:
👋 seeing as all tests have passed are we good to merge this?

grynspan:
> 👋 seeing as all tests have passed are we good to merge this?

Unfortunately, it looks like all CI jobs have failed, and this PR has not been approved by any code owners. So it's not ready to merge yet.

kabiroberai:
looks like CI was re-run today and a few things are out of sync now. will bring it back into working condition.

Sources/SwiftCompilerPluginMessageHandling/CompilerPluginMessageHandler.swift
kabiroberai: Re-added this since it's used by `InProcPluginServer` as of https://github.com/swiftlang/swift/pull/73725. Should be a simple find-and-replace once this PR is merged.

grynspan:
Thanks!

kabiroberai:
🙌 could we please run CI again?

kateinoigakukun:
@swift-ci test

ahoppen:
@swift-ci Please test Windows

Sources/SwiftCompilerPluginMessageHandling/CompilerPluginMessageHandler.swift
rintaro: Ah, my bad. Thanks!

rintaro (approved):
Thank you for patiently dealing with me! LGTM :shipit:

PR #2626, opened 2024-04-25
Improvements to integer literal utilities used in SourceKit-LSP

DougGregor:
SourceKit-LSP's refactoring action has minor extensions for the integer literal utilities. Sink them down here for more general use:
* IntegerLiteralExprSyntax.Radix becomes CaseIterable
* IntegerLiteralExprSyntax.Radix gains a "literalPrefix" property to get the string needed to prefix a literal with this radix.

### DIFF
--- a/Sources/SwiftRefactor/IntegerLiteralUtilities.swift
+++ b/Sources/SwiftRefactor/IntegerLiteralUtilities.swift
@@ -36,10 +36,10 @@ extension IntegerLiteralExprSyntax {
     /// radix in Swift source code, e.g., "0x" for hexadecimal.
     public var literalPrefix: String {
       switch self {
-      case .binary: "0b"
-      case .octal: "0o"
-      case .hex: "0x"
-      case .decimal: ""
+      case .binary: return "0b"
+      case .octal: return "0o"
+      case .hex: return "0x"
+      case .decimal: return ""
       }
     }
   }
--- a/Sources/SwiftRefactor/IntegerLiteralUtilities.swift
+++ b/Sources/SwiftRefactor/IntegerLiteralUtilities.swift
@@ -17,7 +17,7 @@ import SwiftSyntax
 #endif
 
 extension IntegerLiteralExprSyntax {
-  public enum Radix: CaseIterable {
+  public enum Radix {
     case binary
     case octal
     case decimal
--- a/Sources/SwiftRefactor/IntegerLiteralUtilities.swift
+++ b/Sources/SwiftRefactor/IntegerLiteralUtilities.swift
@@ -17,7 +17,7 @@ import SwiftSyntax
 #endif
 
 extension IntegerLiteralExprSyntax {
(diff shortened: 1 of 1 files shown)


DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test Windows

DougGregor:
@swift-ci please test Windows

Sources/SwiftRefactor/IntegerLiteralUtilities.swift
```diff
@@ -17,7 +17,7 @@ import SwiftSyntax
 #endif
 
 extension IntegerLiteralExprSyntax {
-  public enum Radix {
+  public enum Radix: CaseIterable {
```
rintaro, line 20: As for the use-case in SourceKit-LSP, I'd just do:

```swift
for radix: IntegerLiteralExprSyntax.Radix in [.binary, .octal, .hex, .decimal] { ... }
```
This is clearer and more deterministic what it will do, compared to `.allCases`. 

I personally avoid `CaseIterable` unless there's a _strong_ reason to do. Users cannot control the order (maybe it might not matter in this specific use-case)

DougGregor: Hmm, I'm doing this based on the discussion in https://github.com/apple/sourcekit-lsp/pull/1179#discussion_r1569570269. Code that is working with radix values via the APIs here can do so without having to reason about the particulars of the radix; it can handle all of them. That seems like it makes this a reasonable API to provide.

rintaro: Yes I know the discussion. But,

> it can handle all of them.

All of _known_ cases. Perhaps in the future, we might introduce a case that has "suffix". e.g. `0nA24nG_r62` (whatever it is). If we only handle known cases `[.binary, .octal, .hex, .decimal]`, the code would probably still work correctly because we'd not change the syntax of the current radix cases. But using `.allCases` can cause misbehavior.

FWIW https://github.com/apple/swift-syntax/pull/2292, we have a history of removing `CaseIterable`. I want to be really careful to re-adding it.

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test Windows

Sources/SwiftRefactor/IntegerLiteralUtilities.swift
DougGregor: I think it's unlikely to ever cause a problem, given that the literals have been stable for a decade, but neither is it more than an inconvenience if we don't have this. So I've removed it.

DougGregor:
@swift-ci please test Windows

DougGregor:
@swift-ci please test Windows

rintaro (approved):

PR #2630, opened 2024-04-28
[AccessorMacro] disallow accessor macro apply to variable declaration

### LINKED ISSUE #2491: [AccessorMacro] 'let' declarations cannot be computed properties
### Description

The following Swift code fails to compile:

```swift
let x: Int {  //  'let' declarations cannot be computed properties
  1
}
```

It looks like our current `AccessorMacro` implementation adds computed getters to `let` variables without transforming the `let` to `var`:

```swift
extension AccessorMacroTests {
  func testLetDeclarations() {
    assertMacroExpansion(
      """
      @constantOne
      let x: Int = 1
      """,
      expandedSource: """
        var x: Int {
          get {
            return 1
          }
        }
        """,
      macros: ["constantOne": ConstantOneGetter.self],
    )

    // Bit of an odd case, compiler has the type but we don't know it in `MacroSystem`
    assertMacroExpansion(
      """
      @constantOne
      let x = 1
      """,
      expandedSource: """
        var x {
          get {
            return 1
          }
        }
        """,
      macros: ["constantOne": ConstantOneGetter.self],
    )
  }
}
```

This test fails:

```
error: -[SwiftSyntaxMacroExpansionTest.AccessorMacroTests testLetDeclarations] : failed - Macro expansion did not produce the expected expanded source
–var x: Int {
+let x: Int {
   get {
     return 1
   }
 }

Actual expanded source:

let x: Int {
  get {
    return 1
  }
}
error: -[SwiftSyntaxMacroExpansionTest.AccessorMacroTests testLetDeclarations] : failed - Macro expansion did not produce the expected expanded source
–var x {
+let x {
   get {
     return 1
   }
 }

Actual expanded source:

let x {
  get {
    return 1
  }
}
```

Should we update the `AccessorMacro` implementation to produce Swift more consistent with what engineers are able to compile? Should we be transforming that `let` to a `var`?

### Steps to Reproduce

_No response_

RayZhao1998:
Resolve #2491

### DIFF
--- a/Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
+++ b/Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
@@ -633,6 +633,7 @@ private enum MacroApplicationError: DiagnosticMessage, Error {
   case accessorMacroOnVariableWithMultipleBindings
   case peerMacroOnVariableWithMultipleBindings
   case malformedAccessor
+  case accessorMacroOnLetVariable
 
   var diagnosticID: MessageID {
     return MessageID(domain: diagnosticDomain, id: "\(self)")
@@ -650,6 +651,8 @@ private enum MacroApplicationError: DiagnosticMessage, Error {
       return """
         macro returned a malformed accessor. Accessors should start with an introducer like 'get' or 'set'.
         """
+    case .accessorMacroOnLetVariable:
+      return "accessor macro is not allowed to be applied to a 'let' variable"
     }
   }
 }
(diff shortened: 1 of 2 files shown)


ahoppen:
This change will only disallow accessor macros on `let` variables in `assertMacroExpansion`, not during compilation. Disallowing it during compilation needs to be done in the compiler repository. I would prefer not to have a  difference in behavior, so those changes would need to be done in sync.

RayZhao1998:
All right.  Going to close this PR for now. Maybe I can try to make it done in the compiler repository 😀.

ahoppen:
If you are interested in making the change in the compile repo, it’s probably a `var->isLet` check that needs to be added for accessors that aren’t valid for `let` declarations somewhere around here

https://github.com/apple/swift/blob/f7c9966420051fb09926704c2f6d9001242c7e64/lib/Sema/TypeCheckMacros.cpp#L1625

PR #2631, opened 2024-04-29
[CompilerPlugin] Sink error exit logic to "message listener"

rintaro:
And print the error to `stderr` manually, instead of `fatalError()`. We don't want to print the file name / line number for this.


rintaro:
@swift-ci Please test

ahoppen (approved):

PR #2633, opened 2024-04-29
[JSON] 5.7 compatibility

rintaro:
* Add SE-0370 methods
* Return concrete type instead of existential


rintaro:
@swift-ci Please test

ahoppen (approved):

PR #2636, opened 2024-05-01
[Tests] Use '#filePath' as default argument instead of '#file'

rintaro:
Avoid warnings in Swift 6:
```
warning: parameter 'file' with default argument '#file' passed to parameter 'file', whose default argument is '#filePath'
 ```


rintaro:
@swift-ci Please test

ahoppen (approved):

PR #2638, opened 2024-05-02
Rename `errno.h` and `stdio.h` to `swiftsyntax_errno.h` and `swiftsyntax_stdio.h`

ahoppen:
In some build scenarios compilers get confused if a file named `errno.h` includes `errno.h`, even if it’s intended to be from a different directory. Renaming the files in swift-syntax works around that issue.


ahoppen:
@swift-ci Please test

rintaro (approved):

PR #2640, opened 2024-05-03
Add an umbrella header for _SwiftSyntaxCShims

ahoppen:
Having an umbrella header helps build swift-syntax in some build environments.


ahoppen:
@swift-ci Please test

PR #2653, opened 2024-05-14
Don't use an instance property before it's defined

DougGregor:
We're using an instance property before it's been given a value, which wasn't being diagnosed by the compiler.


DougGregor:
@swift-ci please test

ahoppen (approved):

PR #2650, opened 2024-05-09
Remove flag to disable Swift 6 mode

ahoppen:
When https://github.com/apple/swift-package-manager/pull/7545 is in, the `SWIFTSYNTAX_DISABLE_SWIFT_6_MODE` environment variable isn’t used anymore.

PR #2654, opened 2024-05-14
[6.0] Remove flag to disable Swift 6 mode

ahoppen:
* **Explanation**: Remove code that forced swift-syntax to be built in Swift 5 mode because the underlying issue was fixed by https://github.com/apple/swift-package-manager/pull/7504
* **Scope**: None, SwiftPM was the only client to set the environment variable and it has removed the setting in https://github.com/apple/swift-package-manager/pull/7558
* **Risk**: Low, if there are any issues, they will be caught at build time
* **Testing**: n/a
* **Issue**: n/a
* **Reviewer**:  @bnbarham on https://github.com/apple/swift-syntax/pull/2650

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -295,12 +295,7 @@ let package = Package(
       exclude: ["Inputs"]
     ),
   ],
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

PR #2656, opened 2024-05-17
[Macros] Fix column calculation in MacroExpansionContext for plugins

rintaro:
When retriving a location for a node at the first line, column calculation was incorrect. 'column' of 'PluginMessage.Syntax' is 1-based, so we need to '-1' when using it as a offset.


rintaro:
@swift-ci Please test

Examples/Sources/MacroExamples/Playground/main.swift
```diff
@@ -55,3 +55,7 @@ runPeerMacrosPlayground()
 runEnvironmentValueAccessorMacroPlayground()
 
 #endif
+
+// MARK: - SourceLocationMacros
+
+runSourceLoctionMacrosPlayground()
```
ahoppen, line 61: Would it make sense to also have an `XCTest` test for this that will fail if this breaks instead of just a `print` in the macro examples?

rintaro: Unfortunately XCTest case's expansion (`assertMacroExpansion`) doesn't use this code path.

rintaro:
@swift-ci Please test

Examples/Sources/MacroExamples/Playground/main.swift
ahoppen: OK, no worries

ahoppen:
@swift-ci Please test Windows

ahoppen (approved):

PR #2657, opened 2024-05-17
[6.0][Macros] Fix column calculation in MacroExpansionContext for plugins

rintaro:
Cherry-pick https://github.com/apple/swift-syntax/pull/2656 into release/6.0

rdar://128293478 (Source location column can be shifted by 1)
* **Explanation**: When retrieving the location of a node at the first line, column calculation was incorrect. Diagnostics locations and `MacroExpansionContext.location(of:at:filePathMode:)` were often resulted wrong locations. Since `column` of `PluginMessage.Syntax` is 1-based, we need to `-1` when using it as an offset.
* **Scope**: Executable macro plugin expansion
* **Risk**: Low. A simple and clear fix
* **Testing**: Added an reproducing test case to `Examples`, and tested it locally.
* **Issue**: rdar://128293478
* **Reviewer**: TBA


rintaro:
@swift-ci Please test

bnbarham (approved):

PR #2659, opened 2024-05-17
[CMake] Add option to set the target namespace

rintaro:
Each target name is prefixed with `SWIFTSYNTAX_TARGET_NAMESPACE`

Also, add `SWIFTSYNTAX_EMIT_MODULE` setting; `true` for building library-evolution enabled modules with .swiftinterface, `false` for building fragile modules, undefined is `true`


rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

etcwilde:
@swift-ci please test Linux platform

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

rintaro:
@swift-ci Please test Windows

rintaro:
@swift-ci Please test Windows

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

cmake/modules/AddSwiftHostLibrary.cmake
```diff
@@ -17,10 +17,21 @@
 #
 # Remove once rdar://102202478 is fixed.
```
hamishknight, line 18: Presumably we now need to use this function regardless of whether `rdar://102202478` is fixed?

cmake/modules/AddSwiftHostLibrary.cmake
```diff
@@ -17,10 +17,21 @@
...
   cmake_parse_arguments(ARGS "PUBLIC;PRIVATE;INTERFACE" "" "" ${ARGN})
-  foreach(DEPENDENCY ${ARGS_UNPARSED_ARGUMENTS})
+  set(link_type)
+  if(ARGS_PUBLIC)
+    set(link_type PUBLIC)
+  elseif(ARGS_PRIVATE)
+    set(link_type PRIVATE)
+  elseif(ARGS_INTERFACE)
+    set(link_type INTERFACE)
+  endif()
```
hamishknight, line 28: Should it be an error if you omit the link type?

hamishknight (approved):

cmake/modules/AddSwiftHostLibrary.cmake
rintaro: https://cmake.org/cmake/help/latest/command/target_link_libraries.html#libraries-for-both-a-target-and-its-dependents
Since it's optional for `target_link_libraries`, I'm leaving it as is for now

cmake/modules/AddSwiftHostLibrary.cmake
rintaro: 👍  Moved the comment to the `-dep.swift` hack.

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

cmake/modules/AddSwiftHostLibrary.cmake
hamishknight: Fair enough, my main concern would be if a new link type was added that wasn't handled here, it may get silently dropped

PR #2660, opened 2024-05-18
Fix up a mismatch between effect specifier lookahead and parsing

### LINKED ISSUE #2648: swift-format fails to parse valid swift code
The following snippet fails to parse in `swift-format` but `swiftc` is able to compile it:
```swift
func foo<E: Error, T>(a: T, b: T, body: (T, T) throws(E) -> T) throws(E) -> T {
  try body(a, b)
}

struct S: Error { }

let x = try foo(a: 0, b: 1) { a, b throws(S) in 2 }
print(x)
```

repro:

```console
# compile
$ xcrun -toolchain org.swift.59202405011a swiftc test.swift
$ ./test
2

# format
$ xcrun -toolchain org.swift.59202405011a swift-format format test.swift
/Users/rauhul/Desktop/test.swift:7:32: error: unexpected code ', b throws(S) in 2' in closure
```

bnbarham:
Lookahead was not consuming typed throw's error argument during lookahead for closures. `atFunctionTypeArrow` did mostly handle it, but only consumed a single token and thus wouldn't match `throws(any Error)`. Handle this in `consumeEffectsSpecifiers` and then use this for both.

Fixes #2648.
Resolves rdar://127750606.

### DIFF
--- a/Sources/SwiftParser/Expressions.swift
+++ b/Sources/SwiftParser/Expressions.swift
@@ -2428,10 +2428,15 @@ extension Parser.Lookahead {
   // Consume 'async', 'throws', and 'rethrows', but in any order.
   mutating func consumeEffectsSpecifiers() {
     var loopProgress = LoopProgressCondition()
-    while let (_, handle) = self.at(anyIn: EffectSpecifier.self),
+    while let (spec, handle) = self.at(anyIn: EffectSpecifier.self),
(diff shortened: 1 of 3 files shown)


bnbarham:
@swift-ci please test

DougGregor (approved):
Looks great, thank you!

ahoppen (approved):
Thanks

PR #2663, opened 2024-05-22
[SourceLocationConverter] Performance improvement

rintaro:
* Stop using `TokenSequence`. Instead, visit `RawSyntax` tree manually
* Avoid parsing trivia, and creating `Array` of `RawTriviaPiece`. Instead, scan the whole token text directly if available
* Collect `#sourceLocation()` directives in one-pass

### DIFF
--- a/Sources/SwiftSyntax/SourceLocation.swift
+++ b/Sources/SwiftSyntax/SourceLocation.swift
@@ -512,20 +512,29 @@ extension SyntaxProtocol {
 fileprivate func computeLines(
   tree: Syntax
 ) -> ([AbsolutePosition], AbsolutePosition) {
-  var lines: [AbsolutePosition] = []
-  // First line starts from the beginning.
-  lines.append(.startOfFile)
-  var position: AbsolutePosition = .startOfFile
-  let addLine = { (lineLength: SourceLength) in
-    position += lineLength
-    lines.append(position)
+  class ComputeLineVisitor: SyntaxVisitor {
+    var lines: [AbsolutePosition] = [.startOfFile]
+    var position: AbsolutePosition = .startOfFile
+    var curPrefix: SourceLength = .zero
+
+    init() {
+      super.init(viewMode: .sourceAccurate)
+    }
+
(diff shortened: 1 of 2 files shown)


rintaro:
@swift-ci Please test

hamishknight (approved):

Sources/SwiftSyntax/SourceLocation.swift
```diff
@@ -511,21 +481,21 @@ extension SyntaxProtocol {
...
+) -> (
+  lines: [AbsolutePosition],
+  endOfFile: AbsolutePosition,
+  sourceLocationDirectives: [(sourceLine: Int, arguments: SourceLocationDirectiveArguments?)]
+) {
+  var lines: [AbsolutePosition] = [.startOfFile]
   var position: AbsolutePosition = .startOfFile
-  let addLine = { (lineLength: SourceLength) in
+  var sourceLocationDirectives: [(sourceLine: Int, arguments: SourceLocationDirectiveArguments?)] = []
+  let prefix = tree.raw.forEachLineLength { lineLength in
```
ahoppen, line 492: `prefix` doesn’t really make sense as a variable name here, right? Shouldn’t it be `sourceFileLength` or something? Also, the usage of `prefix` in `RawSyntax.forEachLineLength` + the comment on it seem pretty confusing to me. Not your doing but might be worth to clean that up along the way.

ahoppen (approved):

Sources/SwiftSyntax/SourceLocation.swift
rintaro: This `prefix` is more like `lastLineLength`. I changed the naming so.
I didn't changed the naming in `forEachLineLength` functions, because I wasn't able to come up with better naming. It's like "the length of the current line _so far_"

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test macOS

PR #2665, opened 2024-06-01
[SwiftParser] Handle extraneous case keyword after comma in switch case.

gregomni:
For errors of the form `case .a, case .b:` instead of the expected `case .a, .b:`, produce just a single initial unexpected token in the second `SwitchCaseItemSyntax` instead of a bad pattern and then a second separate full case.

See <https://github.com/apple/swift/pull/74071>, which is essentially the same change made to the old parser.

### DIFF
--- a/Sources/SwiftParser/Expressions.swift
+++ b/Sources/SwiftParser/Expressions.swift
@@ -2361,17 +2361,28 @@ extension Parser {
     do {
       var keepGoing: RawTokenSyntax? = nil
       var loopProgress = LoopProgressCondition()
+      var unexpectedPrePatternCase: RawUnexpectedNodesSyntax? = nil
       repeat {
         let (pattern, whereClause) = self.parseGuardedCasePattern()
         keepGoing = self.consume(if: .comma)
         caseItems.append(
           RawSwitchCaseItemSyntax(
+            unexpectedPrePatternCase,
             pattern: pattern,
             whereClause: whereClause,
             trailingComma: keepGoing,
             arena: self.arena
           )
         )
+
+        if keepGoing != nil, let caseToken = self.consume(if: .keyword(.case)) {
+          unexpectedPrePatternCase = RawUnexpectedNodesSyntax(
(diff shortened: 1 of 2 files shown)


ahoppen (approved):
Looking great!

ahoppen:
@swift-ci Please test

gregomni:
Oops! That test didn't fail for me locally, but I see the problem.

Sources/SwiftParser/Expressions.swift
```diff
@@ -2361,17 +2361,28 @@ extension Parser {
...
           )
         )
+
+        if keepGoing != nil, let caseToken = self.consume(if: .keyword(.case)) {
+          unexpectedPrePatternCase = RawUnexpectedNodesSyntax(
+            elements: [RawSyntax(caseToken)],
+            arena: self.arena
+          )
+        } else {
+            unexpectedPrePatternCase = nil
```
ahoppen, line 2384: Can you change indentation to 2 spaces by running swift-format (https://github.com/apple/swift-syntax/blob/main/CONTRIBUTING.md#formatting)?

ahoppen (commented):
Thanks for the quick fix. Could you also squash your commits? We generally prefer that (https://github.com/apple/swift-syntax/blob/main/CONTRIBUTING.md#authoring-commits)

Sources/SwiftParser/Expressions.swift
gregomni: Done! Sorry about the sloppiness.

ahoppen: Not a problem at all.

ahoppen:
By the way, the test didn’t fail locally for you because CI runs test case mutation which isn’t enabled by default. https://github.com/apple/swift-syntax/blob/main/CONTRIBUTING.md#test-case-mutation has a little more documentation about it.

ahoppen:
@swift-ci Please test

PR #2672, opened 2024-06-06
[6.0] Heap allocate our atomics

ahoppen:
* **Explanation**: We used C atomics but these were allocated as Swift variables. Even thought they were atomic, concurrent accesses to them could violate Swift’s exclusivity laws, raising thread sanitizer errors. Allocate the C atomic using malloc to fix this problem.
* **Scope**: Only affects assert builds
* **Risk**: Low, only affects assert builds
* **Testing**: Verified that we no longer hit a thread sanitizer issue in SourceKit-LSP with this change
* **Issue**: rdar://129170128
* **Reviewer**:  @hamishknight on https://github.com/apple/swift-syntax/pull/2671


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #2675, opened 2024-06-09
Make SourceLocationConverter Sendable

j-f1:
This type was not able to be marked as Sendable because its `sourceLocationDirectives` property was declared as a `var`. However, this property was never actually mutated. I changed the initializers to always assign it a value, removing the need for the default `[]` value and enabling it to change to a `let`.


j-f1:
@swift-ci please test

ahoppen (approved):
Thank you! Would you mind also creating a cherry-pick PR to `release/6.0` for this?

PR #2676, opened 2024-06-09
Remove `secondName` parameter from `EnumCaseDeclSyntax`

ahoppen:
Enum cases cannot have a second name.


ahoppen:
I was wrong, enum case declarations can have a second name if the first one is unlabeled, like `case abc(_ internalLable: Int)`.

PR #2678, opened 2024-06-09
Fix renaming enum case parameters with unnamed associated arguments

ahoppen:
Linked to and https://github.com/apple/swift/pull/74241 and https://github.com/apple/sourcekit-lsp/pull/1471

We treated enum case parameters the same way as function parameters and weren’t considering that they can be unlabeled. That caused us to insert eg. `_ ` in front of the case’s type, producing `case myCase(_ String)`, which is invalid. Report `enumCaseParameters` as a special parameter kind so that sourcekitd can work with it.

https://github.com/apple/sourcekit-lsp/issues/1228

### DIFF
--- a/Sources/SwiftIDEUtils/DeclNameLocation.swift
+++ b/Sources/SwiftIDEUtils/DeclNameLocation.swift
@@ -102,6 +102,12 @@ public struct DeclNameLocation: Equatable {
     /// The parameter of a function declaration, like `func foo(a b: Int)`
     case parameters([Argument])
 
+    /// An enum case declaration like `case myCase(label: String)`.
+    ///
+    /// For enum case parameters the argument label is removed when set to empty instead of being changed to eg.
+    /// `myCase(_ label: String)`
+    case enumCaseParameters([Argument])
+
     /// Same as `param` but the parameters can't be collapsed if they are the same. This is the case for subscript
     /// declarations.
     ///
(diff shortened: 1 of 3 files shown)


ahoppen:
https://github.com/apple/swift/pull/74241
https://github.com/apple/sourcekit-lsp/pull/1471

@swift-ci Please test

Sources/SwiftIDEUtils/DeclNameLocation.swift
```diff
@@ -102,6 +102,12 @@ public struct DeclNameLocation: Equatable {
     /// The parameter of a function declaration, like `func foo(a b: Int)`
     case parameters([Argument])
 
+    /// An enum ase declaration like `case myCase(label: String)`.
```
hamishknight, line 105:
```suggestion
    /// An enum case declaration like `case myCase(label: String)`.
```

hamishknight (approved):

Sources/SwiftIDEUtils/DeclNameLocation.swift
ahoppen: I swear I made this type 10 times while writing the PR. One of them seems to have slipped through 😆

ahoppen:
https://github.com/apple/swift/pull/74241
https://github.com/apple/sourcekit-lsp/pull/1471

@swift-ci Please test

ahoppen:
https://github.com/apple/swift/pull/74241
https://github.com/apple/sourcekit-lsp/pull/1471

@swift-ci Please test Windows

PR #2679, opened 2024-06-09
Mark MacroSpec as Sendable

j-f1:
Without making this change, compiling the default `XCTestCase` created for a macro project under Swift 6 will produce the error “Let `testMacros` is not concurrency-safe because non-`Sendable` type `[String : MacroSpec]` may have shared mutable state.” Fortunately there is no shared mutable state here — syntax nodes are already `Sendable`, and metatypes [are also `Sendable`](https://github.com/apple/swift-evolution/blob/main/proposals/0302-concurrent-value-and-concurrent-closures.md#metatype-conformance-to-sendable).


j-f1:
@swift-ci please test

ahoppen (approved):
Thank you!

ahoppen:
Would you mind also cherry-picking this change to `release/6.0`? https://github.com/apple/swift-syntax/blob/main/CONTRIBUTING.md#opening-a-pr-for-release-branch

PR #2689, opened 2024-06-18
Expand trailing comma implementation

mateusrodriguesxyz:
This continues the work that I began in https://github.com/apple/swift-syntax/pull/2515#event-12563352688. After feedback from the language steering group I have expanded the original proposal to allow trailing comma in more places. See the updated proposal text in https://github.com/swiftlang/swift-evolution/pull/2344.

### DIFF
--- a/Sources/SwiftParser/Attributes.swift
+++ b/Sources/SwiftParser/Attributes.swift
@@ -378,7 +378,7 @@ extension Parser {
       return parseAttribute(argumentMode: .customAttribute) { parser in
         let arguments = parser.parseArgumentListElements(
           pattern: .none,
-          allowTrailingComma: parser.experimentalFeatures.contains(.trailingComma)
+          allowTrailingComma: true
         )
         return .argumentList(RawLabeledExprListSyntax(elements: arguments, arena: parser.arena))
       }
--- a/Sources/SwiftParser/Attributes.swift
+++ b/Sources/SwiftParser/Attributes.swift
@@ -376,7 +376,10 @@ extension Parser {
       }
     case nil:
       return parseAttribute(argumentMode: .customAttribute) { parser in
-        let arguments = parser.parseArgumentListElements(pattern: .none, allowTrailingComma: false)
+        let arguments = parser.parseArgumentListElements(
+          pattern: .none,
+          allowTrailingComma: parser.experimentalFeatures.contains(.trailingComma)
+        )
         return .argumentList(RawLabeledExprListSyntax(elements: arguments, arena: parser.arena))
       }
     }
--- a/Sources/SwiftParser/Availability.swift
+++ b/Sources/SwiftParser/Availability.swift
@@ -47,10 +47,15 @@ extension Parser {
             arena: self.arena
           )
         )
-      } while keepGoing != nil && self.hasProgressed(&availabilityArgumentProgress)
+      } while keepGoing != nil && !self.atAvailabilitySpecListTerminator()
+        && self.hasProgressed(&availabilityArgumentProgress)
     }
 
     return RawAvailabilityArgumentListSyntax(elements: elements, arena: self.arena)
+  }
+
+  mutating func atAvailabilitySpecListTerminator() -> Bool {
+    return self.experimentalFeatures.contains(.trailingComma) && self.at(.rightParen)
   }
 
   enum AvailabilityArgumentKind: TokenSpecSet {
--- a/Sources/SwiftParser/Availability.swift
+++ b/Sources/SwiftParser/Availability.swift
@@ -47,15 +47,11 @@ extension Parser {
             arena: self.arena
           )
         )
-      } while keepGoing != nil && !self.atAvailabilitySpecListTerminator()
+      } while keepGoing != nil
         && self.hasProgressed(&availabilityArgumentProgress)
     }
 
     return RawAvailabilityArgumentListSyntax(elements: elements, arena: self.arena)
-  }
-
-  mutating func atAvailabilitySpecListTerminator() -> Bool {
-    return self.experimentalFeatures.contains(.trailingComma) && self.at(.rightParen)
   }
 
   enum AvailabilityArgumentKind: TokenSpecSet {
--- a/Sources/SwiftParser/Declarations.swift
+++ b/Sources/SwiftParser/Declarations.swift
@@ -492,7 +492,7 @@ extension Parser {
             arena: self.arena
           )
         )
-      } while keepGoing != nil && self.hasProgressed(&loopProgress)
+      } while keepGoing != nil && !atGenericParametersListTerminator() && self.hasProgressed(&loopProgress)
     }
 
     let whereClause: RawGenericWhereClauseSyntax?
@@ -517,6 +517,10 @@ extension Parser {
       rightAngle: rangle,
       arena: self.arena
     )
+  }
+
+  mutating func atGenericParametersListTerminator() -> Bool {
+    return self.experimentalFeatures.contains(.trailingComma) && self.at(prefix: ">")
   }
 
   mutating func parseGenericWhereClause() -> RawGenericWhereClauseSyntax {
@@ -700,7 +704,7 @@ extension Parser {
             arena: self.arena
           )
         )
-      } while keepGoing != nil && self.hasProgressed(&loopProgress)
+      } while keepGoing != nil && !self.atWhereClauseListTerminator() && self.hasProgressed(&loopProgress)
     }
 
     return RawGenericWhereClauseSyntax(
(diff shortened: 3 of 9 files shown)


Sources/SwiftParser/Availability.swift
```diff
@@ -48,11 +48,16 @@ extension Parser {
           )
         )
       } while keepGoing != nil && self.hasProgressed(&availabilityArgumentProgress)
+        && !self.atAvailabilitySpecListTerminator()
```
ahoppen, line 51: I would prefer to keep the `hasProgressed` as the last element in the condition.

Same in the other conditions.

Sources/SwiftParser/Attributes.swift
```diff
@@ -376,7 +376,10 @@ extension Parser {
       }
     case nil:
       return parseAttribute(argumentMode: .customAttribute) { parser in
-        let arguments = parser.parseArgumentListElements(pattern: .none, allowTrailingComma: false)
+        let arguments = parser.parseArgumentListElements(
+          pattern: .none,
+          allowTrailingComma: parser.experimentalFeatures.contains(.trailingComma)
+        )
```
ahoppen, line 382: I think you should probably also allow trailing commas in all the custom attributes above, like `@_backDeployed`, `@differentiable` etc.

Sources/SwiftParser/Declarations.swift
```diff
@@ -700,7 +704,7 @@ extension Parser {
             arena: self.arena
           )
         )
-      } while keepGoing != nil && self.hasProgressed(&loopProgress)
+      } while keepGoing != nil && self.hasProgressed(&loopProgress) && !self.atWhereClauseListTerminator()
```
ahoppen, line 707: Generic where clauses don’t necessarily need to be followed by a brace. For example you can have the following, which is valid.

```swift
protocol Foo {
  func test<T>(abc: T) where T: Comparable
}
```

Because of the lack of a proper terminated for the where classes, I’m actually not sure if we should allow trailing commas here. Checking if the next token is a keyword also doesn’t work because it could be a contextual keyword like `nonisolated` on the following function declaration.

Sources/SwiftParser/Declarations.swift
```diff
@@ -880,6 +888,11 @@ extension Parser {
     )
   }
 
+  mutating func atEnumCaseListTerminator() -> Bool {
+    return self.experimentalFeatures.contains(.trailingComma)
+      && (self.atStartOfLine || self.at(.rightBrace) || self.at(.semicolon))
+  }
+
```
ahoppen, line 895: Wouldn’t this mean that we’d stop parsing the following as valid because we classify the comma as a trailing comma and then mark `bar` as unexpected?

```swift
enum A {
  case foo,
       bar
}
```

Similar to the where clause, I’m not sure if we want to allow a trailing comma here because there is no list terminator.

Sources/SwiftParser/Nominals.swift
```diff
@@ -339,6 +339,10 @@ extension Parser {
     )
   }
 
+  mutating func atInheritanceListTerminator() -> Bool {
+    return self.experimentalFeatures.contains(.trailingComma) && (self.at(.leftBrace) || self.at(.keyword(.where)))
+  }
+
```
ahoppen, line 345: I think this doesn’t work in general because `associatedtype` can also have an inheritance clause and it is not followed by `{` or `where`. 

Eg. 

```swift
associatedtype Iterator: IteratorProtocol, 
```

would probably fail to parse wit this change.

I think we should limit the trailing comma to the types where it is guaranteed to be followed by `where` or `{` (ie. everything except `associatedtype`).

ahoppen (commented):
Thanks for your continued work on this @mateusrodriguesxyz!

Sources/SwiftParser/Declarations.swift
mateusrodriguesxyz: Completely missed this, which is ironic because that's exactly how one would use a list with trailing comma 😅

Sources/SwiftParser/Nominals.swift
mateusrodriguesxyz: @ahoppen I believe it's just a matter of update the proposal text mentioning that trailing comma won't work for inheritance clause of an associatedtype, right? The same for where clauses for protocol definition. In general, make it clear that trailing comma will only be allowed when there's a clear terminator.

ahoppen: I’m only speaking as a reviewer of this PR here but I think it would make sense to exclude trailing commas for these situations without a clear terminator from the evolution proposal for now. 

There might be ways of getting them right, but doing so is more complicated than the cases that have a clear terminator. Highlighting these situations in a *Future directions* section would make sense, I think. That way your proposal can focus on the majority of cases and if we do choose to extend trailing commas eg. `case` lists, the corner cases will get the attention they deserve.

Sources/SwiftParser/Attributes.swift
ahoppen: Thanks for adding them.

Sources/SwiftParser/Nominals.swift
mateusrodriguesxyz: I've updated the proposal text highlighting these cases. Thanks for catching them.

Sources/SwiftParser/Declarations.swift
```diff
@@ -710,6 +714,10 @@ extension Parser {
       arena: self.arena
     )
   }
+
+  mutating func atWhereClauseListTerminator() -> Bool {
+    return self.at(.leftBrace)
+  }
```
ahoppen, line 720: As mentioned in https://github.com/swiftlang/swift-syntax/pull/2689#discussion_r1645084232, should we also disallow trailing commas in `where` clauses?

Sources/SwiftParser/Nominals.swift
```diff
@@ -339,6 +339,10 @@ extension Parser {
     )
   }
 
+  mutating func atInheritanceListTerminator() -> Bool {
+    return self.experimentalFeatures.contains(.trailingComma) && (self.at(.leftBrace) || self.at(.keyword(.where)))
+  }
```
ahoppen, line 344: Am I missing something or do we still allow trailing commas in `associatedtype` declarations even though it’s ambiguous in the current state of your PR?

ahoppen (commented):
Looking good, just two discussions from my last review that I would like to continue.

Sources/SwiftParser/Declarations.swift
mateusrodriguesxyz: @ahoppen My intention was not completely rule out `where` clauses and `associatedtype` because there are valid cases with clear terminator and to let the parser just fail as today when there’s not.

mateusrodriguesxyz: I understand there’s a inconsistency here and I'm fine with going another direction but I was thinking that a user of the language with no knowledge of these ambiguous cases would be confuse with the lack of support in these unambiguous cases so I choose the "do the best we can" approach.

Sources/SwiftParser/Nominals.swift
mateusrodriguesxyz: I can't find a example that my implementation would allow trailing commas in `associatedtype`. Do you have you that you think it would?

mateusrodriguesxyz: Indeed it would still parse in the following case because there's a `where` terminator:

```swift
protocol Foo {
    associatedtype T: P1, P2, where P1: Equatable, P2: Equatable
}
```

Sources/SwiftParser/Declarations.swift
mateusrodriguesxyz: It will fail in a where clause for function in a protocol declaration:

```swift
protocol Foo {
  func f<T1, T2>(a: T1, b: T2) where T1: P1, T2: P2, // ❌ Expected type
}
```

But these are more cases with a clear terminator and the parser can handle without ambiguity:

```swift
struct Foo: P1, P2, where P1: Equatable, P2: Equatable, { }

func foo<T1, T2,>() where T1: Equatable, T2: Equatable, { }
      
extension Foo where T1: Equatable, T2: Equatable, { }
```

So there is a inconsistency for `where` clause but we would consistently allow trailing comma for lists with a clear terminator throughout the language.

ahoppen: Ah, thanks for the explanation. Makes sense 👍🏽

ahoppen (approved):

mateusrodriguesxyz:
@ahoppen I've updated the PR to match https://github.com/swiftlang/swift/pull/74522.

ahoppen (approved):

mateusrodriguesxyz:
@ahoppen Don't you prefer that I squash the commits before merging?

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
Could you run swift-format on your changes?

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

Issue #2687, opened 2024-06-17
trimming trivia nodes corrupts source ranges


tayloraswift:
### Description

using the [`trimmed`](https://swiftinit.org/docs/swift-syntax/swiftsyntax/syntaxprotocol.trimmed) family of `SyntaxProtocol` methods corrupts the `ByteSourceRange`s when obtaining syntax classifications.

i was able to reproduce this on `510.0.2`, and on `600.0.0-prerelease-2024-06-12`.

### Steps to Reproduce

```swift
import SwiftIDEUtils
import SwiftParser
import SwiftSyntax

let text:String = """
@_documentation(metadata: blah)
public mutating
func f() {}
"""
let utf8:[UInt8] = .init(text.utf8)

var parser:Parser = utf8.withUnsafeBufferPointer { .init($0) }
let decl:DeclSyntax = .parse(from: &parser)

let function:FunctionDeclSyntax = decl.as(FunctionDeclSyntax.self)!

for span:SyntaxClassifiedRange in function.modifiers.trimmed.classifications
{
    let range:Range<Int> = span.range.offset ..< span.range.offset + span.range.length
    print(String.init(decoding: utf8[range], as: Unicode.UTF8.self))
}
```

```

publi
c
 mutatin
```

ahoppen:
This behaves as intended. Running `trimmed` modifies the syntax tree to remove the leading and trailing trivia of the first and last token, respectively. The trimmed node should probably be detached from the original tree, which would make the behavior more obvious.

I would suggest that you do something like the following

```swift
let trimmed = function.modifiers.detached.trimmed
let trimmedBytes = trimmed.syntaxTextBytes
for span in trimmed.classifications {
    let range = span.range.offset ..< span.range.offset + span.range.length
    print(String.init(decoding: syntaxTextBytes[range], as: Unicode.UTF8.self))
}
```

PR #2690, opened 2024-06-18
Detach node when accessing .trimmed

ahoppen:
Having the trimmed node be attached to a parent was not intuitive because eg. printing the parent node would have the trimmed trivia missing, most likely forming invalid Swift code.

Motivated by https://github.com/apple/swift-syntax/issues/2687

### DIFF
--- a/Release Notes/601.md
+++ b/Release Notes/601.md
@@ -9,6 +9,10 @@
 
 ## API Behavior Changes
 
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

hamishknight (approved):

rintaro (approved):

PR #2695, opened 2024-06-23
Properly deprecate default implementation of `MacroExpansionContext.lexicalContext`

ahoppen:
Now that we have rdar://123403268, we can change the default implementation of `MacroExpansionContext.lexicalContext` to return an empty array and be deprecated, making not implementing `lexicalContext` a compile-time warning instead of a runtime error.

rdar://123410459


ahoppen:
@swift-ci Please test

PR #2696, opened 2024-06-23
[SwiftLexicalScopes][GSoC] Add simple scope queries and initial name lookup API structure

MAJKFL:
This is an initial PR for the `SwiftLexicalScopes`, developed as one of the GSoC 2024 projects.

In this PR:
- Introduced three basic queries `lookupLabeledStmts`, `lookupFallthroughSourceAndDest` and `lookupCatchNode`.
- Added test functions that will guide development of the full scope name lookup.
- Created initial structure of the lexical scope API hierarchy with `FileScope` as the root.

This is my first contribution to the project, so I would greatly appreciate if someone could double-check everything. Thank you!


ahoppen (commented):
Wohooo. Exciting to see this happen. I’ve left a few comments inline. The overall direction that we’re going in is great. And thanks for all the test cases.

Sources/SwiftLexicalScopes/Scope.swift
```diff
@@ -0,0 +1,139 @@
...
+    }
+  }
+}
+
+protocol Scope {
+  var parent: Scope? { get }
+
+  var sourceSyntax: SyntaxProtocol { get }
+
+  func getDeclarationFor(name: String, at syntax: SyntaxProtocol) -> Syntax?
```
CodaFi, line 33: Perhaps we should call this `lookup` and have it return an array of `DeclSyntax` nodes. It is possible to shadow declarations across Scopes in Swift, and I would expect a lookup in a nested scope to be able to see both declarations. That makes it the client's responsibility to overlay a tie-breaking policy (e.g. deepest decl wins)

CodaFi:
I think that this API would emphasize syntax over scope and add entrypoints as such.

My suggestion would be

```
protocol ScopeSyntax: SyntaxProtocol {
  func parent() -> (any ScopeSyntax)?
  func lookup(_ name: String) -> [DeclSyntax]
}

extension SourceFileSyntax: ScopeSyntax { /**/ }
extension CodeBlockSyntax: ScopeSyntax { /**/ }
/*etc*/
```

Internally, we can project a `Scope` tree over the syntax node and perform the lookups against that tree. But I'm not sure it's a good idea to expose the scope tree to clients.

Tests/SwiftLexicalScopesTest/SimpleQueryTests.swift
```diff
@@ -0,0 +1,131 @@
...
+              3️⃣break one
+            } while true
+          }
+          4️⃣break
+        }
+        """,
+      methodUnderTest: { argument in
+        LexicalScopes.lookupLabeledStmts(at: argument)
+      },
+      expected: ["3️⃣": ["2️⃣", "1️⃣"], "4️⃣": ["1️⃣"], "5️⃣": []]
```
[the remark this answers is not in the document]
MAJKFL: The function is supposed to return all the labeled statements accessible at the position of a given node so in the case of the while, which is unlabeled, there aren’t available any labeled statements. The function just uses the `break` location to start lookup. Maybe it was a bit confusing to put unlabeled `break` in the test?

Sources/SwiftLexicalScopes/Scope.swift
```diff
@@ -0,0 +1,139 @@
...
+    var visitedOriginalCase = false
+
+    for child in switchCaseListSyntax.children(viewMode: .sourceAccurate) {
+      if let thisCase = child.as(SwitchCaseSyntax.self) {
+        if thisCase.id == switchCaseSyntax.id {
+          visitedOriginalCase = true
+        } else if visitedOriginalCase {
+          return thisCase
+        }
+      }
```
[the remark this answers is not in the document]
MAJKFL: Do you think it would be possible to e.g. evaluate the `#if` clauses and then perform the lookup on the resulting source code?

Sources/SwiftLexicalScopes/Scope.swift
```diff
@@ -0,0 +1,139 @@
...
+
+  var sourceSyntax: SyntaxProtocol { get }
+
+  func getDeclarationFor(name: String, at syntax: SyntaxProtocol) -> Syntax?
+}
+
+extension Scope {
+  var parent: Scope? {
+    getParentScope(forSyntax: sourceSyntax)
+  }
```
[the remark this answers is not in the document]
MAJKFL: Yeah, I think this is a good point. As I now think about it, a great part of scopes will have a different parent behavior (like possibly `GenericParameterScope`). I made it a requirement to implement `parent` by all subclasses, but I will also leave `getParentScope` method so that subclasses can still use the default behavior.

Sources/SwiftLexicalScopes/LexicalScopes.swift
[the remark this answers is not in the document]
MAJKFL: These 3 methods are reimplementations of methods already present in the compiler scope implementation: [`lookupLabeledStmts`](https://github.com/swiftlang/swift/blob/main/include/swift/AST/ASTScope.h#L294-L296), [`lookupFallthroughSourceAndDest`](https://github.com/swiftlang/swift/blob/main/include/swift/AST/ASTScope.h#L298-L299C3) and [`lookupCatchNode`](https://github.com/swiftlang/swift/blob/main/include/swift/AST/ASTScope.h#L305). @DougGregor suggested to start with these before going to more advanced scope name lookup.

I like also the idea of these more expressive functions though. Would it be possible to maybe somehow separate the API used by the compiler and the API used by other clients? Or maybe we could merge them and have as one? How much could we alter the functions used currently by the compiler?

MAJKFL:
> I think that this API would emphasize syntax over scope and add entrypoints as such.
> 
> My suggestion would be
> 
> ```
> protocol ScopeSyntax: SyntaxProtocol {
>   func parent() -> (any ScopeSyntax)?
>   func lookup(_ name: String) -> [DeclSyntax]
> }
> 
> extension SourceFileSyntax: ScopeSyntax { /**/ }
> extension CodeBlockSyntax: ScopeSyntax { /**/ }
> /*etc*/
> ```
> 
> Internally, we can project a `Scope` tree over the syntax node and perform the lookups against that tree. But I'm not sure it's a good idea to expose the scope tree to clients.

I think it would be a great idea to make syntax in the center of those APIs, but I’m not entirely sure how would this work with more complex scopes like with function declarations. If we’d like to project a scope onto the `FunctionDeclSyntax` we’d have to include some more complex logic handling name lookup for e.g. generic parameters.

Compiler right now for this invalid snippet:
```
func foo<T1, T2: T1>(a: T1, b: T2) {
    print(a, b)
}
```
Produces this scope map:
```
***Complete scope map***
ASTSourceFileScope 0x14d13d000, [1:1 - 4:1] 'test.swift'
`-AbstractFunctionDeclScope 0x14d170020, [1:1 - 3:1] 'foo(a:b:)'
  `-GenericParamScope 0x14d139378, [1:1 - 3:1] param 0 'T1'
    `-GenericParamScope 0x14d139378, [1:1 - 3:1] param 1 'T2 : T1'
      |-ParameterListScope 0x14d139580, [1:21 - 1:34]
      `-FunctionBodyScope 0x14d170020, [1:36 - 3:1]
        `-BraceStmtScope 0x14d1701e8, [1:36 - 3:1]
```
The parameter T1 [should be recognized during lookup as the first generic parameter](https://github.com/swiftlang/swift/blob/5bbf6d05f320e9e83939952557433abfbe1e1b88/include/swift/AST/ASTScope.h#L755).

I assume that’s why there are separate `AbstractFunctionDeclScope` and `FunctionBodyScope` so there’s a clear path for recursive lookup i.e. `BraceStmtScope` -> `FunctionBodyScope` -> `GenericParamScope` -> `GenericParamScope` -> `AbstractFunctionScope` -> …. If we’d consolidate this functionality into one `SyntaxScope` this would require some complex logic for handling lookup at generic parameters differently than in e.g. function body. I started prototyping how a structure resembling the compiler implementation could work in this case with inner (`FunctionBodyScope`) and outer (`AbstractFunctionDeclScope`) scopes both associated to `FunctionDeclSyntax` [here](https://github.com/MAJKFL/swift-syntax/blob/function-scope/Sources/SwiftLexicalScopes/ScopeConcreteImplementations.swift). 

What about a hybrid approach? We could still use the custom structure mapped over the existing syntax nodes internally and additionally introduce a new ScopeSyntax protocol that would add entry points for the API at the nodes themselves (that is move the functionality of `LexicalScopes` into the syntax nodes). We would retain the flexible scope structure and still emphasize syntax nodes as the entry points for the API.

It’s possible I’m wrong and I missed some detail so I’d appreciate to hear your take on this. I’d also like to hear from @DougGregor what he thinks about it.

Sources/SwiftLexicalScopes/LexicalScopes.swift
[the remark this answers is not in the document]
ahoppen: I am not too sure what exactly the compiler needs for name lookup, maybe the compiler could also be simplified by migrating to more expressive functions.

What we could do for now is to mark these functions as `@_spi(Compiler) public` and/or `@_spi(Testing) public`. That way they don’t become part of swift-syntax’s public API, which we want to keep stable but at the same time we can iterate on it.

Sources/SwiftLexicalScopes/Scope.swift
```diff
@@ -0,0 +1,139 @@
...
+
+  // MARK: - lookupLabeledStmts
+
+  func lookupLabeledStmts(at syntax: SyntaxProtocol) -> [LabeledStmtSyntax] {
+    var result = [LabeledStmtSyntax]()
+    lookupLabeledStmtsHelper(at: syntax.parent, accumulator: &result)
+    return result
+  }
+
+  private func lookupLabeledStmtsHelper(at syntax: Syntax?, accumulator: inout [LabeledStmtSyntax]) {
```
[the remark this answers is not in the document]
ahoppen: Which altered implementation are you referring to? I generally try to start with the cleanest implementation and only optimize for performance if we see that the function is a performance bottleneck and can measure that eg. using `inout` actually improves performance. I find it generally helps to keep the code cleaner.

Sources/SwiftLexicalScopes/Scope.swift
```diff
@@ -0,0 +1,139 @@
...
+      accumulator.append(labeledStmtSyntax)
+      lookupLabeledStmtsHelper(at: labeledStmtSyntax.parent, accumulator: &accumulator)
+    } else {
+      lookupLabeledStmtsHelper(at: syntax.parent, accumulator: &accumulator)
+    }
+  }
+
+  // MARK: - lookupFallthroughSourceAndDest
+
+  func lookupFallthroughSourceAndDest(at syntax: SyntaxProtocol) -> (SwitchCaseSyntax?, SwitchCaseSyntax?) {
```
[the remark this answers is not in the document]
ahoppen: Ah, thanks for pointing me to it. 👍🏽

Sources/SwiftLexicalScopes/Scope.swift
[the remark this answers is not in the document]
ahoppen: @DougGregor has a PR up to evaluate `#if` in swift-syntax but I think it’s not quite ready yet. https://github.com/swiftlang/swift-syntax/pull/1816

Let’s wait until Doug is back to discuss how we want to approach this.

Tests/SwiftLexicalScopesTest/SimpleQueryTests.swift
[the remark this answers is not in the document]
ahoppen: Ah, I see, makes sense. I just thought that it would return the location that the `break` would jump to or something like that. But I might also just not have read the test properly.

Sources/SwiftLexicalScopes/Scope.swift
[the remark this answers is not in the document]
MAJKFL: Sorry, I should’ve linked change from the last commit I made.

The [altered solution](https://github.com/swiftlang/swift-syntax/blob/211cb6f90188289bc058aa220a7d9957ec8e19dd/Sources/SwiftLexicalScopes/Scope.swift#L58) builds up the accumulator and then, after finishing lookup, returns it as a result.

```Swift
private func lookupLabeledStmtsHelper(at syntax: Syntax?, accumulator: [LabeledStmtSyntax]) -> [LabeledStmtSyntax] {
  guard let syntax, !syntax.is(MemberBlockSyntax.self) else { return accumulator }
  if let labeledStmtSyntax = syntax.as(LabeledStmtSyntax.self) {
    return lookupLabeledStmtsHelper(at: labeledStmtSyntax.parent, accumulator: accumulator + [labeledStmtSyntax])
  } else {
    return lookupLabeledStmtsHelper(at: syntax.parent, accumulator: accumulator)
  }
}
```

Do you suggest to further simplify it to just building up the return value?

```Swift
private func lookupLabeledStmtsHelper(at syntax: Syntax?) -> [LabeledStmtSyntax] {
  guard let syntax, !syntax.is(MemberBlockSyntax.self) else { return [] }
  if let labeledStmtSyntax = syntax.as(LabeledStmtSyntax.self) {
    return [labeledStmtSyntax] + lookupLabeledStmtsHelper(at: labeledStmtSyntax.parent)
  } else {
    return lookupLabeledStmtsHelper(at: syntax.parent)
  }
}
```

Sources/SwiftLexicalScopes/LexicalScopes.swift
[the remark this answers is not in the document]
MAJKFL: Ok, as for now I marked the functions as `@_spi(Compiler) @_spi(Testing) public`.

Sources/SwiftLexicalScopes/Scope.swift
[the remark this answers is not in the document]
ahoppen: Yes, I was suggesting to use the return value variant. Sometimes it’s just me but I find it easier to reason about.

CodaFi:
> I assume that’s why there are separate AbstractFunctionDeclScope and FunctionBodyScope so there’s a clear path for recursive lookup i.e. BraceStmtScope -> FunctionBodyScope -> GenericParamScope -> GenericParamScope -> AbstractFunctionScope ->

This is rather because ASTScope is designed in a decidedly much more object-oriented fashion than the rest of the compiler. I consider this to be an abstraction leak of the tree, but the structure of the model is nonetheless sound to your point. My preference is still for a Syntax-oriented API.

> What about a hybrid approach?

I agree! Consider the following:

```
protocol ScopeSyntax: SyntaxProtocol {
  var scope: Scope { get } // project this node into a `Scope`
  func parent() -> (any ScopeSyntax)?
  func lookup(_ name: String, in scope: Scope) -> [DeclSyntax]
}
```

That would handle a client that e.g. wants to perform unqualified lookup for a generic parameter by 
1. Walking from the FunctionDeclSyntax node to the GenericParameterListSyntax node
2. Projecting a `Scope`
3. Calling `functionDecl.lookup("x", in: genericParameterList.scope)`

CodaFi:
We also have to be careful here to look at ASTScope as a model because its projection function is top-down (root to leaf), where as a `Scope` is (presumably) projected bottom-up (leaf to root). There are interesting things to consider in the bottom-up case like 

```
guard let x = foo() else {
  return bar
}
```

Projecting the scope of the `else` brace statement has to be careful to walk through the scopes defined in its binding condition list and demand the scope of the next-most syntactic enclosing scope. Another way to put it - you could implement bottom-up scope projection by linking the `else` branch to its binders and marking them as "holes" that a recursive lookup procedure should merely traverse but not collect declarations from.

CodaFi:
Another passing thought: If a `Scope` is sufficiently opaque, you can avoid allocating a `Scope` tree at all/modeling it with separate `class` references by having the `parent` node point to a `ScopeSyntax`. A `Scope` would then just be a value-typed handle returned as the result of projecting up the `ScopeSyntax` parts of one particular branching path of the syntax tree

MAJKFL:
> Another passing thought: If a `Scope` is sufficiently opaque, you can avoid allocating a `Scope` tree at all/modeling it with separate `class` references by having the `parent` node point to a `ScopeSyntax`.

I tried [remodeling the name lookup with shadowing](https://github.com/MAJKFL/swift-syntax/blob/function-scope-new-api-structure/Sources/SwiftLexicalScopes/LookupImplementation.swift) for function declaration with just `ScopeSyntax` nodes and I think it could be a good idea to follow this lead. Specifying lookup caller fixes the problem with name lookup at generic parameters and function parameter list. Additionally, it could also potentially serve as a good starting point for lookup in `CodeBlockSyntax` to filter the declarations according to their relative position.

> As for capturing lookup from a particular part of the generic parameter list, you could implement the projection of the syntax for a generic parameter to a Scope by storing an index into the enclosing generic parameter list and the generic parameter list itself.

I tried a bit different approach with generic parameter scopes. I followed the hierarchical structure of `ASTScope` to make the API more uniform across the lookup and not worry about keeping the indexes of generic parameters. This way, each `GenericParameterSyntax` implements `ScopeSyntax`. On the other hand, it complicates a bit more the lookup on the `FunctionDeclSyntax` side, but I think it’s a great tradeoff.

Also, as for now, I’m just returning the values as `TokenSyntax` nodes. Since the lookup could also result in variable declarations, function parameters, function name, generic parameters etc. Is there a way to represent them uniformly as `DeclSyntax` as you suggested?

I’d love to hear your feedback @CodaFi about the [adjusted structure](https://github.com/MAJKFL/swift-syntax/blob/function-scope-new-api-structure/Sources/SwiftLexicalScopes/ScopeSyntax.swift). I tried to follow your suggestions with these two adjustments, so I hope I haven’t missed anything.

Sources/SwiftLexicalScopes/Scope.swift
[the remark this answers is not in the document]
DougGregor: I had in fact not considered how to handle `#if` clauses. I don't want to rely on any kind of syntactic preprocessing of the syntax tree to remove the `#if`s: we need this query to work on the original, unmodified syntax tree. 

So, it seems like we will need something like #1816 to have any chance of making `lookupNextSwitchCase(at:)` work when there's an `#if` around the case. When we do get #1816, we'll need to parameterize this operation by a `BuildConfiguration`. I'm okay with waiting until #1816 lands to make that change, and just... pick the first one or something... if there are `#if`s around the following case.

Sources/SwiftLexicalScopes/Scope.swift
MAJKFL: Oh yeah, I haven’t thought about this case. You’re right, thanks for pointing this out. I adjusted the `Scope` protocol accordingly.

Sources/SwiftLexicalScopes/Scope.swift
```diff
@@ -0,0 +1,135 @@
...
+  // MARK: - lookupLabeledStmts
+
+  /// Given syntax node position, returns all available labeled statements.
+  func lookupLabeledStmts(at syntax: SyntaxProtocol) -> [LabeledStmtSyntax] {
+    return lookupLabeledStmtsHelper(at: syntax.parent)
+  }
+
+  /// Helper method to recursively collect labeled statements from the syntax node's parents.
+  private func lookupLabeledStmtsHelper(at syntax: Syntax?) -> [LabeledStmtSyntax] {
+    guard let syntax, !syntax.is(MemberBlockSyntax.self) else { return [] }
```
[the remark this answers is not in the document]
MAJKFL: Thanks for pointing this out. I added them with respective test cases.

Sources/SwiftLexicalScopes/Scope.swift
```diff
@@ -0,0 +1,135 @@
...
+    } else {
+      return lookupLabeledStmtsHelper(at: syntax.parent)
+    }
+  }
+
+  // MARK: - lookupFallthroughSourceAndDest
+
+  /// Given syntax node position, returns the current switch case and it's fallthrough destination.
+  func lookupFallthroughSourceAndDestination(at syntax: SyntaxProtocol) -> (SwitchCaseSyntax?, SwitchCaseSyntax?) {
+    guard let originalSwitchCase = syntax.ancestorOrSelf(mapping: { $0.as(SwitchCaseSyntax.self) }) else {
```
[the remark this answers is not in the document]
MAJKFL: I added a generic helper method [`walkParentTree`](https://github.com/swiftlang/swift-syntax/blob/74ddcc47c15ecdc750fe4f1806e20581b9cc77d6/Sources/SwiftLexicalScopes/Scope.swift#L142-L157). It takes an array of syntax node types the lookup should terminate at and specifies type of nodes to collect to make it more reusable in the future. We can then use it to create more specialized functions like [`walkParentTreeUpToFunctionBoundary`](https://github.com/swiftlang/swift-syntax/blob/74ddcc47c15ecdc750fe4f1806e20581b9cc77d6/Sources/SwiftLexicalScopes/Scope.swift#L129-L140).

Sources/SwiftLexicalScopes/Scope.swift
```diff
@@ -0,0 +1,135 @@
...
+      }
+    case .catchClause:
+      return lookupCatchNodeHelper(at: syntax.parent, traversedCatchClause: true)
+    case .tryExpr(let tryExpr):
+      if tryExpr.questionOrExclamationMark != nil {
+        return syntax
+      } else {
+        return lookupCatchNodeHelper(at: syntax.parent, traversedCatchClause: traversedCatchClause)
+      }
+    case .functionDecl(let functionDecl):
```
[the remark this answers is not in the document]
MAJKFL: Thanks for the suggestion. I added those cases. Should we also include source file? Does compiler expect that as a result when we have a `try` in top level code?

Sources/SwiftLexicalScopes/Scope.swift
```diff
@@ -0,0 +1,135 @@
...
+      if functionDecl.signature.effectSpecifiers?.throwsClause != nil {
+        return syntax
```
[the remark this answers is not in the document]
MAJKFL: Thanks for letting me know. I removed the condition and adjusted one of the tests.

Sources/SwiftLexicalScopes/LexicalScopes.swift
```diff
@@ -0,0 +1,42 @@
...
+
+import Foundation
+import SwiftSyntax
+
+public enum LexicalScopes {
+  /// Given syntax node position, returns all available labeled statements.
+  @_spi(Compiler) @_spi(Testing) public static func lookupLabeledStmts(at syntax: SyntaxProtocol) -> [LabeledStmtSyntax] {
+    guard let scope = syntax.scope else { return [] }
+    return scope.lookupLabeledStmts(at: syntax)
+  }
```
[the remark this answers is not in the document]
MAJKFL: I was not entirely sure how much I can adjust the functions already used by the compiler, that’s why I put it in `LexicalScopes` with similar parameter that indicated location to start lookup. I moved them now to `SyntaxProtocol` so clients can just use `syntaxNode.lookupLabeledStmts()`.

Sources/SwiftLexicalScopes/Scope.swift
```diff
@@ -0,0 +1,185 @@
...
+}
+
+/// Provide common functionality for specialized scope implementatations.
+protocol Scope {
+  /// The parent of this scope.
+  var parent: Scope? { get }
+
+  /// Syntax node that introduces this protocol.
+  var sourceSyntax: SyntaxProtocol { get }
+}
```
DougGregor, line 35: As far as I can tell, `Scope` isn't doing anything any more. Can we remove it, and put the various operations on `SyntaxProtocol` directly?

Package.swift
```diff
@@ -28,6 +28,7 @@ let package = Package(
     .library(name: "SwiftSyntaxMacros", targets: ["SwiftSyntaxMacros"]),
     .library(name: "SwiftSyntaxMacroExpansion", targets: ["SwiftSyntaxMacroExpansion"]),
     .library(name: "SwiftSyntaxMacrosTestSupport", targets: ["SwiftSyntaxMacrosTestSupport"]),
+    .library(name: "SwiftLexicalScopes", targets: ["SwiftLexicalScopes"]),
```
DougGregor, line 31: We should talk about the name here a little more. I think I suggested `SwiftLexicalScopes` initially, but given that the "scope" part is actually getting somewhat downplayed... should we call this `SwiftLexicalLookup`? All of these operations are, essentially, lexical lookups for various kinds of things---lookups for declarations, labeled statements, where a `fallthrough` ends up---and most will have "lookup" in their names.

Tests/SwiftLexicalScopesTest/SimpleQueryTests.swift
```diff
@@ -0,0 +1,179 @@
...
+    )
+  }
+
+  func testLookupCatchNodeWithNestedDoCatch() {
+    assertLexicalScopeQuery(
+      source: """
+        1️⃣func foo() rethrows {
+          2️⃣do {
+            3️⃣do {
+              try 4️⃣f()
```
DougGregor, line 163: It's probably worth testing `try x + y` and `try! x + y`, where we look up from the `y`, to make sure we get the appropriate catch node. We just tripped over this [same issue](https://github.com/swiftlang/swift/pull/74454/commits/8d6434410b591cbf6f8cfc0c729ab27e8aa15ebc) with the C++ implementation .

Sources/SwiftLexicalScopes/Scope.swift
MAJKFL: It was a leftover from the original API structure I wanted to implement. I moved now all the query methods to `SyntaxProtocol` and removed parts related to the `Scope` protocol.

Package.swift
MAJKFL: Yeah, I also like the idea. I changed it to `SwiftLexicalLookup`. It’s more broad and not limiting the library to just scope lookup could pave the way for including a broader range of APIs, like the `break` and `continue` lookup you’ve mentioned yesterday that is not necessarily part of `ASTScope` as far as I remember. Maybe it would be a good idea to use this library in the long run to consolidate this sort of logic in one codebase?

DougGregor: Yes, I think this is the right place to consolidate all of the lexical lookup logic over time.

Sources/SwiftLexicalLookup/SimpleLookupQueries.swift
```diff
@@ -0,0 +1,176 @@
...
+      at: syntax,
+      collect: collect,
+      stopWithFirstMatch: stopWithFirstMatch
+    )
+  }
+
+  /// Callect syntax nodes matching the collection type up until encountering one of the specified syntax nodes.
+  private func walkParentTree<T: SyntaxProtocol>(
+    upTo stopAt: [SyntaxProtocol.Type],
+    at syntax: Syntax?,
```
DougGregor, line 150: I think this is a vestige of the initial placement of these functions on `Scope`, because in all of these functions, `self` is unused. They're following the path up the syntax tree via the `syntax` parameter instead. I think we can simplify this by removing the `at syntax` parameter and replacing recursion with iteration.

Sources/SwiftLexicalLookup/SimpleLookupQueries.swift
```diff
@@ -0,0 +1,176 @@
...
+      if stopWithFirstMatch {
+        return [matchedSyntax]
+      } else {
+        return [matchedSyntax]
+          + walkParentTree(
+            upTo: stopAt,
+            at: syntax.parent,
+            collect: collect,
+            stopWithFirstMatch: stopWithFirstMatch
+          )
```
DougGregor, line 165: FWIW, building up an array via `+` here is going to exhibit quadratic behavior, because we need to create a temporary array at each step. I think this recursive function can be replaced with an iterative one that will be more efficient, e.g.,

```swift
var matches: [T] = []
var nextSyntax: Syntax? = syntax // when 'syntax' goes away, replace with Syntax(self)
while let currentSyntax = nextSyntax {
  if stopAt.contains(where: { currentSyntax($0) }) {
    break
  }

  if let matchedSyntax = currentSyntax.as(T.Self) {
    matches.append(matchedSyntax)
    if stopWithFirstMatch {
      break
    }
  }

 nextSyntax = currentSyntax.parent
}
 
return matches
```

Sources/SwiftLexicalLookup/SimpleLookupQueries.swift
```diff
@@ -0,0 +1,176 @@
...
+import SwiftSyntax
+
+extension SyntaxProtocol {
+  /// Given syntax node position, returns all available labeled statements.
+  @_spi(Compiler) @_spi(Testing) public func lookupLabeledStmts() -> [LabeledStmtSyntax] {
+    return lookupLabeledStmts(at: self)
+  }
+
+  /// Given syntax node position, returns the current switch case and it's fallthrough destination.
+  @_spi(Compiler) @_spi(Testing) public func lookupFallthroughSourceAndDest()
```
DougGregor, line 23: We can give this an `@_spi` because we're still working on it, but I would like us to treat it like it's going to become real API (and not just for the compiler). To that end, I think we should replace the `@_spi(Compiler) @_spi(Testing)` on these with one `@_spi(Experimental)` that we can remove later.

Additionally, let's document these entry points a bit more thoroughly here, so that when we drop the "experimental" we won't have to come back and improve things. For example, the goal of this function is to tell us, for a `fallthrough` statement, the case that it's coming from and the case it's going to. The documentation should provide a short code snippet showing the kind of code it's expected to be used on, e.g.,

```swift
switch value {
  case 2:
    doSomething()
    fallthrough
  case 1:
    doSomethingElse()
  default:
    break
}
```

When calling on that `fallthrough`, we'll get back the syntax nodes for `case 2` and `case 1`, in that order. The `nil` results are there to handle ill-formed code: there's no `source` if the fallthrough is outside of a case. There's no `destination` if there is no case or default after the `source`. We'll help our future users considerably if we write this out, and also help prove to ourselves that the API is doing what we need.

Which brings up a question: should this API only be available on `FallthroughStmtSyntax` rather than on `SyntaxProtocol`? It's a bit unlike the other two queries in that it's really for exactly one statement node.

DougGregor (commented):
Thank you! We are very, very close, but I'd like to polish this up just a little bit more before we land this PR.

Sources/SwiftLexicalLookup/SimpleLookupQueries.swift
MAJKFL: Initially, I considered [implementing the method with an `inout` collector](https://github.com/swiftlang/swift-syntax/blob/9f2fab0a6d401f4ba2f7226bc2440b0d80914a15/Sources/SwiftLexicalScopes/Scope.swift#L57-L65) to benefit from tail recursion, but we scrapped that for simplicity. I think it’s a good idea to remove the recursion altogether for simplicity though. I replaced the implementation with your suggestion. Thanks!

Sources/SwiftLexicalLookup/SimpleLookupQueries.swift
MAJKFL: That’s a good point. I added documentation to the three methods. I hope it's sufficient sufficient. I also moved `lookupFallthroughSourceAndDest` to `FallthroughStmtSyntax`. I also think it makes more sense, since it’s quite specific for that one particular node.

DougGregor: All of these changes look great!

DougGregor:
@MAJKFL looks like Package.swift has a conflict. If you're able to resolve that and run formatting locally (if you haven't already), I think we're ready to kick off testing

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test Windows

DougGregor:
macOS failure is this:

```swift
Error: .spi.yml did not contain the same libraries as Package.swift:
 - Missing in .spi.yml: SwiftLexicalLookup
```

DougGregor:
@swift-ci please test

MAJKFL:
> macOS failure is this:
> 
> ```swift
> Error: .spi.yml did not contain the same libraries as Package.swift:
>  - Missing in .spi.yml: SwiftLexicalLookup
> ```

I added `SwiftLexicalLookup` to the `.spi.yml` file. I hope it's fixed now.

DougGregor:
@swift-ci please test

DougGregor:
That looks right to me. Kicked off CI and let's hope for the green checks

DougGregor:
Feel free to merge if it passes

Sources/SwiftLexicalLookup/SimpleLookupQueries.swift
```diff
@@ -0,0 +1,227 @@
...
+
+  // MARK: - walkParentTree helper methods
+
+  /// Callect the first syntax node matching the collection type up to a function boundary.
+  fileprivate func walkParentTreeUpToFunctionBoundary<T: SyntaxProtocol>(
+    at syntax: Syntax?,
+    collect: T.Type
+  ) -> T? {
+    walkParentTreeUpToFunctionBoundary(at: syntax, collect: collect, stopWithFirstMatch: true).first
+  }
```
ahoppen, line 113: This isn’t really collecting anything, so I wouldn’t talk about `collect` here.

```suggestion
  /// Returns the innermost node of the specified type up to a function boundary.
  fileprivate func innermostNodeOfTypeUpToFunctionBoundary<T: SyntaxProtocol>(
    at syntax: Syntax?,
    _ type: T.Type
  ) -> T? {
    walkParentTreeUpToFunctionBoundary(at: syntax, collect: type, stopWithFirstMatch: true).first
  }
```

Sources/SwiftLexicalLookup/SimpleLookupQueries.swift
```diff
@@ -0,0 +1,227 @@
...
+
+  /// Callect syntax nodes matching the collection type up to a function boundary.
+  fileprivate func walkParentTreeUpToFunctionBoundary<T: SyntaxProtocol>(
```
ahoppen, line 116: I would say, the primary action that’s being performed here is the collection of syntax nodes, and walking is the means of achieving that. Thus I’d name the function to something around `collect`, eg.

```swift
func collectNodesOfTypeUpToFunctionBoundary(_ type: T.Type, stopWithFirstMatch: Bool = false)
```

Sources/SwiftLexicalLookup/SimpleLookupQueries.swift
```diff
@@ -0,0 +1,227 @@
...
+        DeinitializerDeclSyntax.self,
+        AccessorDeclSyntax.self,
+        ClosureExprSyntax.self,
+      ],
+      collect: collect,
+      stopWithFirstMatch: stopWithFirstMatch
+    )
+  }
+
+  /// Callect syntax nodes matching the collection type up until encountering one of the specified syntax nodes.
```
ahoppen, line 135:
```suggestion
  /// Collect syntax nodes matching the collection type up until encountering one of the specified syntax nodes. The nodes in the array are inside out, with the innermost node being the first.
```

Same in a couple more locations

Sources/SwiftLexicalLookup/SimpleLookupQueries.swift
```diff
@@ -0,0 +1,227 @@
...
+    stopWithFirstMatch: Bool = false
+  ) -> [T] {
+    walkParentTree(
+      upTo: [
+        MemberBlockSyntax.self,
+        FunctionDeclSyntax.self,
+        InitializerDeclSyntax.self,
+        DeinitializerDeclSyntax.self,
+        AccessorDeclSyntax.self,
+        ClosureExprSyntax.self,
```
ahoppen, line 128: I think `SubscriptDeclSyntax` should also be included in here.

Sources/SwiftLexicalLookup/SimpleLookupQueries.swift
```diff
@@ -0,0 +1,227 @@
...
+      stopWithFirstMatch: stopWithFirstMatch
+    )
+  }
+
+  /// Callect syntax nodes matching the collection type up until encountering one of the specified syntax nodes.
+  private func walkParentTree<T: SyntaxProtocol>(
+    upTo stopAt: [SyntaxProtocol.Type],
+    collect: T.Type,
+    stopWithFirstMatch: Bool = false
+  ) -> [T] {
```
ahoppen, line 140: Similar to my comment on `walkParentTreeUpToFunctionBoundary`, I would phrase this in terms of collecting nodes. Eg.

```suggestion
  private func collectNodes<T: SyntaxProtocol>(
    ofType: type: T.Type,
    upTo stopAt: [SyntaxProtocol.Type],
    stopWithFirstMatch: Bool = false
  ) -> [T] {
```

Sources/SwiftLexicalLookup/SimpleLookupQueries.swift
```diff
@@ -0,0 +1,227 @@
...
+  }
+
+  // MARK: - lookupFallthroughSourceAndDest
+
+  /// Given syntax node position, returns the current switch case and it's fallthrough destination.
+  private func lookupFallthroughSourceAndDestination(
+    at syntax: SyntaxProtocol
+  )
+    -> (SwitchCaseSyntax?, SwitchCaseSyntax?)
+  {
```
ahoppen, line 194:
```suggestion
  ) -> (SwitchCaseSyntax?, SwitchCaseSyntax?) {
```

Sources/SwiftLexicalLookup/SimpleLookupQueries.swift
```diff
@@ -0,0 +1,227 @@
...
+  ///   fallthrough
+  /// case 1:
+  ///   doSomethingElse()
+  /// default:
+  ///   break
+  /// }
+  /// ```
+  /// When calling this function at the `fallthrough`, it returns `case 2` and `case 1` in this exact order. The `nil` results handle ill-formed code: there's no `source` if the `fallthrough` is outside of a case. There's no `destination` if there is no case or `default` after the source case.
+  ///
+  @_spi(Experimental) public func lookupFallthroughSourceAndDest()
```
ahoppen, line 181:
```suggestion
  @_spi(Experimental) public func lookupFallthroughSourceAndDestintation()
```

Sources/SwiftLexicalLookup/SimpleLookupQueries.swift
```diff
@@ -0,0 +1,227 @@
...
+    -> (source: SwitchCaseSyntax?, destination: SwitchCaseSyntax?)
+  {
+    return lookupFallthroughSourceAndDestination(at: self)
+  }
+
+  // MARK: - lookupFallthroughSourceAndDest
+
+  /// Given syntax node position, returns the current switch case and it's fallthrough destination.
+  private func lookupFallthroughSourceAndDestination(
+    at syntax: SyntaxProtocol
```
ahoppen, line 191: Similar to other comments I have, I think having a function that takes a syntax node can be confusing. Personally, I would just implement this in `lookupFallthroughSourceAndDestination` that doesn’t take any parameters.

Sources/SwiftSyntax/SyntaxProtocol.swift
```diff
@@ -238,6 +238,18 @@ extension SyntaxProtocol {
   public var previousToken: TokenSyntax? {
     return self.previousToken(viewMode: .sourceAccurate)
   }
+
+  /// Returns this node or the first ancestor that satisfies `condition`.
+  public func ancestorOrSelf<T>(mapping map: (Syntax) -> T?) -> T? {
```
ahoppen, line 243: If we’re making this public API, could you add an entry to the [release notes](https://github.com/swiftlang/swift-syntax/blob/main/Release%20Notes/601.md)?

Tests/SwiftLexicalLookupTest/Assertions.swift
```diff
@@ -0,0 +1,83 @@
...
+/// Parse `source` and check if the method passed as `methodUnderTest` produces the same results as indicated in `expected`.
+///
+/// The `methodUnderTest` provides test inputs taken from the `expected` dictionary. The closure should return result produced by the tested method as an array with the same ordering.
+///
+/// - Parameters:
+///   - methodUnderTest: Closure with the tested method. Provides test argument from `expected` to the tested function. Should return method result as an array.
+///   - expected: A dictionary with parameter markers as keys and expected results as marker arrays ordered as returned by the test method.
+func assertLexicalScopeQuery(
+  source: String,
+  methodUnderTest: (SyntaxProtocol) -> ([SyntaxProtocol?]),
```
ahoppen, line 29: AFAICT, `methodUnderTest` is always called with a `TokenSyntax` as the argument. Should we change the signature to take a `TokenSyntax`? That also answers questions about which node is picked from the marker.

Package.swift
```diff
@@ -240,6 +241,18 @@ let package = Package(
...
+    .target(
+      name: "SwiftLexicalLookup",
+      dependencies: ["SwiftSyntax"]
+    ),
+
+    .testTarget(
+      name: "SwiftLexicalLookupTest",
+      dependencies: ["_SwiftSyntaxTestSupport", "SwiftLexicalLookup"]
+    ),
+
```
ahoppen, line 255: Can you insert this in alphabetical order?

Tests/SwiftLexicalLookupTest/Assertions.swift
```diff
@@ -0,0 +1,83 @@
...
+
+    // Execute the tested method
+    let result = methodUnderTest(testArgument)
+
+    // Extract the expected results for the test argument
+    let expectedValues: [SyntaxProtocol?] = expectedMarkers.map { expectedMarker in
+      guard let expectedMarker else { return nil }
+
+      guard let expectedPosition = markerDict[expectedMarker],
+        let expectedToken = sourceFileSyntax.token(at: AbsolutePosition(utf8Offset: expectedPosition))
```
ahoppen, line 57: Do we even need to extract the expected tokens? Shouldn’t it be sufficient to make `expectedValues` an `[AbsolutePosition]`? That way you don’t need to do the `firstToken` dance below.

Tests/SwiftLexicalLookupTest/Assertions.swift
```diff
@@ -0,0 +1,83 @@
...
+import XCTest
+import _SwiftSyntaxTestSupport
+
+/// Parse `source` and check if the method passed as `methodUnderTest` produces the same results as indicated in `expected`.
+///
+/// The `methodUnderTest` provides test inputs taken from the `expected` dictionary. The closure should return result produced by the tested method as an array with the same ordering.
+///
+/// - Parameters:
+///   - methodUnderTest: Closure with the tested method. Provides test argument from `expected` to the tested function. Should return method result as an array.
+///   - expected: A dictionary with parameter markers as keys and expected results as marker arrays ordered as returned by the test method.
```
ahoppen, line 26: I found this comment a little confusing. How about?

```suggestion
/// `methodUnderTest` is called with the token at every position marker in the keys of `expected`. It then asserts that the positions of the syntax nodes returned by `methodUnderTest` are the values in `expected`.
```

ahoppen (commented):
I really like the new API. I have a few more comments, mostly around clarity.

Sources/SwiftLexicalLookup/SimpleLookupQueries.swift
```diff
@@ -0,0 +1,227 @@
...
+  fileprivate func walkParentTreeUpToFunctionBoundary<T: SyntaxProtocol>(
+    at syntax: Syntax?,
+    collect: T.Type
+  ) -> T? {
+    walkParentTreeUpToFunctionBoundary(at: syntax, collect: collect, stopWithFirstMatch: true).first
+  }
+
+  /// Callect syntax nodes matching the collection type up to a function boundary.
+  fileprivate func walkParentTreeUpToFunctionBoundary<T: SyntaxProtocol>(
+    at syntax: Syntax?,
```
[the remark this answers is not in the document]
MAJKFL: It was a leftover from the previous implementation. Thanks for pointing this out.

Sources/SwiftLexicalLookup/SimpleLookupQueries.swift
```diff
@@ -0,0 +1,227 @@
...
+    return walkParentTreeUpToFunctionBoundary(
+      at: syntax.parent,
+      collect: LabeledStmtSyntax.self
+    )
+  }
+
+  // MARK: - lookupCatchNode
+
+  /// Given syntax node location, finds where an error could be caught. If set to `true`, `traverseCatchClause`lookup will skip the next do statement.
+  private func lookupCatchNodeHelper(at syntax: Syntax?, traversedCatchClause: Bool) -> Syntax? {
```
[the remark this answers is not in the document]
MAJKFL: I changed it now to just use `self`. Overall, due to the simplifications in this refactor, I also removed some superfluous helper methods.

Sources/SwiftSyntax/SyntaxProtocol.swift
MAJKFL: I added it to the release notes. I hope the description is sufficient.

Tests/SwiftLexicalLookupTest/Assertions.swift
MAJKFL: That's a good point, we don't really need to extract and compare tokens themselves.

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test Windows

DougGregor (approved):
I say "let's merge!"

Sources/SwiftLexicalLookup/SimpleLookupQueries.swift
```diff
@@ -0,0 +1,213 @@
...
+        DeinitializerDeclSyntax.self,
+        AccessorDeclSyntax.self,
+        ClosureExprSyntax.self,
+        SubscriptDeclSyntax.self,
+      ],
+      stopWithFirstMatch: stopWithFirstMatch
+    )
+  }
+
+  /// Callect syntax nodes matching the collection type up until encountering one of the specified syntax nodes.
```
ahoppen, line 131:
```suggestion
  /// Collect syntax nodes matching the collection type up until encountering one of the specified syntax nodes.
```

Sources/SwiftLexicalLookup/SimpleLookupQueries.swift
```diff
@@ -0,0 +1,213 @@
...
+  // MARK: - walkParentTree helper methods
+
+  /// Returns the innermost node of the specified type up to a function boundary.
+  fileprivate func innermostNodeOfTypeUpToFunctionBoundary<T: SyntaxProtocol>(
+    _ type: T.Type
+  ) -> T? {
+    collectNodesOfTypeUpToFunctionBoundary(type, stopWithFirstMatch: true).first
+  }
+
+  /// Collect syntax nodes matching the collection type up until encountering one of the specified syntax nodes. The nodes in the array are inside out, with the innermost node being the first.
```
ahoppen, line 111: Could you also wrap this comment at 120 columns?

Sources/SwiftLexicalLookup/SimpleLookupQueries.swift
```diff
@@ -0,0 +1,213 @@
...
+  /// Calling the function on `bar` results in `try?`.
+  /// When used on `error`, the function returns the function declaration `x`.
+  /// The function returns `nil` when there's no available catch node.
+  @_spi(Experimental) public func lookupCatchNode() -> Syntax? {
+    lookupCatchNodeHelper(traversedCatchClause: false)
+  }
+
+  // MARK: - lookupCatchNode
+
+  /// Given syntax node location, finds where an error could be caught. If `traverseCatchClause` is set to `true` lookup will skip the next do statement.
```
ahoppen, line 71: Same here, 120 columns

Tests/SwiftLexicalLookupTest/Assertions.swift
```diff
@@ -0,0 +1,83 @@
...
+//===----------------------------------------------------------------------===//
+
+import Foundation
+@_spi(Experimental) import SwiftLexicalLookup
+import SwiftParser
+import SwiftSyntax
+import XCTest
+import _SwiftSyntaxTestSupport
+
+/// `methodUnderTest` is called with the token at every position marker in the keys of `expected`. It then asserts that the positions of the syntax nodes returned by `methodUnderTest` are the values in `expected`.
```
ahoppen, line 20: And this comment would be good to keep at 120 columns.

Sources/SwiftLexicalLookup/SimpleLookupQueries.swift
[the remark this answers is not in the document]
ahoppen: Great! That’s what I was expecting as well.

Sources/SwiftSyntax/SyntaxProtocol.swift
ahoppen: Release notes look good. Thank you!

ahoppen (approved):
A few more nitpicky format comments, otherwise looks good to me. Thank you @MAJKFL!

MAJKFL:
Is it ok if we postpone the formatting changes to the next PR to not trigger CI for a couple of comments? I'll remember to include them. If you think the PR is good enough, could you please merge it? I see I would need write access for this repo to do so.

DougGregor:
I'll merge and we can pick up the formatting tweaks in another PR

PR #2702, opened 2024-06-25
[6.0][CMake] Add option to set the target namespace

rintaro:
Cherry-pick #2659  into release/6.0

* **Explanation**: For CMake builds, make an option `SWIFTSYNTAX_TARGET_NAMESPACE` to add a prefix to module names in the package. Also add `SWIFTSYNTAX_EMIT_MODULE` option for build library-evolution enabled modules with `.swiftinterface`. To keep the current behavior, this option is `TRUE` unless it's explicitly specified to `NO` 
* **Scope**: CMake builds
* **Risk**: Low. This adds a couple of CMake settings, but keeping the default behavior as-is.
* **Testing**: Passes the current test suite
* **Issue**: preparation for rdar://130532628
* **Reviewer**: Hamish Knight (@hamishknight)


rintaro:
@swift-ci Please test

PR #2704, opened 2024-06-26
build: parse arguments for `add_swift_host_library`

compnerd:
We use `EXCLUDE_FROM_ALL` in one of the libraries but did not account for that in the installation. When building swift-syntax for the android host for building swift-foundation, this prevents installation of the package. Parse and explicitly pass along the flag.


compnerd:
@swift-ci please test

bnbarham:
CC @rintaro since I know you have a PR that moves `EXCLUDE_FROM_ALL` top level.

PR #2712, opened 2024-06-29
Code action to convert between computed property and stored property

antigluten:
Resolves swiftlang/sourcekit-lsp#1245

### DIFF
--- a/Sources/SwiftRefactor/ConvertComputedPropertyToStored.swift
+++ b/Sources/SwiftRefactor/ConvertComputedPropertyToStored.swift
@@ -1,0 +1,96 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+#if swift(>=6)
+public import SwiftSyntax
+#else
+import SwiftSyntax
+#endif
+
+public struct ConvertComputedPropertyToStored: SyntaxRefactoringProvider {
+  public static func refactor(syntax: VariableDeclSyntax, in context: ()) -> VariableDeclSyntax? {
+    guard syntax.bindings.count == 1, let binding = syntax.bindings.first,
+      let accessorBlock = binding.accessorBlock, case let .getter(body) = accessorBlock.accessors, !body.isEmpty
+    else {
+      return nil
+    }
+
+    let refactored = { (initializer: InitializerClauseSyntax) -> VariableDeclSyntax in
+      let newBinding =
+        binding
+        .with(\.initializer, initializer)
+        .with(\.accessorBlock, nil)
+
+      let bindingSpecifier = syntax.bindingSpecifier
+        .with(\.tokenKind, .keyword(.let))
+
+      return
+        syntax
+        .with(\.bindingSpecifier, bindingSpecifier)
+        .with(\.bindings, PatternBindingListSyntax([newBinding]))
+    }
+
+    guard body.count == 1 else {
+      let closure = ClosureExprSyntax(
+        leftBrace: accessorBlock.leftBrace,
+        statements: body,
+        rightBrace: accessorBlock.rightBrace
+      )
+
+      return refactored(
+        InitializerClauseSyntax(
+          equal: .equalToken(trailingTrivia: .space),
+          value: FunctionCallExprSyntax(callee: closure)
+        )
+      )
+    }
+
+    guard body.count == 1, let item = body.first?.item else {
+      return nil
+    }
+
+    if let item = item.as(ReturnStmtSyntax.self), let expression = item.expression {
+      let trailingTrivia: Trivia = expression.leadingTrivia.isEmpty ? .space : []
+      return refactored(
+        InitializerClauseSyntax(
+          leadingTrivia: accessorBlock.leftBrace.trivia,
+          equal: .equalToken(trailingTrivia: trailingTrivia),
+          value: expression,
+          trailingTrivia: accessorBlock.rightBrace.trivia.droppingTrailingWhitespace
+        )
+      )
+    } else if var item = item.as(ExprSyntax.self) {
+      item.trailingTrivia = item.trailingTrivia.droppingTrailingWhitespace
+      return refactored(
+        InitializerClauseSyntax(
+          equal: .equalToken(trailingTrivia: .space),
+          value: item,
+          trailingTrivia: accessorBlock.trailingTrivia
+        )
+      )
+    }
+
+    return nil
+  }
+}
+
+fileprivate extension TokenSyntax {
+  var trivia: Trivia {
+    return leadingTrivia + trailingTrivia
+  }
+}
+
+fileprivate extension Trivia {
+  var droppingTrailingWhitespace: Trivia {
+    return Trivia(pieces: self.reversed().drop(while: \.isWhitespace).reversed())
+  }
+}
--- a/Sources/SwiftRefactor/ConvertStoredPropertyToComputed.swift
+++ b/Sources/SwiftRefactor/ConvertStoredPropertyToComputed.swift
@@ -1,0 +1,69 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+#if swift(>=6)
+public import SwiftSyntax
+#else
+import SwiftSyntax
+#endif
+
+public struct ConvertStoredPropertyToComputed: SyntaxRefactoringProvider {
+  public static func refactor(syntax: VariableDeclSyntax, in context: ()) -> VariableDeclSyntax? {
+    guard syntax.bindings.count == 1, let binding = syntax.bindings.first, let initializer = binding.initializer else {
+      return nil
+    }
+
+    var codeBlockSyntax: CodeBlockItemListSyntax
+
+    if let functionExpression = initializer.value.as(FunctionCallExprSyntax.self),
+      let closureExpression = functionExpression.calledExpression.as(ClosureExprSyntax.self)
+    {
+      guard functionExpression.arguments.isEmpty else { return nil }
+
+      codeBlockSyntax = closureExpression.statements
+      codeBlockSyntax.leadingTrivia =
+        closureExpression.leftBrace.leadingTrivia + closureExpression.leftBrace.trailingTrivia
+        + codeBlockSyntax.leadingTrivia
+      codeBlockSyntax.trailingTrivia +=
+        closureExpression.trailingTrivia + closureExpression.rightBrace.leadingTrivia
+        + closureExpression.rightBrace.trailingTrivia + functionExpression.trailingTrivia
+    } else {
+      var body = CodeBlockItemListSyntax([
+        CodeBlockItemSyntax(
+          item: .expr(initializer.value)
+        )
+      ])
+      body.leadingTrivia = initializer.equal.trailingTrivia + body.leadingTrivia
+      body.trailingTrivia += .space
+      codeBlockSyntax = body
+    }
+
+    let newBinding =
+      binding
+      .with(\.initializer, nil)
+      .with(
+        \.accessorBlock,
+        AccessorBlockSyntax(
+          accessors: .getter(codeBlockSyntax)
+        )
+      )
+
+    let newBindingSpecifier =
+      syntax.bindingSpecifier
+      .with(\.tokenKind, .keyword(.var))
+
+    return
+      syntax
+      .with(\.bindingSpecifier, newBindingSpecifier)
+      .with(\.bindings, PatternBindingListSyntax([newBinding]))
+  }
+}
--- a/Tests/SwiftRefactorTest/ConvertStoredPropertyToComputedTest.swift
+++ b/Tests/SwiftRefactorTest/ConvertStoredPropertyToComputedTest.swift
@@ -1,0 +1,189 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftRefactor
+import SwiftSyntax
+import SwiftSyntaxBuilder
+import XCTest
+import _SwiftSyntaxTestSupport
+
+final class ConvertStoredPropertyToComputedTest: XCTestCase {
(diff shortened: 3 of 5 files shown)


Sources/SwiftRefactor/ConvertComputedPropertyToStored.swift
```diff
@@ -0,0 +1,55 @@
...
+import SwiftSyntax
+#endif
+
+public struct ConvertComputedPropertyToStored: SyntaxRefactoringProvider {
+  
+  public static func refactor(syntax: VariableDeclSyntax, in context: ()) -> VariableDeclSyntax? {
+    guard syntax.bindings.count == 1, let binding = syntax.bindings.first,
+          let accessorBlock = binding.accessorBlock, case let .getter(body) = accessorBlock.accessors, !body.isEmpty else {
+      return nil
+    }
```
ahoppen, line 25: Could you run swift-format on your changes? https://github.com/swiftlang/swift-syntax/blob/main/CONTRIBUTING.md#formatting

Sources/SwiftRefactor/ConvertComputedPropertyToStored.swift
```diff
@@ -0,0 +1,55 @@
...
+//===----------------------------------------------------------------------===//
+
+#if swift(>=6)
+public import SwiftSyntax
+#else
+import SwiftSyntax
+#endif
+
+public struct ConvertComputedPropertyToStored: SyntaxRefactoringProvider {
+
```
ahoppen, line 20: Superfluous newline. Same in the other types.

Sources/SwiftRefactor/ConvertComputedPropertyToStored.swift
```diff
@@ -0,0 +1,55 @@
...
+    if body.count > 1, let closure = body.with(\.trailingTrivia, .newline).as(ClosureExprSyntax.self) {
+      initializer = InitializerClauseSyntax(
+        value: FunctionCallExprSyntax(
+          leadingTrivia: .space,
+          calledExpression: closure,
+          leftParen: .leftParenToken(),
+          arguments: {[]}(),
+          rightParen: .rightParenToken()
+        )
+      )
```
ahoppen, line 38: I don’t think this works because `body` is a `CodeBlockItemListSyntax` and converting that to a `ClosureExprSyntax` will never succeed. The test case also seems to fail.

Sources/SwiftRefactor/ConvertComputedPropertyToStored.swift
```diff
@@ -0,0 +1,55 @@
...
+    } else if let item = body.first?.item.as(ExprSyntax.self) {
```
ahoppen, line 39: I think we should only enter this case if `body` has a single item because otherwise we would drop everything after the first expression in the getter.

Sources/SwiftRefactor/ConvertComputedPropertyToStored.swift
```diff
@@ -0,0 +1,55 @@
...
+    }
+    
+    guard let initializer else { return nil }
+    
+    let newBinding = binding
+      .with(\.initializer, initializer)
+      .with(\.accessorBlock, nil)
+    
+    return syntax
+      .with(\.bindingSpecifier, .keyword(.let, trailingTrivia: .space))
```
ahoppen, line 52: I think we should retain the trivia of the old binding specifier. I think with the current implementation, we would drop any doc comments on the variable declaration because they are attached to the binding specifiers (since it’s the first token in the variable declaration).

Tests/SwiftRefactorTest/ConvertComputedPropertyToStored.swift
ahoppen, line 1: Let’s name this file `ConvertComputedPropertyToStoredTest.swift` to match the test class.

Tests/SwiftRefactorTest/ConvertComputedPropertyToStored.swift
```diff
@@ -0,0 +1,67 @@
...
+      """
+    
+    let expected: DeclSyntax = """
+      let defaultColor: Color = Color()
+      """
+    
+    try assertRefactorConvert(baseline, expected: expected)
+  }
+  
+  func testToStored2() throws {
```
ahoppen, line 33: How about giving this a little more descriptive name, eg. `testToStoredWithMultipleStatementsInAccessor`

Sources/SwiftRefactor/ConvertComputedPropertyToStored.swift
antigluten: Refactored to enter once if `body` has a single item :)

antigluten:
Thanks for review, @ahoppen!

Sources/SwiftRefactor/ConvertComputedPropertyToStored.swift
```diff
@@ -0,0 +1,81 @@
...
+      let accessorBlock = binding.accessorBlock, case let .getter(body) = accessorBlock.accessors, !body.isEmpty
+    else {
+      return nil
+    }
+
+    var initializer: InitializerClauseSyntax?
+
+    if body.count > 1 {
+      let closure = ClosureExprSyntax(
+        statements: body.with(\.trailingTrivia, .newline)
```
ahoppen, line 31: I think it would be good to re-use the braces from the `accessorBlock`. That way we also don’t loose any trivia (in particular comments) that are associated to those braces

```suggestion
        leftBrace: accessorBlock.leftBrace,
        statements: body,
        rightBrace: accessorBlock.rightBrace
```

Eg. this make sure we don’t loose the comment in

```swift
var x: Int {
  return 1
  // 1 is the correct number here
}
```

Sources/SwiftRefactor/ConvertComputedPropertyToStored.swift
```diff
@@ -0,0 +1,81 @@
...
+      )
+
+      initializer = InitializerClauseSyntax(
+        value: FunctionCallExprSyntax(
+          leadingTrivia: .space,
+          callee: closure
+        )
```
ahoppen, line 38: A little nitpicky but it’s nicer to attach the space as a trailing trivia to the `=` because that’s the way the parser would parse it as well.

Something like

```suggestion
        equal: .equalToken(trailingTrivia: .space).
        value: FunctionCallExprSyntax(callee: closure)
```

Sources/SwiftRefactor/ConvertComputedPropertyToStored.swift
```diff
@@ -0,0 +1,81 @@
...
+      )
+    } else if let item = body.first?.item.as(ExprSyntax.self) {
+      initializer = InitializerClauseSyntax(
+        value: item.with(\.leadingTrivia, .space).with(\.trailingTrivia, Trivia())
```
ahoppen, line 42: Similar here, I would attribute the space the `=`. This also ensures that we don’t loose leading trivia in eg.

```swift
let x: Int = 
  /* One */ 1
```

Sources/SwiftRefactor/ConvertComputedPropertyToStored.swift
```diff
@@ -0,0 +1,81 @@
...
+
+    return
+      syntax
+      .with(\.bindingSpecifier, bindingSpecifier)
+      .with(\.bindings, PatternBindingListSyntax([newBinding]))
+  }
+}
+
+fileprivate extension FunctionCallExprSyntax {
+
```
ahoppen, line 64: Superfluous newline

Sources/SwiftRefactor/ConvertComputedPropertyToStored.swift
```diff
@@ -0,0 +1,81 @@
...
+  init(
+    leadingTrivia: Trivia? = nil,
+    callee: some ExprSyntaxProtocol,
+    argumentList: () -> LabeledExprListSyntax = { [] },
+    trailingTrivia: Trivia? = nil
+  ) {
```
ahoppen, line 70: If we attach the space to the `=`, we should be able to use the `FunctionCallExprSyntax(callee:)` convenience initializer in SwiftSyntaxBuilder and don’t need to define this one.

Sources/SwiftRefactor/ConvertStoredPropertyToComputed.swift
```diff
@@ -0,0 +1,46 @@
...
+      binding
+      .with(\.initializer, nil)
+      .with(
+        \.accessorBlock,
+        AccessorBlockSyntax(
+          leftBrace: .leftBraceToken(trailingTrivia: .space),
+          accessors: .getter(CodeBlockItemListSyntax([body])),
+          rightBrace: .rightBraceToken(leadingTrivia: .space)
+        )
+      )
```
ahoppen, line 39: Would it be worth checking if the initializer is a closure that is immediately called so that converting stored -> computed -> stored results in the same code even if there are multiple statements in the getter?

Sources/SwiftRefactor/ConvertStoredPropertyToComputed.swift
```diff
@@ -0,0 +1,46 @@
...
+
+    return
+      syntax
+      .with(\.bindingSpecifier, .keyword(.var, trailingTrivia: .space))
```
ahoppen, line 43: Similar here, I think we should re-use the trivia of the previous binding specifier. Could you add tests cases to check that we don’t drop doc comments on variables during the refactoring?

ahoppen (commented):
Thanks for the adjustments. I have a few comments, mostly regarding trivia handling. The overall shape looks good. It would also be good to add test cases for the examples that I mentioned.

Sources/SwiftRefactor/ConvertStoredPropertyToComputed.swift
antigluten: You mean add a specific condition to the method or create a test case where convert to computer and vice versa?

antigluten:
Need help with the `testToStoredWithReturnStatementAndTrailingCommentOnNewLine`, I have no idea how to handle the leading trivia, so to not break another test

Test: `testToStoredWithReturnStatementAndTrailingCommentOnNewLine` 
```swift
let baseline: DeclSyntax = """
  var defaultColor: Color {
    return Color()
    /* some text */
  }
  """

let expected: DeclSyntax = """
  let defaultColor: Color = Color()
  /* some text */
  """
```

Test: `testToStoredWithReturnStatementAndTrailingComment`
```swift
let baseline: DeclSyntax = """
  var defaultColor: Color {
    return Color() /* some text */
  }
  """

let expected: DeclSyntax = """
  let defaultColor: Color = Color() /* some text */
  """
```

Sources/SwiftRefactor/ConvertStoredPropertyToComputed.swift
ahoppen: I meant that if you start of with

```swift
var x: Int {
  var intermediate = 1 + 1
  return intermediate + 1
}
```

And then perform the covert to stored property refactoring, you get the following, which is what I would expect

```swift
var x: Int = {
  var intermediate = 1 + 1
  return intermediate + 1
}()
```

But then if you convert that to a stored property again (essentially trying to undo the change without hitting Cmd-Z), you get the following

```swift
var x: Int {
  {
    var intermediate = 1 + 1
    return intermediate + 1
  }()
}
```

And that seems overly complicated. I think in this case we should return to the original code when converting back to a stored property.

ahoppen:
How about adding the trivia of the dropped braces to the initializer?

Ie something like

```swift
initializer.leadingTrivia = accessorBlock.leftBrace.leadingTrivia + accessorBlock.leftBrace.trailing + initializer.leadingTrivia
initializer.trailingTrivia += accessorBlock.rightBrace.leadingTrivia + accessorBlock.rightBrace.trailing
```

antigluten:
@ahoppen, done

Sources/SwiftRefactor/ConvertStoredPropertyToComputed.swift
antigluten: Resolved this in [eac08a4](https://github.com/swiftlang/swift-syntax/pull/2712/commits/eac08a423e06d1153beacfccb1dad6a6c2dfd62b) by checking for a closure and removing it with saving the trivia.
Not understood the idea about converting back to the original code, how could we do that if we don't even save an original state.

FYI, I remember about squashing commits, I will do it after your review :)

Sources/SwiftRefactor/ConvertStoredPropertyToComputed.swift
```diff
@@ -0,0 +1,73 @@
...
+public struct ConvertStoredPropertyToComputed: SyntaxRefactoringProvider {
+  public static func refactor(syntax: VariableDeclSyntax, in context: ()) -> VariableDeclSyntax? {
+    guard syntax.bindings.count == 1, let binding = syntax.bindings.first, let initializer = binding.initializer else {
+      return nil
+    }
+
+    let codeBlockSyntax: CodeBlockItemListSyntax
+
+    if let functionExpression = initializer.value.as(FunctionCallExprSyntax.self),
+      let closureExpression = functionExpression.calledExpression.as(ClosureExprSyntax.self)
```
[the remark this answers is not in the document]
antigluten: Added the check and and a test case

Sources/SwiftRefactor/ConvertStoredPropertyToComputed.swift
```diff
@@ -0,0 +1,73 @@
...
+        .with(
+          \.leadingTrivia,
+          functionExpression.leadingTrivia + closureExpression.leftBrace.leadingTrivia
+            + closureExpression.leftBrace.trailingTrivia + statements.leadingTrivia
+        )
+        .with(
+          \.trailingTrivia,
+          statements.trailingTrivia + closureExpression.trailingTrivia + closureExpression.rightBrace.leadingTrivia
+            + closureExpression.rightBrace.trailingTrivia + functionExpression.trailingTrivia
+        )
```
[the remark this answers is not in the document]
antigluten: Yeah, I agree :)

Sources/SwiftRefactor/ConvertComputedPropertyToStored.swift
```diff
@@ -0,0 +1,97 @@
...
+      )
+    }
+
+    guard let item = body.first?.item else {
+      return nil
+    }
+
+    if let item = item.as(ReturnStmtSyntax.self), let expression = item.expression {
+      let trailingTrivia: Trivia = expression.leadingTrivia.isEmpty ? .space : []
+      let lineIndentation = syntax.firstToken(viewMode: .sourceAccurate)?.indentationOfLine ?? []
```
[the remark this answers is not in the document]
antigluten: Removed

Tests/SwiftRefactorTest/ConvertStoredPropertyToComputedTest.swift
```diff
@@ -0,0 +1,159 @@
...
+    let baseline: DeclSyntax = """
+      var defaultColor: Color = Color()
+      """
+
+    let expected: DeclSyntax = """
+      var defaultColor: Color { Color() }
+      """
+
+    try assertRefactorConvert(baseline, expected: expected)
+  }
```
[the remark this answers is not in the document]
antigluten: Should we leave the test as is, without explicitly adding type of the variable, or there is a special method to know the type?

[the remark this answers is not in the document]
ahoppen: Oh no, I just forgot to add a type annotation.

I have been discussing some ideas of how to determine the inferred type of a variable on the issue (https://github.com/swiftlang/sourcekit-lsp/issues/1245) but that doesn’t need to be addressed in your PR.

ahoppen (approved):
Looking great. Let’s get this merged.

ahoppen:
@swift-ci Please test

Tests/SwiftRefactorTest/ConvertStoredPropertyToComputedTest.swift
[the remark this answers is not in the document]
antigluten: Fixed test to pass with explicit type annotation

kimdv:
@swift-ci please test

kimdv:
@swift-ci please test windows

ahoppen:
@swift-ci Please test Linux

PR #2713, opened 2024-06-29
Add documentation for `FunctionDeclSyntax`, `ExtensionDeclSyntax`

antigluten:
As part of #1528, the documentation is added to `ExtensionDeclSyntax` and `FunctionDeclSyntax`


CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
```diff
@@ -835,6 +835,20 @@ public let DECL_NODES: [Node] = [
     kind: .extensionDecl,
     base: .decl,
     nameForDiagnostics: "extension",
+    documentation: """
+        An `extension` declaration.
+
+        An example of an extension is
```
ahoppen, line 841: I think that other declarations have an `Example` section and it would be good to be consistent here.

```suggestion
        ### Example
```

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
```diff
@@ -835,6 +835,20 @@ public let DECL_NODES: [Node] = [
...
+
+        ```swift
+        extension String {
+
```
ahoppen, line 845: Let’s keep the example short and remove this empty newline 😉

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
```diff
@@ -887,6 +905,27 @@ public let DECL_NODES: [Node] = [
     kind: .functionDecl,
     base: .decl,
     nameForDiagnostics: "function",
+    documentation: """
+        A Swift `func` declaration.
+        
+        An example of a func declaration is
```
ahoppen, line 911:
```suggestion
        ### Examples
```

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
```diff
@@ -844,7 +858,8 @@ public let DECL_NODES: [Node] = [
       Child(
         name: "attributes",
         kind: .collection(kind: .attributeList, collectionElementName: "Attribute", defaultsToEmpty: true),
-        nameForDiagnostics: "attributes"
+        nameForDiagnostics: "attributes",
+        documentation: "Attributes that are attached to the extension declaration."
```
ahoppen, line 862: This just got me wondering: What attributes can be applied to extensions. One thing I could think of was `@MainActor`. What do you think about mentioning that as an example? I generally find that  examples help visualize what the nodes are for.

```suggestion
        documentation: """
          Attributes that are attached to the extension declaration.
          
          ### Example
          
          `@MainActor` in 
          
          ```swift
          @MainActor extension SomeType {}
          ```
          """
```

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
```diff
@@ -898,7 +937,8 @@ public let DECL_NODES: [Node] = [
       Child(
         name: "attributes",
         kind: .collection(kind: .attributeList, collectionElementName: "Attribute", defaultsToEmpty: true),
-        nameForDiagnostics: "attributes"
+        nameForDiagnostics: "attributes",
+        documentation: "Attributes that are attached to the func declaration."
```
ahoppen, line 941: Let’s spell out *function* here because we’re using it as a normal english term instead of as a Swift keyword.

```suggestion
        documentation: "Attributes that are attached to the function declaration."
```

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
```diff
@@ -931,7 +972,11 @@ public let DECL_NODES: [Node] = [
       Child(
         name: "signature",
         kind: .node(kind: .functionSignature),
-        nameForDiagnostics: "function signature"
+        nameForDiagnostics: "function signature",
+        documentation: """
+          The arguments of the initializer.
+          While the function signature allows specifying a return clause, doing so is not semantically valid."
+          """
```
ahoppen, line 979: Did you mix up functions and initializers here? 😉

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
```diff
@@ -953,60 +999,83 @@ public let DECL_NODES: [Node] = [
     kind: .functionParameterList,
     base: .syntaxCollection,
     nameForDiagnostics: "parameter list",
+    documentation: """
+      A list of function parameters that are type annotated and may have an optional label.
```
ahoppen, line 1003: The parameter labels aren’t optional, right? Or am I missing something?

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
```diff
@@ -953,60 +999,83 @@ public let DECL_NODES: [Node] = [
...
         nameForDiagnostics: "internal name",
+        documentation:
+          "If this is specified, it is the name by which the parameter can be referenced inside the function body.",
         isOptional: true
       ),
       Child(
         name: "colon",
-        kind: .token(choices: [.token(.colon)])
+        kind: .token(choices: [.token(.colon)]),
+        documentation: "If the parameter has a label, the colon separating the label from the type."
```
ahoppen, line 1053: `firstName` isn’t optional and neither is `colon`, so I think this should just be

```suggestion
        documentation: "The colon separating the label from the type."
```

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
```diff
@@ -953,60 +999,83 @@ public let DECL_NODES: [Node] = [
...
         kind: .token(choices: [.token(.ellipsis)]),
+        documentation: "If the parameter is variadic, `...` to indicate that.",
         isOptional: true
       ),
       Child(
         name: "defaultValue",
         deprecatedName: "defaultArgument",
         kind: .node(kind: .initializerClause),
         nameForDiagnostics: "default value",
+        documentation: "If the parameter has a default value, the function clause describing the default value.",
```
ahoppen, line 1072: I don’t think *function clause* is a term we use anywhere else. How about we just say it’s the expression, glossing over the fact that the `InitializerClauseSyntax` also contains the `=`. I think that should get the point across good enough.

```suggestion
        documentation: "If the parameter has a default value, the expression describing the default value.",
```

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
```diff
@@ -1016,21 +1085,25 @@ public let DECL_NODES: [Node] = [
...
         name: "parameterClause",
         deprecatedName: "input",
-        kind: .node(kind: .functionParameterClause)
+        kind: .node(kind: .functionParameterClause),
+        documentation: "The parameters of the function."
       ),
       Child(
         name: "effectSpecifiers",
         kind: .node(kind: .functionEffectSpecifiers),
+        documentation: "The side effect indicators of the function. They help clarify the behavior of the function.",
```
ahoppen, line 1099: `async` and `throw` do more than just clarifying the behavior of the function, they actually change the behavior of a function. How about focussing on `async` and `throws` here? I think that’s a concept that most Swift developers understand even if they don’t know the formal term for it.


Something like

```suggestion
        documentation: "The effect indicators of the function, like `async` or `throws`.",
```

ahoppen (commented):
Thanks for adding the documentation, @antigluten!

antigluten:
Thanks for your review, @ahoppen

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
antigluten: Yes, did some copy-pasting, is "A function signature that defines the interface of the function." better for this case, if not, need some help

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
antigluten: I meant specifying argument labels before parameter name, this is an optional action, is it?

ahoppen: Did you mean
- That the argument label could be `_`, eg. `func foo(_ a: Int)`
  - In this case, I would still consider there to be an explicit label on the declaration side, it’s just that it’s `_`.
- That you can omit the internal name, eg. `func foo(a: Int)`
  - In this case the function still has an argument label `a`. It’s the internal or second name that you are omitting, which defines the name of the parameter inside the function’s body, which then defaults to the label

In either case, I would change the documentation to

```suggestion
      A list of function parameters that are type annotated and a label.
```

antigluten: You got it :)

antigluten:
@ahoppen, have commited your suggestions and updated generated code

ahoppen (commented):
One last thing: Could you squash your commits? https://github.com/swiftlang/swift-syntax/blob/main/CONTRIBUTING.md#authoring-commits

antigluten:
> One last thing: Could you squash your commits? https://github.com/swiftlang/swift-syntax/blob/main/CONTRIBUTING.md#authoring-commits

Done  ✅

ahoppen (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

PR #2721, opened 2024-07-09
Code actions to convert between computed properties and zero-parameter functions

antigluten:
A part of https://github.com/swiftlang/sourcekit-lsp/issues/1246

### DIFF
--- a/Sources/SwiftRefactor/ConvertComputedPropertyToZeroParameterFunction.swift
+++ b/Sources/SwiftRefactor/ConvertComputedPropertyToZeroParameterFunction.swift
@@ -1,0 +1,105 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+#if swift(>=6)
+public import SwiftSyntax
+#else
+import SwiftSyntax
+#endif
+
+public struct ConvertComputedPropertyToZeroParameterFunction: SyntaxRefactoringProvider {
+  public static func refactor(syntax: VariableDeclSyntax, in context: Void) -> FunctionDeclSyntax? {
+    guard syntax.bindings.count == 1,
+      let binding = syntax.bindings.first,
+      let identifierPattern = binding.pattern.as(IdentifierPatternSyntax.self)
+    else { return nil }
+
+    var statements: CodeBlockItemListSyntax
+
+    guard let typeAnnotation = binding.typeAnnotation,
+      var accessorBlock = binding.accessorBlock
+    else { return nil }
+
+    var effectSpecifiers: AccessorEffectSpecifiersSyntax?
+
+    switch accessorBlock.accessors {
+    case .accessors(let accessors):
+      guard accessors.count == 1, let accessor = accessors.first,
+        accessor.accessorSpecifier.tokenKind == .keyword(.get), let codeBlock = accessor.body
+      else { return nil }
+      effectSpecifiers = accessor.effectSpecifiers
+      statements = codeBlock.statements
+      let accessorSpecifier = accessor.accessorSpecifier
+      statements.leadingTrivia =
+        accessorSpecifier.leadingTrivia + accessorSpecifier.trailingTrivia.droppingLeadingWhitespace
+        + codeBlock.leftBrace.leadingTrivia.droppingLeadingWhitespace
+        + codeBlock.leftBrace.trailingTrivia.droppingLeadingWhitespace
+        + statements.leadingTrivia
+      statements.trailingTrivia += codeBlock.rightBrace.trivia.droppingLeadingWhitespace
+      statements.trailingTrivia = statements.trailingTrivia.droppingTrailingWhitespace
+    case .getter(let codeBlock):
+      statements = codeBlock
+    }
+
+    let returnType = typeAnnotation.type
+
+    var returnClause: ReturnClauseSyntax?
+    let triviaAfterSignature: Trivia
+
+    if !returnType.isVoid {
+      triviaAfterSignature = .space
+      returnClause = ReturnClauseSyntax(
+        arrow: .arrowToken(
+          leadingTrivia: typeAnnotation.colon.leadingTrivia,
+          trailingTrivia: typeAnnotation.colon.trailingTrivia
+        ),
+        type: returnType
+      )
+    } else {
+      triviaAfterSignature = typeAnnotation.colon.leadingTrivia + typeAnnotation.colon.trailingTrivia
+    }
+
+    accessorBlock.leftBrace.leadingTrivia = accessorBlock.leftBrace.leadingTrivia.droppingLeadingWhitespace
+    accessorBlock.rightBrace.trailingTrivia = accessorBlock.rightBrace.trailingTrivia.droppingTrailingWhitespace
+
+    let body = CodeBlockSyntax(
+      leftBrace: accessorBlock.leftBrace,
+      statements: statements,
+      rightBrace: accessorBlock.rightBrace
+    )
+
+    var parameterClause = FunctionParameterClauseSyntax(parameters: [])
+    parameterClause.trailingTrivia = identifierPattern.identifier.trailingTrivia + triviaAfterSignature
+
+    let functionEffectSpecifiers = FunctionEffectSpecifiersSyntax(
+      asyncSpecifier: effectSpecifiers?.asyncSpecifier,
+      throwsClause: effectSpecifiers?.throwsClause
+    )
+    let functionSignature = FunctionSignatureSyntax(
+      parameterClause: parameterClause,
+      effectSpecifiers: functionEffectSpecifiers,
+      returnClause: returnClause
+    )
+
+    return FunctionDeclSyntax(
+      modifiers: syntax.modifiers,
+      funcKeyword: .keyword(
+        .func,
+        leadingTrivia: syntax.bindingSpecifier.leadingTrivia,
+        trailingTrivia: syntax.bindingSpecifier.trailingTrivia
+      ),
+      name: identifierPattern.identifier.with(\.trailingTrivia, []),
+      signature: functionSignature,
+      body: body
+    )
+  }
+}
--- a/Sources/SwiftRefactor/ConvertZeroParameterFunctionToComputedProperty.swift
+++ b/Sources/SwiftRefactor/ConvertZeroParameterFunctionToComputedProperty.swift
@@ -1,0 +1,70 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+#if swift(>=6)
+public import SwiftSyntax
+#else
+import SwiftSyntax
+#endif
+
+public struct ConvertZeroParameterFunctionToComputedProperty: SyntaxRefactoringProvider {
+  public static func refactor(syntax: FunctionDeclSyntax, in context: ()) -> VariableDeclSyntax? {
+    guard syntax.signature.parameterClause.parameters.isEmpty,
+      let body = syntax.body
+    else { return nil }
+
+    let variableName = PatternSyntax(
+      IdentifierPatternSyntax(
+        leadingTrivia: syntax.funcKeyword.trailingTrivia,
+        identifier: syntax.name
+      )
+    )
+
+    let triviaFromParameters =
+      (syntax.signature.parameterClause.leftParen.trivia + syntax.signature.parameterClause.rightParen.trivia)
+      .droppingTrailingWhitespace
+
+    var variableType: TypeAnnotationSyntax?
+
+    if let returnClause = syntax.signature.returnClause {
+      variableType = TypeAnnotationSyntax(
+        colon: .colonToken(
+          leadingTrivia: triviaFromParameters + returnClause.arrow.leadingTrivia,
+          trailingTrivia: returnClause.arrow.trailingTrivia
+        ),
+        type: returnClause.type
+      )
+    } else {
+      variableType = TypeAnnotationSyntax(
+        colon: .colonToken(
+          leadingTrivia: triviaFromParameters,
+          trailingTrivia: .space
+        ),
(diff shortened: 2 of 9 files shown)


harlanhaskins:
This is awesome!

Can you also add some tests to make sure this doesn't suggest the refactor:

```swift
var x: Int {
    get { 5 }
    set { /*anything */ }
}
```

And that the refactor succeeds with:

```swift
var x: Int {
    get { 5 }
}
```

Sources/SwiftRefactor/ConvertComputedPropertyToZeroParameterFunction.swift
```diff
@@ -0,0 +1,77 @@
...
+#endif
+
+public struct ConvertComputedPropertyToZeroParameterFunction: SyntaxRefactoringProvider {
+  public static func refactor(syntax: VariableDeclSyntax, in context: Void) -> FunctionDeclSyntax? {
+    guard syntax.bindings.count == 1,
+      let binding = syntax.bindings.first,
+      var identifierPattern = binding.pattern.as(IdentifierPatternSyntax.self)
+    else { return nil }
+
+    identifierPattern.leadingTrivia = syntax.bindingSpecifier.trailingTrivia
```
ahoppen, line 26: I assume that you’re doing this so we don’t loose the trivia around the `let`, right? If so, I would just attach the leading trivia of the binding specifier to `func` instead of overriding the leading trivia of the `identifierPattern`.

Sources/SwiftRefactor/ConvertComputedPropertyToZeroParameterFunction.swift
```diff
@@ -0,0 +1,77 @@
...
+      leftBrace: accessorBlock.leftBrace,
+      statements: statements,
+      rightBrace: accessorBlock.rightBrace
+    )
+
+    let functionSignature = FunctionSignatureSyntax(
+      parameterClause: FunctionParameterClauseSyntax(parameters: []),
+      returnClause: returnClause,
+      trailingTrivia: trailingTrivia
+    )
```
ahoppen, line 61: Why do we need to add the space as leading trivia to the arrow? Wouldn’t it be easier if the `)` in the parameter clause always has a trailing space trivia? That would also match how the parser would qualify the trivia.

The arrow token could then just inherit the trivia from the colon. What I would do is something like the following

```suggestion
    let triviaAfterSignature: Trivia

    if !returnType.isVoid {
      triviaAfterSignature = .space
      returnClause = ReturnClauseSyntax(
        arrow: .arrowToken(
          leadingTrivia: typeAnnotation.colon.leadingTrivia,
          trailingTrivia: typeAnnotation.colon.trailingTrivia
        ),
        type: returnType
      )
    } else {
      triviaAfterSignature = typeAnnotation.colon.leadingTrivia + typeAnnotation.colon.trailingTrivia
    }

    let body = CodeBlockSyntax(
      leftBrace: accessorBlock.leftBrace,
      statements: statements,
      rightBrace: accessorBlock.rightBrace
    )

    let functionSignature = FunctionSignatureSyntax(
      parameterClause: FunctionParameterClauseSyntax(parameters: []),
      returnClause: returnClause,
      trailingTrivia: identifierPattern.identifier.trailingTrivia + triviaAfterSignature
    )
```

Sources/SwiftRefactor/ConvertComputedPropertyToZeroParameterFunction.swift
```diff
@@ -0,0 +1,77 @@
...
+  }
+}
+
+fileprivate extension TypeSyntax {
+  var isVoid: Bool {
+    let description = trimmedDescription
+    return description == "Void" || description == "()"
+  }
+}
```
ahoppen, line 77: We could inspect the syntax tree more syntactically. That way, we also match eg. `( )` with a space between the parentheses (not that I expect anyone to write it like that but still).

```suggestion
fileprivate extension TypeSyntax {
  var isVoid: Bool {
    switch self.as(TypeSyntaxEnum.self) {
    case .identifierType(let identifierType) where identifierType.name.text == "Void": return true
    case .tupleType(let tupleType) where tupleType.elements.isEmpty: return true
    default: return false
    }
  }
}
```

Sources/SwiftRefactor/ConvertZeroParameterFunctionToComputedProperty.swift
```diff
@@ -0,0 +1,69 @@
...
+
+    let variableName = PatternSyntax(
+      IdentifierPatternSyntax(
+        leadingTrivia: syntax.funcKeyword.trailingTrivia,
+        identifier: syntax.name
+      )
+    )
+
+    let paramsTrailingTrivia = syntax.signature.parameterClause.rightParen.trailingTrivia
+    let colonLeadingTrivia = paramsTrailingTrivia == .space ? [] : paramsTrailingTrivia
```
ahoppen, line 33: I think we shouldn’t be dropping any trivia from the parameter clause

```suggestion
let triviaFromParameters = (syntax.signature.parameterClause.leftParen.trivia + syntax.signature.parameterClause.rightParen.trivia).droppingTrailingWhitespace
```

With `trivia` and `droppingTrailingWhitespace` from your [other PR](https://github.com/swiftlang/swift-syntax/pull/2712/files#diff-5d23988c50101027622d00dbafeddfe2b73ba76ad27953908617600f8a8c2721R86-R96)

I think that requires a few test updates but I think all of them are improvements.

Sources/SwiftRefactor/ConvertZeroParameterFunctionToComputedProperty.swift
```diff
@@ -0,0 +1,69 @@
...
+
+    var variableType: TypeAnnotationSyntax?
+
+    if let returnClause = syntax.signature.returnClause {
+      variableType = TypeAnnotationSyntax(
+        colon: .colonToken(
+          leadingTrivia: colonLeadingTrivia,
```
ahoppen, line 40: I think it would be good practice to also retain the leading trivia from the arrow.

```suggestion
          leadingTrivia: triviaFromParameters + returnClause.arrow.leadingTrivia,
```

ahoppen (commented):
Very nice. The test suggestion from @harlanhaskins would be good to address as well but I’m pretty sure those test will pass.

Sources/SwiftRefactor/ConvertComputedPropertyToZeroParameterFunction.swift
antigluten: Wow, this is the better way to handle it 👍🏻

antigluten:
@harlanhaskins Thank you, for your test suggestion!
@ahoppen, Thanks for your review, refactoring is done, added checking for `get` accessor when we converting to the function

Sources/SwiftRefactor/ConvertComputedPropertyToZeroParameterFunction.swift
```diff
@@ -0,0 +1,101 @@
...
+    parameterClause.trailingTrivia = identifierPattern.identifier.trailingTrivia + triviaAfterSignature
+
+    let functionSignature = FunctionSignatureSyntax(
+      parameterClause: parameterClause,
+      returnClause: returnClause
+    )
+
+    return FunctionDeclSyntax(
+      modifiers: syntax.modifiers,
+      funcKeyword: .keyword(.func, trailingTrivia: syntax.bindingSpecifier.trailingTrivia),
```
ahoppen, line 79: Should we also inherit the leading trivia of the bindings specifier here? A test case for that would be

```swift
/// Some doc comment
let foo { 1 }
```

Sources/SwiftRefactor/ConvertComputedPropertyToZeroParameterFunction.swift
```diff
@@ -0,0 +1,101 @@
...
+
+    switch accessorBlock.accessors {
+    case .accessors(let accessors):
+      guard accessors.count == 1, let accessor = accessors.first,
+        accessor.accessorSpecifier.tokenKind == .keyword(.get), let codeBlock = accessor.body
+      else { return nil }
+      statements = codeBlock.statements
+      statements.leadingTrivia = accessor.leadingTrivia + codeBlock.leadingTrivia + statements.leadingTrivia
+      statements.trailingTrivia += codeBlock.trailingTrivia + accessor.trailingTrivia
+      statements.trailingTrivia = statements.trailingTrivia.droppingTrailingWhitespace
```
ahoppen, line 40: We are still dropping the trailing trivia here (eg. if you have `get /* comment */ {`). I would suggest that we add the leading trivia of the tokens here.

```suggestion
      statements.leadingTrivia = accessor.accessorSpecifier.trivia + codeBlock.leftBrace.trivia + statements.leadingTrivia
      statements.trailingTrivia += codeBlock.rightBrace.trivia
      statements.trailingTrivia = statements.trailingTrivia.droppingTrailingWhitespace
```

Sources/SwiftRefactor/ConvertComputedPropertyToZeroParameterFunction.swift
```diff
@@ -0,0 +1,101 @@
...
+    var statements: CodeBlockItemListSyntax
+
+    guard let typeAnnotation = binding.typeAnnotation,
+      let accessorBlock = binding.accessorBlock
+    else { return nil }
+
+    switch accessorBlock.accessors {
+    case .accessors(let accessors):
+      guard accessors.count == 1, let accessor = accessors.first,
+        accessor.accessorSpecifier.tokenKind == .keyword(.get), let codeBlock = accessor.body
```
ahoppen, line 35: You also need to transfer the effect specifiers from the `get` accessor to the function, eg. to be able to convert

```swift
var foo: Int {
  get async {
    await someAsyncValue()
  }
}
```

to

```swift
func foo() async -> Int {
  await someAsyncValue()
}
```

ahoppen (commented):
Thanks for also adding support for a variable with an explicit `get` getter.

Sources/SwiftRefactor/ConvertComputedPropertyToZeroParameterFunction.swift
antigluten: Added the test to check this behavior

Sources/SwiftRefactor/ConvertComputedPropertyToZeroParameterFunction.swift
antigluten: Done

Sources/SwiftRefactor/ConvertComputedPropertyToZeroParameterFunction.swift
```diff
@@ -0,0 +1,133 @@
...
+  var droppingTrailingWhitespace: Trivia {
+    return Trivia(pieces: self.reversed().drop(while: \.isWhitespace).reversed())
+  }
+}
+
+fileprivate extension TokenSyntax {
+  var trivia: Trivia {
+    return leadingTrivia + trailingTrivia
+  }
+}
```
ahoppen, line 133: Instead of adding these functions to multiple files, what do you think about adding a SyntaxUtils.swift file to `SwiftRefactor` and adding them there?

Sources/SwiftRefactor/ConvertComputedPropertyToZeroParameterFunction.swift
ahoppen, line 1: Sorry, forgot to mention before but could you also add the new files to [CMakeLists.txt](https://github.com/swiftlang/swift-syntax/blob/main/Sources/SwiftRefactor/CMakeLists.txt). We build swift-syntax using CMake to link it into the compiler.

Sources/SwiftRefactor/ConvertComputedPropertyToZeroParameterFunction.swift
```diff
@@ -0,0 +1,133 @@
...
+
+    var parameterClause = FunctionParameterClauseSyntax(parameters: [])
+    parameterClause.trailingTrivia = identifierPattern.identifier.trailingTrivia + triviaAfterSignature
+
+    let functionEffectSpecifiers = FunctionEffectSpecifiersSyntax(
+      leadingTrivia: effectSpecifiers?.leadingTrivia,
+      asyncSpecifier: effectSpecifiers?.asyncSpecifier,
+      throwsClause: effectSpecifiers?.throwsClause,
+      trailingTrivia: effectSpecifiers?.trailingTrivia.droppingTrailingWhitespace
+    )
```
ahoppen, line 88: A syntax node doesn’t have any inherent trivia, `effectsSpecifiers.leadingTrivia` will be the leading trivia of the first token within that tree. Since that will most likely be the `asyncSpecifier`, this will be duplicating the leading trivia of `asyncSpecifier` (same for `trailingTrivia`).

```suggestion
    let functionEffectSpecifiers = FunctionEffectSpecifiersSyntax(
      asyncSpecifier: effectSpecifiers?.asyncSpecifier,
      throwsClause: effectSpecifiers?.throwsClause
    )
```

Sources/SwiftRefactor/ConvertComputedPropertyToZeroParameterFunction.swift
antigluten: Yeah, you're right, forgot about that :(

Sources/SwiftRefactor/ConvertComputedPropertyToZeroParameterFunction.swift
antigluten: Extracted all repetitive chunks to the new file, should we add a warning about inserting extensions in alphabetical order?

Sources/SwiftRefactor/SyntaxUtils.swift
```diff
@@ -0,0 +1,43 @@
...
+}
+
+extension TypeSyntax {
+  public var isVoid: Bool {
+    switch self.as(TypeSyntaxEnum.self) {
+    case .identifierType(let identifierType) where identifierType.name.text == "Void": return true
+    case .tupleType(let tupleType) where tupleType.elements.isEmpty: return true
+    default: return false
+    }
+  }
```
ahoppen, line 42: Let’s make all of these `internal`. We want to keep them as implementation details of SwiftRefactor and don’t want to let external clients starting to rely on them.

Sources/SwiftRefactor/ConvertComputedPropertyToZeroParameterFunction.swift
ahoppen: I don’t expect that file to grow too much, so I don’t care that much about ordering and we shouldn’t need a warning about the alphabetical ordering.

Sources/SwiftRefactor/SyntaxUtils.swift
antigluten: done

ahoppen (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
Instead of merging `main` into your branch, could you rebase your branch on top of the latest `main`? It makes for a nicer git history and it’s too easy to hide weird subtle changes in merge commits.

antigluten:
@ahoppen, thank you for your hint, done rebasing

antigluten:
Accidentally removed the wrong branch

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #2722, opened 2024-07-09
Fix Accessor Macros Attached to Properties With Trailing Comments

### LINKED ISSUE #2614: Accessor Macros applied to properties with trailing trivia comments lead to Swift that fails to compile.
### Description

The following Swift code fails to compile:

```swift
var x: Int // comment {
  get {
    return 1
  }
}
```

The trailing trivia comment captures the opening round brace… which leads to the compiler error.

We can attach an accessor macro to a variable declaration with similar trailing trivia:

```swift
@constantOne
var x: Int // comment
```

Which expands to:

```swift
var x: Int // comment {
  get {
    return 1
  }
}
```

Maybe we should do something about that comment trivia… could we place the comment trivia after (to the right of) the opening curly brace?

```swift
var x: Int { // comment
  get {
    return 1
  }
}
```

Could we enclose the trailing trivia with asterisks?

```swift
var x: Int /* comment */ {
  get {
    return 1
  }
}
```

### Steps to Reproduce

Here is a diff on `swift-syntax` to repro:

```
diff --git a/Tests/SwiftSyntaxMacroExpansionTest/AccessorMacroTests.swift b/Tests/SwiftSyntaxMacroExpansionTest/AccessorMacroTests.swift
index 471a34c7..97e2b8bc 100644
--- a/Tests/SwiftSyntaxMacroExpansionTest/AccessorMacroTests.swift
+++ b/Tests/SwiftSyntaxMacroExpansionTest/AccessorMacroTests.swift
@@ -43,6 +43,24 @@ fileprivate struct ConstantOneGetter: AccessorMacro {
 
 final class AccessorMacroTests: XCTestCase {
   private let indentationWidth: Trivia = .spaces(2)
+  
+  func testAccessorWithComment() {
+    assertMacroExpansion(
+      """
+      @constantOne
+      var x: Int // comment
+      """,
+      expandedSource: """
+        var x: Int {
+          get {
+            return 1
+          }
+        }
+        """,
+      macros: ["constantOne": ConstantOneGetter.self],
+      indentationWidth: indentationWidth
+    )
+  }
 
   func testAccessorOnVariableDeclWithExistingGetter() {
     assertMacroExpansion(

```

AppAppWorks:
Fixed #2614

### DIFF
--- a/Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
+++ b/Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
@@ -923,7 +923,9 @@ private class MacroApplication<Context: MacroExpansionContext>: SyntaxRewriter {
       return DeclSyntax(node)
     }
 
-    guard node.bindings.count == 1, let binding = node.bindings.first else {
+    guard node.bindings.count == 1,
+      var binding = node.bindings.first
+    else {
       contextGenerator(Syntax(node)).addDiagnostics(
         from: MacroApplicationError.accessorMacroOnVariableWithMultipleBindings,
         node: node
@@ -931,17 +933,27 @@ private class MacroApplication<Context: MacroExpansionContext>: SyntaxRewriter {
       return DeclSyntax(node)
     }
 
-    let expansion = expandAccessors(of: node, existingAccessors: binding.accessorBlock)
+    var expansion = expandAccessors(of: node, existingAccessors: binding.accessorBlock)
+
     if expansion.accessors != binding.accessorBlock {
+      if binding.accessorBlock == nil {
+        // remove the trailing trivia of the variable declaration and move it
+        // to the trailing trivia of the left brace of the newly created accessor block
+        expansion.accessors?.leftBrace.trailingTrivia = binding.trailingTrivia
+        binding.trailingTrivia = []
+      }
+
       if binding.initializer != nil, expansion.expandsGetSet {
         // The accessor block will have a leading space, but there will already be a
         // space between the variable and the to-be-removed initializer. Remove the
         // leading trivia on the accessor block so we don't double up.
-        node.bindings[node.bindings.startIndex].accessorBlock = expansion.accessors?.with(\.leadingTrivia, [])
-        node.bindings[node.bindings.startIndex].initializer = nil
+        binding.accessorBlock = expansion.accessors?.with(\.leadingTrivia, [])
+        binding.initializer = nil
       } else {
-        node.bindings[node.bindings.startIndex].accessorBlock = expansion.accessors
+        binding.accessorBlock = expansion.accessors
       }
+
+      node.bindings = [binding]
     }
 
     return DeclSyntax(node)
--- a/Tests/SwiftSyntaxMacroExpansionTest/AccessorMacroTests.swift
+++ b/Tests/SwiftSyntaxMacroExpansionTest/AccessorMacroTests.swift
@@ -44,6 +44,127 @@ fileprivate struct ConstantOneGetter: AccessorMacro {
 final class AccessorMacroTests: XCTestCase {
   private let indentationWidth: Trivia = .spaces(2)
 
+  func testAccessorOnVariableDeclWithTrailingLineCommentAndNoAccessorBlock() {
+    assertMacroExpansion(
+      """
+      @constantOne
+      var x: Int /*1*/ // hello
+      """,
+      expandedSource: """
+        var x: Int { /*1*/ // hello
+          get {
+            return 1
+          }
+        }
+        """,
+      macros: ["constantOne": ConstantOneGetter.self],
+      indentationWidth: indentationWidth
+    )
+
+    assertMacroExpansion(
+      """
+      @constantOne
+      var x: Int /// hello
+      """,
+      expandedSource: """
+        var x: Int { /// hello
+          get {
+            return 1
+          }
+        }
+        """,
+      macros: ["constantOne": ConstantOneGetter.self],
+      indentationWidth: indentationWidth
+    )
+
+    assertMacroExpansion(
+      """
+      @constantOne
+      var x: Int = 1 /// hello
+      """,
+      expandedSource: """
+        var x: Int { /// hello
+          get {
+            return 1
+          }
+        }
+        """,
+      macros: ["constantOne": ConstantOneGetter.self],
+      indentationWidth: indentationWidth
+    )
+
+    assertMacroExpansion(
+      """
+      @constantOne
+      var x /// hello
+      """,
+      expandedSource: """
+        var x { /// hello
+          get {
+            return 1
+          }
+        }
+        """,
+      macros: ["constantOne": ConstantOneGetter.self],
+      indentationWidth: indentationWidth
+    )
+  }
+
+  func testAccessorOnVariableDeclWithTrailingLineCommentAndAccessorBlock() {
+    assertMacroExpansion(
+      """
+      @constantOne
+      var x: Int /*h*/ { // hello
+        1
+      }
+      """,
+      expandedSource: """
+        var x: Int /*h*/ { // hello
+        get {
+            1
+        }
+          get {
+            return 1
+          }
+        }
+        """,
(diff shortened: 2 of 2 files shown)


Tests/SwiftSyntaxMacroExpansionTest/AccessorMacroTests.swift
```diff
@@ -44,6 +44,95 @@ fileprivate struct ConstantOneGetter: AccessorMacro {
...
+      expandedSource: """
+        var x: Int { // hello
+        get {
+            1
+        }
+          get {
+            return 1
+          }
+        }
+        """,
```
AppAppWorks, line 130: I discovered that adding a trailing comment to the left brace of the accessor block will mess up the indentation levels of the generated getter.

AppAppWorks:
@bnbarham could you review this PR if @ahoppen isn't available?

Tests/SwiftSyntaxMacroExpansionTest/AccessorMacroTests.swift
```diff
@@ -44,6 +44,95 @@ fileprivate struct ConstantOneGetter: AccessorMacro {
...
   private let indentationWidth: Trivia = .spaces(2)
 
+  func testAccessorOnVariableDeclWithTrailingLineCommentAndNoAccessorBlock() {
+    assertMacroExpansion(
+      """
+      @constantOne
+      var x: Int // hello
+      """,
+      expandedSource: """
+        var x: Int /* hello*/ {
```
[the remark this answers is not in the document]
AppAppWorks: I was bikeshedding between keeping the style intact or converting them into block, I picked the conversion route because it was slightly easier to code ;)

I agree source fidelity is important, I'll make some changes in the next commit.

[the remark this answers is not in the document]
AppAppWorks: @bnbarham, please have a look at my latest commit.

Tests/SwiftSyntaxMacroExpansionTest/AccessorMacroTests.swift
```diff
@@ -44,6 +44,95 @@ fileprivate struct ConstantOneGetter: AccessorMacro {
...
   private let indentationWidth: Trivia = .spaces(2)
 
+  func testAccessorOnVariableDeclWithTrailingLineCommentAndNoAccessorBlock() {
+    assertMacroExpansion(
+      """
+      @constantOne
+      var x: Int /*1*/ // hello
+      """,
+      expandedSource: """
+        var x: Int /*1*/  {// hello
```
bnbarham, line 54: I wouldn't worry so much about only moving `//`/`///` style comments and instead just move all of the trailing trivia. That would also help with fixing the double space here (and we can just make sure to insert a space between `Int` and `{`.

AppAppWorks: Thanks, I've made changes accordingly, please have a look.

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
```diff
@@ -923,24 +923,34 @@ private class MacroApplication<Context: MacroExpansionContext>: SyntaxRewriter {
...
       contextGenerator(Syntax(node)).addDiagnostics(
         from: MacroApplicationError.accessorMacroOnVariableWithMultipleBindings,
         node: node
       )
       return DeclSyntax(node)
     }
-
-    let expansion = expandAccessors(of: node, existingAccessors: binding.accessorBlock)
+    
+    let bindingKeyPath = \VariableDeclSyntax.bindings[node.bindings.startIndex]
```
bnbarham, line 934: I mildly prefer just keeping `node.bindings[node.bindings.startIndex]` TBH. Maybe we should have a `mutate` method or something similar to make this a little nicer, or just wait until we can use variadic generics and then update `with`...

bnbarham (approved):
Thanks @AppAppWorks!

bnbarham:
@swift-ci please test

AppAppWorks:
I've run swift-format, please test again.

bnbarham:
@swift-ci please test

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
ahoppen: I agree that the bindings key path is a little confusing, it just seems a little over-complicated

What do you think about having `var binding = node.bindings.first` in the `guard`, then using `binding.accessorBlock = …` etc below and finalizing the new node by setting `node.bindings = [binding]`. That seems like the cleanest solution to me.

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
```diff
@@ -923,24 +923,34 @@ private class MacroApplication<Context: MacroExpansionContext>: SyntaxRewriter {
...
+    let binding = node[keyPath: bindingKeyPath]
+    var expansion = expandAccessors(of: node, existingAccessors: binding.accessorBlock)
+
     if expansion.accessors != binding.accessorBlock {
+      if binding.accessorBlock == nil {
+        // remove the trailing trivia of the variable declaration and move it
+        // to the trailing trivia of the left brace of the newly created accessor block
+        node[keyPath: bindingKeyPath].trailingTrivia = []
+        expansion.accessors?.leftBrace.trailingTrivia = binding.trailingTrivia
+      }
```
ahoppen, line 944: This would still be an issue if the macro expands to an accessor that doesn’t have a newline after the `{`, right? Ie. you have a macro that expands to `{ 1 }` and has formatting disabled, so no newline gets added by `BasicFormat`.

If that’s the case, I would just move the trailing trivia after the closing brace. That way you can conceptually think of the accessor as being inserted before the binding’s trailing trivia.

AppAppWorks: How could I disable formatting for a test case?

bnbarham: You'll need to add a new macro and then set `static let formatMode: FormatMode = .disabled` in its definition. Also, nice catch @ahoppen! I forgot about that case :)

AppAppWorks: I couldn't figure out how to make an `AccessorMacro` expand into something like `{ 1 }`, how could I do this?

ahoppen: The following should test it

```swift
func testAccessorOnVariableDeclWithTrailingLineCommentAndAccessorBlock2() {
  struct ConstantOneSingleLineGetter: AccessorMacro {
    static var formatMode: FormatMode { .disabled }

    static func expansion(
      of node: AttributeSyntax,
      providingAccessorsOf declaration: some DeclSyntaxProtocol,
    ) throws -> [AccessorDeclSyntax] {
      return [
        """
        get { 1 }
        """
      ]
    }
  }

  assertMacroExpansion(
    """
    @constantOne
    var x: Int // hello
    """,
    expandedSource: """
      var x: Int { // hello
        get { 1 }
      }
      """,
    macros: ["constantOne": ConstantOneSingleLineGetter.self],
  )
}
```

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@AppAppWorks Could you add the test case I suggested in https://github.com/swiftlang/swift-syntax/pull/2722#discussion_r1681564244?

AppAppWorks:
> @AppAppWorks Could you add the test case I suggested in [#2722 (comment)](https://github.com/swiftlang/swift-syntax/pull/2722#discussion_r1681564244)?

Done, please have a look.

ahoppen (approved):

ahoppen:
@swift-ci Please test

AppAppWorks:
Could you test Windows too? :)

bnbarham:

PR #2725, opened 2024-07-12
Declare rules_swift max_compatibility_version = 2, set rules_swift to 1.18.0

erikkerber:
This enables repos using `rules_swift` 2.0+ to work with swift-syntax.

swift-syntax does not use any features unique to 1.x, nor have any implicit dependencies on 1.x rules. Tested via overriding `rules_apple` to sha `c8a6f615e37c8781e28c65dafed03ecb378281b3`, which sets rules_swift to 2.1.0.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -6,4 +6,4 @@ module(
 
 bazel_dep(name = "apple_support", version = "1.13.0", repo_name = "build_bazel_apple_support")
 bazel_dep(name = "rules_apple", version = "3.3.0", repo_name = "build_bazel_rules_apple")
-bazel_dep(name = "rules_swift", version = "1.16.0", repo_name = "build_bazel_rules_swift")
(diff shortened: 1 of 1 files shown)


keith:
don't we have to port these? https://github.com/apple/swift-syntax/blob/70e374101af89aaf998d04587455f0f8d4970751/BUILD.bazel#L310

I think we have to bump rules_apple as well here?

keith:
hrm i guess we kept tags compat? because this does build for me locally after bumping rules_apple, also can you buildifier the module.bazel

erikkerber:
> I think we have to bump rules_apple as well here?

I didn't intent to bump `rules_swift` to 2.1.0, and just undid that. I assume any consuming repo that bumps to rules_swift 2.0+ will also override rules_apple to a compatible version as well.

keith:
@swift-ci please test

keith (approved):

PR #2726, opened 2024-07-12
[Perf] Improve SyntaxRewriter visitation performance

rintaro:
* Introduce 'SyntaxNodeFactory'. This is an improved replacement of "`Syntax.Info` reusing mechanism" in `SyntaxVisitor`. Adopt it in both `SyntaxVisitor` and `SyntaxRewriter`
* Improve new layout creation:
  * Use a manually allocated 'UnsafeBufferPointer' to avoid 'Array' related overhead
  * Pre-initialize the buffer with old layout at once, then only update each rewritten child element. This is simpler and faster than appending each element.

### DIFF
--- a/CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxRewriterFile.swift
+++ b/CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxRewriterFile.swift
@@ -44,9 +44,29 @@ let syntaxRewriterFile = SourceFileSyntax(leadingTrivia: copyrightHeader) {
 
     DeclSyntax(
       """
+      /// `Syntax.Info` salvaged from the node being deinitialized in 'visitChildren'.
+      ///
+      /// Instead of deallocating them and allocating memory for new syntax nodes, store the allocated memory in an array.
+      /// We can then re-use them to create new syntax nodes.
+      ///
+      /// The array's size should be a typical nesting depth of a Swift file. That way we can store all allocated syntax
+      /// nodes when unwinding the visitation stack.
+      ///
+      /// The actual `info` stored in the `Syntax.Info` objects is garbage. It needs to be set when any of the `Syntax.Info`
+      /// objects get re-used.
+      ///
+      /// Note: making the element non-nil causes 'swift::runtime::SwiftTLSContext::get()' traffic somehow.
+      private var recyclableNodeInfos: ContiguousArray<Syntax.Info?>
+      """
+    )
+
+    DeclSyntax(
+      """
       public init(viewMode: SyntaxTreeViewMode = .sourceAccurate) {
         self.viewMode = viewMode
         self.arena = nil
+        self.recyclableNodeInfos = []
+        self.recyclableNodeInfos.reserveCapacity(64)
       }
       """
     )
@@ -57,6 +77,8 @@ let syntaxRewriterFile = SourceFileSyntax(leadingTrivia: copyrightHeader) {
       public init(viewMode: SyntaxTreeViewMode = .sourceAccurate, arena: SyntaxArena? = nil) {
         self.viewMode = viewMode
         self.arena = arena
+        self.recyclableNodeInfos = []
+        self.recyclableNodeInfos.reserveCapacity(64)
       }
       """
     )
@@ -65,7 +87,8 @@ let syntaxRewriterFile = SourceFileSyntax(leadingTrivia: copyrightHeader) {
       """
       /// Rewrite `node`, keeping its parent unless `detach` is `true`.
       public func rewrite(_ node: some SyntaxProtocol, detach: Bool = false) -> Syntax {
-        let rewritten = self.dispatchVisit(Syntax(node))
+        var rewritten = Syntax(node)
+        self.dispatchVisit(&rewritten)
         if detach {
           return rewritten
         }
@@ -126,7 +149,9 @@ let syntaxRewriterFile = SourceFileSyntax(leadingTrivia: copyrightHeader) {
       ///   - Returns: the rewritten node
       @available(*, deprecated, renamed: "rewrite(_:detach:)")
       public func visit(_ node: Syntax) -> Syntax {
-        return dispatchVisit(node)
+        var rewritten = node
+        dispatchVisit(&rewritten)
+        return rewritten
       }
       """
     )
@@ -134,7 +159,9 @@ let syntaxRewriterFile = SourceFileSyntax(leadingTrivia: copyrightHeader) {
     DeclSyntax(
       """
       public func visit<T: SyntaxChildChoices>(_ node: T) -> T {
-        return dispatchVisit(Syntax(node)).cast(T.self)
+        var rewritten = Syntax(node)
+        dispatchVisit(&rewritten)
+        return rewritten.cast(T.self)
       }
       """
     )
@@ -177,7 +204,9 @@ let syntaxRewriterFile = SourceFileSyntax(leadingTrivia: copyrightHeader) {
         ///   - Returns: the rewritten node
         \(baseNode.apiAttributes())\
         public func visit(_ node: \(baseKind.syntaxType)) -> \(baseKind.syntaxType) {
-          return dispatchVisit(Syntax(node)).cast(\(baseKind.syntaxType).self)
+          var node: Syntax = Syntax(node)
+          dispatchVisit(&node)
+          return node.cast(\(baseKind.syntaxType).self)
         }
         """
       )
@@ -187,21 +216,16 @@ let syntaxRewriterFile = SourceFileSyntax(leadingTrivia: copyrightHeader) {
       """
       /// Interpret `node` as a node of type `nodeType`, visit it, calling
       /// the `visit` to transform the node.
+      @inline(__always)
       private func visitImpl<NodeType: SyntaxProtocol>(
-        _ node: Syntax,
+        _ node: inout Syntax,
         _ nodeType: NodeType.Type,
         _ visit: (NodeType) -> some SyntaxProtocol
-      ) -> Syntax {
-        let castedNode = node.cast(NodeType.self)
-        // Accessing _syntaxNode directly is faster than calling Syntax(node)
-        visitPre(node)
-        defer {
-          visitPost(node)
-        }
-        if let newNode = visitAny(node) {
-          return newNode
-        }
-        return Syntax(visit(castedNode))
+      ) {
+        let origNode = node
+        visitPre(origNode)
+        node = visitAny(origNode) ?? Syntax(visit(origNode.cast(NodeType.self)))
+        visitPost(origNode)
       }
       """
     )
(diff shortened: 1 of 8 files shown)


rintaro:
@swift-ci Please test

CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxRewriterFile.swift
```diff
@@ -329,10 +353,18 @@ let syntaxRewriterFile = SourceFileSyntax(leadingTrivia: copyrightHeader) {
           }
 
           // Build the Syntax node to rewrite
-          let absoluteRaw = AbsoluteRawSyntax(raw: child, info: info)
+          var childNode: Syntax
+          if !recyclableNodeInfos.isEmpty {
+            let recycledInfo: Syntax.Info = recyclableNodeInfos.removeLast()!
+            recycledInfo.info = .nonRoot(.init(parent: Syntax(node), absoluteInfo: info))
```
rintaro, line 359:
```suggestion
            recycledInfo.info = .nonRoot(.init(parent: syntaxNode, absoluteInfo: info))
```

CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxRewriterFile.swift
```diff
@@ -187,21 +216,16 @@ let syntaxRewriterFile = SourceFileSyntax(leadingTrivia: copyrightHeader) {
...
-          visitPost(node)
-        }
-        if let newNode = visitAny(node) {
-          return newNode
-        }
-        return Syntax(visit(castedNode))
+      ) {
+        let origNode = node
+        visitPre(origNode)
+        node = visitAny(origNode) ?? Syntax(visit(origNode.cast(NodeType.self)))
```
rintaro, line 227:
```suggestion
        node = visitAny(origNode) ?? visit(origNode.cast(NodeType.self))._syntaxNode
```

rintaro: No, no effect.

rintaro:
Too many changes in this PR. I will split this into several PRs.

CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxRewriterFile.swift
```diff
@@ -42,6 +42,13 @@ let syntaxRewriterFile = SourceFileSyntax(leadingTrivia: copyrightHeader) {
       """
     )
 
+    DeclSyntax(
+      """
+      /// 'Syntax' object factory recycling 'Syntax.Info' instances.
+      private let nodeFactory: SyntaxNodeFactory = SyntaxNodeFactory()
```
rintaro, line 48: `SyntaxVisitor` should also use this as well.

rintaro:
@swift-ci Please test

CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxRewriterFile.swift
```diff
@@ -305,77 +312,52 @@ let syntaxRewriterFile = SourceFileSyntax(leadingTrivia: copyrightHeader) {
...
+        for case let (child?, info) in RawSyntaxChildren(node) where viewMode.shouldTraverse(node: child) {
 
           // Build the Syntax node to rewrite
-          let absoluteRaw = AbsoluteRawSyntax(raw: child, info: info)
+          var childNode = nodeFactory.create(parent: node, raw: child, absoluteInfo: info)
 
-          let rewritten = dispatchVisit(Syntax(absoluteRaw, parent: syntaxNode))
-          if rewritten.id != info.nodeId {
+          dispatchVisit(&childNode)
+          if childNode.raw.id != child.id {
```
rintaro, line 327: Previously it compared `SyntaxIdentifier` which takes the entire tree into account.
Now it compares `RawSyntax.ID` because we only care the `RawSyntax` here.

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

rintaro:
@swift-ci Please test Windows

rintaro:
@swift-ci Please clean test Windows

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Linux

CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxRewriterFile.swift
```diff
@@ -187,21 +201,16 @@ let syntaxRewriterFile = SourceFileSyntax(leadingTrivia: copyrightHeader) {
       """
       /// Interpret `node` as a node of type `nodeType`, visit it, calling
       /// the `visit` to transform the node.
+      @inline(__always)
       private func visitImpl<NodeType: SyntaxProtocol>(
-        _ node: Syntax,
+        _ node: inout Syntax,
```
ahoppen, line 206: Out of curiosity: Do you know why `inout` is faster here? Does it avoid reference counting?

Sources/SwiftSyntax/generated/SyntaxRewriter.swift
```diff
@@ -3898,79 +3909,52 @@ open class SyntaxRewriter {
...
     // nodes are being collected.
-    var newLayout: ContiguousArray<RawSyntax?>?
+    var newLayout: UnsafeMutableBufferPointer<RawSyntax?> = .init(start: nil, count: 0)
 
-    // Rewritten children just to keep their 'SyntaxArena' alive until they are
-    // wrapped with 'Syntax'
-    var rewrittens: ContiguousArray<Syntax> = []
+    // Keep 'SyntaxArena' of rewritten nodes alive until they are wrapped
+    // with 'Syntax'
+    var rewrittens: ContiguousArray<RetainedSyntaxArena> = []
```
ahoppen, line 3916: What are the performance characteristics of `ContiguousArray` vs `Array` here? I would expect `Array` to be faster because it doesn’t need to move all elements if it runs out of space. And I don’t think we need the contiguous property here anymore.

Sources/SwiftSyntax/generated/SyntaxRewriter.swift
```diff
@@ -3898,79 +3909,52 @@ open class SyntaxRewriter {
...
+      newLayout.deallocate()
       // 'withExtendedLifetime' to keep 'SyntaxArena's of them alive until here.
       return withExtendedLifetime(rewrittens) {
-        Syntax(raw: newRaw, rawNodeArena: arena).cast(SyntaxType.self)
+        Syntax(raw: newRaw, rawNodeArena: arena)
       }
     } else {
       // No child node was rewritten. So no need to change this node as well.
-      return node
+      return nil
```
ahoppen, line 3957: Is there a benefit from returning `nil` vs `node` here? AFAICT we always to `?? node` on every call and we could avoid that if `visitChildren` didn’t return an optional.

Sources/SwiftSyntax/SyntaxArenaAllocatedBuffer.swift
```diff
@@ -107,4 +107,8 @@ public struct SyntaxArenaAllocatedBufferPointer<Element: Sendable>: RandomAccess
   var unsafeRawBufferPointer: UnsafeRawBufferPointer {
     return UnsafeRawBufferPointer(buffer)
   }
+
+  public func withContiguousStorageIfAvailable<R>(_ body: (UnsafeBufferPointer<Element>) throws -> R) rethrows -> R? {
+    try body(buffer)
+  }
```
ahoppen, line 113: I this used anywhere?

cmake/modules/AddSwiftHostLibrary.cmake
```diff
@@ -92,7 +92,7 @@ function(add_swift_syntax_library name)
   add_custom_command(
     TARGET ${target}
     POST_BUILD
-    COMMAND "${CMAKE_COMMAND}" -E touch_nocreate $<TARGET_FILE:${target}> $<TARGET_OBJECTS:${target}> "${module_base}"
```
ahoppen, line 95: Is this change related?

ahoppen (commented):
Looks good. Just a couple minor comments.

cmake/modules/AddSwiftHostLibrary.cmake
bnbarham: In an unfortunate way. The windows build fails with a fairly cryptic error, my best guess is that we're overflowing the command (either in number of arguments or length) with the extra file @rintaro added. This whole command is really just a hack to help with incremental builds, but we should check if it's needed in 3.26/3.29 and if not, I'm happy enough to just remove it.

rintaro:
@swift-ci Please test Linux

CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxRewriterFile.swift
rintaro: I actually haven't measured performance difference between `inout` or not. I just followed `SyntaxVisitor` pattern here.

Sources/SwiftSyntax/generated/SyntaxRewriter.swift
rintaro: Hm, I haven't tried `Array` here. interesting.

Sources/SwiftSyntax/generated/SyntaxRewriter.swift
rintaro: Oh true, will try.

Sources/SwiftSyntax/SyntaxArenaAllocatedBuffer.swift
rintaro: Yes, `UnsafeMutableBufferPointer.initialize(fromContentsOf:)` have a fast path calling this witness.

CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxRewriterFile.swift
rintaro: Also since we do `isKnownUniquelyReferenced`, we can't do like `let newNode = dispatchVisit(oldNode)` because `newNode` and `oldNode` might be the same. So we need: `node = dispatchVisit(node)` anyway. which is semantically what `inout` does.

Sources/SwiftSyntax/generated/SyntaxRewriter.swift
rintaro: Ah e.g. `visitChildren(node._syntaxNode)?.cast(AccessorEffectSpecifiersSyntax.self) ?? node`
For this pattern, `.cast(AccessorEffectSpecifiersSyntax.self)` have some cost (even after `__shared` change). So being able to use `node` as-is was faster

CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxRewriterFile.swift
ahoppen: Ah, I see. Makes sense 👍🏽

Sources/SwiftSyntax/SyntaxArenaAllocatedBuffer.swift
ahoppen: Ah, I didn’t think about this satisfying a defaulted requirement in `RandomAccessCollection`.

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

Sources/SwiftSyntax/generated/SyntaxRewriter.swift
rintaro: Array always uses contiguous storage unless it's backed by `NSArray`.
Although there's no particular reason to use `ContiguousArray`, its implementation is simpler than `Array`. So let's keep this as-is

rintaro:
@swift-ci Please test Windows

Sources/SwiftSyntax/generated/SyntaxRewriter.swift
rintaro: Sorry I re-measured and I don't see any difference returning `node` instead of `nil`. So I changed it to return `node` and remove `?? node` in the caller.

ahoppen (approved):

PR #2727, opened 2024-07-13
improve diagnostic for unnamed closure parameters

### LINKED ISSUE #2340: Improve diagnostic for unnamed closure parameter
### Issue Kind

Bad Diagnostic Produced

### Source Code

```swift
test { ([X]) in }
```

### Description

New parser:
```
error: unexpected code '[X]' in parameter clause
test { ([X]) in }
        ^~~
```

Old parser:
```
warning: unnamed parameters must be written with the empty name '_' [parameter_unnamed_warn]
test { ([X]) in }
        ^
        _:
```

This is a special case. More generally, the old parser is capable of diagnosing this kind of closure parameter production gracefully when all of the following are true:

* An argument name cannot be parsed ([link](https://github.com/apple/swift/blob/17c30f162d1fd69ae8a511e2496559090829ce51/lib/Parse/ParsePattern.cpp#L311))
* A type can be parsed per `Parser::canParseType()`, and said type is followed by `,`, `)`, or `=` ([link](https://github.com/apple/swift/blob/17c30f162d1fd69ae8a511e2496559090829ce51/lib/Parse/ParsePattern.cpp#L367))
___
Related to #2331.

AppAppWorks:
fix #2340

### DIFF
--- a/Sources/SwiftParser/Parameters.swift
+++ b/Sources/SwiftParser/Parameters.swift
@@ -160,10 +160,23 @@ extension Parser {
     let misplacedSpecifiers = parseMisplacedSpecifiers()
 
     let names = self.parseParameterNames()
-    let colon = self.consume(if: .colon)
+    var colon = self.consume(if: .colon)
+    // try to parse the type regardless of the presence of the preceding colon
+    // to tackle any unnamed parameter or missing colon
+    // e.g. [X], (:[X]) or (x [X])
+    let canParseType = withLookahead { $0.canParseType() }
     let type: RawTypeSyntax?
-    if colon != nil {
+    if canParseType {
       type = self.parseType(misplacedSpecifiers: misplacedSpecifiers)
+      if colon == nil {
+        // mark the preceding colon as missing if the type is present
+        // e.g. [X] or (x [X])
+        colon = missingToken(.colon)
+      }
+    } else if colon != nil {
+      // mark the type as missing if the preceding colon is present
+      // e.g. (:) or (_:)
+      type = RawTypeSyntax(RawMissingTypeSyntax(arena: self.arena))
     } else {
       type = nil
     }
--- a/Tests/SwiftParserTest/ExpressionTests.swift
+++ b/Tests/SwiftParserTest/ExpressionTests.swift
@@ -2875,6 +2875,77 @@ final class StatementExpressionTests: ParserTestCase {
     )
   }
 
+  func testClosureWithMalformedParameters() {
+    assertParse(
+      """
+      test { (1️⃣[X]) in }
+      """,
+      diagnostics: [
+        DiagnosticSpec(message: "expected identifier and ':' in parameter", fixIts: ["insert identifier and ':'"])
+      ],
+      fixedSource: """
+        test { (<#identifier#>: [X]) in }
+        """
+    )
+
+    assertParse(
+      """
+      test { (1️⃣: [X]) in }
+      """,
+      diagnostics: [
+        DiagnosticSpec(message: "expected identifier in parameter", fixIts: ["insert identifier"])
+      ],
+      fixedSource: """
+        test { (<#identifier#>: [X]) in }
+        """
+    )
+
+    assertParse(
+      """
+      test { (1️⃣:2️⃣) in }
+      """,
+      diagnostics: [
+        DiagnosticSpec(
+          locationMarker: "1️⃣",
+          message: "expected identifier in parameter",
+          fixIts: ["insert identifier"]
+        ),
+        DiagnosticSpec(
+          locationMarker: "2️⃣",
+          message: "expected type in parameter",
+          fixIts: ["insert type"]
+        ),
+      ],
+      fixedSource: """
+        test { (<#identifier#>: <#type#>) in }
+        """
+    )
+
+    assertParse(
+      """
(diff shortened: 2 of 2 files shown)


AppAppWorks:
@bnbarham would you mind reviewing this PR too? :)

Sources/SwiftParser/Parameters.swift
```diff
@@ -160,11 +160,20 @@ extension Parser {
...
     let names = self.parseParameterNames()
-    let colon = self.consume(if: .colon)
-    let type: RawTypeSyntax?
-    if colon != nil {
-      type = self.parseType(misplacedSpecifiers: misplacedSpecifiers)
-    } else {
+    var colon = self.consume(if: .colon)
+    // try to parse the type regardless of the presence of the preceding colon
+    // because the parameter may be unnamed
+    // e.g. { [X] in } or { (:[X]) in }
```
rintaro, line 166: Are these relevant here? From the code of `parseParameterClause`, I feel like `parseClosureParameter` is not even called.
If it does, could you add test cases for these patterns?

Sources/SwiftParser/Parameters.swift
```diff
@@ -160,11 +160,20 @@ extension Parser {
...
+    var type: RawTypeSyntax? = self.parseType(misplacedSpecifiers: misplacedSpecifiers)
```
rintaro, line 167: I feel this a bit scary. Could you `Lookahead.canParseType()` and check if the next token is `,` or `)`, instead of checking `.missingType`.

Tests/SwiftParserTest/translated/RecoveryTests.swift
```diff
@@ -3355,4 +3355,19 @@ final class RecoveryTests: ParserTestCase {
       ]
     )
   }
+  
+  func testRecovery185() {
+    // <rdar://1181099123>
+    assertParse(
+      """
+      test { (1️⃣[X]) in }
```
rintaro, line 3363: Do we have a test case for missing label but with colon? e.g. `test { (: FooBar) in }`

bnbarham: With an attribute would also be interesting, ie. `test { (foo @bar baz) in }`.

Also note that all the `translated/*` tests were translated from the compiler parser tests. It'd probably be better to have this one appropriately named in `ExpressionTests` (though in an ideal world we'd split that file up too).

Sources/SwiftParser/Parameters.swift
AppAppWorks: I may just trim the braces from the examples in the comments.

Sources/SwiftParser/Parameters.swift
AppAppWorks: I wasn't aware of the existence of `Lookahead` until earlier today, sorry for that ;(

Tests/SwiftParserTest/translated/RecoveryTests.swift
AppAppWorks: > Do we have a test case for missing label but with colon? e.g. `test { (: FooBar) in }`

For some reason I deleted that test case from the final commit, I'll add it back.

AppAppWorks:
Updated, please have a look.

ahoppen (approved):
Thank you @AppAppWorks. Looks great!

ahoppen:
@swift-ci Please test

AppAppWorks:
@ahoppen I've found that some of my code comments are misleading and the test case for missing both identifier and type is missing.

I've rewritten the comments and added a new test as follows, if you also think they're necessary and good, I'll force-commit again.

```swift
let names = self.parseParameterNames()
var colon = self.consume(if: .colon)
// try to parse the type regardless of the presence of the preceding colon
// to tackle any unnamed parameter or missing colon
// e.g. [X], (:[X]) or (x [X])
let canParseType = withLookahead { $0.canParseType() }
let type: RawTypeSyntax?
if canParseType {
  type = self.parseType(misplacedSpecifiers: misplacedSpecifiers)
  if colon == nil {
    // mark the preceding colon as missing if the type is present
    // e.g. [X] or (x [X])
    colon = missingToken(.colon)
  }
} else if colon != nil {
  // mark the type as missing if the preceding colon is present
  // e.g. (:) or (_:)
  type = RawTypeSyntax(RawMissingTypeSyntax(arena: self.arena))
} else {
  type = nil
}
```

```swift
assertParse(
  """
  test { (1️⃣:2️⃣) in }
  """,
  diagnostics: [
    DiagnosticSpec(
      locationMarker: "1️⃣",
      message: "expected identifier in parameter",
      fixIts: ["insert identifier"]
    ),
    DiagnosticSpec(
      locationMarker: "2️⃣",
      message: "expected type in parameter",
      fixIts: ["insert type"]
    ),
  ],
  fixedSource: """
    test { (<#identifier#>: <#type#>) in }
    """
)
```

ahoppen:
Force push your improved changes and I’ll review them again.

AppAppWorks:
Pushed, please check.

ahoppen (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

PR #2729, opened 2024-07-15
Turn some static variables into static stored constants

rintaro:
Computing `OperatorTable.standardOperators` is not trivial, make it `static let` instead of computed static variable.
So as `SyntaxProtocol.structure` implementations.

### DIFF
--- a/CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxNodesFile.swift
+++ b/CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxNodesFile.swift
@@ -199,17 +199,14 @@ func syntaxNode(nodesStartingWith: [Character]) -> SourceFileSyntax {
           }
         }
 
-        try! VariableDeclSyntax("public static var structure: SyntaxNodeStructure") {
-          let layout = ArrayExprSyntax {
-            for child in node.children {
-              ArrayElementSyntax(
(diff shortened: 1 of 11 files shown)


rintaro:
@swift-ci Please test

Sources/SwiftSyntax/SyntaxProtocol.swift
```diff
@@ -225,13 +225,10 @@ extension SyntaxProtocol {
   }
 
   public var keyPathInParent: AnyKeyPath? {
-    guard let parent = self.parent else {
-      return nil
-    }
-    guard case .layout(let childrenKeyPaths) = parent.kind.syntaxNodeType.structure else {
+    guard case .layout(let childrenKeyPaths) = self._syntaxNode.parent?.kind.syntaxNodeType.structure else {
```
ahoppen, line 228: Is `self._syntaxNode.parent` faster than `parent`?

ahoppen (approved):

Sources/SwiftSyntax/SyntaxProtocol.swift
rintaro: That was my measurement, but I'm not 100% sure tbh. I will double check.

rintaro: Removed this commit from the PR for now

rintaro:
@swift-ci Please test

rintaro:
@swift-ci  Please test Windows

rintaro:
@swift-ci Please test Windows

PR #2731, opened 2024-07-15
fixed-typed-throws-after-arrow-fixit

### LINKED ISSUE #2391: Fix-its for misplaced typed throws could be improved
### Issue Kind

Bad Diagnostic Produced

### Source Code

```swift
func foo() -> throws(MyError) Int {}
```

### Description

Applying the Fix-It for the parser diagnostics here produces malformed code, because the thrown error type gets lost. The "fixed" code is:

```swift
func foo() throws -> (MyError) Int {}
```

It should be

```swift
func foo() throws(MyError) -> Int {}
```

AppAppWorks:
fixes #2391

### DIFF
--- a/Sources/SwiftParser/Parser.swift
+++ b/Sources/SwiftParser/Parser.swift
@@ -826,6 +826,24 @@ extension Parser {
   }
 }
 
+// MARK: Marking Tokens As Missing
+extension Parser {
+  private class TokenMissingMaker: SyntaxRewriter {
+    override func visit(_ token: TokenSyntax) -> TokenSyntax {
+      TokenSyntax(token.tokenKind, presence: .missing)
+    }
+  }
+
+  /// Creates a replicate of `syntax` with all tokens marked as missing.
+  func withAllTokensMarkedMissing<T: RawSyntaxNodeProtocol>(syntax: T) -> T {
+    let tokenMissingMaker = TokenMissingMaker(arena: self.arena)
+    let allMissing = tokenMissingMaker.rewrite(
+      Syntax(raw: RawSyntax(syntax), rawNodeArena: self.arena)
+    ).raw
+    return allMissing.cast(T.self)
+  }
+}
+
 extension SyntaxText {
   func withBuffer<Result>(_ body: (UnsafeBufferPointer<UInt8>) throws -> Result) rethrows -> Result {
     try body(UnsafeBufferPointer<UInt8>(start: self.baseAddress, count: self.count))
--- a/Sources/SwiftParser/Specifiers.swift
+++ b/Sources/SwiftParser/Specifiers.swift
@@ -772,16 +772,17 @@ extension Parser {
     _ effectSpecifiers: inout S?
   ) -> RawUnexpectedNodesSyntax? {
     var synthesizedAsync: RawTokenSyntax? = nil
-    var synthesizedThrows: RawTokenSyntax? = nil
-    var unexpected: [RawTokenSyntax] = []
+    var synthesizedThrowsClause: RawThrowsClauseSyntax? = nil
+    var unexpected: [RawSyntax] = []
     var loopProgress = LoopProgressCondition()
+
     while self.hasProgressed(&loopProgress) {
       if let (spec, handle, matchedSubset) = self.at(
         anyIn: S.MisspelledAsyncTokenKinds.self,
         or: S.CorrectAsyncTokenKinds.self
       ) {
         let misspelledAsync = self.eat(handle)
-        unexpected.append(misspelledAsync)
+        unexpected.append(RawSyntax(misspelledAsync))
         if effectSpecifiers?.asyncSpecifier == nil {
           if matchedSubset == S.CorrectAsyncTokenKinds.self {
             synthesizedAsync = missingToken(spec)
@@ -794,23 +795,82 @@ extension Parser {
         or: S.CorrectThrowsTokenKinds.self
       ) {
         let misspelledThrows = self.eat(handle)
-        unexpected.append(misspelledThrows)
-        if effectSpecifiers?.throwsClause == nil {
-          if matchedSubset == S.CorrectThrowsTokenKinds.self {
-            synthesizedThrows = missingToken(spec)
-          } else {
-            synthesizedThrows = missingToken(.throws)
+        unexpected.append(RawSyntax(misspelledThrows))
+        guard effectSpecifiers?.throwsClause == nil else {
+          continue
+        }
+        if matchedSubset == S.CorrectThrowsTokenKinds.self {
+          let synthesizedThrows = missingToken(spec)
+
+          // avoid generating diagnotics that suggest moving `throws` and the following type if there is any trivia
+          // between `throws` and the following left parenthesis, because the following type is likely to be the
+          // return type.
+          // e.g. -> throws (Int, Int)
+          //
+          // prefer pessimistic diagnostics even if they might be suboptimal.
+          // e.g. -> throws (any Error) Int
+          guard misspelledThrows.trailingTriviaByteLength == 0 && self.currentToken.leadingTriviaByteLength == 0 else {
+            synthesizedThrowsClause = RawThrowsClauseSyntax(
+              throwsSpecifier: synthesizedThrows,
+              leftParen: nil,
+              type: nil,
+              rightParen: nil,
+              arena: self.arena
+            )
+            continue
           }
+
+          guard
+            withLookahead({ lookahead in
+              lookahead.consume(if: .leftParen) != nil && lookahead.canParseType()
+            })
+          else {
+            synthesizedThrowsClause = RawThrowsClauseSyntax(
+              throwsSpecifier: synthesizedThrows,
+              leftParen: nil,
+              type: nil,
+              rightParen: nil,
+              arena: self.arena
+            )
+            continue
+          }
+
+          let (unexpectedBeforeLeftParen, leftParen) = self.expect(.leftParen)
+          if let unexpectedBeforeLeftParen {
+            unexpected += unexpectedBeforeLeftParen.elements
+          }
+          unexpected.append(RawSyntax(leftParen))
+          let type = self.parseType()
+          unexpected.append(RawSyntax(type))
+          let (unexpectedBeforeRightParen, rightParen) = self.expect(.rightParen)
+          if let unexpectedBeforeRightParen {
+            unexpected += unexpectedBeforeRightParen.elements
+          }
+          if !rightParen.isMissing {
+            unexpected.append(RawSyntax(rightParen))
+          }
+
+          synthesizedThrowsClause = RawThrowsClauseSyntax(
+            throwsSpecifier: synthesizedThrows,
+            leftParen: missingToken(.leftParen),
+            type: self.withAllTokensMarkedMissing(syntax: type),
+            rightParen: rightParen.isMissing ? nil : missingToken(.rightParen),
+            arena: self.arena
+          )
+        } else {
+          synthesizedThrowsClause = RawThrowsClauseSyntax(
+            throwsSpecifier: missingToken(.throws),
+            leftParen: nil,
+            type: nil,
+            rightParen: nil,
+            arena: self.arena
+          )
         }
       } else {
         break
       }
     }
-    if synthesizedAsync != nil || synthesizedThrows != nil {
-      let synthesizedThrowsClause = synthesizedThrows.map {
-        RawThrowsClauseSyntax(throwsSpecifier: $0, leftParen: nil, type: nil, rightParen: nil, arena: self.arena)
-      }
-
+    if synthesizedAsync != nil || synthesizedThrowsClause != nil {
       if let specifiers = effectSpecifiers {
         effectSpecifiers = specifiers.withMisplaced(
           async: synthesizedAsync,
--- a/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
+++ b/Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
@@ -171,28 +171,33 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
     return handledNodes.contains(node.id)
   }
 
-  /// Utility function to emit a diagnostic that removes a misplaced token and instead inserts an equivalent token at the corrected location.
+  /// Utility function to emit a diagnostic that removes a misplaced node and instead inserts an equivalent node at the corrected location.
   ///
-  /// If `incorrectContainer` contains some tokens that satisfy `unexpectedTokenCondition`, emit a diagnostic with message `message` that marks this token as misplaced.
-  /// If `correctTokens` contains missing tokens, also emit a Fix-It with message `fixIt` that marks the unexpected token as missing and instead inserts `correctTokens`.
-  public func exchangeTokens(
+  /// If `incorrectContainer` contains some nodes that satisfy `unexpectedNodeCondition`, emit a diagnostic with
+  /// message `message` that marks the present tokens in this node as misplaced.
+  ///
+  /// If `correctNodes` contains missing tokens, also emit a Fix-It with message `fixIt` that marks the present tokens
+  /// in unexpected nodes as missing while inserting the missing tokens in `correctNodes`.
+  public func exchangeNodes<S: SyntaxProtocol>(
     unexpected: UnexpectedNodesSyntax?,
-    unexpectedTokenCondition: (TokenSyntax) -> Bool,
-    correctTokens: [TokenSyntax?],
+    unexpectedNodeCondition: (S) -> Bool,
+    correctNodes: [S?],
     message: (_ misplacedTokens: [TokenSyntax]) -> some DiagnosticMessage,
     moveFixIt: (_ misplacedTokens: [TokenSyntax]) -> FixItMessage,
     removeRedundantFixIt: (_ misplacedTokens: [TokenSyntax]) -> FixItMessage? = { _ in nil }
   ) {
     guard let incorrectContainer = unexpected else { return }
-    let misplacedTokens = incorrectContainer.presentTokens(satisfying: unexpectedTokenCondition)
-    if misplacedTokens.isEmpty { return }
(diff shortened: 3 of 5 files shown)


Sources/SwiftParser/Parser.swift
AppAppWorks, line 1: I lament I have to resort to this dirty trick to get all tokens marked missing in the`RawTypeSyntax` returned from `parseType`.

Sources/SwiftParser/Specifiers.swift
```diff
@@ -797,20 +798,64 @@ extension Parser {
...
+              }
+
+              virtualParsedOffset = this.currentToken.start
+              return RawThrowsClauseSyntax(
+                throwsSpecifier: synthesizedThrows,
+                leftParen: leftParen,
+                type: type,
+                rightParen: rightParen,
+                arena: this.arena
+              )
```
AppAppWorks, line 838: I tried to use a lookahead here to "virtually" parse the throwing type but lookahead methods don't return any `RawSyntax` ;(
I've resorted to rewinding `currentToken` and `lexemes` to emulate virtual parsing which I think is super ugly.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
AppAppWorks, line 1: One big issue here is `correctTokens` can no longer be succinctly described by the set of a handful of `TokenSyntax` and `unexpectedTokenCondition` isn't limited to `EffectSpecifier` any more once the throwing type is added to the mix. Is there a good way to validate all tokens of a throwing type?

Sources/SwiftParser/Parser.swift
AppAppWorks, line 1: I lament I have to resort to this dirty trick to get all tokens marked missing in theRawTypeSyntax returned from parseType.

ahoppen: You should be able to re-use the `TokenMissingMaker` we are using here: https://github.com/swiftlang/swift-syntax/blob/70e374101af89aaf998d04587455f0f8d4970751/Sources/SwiftParser/Declarations.swift#L1284-L1293 

If we’re using `TokenMissingMaker`, it would probably make sense to add a method to `Parser` that takes a `RawSyntaxProtocol` and runs `TokenMissingMaker` on it.

Sources/SwiftParser/Specifiers.swift
```diff
@@ -827,4 +872,14 @@ extension Parser {
...
+
+  mutating func virtualParse<T>(block: (inout Self) -> T) -> T {
+    let oldCurrentToken = currentToken
+    let oldLexemes = lexemes
+    defer {
+      currentToken = oldCurrentToken
+      lexemes = oldLexemes
+    }
+    return block(&self)
+  }
```
ahoppen, line 884: Why do we need this? If you want to look ahead, you should be able to use a `Lookahead` by calling `lookahead()` on `Parser`. There are a number of places where we do that. 

But really, after parsing a misplaces `throws`, I would just check if the next token is a `(` which is not separated by any trivia from `throws` (ie. trailing trivia of `throws` is empty and leading trivia of `(` is empty). If it is, just continue parsing the thrown type.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -257,10 +257,16 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
     effectSpecifiers: (some EffectSpecifiersSyntax)?,
     misplacedSpecifiers: UnexpectedNodesSyntax?
   ) {
+    var correctTokens = [effectSpecifiers?.asyncSpecifier, effectSpecifiers?.throwsClause?.throwsSpecifier]
+    if let throwingType = effectSpecifiers?.throwsClause?.type {
+      correctTokens.append(effectSpecifiers?.throwsClause?.leftParen)
+      correctTokens += Array(throwingType.tokens(viewMode: .all))
+      correctTokens.append(effectSpecifiers?.throwsClause?.rightParen)
+    }
```
ahoppen, line 265: Couldn’t this be simplified to

```suggestion
    if let throwsClause = effectSpecifiers?.throwsClause {
      correctTokens += throwsClause.tokens(viewMode: .all)
    }
```

Sources/SwiftParser/Parser.swift
AppAppWorks: Which source file would be the best place to house this new method?

Sources/SwiftParser/Specifiers.swift
```diff
@@ -827,4 +861,35 @@ extension Parser {
...
+
+  func withAllTokensMarkedMissing<T: RawSyntaxNodeProtocol>(source: T) -> T {
+    let tokenMissingMaker = TokenMissingMaker(arena: self.arena)
+    let allMissing = tokenMissingMaker.rewrite(
+      Syntax(raw: RawSyntax(source), rawNodeArena: self.arena)
+    ).raw
+    return allMissing.cast(T.self)
+  }
+}
```
AppAppWorks, line 879: I tentatively placed the code here, not knowing the ideal source file to house it.

Tests/SwiftParserTest/DeclarationTests.swift
```diff
@@ -951,6 +951,41 @@ final class DeclarationTests: ParserTestCase {
...
+    assertParse(
+      "func test() -> 1️⃣throws /* comment */ (any Error) Int",
+      diagnostics: [
+        DiagnosticSpec(
+          message: "'throws /* comment */ (any Error)' must precede '->'",
+          fixIts: ["move 'throws /* comment */ (any Error)' in front of '->'"]
+        )
+      ],
+      fixedSource: "func test() throws(any Error) -> /* comment */  Int"
+    )
```
AppAppWorks, line 986: @ahoppen, you've mentioned we should only continue parsing the type if there are no whitespace and comment trivia between `throws` and `.leftParen`. Here I noticed that the comment trivia is left in place in the fixed source after applying the fix-it even though `/* comment */` is included in the message and fix-it. Is it because of the issue of `parsedToken` vs `materializedToken`?

However, `throws /* comment */ (any Error)` is also a legit throws clause, shouldn't it be supported by this fix-it?

Sources/SwiftParser/Specifiers.swift
ahoppen: How about placing it in Parser.swift, making `TokenMissingMaker` `private` and also using `withAllTokensMarkedMissing` in [Declarations.swift](https://github.com/swiftlang/swift-syntax/blob/70e374101af89aaf998d04587455f0f8d4970751/Sources/SwiftParser/Declarations.swift#L1284-L1293)?

Sources/SwiftParser/Specifiers.swift
```diff
@@ -797,20 +798,53 @@ extension Parser {
...
+                throwsSpecifier: synthesizedThrows,
+                leftParen: nil,
+                type: nil,
+                rightParen: nil,
+                arena: self.arena
+              )
+              break
+            }
+
+            let leftParen = self.consume(if: .leftParen)!
```
ahoppen, line 816: Let’s make this `self.expect`. That way we avoid the force unwrap.

```suggestion
            let leftParen = self.expect(.leftParen)
```

And for symmetry, I would also use `expect` to get the right parenthesis.

Sources/SwiftParser/Specifiers.swift
```diff
@@ -797,20 +798,53 @@ extension Parser {
...
+            synthesizedThrowsClause = RawThrowsClauseSyntax(
+              throwsSpecifier: synthesizedThrows,
+              leftParen: missingToken(.leftParen),
+              type: self.withAllTokensMarkedMissing(source: type),
+              rightParen: rightParen != nil ? missingToken(.rightParen) : nil,
+              arena: self.arena
+            )
+
+            unexpected.append(leftParen)
+            unexpected += type.allTokens
```
ahoppen, line 829: You should be able to shove the entire `type` into `unexpected` if you make it a `[RawSyntax]` instead of `[RawTokenSyntax]`. You can also remove the `allTokens` property then.

Tests/SwiftParserTest/DeclarationTests.swift
ahoppen: Sorry, I should have elaborate when I suggested the no spacing rule: If a user writes `func foo() -> throws (Int, Int)`, the `(Int, Int)` very likely isn’t the thrown error type but the return type and I want to avoid generating a Fix-It that moves `throws (Int, Int)` in front of the arrow.

And if a user has a space between `throws` and the thrown error type and also writes the clause after the arrow, I am fine with producing a sub-optimal diagnostic because those two choices/mistakes combined are quite unlikely. 

---

And also congrats for noticing the `throws /* comment */ (any Error)` case. Shows a great attention to detail 🤩

Sources/SwiftParser/Specifiers.swift
AppAppWorks: In my current design, a thrown type missing the right parenthesis is allowed to recover, do you think such recovery should be disallowed?

ahoppen: It should be able to recover with `self.expect` as well. The only difference is whether `rightParen` is set to `nil` or a missing right parenthesis.

Sources/SwiftParser/Specifiers.swift
```diff
@@ -795,22 +796,56 @@ extension Parser {
...
-            synthesizedThrows = missingToken(spec)
-          } else {
-            synthesizedThrows = missingToken(.throws)
+        guard effectSpecifiers?.throwsClause == nil else {
+          continue
+        }
+        if matchedSubset == S.CorrectThrowsTokenKinds.self {
+          let synthesizedThrows = missingToken(spec)
+          var lookahead = self.lookahead()
+          guard lookahead.consume(if: .leftParen) != nil,
```
ahoppen, line 805: Could you add a check for whether the is a space between `throws` and the left parenthesis, similar to here

https://github.com/swiftlang/swift-syntax/blob/93df9ed82e07760eb8eceb09a4504d7057110da3/Sources/SwiftParser/Attributes.swift#L184

A test case for this would be that `func foo() -> throws (Int, Int)` has a Fix-It to move the `throws` but not `(Int, Int)`.

Sources/SwiftParser/Specifiers.swift
```diff
@@ -795,22 +796,56 @@ extension Parser {
...
+              throwsSpecifier: synthesizedThrows,
+              leftParen: nil,
+              type: nil,
+              rightParen: nil,
+              arena: self.arena
+            )
+            continue
           }
+
+          let leftParen = self.expect(.leftParen).token
```
ahoppen, line 818: We shouldn’t be discarding the unexpected tokens from `expect`. I know that it can’t really happen here because we check in `lookahead.consume(if:)` that the next token is a left parenthesis but in general, dropping syntax nodes can lead to a syntax tree that doesn’t represent the original source code and it’s good practice to never drop anything. I’d change this to

```swift
let (unexpectedBeforeLeftParen, leftParen) = self.expect(.leftParen)
```
change `unexpected` to be of type `[RawSyntax]` (which requires a few conversions from `RawTokenSyntax` to `RawSyntax` using the `RawSyntax` initializer) and then below do

```swift
if let unexpectedBeforeLeftParen {
  unexpected.append(RawSyntax(unexpectedBeforeLeftParen))
}
unexpected.append(RawSyntax(leftParen))
unexpected.append(RawSyntax(type))
```

Notice how you also don’t need `allTokens` on `type` anymore.

And then you could do the same for `rightParen` to also use `expect`.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
ahoppen: Ah, I see. That’s why you have to use `allTokens` during parsing. 

It think we need to have a generalized version of `exchangeTokens` that works on general `Syntax` nodes instead of `TokenSyntax`.

AppAppWorks: Should I try to generalise `exchangeTokens` in this PR, or just raise another issue?

ahoppen: If you could generalize it in this PR, that would be great

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -171,6 +171,46 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
+  /// into `misplacedTokens`.  All tokens in `correctNodes` will be flattened into `correctTokens`.  After that the
+  /// remaining operations will be the same as `exchangeTokens`.
+  public func exchangeNodes(
+    unexpected: UnexpectedNodesSyntax?,
+    unexpectedNodeCondition: (Syntax) -> Bool,
+    correctNodes: [Syntax?],
+    message: (_ misplacedTokens: [TokenSyntax]) -> some DiagnosticMessage,
+    moveFixIt: (_ misplacedTokens: [TokenSyntax]) -> FixItMessage,
+    removeRedundantFixIt: (_ misplacedTokens: [TokenSyntax]) -> FixItMessage? = { _ in nil }
+  ) {
```
ahoppen, line 187: Instead of implementing `exchangeNodes` in terms of `exchangeTokens`, would it be possible to just generalize `exchangeTokens` to work `Syntax` instead of `TokenSyntax`? I briefly looked through it and I didn’t find any huge blockers but I might have missed something.

Sources/SwiftParser/Specifiers.swift
```diff
@@ -772,16 +772,21 @@ extension Parser {
...
-    var synthesizedThrows: RawTokenSyntax? = nil
-    var unexpected: [RawTokenSyntax] = []
+    var synthesizedThrowsClause: RawThrowsClauseSyntax? = nil
+    var unexpected: [RawSyntax] = []
     var loopProgress = LoopProgressCondition()
+
+    func unexpect(_ node: some RawSyntaxNodeProtocol) {
+      unexpected.append(RawSyntax(node))
+    }
+
```
ahoppen, line 782: I personally think that the code is more readable if you just inline this code. Then it’s easier to follow where `unexpected` is being built up, IMO.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -171,26 +171,38 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
     moveFixIt: (_ misplacedTokens: [TokenSyntax]) -> FixItMessage,
     removeRedundantFixIt: (_ misplacedTokens: [TokenSyntax]) -> FixItMessage? = { _ in nil }
   ) {
     guard let incorrectContainer = unexpected else { return }
-    let misplacedTokens = incorrectContainer.presentTokens(satisfying: unexpectedTokenCondition)
-    if misplacedTokens.isEmpty { return }
+    let misplacedNodes = incorrectContainer.compactMap { $0.as(S.self) }.filter(unexpectedNodeCondition)
+    let misplacedTokens = misplacedNodes.flatMap {
+      Array($0.tokens(viewMode: .sourceAccurate))
+    }
```
ahoppen, line 193: Couldn’t this just be `[Syntax]` as well? `FixIt.MultiNodeChange.makeMissing` supports nodes as well as tokens and I assume that `message`, `moveFixIt`, and `removeRedundantFixIt` could be changed to take `[Syntax]`. And `misplacedTokens.only` could become `misplacedTokens.only.as(TokenSyntax.self)`.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -171,26 +171,38 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
...
+    guard !misplacedTokens.isEmpty else { return }
 
-    let correctTokens = correctTokens.compactMap({ $0 })
+    let correctTokens = correctNodes.flatMap { node -> [TokenSyntax] in
+      if let node {
+        Array(node.tokens(viewMode: .all))
+      } else {
+        []
+      }
+    }
```
ahoppen, line 202: I think this could also be `[Syntax]`:
- `correctTokens.only` would just need to become `correctTokens.only.as(TokenSyntax.self)`
- `correctTokens.filter(\.isMissing)` could become `correctTokens.filter(\.isMissingAllTokens)`

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
ahoppen: Ah, that’s unfortunate. I was hoping that the changes to the `message` etc. closures would be minimal. I still think it would be good to migrate them to `Syntax` but that doesn’t need to be part of this PR. Could you file an issue for that so we can do the migration later?

AppAppWorks:
@ahoppen, could you review my latest commit? :)

ahoppen:
Did you try to address https://github.com/swiftlang/swift-syntax/pull/2731#discussion_r1688770662?

AppAppWorks:
> Did you try to address [#2731 (comment)](https://github.com/swiftlang/swift-syntax/pull/2731#discussion_r1688770662)?

I might have tried that but now I've forgotten it ;)  
I'll have a quick check and come back to you!

AppAppWorks:
[#2731 (comment)](https://github.com/swiftlang/swift-syntax/pull/2731#discussion_r1688770662) addressed.

ahoppen (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

AppAppWorks:
Fixed, please test again.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test macOS

ahoppen:
@swift-ci Please test Windows

PR #2736, opened 2024-07-17
[CMake] Enable package CMO if possible

rintaro:
Package cross-module-optimization allows non-resilient access between modules as long as they share the same package name.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,23 +42,24 @@ option(SWIFT_SYNTAX_ENABLE_WMO_PRE_3_26
        $<IF:$<AND:$<NOT:$<CONFIG:Debug>>,$<PLATFORM_ID:Darwin>>,YES,NO>)
 
 include(AddSwiftHostLibrary)
+include(SwiftCompilerCapability)
 
-# Ensure that we do not link the _StringProcessing module. But we can
-# only pass this flag for new-enough compilers that support it.
-file(WRITE "${CMAKE_BINARY_DIR}/tmp/empty-check-string-processing.swift" "")
-execute_process(
-  COMMAND
-    "${CMAKE_Swift_COMPILER}"
-    -Xfrontend -disable-implicit-string-processing-module-import
-    -Xfrontend -parse-stdlib
-    -typecheck "${CMAKE_BINARY_DIR}/tmp/empty-check-string-processing.swift"
-  OUTPUT_QUIET ERROR_QUIET
-  RESULT_VARIABLE
-    SWIFT_SUPPORTS_DISABLE_IMPLICIT_STRING_PROCESSING_MODULE_IMPORT)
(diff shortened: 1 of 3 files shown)


cmake/modules/AddSwiftHostLibrary.cmake
```diff
@@ -77,8 +77,28 @@ function(add_swift_syntax_library name)
         -emit-module-path;${module_file};
         -emit-module-source-info-path;${module_sourceinfo_file};
         -emit-module-interface-path;${module_interface_file};
-        -emit-private-module-interface-path;${module_private_interface_file}
+        -emit-private-module-interface-path;${module_private_interface_file};
+        -no-verify-emitted-module-interface
```
rintaro, line 81: Without this, the driver invokes `-frontend -typecheck-module-from-interface` job and that emits confusing warnings:
```
<unknown>:0: warning: ignoring -experimental-allow-non-resilient-access (overriden by -compile-module-from-interface or -typecheck-module-from-interface)
<unknown>:0: warning: ignoring -experimental-package-cmo (requires -experimental-allow-non-resilient-access)
```

tshortli: If they are just warnings, is it necessary to opt out of interface verification? We should just fix the diagnostics and I don't want you to get stuck with a configuration where the module interface is not verified (new problems with generated interface could escape as a result).

rintaro: OK, it's not really necessary to disable.

rintaro:
https://github.com/swiftlang/swift/pull/75304
@swift-ci Please test

CMakeLists.txt
```diff
@@ -42,23 +42,24 @@ option(SWIFT_SYNTAX_ENABLE_WMO_PRE_3_26
        $<IF:$<AND:$<NOT:$<CONFIG:Debug>>,$<PLATFORM_ID:Darwin>>,YES,NO>)
 
 include(AddSwiftHostLibrary)
+include(SwiftCompilerCapability)
```
edymtt, line 45: I would suggest to unset `SWIFT_LOADED_MODULE_TRACE_FILE ` like we did in https://github.com/swiftlang/swift/pull/72405 (ideally before we enable the language in line 13)

```suggestion
include(SwiftCompilerCapability)

# we are not interested in logging any Swift module used
# when configuring the build system -- those are not useful
# since they will not contribute to the build of the compiler itself
unset(ENV{SWIFT_LOADED_MODULE_TRACE_FILE})
```

edymtt: Actually, there should not be needed here -- this is important when building `swift-syntax` as part of `swift`, and we are already unsetting that variable there.

bnbarham (approved):

PR #2739, opened 2024-07-18
Fix issue that caused `AccessorDeclSyntax` initializer with header + body to be misparsed

ahoppen:
`get {}` was parsed by `Parser.parseDeclaration()` but it doesn't parse accessors.

`AccessorDeclSyntax` needs a special implementation of `init(header:bodyBuilder:)`.

rdar://131720084

### DIFF
--- a/Sources/SwiftSyntaxBuilder/SyntaxNodeWithBody.swift
+++ b/Sources/SwiftSyntaxBuilder/SyntaxNodeWithBody.swift
@@ -12,8 +12,10 @@
 
 #if swift(>=6)
 public import SwiftSyntax
+internal import SwiftParser
 #else
 import SwiftSyntax
+import SwiftParser
 #endif
 
 // MARK: - PartialSyntaxNode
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

Sources/SwiftSyntaxBuilder/SyntaxNodeWithBody.swift
```diff
@@ -124,7 +126,17 @@ extension WithOptionalCodeBlockSyntax where Self: DeclSyntaxProtocol {
...
   ) throws {
-    let decl = DeclSyntax("\(header) {}")
+    // If the type provides a custom `SyntaxParseable` implementation, use that. Otherwise construct it as a
+    // `DeclSyntax`.
+    let decl: DeclSyntax
+    var stringInterpolation = SyntaxStringInterpolation(literalCapacity: 1, interpolationCount: 1)
+    stringInterpolation.appendInterpolation(header)
+    stringInterpolation.appendLiteral(" {}")
+    if let parsableType = Self.self as? SyntaxParseable.Type {
+      decl = parsableType.init(stringInterpolation: stringInterpolation).cast(DeclSyntax.self)
```
rintaro, line 136: I think upcasting here and down casting again below is a waste. Instead of branching in `where Self: DeclSyntaxProtocol` extension, how about making `where Self: SyntaxParseable` separately?
```swift
extension WithOptionalCodeBlockSyntax where Self: SyntaxParseable & DeclSyntaxProtocol { ... 
```
(`& DeclSyntaxProtocol` is for resolving ambiguity)

ahoppen:
@swift-ci Please test

Sources/SwiftSyntaxBuilder/SyntaxNodeWithBody.swift
ahoppen: Discussed offline and kept the current implementation + added a comment why we need to manually construct the string interpolation.

ahoppen:
@swift-ci Please test Windows

rintaro (approved):

PR #2744, opened 2024-07-19
add fix-its for a bare regex literal with leading and/or trailing spaces

### LINKED ISSUE #1479: Add fix-it for a regex that starts or ends with a space
### Description

We ought to provide a fix-it for a regex such as `/ x/` or `/ x/` to use an extended literal instead e.g `#/ x/#`. For the space at the start we could also suggest to insert a backslash. This depends on #1463 being merged first.

### Steps to Reproduce

_No response_

AppAppWorks:
fix #1479

### DIFF
--- a/Sources/SwiftParserDiagnostics/LexerDiagnosticMessages.swift
+++ b/Sources/SwiftParserDiagnostics/LexerDiagnosticMessages.swift
@@ -325,8 +325,93 @@ extension SwiftSyntax.TokenDiagnostic {
         changes.append(.replaceLeadingTrivia(token: nextToken, newTrivia: []))
       }
       return [FixIt(message: .removeExtraneousWhitespace, changes: changes)]
+    case .spaceAtEndOfRegexLiteral:
+      if let regexLiteral = token.regexParent {
+        // wouldn't suggest `insertBackslash` because the potential presence of (?x) somewhere preceding the trailing
+        // space could mean the trailing space has no semantic meaning. Escaping the space could change the semantics.
+        return [regexLiteral.convertToExtendedRegexLiteralFixIt]
+      } else {
+        return []
+      }
+    case .spaceAtStartOfRegexLiteral:
+      guard let regexLiteral = token.regexParent else {
+        return []
+      }
+
+      let regexText = regexLiteral.regex.text
+      let lastIndex = regexText.index(before: regexText.endIndex)
+      if regexText.startIndex != lastIndex && regexText[lastIndex].isWhitespace {
+        // if the regex has a distinct trailing space, same as the handling at `case .spaceAtEndOfRegexLiteral`
+        return [regexLiteral.convertToExtendedRegexLiteralFixIt]
+      } else {
+        let escapedRegexText = #"\\#(regexText)"#
+        return [
+          regexLiteral.convertToExtendedRegexLiteralFixIt,
+          FixIt(
+            message: .insertBackslash,
+            changes: [
+              .replace(
+                oldNode: Syntax(regexLiteral),
+                newNode: Syntax(
+                  regexLiteral
+                    .with(\.regex, .regexLiteralPattern(escapedRegexText))
+                )
+              )
+            ]
+          ),
+        ]
+      }
     default:
       return []
     }
+  }
+}
+
+private extension TokenSyntax {
+  var regexParent: RegexLiteralExprSyntax? {
+    var parent = Syntax(self)
+    while parent.kind != .regexLiteralExpr, let upper = parent.parent {
+      parent = upper
+    }
+    return parent.as(RegexLiteralExprSyntax.self)
+  }
+}
+
+private extension RegexLiteralExprSyntax {
+  /// Creates a Fix-it that suggests converting to extended regex literal
+  ///
+  /// Covers the following cases:
+  /// ```swift
+  /// let leadingSpaceRegex = / ,/
+  /// // converts to
+  /// let leadingSpaceExtendedRegex = #/ ,/#
+  ///
+  /// let leadingAndTrailingSpaceRegex = / , /
+  /// // converts to
+  /// let leadingAndTrailingSpaceExtendedRegex = #/ , /#
+  ///
+  /// let trailingSpaceRegex = /, /
+  /// // converts to
+  /// let trailingSpaceExtendedRegex = #/ ,/#
+  ///
+  /// let trailingSpaceMissingClosingSlashRegex = /,
+  /// // converts to
+  /// let trailingSpaceExtendedRegex = #/, /#
+  /// ```
+  var convertToExtendedRegexLiteralFixIt: FixIt {
+    FixIt(
+      message: .convertToExtendedRegexLiteral,
+      changes: [
+        .replace(
+          oldNode: Syntax(self),
+          newNode: Syntax(
+            with(\.openingSlash, .regexSlashToken())
+              .with(\.openingPounds, .regexPoundDelimiter("#", leadingTrivia: leadingTrivia))
+              .with(\.closingPounds, .regexPoundDelimiter("#", trailingTrivia: trailingTrivia))
+              .with(\.closingSlash, .regexSlashToken())
+          )
+        )
+      ]
+    )
   }
 }
--- a/Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift
+++ b/Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift
@@ -655,6 +655,9 @@ extension FixItMessage where Self == StaticParserFixIt {
   public static var insertAttributeArguments: Self {
     .init("insert attribute argument")
   }
+  public static var insertBackslash: Self {
+    .init(#"insert '\'"#)
+  }
   public static var insertNewline: Self {
     .init("insert newline")
   }
@@ -690,6 +693,9 @@ extension FixItMessage where Self == StaticParserFixIt {
   }
   public static var wrapInBackticks: Self {
     .init("if this name is unavoidable, use backticks to escape it")
+  }
+  public static var convertToExtendedRegexLiteral: Self {
+    .init("convert to extended regex literal with '#'")
   }
 }
 
--- a/Tests/SwiftParserTest/Assertions.swift
+++ b/Tests/SwiftParserTest/Assertions.swift
@@ -289,9 +289,9 @@ struct DiagnosticSpec {
 }
 
 /// Assert that `location` is the same as that of `locationMarker` in `tree`.
-func assertLocation<T: SyntaxProtocol>(
+func assertLocation(
   _ location: SourceLocation,
-  in tree: T,
+  in tree: some SyntaxProtocol,
   markerLocations: [String: Int],
   expectedLocationMarker locationMarker: String,
   file: StaticString = #filePath,
@@ -315,9 +315,9 @@ func assertLocation<T: SyntaxProtocol>(
 
 /// Assert that the diagnostic `note` produced in `tree` matches `spec`,
 /// using `markerLocations` to translate markers to actual source locations.
-func assertNote<T: SyntaxProtocol>(
+func assertNote(
   _ note: Note,
-  in tree: T,
+  in tree: some SyntaxProtocol,
   markerLocations: [String: Int],
   expected spec: NoteSpec
 ) {
@@ -335,9 +335,9 @@ func assertNote<T: SyntaxProtocol>(
 
 /// Assert that the diagnostic `diag` produced in `tree` matches `spec`,
 /// using `markerLocations` to translate markers to actual source locations.
-func assertDiagnostic<T: SyntaxProtocol>(
+func assertDiagnostic(
   _ diag: Diagnostic,
-  in tree: T,
+  in tree: some SyntaxProtocol,
   markerLocations: [String: Int],
   expected spec: DiagnosticSpec
 ) {
@@ -491,9 +491,9 @@ public struct AssertParseOptions: OptionSet, Sendable {
 extension ParserTestCase {
   /// After a test case has been mutated, assert that the mutated source
   /// round-trips and doesn’t hit any assertion failures in the parser.
-  fileprivate static func assertMutationRoundTrip<S: SyntaxProtocol>(
+  fileprivate static func assertMutationRoundTrip(
     source: [UInt8],
-    _ parse: (inout Parser) -> S,
+    _ parse: (inout Parser) -> some SyntaxProtocol,
     swiftVersion: Parser.SwiftVersion?,
     experimentalFeatures: Parser.ExperimentalFeatures,
     file: StaticString,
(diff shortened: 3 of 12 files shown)


Sources/SwiftParserDiagnostics/LexerDiagnosticMessages.swift
```diff
@@ -325,6 +325,50 @@ extension SwiftSyntax.TokenDiagnostic {
         changes.append(.replaceLeadingTrivia(token: nextToken, newTrivia: []))
       }
       return [FixIt(message: .removeExtraneousWhitespace, changes: changes)]
+    case .spaceAtStartOfRegexLiteral:
+      var ancestor = Syntax(token)
+      while ancestor.kind != .regexLiteralExpr, let parent = ancestor.parent {
+        ancestor = parent
+      }
+
+      let regexLiteral = ancestor.cast(RegexLiteralExprSyntax.self)
```
ahoppen, line 334: We should return with `[]` if we can’t find a parent that’s a `regexLiteralExpr` instead of crashing here, just to be safe.

Sources/SwiftParserDiagnostics/LexerDiagnosticMessages.swift
```diff
@@ -325,6 +325,50 @@ extension SwiftSyntax.TokenDiagnostic {
...
+      let regexText = regexLiteral.regex.text
+      let lastIndex = regexText.index(before: regexText.endIndex)
+      let escapedRegexText: String =
+        if regexText.startIndex != lastIndex && regexText[lastIndex].isWhitespace {
+          #"\\#(regexText[..<lastIndex])\ "#
+        } else {
+          "\\" + regexText
+        }
```
ahoppen, line 342: swift-syntax needs to be able to build with a Swift 5.7 compiler so we can’t use if expressions yet.

Sources/SwiftParserDiagnostics/LexerDiagnosticMessages.swift
```diff
@@ -325,6 +325,50 @@ extension SwiftSyntax.TokenDiagnostic {
...
+      var ancestor = Syntax(token)
+      while ancestor.kind != .regexLiteralExpr, let parent = ancestor.parent {
+        ancestor = parent
+      }
+
+      let regexLiteral = ancestor.cast(RegexLiteralExprSyntax.self)
+      let regexText = regexLiteral.regex.text
+      let lastIndex = regexText.index(before: regexText.endIndex)
+      let escapedRegexText: String =
+        if regexText.startIndex != lastIndex && regexText[lastIndex].isWhitespace {
```
ahoppen, line 338: Can we ever hit this? Isn’t the error about the regex *staring* with a space but you’re checking if it ends with a space. Also would be good to check that this Fix-It behaves correctly when having a regex that starts and ends with a space.

---

`isWhitespace` would also be satisfied for characters other than a space in which case we’d be translating the arbitrary whitespace to a space below. Instead of hardcoding the space there, I would just include `regexText[lastIndex]`.

Tests/SwiftParserTest/translated/ForwardSlashRegexTests.swift
```diff
@@ -1689,14 +1729,22 @@ final class ForwardSlashRegexTests: ParserTestCase {
...
+          ]
+        ),
         DiagnosticSpec(
           locationMarker: "2️⃣",
           message: "expected '/' to end regex literal",
           notes: [NoteSpec(message: "to match this opening '/'")],
           fixIts: ["insert '/'"]
         ),
       ],
+      applyFixIts: ["insert '/'"],
```
ahoppen, line 1747: Typo? The Fix-It is `insert '\'` but here you’re using a forward slash instead of a backslash. Same in the other tests as well.

Sources/SwiftParserDiagnostics/ParserDiagnosticMessages.swift
```diff
@@ -691,6 +694,9 @@ extension FixItMessage where Self == StaticParserFixIt {
   public static var wrapInBackticks: Self {
     .init("if this name is unavoidable, use backticks to escape it")
   }
+  public static var convertToExtendedRegexLiteral: Self {
+    .init("convert to extended regex literal")
```
ahoppen, line 698: I’m not sure if *extended regex literal* is a term that most developers know. What do you think of changing this to

```suggestion
    .init("convert to extended regex literal with '#'")
```

Tests/SwiftParserTest/RegexLiteralTests.swift
```diff
@@ -495,8 +553,37 @@ final class RegexLiteralTests: ParserTestCase {
...
+          message: "bare slash regex literal may not start with space",
+          fixIts: [
+            "convert to extended regex literal",
+            #"insert '\'"#,
+          ]
+        )
+      ],
+      applyFixIts: [#"insert '\'"#],
+      fixedSource: #"""
+        /\ \ /
```
ahoppen, line 585: Could you also add a test case where the regex contains more than 2 spaces. Just to make sure that we only escape the first and last and not the ones in between. (or just make `testForwardSlashRegex164` apply the `insert '\'` Fix-It.

ahoppen (commented):
You should be able to remove the `TODO` in RegexLiteralLexer.swift:259 as well.

Tests/SwiftParserTest/translated/ForwardSlashRegexTests.swift
AppAppWorks: I decided to silence the new fix-its for all translated test cases because I thought these test cases weren't meant for testing these new fix-its?

Sources/SwiftParserDiagnostics/LexerDiagnosticMessages.swift
AppAppWorks: There are two possible cases where the regex lexer might emit the error `.spaceAtEndOfRegexLiteral`, however in each case some other error takes precedence over it,

1. `.spaceAtStartOfRegexLiteral` as in `/  /` (both leading and trailing spaces)
```swift
assertParse(
  """
  /1️⃣  /
  """,
  diagnostics: [
    DiagnosticSpec(
      message: "bare slash regex literal may not start with space",
      fixIts: [
        "convert to extended regex literal with '#'",
        #"insert '\'"#,
      ]
    )
  ],
  applyFixIts: ["convert to extended regex literal with '#'"],
  fixedSource: """
    #/  /#
    """
)
```
The current design will just shadow `.spaceAtEndOfRegexLiteral` at the presence of `.spaceAtStartOfRegexLiteral`.
```swift

builder.recordPatternError(.spaceAtStartOfRegexLiteral, at: cursor)
//
// some further regex lexing
//
builder.recordPatternError(.spaceAtEndOfRegexLiteral, at: lastUnespacedSpaceOrTab)
```
because of this
```swift
extension RegexLiteralLexemes.Builder {
 mutating func recordPatternError(
    _ kind: TokenDiagnostic.Kind,
    at cursor: Lexer.Cursor
  ) {
    precondition(lastLexemeKind == .openingSlash)

    // Only record if we don't already have a pattern error, we want to prefer
    // the first error we encounter.
    if patternError == nil {
      patternError = .init(kind, position: cursor)
    }
  }
}
```
2. `missingNode` as in `/a /` (trailing-space only)
```swift
assertParse(
  """
  /a /1️⃣
  """,
  diagnostics: [
    DiagnosticSpec(message: "expected expression after operator", fixIts: ["insert expression"])
  ],
  fixedSource: """
    /a / <#expression#>
    """
)
```

I decided to make the "insert '\'" fix-it also escape the trailing space in case 1 because escaping only the leading space will turn the fixed regex into case 2, e.g. `/  /` -> `/\  /`, that would be undesirable.

In a previous iteration, I've tried not emitting the "insert `\'" fix-it when there're both leading and trailing spaces in the regex, but for some reason I decided to change the design to the current one.

AppAppWorks:
> You should be able to remove the `TODO` in RegexLiteralLexer.swift:259 as well.

I've found two other `TODO`s in RegexLiteralLexer.swift:210-212 about `.spaceAtEndOfRegexLiteral`

Tests/SwiftParserTest/translated/ForwardSlashRegexTests.swift
AppAppWorks: As a side note, I've duplicated the `assertParse` statement for each test case in `RegexLiteralTests` affected by this PR so as to apply each of the two fix-its at a time, e.g.
```swift
assertParse(
  """
  let x = /1️⃣ a/
  """,
  diagnostics: [
    DiagnosticSpec(
      message: "bare slash regex literal may not start with space",
      fixIts: [
        "convert to extended regex literal with '#'",
        #"insert '\'"#,
      ]
    )
  ],
  applyFixIts: ["convert to extended regex literal with '#'"],
  fixedSource: """
    let x = #/ a/#
    """
)

assertParse(
  """
  let x = /1️⃣ a/
  """,
  diagnostics: [
    DiagnosticSpec(
      message: "bare slash regex literal may not start with space",
      fixIts: [
        "convert to extended regex literal with '#'",
        #"insert '\'"#,
      ]
    )
  ],
  applyFixIts: [#"insert '\'"#],
  fixedSource: #"""
    let x = /\ a/
    """#
)
```

It's become a bit unwieldy.

ahoppen:
> I've found two other `TODO`s in RegexLiteralLexer.swift:210-212 about `.spaceAtEndOfRegexLiteral`

Good catch. You can remove them as well.

Sources/SwiftParserDiagnostics/LexerDiagnosticMessages.swift
ahoppen: Ah, I see. Thanks for the detailed explanation. That makes perfect sense 👍🏽

Tests/SwiftParserTest/translated/ForwardSlashRegexTests.swift
ahoppen: Ah, sorry. I think I realized the same later on as well but then forgot to remove my comment again. Your change makes perfect sense

ahoppen:
Looks good to me, I’d just like to see the following two addressed:
- [ ] https://github.com/swiftlang/swift-syntax/pull/2744#discussion_r1684806632 
- [ ] https://github.com/swiftlang/swift-syntax/pull/2744#issuecomment-2243448358

Tests/SwiftParserTest/Assertions.swift
```diff
@@ -578,6 +611,76 @@ extension ParserTestCase {
...
+    substructure expectedSubstructure: (some SyntaxProtocol)? = Optional<Syntax>.none,
+    substructureAfterMarker: String = "START",
+    diagnostics expectedDiagnostics: [DiagnosticSpec] = [],
+    fixItsApplications: [FixItsApplication] = [],
+    options: AssertParseOptions = [],
+    swiftVersion: Parser.SwiftVersion? = nil,
+    experimentalFeatures: Parser.ExperimentalFeatures? = nil,
+    file: StaticString = #filePath,
+    line: UInt = #line
   ) {
```
AppAppWorks, line 684: I've introduced a more general `assertParse` to avoid the creation of redundant test cases just to test out different combinations of fix-its against fixed sources.

AppAppWorks:
> > I've found two other `TODO`s in RegexLiteralLexer.swift:210-212 about `.spaceAtEndOfRegexLiteral`
> 
> Good catch. You can remove them as well.

```swift
// TODO: We ought to have a fix-it that suggests #/.../#. We could
// suggest escaping, but that would be wrong if the user has written (?x).
// TODO: Should we suggest #/.../# for space-as-first character too?
```

I don't quite understand what "that would be wrong if the user has written (?x)." means, I am hesitating to remove these two `TODO`s.

Tests/SwiftParserTest/Assertions.swift
```diff
@@ -640,27 +743,30 @@ extension ParserTestCase {
...
+            .replacingOccurrences(of: "\r\n", with: "\n")
+            .replacingOccurrences(of: "\r", with: "\n")
+        }
+        assertStringsEqualWithDiff(
+          fixedTreeDescription.trimmingTrailingWhitespace(),
+          fixItsApplication.expectedFixedSource.trimmingTrailingWhitespace(),
+          "Applying \(applyFixIts?.description ?? "all Fix-Its") didn’t produce the expected fixed source",
+          file: file,
+          line: line
+        )
```
AppAppWorks, line 765: Because it's now possible to test out different combinations of fix-its in `assertParse`, I think it's more illustrative to print out what exact fix-its had been applied that led to the failure.

Tests/SwiftParserTest/Assertions.swift
```diff
@@ -578,6 +611,76 @@ extension ParserTestCase {
...
+  ///   - expectedDiagnostics: Asserts the given diagnostics were output, by default it
+  ///     asserts the parse was successful (ie. it has no diagnostics). Note
+  ///     that `DiagnosticsSpec` uses the location marked by `1️⃣` by default.
+  ///   - fixItsApplications: A list of `FixItsApplication` to check for whether applying certain fix-its to the source
+  ///     will generate a certain expected fixed source. An empty list means no fix-its are expected.
+  ///   - swiftVersion: The version of Swift using which the file should be parsed.
+  ///      Defaults to the latest version.
+  ///   - experimentalFeatures: A list of experimental features to enable, or
+  ///     `nil` to enable the default set of features provided by the test case.
+  func assertParse(
```
ahoppen, line 672: You should be able to migrate a few tests to this. Search for `applyFixIts: [testCase.fixIt],`

Tests/SwiftParserTest/Assertions.swift
```diff
@@ -578,6 +611,76 @@ extension ParserTestCase {
...
+  ///     than the whole source.
+  ///   - expectedDiagnostics: Asserts the given diagnostics were output, by default it
+  ///     asserts the parse was successful (ie. it has no diagnostics). Note
+  ///     that `DiagnosticsSpec` uses the location marked by `1️⃣` by default.
+  ///   - fixItsApplications: A list of `FixItsApplication` to check for whether applying certain fix-its to the source
+  ///     will generate a certain expected fixed source. An empty list means no fix-its are expected.
+  ///   - swiftVersion: The version of Swift using which the file should be parsed.
+  ///      Defaults to the latest version.
+  ///   - experimentalFeatures: A list of experimental features to enable, or
+  ///     `nil` to enable the default set of features provided by the test case.
```
ahoppen, line 671: Instead of duplicating the comment (which will inevitably go out-of-sync). I would prefer to have one canonical version of the comment. The other `assertParse` function could then have a comment like

```swift
/// Shorthand version of ``assertParse-<someHash from docc>`` that allows checking a single configuration of applied Fix-Its.
```

Tests/SwiftParserTest/Assertions.swift
```diff
@@ -524,6 +524,39 @@ extension ParserTestCase {
...
+        nil
+      }
+    }
+
+    var expectedFixedSource: String {
+      switch self {
+      case .optIn(_, let fixedSource), .all(let fixedSource):
+        fixedSource
+      }
+    }
```
ahoppen, line 557: swift-syntax needs to be able to build with Swift 5.8, so you can’t use switch expressions here.

```suggestion
    var applyFixIts: [String]? {
      switch self {
      case .optIn(let applyFixIts, _):
        return applyFixIts
      case .all:
        return nil
      }
    }

    var expectedFixedSource: String {
      switch self {
      case .optIn(_, let fixedSource), .all(let fixedSource):
        return fixedSource
      }
    }
```

Tests/SwiftParserTest/Assertions.swift
AppAppWorks: I couldn't get docc build for test targets, would there be any other way to get the hash?

ahoppen: The only I know how to get them is to use code completion in Xcode.

hamishknight:
> I don't quite understand what "that would be wrong if the user has written (?x)." means, I am hesitating to remove these two TODOs

[We support an "extended syntax mode" in regex literals](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0355-regex-syntax-run-time-construction.md#extended-syntax-modes), which can be enabled with `(?x)` and makes whitespace non-semantic, e.g `/(?x) a b c/` is equivalent to `/abc/`. This is relevant for the trailing space case because if you have e.g<br> `/(?x) /`, providing a fix-it of `/(?x)\ /` would change the meaning of the regex (previously it was empty, now it matches a literal space). This likely isn't a common case, but I think the easiest way of handling it is to only suggest adding `#` to the regex delimiters, which won't change the semantic meaning. Checking for the use of `(?x)` isn't entirely trivial since it can appear anywhere in the pattern, can appear with other matching options, and can be disabled later on in the pattern.

AppAppWorks:
> > I don't quite understand what "that would be wrong if the user has written (?x)." means, I am hesitating to remove these two TODOs
> 
> [We support an "extended syntax mode" in regex literals](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0355-regex-syntax-run-time-construction.md#extended-syntax-modes), which can be enabled with `(?x)` and makes whitespace non-semantic, e.g `/(?x) a b c/` is equivalent to `/abc/`. This is relevant for the trailing space case because if you have e.g `/(?x) /`, providing a fix-it of `/(?x)\ /` would change the meaning of the regex (previously it was empty, now it matches a literal space). This likely isn't a common case, but I think the easiest way of handling it is to only suggest adding `#` to the regex delimiters for the trailing space case, which won't change the semantic meaning. Checking for the use of `(?x)` isn't entirely trivial since it can appear anywhere in the pattern, can appear with other matching options, and can be disabled later on in the pattern.

Thanks for your detailed explanation!

Considering that the parser of now will never emit diagnostics for the trailing space case alone, I guess we can safely discard the two `TODO`s ?

hamishknight:
> Considering that the parser of now will never emit diagnostics for the trailing space case alone, I guess we can safely discard the two `TODO`s ?

IIRC it can be emitted for things like this:

```
let x = /, /
```

AppAppWorks:
> > Considering that the parser of now will never emit diagnostics for the trailing space case alone, I guess we can safely discard the two `TODO`s ?
> 
> IIRC it can be emitted for things like this:
> 
> ```swift
> let x = /, /
> ```
> 
> (Oh huh I never added a test for that, my bad)

I got it, I will revert back to my earlier design that the presence of a trailing space will mandate only "convert to extended regex literal with '#'" as the only Fix-it.

ahoppen (commented):
Could you also address this comment?

- [ ] https://github.com/swiftlang/swift-syntax/pull/2744/files#r1688783702

Tests/SwiftParserTest/Assertions.swift
AppAppWorks: > ```swift
> /// Shorthand version of ``assertParse-<someHash from docc>`` that allows checking a single configuration of applied Fix-Its.
> ```

Could I just omit the the hash for now and go for the full name of `assertParse` instead?

ahoppen: Ah, I forgot that `assertParse` is an internal function and not `public`, hence a docc link won’t work. 

So just use 
```
`assertParse`
```
with single backticks and make it a normal code font.

AppAppWorks:
Done, please check.

ahoppen (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/SwiftParserDiagnostics/LexerDiagnosticMessages.swift
```diff
@@ -325,8 +325,92 @@ extension SwiftSyntax.TokenDiagnostic {
...
+  /// let trailingSpaceExtendedRegex = #/, /#
+  /// ```
+  var convertToExtendedRegexLiteralFixIt: FixIt {
+    FixIt(
+      message: .convertToExtendedRegexLiteral,
+      changes: [
+        .replace(
+          oldNode: Syntax(self),
+          newNode: Syntax(
+            with(\.openingPounds, .poundToken())
```
ahoppen, line 408: Tests are failing because this should be `regexPoundDelimiter` instead of `poundToken`.

```suggestion
            with(\.openingPounds, .regexPoundDelimiter("#"))
```

Same for the closing pounds.

AppAppWorks: Fixed, please have a look.

ahoppen:
@swift-ci Please test

AppAppWorks:
The check statuses have stayed unchanged for 5 days, could there be any problem with the CI?

bnbarham:
@swift-ci please test

PR #2748, opened 2024-07-23
[SwiftLexicalLookup][GSoC] Add proper guard scope and implicit name lookup

MAJKFL:
This PR introduces proper `guard` handling and implicit name lookup.
- Scopes of new internal `IntroducingToSequentialParentScopeSyntax` kind introduce their lookup results to their `SequentialScopeSyntax` parents. Currently, the only `IntroducingToSequentialParentScopeSyntax` is `guard` scope and the only `SequentialScopeSyntax`s are code block and file scope (the latter one uses `SequentialScopeSyntax` properties depending on the configuration).
- Lookup from within `guard … else` clause specifies `skipSequentialIntroductionFrom` property in the new `LookupState` object passed between `lookup` functions. I think it would be great to make this more opaque for the client, so I'd love to hear suggestions/other ideas how we could represent and pass the state between lookup methods.
- Implicit name handling works through a new `LookupName` `implicit` case. Associated with it, a new enum `LookupImplicitNameKind` describes all possible implicit names that could be referenced during lexical name lookup.
- `TypeScopeSyntax` defines behavior of type declaration and extensions scopes. In the subsequent PRs could be used to add generic parameter handling and possibly other common behavior.
- This PR also includes some not addressed changes proposed in #2719


Sources/SwiftLexicalLookup/ScopeImplementations.swift
```diff
@@ -96,28 +96,80 @@ import SwiftSyntax
...
       LookupName.getNames(from: codeBlockItem.item, accessibleAfter: codeBlockItem.endPosition)
     }
   }
+
+  @_spi(Experimental) public func _lookup(
+    for identifier: Identifier?,
+    at syntax: SyntaxProtocol,
+    with config: LookupConfig,
+    state: LookupState
+  ) -> [LookupResult] {
```
ahoppen, line 163: Open question:

I find it quite confusing that we always have `self`, which is a syntax node and `syntax`, which  is also a syntax node. As far as I can tell the purpose of `syntax` is to define the position at which lookup occurs so that declarations introduced afterwards are not included. If this is the case, would it be clearer if `at` was an `AbsolutePosition`? If not, do you think some other name than `syntax` could make the distinction from `self` clearer?

Sources/SwiftLexicalLookup/SequentialScopeSyntax.swift
```diff
@@ -0,0 +1,93 @@
...
+        // Extract new names from encountered node.
+        currentChunk.append(
+          contentsOf:
+            LookupName.getNames(
+              from: codeBlockItem.item,
+              accessibleAfter: codeBlockItem.endPosition
+            ).filter { introducedName in
+              checkName(identifier, refersTo: introducedName, at: syntax)
+            }
+        )
```
ahoppen, line 80: Do you think this would read nicer with `+=`? Removes one nesting level. FWIW, I usually prefer `+=` over `append(contentsOf:)`

```suggestion
        currentChunk += LookupName.getNames(
          from: codeBlockItem.item,
          accessibleAfter: codeBlockItem.endPosition
        ).filter { introducedName in
          checkName(identifier, refersTo: introducedName, at: syntax)
        }
```

Sources/SwiftLexicalLookup/SequentialScopeSyntax.swift
```diff
@@ -0,0 +1,93 @@
...
+      }
+    }
+
+    // If there are some names collected, create a new result for this scope.
+    if !currentChunk.isEmpty {
+      result.append(getResults(currentChunk))
+      currentChunk = []
```
ahoppen, line 87: Why do we need to reset `currentChunk` to `[]` here?

Sources/SwiftLexicalLookup/SequentialScopeSyntax.swift
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    return (result.isEmpty ? [] : result.reversed())
```
ahoppen, line 90: Any reason we need to check for `result.isEmpty` at all? Shouldn’t `result.reversed()` also return an empty array if `result` is empty?

Sources/SwiftSyntax/Identifier.swift
```diff
@@ -12,23 +12,58 @@
 
 /// A canonicalized representation of an identifier that strips away backticks.
 public struct Identifier: Equatable, Hashable, Sendable {
-  /// The sanitized `text` of a token.
+  enum IdentifierKind: Hashable {
+    case token(raw: RawIdentifier, arena: SyntaxArenaRef)
+    case string(String)
```
ahoppen, line 17: How often do you create `Identifier` from `String`. I think it might be easier to just write the name to a new `SyntaxArena` and then use it.

ahoppen (commented):
- [ ] Could you update `name` -> `identifier` in various doc comment for which you changed the parameter name?

Tests/SwiftLexicalLookupTest/NameLookupTests.swift
```diff
@@ -553,4 +579,132 @@ final class testNameLookup: XCTestCase {
...
+
+        3️⃣class a {}
+
+        let x = 4️⃣a
+        """,
+      references: [
+        "4️⃣": [
+          .fromFileScope(expectedNames: ["1️⃣"]),
+          .fromFileScope(expectedNames: ["3️⃣"]),
+          .fromScope(GuardStmtSyntax.self, expectedNames: ["2️⃣"]),
```
DougGregor, line 600: Is the order of these meant to be meaningful? They're all effectively in the same scope, right?

Tests/SwiftLexicalLookupTest/NameLookupTests.swift
```diff
@@ -553,4 +579,132 @@ final class testNameLookup: XCTestCase {
...
+            8️⃣oldValue
+          }
+        }
+        """,
+      references: [
+        "2️⃣": [.fromScope(AccessorDeclSyntax.self, expectedNames: [NameExpectation.implicit(.newValue("1️⃣"))])],
+        "4️⃣": [.fromScope(AccessorDeclSyntax.self, expectedNames: [NameExpectation.identifier("3️⃣")])],
+        "6️⃣": [.fromScope(AccessorDeclSyntax.self, expectedNames: [NameExpectation.implicit(.newValue("5️⃣"))])],
+        "8️⃣": [.fromScope(AccessorDeclSyntax.self, expectedNames: [NameExpectation.implicit(.oldValue("7️⃣"))])],
+      ]
```
DougGregor, line 668: This is great!

Sources/SwiftLexicalLookup/LookupName.swift
```diff
@@ -121,9 +173,15 @@ import SwiftSyntax
...
-  {
-    if identifiable.identifier.tokenKind != .wildcard {
+  private static func handle(
+    identifiable: IdentifiableSyntax,
+    accessibleAfter: AbsolutePosition? = nil
+  ) -> [LookupName] {
+    if let closureCapture = identifiable as? ClosureCaptureSyntax,
+      closureCapture.identifier.tokenKind == .keyword(.self)
+    {
+      return [.implicit(.self(closureCapture))]  // Handle `self` closure capture.
```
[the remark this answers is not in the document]
MAJKFL: You're right. This was a leftover from before when we couldn't create `Identifier` from String, since we can't initialize an `Identifier` with a keyword token. Now, I removed the special case.

Sources/SwiftLexicalLookup/TypeScopeSyntax.swift
```diff
@@ -0,0 +1,29 @@
...
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
+@_spi(Experimental) public protocol TypeScopeSyntax: ScopeSyntax, DeclSyntaxProtocol {
+  /// `self` and `Self` names referring to
+  /// this scope.
+  var implicitInstanceAndTypeNames: [LookupName] { get }
```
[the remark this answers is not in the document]
MAJKFL: You're right. I don't think any conforming types will want to override this one. I removed it from the protocol requirements.

Sources/SwiftLexicalLookup/ScopeSyntax.swift
```diff
@@ -59,59 +59,88 @@ extension SyntaxProtocol {
   /// Finds all declarations `name` refers to. `syntax` specifies the node lookup was triggered with.
   /// If `name` set to `nil`, returns all available names at the given node.
   func lookup(
```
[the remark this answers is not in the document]
MAJKFL: I think you're right. It will always just make a call to `_lookup` method with new state. I removed it from the requirements.

Sources/SwiftLexicalLookup/ScopeSyntax.swift
```diff
@@ -59,59 +59,88 @@ extension SyntaxProtocol {
...
+    at syntax: SyntaxProtocol,
+    with config: LookupConfig,
+    state: LookupState
+  ) -> [LookupResult]
 }
 
 @_spi(Experimental) extension ScopeSyntax {
   @_spi(Experimental) public var parentScope: ScopeSyntax? {
     self.parent?.scope
   }
```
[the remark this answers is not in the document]
MAJKFL: For most of the scopes, the default `parentScope` implementation is sufficient, though in some cases we'd expect it to be overridden. Currently, `IfExprSyntax` scope provides such [custom behavior](https://github.com/swiftlang/swift-syntax/blob/059860a66135ee92954ab42c34d82a7100a88151/Sources/SwiftLexicalLookup/ScopeImplementations.swift#L226-L259) to handle `if else` chaining. I think I'd prefer `parentScope` to remain a requirement.

Sources/SwiftLexicalLookup/SequentialScopeSyntax.swift
```diff
@@ -0,0 +1,93 @@
...
+}
+
+extension SequentialScopeSyntax {
+  func sequentialLookup(
+    in codeBlockItems: any Collection<CodeBlockItemSyntax>,
+    for identifier: Identifier?,
+    at syntax: SyntaxProtocol,
+    with config: LookupConfig,
+    state: LookupState,
+    createResultsForThisScopeWith getResults: ([LookupName]) -> (LookupResult)
```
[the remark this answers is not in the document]
MAJKFL: It's here particularly because of the file scope. Since it's results are of `.fromFileScope` kind, we need a way to define which `LookupResult` should we use. Do you think there could be a cleaner way to provide this?

Sources/SwiftLexicalLookup/SequentialScopeSyntax.swift
```diff
@@ -0,0 +1,93 @@
...
+  ) -> [LookupResult] {
+    var result = [LookupResult]()
+    var currentChunk = [LookupName]()
+
+    for codeBlockItem in codeBlockItems {
```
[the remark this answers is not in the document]
MAJKFL: `skipSequentialIntroductionFrom` was only [set when triggering lookup from within `guard ... else` clause](https://github.com/swiftlang/swift-syntax/blob/059860a66135ee92954ab42c34d82a7100a88151/Sources/SwiftLexicalLookup/ScopeImplementations.swift#L342). The condition would still hold for lookup that started there.

But I liked the idea of removing `skipSequentialIntroductionFrom` as it added another layer of complexity to sequential lookup, so I moved the check for lookup triggered in `else` body to [`introducesToSequentialParent` in `guard` scope implementation](https://github.com/swiftlang/swift-syntax/blob/252d3be80c010add2bac9f9712ac32d9892ea574/Sources/SwiftLexicalLookup/ScopeImplementations.swift#L321-L322). 

Should we now remove `LookupState`, or have it available across the API in case we need it in the future?

Sources/SwiftLexicalLookup/ScopeImplementations.swift
MAJKFL: My intent to provide syntax node instead of just position was to have some flexibility in case we want to use e.g. it's `endPosition` instead of `position`. But you're right we're not using it for anything more than just accessing `position` currently. What would you say about [renaming it to `origin`](https://github.com/swiftlang/swift-syntax/blob/252d3be80c010add2bac9f9712ac32d9892ea574/Sources/SwiftLexicalLookup/ScopeSyntax.swift#L66).

Sources/SwiftLexicalLookup/SequentialScopeSyntax.swift
MAJKFL: I'm not sure why did I put it here. Thanks for catching this!

Sources/SwiftLexicalLookup/LookupName.swift
```diff
@@ -12,31 +12,87 @@
 
 import SwiftSyntax
 
+@_spi(Experimental) public enum LookupImplicitNameKind {
```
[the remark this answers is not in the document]
MAJKFL: Thanks for the suggestion. I like the simpler name and also added proper documentation.

Sources/SwiftLexicalLookup/LookupName.swift
```diff
@@ -12,31 +12,87 @@
...
+  /// `self` keyword representing object instance.
+  case `self`(SyntaxProtocol)
```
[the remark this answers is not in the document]
MAJKFL: It had to be `SyntaxProtocol` since we didn't have a way to represent an identifier for `self` keyword in closure capture. Now, as `self` in closure capture is just an explicit `identifier` name kind, I changed it from [`SyntaxProtocol` to `DeclSyntaxProtocol `](https://github.com/swiftlang/swift-syntax/blob/252d3be80c010add2bac9f9712ac32d9892ea574/Sources/SwiftLexicalLookup/LookupName.swift#L20) as in `.Self` case.

Sources/SwiftLexicalLookup/LookupName.swift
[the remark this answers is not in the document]
MAJKFL: You're right, it doesn't really makes sense as the `self` here is explicit. I modeled this here because we didn't have a way to represent non-identifier kind tokens as `Identifier`s. Now, since we can use `String` to create an `Identifier`, we can finally properly [represent this behavior](https://github.com/swiftlang/swift-syntax/blob/252d3be80c010add2bac9f9712ac32d9892ea574/Sources/SwiftLexicalLookup/LookupName.swift#L119).

Sources/SwiftLexicalLookup/LookupState.swift
```diff
@@ -0,0 +1,21 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+
+@_spi(Experimental) public struct LookupState {
```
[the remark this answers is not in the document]
MAJKFL: I added an appropriate doc comment. 

On a side note, thanks to @ahoppen suggestion, I've simplified how we handle lookup from inside `guard ... else` body to [filter those inside `introducesToSequentialParent` in `guard` scope implementation](https://github.com/swiftlang/swift-syntax/blob/252d3be80c010add2bac9f9712ac32d9892ea574/Sources/SwiftLexicalLookup/ScopeImplementations.swift#L321-L322). 

Do you think we should remove `LookupState` now, as it doesn't provide any meaningful functionality, or should we leave it as is. It could potentially be beneficial in the future to have it already accessible across the lookup API in case we want to utilize it.

Sources/SwiftLexicalLookup/SequentialScopeSyntax.swift
```diff
@@ -0,0 +1,93 @@
...
+
+/// Scope that, in addition to names introduced by itself,
+/// also handles names introduced by
+/// `IntroducingToSequentialParentScopeSyntax` children scopes.
+protocol SequentialScopeSyntax: ScopeSyntax {
+  /// Returns names introduced by `codeBlockItems`
+  /// and included `IntroducingToSequentialParentScopeSyntax` children
+  /// scopes that match the lookup.
+  func sequentialLookup(
+    in codeBlockItems: any Collection<CodeBlockItemSyntax>,
```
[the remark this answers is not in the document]
MAJKFL: I changed it to `some Collection` and added a code example.

Sources/SwiftLexicalLookup/SequentialScopeSyntax.swift
MAJKFL: Thanks for catching this! I forgot to remove it presumably after copying the if statement from inside the `for` loop.

Sources/SwiftLexicalLookup/ScopeImplementations.swift
```diff
@@ -96,28 +96,80 @@ import SwiftSyntax
...
+          checkName(identifier, refersTo: lookupName, at: syntax)
+        }
+
+      return names.isEmpty ? [] : [.fromFileScope(self, withNames: names)]
+    case .memberBlockUpToLastDecl:
+      var members = [LookupName]()
+      var sequentialItems = [CodeBlockItemSyntax]()
+      var encounteredNonDeclaration = false
+
+      for codeBlockItem in statements {
```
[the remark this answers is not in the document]
MAJKFL: I think I'm a little confused how should we treat the file scope. When I run this code snippet in a playground
```Swift
class Foo {
  func bar() {
    x.test()
  }
}

let x: X = X()

class X {
  func test() {
    print("It works")
  }
}

Foo().bar()
```
The result is:
```
It works
```
But when I move `Foo().bar()` to before `let x`:
```Swift
class Foo {
  func bar() {
    x.test()
  }
}

Foo().bar()

let x: X = X()

class X {
  func test() {
    print("It works")
  }
}
```
The code compiles (so x is already visible inside `Foo`), but we get `EXC_BAD_ACCESS`. 
Lastly, when we run
```Swift
class Foo {
  func bar() {
    print(x, y)
  }
}

let x: Int? = 1

Foo().bar()

let y: Int? = 2
```
The output is:
```
Optional(1) Optional(0)
```
So does the compiler hoist variables defined after the first non-declaration to the top of the file and default some of their values to some particular ones? Should we maybe represent this playground behavior with some kind of result flag saying "this variable hasn't been initialized yet"? From what I've seen here, the name availability should be similar to the one of member scope (but then also not really, because we also have to somehow handle `guard`). I'm curious what do you think about this.

Tests/SwiftLexicalLookupTest/NameLookupTests.swift
MAJKFL: Yes, the only difference is that the first result represents the "member scope" within that file scope and the second one is from the "code block scope" part after the first non-declaration. 

As I mentioned in [this comment](https://github.com/swiftlang/swift-syntax/pull/2748#discussion_r1700084713), I'm not entirely sure now how should we handle this sort of sequential code that is e.g. used by playgrounds as it seems like all the names should be visible across the whole scope like in a member block.

Sources/SwiftLexicalLookup/SequentialScopeSyntax.swift
```diff
@@ -0,0 +1,93 @@
...
+    at syntax: SyntaxProtocol,
+    with config: LookupConfig,
+    state: LookupState,
+    createResultsForThisScopeWith getResults: ([LookupName]) -> (LookupResult)
+  ) -> [LookupResult]
+}
+
+extension SequentialScopeSyntax {
+  func sequentialLookup(
+    in codeBlockItems: any Collection<CodeBlockItemSyntax>,
```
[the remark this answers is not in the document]
MAJKFL: It's because of how we currently handle the file scope i.e. we treat the part before the first non-declaration as a member block scope and then we use sequential lookup for the other part. That's why we need to be able to pass here just a partition of all the available code block items.

Though it might need to get changed, as I mentioned in [this comment](https://github.com/swiftlang/swift-syntax/pull/2748#discussion_r1700084713), it seems like the visibility in playgrounds is a bit different than what we've modeled so far.

Sources/SwiftSyntax/Identifier.swift
MAJKFL: I think it would be nice to just be able to reuse `RawIdentifier` for this purpose. I can't only think of the place we should store the newly created `SyntaxArena` to keep it alive. Can you think of a way to put it inside `Identifier` and remain `Hashable` and `Sendable` conformance?

Sources/SwiftLexicalLookup/LookupName.swift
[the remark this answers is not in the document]
DougGregor: Sounds good!

Sources/SwiftLexicalLookup/LookupState.swift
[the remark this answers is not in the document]
DougGregor: If we don't need it now, let's remove it! The fewer parameters and less state we need to track, the better. If we find we need to bring it back... well, we know how to do it.

Sources/SwiftLexicalLookup/ScopeImplementations.swift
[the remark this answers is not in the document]
DougGregor: The name lookup rules at file scope are a bit... unfortunate, and you're seeing the effects of that in your example. Long-term, we'd like to fix the mistakes here.

I don't think we should try to determine "hasn't been initialized yet" here in this library. It's something that should be done in a later pass in the compiler that can reason about control flow more effectively.

For this library, we should try to match the existing lookup behavior, and document what it's doing. When it comes time to revisit this behavior (if ever), it'll be a lot easier to fix here than in the C++ implementation of unqualified lookup.

Tests/SwiftLexicalLookupTest/NameLookupTests.swift
DougGregor: Thanks for clarifying.

DougGregor:
@swift-ci please test

Sources/SwiftLexicalLookup/ScopeImplementations.swift
ahoppen: > My intent to provide syntax node instead of just position was to have some flexibility in case we want to use e.g. it's `endPosition` instead of `position`.

Wouldn’t that be an argument for making the parameter an `AbsolutePosition`? If we want to use `endPosition` instead of `position` in some cases, we would just need to change it at the location that calls `lookup` instead of at every location that does a position conversion to `origin`.

Sources/SwiftLexicalLookup/ScopeImplementations.swift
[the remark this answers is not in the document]
ahoppen: > I'm a little confused myself as to why we're gathering declarations up to the first non-declaration, then gathering everything else as code block items, because I don't consider the first non-declaration to be special in Swift's lookup rules.

Does this mean that we should remove the `memberBlockUpToLastDecl` lookup kind?

Sources/SwiftLexicalLookup/ScopeImplementations.swift
```diff
@@ -96,28 +96,80 @@ import SwiftSyntax
...
+      var sequentialItems: [CodeBlockItemSyntax] = []
+      var encounteredNonDeclaration = false
+
+      for codeBlockItem in statements {
+        let item = codeBlockItem.item
+
+        if encounteredNonDeclaration {
+          sequentialItems.append(codeBlockItem)
+        } else {
+          if item.is(DeclSyntax.self) || item.is(VariableDeclSyntax.self) {
```
ahoppen, line 124:
```suggestion
          if item.is(DeclSyntax.self) {
```

Sources/SwiftLexicalLookup/LookupState.swift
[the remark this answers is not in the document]
ahoppen: I agree that we should remove `LookupState` altogether.

Sources/SwiftLexicalLookup/IntroducingToSequentialParentScopeSyntax.swift
```diff
@@ -0,0 +1,24 @@
...
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
+protocol IntroducingToSequentialParentScopeSyntax: ScopeSyntax {
+  /// Returns names matching lookup that should be
+  /// handled by it's parent sequential scope.
+  func introducesToSequentialParent(
```
ahoppen, line 18: I find it a little surprising that `ScopeSyntax` has an API that returns all names introduce in its scope (`introducedNames`) while this API contains a filter. I think it would be more consistent if this also returned all the names introduced to the parent scope and then be named `namesIntroducedToParentScope`. As far as I can tell it wouldn’t need to take any parameters at all then and could be a computed property as well.

Sources/SwiftLexicalLookup/ScopeImplementations.swift
```diff
@@ -265,15 +312,57 @@ import SwiftSyntax
...
+
+@_spi(Experimental) extension ActorDeclSyntax: TypeScopeSyntax {}
+@_spi(Experimental) extension ClassDeclSyntax: TypeScopeSyntax {}
+@_spi(Experimental) extension StructDeclSyntax: TypeScopeSyntax {}
+@_spi(Experimental) extension EnumDeclSyntax: TypeScopeSyntax {}
+@_spi(Experimental) extension ExtensionDeclSyntax: TypeScopeSyntax {}
+
+@_spi(Experimental) extension AccessorDeclSyntax: ScopeSyntax {
+  /// Implicit and/or explicit names introduced
+  /// within the accessor.
```
ahoppen, line 346: I think this would also fit on one line.

Sources/SwiftLexicalLookup/ScopeImplementations.swift
```diff
@@ -265,15 +312,57 @@ import SwiftSyntax
...
+  @_spi(Experimental) public var introducedNames: [LookupName] {
+    if let parameters {
+      return LookupName.getNames(from: parameters)
     } else {
-      return defaultLookupImplementation(for: name, at: syntax, with: config)
+      switch accessorSpecifier.tokenKind {
+      case .keyword(.set), .keyword(.willSet):
+        return [.implicit(.newValue(self))]
+      case .keyword(.didSet):
+        return [.implicit(.oldValue(self))]
```
ahoppen, line 355: All these accessors can also specify custom names instead of using the implicit `newValue` or `oldValue`. Eg. 

```swift
var x: Int = 2 {
  didSet(myNewValue) {
    print(myNewValue)
  }
}
```

Sources/SwiftLexicalLookup/SequentialScopeSyntax.swift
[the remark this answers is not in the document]
ahoppen: Why do we need to differentiate between `fromScope` and `fromFileScope` at all? Couldn’t we conform `SourceFileSyntax` to `ScopeSyntax` and then consider the cases the same?

Sources/SwiftLexicalLookup/SequentialScopeSyntax.swift
```diff
@@ -0,0 +1,108 @@
...
+          accessibleAfter: codeBlockItem.endPosition
+        ).filter { introducedName in
+          checkName(identifier, refersTo: introducedName, at: origin)
+        }
+      } else {
+        itemsWithoutNamedDecl.append(codeBlockItem)
+      }
+    }
+
+    for codeBlockItem in itemsWithoutNamedDecl {
```
ahoppen, line 66: Would it make sense to do this in the `else` clause of the loop above? Also, technically we could have a type that conforms to `NamedDeclSyntax` and `IntroducingToSequentialParentScopeSyntax`.

Sources/SwiftLexicalLookup/SequentialScopeSyntax.swift
```diff
@@ -0,0 +1,108 @@
...
+        // If there are some names collected, create a new result for this scope.
+        if !currentChunk.isEmpty {
+          result.append(getResults(currentChunk))
+          currentChunk = []
+        }
+
+        result += introducedResults
+      } else {
+        // Extract new names from encountered node.
+        currentChunk += LookupName.getNames(
```
ahoppen, line 92: Why do we need `currentChunk` at all? Couldn’t we add the names directly to `result`?

Sources/SwiftLexicalLookup/SequentialScopeSyntax.swift
```diff
@@ -0,0 +1,108 @@
...
+          from: codeBlockItem.item,
+          accessibleAfter: codeBlockItem.endPosition
+        ).filter { introducedName in
+          checkName(identifier, refersTo: introducedName, at: origin)
```
ahoppen, line 96: I guess we could also move the `filter` all the way to the end instead of duplicating it.

Sources/SwiftLexicalLookup/SequentialScopeSyntax.swift
```diff
@@ -0,0 +1,108 @@
...
+        }
+      }
+    }
+
+    // If there are some names collected, create a new result for this scope.
+    if !currentChunk.isEmpty {
+      result.append(getResults(currentChunk))
+    }
+
+    return result.reversed() + lookupInParent(for: identifier, at: origin, with: config, state: state)
```
ahoppen, line 106: Instead of reversing the result here (which requires a copy of the array), have you considered changing the semantics of the result to say a later element overrides earlier elements? I think I would also find that more intuitive because it matches the source order.

Sources/SwiftLexicalLookup/SequentialScopeSyntax.swift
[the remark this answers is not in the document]
ahoppen: I was also thinking about performance here and that code block items after the lookup position can’t be relevant but I forgot that there could be a local class declaration after the lookup position that the lookup position refers to.

Sources/SwiftSyntax/Identifier.swift
ahoppen: I just checked and I think that in production all usages of `Identifier(String)` use statically known strings. For that we can continue using the current `RawIdentifier` because we can initialize a `SyntaxText` from a `StaticString`

```swift
public struct RawIdentifier: Equatable, Hashable, Sendable {
  … 
  fileprivate init(_ raw: StaticString) {
    name = SyntaxText(raw)
  }
  …
}
```

And I don’t want to change the representation for tests only. Also, in tests performance does not really matter, so there we can add

```swift
fileprivate extension Identifier {
  @_spi(Testing) public init?(string: String) {
    self.init(TokenSyntax.identifier(string))
  }
}
```

Sources/SwiftLexicalLookup/LookupName.swift
```diff
@@ -27,16 +107,20 @@ import SwiftSyntax
...
     }
   }
 
-  /// Introduced name.
+  /// Identifier used for name comparison.
   @_spi(Experimental) public var identifier: Identifier? {
     switch self {
     case .identifier(let syntax, _):
-      return Identifier(syntax.identifier)
+      return Identifier(syntax.identifier) ?? Identifier(syntax.identifier.text)
```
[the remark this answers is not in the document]
MAJKFL: I added new `LookupName` kinds: `self` and `Self`. As for now it's only really used in closure captures, but I agree it's nice to have a separate representation for both of those.

Sources/SwiftLexicalLookup/ScopeImplementations.swift
MAJKFL: I was thinking more of the API deciding whether to use `position` or `endPosition`, but more I think about now, the less I actually like this idea since it would be unnecessarily complex and ambiguous for clients. I changed it now to be just `AbsolutePosition`.

Sources/SwiftLexicalLookup/LookupState.swift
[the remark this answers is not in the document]
MAJKFL: Ok. I removed `LookupState` and the internal `_lookup` method.

Sources/SwiftLexicalLookup/IntroducingToSequentialParentScopeSyntax.swift
MAJKFL: I think that was poor naming on my part as `introducesToSequentialParent` worked more like `lookup` producing `[LookupResult]` to be interleaved with results of the sequential parent, rather than `introducedNames` that just return all `LookupName`s available at that scope. I renamed `introducesToSequentialParent` to less misleading `lookupFromSequentialParent` and added `namesIntroducedToSequentialParent` that closely resembles `introducedNames` of a normal scope.

Sources/SwiftLexicalLookup/ScopeImplementations.swift
MAJKFL: It already actually [works](https://github.com/swiftlang/swift-syntax/blob/cf273f8943c52a8df8c7348c81a5120b16c577c7/Tests/SwiftLexicalLookupTest/NameLookupTests.swift#L649-L651), as we first try to [extract names from `parameters` if available ](https://github.com/swiftlang/swift-syntax/blob/cf273f8943c52a8df8c7348c81a5120b16c577c7/Sources/SwiftLexicalLookup/ScopeImplementations.swift#L347).

Sources/SwiftLexicalLookup/SequentialScopeSyntax.swift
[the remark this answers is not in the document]
MAJKFL: It was a design decision we agreed on with @DougGregor to further differentiate the weird behavior of file scope from the rest of the scopes. `SourceFileSyntax` already conforms to `ScopeSyntax`, but I think it's nice to have it's results separated.

Sources/SwiftLexicalLookup/SequentialScopeSyntax.swift
MAJKFL: Sequential scope needs to ensure all type declarations are available in the whole scope (first loop) and then that results from `guard` (and possibly other `IntroducingToSequentialParentScopeSyntax` in the future) are properly interleaved with the results produced by this scope. That's also why we can't add names directly to `results` as then we would loose this sort of partitioning and proper interleaving with the results from `guard` etc. I think [this test](https://github.com/swiftlang/swift-syntax/blob/c83b3c6f3454a05b0a22ce26f12de594de19d68b/Tests/SwiftLexicalLookupTest/NameLookupTests.swift#L803-L832) nicely exemplifies this behavior.

Sources/SwiftLexicalLookup/SequentialScopeSyntax.swift
MAJKFL: Again, I think it was a bit poor naming on my side as `result` was an array of `LookupResult`. I renamed it now to `results`. 

As I briefly mentioned in [this comment](https://github.com/swiftlang/swift-syntax/pull/2748#discussion_r1701626317), we can't add the names directly to `results` as we need to partition them based on results obtained from `IntroducingToSequentialParentScopeSyntax`.

Sources/SwiftLexicalLookup/SequentialScopeSyntax.swift
MAJKFL: As we already represent results from inner-most to outer-most, I think it makes most sense to reverse it. Like in [this example](https://github.com/swiftlang/swift-syntax/blob/c83b3c6f3454a05b0a22ce26f12de594de19d68b/Tests/SwiftLexicalLookupTest/NameLookupTests.swift#L803-L832):
```Swift
func foo() {
  let a = 1 // <-- Code block scope
  let b = 2

  guard let b = a else { return }

  let a = 3 // <-- Code block scope
  let b = 4

  guard let a = b else { return } // <-- Guard scope

  print(a, b) // <-- lookup for a
}
```
I think the order of results should be: `guard` scope and exactly one code block scope result (as the first `guard` is not relevant for `a` lookup).

The order of `LookupName`s within `LookupResult`s is the way you described actually. In the result from code block scope the order is exactly: `let a = 1` and then `let a = 3`.

Sources/SwiftLexicalLookup/SequentialScopeSyntax.swift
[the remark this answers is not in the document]
MAJKFL: That's why I added now the first loop to scan for class declarations (as we need them in the first code block result to be available in the whole scope no matter the partition). Because to this change, we can now break early. Thanks for the suggestion!

Sources/SwiftSyntax/Identifier.swift
MAJKFL: That's a good idea, thanks for helping with simplifying this!

Turns out we don't even need a `String` initializer for the tests as we can just directly use keyword tokens' `rawText` property.
```Swift
extension Identifier {
  @_spi(Testing) public init(anyToken token: TokenSyntax) {
    self.raw = RawIdentifier(token.tokenView.rawText)
    self.arena = token.raw.arenaReference
  }
}
```

Sources/SwiftLexicalLookup/SequentialScopeSyntax.swift
MAJKFL: As we need to create multiple `LookupResult`s here, I think the only other possible place to move filtering to, would be when creating a new result from collected `currentChunk`. Either way, we'd need to traverse the same amount of items when filtering so the time complexity would be the same, but this way we can maintain smaller sizes of `currentChunk`.

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test Windows

Sources/SwiftLexicalLookup/ScopeImplementations.swift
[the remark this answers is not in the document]
DougGregor: Maybe? There's definitely a difference here for the source file scope when top-level code is involved, and we need a way to model it. I think we should leave this in place for now and we can rename/revise as we learn more.

Sources/SwiftLexicalLookup/SequentialScopeSyntax.swift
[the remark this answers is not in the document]
DougGregor: Part of the reason for having `fromFileScope` be separate is because clients need to do extra work to do lookups into other source files and in visible imports from that scope, so it's both "here's what I found in this file" and "you need to go look elsewhere, too". I'm not totally settled on what the right answer is for dealing with file scope, so I'd rather keep thinking about it (but not block this PR on a resolution).

DougGregor (approved):
While we have some open questions about the handling of top-level code, I'm happy with where this pull request is now and I think we should go ahead and land it once CI passes.

DougGregor:
```
/home/build-user/swift-syntax/Tests/SwiftSyntaxTest/IdentifierTests.swift:42:43: error: cannot use optional chaining on non-optional value of type 'RawIdentifier'
40 |     let token = TokenSyntax(stringLiteral: "sometoken")
41 |     withExtendedLifetime(token) { token in
42 |       XCTAssertEqual(token.identifier?.raw?.name, SyntaxText("sometoken"))
   |                                           `- error: cannot use optional chaining on non-optional value of type 'RawIdentifier'
43 |     }
44 |   }
```

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

MAJKFL:
Not sure what failed here now. I see in the console output all tests passed
```
Test Suite 'All tests' passed at 2024-08-02 23:13:38.842
	 Executed 640 tests, with 24 tests skipped and 0 failures (0 unexpected) in 448.763 (448.763) seconds
```
Though there are multiple „message decoding errors” from LSP like
```
Failed to decode message: MessageDecodingError(code: LanguageServerProtocol.ErrorCode(rawValue: -32602), message: "missing expected parameter: string", id: nil, messageKind: LanguageServerProtocol.MessageDecodingError.MessageKind.notification)
```
Could this be it has something to do with the new `Identifier` initializers?

ahoppen:
It crashed with 

```
SourceKitLSPPackageTests.xctest: /home/build-user/swift/stdlib/public/Concurrency/TaskPrivate.h:803: void swift::AsyncTask::PrivateStorage::complete(AsyncTask *): Assertion `oldStatus.getInnermostRecord() == NULL' failed.
```

It’s a non-deterministic failure on Linux, though I haven’t seen it for months. Unrelated to your changes though.

DougGregor:
@swift-ci please test Linux

Sources/SwiftSyntax/Identifier.swift
```diff
@@ -12,23 +12,38 @@
...
+  /// The sanitized name of the identifier.
   public var name: String {
     String(syntaxText: raw.name)
   }
 
   @_spi(RawSyntax)
   public let raw: RawIdentifier
-
-  private let arena: SyntaxArenaRef
+  let arena: SyntaxArenaRef?
```
ahoppen, line 22: Why do we need to make `arena` non-private? It should only be used to keep the arena alive and we shouldn’t expose it.

Sources/SwiftSyntax/Identifier.swift
```diff
@@ -12,23 +12,38 @@
...
+
+  public static func == (lhs: Self, rhs: Self) -> Bool {
+    lhs.name == rhs.name
+  }
+}
+
+extension Identifier {
+  @_spi(Testing) public init(anyToken token: TokenSyntax) {
+    self.raw = RawIdentifier(token.tokenView.rawText)
+    self.arena = token.raw.arenaReference
```
ahoppen, line 46: I’m not a fan of this initializer because it deliberately discards the difference between keywords and identifiers in a very subtle way. I assume that we only need it for `self` and `Self`, right? Since these behave identifier-like, what do you think about changing the existing `init?(_ token: TokenSyntax)` to initialize itself with `self` and `Self` if the passed in token is `self` or `Self`?

I haven’t thought it through but this might also simplify the name lookup logic because you no longer need to differentiate between the `self` keyword and `self` identifier`. Maybe `LookupName.self` and `LookupName.Self` won’t be needed anymore.

DougGregor:
Alright, let's go ahead and merge this!

MAJKFL:
Thanks for merging! I included suggested changes by @ahoppen in #2782.

Sources/SwiftSyntax/Identifier.swift
MAJKFL: I think this is a good idea indeed. Not being able to create an `Identifier` from `self` and `Self` caused a lot of trouble that now could be handled out of the box.

Sources/SwiftLexicalLookup/SequentialScopeSyntax.swift
MAJKFL: Hmm I think you have a point here. Though, I think it's a bit hard to nail all the nuances without comparing them directly to the current implementation. I think I'd prefer to wait with fixing details like this until we hook up `SwiftLexicalLookup` to the compiler and start testing it there.

PR #2750, opened 2024-07-24
Fix build warnings

ahoppen:
This gets swift-syntax building without warnings again.

rdar://131047703


ahoppen:
@swift-ci Please test

PR #2751, opened 2024-07-24
[6.0] Fix build warnings

ahoppen:
- **Explanation**: Fix build warnings so that the version of swift-syntax we release as a package dependency doesn’t contain any warnings
- **Scope**: Tests only
- **Risk**: Very low, test only change with no functionality change
- **Testing**: Verified that swift-syntax now builds without warnings on macOS and Linux
- **Issue**: rdar://131047703
- **Reviewer**:   @bnbarham


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #2752, opened 2024-07-24
Allow building swift-syntax in C++ interop mode

ahoppen:
If swift-syntax is being built with `-cxx-interoperability-mode=default`, it fails because of `annot initialize a variable of type 'AtomicBool *' with an rvalue of type 'void *'`. Add a cast here to fix the build issue.

rdar://129252735


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #2754, opened 2024-07-25
Emit a warning if a conformance to `MemberMacro` does not contain the `expansion` method that takes a `conformingTo` parameter

ahoppen:
The `expansion` method that takes a `conformingTo` has been defaulted to be source-compatible with older versions of swift-syntax that required a version of the `expansion` function without the `conformingTo` parameter.

That meant that you could state a conformance to `MemberMacro` without implementing any methods and you wouldn’t get any compile time diagnostics.

Deprecate the forwarding default implementation so that you get a warning if you are relying on it.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #2758, opened 2024-07-27
Introduce FixIt.Change API that replaces the child of a syntax node

### LINKED ISSUE #2205: Introduce `FixIt.Change` API that replaces the child of a syntax node
Currently, Fix-Its are only able to replace syntax nodes but that makes it impossible to replace a `nil` node by a proper syntax node. 

We should offer a `FixIt.Change` that replaces the child of a syntax node, with a signature that’s something like this (written in GitHub, not sure if it compiles, but it gets the idea across).

```swift
FixIt.Change.replaceChild<Parent, Child>(_: KeyPath<Parent, Child>, of parent: Parent, with: Child)
```

`FixItApplier` will need to be updated to also handle this new change kind

rdar://115562892

AppAppWorks:
fix #2205

### DIFF
--- a/Sources/SwiftCompilerPluginMessageHandling/PluginMacroExpansionContext.swift
+++ b/Sources/SwiftCompilerPluginMessageHandling/PluginMacroExpansionContext.swift
@@ -170,19 +170,24 @@ class SourceManager {
     from startKind: PositionInSyntaxNode = .afterLeadingTrivia,
     to endKind: PositionInSyntaxNode = .beforeTrailingTrivia
   ) -> SourceRange? {
+    range(node.position(at: startKind)..<node.position(at: endKind), in: node)
+  }
+
+  /// Get ``SourceRange`` (file name + UTF-8 offset range) of `localRange` in `node`'s root node, which must be one
+  /// of the returned values from `add(_:)`.
+  func range(
+    _ localRange: @autoclosure () -> Range<AbsolutePosition>,
+    in node: some SyntaxProtocol
+  ) -> SourceRange? {
     guard let base = self.knownSourceSyntax[node.root.id] else {
       return nil
     }
-    let localStartPosition = node.position(at: startKind)
-    let localEndPosition = node.position(at: endKind)
-    precondition(localStartPosition <= localEndPosition)
-
     let positionOffset = base.location.offset
-
+    let localRange = localRange()
     return SourceRange(
       fileName: base.location.fileName,
-      startUTF8Offset: localStartPosition.advanced(by: positionOffset).utf8Offset,
-      endUTF8Offset: localEndPosition.advanced(by: positionOffset).utf8Offset
+      startUTF8Offset: localRange.lowerBound.advanced(by: positionOffset).utf8Offset,
+      endUTF8Offset: localRange.upperBound.advanced(by: positionOffset).utf8Offset
     )
   }
 
--- a/Sources/SwiftDiagnostics/FixIt.swift
+++ b/Sources/SwiftDiagnostics/FixIt.swift
@@ -27,15 +27,56 @@ public protocol FixItMessage: Sendable {
   var fixItID: MessageID { get }
 }
 
+/// Types conforming to this protocol provide the data required for replacing a child node of a parent node.
+///
+/// Conforming types should ensure the child of ``parent`` to be replaced at ``replacementRange`` is type-compatible
+/// with ``newChild``. Conforming types are stored as type-erased existentials (i.e. `any ReplacingChildData`) in
+///  ``FixIt/Change/replaceChild(data:)`` to keep ``FixIt`` type-erased.
+public protocol ReplacingChildData: Sendable {
+  associatedtype Parent: SyntaxProtocol
+  associatedtype Child: SyntaxProtocol
+
+  /// The node whose child node at ``replacementRange`` to be replaced by ``newChild``.
+  var parent: Parent { get }
+
+  /// The node to replace the child node of ``parent`` at ``replacementRange``.
+  var newChild: Child { get }
+
+  /// The absolute position range of the child node to be replaced.
+  ///
+  /// If a nil child node is to be replaced, conforming types should provide a zero-length range with both bounds
+  /// denoting the start position of ``newChild`` in ``parent`` after replacement.
+  var replacementRange: Range<AbsolutePosition> { get }
+}
+
 /// A Fix-It that can be applied to resolve a diagnostic.
 public struct FixIt: Sendable {
   public enum Change: Sendable {
+    struct ReplacingOptionalChildData<Parent: SyntaxProtocol, Child: SyntaxProtocol>: ReplacingChildData {
+      let parent: Parent
+      let newChild: Child
+      let keyPath: WritableKeyPath<Parent, Child?> & Sendable
+
+      var replacementRange: Range<AbsolutePosition> {
+        // need to upcast keyPath to strip Sendable for older Swift versions
+        let keyPath: WritableKeyPath<Parent, Child?> = keyPath
+        if let oldChild = parent[keyPath: keyPath] {
+          return oldChild.range
+        } else {
+          let newChild = parent.with(keyPath, newChild)[keyPath: keyPath]!
+          return newChild.position..<newChild.position
+        }
+      }
+    }
+
     /// Replace `oldNode` by `newNode`.
     case replace(oldNode: Syntax, newNode: Syntax)
     /// Replace the leading trivia on the given token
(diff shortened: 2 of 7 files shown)


Tests/SwiftSyntaxMacroExpansionTest/PeerMacroTests.swift
```diff
@@ -62,9 +60,10 @@ final class PeerMacroTests: XCTestCase {
                 "add 'async'"
               ),
               changes: [
-                FixIt.Change.replace(
-                  oldNode: Syntax(funcDecl.signature),
-                  newNode: Syntax(newSignature)
+                FixIt.Change(
+                  parent: funcDecl,
+                  replacingChildAt: \.signature.effectSpecifiers,
+                  with: newEffects
```
AppAppWorks, line 66: I guess `FunctionDeclSyntax.signature.effectSpecifiers` is a valid use case of this new API?

Sources/SwiftDiagnostics/FixIt.swift
```diff
@@ -36,6 +36,33 @@ public struct FixIt: Sendable {
     case replaceLeadingTrivia(token: TokenSyntax, newTrivia: Trivia)
     /// Replace the trailing trivia on the given token
     case replaceTrailingTrivia(token: TokenSyntax, newTrivia: Trivia)
+    /// Replace a child of a parent node
+    case replaceChild(
+      _ keyPath: AnyKeyPath & Sendable,
+      parent: Syntax,
+      child: Syntax,
+      replacingRange: @Sendable () -> Range<AbsolutePosition>
+    )
```
ahoppen, line 45: Interesting idea with making `replacingChild` a member of `replaceChild`. I am a little worried that it’s possible to construct a `replaceChild` case with `parent`, `child`, `keyPath` and `replacingRange` members that are incompatible with each other if a user doesn’t use the initializer you defined.

Instead, what do you think about storing the type safe generic version of the data and then expose all the necessary information through a protocol? Something along the lines of the following (really not fleshed out yet, just to give an idea).

```swift
protocol AnyReplacingChildData {
  var replacementRange: Range<AbsoluetePosition>
  var newChild: Syntax
}

struct ReplacingChildData<Parent, Child>: AnyReplacingChildData {
  let parent: Parent
  let newChild: Child
  let keyPath: WritableKeyPath<Parent, Child>
}

case replaceChild(any AnyReplacingChildData)
```

Sources/SwiftCompilerPluginMessageHandling/PluginMacroExpansionContext.swift
```diff
@@ -186,6 +186,23 @@ class SourceManager {
     )
   }
 
+  func range(
+    of node: some SyntaxProtocol,
+    localRange: Range<AbsolutePosition>
+  ) -> SourceRange? {
```
ahoppen, line 192: Could you add a doc comment to this method? 

Also: Could `range(of:from:to:)` call into this method instead of duplicating logic?

Sources/SwiftDiagnostics/FixIt.swift
```diff
@@ -27,15 +27,45 @@ public protocol FixItMessage: Sendable {
...
 }
 
+public protocol ReplacingChildData: Sendable {
+  associatedtype Parent: SyntaxProtocol
+  associatedtype Child: SyntaxProtocol
+
+  var replacementRange: Range<AbsolutePosition> { get }
+  var parent: Parent { get }
+  var newChild: Child { get }
+}
```
ahoppen, line 37: Could you add comments to the protocol and its members? Also, I think it would be good to describe why this is a protocol and can’t be members on the `FixIt.Change.replaceChild` case.

Sources/SwiftDiagnostics/FixIt.swift
```diff
@@ -27,15 +27,45 @@ public protocol FixItMessage: Sendable {
...
+      var replacementRange: Range<AbsolutePosition> {
+        if let oldChild = parent[keyPath: keyPath] {
+          return oldChild.range
+        } else {
+          let newParent = parent.with(keyPath, newChild)
+          if let previousToken = newParent[keyPath: keyPath]?.previousToken(viewMode: .all) {
+            return previousToken.endPosition..<previousToken.endPosition
+          } else {
+            return parent.position..<parent.position
+          }
```
ahoppen, line 56: Is it necessary to use `previousToken().endPosition`? Can’t you use `newParent[keyPath: keyPath].position`?

ahoppen (commented):
Very nice stuff.

Could you also add an entry to the 601 release notes? Adding a new case to `FixIt.Change` is technically source breaking because clients might exhaustively switch over it.

Sources/SwiftDiagnostics/FixIt.swift
AppAppWorks: Agree, but should we add more test cases for this new API?

ahoppen: I think the existing test in `PeerMacroTests` that generates a `replaceChild` Fix-It should be sufficient.

Sources/SwiftCompilerPluginMessageHandling/PluginMacroExpansionContext.swift
```diff
@@ -170,19 +170,35 @@ class SourceManager {
     from startKind: PositionInSyntaxNode = .afterLeadingTrivia,
     to endKind: PositionInSyntaxNode = .beforeTrailingTrivia
   ) -> SourceRange? {
-    guard let base = self.knownSourceSyntax[node.root.id] else {
+    if let base = self.knownSourceSyntax[node.root.id] {
+      return range(node.position(at: startKind)..<node.position(at: endKind), in: base)
+    } else {
       return nil
     }
```
ahoppen, line 177: I feel like I’m missing something. Can’t this call

```suggestion
    return range(node.position(at: startKind)..<node.position(at: endKind), in: node)
```

And then we wouldn’t need the version that takes a `KnownSourceSyntax` at all.

Sources/SwiftCompilerPluginMessageHandling/PluginMacroExpansionContext.swift
```diff
@@ -170,19 +170,35 @@ class SourceManager {
...
-    let localStartPosition = node.position(at: startKind)
-    let localEndPosition = node.position(at: endKind)
-    precondition(localStartPosition <= localEndPosition)
```
ahoppen, line 178: Should we keep this precondition?

AppAppWorks: Hasn't this precondition been enforced by the range constructor?

ahoppen: True

ahoppen (approved):

ahoppen:
@swift-ci Please test

AppAppWorks:
Fixed the type checker issue, please test again.

bnbarham:
@swift-ci please test

ahoppen:
@swift-ci Please test Windows

AppAppWorks:
I figured that older Swift versions wouldn't compile for `[keyPath: (WritableKeyPath<Parent, Child?> & Sendable)]` but `[keyPath: WritableKeyPath<Parent, Child?>]`.

ahoppen:
@swift-ci Please test

AppAppWorks:
The CI prompt seems to have failed again ;(

ahoppen:
@swift-ci Please test

ahoppen:
Adding a new case to `FixIt.Change` means that the compiler needs to be handle it [here](https://github.com/swiftlang/swift/blob/fe238cf2d1cedb0f46dc7d431727d25ec7f16253/lib/ASTGen/Sources/ASTGen/DiagnosticsBridge.swift#L60-L82). Would you like to try implementing that change and opening a PR to [swiftlang/swift](https://github.com/swiftlang/swift)? We can then test the two PRs together.

AppAppWorks:
It'd be a great idea, but I wonder how could I orchestrate the version bump across two projects?

ahoppen:
We can test the PRs together using [Cross Repository Testing](https://github.com/swiftlang/swift/blob/main/docs/ContinuousIntegration.md#cross-repository-testing) and then we’ll merge them at the same time.

ahoppen:
https://github.com/swiftlang/swift/pull/75582

@swift-ci Please test

ahoppen:
https://github.com/swiftlang/swift/pull/75582

@swift-ci Please test Linux

PR #2763, opened 2024-07-29
Replace expression in `ClosureCaptureSyntax` with initializer clause

MAJKFL:
This PR adjusts `ClosureCaptureSyntax` node.
- `name` is no longer optional, making it more convenient to refer it by name (as discussed under #2719).
- `equal` and `expression` were replaced by the optional `initializer` property of the `InitializerClauseSyntax` type, making it more uniform with variable declarations like `let a = b` and function parameters like `a: Int = 123`.
- Due to the new node structure, removed the convenience initializer introduced with #2127 as `name` is not optional and equal sign is no longer part of the node itself.


Sources/SwiftParser/Expressions.swift
```diff
@@ -1704,29 +1704,28 @@ extension Parser {
...
           // The thing being capture specified is an identifier, or as an identifier
-          // followed by an expression.
+          // followed by an initializer clause.
           let unexpectedBeforeName: RawUnexpectedNodesSyntax?
-          let name: RawTokenSyntax?
-          let unexpectedBeforeEqual: RawUnexpectedNodesSyntax?
-          let equal: RawTokenSyntax?
-          let expression: RawExprSyntax
+          let name: RawTokenSyntax
+          let initializer: RawInitializerClauseSyntax?
```
ahoppen, line 1710: You should be able to hoist the parsing of `name` out of the `if` statement and then use `self.at(.equal)` instead of `self.peek(isAt: .equal)`

ahoppen (commented):
Instead of removing any public API, could you add the existing API signatures as deprecated methods to SwiftSyntaxCompatibility.swift with the most reasonable implementation possible? 

Also, could you add the changes to the [601 release notes](https://github.com/swiftlang/swift-syntax/blob/main/Release%20Notes/601.md)?

Sources/SwiftParser/Expressions.swift
MAJKFL: Thanks for the suggestion. I moved name parsing to before the `if` statement.

As an interesting artifact, `self` token will not be remapped to `identifier` when parsing capture of `[self = 123]` form. Out of curiosity, was there a particular reason for this remapping? As far as I checked, e.g. in `let self = 123` `self` is classified as a keyword.

Sources/SwiftSyntax/generated/RenamedChildrenCompatibility.swift
```diff
@@ -1167,73 +1167,6 @@ extension ClassDeclSyntax {
...
-        unexpectedBetweenAssignTokenAndExpression, 
-        expression: expression, 
-        unexpectedBetweenExpressionAndTrailingComma, 
-        trailingComma: trailingComma, 
-        unexpectedAfterTrailingComma, 
-        trailingTrivia: trailingTrivia
-      )
-  }
-}
-
```
ahoppen, line 1236: People might still be relying on these compatibility overloads, so we shouldn’t remove them.

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxNodesC.swift
ahoppen, line 1: We should continue to offer deprecated compatibility versions of all the old signatures in this file. SwiftSyntaxCompatibility.swift should have a few examples.

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxNodesC.swift
```diff
@@ -1635,16 +1628,16 @@ public struct ClosureCaptureSyntax: SyntaxProtocol, SyntaxHashable, _LeafSyntaxN
   /// ### Tokens
   /// 
   /// For syntax trees generated by the parser, this is guaranteed to be `<identifier>`.
-  public var name: TokenSyntax? {
+  public var name: TokenSyntax {
```
ahoppen, line 1631: Changing `name` to be non-optional is an API-breaking change because eg. `closuereCapture.name?.text` will stop compiling. There is no way of making it API-compatible. Could you highlight it in the release notes?

Release Notes/601.md
```diff
@@ -14,6 +14,10 @@
 - `Error` protocol now has an `asDiagnostics(at:)` method.
   - Description: This method translates an error into one or more diagnostics, recognizing `DiagnosticsError` and `DiagnosticMessage` instances or providing its own `Diagnostic` as needed.
   - Pull Request: https://github.com/swiftlang/swift-syntax/pull/1816
+  
+- `ClosureCaptureSyntax` now has a new node structure.
+  - Description: `ClosureCaptureSyntax` now has an `initializer` property instead of `equal` and `expression`. Additionally, the `name` property is no longer optional.
+  - Pull request: https://github.com/swiftlang/swift-syntax/pull/2763
```
ahoppen, line 20: I would move this to the `Deprecations` section because that’s the way that users are most likely to hit it: They were using it before and now they have a deprecation warning.

*New APIs* is intended for features that didn’t exist before.

Sources/SwiftSyntax/generated/RenamedChildrenCompatibility.swift
MAJKFL: I added those to `SwiftSyntaxCompatibility.swift`.

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxNodesC.swift
MAJKFL: I added those to `SwiftSyntaxCompatibility.swift`.

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxNodesC.swift
MAJKFL: Thanks for pointing this out. I added a note under "API-Incompatible Changes" in the release notes.

Release Notes/601.md
MAJKFL: Thanks for the suggestion. I moved it to the note under "Deprecations".

ahoppen (commented):
Thanks for adding the compatibility layers. Looks good to me 👍🏽

ahoppen (approved):

ahoppen:
@swift-ci Please test

MAJKFL:
The CI failed because we removed the `self` remapping to `identifier` earlier in this PR and I didn't add it as a possible kind in CodeGeneration. Should be fixed now.

MAJKFL:
Should be good to go now. Could you please test it @ahoppen?

ahoppen:
@swift-ci Please test

MAJKFL:
Had to fix failing `testConsistentNamingOfChildren` test. Seems like `ClosureCaptureSyntax` was picked up as `mostCommonChild` over `MatchingPatternConditionSyntax` because of alphabetical precedence.

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test Windows

MAJKFL:
Hmm... this is an interesting one. Inside `TokenStreamCreator`, when [visiting `ClosureCaptureSyntax` node](https://github.com/swiftlang/swift-format/blob/b60a4eeb16010cc55f30a988913e58d5806fc520/Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift#L1271-L1283), the method adds spaces around the `=` token which seem to be not necessary anymore. I was able to fix the failure locally just by removing those two lines of code (which used deprecated by this PR `ClosureCaptureSyntax.equal` property btw).
```Swift
override func visit(_ node: ClosureCaptureSyntax) -> SyntaxVisitorContinueKind {
  before(node.firstToken(viewMode: .sourceAccurate), tokens: .open)
  after(node.specifier?.lastToken(viewMode: .sourceAccurate), tokens: .break)
  // before(node.equal, tokens: .break)
  // after(node.equal, tokens: .break)
  if let trailingComma = node.trailingComma {
    before(trailingComma, tokens: .close)
    after(trailingComma, tokens: .break(.same))
  } else {
    after(node.lastToken(viewMode: .sourceAccurate), tokens: .close)
  }
  return .visitChildren
}
```
Does it mean we need to fix it there? Or do you think there is a way to fix it in this PR and I'm missing something?

DougGregor:
> Hmm... this is an interesting one. Inside `TokenStreamCreator`, when [visiting `ClosureCaptureSyntax` node](https://github.com/swiftlang/swift-format/blob/b60a4eeb16010cc55f30a988913e58d5806fc520/Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift#L1271-L1283), the method adds spaces around the `=` token which seem to be not necessary anymore. I was able to fix the failure locally just by removing those two lines of code (which used deprecated by this PR `ClosureCaptureSyntax.equal` property btw).
> 
> ```swift
> override func visit(_ node: ClosureCaptureSyntax) -> SyntaxVisitorContinueKind {
>   before(node.firstToken(viewMode: .sourceAccurate), tokens: .open)
>   after(node.specifier?.lastToken(viewMode: .sourceAccurate), tokens: .break)
>   // before(node.equal, tokens: .break)
>   // after(node.equal, tokens: .break)
>   if let trailingComma = node.trailingComma {
>     before(trailingComma, tokens: .close)
>     after(trailingComma, tokens: .break(.same))
>   } else {
>     after(node.lastToken(viewMode: .sourceAccurate), tokens: .close)
>   }
>   return .visitChildren
> }
> ```
> 
> Does it mean we need to fix it there? Or do you think there is a way to fix it in this PR and I'm missing something?

Yes, I think it's okay to fix it there.

ahoppen:
> Edit: Argh, I'm not really sure now. Could it be that the now-deprecated `equal` should produce the node without whitespace? That doesn't seem quite right, either. @ahoppen or @allevato , do either of you have an opinion?

The `equal` property should return the equal token with its trivia because the equal token has trivia attached to it. The compatibility layers are best-effort and this sort of issue might happen because we’re now effectively visiting the equal token twice. We should just fix swift-format like you suggested. @MAJKFL Could you open a PR with that change to swift-format and then we can merge it together with the swift-syntax PR.

allevato:
I agree that we should just update swift-format to stop using the compatibility layer nodes. In cases like this where the fundamental structure of a node changes, it's important that we visit the tree in the way that reflects the actual structure.

ahoppen:
https://github.com/swiftlang/swift-format/pull/790

@swift-ci Please test

PR #2764, opened 2024-07-30
[CMake] Enable package CMO only in newer compiler

rintaro:


rintaro:
@swift-ci Please test

ahoppen (approved):

bnbarham (approved):

PR #2766, opened 2024-07-30
Update MODULE.bazel for 5.10 to support rules_swift 2.x+

luispadron:
This is required to support rules_swift 2.x on the 5.10 releases. The `main` branch already contains this change so it should go out with the version 6 releases

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -5,4 +5,4 @@ module(
 )
 
 bazel_dep(name = "rules_swift", version = "1.16.0", repo_name = "build_bazel_rules_swift")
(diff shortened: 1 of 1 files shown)


keith:
@swift-ci please test

keith (approved):

bnbarham (approved):

luispadron:
@ahoppen would it be possible to tag a release with this change so it can be uploaded to the Bazel registry? 🙏🏼

ahoppen:
Here we go: https://github.com/swiftlang/swift-syntax/releases/tag/510.0.3

Issue #1805, opened 2023-06-16
SyntaxProtocol.expand() does not expand peer macros introduced by other macros


tgoyne:
### Description

This test fails due to `Foo` not having the second `bar` field (while `Bar` passes). This works correctly when expanded by the compiler plugin (although in this example it obviously result in a compilation error due to the property being redeclared).

```swift
public struct DuplicateAll: MemberAttributeMacro {
    public static func expansion(
        of node: AttributeSyntax,
        attachedTo declaration: some DeclGroupSyntax,
        providingAttributesFor member: some DeclSyntaxProtocol,
        in context: some MacroExpansionContext
    ) throws -> [AttributeSyntax] {
        return [AttributeSyntax(attributeName: SimpleTypeIdentifierSyntax(name: .identifier("Duplicate")))]
    }
}

public struct Duplicate: PeerMacro {
    public static func expansion(of node: AttributeSyntax, providingPeersOf declaration: some DeclSyntaxProtocol, in context: some MacroExpansionContext) throws -> [DeclSyntax] {
        if var decl = declaration.as(VariableDeclSyntax.self) {
            decl.attributes = nil
            return [DeclSyntax(decl)]
        }
        return []
    }
}

final class reproTests: XCTestCase {
    func testMacro() {
        assertMacroExpansion(
            """
            @DuplicateAll
            class Foo {
                var bar: Int = 0
            }
            class Bar {
                @Duplicate
                var bar: Int = 0
            }
            """,
            expandedSource: """

            class Foo {
                var bar: Int = 0
                var bar: Int = 0
            }
            class Bar {
                var bar: Int = 0
                var bar: Int = 0
            }
            """,
            macros:  [
                "DuplicateAll": DuplicateAll.self,
                "Duplicate": Duplicate.self,
            ]
        )
    }
}
```

ahoppen:
Tracked in Apple’s issue tracker as rdar://110923449

DougGregor:
`SyntaxProtocol.expand` is really a "full" macro expansion system, and only handles one level of expansion. We could perhaps make it a full expansion system at some point.

liamnichols:
In `509.0.0-swift-5.9-DEVELOPMENT-SNAPSHOT-2023-04-25-b`, i'm currently observing inconsistent behaviour. 

[...]

PR #2767, opened 2024-07-31
Detect circular macro expansion

### LINKED ISSUE #2018: Detect circular macro expansion in `MacroSystem`
The following recursive expansion of a macro causes an infinite loop in `MacroSystem`, which is used by `assertMacroExpansion`. The compiler diagnoses the circular evaluation of the macro. `MacroSystem` should do the same.

```swift
struct InfiniteRecursionMacro: ExpressionMacro {
  static func expansion(of node: some FreestandingMacroExpansionSyntax, in context: some MacroExpansionContext) throws -> ExprSyntax {
    return "#infiniteRecursion"
  }
}

final class MacroSystemTests: XCTestCase {
  func testInfiniteRecursion() {
    assertMacroExpansion("#infiniteRecursion", expandedSource: "", macros: ["infiniteRecursion": InfiniteRecursionMacro.self])
  }
}

### PR DESCRIPTION
**Pull request by AppAppWorks**
fixes #2018

### DIFF
--- a/Sources/SwiftSyntaxMacroExpansion/MacroExpansion.swift
+++ b/Sources/SwiftSyntaxMacroExpansion/MacroExpansion.swift
@@ -63,6 +63,7 @@ enum MacroExpansionError: Error, CustomStringConvertible {
   case noFreestandingMacroRoles(Macro.Type)
   case moreThanOneBodyMacro
   case preambleWithoutBody
+  case recursiveExpansion(any Macro.Type)
 
   var description: String {
     switch self {
@@ -92,6 +93,9 @@ enum MacroExpansionError: Error, CustomStringConvertible {
 
     case .preambleWithoutBody:
       return "preamble macro cannot be applied to a function with no body"
+
+    case .recursiveExpansion(let type):
+      return "recursive expansion of macro '\(type)'"
     }
   }
 }
--- a/Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
+++ b/Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
@@ -667,6 +667,12 @@ private class MacroApplication<Context: MacroExpansionContext>: SyntaxRewriter {
   /// added to top-level 'CodeBlockItemList'.
   var extensions: [CodeBlockItemSyntax] = []
 
+  /// Stores the types of the freestanding macros that are currently expanding.
+  ///
+  /// As macros are expanded by DFS, `expandingFreestandingMacros` always represent the expansion path starting from
+  /// the root macro node to the last macro node currently expanding.
+  var expandingFreestandingMacros: [any Macro.Type] = []
+
   init(
     macroSystem: MacroSystem,
     contextGenerator: @escaping (Syntax) -> Context,
@@ -683,7 +689,7 @@ private class MacroApplication<Context: MacroExpansionContext>: SyntaxRewriter {
   }
 
   override func visitAny(_ node: Syntax) -> Syntax? {
-    if skipVisitAnyHandling.contains(node) {
+    guard !skipVisitAnyHandling.contains(node) else {
       return nil
     }
 
@@ -692,8 +698,10 @@ private class MacroApplication<Context: MacroExpansionContext>: SyntaxRewriter {
     // position are handled by 'visit(_:CodeBlockItemListSyntax)'.
     // Only expression expansions inside other syntax nodes is handled here.
     switch expandExpr(node: node) {
-    case .success(let expanded):
-      return Syntax(visit(expanded))
+    case .success(let expansion):
+      return expansion.withExpandedNode { expandedNode in
+        Syntax(visit(expandedNode))
+      }
     case .failure:
       return Syntax(node)
     case .notAMacro:
@@ -794,9 +802,11 @@ private class MacroApplication<Context: MacroExpansionContext>: SyntaxRewriter {
     func addResult(_ node: CodeBlockItemSyntax) {
       // Expand freestanding macro.
       switch expandCodeBlockItem(node: node) {
-      case .success(let expanded):
-        for item in expanded {
-          addResult(item)
+      case .success(let expansion):
+        expansion.withExpandedNode { expandedNode in
+          for item in expandedNode {
+            addResult(item)
+          }
         }
         return
       case .failure:
@@ -839,9 +849,11 @@ private class MacroApplication<Context: MacroExpansionContext>: SyntaxRewriter {
     func addResult(_ node: MemberBlockItemSyntax) {
       // Expand freestanding macro.
       switch expandMemberDecl(node: node) {
-      case .success(let expanded):
-        for item in expanded {
-          addResult(item)
+      case .success(let expansion):
+        expansion.withExpandedNode { expandedNode in
+          for item in expandedNode {
+            addResult(item)
+          }
         }
         return
       case .failure:
@@ -1205,9 +1217,36 @@ extension MacroApplication {
 // MARK: Freestanding macro expansion
 
 extension MacroApplication {
+  /// Encapsulates an expanded node, the type of the macro from which the node was expanded, and the macro application,
+  /// such that recursive macro expansion can be consistently detected.
+  struct MacroExpansion<ResultType> {
+    private let expandedNode: ResultType
+    private let macro: any Macro.Type
+    private unowned let macroApplication: MacroApplication
+
+    fileprivate init(expandedNode: ResultType, macro: any Macro.Type, macroApplication: MacroApplication) {
+      self.expandedNode = expandedNode
+      self.macro = macro
+      self.macroApplication = macroApplication
+    }
+
+    /// Invokes the given closure with the node resulting from a macro expansion.
+    ///
+    /// This method inserts a pair of push and pop operations immediately around the invocation of `body` to maintain
+    /// an exact stack of expanding freestanding macros to detect recursive macro expansion. Callers should perform any
+    /// further macro expansion on `expanded` only within the scope of `body`.
+    func withExpandedNode<T>(_ body: (_ expandedNode: ResultType) throws -> T) rethrows -> T {
+      macroApplication.expandingFreestandingMacros.append(macro)
+      defer {
+        macroApplication.expandingFreestandingMacros.removeLast()
+      }
+      return try body(expandedNode)
+    }
+  }
+
   enum MacroExpansionResult<ResultType> {
     /// Expansion of the macro succeeded.
-    case success(ResultType)
+    case success(expansion: MacroExpansion<ResultType>)
 
     /// Macro system found the macro to expand but running the expansion threw
     /// an error and thus no expansion result exists.
@@ -1217,18 +1256,37 @@ extension MacroApplication {
     case notAMacro
   }
 
+  /// Expands the given freestanding macro node into a syntax node by invoking the given closure.
+  ///
+  /// Any error thrown by `expandMacro` and circular expansion error will be added to diagnostics.
+  ///
+  /// - Parameters:
+  ///   - node: The freestanding macro node to be expanded.
+  ///   - expandMacro: The closure that expands the given macro type and macro node into a syntax node.
+  ///
+  /// - Returns:
+  /// Returns `.notAMacro` if `node` is `nil` or `node.macroName` isn't registered with any macro type.
+  /// Returns `.failure` if `expandMacro` throws an error or returns `nil`, or recursive expansion is detected.
+  /// Returns `.success` otherwise.
   private func expandFreestandingMacro<ExpandedMacroType: SyntaxProtocol>(
     _ node: (any FreestandingMacroExpansionSyntax)?,
-    expandMacro: (_ macro: Macro.Type, _ node: any FreestandingMacroExpansionSyntax) throws -> ExpandedMacroType?
+    expandMacro: (_ macro: any Macro.Type, _ node: any FreestandingMacroExpansionSyntax) throws -> ExpandedMacroType?
   ) -> MacroExpansionResult<ExpandedMacroType> {
     guard let node,
       let macro = macroSystem.lookup(node.macroName.text)?.type
     else {
       return .notAMacro
     }
+
     do {
+      guard !expandingFreestandingMacros.contains(where: { $0 == macro }) else {
+        // We may think of any ongoing macro expansion as a tree in which macro types being expanded are nodes.
+        // Any macro type being expanded more than once will create a cycle which the compiler as of now doesn't allow.
+        throw MacroExpansionError.recursiveExpansion(macro)
+      }
+
       if let expanded = try expandMacro(macro, node) {
-        return .success(expanded)
+        return .success(expansion: MacroExpansion(expandedNode: expanded, macro: macro, macroApplication: self))
       } else {
         return .failure
       }
--- a/Tests/SwiftSyntaxMacroExpansionTest/ExpressionMacroTests.swift
+++ b/Tests/SwiftSyntaxMacroExpansionTest/ExpressionMacroTests.swift
@@ -37,6 +37,46 @@ fileprivate struct StringifyMacro: ExpressionMacro {
   }
 }
 
+private struct InfiniteRecursionMacro: ExpressionMacro {
+  static func expansion(
+    of node: some FreestandingMacroExpansionSyntax,
+    in context: some MacroExpansionContext
+  ) throws -> ExprSyntax {
+    if let i = node.arguments.first?.expression.as(IntegerLiteralExprSyntax.self)?.representedLiteralValue {
+      return "\(raw: i) + #infiniteRecursion(i: \(raw: i + 1))"
+    } else {
+      return "#nested1"
+    }
+  }
+}
+
+private struct Nested1RecursionMacro: ExpressionMacro {
+  static func expansion(
+    of node: some FreestandingMacroExpansionSyntax,
+    in context: some MacroExpansionContext
+  ) throws -> ExprSyntax {
+    "(#nested2, #nested3, #infiniteRecursion(i: 1), #infiniteRecursion)"
+  }
+}
+
+private struct Nested2RecursionMacro: ExpressionMacro {
+  static func expansion(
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

**Review by ahoppen — COMMENTED** [review]
Wouldn’t it be simpler to just maintain a stack of expanded macros, add a macro’s name to it whenever we are inside its expansion and then pop it again afterwards? And if we try to expand a macro that’s in that list, emit an error?

File: Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift [hunk 1]
```diff
@@ -1226,7 +1259,13 @@ extension MacroApplication {
     else {
       return .notAMacro
     }
+
     do {
+      guard expandingFreestandingMacros.allSatisfy({ $0 != macro }) else {
```
CodaFi, line 1264: This implementation is quadratic in the depth of the stack.

Sources/SwiftSyntaxMacroExpansion/MacroExpansion.swift
```diff
@@ -92,6 +93,9 @@ enum MacroExpansionError: Error, CustomStringConvertible {
 
     case .preambleWithoutBody:
       return "preamble macro cannot be applied to a function with no body"
+
+    case .circularExpansion(let type, let syntax):
+      return "circular expansion detected: '\(syntax)' with macro implementation type '\(type)'"
```
CodaFi, line 98: Note: I think the phrasing here could be improved. Maybe not in this PR.

CodaFi:
Do you consider a tail-recursive macro whose expansion eventually terminates to be bannable? I think we have to be very careful about what counts as non-productive recursion.

CodaFi:
> Wouldn’t it be simpler to just maintain a stack of expanded macros, add a macro’s name to it whenever we are inside its expansion and then pop it again afterwards? 

Now you're approaching the territory of a termination checker. I suggest we take this approach, but rather than query the stack let's use it to install a reasonable depth limit and call it a day. The full scope of efficiently rejecting general recursion in programs is an open area of research 😓

CodaFi:
> a reasonable depth limit

Unfortunately, I'm not sure what this number should be, and it seems neither do our peer compilers.

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
AppAppWorks: I used an array for the stack and a dictionary for memoizing macro cyclicalities in a previous commit, would that approach be better?

AppAppWorks:
> Do you consider a tail-recursive macro whose expansion eventually terminates to be bannable? I think we have to be very careful about what counts as non-productive recursion.

@ahoppen and I have discussed in [#2018](https://github.com/swiftlang/swift-syntax/issues/2018#issuecomment-2259180536) that recursive macro expansion is currently prohibited by the compiler. Notwithstanding, I think determining whether a tail-recursive macro will eventually terminate is akin to the halting problem?

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
AppAppWorks: I've force-pushed my previous approach, please have a look.

CodaFi:
>> That’s because #infiniteRecursion isn’t allowed to expand to #infiniteRecursion again, even if it contains different argument.

I see, well then that is really quite restrictive and I do disagree with it but I defer to Alex on this one.

> I think determining whether a tail-recursive macro will eventually terminate is akin to the halting problem?

Absolutely, but there is a lot of research in restricted forms of termination checking, mostly in service of proof assistants. There, you're absolutely willing to sacrifice a certain class of (often, ill-typed anyways) terms that will happen to halt but that your termination checker can't really prove. There's heuristics like syntactic decrease all the way up to stuff like Andreas Abel's research on termination checking for Agda which reduces inter-procedural analysis to a series of matrix multiplies. It's really neat stuff!

ahoppen:
To be clear, we currently have the restriction that a macro can’t expand to itself in the compiler. And this PR is just implementing the same behavior in MacroSystem, which is used for `assertMacroExpansion`. The goal here is to make the compiler and MacroSystem match, not to change any limitations on recursive macro expansion.

> Wouldn’t it be simpler to just maintain a stack of expanded macros, add a macro’s name to it whenever we are inside its expansion and then pop it again afterwards? And if we try to expand a macro that’s in that list, emit an error?

Did you address this comment yet, @AppAppWorks? I couldn’t find anything related to it in your latest commits. Just checking that we’re not mutually waiting for each other.

AppAppWorks:
> Did you address this comment yet, @AppAppWorks? I couldn’t find anything related to it in your latest commits. Just checking that we’re not mutually waiting for each other.

I've just reverted to a previous commit as well as correcting the sites of stack popping, more precisely we only pop a macro type name from the stack whenever such freestanding macro has been fully expanded.

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
ahoppen: We could make `expandingFreestandingMacros` a `Set` to avoid the quadratic complexity and use `!expandingFreestandingMacros.contains(macro)`

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
```diff
@@ -1226,7 +1244,13 @@ extension MacroApplication {
...
+        throw MacroExpansionError.circularExpansion(macro, node)
+      }
+      expandingFreestandingMacros.append(macro)
+
       if let expanded = try expandMacro(macro, node) {
```
ahoppen, line 1254: Instead of removing macros in a number of places, can’t we just remove `macro` from `expandingFreestandingMacros` after the `if`?

AppAppWorks: > Instead of removing macros in a number of places, can’t we just remove `macro` from `expandingFreestandingMacros` after the `if`?

I've tried this approach before only to find out there could be non-intermediate recursive macro expansion hidden inside `expanded`, e.g. `#macro1 -> #macro2 -> #macro1`. Since `#macro1` expands into `#macro2`, if we pop `#macro1` from the stack right after `expandMacro(macro, node)`, by the time `#macro2` is expanded into `#macro1`, the stack won't contain `#macro1`, resulting in infinite expansion.

So we can safely discard a macro name only when we can be sure the macro expansion has fully completed, i.e. the sites I've placed the removal logic. (I might have missed some other sites though)

> We could make `expandingFreestandingMacros` a `Set` to avoid the quadratic complexity and use `!expandingFreestandingMacros.contains(macro)`

The reason I'm using an `Array` instead of a `Set` here is I need to keep track of the insertion order, such that I can safely discard all macro names indexed on or after `macroCount` because they must be inserted during the expansion.

If asymptotic performance is a concern, I can use a `Set` for membership checking alongside an `Array` for keeping track of insertion order.

ahoppen:
- I would really prefer if the place that adds elements to `expandingFreestandingMacros` is connected to the place that removes them. It's just too easy to forget to remove them otherwise.

AppAppWorks:
> * I would really prefer if the place that adds elements to `expandingFreestandingMacros` is connected to the place that removes them. It's just too easy to forget to remove them otherwise.

True, I'll put some more thought into refactoring.

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
```diff
@@ -1204,10 +1218,33 @@ extension MacroApplication {
 
 // MARK: Freestanding macro expansion
 
+protocol MacroExpansion<ResultType> {
+  associatedtype ResultType
+  func map<T>(_ transform: (ResultType) throws -> T) rethrows -> T
+}
+
```
ahoppen, line 1225: Is this protocol needed. I would have expected that it isn’t because the case in `MacroExpansionResult` could just be the `StackMacroExpansion` with the generic type specialized to `ResultType`.

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
```diff
@@ -1204,10 +1218,33 @@ extension MacroApplication {
...
+    let expandedNode: ResultType
+    unowned let macroApplication: MacroApplication
+
+    init(expandedNode: ResultType, macro: any Macro.Type, macroApplication: MacroApplication) {
+      self.expandedNode = expandedNode
+      self.macroApplication = macroApplication
+      macroApplication.expandingFreestandingMacros.append(macro)
+    }
+
+    func map<T>(_ transform: (ResultType) throws -> T) rethrows -> T {
```
ahoppen, line 1237: I think `map` is a little misleading because I’d expect `map` to be side-effect free. It’s more a `takeResult` that doesn’t have a return value. With that name, it would also be good to have an assertion and ensure that  we don’t take the result twice (and thus pop from `macroApplication.expandingFreestandingMacros` twice).

Also, I think it would be worth adding a doc comment why we need to handle the result in a closure, ie. explain about how we detect circular macro expansions.

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
```diff
@@ -687,13 +690,20 @@ private class MacroApplication<Context: MacroExpansionContext>: SyntaxRewriter {
       return nil
     }
 
+    let macroCount = expandingFreestandingMacros.count
+    defer {
+      expandingFreestandingMacros.removeLast(expandingFreestandingMacros.count - macroCount)
+    }
+
```
ahoppen, line 697: Is this still needed?

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
```diff
@@ -1204,10 +1218,33 @@ extension MacroApplication {
...
+    unowned let macroApplication: MacroApplication
+
+    init(expandedNode: ResultType, macro: any Macro.Type, macroApplication: MacroApplication) {
+      self.expandedNode = expandedNode
+      self.macroApplication = macroApplication
+      macroApplication.expandingFreestandingMacros.append(macro)
+    }
+
+    func map<T>(_ transform: (ResultType) throws -> T) rethrows -> T {
+      defer {
```
ahoppen, line 1238: Would it be possible to run `macroApplication.expandingFreestandingMacros.append(macro)` right before `transform` instead of in `init`? That way, we’d guarantee that the append is always balanced with a `remove` and they can never get out-of-sync if somebody forgets to calls `map`.

ahoppen (commented):
FWIW the compiler also emits an error if an attached macro is recursively expanded in the same role. https://github.com/swiftlang/swift/blob/454dfb9a0b19917a64e2ecda31e714203a05231e/lib/Sema/TypeCheckMacros.cpp#L1364-L1369

I think it would be good to implement that in MacroSystem as well but that doesn’t have to be this PR.

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
AppAppWorks: The reason why I used a protocol here instead of a concrete type is I want to make sure the only blessed way to access the expanded node is through the closure.

ahoppen: You could just make everything but `map` private, right?

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
AppAppWorks: > I think `map` is a little misleading because I’d expect `map` to be side-effect free. It’s more a `takeResult` that doesn’t have a return value. 

Actually there's at least one call site that returns a value ;) I'm typing this comment on a phone and have no easy access to the source, but I vaguely remember that's 'expandExpr'? 

Btw should we rename the method to 'consume'?

> With that name, it would also be good to have an assertion and ensure that  we don’t take the result twice (and thus pop from `macroApplication.expandingFreestandingMacros` twice).

I had this in my mind and wanted to implement it with lifetime specifiers. I forgot to fallback to assertions after I noticed 'consuming' is not a thing in earlier Swift versions.

> Also, I think it would be worth adding a doc comment why we need to handle the result in a closure, ie. explain about how we detect circular macro expansions.

True!

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
AppAppWorks: True, I was carried away by the thought of storing one less field in the struct...

Sources/SwiftSyntaxMacroExpansion/MacroExpansion.swift
AppAppWorks: > FWIW the compiler’s error message is `recursive expansion of macro 'MacroName'`. I think we should copy that.

I think including both the macro name and type in the error message is more helpful for debugging?

AppAppWorks:
> FWIW the compiler also emits an error if an attached macro is recursively expanded in the same role. https://github.com/swiftlang/swift/blob/454dfb9a0b19917a64e2ecda31e714203a05231e/lib/Sema/TypeCheckMacros.cpp#L1364-L1369
> 
> 
> 
> I think it would be good to implement that in MacroSystem as well but that doesn’t have to be this PR.

I wrote this in #2018,

> As reported in #1805, even though SyntaxProtocol.expand() isn't yet a "full" expansion system, the expansion behaviours of different subtypes of AttachedMacro vary as some of them will expand macros recursively to some extent.

Should the upcoming PR also solve this incoherent behaviour across different types of attached macros?

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
AppAppWorks: > You could just make everything but `map` private, right?

Despite this, call sites in this file could still access the private fields?

Maybe I've gone too far to entertain my paranoia for discouraging unintended usages...

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
AppAppWorks: My bad, gotta remove this vestige

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
AppAppWorks: > I do think this would read more intuitive as `!expandingFreestandingMacros.contains(macro)` though.

That means we'd be storing strings instead of Macro.Type in the stack?

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
ahoppen: > I had this in my mind and wanted to implement it with lifetime specifiers. I forgot to fallback to assertions after I noticed 'consuming' is not a thing in earlier Swift versions.

> Sans the support of move-only value types, we need to change StackMacroExpansion from a struct to a class too.

Yeah, we can’t use those yet and even then, I think it might be tricky to keep the move-only type alive until all the nested macro expansion are performed. The current design is probably best for now.

ahoppen:
> Should the upcoming PR also solve this incoherent behaviour across different types of attached macros?

Yes, let’s do a separate PR for that.

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
ahoppen: No, `private` can only be accessed from the type itself, `fileprivate` can be accessed from anywhere in the file.

AppAppWorks: I need to brush up my knowledge of Swift access modifiers...

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
ahoppen: > That means we'd be storing strings instead of Macro.Type in the stack?

Oh, I forgot that metatypes don’t conform to `Comparable`. I’d still do `!expandingFreestandingMacros.contains({ $0 == macro })

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
```diff
@@ -1205,9 +1214,39 @@ extension MacroApplication {
...
+    /// Invokes the given closure with the node resulting from a macro expansion.
+    ///
+    /// - Precondition: This method has never been called for this instance.
+    ///
+    /// This method inserts a pair of push and pop operations immediately around the closure invocation to maintain an
+    /// exact stack of expanding freestanding macros to detect circluar macro expansion. Callers should only call this
+    /// method once and perform all further expansions on `expanded` only within the scope of `body`.
+    func consume<T>(_ body: (_ expanded: ResultType) throws -> T) rethrows -> T {
+      precondition(!isConsumed)
+      isConsumed = true
```
ahoppen, line 1238: Sorry, should have noticed this earlier but if we have the `append` and `removeLast` paired up like this, I don’t it’s not as fatal anymore if a `MacroExpansion` is consumed twice – the state would still be left intact. So, I guess we don’t need the `isConsumed` check anymore, which removes one potential cause of crashes – the parser tries to be very fault-tolerant and never crash. And then, this could also become a `struct` again.

Sources/SwiftSyntaxMacroExpansion/MacroExpansion.swift
ahoppen: I think usually the mating between macro name and macro type is quite obvious, so I don’t think we need to include both. But I don’t have very strong opinions here.

ahoppen (approved):
Looking great

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
Could you run swift-format? https://github.com/swiftlang/swift-syntax/blob/main/CONTRIBUTING.md#formatting

AppAppWorks:
Formatted, please test again.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #2772, opened 2024-08-03
Minor fixes for SwiftIfConfig

DougGregor:
While integrating this further in the compiler, found a few issues to be fixed here:
* `IfConfigRegionState.evaluate()` needs to return `syntaxErrorsAllowed`, too
* Version parsing needs to deal with weird malformed versions like `...`
* `#if _compiler_version(>=6.0)` is supported syntax, so handle it


DougGregor:
@swift-ci please test

PR #2773, opened 2024-08-04
Make the "are syntax errors allowed?" condition independent of evaluation

DougGregor:
The compiler needs to be more careful about when it actually evaluates an #if condition, because canImport can have side effects. Make it easier to answer the question "are syntax errors allowed?" without having to evaluate the condition every time.

### DIFF
--- a/Sources/SwiftIfConfig/IfConfigEvaluation.swift
+++ b/Sources/SwiftIfConfig/IfConfigEvaluation.swift
@@ -11,6 +11,7 @@
 //===----------------------------------------------------------------------===//
 
 import SwiftDiagnostics
+import SwiftOperators
 import SwiftSyntax
 
 /// Evaluate the condition of an `#if`.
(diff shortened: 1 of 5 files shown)


DougGregor:
@swift-ci please test

Sources/SwiftIfConfig/IfConfigEvaluation.swift
```diff
@@ -433,20 +434,77 @@ func evaluateIfConfig(
...
 extension IfConfigClauseSyntax {
-  /// Determine whether this condition is "syntaxErrorsAllowed".
-  func syntaxErrorsAllowed(
-    configuration: some BuildConfiguration
+  /// Fold the operators within an #if condition, turning sequence expressions
+  /// involving the various allowed operators (&&, ||, !) into well-structured
+  /// binary operators.
+  public static func foldOperators(
+    _ condition: some ExprSyntaxProtocol
+  ) -> (folded: ExprSyntax, diagnostics: [Diagnostic]) {
```
ahoppen, line 442: Did you intentionally make this method public? I don’t think it provides sufficient value to increase the API surface area of swift-syntax.

Sources/SwiftIfConfig/IfConfigEvaluation.swift
```diff
@@ -433,20 +434,77 @@ func evaluateIfConfig(
...
+    let foldedCondition = OperatorTable.logicalOperators.foldAll(condition) { error in
+      foldingDiagnostics.append(contentsOf: error.asDiagnostics(at: condition))
+    }.cast(ExprSyntax.self)
+    return (folded: foldedCondition, diagnostics: foldingDiagnostics)
+  }
+
+  /// Determine whether the given expression, when used as the condition in
+  /// an inactive `#if` clause, implies that syntax errors are permitted within
+  /// that region.
+  public static func syntaxErrorsAllowed(
```
ahoppen, line 453: Similar here: Does this method have to be public? I think internal would suffice.

Sources/SwiftIfConfig/IfConfigEvaluation.swift
DougGregor: https://github.com/swiftlang/swift-syntax/pull/2806 makes it internal

Sources/SwiftIfConfig/IfConfigEvaluation.swift
DougGregor: https://github.com/swiftlang/swift-syntax/pull/2806 makes it internal

PR #2774, opened 2024-08-05
Various improvements and tweaks to bring SwiftIfConfig in line with the compiler

DougGregor:
Implement a number of improvements and minor wording tweaks to bring SwiftIfConfig in line with the expectations of the compiler, as determined by the compiler's test suite. This includes:

* Add warning for `#if` conditions like `(os(iOS) || os(tvOS)) && (arch(i386) || arch(x86_64))` that should be replaced by `targetEnvironment(simulator)`.
* Add warning and remapping for the rename of the `macabi` target environment to `macCatalyst`.
* Downgrade the error about `_compiler_version`'s second version component being non-`*` to a warning.
* Downgrade error about bad `_endian` platform condition to a warning.
* Properly process expressions as import paths, replacing my horrible "grab the string and split on `.`" hack.
* Ensure that we diagnose compound names where they aren't permitted.
* Improve diagnostics for uses of operators other than `&&`, `||`, or `!`.
* Avoid evaluating `canImport` when it won't affect the result, because `canImport` in the compiler has side effects.

### DIFF
--- a/Sources/SwiftIfConfig/IfConfigError.swift
+++ b/Sources/SwiftIfConfig/IfConfigError.swift
@@ -31,6 +31,7 @@ enum IfConfigError: Error, CustomStringConvertible {
   case ignoredTrailingComponents(version: VersionTuple, syntax: ExprSyntax)
   case integerLiteralCondition(syntax: ExprSyntax, replacement: Bool)
   case likelySimulatorPlatform(syntax: ExprSyntax)
+  case endiannessDoesNotMatch(syntax: ExprSyntax, argument: String)
 
   var description: String {
     switch self {
@@ -81,6 +82,9 @@ enum IfConfigError: Error, CustomStringConvertible {
     case .likelySimulatorPlatform:
       return
         "platform condition appears to be testing for simulator environment; use 'targetEnvironment(simulator)' instead"
+
+    case .endiannessDoesNotMatch:
+      return "unknown endianness for build configuration '_endian' (must be 'big' or 'little')"
     }
   }
 
@@ -100,7 +104,8 @@ enum IfConfigError: Error, CustomStringConvertible {
       .canImportTwoParameters(syntax: let syntax),
       .ignoredTrailingComponents(version: _, syntax: let syntax),
       .integerLiteralCondition(syntax: let syntax, replacement: _),
-      .likelySimulatorPlatform(syntax: let syntax):
+      .likelySimulatorPlatform(syntax: let syntax),
+      .endiannessDoesNotMatch(syntax: let syntax, argument: _):
       return Syntax(syntax)
 
     case .unsupportedVersionOperator(name: _, operator: let op):
@@ -118,7 +123,8 @@ extension IfConfigError: DiagnosticMessage {
 
   var severity: SwiftDiagnostics.DiagnosticSeverity {
     switch self {
-    case .compilerVersionSecondComponentNotWildcard, .ignoredTrailingComponents, .likelySimulatorPlatform:
+    case .compilerVersionSecondComponentNotWildcard, .ignoredTrailingComponents,
+      .likelySimulatorPlatform, .endiannessDoesNotMatch:
       return .warning
     default: return .error
     }
--- a/Sources/SwiftIfConfig/IfConfigError.swift
+++ b/Sources/SwiftIfConfig/IfConfigError.swift
@@ -32,6 +32,7 @@ enum IfConfigError: Error, CustomStringConvertible {
   case integerLiteralCondition(syntax: ExprSyntax, replacement: Bool)
   case likelySimulatorPlatform(syntax: ExprSyntax)
   case endiannessDoesNotMatch(syntax: ExprSyntax, argument: String)
+  case macabiIsMacCatalyst(syntax: ExprSyntax)
 
   var description: String {
     switch self {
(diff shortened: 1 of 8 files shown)


DougGregor:
@swift-ci please test

CodaFi:
> canImport in the compiler has side effects.

:sadblob:

Sources/SwiftIfConfig/SyntaxLiteralUtils.swift
```diff
@@ -36,13 +36,23 @@ extension LabeledExprListSyntax {
...
+
+extension DeclReferenceExprSyntax {
+  /// If this declaration reference is a simple identifier, return that
+  /// string.
+  var simpleIdentifier: String? {
+    guard argumentNames == nil else {
+      return nil
+    }
+
+    /// FIXME: Make this an Identifier so we handle escaping properly.
```
ahoppen, line 55: Any reason to not make it an `Identifier` now?

Sources/SwiftIfConfig/IfConfigEvaluation.swift
```diff
@@ -96,71 +97,107 @@ func evaluateIfConfig(
...
-      case (true, "||"):
-        return (active: true, syntaxErrorsAllowed: lhssyntaxErrorsAllowed, diagnostics: lhsDiagnostics)
-      case (false, "&&"):
-        return (active: false, syntaxErrorsAllowed: lhssyntaxErrorsAllowed, diagnostics: lhsDiagnostics)
-      default:
-        break
-      }
+    // Short-circuit evaluation if we know the answer. We still recurse into
+    // the right-hand side, but with a dummy configuration that won't have
+    // side effects, so we only get validation-related errors.
```
ahoppen, line 151: > We still recurse into the right-hand side

Isn’t this in contradiction to the following below?

> If we are supposed to short-circuit and the left-hand side of this operator with inactive &&, stop now: we shouldn't evaluate the right-hand side at all.

Sources/SwiftIfConfig/IfConfigEvaluation.swift
```diff
@@ -96,71 +97,107 @@ func evaluateIfConfig(
...
+    switch (lhsActive, op.operator.text) {
+    case (true, "||"): shortCircuitResult = true
+    case (false, "&&"): shortCircuitResult = false
+    default: shortCircuitResult = nil
+    }
+
+    // If we are supposed to short-circuit and the left-hand side of this
+    // operator with inactive &&, stop now: we shouldn't evaluate the right-
+    // hand side at all.
+    if let isActive = shortCircuitResult, lhsSyntaxErrorsAllowed {
```
ahoppen, line 162: Why don’t we want to short-circuit if the left side doesn’t allow syntax errors?

Sources/SwiftIfConfig/IfConfigEvaluation.swift
```diff
@@ -30,7 +30,8 @@ import SwiftSyntax
 ///   diagnose syntax errors in blocks where the check fails.
 func evaluateIfConfig(
   condition: ExprSyntax,
-  configuration: some BuildConfiguration
+  configuration: some BuildConfiguration,
+  outermostCondition: Bool = true
```
ahoppen, line 34: Do we need to pass this around? Couldn’t you check `condition.keyPathInParent == \IfConfigClauseSyntax.condition`?

Sources/SwiftIfConfig/IfConfigError.swift
```diff
@@ -12,6 +12,7 @@
 
 import SwiftDiagnostics
 import SwiftSyntax
+import SwiftSyntaxBuilder
 
 /// Describes the kinds of errors that can occur when processing #if conditions.
 enum IfConfigError: Error, CustomStringConvertible {
```
ahoppen, line 18: Since `IfConfigError` can also represent warnings now, I think a better name would be `IfConfigDiagnostic`.

Sources/SwiftIfConfig/IfConfigEvaluation.swift
DougGregor: The semantics of the compiler were a bit surprising here, and the code and comments don't match now. I've updated the comments to describe what's going on in https://github.com/swiftlang/swift-syntax/pull/2806 makes it internal

Sources/SwiftIfConfig/IfConfigEvaluation.swift
DougGregor: It's covered in the comments of https://github.com/swiftlang/swift-syntax/pull/2806 now, but the short answer is that we aren't supposed to trigger any side effects or any diagnostics for ill-formed `#if` conditions if we already know the result of evaluation and the left-hand condition allows syntax errors

Sources/SwiftIfConfig/IfConfigEvaluation.swift
DougGregor: It's more involved than checking the immediate parent, because we pass this through parentheses and prefix `!`. However, you're right that we can get this information from the syntax tree itself without passing it down as state. I've done that in https://github.com/swiftlang/swift-syntax/pull/2806

Sources/SwiftIfConfig/IfConfigError.swift
DougGregor: Good idea! Done in https://github.com/swiftlang/swift-syntax/pull/2806

Sources/SwiftIfConfig/SyntaxLiteralUtils.swift
DougGregor: Nope, we can do it now: https://github.com/swiftlang/swift-syntax/pull/2806

PR #2775, opened 2024-08-05
Support body macros that don’t return any statements

ahoppen:
Previously, when a body macro returned no statements, we would return an empty expansion and not add the braces in `collapse`. Fix that.

### DIFF
--- a/Tests/SwiftSyntaxMacroExpansionTest/BodyMacroTests.swift
+++ b/Tests/SwiftSyntaxMacroExpansionTest/BodyMacroTests.swift
@@ -125,4 +125,34 @@ final class BodyMacroTests: XCTestCase {
       indentationWidth: indentationWidth
     )
   }
+
+  func testEmptyBodyMacro() {
+    struct EmptyBodyMacro: BodyMacro {
+      public static var formatMode: FormatMode { .disabled }
+
+      public static func expansion(
+        of node: AttributeSyntax,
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Tests/SwiftSyntaxMacroExpansionTest/BodyMacroTests.swift
```diff
@@ -125,4 +125,30 @@ final class BodyMacroTests: XCTestCase {
...
+      }
+    }
+
+    assertMacroExpansion(
+      """
+      @EmptyBody
+      func f() {}
+      """,
+      expandedSource: """
+        func f() {}
```
rintaro, line 148: Is expected behavior is replacing the body in the original code to `{}`? If so, could you add some statements in the original source code?

ahoppen: It’s expected to leave the body untouched. This matches the behavior of accessors

https://github.com/swiftlang/swift-syntax/blob/28794d2dc574b4d97aed235c43d852be2957d36e/Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift#L329

I updated the test.

ahoppen:
@swift-ci Please test Windows

rintaro (approved):

PR #2707, opened 2024-06-27
[android] add android build support

hyp:
this is needed to build swift-corelibs-foundation for android, using the 'package' branch that is based on new swift-foundation

PR #2777, opened 2024-08-06
[Android] Use the new overlay and Bionic module from Swift 6

finagolfin:
The new overlay was merged into Swift 6 in swiftlang/swift#74758.

This supercedes the earlier #2707 with more complete support.

@hyp and @bnbarham, please review.

### DIFF
--- a/Sources/SwiftLibraryPluginProvider/LibraryPluginProvider.swift
+++ b/Sources/SwiftLibraryPluginProvider/LibraryPluginProvider.swift
@@ -22,6 +22,8 @@ private import Darwin
 private import Glibc
 #elseif canImport(Musl)
 private import Musl
+#elseif canImport(Android)
+private import Android
 #endif
 #else
 import SwiftSyntaxMacros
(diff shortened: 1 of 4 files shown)


Sources/SwiftLibraryPluginProvider/LibraryPluginProvider.swift
```diff
@@ -137,7 +139,7 @@ private func _loadLibrary(_ path: String) throws -> UnsafeMutableRawPointer {
 #else
 private func _loadLibrary(_ path: String) throws -> UnsafeMutableRawPointer {
   guard let dlHandle = dlopen(path, RTLD_LAZY | RTLD_LOCAL) else {
-    throw LibraryPluginError(message: "loader error: \(String(cString: dlerror()))")
+    throw LibraryPluginError(message: "loader error: \(String(cString: dlerror()!))")
```
compnerd, line 142: Does this work with glibc? I thought that it was annotated with non null in some cases (at least I ran into one with foundation)

compnerd (approved):

compnerd:
@swift-ci please test

Sources/SwiftLibraryPluginProvider/LibraryPluginProvider.swift
finagolfin: Yep, worked fine in Fedora 40 x86_64, where I tested it before submitting, which uses a recent glibc 2.39-17 from earlier this year. Bionic simply added the `_Nullable` annotation to the output in the last year, so it _has_ to be unwrapped for Android here, but unless glibc somehow can guarantee this is `_Nonnull`, this force unwrap should work for both.

finagolfin:
@ahoppen, ready for merge.

finagolfin:
Thanks for the quick review, @compnerd.

PR #2780, opened 2024-08-06
Fix bad module dep version in 5.10

luispadron:
This was using the wrong versions for `apple_support`. We **don't** need a new tag here as we'll be patching the [bazel-central-registry](https://github.com/bazelbuild/bazel-central-registry/pull/2552) but fixing in case there is another 5.10 patch

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -4,5 +4,5 @@ module(
     compatibility_level = 1,
 )
 
-bazel_dep(name = "rules_swift", version = "1.16.0", repo_name = "build_bazel_rules_swift")
-bazel_dep(name = "apple_support", version = "1.18.0", max_compatibility_level = 2, repo_name = "build_bazel_apple_support")
+bazel_dep(name = "rules_swift", version = "1.18.0", max_compatibility_level = 2, repo_name = "build_bazel_rules_swift")
+bazel_dep(name = "apple_support", version = "1.11.1", repo_name = "build_bazel_apple_support")


keith:
@swift-ci please test

keith (approved):

bnbarham (approved):

bnbarham:
@luispadron / @keith is there anyway that we can check these earlier? 5.10 was released a while back now, did something else change that needed these to be patched?

keith:
we technically didn't need the change here unless there was going to be other natural 5.10 releases

bnbarham:
> we technically didn't need the change here unless there was going to be other natural 5.10 releases

I was more talking about the other change rather than this one. And doesn't hurt to have this in just in case we do.

keith:
sorry which one do you mean? the 6.x one? FWIW something did change with these rules, repos have to declare compatibility on the new version of rules_swift since there were breaking changes, so it wasn't that swift-syntax changed, but if a repo that consumes it wants to use a newer version of rules_swift it would have failed w/o this change. These can technically be patched in the bazel-central-registry repo instead of pushed here, but on main it's nice to land to reduce the patching that has to happen there, and make it easier for folks to test dev versions

bnbarham:
> sorry which one do you mean?

https://github.com/swiftlang/swift-syntax/pull/2766

> so it wasn't that swift-syntax changed, but if a repo that consumes it wants to use a newer version of rules_swift it would have failed w/o this change

Ah I see, good to know. Thanks.

PR #2781, opened 2024-08-07
Fix build issue when using a Swift 5.8 compiler

ahoppen:
We were unable to infer the base of the subscript with a Swift 5.8 compiler.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #2782, opened 2024-08-07
[SwiftLexicalLookup][GSoC] Add switch, generic parameter and function scopes.

MAJKFL:
This PR introduces `switch case`, generic parameter/primary associated types and function scopes.
- `WithGenericParametersOrAssociatedTypesScopeSyntax` and `GenericParameterOrAssociatedTypeScopeSyntax` are used to properly handle generic parameters/primary associated types by providing an alternative route for the lookup method before propagating it to parent scope.
- `switch` now introduces names in it's `case` bodies
- Function parameters are now passed into to the body
- Includes some changes suggested in #2748


ahoppen:
Would it be possible to do the changes that address my review comments from https://github.com/swiftlang/swift-syntax/pull/2748 in a separate PR again? It just makes it easier to review.

MAJKFL:
>Would it be possible to do the changes that address my review comments from https://github.com/swiftlang/swift-syntax/pull/2748 in a separate PR again?

Ok, I think this is a great idea. I've opened #2789.

Tests/SwiftLexicalLookupTest/NameLookupTests.swift
```diff
@@ -835,4 +835,137 @@ final class testNameLookup: XCTestCase {
...
+  }
+
+  func testSimpleGenericParameterScope() {
+    assertLexicalNameLookup(
+      source: """
+        class A<1️⃣T1, 2️⃣T2> {
+          let 7️⃣x: 3️⃣T1 = v
+          let y: 4️⃣T2 = v
+
+          class B<5️⃣T1> {
```
DougGregor, line 869: I am curious what the compiler (and this implementation) would do for the second `T2` in `class B<T1, T2: T2>`.

DougGregor (approved):
I have a few comments, but in general this looks great. There are a few constructs that probably still need to be covered here, but they can be in a follow-up PR:
* subscripts
* type aliases
* where clauses and inheritance clauses of types/extensions to make sure they're seeing the right things

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test Windows

Sources/SwiftLexicalLookup/Scopes/GenericParameterScopeSyntax.swift
```diff
@@ -0,0 +1,91 @@
...
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
+/// Scope that introduces generic parameter names and directs
+/// futher lookup to it's `WithGenericParametersScopeSyntax`
+/// parent scope's parent scope (i.e. on return, bypasses names
+/// introduced by it's parent).
```
ahoppen, line 18:
```suggestion
/// introduced by its parent).
```

Sources/SwiftLexicalLookup/Scopes/GenericParameterScopeSyntax.swift
```diff
@@ -0,0 +1,91 @@
...
+  /// and effectively bypass names already looked up before.
+  private func lookupBypassingParentResults(
+    _ identifier: Identifier?,
+    at lookUpPosition: AbsolutePosition,
+    with config: LookupConfig
+  ) -> [LookupResult] {
+    guard let parentScope else { return [] }
+
+    if let parentScope = Syntax(parentScope).asProtocol(SyntaxProtocol.self)
+      as? WithGenericParametersScopeSyntax
```
ahoppen, line 84: Out of curiosity: Do we ever expect a case where the parent of a `GenericParameterScopeSyntax` is not a `WithGenericParametersScopeSyntax`?

Sources/SwiftLexicalLookup/Scopes/GenericParameterScopeSyntax.swift
```diff
@@ -0,0 +1,91 @@
...
+/// futher lookup to it's `WithGenericParametersScopeSyntax`
+/// parent scope's parent scope (i.e. on return, bypasses names
+/// introduced by it's parent).
+@_spi(Experimental) public protocol GenericParameterScopeSyntax: ScopeSyntax {}
+
+@_spi(Experimental) extension GenericParameterScopeSyntax {
+  /// Returns names matching lookup and bypasses
+  /// `WithGenericParametersScopeSyntax` parent scope in futher lookup.
+  ///
+  /// example:
```
ahoppen, line 25: Let’s take advantage of markdown formatting. We do this in other places already. I think there are a few more places in this PR that use `example:`.

```suggestion
  /// ### Example
```

Sources/SwiftLexicalLookup/Scopes/ScopeImplementations.swift
```diff
@@ -356,7 +374,79 @@ import SwiftSyntax
...
+    }
+
+    return results + defaultLookupImplementation(identifier, at: lookUpPosition, with: config)
+  }
+}
+
+@_spi(Experimental) extension GenericParameterClauseSyntax: GenericParameterScopeSyntax {
+  /// Generic parameter names introduced by this clause.
+  @_spi(Experimental) public var introducedNames: [LookupName] {
+    parameters.children(viewMode: .sourceAccurate).flatMap { child in
```
ahoppen, line 439: I think the `.fixedUp` view mode would make more sense here. It probably doesn’t make a huge difference but conceptually, the `.fixedUp` tree contains the missing tokens that the parser introduced to recover from syntax errors while `.sourceAccurate` contains unexpected tokens as well. Since we’re looking at the meaning the user intended here, I think `.fixedUp` makes more sense.

Sources/SwiftLexicalLookup/Scopes/ScopeImplementations.swift
```diff
@@ -356,7 +374,79 @@ import SwiftSyntax
...
+      LookupName.getNames(from: child, accessibleAfter: child.endPosition)
```
ahoppen, line 440: I think generic parameters need to be visible for later generic parameters. For example, consider the following

```swift
class X {}

func foo<X: AnyObject, Y: 1️⃣X>(_: X, _: Y) {}

```

Performing lookup at 1️⃣ in the compiler returns the `X` declared in `myMethod` and it fails to compile. I think with the current implementation in `SwiftLexicalLookup`, it would pick the `class X {}` declaration and thus probably compile.

ahoppen (commented):
Nice. I’m really curious how close we are to being feature-equivalent with the C++ name lookup and how many more edge cases we might hit.

Sources/SwiftLexicalLookup/Scopes/ScopeImplementations.swift
```diff
@@ -330,10 +348,10 @@ import SwiftSyntax
...
 }
 
-@_spi(Experimental) extension ActorDeclSyntax: TypeScopeSyntax {}
-@_spi(Experimental) extension ClassDeclSyntax: TypeScopeSyntax {}
-@_spi(Experimental) extension StructDeclSyntax: TypeScopeSyntax {}
-@_spi(Experimental) extension EnumDeclSyntax: TypeScopeSyntax {}
+@_spi(Experimental) extension ActorDeclSyntax: TypeScopeSyntax, WithGenericParametersScopeSyntax {}
+@_spi(Experimental) extension ClassDeclSyntax: TypeScopeSyntax, WithGenericParametersScopeSyntax {}
+@_spi(Experimental) extension StructDeclSyntax: TypeScopeSyntax, WithGenericParametersScopeSyntax {}
+@_spi(Experimental) extension EnumDeclSyntax: TypeScopeSyntax, WithGenericParametersScopeSyntax {}
```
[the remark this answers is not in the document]
MAJKFL: It wasn't a big change, so I added `typealias` to this PR.

Sources/SwiftLexicalLookup/Scopes/ScopeImplementations.swift
```diff
@@ -356,7 +374,79 @@ import SwiftSyntax
...
+@_spi(Experimental) extension GenericParameterClauseSyntax: GenericParameterScopeSyntax {
+  /// Generic parameter names introduced by this clause.
+  @_spi(Experimental) public var introducedNames: [LookupName] {
+    parameters.children(viewMode: .sourceAccurate).flatMap { child in
+      LookupName.getNames(from: child, accessibleAfter: child.endPosition)
+    }
+  }
+}
+
+@_spi(Experimental) extension FunctionDeclSyntax: WithGenericParametersScopeSyntax {
```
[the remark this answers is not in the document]
MAJKFL: You're right, I forgot about `subscript`. As with `typealias`, it wasn't a big change, so I added to this PR.

Tests/SwiftLexicalLookupTest/NameLookupTests.swift
MAJKFL: At least when it comes to this implementation, the result [would be empty](https://github.com/swiftlang/swift-syntax/blob/57183fc995200d6ed3a6cd8755a9c761883064fd/Tests/SwiftLexicalLookupTest/NameLookupTests.swift#L897).

Sources/SwiftLexicalLookup/Scopes/GenericParameterScopeSyntax.swift
MAJKFL: I don't think we should. Those protocols only make sense when coupled together.

Sources/SwiftLexicalLookup/Scopes/ScopeImplementations.swift
MAJKFL: They are actually now visible to subsequent generic parameters and it's exercised by [this test case](https://github.com/swiftlang/swift-syntax/blob/5e8c6d404a44e780d26f2040d16b7dbbd7de4760/Tests/SwiftLexicalLookupTest/NameLookupTests.swift#L891-L904).

ahoppen: Thanks!

Sources/SwiftLexicalLookup/Scopes/GenericParameterScopeSyntax.swift
ahoppen: OK, thanks for clarifying

ahoppen (approved):

DougGregor:
@swift-ci please test

Tests/SwiftLexicalLookupTest/NameLookupTests.swift
DougGregor: So, it seems that there's no ordering amongst the various generic parameters. Given this:
```swift
class A<X> {
  class B<T2: T1, T1> { }
}  
```

unqualified lookup is finding `T1` and the type checker is rejecting it:

```
1 | class A<X> {
2 |   class B<T1, T2: T1> { }
  |                   `- error: type 'T2' constrained to non-protocol, non-class type 'T1'
3 | }  
4 | 
```

DougGregor: FWIW, here's a case where it's also semantically well-formed to refer to generic parameters out-of-order:

```
struct X<C: Collection<T>, T> { }
```

DougGregor:
@swift-ci please test Windows

PR #2792, opened 2024-08-08
Use root node for detached node lookup in location retrieval

### LINKED ISSUE #2791: Source location retrieving doesn't work in body macro expansion assertion
### Description

The [location](https://github.com/swiftlang/swift-syntax/blob/28794d2dc574b4d97aed235c43d852be2957d36e/Sources/SwiftSyntaxMacroExpansion/BasicMacroExpansionContext.swift#L200) method of `BasicMacroExpansionContext` returns `nil` when a body macro is expanded using `assertMacroExpansion`. This makes it difficult to write tests for macros that depend on non-nil location value.

### Steps to Reproduce

Consider the following macro:
```swift
struct SourceLocationMacro: BodyMacro {
  public static var formatMode: FormatMode { .disabled }

  public static func expansion(
    of node: AttributeSyntax,
    providingBodyFor declaration: some DeclSyntaxProtocol & WithOptionalCodeBlockSyntax,
  ) throws -> [CodeBlockItemSyntax] {
    guard let statements = declaration.body?.statements else {
      return []
    }
    let body =
      if let location = context.location(of: statements, at: .afterLeadingTrivia, filePathMode: .filePath) {
        CodeBlockItemListSyntax {
          "#sourceLocation(file: \(location.file), line: \(location.line))"
          statements
          "#sourceLocation()"
        }
      } else {
        statements
      }
    return body.map(\.self)
  }
}
```

Assertion of it produces an unexpected expanded source code:
```swift
assertMacroExpansion(
  """
  @SourceLocationMacro
  func f() {
    let x: Int = 1
  }
  """,
  expandedSource:
    """
    func f() {
      #sourceLocation(file: "test.swift", line: 3)
      let x: Int = 1
      #sourceLocation()
    }
    """,
  macros: ["SourceLocationMacro": SourceLocationMacro.self],
)

❌ failed - Macro expansion did not produce the expected expanded source

Actual expanded source:

func f() {
  let x: Int = 1
}
```

mateusrodriguesxyz:
Fixes #2791

Maybe just `let nodeInOriginalTree = sharedState.detachedNodes[Syntax(node.root)]` is sufficient? I tried and no existing tests failed.

### DIFF
--- a/Sources/SwiftSyntaxMacroExpansion/BasicMacroExpansionContext.swift
+++ b/Sources/SwiftSyntaxMacroExpansion/BasicMacroExpansionContext.swift
@@ -210,7 +210,7 @@ extension BasicMacroExpansionContext: MacroExpansionContext {
       // The syntax node came from the source file itself.
       rootSourceFile = directRootSourceFile
       offsetAdjustment = .zero
-    } else if let nodeInOriginalTree = sharedState.detachedNodes[Syntax(node)] {
+    } else if let nodeInOriginalTree = sharedState.detachedNodes[node.root] {
       // The syntax node came from a disconnected root, so adjust for that.
       rootSourceFile = nodeInOriginalTree.root.as(SourceFileSyntax.self)
       offsetAdjustment = SourceLength(utf8Length: nodeInOriginalTree.position.utf8Offset)
--- a/Tests/SwiftSyntaxMacroExpansionTest/BodyMacroTests.swift
+++ b/Tests/SwiftSyntaxMacroExpansionTest/BodyMacroTests.swift
@@ -155,4 +155,50 @@ final class BodyMacroTests: XCTestCase {
       macros: ["EmptyBody": EmptyBodyMacro.self]
     )
   }
+
+  func testBodyNodeLocationFromContext() {
+    struct SourceLocationMacro: BodyMacro {
+      public static var formatMode: FormatMode { .disabled }
+
+      public static func expansion(
+        of node: AttributeSyntax,
+        providingBodyFor declaration: some DeclSyntaxProtocol & WithOptionalCodeBlockSyntax,
+        in context: some MacroExpansionContext
+      ) throws -> [CodeBlockItemSyntax] {
+        guard let statements = declaration.body?.statements else {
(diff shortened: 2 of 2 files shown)


Sources/SwiftSyntaxMacroExpansion/BasicMacroExpansionContext.swift
```diff
@@ -210,7 +210,9 @@ extension BasicMacroExpansionContext: MacroExpansionContext {
       // The syntax node came from the source file itself.
       rootSourceFile = directRootSourceFile
       offsetAdjustment = .zero
-    } else if let nodeInOriginalTree = sharedState.detachedNodes[Syntax(node)] {
+    } else if let nodeInOriginalTree = sharedState.detachedNodes[Syntax(node)]
+      ?? sharedState.detachedNodes[Syntax(node.root)]
+    {
```
ahoppen, line 215: Good catch. I think this should just check `node.root` though, similar to `location(for:anchoredAt:fileName:)`

```suggestion
    } else if let nodeInOriginalTree = sharedState.detachedNodes[node.root]
    {
```

Tests/SwiftSyntaxMacroExpansionTest/BodyMacroTests.swift
```diff
@@ -125,4 +125,51 @@ final class BodyMacroTests: XCTestCase {
       indentationWidth: indentationWidth
     )
   }
+
+  func testBodyNodeLocationFromContext() {
+
```
ahoppen, line 130: Superfluous newline

ahoppen (commented):
Thank you @mateusrodriguesxyz 🙏🏽

ahoppen (approved):

ahoppen:
@swift-ci Please test

ahoppen:
Thanks for fixing this @mateusrodriguesxyz! 🙏🏽

AdamCmiel:
@mateusrodriguesxyz still seeing this erroneous behavior for attributes of a functionDeclSyntax, I'll write up a test and see if I can't amend the context

PR #2793, opened 2024-08-09
Revert "Replace expression in `ClosureCaptureSyntax` with initializer clause"

rintaro:
Reverts swiftlang/swift-syntax#2763

```swift
func test(expr: some ExprSyntaxProtocol) {
  _ = ClosureCaptureSyntax(expression: expr)
}
```

became ambiguous by this.


ahoppen:
CC @MAJKFL Could you check why Rintaro’s example is ambiguous and how we can disambiguate it, eg. by adding `@_disfavoredOverload` to some declaration?

MAJKFL:
It's ambiguous because we have two deprecated initializers: one [with `assignToken`](https://github.com/swiftlang/swift-syntax/blob/f8075dac5bc64ab9495c8f8f23d5fe764b3bedc8/Sources/SwiftSyntax/SwiftSyntaxCompatibility.swift#L178-L217) and the other [with `equal`](https://github.com/swiftlang/swift-syntax/blob/f8075dac5bc64ab9495c8f8f23d5fe764b3bedc8/Sources/SwiftSyntax/SwiftSyntaxCompatibility.swift#L297-L335) and both are marked with `@_disfavoredOverload`. I think the easiest way to fix this would be to remove `@_disfavoredOverload` from one of them (I think from the one with `equal` parameter, as it's the most recent one). Should I open a PR?

ahoppen:
> I think the easiest way to fix this would be to remove `@_disfavoredOverload` from one of them (I think from the one with equal parameter, as it's the most recent one). Should I open a PR?

I think we need the `@ _disfavoredOverload` for some other ambiguity resolution. Let’s just remove the deprecated initializer with `assignToken`. That one has been deprecated since 509 (ie. for a year), so I don’t expect that many people still use it. If you could open a PR for that, that would be great!

MAJKFL:
Ok, I've created #2795 without the initializer and sorry for causing the trouble.

PR #2794, opened 2024-08-09
[SwiftParser] Improve diagnostics for misspelled keywords

### LINKED ISSUE #2198: [SR-8510] 'guard' is easy to misspell as 'gaurd', results in large number of follow-on errors, and could have a fixit.
|                  |                 |
   |------------------|-----------------|
   |Previous ID       | SR-8510      |
   |Radar             | rdar://problem/32419789         |
   |Original Reporter | @huonw      |
   |Type              | Bug    |
   |Status            | In Progress        |
   |Resolution        |     |
        
<details>
  <summary>Environment</summary>

Master at 2018-08-09

</details>

   <details>
  <summary>Additional Detail from JIRA</summary>

   |                  |                 |
   |------------------|-----------------|
   |Votes             | 1         |
   |Component/s       | Compiler    |
   |Labels            | Bug, DiagnosticsQoI, Parser, StarterBug        |
   |Assignee          | mkita (JIRA)      |
   |Priority          | Medium      |

   md5: 83854736e7d37441acce9e16b297a24a

  </details>

**Issue Description:**

In the following code, the guard statement is misspelled, and it results in 6 errors, plus notes and warnings! It'd be neat if the compiler detected when the "consecutive statements" error starts with the identifier `gaurd`, and assumed that `guard` was what was meant (along with a fixit so the user swaps it). I think that only doing this when that error occurs is important, because someone could legitimately have `var gaurd = 1; gaurd = 2; print(gaurd)`, and none of those should change to `guard`.

```swift
func foo() -> Int? { return 0 }
func bar() {
    gaurd let x = foo() else { return }
    print(x)
}
```

```
guard.swift:3:10: error: consecutive statements on a line must be separated by ';'
    gaurd let x = foo() else { return }
         ^
         ;
guard.swift:3:24: error: consecutive statements on a line must be separated by ';'
    gaurd let x = foo() else { return }
                       ^
                       ;
guard.swift:3:25: error: expected expression
    gaurd let x = foo() else { return }
                        ^
guard.swift:3:5: error: use of unresolved identifier 'gaurd'
    gaurd let x = foo() else { return }
    ^~~~~
guard.swift:3:30: error: closure expression is unused
    gaurd let x = foo() else { return }
                             ^
guard.swift:3:30: note: did you mean to use a 'do' statement?
    gaurd let x = foo() else { return }
                             ^
                             do 
guard.swift:4:11: warning: expression implicitly coerced from 'Int?' to 'Any'
    print(x)
          ^
guard.swift:4:11: note: provide a default value to avoid this warning
    print(x)
          ^
            ?? <#default value#>
guard.swift:4:11: note: force-unwrap the value to avoid this warning
    print(x)
          ^
           !
guard.swift:4:11: note: explicitly cast to 'Any' with 'as Any' to silence this warning
    print(x)
          ^
            as Any
```

Possible new version:

```
guard.swift:3:10: error: consecutive statements on a line must be separated by ';'; did you misspell 'guard'?
    gaurd let x = foo() else { return }
    ^~~~~
    guard
```

AppAppWorks:
fixes #2198 #2180

This is preliminary work on emitting better diagnostics for misspelled keywords. More discussions would be needed to ascertain the scope of misspelling corrections.

### DIFF
--- a/Sources/SwiftParser/TokenSpecSet.swift
+++ b/Sources/SwiftParser/TokenSpecSet.swift
@@ -26,6 +26,20 @@ protocol TokenSpecSet: CaseIterable {
   init?(lexeme: Lexer.Lexeme, experimentalFeatures: Parser.ExperimentalFeatures)
 }
 
+protocol MisspelledTokenSpecSet: TokenSpecSet {
+  associatedtype CorrectSpecSet: TokenSpecSet
+
+  typealias FuzzyMatchSpecSet = EitherTokenSpecSet<Self, CorrectSpecSet>
+
+  var correctSpecSet: CorrectSpecSet { get }
+}
+
+extension MisspelledTokenSpecSet {
+  var spec: TokenSpec {
+    TokenSpec(.identifier, recoveryPrecedence: correctSpecSet.spec.recoveryPrecedence)
+  }
+}
+
 /// A way to combine two token spec sets into an aggregate token spec set.
 enum EitherTokenSpecSet<LHS: TokenSpecSet, RHS: TokenSpecSet>: TokenSpecSet {
   case lhs(LHS)
@@ -58,6 +72,17 @@ enum EitherTokenSpecSet<LHS: TokenSpecSet, RHS: TokenSpecSet>: TokenSpecSet {
   }
 }
 
+extension EitherTokenSpecSet where LHS: MisspelledTokenSpecSet, RHS == LHS.CorrectSpecSet {
+  var correctSpecSet: RHS {
+    switch self {
+    case .lhs(let misspelled):
+      return misspelled.correctSpecSet
+    case .rhs(let correct):
+      return correct
+    }
+  }
+}
+
 // MARK: - Subsets
 
 enum AccessorModifier: TokenSpecSet {
@@ -85,6 +110,42 @@ enum AccessorModifier: TokenSpecSet {
     case .borrowing: return .keyword(.borrowing)
     case .mutating: return .keyword(.mutating)
     case .nonmutating: return .keyword(.nonmutating)
+    }
+  }
+}
+
+enum MisspelledAccessorModifier: MisspelledTokenSpecSet {
+  case consuming
+  case borrowing
+  case nonmutating
+
+  var correctSpecSet: AccessorModifier {
+    switch self {
+    case .consuming: return .consuming
+    case .borrowing: return .borrowing
+    case .nonmutating: return .nonmutating
+    }
+  }
+
+  init?(lexeme: Lexer.Lexeme, experimentalFeatures: Parser.ExperimentalFeatures) {
+    let text = lexeme.tokenText
+    switch text.count {
+    case 6:
+      switch text {
+      case "borrow": self = .borrowing
+      default: return nil
+      }
+    case 7:
+      switch text {
+      case "consume": self = .consuming
+      default: return nil
+      }
+    case 11:
+      switch text {
+      case "nonMutating": self = .nonmutating
+      default: return nil
+      }
+    default: return nil
     }
   }
 }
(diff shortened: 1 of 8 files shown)


Sources/SwiftParser/TokenSpecSet.swift
```diff
@@ -339,6 +428,80 @@ enum PureDeclarationKeyword: TokenSpecSet {
...
+      switch text {
+      case "deInit": self = .deinit
+      case "object": self = .class
+      default: return nil
+      }
+    case 8:
+      switch text {
+      case "function": self = .func
+      default: return nil
+      }
```
AppAppWorks, line 473: I think we could use code generation in the future for creating matching statements like these.

ahoppen (commented):
Great changes! One thought after skimming over the PR:

I think it would be great if we could make misspelling correction even easier to use. What do you think of the following:
- Every keyword has a list of possible misspellings + maybe an implicit list of single-character typos (the latter is probably more involved and could be a separate PR)
- `canRecover` always takes misspellings into account and automatically generates a `TokenConsumptionHandle` that generates a missing token if it discovers a misspelling.
- `expect` takes a parameter to decide whether it should take misspellings into account.

mateusrodriguesxyz:
FWIW I've implemented a few months ago a general solution for keyword misspelling using levenshtein distance]:
https://github.com/mateusrodriguesxyz/swift-syntax/tree/keywords-correction

[MisspelledKeywordsTest](https://github.com/mateusrodriguesxyz/swift-syntax/blob/keywords-correction/Tests/SwiftParserTest/_MisspelledKeywordsTest.swift)

AppAppWorks:
> * Every keyword has a list of possible misspellings

I think we could store the mapping of keyword misspellings in some resource files and make use of code generation to create boilerplate code (there would be a lot of them). Resource files would be friendly for crowdsourcing too.

>  maybe an implicit list of single-character typos (the latter is probably more involved and could be a separate PR)

@mateusrodriguesxyz has done some great work on [correcting keywords with single-character typos based on Levenshtein Distance](https://github.com/mateusrodriguesxyz/swift-syntax/tree/keywords-correction), but I'm not sure if we should precompute typo permutations statically as computation for Levenshtein Distance takes O(mn) time.

> * `canRecover` always takes misspellings into account and automatically generates a `TokenConsumptionHandle` that generates a missing token if it discovers a misspelling.

Would it lead to performance regression as the search space for the parser might explode?

AppAppWorks:
> FWIW I've implemented a few months ago a general solution for keyword misspelling using levenshtein distance: https://github.com/mateusrodriguesxyz/swift-syntax/tree/keywords-correction
> 
> [MisspelledKeywordsTest](https://github.com/mateusrodriguesxyz/swift-syntax/blob/keywords-correction/Tests/SwiftParserTest/_MisspelledKeywordsTest.swift)

Great work! It'll need some time for me to digest it.

mateusrodriguesxyz:
> Great work! It'll take some time for me to digest it.

Thanks! Most of the relevance code is in [TokenSpec](https://github.com/mateusrodriguesxyz/swift-syntax/blob/bf14dbc71a96378a48fed2a3e7de33322372be02/Sources/SwiftParser/TokenSpec.swift#L164). I do quite a few check before even trying to find the correct keyword to avoid false diagnostics and unnecessary distance computation.

AppAppWorks:
With the latest commit, I've created a kitchen sink to facilitate crowdsourcing of keyword "false friends" in popular programming languages and other general misspellings that won't be captured by one-character Levenshtein Distance permutations.

ahoppen (commented):
That’s a great list. I don’t think it needs to be 100% complete but just having the infrastructure to add more typo Fix-Its is amazing because it makes it very easy to extend. 

I think the next big step is to integrate this into the main parsing infrastructure so typo-correction is taken care of automatically in a variety of places, as I described in https://github.com/swiftlang/swift-syntax/pull/2794#pullrequestreview-2230576956.

PR #2798, opened 2024-08-10
Replace self key path as function usage

### LINKED ISSUE #2796: KeyPath `.self` cannot be resolved in earlier Swift versions
### Description

#2792 introduced a bug in BodyMacroTests.swift:181,

```swift
return body.map(\.self)
```
where `\.self` cannot be resolved by the compiler of some earlier Swift versions.

### Steps to Reproduce

_No response_

mateusrodriguesxyz:
Fixes #2796

I used `map(\.self)` but that's only possible since [Subtyping for keypath literals as functions](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0416-keypath-function-subtyping.md), which landed on Swift 6.0. Thanks @AppAppWorks for catching this.

### DIFF
--- a/Tests/SwiftSyntaxMacroExpansionTest/BodyMacroTests.swift
+++ b/Tests/SwiftSyntaxMacroExpansionTest/BodyMacroTests.swift
@@ -178,7 +178,7 @@ final class BodyMacroTests: XCTestCase {
           } else {
             statements
           }
-        return body.map(\.self)
+        return Array(body)
       }
     }
 


Tests/SwiftSyntaxMacroExpansionTest/BodyMacroTests.swift
```diff
@@ -178,7 +178,7 @@ final class BodyMacroTests: XCTestCase {
           } else {
             statements
           }
-        return body.map(\.self)
+        return body.map({ $0.self })
```
ahoppen, line 181: Do you think `Array(body)` would convey the intention better than mapping to `self`?

mateusrodriguesxyz: I hadn't thought of that, it's better indeed. I will update it.

ahoppen (approved):
Thank you!

ahoppen:
@swift-ci Please test

PR #2799, opened 2024-08-11
[CodeGeneration] Improve ergonomics for initializers of raw syntax nodes

AppAppWorks:
This PR aims to improve developer ergonomics when using the designated initializers of raw syntax nodes. Simply put, it's through achieving feature parity with the non-raw side.

### Prefer generic types over type erasure types

Currently the designated initializer of `RawXXXSyntax` always uses type-erased types instead of generic types as parameter types, leading to much clunkiness.

```swift
public struct RawPatternExprSyntax: RawExprSyntaxNodeProtocol {
  public init(
    _ unexpectedBeforePattern: RawUnexpectedNodesSyntax? = nil, 
    pattern: RawPatternSyntax, // type erasure
    _ unexpectedAfterPattern: RawUnexpectedNodesSyntax? = nil, 
    arena: __shared SyntaxArena
  )
}

let pattern = RawIdentifierPatternSyntax(
  identifier: identifier,
  arena: self.arena
)
return RawExprSyntax(RawPatternExprSyntax(pattern: RawPatternSyntax(pattern), arena: self.arena))
```

The initializer body, however, only invokes `pattern.raw`, therefore we could just replace `pattern: RawPatternSyntax` with `pattern: some RawPatternSyntaxNodeProtocol` transparently,

```swift
public struct RawPatternExprSyntax: RawExprSyntaxNodeProtocol {
  public init(
    _ unexpectedBeforePattern: RawUnexpectedNodesSyntax? = nil, 
    pattern: some RawPatternSyntaxNodeProtocol, // generic type
    _ unexpectedAfterPattern: RawUnexpectedNodesSyntax? = nil, 
    arena: __shared SyntaxArena
  )
}

let pattern = RawIdentifierPatternSyntax(
  identifier: identifier,
  arena: self.arena
)
// clunky type-erasure removed
```

Actually it's what `PatternExprSyntax.init` has been doing,
```swift
public init(
  leadingTrivia: Trivia? = nil,
  _ unexpectedBeforePattern: UnexpectedNodesSyntax? = nil,
  pattern: some PatternSyntaxProtocol,
  _ unexpectedAfterPattern: UnexpectedNodesSyntax? = nil,
  trailingTrivia: Trivia? = nil
)
```

### Generate convenience generic initializers for choices enum

e.g.
```swift
public struct RawCodeBlockItemSyntax: RawSyntaxNodeProtocol {
  public enum Item: RawSyntaxNodeProtocol {
    case `decl`(RawDeclSyntax)
    case `stmt`(RawStmtSyntax)
    case `expr`(RawExprSyntax)

    public init(_ other: some RawDeclSyntaxNodeProtocol) {
      self = .decl(RawDeclSyntax(other))
    }
    
    public init(_ other: some RawStmtSyntaxNodeProtocol) {
      self = .stmt(RawStmtSyntax(other))
    }
    
    public init(_ other: some RawExprSyntaxNodeProtocol) {
      self = .expr(RawExprSyntax(other))
    }
```

So call sites might get rid of type erasure as follows,
```swift
// requires type erasure
.expr(RawExprSyntax(RawMissingExprSyntax(arena: self.arena)))

// supports generic type 
.init(RawMissingExprSyntax(arena: self.arena))
```

### Remove type erasure in call sites

This PR has removed unnecessary uses of type-erased types in many call sites thanks to the new initializers.

### DIFF
--- a/CodeGeneration/Sources/generate-swift-syntax/templates/swiftparser/LayoutNodesParsableFile.swift
+++ b/CodeGeneration/Sources/generate-swift-syntax/templates/swiftparser/LayoutNodesParsableFile.swift
@@ -65,13 +65,11 @@ let layoutNodesParsableFile = SourceFileSyntax(leadingTrivia: copyrightHeader) {
           // The missing item is not necessary to be a declaration,
           // which is just a placeholder here
           return RawCodeBlockItemSyntax(
-            item: .decl(
-              RawDeclSyntax(
-                RawMissingDeclSyntax(
-                  attributes: self.emptyCollection(RawAttributeListSyntax.self),
-                  modifiers: self.emptyCollection(RawDeclModifierListSyntax.self),
-                  arena: self.arena
-                )
+            item: .init(
+              RawMissingDeclSyntax(
+                attributes: self.emptyCollection(RawAttributeListSyntax.self),
+                modifiers: self.emptyCollection(RawDeclModifierListSyntax.self),
+                arena: self.arena
               )
             ),
             semicolon: nil,
--- a/CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/RawSyntaxNodesFile.swift
+++ b/CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/RawSyntaxNodesFile.swift
@@ -102,6 +102,26 @@ func rawSyntaxNodesFile(nodesStartingWith: [Character]) -> SourceFileSyntax {
 
               StmtSyntax("return nil")
             }
+
+            for (swiftName, kind) in choices {
+              if let choiceNode = SYNTAX_NODE_MAP[kind], choiceNode.kind.isBase {
+                DeclSyntax(
+                  """
+                  public init(_ other: some \(choiceNode.kind.rawProtocolType)) {
+                    self = .\(swiftName)(\(choiceNode.kind.rawType)(other))
+                  }
+                  """
+                )
+              } else {
+                DeclSyntax(
+                  """
+                  public init(_ other: \(kind.rawType)) {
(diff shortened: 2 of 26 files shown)


ahoppen (commented):
I would like to ensure that this doesn’t result in a performance regression in the parser. Could you build `swift-parser-cli` in release mode and run `swift-parser-cli performance-test` on some decently-sized Swift project a couple of times before and after this change? I would mostly check if the instruction count increased. https://github.com/Ranchero-Software/NetNewsWire is a good sample project.

AppAppWorks:
> I would like to ensure that this doesn’t result in a performance regression in the parser. Could you build `swift-parser-cli` in release mode and run `swift-parser-cli performance-test` on some decently-sized Swift project a couple of times before and after this change? I would mostly check if the instruction count increased. https://github.com/Ranchero-Software/NetNewsWire is a good sample project.

True, I was aware this PR might lead to more code bloat due to more generic monomorphization but I didn't know how to measure the impact!

AppAppWorks:
```
(base) resonancel@dhcp-128-189-71-173 Release % ./swift-parser-cli performance-test --directory /Users/resonancel/Documents/Developer_Neo/NetNewsWire --iterations 1000
Time:         98.1874829530716ms
Instructions: 1299785961.564
(base) resonancel@dhcp-128-189-71-173 Release % ./swift-parser-cli performance-test --directory /Users/resonancel/Documents/Developer_Neo/NetNewsWire --iterations 1000
Time:         97.24006307125092ms
Instructions: 1298639072.112
```

There was almost 1ms in performance regression after this change, I guess I have to scrap this PR...

ahoppen:
Out of curiosity: What did your measurements show?

PR #2800, opened 2024-08-11
remove lazy modifier from computed property produced by accessor macro

Gotyanov:
https://forums.swift.org/t/accessor-macros-and-lazy-properties/73869


ahoppen:
Thanks for the PR @Gotyanov.

Accessor macros on lazy properties should be invalid because they can remove the laziness of the property, which has can cause surprising behavior. We could probably improve the error message in the compiler.

PR #2804, opened 2024-08-14
[SwiftParser] Improve developer ergonomics for type erasure with `RawXXX<Kind>Syntax`

AppAppWorks:
### Overview

This PR aims to improve developer ergonomics when performing type erasure with `RawXXX<Kind>Syntax` nodes. In short, a convenience property `raw<Kind>Expr` will be added to all conforming types of `Raw<Kind>SyntaxNodeProtocol` where `<Kind>` is one of `Decl`, `Expr`, `Stmt` or `Pattern`.

For example, 
```swift
public protocol RawDeclSyntaxNodeProtocol: RawSyntaxNodeProtocol {
  var rawDeclSyntax: RawDeclSyntax {
    get
  }
}

public extension RawDeclSyntaxNodeProtocol {
  @inline(__always)
  var rawDeclSyntax: RawDeclSyntax {
    RawDeclSyntax(self)
  }
}
```

Instead of wrapping a subtype with the constructor of its type-erased base type,
```swift
RawDeclSyntax(RawMissingDeclSyntax(arena: arena))
```
we may now use the convenience property `rawDeclSyntax`.
```swift
RawMissingDeclSyntax(arena: arena).rawDeclSyntax
```

### Rationale

This PR seems to be at odds with the established idiom of directly wrapping a subtype with the constructor of its type-erased base type for upcasting. Despite this, I'd like to argue with the following points,

#### Existing usage of `.raw`
In the codebase, it's not uncommon to see `.raw` be used instead of `RawSyntax.init` when it comes to type-erasure even outside generated code. 

#### Readability
Maybe it's only my preference, but I'd contend that the current idiom has somewhat increased the noise-to-information ratio as `RawExprSyntax` shouldn't be the focus here but it definitely has caught my attention (or distraction one might consider?),
```swift
RawExprSyntax(
  RawTryExprSyntax(
    tryKeyword: tryKeyword,
    questionOrExclamationMark: mark,
    expression: expression,
    arena: self.arena
  )
)
```
Arguably the following usage would better emphasize `RawTryExprSyntax` than the one above,
```swift
RawTryExprSyntax(
  tryKeyword: tryKeyword,
  questionOrExclamationMark: mark,
  expression: expression,
  arena: self.arena
).rawExprSyntax
```
In highly nested code, the use of `.rawExprSyntax` would also help control nesting levels.

#### Composability
Appending/trimming `.rawExprSyntax` would be easier than wrapping/unwrapping a `RawExprSyntax(...)` when one would like to opt in/out of type erasure. I'd also argue that `someOptionalExpr.map(\.rawExprSyntax)` would be less unwieldy than `someOptionalExpr.map(RawExprSyntax.init)`.

### Performance Regression
By always inlining the convenience properties, there has been no observable performance regression in terms of instruction count after the change. The binary size of `swift-parser-cli` has increased by a tiny margin, possibly due to the new protocol requirements.

### Alternative Approaches
In a prior PR, I'd replaced the occurrences of `Raw<Kind>Syntax` with `some Raw<Kind>SyntaxNodeProtocol` in function signatures and got rid of the use of type erasure in most call sites. This had resulted in a performance regression of ~0.1% increase in instruction count.

### DIFF
--- a/CodeGeneration/Sources/SyntaxSupport/SyntaxNodeKind.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/SyntaxNodeKind.swift
@@ -388,6 +388,11 @@ public enum SyntaxNodeKind: String, CaseIterable {
     return "Raw\(syntaxType)"
   }
 
+  /// The name of this node at the `RawSyntax` level as an variable name.
+  public var rawTypeAsVarName: TypeSyntax {
+    "raw\(syntaxType)"
+  }
+
   /// For base nodes, the name of the corresponding raw protocol to which all the
   /// concrete raw nodes that have this base kind, conform.
   public var rawProtocolType: TypeSyntax {
--- a/CodeGeneration/Sources/generate-swift-syntax/templates/swiftparser/LayoutNodesParsableFile.swift
+++ b/CodeGeneration/Sources/generate-swift-syntax/templates/swiftparser/LayoutNodesParsableFile.swift
@@ -65,14 +65,12 @@ let layoutNodesParsableFile = SourceFileSyntax(leadingTrivia: copyrightHeader) {
           // The missing item is not necessary to be a declaration,
           // which is just a placeholder here
           return RawCodeBlockItemSyntax(
-            item: .decl(
-              RawDeclSyntax(
-                RawMissingDeclSyntax(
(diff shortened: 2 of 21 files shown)


ahoppen:
I would like to see the performance measurements of https://github.com/swiftlang/swift-syntax/pull/2799 before considering this option.

PR #2805, opened 2024-08-15
[SwiftParser] Improve ergonomics for initializers of `RawXXXSyntax`

AppAppWorks:
This PR aims to improve developer ergonomics when using the designated initializers of raw syntax nodes. Simply put, it's through achieving feature parity with the non-raw side.

### Prefer generic types over type erasure types

Currently the designated initializer of `RawXXXSyntax` always uses type-erased types instead of generic types as parameter types, leading to much clunkiness, e.g.

```swift
public struct RawPatternExprSyntax: RawExprSyntaxNodeProtocol {
  public init(
    _ unexpectedBeforePattern: RawUnexpectedNodesSyntax? = nil, 
    pattern: RawPatternSyntax, // type erasure
    _ unexpectedAfterPattern: RawUnexpectedNodesSyntax? = nil, 
    arena: __shared SyntaxArena
  )
}

let pattern = RawIdentifierPatternSyntax(
  identifier: identifier,
  arena: self.arena
)
return RawExprSyntax(RawPatternExprSyntax(pattern: RawPatternSyntax(pattern), arena: self.arena))
```

The initializer body, however, only invokes `pattern.raw`, therefore we could just replace `pattern: RawPatternSyntax` with `pattern: some RawPatternSyntaxNodeProtocol` transparently,

```swift
public struct RawPatternExprSyntax: RawExprSyntaxNodeProtocol {
  public init(
    _ unexpectedBeforePattern: RawUnexpectedNodesSyntax? = nil, 
    pattern: some RawPatternSyntaxNodeProtocol, // generic type
    _ unexpectedAfterPattern: RawUnexpectedNodesSyntax? = nil, 
    arena: __shared SyntaxArena
  )
}

let pattern = RawIdentifierPatternSyntax(
  identifier: identifier,
  arena: self.arena
)
// clunky type-erasure removed
```

Actually it's what `PatternExprSyntax.init` has been doing,
```swift
public init(
  leadingTrivia: Trivia? = nil,
  _ unexpectedBeforePattern: UnexpectedNodesSyntax? = nil,
  pattern: some PatternSyntaxProtocol,
  _ unexpectedAfterPattern: UnexpectedNodesSyntax? = nil,
  trailingTrivia: Trivia? = nil
)
```

### Generate convenience generic initializers for choices enum

e.g.
```swift
public struct RawCodeBlockItemSyntax: RawSyntaxNodeProtocol {
  public enum Item: RawSyntaxNodeProtocol {
    case `decl`(RawDeclSyntax)
    case `stmt`(RawStmtSyntax)
    case `expr`(RawExprSyntax)

    public init(_ other: some RawDeclSyntaxNodeProtocol) {
      self = .decl(RawDeclSyntax(other))
    }
    
    public init(_ other: some RawStmtSyntaxNodeProtocol) {
      self = .stmt(RawStmtSyntax(other))
    }
    
    public init(_ other: some RawExprSyntaxNodeProtocol) {
      self = .expr(RawExprSyntax(other))
    }
```

So call sites might get rid of type erasure as follows,
```swift
// requires type erasure
.expr(RawExprSyntax(RawMissingExprSyntax(arena: self.arena)))

// supports generic type 
.init(RawMissingExprSyntax(arena: self.arena))
```

### Remove type erasure in call sites

This PR has removed unnecessary uses of type-erased types in many call sites thanks to the new initializers.


CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/RawSyntaxNodesFile.swift
```diff
@@ -270,13 +290,23 @@ func rawSyntaxNodesFile(nodesStartingWith: [Character]) -> SourceFileSyntax {
...
 fileprivate extension Child {
   var rawParameterType: TypeSyntax {
-    let paramType: TypeSyntax
-    if case ChildKind.nodeChoices = kind {
-      paramType = syntaxChoicesType
+    var paramType: TypeSyntax
+    if !kind.isNodeChoicesEmpty {
+      paramType = "\(syntaxChoicesType)"
+    } else if hasBaseType && !isOptional {
+      paramType = "some \(syntaxNodeKind.rawProtocolType)"
```
ahoppen, line 297: Should we also implement this for optional nodes with a base type using `(some XXXX)?`?

CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/RawSyntaxNodesFile.swift
```diff
@@ -270,13 +290,23 @@ func rawSyntaxNodesFile(nodesStartingWith: [Character]) -> SourceFileSyntax {
...
     } else {
       paramType = syntaxNodeKind.rawType
     }
 
-    return buildableType.optionalWrapped(type: paramType)
+    if isOptional {
+      if paramType.is(SomeOrAnyTypeSyntax.self) {
```
ahoppen, line 303: Why is this special case needed?

Sources/SwiftSyntax/generated/raw/RawSyntaxNodesAB.swift
```diff
@@ -42,6 +42,14 @@ public struct RawAccessorBlockSyntax: RawSyntaxNodeProtocol {
       }
       return nil
     }
+    
+    public init(_ other: RawAccessorDeclListSyntax) {
+      self = .accessors(other)
+    }
```
ahoppen, line 48: Why are these initializers needed? Couldn’t we just continue using the case names?

ahoppen (commented):
Thank you. That does simplify the construction of raw syntax nodes.

I am not personally a huge fan of using `.init` to create objects. It’s easy to write but I think it hurts readability along the way because you don’t know which object is being constructed. Could you change those to the actual type names or stick with the enum case name if we don’t actually need to go through the initializer?

CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/RawSyntaxNodesFile.swift
AppAppWorks: If we replace `Raw<Kind>Syntax?` with `(some Raw<Kind>SyntaxNodeProtocol)?`, call sites will no longer be able to just pass `nil` to this parameter because we'll need to specialize the generic type.

A workaround for this is what `XXX<Kind>Syntax` are doing,  e.g.
```swift
type: (some TypeSyntaxProtocol)? = TypeSyntax?.none,
```
I didn't know whether it'd be desirable to change all parameters like this into optional parameters for `RawXXX<Kind>Syntax`, so I've kept them as-is.

Sources/SwiftSyntax/generated/raw/RawSyntaxNodesAB.swift
AppAppWorks: True, I'll change the code generation logic so only enum cases with the payload type `Raw<Kind>Syntax` will get a convenience `init`.

CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/RawSyntaxNodesFile.swift
ahoppen: I see. I think requiring the explicit `nil` for raw syntax makes sense because it prompts you to think whether you should parse that part of the node. Could you add this explanation as a comment?

AppAppWorks:
>Could you change those to the actual type names or stick with the enum case name if we don’t actually need to go through the initializer?

I've reverted most changes back to their original states, however for the convenience initializers I've implemented another idea, e.g.
```swift
/// convenience init for `decl`
public init(decl: some RawDeclSyntaxNodeProtocol) {
  self = .decl(RawDeclSyntax(decl))
}
``` 
And the call sites are more readable now?
```swift
item: .init(expr: RawMissingExprSyntax(arena: self.arena)),
```

Sources/SwiftSyntax/generated/raw/RawSyntaxNodesTUVWXYZ.swift
```diff
@@ -2545,6 +2545,11 @@ public struct RawYieldStmtSyntax: RawStmtSyntaxNodeProtocol {
       }
       return nil
     }
+    
+    /// convenience init for `single`
```
ahoppen, line 2549: I’m not sure this doc comment adds much. I’d say we can remove it.

Sources/SwiftParser/Expressions.swift
```diff
@@ -1793,7 +1790,7 @@ extension Parser {
           } while keepGoing != nil && self.hasProgressed(&loopProgress)
         }
 
-        parameterClause = .simpleInput(RawClosureShorthandParameterListSyntax(elements: params, arena: self.arena))
+        parameterClause = .init(RawClosureShorthandParameterListSyntax(elements: params, arena: self.arena))
```
ahoppen, line 1793: Should this be `.init(simpleInput:)` or continue being `.simpleInput`?

Sources/SwiftParser/Expressions.swift
```diff
@@ -2137,11 +2134,11 @@ extension Parser {
     let elseBody: RawIfExprSyntax.ElseBody?
     if elseKeyword != nil {
       if self.at(.keyword(.if)) {
-        elseBody = .ifExpr(
+        elseBody = .init(
```
ahoppen, line 2137: Should this continue being `ifExpr`?

Sources/SwiftParser/Expressions.swift
```diff
@@ -2137,11 +2134,11 @@ extension Parser {
...
           self.parseIfExpression(ifHandle: .constant(.keyword(.if)))
         )
       } else {
-        elseBody = .codeBlock(self.parseCodeBlock(introducer: ifKeyword))
+        elseBody = .init(self.parseCodeBlock(introducer: ifKeyword))
```
ahoppen, line 2141: Same here

Sources/SwiftParser/Expressions.swift
```diff
@@ -1765,7 +1762,7 @@ extension Parser {
         let params = self.parseParameterClause(RawClosureParameterClauseSyntax.self) { parser in
           parser.parseClosureParameter()
         }
-        parameterClause = .parameterClause(params)
+        parameterClause = .init(params)
```
ahoppen, line 1765: Should this continue being `.parameterClause`?

ahoppen (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #2809, opened 2024-08-17
Add support for parsing and emitting diagnostics for #if defined(...)

j-f1:
SwiftIfConfig now supports parsing `defined(FOO)` and interpreting it as just `FOO` for the purposes of active/inactive evaluation. All uses of `defined()` will produce an error with a fix-it that removes the `defined()` call and replaces it with its argument.

This produces a much clearer error message than the current version of Swift. It also un-suppresses syntax errors from the contents of the `#if` block.

```swift
#if defined(FOO)
// before: error: Unexpected platform condition (expected 'os', 'arch', or 'swift')
// after:  error: compilation conditions in Swift are always boolean and do not need to be checked for existence with 'defined()'
//         fix-it: remove 'defined()'
```


j-f1:
@swift-ci please test

j-f1:
@swift-ci please test

ahoppen (approved):
Thanks. Looks good to me

PR #2810, opened 2024-08-19
Add a warning for TARGET_OS_* compilation conditions

j-f1:
This helps people avoid silent mistakes when migrating from Objective-C:

```swift
#if TARGET_OS_IOS
//  ^ warning: 'TARGET_OS_*' preprocessor macros are not available in Swift; use 'os(iOS)' instead
//  ^ fix-it: replace with 'os(iOS)'
print("I will never be executed, even on iOS!")
#endif
```

### DIFF
--- a/Sources/SwiftIfConfig/IfConfigEvaluation.swift
+++ b/Sources/SwiftIfConfig/IfConfigEvaluation.swift
@@ -99,6 +99,9 @@ func evaluateIfConfig(
   if let identExpr = condition.as(DeclReferenceExprSyntax.self),
     let ident = identExpr.simpleIdentifier?.name
   {
+    if let targetOSDiagnostic = diagnoseLikelyTargetOSTest(at: identExpr, name: ident) {
+      extraDiagnostics.append(targetOSDiagnostic)
+    }
     // Evaluate the custom condition. If the build configuration cannot answer this query, fail.
     return checkConfiguration(at: identExpr) {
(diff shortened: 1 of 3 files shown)


Sources/SwiftIfConfig/IfConfigEvaluation.swift
```diff
@@ -627,6 +630,57 @@ private func diagnoseLikelySimulatorEnvironmentTest(
...
+  "WINDOWS": ("os", "Windows"),
+  "LINUX": ("os", "Linux"),
+  "OSX": ("os", "macOS"),
+  "IOS": ("os", "iOS"),
+  "MACCATALYST": ("targetEnvironment", "macCatalyst"),
+  "TV": ("os", "tvOS"),
+  "WATCH": ("os", "watchOS"),
+  "VISION": ("os", "visionOS"),
+  "SIMULATOR": ("targetEnvironment", "simulator"),
+]
```
ahoppen, line 682: Would it be simpler if the value of the dictionary is an `ExprSyntax` directly and then construct that from a string literal?

```suggestion
private let targetOSNameMap: [String: ExprSyntax] = [
  "WINDOWS": "os(Windows)",
  ...
]
```

j-f1:
@swift-ci please test

ahoppen (approved):
Looks good

j-f1:
@swift-ci please test

ahoppen:
@swift-ci Please test Windows

PR #2811, opened 2024-08-19
Improve unexpected tokens diagnostics

mateusrodriguesxyz:
While investigating https://github.com/swiftlang/swift-syntax/issues/1373 I noticed that replacing a `break` with a `continue` in `canRecoverTo` improves diagnostic in a number of recovery tests.
I've also added a little logic to call `skipSingle` so we achieve desired diagnostic in `testRecovery105()`, but only if possible at the same line, so cases like `testRecovery100()` is not messed up.

### DIFF
--- a/Sources/SwiftParser/Recovery.swift
+++ b/Sources/SwiftParser/Recovery.swift
@@ -123,13 +123,33 @@ extension Parser.Lookahead {
       if currentTokenPrecedence >= recoveryPrecedence {
         break
       }
-      self.consumeAnyToken()
       if let closingDelimiter = currentTokenPrecedence.closingTokenKind {
         let closingDelimiterSpec = TokenSpec(closingDelimiter)
+        let canCloseAtSameLine: Int? = self.withLookahead { lookahead in
+          var tokensToSkip = 0
+          while !lookahead.at(.endOfFile), !lookahead.currentToken.isAtStartOfLine {
+            tokensToSkip += 1
+            if lookahead.at(closingDelimiterSpec) {
+              return tokensToSkip
+            } else {
+              lookahead.consumeAnyToken()
+            }
+          }
+          return nil
+        }
+        if let tokensToSkip = canCloseAtSameLine {
+          for _ in 0..<tokensToSkip {
+            self.consumeAnyToken()
+          }
+          continue
+        }
+        self.consumeAnyToken()
         guard self.canRecoverTo(closingDelimiterSpec) != nil else {
           continue
         }
         self.eat(closingDelimiterSpec)
+      } else {
+        self.consumeAnyToken()
       }
     }
 
--- a/Sources/SwiftParser/Recovery.swift
+++ b/Sources/SwiftParser/Recovery.swift
@@ -127,7 +127,7 @@ extension Parser.Lookahead {
       if let closingDelimiter = currentTokenPrecedence.closingTokenKind {
         let closingDelimiterSpec = TokenSpec(closingDelimiter)
         guard self.canRecoverTo(closingDelimiterSpec) != nil else {
-          break
+          continue
         }
         self.eat(closingDelimiterSpec)
       }
--- a/Tests/SwiftBasicFormatTest/BasicFormatTests.swift
+++ b/Tests/SwiftBasicFormatTest/BasicFormatTests.swift
@@ -53,15 +53,6 @@ fileprivate func assertFormattingRoundTrips(
 }
 
 final class BasicFormatTest: XCTestCase {
-  func testClosureParameter() {
-    let source = """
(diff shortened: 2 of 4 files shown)


Sources/SwiftParser/Recovery.swift
```diff
@@ -123,13 +123,29 @@ extension Parser.Lookahead {
       if currentTokenPrecedence >= recoveryPrecedence {
         break
       }
-      self.consumeAnyToken()
       if let closingDelimiter = currentTokenPrecedence.closingTokenKind {
         let closingDelimiterSpec = TokenSpec(closingDelimiter)
+        let canSkipSingleAtSameLine = self.withLookahead { lookahead in
```
ahoppen, line 128: I don’t quite understand why this works. `canSkipSingleAtSameLine` checks if we have the closing delimiter on the  same line and allows skipping over tokens with a higher precedence. For example, when `closingDelimiter` was `)` and there was a `}` in between, it would skip over that. But then `skipSingle` uses a different logic to essentially discover the closing token again (which doesn’t match for the `)` and `}` I described). 

If the intended behavior is to skip until the closing token on the same line, skipping over any higher precedence tokens (which I think makes sense, it produces good results in `testRecovery105`), I think we should count the number of tokens consumed in the nested lookahead and consume them, not calling into `skipSingle`.

Sources/SwiftParser/Recovery.swift
```diff
@@ -123,13 +123,29 @@ extension Parser.Lookahead {
...
+          return false
+        }
+        if canSkipSingleAtSameLine {
+          skipSingle()
+          continue
+        }
+        self.consumeAnyToken()
         guard self.canRecoverTo(closingDelimiterSpec) != nil else {
-          break
+          continue
```
ahoppen, line 144: Yes, I think this makes a lot of sense so we don’t completely abort recovery on a mismatched `(`. Good catch!

Sources/SwiftParser/Recovery.swift
mateusrodriguesxyz: TBH I came up with this solution after trying just `skipSingle`, which improved `testRecovery105' but broke other test, so I added this lookahead to prevent too much skipping. Your suggestion does make more sense. I will try it, thanks!

ahoppen (approved):
Looks great. Thank you @mateusrodriguesxyz!

ahoppen:
@swift-ci Please test

mateusrodriguesxyz:
Couldn't figure out what caused the tests to fail 🤔

ahoppen:
Looks like `BasicFormatTest.testClosureParameter` failed. That test did look unrelated to your other changes anyway.

```
/Users/ec2-user/jenkins/workspace/swift-syntax-PR-macOS/branch-main/swift-syntax/Tests/SwiftBasicFormatTest/BasicFormatTests.swift:62: error: -[SwiftBasicFormatTest.BasicFormatTest testClosureParameter] : failed - Actual output does not match the expected
 myFunc({
     return true
–})
+    })
```

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen (commented):
Looks like a few files that used to parse correctly now no longer do.

```
Failed Tests (9):
  Swift(macosx-x86_64) :: ModuleInterface/features.swift
  Swift(macosx-x86_64) :: Parse/ConditionalCompilation/compiler.swift
  Swift(macosx-x86_64) :: Parse/ConditionalCompilation/language_version_explicit.swift
  Swift(macosx-x86_64) :: Parse/ConditionalCompilation/sequence_version.swift
  Swift(macosx-x86_64) :: Parse/features.swift
  Swift(macosx-x86_64) :: SILGen/specialize_attr.swift
  Swift(macosx-x86_64) :: SILOptimizer/pre_specialize-macos.swift
  Swift(macosx-x86_64) :: SILOptimizer/pre_specialize.swift
  Swift(macosx-x86_64) :: SILOptimizer/pre_specialize_layouts.swift
```

These files are located in https://github.com/swiftlang/swift/tree/main/test

ahoppen:
Oh, looks like the errors are unrelated to your PR, they also seem to fail in https://github.com/swiftlang/swift-syntax/pull/2814.

DougGregor:
I think https://github.com/swiftlang/swift/pull/76040 will fix this issue

mateusrodriguesxyz:
@ahoppen I think this should pass now that https://github.com/swiftlang/swift/pull/76040 has been merged.

DougGregor:
@swift-ci please test

mateusrodriguesxyz:
Thanks @DougGregor !

PR #2816, opened 2024-08-22
Parse @_specialize with a complete availability clause

DougGregor:
`@_specialize` can also use the form of the availability clause that separates the platform name from the `introduced:` argument label. Parse this more general form.


DougGregor:
@swift-ci please test

DougGregor:
I prefer this to the compiler revert, assuming it's correct

ahoppen (approved):

PR #2817, opened 2024-08-23
Remove unnecessary operators.

johnno1962:
Hi @ahoppen, follow-up PR to https://github.com/swiftlang/swift-syntax/pull/2439 removing some operators that turn out to not be necessary as I just noticed they are implied by their optional counterparts. Does not seem to affect compile/runtime performance.


ahoppen (approved):
Thanks for noticing the cleanup opportunity!

ahoppen:
@swift-ci Please test

PR #2819, opened 2024-08-24
SwiftIfConfig improvements to suit usage in the compiler

DougGregor:
A couple more semantic tweaks and improvements to address the needs of the compiler:
* Rework the configured regions computation to better match the compiler with respect to inactive vs. unparsed
* Turn "configured regions" into a proper struct that also carries with it diagnostics and an `isActive` operation
* Reimplement `SyntaxProtocol.isActive(in:)` in terms of configured regions, and warn that it's slow
* Provide `TokenSyntax` nodes for the import path in `BuildConfiguration.canImport` because the compiler really needs location information

### DIFF
--- a/Sources/SwiftIfConfig/BuildConfiguration.swift
+++ b/Sources/SwiftIfConfig/BuildConfiguration.swift
@@ -9,6 +9,7 @@
 // See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
 //
 //===----------------------------------------------------------------------===//
+import SwiftSyntax
 
 /// Describes the ordering of a sequence of bytes that make up a word of
 /// storage for a particular architecture.
@@ -114,15 +115,15 @@ public protocol BuildConfiguration {
   /// information, which will translate into the `version` argument.
   ///
   /// - Parameters:
-  ///   - importPath: A nonempty sequence of identifiers describing the
-  ///     imported module, which was written in source as a dotted sequence,
-  ///     e.g., `UIKit.UIViewController` will be passed in as the import path
-  ///     array `["UIKit", "UIViewController"]`.
+  ///   - importPath: A nonempty sequence of (token, identifier) pairs
+  ///     describing the imported module, which was written in source as a
+  ///     dotted sequence, e.g., `UIKit.UIViewController` will be passed in as
+  ///     the import path array `[(token, "UIKit"), (token, "UIViewController")]`.
   ///   - version: The version restriction on the imported module. For the
   ///     normal `canImport(<import-path>)` syntax, this will always be
   ///     `CanImportVersion.unversioned`.
   /// - Returns: Whether the module can be imported.
-  func canImport(importPath: [String], version: CanImportVersion) throws -> Bool
+  func canImport(importPath: [(TokenSyntax, String)], version: CanImportVersion) throws -> Bool
 
   /// Determine whether the given name is the active target OS (e.g., Linux, iOS).
   ///
--- a/Sources/SwiftIfConfig/ConfiguredRegions.swift
+++ b/Sources/SwiftIfConfig/ConfiguredRegions.swift
@@ -56,57 +56,95 @@ fileprivate class ConfiguredRegionVisitor<Configuration: BuildConfiguration>: Sy
   /// Whether we are currently within an active region.
   var inActiveRegion = true
 
+  // All diagnostics encountered along the way.
+  var diagnostics: [Diagnostic] = []
+
   init(configuration: Configuration) {
     self.configuration = configuration
     super.init(viewMode: .sourceAccurate)
   }
 
   override func visit(_ node: IfConfigDeclSyntax) -> SyntaxVisitorContinueKind {
-    // If we're in an active region, find the active clause. Otherwise,
-    // there isn't one.
-    let activeClause = inActiveRegion ? node.activeClause(in: configuration).clause : nil
+    // Walk through the clauses to find the active one.
     var foundActive = false
     var syntaxErrorsAllowed = false
     for clause in node.clauses {
-      // If we haven't found the active clause yet, syntax errors are allowed
-      // depending on this clause.
-      if !foundActive {
-        syntaxErrorsAllowed =
-          clause.condition.map {
-            IfConfigClauseSyntax.syntaxErrorsAllowed($0).syntaxErrorsAllowed
-          } ?? false
-      }
+      let isActive: Bool
+      if let condition = clause.condition {
+        if !foundActive {
+          // Fold operators so we can evaluate this #if condition.
+          let (foldedCondition, foldDiagnostics) = IfConfigClauseSyntax.foldOperators(condition)
+          diagnostics.append(contentsOf: foldDiagnostics)
 
-      // If this is the active clause, record it and then recurse into the
-      // elements.
-      if clause == activeClause {
-        assert(inActiveRegion)
+          // In an active region, evaluate the condition to determine whether
+          // this clause is active. Otherwise, this clause is inactive.
+          // inactive.
+          if inActiveRegion {
+            let (thisIsActive, _, evalDiagnostics) = evaluateIfConfig(
+              condition: foldedCondition,
+              configuration: configuration
+            )
+            diagnostics.append(contentsOf: evalDiagnostics)
 
-        regions.append((clause, .active))
+            // Determine if there was an error that prevented us from
+            // evaluating the condition. If so, we'll allow syntax errors
+            // from here on out.
+            let hadError =
+              foldDiagnostics.contains { diag in
+                diag.diagMessage.severity == .error
+              }
+              || evalDiagnostics.contains { diag in
+                diag.diagMessage.severity == .error
+              }
 
-        if let elements = clause.elements {
-          walk(elements)
+            if hadError {
+              isActive = false
+              syntaxErrorsAllowed = true
+            } else {
+              isActive = thisIsActive
+
+              // Determine whether syntax errors are allowed.
+              syntaxErrorsAllowed = foldedCondition.allowsSyntaxErrorsFolded
+            }
+          } else {
+            isActive = false
+
+            // Determine whether syntax errors are allowed, even though we
+            // skipped evaluation of the actual condition.
+            syntaxErrorsAllowed = foldedCondition.allowsSyntaxErrorsFolded
+          }
+        } else {
+          // We already found an active condition, so this is inactive.
+          isActive = false
         }
-
-        foundActive = true
-        continue
+      } else {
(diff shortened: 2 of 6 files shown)


DougGregor:
https://github.com/swiftlang/swift/pull/76074

@swift-ci please test

DougGregor:
https://github.com/swiftlang/swift/pull/76074

@swift-ci please test

DougGregor:
https://github.com/swiftlang/swift/pull/76074

@swift-ci please test Windows

Sources/SwiftIfConfig/BuildConfiguration.swift
```diff
@@ -114,15 +115,15 @@ public protocol BuildConfiguration {
...
+  ///   - importPath: A nonempty sequence of (token, identifier) pairs
+  ///     describing the imported module, which was written in source as a
+  ///     dotted sequence, e.g., `UIKit.UIViewController` will be passed in as
+  ///     the import path array `[(token, "UIKit"), (token, "UIViewController")]`.
   ///   - version: The version restriction on the imported module. For the
   ///     normal `canImport(<import-path>)` syntax, this will always be
   ///     `CanImportVersion.unversioned`.
   /// - Returns: Whether the module can be imported.
-  func canImport(importPath: [String], version: CanImportVersion) throws -> Bool
+  func canImport(importPath: [(TokenSyntax, String)], version: CanImportVersion) throws -> Bool
```
ahoppen, line 126: Why do we need both the token and the string? Couldn’t we extract the identifier from the token?

Sources/SwiftIfConfig/BuildConfiguration.swift
```diff
@@ -9,6 +9,7 @@
 // See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
 //
 //===----------------------------------------------------------------------===//
+import SwiftSyntax
```
ahoppen, line 12: Could you add a newline between the copyright header and the import for consistency with the other files?

Sources/SwiftIfConfig/ConfiguredRegions.swift
```diff
@@ -13,6 +13,90 @@
...
+///
+/// If the configuration options `DEBUG` and `B` are provided, but `A` is not,
+/// and the compiler version is less than 12.0, the results will be contain:
+///   - Active region for the `#if DEBUG`.
+///   - Inactive region for the `#if A`.
+///   - Active region for the `#elseif B`.
+///   - Unparsed region for the `#elseif compiler(>= 12.0)`.
+///   - Inactive region for the final `#else`.
+public struct ConfiguredRegions {
+  let regions: [Element]
```
ahoppen, line 40: Defining this as `Element` where `Element` is only defined in the conformance to `RandomAccessCollection` below seems a little confusing to me (I wasn’t able to figure it out on GitHub alone). I think it would be clearer to explicitly spell out the tuple type explicitly here 

```swift
let regions: [(IfConfigClauseSyntax, IfConfigRegionState)]
```

Sources/SwiftIfConfig/ConfiguredRegions.swift
```diff
@@ -13,6 +13,90 @@
...
+  /// Determine whether the given syntax node is active within the configured
+  /// regions.
+  public func isActive(_ node: some SyntaxProtocol) -> IfConfigRegionState {
+    var currentState: IfConfigRegionState = .active
+    for (ifClause, state) in regions {
+      if node.position < ifClause.position {
+        return currentState
+      }
+
+      if node.position >= ifClause.regionStart && node.position <= ifClause.endPosition {
```
ahoppen, line 54: Do you think of it but do you think it would be clearer to check use `contains here?

```swift
if (ifClause.regionStart..<ifClause.endPosition).contains(node.position) {
```

Sources/SwiftIfConfig/ConfiguredRegions.swift
```diff
@@ -13,6 +13,90 @@
...
+}
+
+extension ConfiguredRegions: CustomDebugStringConvertible {
+  /// Provides source ranges for each of the configured regions.
+  public var debugDescription: String {
+    guard let firstRegion = first else {
+      return "[]"
+    }
+
+    let root = firstRegion.0.root
```
ahoppen, line 80: If you add labels to the tuple, you wouldn’t need to do `.0` here, which I usually think decreases readability.

Sources/SwiftIfConfig/ConfiguredRegions.swift
```diff
@@ -56,58 +143,111 @@ fileprivate class ConfiguredRegionVisitor<Configuration: BuildConfiguration>: Sy
...
+      let isActive: Bool
+      if let condition = clause.condition {
+        if !foundActive {
+          // Fold operators so we can evaluate this #if condition.
+          let (foldedCondition, foldDiagnostics) = IfConfigClauseSyntax.foldOperators(condition)
+          diagnostics.append(contentsOf: foldDiagnostics)
+
+          // In an active region, evaluate the condition to determine whether
+          // this clause is active. Otherwise, this clause is inactive.
+          // inactive.
```
ahoppen, line 179: `inactive` duplicated with the previous line.

Sources/SwiftIfConfig/ConfiguredRegions.swift
```diff
@@ -56,58 +143,111 @@ fileprivate class ConfiguredRegionVisitor<Configuration: BuildConfiguration>: Sy
...
-          clause.condition.map {
-            IfConfigClauseSyntax.syntaxErrorsAllowed($0).syntaxErrorsAllowed
-          } ?? false
-      }
+      let isActive: Bool
+      if let condition = clause.condition {
+        if !foundActive {
+          // Fold operators so we can evaluate this #if condition.
+          let (foldedCondition, foldDiagnostics) = IfConfigClauseSyntax.foldOperators(condition)
+          diagnostics.append(contentsOf: foldDiagnostics)
```
ahoppen, line 175: This is now duplicating logic from `IfConfigDeclSyntax.activeClause(in:)`, right? Do you think there’s any way to avoid the duplication? I’m a little worried that they will get out of sync eventually.

Sources/SwiftIfConfig/ConfiguredRegions.swift
```diff
@@ -56,58 +143,111 @@ fileprivate class ConfiguredRegionVisitor<Configuration: BuildConfiguration>: Sy
...
+            syntaxErrorsAllowed = foldedCondition.allowsSyntaxErrorsFolded
+          }
+        } else {
+          // We already found an active condition, so this is inactive.
+          isActive = false
         }
+      } else {
+        // This is an #else. It's active if we haven't found an active clause
+        // yet and are in an active region.
+        isActive = !foundActive && inActiveRegion
```
ahoppen, line 221: Do we still need to determine `syntaxErrorsAllowed` in that case. Eg. assuming that `DEBUG` is set, do we report `print debugDescription` as `syntaxErrorsAllowed` for the following

```swift
#if DEBUG
print(debugDescription)
#elseif compiler(>99.0)
print debugDescription
#endif
```

Sources/SwiftIfConfig/SyntaxProtocol+IfConfig.swift
```diff
@@ -82,20 +54,10 @@ extension SyntaxProtocol {
   /// If you are querying whether many syntax nodes in a particular file are
   /// active, consider calling `configuredRegions(in:)` once and using
   /// this function. For occasional queries, use `isActive(in:)`.
+  @available(*, deprecated, message: "Please use ConfiguredRegions.isActive(_:)")
```
ahoppen, line 57: Is anyone using this yet (the compiler?). I’d be fine with removing it straight away since we don’t have a release with that API yet.

Sources/SwiftIfConfig/ConfiguredRegions.swift
DougGregor: Sure

Sources/SwiftIfConfig/BuildConfiguration.swift
DougGregor: Smells like a linter rule to me, but okay

Sources/SwiftIfConfig/ConfiguredRegions.swift
DougGregor: That should be `...` to match the semantics of my code, but okay.

Sources/SwiftIfConfig/ConfiguredRegions.swift
DougGregor: sure, I'll label the tuple

Sources/SwiftIfConfig/SyntaxProtocol+IfConfig.swift
DougGregor: Nobody is using it; we can zap it now.

Sources/SwiftIfConfig/ConfiguredRegions.swift
DougGregor: The compiler doesn't determine `syntaxErrorsAllowed` here, and prints an error for `print debugDescription`. We could choose to deviate from the compiler here (and be more lax) if we want to.

Sources/SwiftIfConfig/BuildConfiguration.swift
DougGregor: The `identifier` on the token is optional, so we'd be asking clients to either force the optional every time, or add a spurious "guard". I'd rather just duplicate the information.

Sources/SwiftIfConfig/ConfiguredRegions.swift
DougGregor: We can do a little bit better here, because there were 4 copies of code like this.

Sources/SwiftIfConfig/BuildConfiguration.swift
ahoppen: Ah, I see. Makes sense. I think that explanation in a doc comment would be useful.

Sources/SwiftIfConfig/ConfiguredRegions.swift
ahoppen: OK, that’s surprising to me but matching the compiler here for now makes sense for now to make switching between the two implementations easier. 

Should we gather these oddities that we find somewhere so we can consider fixing them in the future?

DougGregor: It turns out that this behavior changed in the compiler very recently (https://github.com/swiftlang/swift/pull/74415). We could consider whether we need to fix the original issue in some other way.

PR #2823, opened 2024-08-26
Fix `CodeGenerationFormat.formatChildrenSeparatedByNewline()` to not add an extra newline to the end of lists that already have one

roopekv:
This fixes a bug where `CodeGenerationFormat.formatChildrenSeparatedByNewline()` adds an extra newline to the end of lists that already have one.

### DIFF
--- a/CodeGeneration/Sources/Utils/CodeGenerationFormat.swift
+++ b/CodeGeneration/Sources/Utils/CodeGenerationFormat.swift
@@ -136,11 +136,14 @@ public class CodeGenerationFormat: BasicFormat {
       }
     }
     decreaseIndentationLevel()
-    if !formattedChildren.isEmpty {
-      formattedChildren[formattedChildren.count - 1] = formattedChildren[formattedChildren.count - 1].with(
-        \.trailingTrivia,
-        indentedNewline
-      )
+    if let lastChild = formattedChildren.last {
+      let nextTokenStartsWithNewline = lastChild.nextToken(viewMode: .all)?.leadingTrivia.first?.isNewline ?? false
+      if !nextTokenStartsWithNewline {
+        formattedChildren[formattedChildren.count - 1] = lastChild.with(
+          \.trailingTrivia,
+          indentedNewline
+        )
+      }
     }
     return formattedChildren
   }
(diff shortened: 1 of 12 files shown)


CodeGeneration/Sources/Utils/CodeGenerationFormat.swift
```diff
@@ -136,11 +136,14 @@ public class CodeGenerationFormat: BasicFormat {
       }
     }
     decreaseIndentationLevel()
-    if !formattedChildren.isEmpty {
-      formattedChildren[formattedChildren.count - 1] = formattedChildren[formattedChildren.count - 1].with(
-        \.trailingTrivia,
-        indentedNewline
-      )
+    if let lastChild = formattedChildren.last {
+      let nextTokenStartsWithNewline = lastChild.nextToken(viewMode: .all)?.leadingTrivia.first?.isNewline ?? false
```
ahoppen, line 140: Nitpick, I would use `.sourceAccurate` here because we only care about tokens that will actually be printed (not missing tokens). It doesn’t make any difference in practice because we shouldn’t have any missing tokens in the tree but still.

```suggestion
      let nextTokenStartsWithNewline = lastChild.nextToken(viewMode: .sourceAccurage)?.leadingTrivia.first?.isNewline ?? false
```

ahoppen (commented):
Thank you ❤️ Nice clean up

CodeGeneration/Sources/Utils/CodeGenerationFormat.swift
roopekv: Sure thing, can you fix that typo (g -> t) so I can commit the suggestion from here?

ahoppen: Done

roopekv:
Do I squash commits, or is this ok?

roopekv:
The view mode change didn't affect generated code, but it made the line longer so that it should wrap. I'll run format and commit once more.

ahoppen (approved):
Thanks!

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

PR #2825, opened 2024-08-26
Don't always treat a closure following a variable declaration as an accessor block

DougGregor:
Introduce some lookahead to make sure we actually have something that looks like an accessor block when the `{` is on the next line, matching what the C++ parser does.

Fixes https://github.com/swiftlang/swift/issues/76089


DougGregor:
@swift-ci please test

ahoppen (approved):

PR #2827, opened 2024-08-27
[CodeGeneration] Unify handling for node choices

AppAppWorks:
Currently there are three separate implementations for generating node choices enum on the raw and non-raw sides, resulting in code duplication and inconsistencies. Besides, documentation, experimental note, and api attributes have also been omitted in the generation of node choices enum.

This PR aims to fix these issues by unifying and rectifying the facilities for code generation.


CodeGeneration/Sources/SyntaxSupport/RawSyntaxNodeKind.swift
```diff
@@ -0,0 +1,47 @@
...
+  var protocolType: TypeSyntax {
+    switch self {
+    case .syntax, .syntaxCollection:
+      return "RawSyntaxNodeProtocol"
+    default:
+      return "\(self.syntaxType)NodeProtocol"
+    }
+  }
+
+  var isAvailableInDocc: Bool {
```
ahoppen, line 40: Are any raw nodes available in docc? They are all marked as `@_spi(RawSyntax)`, so I would expect that none of them are available in docc.

CodeGeneration/Sources/SyntaxSupport/SyntaxNodeKind.swift
```diff
@@ -507,4 +474,8 @@ public enum SyntaxNodeKind: String, CaseIterable, IdentifierConvertible {
       AttributeSyntax(#"@available(*, deprecated, renamed: "\#(syntaxType)")"#)
     }
   }
+
+  public var raw: some IdentifierConvertible & TypeConvertible {
```
ahoppen, line 478: Why not just return `RawSyntaxNodeKind` here?

CodeGeneration/Sources/SyntaxSupport/TypeConvertible.swift
```diff
@@ -0,0 +1,50 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
+/// Instances of a conforming type should provide `TypeSyntax` nodes  to be used in code generation.
```
ahoppen, line 15:
```suggestion
/// Instances of a conforming type should provide `TypeSyntax` nodes to be used in code generation.
```

Sources/SwiftSyntax/generated/SyntaxCollections.swift
```diff
@@ -124,22 +125,18 @@ public struct AttributeListSyntax: SyntaxCollection, SyntaxHashable {
       self = .ifConfigDecl(node)
     }
     
-    public init?(_ node: some SyntaxProtocol) {
-      if let node = node.as(AttributeSyntax.self) {
+    public init?(_ node: __shared some SyntaxProtocol) {
+      if let node = AttributeSyntax(node) {
```
ahoppen, line 129: I personally prefer the `node.as(AttributeSyntax.self)` casting style because it’s closer to standard Swift.

CodeGeneration/Sources/SyntaxSupport/Child.swift
```diff
@@ -129,6 +129,16 @@ public class Child: IdentifierConvertible {
   /// The first line of the child's documentation
   public let documentationAbstract: String
 
+  /// The documentation note to print for an experimental feature.
+  public var experimentalDocNote: SwiftSyntax.Trivia {
+    let comment = experimentalFeature.map {
+      """
+      - Experiment: Requires experimental feature `\($0.token)`.
```
ahoppen, line 136: IIRC the `- Experiment:` comment in docc is intended for an experiment a user of the API might perform to get familiar with it. That’s  not what experimental means here. I would suggest using `- Note:` or `- Warning:` instead.

Sources/SwiftSyntax/generated/raw/RawSyntaxNodesOP.swift
```diff
@@ -2006,20 +2009,16 @@ public struct RawPrecedenceGroupAttributeListSyntax: RawSyntaxNodeProtocol {
       }
     }
     
-    public init?(_ other: some RawSyntaxNodeProtocol) {
-      if let node = RawPrecedenceGroupRelationSyntax(other) {
+    public init?(_ node: __shared some RawSyntaxNodeProtocol) {
```
ahoppen, line 2012: Nice catch with the missing `__shared` attributes here. 👍🏽

CodeGeneration/Sources/generate-swift-syntax/ChildNodeChoices.swift
```diff
@@ -0,0 +1,210 @@
...
+      public init(_ node: \(self.syntaxType))
+      """
+    ) {
+      ExprSyntax("self = .\(self.memberCallName)(node)")
+    }
+  }
+}
+
+private extension ArraySlice where Element == ChildNodeChoices.Choice {
+  mutating func elseBody() -> IfExprSyntax.ElseBody {
```
ahoppen, line 194: Mutating the array seems very surprising to me. I would prefer something that doesn’t modify the underlying array.

CodeGeneration/Sources/generate-swift-syntax/ChildNodeChoices.swift
```diff
@@ -0,0 +1,210 @@
...
+      \(self.documentation)\
+      \(self.experimentalDocNote)\
+      \(self.apiAttributes)\
+      case \(self.enumCaseDeclName)(\(self.syntaxType))
+      """
+    )
+  }
+
+  func baseTypeInitDecl(hasArgumentName: Bool) -> InitializerDeclSyntax? {
+    if let choiceType = self.choiceType, choiceType.isBase {
```
ahoppen, line 154: Could make this a `guard` to remove one nesting level below.

ahoppen (commented):
Very nice cleanup. I only have a few minor comments.

CodeGeneration/Sources/generate-swift-syntax/ChildNodeChoices.swift
AppAppWorks: I did it for performance sake, maybe I could wrap this ArraySlice api with an immutable façade under Array?

Sources/SwiftSyntax/generated/SyntaxCollections.swift
AppAppWorks: I just blindly copied that from the non-raw side ;)

CodeGeneration/Sources/SyntaxSupport/RawSyntaxNodeKind.swift
AppAppWorks: I didn't know that, I'll just set it to false then.

Sources/SwiftSyntax/generated/raw/RawSyntaxNodesOP.swift
AppAppWorks: Benefits of code deduplication reaped :)

CodeGeneration/Sources/generate-swift-syntax/ChildNodeChoices.swift
ahoppen: Can’t you do something loop-based instead of recursive, like

```swift
func dunnoAboutTheName(_ choices: [ChildNodeChoices.Choice]) -> IfExprSyntax {
  precondition(!choices.isEmpty)
  var result: IfExprSyntax? = nil

  for choice in choices.reversed() {
    let elseBody: IfExprSyntax.ElseBody
    if let result {
      elseBody = .ifExpr(result)
    } else {
      elseBody = .codeBlock("return nil")
    }
    let ifExpr = IfExprSyntax(
      conditions: [
        ConditionElementSyntax(
          condition: .expression("let node = node.as(\(choice.syntaxType).self) "))
      ],
      body: CodeBlockSyntax(statements: ["self = .\(choice.memberCallName)(node)"]),
      elseKeyword: .keyword(.else),
      elseBody: elseBody
    )
  }
  return result!
}
```

ahoppen (commented):
Only open discussion is https://github.com/swiftlang/swift-syntax/pull/2827/files#r1733633656, otherwise LGTM.

AppAppWorks:
I've changed the generation of ifExpr from recursion-based to iteration-based, LGTM :)

AppAppWorks:
This PR should be ready for ci-test now :)

ahoppen (approved):
LGTM

ahoppen:
@swift-ci Please test

PR #2829, opened 2024-08-28
Add `buildFinalResult` to `CodeBlockItemListBuilder` to ensure newline between expressions

mateusrodriguesxyz:
This ensures that the following usage of `CodeBlockItemListBuilder` produces a valid block where expressions are separated by newlines:

```swift
CodeBlockItemListSyntax {
    "let a = 1"
    "let b = 2"
    "let c = 3"
}
```
Something similar it's already done for comma separated lists [here](https://github.com/swiftlang/swift-syntax/blob/8740e1555d3202ec65bb6ccc2721a87ec450a7e3/Sources/SwiftSyntaxBuilder/ListBuilder.swift#L108-L116).

I've also modified the `collapse` function to avoid adding unnecessary separators to a macro expansion. This change is important because if someone uses `CodeBlockItemListBuilder` as in [testDontAddIndentationWhenCollapsingBody](https://github.com/swiftlang/swift-syntax/blob/c2c798059a3787694d1fd22668d55d09d725c101/Tests/SwiftSyntaxMacroExpansionTest/BodyMacroTests.swift#L159C8-L159C48), they might be surprised by unexpected empty lines if `collapse` adds separator unconditionally.

```swift
func f() {
#sourceLocation(file: "test.swift", line: 3)

        let x: Int = 1

#sourceLocation()
}
```

### DIFF
--- a/Sources/SwiftSyntaxMacroExpansion/MacroExpansion.swift
+++ b/Sources/SwiftSyntaxMacroExpansion/MacroExpansion.swift
@@ -528,5 +528,15 @@ public func collapse<Node: SyntaxProtocol>(
     break
   }
 
-  return expansions.joined(separator: separator)
+  // Join the expansions ensuring `separator` between them.
+  var collapsed = ""
+  for expansion in expansions {
+    if collapsed.isEmpty || expansion.hasPrefix(separator) {
+      collapsed.append(expansion)
+    } else {
+      collapsed.append(separator + expansion)
+    }
+  }
+
+  return collapsed
 }
--- a/Sources/SwiftSyntaxBuilder/ResultBuilderExtensions.swift
+++ b/Sources/SwiftSyntaxBuilder/ResultBuilderExtensions.swift
@@ -40,6 +40,18 @@ extension CodeBlockItemListBuilder {
   public static func buildExpression(_ expression: some Sequence<DeclSyntaxProtocol>) -> Component {
     buildExpression(expression.map { CodeBlockItemSyntax(item: .decl(DeclSyntax($0))) })
   }
+
+  public static func buildFinalResult(_ component: Component) -> CodeBlockItemListSyntax {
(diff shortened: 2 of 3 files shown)


ahoppen (commented):
Thank you @mateusrodriguesxyz. Could you add a test case for this?

Sources/SwiftSyntaxMacroExpansion/MacroExpansion.swift
```diff
@@ -528,5 +528,15 @@ public func collapse<Node: SyntaxProtocol>(
     break
   }
 
-  return expansions.joined(separator: separator)
+  // Join the expansions ensuring `separator` between them.
+  var collapsed = ""
+  for expansion in expansions {
+    if (collapsed.isEmpty || expansion.hasPrefix(separator)) {
```
ahoppen, line 534: Nitpick, sorry for not noticing earlier: The parentheses are not needed.

```suggestion
    if collapsed.isEmpty || expansion.hasPrefix(separator) {
```

ahoppen (approved):
Just one nitpick, otherwise looks good to me.

ahoppen:
@swift-ci Please test

ahoppen:
Not entirely sure why but there seems to be a build failure on Linux only. 🤔 Could you take a look at it @mateusrodriguesxyz?

mateusrodriguesxyz:
> Not entirely sure why but there seems to be a build failure on Linux only.

It's because I used [SE-0380](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0380-if-switch-expressions.md) which it's not available in Swift 5.8 😅 It's fixed now. Could you test again please? @ahoppen

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

PR #2830, opened 2024-08-29
Braces following a variable declaration with no initializer clause are accessors

DougGregor:
Fixes rdar://134877843, a regression introduced recently as part of the fix to https://github.com/swiftlang/swift/issues/76089.


DougGregor:
@swift-ci please test

PR #2832, opened 2024-08-30
Implementation of the compiler's "extract inlinable text"

DougGregor:
The compiler has an operation to extract "inlinable" text for use in Swift interface generation. The operation removes all "inactive" code (per a build configuration) that doesn't involve the compiler version or feature checks, as well as stripping comments and `#sourceLocation`. 

Provide those APIs here based on swift-syntax.

### DIFF
--- a/Sources/SwiftIfConfig/ActiveSyntaxRewriter.swift
+++ b/Sources/SwiftIfConfig/ActiveSyntaxRewriter.swift
@@ -18,15 +18,20 @@ extension SyntaxProtocol {
   /// are inactive according to the given build configuration, leaving only
   /// the code that is active within that build configuration.
   ///
-  /// Returns the syntax node with all inactive regions removed, along with an
-  /// array containing any diagnostics produced along the way.
-  ///
   /// If there are errors in the conditions of any configuration
   /// clauses, e.g., `#if FOO > 10`, then the condition will be
   /// considered to have failed and the clauses's elements will be
   /// removed.
+  /// - Parameters:
+  ///   - configuration: the configuration to apply.
+  ///   - retainFeatureCheckIfConfigs: whether to retain `#if` blocks involving
+  ///     compiler version checks (e.g., `compiler(>=6.0)`) and `$`-based
+  ///     feature checks.
+  /// - Returns: the syntax node with all inactive regions removed, along with
+  ///   an array containing any diagnostics produced along the way.
   public func removingInactive(
-    in configuration: some BuildConfiguration
+    in configuration: some BuildConfiguration,
+    retainFeatureCheckIfConfigs: Bool = false
   ) -> (result: Syntax, diagnostics: [Diagnostic]) {
     // First pass: Find all of the active clauses for the #ifs we need to
     // visit, along with any diagnostics produced along the way. This process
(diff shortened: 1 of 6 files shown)


DougGregor:
@swift-ci please test

Sources/SwiftIfConfig/ActiveSyntaxRewriter.swift
```diff
@@ -307,3 +327,104 @@ class ActiveSyntaxRewriter<Configuration: BuildConfiguration>: SyntaxRewriter {
...
+
+extension SyntaxProtocol {
+  // Produce the source code for this syntax node with all of the comments
+  // and #sourceLocations removed. Each comment will be replaced with either
+  // a newline or a space, depending on whether the comment involved a newline.
+  @_spi(Compiler)
+  public var descriptionWithoutCommentsAndSourceLocations: String {
+    var result = ""
+    var skipUntilRParen = false
+    for token in tokens(viewMode: .sourceAccurate) {
```
rintaro, line 391: `SyntaxVisitor` is _probably_ faster because it uses `SyntaxNodeFactory` hack.
```swift
class DescriptionWithoutCommentsAndSourceLocationsVisotr: SyntaxVisitor {
  var result: String = ""
  override func visit(_ token: TokenSyntax) -> SyntaxVisitorContinueKind {
    token.leadingTrivia.writeWithoutComments(to: &result)
    token.text.write(to: &result)
    token.trailingTrivia.writeWithoutComments(to: &result)
    return .skipChildren
  }
  override func visit(_ node: PoundSourceLocationSyntax) -> SyntaxVisitorContinueKind {
    return .skipChildren
  }
}
```

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test Windows

Sources/SwiftIfConfig/ActiveSyntaxRewriter.swift
DougGregor: This is much cleaner than my implementation, too!

Sources/SwiftIfConfig/ActiveSyntaxRewriter.swift
```diff
@@ -307,3 +346,104 @@ class ActiveSyntaxRewriter<Configuration: BuildConfiguration>: SyntaxRewriter {
...
+    return visitor.foundFeatureCheck
+  }
+}
+
+extension IfConfigDeclSyntax {
+  /// Whether any of the clauses in this #if contain a feature check.
+  var containsFeatureCheck: Bool {
+    return clauses.contains { clause in
+      if let condition = clause.condition {
+        return condition.containsFeatureCheck
```
ahoppen, line 394: Just checking: If we have the following we don’t want to remove `secretCode` right?

```swift
#if SECRET
secretCode()
#elseif compiler(>=6.0)
swift6Code()
#else
swift5Code()
#endif
```

DougGregor: Oh my, I really screwed that up. Thank you!

DougGregor: Okay, now I'm looking back at this more closely. I thought I got the basic algorithm wrong, but I did not: this matches what the compiler is doing today, where `secretCode()` is not removed. We could decide to do better than the compiler here if we want, and it probably wouldn't break anything.

ahoppen: I think that might be worth it 🤔

PR #2833, opened 2024-08-31
Fix `CodeGenerationFormat` to not add extra indentation, leave trailing spaces when adding newlines, or indent empty lines

roopekv:
This fixes multiple issues with `CodeGenerationFormat`. Files were re-generated with each commit, so the differences in formatting caused by each change to `CodeGenerationFormat` should be easy to review.

[Fix CodeGenerationFormat to not add extra indentation](https://github.com/swiftlang/swift-syntax/commit/1477e57387965776dd2cafa516543143bd2f33a5)

- Indentation used to be all over the place when formatting children to be separated by newline, this should fix it.

[Fix CodeGenerationFormat to not leave trailing spaces when adding newlines](https://github.com/swiftlang/swift-syntax/commit/0cdde9c52f1f2ffefca4f5f008832be0e542cb68)

- Formatting children to be separated by newline used to leave spaces after each line's comma, this should trim them.

[Fix CodeGenerationFormat to not indent empty lines](https://github.com/swiftlang/swift-syntax/commit/9b51a7b864e83ca418fbe50a31f52334b9802cb6)

- `ensuringTwoLeadingNewlines` used to add indented newlines, resulting in empty lines being indented. Not anymore.


CodeGeneration/Sources/Utils/CodeGenerationFormat.swift
```diff
@@ -124,16 +143,16 @@ public class CodeGenerationFormat: BasicFormat {
     children: SyntaxChildren,
     elementType: SyntaxType.Type
   ) -> [SyntaxType] {
-    increaseIndentationLevel()
-    var formattedChildren = children.map {
-      self.rewrite($0.cast(SyntaxType.self)).cast(SyntaxType.self)
+    var formattedChildren = children.map { child in
+      var child = child.cast(SyntaxType.self)
+      child.leadingTrivia = Trivia(pieces: child.leadingTrivia.drop(while: \.isWhitespace))
```
ahoppen, line 148: If a child would span multiple lines, this only removes and re-adds whitespace for the first line and the subsequent lines would still be off. Can you think of a solution that immediately assigns the right indentation to the child? How many whitespaces (compared to `indentedNewline`) are we stripping here? Are we just off one indentation level? If so, could we fix that by arranging the calls to `increaseIndentationLevel` and `decreaseIndentationLevel` correctly?

ahoppen (commented):
Great changes. These have always bothered me but never enough that I set time aside to fix them. So, thank you for doing it 🙏🏽 

First two commits look good, I have one comment about the indentation change.

ahoppen (approved):
Looks good to me. Thanks! Also good to see that the latest commit fixed some more formatting issues.

Could you rebase your changes onto `main` to fix the merge conflict?

CodeGeneration/Sources/Utils/CodeGenerationFormat.swift
roopekv: Thanks for the notice. Calling `increaseIndentationLevel()` before rewriting the children seems to fix the issue of children that span multiple lines not getting indented probably.

Before:
https://github.com/swiftlang/swift-syntax/blob/9b51a7b864e83ca418fbe50a31f52334b9802cb6/Sources/SwiftSyntaxBuilder/generated/BuildableNodes.swift#L53-L55

After:
https://github.com/swiftlang/swift-syntax/blob/b63eaaf38da029715f045730405ed77f79d0c006/Sources/SwiftSyntaxBuilder/generated/BuildableNodes.swift#L53-L55

Regarding the removal and re-adding of leading whitespace, it was to prevent adding newlines and indentation to children that were already separated by newlines. I changed it so that indented newlines are only added to children that don't already start on newlines instead of removing the whitespace and adding it back to all children, although it should work either way.

Both of these changes are included in:
[Fix CodeGenerationFormat to correctly indent certain children that span multiple lines](https://github.com/swiftlang/swift-syntax/pull/2833/commits/b63eaaf38da029715f045730405ed77f79d0c006)

**However...**
... I found another issue while inspecting the generated code:
https://github.com/swiftlang/swift-syntax/blob/b63eaaf38da029715f045730405ed77f79d0c006/Sources/SwiftSyntax/generated/syntaxNodes/SyntaxNodesQRS.swift#L854-L861

The arguments and the right parenthesis of `.replacingChild(...)` are missing a level of indentation. Prior to this pull request this particular case looked OK, but I believe it was a false positive, as most of the code formatted by `CodeGenerationFormat` was over-indented. 

I had fixed the over-indentation by overriding `requiresIndent(_:)`,

https://github.com/swiftlang/swift-syntax/blob/b63eaaf38da029715f045730405ed77f79d0c006/CodeGeneration/Sources/Utils/CodeGenerationFormat.swift#L107-L124

which revealed that `.replacingChild(...)` is under-indented. The issue might have to do with the fact that the member function call starts on a newline, as it seems to be the only place where this issue manifests. Perhaps it has something to do with how `BasicFormat` handles the indentation for this case? Any insight on this issue @ahoppen ? Either way, I'll look into this issue during the weekend and try to resolve it.

ahoppen: Thanks for the detailed analysis.

My guess would be that this is because the function call is modeled as follows (simplified):
- callee: `Syntax(self).replacingChild`
- arguments: `(at: 1 ….)`

Thus, the call’s indentation is anchored at the base of the callee, which is the indentation level of `Syntax` here, not the indentation level of `replacingChild`. I don’t have an immediate suggestion of how to fix this but didn’t think about it a lot either. Maybe you can come up with something. If my guess is correct, this would be an issue in `BasicFormat`, not just `CodeGenerationFormat`.

ahoppen: Happy to take this PR as-is and fix that in a follow-up PR. That helps keep the reviews easier if the PRs are narrow in scope.

roopekv: > Thanks for the detailed analysis.
> 
> My guess would be that this is because the function call is modeled as follows (simplified):
> 
> * callee: `Syntax(self).replacingChild`
> * arguments: `(at: 1 ….)`
> 
> Thus, the call’s indentation is anchored at the base of the callee, which is the indentation level of `Syntax` here, not the indentation level of `replacingChild`. I don’t have an immediate suggestion of how to fix this but didn’t think about it a lot either. Maybe you can come up with something. If my guess is correct, this would be an issue in `BasicFormat`, not just `CodeGenerationFormat`.

Alright, thanks for the insight 👍

roopekv: > Happy to take this PR as-is and fix that in a follow-up PR. That helps keep the reviews easier if the PRs are narrow in scope.

Sounds good to me. Shall I fix the merge issues now?

roopekv:
@ahoppen is this good to go now?

ahoppen:
Yes, let’s 🚢it

ahoppen:
@swift-ci Please test

ahoppen:
The macOS test failure is unrelated to your changes. I’m looking into them.

ahoppen:
Sorry, there’s a merge conflict now. Could you rebase?

roopekv:
Sure, just a moment.

roopekv:
Done

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test macOS

ahoppen:
@swift-ci Please test macOS

ahoppen:
@swift-ci Please test macOS

PR #2834, opened 2024-09-03
[SwiftIfConfig] Make ConfiguredRegions and a BuildConfiguration generally interchangeable

DougGregor:
Make the API surface area available for build configurations also available for configured regions. This allows clients that are going to provide repeated `#if`-related queries to produce configured regions once (for the whole syntax tree) and then perform more queries and traversals against it. Overall, this affects a few APIs:

* The `removingInactive(in:)` API is now available on `ConfiguredRegions` as well as on a syntax node (with build configuration). In both cases, it requires a single pass (whereas it was previously two-pass and evaluated `#if`s multiple times). This require a bit of a rework in how the recursion works in this operation.
* The `ActiveSyntax(Any)Visitor` visitor classes can be initialized with either `ConfiguredRegions` or `some BuildConfiguration`. These visitor classes are no longer generic, which makes them simpler to use.

### DIFF
--- a/Sources/SwiftIfConfig/ActiveClauseEvaluator.swift
+++ b/Sources/SwiftIfConfig/ActiveClauseEvaluator.swift
@@ -1,0 +1,50 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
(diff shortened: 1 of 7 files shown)


DougGregor:
https://github.com/swiftlang/swift/pull/76187

@swift-ci please test

DougGregor:
https://github.com/swiftlang/swift/pull/76187

@swift-ci please test macOS

DougGregor:
https://github.com/swiftlang/swift/pull/76187

@swift-ci please test Linux

ahoppen (commented):
Nice. Looks good to me 👍🏽

PR #2842, opened 2024-09-06
[Syntax] Perf: Optimize 'root' and 'ancestorOrSelf'

rintaro:
Accessing `Syntax.parent` requires retain/release traffic. Especially when traversing ancestors to the root, that ARC traffic is not negligible.

Introduce `UnownedSyntax` that provides a way to access the parents temporarily without causing ARC traffics.

### DIFF
--- a/Sources/SwiftSyntax/Syntax.swift
+++ b/Sources/SwiftSyntax/Syntax.swift
@@ -81,10 +81,13 @@ public struct Syntax: SyntaxProtocol, SyntaxHashable {
     }
   }
 
-  private var root: Syntax {
-    switch info.info! {
-    case .root(_): return self
-    case .nonRoot(let info): return info.parent.root
+  public var root: Syntax {
+    return self.withUnownedSyntax {
+      var node = $0
+      while let parent = node.parent {
+        node = parent
+      }
+      return node.value
     }
   }
 
@@ -129,7 +132,10 @@ public struct Syntax: SyntaxProtocol, SyntaxHashable {
   }
 
   /// "designated" memberwise initializer of `Syntax`.
-  init(_ raw: RawSyntax, info: Info) {
+  // transparent because normal inlining is too late for eliminating ARC traffic for Info.
+  // FIXME: Remove @_transparent after OSSA enabled.
+  @_transparent
+  init(_ raw: RawSyntax, info: __shared Info) {
     self.raw = raw
     self.info = info
   }
@@ -309,7 +315,7 @@ public struct Syntax: SyntaxProtocol, SyntaxHashable {
   /// Create a ``Syntax`` node from a specialized syntax node.
   // Inline always so the optimizer can optimize this to a member access on `syntax` without having to go through
   // generics.
-  @inline(__always)
+  @_transparent
   public init(_ syntax: __shared some SyntaxProtocol) {
     self = syntax._syntaxNode
   }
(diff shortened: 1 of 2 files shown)


rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test macOS

Sources/SwiftSyntax/Syntax.swift
```diff
@@ -129,6 +132,7 @@ public struct Syntax: SyntaxProtocol, SyntaxHashable {
   }
 
   /// "designated" memberwise initializer of `Syntax`.
+  @_transparent
```
ahoppen, line 135: Do we need `@_transparent` here? It does make debugging odd because you can no longer step into the function. Isn‘t `@inline(__always)` enough? Same for the other `@_transparent` functions. https://github.com/swiftlang/swift/blob/main/docs/TransparentAttr.md for reference.

Sources/SwiftSyntax/Syntax.swift
```diff
@@ -380,6 +384,57 @@ extension Syntax {
   }
 }
 
+/// Temporary non-owning Syntax.
+///
+/// This can be used for handing Syntax node without ARC traffic.
```
ahoppen, line 389:
```suggestion
/// This can be used for handling Syntax node without ARC traffic.
```

Sources/SwiftSyntax/Syntax.swift
```diff
@@ -380,6 +384,57 @@ extension Syntax {
...
+  }
+
+  /// Get the parent of the Syntax value, but without retaining it.
+  @_transparent
+  var parent: UnownedSyntax? {
+    return info._withUnsafeGuaranteedRef {
+      switch $0.info.unsafelyUnwrapped {
+      case .nonRoot(let info):
+        return UnownedSyntax(info.parent)
+      case .root(_):
```
ahoppen, line 413:
```suggestion
      case .root:
```

Sources/SwiftSyntax/SyntaxProtocol.swift
```diff
@@ -241,14 +237,18 @@ extension SyntaxProtocol {
...
   public func ancestorOrSelf<T>(mapping map: (Syntax) -> T?) -> T? {
-    var walk: Syntax? = Syntax(self)
-    while let unwrappedParent = walk {
-      if let mapped = map(unwrappedParent) {
-        return mapped
-      }
-      walk = unwrappedParent.parent
+    return withUnownedSyntax(self) {
+      var node = $0
+      repeat {
```
ahoppen, line 242: If this is an infinite loop, would `while true` be easier to read than `repeat { } while true`?

ahoppen (commented):
Nice idea

Sources/SwiftSyntax/Syntax.swift
rintaro: Looks like we're inconsistent. `var nonRootInfo` uses `case .root(_)`

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

Sources/SwiftSyntax/Syntax.swift
ahoppen: Hmm, that’s unfortunate. Could you add a comment with this explanation to the `@_transparent` attribute?

Sources/SwiftSyntax/Syntax.swift
```diff
@@ -380,6 +384,57 @@ extension Syntax {
   }
 }
 
+/// Temporary non-owning Syntax.
+///
+/// This can be used for handling Syntax node without ARC traffic.
+struct UnownedSyntax {
+  var raw: RawSyntax
+  var info: Unmanaged<Syntax.Info>
```
ahoppen, line 392: Should these be `private` (or at least `info` be `private`). Seems like the intended access paths are via `value` and `parent`.

Sources/SwiftSyntax/Syntax.swift
```diff
@@ -380,6 +384,57 @@ extension Syntax {
...
+      body(Syntax(self.raw, info: $0))
+    }
+  }
+}
+
+/// Execute the `body` with ``UnownedSyntax`` of `node`.
+///
+/// This guarantees the life time of the `node` during the `body` is executed.
+@inline(__always)
+func withUnownedSyntax<T>(_ node: some SyntaxProtocol, _ body: (UnownedSyntax) -> T) -> T {
```
ahoppen, line 432: Would this be better as a member on `SyntaxProtocol` so that you would do `self.withUnownedSyntax` instead of `withUnownedSyntax(self)`?

ahoppen (approved):

rintaro:
https://github.com/swiftlang/swift/pull/76360
@swift-ci Please test

rintaro:
https://github.com/swiftlang/swift/pull/76360
@swift-ci Please test Windows

PR #2848, opened 2024-09-12
[Diagnostics] fix trivia merging and transfer

AppAppWorks:
This PR will deprecate the incorrectly implemented methods `Trivia.merging(_:)` and `Trivia.merging(triviaOf:)`. Making use of the newly introduced methods`Trivia.mergingCommonPrefix` and `Trivia.mergingCommonSuffix`, this PR will also fix issues in transferring trivia when refactoring code or applying Fix-it.


Sources/SwiftParserDiagnostics/DiagnosticExtensions.swift
```diff
@@ -52,13 +52,39 @@ extension FixIt {
...
 // MARK: - Make missing
 
+private extension FixIt.Change {
+  /// Transfers the leading and trailing trivia of `nodes` to the previous token
+  /// While doing this, it tries to be smart, merging trivia where it makes sense
+  /// and refusing to add e.g. a space after punctuation, where it usually
+  /// doesn't make sense.
+  static func transferTriviaAtSides(from nodes: [Syntax]) -> Self? {
+    guard let first = nodes.first, let last = nodes.last else {
+      preconditionFailure()
```
ahoppen, line 62:
```suggestion
      preconditionFailure("nodes must not be empty")
```

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
```diff
@@ -757,7 +757,7 @@ public class ParseDiagnosticsGenerator: SyntaxAnyVisitor {
         node,
         .standaloneSemicolonStatement,
         fixIts: [
-          FixIt(message: RemoveNodesFixIt(semicolon), changes: .makeMissing(semicolon))
+          FixIt(message: RemoveNodesFixIt(semicolon), changes: .makeMissing(semicolon, transferTrivia: false))
```
ahoppen, line 760: Why do we need `transferTrivia: false` here? Shouldn’t we keep the trivia? For example the comment in the following case?

```swift
let x = 1 ; // some comment
```

Sources/SwiftSyntax/Trivia.swift
ahoppen, line 1: Could you add an entry in `Release Notes/601.md` for the public API changes?

Sources/SwiftSyntax/Trivia.swift
```diff
@@ -69,6 +69,7 @@ public struct Trivia: Sendable {
 
   /// Creates a new ``Trivia`` by merging in the given trivia. Only includes one
   /// copy of a common prefix of `self` and `trivia`.
+  @available(*, deprecated, message: "Use mergingCommonPrefix(trivia) or mergingCommonSuffix(trivia) instead")
   public func merging(_ trivia: Trivia?) -> Trivia {
```
ahoppen, line 73: What do you think of continuing to have a single `merging` function that merges both a common prefix and a common suffix? I would expect that that’s what most callers would prefer. I doubt that anyone is actively thinking about whether they want to merge the prefix or the suffix?

Sources/SwiftParserDiagnostics/DiagnosticExtensions.swift
```diff
@@ -68,57 +94,42 @@ extension FixIt.MultiNodeChange {
...
+    self.makeMissing(nodes.map(Syntax.init), transferTrivia: transferTrivia)
   }
 
   /// Replace present nodes with their missing equivalents.
   ///
   /// If `transferTrivia` is `true`, the leading trivia of the first node and
   /// the trailing trivia of the last node will be transferred to their adjecent
   /// tokens.
-  static func makeMissing(_ nodes: [Syntax], transferTrivia: Bool = true) -> Self {
+  private static func makeMissing(_ nodes: [Syntax], transferTrivia: Bool = true) -> Self {
```
ahoppen, line 123: Do we even need this or could `makeMissing(_ nodes: [some SyntaxProtocol], transferTrivia: Bool)` and `makeMissing(_ nodes: [Syntax], transferTrivia: Bool)` be merged into one function.

Sources/SwiftSyntax/Trivia.swift
AppAppWorks: The issue here is `merging` doesn't actually merge by common prefix or common suffix despite its documentation, changing its behavior may break client code that depends on it, that's why I suggested to deprecate it.

Sources/SwiftParserDiagnostics/ParseDiagnosticsGenerator.swift
AppAppWorks: I added `transferTrivia: false` to kinda sweep a nasty issue under the rug but I forgot to fix it afterwards. I'll need to some more time to properly handle,
```swift
for 
  ;
} 
// after Fix-its
for 
  <#pattern#> in <#expression#> {
}
// desired
for <#pattern#> in <#expression#> {
}
```

ahoppen: I think the `after Fix-its` result is acceptable in this case. It’s a silly example.

Sources/SwiftSyntax/Trivia.swift
ahoppen: I would consider that a bug fix and thus an acceptable behavior change.

PR #2851, opened 2024-09-13
Fix `CodeGenerationFormat` to correctly indent `LabeledExprListSyntax` in method calls that start on a new line

roopekv:
This fixes a case where  `CodeGenerationFormat` would incorrectly indent the arguments (`LabeledExprListSyntax`) and the right parenthesis of method calls that start on a new line when formatting children to be separated by a newline.

### DIFF
--- a/CodeGeneration/Sources/Utils/CodeGenerationFormat.swift
+++ b/CodeGeneration/Sources/Utils/CodeGenerationFormat.swift
@@ -96,6 +96,18 @@ public class CodeGenerationFormat: BasicFormat {
     let children = node.children(viewMode: .all)
     // Short tuple element list literals are presented on one line, list each element on a different line.
     if children.count > maxElementsOnSameLine {
+      let inMethodCallThatStartsOnNewline =
+        node.parent?.as(FunctionCallExprSyntax.self)?.calledExpression.as(MemberAccessExprSyntax.self)?.period
+        .startsOnNewline ?? false
+      if inMethodCallThatStartsOnNewline {
+        increaseIndentationLevel()
+      }
+      defer {
+        if inMethodCallThatStartsOnNewline {
+          decreaseIndentationLevel()
+        }
+      }
+
       return LabeledExprListSyntax(
         formatChildrenSeparatedByNewline(children: children, elementType: LabeledExprSyntax.self)
       )
@@ -111,10 +123,7 @@ public class CodeGenerationFormat: BasicFormat {
       if indentManually {
         return false
       }
-      let startsOnNewline =
-        node.leadingTrivia.contains(where: \.isNewline)
-        || node.previousToken(viewMode: .sourceAccurate)?.trailingTrivia.contains(where: \.isNewline) ?? false
-      if !startsOnNewline {
+      if !node.startsOnNewline {
         return false
       }
     default:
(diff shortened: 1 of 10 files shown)


roopekv:
This fixes the remaining issue discussed in #2833 with @ahoppen and _should_ be mergeable without conflicts.

CodeGeneration/Sources/Utils/CodeGenerationFormat.swift
```diff
@@ -96,6 +96,24 @@ public class CodeGenerationFormat: BasicFormat {
...
     if children.count > maxElementsOnSameLine {
+      let inMethodCallThatStartsOnNewline =
+        if let functionCallExpr = node.parent?.as(FunctionCallExprSyntax.self),
+          let memberAccessExpr = functionCallExpr.calledExpression.as(MemberAccessExprSyntax.self),
+          startsOnNewline(memberAccessExpr.period)
+        {
+          true
+        } else {
+          false
+        }
```
ahoppen, line 107: If you mad `startsOnNewline` an extension on `SyntaxProtocol` this could be done without an `if` expression:

```swift
node.parent?.as(FunctionCallExprSyntax.self).calledExpression.as(MemberAccessExprSyntax.self)?.startsOnNewline ?? false
```

---

Why does this only apply if the called expression is a `MemberAccessExprSyntax`? Shouldn’t the same be true for top level function calls as well?

ahoppen (commented):
Nice, just a syntactic comment. The approach looks good to me

CodeGeneration/Sources/Utils/CodeGenerationFormat.swift
roopekv: > If you mad `startsOnNewline` an extension on `SyntaxProtocol` this could be done without an `if` expression:
> 
> ```swift
> node.parent?.as(FunctionCallExprSyntax.self).calledExpression.as(MemberAccessExprSyntax.self)?.startsOnNewline ?? false
> ```

Done 👍

> Why does this only apply if the called expression is a `MemberAccessExprSyntax`? Shouldn’t the same be true for top level function calls as well?

If you mean top level functions calls by specifying the module, they also parse into `MemberAccessExprSyntax`. Example:
```swift
Swift
  .print(
    "1",
    "2",
    separator: ",",
    terminator: ";"
  )
``` 
image

ahoppen: I mean if you have something like

```swift
print(
  "1", 
  "2"
)
```

Or do we format that correctly already?

roopekv: > Or do we format that correctly already?

Yes we do.

roopekv:
Should I squash the commits to remove the use of the `if` expression etc. @ahoppen ?

ahoppen:
Yes, that would be good

CodeGeneration/Sources/Utils/CodeGenerationFormat.swift
ahoppen: Oh, great. Thanks for confirming

roopekv:
Done 👍

ahoppen:
@swift-ci Please test

ahoppen (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #2852, opened 2024-09-17
[SwiftLexicalLookup] Add ASTScope related fixes and lookInMembers result kind.

MAJKFL:
This PR brings closer the current `SwiftLexicalLookup` unqualified lookup implementation to `ASTScope`. Changes include:
- New `lookInMembers` result kind. It prompts client where to perform member lookup.
- Add configuration to not include member names. Compiler doesn't introduce members during unqualified lookup itself.
- Add configuration to terminate lookup early in first encountered sequential scope. It's supposed to match behavior of `ASTScope::lookupLocalDecls` `stopAfterInnermostBraceStmt` parameter.
- Fix `self` keyword behavior. Compiler introduces it on a function boundary, that's why this PR moves it into function declaration scope.
- Reverse order of name introductions in sequential scopes.
- Make closure expression a sequential scope to properly handle it's children.
- Remove setting of `accessibleAfter` for generic parameters.
 
This PR also includes a simple debug description functionality for all lookup components to make it easier to work with the library.

### DIFF
--- a/Sources/SwiftLexicalLookup/Configurations/LookupConfig.swift
+++ b/Sources/SwiftLexicalLookup/Configurations/LookupConfig.swift
@@ -14,8 +14,43 @@
   /// Specifies behavior of file scope.
   @_spi(Experimental) public var fileScopeHandling: FileScopeHandlingConfig
   /// Specifies whether lookup should finish in the closest sequential scope.
+  ///
+  /// ### Example
+  /// ```swift
+  /// class X {
+  ///   let a = 42
+  ///
+  ///   func (a: Int) {
+  ///     let a = 123
+  ///
+  ///     a // <-- lookup here
+  ///   }
+  /// }
+  /// ```
+  /// When looking up at the specified position with `finishInSequentialScope`
+  /// set to `false`, lookup will return declaration from inside function body,
+  /// function parameter and the `a` declaration from `class X` member block.
+  /// If `finishInSequentialScope` would be set to `false`, the only name
+  /// returned by lookup would be the `a` declaration from inside function body.
   @_spi(Experimental) public var finishInSequentialScope: Bool
   /// Specifies whether to include results generated in file and member block scopes.
+  ///
+  /// ### Example
+  /// ```swift
+  /// class X {
+  ///   let a = 42
+  ///
+  ///   func (a: Int) {
+  ///     let a = 123
+  ///
+  ///     a // <-- lookup here
+  ///   }
+  /// }
+  /// ```
+  /// When looking up at the specified position with `includeMembers`
+  /// set to `true`, lookup will return declaration from inside function body,
+  /// function parameter and the `a` declaration from `class X` member block.
+  /// If `includeMembers` would be set to `false`, the latter name would be omitted.
   @_spi(Experimental) public var includeMembers: Bool
 
   /// Creates a new lookup configuration.
--- a/Sources/SwiftLexicalLookup/Configurations/LookupConfig.swift
+++ b/Sources/SwiftLexicalLookup/Configurations/LookupConfig.swift
@@ -13,14 +13,26 @@
 @_spi(Experimental) public struct LookupConfig {
   /// Specifies behavior of file scope.
   @_spi(Experimental) public var fileScopeHandling: FileScopeHandlingConfig
+  /// Specifies whether lookup should finish in the closest sequential scope.
+  @_spi(Experimental) public var finishInSequentialScope: Bool
+  /// Specifies whether to include results generated in file and member block scopes.
+  @_spi(Experimental) public var includeMembers: Bool
 
   /// Creates a new lookup configuration.
   ///
   /// - `fileScopeHandling` - specifies behavior of file scope.
   ///   `memberBlockUpToLastDecl` by default.
+  /// - `finishInSequentialScope` - specifies whether lookup should finish
+  ///   in the closest sequential scope. `false` by default.
+  /// - `includeMembers` - specifies whether to include results generated
+  ///   in file and member block scopes. `true` by default.
   @_spi(Experimental) public init(
-    fileScopeHandling: FileScopeHandlingConfig = .memberBlockUpToLastDecl
+    fileScopeHandling: FileScopeHandlingConfig = .memberBlockUpToLastDecl,
+    finishInSequentialScope: Bool = false,
+    includeMembers: Bool = true
   ) {
     self.fileScopeHandling = fileScopeHandling
+    self.finishInSequentialScope = finishInSequentialScope
+    self.includeMembers = includeMembers
   }
 }
--- a/Sources/SwiftLexicalLookup/LookupName.swift
+++ b/Sources/SwiftLexicalLookup/LookupName.swift
@@ -16,8 +16,8 @@ import SwiftSyntax
 @_spi(Experimental) public enum ImplicitDecl {
   /// `self` keyword representing object instance.
   /// Could be associated with type declaration, extension,
-  /// or closure captures.
-  case `self`(DeclSyntaxProtocol)
+  /// or closure captures. Introduced at function edge.
+  case `self`(FunctionDeclSyntax)
   /// `Self` keyword representing object type.
   /// Could be associated with type declaration or extension.
   case `Self`(DeclSyntaxProtocol)
@@ -135,6 +135,35 @@ import SwiftSyntax
     }
   }
 
+  /// Position of this name.
+  /// 
+  /// For some syntax nodes, their position doesn't reflect
+  /// the position at which a particular name was introduced at.
+  /// Such cases are function parameters (as they can
+  /// contain two identifiers) and function declarations (where name
+  /// is precided by access modifiers and `func` keyword).
+  @_spi(Experimental) public var position: AbsolutePosition {
+    switch self {
+    case .identifier(let syntax, _):
+      switch Syntax(syntax).as(SyntaxEnum.self) {
+      case .functionParameter(let functionParameter):
+        return functionParameter.secondName?.positionAfterSkippingLeadingTrivia
+          ?? functionParameter.firstName.positionAfterSkippingLeadingTrivia
+      default:
+        return syntax.position
+      }
+    case .declaration(let syntax):
+      return syntax.name.position
+    case .implicit(let implicitName):
+      switch implicitName {
+      case .self(let functionDecl):
+        return functionDecl.name.position
+      default:
+        return implicitName.syntax.position
+      }
+    }
+  }
+
   /// Point, after which the name is available in scope.
   /// If set to `nil`, the name is available at any point in scope.
   var accessibleAfter: AbsolutePosition? {
@@ -224,5 +253,28 @@ import SwiftSyntax
     accessibleAfter: AbsolutePosition? = nil
   ) -> [LookupName] {
     [.declaration(namedDecl)]
+  }
+
+  /// Debug description of this lookup name.
+  @_spi(Experimental) public var debugDescription: String {
+    let sourceLocationConverter = SourceLocationConverter(fileName: "", tree: syntax.root)
+    let location = sourceLocationConverter.location(for: position)
+    let strName = (identifier != nil ? identifier!.name : "NO-NAME") + " at: \(location.line):\(location.column)"
+
+    switch self {
+    case .identifier:
+      let str = "identifier: \(strName)"
+
+      if let accessibleAfter {
+        let location = sourceLocationConverter.location(for: accessibleAfter)
+        return str + " after: \(location.line):\(location.column)"
+      } else {
+        return str
+      }
+    case .declaration:
+      return "declaration: \(strName)"
+    case .implicit:
+      return "implicit: \(strName)"
+    }
   }
 }
--- a/Sources/SwiftLexicalLookup/LookupName.swift
+++ b/Sources/SwiftLexicalLookup/LookupName.swift
@@ -154,8 +154,20 @@ import SwiftSyntax
         switch Syntax(declSyntax).as(SyntaxEnum.self) {
         case .functionDecl(let functionDecl):
           return functionDecl.name.position
+        case .initializerDecl(let initializerDecl):
+          return initializerDecl.initKeyword.positionAfterSkippingLeadingTrivia
         case .subscriptDecl(let subscriptDecl):
           return subscriptDecl.accessorBlock?.position ?? subscriptDecl.endPosition
+        case .variableDecl(let variableDecl):
+          return variableDecl.bindings.first?.accessorBlock?.positionAfterSkippingLeadingTrivia
+            ?? variableDecl.endPosition
+        default:
+          return declSyntax.positionAfterSkippingLeadingTrivia
+        }
+      case .Self(let declSyntax):
+        switch Syntax(declSyntax).as(SyntaxEnum.self) {
+        case .protocolDecl(let protocolDecl):
+          return protocolDecl.name.positionAfterSkippingLeadingTrivia
         default:
           return declSyntax.positionAfterSkippingLeadingTrivia
         }
@@ -226,6 +238,8 @@ import SwiftSyntax
       return functionCallExpr.arguments.flatMap { argument in
         getNames(from: argument.expression, accessibleAfter: accessibleAfter)
       }
+    case .optionalChainingExpr(let optionalChainingExpr):
+      return getNames(from: optionalChainingExpr.expression, accessibleAfter: accessibleAfter)
     default:
       if let namedDecl = Syntax(syntax).asProtocol(SyntaxProtocol.self) as? NamedDeclSyntax {
         return handle(namedDecl: namedDecl, accessibleAfter: accessibleAfter)
--- a/Sources/SwiftLexicalLookup/LookupName.swift
+++ b/Sources/SwiftLexicalLookup/LookupName.swift
@@ -20,7 +20,7 @@ import SwiftSyntax
   case `self`(DeclSyntaxProtocol)
   /// `Self` keyword representing object type.
   /// Could be associated with type declaration or extension.
-  case `Self`(ProtocolDeclSyntax)
+  case `Self`(DeclSyntaxProtocol)
   /// `error` value caught by a `catch`
   /// block that does not specify a catch pattern.
   case error(CatchClauseSyntax)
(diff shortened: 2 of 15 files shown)


MAJKFL:
With the recent changes in this PR, `SwiftLexicalLookup` actually matches results of ASTScope unqualified lookup for almost all source files of `SwiftIfConfig` when compared with [this](https://github.com/MAJKFL/swift/blob/swift-lexical-lookup/lib/ASTGen/Sources/ASTGen/LexicalLookup.swift) validation function! The only exception is handling of `$0` in closures.
```swift
public var description: String {
  return components.map { String($0) }.joined(separator: ".")
}
```
As for now, `SwiftLexicalLookup` is missing the representation of almost visible names in scope. Additionally the [validation method](https://github.com/MAJKFL/swift/blob/swift-lexical-lookup/lib/ASTGen/Sources/ASTGen/LexicalLookup.swift) used for comparison accounts for some ASTScope behavior:
- It skips the same (generic parameter) names introduced more than once by ASTScope.
- It skips variable declaration names visible inside of their declarations.
- It reorders results returned from generic parameter clauses of function declarations for consistency.

It might be necessary to revisit some of those cases in the future, but I agreed with @DougGregor to ignore them for now.

Here’s a sample output of properly matched results for one of the references from `ConfiguredRegions.swift`:
```
-----> Lookup started at: 215:19 ("node")
     |      ASTScope      | SwiftLexicalLookup 
> ✅ |  outerState 214:9  |  outerState 214:9  
> ✅ |syntaxErrorsAllowed 213:9|syntaxErrorsAllowed 213:9
> ✅ | foundActive 212:9  | foundActive 212:9  
> ✅ |priorInAnyIfConfig 205:9|priorInAnyIfConfig 205:9
> ✅ |End here: node 203:25|    node 203:25     
> ⏩ |       -----        |    self 203:17     
> ⏩ |       -----        |Configuration 177:43
> ⏩ |       -----        | Look memb: 177:19  
```

DougGregor:
@swift-ci please test

DougGregor:
@MAJKFL looks like you need to run the format script

Sources/SwiftLexicalLookup/Configurations/LookupConfig.swift
```diff
@@ -13,14 +13,26 @@
 @_spi(Experimental) public struct LookupConfig {
   /// Specifies behavior of file scope.
   @_spi(Experimental) public var fileScopeHandling: FileScopeHandlingConfig
+  /// Specifies whether lookup should finish in the closest sequential scope.
+  @_spi(Experimental) public var finishInSequentialScope: Bool
+  /// Specifies whether to include results generated in file and member block scopes.
+  @_spi(Experimental) public var includeMembers: Bool
```
ahoppen, line 19: I don’t know if you have examples that do set these to `true` but if you do, I think they would help readers understand what the options are used for.

Sources/SwiftLexicalLookup/LookupName.swift
```diff
@@ -135,6 +135,50 @@ import SwiftSyntax
...
+          return functionDecl.name.position
+        case .initializerDecl(let initializerDecl):
+          return initializerDecl.initKeyword.positionAfterSkippingLeadingTrivia
+        case .subscriptDecl(let subscriptDecl):
+          return subscriptDecl.accessorBlock?.position ?? subscriptDecl.endPosition
+        case .variableDecl(let variableDecl):
+          return variableDecl.bindings.first?.accessorBlock?.positionAfterSkippingLeadingTrivia
+            ?? variableDecl.endPosition
+        default:
+          return declSyntax.positionAfterSkippingLeadingTrivia
```
ahoppen, line 165: Not that it matters much but some cases use `position` and some `positionAfterSkippingLeadingTrivia`. That makes me wonder why there’s a difference. If there is no meaningful distinction, I would suggest that they all use the same, maybe defaulting to `position` because it doesn’t need to add any trivia length.

Sources/SwiftLexicalLookup/LookupName.swift
```diff
@@ -135,6 +135,50 @@ import SwiftSyntax
...
+        }
+      case .Self(let declSyntax):
+        switch Syntax(declSyntax).as(SyntaxEnum.self) {
+        case .protocolDecl(let protocolDecl):
+          return protocolDecl.name.positionAfterSkippingLeadingTrivia
```
ahoppen, line 170: Why do we only need this for protocols and not other type declarations?

Sources/SwiftLexicalLookup/LookupName.swift
```diff
@@ -225,4 +266,27 @@ import SwiftSyntax
   ) -> [LookupName] {
     [.declaration(namedDecl)]
   }
+
+  /// Debug description of this lookup name.
+  @_spi(Experimental) public var debugDescription: String {
+    let sourceLocationConverter = SourceLocationConverter(fileName: "", tree: syntax.root)
+    let location = sourceLocationConverter.location(for: position)
+    let strName = (identifier != nil ? identifier!.name : "NO-NAME") + " at: \(location.line):\(location.column)"
```
ahoppen, line 274: How about using `??` and string interpolation

```suggestion
    let strName = "\(identifier?.name ?? "NO-NAME") at: \(location.line):\(location.column)"
```

Sources/SwiftLexicalLookup/LookupResult.swift
```diff
@@ -46,4 +52,54 @@ import SwiftSyntax
...
+@_spi(Experimental) extension [LookupResult] {
+  /// Debug description this array of lookup results.
+  @_spi(Experimental) public var debugDescription: String {
+    var str: String = ""
+
+    for (index, result) in self.enumerated() {
+      str += result.debugDescription + (index + 1 == self.count ? "" : "\n")
+    }
+
+    return str
```
ahoppen, line 103: You should be able to use `joined(separator: )` for this

```suggestion
    return self.map(\.debugDescription).joined(separator: "\n")
```

Sources/SwiftLexicalLookup/Scopes/ScopeImplementations.swift
```diff
@@ -348,27 +459,107 @@ import SwiftSyntax
...
       switch accessorSpecifier.tokenKind {
       case .keyword(.set), .keyword(.willSet):
-        return [.implicit(.newValue(self))]
+        names = [.implicit(.newValue(self))]
       case .keyword(.didSet):
-        return [.implicit(.oldValue(self))]
+        names = [.implicit(.oldValue(self))]
       default:
-        return []
+        names = []
```
ahoppen, line 522: Any reason why you changed this? I liked the version that directly returned better 😉

Sources/SwiftLexicalLookup/Scopes/ScopeImplementations.swift
```diff
@@ -348,27 +459,107 @@ import SwiftSyntax
...
+    with config: LookupConfig
+  ) -> [LookupResult] {
+    guard let parentAccessorBlockScope = parentScope?.as(AccessorBlockSyntax.self) else {
+      return defaultLookupImplementation(identifier, at: lookUpPosition, with: config)
+    }
+
+    let implicitSelf: [LookupName] = [.implicit(.self(self))]
+      .filter { name in
+        checkIdentifier(identifier, refersTo: name, at: lookUpPosition)
+      }
```
ahoppen, line 549: Should we only add `self` if this is a member? Ie. we shouldn’t add `self` in a top level variable with an accessor, right?

ahoppen (commented):
Left a few comments of things I noticed while reading through the change. 

Super exciting to see this progressing and seeing you continue working on it after the GSoC project is over, @MAJKFL ❤️

Sources/SwiftLexicalLookup/Configurations/LookupConfig.swift
MAJKFL: I added some examples to the documentation comments of those new configurations. 

I think `includeMembers` is particularly interesting because if we just consider the compiler, it always excludes members and file scope classes in unqualified lookup. I suppose those are handled by a separate member lookup. 

Should we then consider members in `SwiftLexicalLookup` and allow this sort of „semi-member lookup” (as modeled by this `includeMembers` config) or should we just skip them and don’t provide this functionality of ASTScope unqualified lookup? I think I'd also like to hear @DougGregor opinion on this one too.

Sources/SwiftLexicalLookup/LookupName.swift
MAJKFL: As I tested new lookup with different source code inputs, I’ve discovered that using `positionAfterSkippingLeadingTrivia` over `position` did fix discrepancies in some cases. This is an example of not matching results from `ActiveSyntaxRewriter.swift` after I changed `initializerDeclSyntax` to use `position` for `self` keyword in this `switch`.

```swift
class ActiveSyntaxRewriter: SyntaxRewriter {
  var retainFeatureCheckIfConfigs: Bool // <— line 153

  init(configuredRegions: ConfiguredRegions, retainFeatureCheckIfConfigs: Bool) { // <— line 155
    self.activeClauses = .configuredRegions(configuredRegions) 
    self.diagnostics = activeClauses.priorDiagnostics
    self.retainFeatureCheckIfConfigs = retainFeatureCheckIfConfigs
  }
}
```

```
-----> Lookup started at: 156:5 ("self")
     |      ASTScope      | SwiftLexicalLookup 
> ✅ |configuredRegions 155:8|configuredRegions 155:8
> ✅ |retainFeatureCheckIfConfigs 155:46|retainFeatureCheckIfConfigs 155:46
> ⚠️ |End here: self 155:3|    self 153:40     
> ⏩ |       -----        |  Look memb: 148:7  
```

Do you think there could be some better way to fix this, or should we maybe start using `positionAfterSkippingLeadingTrivia`  consistently?

Sources/SwiftLexicalLookup/LookupName.swift
MAJKFL: In this PR, protocol declarations are the only places we are introducing `Self` keyword, so handling other declarations wouldn’t matter in this case. I changed associated syntax node type of `ImplicitDecl.Self` to `ProtocolDeclSyntax`, to be more explicit and remove this nested `switch`.

Sources/SwiftLexicalLookup/Scopes/ScopeImplementations.swift
MAJKFL: Oh, before properly handling `self` keyword in accessors I implemented it there in `introducedNames`. I must’ve forgotten to change it back. Thanks for noticing! Reverted that.

Sources/SwiftLexicalLookup/Scopes/ScopeImplementations.swift
MAJKFL: This case was handled here implicitly. The result with `self` was handed over to a special function that passed it to `AccessorBlockSyntax` parent which later handed it off to it’s `SubscriptDeclSyntax` parent. 

I’ve noticed now this solution wasn’t perfect though. It was quite inflexible and didn’t work for variable declaration accessors. To fix this, I introduced now a new specialized scope kind: [`CanInterleaveResultsLaterScopeSyntax`](https://github.com/swiftlang/swift-syntax/blob/433f14bbd85810efd88954e0cb62c3b655ca7037/Sources/SwiftLexicalLookup/Scopes/CanInterleaveResultsLaterScopeSyntax.swift#L15-L22) with `lookupWithInterleavedResults` function to further generalize this behavior. [Now accessor can hand the result with `self` keyword to it’s `CanInterleaveResultsLaterScopeSyntax` parent](https://github.com/swiftlang/swift-syntax/blob/433f14bbd85810efd88954e0cb62c3b655ca7037/Sources/SwiftLexicalLookup/Scopes/ScopeImplementations.swift#L556-L561). The `self` keyword is introduced properly by accessors inside `subscript` and variables and it does differentiate whether it’s a member or not.

Sources/SwiftLexicalLookup/LookupName.swift
ahoppen: Oh, I see. 

`positionAfterSkippingLeadingTrivia` probably matches the AST’s `position` more closely because the AST does not care about trivia. But I think outside of your validation tooling `position` and `positionAfterSkippingLeadingTrivia` should produce the same lookup results, right? Or can you think of cases where the two produce fundamentally different results?

Anyway, I would suggest to use `positionAfterSkippingLeadingTrivia` consistently for now because it’s also more natural. You really need to think twice about trivia handling to figure out why `self` is declared at `153:40` in the above example, should we use `position`.

Sources/SwiftLexicalLookup/LookupName.swift
ahoppen: So, you’re not handling `Self` in classes yet?

```swift
class Foo {
  func bar() -> Self { fatalError() }
}
```

Sources/SwiftLexicalLookup/LookupName.swift
MAJKFL: I agree, there's no fundamental difference between using `position` over `positionAfterSkippingLeadingTrivia`. I changed to `positionAfterSkippingLeadingTrivia` for consistency.

Sources/SwiftLexicalLookup/LookupName.swift
MAJKFL: Interestingly, it seems to me like `ASTScope` doesn’t introduce implicit `Self` in nominal type declarations. I suppose this could be implicitly modeled by the `lookInMembers` result? Here’s the entire output of the validation function for the snippet you’ve provided
```
-----> Lookup started at: 2:17 ("Self")
     |      ASTScope      | SwiftLexicalLookup 
> ✅ |   Look memb: 1:7   |   Look memb: 1:7   

-----> Lookup started at: 2:17 ("Self")
     |      ASTScope      | SwiftLexicalLookup 
> ✅ |   Look memb: 1:7   |   Look memb: 1:7   

-----> Lookup started at: 2:24 ("fatalError")
     |      ASTScope      | SwiftLexicalLookup 
> ✅ |      self 2:8      |      self 2:8      
> ✅ |End here: Look memb: 1:7|   Look memb: 1:7   

-----> Lookup started at: 2:24 ("fatalError")
     |      ASTScope      | SwiftLexicalLookup 
> ✅ |      self 2:8      |      self 2:8      
> ✅ |   Look memb: 1:7   |   Look memb: 1:7   
```
Nevertheless, there certainly are more places `Self` is introduced apart from protocol declarations. This seems to be also the case for extension declarations. I fixed it now as well and this is how the output looks like for the following snippet (you can also notice here the duplicate omission I’ve mentioned in my [previous comment](https://github.com/swiftlang/swift-syntax/pull/2852#issuecomment-2360959157))
```swift
protocol Foo {}
extension Foo {
  func bar() -> Self { fatalError() }
}
```
```
-----> Lookup started at: 2:11 ("Foo")
     |      ASTScope      | SwiftLexicalLookup 

-----> Lookup started at: 2:11 ("Foo")
     |      ASTScope      | SwiftLexicalLookup 

-----> Lookup started at: 3:17 ("Self")
     |      ASTScope      | SwiftLexicalLookup 
> ✅ | End here: Self 2:1 |      Self 2:1      
> ⏩ |       -----        |  Look memb: 2:11   

-----> Lookup started at: 3:24 ("fatalError")
     |      ASTScope      | SwiftLexicalLookup 
> ✅ |      self 3:8      |      self 3:8      
> ✅ |      Self 2:1      |      Self 2:1      
> ✅ |End here: Look memb: 2:11|  Look memb: 2:11   

-----> Lookup started at: 3:24 ("fatalError")
     |      ASTScope      | SwiftLexicalLookup 
> ✅ |      self 3:8      |      self 3:8      
> ✅ |      Self 2:1      |      Self 2:1      
> ✅ |  Look memb: 2:11   |  Look memb: 2:11   
> ℹ️ | Omitted ASTScope name: Self 2:1
> ℹ️ | Omitted ASTScope name: Self 2:1
```

Sources/SwiftLexicalLookup/Scopes/CanInterleaveResultsLaterScopeSyntax.swift
```diff
@@ -12,14 +12,11 @@
...
-
-extension TypeScopeSyntax {
-  @_spi(Experimental) public var implicitInstanceAndTypeNames: [LookupName] {
-    [.implicit(.self(self)), .implicit(.Self(self))]
-  }
-
-  @_spi(Experimental) public var introducedNames: [LookupName] {
-    implicitInstanceAndTypeNames
-  }
+protocol CanInterleaveResultsLaterScopeSyntax: ScopeSyntax {
```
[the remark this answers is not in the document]
MAJKFL: Thanks for pointing this out! I added documentation comments to the definition of `lookupWithInterleavedResults` function and it's concrete implementations.

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test Windows

Sources/SwiftLexicalLookup/Scopes/ScopeImplementations.swift
```diff
@@ -665,13 +671,16 @@ import SwiftSyntax
       )
     }
 
+    let lookInMembers: [LookupResult] =
+      inheritanceClause?.range.contains(lookUpPosition) ?? false ? [] : [.lookInMembers(self)]
```
ahoppen, line 675: Instead of blacklisting the inheritance clause, wouldn’t it be better to whitelist the `memberBlock`? I got to this because I was thinking that `self` shouldn’t be available in the generic parameters of the type.

DougGregor:
macOS failure is a formatting issue:

```
syntax/Tests/SwiftLexicalLookupTest/NameLookupTests.swift:897:34: warning: [EndOfLineComment] move end-of-line comment that exceeds the line length
```

DougGregor:
@swift-ci please build toolchain

MAJKFL:
Oh, so the comments were the problem, so that's why running formatting script didn't help. Thanks for pointing this out! Hopefully should be fixed now.

Sources/SwiftLexicalLookup/Scopes/ScopeImplementations.swift
MAJKFL: That's a good point. As for now, I couldn't find a specific case this wouldn't work and it definitely improves readability of this part of code as well. I changed it to check if lookup was started inside `memberBlock`.

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci test macOS

MAJKFL:
@swift-ci please test

MAJKFL:
Seems like everything passed now, do you think @ahoppen it's good to merge?

ahoppen (approved):

PR #2856, opened 2024-09-19
[CMake] Enable CMP0157 NEW if available

rintaro:
Use new CMake Swift compilation mode if possible.
Disable sowe workaround when CMP0157 is enabled.


rintaro:
https://github.com/swiftlang/swift/pull/76587
@swift-ci Please test

bnbarham (approved):

PR #2857, opened 2024-09-22
Parse raw identifiers.

allevato:
The parser implementation of [SE-0451](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0451-escaped-identifiers.md).

### DIFF
--- a/Sources/SwiftParser/Lexer/Cursor.swift
+++ b/Sources/SwiftParser/Lexer/Cursor.swift
@@ -2025,14 +2025,57 @@ extension Lexer.Cursor {
     // Check whether we have an identifier followed by another backtick, in which
     // case this is an escaped identifier.
     let identifierStart = self
-    if self.advance(if: { $0.isValidIdentifierStartCodePoint }) {
-      // Keep continuing the identifier.
-      self.advance(while: { $0.isValidIdentifierContinuationCodePoint })
 
-      // If we have the terminating "`", it's an escaped identifier.
-      if self.advance(matching: "`") {
-        return Lexer.Result(.identifier)
+    // Scan until we see either a closing backtick or the end of the line. Do
+    // additional validation for raw identifiers along the way; if we see
+    // characters that aren't allowed (prohibited whitespace or unprintable
+    // ASCII characters) or if the identifier is an operator, provide a more
+    // precise diagnostic and location, but otherwise keep trying to tokenize
+    // it as a raw identifier as long as we see the closing backtick because
+    // it more likely represents what the user was trying to do.
+    var hasNonOperatorCharacter = false
+    var hasNonWhitespaceCharacter = false
+    var isEmpty = true
+    var error: LexingDiagnostic? = nil
+    while true {
+      let ch = self.peek()
+      if ch == nil || ch == "`" || ch == "\n" || ch == "\r" {
+        break
       }
+      let position = self
+      guard let scalar = self.advanceValidatingUTF8Character() else {
+        error = LexingDiagnostic(.invalidUtf8, position: position)
+        continue
+      }
+      if error == nil {
+        if scalar.isForbiddenRawIdentifierWhitespace {
+          error = LexingDiagnostic(.invalidWhitespaceInRawIdentifier, position: position)
+        } else if scalar == "\\" {
+          error = LexingDiagnostic(.invalidBackslashInRawIdentifier, position: position)
+        } else if scalar.isASCII && !scalar.isPrintableASCII {
+          error = LexingDiagnostic(.unprintableAsciiCharacter, position: position)
+        }
+      }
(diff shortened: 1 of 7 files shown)


Sources/SwiftParser/Lexer/Cursor.swift
```diff
@@ -2025,14 +2025,51 @@ extension Lexer.Cursor {
...
+      if error == nil {
+        if scalar.isForbiddenRawIdentifierWhitespace {
+          error = LexingDiagnostic(.invalidWhitespaceInRawIdentifier, position: position)
+        } else if scalar == "\\" {
+          error = LexingDiagnostic(.invalidBackslashInRawIdentifier, position: position)
+        } else if scalar.isASCII && !scalar.isPrintableASCII {
+          error = LexingDiagnostic(.unprintableAsciiCharacter, position: position)
+        }
+      }
+      if isEmpty && !scalar.isOperatorStartCodePoint || !scalar.isOperatorContinuationCodePoint {
```
ahoppen, line 2058: I think adding parenthesis here would help readability.

```suggestion
      if (isEmpty && !scalar.isOperatorStartCodePoint) || !scalar.isOperatorContinuationCodePoint {
```

Tests/SwiftParserTest/translated/EscapedIdentifiersTests.swift
```diff
@@ -99,4 +99,201 @@ final class EscapedIdentifiersTests: ParserTestCase {
...
+      1️⃣`` = 5
+      """,
+      diagnostics: [
+        DiagnosticSpec(
+          locationMarker: "1️⃣",
+          message: "a raw identifier cannot be empty"
+        )
+      ]
+    )
+  }
```
ahoppen, line 250: Do we diagnose raw identifiers that only consist of whitespace? As far as I can tell, they are forbidden by the proposal.

allevato: We weren't; thanks for catching that. Fixed.

grynspan:
I _think_ I have a hard need for determining if an identifier wrapped in backticks is raw or just an escaped keyword. Having a member such as `TokenKind.isRawIdentifier` would be very useful here.

Sources/SwiftParser/Lexer/UnicodeScalarExtensions.swift
```diff
@@ -93,6 +93,31 @@ extension Unicode.Scalar {
...
   }
 
+  var isForbiddenRawIdentifierWhitespace: Bool {
+    let c = self.value
+    // This is the set of code points satisfying the `White_Space` property,
+    // excluding the set satisfying the `Pattern_White_Space` property, and
+    // excluding any other ASCII non-printables and Unicode separators. In
+    // other words, the only whitespace code points allowed in a raw
+    // identifier are U+0020, and U+200E/200F (LTR/RTL marks).
+    return (c >= 0x0009 && c <= 0x000D) as Bool
```
grynspan, line 103: Would it be better to express this as `isWhitespace && !isPermittedRawIdentifierWhitespace`?

allevato: The parser explicitly avoids the use of `Unicode.Scalar.Properties` because they're not necessarily stable between Unicode versions and may change depending on the Unicode tables that are built into `libswiftCore`, so we don't want the behavior of the parser to change depending on things like what version of the operating system the compiler is running on.

It's probably _exceedingly rare_ that new whitespace will be added (or worse, that a non-whitespace code unit would become whitespace or vice versa), but hardcoding the code points means the behavior is deterministic and we can choose to make breaks via new language modes, if needed in the future.

allevato:
> I _think_ I have a hard need for determining if an identifier wrapped in backticks is raw or just an escaped keyword. Having a member such as `TokenKind.isRawIdentifier` would be very useful here.

I think you can do this today, in a somewhat roundabout fashion, using [existing APIs](https://github.com/swiftlang/swift-syntax/blob/main/Sources/SwiftParser/IsValidIdentifier.swift) from `SwiftParser`:

```
let id = "<some keyword or raw identifier without backticks>"
let quotedID = "`\(id)`"
let isKeywordish = id.isValidSwiftIdentifier(for: .memberAccess)
  && quotedID.isValidSwiftIdentifier(for: .memberAccess)
```

...since `x.class` and ``x.`class` `` are both valid for escaped keywords but `x.some raw identifier` and ``x.`some raw identifier` `` would fail the first test. There may be some failure cases I'm missing here, though...

It's not the most efficient though since it kicks off two parsers. I agree it's probably a good idea to have a separate API for this that does the same single-pass over the identifier that the lexer does.

ahoppen (approved):

allevato:
@swift-ci please test

PR #2858, opened 2024-09-23
Don’t add an implicit leading newline in `CodeBlockItemList` result builder if previous element already has a trailing newline

ahoppen:


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #2861, opened 2024-09-25
[6.0] Allow building swift-syntax in C++ interop mode

ahoppen:
- **Explanation**: If swift-syntax is being built with `-cxx-interoperability-mode=default`, it fails because of `cannot initialize a variable of type 'AtomicBool *' with an rvalue of type 'void *'`. Add a cast here to fix the build issue.
- **Scope**: Build of swift-syntax in C++ interop mode
- **Risk**: Very low, just adding an explicit cast from `void *` to `AtomicBool *`
- **Testing**: n/a
- **Issue**: rdar://129252735
- **Reviewer**:   @bnbarham on https://github.com/swiftlang/swift-syntax/pull/2752


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #2865, opened 2024-09-30
Silence `switch covers known cases` warnings when building with resilience

tshortli:


tshortli:
@swift-ci please smoke test

Sources/SwiftCompilerPluginMessageHandling/Macros.swift
```diff
@@ -167,6 +167,7 @@ extension PluginProviderMessageHandler {
         switch macroDefinition.formatMode {
         case .auto: collapseIndentationWidth = nil
         case .disabled: collapseIndentationWidth = []
+        @unknown default: fatalError()
```
ahoppen, line 170:
```suggestion
        #if RESILIENT_LIBRARIES
        @unknown default: fatalError()
        #endif
```

Same for the other change.

tshortli: Ah right, I forgot I added that. FWIW, the Swift 6 compiler will no longer warning about this code when building non-resiliently (the warning is suppressed now when the default is `@unknown`) so depending on how we expect people to be developing swift-syntax we could potentially drop the verbosity of the `RESILIENT_LIBRARIES` guards.

ahoppen: We still support building swift-syntax with 5.8.

ahoppen: I would like to keep `RESILIENT_LIBRARIES` for now so that swift-syntax can build without warnings with a Swift 5 compiler.

tshortli:
@swift-ci please smoke test

tshortli:
@swift-ci please test

tshortli:
@swift-ci please smoke test macOS

tshortli:
@swift-ci please smoke test

bnbarham:
@swift-ci please test

PR #2869, opened 2024-10-02
Re-enable three swift-format rules.

allevato:
This PR re-enables the following three rules, with the associated fixes:

* `UseSynthesizedInitializer`
* `NoBlockComments`
* `UseLetInEveryBoundCaseVariable`


Sources/SwiftCompilerPluginMessageHandling/PluginMacroExpansionContext.swift
```diff
@@ -129,7 +129,7 @@ class SourceManager {
 
     var node = syntaxRegistry.get(source: syntaxInfo.source, kind: syntaxInfo.kind)
     if let operatorTable {
-      node = operatorTable.foldAll(node, errorHandler: { _ in /*ignore*/ })
+      node = operatorTable.foldAll(node, errorHandler: { _ in })  // ignore
```
ahoppen, line 132: I think this is more readable with the block comment, same for the other `ignore` comments.

Sources/SwiftLibraryPluginProvider/LibraryPluginProvider.swift
```diff
@@ -151,7 +151,11 @@ private func _findAnyType(_ moduleName: String, _ typeName: String) -> Any.Type?
   // actual symbol name doesn't match with it. i.e. We don't need to perform
   // punycode encodings or word substitutions.
   // FIXME: This is process global. Can we limit it to a specific .dylib ?
-  for suffix in [ /*struct*/"V", /*enum*/ "O", /*class*/ "C"] {
+  for suffix in [
+    "V",  // struct
+    "O",  // enum
+    "C",  // class
+  ] {
```
ahoppen, line 158: Similarly here, I think disallowing block comments decreases readability.

Sources/SwiftSyntax/SyntaxArena.swift
```diff
@@ -202,7 +202,7 @@ public class ParsingSyntaxArena: SyntaxArena {
   public func internSourceBuffer(_ buffer: UnsafeBufferPointer<UInt8>) -> UnsafeBufferPointer<UInt8> {
     let allocated = allocator.allocate(
       UInt8.self,
-      count: buffer.count + /* for NULL */ 1
+      count: buffer.count + 1  // +1 for NULL
```
ahoppen, line 205: And here I think the block comment also makes sense.

ahoppen (commented):
- `UseSynthesizedInitializer`: Looks good to me
- `NoBlockComments`: I highlighted the cases where I think that block comments are actually the right choice. I’m mildly leaning towards keeping that rule off. Pretty impartial to migrating the block comments that I didn’t comment on to line comments.
- `UseLetInEveryBoundCaseVariable`: I really don’t have an opinion here, happy with it on or off

Sources/SwiftCompilerPluginMessageHandling/PluginMacroExpansionContext.swift
Matejkob: I totally agree. I think in most cases if the comment block is placed between nodes, it shouldn’t be moved to the end?

.swift-format
```diff
@@ -10,9 +10,9 @@
     "rules": {
         "AlwaysUseLowerCamelCase": false,
         "AmbiguousTrailingClosureOverload": false,
-        "NoBlockComments": false,
+        "NoBlockComments": true,
         "OrderedImports": true,
-        "UseLetInEveryBoundCaseVariable": false,
-        "UseSynthesizedInitializer": false
+        "UseLetInEveryBoundCaseVariable": true,
+        "UseSynthesizedInitializer": true
```
Matejkob, line 16: I’ve noticed that @ahoppen mentioned not having a strong opinion here, so I’d like to share my perspective. In my projects, when a case in a closure has multiple parameters, and I want to designate one parameter as a variable (var) and the others as constants (let), applying this rule makes sense since you can see what is mutable and what's not–don't know if this statment makes sense to u XD hope so. 

Code example:
```swift
enum Foo {
  case bar(name: String: age: Int)
}

let foo = Foo.bar(name: "Tom", age: 22)

if case .bar(let name, var age) = foo {
  age += 1
  // do sth with increased age and not mutated name? 
}
```

Matejkob (commented):
Love the clear git history — the commits are awesome to review <3

.swift-format
allevato: This is one of the reasons I prefer `UseLetInEveryBoundCaseVariable` as well. Wrapping the whole pattern with `let/var` forces you into an all-or-nothing situation with the bindings. Once we have other kinds of bindings like `borrowing`, `inout`, etc., I think it'll be even more common for those to differ within a payload.

The other reason for this rule is that having the binding specifier immediately adjacent to the binding makes it easier to see what the intended outcome is, and avoids mistakes like this:

```swift
let age = 20
if case let .bar(name, age) = foo { ... }
// oops I wanted to bind `name` and match `age`, not create a
// new `age` binding

if case .bar(let name, age) = foo { ... }
// much clearer about which one is a new binding and which
// is an expression to be matched
```

allevato:
> * `NoBlockComments`: I highlighted the cases where I think that block comments are actually the right choice. I’m mildly leaning towards keeping that rule off. Pretty impartial to migrating the block comments that I didn’t comment on to line comments.

After making the actual changes, I tend to agree for these cases. I've dropped that commit from the PR.

I think it reveals a potential improvement to the `NoBlockComments` rule: it could be refined to diagnose multiline block comments and block comments that are entirely on their own line, but not small inline comments between other tokens.

.swift-format
```diff
@@ -12,7 +12,7 @@
         "AmbiguousTrailingClosureOverload": false,
         "NoBlockComments": false,
         "OrderedImports": true,
-        "UseLetInEveryBoundCaseVariable": false,
-        "UseSynthesizedInitializer": false
+        "UseLetInEveryBoundCaseVariable": true,
+        "UseSynthesizedInitializer": true
```
ahoppen, line 16: These are `true` by default, right? So we should be able to remove them from `.swift-format` to make it shorter.

allevato: That's what I thought as well, but I just noticed that if I remove the line entirely and insert a violation, it isn't flagged.

Oof. I think this is a consequence of the decoding logic; if the `rules` dictionary is present in the JSON, its value is used verbatim. Only if it isn't present at all do we use the default dictionary. Then, when we check if a rule should be applied at a node, if the rule name isn't in the dictionary, it's _always_ false.

We should fix this by always taking the default rule settings and then _merging_ what we read from the JSON into that. That would ensure that the in-memory dictionary always contains all rules. That might reveal some other places in swift-syntax/swift-format where the rules aren't flagging violations, though.

So my thinking is we merge this as-is ("OrderedImports" was already this way) and then fix that for the next release. WDYT?

ahoppen: Sounds good to me. But yes, this seems like a swift-format bug to me to not merge the dictionaries that’s worth fixing. If you don’t have time to look at that now, could you file an issue for it?

ahoppen:
@swift-ci Please test

ahoppen (approved):

allevato:

PR #2871, opened 2024-10-02
Support parsing of key path subscripts on metatypes

ahoppen:
This is needed to support parsing eg.
https://github.com/swiftlang/swift/pull/73242/files#diff-a94f38549dc87ce27c438b255ae2f29c69cc1c2c5e5069f327c5c7fae7bc553dR667

### DIFF
--- a/Sources/SwiftParser/Expressions.swift
+++ b/Sources/SwiftParser/Expressions.swift
@@ -925,14 +925,16 @@ extension Parser {
 
 extension Parser {
   /// Determine if this is a key path postfix operator like ".?!?".
-  private func getNumOptionalKeyPathPostfixComponents(
-    _ tokenText: SyntaxText
-  ) -> Int? {
+  private func getNumOptionalKeyPathPostfixComponents(_ tokenText: SyntaxText, mayBeAfterTypeName: Bool) -> Int? {
+    var mayBeAfterTypeName = mayBeAfterTypeName
     // Make sure every character is ".", "!", or "?", without two "."s in a row.
     var numComponents = 0
     var lastWasDot = false
     for byte in tokenText {
       if byte == UInt8(ascii: ".") {
+        if !mayBeAfterTypeName {
+          break
+        }
         if lastWasDot {
           return nil
         }
@@ -942,6 +944,7 @@ extension Parser {
       }
 
       if byte == UInt8(ascii: "!") || byte == UInt8(ascii: "?") {
+        mayBeAfterTypeName = false
         lastWasDot = false
         numComponents += 1
         continue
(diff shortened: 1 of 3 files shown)


ahoppen:
@swift-ci Please test

Sources/SwiftParser/Types.swift
```diff
@@ -260,7 +259,9 @@ extension Parser {
 
     var loopProgress = LoopProgressCondition()
     while self.hasProgressed(&loopProgress) {
-      if !stopAtFirstPeriod, self.at(.period) {
+      if allowMemberTypes || (self.at(.period) && self.peek(isAt: .keyword(.Type), .keyword(.Protocol))),
```
rintaro, line 262: What about Nested types like: `\Foo.Bar.[0]`?

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Linux

ahoppen:
@swift-ci Please test Linux

Sources/SwiftParser/Expressions.swift
```diff
@@ -997,18 +997,20 @@ extension Parser {
...
     var loopProgress = LoopProgressCondition()
     while self.hasProgressed(&loopProgress) {
-      // Check for a [] or .[] suffix. The latter is only permitted when there
-      // are no components.
+      // Check for a [] or .[] suffix. The latter is only permitted after the type name. Since we don't know what
+      // constitutes a nested type reference and what constitutes a type's member, we can only disallow it in the parser
+      // after seeing the first non-property component.
       if self.at(TokenSpec(.leftSquare, allowAtStartOfLine: false))
-        || (components.isEmpty && self.at(.period) && self.peek(isAt: .leftSquare))
+        || (components.allSatisfy({ $0.component.is(RawKeyPathPropertyComponentSyntax.self) }) && self.at(.period)
```
rintaro, line 1012: I found `\T.?.?` is currently accepted, it also need this check. 

Also, could this `components.allSatisfy()` check a flag outside the loop? Like

ahoppen: Updated to disallow these key paths. I found a fun example along the way that’s valid 😭. This prevents us from diagnosing the `.` as an unexpected token, which would generally be nicer diagnostics.

```swift
struct Foo {
  var bar: Int?
}

infix operator .?.

func .?.(_ x: AnyKeyPath, _ a: Int) {}

var blah = 2
\Foo?.?.bar.?.blah
```

rintaro (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test macOS

PR #2876, opened 2024-10-08
Delete dependsOn support

meg-gupta:


meg-gupta:
@swift-ci test

Sources/SwiftCompilerPluginMessageHandling/CompilerPluginMessageHandler.swift
```diff
@@ -263,7 +263,7 @@ nonisolated(unsafe) private var readabilityHandler: () -> Void = {
   )
 }
 
-@_expose(wasm, "swift_wasm_macro_v1_pump")
+@_expose(wasm,"swift_wasm_macro_v1_pump")
```
ahoppen, line 266: I assume this will cause formatting verification to fail.

Sources/SwiftSyntax/generated/SyntaxCollections.swift
```diff
@@ -1675,113 +1652,13 @@ public struct TupleTypeElementListSyntax: SyntaxCollection, SyntaxHashable {
...
-    /// - Warning: This function will crash if the cast is not possible. Use `as` to safely attempt a cast.
-    /// - Note: Requires experimental feature `nonescapableTypes`.
-    #if compiler(>=5.8)
-    @_spi(ExperimentalLanguageFeatures)
-    #endif
-    public func cast(_ syntaxType: LifetimeTypeSpecifierSyntax.Type) -> LifetimeTypeSpecifierSyntax {
-      return self.as(LifetimeTypeSpecifierSyntax.self)!
-    }
-  }
+  public typealias Element = SimpleTypeSpecifierSyntax
```
ahoppen, line 1661: This is an API-breaking change because clients might be switching over the element enum and using the `simpleTypeSpecifier` case to extract the value. It’s a little unfortunate that we did release swift-syntax 600 with `TypeSpecifierListSyntax.Element` as a `SyntaxChildChoices` and now need to undo it again.

@rintaro @hamishknight @bnbarham Do you think we should take the API breakage again or should we somehow keep `Element: SyntaxChildChoices` even if it only has a single case?

meg-gupta:
@swift-ci test

Sources/SwiftSyntax/generated/SyntaxCollections.swift
ahoppen: @rintaro and I just discussed this and decided that what you did, ie. having a `SyntacChildChoices` with a single element, here is the best choice. It is one less API break and set us up in case we want to add type specifiers with detail in the future.

meg-gupta:
https://github.com/swiftlang/swift/pull/77030
https://github.com/swiftlang/swift-testing/pull/765

@swift-ci smoke test

Sources/SwiftSyntax/generated/SyntaxCollections.swift
ahoppen: Sorry, I mixed up what we wanted and what we have here. `TypeSpecifierListSyntax. Element` should continue being a `SyntaxChildChoices` enum, just with a single case and not change to `SimpleTypeSpecifierSyntax` in order to maintain API compatibility.

meg-gupta:
@swift-ci test

meg-gupta:
@swift-ci test

meg-gupta:
@swift-ci smoke test windows platform

CodeGeneration/Sources/generate-swift-syntax/ChildNodeChoices.swift
```diff
@@ -89,7 +89,9 @@ extension Node {
           return nil
         }
       }
-    } else if let node = self.collectionNode, node.elementChoices.count > 1 {
+    } else if let node = self.collectionNode, node.elementChoices.count == 1, !node.disableSameTypeForUniqueChoice {
+      return []
+    } else if let node = self.collectionNode, node.elementChoices.count >= 1 {
```
ahoppen, line 94: I think this can be simplified to 

```suggestion
    } else if let node = self.collectionNode, (node.elementChoices.count > 1 || node.disableSameTypeForUniqueChoice) {
```

CodeGeneration/Sources/SyntaxSupport/Node.swift
```diff
@@ -55,6 +55,8 @@ public class Node: NodeChoiceConvertible {
   /// function that should be invoked to create this node.
   public let parserFunction: TokenSyntax?
 
+  public let disableSameTypeForUniqueChoice: Bool
```
ahoppen, line 58: Please correct me if I’m wrong but I think we only support this for syntax collections, right? If so, I think it would be a cleaner design to modify [`Node.Data.collection`](https://github.com/swiftlang/swift-syntax/blob/d145cb2a6efb5c8d45cd5df5f494a89c93c90758/CodeGeneration/Sources/SyntaxSupport/Node.swift#L27) to

```swift
fileprivate enum SyntaxCollectionElement {
  /// The syntax collection can only have a single kind of syntax node as its element.
  case singleOption(SyntaxNodeKind)
  /// Create a `SyntaxChildChoices` enum for the collection’s elements.
  case choices([SyntaxNodeKind])
}

case collection(SyntaxCollectionElement)
```

And with that `CollectionNode` can gain

```swift
var generateSyntaxChoicesEnum: Bool {
    switch node.data {
    case .layout:
      preconditionFailure("NodeLayoutView must wrap a Node with data `.collection`")
    case .collection(.singleOption(_)): 
      return false
    case .collection(.choices(_)): 
      return true
    }
} 
```

That way we don’t need to repeat the `> 1 || disableSameTypeForUniqueChoice` check everywhere.

PR #2877, opened 2024-10-09
[SwiftLexicalLookup] Add dollar identifier, better extension handling and more `ASTScope` related fixes

MAJKFL:
This PR brings more `ASTScope` related fixes to `SwiftLexicalLookup`. Those include:
- Prompting clients to lookup generic parameters of extended types inside extensions with `LookupResult.lookInGenericParametersOfExtendedType`.
- Dollar identifier handling through `LookupResult.mightIntroduceDollarIdentifiers` result kind when looking up for `identifier = nil` and more explicit `LookupName.dollarIdentifier` when looking up for a particular dollar identifier.
- Remove member and file scope handling to avoid confusion and incomplete member lookup.
- Fix behavior of lookup from `where` clauses.
- And more fine grained fixes.


MAJKFL:
@swift-ci Please test

Sources/SwiftLexicalLookup/LookupResult.swift
```diff
@@ -20,6 +20,10 @@ import SwiftSyntax
   case fromFileScope(SourceFileSyntax, withNames: [LookupName])
   /// Indicates where to perform member lookup.
   case lookInMembers(LookInMembersScopeSyntax)
+  /// Indicates to lookup generic parameters of extended type.
+  case lookInGenericParametersOfExtendedType(ExtensionDeclSyntax)
+  /// Indicates this closure expression could introduce dollar identifiers.
+  case mightIntroduceDollarIdentifiers(ClosureExprSyntax)
```
DougGregor, line 26: This enum is turning into pseudo-documentation for all of the weird lookup rules. Could you add a code example for these last two cases? Folks using the lookup APIs are likely to have to handle each of these cases, and a code example will help them understand why they exist and how to test for them.

Sources/SwiftLexicalLookup/Scopes/ScopeImplementations.swift
```diff
@@ -503,17 +415,46 @@ import SwiftSyntax
...
         }
 
-      return (implicitSelf.isEmpty ? [] : [.fromScope(self, withNames: implicitSelf)]) + [.lookInMembers(self)]
-        + defaultLookupImplementation(identifier, at: lookUpPosition, with: config)
+      return LookupResult.getResultArray(for: self, withNames: implicitSelf)
+        + [.lookInGenericParametersOfExtendedType(self)]
+        + defaultLookupImplementation(identifier, at: lookUpPosition, with: config, propagateToParent: false)
+        + [.lookInMembers(self)]
+        + lookupInParent(identifier, at: lookUpPosition, with: config)
+    } else if !extendedType.range.contains(lookUpPosition) && genericWhereClause != nil {
```
DougGregor, line 423: Okay, it's a stylistic thing, but I prefer that an "if" ending with a "return" not have an "else" following it, because it means unnecessary nesting. So instead of

```swift
if x {
 // ...
 return y
} else {
  // rest of body
}
```

we de-nest as:

```swift
if x {
  // ...
  return y
}

// rest of body
```

Sources/SwiftLexicalLookup/Scopes/ScopeImplementations.swift
```diff
@@ -503,17 +415,46 @@ import SwiftSyntax
           checkIdentifier(identifier, refersTo: name, at: lookUpPosition)
         }
 
-      return (implicitSelf.isEmpty ? [] : [.fromScope(self, withNames: implicitSelf)]) + [.lookInMembers(self)]
-        + defaultLookupImplementation(identifier, at: lookUpPosition, with: config)
+      return LookupResult.getResultArray(for: self, withNames: implicitSelf)
+        + [.lookInGenericParametersOfExtendedType(self)]
+        + defaultLookupImplementation(identifier, at: lookUpPosition, with: config, propagateToParent: false)
+        + [.lookInMembers(self)]
+        + lookupInParent(identifier, at: lookUpPosition, with: config)
```
DougGregor, line 422: That's a lot of array concatenation, but I like how it very clearly shows each of the necessary lookup steps in the appropriate order. (No action needed)

Tests/SwiftLexicalLookupTest/ExpectedName.swift
```diff
@@ -79,6 +81,11 @@ enum NameExpectation: ExpectedName {
     case (.declaration, .declaration): break
     case (.implicit(let implicitName), .implicit(let implicitNameExpectation)):
       implicitNameExpectation.assertExpectation(marker: marker, for: implicitName)
+    case (.dollarIdentifier(_, let acutalStr), .dollarIdentifier(_, let expectedStr)):
```
DougGregor, line 84: Typo "acutalStr"

DougGregor (approved):
Just a couple of minor comments, these are great improvements.

DougGregor:
@swift-ci please test Windows

Sources/SwiftLexicalLookup/LookupResult.swift
MAJKFL: Thanks for the suggestion. I added some example scenarios showing this behavior.

Sources/SwiftLexicalLookup/Scopes/ScopeImplementations.swift
MAJKFL: That's a good idea. I de-nested those couple of `else` statements inside `ExtensionDeclSyntax` scope extension.

Sources/SwiftLexicalLookup/Scopes/ScopeImplementations.swift
MAJKFL: I also like how it shows the steps, but I think we could try to improve readability here a bit. I think I’d like to try to rewrite it to some kind of result builder method at some point in the future when we have a good suite of integration and unit tests (I think I’ve mentioned it once during a meeting). I was thinking about something like this:

```swift
return lookupResult(identifier, at: lookupPosition, with: config)
  .implicitSelf()
  .lookInGenericParametersOfExtendedType()
  .defaultLookup()
  .lookInMembers()
  .lookupInParent()
```

Each of the methods could also take an optional `if: bool` parameter that would conditionally include that step. This way we could basically remove all those complicated `if … else if … else` statements from `lookup` methods. Additionally, we could try to enforce some pre/post conditions related to `config` in those functions (like `LookupConfig.finishInSequentialScope` being checked inside `.lookupInParent()` call for all sequential scopes).

MAJKFL:
@swift-ci Please test

DougGregor:
@swift-ci please test Windows

Sources/SwiftLexicalLookup/LookupResult.swift
DougGregor: Thank you!

Sources/SwiftLexicalLookup/Scopes/ScopeImplementations.swift
```diff
@@ -424,14 +424,14 @@ import SwiftSyntax
...
-          + defaultLookupImplementation(identifier, at: lookUpPosition, with: config)
       }
-    } else {
+
       return [.lookInGenericParametersOfExtendedType(self)]
-        + lookupInParent(identifier, at: lookUpPosition, with: config)
+        + defaultLookupImplementation(identifier, at: lookUpPosition, with: config)
     }
+
+    return [.lookInGenericParametersOfExtendedType(self)]
```
DougGregor, line 433: Much nicer de-nested, thank you!

DougGregor (approved):

Sources/SwiftLexicalLookup/Scopes/ScopeImplementations.swift
DougGregor: That does look nice! I don't know whether it's "worth" the effort of doing (vs. other things we've talked about for this library), but the result would be cleaner code in this part of the library.

DougGregor:
@swift-ci please test Windows

PR #2879, opened 2024-10-10
Add option to build a single dynamic library containing all swift-syntax modules

ahoppen:
My previous approach of building each product as a dynamic library did not work because SwiftPM includes symbols from all dependencies in each `.dll`. Because of this, we ended up with duplicate symbols for each type in eg. SwiftSyntax at runtime.

Build a single `.dll` with all required products instead to avoid that issue.

To enforce that this is an internal target that should not be relied upon, only enable it when an environment variable is set.


ahoppen:
@swift-ci Please test

rintaro:
I wonder how autolink works in this case. I.e. I assume `SwiftParser.swiftmodule` has a single module link name. But for example, `SwiftParser` belongs to 2 products. Does `import SwiftParser` causes `-lSwiftParser` or `-l_SwiftSyntaxDynamic`?

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

PR #2881, opened 2024-10-14
Fix all documentation warnings about mislabeled parameters

jaredgrubb:
I was trying to understand one of the macro API and the web documentation was missing the doc I was looking at, so I went to source and saw it was there, but wasn't being used because the name of the parameter was wrong.

I fixed all "not found" documentation warnings in the project. (Note that there are about 1400 more warnings about -missing- parameter documentation, but that's another PR for another day lol)


jaredgrubb:
When there were two options, I made my best call. Happy to adjust if the reviewers prefer me to use the other option (token/node and stream/target).

ahoppen (approved):
Thank you!

ahoppen:
@swift-ci Please test

PR #2882, opened 2024-10-16
SwiftSyntax support for experimental @abi attribute

beccadax:
Matches swiftlang/swift#76878.

Note that `@abi` introduces a situation we've never had before: the various `DeclGroupSyntax` decls can now have a `nil` `memberBlock`. This can only occur for children of an `ABIAttributeArgumentsSyntax` node, not any nodes that would occur in existing constructs. To preserve source compatibility with existing clients, I've made `memberBlock` an implicitly-unwrapped optional, but I'm open to opinions if that's not the design we'd like to have.

Another place I'd like feedback on is the way I'm parsing malformed `@abi` argument lists. Because the argument is literally an entire decl, I found that the default recovery behavior when the left paren was missing didn't work well, so I built something a little different. If you have better suggestions, I'd love to hear them.

### DIFF
--- a/CodeGeneration/Sources/SyntaxSupport/AttributeNodes.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/AttributeNodes.swift
@@ -141,6 +141,11 @@ public let ATTRIBUTE_NODES: [Node] = [
             name: "documentationArguments",
             kind: .node(kind: .documentationAttributeArgumentList)
           ),
+          Child(
+            name: "abiArguments",
+            kind: .node(kind: .abiAttributeArguments),
+            experimentalFeature: .abiAttribute
+          )
         ]),
         documentation: """
           The arguments of the attribute.
@@ -247,6 +252,20 @@ public let ATTRIBUTE_NODES: [Node] = [
           deprecatedCollectionElementName: "Availability"
         ),
         documentation: "The list of OS versions in which the declaration became ABI stable."
+      ),
+    ]
+  ),
+
+  Node(
+    kind: .abiAttributeArguments,
+    base: .syntax,
+    experimentalFeature: .abiAttribute,
+    nameForDiagnostics: "ABI-providing declaration",
+    documentation: "The arguments of the '@abi' attribute",
+    children: [
+      Child(
+        name: "provider",
+        kind: .node(kind: .decl)
       ),
     ]
   ),
--- a/CodeGeneration/Sources/SyntaxSupport/Child.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/Child.swift
@@ -36,6 +36,16 @@ public enum TokenChoice: Equatable, IdentifierConvertible {
   }
 }
 
+/// If a child can be optional, what flavor of optional is it? Used to make
+/// some children optional without breaking source compatibility.
+public enum Optionality: String {
+  /// An ordinary `Optional` which must be explicitly unwrapped by users.
+  case normal = "?"
+  /// An implicitly-unwrapped `Optional` which maintains source compatibility
+  /// if the child's optionality has changed.
+  case implicitlyUnwrapped = "!"
+}
+
 public enum ChildKind {
   /// The child always contains a node of the given `kind`.
   case node(kind: SyntaxNodeKind)
@@ -95,7 +105,7 @@ public class Child: NodeChoiceConvertible {
   public let kind: ChildKind
 
   /// Whether this child is optional and can be `nil`.
-  public let isOptional: Bool
+  public let optionality: Optionality?
 
   public let experimentalFeature: ExperimentalFeature?
 
@@ -256,7 +266,7 @@ public class Child: NodeChoiceConvertible {
     experimentalFeature: ExperimentalFeature? = nil,
     nameForDiagnostics: String? = nil,
     documentation: String? = nil,
-    isOptional: Bool = false
+    optionality: Optionality? = nil
   ) {
     precondition(name.first?.isLowercase ?? true, "The first letter of a child’s name should be lowercase")
     precondition(
@@ -270,6 +280,31 @@ public class Child: NodeChoiceConvertible {
     self.nameForDiagnostics = nameForDiagnostics
     self.documentationSummary = SwiftSyntax.Trivia.docCommentTrivia(from: documentation)
     self.documentationAbstract = String(documentation?.split(whereSeparator: \.isNewline).first ?? "")
-    self.isOptional = isOptional
+    self.optionality = optionality
+  }
+
+  /// If a classification is passed, it specifies the color identifiers in
+  /// that subtree should inherit for syntax coloring. Must be a member of
+  /// ``SyntaxClassification``.
+  /// If `forceClassification` is also set to true, all child nodes (not only
+  /// identifiers) inherit the syntax classification.
+  convenience init(
+    name: String,
+    deprecatedName: String? = nil,
+    kind: ChildKind,
+    experimentalFeature: ExperimentalFeature? = nil,
+    nameForDiagnostics: String? = nil,
+    documentation: String? = nil,
+    isOptional: Bool
+  ) {
+    self.init(
+      name: name,
+      deprecatedName: deprecatedName,
+      kind: kind,
+      experimentalFeature: experimentalFeature,
+      nameForDiagnostics: nameForDiagnostics,
+      documentation: documentation,
+      optionality: isOptional ? .normal : nil
+    )
   }
 }
--- a/CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
@@ -227,7 +227,8 @@ public let DECL_NODES: [Node] = [
       ),
       Child(
         name: "memberBlock",
-        kind: .node(kind: .memberBlock)
+        kind: .node(kind: .memberBlock),
+        optionality: .implicitlyUnwrapped
       ),
     ]
   ),
@@ -401,7 +402,8 @@ public let DECL_NODES: [Node] = [
         name: "memberBlock",
         kind: .node(kind: .memberBlock),
         documentation:
-          "The members of the class declaration. As class extension declarations may declare additional members, the contents of this member block isn't guaranteed to be a complete list of members for this type."
+          "The members of the class declaration. As class extension declarations may declare additional members, the contents of this member block isn't guaranteed to be a complete list of members for this type.",
+        optionality: .implicitlyUnwrapped
       ),
     ]
   ),
@@ -827,7 +829,8 @@ public let DECL_NODES: [Node] = [
         name: "memberBlock",
         kind: .node(kind: .memberBlock),
         documentation:
-          "The cases and other members associated with this enum declaration. Because enum extension declarations may declare additional members the contents of this member block isn't guaranteed to be a complete list of members for this type."
+          "The cases and other members associated with this enum declaration. Because enum extension declarations may declare additional members the contents of this member block isn't guaranteed to be a complete list of members for this type.",
+        optionality: .implicitlyUnwrapped
       ),
     ]
   ),
@@ -906,7 +909,8 @@ public let DECL_NODES: [Node] = [
         name: "memberBlock",
         kind: .node(kind: .memberBlock),
         documentation:
-          "The members of the extension declaration. As this is an extension, the contents of this member block isn't guaranteed to be a complete list of members for this type."
+          "The members of the extension declaration. As this is an extension, the contents of this member block isn't guaranteed to be a complete list of members for this type.",
+        optionality: .implicitlyUnwrapped
       ),
     ]
   ),
@@ -2011,7 +2015,8 @@ public let DECL_NODES: [Node] = [
       Child(
         name: "memberBlock",
         kind: .node(kind: .memberBlock),
-        documentation: "The members of the protocol declaration."
+        documentation: "The members of the protocol declaration.",
+        optionality: .implicitlyUnwrapped
       ),
     ]
   ),
@@ -2182,7 +2187,8 @@ public let DECL_NODES: [Node] = [
         name: "memberBlock",
         kind: .node(kind: .memberBlock),
         documentation:
-          "The members of the struct declaration. Because struct extension declarations may declare additional members the contents of this member block isn't guaranteed to be a complete list of members for this type."
+          "The members of the struct declaration. Because struct extension declarations may declare additional members the contents of this member block isn't guaranteed to be a complete list of members for this type.",
+        optionality: .implicitlyUnwrapped
       ),
     ]
   ),
--- a/CodeGeneration/Sources/SyntaxSupport/SyntaxNodeKind.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/SyntaxNodeKind.swift
@@ -22,6 +22,7 @@ public enum SyntaxNodeKind: String, CaseIterable, IdentifierConvertible, TypeCon
 
   case _canImportExpr
   case _canImportVersionInfo
+  case abiAttributeArguments
   case accessorBlock
   case accessorDecl
   case accessorDeclList
@@ -340,6 +341,15 @@ public enum SyntaxNodeKind: String, CaseIterable, IdentifierConvertible, TypeCon
     return .identifier(rawValue)
   }
 
+  public var uppercasedFirstWordRawValue: String {
+    switch self {
+    case .abiAttributeArguments:
+      "ABIAttributeArguments"
+    default:
+      rawValue.withFirstCharacterUppercased
+    }
+  }
+
   public var syntaxType: TypeSyntax {
     switch self {
     case .syntax:
@@ -347,7 +357,7 @@ public enum SyntaxNodeKind: String, CaseIterable, IdentifierConvertible, TypeCon
     case .syntaxCollection:
       return "SyntaxCollection"
     default:
-      return "\(raw: rawValue.withFirstCharacterUppercased)Syntax"
+      return "\(raw: uppercasedFirstWordRawValue)Syntax"
     }
   }
 
--- a/Sources/SwiftParser/Attributes.swift
+++ b/Sources/SwiftParser/Attributes.swift
@@ -11,9 +11,9 @@
 //===----------------------------------------------------------------------===//
 
 #if swift(>=6)
-@_spi(RawSyntax) internal import SwiftSyntax
+@_spi(ExperimentalLanguageFeatures) @_spi(RawSyntax) internal import SwiftSyntax
 #else
-@_spi(RawSyntax) import SwiftSyntax
+@_spi(ExperimentalLanguageFeatures) @_spi(RawSyntax) import SwiftSyntax
 #endif
 
 extension Parser {
@@ -58,6 +58,7 @@ extension Parser {
     case _typeEraser
     case _unavailableFromAsync
     case `rethrows`
+    case abi
     case attached
     case available
     case backDeployed
@@ -95,6 +96,7 @@ extension Parser {
       case TokenSpec(._typeEraser): self = ._typeEraser
       case TokenSpec(._unavailableFromAsync): self = ._unavailableFromAsync
       case TokenSpec(.`rethrows`): self = .rethrows
+      case TokenSpec(.abi) where experimentalFeatures.contains(.abiAttribute): self = .abi
       case TokenSpec(.attached): self = .attached
       case TokenSpec(.available): self = .available
       case TokenSpec(.backDeployed): self = .backDeployed
@@ -136,6 +138,7 @@ extension Parser {
       case ._typeEraser: return .keyword(._typeEraser)
       case ._unavailableFromAsync: return .keyword(._unavailableFromAsync)
       case .`rethrows`: return .keyword(.rethrows)
+      case .abi: return .keyword(.abi)
       case .attached: return .keyword(.attached)
       case .available: return .keyword(.available)
       case .backDeployed: return .keyword(.backDeployed)
@@ -176,9 +179,16 @@ extension Parser {
     case noArgument
   }
 
+  /// Parse the argument of an attribute, if it has one.
+  ///
+  /// - Parameters:
+  ///   - argumentMode: Indicates whether the attribute must, may, or may not have an argument.
+  ///   - parseArguments: Called to parse the argument list. If there is an opening parenthesis, it will have already been consumed.
+  ///   - parseMissingArguments: If provided, called instead of `parseArgument` when an argument list was required but no opening parenthesis was present.
   mutating func parseAttribute(
     argumentMode: AttributeArgumentMode,
-    parseArguments: (inout Parser) -> RawAttributeSyntax.Arguments
+    parseArguments: (inout Parser) -> RawAttributeSyntax.Arguments,
+    parseMissingArguments: ((inout Parser) -> RawAttributeSyntax.Arguments)? = nil
   ) -> RawAttributeListSyntax.Element {
     var (unexpectedBeforeAtSign, atSign) = self.expect(.atSign)
     if atSign.trailingTriviaByteLength > 0 || self.currentToken.leadingTriviaByteLength > 0 {
@@ -213,7 +223,11 @@ extension Parser {
         )
         leftParen = leftParen.tokenView.withTokenDiagnostic(tokenDiagnostic: diagnostic, arena: self.arena)
       }
-      let argument = parseArguments(&self)
+      let argument = if let parseMissingArguments, leftParen.presence == .missing {
+        parseMissingArguments(&self)
+      } else {
+        parseArguments(&self)
+      }
       let (unexpectedBeforeRightParen, rightParen) = self.expect(.rightParen)
       return .attribute(
         RawAttributeSyntax(
@@ -255,6 +269,12 @@ extension Parser {
     }
 
     switch peek(isAtAnyIn: DeclarationAttributeWithSpecialSyntax.self) {
+    case .abi:
+      return parseAttribute(argumentMode: .required) { parser in
+        return .abiArguments(parser.parseABIAttributeArguments())
+      } parseMissingArguments: { parser in
+        return .abiArguments(parser.parseABIAttributeArguments(missing: true))
+      }
     case .available, ._spi_available:
       return parseAttribute(argumentMode: .required) { parser in
         return .availability(parser.parseAvailabilityArgumentSpecList())
@@ -915,6 +935,33 @@ extension Parser {
       elements: [isolationKindElement],
       arena: self.arena
     )
+  }
+}
+
+extension Parser {
+  mutating func parseABIAttributeArguments(missing: Bool = false) -> RawABIAttributeArgumentsSyntax {
+    let providerDecl: RawDeclSyntax = if missing {
+      // There are two situations where the left paren might be missing:
+      //   1. The user just forgot the paren: `@abi var x_abi: Int) var x: Int`
+      //   2. The user forgot the whole argument list: `@abi var x: Int`
+      // Normally, we would distinguish these by seeing if whatever comes after
+      // the attribute parses as an argument. However, for @abi the argument is
+      // a decl, so in #2, we would consume the decl the attribute is attached
+      // to! This leads to a lousy diagnostic in that situation.
+      // Avoid this problem by simply returning a missing decl immediately.
+      // FIXME: Could we look ahead to find an unbalanced parenthesis?
+      RawDeclSyntax(
+        RawMissingDeclSyntax(
+          attributes: self.emptyCollection(RawAttributeListSyntax.self),
+          modifiers: self.emptyCollection(RawDeclModifierListSyntax.self),
+          arena: arena
+        )
+      )
+    } else {
+      parseDeclaration(in: .attribute)
+    }
+
+    return RawABIAttributeArgumentsSyntax(provider: providerDecl, arena: arena)
   }
 }
 
--- a/Sources/SwiftParser/Declarations.swift
+++ b/Sources/SwiftParser/Declarations.swift
@@ -168,11 +168,24 @@ extension Parser {
     }
   }
 
+  /// Information about the syntactic position of the declaration being parsed.
+  /// Used to tweak recovery and to permit missing bodies.
+  enum DeclarationParsingContext {
+    /// The declaration is at the top level of a file.
+    case topLevelCode
+
+    /// The declaration is nested inside the member list of a DeclGroupSyntax.
+    case memberList
+
+    /// The declaration is nested inside the argument list of an attribute.
+    case attribute
+  }
+
   /// Parse a declaration.
   ///
-  /// If `inMemberDeclList` is `true`, we know that the next item must be a
+  /// If `parseContext` is `memberList`, we know that the next item must be a
   /// declaration and thus start with a keyword. This allows further recovery.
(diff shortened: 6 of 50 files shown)


beccadax:
With https://github.com/swiftlang/swift/pull/76878

@swift-ci please test

ahoppen:
I really dislike using implicitly unwrapped optionals here. It’s just waiting to haunt us later when we forget to check for `nil`. As you already noted, changing `memberBlock` to be optional has a pretty big API impact (because `memberBlock` is pretty widely used) and also feels a little weird if the only case where it can actually be `nil` is inside the relatively low-frequency `@abi` attribute. Thus, I’d prefer to not make `memberBlock` optional.

This leaves us with one more option: Add new syntax nodes for all the type declarations. Maybe it would be OK if they are completely separate from the existing type nodes but just shared the parsing logic. I’ll need to think about that some more.

allevato:
Is there a way that we could use a non-optional `MemberBlockSyntax` with an empty item list, and some kind of dummy/missing-but-not-nil `leftBrace` and `rightBrace` tokens?

beccadax:
Another option I considered was adding a new `MemberBlockSyntax?` property to replace `memberBlock`, and making the backwards-compatibility `memberBlock` create a dummy missing MemberBlockSyntax when the new property is nil. I'm not sure what we'd call the new property, though. `members` has already been expended.

beccadax:
> Add new syntax nodes for all the type declarations. Maybe it would be OK if they are completely separate from the existing type nodes but just shared the parsing logic. I’ll need to think about that some more.

I guess this would look something like extracting `ClassDeclHeaderSyntax` out of `ClassDeclSyntax` so that you end up with:

```
// `class SomeClass {}`
- SourceFileSyntax
├─statements: CodeBlockItemListSyntax
│ ├─[0]: CodeBlockItemSyntax
│ │ ╰─item: ClassDeclSyntax
│ │   ├─header: ClassDeclHeaderSyntax
│ │   │ ├─attributes: AttributeListSyntax
│ │   │ ├─modifiers: DeclModifierListSyntax
│ │   │ ├─classKeyword: keyword(_CompilerSwiftSyntax.Keyword.class)
│ │   │ ╰─name: identifier("SomeClass")
│ │   ╰─memberBlock: MemberBlockSyntax
│ │     ├─leftBrace: leftBrace
│ │     ├─members: MemberBlockItemListSyntax
│ │     ╰─rightBrace: rightBrace
...
```

Then you could put compatibility properties on `ClassDeclSyntax` that forward to `header`'s corresponding properties. Repeat for each DeclGroup (or maybe each Decl period?), and then have `ABIAttributeArguments` use a child of type `DeclHeaderSyntax` instead of `DeclSyntax`.

This is a rather large refactor, but I think it could work.

beccadax:
This has been changed from the original implementation. The new version extracts a new "header" child node from nodes like `DeclClassSyntax` which contains everything but the member block. All of the `DeclGroup` nodes have this change applied. As proof that this works, I've updated the `HasTrailingMemberDeclBlock` initializer to take a header node, rather than a syntax string.

Generating the new decls I wanted—particularly their compatibility layers—took substantial refactoring of CodeGenerator, including the introduction of a new `childHistory` mechanism to replace `deprecatedName` and its ilk, extending it with the capability to model this kind of "extraction" of a child node from a parent. I've also made it so that base nodes can have children (these become protocol requirements), traits can optionally declare a memberwise failable initializer (this must be manually implemented at the moment), and deprecated properties are automatically generated for traits.

beccadax:
@ahoppen I'd love your opinion on this modification.

beccadax:
With https://github.com/swiftlang/swift/pull/76878

@swift-ci please test

CodeGeneration/Sources/SyntaxSupport/CommonNodes.swift
```diff
@@ -180,13 +180,98 @@ public let COMMON_NODES: [Node] = [
     parserFunction: "parseDeclaration"
   ),
 
+  Node(
+    kind: .declGroupHeader,
```
ahoppen, line 184: None of the other base nodes (`DeclSyntax`, `ExprSyntax` etc.) have any children and I think we will get crashes if any of the nodes whose base is `.declGroupHeader` are layout-incompatible with `declGroupHeader`. `DeclGroupHeaderSyntax` will assume that the first child in the node is `attributes` but that might not be the case for a node inheriting from `.declGroupHeader`. 

That said, I would make the base of all header nodes `.syntax` and create a trait for the decl group headers instead.

CodeGeneration/Sources/SyntaxSupport/Child.swift
```diff
@@ -244,32 +263,118 @@ public class Child: NodeChoiceConvertible {
...
+  
+  public func hash(into hasher: inout Hasher) {
+    hasher.combine(ObjectIdentifier(self))
+  }
 }
+
+extension Child {
+  /// A set of changes to the children that were all made simultaneously. The
+  /// key is the *new* name of the child; any names in the value are old names.
+  public typealias ChangeSet = [String: Refactoring]
```
ahoppen, line 340: I think I like the idea of the change set here but it’s a little hard to review because it’s combined with the `@abi` attribute changes. Do you think you could pull out the refactoring to `childHistory` into a separate, preliminary PR?

ahoppen (commented):
Two preliminary comments from an initial look at the changes. I think we’re on the right track but I don’t think it makes sense to give this PR a full review until they are addressed.

CodeGeneration/Sources/SyntaxSupport/Child.swift
beccadax: Done here: #2888

CodeGeneration/Sources/SyntaxSupport/CommonNodes.swift
beccadax: I'm a little reluctant to do this because I get a fair bit of mileage out of having `DeclGroupHeaderSyntax` be a concrete type instead of an existential. For instance, both `ABIAttributeSyntax.Provider` and the modified version of `HasTrailingMemberBlock` use the concrete `DeclGroupHeaderSyntax`; the former could not easily use an existential because there's no way to constrain a child node to have a trait, and the latter because a concrete type is needed to provide a contextual type for the string literal.

(If it helps, the `DeclGroupHeaderSyntax` accessors don't directly access the children—they indirect through the existential so they use the leaf type's layout.)

If you can suggest another design that would still give me a concrete node which could contain any of the headers, I'd be open to that change.

ahoppen: Let me think about it some more. It might also become clearer to me once https://github.com/swiftlang/swift-syntax/pull/2888 is merged and this PR is rebased so it doesn’t contain those changes.

nkcsgexi: Hey @ahoppen , we now have the other PR merged. Do you have more specific suggestions in mind?

Sources/SwiftSyntax/CustomTraits.swift
```diff
@@ -32,7 +60,35 @@ extension ClassDeclSyntax {
   }
 }
 
-extension EnumDeclSyntax {
+extension ClassDeclSyntax {
+  public var header: DeclGroupHeaderSyntax {
```
ahoppen, line 64: This ties in to my general comment on whether we need `DeclGroupHeaderSyntax`. I would prefer for `ClassDeclSyntax.header` to be of type `ClassDeclHeaderSyntax` instead of `DeclGroupHeaderSyntax`.

Sources/SwiftSyntax/generated/SyntaxTraits.swift
```diff
@@ -62,47 +62,28 @@ extension SyntaxProtocol {
...
-  /// ### Tokens
-  /// 
-  /// For syntax trees generated by the parser, this is guaranteed to be one of the following kinds:
-  ///  - `actor`
-  ///  - `class`
-  ///  - `enum`
-  ///  - `extension`
-  ///  - `protocol`
-  ///  - `struct`
-  var introducer: TokenSyntax {
```
ahoppen, line 86: Do we provide a compatibility layer for `DeclGroupSyntax.introducer` somewhere?

Tests/SwiftParserTest/AttributeTests.swift
```diff
@@ -965,6 +965,272 @@ final class AttributeTests: ParserTestCase {
...
+          returnClause: ReturnClauseSyntax(type: TypeSyntax("Int"))
+        )
+      ) {},
+      experimentalFeatures: [.abiAttribute]
+    )
+
+    assertParse(
+      """
+      @abi(associatedtype AssocTy)
+      associatedtype AssocTy
```
ahoppen, line 1016: Could you split these into multiple `assertParse` calls? That makes debugging a lot simpler if one of the decls fails to parse because you don’t need to reduce the input anymore.

Same below.

CodeGeneration/Sources/SyntaxSupport/AttributeNodes.swift
```diff
@@ -267,6 +272,31 @@ public let ATTRIBUTE_NODES: [Node] = [
...
+          Child(name: "associatedType", kind: .node(kind: .associatedTypeDecl)),
+          Child(name: "declGroup", kind: .node(kind: .declGroupHeader)),
+          Child(name: "deinitializer", kind: .node(kind: .deinitializerDecl)),
+          Child(name: "enumCase", kind: .node(kind: .enumCaseDecl)),
+          Child(name: "function", kind: .node(kind: .functionDecl)),
+          Child(name: "initializer", kind: .node(kind: .initializerDecl)),
+          Child(name: "subscript", kind: .node(kind: .subscriptDecl)),
+          Child(name: "typeAlias", kind: .node(kind: .typeAliasDecl)),
+          Child(name: "variable", kind: .node(kind: .variableDecl)),
+          Child(name: "unsupported", kind: .node(kind: .decl)),
```
ahoppen, line 294: Instead of having an `unsupported` case here, I think we should put the unsupported declaration into unexpected nodes. That way we can also generate a parser diagnostic for eg. `@api(import Foo)`.

Tests/SwiftParserTest/AttributeTests.swift
```diff
@@ -965,6 +965,272 @@ final class AttributeTests: ParserTestCase {
...
+          locationMarker: "7️⃣",
+          message: "expected '(', ABI-providing declaration, and ')' in attribute",
+          fixIts: ["insert '(', ABI-providing declaration, and ')'"]
+        ),
+        DiagnosticSpec(
+          locationMarker: "8️⃣",
+          message: "unexpected code ')' before function"
+        ),
+      ],
+      // FIXME: Would love to have the fix-it delete the space before the close paren on the first line
```
ahoppen, line 1209: You should be able to do this with a logic similar to the following

https://github.com/swiftlang/swift-syntax/blob/0386d87aa047ff4bd8bff4d651463657523485a9/Sources/SwiftParserDiagnostics/DiagnosticExtensions.swift#L95-L100

Tests/SwiftSyntaxBuilderTest/ClassDeclSyntaxTests.swift
```diff
@@ -35,7 +35,7 @@ final class ClassDeclSyntaxTests: XCTestCase {
   }
 
   func testThrowableClass() throws {
-    let buildable = try ClassDeclSyntax(name: .identifier("Foo")) {
+    let buildable = try ClassDeclSyntax(classHeader: .init(name: .identifier("Foo"))) {
```
ahoppen, line 38: In all of these test cases, I would prefer to not write `.init`. It’s easier to write but makes you wonder which type the header node has when reading and that’s essential information for tests that test the syntax tree structure.

Sources/SwiftParser/Declarations.swift
```diff
@@ -311,6 +325,185 @@ extension Parser {
...
+  /// For a header that will have a member list attached, use ``Parser/parseHeaderForDeclarationGroup(attrs:keyword:handle:)``, which this method wraps.
+  ///
+  /// If `inMemberDeclList` is `true`, we know that the next item must be a
+  /// declaration and thus start with a keyword. This allows further recovery.
+  mutating func parseDeclarationGroupHeader(inMemberDeclList: Bool = false) -> RawDeclGroupHeaderSyntax {
+    // If we are at a `#if` of attributes, the `#if` directive should be
+    // parsed when we're parsing the attributes.
+    if self.at(.poundIf) && !self.withLookahead({ $0.consumeIfConfigOfAttributes() }) {
+      // If we're at a #if, that's not actually allowed here; parse it and add
+      // it as an unexpected node of a MissingDeclGroupHeaderSyntax.
```
ahoppen, line 338: Why is this needed?

Sources/SwiftParser/Declarations.swift
```diff
@@ -311,6 +325,185 @@ extension Parser {
...
+      )
+    }
+
+    let attrs = DeclAttributes(
+      attributes: self.parseAttributeList(),
+      modifiers: self.parseDeclModifierList()
+    )
+
+    let recoveryResult: (match: DeclGroupHeaderSyntax.IntroducerOptions, handle: RecoveryConsumptionHandle)?
+    if let atResult = self.at(anyIn: DeclGroupHeaderSyntax.IntroducerOptions.self) {
```
ahoppen, line 382: Why can’t we always use `self.canRecoverTo`? That should already contain the case for if we are at a `DeclGroupHeaderSyntax.IntroducerOptions.self`.

Sources/SwiftParser/Declarations.swift
```diff
@@ -311,6 +325,185 @@ extension Parser {
...
+      )
+    }
+
+    if let (match, handle) = recoveryResult {
+      return parseHeaderForDeclarationGroup(
+        attrs: attrs,
+        introducer: match,
+        introducerHandle: handle,
+        allowsMemberBlock: false
+      ).0
```
ahoppen, line 403: I would name that tuple element so we don’t need to do an ambiguous `.0` here.

Sources/SwiftParser/Attributes.swift
```diff
@@ -918,6 +939,75 @@ extension Parser {
...
+            attributes: self.emptyCollection(RawAttributeListSyntax.self),
+            modifiers: self.emptyCollection(RawDeclModifierListSyntax.self),
+            arena: arena
+          )
+        )
+      )
+    } else {
+      switch nextDeclarationKeyword() {
+      case .group?:
+        providerDecl = RawABIAttributeArgumentsSyntax.Provider(parseDeclarationGroupHeader())!
```
ahoppen, line 967: I don’t like the use of the force unwrap here and thus the lack of type system checking that `RawABIAttributeArgumentsSyntax.Provider` can actually represent the parsed declaration header.

I don’t think the force unwrap is needed if you created the `RawABIAttributeArgumentsSyntax.Provider.declGroup` case explicitly.

Sources/SwiftParser/Attributes.swift
```diff
@@ -918,6 +939,75 @@ extension Parser {
...
+
+      case .simple?, .binding?, nil:
+        providerDecl = RawABIAttributeArgumentsSyntax.Provider(parseDeclaration(in: .argumentList))!
```
ahoppen, line 970: Similar here, I would prefer a solution that doesn’t require force unwrap here because it’s not possible to determine that `RawABIAttributeArgumentsSyntax.Provider` will be non-nil based on local reasoning.

Sources/SwiftParser/Declarations.swift
```diff
@@ -311,6 +325,185 @@ extension Parser {
...
+          )
+        )
+      }
+    }
+
+    return RawDeclGroupHeaderSyntax(
+      RawMissingDeclHeaderSyntax(
+        attributes: attrs.attributes,
+        modifiers: attrs.modifiers,
+        placeholder: missingToken(.identifier, text: "decl"),
```
ahoppen, line 426: I think the text should be `<#decl#>` to render a placeholder when the Fix-It to add the `MissingDeclHeaderSyntax` is applied.

Sources/SwiftParser/Declarations.swift
```diff
@@ -311,6 +325,185 @@ extension Parser {
...
+  ///     expected after this header and if one is found, it should be
+  ///     proactively consumed.
+  mutating func parseHeaderForDeclarationGroup(
+    attrs: DeclAttributes,
+    introducer: DeclGroupHeaderSyntax.IntroducerOptions,
+    introducerHandle: RecoveryConsumptionHandle,
+    allowsMemberBlock: Bool
+  ) -> (RawDeclGroupHeaderSyntax, shouldContinueParsing: Bool) {
+    func eraseToRawDeclGroupHeaderSyntax(
+      _ result: (some RawDeclGroupHeaderSyntaxNodeProtocol, Bool)
```
ahoppen, line 447: Could you add labels to the tuple elements here? It’s not really obvious what the `Bool` represents here.

Sources/SwiftParser/Declarations.swift
```diff
@@ -311,6 +325,185 @@ extension Parser {
...
+    ) -> (RawDeclGroupHeaderSyntax, shouldContinueParsing: Bool) {
```
ahoppen, line 448: Same here, I would prefer to also have a label for the `RawDeclGroupHeaderSyntax`.

Sources/SwiftParser/Declarations.swift
```diff
@@ -373,12 +566,122 @@ extension Parser {
...
+      RawMissingDeclHeaderSyntax.self,
+    ]
+
+    for subtype in subtypes {
+      if let result = subtype.init(self) {
+        return result
+      }
+    }
+
+    fatalError("Node \(self) does not have a known subtype")
```
ahoppen, line 638: `preconditionFailure` is mildly preferred over `fatalError`.

Sources/SwiftParser/Declarations.swift
```diff
@@ -373,12 +566,122 @@ extension Parser {
...
+    return asConcreteExistentialSelf.introducer
+  }
+}
+
 extension Parser {
+  /// Parse a declaration group (class, struct, enum, actor, protocol, extension) body given its header.
+  mutating func parseDeclarationGroup<T>(
+    for header: T,
+    shouldParseMemberBlock: Bool = true
+  ) -> T.Declaration where T: DeclarationGroupHeaderTrait {
```
ahoppen, line 655:
```suggestion
  mutating func parseDeclarationGroup<T: DeclarationGroupHeaderTrait>(
    for header: T,
    shouldParseMemberBlock: Bool = true
  ) -> T.Declaration {
```

Sources/SwiftParser/Declarations.swift
```diff
@@ -373,12 +566,122 @@ extension Parser {
...
+    let memberBlock =
+      if shouldParseMemberBlock {
+        self.parseMemberBlock(introducer: header.introducer)
```
ahoppen, line 658: Same here, we can’t use `if` expressions yet.

ahoppen (commented):
Left a few comments inline. 

Overall, I’m still skeptical of the new `DeclGroupHeaderSyntax` header base type because:
- The reasons from https://github.com/swiftlang/swift-syntax/pull/2882#discussion_r1823113927
- Accessing any properties of it requires an existential conversion, which may be unexpected performance-wise.

I would prefer if `ABIAttributeSyntax.Provider` listed all the headers nodes it supports, just like it already lists types like `typealiasDecl` and `funcDecl`.

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
```diff
@@ -232,15 +231,38 @@ public let DECL_NODES: [Node] = [
...
+    traits: [
+      "DeclGroup",
+      "NamedDecl",
+      "WithAttributes",
+      "WithGenericParameters",
+      "WithModifiers",
+    ],
+    children: [
+      Child(
+        name: "actorHeader",
```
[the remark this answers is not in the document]
beccadax: If there is anything in DeclGroupSyntax that allows access to the header, it needs to have a different name from the header children in all of the concrete nodes since it will need to have a broader type.

CodeGeneration/Sources/SyntaxSupport/SyntaxNodeKind.swift
```diff
@@ -340,14 +349,32 @@ public enum SyntaxNodeKind: String, CaseIterable, IdentifierConvertible, TypeCon
...
     default:
-      return "\(raw: rawValue.withFirstCharacterUppercased)Syntax"
+      return "\(raw: uppercasedFirstWordRawValue)Syntax"
+    }
+  }
+
+  public var baseTypeSuffix: String? {
+    switch self {
+    case .declGroupHeader:
+      return "DeclHeaderSyntax"
```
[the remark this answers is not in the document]
beccadax: I'm using "DeclGroupHeader" because only the DeclGroup nodes have separate headers. Do you think I'm being too picky with that?

CodeGeneration/Sources/generate-swift-syntax/InitSignature+Extensions.swift
```diff
@@ -63,19 +65,25 @@ extension InitSignature {
       )
     }
 
+    func transformParam(_ param: FunctionParameterSyntax) -> FunctionParameterSyntax {
```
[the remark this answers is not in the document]
beccadax: That feels at odds with how we normally use argument labels because the word `parameter` is redundant information given the type. Should it just be `transform(_:)`?

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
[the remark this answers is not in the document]
ahoppen: I realized that as well as I continued reviewing and think this is an argument for not having a `DeclGroupSyntax` node.

CodeGeneration/Sources/generate-swift-syntax/InitSignature+Extensions.swift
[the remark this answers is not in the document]
ahoppen: `transform(_:)` seems too ambiguous to me. If you don’t like `transform(parameter:)`, I don’t have super strong opinions on it and we can keep `transformParam(_:)`

Sources/SwiftParser/Declarations.swift
beccadax: If somebody—or some tool—mistakenly writes:
```
@abi(
    #if SomeCond
    class C1
    #endif
)
class C1 { ... }
```

I want the parser to recover and emit a good error.

I especially care about this because at one point, the compiler was emitting `#if $LangFeature` checks inside `@abi` attributes in module interfaces. I've fixed that by having the AST printer treat a usage inside an `@abi` attribute as a usage in the decl the `@abi` attribute is attached to, but I want to make sure that if there's a similar bug in the future, we parse it in a way that will produce a good diagnostic.

Sources/SwiftParser/Declarations.swift
beccadax: I must admit that I based this on `parseDeclaration()` and didn't look closely at whether this method could be written to always use `canRecoverTo`. If you think it can, I'll try to modify it.

Sources/SwiftParser/Attributes.swift
beccadax: We know that it will be non-nil because `parseDeclaration()` returns a `RawDeclSyntax` and `RawABIAttributeArgumentsSyntax.Provider` has a catch-all `unsupported` case which can contain any `RawDeclSyntax`. In other words, the real issue here is that `RawABIAttributeArgumentsSyntax.Provider.init(_:)`.

Would it suffice to add something like this somewhere to limit, document, and encapsulate the force unwrap:

```
extension RawABIAttributeArgumentsSyntax.Provider {
    init(decl: some RawDeclSyntaxNodeProtocol) {
        // Guaranteed to be non-nil because the `unsupported` case which will accept any `RawDeclSyntax`
        self.init(decl)!
    }
}
```

Or do you really think I should add a boilerplate `switch` or `if`-`else` chain solely to avoid a force unwrap? This seems like exactly the situation force unwrap is intended for: The programmer knows for certain that this operation will not be nil, but the type system cannot verify it from the declarations available.

Sources/SwiftParser/Declarations.swift
beccadax: Perhaps I should add a little commentary:

I've noticed that a weakness of the `@abi` syntax is that the parser struggles to recover when its syntax is malformed. I *think* the problem is that the parser's recovery system assumes that any parentheses it finds are part of the declaration, rather than part of an enclosing syntactic construct (the `@abi` attribute). I've therefore had to add some lookaheads and special cases which might not otherwise have been necessary.

If I'm holding it wrong or there is some way to modify the recovery system to work more gracefully for `@abi`, I'd love to hear it—I haven't worked much with this parser yet.

Sources/SwiftSyntax/generated/RenamedChildrenCompatibility.swift
```diff
@@ -103,37 +103,235 @@ extension AccessorDeclSyntax {
...
+    get {
+      return actorHeader.unexpectedBetweenActorKeywordAndName
+    }
+    set {
+      actorHeader.unexpectedBetweenActorKeywordAndName = newValue
+    }
+  }
+
+  @available(*, deprecated, renamed: "actorHeader.name")
+  public var name: TokenSyntax {
```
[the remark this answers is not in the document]
beccadax: I do understand your concerns here; I separated out the deprecation changes into their own commit specifically so you could evaluate this impact.

I could add something in CodeGenerator to allow certain compatibility layer APIs to be non-deprecated. Just let me know what you'd like me to do here.

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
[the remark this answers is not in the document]
beccadax: An alternative I've been considering is making the `DeclGroupHeader` node *non*-base and having it be something like:

```
  Node(
    kind: .declGroupHeader,
    base: .syntax,
    nameForDiagnostics: "declaration group header",
    parserFunction: "parseDeclarationGroupHeader",
    traits: [
      "WithAttributes",
      "WithModifiers",
      "DeclGroupHeaderProtocol", // slight naming issue I'll have to resolve
    ],
    children: [
      Child(
        name: "header",
        kind: .nodeChoices(choices: [
          Child(name: "actorHeader", kind: .node(kind: .actorDeclHeader)),
          Child(name: "classHeader", kind: .node(kind: .classDeclHeader)),
          Child(name: "enumHeader", kind: .node(kind: .enumDeclHeader)),
          Child(name: "extensionHeader", kind: .node(kind: .extensionDeclHeader)),
          Child(name: "protocolHeader", kind: .node(kind: .protocolDeclHeader)),
          Child(name: "structHeader", kind: .node(kind: .structDeclHeader)),
        ])
      )
    ]
  )
```

I would then add `modifiers`, `attributes`, etc. using a handwritten extension switching over the choices, so there'd be no existential overhead. Not quite as elegant, but it still gives us a concrete type that can represent any of the headers.

Sources/SwiftParser/Declarations.swift
ahoppen: I see about the `#if`. It makes sense with your explanation that we might incorrectly generate `#if` inside `@abi`. Otherwise, you could equally argue that the developer just forgot to finish the `@abi` attribute here:

```swift
@abi(

#if blah
class Foo {}
#endif
```

---

Do you have any concrete examples here? The problem is probably that `)` has a relatively low recovery precedence and decl keyword have a pretty high recovery precedence. Depends on the examples whether something can be done here.

Sources/SwiftParser/Declarations.swift
ahoppen: Yes, I think `canRecoverTo` should work.

Sources/SwiftParser/Attributes.swift
ahoppen: I want to be able to do this sort of reasoning locally without needing to look at multiple declarations (because the implementation of `RawABIAttributeArgumentsSyntax.Provider` or `parseDeclaration` might change in the future without re-visiting this invariant. Moving the force unwrap into an initializer of `RawABIAttributeArgumentsSyntax.Provider` seem like a reasonable solution.

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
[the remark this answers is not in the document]
ahoppen: Do we really need a `DeclGroupHeader` node at all? Correct me if I’m wrong, but I think the primary use for it is to be an option to `ABIAttributeArgumentsSyntax`, which already lists all sorts of other declarations. If we listed all the nodes that are subtypes of `DeclGroupHeaderSyntax` in there as well, I don’t think there’s real need for `DeclGroupHeaderSyntax` anymore. Or am I missing something?

Sources/SwiftParser/Attributes.swift
```diff
@@ -918,6 +939,75 @@ extension Parser {
...
+      }
+    }
+
+    return RawABIAttributeArgumentsSyntax(provider: providerDecl, arena: arena)
+  }
+
+  mutating func nextDeclarationKeyword() -> DeclarationKeyword? {
```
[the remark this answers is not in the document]
ahoppen: I think it should be possible to share the part that consumes attributes and modifiers pretty easily and that’s mostly what I care about because it’s the non-trivial part of this function. I’m not too concerned about having a function that calls some new functions and then `return subparser.at(anyIn: DeclarationKeyword.self)?.spec`.

Sources/SwiftSyntax/generated/RenamedChildrenCompatibility.swift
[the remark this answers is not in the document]
ahoppen: @rintaro Just chatted and agree that we shouldn’t deprecate `ClassDeclSyntax.name`.

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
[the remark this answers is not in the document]
beccadax: I also use it in the new implementation for `HasTrailingMemberDeclBlock`, and I believe I use it in ASTGen to generalize some of the lowering of DeclGroups, but those probably aren't mandatory—they're just nice little tricks I was able to do now that the new nodes existed.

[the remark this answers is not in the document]
ahoppen: I think none of those are sufficient reason to introduce `DeclGroupHeaderSyntax`.

CodeGeneration/Sources/SyntaxSupport/AttributeNodes.swift
```diff
@@ -267,6 +272,31 @@ public let ATTRIBUTE_NODES: [Node] = [
     ]
   ),
 
+  Node(
```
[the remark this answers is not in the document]
beccadax: At this point, the node is hidden behind an experimental feature. Should I still document it now?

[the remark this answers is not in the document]
ahoppen: I probably didn’t put the comment at the right place. I meant adding a release note entry for the nodes that change and that are not SPI, eg. the introduction of `ClassDeclSyntax.header`.

CodeGeneration/Sources/SyntaxSupport/AttributeNodes.swift
ahoppen: Yes, adding a choice for `.missingDecl` sounds like a good solution to me.

PR #2883, opened 2024-10-17
[SwiftLexicalLookup] Add support for `#if` and macro declarations. More `ASTScope` related fixes.

MAJKFL:
This PR includes more fixes that bring `SwiftLexicalLookup` unqualified lookup implementation closer to the `ASTScope` counterpart. Those include:
- Handling of `#if` clauses thanks to `SwiftIfConfig`.
- Macro declaration support.
- Sequential scope improvements. Now, only uses `.reversed()` for iteration. Since it's more likely code block contains more variable decls than named decls, sequential lookup first goes through all items, and collects named decls.
- Bring back `wildcard` token filtering for names from `IdentifiableSyntax`.
- More minor fixes.

With this PR, I was able to run validation for the entire Swift standard library with just **two** remaining bugs!

### DIFF
--- a/Sources/SwiftLexicalLookup/LookupConfig.swift
+++ b/Sources/SwiftLexicalLookup/LookupConfig.swift
@@ -10,6 +10,8 @@
 //
 //===----------------------------------------------------------------------===//
 
+import SwiftIfConfig
+
 @_spi(Experimental) public struct LookupConfig {
   /// Specifies whether lookup should finish in the closest sequential scope.
   ///
@@ -31,14 +33,17 @@
   /// If `finishInSequentialScope` would be set to `false`, the only name
   /// returned by lookup would be the `a` declaration from inside function body.
   @_spi(Experimental) public var finishInSequentialScope: Bool
+  @_spi(Experimental) public var buildConfiguration: BuildConfiguration?
 
   /// Creates a new lookup configuration.
   ///
   /// - `finishInSequentialScope` - specifies whether lookup should finish
   ///   in the closest sequential scope. `false` by default.
   @_spi(Experimental) public init(
-    finishInSequentialScope: Bool = false
+    finishInSequentialScope: Bool = false,
+    buildConfiguration: BuildConfiguration? = nil
   ) {
     self.finishInSequentialScope = finishInSequentialScope
+    self.buildConfiguration = buildConfiguration
   }
 }
--- a/Sources/SwiftLexicalLookup/LookupConfig.swift
+++ b/Sources/SwiftLexicalLookup/LookupConfig.swift
@@ -33,7 +33,7 @@ import SwiftIfConfig
   /// If `finishInSequentialScope` would be set to `false`, the only name
   /// returned by lookup would be the `a` declaration from inside function body.
   @_spi(Experimental) public var finishInSequentialScope: Bool
-  @_spi(Experimental) public var buildConfiguration: BuildConfiguration?
+  @_spi(Experimental) public var configuredRegions: ConfiguredRegions?
 
   /// Creates a new lookup configuration.
   ///
@@ -41,9 +41,9 @@ import SwiftIfConfig
   ///   in the closest sequential scope. `false` by default.
   @_spi(Experimental) public init(
     finishInSequentialScope: Bool = false,
-    buildConfiguration: BuildConfiguration? = nil
+    configuredRegions: ConfiguredRegions? = nil
   ) {
     self.finishInSequentialScope = finishInSequentialScope
-    self.buildConfiguration = buildConfiguration
+    self.configuredRegions = configuredRegions
   }
 }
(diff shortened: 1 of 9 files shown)


MAJKFL:
@swift-ci Please test

Sources/SwiftLexicalLookup/LookupConfig.swift
```diff
@@ -31,14 +33,17 @@
...
+  @_spi(Experimental) public var buildConfiguration: BuildConfiguration?
 
   /// Creates a new lookup configuration.
   ///
   /// - `finishInSequentialScope` - specifies whether lookup should finish
   ///   in the closest sequential scope. `false` by default.
   @_spi(Experimental) public init(
-    finishInSequentialScope: Bool = false
+    finishInSequentialScope: Bool = false,
+    buildConfiguration: BuildConfiguration? = nil
```
DougGregor, line 44: It's probably worth taking in a [`ConfiguredRegions`](https://swiftpackageindex.com/swiftlang/swift-syntax/main/documentation/swiftifconfig/configuredregions) here rather than a `BuildConfiguration`. `ConfiguredRegions` caches results so we aren't re-evaluating the same `#if`s repeatedly.

Sources/SwiftLexicalLookup/Scopes/ScopeImplementations.swift
```diff
@@ -512,13 +500,19 @@ import SwiftSyntax
...
   @_spi(Experimental) public var defaultIntroducedNames: [LookupName] {
+    var containsExpressionSyntax = false
+
     let extractedNames = catchItems.flatMap { item in
       guard let pattern = item.pattern else { return [LookupName]() }
 
+      if !containsExpressionSyntax && pattern.is(ExpressionPatternSyntax.self) {
+        containsExpressionSyntax = true
+      }
+
```
DougGregor, line 511: This is because we suppress the implicit `error` when there is some kind of pattern? Please leave such a comment with a code example to remind our future selves.

Sources/SwiftLexicalLookup/Scopes/ScopeImplementations.swift
```diff
@@ -887,10 +904,99 @@ extension SubscriptDeclSyntax: WithGenericParametersScopeSyntax, CanInterleaveRe
...
+  @_spi(Experimental) public var defaultIntroducedNames: [LookupName] {
+    [.implicit(.self(self))]
+  }
+
+  @_spi(Experimental) public var scopeDebugName: String {
+    "DeinitializerScope"
+  }
+}
+
+@_spi(Experimental) extension IfConfigDeclSyntax: IntroducingToSequentialParentScopeSyntax, SequentialScopeSyntax {
```
DougGregor, line 925: This was quite straightforward, I love it!

DougGregor (approved):

DougGregor:
@swift-ci please test Windows

DougGregor:
@swift-ci please test

Sources/SwiftLexicalLookup/Scopes/ScopeImplementations.swift
```diff
@@ -887,10 +923,99 @@ extension SubscriptDeclSyntax: WithGenericParametersScopeSyntax, CanInterleaveRe
...
+    let clause: IfConfigClauseSyntax?
+
+    if let configuredRegions = config.configuredRegions {
+      clause = configuredRegions.activeClause(for: self)
+    } else {
+      clause =
+        clauses
+        .first { clause in
+          clause.poundKeyword.tokenKind == .poundElse
+        }
```
ahoppen, line 996: Do we want to use the `#else` clause if no configured regions are specified? It’s a question. I don’t know what the correct answer here should be.

ahoppen (commented):
I only skimmed through it and that looked. Thank you, @MAJKFL 😍.

Sources/SwiftLexicalLookup/Scopes/ScopeImplementations.swift
MAJKFL: Yes, that was my intention that when `config.buildConfiguration` is set to `nil` it should provide some "default" behavior. I think it makes sense to use `#else` as the only other logical solution here I can think of would be maybe not including any names from `#if` clauses? Either way, I'm open to debate and I'm curious what do you think about it.

ahoppen: The only alternative I can think of would be to throw an error or something like that. @DougGregor Do you remember what we do in `SwiftIfConfig` when we’re unable to answer a query in the build configuration?

DougGregor: SwiftIfConfig treats the condition as having failed if the configuration can't answer the query.

PR #2886, opened 2024-10-29
Add TypeSyntax util functions

hnrklssn:
This moves isVoid from SwiftRefactor to be publicly accessible in SwiftSyntax. Also adds isSwiftCoreModule and isSwiftInt and calls isSwiftCoreModule in isVoid and isSwiftInt to properly handle the qualified cases of Swift.Void and Swift.Int.

### DIFF
--- a/Sources/SwiftSyntax/Utils.swift
+++ b/Sources/SwiftSyntax/Utils.swift
@@ -103,28 +103,64 @@ extension RawUnexpectedNodesSyntax {
   }
 }
 
-extension TypeSyntax {
+extension TypeSyntaxProtocol {
+  /// Check if this syntax matches any of the standard names for `Void`:
+  /// * Void
+  /// * Swift.Void
+  /// * ()
   public var isVoid: Bool {
-    switch self.as(TypeSyntaxEnum.self) {
-    case .identifierType(let identifierType) where identifierType.name.text == "Void": return true
-    case .tupleType(let tupleType) where tupleType.elements.isEmpty: return true
-    case .memberType(let memberType) where memberType.name.text == "Void": return memberType.baseType.isSwiftCoreModule
-    default: return false
+    if let identifierType = self.as(IdentifierTypeSyntax.self) {
+      return identifierType.name.text == "Void"
     }
+    if let memberType = self.as(MemberTypeSyntax.self) {
+      return memberType.baseType.isSwiftCoreModule && memberType.name.text == "Void"
+    }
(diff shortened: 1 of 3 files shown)


hnrklssn:
@swift-ci Please smoke test

Sources/SwiftSyntax/Utils.swift
```diff
@@ -102,3 +102,29 @@ extension RawUnexpectedNodesSyntax {
     self.init(raw: raw)
   }
 }
+
+extension TypeSyntax {
```
DougGregor, line 106: I think we want these to go on `TypeSyntaxProtocol`?

Sources/SwiftSyntax/Utils.swift
```diff
@@ -102,3 +102,29 @@ extension RawUnexpectedNodesSyntax {
...
+  public var isVoid: Bool {
```
DougGregor, line 107: Please provide documentation comments for all public APIs. Once we settle on them, please add an entry to the [release notes](https://github.com/swiftlang/swift-syntax/blob/main/Release%20Notes/601.md).

Sources/SwiftSyntax/Utils.swift
```diff
@@ -102,3 +102,29 @@ extension RawUnexpectedNodesSyntax {
...
+    }
+  }
+
+  public var isSwiftInt: Bool {
+    switch self.as(TypeSyntaxEnum.self) {
+    case .identifierType(let identifierType) where identifierType.name.text == "Int": return true
+    case .memberType(let memberType) where memberType.name.text == "Int": return memberType.baseType.isSwiftCoreModule
+    default: return false
+    }
+  }
```
DougGregor, line 122: This is largely a copy of the `isVoid` implementation above. I'm concerned that this approach to the API doesn't really scale: are we going to end up growing `isSwiftUInt`, `isSwiftBool`, `isSwiftSet`, and others as separate properties? That doesn't seem like a desirable end state.

hnrklssn: I found a way to generalise it to types where the canonical name only contains `MemberTypeSyntax` or `IdentifierTypeSyntax`. This doesn't include `Void` since it's a tuple, and there doesn't seem to be a good way to generalise meta types for tuples; I couldn't find anything like `Tuple.Type`, so handling tuple types would involve enumerating every arity. Not sure how to handle type shadowing either, since there doesn't seem to be a way to reify `module<Swift>.self`.

hnrklssn:
@swift-ci Please smoke test

PR #2887, opened 2024-10-30
Import os instead of OSLog

ahoppen:
`OSLog` is a module to consume logs and `os` is the module to produce them. Also, make the import `private`

rdar://137790052


ahoppen:
@swift-ci Please test

bnbarham (approved):

ahoppen:
@swift-ci Please test Windows

PR #2888, opened 2024-10-31
[NFC-ish] Rework compatibility layer generation

beccadax:
This PR replaces the `deprecatedName` property on `Child` with a new, more sophisticated mechanism which supports:

• Multiple past changes (for instance, a property that has been renamed more than once) with a different initializer for each past state.
• The ability to represent the order of multiple changes and group together simultaneous changes.
• Multiple kinds of changes, including an `extracted` change where a group of adjacent nodes are moved into a newly-created child node.

This new `childHistory` mechanism also works on `Trait`s, where it generates compatibility properties as one would expect. (We previously silently ignored `deprecatedName` on a child of a trait.)

This PR does not add or remove any SwiftSyntax APIs, but it does cause small, functionally equivalent changes to the generated code.

Extracted from #2882.


CodeGeneration/Sources/SyntaxSupport/Child.swift
```diff
@@ -150,28 +146,51 @@ public class Child: NodeChoiceConvertible {
...
+    return "\(raw: newestName.withFirstCharacterUppercased)Options"
   }
 
-  /// The deprecated name of this child that's suitable to be used for variable or enum case names.
-  public var deprecatedVarName: TokenSyntax? {
-    guard let deprecatedName = deprecatedName else {
-      return nil
+  /// If this child is actually part of another child's history, links back
+  /// to that child (and its parent, if this child was extracted to another
+  /// node). Empty if this is the newest version of the child.
```
ahoppen, line 162: I had a hard time grasping this (I think because I thought that the array traced the name across multiple change sets). I think an example would go a long way here to illustrate when `newerChildPath` can have multiple entries.

CodeGeneration/Sources/SyntaxSupport/Child.swift
```diff
@@ -150,28 +146,51 @@ public class Child: NodeChoiceConvertible {
...
+
+  /// True if this child was created by a `Refactoring`. Such children are part
+  /// of a compatibility layer and are therefore deprecated.
+  public var isHistorical: Bool {
+    !newerChildPath.isEmpty
+  }
+
+  /// Replaces the nodes in `newerChildPath` with their own `newerChildPath`s,
+  /// if any, to form a child path enitrely of non-historical nodes.
+  public func makeNewestChildPath() -> [Child] {
```
ahoppen, line 173: Should either `newerChildPath` or `makeNewestChildPath()` be `private`. It seems like it’s non-trivial for clients to decide whether they want `makeNewestChildPath()` or `newerChildPath`. Or do we need both in different circumstances?

CodeGeneration/Sources/SyntaxSupport/InitSignature.swift
```diff
@@ -2,18 +2,27 @@
...
+// Copyright (c) 2014 - 2024 Apple Inc. and the Swift project authors
 // Licensed under Apache License v2.0 with Runtime Library Exception
 //
 // See https://swift.org/LICENSE.txt for license information
 // See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
 //
 //===----------------------------------------------------------------------===//
 
-import SyntaxSupport
+public struct InitSignature {
```
ahoppen, line 13: Could you add a doc comment here?

CodeGeneration/Sources/SyntaxSupport/CompatibilityLayers.swift
```diff
@@ -0,0 +1,192 @@
...
+  private func extractedChildren(in layoutNode: LayoutNode) -> ArraySlice<Child> {
+    var children = layoutNode.children[...]
+    if children.first?.isUnexpectedNodes ?? false {
+      children.removeFirst()
+    }
+    if children.last?.isUnexpectedNodes ?? false {
+      children.removeLast()
+    }
+    return children
+  }
```
ahoppen, line 45: I would inline this. This function only makes sense with the comment on the only call side for it, so I think it’s easier to understand if the implementation lives there as well.

CodeGeneration/Sources/SyntaxSupport/CompatibilityLayers.swift
```diff
@@ -0,0 +1,192 @@
...
+        }
+      }
+    }
+
+    if cachedReplacementChildren[newerChild] == nil {
+      cachedReplacementChildren[newerChild] = make()
+    }
+    return cachedReplacementChildren[newerChild]!
+  }
+  private var cachedReplacementChildren: [Child: [Child]] = [:]
```
ahoppen, line 78: Could you move this up to the other members?

CodeGeneration/Sources/SyntaxSupport/CompatibilityLayers.swift
```diff
@@ -0,0 +1,192 @@
...
+        // Drop the leading and trailing unexpected nodes--these will be newly introduced.
+        let newerGrandchildren = extractedChildren(in: extractedNode.layoutNode!)
+
+        return newerGrandchildren.map { newerGrandchild in
+          Child(newerChildPath: [newerChild, newerGrandchild])
+        }
+      }
+    }
+
+    if cachedReplacementChildren[newerChild] == nil {
```
ahoppen, line 73: I assume we need to cache here to get pointer identity for the `Child` nodes? If that’s the case, could you explain that in a comment?

CodeGeneration/Sources/SyntaxSupport/CompatibilityLayers.swift
```diff
@@ -0,0 +1,192 @@
...
+// Copyright (c) 2014 - 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// Computes and caches information about properties and initializers that ought to be generated for the compatibility layer.
+public struct CompatibilityLayers /*: Sendable*/ {
```
ahoppen, line 14: I’m a little surprised by the use of `Layers` (plural). Layers apply to me that there are multiple layers on top of each other where one shadows the other. But all the data structures in here are flat, so would `CompatibilityLayer` make more sense?

CodeGeneration/Sources/SyntaxSupport/CompatibilityLayers.swift
```diff
@@ -0,0 +1,192 @@
...
+      // First pass: Apply the changes explicitly specified in the change set.
+      for (currentName, refactoring) in changeSet {
+        let i = firstIndexOfChild(named: currentName)
+
+        let replacementChildren = replacementChildren(for: children[i], by: refactoring)
+        children.replaceSubrange(i...i, with: replacementChildren)
+
+        if !areRequirements {
+          // Mark adjacent unexpected node children whose names have changed too.
+          if currentName != (replacementChildren.first?.name ?? "") {
```
ahoppen, line 147: I don’t think you need the `?? ""` here. We should be able to compare the non-optional `currentName` to an optional on the rhs. (same a few lines below)

CodeGeneration/Sources/SyntaxSupport/CompatibilityLayers.swift
```diff
@@ -0,0 +1,192 @@
...
+        precondition(unexpectedChild.isUnexpectedNodes)
+        let i = firstIndexOfChild(named: unexpectedChild.name)
+
+        let earlier = children[checked: i - 1]
+        let later = children[checked: i + 1]
+        precondition(!(earlier?.isUnexpectedNodes ?? false) && !(later?.isUnexpectedNodes ?? false))
+
+        let newChild = Child(forUnexpectedBetween: earlier, and: later, newerChildPath: [unexpectedChild])
+        precondition(newChild.name != unexpectedChild.name)
+        assert(!children.contains { $0.name == newChild.name })
```
ahoppen, line 167: Any reason why this is `assert` and the other is `precondition`. I would prefer `precondition` so that we also hit it when building CodeGeneration in release mode.

CodeGeneration/Sources/generate-swift-syntax/InitSignature+Extensions.swift
```diff
@@ -15,10 +15,24 @@ import SwiftSyntaxBuilder
...
+        } else {
+          return "\(child.labelDeclName):"
+        }
+      }.joined(separator: "")
+
+    return "init(leadingTrivia:\(renamedArguments)trailingTrivia:)"
+  }
+
+  func generateInitializerDeclHeader() -> SyntaxNodeString {
+    let stub = "public init"
```
ahoppen, line 32: Do you think that extracting `public init` helps readability here? I would argue that it’s easier to understand if `public init` was inlined in the two locations it’s used.

CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/RenamedChildrenCompatibilityFile.swift
```diff
@@ -16,90 +16,112 @@ import SyntaxSupport
...
+      let renamedName = InitSignature(layoutNode).compoundName
+      for signature in deprecatedInitSignatures {
+        makeCompatibilityInit(for: signature, renamedName: renamedName)
+      }
+    }
+  }
+
+  for trait in TRAITS.filter({ !$0.childHistory.isEmpty }) {
+    var deprecatedVars = SYNTAX_COMPATIBILITY_LAYERS.deprecatedVarsByTrait[trait.traitName] ?? []
+    var deprecatedArgLists = SYNTAX_COMPATIBILITY_LAYERS.deprecatedInitSignaturesByTrait[trait.traitName] ?? []
```
ahoppen, line 40: I assume this raises an unused variable warning, right? Could you remove `deprecatedArgLists` to fix that?

ahoppen (commented):
Looks like a great clean-up to the somewhat ad-hoc deprecation design we had before.

CodeGeneration/Sources/SyntaxSupport/Child.swift
```diff
@@ -40,7 +40,8 @@ public enum ChildKind {
   /// The child always contains a node of the given `kind`.
   case node(kind: SyntaxNodeKind)
   /// The child always contains a node that matches one of the `choices`.
-  case nodeChoices(choices: [Child])
+  case nodeChoices(choices: [Child], childHistory: Child.History = [])
+  // FIXME: We don't appear to have ever generated compatibility layers for children of node choices!
```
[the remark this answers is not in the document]
beccadax: `nodeChoices` still doesn't have compatibility layers. I think largely because Swift can't alias cases—you can use a deprecated static method to construct the replacement case, but not to pattern-match it.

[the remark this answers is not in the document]
ahoppen: Instead of having a `FIXME` for it (which is very easy to miss), could we raise a `fatalError` if you try to add a compatibility layer of node choices? That way you can’t miss it if you try to add one.

CodeGeneration/Sources/SyntaxSupport/Child.swift
```diff
@@ -244,32 +263,117 @@ public class Child: NodeChoiceConvertible {
...
+  convenience init(forUnexpectedBetween earlier: Child?, and later: Child?, newerChildPath: [Child] = []) {
+    let name = switch (earlier, later) {
+    case (nil, let later?):
+      "unexpectedBefore\(later.name.withFirstCharacterUppercased)"
+    case (let earlier?, nil):
+      "unexpectedAfter\(earlier.name.withFirstCharacterUppercased)"
+    case (let earlier?, let later?):
+      "unexpectedBetween\(earlier.name.withFirstCharacterUppercased)And\(later.name.withFirstCharacterUppercased)"
+    case (nil, nil):
+      "unexpected"
```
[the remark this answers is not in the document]
beccadax: This matches current behavior, but I'll check that this never currently happens and `fatalError()` in that case.

CodeGeneration/Sources/SyntaxSupport/Child.swift
beccadax: We may be able to get away with computing and storing the `newestChildPath` at initialization time. I'll consider that.

CodeGeneration/Sources/SyntaxSupport/CompatibilityLayers.swift
beccadax: I wasn't sure about how we meant the term "compatibility layer" to be used—is there one compatibility layer that covers the entire project, or are there many "compatibility layers", each covering one type or API? I took the latter interpretation, and named this `CompatibilityLayers` because it has knowledge of all of them, but I can rename it if I've gotten this wrong.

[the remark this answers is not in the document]
ahoppen: Let’s just remove it for now. Making CodeGeneration build in Swift 6 mode would be a different endeavor.

CodeGeneration/Sources/SyntaxSupport/Child.swift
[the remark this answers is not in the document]
ahoppen: Ah, I didn’t realize that. Let’s try if `fatalError` works and if not, let’s just leave it as-is.

CodeGeneration/Sources/SyntaxSupport/CompatibilityLayers.swift
ahoppen: I consider there to be a single compatibility layer for the entire project.

CodeGeneration/Sources/SyntaxSupport/Child.swift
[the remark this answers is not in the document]
beccadax: So, there actually *are* existing cases where people have specified a `deprecatedName:` for a child of a `ChildKind.nodeChoices`; this information was simply ignored until now. Rather than deleting it, I thought it'd be better to preserve the info in the node definition files, but if we'd rather just delete it (which—again—would be an NFC change), I can remove the `childHistory` value here entirely.

CodeGeneration/Sources/SyntaxSupport/CompatibilityLayers.swift
beccadax: Gotcha.

CodeGeneration/Sources/SyntaxSupport/CompatibilityLayers.swift
beccadax: It just looked relatively expensive, but sure, I can change it.

CodeGeneration/Sources/generate-swift-syntax/InitSignature+Extensions.swift
beccadax: This is sort of preparing for a change in the next PR, where we sometimes end up using `init?` instead of `public init`. I can move this specific change into that commit, though, so we can figure out how we want to structure this once we get there.

CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/RenamedChildrenCompatibilityFile.swift
beccadax: This might again be a change that ought to be deferred into a commit in the other PR—I'll take a look at it.

beccadax: Ended up tweaking this so there's a small struct grouping the vars and inits together; I'd been thinking about making that change for a while anyway.

CodeGeneration/Sources/SyntaxSupport/Child.swift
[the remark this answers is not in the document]
beccadax: I haven't done anything with this yet, but I think I've addressed everything else you suggested.

CodeGeneration/Sources/SyntaxSupport/CompatibilityLayers.swift
ahoppen: Since CodeGeneration is only run occasionally during swift-syntax development, I don’t really care about performance all that much.

CodeGeneration/Sources/generate-swift-syntax/InitSignature+Extensions.swift
ahoppen: Oh, I see. Happy to leave it in that case.

ahoppen (approved):
Thanks for addressing my comments, @beccadax.

CodeGeneration/Sources/SyntaxSupport/CompatibilityLayer.swift
```diff
@@ -0,0 +1,204 @@
...
+    return deprecatedMembersByNode[node.kind] ?? DeprecatedMemberInfo()
+  }
+
+  /// Returns the deprecated members that the compatibility layer needs for `trait`.
+  public func deprecatedMembers(for trait: Trait) -> DeprecatedMemberInfo {
+    return deprecatedMembersByTrait[trait.traitName] ?? DeprecatedMemberInfo()
+  }
+
+  internal init(nodes: [Node], traits: [Trait]) {
+    // Note that we compute these up-front, greedily, so that all of the mutation is isolated by the lazy var's state.
```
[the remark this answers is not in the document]
beccadax: I was referring to the `SYNTAX_COMPATIBILITY_LAYER` global, which like all Swift globals, is implicitly lazily initialized with synchronization.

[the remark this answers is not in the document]
beccadax: Would this be a better way to describe the situation?
```suggestion
    // This instance will be stored in a global that's used from multiple threads simultaneously, so it won't be safe
    // to mutate once the initializer returns. We therefore do all the work to populate its tables up front, rather
    // than computing it lazily on demand.
```

beccadax:
@swift-ci please test

CodeGeneration/Sources/SyntaxSupport/CompatibilityLayer.swift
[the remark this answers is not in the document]
ahoppen: Sounds great to me. Thanks!

beccadax:
```
SyntaxSupport/Child.swift:345: Fatal error: unexpected node has no siblings?
```

Turned out this was happening with base nodes, since they don't have any children (but their unexpected nodes are ignored anyway). Pulled in a fix that avoids trying to create unexpected children for base nodes.

beccadax:
@swift-ci please test

beccadax:
Needed to placate swift-format.

beccadax:
@swift-ci please smoke test

beccadax:
@swift-ci please test

PR #2891, opened 2024-11-06
Sanitize dollar identifiers with leading zeros in `Identifier`.

MAJKFL:
This PR introduces sanitization of dollar identifiers with leading zeros to `Identifier`. 

The compiler doesn't treat dollar identifiers with leading zeros differently than the "canonical" ones, so for the following piece of code:

```swift
var foo = {
  print($0, $00000000, $1, $00001)
}

foo(4, 2)
```

The result will be:

```
4 4 2 2
```

This PR will also enable `SwiftLexicalLookup` to properly handle this edge case when introducing dollar identifiers as names in closures.

### DIFF
--- a/Sources/SwiftSyntax/Identifier.swift
+++ b/Sources/SwiftSyntax/Identifier.swift
@@ -19,7 +19,7 @@ public struct Identifier: Equatable, Hashable, Sendable {
 
   /// `true` if the identifier is a dollar identifier.
   public var isDollarIdentifier: Bool {
-    raw.name.hasPrefix(SyntaxText("$"))
+    raw.original.hasPrefix(SyntaxText("$")) && Int(String(syntaxText: raw.original).dropFirst()) != nil
   }
 
   @_spi(RawSyntax)
@@ -49,17 +49,17 @@ public struct Identifier: Equatable, Hashable, Sendable {
     }
   }
 
-  /// Create a new `Identifier` from given `staticString`.
+  /// Create a new `Identifier` from given `canonicalName`.
   ///
-  /// - Precondition: `staticString` is a canonical identifier i.e. doesn't
+  /// - Precondition: `canonicalName` is a canonical identifier i.e. doesn't
   ///   use backticks and is not a dollar identifier with leading zeros.
-  public init(staticString: StaticString) {
+  public init(canonicalName: StaticString) {
     precondition(
-      Self.isCanonicalRepresentation(staticString),
-      "\(staticString) is not a canonical identifier."
+      Self.isCanonicalRepresentation(canonicalName),
+      "\(canonicalName) is not a canonical identifier."
     )
 
-    self.raw = RawIdentifier(SyntaxText(staticString))
+    self.raw = RawIdentifier(SyntaxText(canonicalName))
     self.arena = nil
   }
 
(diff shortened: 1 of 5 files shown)


MAJKFL:
@swift-ci Please test

Sources/SwiftSyntax/Identifier.swift
```diff
@@ -26,41 +29,67 @@ public struct Identifier: Equatable, Hashable, Sendable {
...
-  public init(_ staticString: StaticString) {
-    self.raw = RawIdentifier(staticString)
-    self.arena = nil
-
-    let name = String(syntaxText: raw.name)
+  /// Create a new `Identifier` from given `staticString`.
+  ///
+  /// - Precondition: `staticString` is a canonical identifier i.e. doesn't
+  ///   use backticks and is not a dollar identifier with leading zeros.
+  public init(staticString: StaticString) {
```
ahoppen, line 56: What’s the reason why you added the `staticString` argument label?

ahoppen (commented):
How did you find that, Jakub?

Sources/SwiftSyntax/Identifier.swift
MAJKFL: I noticed this used to produce a compilation error
```swift
let identifier = Identifier("smth") // Error: Ambiguous use of 'init(_:)'
```
Interestingly, these two worked absolutely fine
```swift
let x: StaticString = "smth"
let identifier1 = Identifier(x)
let identifier2: Identifier = Identifier("smth")
```
It seemed to me a bit odd overloading the initializer worked like that in this case. Is there a reason where a string literal could be used with the other `init?(_ token: TokenSyntax)`?

This behavior doesn't happen with an explicit argument label so that's why I changed that. Do you think there could be a better way to fix this?

MAJKFL:
> How did you find that, Jakub?

We’ve discovered this with @DougGregor when validating `SwiftLexicalLookup` against `ASTScope`. We were trying out different edge cases and noticed this behavior. 

`SwiftLexicalLookup` handles dollar identifiers slightly differently for anonymous lookup than the compiler. When encountering a closure without parameters, the lookup just prompts clients with `LookupResult.mightIntroduceDollarIdentifiers` result flag.

For a lookup with a particular identifier, it does introduce a specific dollar identifier name (`LookupName.dollarIdentifier`), and that’s where we need its canonical representation to not introduce a monstrosity such as `$000000000042`.

MAJKFL:
@swift-ci Please test Windows

DougGregor (approved):
Looks reasonable to me, thank you

Sources/SwiftSyntax/Identifier.swift
ahoppen: Oh, I see. `TokenSyntax` is expressible by string literal as well, that’s why it’s ambiguous.

I’m not a fan of the `staticString` label. How about we use the argument label to convey that the string needs to be canonical? That way a user will get the overload via `TokenSyntax` that canonicalizes the name by default and they need to opt into this faster but more restrictive overload.

Ie. change it to `init(canonicalName:)` or something of the sort.

MAJKFL: Oh that makes sense why it didn't work with string literals, thanks for explanation. I also like the idea of `init(canonicalName:)`. I changed it.

MAJKFL:
@swift-ci Please test

MAJKFL:
@swift-ci Please test Windows

ahoppen (approved):

ahoppen:
@swift-ci Please test Windows

DougGregor:
@swift-ci Please test Windows

PR #2895, opened 2024-11-08
Add a string-based initializer for `ExperimentalFeatures`.

allevato:
Tools (like swift-format) can use this to let users enable experimental features in the parser via interfaces like the command line, without having to handle the translation from string to option set themselves.

### DIFF
--- a/Sources/SwiftParser/generated/ExperimentalFeatures.swift
+++ b/Sources/SwiftParser/generated/ExperimentalFeatures.swift
@@ -33,15 +33,38 @@ extension Parser.ExperimentalFeatures {
   /// Whether to enable the parsing of 'do' expressions.
   public static let doExpressions = Self (rawValue: 1 << 2)
 
-  /// Whether to enable the parsing of NonEscableTypes.
+  /// Whether to enable the parsing of non-escapable types.
   public static let nonescapableTypes = Self (rawValue: 1 << 3)
 
-  /// Whether to enable the parsing of trailing comma.
+  /// Whether to enable the parsing of trailing commas.
   public static let trailingComma = Self (rawValue: 1 << 4)
 
-  /// Whether to enable the parsing of CoroutineAccessors.
+  /// Whether to enable the parsing of coroutine accessors.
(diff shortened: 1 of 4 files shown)


Sources/SwiftParser/generated/ExperimentalFeatures.swift
```diff
@@ -44,4 +44,27 @@ extension Parser.ExperimentalFeatures {
 
   /// Whether to enable the parsing of ValueGenerics.
   public static let valueGenerics = Self (rawValue: 1 << 6)
+
+  /// Creates a new value representing the experimental feature with the
+  /// given name, or returns nil if the name is not recognized.
+  public init?(name: String) {
+    switch name {
+    case "referenceBindings":
```
ahoppen, line 52: I think these feature names should match the feature names in the compiler (ie. experimental or upcoming features), which means that they will need to be manually specified.

Sources/SwiftParser/generated/ExperimentalFeatures.swift
```diff
@@ -33,15 +33,38 @@ extension Parser.ExperimentalFeatures {
...
+  /// Whether to enable the parsing of coroutine accessors.
   public static let coroutineAccessors = Self (rawValue: 1 << 5)
 
-  /// Whether to enable the parsing of ValueGenerics.
+  /// Whether to enable the parsing of value generics.
   public static let valueGenerics = Self (rawValue: 1 << 6)
+
+  /// Creates a new value representing the experimental feature with the
+  /// given name, or returns nil if the name is not recognized.
+  public init?(name: String) {
```
ahoppen, line 50: Could you add a release note entry for this new method?

ahoppen (approved):
Just one request for a release notes entry, otherwise LGTM.

Sources/SwiftParser/generated/ExperimentalFeatures.swift
allevato: Done.

allevato:
@swift-ci please test

ahoppen (approved):

Issue #2803, opened 2024-08-13
Macro protocols lack `async` from `expansion` function requirements despite proposals stating they should have it


roopekv:
## Issue
Swift's concurrency features cannot be fully utilized in macros without being able to implement `expansion` function requirements of macro protocols as `async` functions.

## Cause
While the Swift Evolution proposals for expression and attached macros state that the `expansion` function requirements of macro protocols should be `async`, the macro protocol declarations in swift-syntax are missing the `async` keyword from their `expansion` function requirements ([see comparison below](#differences-between-swift-evolution-proposals-and-swift-syntax)), making it impossible to implement them as `async` functions.

## Solution
Changing the declarations in swift-syntax to `async` shouldn't break existing macros, as `async` function requirements can be implemented as non-async functions. 

### Example
```swift
protocol Macro {
    func expansion()
}

struct SyncMacro : Macro {
    func expansion() {} // OK ✅
}

struct AsyncMacro : Macro {
    func expansion() async {} // ERROR ❌
}
```

```swift
protocol Macro {
    func expansion() async
}

struct SyncMacro : Macro {
    func expansion() {} // OK ✅
}

struct AsyncMacro : Macro {
    func expansion() async {} // OK ✅
}
```

I acknowledge that additional changes would have to be made within the swift-syntax package (changes at the expansion function call sites and possible further structural changes to facilitate asynchronous code), but depending on how they are implemented, the changes don't have to affect the user-facing parts of the package in a source breaking way.

<h2 id="differences-between-swift-evolution-proposals-and-swift-syntax">Differences between Swift Evolution proposals and swift-syntax</h2>

### ExpressionMacro

[**Proposal**](https://github.com/swiftlang/swift-evolution/blob/2c10ffea15f5fdc41b32f0e78fe17b0b975adc8e/proposals/0382-expression-macros.md?plain=1#L88-L97)

```swift
public protocol ExpressionMacro: FreestandingMacro {
  /// Expand a macro described by the given freestanding macro expansion
  /// within the given context to produce a replacement expression.
  static func expansion(
    of node: some FreestandingMacroExpansionSyntax,
    in context: some MacroExpansionContext
  ) async throws -> ExprSyntax
}
```

**swift-syntax**

https://github.com/swiftlang/swift-syntax/blob/0b324f86cb39dacaf48ed263619826183c25977b/Sources/SwiftSyntaxMacros/MacroProtocols/ExpressionMacro.swift#L20-L27

[...]

PR #2896, opened 2024-11-14
Add support for asynchronous macro expansion

roopekv:
This adds support for asynchronous macro expansion by declaring async overloads for `assertMacroExpansion` and the various `expansion` functions discussed with @ahoppen at #2803, making it possible to utilize Swift's concurrency features in macros.

## [Add `AsyncSyntaxRewriter`](https://github.com/swiftlang/swift-syntax/commit/276c8d40e505c8ef287373734d02e1118f39dbe1)

Adds an asynchronous version of `SyntaxRewriter` named `AsyncSyntaxRewriter` marked with the `@_spi(MacroExpansion)` attribute for implementing an asynchronous version of `MacroApplication` in the `SwiftSyntaxMacroExpansion` module.  

If the `package` access level supported subclassing and overriding of functions from other modules, it could be used instead of `@_spi(...) open` to restrict the use of `AsyncSyntaxRewriter` from outside of the package, but currently, it isn't supported.

## [Add async overloads of `expansion` functions to macro protocols](https://github.com/swiftlang/swift-syntax/commit/6f4cc1dfae6752c6738bfc20ef71a8eec9e71cd0)

Adds asynchronous overloads for `expansion` functions of the various macro protocols with default implementations that call the synchronous versions.

## [Add support for async macro expansion](https://github.com/swiftlang/swift-syntax/commit/d98d1fee0398a78cdc6dde6f931a4cf436f3f858)

Adds asynchronous overloads for `assertMacroExpansion` functions and support for asynchronous macro expansion down to the `SwiftCompilerPlugin` level.


ahoppen (commented):
Thank you for the PR, @roopekv. 

This PR adds a lot of code, which has cost associated with it, such as an increased binary size, higher maintainance cost because there are an async and a sync version of a function that need to be kept in sync, and because updating generated code is harder than non-generated code. At the same time, only a subset of swift-syntax works well with async functions (for example there is no async visitor). So, you could argue that this PR is too large and too small at the same time.

Furthermore, given this non-trivial cost, this change needs to provide some significant, real-world benefits and I don’t see those right now: The two scenarios in which Swift Concurrency provides the biggest benefits are (1) if the program needs to halt to wait for network requests, file access etc. and (2) to execute expensive work across multiple cores. (1) does not apply to macros because they should not read data from the outside world. Regarding (2), macros should not perform so much work that they benefit from multi-threading. They are only given a very limited subset of the source file and processing that should not take so much time that it makes sense to go through the overhead of spinning up mulitple cores. Multi-threading might even be harmful because the build system will schedule compiler invocations in parallel to use all CPU cores. If one of these invocations uses multiple cores, it can lead to machine over-subscription.

roopekv:
> This PR adds a lot of code, which has cost associated with it, such as an increased binary size, higher maintainance cost because there are an async and a sync version of a function that need to be kept in sync, and because updating generated code is harder than non-generated code.

The synchronous version of the code and the added code generation step is only required for the synchronous version of `assertMacroExpansion`. If the synchronous version of `assertMacroExpansion` could be removed, so could all the synchronous code facilitating it, in addition to the added code generation step.

As `assertMacroExpansion` is mainly used in tests, and all the major testing frameworks (Swift Testing, XCTest) support asynchronous tests, the migration process would be as easy as marking the test functions as async and adding the await keyword before the `assertMacroExpansion` calls. As a compromise, the users could be given a heads-up by marking the synchronous `assertMacroExpansion` as deprecated, and removing it in a later release along with all the extra code.

> At the same time, only a subset of swift-syntax works well with async functions (for example there is no async visitor). So, you could argue that this PR is too large and too small at the same time.

The scope of this PR can be expanded to other parts of the package if required, or those issues could be tackled in separate ones.

> Furthermore, given this non-trivial cost, this change needs to provide some significant, real-world benefits and I don’t see those right now: The two scenarios in which Swift Concurrency provides the biggest benefits are (1) if the program needs to halt to wait for network requests, file access etc. and (2) to execute expensive work across multiple cores. (1) does not apply to macros because they should not read data from the outside world. Regarding (2), macros should not perform so much work that they benefit from multi-threading. They are only given a very limited subset of the source file and processing that should not take so much time that it makes sense to go through the overhead of spinning up mulitple cores. Multi-threading might even be harmful because the build system will schedule compiler invocations in parallel to use all CPU cores. If one of these invocations uses multiple cores, it can lead to machine over-subscription.

The accepted Swift Evolution proposals for macros stated that the macro expansion operation can be asynchronous before I pointed out that the implementation in swift-syntax didn't match those proposals in #2803. Perhaps the potential benefits and drawbacks should be discussed with the wider Swift community before completely ruling out the idea?

roopekv:
> You never know in which context people might be using assertMacroExpansion (they could be calling it from a closure that doesn’t support async calls) and forcing async/await on them when there is no reason to unless their macro is using async functions (which I expect 99% of all macros to not) does not seem right to me.

I cannot think of a valid use case for `assertMacroExpansion` outside of testing. After all, it's part of the TestSupport modules. I would even go as far as to say that `assertMacroExpansion` shouldn't be used outside of tests at all, as the purpose of macros is not to be called as regular functions. If such behavior is desired, the implementation of the macro should be separated from the macro itself, allowing its functionality to be used from regular code.

Is there a case where a macro expansion test couldn't be asynchronous?

> My understanding is that first macro proposals included an async expansion function because there was the idea that macros could call back into the compiler to retrieve additional information (such as types). As we gained more experience with macros, we realized that such facilities are not needed or wanted, so the need for async never arose. But since I’m not the author of these proposals, I’m not the right person to give the final decision here. The language steering group has accepted the revision to the macro proposals, making the expansion functions synchronous https://github.com/swiftlang/swift-evolution/pull/2546.

Thanks for the clarification. There is an announcement about the revision on the Swift Forums that says:

> In line with our ["shrink-to-fit" process](https://forums.swift.org/t/addressing-unimplemented-evolution-proposals/40322), we are incorporating edits to SE-0382 and SE-0389 that reflect the API as implemented. Notably, this does not necessarily preclude "async-ified" APIs being included at some point in the future, but it would not be as part of these proposals.

[https://forums.swift.org/t/update-on-se-0382-and-se-0389-expression-macros-and-attached-macros/74094](https://forums.swift.org/t/update-on-se-0382-and-se-0389-expression-macros-and-attached-macros/74094)

Does this mean that an additional Swift Evolution proposal would be required for asynchronous expansion to be included in swift-syntax?

roopekv:
> I think we would need to see some significant real-world benefits from the switch to asynchronous macros in order to take this change.

I may have found a significant real-world benefit for all users of macros that could be made possible by asynchronous expansion: processing macro invocations in parallel within a single process, instead of launching each of them as separate processes.

I found a few threads on the Swift Forums lamenting the performance of Macros, mostly about the compile time of the macros themselves, but also concerning the macro invocation/expansion phase of the compilation ([example](https://forums.swift.org/t/emitting-module-takes-25x-the-time-in-xcode-15-1-beta/67671/23)). I believe that the introduction of asynchronous expansion could allow for a faster, data race safe, and possibly a more parallel expansion of macros, with the opportunity for significant compile time reduction for all macros, both synchronous and asynchronous.

The act of launching processes in itself has overhead, so not having to launch more than one process per plugin should be a performance win regardless of whether parallelism is utilized. And while multiple processes can be safely run in parallel too, that would occur with redundant memory usage compared to running multiple macro invocations in a single process on multiple threads. Asynchronous macro expansion via the Swift Concurrency model also guarantees data race safety and forward progress, making it safer than running macro invocations in parallel with other in-process methods like GCD. Macros would still run sandboxed, isolated to the plugin which they are defined in with no way to tamper with the memory of other processes. Consecutive and parallel macro invocations could affect each other if they explicitly modify and read shared memory within the plugin, but this could be considered a positive as it could allow for some clever caching behavior for advanced users.

The concept could be taken even further by linking macro plugins directly to the compiler as dynamic libraries, eliminating interprocess communication entirely, but it might not be feasible due to security concerns compared to running them in separate sandboxed processes.

What do you think, would this make any sense?

roopekv:
Yes, utilizing parallelism within the macro expansion functions would only be beneficial in certain cases, which we have already discussed.

But in my previous comment I was talking about parallelism in relation to processing multiple macro invocations within a single macro plugin process, without the overhead of launching multiple processes or being limited to running them sequentially.

roopekv:
But if the synchronous macro expansion calls can be wrapped and run in asynchronous tasks within the process, then the macro expansion wouldn't have to asynchronous to facilitate parallelism at the plugin server level after all I guess?

ahoppen:
> But if the synchronous macro expansion calls can be wrapped and run in asynchronous tasks within the process, then the macro expansion wouldn't have to asynchronous to facilitate parallelism at the plugin server level after all I guess?

Exactly.

PR #2897, opened 2024-11-17
Allow customization of expansion for function-typed placeholders

woolsweater:
This addresses <https://github.com/swiftlang/sourcekit-lsp/issues/1788>, refining code completion for closure placeholders.

By default function-typed placeholders will continue to expand to multi-line trailing form. A caller, such as sourcekit-lsp, may now customize the behavior by passing its own formatter. It may additionally request that the closure itself be marked as a placeholder, with the argument and return type placeholders nested inside.

### DIFF
--- a/Release Notes/602.md
+++ b/Release Notes/602.md
@@ -1,0 +1,26 @@
+# Swift Syntax 602 Release Notes
+
+## New APIs
+
+## API Behavior Changes
+
+## Deprecations
+
+## API-Incompatible Changes
+
+- `ExpandEditorPlaceholdersToTrailingClosures` has changed to `ExpandEditorPlaceholdersToLiteralClosures`
+  - Description: Whether function-typed placeholders are expanded to trailing closures is now configurable using a `format` argument to this rewriter. Additionally clients that support nested placeholders may request that the entire expanded closure be wrapped in an outer placeholder, e.g. `<#{ <#foo#> in <#Bar#> }#>`.
+  - Pull Request: https://github.com/swiftlang/swift-syntax/pull/2897
+  - Migration steps: Replace uses of `ExpandEditorPlaceholdersToTrailingClosures` with `ExpandEditorPlaceholdersToLiteralClosures`. The initializer does not need to change: `.init(indentationWidth:)` on the new type provides the same behavior as the old type.
+  - Notes: This improves code completion in a SourceKitLSP session where the trailing closure form may be undesirable. The nested placeholders offer more flexibility to end users, in editors that support it.
+
+## Template
+
+- *Affected API or two word description*
+  - Description: *A 1-2 sentence description of the new/modified API*
+  - Issue: *If an issue exists for this change, a link to the issue*
+  - Pull Request: *Link to the pull request(s) that introduces this change*
+  - Migration steps: Steps that adopters of swift-syntax should take to move to the new API (required for deprecations and API-incompatible changes).
+  - Notes: *In case of deprecations or API-incompatible changes, the reason why this change was made and the suggested alternative*
+
+*Insert entries in chronological order, with newer entries at the bottom*
--- a/Sources/SwiftRefactor/ExpandEditorPlaceholder.swift
+++ b/Sources/SwiftRefactor/ExpandEditorPlaceholder.swift
@@ -40,7 +40,9 @@ import SwiftSyntaxBuilder
 /// `type-for-expansion-string`), is parsed into a syntax node. If that node is
 /// a `FunctionTypeSyntax` then the placeholder is expanded into a
 /// `ClosureExprSyntax`. Otherwise it is expanded as is, which is also the case
-/// for when only a display string is provided.
+/// for when only a display string is provided. You may customize the formatting
+/// of a closure expansion via ``Context/closureLiteralFormat``, for example to
+/// change whether it is split onto multiple lines.
 ///
 /// ## Function Typed Placeholder
 /// ### Before
@@ -78,12 +80,29 @@ import SwiftSyntaxBuilder
 /// ```
 struct ExpandSingleEditorPlaceholder: EditRefactoringProvider {
   struct Context {
-    let indentationWidth: Trivia?
-    let initialIndentation: Trivia
+    /// The formatter to use when expanding a function-typed placeholder.
+    let closureLiteralFormat: BasicFormat
+    /// When true, the expansion will wrap a function-typed placeholder's entire
+    /// expansion in placeholder delimiters, in addition to any placeholders
+    /// inside the expanded closure literal.
+    ///
+    /// With `allowNestedPlaceholders = false`
+    /// ```swift
+    /// { someInt in <#String#> }
+    /// ```
+    ///
+    /// With `allowNestedPlaceholders = true`
+    /// ```swift
+    /// <#{ someInt in <#String#> }#>
+    /// ```
+    let allowNestedPlaceholders: Bool
 
-    init(indentationWidth: Trivia? = nil, initialIndentation: Trivia = []) {
-      self.indentationWidth = indentationWidth
-      self.initialIndentation = initialIndentation
+    init(
+      closureLiteralFormat: BasicFormat = BasicFormat(),
+      allowNestedPlaceholders: Bool = false
+    ) {
+      self.closureLiteralFormat = closureLiteralFormat
+      self.allowNestedPlaceholders = allowNestedPlaceholders
     }
   }
 
@@ -94,16 +113,17 @@ struct ExpandSingleEditorPlaceholder: EditRefactoringProvider {
 
     let expanded: String
     if let functionType = placeholder.typeForExpansion?.as(FunctionTypeSyntax.self) {
-      let basicFormat = BasicFormat(
-        indentationWidth: context.indentationWidth,
-        initialIndentation: context.initialIndentation
-      )
-      var formattedExpansion = functionType.closureExpansion.formatted(using: basicFormat).description
+      let format = context.closureLiteralFormat
+      let initialIndentation = format.currentIndentationLevel
+      var formattedExpansion = functionType.closureExpansion.formatted(using: format).description
       // Strip the initial indentation from the placeholder itself. We only introduced the initial indentation to
       // format consecutive lines. We don't want it at the front of the initial line because it replaces an expression
       // that might be in the middle of a line.
-      if formattedExpansion.hasPrefix(context.initialIndentation.description) {
-        formattedExpansion = String(formattedExpansion.dropFirst(context.initialIndentation.description.count))
+      if formattedExpansion.hasPrefix(initialIndentation.description) {
+        formattedExpansion = String(formattedExpansion.dropFirst(initialIndentation.description.count))
+      }
+      if context.allowNestedPlaceholders {
+        formattedExpansion = wrapInPlaceholder(formattedExpansion)
       }
       expanded = formattedExpansion
     } else {
@@ -161,20 +181,24 @@ public struct ExpandEditorPlaceholder: EditRefactoringProvider {
       let arg = placeholder.parent?.as(LabeledExprSyntax.self),
       let argList = arg.parent?.as(LabeledExprListSyntax.self),
       let call = argList.parent?.as(FunctionCallExprSyntax.self),
-      let expandedTrailingClosures = ExpandEditorPlaceholdersToTrailingClosures.expandTrailingClosurePlaceholders(
+      let expandedClosures = ExpandEditorPlaceholdersToLiteralClosures.expandClosurePlaceholders(
         in: call,
         ifIncluded: arg,
-        indentationWidth: context.indentationWidth
+        context: ExpandEditorPlaceholdersToLiteralClosures.Context(
+          format: .trailing(indentationWidth: context.indentationWidth)
+        )
       )
     else {
       return ExpandSingleEditorPlaceholder.textRefactor(syntax: token)
     }
 
-    return [SourceEdit.replace(call, with: expandedTrailingClosures.description)]
+    return [SourceEdit.replace(call, with: expandedClosures.description)]
   }
 }
 
-/// Expand all the editor placeholders in the function call that can be converted to trailing closures.
+/// Expand all the editor placeholders in the function call to literal closures.
+/// By default they will be expanded to trailing form; if you provide your own
+/// formatter via ``Context/format`` they will be expanded inline.
 ///
 /// ## Before
 /// ```swift
@@ -185,7 +209,7 @@ public struct ExpandEditorPlaceholder: EditRefactoringProvider {
 /// )
 /// ```
 ///
-/// ## Expansion of `foo`
+/// ## Expansion of `foo`, default behavior
 /// ```swift
 /// foo(
 ///   arg: <#T##Int#>,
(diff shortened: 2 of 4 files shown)


Sources/SwiftBasicFormat/ClosureLiteralFormat.swift
```diff
@@ -0,0 +1,65 @@
...
+#else
+@_spi(RawSyntax) import SwiftSyntax
+#endif
+
+/// A specialization of `BasicFormat` for closure literals, which is more
+/// conservative about newline insertion.
+///
+/// A closure that seems to be a simple predicate or transform — based on the
+/// number of enclosed statements — will not be reformatted to multiple lines.
+open class ClosureLiteralFormat: BasicFormat {
```
woolsweater, line 24: ℹ️ I wasn't confident that tweaking `BasicFormat` directly wouldn't break something else, so I opted to introduce a new formatted. It's easy enough to roll this into `BasicFormat` if preferred.

Sources/SwiftRefactor/ExpandEditorPlaceholder.swift
```diff
@@ -94,18 +92,18 @@ struct ExpandSingleEditorPlaceholder: EditRefactoringProvider {
...
-      var formattedExpansion = functionType.closureExpansion.formatted(using: basicFormat).description
+      var formattedExpansion = functionType.closureExpansion.formatted(using: format).description
       // Strip the initial indentation from the placeholder itself. We only introduced the initial indentation to
       // format consecutive lines. We don't want it at the front of the initial line because it replaces an expression
       // that might be in the middle of a line.
       if formattedExpansion.hasPrefix(context.initialIndentation.description) {
         formattedExpansion = String(formattedExpansion.dropFirst(context.initialIndentation.description.count))
       }
-      expanded = formattedExpansion
+      expanded = wrapInPlaceholder(formattedExpansion)
```
woolsweater, line 106: ℹ️ I think wrapping the whole closure in a placeholder, with inner placeholders for the args and body, provides an optimal UX: you can immediately delete/replace the entire thing or you can tab to the constituent parts. This does necessitate a change in sourcekit-lsp's placeholder translation routine to handle the nesting; this is done in https://github.com/swiftlang/sourcekit-lsp/pull/1831

Sources/SwiftSyntax/SyntaxProtocol.swift
```diff
@@ -235,7 +235,10 @@ extension SyntaxProtocol {
     return self.previousToken(viewMode: .sourceAccurate)
   }
 
-  /// Returns this node or the first ancestor that satisfies `condition`.
+  /// Applies `map` to this node and each of its ancestors until a non-`nil`
```
woolsweater, line 238: ℹ️ While working on the change I noticed this doc comment was inaccurate.

Tests/SwiftRefactorTest/ExpandEditorPlaceholderTests.swift
```diff
@@ -18,6 +18,12 @@ import SwiftSyntaxBuilder
 import XCTest
 import _SwiftSyntaxTestSupport
 
+fileprivate extension DefaultStringInterpolation {
+  mutating func appendInterpolation(placeholder string: String) {
+    self.appendLiteral(wrapInPlaceholder(string))
```
woolsweater, line 23: ℹ️ Helper to keep literal `<#` and `#>` delimiters out of the test assertions.

Sources/SwiftBasicFormat/ClosureLiteralFormat.swift
```diff
@@ -0,0 +1,65 @@
...
+      return super.requiresNewline(between: first, and: second)
+    }
+  }
+
+  /// Returns `true` if `token` is an opening or closing brace (according to
+  /// `kind`) of a closure, and that closure has no more than one statement in
+  /// its body.
+  private func isSmallClosureDelimiter(
+    _ token: TokenSyntax,
+    kind: KeyPath<ClosureExprSyntax, TokenSyntax>
```
ahoppen, line 42: Do we need to pass in the key path here? It doesn’t really look very dynamic, so I’d just check for `\.leftBrace` and `\.rightBrace` in here.

Sources/SwiftBasicFormat/ClosureLiteralFormat.swift
```diff
@@ -0,0 +1,65 @@
...
+      return false
+    }
+
+    return closure.statements.count <= 1
+  }
+
+  /// Returns `true` if `token` is the last token in the signature of a closure,
+  /// and that closure has no more than one statement in its body.
+  private func isEndOfSmallClosureSignature(_ token: TokenSyntax) -> Bool {
+    guard let signature = token.ancestorOrSelf(mapping: { $0.as(ClosureSignatureSyntax.self) }),
```
ahoppen, line 56: Do we need to generically look for the last token in the closure signature here or could we just look for the `in` keyword that terminates the signature?

Sources/SwiftRefactor/ExpandEditorPlaceholder.swift
```diff
@@ -50,9 +50,7 @@ import SwiftSyntaxBuilder
 ///
 /// ### After
 /// ```swift
-/// { someInt in
-///   <#T##String#>
-/// }
+/// <#{ <#someInt#> in <#T##String#> }#>
```
ahoppen, line 53: If we expand the editor placeholder, we shouldn’t end up with an editor placeholder again.

Sources/SwiftRefactor/ExpandEditorPlaceholder.swift
ahoppen: Unfortunately, we can’t do that. The Swift compiler and Xcode don’t support nested editor placeholders using the `<# … #>` syntax. But I do agree that wrapping the closure in a placeholder again makes sense in the context of automatic SourceKit-LSP expansions. Do you think we could do the re-wrapping on the SourceKit-LSP side?

Sources/SwiftRefactor/ExpandEditorPlaceholder.swift
```diff
@@ -94,18 +92,18 @@ struct ExpandSingleEditorPlaceholder: EditRefactoringProvider {
 
     let expanded: String
     if let functionType = placeholder.typeForExpansion?.as(FunctionTypeSyntax.self) {
-      let basicFormat = BasicFormat(
+      let format = ClosureLiteralFormat(
```
ahoppen, line 95: Could we make the format that’s used to expand the editor placeholder an option that still defaults to `BasicFormat`? For example, when you explicitly run the refactoring to expand a placeholder to a trailing closure, I think it makes sense to have the multi-line behavior. Then we could just override it with `ClosureLiteralFormat` for SourceKit-LSP code completion expansion. That way `ClosureLiteralFormat` could also be specified in SourceKit-LSP and wouldn’t become public API of swift-syntax.

Sources/SwiftRefactor/ExpandEditorPlaceholder.swift
```diff
@@ -137,12 +135,10 @@ struct ExpandSingleEditorPlaceholder: EditRefactoringProvider {
...
-///   closure2: { ... }
-/// ) { someInt in
-///   <#T##String#>
-/// } closure2: { someInt in
-///   <#T##String#>
-/// }
+///   closure2: { ... },
+///   closure3: { <#someInt#> in <#T##String#> },
+///   closure4: { <#someInt#> in <#T##String#> }
+/// )
```
ahoppen, line 141: Similar here, I think that the general `ExpandSingleEditorPlaceholder` refactoring should produce trailing closures, we just want to tweak the behavior from SourceKit-LSP. Maybe we need an option for that.

ahoppen (commented):
Thanks for the changes @woolsweater. I don’t want to change the general behavior of the refactorings because there are applications where the current behavior makes sense, so we’d need a way to only get the modified behavior for SourceKit-LSP.

Tests/SwiftRefactorTest/ExpandEditorPlaceholderTests.swift
woolsweater: @ahoppen I assume you would prefer I remove this, per [your comment on the same method in the sourcekit-lsp PR](https://github.com/swiftlang/sourcekit-lsp/pull/1831#discussion_r1847448599)?

Sources/SwiftRefactor/ExpandEditorPlaceholder.swift
woolsweater: Darn, I was afraid that might be the case. Okay, following this and your other comments, I'll figure out how to move the bulk of the behavior change into sourcekit-lsp.

woolsweater (commented):
Thank you for the prompt reviews here and in sourcekit-lsp, @ahoppen! I will work on the revisions you've proposed.

Tests/SwiftRefactorTest/ExpandEditorPlaceholderTests.swift
ahoppen: Yes, correct.

Sources/SwiftBasicFormat/ClosureLiteralFormat.swift
woolsweater: That seems fine; done!

Sources/SwiftRefactor/ExpandEditorPlaceholder.swift
woolsweater: Oops, reverted!

Sources/SwiftRefactor/ExpandEditorPlaceholder.swift
woolsweater: I don't see a (good) way to move the nested wrapping into SourceKit-LSP; we'd essentially have to re-do most of this work: re-parsing the expansion and finding the closure body to wrap. I have added a flag to the `Context` to control this. It is off by default and `CodeCompletionSession` will set it for its own expansion requests.

Sources/SwiftRefactor/ExpandEditorPlaceholder.swift
```diff
@@ -78,12 +80,15 @@ import SwiftSyntaxBuilder
...
 struct ExpandSingleEditorPlaceholder: EditRefactoringProvider {
   struct Context {
-    let indentationWidth: Trivia?
-    let initialIndentation: Trivia
-
-    init(indentationWidth: Trivia? = nil, initialIndentation: Trivia = []) {
-      self.indentationWidth = indentationWidth
-      self.initialIndentation = initialIndentation
+    let closureLiteralFormat: BasicFormat
+    let allowNestedPlaceholders: Bool
```
ahoppen, line 84: Could you add a short doc comment what `allowNestedPlaceholders` does, ideally with a short example. I think that would be useful in the future.

Sources/SwiftRefactor/ExpandEditorPlaceholder.swift
```diff
@@ -195,45 +205,72 @@ public struct ExpandEditorPlaceholder: EditRefactoringProvider {
 ///   <#T##String#>
 /// }
 /// ```
-public struct ExpandEditorPlaceholdersToTrailingClosures: SyntaxRefactoringProvider {
+public struct ExpandEditorPlaceholdersToLiteralClosures: SyntaxRefactoringProvider {
```
ahoppen, line 208: Because this is a change to a public API, we should have a release note entry for it. Could you create a `Release Notes/602.md` file and add the change to it?

ahoppen (commented):
Nice. Thank you, @woolsweater. Just a couple minor comments.

Sources/SwiftRefactor/ExpandEditorPlaceholder.swift
woolsweater: Done!

Sources/SwiftRefactor/ExpandEditorPlaceholder.swift
woolsweater: Absolutely; done!

Release Notes/602.md
ahoppen, line 1: Sorry, I should have maybe been a little more specific. Could you also include the other headings and the template? Ie. this file should look as follows before adding your entry:

```md
# Swift Syntax 601 Release Notes

## New APIs

## API Behavior Changes

## Deprecations

## Template

- *Affected API or two word description*
  - Description: *A 1-2 sentence description of the new/modified API*
  - Issue: *If an issue exists for this change, a link to the issue*
  - Pull Request: *Link to the pull request(s) that introduces this change*
  - Migration steps: Steps that adopters of swift-syntax should take to move to the new API (required for deprecations and API-incompatible changes).
  - Notes: *In case of deprecations or API-incompatible changes, the reason why this change was made and the suggested alternative*

*Insert entries in chronological order, with newer entries at the bottom*

```

Sources/SwiftRefactor/ExpandEditorPlaceholder.swift
```diff
@@ -78,12 +80,30 @@ import SwiftSyntaxBuilder
...
+    /// ## With nested placeholders allowed
+    /// ### Before
+    /// ```swift
+    /// <#T##(Int) -> String##(Int) -> String##(_ someInt: Int) -> String#>
+    /// ```
+    ///
+    /// ### After
+    /// ```swift
+    /// <#{ someInt in <#String#> }#>
+    /// ```
```
ahoppen, line 98: I’m not sure `Before` and `After` are the correct headings here. I would suggest

```suggestion
    /// With `allowNestedPlaceholders = false`
    /// ```swift
    /// <#T##(Int) -> String##(Int) -> String##(_ someInt: Int) -> String#>
    /// ```
    ///
    /// With `allowNestedPlaceholders = true`
    /// ```swift
    /// <#{ someInt in <#String#> }#>
    /// ```
```

ahoppen (commented):
Thanks for addressing my last comments. Two more quick follow-ups. We’re really close.

Sources/SwiftRefactor/ExpandEditorPlaceholder.swift
woolsweater: Sure, was following the wording from the doc comment above this; changed!

Release Notes/602.md
woolsweater: No problem; done!

ahoppen:
https://github.com/swiftlang/sourcekit-lsp/pull/1831

@swift-ci Please test

ahoppen:
https://github.com/swiftlang/sourcekit-lsp/pull/1831

@swift-ci Please test

ahoppen:
Looks like CI is failing for two reasons:
1. Could you format your changes using swift-format
2. Compilation of SourceKit-LSP is failing with (+ maybe other similar issues, I didn’t scan the entire log)

```
/home/build-user/sourcekit-lsp/Tests/SourceKitLSPTests/ClosureCompletionFormatTests.swift:120:62: error: initializer 'init(integerLiteral:)' is not available due to missing import of defining module 'SwiftSyntaxBuilder'
 17 | @_spi(Testing) import SourceKitLSP
 18 | import XCTest
 19 | 
    | `- note: add import of module 'SwiftSyntaxBuilder'
 20 | fileprivate func assertFormatted<T: SyntaxProtocol>(
 21 |   tree: T,
    :
118 |       tree: ClosureExprSyntax(
119 |         statements: CodeBlockItemListSyntax([
120 |           CodeBlockItemSyntax(item: CodeBlockItemSyntax.Item(IntegerLiteralExprSyntax(integerLiteral: 2)))
    |                                                              `- error: initializer 'init(integerLiteral:)' is not available due to missing import of defining module 'SwiftSyntaxBuilder'
121 |         ])
122 |       ),
```

woolsweater:
> 1. Could you format your changes using swift-format

Oops, had a stray empty line, sorry! Fixed.

> 2. Compilation of SourceKit-LSP is failing with (+ maybe other similar issues, I didn’t scan the entire log)
> 
> ```
> /home/build-user/sourcekit-lsp/Tests/SourceKitLSPTests/ClosureCompletionFormatTests.swift:120:62: error: initializer 'init(integerLiteral:)' is not available due to missing import of defining module 'SwiftSyntaxBuilder'

Huh, is import behavior different on Linux? I only have a Mac to test on. Well, I have added that import and that's the only error I see in the logs.

ahoppen:
> Huh, is import behavior different on Linux? I only have a Mac to test on. Well, I have added that import and that's the only error I see in the logs.

I think it’s because the diagnostic is created by the [`MemberImportVisibility`](https://github.com/swiftlang/sourcekit-lsp/blob/3131ca3c81c2895d7e319bb10abab1d9801f4bd9/Package.swift#L9) upcoming feature and your local compiler probably doesn’t have it (I can’t remember whether it was introduced by 6.0 or 6.1).

ahoppen:
https://github.com/swiftlang/sourcekit-lsp/pull/1831

@swift-ci Please test

woolsweater:
> > Huh, is import behavior different on Linux? I only have a Mac to test on. Well, I have added that import and that's the only error I see in the logs.
> 
> I think it’s because the diagnostic is created by the [`MemberImportVisibility`](https://github.com/swiftlang/sourcekit-lsp/blob/3131ca3c81c2895d7e319bb10abab1d9801f4bd9/Package.swift#L9) upcoming feature and your local compiler probably doesn’t have it (I can’t remember whether it was introduced by 6.0 or 6.1).

Ah, I see, thanks. I am an Xcode point release or two behind the latest 🙂

ahoppen:
https://github.com/swiftlang/sourcekit-lsp/pull/1831

@swift-ci Please test Windows

ahoppen:
https://github.com/swiftlang/sourcekit-lsp/pull/1831

@swift-ci Please test Windows

PR #2899, opened 2024-11-19
Fix an infinite loop if a conflict marker is found but it's not at the start of a new line

ahoppen:
Until I found the underlying issue, I was looking at performance issues with conflict marker lexing and was able to improve the parsing time of a file consisting of only conflict markers by 2x (not that it matters much in practice but doesn’t hurt either). That’s included in a second commit.

rdar://137746823


ahoppen:
@swift-ci Please test

rintaro (approved):
nice

DougGregor (approved):
Nice catch

PR #2901, opened 2024-11-20
[6.1] Fix an infinite loop if a conflict marker is found but it's not at the start of a new line

ahoppen:
- **Explanation**: The parser would enter into an infinite loop when a closing conflict marker was not at the start of a line. Until I found the underlying issue, I was looking at performance issues with conflict marker lexing and was able to improve the parsing time of a file consisting of only conflict markers by 2x (not that it matters much in practice but doesn’t hurt either). That’s included in a second commit.
- **Scope**: Lexing of conflict markers
- **Risk**: Low
- **Testing**: Added a test case / manually tested performance of conflict marker parsing
- **Issue**: rdar://137746823
- **Reviewer**:   @DougGregor @rintaro on https://github.com/swiftlang/swift-syntax/pull/2899

### DIFF
--- a/Sources/SwiftParser/Lexer/Cursor.swift
+++ b/Sources/SwiftParser/Lexer/Cursor.swift
@@ -292,7 +292,7 @@ extension Lexer {
       self.stateStack.perform(stateTransition: stateTransition, stateAllocator: stateAllocator)
     }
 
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

PR #2904, opened 2024-12-01
Swift package manifest refactoring actions

DougGregor:
Port the package manifest editing refactoring actions over to `SwiftRefactor`. These refactoring operations come from the swift-package-manager repository, and are being moved here for several reasons:

1. They are shared by SwiftPM and SourceKit-LSP already, and are likely to also get used in the compiler, so swift-syntax is a good place for them to be to avoid code duplication.
2. The `SwiftRefactor` module is where we are collecting various refactoring.
3. Putting them here completely separates them from the semantic package model, so it's clear that we are just dealing with syntactic transforms.

This ports all of the package manifest refactoring actions over from swift-package-manager (add package dependency, add target, add product, add target dependency) and introduces a new one that will be used by swift-package-manager in the future, "add plugin usage". The code itself is little changed from the SwiftPM version, although the `ManifestEditRefactoringProvider` protocol that unifies them all is new.

Note that we end up bringing over syntactic versions of various SwiftPM types like `TargetDescription`,  but in a string-backed, simplified form that is purely syntactic. These are needed to community with the refactoring actions, and can easily be constructed by clients.


DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test Windows

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test macOS

Tests/SwiftRefactorTest/ManifestEditTests.swift
```diff
@@ -0,0 +1,728 @@
...
+  }
+
+  XCTAssertEqual(
+    edits.auxiliaryFiles.count,
+    expectedAuxiliarySources.count,
+    "didn't get all of the auxiliary files we expected"
+  )
+}
+
+func XCTAssertThrows<T: Swift.Error, Ignore>(
```
ahoppen, line 714: Why doesn’t `XCTAssertThrowsError` from XCTest work here? If it doesn’t work, I would name this `assertThrows` because we don’t want to define local functions with the `XCT` prefix. Also, I would make this `fileprivate` to avoid it leaking into other tests.

ahoppen (commented):
Thanks for moving the code, Doug.

Sources/SwiftRefactor/PackageManifest/AbsolutePath.swift
```diff
@@ -0,0 +1,21 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// Syntactic wrapper type that describes an absolute path for refactoring
+/// purposes but does not interpret its contents.
+public struct AbsolutePath: CustomStringConvertible, Equatable, Hashable, Sendable {
```
bnbarham, line 15: I'd prefer not introducing so many fairly generic public types here (AbsolutePath, PackageIdentity, RelativePath, SemanticVersion, SourceControlURL). How much do you prefer this over just strings for these cases? Another alternative would be to namespace them in an enum, but I believe @ahoppen is generally not a fan of that.

Also, could we make `Target` -> `PackageTarget`?

Tests/SwiftRefactorTest/ManifestEditTests.swift
xedin: ✅ replaced `XCTAssertThrows` with `XCTAssertThrowsError`

Sources/SwiftRefactor/PackageManifest/AddPackageTarget.swift
```diff
@@ -0,0 +1,418 @@
...
+                range: insertPos..<insertPos,
+                replacement: newImport
+              )
+            )
+            break
+          }
+        }
+      }
+
+    default: break;
```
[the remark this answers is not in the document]
xedin: ✅

Sources/SwiftRefactor/PackageManifest/AddPackageTarget.swift
```diff
@@ -0,0 +1,418 @@
...
+
+    // Add dependencies needed for various targets.
+    switch target.type {
+    case .macro:
+      // Macro targets need to depend on a couple of libraries from
+      // SwiftSyntax.
+      target.dependencies.append(contentsOf: macroTargetDependencies)
+
+    default:
+      break;
```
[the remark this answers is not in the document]
xedin: ✅

Sources/SwiftRefactor/PackageManifest/SyntaxEditUtils.swift
```diff
@@ -0,0 +1,552 @@
...
+
+extension Trivia {
+  /// Determine whether this trivia has newlines or not.
+  var hasNewlines: Bool {
+    contains(where: \.isNewline)
+  }
+
+  /// Produce trivia from the last newline to the end, dropping anything
+  /// prior to that.
+  func onlyLastLine() -> Trivia {
```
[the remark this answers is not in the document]
xedin: ✅

Sources/SwiftRefactor/PackageManifest/SyntaxEditUtils.swift
```diff
@@ -0,0 +1,552 @@
...
+  /// position in this argument list.
+  ///
+  /// This operation will attempt to introduce trivia to match the
+  /// surrounding context where possible. The actual argument will be
+  /// created by the `generator` function, which is provided with leading
+  /// trivia and trailing comma it should use to match the surrounding
+  /// context.
+  func insertingArgument(
+    at position: SyntaxChildrenIndex,
+    generator: (Trivia, TokenSyntax?) -> LabeledExprSyntax
```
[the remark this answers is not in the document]
xedin: ✅

Sources/SwiftRefactor/PackageManifest/SyntaxEditUtils.swift
```diff
@@ -0,0 +1,552 @@
...
+  ) -> LabeledExprListSyntax {
+    // Turn the arguments into an array so we can manipulate them.
+    var arguments = Array(self)
+
+    let positionIdx = distance(from: startIndex, to: position)
+
+    let commaToken = TokenSyntax.commaToken()
```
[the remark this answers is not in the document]
xedin: ✅  removed this in favor of using `.commaToken()` where needed.

Sources/SwiftRefactor/PackageManifest/SyntaxEditUtils.swift
```diff
@@ -0,0 +1,552 @@
...
+    // Look for a call whose name is a string literal matching the
+    // requested target name.
+    func matchesTargetCall(call: FunctionCallExprSyntax) -> Bool {
+      guard let nameArgument = call.findArgument(labeled: "name") else {
+        return false
+      }
+
+      guard let stringLiteral = nameArgument.expression.as(StringLiteralExprSyntax.self),
+        let literalValue = stringLiteral.representedLiteralValue
+      else {
```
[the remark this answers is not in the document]
xedin: ✅

Sources/SwiftRefactor/PackageManifest/SyntaxEditUtils.swift
```diff
@@ -0,0 +1,552 @@
...
+extension ArrayExprSyntax {
+  /// Produce a new array literal expression that appends the given
+  /// element, while trying to maintain similar indentation.
+  func appending(
+    element: ExprSyntax,
+    outerLeadingTrivia: Trivia
+  ) -> ArrayExprSyntax {
+    var elements = self.elements
+
+    let commaToken = TokenSyntax.commaToken()
```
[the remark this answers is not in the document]
xedin: ✅  removed this in favor of using `.commaToken()` where needed.

Sources/SwiftRefactor/PackageManifest/SyntaxEditUtils.swift
```diff
@@ -0,0 +1,552 @@
...
+      // last element.
+      leadingTrivia = last.leadingTrivia.onlyLastLine()
+
+      // Add a trailing comma to the last element if it isn't already
+      // there.
+      if last.trailingComma == nil {
+        var newElements = Array(elements)
+        newElements[newElements.count - 1].trailingComma = commaToken
+        newElements[newElements.count - 1].expression.trailingTrivia =
+          Trivia()
```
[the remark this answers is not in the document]
xedin: ✅

Sources/SwiftRefactor/PackageManifest/SyntaxEditUtils.swift
```diff
@@ -0,0 +1,552 @@
...
+      append(label: label, stringLiteral: stringLiteral)
+    }
+  }
+
+  /// Append an array literal containing elements that can be rendered
+  /// into expression syntax nodes.
+  mutating func append<T>(
+    label: String?,
+    arrayLiteral: [T]
+  ) where T: ManifestSyntaxRepresentable, T.PreferredSyntax == ExprSyntax {
```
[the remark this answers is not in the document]
xedin: ✅

Sources/SwiftRefactor/PackageManifest/SyntaxEditUtils.swift
```diff
@@ -0,0 +1,552 @@
...
+
+  /// Append a potentially labeled argument with a string literal.
+  mutating func append(label: String?, stringLiteral: String) {
+    append(label: label, expression: "\(literal: stringLiteral)")
+  }
+
+  /// Append a potentially labeled argument with a string literal, but only
+  /// when the string literal is not nil.
+  mutating func appendIf(label: String?, stringLiteral: String?) {
+    if let stringLiteral {
```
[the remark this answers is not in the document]
xedin: ✅

Sources/SwiftRefactor/PackageManifest/SyntaxEditUtils.swift
```diff
@@ -0,0 +1,552 @@
...
+    )
+    append(label: label, expression: ExprSyntax(array))
+  }
+
+  /// Append an array literal containing elements that can be rendered
+  /// into expression syntax nodes.
+  mutating func appendIf<T>(
+    label: String?,
+    arrayLiteral: [T]?
+  ) where T: ManifestSyntaxRepresentable, T.PreferredSyntax == ExprSyntax {
```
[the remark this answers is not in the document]
xedin: ✅

Sources/SwiftRefactor/PackageManifest/SyntaxEditUtils.swift
```diff
@@ -0,0 +1,552 @@
...
+    guard let arrayLiteral else { return }
+    append(label: label, arrayLiteral: arrayLiteral)
+  }
+
+  /// Append an array literal containing elements that can be rendered
+  /// into expression syntax nodes, but only if it's not empty.
+  mutating func appendIfNonEmpty<T>(
+    label: String?,
+    arrayLiteral: [T]
+  ) where T: ManifestSyntaxRepresentable, T.PreferredSyntax == ExprSyntax {
```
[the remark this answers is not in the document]
xedin: ✅

Sources/SwiftRefactor/PackageManifest/SyntaxEditUtils.swift
```diff
@@ -0,0 +1,552 @@
...
+  ///   - label: The argument label for the argument whose array will be
+  ///     added or modified.
+  ///   - trailingLabels: The argument labels that could follow the label,
+  ///     which helps determine where the argument should be inserted if
+  ///     it doesn't exist yet.
+  ///   - newElement: The new element.
+  /// - Returns: the function call after making this change.
+  func appendingToArrayArgument(
+    label: String,
+    trailingLabels: Set<String>,
```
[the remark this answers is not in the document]
xedin: ✅

Sources/SwiftRefactor/PackageManifest/SyntaxEditUtils.swift
```diff
@@ -0,0 +1,552 @@
...
+      guard let argArray = arg.expression.findArrayArgument() else {
+        throw ManifestEditError.cannotFindArrayLiteralArgument(
+          argumentName: label,
+          node: Syntax(arg.expression)
+        )
+      }
+
+      // Format the element appropriately for the context.
+      let indentation = Trivia(
+        pieces: arg.leadingTrivia.filter { $0.isSpaceOrTab }
```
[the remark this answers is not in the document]
xedin: ✅

Sources/SwiftRefactor/PackageManifest/SyntaxEditUtils.swift
```diff
@@ -0,0 +1,552 @@
...
+
+    // Insert the new argument at the appropriate place in the call.
+    let insertionPos = arguments.findArgumentInsertionPosition(
+      labelsAfter: trailingLabels
+    )
+    let newArguments = arguments.insertingArgument(
+      at: insertionPos
+    ) { (leadingTrivia, trailingComma) in
+      // Format the element appropriately for the context.
+      let indentation = Trivia(pieces: leadingTrivia.filter { $0.isSpaceOrTab })
```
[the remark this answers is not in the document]
xedin: ✅

Sources/SwiftRefactor/PackageManifest/SyntaxEditUtils.swift
```diff
@@ -0,0 +1,552 @@
...
+      )
+      let formattedElement = newElement.formatted(using: format)
+        .cast(ExprSyntax.self)
+
+      // Form the array.
+      let newArgument = ArrayExprSyntax(
+        leadingTrivia: .space,
+        leftSquare: .leftSquareToken(
+          trailingTrivia: .newline
+        ),
```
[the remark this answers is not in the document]
xedin: ✅

Sources/SwiftRefactor/PackageManifest/SyntaxEditUtils.swift
```diff
@@ -0,0 +1,552 @@
...
+          [
+            ArrayElementSyntax(
+              expression: formattedElement,
+              trailingComma: .commaToken()
+            )
+          ]
+        ),
+        rightSquare: .rightSquareToken(
+          leadingTrivia: leadingTrivia
+        )
```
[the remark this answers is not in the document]
xedin: ✅

Sources/SwiftRefactor/PackageManifest/SyntaxEditUtils.swift
```diff
@@ -0,0 +1,552 @@
...
+  func replacingChild(_ childNode: Syntax, with newChildNode: Syntax) -> Self {
+    return ReplacingRewriter(
+      childNode: childNode,
+      newChildNode: newChildNode
+    ).rewrite(self).cast(Self.self)
+  }
+}
+
+extension FunctionCallExprSyntax {
+  /// Produce source edits that will add the given new element to the
```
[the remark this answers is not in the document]
xedin: ✅

Sources/SwiftRefactor/PackageManifest/SyntaxEditUtils.swift
```diff
@@ -0,0 +1,552 @@
...
+      if leftSquare.trailingTrivia.hasNewlines {
+        leftSquareTrailingTrivia = leftSquare.trailingTrivia
+      } else {
+        leftSquareTrailingTrivia = Trivia()
+      }
+    }
+
+    elements.append(
+      ArrayElementSyntax(
+        expression: element.with(\.leadingTrivia, leadingTrivia),
```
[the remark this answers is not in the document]
xedin: ✅ Same as above, the leading trivia for the new element would now only include indentation.

Sources/SwiftRefactor/PackageManifest/SyntaxEditUtils.swift
```diff
@@ -0,0 +1,552 @@
...
+
+    // Figure out leading trivia and adjust the prior argument (if there is
+    // one) by adding a comma, if necessary.
+    let leadingTrivia: Trivia
+    if position > startIndex {
+      let priorArgument = arguments[positionIdx - 1]
+
+      // Our leading trivia will be based on the prior argument's leading
+      // trivia.
+      leadingTrivia = priorArgument.leadingTrivia
```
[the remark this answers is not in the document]
xedin: ✅ Fixed this to only included indentation.

Sources/SwiftRefactor/PackageManifest/ProductDescription.swift
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    arguments.appendIfNonEmpty(
+      label: "targets",
+      arrayLiteral: targets
+    )
+
+    let separateParen: String = arguments.count > 1 ? "\n" : ""
+    let argumentsSyntax = LabeledExprListSyntax(arguments)
+    return ".\(raw: functionName)(\(argumentsSyntax)\(raw: separateParen))"
```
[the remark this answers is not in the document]
xedin: This is all about how `.append(label:expression:)` works, it appends a `\n` as a leadingTrivia of the first element if there is more than one of them, so what we end up with is:

- For one element: `<<name>>(<<argument>>)`
- For multiple: `<<name>>(\n<<argument1>>,\n<<argument2>>,\n ...\n)`

There is a test for this located in `ManifestEditTests.testAddLibraryProduct`

Sources/SwiftRefactor/PackageManifest/AbsolutePath.swift
xedin: ✅ 

- Renamed Target to PackageTarget
- Remove all of the placeholder types in favor of strings.

Sources/SwiftRefactor/PackageManifest/AbsolutePath.swift
```diff
@@ -0,0 +1,21 @@
...
+  public private(set) var description: String
```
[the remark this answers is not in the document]
xedin: This is no longer applicable since I removed the type.

Sources/SwiftRefactor/PackageManifest/AbsolutePath.swift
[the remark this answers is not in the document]
xedin: This is no longer applicable since I removed all the placeholder types.

Sources/SwiftRefactor/PackageManifest/PackageDependency.swift
```diff
@@ -0,0 +1,159 @@
...
+  }
+}
+
+extension PackageDependency.SourceControl: ManifestSyntaxRepresentable {
+  func asSyntax() -> ExprSyntax {
+    // TODO: Not handling identity, nameForTargetDependencyResolutionOnly,
+    // or productFilter yet.
+    switch location {
+    case .local:
+      fatalError()
```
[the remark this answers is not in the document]
xedin: ✅  I changed source control dependencies to accept string only since `PackageDescription` does that as well.

Sources/SwiftRefactor/PackageManifest/AddTargetDependency.swift
```diff
@@ -0,0 +1,90 @@
...
+    public init(dependency: Target.Dependency, targetName: String) {
+      self.dependency = dependency
+      self.targetName = targetName
+    }
+  }
+
+  /// The set of argument labels that can occur after the "dependencies"
+  /// argument in the various target initializers.
+  ///
+  /// TODO: Could we generate this from the the PackageDescription module, so
```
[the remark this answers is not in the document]
xedin: ✅ Removed all of the TODOs throughout.

Sources/SwiftRefactor/PackageManifest/AddPackageTarget.swift
```diff
@@ -0,0 +1,418 @@
...
+    public var testHarness: TestHarness
+
+    public let swiftSyntaxVersion: SemanticVersion
+    public let swiftTestingVersion: SemanticVersion
+
+    public init(
+      testHarness: TestHarness = .default,
+      swiftSyntaxVersion: SemanticVersion = SemanticVersion("600.0.0-latest"),
+      swiftTestingVersion: SemanticVersion = SemanticVersion("0.8.0")
+    ) {
```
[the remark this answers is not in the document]
xedin: ✅ Removed `SemanticVersion` and its uses, instead of this when `.macro` target needs to inject `swift-syntax` dependency we are going to use to a placeholder version and let the developers specify it.

xedin:
@swift-ci please test

Sources/SwiftRefactor/PackageManifest/ManifestEditRefactoringProvider.swift
```diff
@@ -0,0 +1,25 @@
...
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
+public protocol ManifestEditRefactoringProvider: EditRefactoringProvider
+where Self.Input == SourceFileSyntax {
+
+  static func manifestRefactor(syntax: SourceFileSyntax, in context: Context) throws -> PackageEdit
```
xedin, line 19: Note: I'm planning to follow-up to this PR and remove this, I think all we need to do is change result type of `textRefactor` to not only produce edits but also auxiliary files.

Sources/SwiftRefactor/PackageManifest/AddPackageDependency.swift
```diff
@@ -0,0 +1,138 @@
...
+    guard
+      let id = dependencyFnSyntax.arguments.first(where: {
+        $0.label?.text == "url" || $0.label?.text == "id" || $0.label?.text == "path"
+      })
+    else {
+      throw ManifestEditError.malformedManifest(error: "missing id or url argument in dependency syntax")
+    }
+
+    if let existingDependencies = packageCall.findArgument(labeled: "dependencies") {
+      // If we have an existing dependencies array, we need to check if
```
bnbarham, line 89: ... it's already added. I assume?

Sources/SwiftRefactor/PackageManifest/AddPackageDependency.swift
```diff
@@ -0,0 +1,138 @@
...
+//===----------------------------------------------------------------------===//
+
+import SwiftParser
+import SwiftSyntax
+import SwiftSyntaxBuilder
+
+/// Add a package dependency to a package manifest's source code.
+public struct AddPackageDependency: ManifestEditRefactoringProvider {
+  public struct Context {
+    public var dependency: PackageDependency
```
bnbarham, line 21: I'm tempted to say this should be optional and we could add placeholder if not given?

Sources/SwiftRefactor/PackageManifest/PackageDependency.swift
```diff
@@ -0,0 +1,149 @@
...
+
+/// Describes a package dependency for refactoring purposes. This is a syntactic
+/// subset of the full package manifest's description of a package dependency.
+public enum PackageDependency: Sendable {
+  case fileSystem(FileSystem)
+  case sourceControl(SourceControl)
+  case registry(Registry)
+
+  public struct FileSystem: Sendable {
+    public let identity: String
```
bnbarham: `identity` is set fairly rarely, seems like it should be optional

Sources/SwiftRefactor/PackageManifest/AddPackageTarget.swift
```diff
@@ -0,0 +1,400 @@
...
+    /// Create a test using the swift-testing package.
+    case swiftTesting = "swift-testing"
+
+    /// The default testing library to use.
+    public static let `default`: TestHarness = .swiftTesting
+  }
+
+  /// Additional configuration information to guide the package editing
+  /// process.
+  public struct Configuration {
```
bnbarham: Can this just be inlined into `Context`?

Sources/SwiftRefactor/PackageManifest/AddPackageTarget.swift
```diff
@@ -0,0 +1,400 @@
...
+//===----------------------------------------------------------------------===//
+
+import SwiftParser
+import SwiftSyntax
+import SwiftSyntaxBuilder
+
+/// Add a target to a manifest's source code.
+public struct AddPackageTarget: ManifestEditRefactoringProvider {
+  public struct Context {
+    public let target: PackageTarget
```
bnbarham, line 21: Same comment as adding a package dependency, I could see it being useful to add a placeholder target. Although maybe I'm wrong and you could just use completion for that. Thoughts?

Sources/SwiftRefactor/PackageManifest/AddPluginUsage.swift
```diff
@@ -0,0 +1,62 @@
...
+
+    public init(targetName: String, pluginUsage: PackageTarget.PluginUsage) {
+      self.targetName = targetName
+      self.pluginUsage = pluginUsage
+    }
+  }
+
+  /// The set of argument labels that can occur after the "plugins"
+  /// argument in the Target initializers. (There aren't any right now)
+  private static let argumentLabelsAfterPluginUsages: Set<String> = []
```
bnbarham: Do we need to keep this 🤔?

Sources/SwiftRefactor/PackageManifest/AddPackageDependency.swift
xedin: I'd think that the whole point of the refactoring is to actually add a dependency that's why it's non-optional at the moment, I'm not sure how we'd placeholder that efficiently because different dependency kinds have different arguments...

Sources/SwiftRefactor/PackageManifest/PackageDependency.swift
xedin: I think we can remove this and FileSystem and SourceControl could be identified by their path and location.

Sources/SwiftRefactor/PackageManifest/AddPackageTarget.swift
xedin: Sure, since the `Configuration` is just one thing now really, we can just inline it for now.

Sources/SwiftRefactor/PackageManifest/AddPluginUsage.swift
xedin: I don't think so, let's remove it for now.

Sources/SwiftRefactor/PackageManifest/AddPackageDependency.swift
xedin: ✅

Sources/SwiftRefactor/PackageManifest/PackageDependency.swift
xedin: ✅

Sources/SwiftRefactor/PackageManifest/AddPackageTarget.swift
xedin: ✅

Sources/SwiftRefactor/PackageManifest/AddPluginUsage.swift
xedin: ✅

Sources/SwiftRefactor/PackageManifest/ManifestEditRefactoringProvider.swift
```diff
@@ -0,0 +1,25 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
+public protocol ManifestEditRefactoringProvider: EditRefactoringProvider
```
ahoppen, line 16: Does this need to be a public protocol? I think it would be better of `EditRefactoringProvider` would gain the ability to modify auxiliary files. At that point a package refactoring wouldn’t be anything special, it would just be a normal refactoring.

Sources/SwiftRefactor/PackageManifest/ManifestEditRefactoringProvider.swift
```diff
@@ -0,0 +1,25 @@
...
+where Self.Input == SourceFileSyntax {
+
+  static func manifestRefactor(syntax: SourceFileSyntax, in context: Context) throws -> PackageEdit
+}
+
+extension EditRefactoringProvider where Self: ManifestEditRefactoringProvider {
```
ahoppen: Couldn’t this be simplified to the following? Or doesn’t that work?

```suggestion
extension ManifestEditRefactoringProvider {
```

Sources/SwiftRefactor/PackageManifest/PackageDependency.swift
```diff
@@ -0,0 +1,146 @@
...
+
+/// Describes a package dependency for refactoring purposes. This is a syntactic
+/// subset of the full package manifest's description of a package dependency.
+public enum PackageDependency: Sendable {
+  case fileSystem(FileSystem)
+  case sourceControl(SourceControl)
+  case registry(Registry)
+
+  public struct FileSystem: Sendable {
+    public let nameForTargetDependencyResolutionOnly: String?
```
ahoppen: That name is quite a mouthful. If this is the best name, could you add a comment that describes what it does? Maybe an example would be the best way to convey what both of these members do – also for `SourceControl` and `Registry`.

Sources/SwiftRefactor/PackageManifest/PackageDependency.swift
```diff
@@ -0,0 +1,146 @@
...
+    public enum Requirement: Sendable {
+      case exact(String)
+      case rangeFrom(String)
+      case range(lowerBound: String, upperBound: String)
+    }
+  }
+}
+
+extension PackageDependency: ManifestSyntaxRepresentable {
+  func asSyntax() -> ExprSyntax {
```
ahoppen, line 61: Would this make more sense as a computed property?

Sources/SwiftRefactor/PackageManifest/PackageDependency.swift
```diff
@@ -0,0 +1,146 @@
...
+
+extension PackageDependency.FileSystem: ManifestSyntaxRepresentable {
+  func asSyntax() -> ExprSyntax {
+    ".package(path: \(literal: path.description))"
+  }
+}
+
+extension PackageDependency.SourceControl: ManifestSyntaxRepresentable {
+  func asSyntax() -> ExprSyntax {
+    // TODO: Not handling identity, nameForTargetDependencyResolutionOnly,
```
ahoppen: Could you file an issue for this if you’re not planning to address it in this PR. My experience is that these TODOs will just be forgotten. And this would probably also make for a nice good first issue.

Sources/SwiftRefactor/PackageManifest/PackageDependency.swift
```diff
@@ -0,0 +1,146 @@
...
+
+    case .rangeFrom(let range):
+      return LabeledExprSyntax(
+        label: "from",
+        expression: range.asSyntax()
+      )
+
+    case .range(let lowerBound, let upperBound):
+      return LabeledExprSyntax(
+        expression: "\(lowerBound.asSyntax())..<\(upperBound.asSyntax())" as ExprSyntax
```
ahoppen: I was thinking a bit about whether we need to handle operator precedence here until I realized that `lowerBound.asSyntax()` always results in a string literal. Do we need the conformance of `String` to `ManifestSyntaxRepresentable` or wouldn’t it be clearer if we just used the `literal:` interpolation here, ie.


```suggestion
        expression: "\(literal: lowerBound)..<\(literal: upperBound)" as ExprSyntax
```

Sources/SwiftRefactor/PackageManifest/PackageEdit.swift
```diff
@@ -0,0 +1,23 @@
...
+import SwiftSyntax
+
+/// The result of editing a package, including any edits to the package
+/// manifest and any new files that are introduced.
+public struct PackageEdit {
+  /// Edits to perform to the package manifest.
+  public var manifestEdits: [SourceEdit] = []
+
+  /// Auxiliary files to write.
+  public var auxiliaryFiles: [(String, SourceFileSyntax)] = []
```
ahoppen: Is the `String` a file name or fully qualified file path? Also, I think adding tuple labels would make this easier to work with, eg.


```suggestion
  public var auxiliaryFiles: [(filePath: String, contents: SourceFileSyntax)] = []
```

Sources/SwiftRefactor/PackageManifest/PackageTarget.swift
```diff
@@ -0,0 +1,142 @@
...
+    arguments.appendIf(label: "url", stringLiteral: url)
+
+    // Only for plugins
+    arguments.appendIf(label: "checksum", stringLiteral: checksum)
+
+    if let pluginUsages {
+      arguments.appendIfNonEmpty(label: "plugins", arrayLiteral: pluginUsages)
+    }
+
+    let separateParen: String = arguments.count > 1 ? "\n" : ""
```
ahoppen, line 109: Instead of doing some formatting here, should we run `BasicFormat`?

Sources/SwiftRefactor/PackageManifest/ProductDescription.swift
```diff
@@ -0,0 +1,114 @@
...
+
+extension ProductDescription: ManifestSyntaxRepresentable {
+  /// The function name in the package manifest.
+  ///
+  /// Some of these are actually invalid, but it's up to the caller
+  /// to check the precondition.
+  private var functionName: String {
+    switch type {
+    case .executable: return "executable"
+    case .library(_): return "library"
```
ahoppen:
```suggestion
    case .library: return "library"
```

Sources/SwiftRefactor/PackageManifest/StringUtils.swift
```diff
@@ -0,0 +1,223 @@
...
+        // Hangul,
+        0xAC00...0xD7A3:
+        return $0
+      default:
+        return replacementUnichar
+      }
+    })
+
+    // Apply further restrictions to the prefix.
+    loop: for (idx, c) in mangledUnichars.enumerated() {
```
ahoppen: I think in the couple cases where we have control flow labels, we have made them all uppercase. Could we be consistent here as well and name this `LOOP`?

Sources/SwiftRefactor/PackageManifest/StringUtils.swift
```diff
@@ -0,0 +1,223 @@
...
+
+extension String {
+  /// Returns a form of the string that is valid C99 Extended Identifier (by
+  /// replacing any invalid characters in an unspecified but consistent way).
+  /// The output string is guaranteed to be non-empty as long as the input
+  /// string is non-empty.
+  func mangledToC99ExtendedIdentifier() -> String {
+    // Map invalid C99-invalid Unicode scalars to a replacement character.
+    let replacementUnichar: UnicodeScalar = "_"
+    var mangledUnichars: [UnicodeScalar] = self.unicodeScalars.map({
```
ahoppen:
```suggestion
    var mangledUnichars: [UnicodeScalar] = self.unicodeScalars.map {
```

Plus closing parenthesis

Sources/SwiftRefactor/PackageManifest/StringUtils.swift
```diff
@@ -0,0 +1,223 @@
...
+      switch $0.value {
+      case  // A-Z
```
ahoppen, line 24: Is there a definition of this list in some spec that we could link to? It feels pretty arbitrary as a standalone switch.

Sources/SwiftRefactor/PackageManifest/StringUtils.swift
```diff
@@ -0,0 +1,223 @@
...
+      default:
+        break loop
+      }
+    }
+
+    // Combine the characters as a string again and return it.
+    // FIXME: We should only construct a new string if anything changed.
+    // FIXME: There doesn't seem to be a way to create a string from an
+    //        array of Unicode scalars; but there must be a better way.
+    return mangledUnichars.reduce("") { $0 + String($1) }
```
ahoppen: How about

```swift
String(decoding: mangledUnichars.flatMap { $0.utf8 }, as: UTF8.self)
```

Not that it’s really relevant here but it would be interesting to see the different performance characteristics of these two implementations.

Sources/SwiftRefactor/PackageManifest/SyntaxEditUtils.swift
```diff
@@ -0,0 +1,546 @@
...
+
+extension Trivia {
+  /// Determine whether this trivia has newlines or not.
+  var hasNewlines: Bool {
+    contains(where: \.isNewline)
+  }
+
+  /// Produce trivia from the last newline to the end, dropping anything
+  /// prior to that.
+  var onlyLastLine: Trivia {
```
ahoppen: Is this used anywhere?

Sources/SwiftRefactor/PackageManifest/SyntaxEditUtils.swift
```diff
@@ -0,0 +1,546 @@
...
+    }
+
+    return Trivia(pieces: pieces[lastNewline...])
+  }
+}
+
+/// Syntax walker to find the first occurrence of a given node kind that
+/// matches a specific predicate.
+private class FirstNodeFinder<Node: SyntaxProtocol>: SyntaxAnyVisitor {
+  var predicate: (Node) -> Bool
```
ahoppen:
```suggestion
  let predicate: (Node) -> Bool
```

?

Sources/SwiftRefactor/PackageManifest/SyntaxEditUtils.swift
```diff
@@ -0,0 +1,546 @@
...
+  func findArgumentIndex(labeled label: String) -> LabeledExprListSyntax.Index? {
+    arguments.firstIndex { $0.label?.text == label }
+  }
+}
+
+extension LabeledExprListSyntax {
+  /// Find the index at which the one would insert a new argument given
+  /// the set of argument labels that could come after the argument we
+  /// want to insert.
+  func findArgumentInsertionPosition(
```
ahoppen: This could be `fileprivate` function, right?

Sources/SwiftRefactor/PackageManifest/SyntaxEditUtils.swift
```diff
@@ -0,0 +1,546 @@
...
+
+  /// Form a new argument list that inserts a new argument at the specified
+  /// position in this argument list.
+  ///
+  /// This operation will attempt to introduce trivia to match the
+  /// surrounding context where possible. The actual argument will be
+  /// created by the `generator` function, which is provided with leading
+  /// trivia and trailing comma it should use to match the surrounding
+  /// context.
+  func insertingArgument(
```
ahoppen: Could also be `fileprivate`, right?

Sources/SwiftRefactor/PackageManifest/SyntaxEditUtils.swift
```diff
@@ -0,0 +1,546 @@
...
+    // one) by adding a comma, if necessary.
+    let leadingTrivia: Trivia
+    if position > startIndex {
+      let priorArgument = arguments[positionIdx - 1]
+
+      // Our indentation will be based on the prior argument's.
+      if priorArgument.leadingTrivia.hasNewlines {
+        leadingTrivia = .newline + (priorArgument.firstToken(viewMode: .sourceAccurate)?.indentationOfLine ?? Trivia())
+      } else {
+        leadingTrivia = priorArgument.leadingTrivia.indentation(isOnNewline: false) ?? Trivia()
```
ahoppen, line 136: Ah, interesting misuse of `indentation(isOnNewline:)`. I suppose that it does what you want here (or something that’s at least good enough, I don’t know how we would like to handle the prior argument having comments in its leading trivia. But I think a short comment explaining what’s happening here would be useful.

Same for `ArrayExprSyntax.appending`

Sources/SwiftRefactor/PackageManifest/SyntaxEditUtils.swift
```diff
@@ -0,0 +1,546 @@
...
+        leadingTrivia = last.leadingTrivia.indentation(isOnNewline: false) ?? Trivia()
+      }
+
+      // Add a trailing comma to the last element if it isn't already
+      // there.
+      if last.trailingComma == nil {
+        var newElements = Array(elements)
+        newElements[newElements.count - 1].trailingComma = .commaToken()
+        newElements[newElements.count - 1].expression.trailingTrivia = Trivia()
+        newElements[newElements.count - 1].trailingTrivia = last.trailingTrivia
```
ahoppen: Would something like the following be simpler?

```suggestion
        newElements[newElements.count - 1].trailingComma = .commaToken(trailingTrivia: last.expression.trailingTrivia)
        newElements[newElements.count - 1].expression.trailingTrivia = Trivia()
```

Sources/SwiftRefactor/PackageManifest/SyntaxEditUtils.swift
```diff
@@ -0,0 +1,546 @@
...
+  }
+}
+
+// MARK: Utilities to oeprate on arrays of array literal elements.
+extension Array<ArrayElementSyntax> {
+  /// Append a new argument expression.
+  mutating func append(expression: ExprSyntax) {
+    // Add a comma on the prior expression, if there is one.
+    let leadingTrivia: Trivia?
+    if count > 0 {
```
ahoppen:
```suggestion
    if !isEmpty {
```

Sources/SwiftRefactor/PackageManifest/SyntaxEditUtils.swift
```diff
@@ -0,0 +1,546 @@
...
+      self[count - 1].trailingComma = TokenSyntax.commaToken()
+      leadingTrivia = .newline
+
+      // Adjust the first element to start with a newline
+      if count == 1 {
+        self[0].leadingTrivia = .newline
```
ahoppen, line 301: Do we need to insert indentation here?

Sources/SwiftRefactor/PackageManifest/SyntaxEditUtils.swift
```diff
@@ -0,0 +1,546 @@
...
+// MARK: Utilities to operate on arrays of call arguments.
+
+extension Array<LabeledExprSyntax> {
+  /// Append a potentially labeled argument with the argument expression.
+  mutating func append(label: String?, expression: ExprSyntax) {
+    // Add a comma on the prior expression, if there is one.
+    let leadingTrivia: Trivia
+    if count > 0 {
+      self[count - 1].trailingComma = TokenSyntax.commaToken()
+      leadingTrivia = .newline
```
ahoppen, line 326: Same here, do we need to add indentation?

Sources/SwiftRefactor/PackageManifest/SyntaxEditUtils.swift
```diff
@@ -0,0 +1,546 @@
...
+  }
+
+  /// Append a potentially labeled argument with a string literal.
+  mutating func append(label: String?, stringLiteral: String) {
+    append(label: label, expression: "\(literal: stringLiteral)")
+  }
+
+  /// Append a potentially labeled argument with a string literal, but only
+  /// when the string literal is not nil.
+  mutating func appendIf(label: String?, stringLiteral: String?) {
```
ahoppen, line 352: Are the `appendIf` methods used anywhere?

Sources/SwiftRefactor/PackageManifest/SyntaxEditUtils.swift
```diff
@@ -0,0 +1,546 @@
...
+    }
+
+    // Figure out the trivia for the left and right square
+    let leftSquareTrailingTrivia: Trivia
+    let rightSquareLeadingTrivia: Trivia
+    switch elements.count {
+    case 0:
+      // Put a single space between the square brackets.
+      leftSquareTrailingTrivia = Trivia()
+      rightSquareLeadingTrivia = .space
```
ahoppen, line 375: Do we really want a space between the brackets? I think swift-format doesn’t include a space in `[]` (nor do we include it in all of swift-syntax’s codebase).

Sources/SwiftRefactor/PackageManifest/SyntaxEditUtils.swift
```diff
@@ -0,0 +1,546 @@
...
+
+    case 1:
+      // Put spaces around the single element
+      leftSquareTrailingTrivia = .space
+      rightSquareLeadingTrivia = .space
```
ahoppen, line 380: Same here, odd choice to add a space around the element, IMO.

Sources/SwiftRefactor/PackageManifest/SyntaxEditUtils.swift
```diff
@@ -0,0 +1,546 @@
...
+      return newChildNode
+    }
+
+    return nil
+  }
+}
+
+fileprivate extension SyntaxProtocol {
+  /// Replace the given child with a new child node.
+  func replacingChild(_ childNode: Syntax, with newChildNode: Syntax) -> Self {
```
ahoppen, line 446: Do we ever replace nodes with nodes of a different type? If so, this would fail the `cast(Self.self)` if `self == childNode`. If we don’t need to replace nodes of different types, this could be

```suggestion
  func replacingChild<T: SyntaxProtocol>(_ childNode: T, with newChildNode: T) -> Self {
```

Sources/SwiftRefactor/PackageManifest/SyntaxEditUtils.swift
```diff
@@ -0,0 +1,546 @@
...
+    if let arg = findArgument(labeled: label) {
+      guard let argArray = arg.expression.findArrayArgument() else {
+        throw ManifestEditError.cannotFindArrayLiteralArgument(
+          argumentName: label
+        )
+      }
+
+      // Format the element appropriately for the context.
+      let indentation = Trivia(
+        pieces: arg.leadingTrivia.filter(\.isSpaceOrTab)
```
ahoppen, line 485: Should we be using `indentaitonOfLine` or something like that to cover if `arg` is not on a new line?

Sources/SwiftRefactor/PackageManifest/AddPackageTarget.swift
```diff
@@ -0,0 +1,388 @@
...
+      case .xctest:
+        importModuleNames.append("XCTest")
+
+      case .swiftTesting:
+        importModuleNames.append("Testing")
+      }
+    }
+
+    let importDecls = importModuleNames.lazy.sorted().map { name in
+      DeclSyntax("import \(raw: name)").with(\.trailingTrivia, .newline)
```
ahoppen: Can’t this just be

```suggestion
      DeclSyntax("import \(raw: name)\n")
```

Sources/SwiftRefactor/PackageManifest/AddPackageTarget.swift
```diff
@@ -0,0 +1,388 @@
...
+    let imports = CodeBlockItemListSyntax {
+      for importDecl in importDecls {
+        importDecl
+      }
+    }
+
+    let sourceFileText: SourceFileSyntax
+    switch target.type {
+    case .binary, .plugin, .system:
+      fatalError("should have exited above")
```
ahoppen, line 206: Where is `above`? I don’t see the check where this should have exited.

Sources/SwiftRefactor/PackageManifest/AddPackageTarget.swift
```diff
@@ -0,0 +1,388 @@
...
+fileprivate extension PackageTarget {
+  var sanitizedName: String {
+    name
+      .mangledToC99ExtendedIdentifier()
+      .localizedFirstWordCapitalized()
+  }
+}
+
+fileprivate extension String {
+  func localizedFirstWordCapitalized() -> String { prefix(1).uppercased() + dropFirst() }
```
ahoppen, line 358: What’s the `localize` part of this function name?

Sources/SwiftRefactor/PackageManifest/ManifestEditRefactoringProvider.swift
xedin: See my comment about this - https://github.com/swiftlang/swift-syntax/pull/2904#discussion_r2261560655

Sources/SwiftRefactor/PackageManifest/PackageDependency.swift
bnbarham: I was assuming `identity` is `name` here, which you can pass (though it's deprecated for the URL case). So at least for `FileSystem` it seems like it should be optional. For `SourceControl`, probably fine to leave out.

EDIT: Given there's the `nameForDependencyResolutionOnly` though, maybe that's not actually true?

Sources/SwiftRefactor/PackageManifest/AddPackageDependency.swift
bnbarham: > I'm not sure how we'd placeholder that efficiently

We'd just pick a reasonable default, ie. `SourceControl` and `from`:
```
.package(url: <#url#>, from: <#version#>)
```
(though those placeholders could also be typed)

Sources/SwiftRefactor/PackageManifest/PackageDependency.swift
bnbarham: Honestly seems like this could just be `name`, I don't think the "for target dependency resolution" part really matters

Sources/SwiftRefactor/PackageManifest/PackageDependency.swift
bnbarham: We should handle `name` now and remove the TODO

Sources/SwiftRefactor/PackageManifest/SyntaxEditUtils.swift
xedin: They are used in `PackageTarget`.

Sources/SwiftRefactor/PackageManifest/AddPackageDependency.swift
xedin: Maybe that is the way to go here but instead of having it optional we can make it a defaulted parameter on initializer in both cases.

Sources/SwiftRefactor/PackageManifest/ManifestEditRefactoringProvider.swift
xedin: ✅ You are right, it doesn't require `Self:` could be placed directly.

Sources/SwiftRefactor/PackageManifest/PackageDependency.swift
xedin: ✅

Sources/SwiftRefactor/PackageManifest/PackageEdit.swift
xedin: ✅ I think it's a relative path there so I labeled it as such.

Sources/SwiftRefactor/PackageManifest/ProductDescription.swift
xedin: ✅

Sources/SwiftRefactor/PackageManifest/StringUtils.swift
xedin: ✅

Sources/SwiftRefactor/PackageManifest/StringUtils.swift
xedin: ✅

Sources/SwiftRefactor/PackageManifest/StringUtils.swift
xedin: ✅

Sources/SwiftRefactor/PackageManifest/SyntaxEditUtils.swift
xedin: ✅  Nope, I changed the leadingTrivia handling and forgot to remove it, thank you!

Sources/SwiftRefactor/PackageManifest/SyntaxEditUtils.swift
xedin: ✅

Sources/SwiftRefactor/PackageManifest/SyntaxEditUtils.swift
xedin: ✅  Yep

Sources/SwiftRefactor/PackageManifest/SyntaxEditUtils.swift
xedin: ✅

Sources/SwiftRefactor/PackageManifest/SyntaxEditUtils.swift
xedin: ✅

Sources/SwiftRefactor/PackageManifest/SyntaxEditUtils.swift
xedin: ✅

Sources/SwiftRefactor/PackageManifest/AddPackageTarget.swift
xedin: ✅

xedin:
We have discussed this and decided to SPI new refactorings for now.

xedin:
@swift-ci please test

xedin:

ahoppen (approved):
Good with me, assuming all declarations are at most SPI (ie. no public declarations).

Sources/SwiftRefactor/PackageManifest/PackageDependency.swift
xedin: ✅ This TODO was outdated and I removed it, nameFor* is SwiftPM concept that shouldn't be involved here and source control could be identified by its location.

Sources/SwiftRefactor/PackageManifest/PackageDependency.swift
xedin: I removed this completely since we don't really need that for refactorings.

PR #2905, opened 2024-12-02
[SwiftLexicalLookup] Composite names

MAJKFL:
This PR introduces notion of composite names to `SwiftLexicalLookup`. In Swift, it’s possible to have a switch case introducing one name for several enum cases:
```swift
switch X {
case .x(let a), .y(let a):
  print(a) // Refers to both introduced names above
}
```
`ASTScope` currently deals with this special case by just introducing the first name from the list. Composite names are supposed to instead merge all names with the same semantic meaning into one, unified name representation. It stores an array of related names and, to match the behavior of the compiler, a composite name is *represented* by the first name. 

Combined with some additional minor fixes included in this PR, I was able to finally perform entire Swift compilation with fully enabled validation! There still seem to be some discrepancies when running entire test suite with `lit`, but I’m happy to see the library finally maturing.


MAJKFL:
@swift-ci Please test

Sources/SwiftLexicalLookup/LookupName.swift
```diff
@@ -139,6 +139,19 @@ import SwiftSyntax
...
+  /// ### Example:
+  /// ```swift
+  /// switch X {
+  /// case .a(let x), .b(let x):
+  ///   print(x) // <-- lookup here
+  /// }
+  /// ```
+  /// For lookup at the given position, the result
+  /// contains only one (composite) name
+  case compositeName([LookupName])
```
ahoppen, line 154: I don’t like to use `compositeName` here because we also call function names with argument labels (like `foo(a:b:)`) compound names, which is very close to composite name. How about something that like `multipleDeclarations` or something like that?

Sources/SwiftLexicalLookup/LookupName.swift
```diff
@@ -319,6 +338,20 @@ import SwiftSyntax
...
+
+      for (index, name) in names.enumerated() {
+        result += name.debugDescription
+
+        if index < names.count - 1 {
+          result += ", "
+        } else {
+          result += " ]"
+        }
+      }
```
ahoppen, line 352: Any reason why we can’t do 
```suggestion
      return "Composite name: [ \(names.map(\.debugDescription).joined(separator: ", "))]"
```

Sources/SwiftLexicalLookup/Scopes/ScopeImplementations.swift
```diff
@@ -574,13 +574,44 @@ import SwiftSyntax
...
+
+      if !partitioned.keys.contains(identifier) {
+        orderedKeys.append(identifier)
+      }
+
+      if partitioned[identifier] == nil {
+        partitioned[identifier] = [extractedName]
+      } else {
+        partitioned[identifier]?.append(extractedName)
+      }
```
ahoppen, line 605: I think this can be 
 
```suggestion
      partitioned[identifier, default: []].append(extractedName)
```

Tests/SwiftLexicalLookupTest/NameLookupTests.swift
```diff
@@ -803,13 +805,45 @@ final class testNameLookup: XCTestCase {
...
   }
 
+  func testCompositeNames() {
+    assertLexicalNameLookup(
+      source: """
+        switch X {
+        case .x(let 1️⃣a, let 2️⃣b), .y(let 3️⃣a, let 4️⃣b):
+          print(5️⃣x)
+        case .z(let 7️⃣a), .smth(let 8️⃣a):
+          print(9️⃣x)
```
ahoppen, line 822: Shouldn’t this be `print(a)`? `x` is never defined in this test case.

ahoppen (commented):
Nice

Sources/SwiftLexicalLookup/LookupName.swift
MAJKFL: Thanks for the suggestion! I replaced the old concatenation with this one.

Tests/SwiftLexicalLookupTest/NameLookupTests.swift
MAJKFL: I purposefully called it `x` to further indicate that we are not performing name matching in this case. We use `useNilAsTheParameter: true` for that. This test is supposed to check if we properly partition names equivalent names (like in the first case here).

MAJKFL:
@swift-ci Please test

Sources/SwiftLexicalLookup/LookupName.swift
ahoppen: `equivalentNames` sounds good to me. 👍🏽

Tests/SwiftLexicalLookupTest/NameLookupTests.swift
ahoppen: Oh, I missed that and thought that we were looking up `x`, not that we were collecting all names available at that location. The test makes sense now 😄

ahoppen (approved):
Looks good 👍🏽

ahoppen:
@swift-ci Please test Windows

MAJKFL:
@swift-ci Please test

MAJKFL:
@swift-ci Please test Windows

DougGregor:
@swift-ci Please test Windows

DougGregor:
@swift-ci Please test Windows

DougGregor:
@swift-ci please test platform Windows

DougGregor:
@swift-ci please test platform Windows

PR #2911, opened 2024-12-04
Fix a stack overflow when using `canRecoveryTo` with too many open delimiters

ahoppen:
`canRecoverTo` calls itself recursively if it finds a nested opening token, eg. when calling `canRecoverTo` on `{{{`. To avoid stack overflowing, limit the number of nested `canRecoverTo` calls we make. Since returning a recovery handle from this function only improves error recovery but is not necessary for correctness, bailing from recovery is safe.


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

rintaro (approved):

PR #2917, opened 2024-12-13
Use access levels on imports when `compiler(>=6)`

tshortli:
Access levels on `import` declarations are a language feature that was introduced with the Swift 6 compiler but they don't depend on the Swift 6 language mode. `@_implementationOnly import` is now deprecated and `internal import` should be used instead, so SwiftSyntax's use of access levels on imports should depend on the compiler version, not the language version, to ensure it no longer uses `@_implementationOnly` unnecessarily.

As a notable exception, `_SwiftSyntaxCShims` must continue to be imported `@_implementationOnly` since it's a local module that cannot be loaded by clients, and `internal import` does not hide modules from dependents when SwiftSyntax is built without library evolution.

### DIFF
--- a/CodeGeneration/Sources/generate-swift-syntax/templates/swiftparser/IsLexerClassifiedFile.swift
+++ b/CodeGeneration/Sources/generate-swift-syntax/templates/swiftparser/IsLexerClassifiedFile.swift
@@ -18,7 +18,7 @@ import Utils
 let isLexerClassifiedFile = SourceFileSyntax(leadingTrivia: copyrightHeader) {
   DeclSyntax(
     """
-    #if swift(>=6)
+    #if compiler(>=6)
     public import SwiftSyntax
     #else
     import SwiftSyntax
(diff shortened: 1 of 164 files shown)


tshortli:
We should also consider enabling the upcoming `InternalImportsByDefault` feature regardless of language mode, but I think that should be done in a separate PR.

tshortli:
@swift-ci please test

tshortli:
@swift-ci please test

ahoppen (approved):
Thanks, Allan.

ahoppen:
@swift-ci Please test Windows

tshortli:
I forgot to update code generation - fixed.

tshortli:
@swift-ci please test

tshortli:
@swift-ci please test

xymus (approved):

tshortli:
@swift-ci please test

PR #2922, opened 2024-12-18
Improve documentation comments of some commonly used types

Jeehut:
This resolves https://github.com/swiftlang/swift-syntax/issues/2906 for an initial set of types. To keep the PR relatively small, I have focused on improving docs of those types that I have personally needed & used for my first macro (which I haven't released yet, as it's not finished).

@ahoppen I didn't quite understand if your "I would" [in this comment](https://github.com/swiftlang/swift-syntax/issues/2906#issuecomment-2524190993) literally meant that it's something you would do after my PR is merged, of if it's something you suggested me to do.

In any case, I found it very confusing that all the types I searched for were inside a `generated` folder and the file names were suffixed by alphabetic characters like `QRS` in `SyntaxNodesQRS.swift`. I have a feeling that I added the documentation comments to the wrong place. But I couldn't find any mentions of `StringSegmentSyntax`, for example inside the `CodeGeneration` folder. In that particular case, I did find the text of the comment "A literal segment inside a string segment." inside the `ExprNodes.swift` file, but then again where to add the documentation comments for the initializer, which are very important when discovering APIs via code completion in Xcode? 🤔

I think I need some guidance here to adjust this PR properly. Thank you in advance for your patience!

### DIFF
--- a/Sources/SwiftSyntax/generated/SyntaxCollections.swift
+++ b/Sources/SwiftSyntax/generated/SyntaxCollections.swift
@@ -1397,21 +1397,56 @@ public struct SpecializeAttributeArgumentListSyntax: SyntaxCollection, SyntaxHas
   public static let syntaxKind = SyntaxKind.specializeAttributeArgumentList
 }
 
+/// A collection of string literal segments representing both plain text and interpolated expressions.
+///
+/// This collection is used to represent the contents of string literals, including both static text segments and interpolated expressions.
+/// For example, in the string literal `"Hello \(name)"`, the collection would contain a string segment for `"Hello "` followed by an expression segment for `\(name)`.
+///
+/// ```swift
+/// let segments = StringLiteralSegmentListSyntax([
+///   .stringSegment(.init(content: .stringSegment("Hello "))),
+///   .expressionSegment(/* interpolation expression */)
+/// ])
+/// ```
+///
+/// This type is commonly used in macros for:
+/// - Building string literals with both static text and interpolations
+/// - Validating string literal contents at compile time
+/// - Analyzing or transforming string interpolations
+///
 /// ### Children
-/// 
+///
 /// (``StringSegmentSyntax`` | ``ExpressionSegmentSyntax``) `*`
 ///
 /// ### Contained in
-/// 
+///
 ///  - ``StringLiteralExprSyntax``.``StringLiteralExprSyntax/segments``
 public struct StringLiteralSegmentListSyntax: SyntaxCollection, SyntaxHashable {
+  /// Represents either a literal string segment or an interpolated expression segment within a string literal.
+  ///
+  /// This enum is used to handle the two possible types of segments that can appear in a string literal:
+  /// - Plain text segments (like `"Hello "` in `"Hello \(name)"`)
+  /// - Expression segments (like `\(name)` in `"Hello \(name)"`)
+  ///
+  /// Example:
+  /// ```swift
+  /// let segments = [
+  ///   StringLiteralSegmentListSyntax.Element.stringSegment(.init(content: .stringSegment("Hello "))),
+  ///   StringLiteralSegmentListSyntax.Element.expressionSegment(/* expression */)
+  /// ]
+  /// ```
   public enum Element: SyntaxChildChoices, SyntaxHashable {
-    /// A literal segment inside a string segment.
-    /// 
+    /// A literal text segment inside a string literal.
+    ///
+    /// This case represents static text content like `"Hello "` in `"Hello \(name)"`.
+    ///
     /// - SeeAlso: ``ExpressionSegmentSyntax``
     case stringSegment(StringSegmentSyntax)
-    /// An interpolated expression inside a string literal.
-    /// 
+
+    /// An interpolated expression segment inside a string literal.
+    ///
+    /// This case represents interpolated expressions like `\(name)` in `"Hello \(name)"`.
+    ///
     /// - SeeAlso: ``StringSegmentSyntax``
     case expressionSegment(ExpressionSegmentSyntax)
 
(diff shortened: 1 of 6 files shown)


Sources/SwiftSyntax/generated/SyntaxCollections.swift
```diff
@@ -1424,14 +1504,59 @@ public struct StringLiteralSegmentListSyntax: SyntaxCollection, SyntaxHashable {
...
+    ///
+    /// - Parameters:
+    ///   - node: The expression segment to wrap in this element.
     public init(_ node: ExpressionSegmentSyntax) {
       self = .expressionSegment(node)
     }
 
+    /// Creates a new Element from an existing syntax node by attempting to cast it to either a string or expression segment.
+    ///
+    /// Example from string literal parsing:
```
ahoppen, line 1547: Similar to my comments above, I would prefer to keep this a standalone comment without referencing where it’s from.

Sources/SwiftSyntax/generated/SyntaxCollections.swift
```diff
@@ -1493,6 +1618,23 @@ public struct StringLiteralSegmentListSyntax: SyntaxCollection, SyntaxHashable {
...
+  ///   openingQuote: .stringQuoteToken(),
+  ///   segments: StringLiteralSegmentListSyntax([
+  ///     .stringSegment(.init(content: .stringSegment("Hello")))
+  ///   ]),
+  ///   closingQuote: .stringQuoteToken()
+  /// )
+  /// ```
+  ///
+  /// - Parameters:
+  ///   - node: The syntax node to create the list from. Must be of kind `.stringLiteralSegmentList`.
```
ahoppen, line 1637: I don’t think this initializer is being called in the example you give.

Sources/SwiftSyntax/generated/SyntaxCollections.swift
```diff
@@ -1424,14 +1504,59 @@ public struct StringLiteralSegmentListSyntax: SyntaxCollection, SyntaxHashable {
...
+    ///   let stringLiteral = node.as(StringLiteralExprSyntax.self),
+    ///   let firstSegment = stringLiteral.segments.first,
+    ///   case .stringSegment(let segment) = firstSegment
+    /// else {
+    ///   throw CustomError.message("Expected string literal")
+    /// }
+    /// ```
+    ///
+    /// - Parameters:
+    ///   - node: The syntax node to attempt to convert into an element. Must be either a StringSegmentSyntax or ExpressionSegmentSyntax.
```
ahoppen, line 1559: This sounds like it would crash if `node` is neither `StringSegmentSyntax` nor `ExpressionSegmentSyntax`. I would instead say something like `Returns `nil` if `node` is neither `StringSegmentSyntax` nor `ExpressionSegmentSyntax`.

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxNodesQRS.swift
```diff
@@ -2277,31 +2277,67 @@ public struct StringLiteralExprSyntax: ExprSyntaxProtocol, SyntaxHashable, _Leaf
 
 // MARK: - StringSegmentSyntax
 
-/// A literal segment inside a string segment.
-/// 
-/// - SeeAlso: ``ExpressionSegmentSyntax``
+/// A string literal segment that represents plain text content within a string literal expression.
+///
+/// Used to construct string literals programmatically, especially in macros that need to generate or manipulate string content. For example, in a string literal `"Hello \(name)"`, the `"Hello "` part would be represented by a `StringSegmentSyntax`.
```
ahoppen, line 2282: I wouldn’t say `Used to construct` in the type’s definition because you aren’t necessarily creating syntax trees, you might also just be walking existing trees.

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxNodesQRS.swift
```diff
@@ -2277,31 +2277,67 @@ public struct StringLiteralExprSyntax: ExprSyntaxProtocol, SyntaxHashable, _Leaf
...
+/// Creating a string segment requires special attention to use `.stringSegment()` for the content token to ensure proper formatting. For example, when creating a simple string segment:
+/// ```swift
+/// let segment = StringSegmentSyntax(content: .stringSegment("Hello World"))
+/// ```
+///
+/// This type is commonly used in macros for:
+/// - Creating and validating string literals at compile time (e.g., URL validation)
+/// - Building string-based error messages and diagnostics
+/// - Generating code that contains string literals
+/// - Constructing string interpolations
```
ahoppen, line 2293: Similar to other comment, I’m not sure this list provides a lot of value.

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxNodesQRS.swift
```diff
@@ -2277,31 +2277,67 @@ public struct StringLiteralExprSyntax: ExprSyntaxProtocol, SyntaxHashable, _Leaf
...
+///
+/// Important: When creating a string segment from a string literal, always use `.stringSegment(string)` rather than just passing the string directly. Using the raw string will create an identifier token instead of a string segment token, which can lead to formatting issues.
```
ahoppen, line 2295:
```suggestion
/// - Important: When creating a string segment from a string literal, always use `.stringSegment(string)` rather than just passing the string directly. Using the raw string will create an identifier token instead of a string segment token, which can lead to formatting issues.
```

Renders nicer in docc

ahoppen (commented):
Thanks for adding the documentation @Jeehut 🙏🏽 You added the documentation to generated files. Could you add the documentation in the corresponding places in [CodeGeneration](https://github.com/swiftlang/swift-syntax/tree/main/CodeGeneration) and run code generation to make sure the committed files match what would be generated?

Jeehut:
@ahoppen I've really done my best to accommodate all your comments and address them all. But please note that, like I've mentioned in my initial comment on this PR, I'm having a hard time finding the right places in the CodeGeneration package to place the documentation comments. I have now done my best to find the right places and add the documentation code to the type, which I could find some of, but it was a very time-consuming task.

But there were a few reuses of some comments in multiples places (such as `StringSegmentSyntax` once for the type and once for the enum case), although they should probably be more detailed in the type and shorter for the case.

But most importantly, pretty much ALL of my provided documentation comments for any initializers I couldn't find a place to add them in the CodeGeneration package. It doesn't seem to be trivial to add documentation comments, which is a bummer. If you checkout my branch and run `swift run generate-swift-syntax` from the CodeGeneration folder, you'll see in the git diff that all of the initializers that are missing their documentation (and few other things).

If you can guide me how to add documentation comments for those, e.g. by giving an example for one of those, I might be able to get more comments into the CodeGeneration package. For now, this is all I could do.

ahoppen:
Thanks for your effort @Jeehut.  

The initializers are generated here. 

https://github.com/swiftlang/swift-syntax/blob/eae92fed84d85325372ce41167cde3299c2ec90a/CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxNodesFile.swift#L46-L62



I didn’t check all the comments you wrote on initializers but I think we should be able to re-use some of the existing infrastructure to generate them:
- Comments for the parameters (eg. https://github.com/swiftlang/swift-syntax/pull/2922/files#diff-c5a0310c344fd64bfa2e1a25205f2faa3ee2a6f2a77686ada19dd4bfa8e40842R671): If you add those comments to the the corresponding `Child` types in SyntaxSupport, they should automatically show up
- Documentation for `unexpected*`. I wonder if we could automatically generate them based on the information that we have.
- The documentation for the conversion initializer should be possible to add to all `init?(_ node: __shared some SyntaxProtocol)` initializers
- The remaining hand-written initializer comments might make sense to put on the type themselves. Alternatively, we could add a field to `Node` and then pass that through `InitSignature` to add it the generated memberwise initializer. But if there’s a way to generate equivalent information using the fields we already have, I would prefer that because it’s less maintenance burden.

PR #2927, opened 2024-12-23
Fix parsing invalid function parameters

### LINKED ISSUE #2254: swift syntax should be able to recover function parameters without argument labels
### Description

SwiftParser currently can’t handle DocC-style abridged signatures such as:

```
init(String.Index, x:(Int) -> Int) rethrows
```

as they are considered malformed parse trees. 

more details at https://forums.swift.org/t/swiftsyntax-how-to-repair-abbreviated-function-signatures/67602

kimdv:
Fixes #2254

### DIFF
--- a/Sources/SwiftParser/Parameters.swift
+++ b/Sources/SwiftParser/Parameters.swift
@@ -96,10 +96,32 @@ extension Parser {
     let modifiers = parseParameterModifiers(isClosure: false)
     let misplacedSpecifiers = parseMisplacedSpecifiers()
 
-    var names = self.parseParameterNames()
-    let (unexpectedBeforeColon, colon) = self.expect(.colon)
+    var names: ParameterNames
 
+    let unexpectedBeforeColon: RawUnexpectedNodesSyntax?
+    let colon: RawTokenSyntax
     let type: RawTypeSyntax
+
+    // try to parse the type regardless of the presence of the preceding colon
+    // to tackle any unnamed parameter or missing colon
+    // e.g. [X], (:[X]) or (x [X])
+    let canParseType = withLookahead {
+      $0.currentToken.tokenText.isStartingWithUppercase && $0.canParseType() && $0.at(.comma, .rightParen)
+    }
+
+    if canParseType {
+      names = ParameterNames(
+        unexpectedBeforeFirstName: nil,
+        firstName: nil,
+        unexpectedBeforeSecondName: nil,
+        secondName: nil
+      )
+      unexpectedBeforeColon = nil
+      colon = missingToken(.colon)
+    } else {
+      names = self.parseParameterNames()
+      (unexpectedBeforeColon, colon) = self.expect(.colon)
+    }
 
     if colon.presence == .missing,
       let secondName = names.secondName,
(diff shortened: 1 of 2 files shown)


Sources/SwiftParser/Parameters.swift
```diff
@@ -96,11 +96,31 @@ extension Parser {
...
+    var names: ParameterNames
 
+    let unexpectedBeforeColon: RawUnexpectedNodesSyntax?
+    let colon: RawTokenSyntax
     let type: RawTypeSyntax
 
+    // try to parse the type regardless of the presence of the preceding colon
+    // to tackle any unnamed parameter or missing colon
+    // e.g. [X], (:[X]) or (x [X])
+    let canParseType = withLookahead { $0.canParseType() && $0.at(.comma) }
```
rintaro, line 108: Could you handle `, <type>)` as well? 
```suggestion
    let canParseType = withLookahead { $0.canParseType() && $0.at(.comma, .rightParen) }
```

kimdv: Also added a test case for that. Thanks

kimdv:
@swift-ci please test

ahoppen (approved):
Thanks. Looks good to me. I think just a few tests need to be updated.

kimdv:
> Thanks. Looks good to me. I think just a few tests need to be updated.

Not sure what is right. 
We have the test case `testInitAccessorsWithDefaultValues`. 

It will change `init(initialValue: <#type#>) {}` -> `init(<#identifier#>: initialValue) {}`. 

I think this is more wrong. 

Would it make sense to assume that if it starts with a lower case, it's an identifier and if it starts with an uppercase it's a type?

ahoppen:
> Would it make sense to assume that if it starts with a lower case, it's an identifier and if it starts with an uppercase it's a type?

I think that’s very reasonable behavior for error recovery. I have wanted to do this in some case in the past but I think that change never made to to a PR.

kimdv:
> > Would it make sense to assume that if it starts with a lower case, it's an identifier and if it starts with an uppercase it's a type?
> 
> I think that’s very reasonable behavior for error recovery. I have wanted to do this in some case in the past but I think that change never made to to a PR.

I was digging a bit into it. 
Where do you think is a propper place, would it be `canParseType()`?

ahoppen:
I would try shoving the uppercase check into the line that currently has.

```swift
 let canParseType = withLookahead { $0.canParseType() && $0.at(.comma, .rightParen) }
```

kimdv:
@swift-ci please test

ahoppen (approved):
Thanks, Kim. This looks good to me.

ahoppen:
The macOS formatting failure is unrelated. We’re currently working on a fix, should be fixed tomorrow-ish, could you re-trigger CI then?

kimdv:
> The macOS formatting failure is unrelated. We’re currently working on a fix, should be fixed tomorrow-ish, could you re-trigger CI then?

Yes no problem. Will rebase and re-trigger.

ahoppen:
For reference, this is the PR that fixes the format issue: https://github.com/swiftlang/swift/pull/79380

kimdv:
@swift-ci please test

PR #2928, opened 2024-12-27
[DNM] Support box-drawing-based blocks

beccadax:
There’s a social media post circulating where `#define` is used to write blocks using box-drawing characters in C code. Swift macros aren’t suitable for that kind of thing, but I wanted to see how difficult it would be to implement, just for funsies.

This is—and I cannot believe I am about to write this sentence—perhaps not the way we would actually want to implement box-drawing-based blocks, as it works by treating most of the characters as trivia and only assigning a semantic meaning to a few corners. That means it doesn’t actually enforce the validity of the box shapes.

Nor is this—and I cannot believe I am about to write *this* sentence, either—a production-quality implementation of box-drawing-based blocks. There’s a lot I haven’t tested, the diagnostics are probably a mess, and there are a number of places where I suspect I made changes that would cost us some speed.

However, it’s certainly a fun little hack:

```swift
    assertParse(
      #"""
      ╔═════════════ func fizzBuzz() ═════════════╗
      ║ ╔═══════════ for i in 0..<100 ══════════╗ ║
      ║ ║ ╔═════ if i.isMultiple(of: 15) ═════╗ ║ ║
      ║ ║ ║        print("fizzbuzz ")         ║ ║ ║
      ║ ║ ╠═══ else if i.isMultiple(of: 3) ═══╣ ║ ║
      ║ ║ ║          print("fizz ")           ║ ║ ║
      ║ ║ ╠═══ else if i.isMultiple(of: 5) ═══╣ ║ ║
      ║ ║ ║          print("buzz ")           ║ ║ ║
      ║ ║ ╠══════════════ else ═══════════════╣ ║ ║
      ║ ║ ║          print("\(i) ")           ║ ║ ║
      ║ ║ ╚═══════════════════════════════════╝ ║ ║
      ║ ╚═══════════════════════════════════════╝ ║
      ╚═══════════════════════════════════════════╝
      """#
    )
```


hamishknight:
SE-NNNN: Allow control flow using arrows 😄

PR #2926, opened 2024-12-21
[NFC] Rename 'SyntaxArena' to 'RawSyntaxArena'

rintaro:
The current `SyntaxArena` only manages `RawSyntax` green tree things. In swift-syntax, "Syntax" refers to `Syntax` the red tree. So the naming was a bit confusing.

As for `arena: ` argument labels, rename it to `rawAllocationArena: ` expect in the `RawSyntax` world (i.e. `SwiftSyntax/raw/*` and `SwiftParser/*`). 

Also rename `SyntaxArenaAllocated(Buffer)Pointer` to `ArenaAllocated*` in preparation for future introduction of other "arenas". (i.e. #2925)

PR #2930, opened 2025-01-02
Make `SyntaxArena` SPI

rintaro:
(Preparation for #2926)

Mark `SyntaxArena` and `ParsingSyntaxArena` with `@_spi(RawSyntax)`.

`SyntaxArena` and its subclass `ParsingSyntaxArena` were only meant be used when dealing with `RawSyntax` which is already a SPI. Although the type itself was `public`, most initializers were SPI, also there was no way to retrieve it from existing types via public API. So we assume no one was actually using it directly.


rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

ahoppen (approved):
Thanks

PR #2931, opened 2025-01-03
Remove default values for leftParen and rightParen parameters in the FunctionCallExprSyntax initializer

### LINKED ISSUE #2912: `FunctionCallExprSyntax` result builder-based initializer does not insert parentheses tokens
### Description

The convenience initializer for `FunctionCallExprSyntax` that takes a result builder doesn't insert the parentheses tokens by default.

### Steps to Reproduce

```swift
let functionCallExpr = FunctionCallExprSyntax(calledExpression: DeclReferenceExprSyntax(baseName: .identifier("f"))) {
  LabeledExprSyntax(expression: NilLiteralExprSyntax())
}
assert(functionCallExpr.leftParen != nil)
assert(functionCallExpr.rightParen != nil)
```

TTOzzi:
Resolve #2912 

Initially, I attempted to fix the issue by modifying only the result builder-based initializer of `FunctionCallExprSyntax` where the issue occurred. 
However, since the initializer is automatically generated by `CodeGeneration`, it seemed impossible to customize it for `FunctionCallExprSyntax` alone.

When the parentheses argument is not provided, setting the default parentheses based on the `isEmpty` value of the provided `arguments` appears to be part of the common logic for `FunctionCallExprSyntax`, so I modified the logic in the designated initializer accordingly.


Tests/SwiftSyntaxTest/SyntaxCreationTests.swift
```diff
@@ -176,13 +176,6 @@ class SyntaxCreationTests: XCTestCase {
     let call2 = call1.with(\.leftParen, nil).with(\.rightParen, nil)
     XCTAssertNil(call2.leftParen)
     XCTAssertNil(call2.rightParen)
-
-    let call3 = FunctionCallExprSyntax(
-      calledExpression: printID,
-      arguments: LabeledExprListSyntax([arg])
-    )
-    XCTAssertNil(call3.leftParen)
-    XCTAssertNil(call3.rightParen)
```
TTOzzi, line 185: It seems like a test to verify that the default value `nil` is assigned when the initializer argument is omitted. 
However, due to changes in the internal logic of the initializer, it was failing, and since the test no longer seems meaningful, I have removed it 🤔

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxNodesEF.swift
```diff
@@ -3162,6 +3162,9 @@ public struct FunctionCallExprSyntax: ExprSyntaxProtocol, SyntaxHashable, _LeafE
     _ unexpectedAfterAdditionalTrailingClosures: UnexpectedNodesSyntax? = nil,
     trailingTrivia: Trivia? = nil
   ) {
+    let shouldOmitParens = arguments.isEmpty && trailingClosure != nil
+    let leftParen = leftParen ?? (shouldOmitParens ? nil : .leftParenToken())
+    let rightParen = rightParen ?? (shouldOmitParens ? nil : .rightParenToken())
```
ahoppen, line 3167: Since this code is generated, you won’t be able to modify it. Also, these memberwise initializers should not contain smart logic (eg. consider that maybe you want to omit the parentheses to build an invalid syntax tree for some test case). So, I think any logic that tries to be smart should live only in the convenience initializers.

ahoppen (commented):
Thanks for approaching this @TTOzzi but I don’t think this is the correct fix.

Sources/SwiftSyntax/generated/syntaxNodes/SyntaxNodesEF.swift
TTOzzi: Ah, I see. Thank you for your feedback 😄 
It seems that I can't make a simple modification while maintaining the current structure.
How about declaring a `Bool` property (for example, `shouldCreateBuilderInitializer`) in `CodeGeneration`'s `Node` to indicate whether the result builder-based initializer should be automatically generated, so that it wouldn't be generated for `functionCallExpr`, and instead, manually creating the result builder-based initializer for `functionCallExpr` in [ConvenienceInitializers.swift](https://github.com/swiftlang/swift-syntax/blob/main/Sources/SwiftSyntaxBuilder/ConvenienceInitializers.swift#L194)?
Would it be acceptable to introduce a conditional value just for a specific node?

ahoppen: As I said above, I think we should have a way of creating a function call with arguments but without parentheses (eg. to test a syntax tree for invalid Swift code) and this memberwise initializer is the correct one to do. If anything, you could argue that we shouldn’t default `leftParen` and `rightParen` to be `nil`, but I’m not sure that’s all that helpful.

TTOzzi: So you're saying that the behavior where the result builder-based initializer doesn't insert parentheses tokens by default (in #2912) is intentional.
I understand. I will go ahead and close the PR.
Thank you for the kind explanation 🙇

ahoppen: I’m saying that it’s intentional that the non-result builder initializer should have a way to have a `nil` parenthesis but arguments. I’m not as fixed on the result-builder initializers since they already have some special logic, like adding  commas between elements in the result builder.

TTOzzi: I implemented it by removing the automatically generated code in order to maintain the existing result builder initializer interface. However, if you're reluctant to add an option to `CodeGeneration`'s `Node`, another approach could be to add a convenience initializer that defines the argument label slightly differently 🤔 
It seems that CatchClauseSyntax(https://github.com/swiftlang/swift-syntax/blob/main/Sources/SwiftSyntaxBuilder/ConvenienceInitializers.swift#L93-L107) is already implemented this way.

ahoppen:
> Would removing `= nil` to enforce passing all arguments be a better solution than adding a result builder-based convenience initializer that generates a default `paren` even when it is not explicitly provided?

I think these are orthogonal ways to approach the problem and both might be worthwhile. As I mentioned before, the result builder initializer might perform some clever logic to infer whether the parentheses are needed. Removing `= nil` from the memberwise initializer reduces the chances of surprising behavior when a user uses that initializer because they need to explicitly specify if they need parentheses.

TTOzzi:
> I think these are orthogonal ways to approach the problem and both might be worthwhile. As I mentioned before, the result builder initializer might perform some clever logic to infer whether the parentheses are needed. Removing `= nil` from the memberwise initializer reduces the chances of surprising behavior when a user uses that initializer because they need to explicitly specify if they need parentheses.

Aha, I see! By forcing argument passing, the initializer's behavior will always be within the user's expectations, which will result in a meaningful implementation!
I'll try to reimplement it in the direction you mentioned. Thanks for sharing your opinions.

TTOzzi:
During implementation, I encountered a few issues:

1. I cannot declare the existing builder-based initializer for `FunctionCallExprSyntax` in `SwiftSyntaxCompatibility.swift`. The `@LabeledExprListBuilder` argument is part of `SwiftSyntaxBuilder` code, and importing `SwiftSyntaxBuilder` into `SwiftSyntax` causes a circular reference between modules.
2. The existing initializer, which needs to maintain compatibility, has the same argument labels as the automatically generated initializer from `CodeGeneration`, leading to a redeclaration error.

To address these, I wrote compatibility-maintaining code for `FunctionCallExprSyntax` in `SwiftSyntaxBuilderCompatibility.swift`.
And the redeclaration error was resolved by changing the label names of the builder arguments. Since the builder initializer uses trailing closures without argument labels, I expect the compatibility layer will work as intended.

However, the issue is that while I managed to maintain compatibility with the builder init, I can’t create another initializer with the same argument labels and different default values for other member initializers, so it seems impossible to maintain compatibility there.
Would it be better to apply the removal of default values only to the builder-based convenience initializer that can maintain compatibility?
(The implementation so far has been committed.)

I’m concerned that the changes have grown larger than the issue I was trying to resolve 🤯

Sources/SwiftSyntax/SwiftSyntaxCompatibility.swift
```diff
@@ -350,6 +350,246 @@ extension GenericArgumentSyntax {
...
+    )
+  }
+
+  @available(
+    *,
+    deprecated,
+    renamed:
+      "init(leadingTrivia:_:calledExpression:_:leftParen:_:arguments:_:rightParen:_:trailingClosure:_:additionalTrailingClosures:_:trailingTrivia:)"
+  )
+  public init(
```
ahoppen, line 558: Could you mark these compatibility layers as `@_disfavoredOverload` because the original method was also marked as `@_disfavoredOverload`?

Sources/SwiftSyntax/SwiftSyntaxCompatibility.swift
```diff
@@ -350,6 +350,246 @@ extension GenericArgumentSyntax {
   }
 }
 
+extension FunctionCallExprSyntax {
+  @available(
+    *,
+    deprecated,
+    message: "Use the initializer that does not provide default values for leftParen and rightParen."
+  )
+  public init(
```
ahoppen, line 359: Do we need all variations of `leftParen` and `rightParen` missing? Couldn’t we have one compatibility function that still has the `= nil` default value?

Sources/SwiftSyntaxBuilder/generated/RenamedChildrenBuilderCompatibility.swift
```diff
@@ -340,10 +340,10 @@ extension FunctionCallExprSyntax {
     unexpectedBeforeCalledExpression: UnexpectedNodesSyntax? = nil,
```
ahoppen, line 340: Could you also add a compatibility layer for this initializer?

ahoppen (commented):
Thank you. I left a few ideas inline.

Sources/SwiftSyntaxBuilder/generated/RenamedChildrenBuilderCompatibility.swift
TTOzzi: If the builder-based initializer is used with a trailing closure like this:

```swift
FunctionCallExprSyntax(calledExpression: DeclReferenceExprSyntax(baseName: .identifier("f"))) {
  LabeledExprSyntax(expression: NilLiteralExprSyntax())
}
```
then the argument label of `@LabeledExprListBuilder` (whether it's `argumentListBuilder` or `argumentList`) doesn't matter, and it can be used the same way as before without requiring a compatibility layer.

However, I realized that in cases where the argument label is explicitly specified, like this:

```swift
FunctionCallExprSyntax(calledExpression: DeclReferenceExprSyntax(baseName: .identifier("f")), argumentListBuilder: {
  LabeledExprSyntax(expression: NilLiteralExprSyntax())
})
```
it could cause issues. I'll add compatibility handling for such cases.

Sources/SwiftSyntax/SwiftSyntaxCompatibility.swift
TTOzzi: I initially intended to implement it as you suggested.
However, when all parameters in an initializer share the same labels and only a default value(`= nil`) is added, the compiler considers them identical and raises a redeclaration error 🤔 
To resolve this, I had to define separate initializers for each case.

ahoppen: Oh, that’s unfortunate. But fair enough. Thanks for spelling them out.

ahoppen (approved):
LGTM. Thanks.

ahoppen:
@swift-ci Please test

Sources/SwiftSyntax/SwiftSyntaxCompatibility.swift
TTOzzi: Thanks for taking the time to review 😊

ahoppen:
@swift-ci Please test Windows

PR #2935, opened 2025-01-10
[DO NOT MERGE] Revert "Workaround for rdar://141977987"

rintaro:
Just testing for now after https://github.com/swiftlang/swift/pull/78482


rintaro:
@swift-ci Please smoke test

rintaro:
@swift-ci Please test

rintaro:
Nice we don't see the crash anymore. But I will keep it as-is for now just in case someone is using the host compiler without the fix. Also the workaround shouldn't affect the performance :)

PR #2936, opened 2025-01-10
Compute the lexical context of an accessor macro based on the pre-rewritten node

ahoppen:
When an accessor macro contains a closure, it gets rewritten to a new node in `MacroSystem` by calling `super.visit`. That rewritten node doesn’t have any parents (because `SyntaxRewriter` is expecting that node to get inserted into the tree when the node’s parent gets rewritten). This caused us to generate an empty lexical context in these situations.

To fix this, generate the lexical context based on the node before we rewrite it.

rdar://142639902

### DIFF
--- a/Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
+++ b/Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
@@ -930,23 +930,30 @@ private class MacroApplication<Context: MacroExpansionContext>: SyntaxRewriter {
   }
 
   override func visit(_ node: VariableDeclSyntax) -> DeclSyntax {
-    var node = super.visit(node).cast(VariableDeclSyntax.self)
+    var rewrittenNode = super.visit(node).cast(VariableDeclSyntax.self)
 
-    guard !macroAttributes(attachedTo: DeclSyntax(node), ofType: AccessorMacro.Type.self).isEmpty else {
-      return DeclSyntax(node)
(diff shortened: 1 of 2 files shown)


Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
```diff
@@ -930,23 +930,30 @@ private class MacroApplication<Context: MacroExpansionContext>: SyntaxRewriter {
...
+      return DeclSyntax(rewrittenNode)
     }
 
-    guard node.bindings.count == 1,
-      var binding = node.bindings.first
+    guard rewrittenNode.bindings.count == 1,
+      var binding = rewrittenNode.bindings.first
     else {
-      contextGenerator(Syntax(node)).addDiagnostics(
+      contextGenerator(Syntax(rewrittenNode)).addDiagnostics(
```
rintaro, line 942: I think this should also be `contextGenerator(Syntax(node))`, no?

ahoppen: Yes, it should. Thanks for catching it. Not sure how I missed it.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

rintaro (approved):

ahoppen:
@swift-ci Please test Windows

PR #2938, opened 2025-01-10
[Syntax] Let SyntaxIdentifier use RawSyntax.ID directly

rintaro:
There was no reason to convert it to UInt.


rintaro:
@swift-ci Please test

ahoppen (approved):

PR #2943, opened 2025-01-18
[Syntax] Fix TriviaPiece.isNewline

rintaro:
In Swift, only `CR`, `LF`, and `CR`+`LF` are considered newline characters. `TriviaPiece.isNewline` should not return true for `VT` and `FF`.

Instead of using `Swift.Character.isNewline`, introduce `TriviaTraits` in CodeGeneration, and manually set them for each trivia piece kind.

Also some drive-by cleanups:
* Eliminate `Trivia.swiftCharacters` because it was unused.
* Use `String?` for `Token.characters` because string is naturally a collection of characters.
* Eliminate `Trivia.CharactersLen` because `Trivia.characters.utf8.count` is straightforward enough, and it was technically wrong because it returned character count, but not UTF8 byte count.

### DIFF
--- a/CodeGeneration/Sources/SyntaxSupport/Trivia.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/Trivia.swift
@@ -44,11 +44,6 @@ public class Trivia {
   /// Useful for multi-character trivias like `\r\n`.
   public let characters: [Character]
 
-  /// The list of characters as they would appear in Swift code.
-  ///
-  /// This might differ from `characters` due to Swift's character escape requirements.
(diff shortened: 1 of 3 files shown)


rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

ahoppen (approved):

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

rintaro:
@swift-ci Please test Windows

PR #2945, opened 2025-01-20
Don’t remove `#if` attributes with `AttributeRemover`

### LINKED ISSUE #2923: `SyntaxProtocol.expand(macros:)` drops `IfConfigDeclSyntax` nodes from result
### Description

Given the following code:

```swift
let sourceCode = """
#if true
@inlinable
#endif
func f() {}
"""

let originalSyntax = Parser.parse(source: sourceCode)
print(originalSyntax)
let context = BasicMacroExpansionContext(lexicalContext: [], expansionDiscriminator: "", sourceFiles: [:])
let syntax = originalSyntax.expand(macros: [:]) { syntax in
  BasicMacroExpansionContext(sharingWith: context, lexicalContext: syntax.allMacroLexicalContexts())
}
print(syntax)
```

I would expect to see (ignoring potential reformatting):

```swift
#if true
@inlinable
#endif
func f() {}
```

But `syntax` drops everything in the `IfConfigDeclSyntax` node (`#if` onward) and just produces:

```swift
func f() {}
```

`originalSyntax` is correct, so the issue is during macro expansion, not the initial parse.

### Steps to Reproduce

See above.

ahoppen:
Fixes #2923
rdar://141840779

### DIFF
--- a/Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
+++ b/Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
@@ -530,7 +530,11 @@ public class AttributeRemover: SyntaxRewriter {
 
   public override func visit(_ node: AttributeListSyntax) -> AttributeListSyntax {
     var filteredAttributes: [AttributeListSyntax.Element] = []
-    for case .attribute(let attribute) in node {
+    for attribute in node {
+      guard case .attribute(let attribute) = attribute else {
+        filteredAttributes.append(attribute)
+        continue
+      }
       if self.predicate(attribute) {
         var leadingTrivia = attribute.leadingTrivia
 
--- a/Tests/SwiftSyntaxMacroExpansionTest/AttributeRemoverTests.swift
+++ b/Tests/SwiftSyntaxMacroExpansionTest/AttributeRemoverTests.swift
@@ -470,4 +470,21 @@ final class AttributeRemoverTests: XCTestCase {
         """
     )
   }
+
+  func testKeepPoundIfInAttributes() {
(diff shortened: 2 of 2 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
```diff
@@ -530,7 +530,11 @@ public class AttributeRemover: SyntaxRewriter {
 
   public override func visit(_ node: AttributeListSyntax) -> AttributeListSyntax {
     var filteredAttributes: [AttributeListSyntax.Element] = []
-    for case .attribute(let attribute) in node {
+    for attribute in node {
+      guard case .attribute(let attribute) = attribute else {
+        filteredAttributes.append(attribute)
```
rintaro, line 535: What if `#if ... #endif` contained attributes matching `self.predicate(attribute)`? Or it never happens because MacroSystem doesn't support `#if` attributes?

ahoppen: My thought is that `MacroSystem` doesn’t know how to evaluate `#if` conditions, so just keeping them verbatim is the best we can do…

rintaro: I see `MacroApplication.macroAttributes(attachedTo:)` doesn't look into `#if` at all at this point. So yeah, this is fine.

rintaro (approved):

PR #2946, opened 2025-01-21
[Build] Remove reference to implicit backtracing import.

al45tair:
This wasn't necessary anyway, since it's off by default and we never changed that because of the plan to move the backtracing code elsewhere. I don't think this is in fact a blocker because it looks like we test for it first, but we should remove it anyway.

rdar://143310345


al45tair:
@swift-ci Please test

al45tair:
@swift-ci Please test

ahoppen (approved):
Sounds reasonable to me.

PR #2947, opened 2025-01-21
[6.1] Don’t remove `#if` attributes with `AttributeRemover`

ahoppen:
- **Explanation**: When expanding macros using MacroSystem, we would always remove attributes inside `#if`. Change the behavior to never remove them since we can’t evaluate `#if`.
- **Scope**: Expansion of macros in `assertMacroExpansion`
- **Issue**: #2923 / rdar://141840779
- **Original PR**:https://github.com/swiftlang/swift-syntax/pull/2945
- **Risk**: Does not affect compile-time expansion of macros
- **Testing**: Added test case 
- **Reviewer**:  @bnbarham


ahoppen:
@swift-ci Please test

kimdv (approved):

bnbarham (approved):

PR #2950, opened 2025-01-22
Add keypath method and initializer syntax under an experimental feature flag `keypathWithMethodMembers`

amritpan:
This accompanies the [swift-foundation](https://github.com/swiftlang/swift-foundation/pull/1179) and [compiler implementation](https://github.com/swiftlang/swift/pull/78823) for [Method and Initializer Keypaths](https://github.com/swiftlang/swift-evolution/pull/2675) which extends keypath usage to include references to methods and initializers:

```
struct S {
  static let millenium = 3
  var year = 2024
  init() {}
  init(val value: Int = 2024) { year = value }
  
  func add(this: Int) -> Int { this + this}
  func add(that: Int) -> Int { that + that }
  static func subtract(_ val: Int) -> Int { return millenium - val }
  nonisolated func nonisolatedNextYear() -> Int { return year + 1 }
  consuming func consume() { print(year) }

  subscript(index: Int) -> Int { return year + index }
}

let kp1: KeyPath<S, () -> S> = \S.Type.init
let kp2: KeyPath<S, S> = \S.Type.init()
let kp3: KeyPath<S, (Int) -> S> = \S.Type.init(val:)
let kp4: KeyPath<S, S> = \S.Type.init(val: 2025)
let kp5: KeyPath<S, (Int) -> Int> = \S.add(this:)
let kp6: KeyPath<S, Int> = \S.add(that: 1)
let kp7: KeyPath<S, Int> = \S.Type.subtract(1)
let kp8: KeyPath<S, Int> = \S.nonisolatedNextYear()
let kp9: KeyPath<S, Int> = \S.nonisolatedNextYear().signum()
let kp10: KeyPath<S, String> = \S.nonisolatedNextYear().description
let kp11: KeyPath<S, ()> = \S.consume()
let kp12: KeyPath<S, Int> = \S.Type.init()[1]
```

### DIFF
--- a/CodeGeneration/Sources/SyntaxSupport/ExprNodes.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/ExprNodes.swift
@@ -1209,6 +1209,11 @@ public let EXPR_NODES: [Node] = [
             kind: .node(kind: .keyPathPropertyComponent)
           ),
           Child(
+            name: "method",
+            kind: .node(kind: .keyPathMethodComponent),
+            experimentalFeature: .keypathWithMethodMembers
+          ),
+          Child(
             name: "subscript",
             kind: .node(kind: .keyPathSubscriptComponent)
           ),
@@ -1309,6 +1314,33 @@ public let EXPR_NODES: [Node] = [
         "arguments": .renamed(from: "argumentList"),
         "rightSquare": .renamed(from: "rightBracket"),
       ]
+    ]
+  ),
+
+  Node(
+    kind: .keyPathMethodComponent,
+    base: .syntax,
+    experimentalFeature: .keypathWithMethodMembers,
+    nameForDiagnostics: "key path method component",
+    documentation: "A key path component like `.method()`, `.method(10)`, or `.method(val: 10)`.",
+    children: [
+      Child(
+        name: "declName",
+        kind: .node(kind: .declReferenceExpr)
+      ),
+      Child(
+        name: "leftParen",
+        kind: .token(choices: [.token(.leftParen)])
+      ),
+      Child(
+        name: "arguments",
+        kind: .collection(kind: .labeledExprList, collectionElementName: "Argument"),
+        nameForDiagnostics: "arguments"
+      ),
+      Child(
+        name: "rightParen",
+        kind: .token(choices: [.token(.rightParen)])
+      ),
     ]
   ),
 
--- a/Sources/SwiftParser/Expressions.swift
+++ b/Sources/SwiftParser/Expressions.swift
@@ -11,9 +11,9 @@
 //===----------------------------------------------------------------------===//
 
 #if compiler(>=6)
-@_spi(RawSyntax) internal import SwiftSyntax
+@_spi(RawSyntax) @_spi(ExperimentalLanguageFeatures) internal import SwiftSyntax
 #else
-@_spi(RawSyntax) import SwiftSyntax
+@_spi(RawSyntax) @_spi(ExperimentalLanguageFeatures) import SwiftSyntax
 #endif
 
 extension TokenConsumer {
@@ -1115,11 +1115,54 @@ extension Parser {
         continue
       }
 
-      // Check for a .name or .1 suffix.
+      // Check for a .name, .1, .name(), .name("Kiwi"), .name(fruit:),
+      // .name(_:), .name(fruit: "Kiwi) suffix.
       if self.at(.period) {
         let (unexpectedPeriod, period, declName, generics) = parseDottedExpressionSuffix(
           previousNode: components.last?.raw ?? rootType?.raw ?? backslash.raw
         )
+
+        // If fully applied method component, parse as a keypath method.
+        if self.experimentalFeatures.contains(.keypathWithMethodMembers)
+          && self.at(.leftParen)
+        {
+          var (unexpectedBeforeLParen, leftParen) = self.expect(.leftParen)
+          if let generics = generics {
+            unexpectedBeforeLParen = RawUnexpectedNodesSyntax(
+              (unexpectedBeforeLParen?.elements ?? []) + [RawSyntax(generics)],
+              arena: self.arena
+            )
+          }
+          let args = self.parseArgumentListElements(
+            pattern: pattern,
+            allowTrailingComma: true
+          )
+          let (unexpectedBeforeRParen, rightParen) = self.expect(.rightParen)
+
+          components.append(
+            RawKeyPathComponentSyntax(
+              unexpectedPeriod,
+              period: period,
+              component: .method(
+                RawKeyPathMethodComponentSyntax(
+                  declName: declName,
+                  unexpectedBeforeLParen,
+                  leftParen: leftParen,
+                  arguments: RawLabeledExprListSyntax(
+                    elements: args,
+                    arena: self.arena
+                  ),
+                  unexpectedBeforeRParen,
+                  rightParen: rightParen,
+                  arena: self.arena
+                )
+              ),
+              arena: self.arena
+            )
+          )
+          continue
+        }
+        // Else, parse as a property.
         components.append(
           RawKeyPathComponentSyntax(
             unexpectedPeriod,
@@ -1140,7 +1183,6 @@ extension Parser {
       // No more postfix expressions.
       break
     }
-
     return RawKeyPathExprSyntax(
       unexpectedBeforeBackslash,
       backslash: backslash,
--- a/Tests/SwiftParserTest/ExpressionTests.swift
+++ b/Tests/SwiftParserTest/ExpressionTests.swift
@@ -11,6 +11,7 @@
 //===----------------------------------------------------------------------===//
 
 @_spi(RawSyntax) @_spi(ExperimentalLanguageFeatures) import SwiftParser
+@_spi(RawSyntax) @_spi(ExperimentalLanguageFeatures) import SwiftSyntax
 @_spi(RawSyntax) import SwiftSyntax
 import XCTest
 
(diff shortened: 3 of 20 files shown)


Tests/SwiftParserTest/ExpressionTests.swift
```diff
@@ -256,6 +256,167 @@ final class ExpressionTests: ParserTestCase {
...
+
+    assertParse(
+      #"\Foo.1️⃣()"#,
+      diagnostics: [
+        DiagnosticSpec(message: "expected identifier in key path method component", fixIts: ["insert identifier"])
+      ],
+      fixedSource: #"\Foo.<#identifier#>()"#,
+      experimentalFeatures: .keypathWithMethodMembers
+    )
+  }
```
ahoppen, line 418: Some suggestions for more test cases
- `\Foo.method<Int>()`
- `\Foo.method<Int>(arg:)`
- `\Foo.method(_:)` for unnamed arguments

ahoppen (commented):
The implementation looks good to me. I have a few small comments inline.

CodeGeneration/Sources/SyntaxSupport/ExprNodes.swift
```diff
@@ -1313,6 +1317,39 @@ public let EXPR_NODES: [Node] = [
...
+      Child(
+        name: "genericArgumentClause",
+        kind: .node(kind: .genericArgumentClause),
+        isOptional: true
+      ),
+      Child(
+        name: "leftParen",
+        kind: .token(choices: [.token(.leftParen)]),
+        isOptional: true
+      ),
```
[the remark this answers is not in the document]
amritpan: I modeled this after `.funcCallExpr` but you're right and I will make this non optional. @ahoppen I also have 2 related questions:

`.keyPathPropertyComponent` [has a `.genericArgumentClause`](https://github.com/swiftlang/swift-syntax/blob/a84f21d0d6f6b2f0774af605405fde07179a7715/CodeGeneration/Sources/SyntaxSupport/ExprNodes.swift#L1269) but I don't see any examples of it being used in the tests. Do you happen to know when that is used? It seems like a generic on the root (eg: `\Box<Int>.item`) is already being handled by `KeyPathExpr`.

Is there a way to make `.labeledExprList` optional? I forgot to account for this - partially applied methods parse their method name and arg list into `DeclReferenceExpr` and do not have a `LabeledExprList`. I'm inclined to create a second node to handle this (eg: `.keyPathPartiallyAppliedComponent`).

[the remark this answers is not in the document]
ahoppen: > `.keyPathPropertyComponent` [has a `.genericArgumentClause`](https://github.com/swiftlang/swift-syntax/blob/a84f21d0d6f6b2f0774af605405fde07179a7715/CodeGeneration/Sources/SyntaxSupport/ExprNodes.swift#L1269) but I don't see any examples of it being used in the tests. Do you happen to know when that is used? It seems like a generic on the root (eg: `\Box<Int>.item`) is already being handled by `KeyPathExpr`.

I don’t think a generic clause is actually valid in key path properties and maybe we should remove it. @rintaro Can you think of a reason why `KeyPathPropertyComponentSyntax` should have a generic arguments clause?

> Is there a way to make `.labeledExprList` optional? I forgot to account for this - partially applied methods parse their method name and arg list into `DeclReferenceExpr` and do not have a `LabeledExprList`. I'm inclined to create a second node to handle this (eg: `.keyPathPartiallyAppliedComponent`).

I’m not sure I understand. If there are no parentheses, we should be able to parse the component as a `KeyPathPropertyComponentSyntax`. If there are parentheses and no arguments, `arguments` can just be an empty list. And if there are arguments … then it doesn’t need to be optional.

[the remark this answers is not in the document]
amritpan: > I’m not sure I understand. If there are no parentheses, we should be able to parse the component as a KeyPathPropertyComponentSyntax. If there are parentheses and no arguments, arguments can just be an empty list. And if there are arguments … then it doesn’t need to be optional.

Should partial arguments should also be KeyPathPropertyComponentSyntax? `method(_:)` and `method(arg:)` for unapplied keypath methods are being parsed as `DeclReferenceExpr` with `DeclNameArgumentListSyntax` arguments instead of `LabeledExprList`. I have pushed changes that reflect this, unsure if there are any ramifications of doing it this way, but the Swift and c++ parser outputs match.

Tests/SwiftParserTest/ExpressionTests.swift
amritpan: > - \Foo.method\<Int>()
> - \Foo.method\<Int>(arg:)

I'm getting an unable to round trip message for these 2 invalid test cases. How do I resolve this?

amritpan:
Added additional tests to ensure Swift and c++ parsers match.

Sources/SwiftParser/Expressions.swift
```diff
@@ -2017,6 +2056,37 @@ extension Parser {
...
+        }
+      }
+      if token == .rightParen {
+        hasRParen = true
+      }
+      lookahead.consumeAnyToken()
+    }
+    // If parentheses exist with no partial application pattern,
+    // parse as a key path method.
+    return hasLParen && hasRParen ? true : false
```
ahoppen, line 2087: This seems problematic to me. I haven’t fully though it through but it has the potential for an unbounded lookahead, which could be very bad for parser performance and the use of `consumeAnyToken` to eat any uncovered token also seems like a pitfall to me.

What I would suggest is that in your actual `parse*` method, you call `parseDottedExpressionSuffix` outside of the `if`. That would parse any compound names, ie. unapplied function references. If we are at a parenthesis after that, continue parsing arguments for an applied key path reference. That way, you can also parse key paths like `\Foo.t(a:)(2)`, mirroring the following valid Swift code

```swift
struct Foo {
  func t(a: Int) {
    self.t(a:)(2)
  }
}
```

Or am I missing something with that approach?

amritpan: This makes sense and I've updated the code to reflect this. For reasons I can't reproduce now, `parseDottedExpressionSuffix` did not work for me initially and so I wrote this new method. Regardless it works now. All good things to know about `Lookahead` as well. Thank you!

Sources/SwiftParser/Expressions.swift
```diff
@@ -15,7 +15,9 @@
 #else
 @_spi(RawSyntax) import SwiftSyntax
 #endif
-
+#if compiler(>=5.8)
+@_spi(RawSyntax) @_spi(ExperimentalLanguageFeatures) internal import SwiftSyntax
```
ahoppen, line 19: Could you update the import above instead of adding another import to `SwiftSyntax`?

Tests/SwiftParserTest/ExpressionTests.swift
```diff
@@ -256,6 +257,335 @@ final class ExpressionTests: ParserTestCase {
...
+        ),
+      ],
+      fixedSource: #"\Foo.method<Int>(arg: <#expression#>)"#,
+      experimentalFeatures: .keypathWithMethodMembers
+    )
+
+    assertParse(
+      #"""
+      \Foo.t(a:)(2)
+      """#,
```
ahoppen, line 506: I think a substructure test would be beneficial here because it’s not immediately obvious how this should parse.

Release Notes/602.md
```diff
@@ -25,6 +25,12 @@
   - Pull Request: https://github.com/swiftlang/swift-syntax/pull/2925
   - Migration steps: Use `SyntaxChildren.endIndex` instead.
   - Notes: `ExpressibleByNilLiteral` was a mistake. In general, `Collection.Index` should only be created and managed by the collection itself. For example, `Collection.index(after:)` exists, but `Index.advanced(by:)` does not.
+  
+  - *Affected API or two word description*
+  - Description: *A new keypath component (`.keyPathMethodComponent`) has been added to handle fully applied keypath methods and initializers.*
+  - Pull Request: *https://github.com/swiftlang/swift-syntax/pull/2950*
+  - Migration steps: To switch over KeyPathComponent.Component exhaustively, please add handling for `.keyPathMethodComponent`.
+  - Notes: Keypath methods and initializers are an uncoming Swift language feature.
```
ahoppen, line 33: I don’t think we need this anymore now that `.keyPathMethodComponent` is SPI.

ahoppen (commented):
That looks great and is much simpler as well. Just two minor comments, everything else is great.

ahoppen (approved):
Looks good to me 🚀

ahoppen:
@swift-ci Please test

amritpan:
Not sure why this test is failing - is it due to my imports in `Expressions.swift?`

ahoppen:
Oh, I just didn’t test it with your PR in the compiler repo. Let’s try again.

https://github.com/swiftlang/swift/pull/78823

@swift-ci Please test

ahoppen:
Looks like you need to rebase this PR + re-generate the sources to adjust it for https://github.com/swiftlang/swift-syntax/pull/2926.

amritpan:
Rebased and regenerated, but running into this failure locally: 

```
./swift-syntax-dev-utils local-pr-precheck
Building for debugging...
[1/1] Write swift-version--58304C5D6DBC2206.txt
Build of product 'swift-syntax-dev-utils' complete! (0.14s)
** Running code generation **
** Building target SwiftSyntax-all **
Building for debugging...
[55/55] Emitting module SwiftSyntaxBuilder
Build of target: 'SwiftSyntax-all' complete! (9.33s)
** Building target Examples-all **
Building for debugging...
error: couldn't build /.../swift-project/swift-syntax/Examples/.build/arm64-apple-macosx/debug/SwiftSyntax.build/sources because of missing inputs: /.../swift-project/swift-syntax/Sources/SwiftSyntax/Raw/RawSyntaxArena.swift, /.../swift-project/swift-syntax/Sources/SwiftSyntax/ArenaAllocatedBuffer.swift
```

ahoppen:
Could you try if deleting `Examples/.build` fixes the issue you’re seeing locally?

ahoppen:
https://github.com/swiftlang/swift/pull/78823

@swift-ci Please test

ahoppen:
https://github.com/swiftlang/swift/pull/78823

@swift-ci Please test

amritpan:
Yes, deleting Examples/.build fixed the issue locally.

ahoppen:
Sorry, looks like you picked a bad time for your PR right now as you’re racing with @rintaro’s refactorings. Can you rebase + re-generate once more to pick up https://github.com/swiftlang/swift-syntax/pull/2926?

amritpan:
Yes, rebased and regenerated!

ahoppen:
https://github.com/swiftlang/swift/pull/78823

@swift-ci Please test

Sources/SwiftParser/Expressions.swift
```diff
@@ -10,11 +10,15 @@
...
-@_spi(RawSyntax) internal import SwiftSyntax
+#if swift(>=6)
+@_spi(RawSyntax) @_spi(ExperimentalLanguageFeatures) internal import SwiftSyntax
+#else
+#if swift(>=5.8)
+@_spi(RawSyntax) @_spi(ExperimentalLanguageFeatures) import SwiftSyntax
 #else
 @_spi(RawSyntax) import SwiftSyntax
 #endif
+#endif
```
ahoppen, line 21: This should just be the following, similar to `Attributes.swift`.

```
#if compiler(>=6)
@_spi(RawSyntax) @_spi(ExperimentalLanguageFeatures) internal import SwiftSyntax
#else
@_spi(RawSyntax) @_spi(ExperimentalLanguageFeatures) import SwiftSyntax
#endif
```

ahoppen (commented):
Looks like we need to cover `.method` in this swift for the `#predicate` macro as well: https://github.com/swiftlang/swift-foundation/blob/2f4f5b80cfb4cd4540e1e75bc07dff4a54a5ea47/Sources/FoundationMacros/PredicateMacro.swift#L385

amritpan:
Opened [swift-foundation/#1179](https://github.com/swiftlang/swift-foundation/pull/1179) to add .method handling to predicate macro.

amritpan:
I believe the swift-foundation change is ready to be merged but [they are unable to run CI and merge it](https://github.com/swiftlang/swift-foundation/pull/1179#discussion_r1960397401) without these changes in swift syntax main, so perhaps swift-syntax needs to merge first? I've rebased all 3 PRs (compiler/foundation/syntax) to fix merge conflicts.

amritpan:
https://github.com/swiftlang/swift-foundation/pull/1179

https://github.com/swiftlang/swift/pull/78823

amritpan:
https://github.com/swiftlang/swift-foundation/pull/1179

https://github.com/swiftlang/swift/pull/78823

@swift-ci Please test windows platform.

amritpan:
https://github.com/swiftlang/swift-foundation/pull/1179

https://github.com/swiftlang/swift/pull/78823

amritpan:
@ahoppen Unsure why the Windows platform test keeps failing.

amritpan:
Are you able to share any failure output? I only see "Build finished. No test results found." and nothing is failing for me locally.

bnbarham:
> Are you able to share any failure output? I only see "Build finished. No test results found." and nothing is failing for me locally.

The failing job was https://ci-external.swift.org/job/swift-syntax-PR-windows/4074/ with two test failures:
```
Failed Tests (2):
  Swift(windows-x86_64) :: Interpreter/keypath.swift
  Swift(windows-x86_64) :: SILGen/keypaths.swift
```

`SILGen/keypaths.swift` is because:
```
# | C:\Users\swift-ci\jenkins\workspace\swift-syntax-PR-windows\swift\test\SILGen\keypaths.swift:677:12: error: CHECK: expected string not found in input
# |  // CHECK: %0 = keypath $WritableKeyPath<K.Type, () -> K>, (objc "self"; root $K.Type; gettable_property $() -> K, id @$s8keypaths1KVACycfC : $@convention(method) (@thin K.Type) -> K, getter @$s8keypaths1KVACycfcACmTkm : $@convention(keypath_accessor_getter) (@in_guaranteed @thick K.Type) -> @out @callee_guaranteed @substituted <τ_0_0> () -> @out τ_0_0 for <K>)
```

Seems like the difference is the lack of `objc "self"`.

`Interpreter/keypath.swift` is because:
```
# | C:\Users\swift-ci\jenkins\workspace\swift-syntax-PR-windows\swift\test\Interpreter\keypath.swift:240:11: error: CHECK: expected string not found in input
# | // CHECK: \S.Type.<computed {{.*}} (S)>.subscript(_: Int)
```

That one is less obvious to me though - I don't see a `subscript` (in the logged output anyway).

bnbarham:
You can add `-disable-experimental-parser-round-trip` to this particular test case. I would have said that we should have a proper diagnostic here (to eg. add `()` as unexpected), but given that error wouldn't exist any more with your proposal, that's not really worth doing.

amritpan:
@bnbarham Would it be `// RUN: %target-typecheck-verify-swift -disable-experimental-parser-round-trip` ? This is still producing the parser diagnostics.

bnbarham:
Yeah, that should definitely work 🤔. Locally I see eg.
```
> swift-frontend t2.swift -typecheck
t2.swift:1:28: error: consecutive statements on a line must be separated by newline or ';'
1 | _ = \Int.byteSwapped.signum()
  |                            |- error: consecutive statements on a line must be separated by newline or ';'
  |                            |- note: insert newline
  |                            `- note: insert ';'
2 |
```

And then:
```
> swift-frontend t2.swift -typecheck -disable-experimental-parser-round-trip
t2.swift:1:22: error: invalid component of Swift key path
1 | _ = \Int.byteSwapped.signum()
  |                      `- error: invalid component of Swift key path
2 |
```

`target-typecheck-verify-swift` just expands to `swift-frontend -typecheck -verify -disable-objc-attr-requires-foundation-module %s`

amritpan:
Still getting errors. I also tried the expanded version.

Could the [experimental flags](https://github.com/swiftlang/swift/blob/7b8338072274a19943cb156375aed159e899233a/lib/ASTGen/Sources/ASTGen/SourceFile.swift#L80) be causing issues here?

```    
mapFeature(.KeyPathWithMethodMembers, to: .keypathWithMethodMembers)
```

ahoppen:
What are the errors you’re seeing now? It might be easiest if you could push your latest revisions.

amritpan:
The errors are the same as before adding the flag. Here is full output: 

```
/.../swift/test/expr/unary/keypath/keypath-unsupported-methods.swift:48:6: error: expected error not produced
  // expected-error@-2 {{consecutive statements on a line must be separated by newline or ';'}}
~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/.../swift/test/expr/unary/keypath/keypath-unsupported-methods.swift:49:6: error: expected note not produced
  // expected-note@-3 {{insert newline}}
~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/.../swift/test/expr/unary/keypath/keypath-unsupported-methods.swift:50:6: error: expected note not produced
  // expected-note@-4 {{insert ';'}}
~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

--
```

I've added `-disable-experimental-parser-round-trip` to the test and pushed changes as well.

amritpan:
https://github.com/swiftlang/swift/pull/78823

@swift-ci Please test.

amritpan:
https://github.com/swiftlang/swift/pull/78823

@swift-ci Please test.

amritpan:
https://github.com/swiftlang/swift/pull/78823

@swift-ci Please test.

ahoppen:
That’s an unrelated test failure which was fixed by https://github.com/swiftlang/llvm-project/pull/10189. Let’s try again

@swift-ci Please test Windows

amritpan:
Do update-checkout errors need a rebase? I'm now seeing: 

```
Error on repo "C:\Users\swift-ci\jenkins\workspace\swift-syntax-PR-windows\llvm-project": Traceback (most recent call last):
  File "C:\Users\swift-ci\jenkins\workspace\swift-syntax-PR-windows\swift\utils\update_checkout\update_checkout\update_checkout.py", line 167, in update_single_repository
    config, repo_name, scheme_name, scheme_map, cross_repos_pr)
  File "C:\Users\swift-ci\jenkins\workspace\swift-syntax-PR-windows\swift\utils\update_checkout\update_checkout\update_checkout.py", line 145, in get_branch_for_repo
    .format(pr_id, repo_branch), "--tags"], echo=True)
  File "C:\Users\swift-ci\jenkins\workspace\swift-syntax-PR-windows\swift\utils\swift_build_support\swift_build_support\shell.py", line 257, in run
    raise eout
Exception: ['git', 'fetch', 'origin', 'pull/10189/merge:ci_pr_10189', '--tags']
```

ahoppen:
No, seems to be some infrastructure issue. Let’s try again. And sorry for those unrelated CI failures.

@swift-ci Please test Windows

amritpan:
Oh, we forgot to link the compiler PR.

https://github.com/swiftlang/swift/pull/78823

@swift-ci Please test Windows.

amritpan:
https://github.com/swiftlang/swift/pull/78823

@swift-ci Please test Windows.

amritpan:
https://github.com/swiftlang/swift/pull/78823

@swift-ci Please test Windows.

PR #2962, opened 2025-02-10
Remove warnings-as-errors from DocC

shahmishal:
Xcode 16.2 introduced a large set of the for `missing documentation`

Example: https://github.com/swiftlang/swift-syntax/issues/2961


ahoppen:
https://github.com/swiftlang/swift-syntax/pull/2963 should fix this in a more tightly restricted manner.

Issue #1890, opened 2023-07-07
Please add a `TriviaPiece.commentValue` property


grynspan:
### Description

There are four kinds of comment tracked by `TriviaPiece`. It'd be a nice convenience to be able to extract the string values of those comments. May I suggest something like:

```swift
extension TriviaPiece {
  public var commentValue: String? { get }
}
```

ahoppen:
Tracked in Apple’s issue tracker as rdar://111922531

grynspan:
Well, the latter case gets hard for multi-line comments. What's the content of:

```swift
/**
 * Hello, folks!
 * I am a banana!
 */
```

I think just returning the string verbatim is fine—anybody who cares about processing it can do so. Or, if it _is_ easy enough, you could provide `rawCommentValue` and `commentValue`, the latter doing the processing?

ahoppen:
I think stripping off the `///` makes sense because if you wanted to get the entire text of the `TriviaPiece` you could do `Trivia(pieces: [piece]).description` (and we could even add `description` to `TriviaPiece` but I’m not sure if that’s super useful). 

Now the question is: What should we do with the space after `///`? My proposal is the following: We implement `commentValue` on `Trivia` instead of `TriviaPiece`. What it will do is the following: Check if all line comments have a space right after `//` or `///`. If that’s the case, also remove that space from the `commentValue`. If not, we only remove the `//` or `///`. 

We could do the same for block comments and use the same logic whether to remove a space after `/*` or before `*/`.

Matejkob:
> I think stripping off the `///` makes sense because if you wanted to get the entire text of the `TriviaPiece` you could do `Trivia(pieces: [piece]).description` (and we could even add `description` to `TriviaPiece` but I’m not sure if that’s super useful).
> 
> Now the question is: What should we do with the space after `///`? My proposal is the following: We implement `commentValue` on `Trivia` instead of `TriviaPiece`. What it will do is the following: Check if all line comments have a space right after `//` or `///`. If that’s the case, also remove that space from the `commentValue`. If not, we only remove the `//` or `///`.
> 
> We could do the same for block comments and use the same logic whether to remove a space after `/*` or before `*/`.

Sounds good to me. I had the same idea in mind.

I can try implement it 🤗

Matejkob:
[...]

PR #2966, opened 2025-02-14
Add `commentValue` property to `Trivia` for cleaned comments

zyadtaha:
This PR continues the work from #2578, addressing the review comments and refining the behavior of `Trivia.commentValue`.


Sources/SwiftSyntax/Trivia.swift
```diff
@@ -42,6 +42,28 @@ public struct Trivia: Sendable {
...
   }
 
+  /// The string contents of all the comment pieces with any comments tokens trimmed.
+  public var commentValue: String {
+    var comments = [String]()
+
+    for piece in pieces {
+      switch piece {
+      case .blockComment(let text), .docBlockComment(let text):
+        let text = text.trimmingCharacters(in: "/* ").trimmingAnyCharacters(in: ["\n"])
```
ahoppen, line 52: Should we only trim `/*` as a prefix? If you have a comment with `/* ///// abc */`, I would expect the `commentValue` to be `///// abc`, not `abc`

Tests/SwiftSyntaxTest/TriviaTests.swift
```diff
@@ -69,4 +70,272 @@ class TriviaTests: XCTestCase {
...
+        Some block comment
+        * // A line comment in a block
+        * spread on many lines
+        Some doc block comment
+        * /// A doc line comment in a block
+        * spread on
+        * many lines
+        """
+    )
+  }
```
ahoppen, line 312: Comments often occur nested in source code, eg. 

```swift
struct Foo {
  /* 
  Some block comment
  with another line
  */
  func test() {}
}
```

I think in this case, we should strip the leading indentation of the `/*` from the subsequent lines. We should still maintain any indentation relative to `/*` because it’s relevant for eg. markdown lists with sub-bullet, like

```
* Task 
  * Subtask
* Next Task
```

Sources/SwiftSyntax/Trivia.swift
```diff
@@ -42,6 +42,28 @@ public struct Trivia: Sendable {
...
+        comments.append(String(text))
+
+      case .lineComment(let text), .docLineComment(let text):
+        let text = text.trimmingPrefix("/ ")
```
ahoppen, line 56: I think we want to check if the `/` are followed by a space in all `.lineComment` pieces in this `Trivia` and if so, trim a single space, if not, not trim any space. For example both

```
// - Task
//   - Subtask
// - Task 2
```

and 

```
//- Task
//  - Subtask
//- Task 2
```

should return

```
- Task
  - Subtask
- Task 2
```

as the comment value.

Sources/SwiftSyntax/Utils.swift
```diff
@@ -102,3 +102,49 @@ extension RawUnexpectedNodesSyntax {
     self.init(raw: raw)
   }
 }
+
+extension String {
+  func trimmingCharacters(in charactersToTrim: any BidirectionalCollection<Character>) -> Substring {
```
ahoppen, line 107: Based on my comments above, I’m not sure we ever want to trim all characters from a collection. If so, I’m not sure if these extensions are needed 😉

ahoppen (commented):
Thanks for addressing the issue, @zyadtaha. I think the PR still needs some polishing regarding indentation handling. I left a few comments inline.

Tests/SwiftSyntaxTest/TriviaTests.swift
zyadtaha: Thanks for pointing this out!

zyadtaha:
Thanks for the feedback, @ahoppen! I've addressed the comments—let me know if anything else needs attention.

Sources/SwiftSyntax/Trivia.swift
```diff
@@ -42,6 +44,80 @@ public struct Trivia: Sendable {
...
+    let allLineCommentsHaveSpace = pieces.allSatisfy { piece in
+      guard case .lineComment(let text) = piece else { return true }
+      return text.hasPrefix("// ")
+    }
+
+    // Determine if all doc line comments have a single space
+    let allDocLineCommentsHaveSpace = pieces.allSatisfy { piece in
+      guard case .docLineComment(let text) = piece else { return true }
+      return text.hasPrefix("/// ")
+    }
```
ahoppen, line 61: I would combine this into a single property so that we also handle comments like the following.

```
/// first
// second
```

```
/// first
//second
```

Also, that only requires one iteration over the pieces which is more efficient.

--- 

Also, this could be a `lazy` variable so that we don’t need to compute it if the comment is a block comment.

Sources/SwiftSyntax/Trivia.swift
```diff
@@ -10,6 +10,8 @@
 //
 //===----------------------------------------------------------------------===//
 
+import Foundation
```
ahoppen, line 13: SwiftSyntax is a low-level library that’s below `Foundation` in the stack and thus can’t depend on `Foundation` … unfortunately.

Sources/SwiftSyntax/Trivia.swift
```diff
@@ -42,6 +44,80 @@ public struct Trivia: Sendable {
...
+    }
+
+    // Helper function to process multiline block comments
+    func processMultilineBlockComment(_ text: String) -> String {
+      var lines = text.components(separatedBy: .newlines)
+
+      let lastLine = lines.last ?? ""
+      let leadingSpaces = lastLine.prefix { $0 == " " }.count
+
+      lines.removeFirst()
```
ahoppen, line 80: Does this mean we always drop the first line, even if it is

```
/* Something
 Does a thing
 */
```

--- 

Same for the last line, eg.

```
/* 
 Something
 Does a thing */
```

Sources/SwiftSyntax/Trivia.swift
```diff
@@ -42,6 +44,80 @@ public struct Trivia: Sendable {
...
+      let newlineCharacterSet = CharacterSet(charactersIn: "\n")
+      return text.trimmingCharacters(in: newlineCharacterSet)
+    }
+
+    // Helper function to process multiline block comments
+    func processMultilineBlockComment(_ text: String) -> String {
+      var lines = text.components(separatedBy: .newlines)
+
+      let lastLine = lines.last ?? ""
+      let leadingSpaces = lastLine.prefix { $0 == " " }.count
```
ahoppen, line 78: Does this drop all characters based on the indentation of the last line, eg.

```
  /* 
unindented line
  */
```

Would loose `un`, right?

ahoppen (commented):
Thanks for the adjustments. I left a few more edge cases to consider as comments.

Sources/SwiftSyntax/Trivia.swift
zyadtaha: @ahoppen Yes, the current code would drop `un` from `unindented line`. We can fix this by calculating the minimum indentation across all lines and removing it. Is this mixed indentation use case something you've encountered in real-world scenarios?

ahoppen: swift-syntax is a widely used library and it should be correct in all Swift code. I haven’t encountered this exact case but my experience is that every code that you can possible imagine is somewhere out there.

Sources/SwiftSyntax/Trivia.swift
zyadtaha: For the first line, no, it won't be dropped because the `processMultilineBlockComment` function is only invoked when the text starts with `/*\n` or `/**\n`. As for the last line, I'll address that.

zyadtaha:
Hey @ahoppen, just wanted to check if there's anything else needed from my side on this PR. Let me know if you have any concerns. Thanks!

Sources/SwiftSyntax/Trivia.swift
```diff
@@ -42,6 +42,112 @@ public struct Trivia: Sendable {
...
+      let trimmed = text.drop(while: { $0 == " " })
+        .reversed()
+        .drop(while: { $0 == " " })
+        .reversed()
+      return String(trimmed)
+    }
+
+    // Helper function to trim leading and trailing newlines
+    func trimNewlines(_ text: String) -> String {
+      let trimmed = text.drop(while: { $0 == "\n" })
```
ahoppen, line 74: You should also cover `\r\n` (Windows-style newlines) here as well.

Sources/SwiftSyntax/Trivia.swift
```diff
@@ -42,6 +42,112 @@ public struct Trivia: Sendable {
...
+        .drop(while: { $0 == "\n" })
+        .reversed()
+      return String(trimmed)
+    }
+
+    // Helper function to process block comments
+    func processBlockComment(_ text: String, prefix: String, suffix: String) -> String {
+      var text = text
+      text.removeFirst(prefix.count)
+      text.removeLast(suffix.count)
```
ahoppen, line 85: We can’t generally assume that the comment will end with `*/`. For example, if you have a source file with an unterminated block comment, the end of file token will have the block comment as trivia but that trivia doesn’t end with `*/`.

Sources/SwiftSyntax/Trivia.swift
```diff
@@ -42,6 +42,112 @@ public struct Trivia: Sendable {
...
+      text = trimWhitespace(text)
+      text = trimNewlines(text)
+      return text
+    }
+
+    // Helper function to process multiline block comments
+    func processMultilineBlockComment(_ text: String) -> String {
+      var lines = text.split(separator: "\n", omittingEmptySubsequences: false).map(String.init)
```
ahoppen, line 93: Could you avoid constructing `String` here and instead work on `Substring`. `Substring` avoids memory allocations and is thus more performant.

Sources/SwiftSyntax/Trivia.swift
```diff
@@ -42,6 +42,112 @@ public struct Trivia: Sendable {
...
+
+      lines.removeFirst()
+
+      let minIndentation =
+        lines
+        .filter { !$0.isEmpty }
+        .map { $0.prefix { $0 == " " }.count }
```
ahoppen, line 100: I guess we should also handle tabs as indentation.

Sources/SwiftSyntax/Trivia.swift
```diff
@@ -42,6 +42,112 @@ public struct Trivia: Sendable {
...
+    func processBlockComment(_ text: String, prefix: String, suffix: String) -> String {
+      var text = text
+      text.removeFirst(prefix.count)
+      text.removeLast(suffix.count)
+      text = trimWhitespace(text)
+      text = trimNewlines(text)
+      return text
+    }
+
+    // Helper function to process multiline block comments
```
ahoppen, line 91: Could you be a little more specific in the comment what this function does? As it stands right now, the comment doesn’t really add any value, it just rephrases the function name.

Sources/SwiftSyntax/Trivia.swift
```diff
@@ -42,6 +42,112 @@ public struct Trivia: Sendable {
...
+        .filter { !$0.isEmpty }
+        .map { $0.prefix { $0 == " " }.count }
+        .min() ?? 0
+
+      if let lastLine = lines.last {
+        if trimWhitespace(lastLine) == "*/" {
+          lines.removeLast()
+        } else {
+          lines[lines.count - 1].removeLast(2)
+          lines[lines.count - 1] = trimWhitespace(lines[lines.count - 1])
```
ahoppen, line 108: This removes all whitespace from the last line, not just `minIndentation`, right?

Actually, on second reading, does their trim whitespace and then drop the prefix again in the `map` below, potentially removing contents of the comment?

Sources/SwiftSyntax/Trivia.swift
```diff
@@ -42,6 +42,112 @@ public struct Trivia: Sendable {
...
+      return unindentedLines.joined(separator: "\n")
+    }
+
+    for piece in pieces {
+      switch piece {
+      case .blockComment(let text):
+        let processedText =
+          text.hasPrefix("/*\n")
+          ? processMultilineBlockComment(text)
+          : processBlockComment(text, prefix: "/*", suffix: "*/")
```
ahoppen, line 126: I don’t think we need a general distinction here. Say if you have

```
  /* abc
      def
      ghi
   */
```

We should still be able to strip the indentation. We just need to exclude the first line from the indentation stripping.

Sources/SwiftSyntax/Trivia.swift
```diff
@@ -42,6 +42,112 @@ public struct Trivia: Sendable {
     pieces.isEmpty
   }
 
+  /// The string contents of all the comment pieces with any comments tokens trimmed.
+  public var commentValue: String {
+    var comments = [String]()
```
ahoppen, line 47: Could this be an array of `Substring` instead of `String` to avoid intermediate memory allocations for the comment value pieces?

ahoppen (commented):
Thanks for the updates. I think there are still a few more issues that need to be addressed.

Sources/SwiftSyntax/Trivia.swift
zyadtaha: Then maybe we should count each tab as 2 spaces?

Sources/SwiftSyntax/Trivia.swift
zyadtaha: Yep, you’re totally right! It’s all fixed now, and I’ve added a test to cover it.

zyadtaha:
Hey @ahoppen, I've pushed my updates. For your last two comments, I've added responses—let me know your feedback when you get a chance. Thanks!

Sources/SwiftSyntax/Trivia.swift
ahoppen: We shouldn’t assume any tab setting. Instead, I think we should find the lowest common indentation that all lines have. 

For example:
- If all lines start with 2 tabs, then `minIndentation` should be two tabs.
- If we have one line starting with 1 tab + 2 spaces and one with 2 tabs, then `minIndentation` should be one tab because we shouldn’t make assumptions about how the tab width
- If one line has a tab and another line has 4 spaces, `minIndentation` should be empty because there is no common indentation here.

Sources/SwiftSyntax/Trivia.swift
```diff
@@ -42,6 +42,101 @@ public struct Trivia: Sendable {
...
+      var lines = text.split(separator: "\n", omittingEmptySubsequences: false)
+
+      let minIndentation =
+        lines
+        .dropFirst()
+        .filter { !$0.isEmpty }
+        .map { $0.prefix { $0 == " " }.count }
+        .min() ?? 0
+
+      var firstLineRemoved = false;
```
ahoppen, line 83: Superfluous semicolon (in other places as well).

Sources/SwiftSyntax/Trivia.swift
```diff
@@ -42,6 +42,101 @@ public struct Trivia: Sendable {
...
+      var firstLine = lines[0]
+      if trimWhitespace(firstLine) == "/*" || trimWhitespace(firstLine) == "/**" {
+          lines.removeFirst()
+          firstLineRemoved = true;
+      } else {
+          firstLine = firstLine.hasPrefix("/**") ? firstLine.dropFirst(3) : firstLine.dropFirst(2)
+          while firstLine.first?.isWhitespace == true {
+              firstLine = firstLine.dropFirst()
+          }
```
ahoppen, line 92: Why do we need to drop whitespace-only lines here? We don’t drop them in the middle of the comment either. Or if the first line has `/*` on its own line.

Sources/SwiftSyntax/Trivia.swift
```diff
@@ -42,6 +42,101 @@ public struct Trivia: Sendable {
...
+      let minIndentation =
+        lines
+        .dropFirst()
+        .filter { !$0.isEmpty }
+        .map { $0.prefix { $0 == " " }.count }
+        .min() ?? 0
+
+      var firstLineRemoved = false;
+      var firstLine = lines[0]
+      if trimWhitespace(firstLine) == "/*" || trimWhitespace(firstLine) == "/**" {
```
ahoppen, line 85: Do we really need this check here? Wouldn’t it be easier to just do normal processing and right at the end, check if the processed comment starts with a whitespace-only line and, if so, drop that?

Sources/SwiftSyntax/Trivia.swift
```diff
@@ -42,6 +42,101 @@ public struct Trivia: Sendable {
...
+      } else {
+          firstLine = firstLine.hasPrefix("/**") ? firstLine.dropFirst(3) : firstLine.dropFirst(2)
+          while firstLine.first?.isWhitespace == true {
+              firstLine = firstLine.dropFirst()
+          }
+          lines[0] = firstLine
+      }
+      
+      if let lastLine = lines.last {
+          if trimWhitespace(lastLine) == "*/" {
```
ahoppen, line 97: Similar here, would it be easier to do normal processing and drop the last line from `unindentedLines` if it is whitespace-only?

Sources/SwiftSyntax/Trivia.swift
```diff
@@ -42,6 +42,101 @@ public struct Trivia: Sendable {
...
+
+      case .docLineComment(let text):
+        let prefix = allLineCommentsHaveSpace ? "/// " : "///"
+        comments.append(text.dropFirst(prefix.count))
+
+      default:
+        break
+      }
+    }
+    return comments.joined(separator: "\n")
```
ahoppen, line 137: I just realized that if you have `/* abc */ /* def */` as your trivia, then this will return 

```
abc
def
```
as the comment value (concatenated by newline), which isn’t what I would expect. I’m leaning towards making `commentValue` return an Optional and return `nil` if:
- No comment is in the trivia
- The trivia contains more than one block comment or a block comment and a line comment.

Sources/SwiftSyntax/Trivia.swift
```diff
@@ -42,6 +42,137 @@ public struct Trivia: Sendable {
...
+            case " ":
+              spaceCount += 1
+            case "\t":
+              tabCount += 1
+            default:
+              inLeadingWhitespace = false
+            }
+          }
+
+          return (Swift.min(currentMin.0, spaceCount), Swift.min(currentMin.1, tabCount))
```
ahoppen, line 99: I think this doesn’t work if you have one line that contains a space as indentation and another one that only contains a tab as indentation.

zyadtaha: In this case:
```
/* comment
\tcomment*/
```
The `minIndentation` would be 1 because it only starts checking from the second line, which contains a tab. Since the first line is ignored during trimming if it contains the `/*`, the output would be:
```
 comment
comment
```
But in this case:
```
/* 
 comment
\tcomment*/
``` 
The `minIndentation` would be 0, as expected, and the output would be:
```
 comment
\tcomment
```
If we want both cases to produce consistent output, maybe we could insert a `\n` after the `/*` if it's not already present, to ensure that `minIndentation` is 0. What do you think?

zyadtaha:
> I addressed my own review comments in a [2197103](https://github.com/swiftlang/swift-syntax/commit/2197103cbcd2a3f89901734ecfca76010b73140e), which I pushed to your branch. Could you check if these changes make sense to you?

Thank you! I pulled the changes and reviewed the new logic and tests, and everything looks great to me. I really appreciated the detailed doc comments; they helped me understand your approach much better. I also liked how you split the tests into separate functions since it makes things much clearer.

I just had one question: why did you move the `commentValue` property to a separate file, and similarly, the test cases to a separate file?"

Also, if you have any feedback on my code, communication, or anything I could improve, I’d be very grateful. I want to make sure I’m not wasting your time and that I'm learning as much as possible through this process.

ahoppen:
> Thank you! I pulled the changes and reviewed the new logic and tests, and everything looks great to me. I really appreciated the detailed doc comments; they helped me understand your approach much better. I also liked how you split the tests into separate functions since it makes things much clearer.

Thanks for checking that the changes look good to you.

> I just had one question: why did you move the `commentValue` property to a separate file, and similarly, the test cases to a separate file?"

The implementation of `commentValue` ended up being non-trivial and since it isn’t all that related to the other `Trivia` functionality (which is a lot more low-level), I decided to put it into its own file. This way we can also communicate that the extensions defined in that file are only intended for the `commentValue` property.

> Also, if you have any feedback on my code, communication, or anything I could improve, I’d be very grateful. I want to make sure I’m not wasting your time and that I'm learning as much as possible through this process.

You did great work to get things started. I just figured that it was easier to fix the edge cases instead of repeatedly pointing out edge cases and then discovering more after you addressed them. Thinking of all possible edge cases is a big part of developing compiler-related tooling and it’s hard to teach that general way of thinking in pull request comments.

Sources/SwiftSyntax/Trivia+commentValue.swift
```diff
@@ -0,0 +1,143 @@
...
+        if hasBlockComment || hasLineComment {
+          return nil
+        }
+        hasBlockComment = true
+        guard let processedText = processBlockComment(text, isDocComment: piece.isDocComment) else {
+          return nil
+        }
+        comments.append(processedText[...])
+      case .lineComment(let text), .docLineComment(let text):
+        if hasBlockComment {
```
rintaro, line 94: So we allow
```swift
/// Foo Bar
/// This is a doc-comment.
// @_spi(Testing)
public fooBar() {}
```
resulting `"Foo Bar\nThis is a doc-comment\n@_spi(Testing)"`, but not
```swift
/**
 * Foo Bar
 * This is a doc-comment.
 */
// @_spi(Testing)
public fooBar() {}
```

Not sure about this case.

rintaro (commented):
I'm not sure about concatenating the pieces. The current behavior sounds fragile and not predictable.

`TriviaPiece.commentValue: String?` (as https://github.com/swiftlang/swift-syntax/issues/1890 requested) would make more sense and we should implement that at least. But it's also debatable what to do with the space after `//`.

`Trivia.commentValue: String?` should be IMO a high-level API based on `TriviaPiece.commentValue`, but the concatenating behavior mixing trivia kinds should be discussed more.

Maybe `Trivia.docCommentValue: String?` that would only handle `.docLineComment`/`.docBlockComment` makes more sense.

zyadtaha:
> `TriviaPiece.commentValue: String?` (as #1890 requested) would make more sense and we should implement that at least. But it's also debatable what to do with the space after `//`.

I appreciate the thoughtful feedback! I agree that `TriviaPiece.commentValue: String?` (per #1890) is a great starting point. On the space after `//`, I'm not sure we should preserve it. In my view, something like `TriviaPiece.lineComment("// Hello")` could simply result in `commentValue = "Hello"`, not `" Hello"`. Keeping the space feels unnecessary with the likely intent of accessing the actual comment content. What’s your take on this?

> `Trivia.commentValue: String?` should be IMO a high-level API based on `TriviaPiece.commentValue`, but the concatenating behavior mixing trivia kinds should be discussed more.
> 
> Maybe `Trivia.docCommentValue: String?` that would only handle `.docLineComment`/`.docBlockComment` makes more sense.


Your suggestion of `Trivia.docCommentValue: String?` for `.docLineComment` and `.docBlockComment` makes sense to me. As Alex noted (https://github.com/swiftlang/swift-syntax/pull/2966#discussion_r2049453045), it could return `nil` for multiple block comments or mixed comment types. Similarly, `Trivia.commentValue: String?` could focus on `.lineComment` and `.blockComment`, returning `nil` in those same cases. What do you think?

ahoppen:
@rinataro and I just chatted about this and agreed that it would be best if we focus on `docCommentValue` first – its semantics are a lot clearer and we don’t need to decide how to concatenate mixtures of line comments and doc line comments etc.

The doc line comment should return the value of the last doc comment that’s inside the `Trivia`: 
- For line comments, this means that it should return the last consecutive block of doc line comments, stopping at empty lines or lines with other trivia types. In all of the cases below, we should return `X Y` and not `A`
```swift
/// not included

/// included
/// also included
func test() {}

/// not included
//
/// included
func test() {}


/// not included
/* abc */
/// included
func test() {}

/** not included */
/// included
func test() {}

/// included
// ignored
func test() {}
```
- For block comments, we were less sure whether we should only return the contents of the last doc block comment or whether we should concatenate the different doc block comments. We were leaning towards concatenating them because that’s what the current compiler does, though the concatenation behavior is still up for debate. Again, details to be decided, I think that if there are multiple block comments, we should pretend that the text between `*/` and `/**` doesn’t exist.

zyadtaha:
Hi @ahoppen and @rintaro,
Sorry for the late reply - I was busy with final exams, but summer is finally here and I'm excited to get back to contributing!

I've just pushed my latest changes. The implementation now handles concatenation for doc block comments as the compiler does. If no doc block comments are found, it falls back to returning the last consecutive block of doc line comments. Please let me know if there are any new details!

Sources/SwiftSyntax/Trivia+commentValue.swift
```diff
@@ -0,0 +1,144 @@
...
+          if hasBlockComment {
+            comments.append(processedComment)
+          } else {
+            hasBlockComment = true
+            comments = [processedComment]
+          }
+        }
+        currentLineComments = [] // Reset line comments when encountering a block comment
+      case .docLineComment(let text):
+        let prefixToDrop = ("///") + (allLineCommentsHaveSpace ? " " : "")
```
ahoppen: Hmm, with the new implementation of only looking at the doc comments and only using the last run of line comments as the doc comment value, I don’t think we should check all line comments for spaces anymore. For example consider

```swift
//Blah, unrelated
/// This is the
/// real comment
```

or

```swift
///Blah, unrelated

/// This is the
/// real comment
```

I think in this case we should strip the leading newline. I would suggest doing this stripping as a post-processing step down below in `if !currentLineComments.isEmpty`.

Sources/SwiftSyntax/Trivia+commentValue.swift
```diff
@@ -0,0 +1,144 @@
...
+      if unindentedLines.last?.allSatisfy({ $0 == " " }) ?? false {
+        unindentedLines.removeLast()
+      }
+      // We canonicalize the line endings to `\n` here. This matches how we concatenate the different line comment
+      // pieces using \n as well.
+      return unindentedLines.joined(separator: "\n")[...]
+    }
+
+    var currentLineComments: [Substring] = []
+    var foundStop = false
```
ahoppen: Would the name `isInsideDocLineCommentSection` be more descriptive here? I think that would entail inverting the value down below.

Sources/SwiftSyntax/Trivia+commentValue.swift
```diff
@@ -0,0 +1,144 @@
...
+      case .docLineComment(let text):
+        let prefixToDrop = ("///") + (allLineCommentsHaveSpace ? " " : "")
+        if foundStop {
+          currentLineComments = [text.dropPrefix(prefixToDrop)]
+          foundStop = false
+        } else {
+          currentLineComments.append(text.dropPrefix(prefixToDrop))
+        }
+      case .newlines(let count) where count == 1:
+        continue
```
ahoppen: 1. This can be simplified to `case .newlines(1)`
2. We should also handle `carriageReturns` and `carriageReturnLineFeeds` here for different style line endings.
3. The trivia could also contain `[.newlines(1), .newlines(1)]`. We should detect that as well and shouldn’t rely on trivia being normalized to `[.newlines(2)]`.

Sources/SwiftSyntax/Trivia+commentValue.swift
zyadtaha: > I don’t think we should check all line comments for spaces anymore.

Maybe we could instead check `currentLineComments` for spaces as an also post-processing step in `if !currentLineComments.isEmpty`?

This would better handle cases like:
```
/// - Task
///   - Subtask
/// - Task 2
```

ahoppen: Yes, if I understand you correctly, that’s exactly what I meant.

zyadtaha: I had to do the stripping outside `if !currentLineComments.isEmpty` because `currentLineComments` could be empty while `comments` still contains line comments that need to be stripped.

For example:
```
/// included
// ignored
```

Sources/SwiftSyntax/Trivia+commentValue.swift
```diff
@@ -0,0 +1,153 @@
...
+
+extension Trivia {
+  /// The contents of the last doc comment piece with any comment markers removed and indentation whitespace stripped.
+  public var docCommentValue: String? {
+    var comments: [Substring] = []
+
+    /// Keep track of whether we have seen a line or block comment trivia piece.
+    var hasBlockComment = false
+
+    var currentLineComments: [Substring] = []
```
ahoppen, line 21: This could now be of type `[String]`, right? We don’t ever add substrings to this array if I understand correctly.

Sources/SwiftSyntax/Trivia+commentValue.swift
```diff
@@ -0,0 +1,153 @@
...
+    var isInsideDocLineCommentSection = true
+    var consecutiveNewlines = 0 
+
+    for piece in pieces {
+      switch piece {
+      case .docBlockComment(let text):
+        if let processedComment = processBlockComment(text) {
+          if hasBlockComment {
```
ahoppen, line 29: Do we need the variable `hasBlockComment`? If I understand correctly, we never set it to `false` and if it’s `false` then we always have `comments == []`, which means that the `else`-block is equivalent to this `if` branch.

Tests/SwiftSyntaxTest/TriviaCommentValueTests.swift
```diff
@@ -0,0 +1,304 @@
...
+        Some line comment
+        """
+    )
+
+    assertCommentValue(
+      """
+      /** abc */
+      /** def */
+      """, 
+    docCommentValue: "abc\ndef")
```
ahoppen, line 267: Sorry in case we discussed this before and I just don’t remember but my gut feeling is that the result of this should be `def`, not `abc\ndef`. @rintaro What do you think?

rintaro: I vote for `def`, i.e. only use the last block doc-comment.
FWIW, the current SourceKit seems to concat them as long as there's no blank lines between them, but that's also true for this case.
```swift
/** not included */
/// included
```

PR #2971, opened 2025-02-24
[SE-0458] Implement "unsafe" effect for the for-in loop

DougGregor:


DougGregor:
https://github.com/swiftlang/swift/pull/79573

@swift-ci please test

DougGregor:
https://github.com/swiftlang/swift/pull/79573

@swift-ci please test Linux

DougGregor:
https://github.com/swiftlang/swift/pull/79573

@swift-ci please test

DougGregor:
https://github.com/swiftlang/swift/pull/79573

@swift-ci please test

DougGregor:
https://github.com/swiftlang/swift/pull/79573

@swift-ci please test macOS

DougGregor:
https://github.com/swiftlang/swift/pull/79573

@swift-ci please test Windows

Sources/SwiftParser/Statements.swift
```diff
@@ -570,6 +570,16 @@ extension Parser {
     let tryKeyword = self.consume(if: .keyword(.try))
     let awaitKeyword = self.consume(if: .keyword(.await))
 
+    let unsafeKeyword: RawTokenSyntax?
+    if let modifierKeyword = ExpressionModifierKeyword(
+      lexeme: self.currentToken,
+      experimentalFeatures: self.experimentalFeatures
+    ), modifierKeyword == .unsafe, !self.peek(isAt: .keyword(.in)) {
+      unsafeKeyword = self.consumeAnyToken(remapping: .keyword)
```
ahoppen, line 578: I think we also need to check at least for `:` to make sure the following continues to parse:

```swift
for unsafe: Int in [1, 2] {
  print(unsafe)
}
```

Also, I’d prefer to do `self.expect(.keyword(.unsafe))` instead of `consumeAnyToken`. It just adds more context to what kind of keyword `unsafeKeyword` is and another layer of safety if the current token should not be `unsafe` for any reason in the future.

Tests/SwiftParserTest/StatementTests.swift
```diff
@@ -957,4 +957,16 @@ final class StatementTests: ParserTestCase {
...
+  func testForUnsafeStatement() {
+    assertParse(
+      "for try await unsafe x in e { }",
+      experimentalFeatures: [.unsafeExpression]
+    )
+
+    assertParse(
+      "for try await unsafe in e { }",
+      experimentalFeatures: [.unsafeExpression]
+    )
```
ahoppen, line 970: Could you also add a test for `for unsafe in e { }`?

DougGregor: Sure. There's another ambiguity I also need to check for, i.e., `for unsafe: Int in e { }` needs to not parse as the `unsafe` keyword.

Sources/SwiftParser/Statements.swift
DougGregor: Hah, I read your comments out of order. We indeed need to deal with the `:`. Thanks for the `self.expect(.keyword(.unsafe))` hint.

DougGregor:
@ahoppen I'll address your comments in a follow-up PR, because I want to get the bones of this in place for the compiler PR to land

DougGregor:
Code review comments addressed via https://github.com/swiftlang/swift-syntax/pull/2977

PR #2973, opened 2025-02-24
[CodeGeneration] Stop generating `add{ElementName}(_:)` methods

rintaro:
They have been deprecated for a while. Now it's time to remove.


ahoppen (approved):
Thanks!

PR #2977, opened 2025-02-25
[SE-0458] Improve disambiguation for "unsafe" effect on for..in loop

DougGregor:


DougGregor:
@swift-ci please test

ahoppen (approved):

PR #2978, opened 2025-02-26
[SE-0458] Enable `unsafe` expressions and for..in effects everywhere

DougGregor:
With the acceptance of SE-0458, stop using the `unsafeExpression` experimental feature to gate the `unsafe` expression and for..in effect.

Improve disambiguation for the `unsafe` expression slightly to account for this.


DougGregor:
@swift-ci please test

ahoppen (approved):

PR #2979, opened 2025-02-28
Remove experimental feature unsafeExpression

DougGregor:
With SE-0458 being enabled by default, we no longer need this.


DougGregor:
@swift-ci please test

PR #2983, opened 2025-03-05
_SwiftSyntaxCShims: rename Swift getter

compnerd:
`_errno` is a C reserved identifier. Windows uses this reserved namespace and results in a collision. Rename the getter to namespace it and avoid the collision.

### DIFF
--- a/Sources/SwiftCompilerPluginMessageHandling/CompilerPluginMessageHandler.swift
+++ b/Sources/SwiftCompilerPluginMessageHandling/CompilerPluginMessageHandler.swift
@@ -121,7 +121,7 @@ public class CompilerPluginMessageListener<Connection: MessageConnection, Handle
     } catch {
       // Emit a diagnostic and indicate failure to the plugin host,
       // and exit with an error code.
-      fputs("Internal Error: \(error)\n", _stderr)
+      fputs("Internal Error: \(error)\n", swift_syntax_stderr)
       exit(1)
     }
   }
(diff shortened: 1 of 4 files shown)


compnerd:
@swift-ci please test

rintaro (approved):

ahoppen (commented):
Why is `_errno` a reserved C identifier? As far as I know only identifiers with two underscores and a lowercase letter or a single underscore followed by an uppercase letter are reserved.

Also, if we change this, I think we should also prefix `_stdout` etc in swiftsyntax_stdio.h for consistency.

compnerd:
I can change `_stdout` et al as well.

`_` symbols are always reserved: https://en.cppreference.com/w/c/language/identifier
> All external identifiers that begin with an underscore.

The same applies to C++. However, C++ additionally reserves any identifier with `__` in the name (as opposed to C where it is just a leader).

compnerd:
@swift-ci please test

ahoppen:
Hmm, interesting. I didn’t know about the _external_ identifiers rule, assuming that _external_ means external linkage here, the spec doesn’t seem to be clear on that.

But then we shouldn’t prefix the symbol with an underscore at all, ie. change it to `swift_syntax_errno` instead of `_swift_syntax_errno`, which is technically still reserved, right?

compnerd:
Yeah, anything with external linkage (though I think that there would be a conflict since `_errno` is defined. You should however be able to use it within a function as a variable name.

I can strip the `_` on these decls.

compnerd:
@swift-ci please test

ahoppen (approved):

ahoppen:
@swift-ci Please test

compnerd:

PR #2993, opened 2025-03-06
Enable Windows tests in GitHub Actions

### LINKED ISSUE #2992: Enable Windows CI checks in GitHub Actions
Make the tests pass and enable them.

ahoppen:
Fixes #2992

This exposed a bug in swift-syntax, which I’m fixing:

Normalize line ending to `\n` in `StringLiteralExprSyntax.representedLiteralValue`

This matches the specification of multi-line string literal handling:

From https://docs.swift.org/swift-book/documentation/the-swift-programming-language/lexicalstructure/#String-Literals:
> Line breaks in a multiline string literal are normalized to use the line feed character. Even if your source file has a mix of carriage returns and line feeds, all of the line breaks in the string will be the same.

From https://lists.swift.org/pipermail/swift-evolution/Week-of-Mon-20170417/035923.html:

> The quoted string should normalize newlines to \n in the value of the literal, regardless of whether the source file uses \n (Unix), \r\n (Windows), or \r (classic Mac) line endings. Likewise, when the compiler strips the initial and final newline from the literal value, it will strip one of any of the \n, \r\n, or \r line-ending sequences from both ends of the literal.

### DIFF
--- a/Sources/SwiftParser/StringLiteralRepresentedLiteralValue.swift
+++ b/Sources/SwiftParser/StringLiteralRepresentedLiteralValue.swift
@@ -83,8 +83,8 @@ extension StringSegmentSyntax {
     precondition(!hasError, "appendUnescapedLiteralValue relies on properly parsed literals")
 
     let rawText = content.rawText
-    if !rawText.contains("\\") {
-      // Fast path. No escape sequence.
+    if !rawText.contains(where: { $0 == "\\" || $0 == "\r" }) {
+      // Fast path. No escape sequence that need to be interpreted or line endings that need to be normalized to \n.
       output.append(String(syntaxText: rawText))
       return
     }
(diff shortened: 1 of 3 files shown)


Sources/SwiftParser/StringLiteralRepresentedLiteralValue.swift
```diff
@@ -105,6 +105,18 @@ extension StringSegmentSyntax {
         )
 
         switch lex {
+        case .success(Unicode.Scalar("\r")):
+          // Line endings in multi-line string literals are normalized to line feeds even if the source file has a
+          // different encoding few new lines.
```
bnbarham, line 110:
```suggestion
          // different encoding for new lines.
```

ahoppen:
@swift-ci Please test

PR #3005, opened 2025-03-07
SE-0458: Disambiguate postfix expressions vs. unsafe expressions more generally

DougGregor:
Handle call-vs-tuple and subscript-vs-collection-expr disambiguation using the same "no trivia" rule that we used to disambiguite "unsafe.x" (which we treat as a member access) from "unsafe .x" (which we treat as an unsafe expression with a leading-dot member access).

Fixes rdar://146459104.

### DIFF
--- a/Tests/SwiftParserTest/ExpressionTests.swift
+++ b/Tests/SwiftParserTest/ExpressionTests.swift
@@ -2295,6 +2295,46 @@ final class StatementExpressionTests: ParserTestCase {
       """,
       substructure: DeclReferenceExprSyntax(baseName: .identifier("unsafe"))
     )
+
+    assertParse(
+      """
+      func f() {
+        unsafe()
+      }
+      """,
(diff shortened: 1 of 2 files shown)


DougGregor:
@swift-ci please test

Tests/SwiftParserTest/ExpressionTests.swift
```diff
@@ -2295,6 +2295,46 @@ final class StatementExpressionTests: ParserTestCase {
...
+        unsafe()
+      }
+      """,
+      substructure: DeclReferenceExprSyntax(baseName: .identifier("unsafe"))
+    )
+
+    assertParse(
+      """
+      func f() {
+        unsafe ()
```
ahoppen, line 2311: Technically, this is still source breaking because we parse it as a call to `unsafe` right now. But maybe it’s not too bad in practice.

ahoppen (approved):

PR #3013, opened 2025-03-11
[CMake] Fix package CMO capability checking logic

rintaro:
Package CMO requires '-enable-library-evolution' in 6.1+ compilers.

https://github.com/swiftlang/swift/pull/79884 but for swift-syntax repository.

Note that this is not used in compiler build because CMakeLists.txt in `swift` repository provides the result by FetchContent. This is for swift-syntax is built for other scenarios that wants to build `swift-syntax` as a LE enabled library, which pretty unlikely happpens.


rintaro:
@swift-ci Please test

bnbarham (approved):

PR #3016, opened 2025-03-13
[Macros] Support function body macros on closures.

hborla:
This change does 2 things:

1. Generalize `expandAttachedMacro` and `expandAttachedMacroWithoutCollapsing` to accept `some SyntaxProtocol` for the `declarationNode` parameter to allow an argument type of `ClosureExprSyntax`.
2. Add a new requirement on `BodyMacro` that provides a function body for a given `ClosureExprSyntax`. This method has a default implementation that throws an error.


hborla:
@swift-ci please test

hborla:
@swift-ci please test

hborla:
@swift-ci please test

Sources/SwiftSyntaxMacroExpansion/MacroExpansion.swift
```diff
@@ -215,7 +215,7 @@ public func expandAttachedMacroWithoutCollapsing<Context: MacroExpansionContext>
   definition: Macro.Type,
   macroRole: MacroRole,
   attributeNode: AttributeSyntax,
-  declarationNode: DeclSyntax,
+  declarationNode: some SyntaxProtocol,
```
ahoppen, line 218: `declarationNode` is no longer the correct argument label here if this can be a closure as well, right?

Sources/SwiftSyntaxMacroExpansion/MacroExpansion.swift
```diff
@@ -380,7 +390,7 @@ public func expandAttachedMacro<Context: MacroExpansionContext>(
...
```
ahoppen, line 393: Same here, we should change the argument label, right?

Sources/SwiftSyntaxMacros/MacroProtocols/BodyMacro.swift
```diff
@@ -29,4 +29,31 @@ public protocol BodyMacro: AttachedMacro {
...
+}
+
+private struct ClosureNotSupported: Error, CustomStringConvertible {
+  var description: String {
+    "Function body macro cannot be applied to closure"
+  }
+}
+
+extension BodyMacro {
+  public static func expansion(
```
ahoppen, line 52: Should we mark this default implementation as deprecated to prompt clients to add it?

ahoppen (commented):
Changes look good to me, just a few comments. Also, could you create a short RFC for this? https://github.com/swiftlang/swift-syntax/blob/main/Contributor%20Documentation/RFC%20Process.md

Sources/SwiftSyntaxMacroExpansion/MacroExpansion.swift
hborla: That's correct, ideally the argument label should be called `attachedTo` or just `node` or similar. However, I was trying to maintain source compatibility since this is a public function, per @rintaro 's recommendation, though I'm not sure what kind of code actually calls `expandAttachedMacro{WithoutCollapsing}` from outside of swift-syntax. I'm happy to change the argument label in both APIs if you aren't concerned about source compatibility.

ahoppen: I think we should change it and have a deprecated version of this function with the old signature that forwards to the new version.

hborla: Done!

Sources/SwiftSyntaxMacros/MacroProtocols/BodyMacro.swift
hborla: It's reasonable for a function body macro to support function declarations and not closures, but yeah that should probably be a deliberate decision by the macro author by implementing the requirement explicitly rather than the default behavior.

hborla: I'll wait to add the deprecation until the RFC and SE pitch discussion in case the requirement ends up being tweaked in some way.

ahoppen (approved):
Looks good to me, I’d just like to see a short RFC post.

hborla:
@swift-ci please test

hborla:
>Looks good to me, I’d just like to see a short RFC post.

Yes, I'm going to write one. But the APIs I've added here are very closely related to the design of (experimental) closure body macros in the language, so I'm going to post an evolution pitch that includes the new `BodyMacro` requirement, and I'll link to the pitch from the RFC that covers all of the API changes. I'm working on the pitch now and hope to have them both posted tomorrow!

ahoppen:
OK, no worries. Just wanted to make sure it doesn’t get forgotten 😉

hborla:
Pitched at https://forums.swift.org/t/pre-pitch-closure-body-macros/78534

RFC at https://forums.swift.org/t/rfc-generalize-expandattachedmacro-apis-to-accept-closureexprsyntax/78535

hborla:
@swift-ci please test Windows

hborla:
@swift-ci please test Windows

PR #3019, opened 2025-03-17
Check `thenStatements` feature in `CanBeStatementStart`

hamishknight:
I added the `experimentalFeatures` parameter in the TokenSpecSet initializer for `do` statements but never went back to apply the change to `then` statements. Should be NFC.


hamishknight:
@swift-ci please test

ahoppen (approved):

PR #3020, opened 2025-03-17
[Syntax] Workaround for 'cannot bypass resilience' warnings

rintaro:
`RawSyntaxArena` and `SyntaxDataaArena` have properties with types from `_SwiftSyntaxCShims` which is `@_implementationOnly` when compiling with `-enable-library-evolution`.

But for `-allow-non-resilient-access` for the package cross module optimizations, they caused warnings in 6.1+ compilers:

```
warning: cannot bypass resilience due to member deserialization failure while attempting to access member 'mutex' of 'SyntaxDataArena' in module 'SwiftSyntax' from module 'SwiftSyntax'
warning: cannot bypass resilience due to member deserialization failure while attempting to access member 'hasParent' of 'RawSyntaxArena' in module 'SwiftSyntax' from module 'SwiftSyntax'
```

and non-resilient-access are disabled for these types.

To workaround that, use `UnsafeRawPointer` for those stored properties and use casting methods to access the actual types. Note the using computed properties were not enough to silience the warnings.

### DIFF
--- a/Sources/SwiftSyntax/Syntax.swift
+++ b/Sources/SwiftSyntax/Syntax.swift
@@ -382,7 +382,10 @@ struct SyntaxData: Sendable {
 /// `SyntaxDataArena` manages the entire data of a `Syntax` tree.
 final class SyntaxDataArena: @unchecked Sendable {
   /// Mutex for locking the data when populating layout buffers.
-  private let mutex: PlatformMutex
+  ///
+  /// - Note: `UnsafeMutableRawPointer` + casting accessor is a workaround to silence the warning 'cannot bypass resilience'.
+  private let _mutex: UnsafeMutableRawPointer?
+  private func mutex() -> PlatformMutex { PlatformMutex(opaque: self._mutex) }
 
   /// Allocator.
   private let allocator: BumpPtrAllocator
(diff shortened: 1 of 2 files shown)


Sources/SwiftSyntax/Syntax.swift
```diff
@@ -382,7 +382,10 @@ struct SyntaxData: Sendable {
 /// `SyntaxDataArena` manages the entire data of a `Syntax` tree.
 final class SyntaxDataArena: @unchecked Sendable {
   /// Mutex for locking the data when populating layout buffers.
-  private let mutex: PlatformMutex
+  ///
+  /// - Note: `UnsafeMutableRawPointer` + casting accessor is a workaround to silence the warning 'cannot bypass resilience'.
+  private let _mutex: UnsafeMutableRawPointer!
```
ahoppen, line 387: Why does this nave to be an IUO? Shouldn’t we be able to make this a `UnsafeMutableRawPointer`?

rintaro: We have a platforms it is actually `nil`. I used `!` because the `PlatformMutex.opaque` type is `UnsafeMutableRawPointer!` (by default by ClangImporter)

rintaro:
@swift-ci Please test

Sources/SwiftSyntax/Syntax.swift
ahoppen: Oh, I see. Should we just make it a proper optional in that case?

rintaro:
@swift-ci Please test

Sources/SwiftSyntax/Syntax.swift
rintaro: Okay, I don't see any drawbacks for using optional. Done

rintaro:
@swift-ci Please test Linux

rintaro:
@swift-ci Please test Windows

rintaro:
@swift-ci Please test Linux

PR #3023, opened 2025-03-19
[Parser] Add 'atContextualKeywordPrefixedSyntax' method

rintaro:
Unified disambiguation method for contextual-keyword prefixed syntax.

### DIFF
--- a/Sources/SwiftParser/Statements.swift
+++ b/Sources/SwiftParser/Statements.swift
@@ -909,15 +909,17 @@ extension TokenConsumer {
   ///   - acceptClosure: When the next token is '{' and it looks like a closure, use this value as the result.
   ///   - preferPostfixExpr: When the next token is '.', '(', or '[' and there is a space between the word,
   ///     use `!preferPostfixExpr` as the result.
+  ///   - allowNextLineOperand: Whether the keyword-prefixed syntax accepts the operand on the next line.
   mutating func atContextualKeywordPrefixedSyntax(
     exprFlavor: Parser.ExprFlavor,
     acceptClosure: Bool = false,
-    preferPostfixExpr: Bool = true
+    preferPostfixExpr: Bool = true,
+    allowNextLineOperand: Bool = false
   ) -> Bool {
     let next = peek()
 
     // The next token must be at the same line.
-    if next.isAtStartOfLine {
+    if next.isAtStartOfLine && !allowNextLineOperand {
       return false
     }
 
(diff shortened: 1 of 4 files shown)


Sources/SwiftParser/Statements.swift
```diff
@@ -900,6 +900,112 @@ extension Parser {
...
+      //   - Trailing closure vs. immediately-invoked closure
+      //
+      // Existence of whitespace can help this because people usually put a space
+      // before trailing closures. Although it is source breaking but we prefer
+      // parsing it as a keyword.
+      let isClosure =  self.withLookahead {
+        $0.consumeAnyToken()
+        return $0.atValidTrailingClosure(flavor: exprFlavor)
+      }
+      return isClosure && acceptClosure
```
ahoppen, line 1004: Can we avoid the `isClosure` check if `acceptClosure` is `false`?

ahoppen (approved):
Thank you. Much cleaner than the old rules

rintaro:
@swift-ci Please test

Tests/SwiftParserTest/ThenStatementTests.swift
```diff
@@ -685,7 +673,10 @@ final class ThenStatementTests: ParserTestCase {
       then
         .foo
       """,
-      substructure: ThenStmtSyntax(expression: MemberAccessExprSyntax(name: .identifier("foo")))
+      substructure: MemberAccessExprSyntax(
+        base: DeclReferenceExprSyntax(baseName: .identifier("then")),
+        name: .identifier("foo")
```
hamishknight, line 678: Hmm, this does conflict [with the proposal](https://github.com/swiftlang/swift-evolution/blob/7bb5b9d6e02caaf2e7dbd9e9c084ce969be66825/proposals/NNNN-then-keyword.md#parsing-ambiguities-with-then):

> `then.foo` is a member access, `then .foo` is a then statement, as is:
> 
> ```swift
> then
>   .member
> ```
> 
> If member access was still desired, back ticks could be used:
> 
> ```swift
> `then`
>   .member
> ```
> 
> This is a potential (albeit unlikely) source break, but the back tick fix can be applied to the 5.9 compiler today to ensure existing code can compile with both the old and new compiler.

IIRC it was felt that `then` would much more likely be a function name than a variable name, which made the member interpretation fairly unlikely

Sources/SwiftParser/Statements.swift
```diff
@@ -949,91 +1067,16 @@ extension Parser.Lookahead {
...
-        // later on during type checking.
-        return true
-      default:
-        // any other token following "discard" means it's not the statement.
-        // For example, could be the function call "discard()".
-        return false
-      }
-
-    case .then:
-      return atStartOfThenStatement(preferExpr: preferExpr)
```
hamishknight, line 990: IIRC this was the only use of `preferExpr`, and it was to disambiguate the multi-line case when `then` appeared in expression position.

rintaro:
@swift-ci Please test

hamishknight (approved):
Thanks!

rintaro:
@swift-ci Please test Windows

PR #3024, opened 2025-03-19
Enable library evolution in Bazel builds

pennig:
Updating Bazel builds to use library evolution

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -16,6 +16,6 @@ module(
     compatibility_level = 1,
 )
 
-bazel_dep(name = "apple_support", version = "1.13.0", repo_name = "build_bazel_apple_support")
-bazel_dep(name = "rules_apple", version = "3.3.0", repo_name = "build_bazel_rules_apple")
-bazel_dep(name = "rules_swift", version = "1.18.0", max_compatibility_level = 2, repo_name = "build_bazel_rules_swift")
+bazel_dep(name = "apple_support", version = "1.15.1", repo_name = "build_bazel_apple_support")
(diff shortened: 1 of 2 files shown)


pennig:
I'm hoping to get this into the 6.1 release branch as well

keith:
is there more context about this somewhere I can read up on?

On the surface this breaks my understanding of what should have this set.

My mental model for this feature is that you should only ever enable it if you're planning on redistributing a prebuilt version of the library for potential consumption by multiple swift versions. I also believe there is some downside to setting this, I don't have a citation off hand but if that's true, I don't think this would be the right default.

Specifically I'm thinking about projects that pull this in directly and build it all the time, and just rely on the remote cache to solve the prebuilt concern. Projects like SwiftLint or Lyft's iOS project ofc never redistribute this for use by other swift compilers so they don't need this feature.

Would it be better to force a transition on this from the consumers BUILD file if this is desired? (I believe one of the `*_framework` rules actually tries to do this by default for you?)

Similarly I guess what makes SwiftSyntax special versus other third parties that we'd want to set it here but not more widely? (maybe you are just setting it for all third parties?)

pennig:
Thanks for sharing your thoughts. This was an attempt to address the warnings about `@_implementationOnly` imports when consuming the project with the Swift 6 language mode. I admit I don't have a complete understanding of library evolution, particularly why you would _not_ want it enabled. 

With that in mind, closing this seems fine.

ahoppen:
Note that I addressed the `@_implementationOnly` warnings in https://github.com/swiftlang/swift-syntax/pull/3002, though that change won’t be in 6.1

keith:
I'd probably patch that in to my project in the meantime to workaround

PR #3027, opened 2025-03-22
[SwiftSyntax] Add #_objectFileFormat compilation conditional

kubamracek:
SwiftSyntax part for https://github.com/swiftlang/swift/pull/80212.

> This implements the relevant part of the "Section Placement" proposal (draft: https://github.com/kubamracek/swift-evolution/blob/section-placement-control/proposals/0nnn-section-control.md, forum thread: https://forums.swift.org/t/pitch-3-section-placement-control/77435). Namely the ability to conditionalize on the object file format that the compiler is targeting (MachO, ELF, COFF of Wasm).

> Adding with an underscored name (#_objectFileFormat) until the proposal is codified.

rdar://147505228

### DIFF
--- a/Sources/SwiftIfConfig/BuildConfiguration.swift
+++ b/Sources/SwiftIfConfig/BuildConfiguration.swift
@@ -216,7 +216,7 @@ public protocol BuildConfiguration {
 
   /// Determine whether the given name is the active target object file format (e.g., ELF).
   ///
-  /// The target object file format can only be queried by an experimental 
+  /// The target object file format can only be queried by an experimental
   /// syntax `_objectFileFormat(<name>)`, e.g.,
   ///
   /// ```swift
--- a/Sources/SwiftIfConfig/BuildConfiguration.swift
+++ b/Sources/SwiftIfConfig/BuildConfiguration.swift
@@ -291,3 +291,11 @@ public protocol BuildConfiguration {
   /// #endif
   var compilerVersion: VersionTuple { get }
 }
+
+/// Default implementation of BuildConfiguration, to avoid a revlock with the
+/// swift repo.
+extension BuildConfiguration {
+  func isActiveTargetObjectFileFormat(name: String) throws -> Bool {
+    return false
+  }
+}
--- a/Sources/SwiftIfConfig/BuildConfiguration.swift
+++ b/Sources/SwiftIfConfig/BuildConfiguration.swift
@@ -214,6 +214,21 @@ public protocol BuildConfiguration {
   /// pointer authentication scheme.
   func isActiveTargetPointerAuthentication(name: String) throws -> Bool
 
+  /// Determine whether the given name is the active target object file format (e.g., ELF).
+  ///
+  /// The target object file format can only be queried by an experimental 
+  /// syntax `_objectFileFormat(<name>)`, e.g.,
+  ///
+  /// ```swift
+  /// #if _objectFileFormat(ELF)
+  /// // Special logic for ELF object file formats
+  /// #endif
+  /// ```
+  /// - Parameters:
+  ///   - name: The name of the object file format.
+  /// - Returns: Whether the target object file format matches the given name.
+  func isActiveTargetObjectFileFormat(name: String) throws -> Bool
+
   /// The bit width of a data pointer for the target architecture.
   ///
   /// The target's pointer bit width (which also corresponds to the number of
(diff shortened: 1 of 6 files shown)


ahoppen (commented):
Should this also be guarded behind and experimental feature?

Edit: This should be guarded behind an experimental feature.

Sources/SwiftIfConfig/BuildConfiguration.swift
```diff
@@ -214,6 +214,21 @@ public protocol BuildConfiguration {
...
+  ///
+  /// ```swift
+  /// #if _objectFileFormat(ELF)
+  /// // Special logic for ELF object file formats
+  /// #endif
+  /// ```
+  /// - Parameters:
+  ///   - name: The name of the object file format.
+  /// - Returns: Whether the target object file format matches the given name.
+  func isActiveTargetObjectFileFormat(name: String) throws -> Bool
```
rintaro, line 230: Could you add a default implementation returning `false` so that we can land this `swift-syntax` PR independently before swift-syntax counterpart?

rintaro (commented):
Regarding the experimental feature flag, we currently don't have infrastructure to handle feature flags in `SwiftIfConfig`, and I don't think we should include such change in this specific PR. So IMO we should take this as-is, then establish the feature flag facility asap, _if_ we want to hide this in a feature flag.

rintaro:
As for the "format check" failure, please just do `swift format -ipr .` in the repository.

Sources/SwiftIfConfig/BuildConfiguration.swift
kubamracek: Added!

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

Sources/SwiftIfConfig/BuildConfiguration.swift
rintaro: The test failure is because this default implementation is not `public` 🙏

ahoppen: We should also make this `@_spi(ExperimentalLanguageFeatures)` so that clients can’t call it as public API and start relying on it.

Sources/SwiftIfConfig/BuildConfiguration.swift
```diff
@@ -276,3 +291,11 @@ public protocol BuildConfiguration {
   /// #endif
   var compilerVersion: VersionTuple { get }
 }
+
+/// Default implementation of BuildConfiguration, to avoid a revlock with the
+/// swift repo.
+extension BuildConfiguration {
+  func isActiveTargetObjectFileFormat(name: String) throws -> Bool {
+    return false
```
ahoppen, line 299: I think this method should throw an error that has a description like `'_objectFileFormat' is an experimental feature`. That way `_objectFileFormat` will not be allowed unless `BuildConfiguration` actively implements it. This way we also don’t silently assume that there is no active target object file format, as we do right now.

The `BuildConfiguration` implementation in ASTGen can then check whether the `_objectFileFormat` experimental feature is enabled. If not, it can throw a similar error. And if the feature is enabled, it can return `true` or `false`.

Sources/SwiftIfConfig/BuildConfiguration.swift
```diff
@@ -276,3 +291,11 @@ public protocol BuildConfiguration {
   /// #endif
   var compilerVersion: VersionTuple { get }
 }
+
+/// Default implementation of BuildConfiguration, to avoid a revlock with the
+/// swift repo.
```
ahoppen, line 296: This is not only to avoid revlock with the compiler repo. It is also to not introduce a new requirement to `BuildConfiguration` for an experimental feature that would be API breaking.

Sources/SwiftIfConfig/IfConfigDiagnostic.swift
```diff
@@ -36,6 +36,7 @@ enum IfConfigDiagnostic: Error, CustomStringConvertible {
   case likelySimulatorPlatform(syntax: ExprSyntax)
   case likelyTargetOS(syntax: ExprSyntax, replacement: ExprSyntax?)
   case endiannessDoesNotMatch(syntax: ExprSyntax, argument: String)
+  case objectFileFormatDoesNotMatch(syntax: ExprSyntax, argument: String)
```
ahoppen, line 39: Is this used anywhere?

Sources/SwiftIfConfig/BuildConfiguration.swift
kubamracek: Done

ahoppen (approved):

kubamracek:
@swift-ci please test

kubamracek:
@swift-ci please test windows

PR #3033, opened 2025-03-29
Fix build warning

SimplyDanny:
This fixes the build warning

```bash
    .../.build/checkouts/swift-syntax/Sources/SwiftLexicalLookup/CMakeLists.txt
```

which was fixed on `main` but not for the 601.0.0 release.


kimdv:
@swift-ci please test

ahoppen (approved):
Thanks!

PR #3034, opened 2025-04-01
Specify SwiftSyntax601 in more places.

mbrandonw:
Currently the trick of using version modules to conditionally access Swift syntax is broken for version 601:

```swift
#if canImport(SwiftSyntax601)
  // Code in here is never compiled
#endif
```

We use these version modules heavily to support SwiftSyntax all the way back to 509, but now multiple packages in our ecosystem are broken with no way to fix them other than rolling back all SwiftSyntax support back to 600.

I believe the fix is to simply update the dependencies of SwiftSyntax to use SwiftSyntax601, but I went ahead and fixed up other inconsistencies with version modules throughout the repo. If this is the correct fix then hopefully we can get a patch release of SwiftSyntax released soon?

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -211,7 +211,7 @@ let package = Package(
     .target(
       name: "SwiftSyntax",
       dependencies: [
-        "_SwiftSyntaxCShims", "SwiftSyntax509", "SwiftSyntax510", "SwiftSyntax600", "SwiftSyntax601", "SwiftSyntax602"
+        "_SwiftSyntaxCShims", "SwiftSyntax509", "SwiftSyntax510", "SwiftSyntax600", "SwiftSyntax601", "SwiftSyntax602",
       ],
       exclude: ["CMakeLists.txt"],
       swiftSettings: swiftSyntaxSwiftSettings
--- a/Package.swift
+++ b/Package.swift
@@ -210,7 +210,8 @@ let package = Package(
 
     .target(
       name: "SwiftSyntax",
-      dependencies: ["_SwiftSyntaxCShims", "SwiftSyntax509", "SwiftSyntax510", "SwiftSyntax600"],
+      dependencies: ["_SwiftSyntaxCShims", "SwiftSyntax509", "SwiftSyntax510", "SwiftSyntax600",
+      "SwiftSyntax601"],
       exclude: ["CMakeLists.txt"],
       swiftSettings: swiftSyntaxSwiftSettings
     ),
--- a/Package.swift
+++ b/Package.swift
@@ -210,8 +210,7 @@ let package = Package(
 
     .target(
       name: "SwiftSyntax",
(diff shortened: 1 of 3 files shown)


bnbarham:
@swift-ci please test

bnbarham:
Thanks @mbrandonw

> If this is the correct fix then hopefully we can get a patch release of SwiftSyntax released soon?

Yes, would you be able to cherry-pick this to release/6.1 🙏?

bnbarham:
Also CC @keith for the bazel changes, but they look fine to me.

bnbarham (approved):

mbrandonw:
> Yes, would you be able to cherry-pick this to release/6.1 🙏?

Yep can do. Just so that I understand, you want me to make a new PR with these changes that targets release/6.1?

keith:
Yep bazel changes look good! Thanks for making them!

bnbarham:
> Yep can do. Just so that I understand, you want me to make a new PR with these changes that targets release/6.1?

Yep 🙇‍♂️!

mbrandonw:
@bnbarham Done here #3036.

Package.swift
```diff
@@ -210,7 +210,7 @@ let package = Package(
 
     .target(
       name: "SwiftSyntax",
-      dependencies: ["_SwiftSyntaxCShims", "SwiftSyntax509", "SwiftSyntax510", "SwiftSyntax600"],
+      dependencies: ["_SwiftSyntaxCShims", "SwiftSyntax509", "SwiftSyntax510", "SwiftSyntax600", "SwiftSyntax601"],
```
bnbarham, line 213: Oh while you're here, want to add SwiftSyntax602 as well (just on main)? Otherwise I'll wait until this is merged and put up a similar patch for 602.

mbrandonw: Ah, I didn't realize SwiftSyntax602 was already in place. Just updated the PR to add references to it in all the appropriate places.

bnbarham: Thanks! Just need to split the array onto a separate line now 😅

mbrandonw: Hi @bnbarham, I don't know if y'all are open to this kind of feedback, but fixing minor formatting issues can be a little annoying for PRs. I moved the dependencies to a separate line only to later find out a trailing comma is required. The test failure message can be difficult to parse, and I don't get emails for CI failures. So I rely on you and other repo owners to notify me of problems, and that seems like a bottleneck.

Now that this repo is on GitHub actions, would there be the possibility of auto formatting the entire repo on every commit to main? [Here](https://github.com/pointfreeco/swift-dependencies/commit/a9c9b583c15468335da2f47a5dea3ac4c08f425e) is an example from one of our repos. It's nice to not have to ask contributors to format their code since it happens automatically once PRs are merged.

bnbarham: Thanks for the feedback :)! It's an interesting idea, I'll pass it around

bnbarham:

bnbarham:

bnbarham:
Thanks again @mbrandonw!

PR #3037, opened 2025-04-02
Introduce `try` and `await` macro lexical contexts with unfolded sequence handling

hamishknight:
This is #2724 with additional logic for handling `try` and `await` in unfolded sequence expressions. We can say that any `try`/`await` element also covers all elements to the right of it (matching the logic in swiftlang/swift#80461). Cases where this isn't true will be rejected by the compiler, e.g:

```swift
0 * try foo() + bar()
_ = try foo() ~~~ bar() // Assuming `~~~` has lower precedence than `=`
```

rdar://109470248

### DIFF
--- a/Sources/SwiftSyntaxMacros/Syntax+LexicalContext.swift
+++ b/Sources/SwiftSyntaxMacros/Syntax+LexicalContext.swift
@@ -106,6 +106,39 @@ extension SyntaxProtocol {
       if let parentContext = parentNode.asMacroLexicalContext() {
         parentContexts.append(parentContext)
       }
+      if let sequence = parentNode.as(SequenceExprSyntax.self) {
+        var sequenceExprContexts: [Syntax] = []
+        for elt in sequence.elements {
+          if elt.range.contains(self.position) {
+            // `sequenceExprContexts` is built from the top-down, but we
+            // build the rest of the contexts bottom-up. Reverse for
+            // consistency.
+            parentContexts += sequenceExprContexts.reversed()
+            break
+          }
+          var elt = elt
+          while true {
+            if let tryElt = elt.as(TryExprSyntax.self) {
+              sequenceExprContexts.append(tryElt.asMacroLexicalContext()!)
+              elt = tryElt.expression
+              continue
+            }
+            if let awaitElt = elt.as(AwaitExprSyntax.self) {
+              sequenceExprContexts.append(awaitElt.asMacroLexicalContext()!)
+              elt = awaitElt.expression
+              continue
+            }
+            if let unsafeElt = elt.as(UnsafeExprSyntax.self) {
+              // No scope for this currently, but we need to look through it
+              // since it's similar to 'try' in that it's hoisted above a
+              // binary operator when appearing on the LHS.
+              elt = unsafeElt.expression
+              continue
+            }
+            break
+          }
+        }
+      }
 
       currentNode = parentNode
     }
--- a/Sources/SwiftSyntaxMacros/Syntax+LexicalContext.swift
+++ b/Sources/SwiftSyntaxMacros/Syntax+LexicalContext.swift
@@ -67,11 +67,11 @@ extension SyntaxProtocol {
     case let freestandingMacro as FreestandingMacroExpansionSyntax:
       return Syntax(freestandingMacro.detached) as Syntax
 
-    // Try and await are preserved: A freestanding expression macro preceded
-    // by try or await may need to know whether those keywords are present so it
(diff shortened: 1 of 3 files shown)


hamishknight:
@swift-ci please test

Sources/SwiftSyntaxMacros/Syntax+LexicalContext.swift
```diff
@@ -92,6 +106,39 @@ extension SyntaxProtocol {
...
+              sequenceExprContexts.append(tryElt.asMacroLexicalContext()!)
+              elt = tryElt.expression
+              continue
+            }
+            if let awaitElt = elt.as(AwaitExprSyntax.self) {
+              sequenceExprContexts.append(awaitElt.asMacroLexicalContext()!)
+              elt = awaitElt.expression
+              continue
+            }
+            if let unsafeElt = elt.as(UnsafeExprSyntax.self) {
```
DougGregor, line 131: Not for this PR, but this makes me think we need a protocol for the "effect-like" expression nodes. I feel like this pattern is going to come up a bit.

Sources/SwiftSyntaxMacros/Syntax+LexicalContext.swift
```diff
@@ -92,6 +106,39 @@ extension SyntaxProtocol {
...
+              // No scope for this currently, but we need to look through it
+              // since it's similar to 'try' in that it's hoisted above a
+              // binary operator when appearing on the LHS.
+              elt = unsafeElt.expression
```
DougGregor, line 135: I would like us to include `unsafe` expressions in the lexical context as well.

DougGregor (approved):
I'd like to include `unsafe` expressions in the lexical context, but otherwise LGTM!

hamishknight:
@swift-ci please test

Sources/SwiftSyntaxMacros/Syntax+LexicalContext.swift
hamishknight: Updated to include `unsafe`

Sources/SwiftSyntaxMacros/Syntax+LexicalContext.swift
hamishknight: Filed #3040

hamishknight:
@swift-ci please test Windows

Sources/SwiftSyntaxMacros/Syntax+LexicalContext.swift
grynspan: I requested that a while back, actually. :)

grynspan: #2549

Sources/SwiftSyntaxMacros/Syntax+LexicalContext.swift
```diff
@@ -92,6 +110,43 @@ extension SyntaxProtocol {
...
+      // be diagnosed by the compiler. This matches the compiler's ASTScope
+      // handling logic.
+      if let sequence = parentNode.as(SequenceExprSyntax.self) {
+        var sequenceExprContexts: [Syntax] = []
+        for elt in sequence.elements {
+          if elt.range.contains(self.position) {
+            // `sequenceExprContexts` is built from the top-down, but we
+            // build the rest of the contexts bottom-up. Reverse for
+            // consistency.
+            parentContexts += sequenceExprContexts.reversed()
```
ahoppen, line 126: Wouldn’t it make more sense to move this after the `for` loop? Just to make sure that we add elements to `parentContext` in case we should terminate the loop for some other reason than the `break` below.

hamishknight: Are there cases where we might not be able to determine which element the node is within? The main goal here is to only add the effects for nodes that occur after them in the sequence

ahoppen: No, I just thought that it would read nicer to have 

```swift
            // `sequenceExprContexts` is built from the top-down, but we
            // build the rest of the contexts bottom-up. Reverse for
            // consistency.
            parentContexts += sequenceExprContexts.reversed()
```

after the `for` loop and only having a `break` in here because the addition of elements to `parentContexts` conceptually happens after iterating over the elements.

hamishknight: Ah okay, I don't really feel all that strongly about it, happy to change

PR #3038, opened 2025-04-02
[Diagnostics] Make the node and position of Diagnostic and Note optional

DougGregor:
Some tools that wish to make use of the diagnostics machinery don't have source locations, or even a Swift source file to point into. For example, the compiler and driver both have diagnostics that don't involve anything in source code, e.g., because they are diagnosing problems at the module level such as conflicting or incorrect compiler options. At present, the driver uses a different scheme for emitting diagnostics, and the compiler falls back to using the LLVM diagnostics machinery when there is no source location information.

Making the "node" and "position" of a Diagnostic optional make it possible to express diagnostics that aren't associated with sources. Do so, and update the diagnostic formatter to support displaying diagnostics with no source location using the same `<unknown>:0:0:` scheme the compiler users.

### DIFF
--- a/Sources/SwiftDiagnostics/Diagnostic.swift
+++ b/Sources/SwiftDiagnostics/Diagnostic.swift
@@ -21,11 +21,11 @@ public struct Diagnostic: CustomDebugStringConvertible, Sendable {
   public let diagMessage: DiagnosticMessage
 
   /// The node at whose start location the message should be displayed.
-  public let node: Syntax
+  public let node: Syntax?
 
   /// The position at which the location should be anchored.
   /// By default, this is the start location of `node`.
-  public let position: AbsolutePosition
+  public let position: AbsolutePosition?
 
   /// Nodes that should be highlighted in the source code.
   public let highlights: [Syntax]
@@ -40,17 +40,17 @@ public struct Diagnostic: CustomDebugStringConvertible, Sendable {
   /// If `highlights` is `nil` then `node` will be highlighted. This is a
   /// reasonable default for almost all diagnostics.
   public init(
-    node: some SyntaxProtocol,
+    node: (some SyntaxProtocol)?,
     position: AbsolutePosition? = nil,
     message: DiagnosticMessage,
     highlights: [Syntax]? = nil,
     notes: [Note] = [],
     fixIts: [FixIt] = []
   ) {
-    self.node = Syntax(node)
-    self.position = position ?? node.positionAfterSkippingLeadingTrivia
+    self.node = node.map { Syntax($0) }
+    self.position = position ?? node?.positionAfterSkippingLeadingTrivia
     self.diagMessage = message
-    self.highlights = highlights ?? [Syntax(node)]
+    self.highlights = highlights ?? node.map { [Syntax($0)] } ?? []
     self.notes = notes
     self.fixIts = fixIts
   }
@@ -67,13 +67,24 @@ public struct Diagnostic: CustomDebugStringConvertible, Sendable {
   }
 
   /// The location at which the diagnostic should be displayed.
-  public func location(converter: SourceLocationConverter) -> SourceLocation {
+  public func location(converter: SourceLocationConverter) -> SourceLocation? {
+    guard let position else {
+      return nil
+    }
+
     return converter.location(for: position)
   }
 
   public var debugDescription: String {
+    guard let node else {
+      return "\(DiagnosticsFormatter.unknownFileName):0:0: \(message)"
+    }
+
     let locationConverter = SourceLocationConverter(fileName: "", tree: node.root)
-    let location = location(converter: locationConverter)
+    guard let location = location(converter: locationConverter) else {
+      return "\(DiagnosticsFormatter.unknownFileName):0:0: \(message)"
+    }
+
     return "\(location.line):\(location.column): \(message)"
   }
 }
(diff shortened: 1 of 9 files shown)


DougGregor:
https://github.com/swiftlang/swift/pull/80477

@swift-ci please test

DougGregor:
SwiftFormat breaks as well, which shouldn't be a surprise

Sources/SwiftDiagnostics/Diagnostic.swift
```diff
@@ -40,17 +40,17 @@ public struct Diagnostic: CustomDebugStringConvertible, Sendable {
...
+    node: (some SyntaxProtocol)?,
     position: AbsolutePosition? = nil,
     message: DiagnosticMessage,
     highlights: [Syntax]? = nil,
     notes: [Note] = [],
     fixIts: [FixIt] = []
   ) {
-    self.node = Syntax(node)
-    self.position = position ?? node.positionAfterSkippingLeadingTrivia
+    self.node = node.map { Syntax($0) }
```
ahoppen, line 50: I think we have a `Syntax` initializer that just proposes `nil`, so it should be possible to leave this as-is.

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
```diff
@@ -299,12 +307,14 @@ public struct DiagnosticsFormatter {
       }
 
       let columnsWithDiagnostics = Set(
-        annotatedLine.diagnostics.map {
-          annotatedLine.characterColumn(ofUtf8Column: $0.location(converter: slc).column)
+        annotatedLine.diagnostics.compactMap {
+          $0.location(converter: slc).map {
+            annotatedLine.characterColumn(ofUtf8Column: $0.column)
+          }
```
ahoppen, line 313: Maybe it’s just me but I find `Optional.map` always hard to read and would prefer.

```suggestion
          if guard let location = $0.location(converter: slc) else {
            return nil
          }
          return annotatedLine.characterColumn(ofUtf8Column: $0.column)
```

Sources/SwiftDiagnostics/GroupedDiagnostics.swift
```diff
@@ -275,9 +293,25 @@ extension GroupedDiagnostics {
...
+      let annotated = group.annotateSource(
+        rootSourceFileID,
+        formatter: self,
+        indentString: ""
+      )
+      return annotated.isEmpty ? nil : annotated
+    }
+
+    // Render any floating diagnostics, which have no anchor.
+    renderedDiags.append(
```
ahoppen, line 307: Opinionated question: Should we render the floating diagnostics before or after the diagnostics with source locations? I would have rendered them before because I visualize them on the first line.

Also, using `+=` instead of `append(contentsOf:)` would remove a nesting level here.

Tests/SwiftDiagnosticsTest/GroupDiagnosticsFormatterTests.swift
```diff
@@ -85,6 +85,43 @@ final class GroupedDiagnosticsFormatterTests: XCTestCase {
...
+          )
+        )
+      )
+    )
+
+    let annotated = DiagnosticsFormatter.annotateSources(in: group)
+    assertStringsEqualWithDiff(
+      annotated,
+      """
+      <unknown>:0:0: error: this is bad [#Badness]
```
ahoppen, line 120: Can we suppress the rendering of  `0:0` here? It doesn’t add any value.

Tests/SwiftParserTest/Assertions.swift
```diff
@@ -324,7 +324,7 @@ func assertNote(
   XCTAssertEqual(note.message, spec.message, file: spec.file, line: spec.line)
   let locationConverter = SourceLocationConverter(fileName: "", tree: tree)
   assertLocation(
-    note.location(converter: locationConverter),
+    note.location(converter: locationConverter)!,
```
ahoppen, line 327: We should make sure that the test doesn’t crash if a note without a source location should be generated. Instead, we should raise a test failure.

Sources/SwiftDiagnostics/GroupedDiagnostics.swift
```diff
@@ -57,6 +57,10 @@ public struct GroupedDiagnostics {
   /// source file IDs.
   var rootIndexes: [Syntax: SourceFileID] = [:]
 
+  /// Diagnostics that aren't associated with any source file because they have
+  /// no location information.
+  var floatingDiagnostics: [Diagnostic] = []
```
ahoppen, line 62: Instead of introducing a new term _floating_ here, how about calling this something like `unknownPositionDiagnostics`?

ahoppen (commented):
I am a little worried about the source breakage this might cause and also about ambiguities about diagnostics that have `node == nil && position != nil` or `node != nil && position == nil` – I don’t know what that should signify and the diagnostic formatter also seems to be inconsistent about this. 

Maybe we need to introduce another type like `DiagnosticLocation` to wrap `node` and `position`. That would also have the advantage that we could keep `node` and `position` as deprecated non-optional compatibility accessors.

Something like

```swift
enum DiagnosticLocation {
  /// The diagnostic is anchored at the start of the given syntax node.
  case startOfNode(Syntax)
  /// The diagnostic is anchored at the given position within the syntax tree that contains the given node.
  case position(AbsolutePosition, inTreeOf: Syntax)
  /// The diagnostic does not have a source location associated with it.
  case unknown

  /// The node at whose start location the message should be displayed.
  public var node: Syntax? { … }

  /// The position at which the location should be anchored.
  public var position: AbsolutePosition? { … }
}
```

DougGregor:
I decided not to pursue this.

PR #3039, opened 2025-04-03
[Windows] Remove extra newlines from diagnostic output.

al45tair:
We are outputting lots of extra newlines on Windows, which makes it hard to read the diagnostics.

rdar://148520063

### DIFF
--- a/Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
+++ b/Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
@@ -293,10 +293,13 @@ public struct DiagnosticsFormatter {
         )
       )
 
-      // If the line did not end with \n (e.g. the last line), append it manually
-      if annotatedSource.last != "\n" {
-        annotatedSource.append("\n")
+      // Remove any trailing newline and replace it; this may seem
+      // counterintuitive, but if we're running within CMake and we let a
+      // '\r\n' through, CMake will turn that into *two* newlines.
+      if let last = annotatedSource.last, last.isNewline {
+        annotatedSource.removeLast()
       }
+      annotatedSource.append("\n")
 
       let columnsWithDiagnostics = Set(
         annotatedLine.diagnostics.map {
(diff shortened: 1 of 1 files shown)


al45tair:
@swift-ci Please test

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
```diff
@@ -293,10 +310,7 @@ public struct DiagnosticsFormatter {
         )
       )
 
-      // If the line did not end with \n (e.g. the last line), append it manually
```
bnbarham, line 296: Stripping is maybe fine, but it does seem a little overkill. Can't we just check `annotatedLine` instead?

bnbarham (commented):
Nice catch @al45tair!

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
al45tair: That's what I tried initially, but it didn't completely eliminate the problem. Let me have another go at it and see if I can make that work.

al45tair: So… making it check the line ending properly didn't work before — it left an extra line ending still. Digging a bit further, it appears that the problem is that something upstream of this code is turning

```
... 0d 0a
```

into _two_ line endings. Currently looking to see if I can work out where that's happening (I'm guessing it's on the output side).

al45tair: It seems to be the `llvm::raw_ostream` that's causing this issue. Haven't found where the problem is exactly though; MS's C library doesn't exhibit the same problem with outputting strings containing `0d 0a` sequences instead of just `0a`.

al45tair: Stripping the `0d 0a` line endings and replacing them with `0a` fixes that problem, I might add.

al45tair: OK. I think I have a more efficient approach; we'll still always remove the trailing newline and add our own, but we'll have one less string copy.

al45tair:
@swift-ci Please test

al45tair:

Sources/SwiftDiagnostics/DiagnosticsFormatter.swift
al45tair: The problem wasn't `colorizeSourceLine`, thankfully; it was a combination of the Swift Syntax code not looking for `\r\n` (it should have been using `.last.isNewline`) and CMake messing up if you send `\r\n` through it (it doubles that up to `\r\n\r\n`).

al45tair: Anyway, the fix works now, and it doesn't forcibly strip every time.

PR #3043, opened 2025-04-05
Removes duplicate trivia from UnsafeExprSyntax

TTOzzi:
When creating `UnsafeExprSyntax`, the `leadingTrivia` passed as an argument from `unsafeExpr.leadingTrivia` is already included in `unsafeExpr.unsafeKeyword`, which causes the trivia to be duplicated if it exists. To fix this, I modified the code to avoid passing `leadingTrivia` as an argument. (Similarly, the same applies to `trailingTrivia`.)

I couldn’t find a suitable way to directly test `UnsafeExprSyntax` created via `OperatorTable`, so I verified the change by integrating my local modifications into `swift-format`.
Please let me know if there’s a better way to test this.

### DIFF
--- a/Sources/SwiftOperators/OperatorTable+Folding.swift
+++ b/Sources/SwiftOperators/OperatorTable+Folding.swift
@@ -141,7 +141,6 @@ extension OperatorTable {
     if let unsafeExpr = lhs.as(UnsafeExprSyntax.self) {
       return ExprSyntax(
         UnsafeExprSyntax(
-          leadingTrivia: unsafeExpr.leadingTrivia,
           unsafeExpr.unexpectedBeforeUnsafeKeyword,
           unsafeKeyword: unsafeExpr.unsafeKeyword,
           unsafeExpr.unexpectedBetweenUnsafeKeywordAndExpression,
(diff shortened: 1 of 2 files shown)


ahoppen (commented):
Thanks for the fix @TTOzzi.

You should be able to test this using

```swift
func testTriviaAroundUnsafe() throws {
  let original = ExprSyntax("/*leading*/ unsafe a + b")
  let folded = try OperatorTable.standardOperators.foldAll(original)
  XCTAssertEqual(original.description, folded.description)
}
```

TTOzzi:
> Thanks for the fix @TTOzzi.
> 
> You should be able to test this using
> 
> ```swift
> func testTriviaAroundUnsafe() throws {
>   let original = ExprSyntax("/*leading*/ unsafe a + b")
>   let folded = try OperatorTable.standardOperators.foldAll(original)
>   XCTAssertEqual(original.description, folded.description)
> }
> ```

Thank you for letting me know! 🙇 
I've added test cases for the changes.

ahoppen (approved):
Thanks for the fix!

PR #3047, opened 2025-04-07
 [SwiftParser] Implement `nonisolated(nonsending)` specifier

xedin:
This is part of SE-0461 proposal where `nonisolated` has to be
marked as `nonsending` in type context to indicate that the
function type it's attached to is caller isolated.


xedin:
https://github.com/swiftlang/swift/pull/80668
@swift-ci please test

Sources/SwiftParser/Types.swift
```diff
@@ -691,11 +691,27 @@ extension Parser.Lookahead {
...
       self.hasProgressed(&specifierProgress)
     {
-      self.consumeAnyToken()
+      switch self.currentToken {
+      case .keyword(.nonisolated), .keyword(.dependsOn):
+        self.consumeAnyToken()
+
+        // The argument is missing but it still could be a valid modifier,
+        // i.e. `nonisolated` in an inheritance clause.
+        guard self.at(.leftParen) else {
```
ahoppen, line 704: Should we check that the left parenthesis on the same line as `nonisolated` or `dependsOn`? Eg. to make sure that we parse the following as two statements

```swift
let x: nonisolated
(blah)
```

Sources/SwiftParser/Types.swift
```diff
@@ -1056,6 +1072,88 @@ extension Parser {
...
 
+  private mutating func parseNonisolatedTypeSpecifier() -> RawTypeSpecifierListSyntax.Element {
+    let (unexpectedBeforeNonisolatedKeyword, nonisolatedKeyword) = self.expect(.keyword(.nonisolated))
+
+    // If the next token is not '(' this could mean two things:
+    //  - What follows is a type and we should allow it because
+    //    using `nonsisolated` without an argument is allowed in
+    //    an inheritance clause.
+    //  - The '(nonsending)' was omitted.
+    if !self.at(.leftParen) {
```
ahoppen, line 1083: Similar here, should we check that `(` is on the same line?

Sources/SwiftParser/Types.swift
```diff
@@ -1056,6 +1072,88 @@ extension Parser {
...
+        arena: self.arena
+      )
+
+      return .nonisolatedTypeSpecifier(nonisolatedSpecifier)
+    }
+
+    // Avoid being to greedy about `(` since this modifier should be associated with
+    // function types, it's possible that the argument is omitted and what follows
+    // is a function type i.e. `nonisolated () async -> Void`.
+    if self.at(.leftParen) && !withLookahead({ $0.atAttributeOrSpecifierArgument() }) {
```
ahoppen, line 1116: Last place where we might want to check that `(` is on the same line.

Tests/SwiftParserTest/DeclarationTests.swift
```diff
@@ -271,7 +271,36 @@ final class DeclarationTests: ParserTestCase {
...
+      struct A {
+        nonisolated(nonsending) let b = 0
+        nonisolated(nonsending) var c: Int { 0 }
+        nonisolated(1️⃣sending) let d = 0
+      }
+      """,
+      diagnostics: [
+        DiagnosticSpec(
+          message: "expected identifier in modifier",
+          fixIts: ["replace 'sending' with identifier"]
```
ahoppen, line 294: Hmm, this error message isn’t great. I would prefer if we just accept any identifier here in the parser and let ASTGen diagnose unknown identifiers as the detail.

ahoppen (commented):
Looks good to me, just a few comments inline.

Sources/SwiftParser/Types.swift
```diff
@@ -691,11 +691,27 @@ extension Parser.Lookahead {
     var specifierProgress = LoopProgressCondition()
     // TODO: Can we model isolated/_const so that they're specified in both canParse* and parse*?
     while canHaveParameterSpecifier,
-      self.at(anyIn: SimpleTypeSpecifierSyntax.SpecifierOptions.self) != nil || self.at(.keyword(.isolated))
-        || self.at(.keyword(._const)),
```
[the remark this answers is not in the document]
xedin: `_const` and `isolated` are listed in `SimpleTypeSpecifierSyntax.SpecifierOptions`, that's why listing them here seems like no-op?

Tests/SwiftParserTest/DeclarationTests.swift
xedin: Yeah, that's why I was asking how to make it so it lists both `unsafe` and `nonsending` now, `self.expect` accepts multiple choices but only one "default".

Sources/SwiftParser/Types.swift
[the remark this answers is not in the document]
ahoppen: Ah, I see. Thanks for the explanation.

Sources/SwiftParser/Types.swift
```diff
@@ -689,13 +689,28 @@ extension Parser.Lookahead {
...
       self.hasProgressed(&specifierProgress)
     {
-      self.consumeAnyToken()
+      switch self.currentToken {
+      case .keyword(.nonisolated), .keyword(.dependsOn):
+        self.consumeAnyToken()
+
+        // The argument is missing but it still could be a valid modifier,
+        // i.e. `nonisolated` in an inheritance clause.
+        guard self.at(.leftParen), !self.currentToken.isAtStartOfLine else {
```
ahoppen, line 703: This can be unified slightly to 

```suggestion
        guard self.at(TokenSpec(.leftParen, allowAtStartOfLine: false)) else {
```

Sources/SwiftParser/Types.swift
```diff
@@ -1079,6 +1176,13 @@ extension Parser {
         } else {
           break SPECIFIER_PARSING
         }
+      } else if self.at(.keyword(.nonisolated)) {
+        // If '(' is located on the new line 'nonisolated' cannot be parsed
+        // as a specifier.
+        if self.peek(isAt: .leftParen) && self.peek().isAtStartOfLine {
```
ahoppen, line 1182:
```suggestion
        if self.peek(isAt: TokenSpec(.leftParen, allowAtStartOfLine: false)) {
```

Sources/SwiftParser/Types.swift
xedin: Nice, thank you!

Sources/SwiftParser/Types.swift
xedin: This is actually not equivalent to what is written, we'd like to stop only if the next token is `(` on the new line, it's not the same condition as above.

ahoppen: Sorry, please ignore my comment

Sources/SwiftParser/Types.swift
```diff
@@ -691,11 +691,27 @@ extension Parser.Lookahead {
     var specifierProgress = LoopProgressCondition()
     // TODO: Can we model isolated/_const so that they're specified in both canParse* and parse*?
```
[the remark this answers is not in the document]
xedin: ✅

Sources/SwiftParser/Types.swift
xedin: No worries!

Sources/SwiftParser/Types.swift
xedin: ✅

Tests/SwiftParserTest/DeclarationTests.swift
xedin: Something to consider here before we make a final decision: if we did allow any and all identifiers here, it would mean that we'd always parse nonisolated as a declaration modifier, which is a break from what C++ parser is doing and is not always correct. 

Consider this situation:

```
func nonisolated() {
}

let v: Int = 42

nonisolated(v)
func test() -> Int { 42 }
```

This is currently a call + a decl but it would be interpreted as a a decl with a modifier with this change.

ahoppen: OK, let’s just stick with the current solution. It’s not the best but maybe good enough for now. And I don’t want to block this PR on improving this.

Sources/SwiftParser/Types.swift
xedin: ✅

Sources/SwiftParser/Types.swift
xedin: ✅

Sources/SwiftParser/Types.swift
xedin: ✅

Tests/SwiftParserTest/DeclarationTests.swift
xedin: Sounds good!

xedin:
https://github.com/swiftlang/swift/pull/80668
@swift-ci please test

ahoppen (approved):

xedin:
https://github.com/swiftlang/swift/pull/80668

Sources/SwiftParser/Expressions.swift
```diff
@@ -381,13 +381,37 @@ extension Parser {
     }
   }
 
+  /// Make sure that we only accept `nonisolated(nonsending)` as a valid type specifier
+  /// in expression context to minimize source compatibility impact.
+  func canParseNonisolatedAsSpecifierInExpressionContext() -> Bool {
+    return withLookahead {
+      $0.consumeAnyToken()
```
ahoppen, line 388: Should we just double  check that we are indeed on a `nonisolated` here? Just for another line of defense.

Ie. 

```swift
      guard $0.consume(if: .keyword(.nonisolated)) != nil else {
        return false
      }
```

Sources/SwiftParser/Types.swift
```diff
@@ -687,27 +687,74 @@ extension Parser.Lookahead {
...
+            // keyword to be a modifier when it isn't i.e. `[nonisolated(42)]`
+            guard $0.consume(if: TokenSpec(.nonsending, allowAtStartOfLine: false)) != nil else {
+              return false
+            }
+
+            return $0.consume(if: TokenSpec(.rightParen, allowAtStartOfLine: false)) != nil
+          })
+        else {
+          return false
+        }
```
ahoppen, line 724: I’m not a huge fan of >10 LOC conditionals and would pull this out


```swift
          let canParseNonisolated = self.withLookahead {
            // Consume 'nonisolated'
            $0.consumeAnyToken()

            // The argument is missing but it still could be a valid modifier,
            // i.e. `nonisolated` in an inheritance clause.
            guard $0.at(TokenSpec(.leftParen, allowAtStartOfLine: false)) else {
              return true
            }

            // Consume '('
            $0.consumeAnyToken()

            // nonisolated accepts a single modifier at the moment: 'nonsending'
            // we need to check for that explicitly to avoid misinterpreting this
            // keyword to be a modifier when it isn't i.e. `[nonisolated(42)]`
            guard $0.consume(if: TokenSpec(.nonsending, allowAtStartOfLine: false)) != nil else {
              return false
            }

            return $0.consume(if: TokenSpec(.rightParen, allowAtStartOfLine: false)) != nil
          }
        guard canParseNonisolated else {
          return false
        }
```

Same for `dependsOn`

Sources/SwiftParser/Types.swift
```diff
@@ -687,27 +687,74 @@ extension Parser.Lookahead {
     return true
   }
 
-  mutating func skipTypeAttributeList() {
+  mutating func skipTypeAttributeList() -> Bool {
```
ahoppen, line 690: Could you add a short comment to indicate what the bool return value means?

bnbarham: I think we can probably just rename this to `canParseTypeAttributeList` which would then make it obvious?

Sources/SwiftParser/Types.swift
xedin: Sure, done!

Sources/SwiftParser/Types.swift
xedin: Yeah, I like `canParseTypeAttributeList`, renamed!

Sources/SwiftParser/Expressions.swift
xedin: ✅

xedin:
https://github.com/swiftlang/swift/pull/80668

xedin:
https://github.com/swiftlang/swift/pull/80668
@swift-ci please test

xedin:
https://github.com/swiftlang/swift/pull/80668

PR #3055, opened 2025-04-15
[6.2][SwiftParser] Implement `nonisolated(nonsending)` specifier

xedin:
- Explanation:

  This is part of SE-0461 proposal where `nonisolated` has to be
  marked as `nonsending` in type context to indicate that the
  function type it's attached to is caller isolated.
 
- Main Branch PR: https://github.com/swiftlang/swift-syntax/pull/3047

- Risk: Low (The only risk here is related to `nonisolated` keyword expansion to accept `nonsending` modifier, this could cause a source break if there exists a function which such a name an an argument with name `nonsending`, which seems like a very low probability).

- Reviewed By: @hamishknight @ahoppen @bnbarham 

- Testing: Added new tests to the test suite.


xedin:
https://github.com/swiftlang/swift/pull/80807
@swift-ci please test

bnbarham (approved):

PR #3059, opened 2025-04-17
Add a new API DiagnosticsFormatter.formattedMessage(_:) for messages without source location info

DougGregor:
Some tools want to use the diagnostics formatter to produce diagnostics that don't relate to source code, or for which the source code isn't available. This API allows them to do so while maintaining consistent presentation. The first client is the compiler itself, when it doesn't have a source location for a specific diagnostic.

### DIFF
--- a/Tests/SwiftDiagnosticsTest/DiagnosticFormatterTests.swift
+++ b/Tests/SwiftDiagnosticsTest/DiagnosticFormatterTests.swift
@@ -1,0 +1,31 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2025 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftDiagnostics
+import XCTest
+
+final class DiagnosticFormatterTests: XCTestCase {
+  func testFormattedMessage() {
+    let message = SimpleDiagnosticMessage(
+      message: "something went wrong",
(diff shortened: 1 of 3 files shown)


DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test macOS

DougGregor:
@swift-ci please test macOS

Tests/SwiftDiagnosticsTest/DiagnosticFormatterTests.swift
```diff
@@ -0,0 +1,31 @@
...
+      diagnosticID: MessageID(domain: "swift-syntax", id: "testing"),
+      severity: .error,
+      category: DiagnosticCategory(
+        name: "Testing",
+        documentationURL: "http://example.com"
+      )
+    )
+
+    let formattedText = DiagnosticsFormatter().formattedMessage(message)
+    XCTAssertEqual(formattedText, "error: something went wrong [#Testing]")
```
ahoppen, line 29: Is the idea that the caller of this function will be responsible for emitting the link between `[#Testing]` and `http://example.com`?

ahoppen (approved):

Tests/SwiftDiagnosticsTest/DiagnosticFormatterTests.swift
DougGregor: If your terminal supports it, that `[#Testing]` will be a hyperlink. There's also existing API for emitting the footnote style for a set of categoryies on DiagnosticsFormatter:

```swift
public func categoryFootnotes(
    _ categories: some Sequence<DiagnosticCategory>,
    leadingText: String = "\n"
  ) -> String
```

because you don't want to emit the footnotes after every diagnostic, you want to do it once at the end.

ahoppen: Oh, great. I didn’t realize that we had an API for the footnotes. 👍🏽

DougGregor:
@swift-ci please test macOS

PR #3061, opened 2025-04-19
Fix CaseDetectionMacro to handle multiple elements and improve casing

Stof83:
This patch fixes a bug in CaseDetectionMacro where only the first case element was considered. It now handles multiple case elements and uses proper PascalCase formatting for isX computed properties.

### DIFF
--- a/Examples/Sources/MacroExamples/Implementation/Member/CaseDetectionMacro.swift
+++ b/Examples/Sources/MacroExamples/Implementation/Member/CaseDetectionMacro.swift
@@ -22,16 +22,14 @@ public enum CaseDetectionMacro: MemberMacro {
   ) throws -> [DeclSyntax] {
     declaration.memberBlock.members
       .compactMap { $0.decl.as(EnumCaseDeclSyntax.self) }
-      .map { $0.elements.first!.name }
-      .map { ($0, $0.initialUppercased) }
-      .map { original, uppercased in
+      .map { $0.elements }
+      .flatMap { $0 }
+      .map { $0.name }
+      .map { ($0, $0.pascalcased) }
+      .map { original, pascalcased in
         """
-        var is\(raw: uppercased): Bool {
-          if case .\(raw: original) = self {
-            return true
-          }
-
-          return false
+        var is\(raw: pascalcased): Bool {
+            self == .\(raw: original)
         }
         """
       }
@@ -39,12 +37,13 @@ public enum CaseDetectionMacro: MemberMacro {
 }
 
 extension TokenSyntax {
-  fileprivate var initialUppercased: String {
-    let name = self.text
-    guard let initial = name.first else {
-      return name
+    fileprivate var pascalcased: String {
+        self.text
+            .replacingOccurrences(of: "_", with: " ")
+            .replacingOccurrences(of: "(?<=.)([A-Z])", with: " $1", options: .regularExpression)
+            .lowercased()
+            .split(separator: " ")
+            .map { $0.capitalized }
+            .joined()
     }
-
-    return "\(initial.uppercased())\(name.dropFirst())"
-  }
 }


Examples/Sources/MacroExamples/Implementation/Member/CaseDetectionMacro.swift
```diff
@@ -22,29 +22,28 @@ public enum CaseDetectionMacro: MemberMacro {
...
       }
   }
 }
 
 extension TokenSyntax {
-  fileprivate var initialUppercased: String {
-    let name = self.text
-    guard let initial = name.first else {
-      return name
+    fileprivate var pascalcased: String {
```
ahoppen, line 40: It looks like *pascal case* always starts with an uppercase letter. Also, from my experience, the term *camel case* is more common, so I would call this property `asCamelCase`.

Examples/Sources/MacroExamples/Implementation/Member/CaseDetectionMacro.swift
```diff
@@ -22,29 +22,28 @@ public enum CaseDetectionMacro: MemberMacro {
   ) throws -> [DeclSyntax] {
     declaration.memberBlock.members
       .compactMap { $0.decl.as(EnumCaseDeclSyntax.self) }
-      .map { $0.elements.first!.name }
-      .map { ($0, $0.initialUppercased) }
-      .map { original, uppercased in
+      .map { $0.elements }
+      .flatMap { $0 }
```
ahoppen, line 26: Can this just be the following?

```suggestion
      .flatMap { $0.elements }
```

Examples/Sources/MacroExamples/Implementation/Member/CaseDetectionMacro.swift
```diff
@@ -22,29 +22,28 @@ public enum CaseDetectionMacro: MemberMacro {
...
-        var is\(raw: uppercased): Bool {
-          if case .\(raw: original) = self {
-            return true
-          }
-
-          return false
+        var is\(raw: pascalcased): Bool {
+            self == .\(raw: original)
         }
         """
```
ahoppen, line 34: Would it be easier to just unify these three `map` calls into one? Also, I think we were using `if case` instead of `==` here so that this also works for enums with associated values.

```swift
.map { 
  """
  var is\(raw: $0.name.asCamelCase): Bool {
    if case .\(raw: $0.name) = self {
      return true
    }
    return false
  }
  """
}
```

Examples/Sources/MacroExamples/Implementation/Member/CaseDetectionMacro.swift
```diff
@@ -22,29 +22,28 @@ public enum CaseDetectionMacro: MemberMacro {
...
-  fileprivate var initialUppercased: String {
-    let name = self.text
-    guard let initial = name.first else {
-      return name
+    fileprivate var pascalcased: String {
+        self.text
+            .replacingOccurrences(of: "_", with: " ")
+            .replacingOccurrences(of: "(?<=.)([A-Z])", with: " $1", options: .regularExpression)
+            .lowercased()
+            .split(separator: " ")
```
ahoppen, line 45: Do we need to insert space as the separator character here? Couldn’t we use `split(whereSeparator:)` instead? Something like

```suggestion
            .split { $0 == "_" || $0.isUppercase }
            .map { $0.lowercased().capitalized }
```

ahoppen (commented):
Thanks for the improvement. I have left a couple suggestion inline.

Examples/Sources/MacroExamples/Implementation/Member/CaseDetectionMacro.swift
Stof83: Thanks for the feedback!

You're right that naming consistency matters, and camelCase is indeed a commonly used term. However, in this context, I believe PascalCase is actually the more accurate naming.

The transformation here turns identifiers like en_route, EN_ROUTE, or even enRoute into EnRoute—that is, with each word capitalized and no underscores. This matches the definition of PascalCase, where each word starts with an uppercase letter (including the first), as opposed to camelCase, which typically starts with a lowercase first word (e.g., enRoute).

For example:
```
en_route → EnRoute → isEnRoute
EN_ROUTE → EnRoute → isEnRoute
on_duty → OnDuty → isOnDuty
```
So naming it pascalcased better reflects what the transformation is actually doing.

Examples/Sources/MacroExamples/Implementation/Member/CaseDetectionMacro.swift
Stof83: Thanks for the suggestion! I think we could simplify this a bit further and make it more reliable for cases like enRoute or EN_ROUTE by using a regex to normalize everything into a consistent underscore-separated format first. That way, we avoid dealing with space insertion or manual splitting logic.

What do you think about something like this instead?

```
.replacingOccurrences(of: "([a-z])([A-Z])", with: "$1_$2", options: .regularExpression)
.split(separator: "_")
.map { $0.lowercased().capitalized }
```

This converts enRoute to EnRoute, EN_ROUTE to EnRoute, en_route to EnRoute, etc., without needing to match on uppercase characters manually or insert spaces. Let me know what you think!

Examples/Sources/MacroExamples/Implementation/Member/CaseDetectionMacro.swift
Stof83: Good point. Happy to update the name accordingly. Thanks!

Examples/Sources/MacroExamples/Implementation/Member/CaseDetectionMacro.swift
ahoppen: Actually thinking about this again, I would argue to stick with the old `initialUppercased` approach because cases that adhere to Swift’s standard naming conventions would be lowercase camel case anyway, so none of this matters. For anything else, `initialUppercased` still produces understandable results and doesn’t break cases like `v5_2`, which might be named like this to indicate a version `5.2`. Anything but `initialUppercased` would require us to implement a fully-fledged snake_case to camelCase conversion function with all the pitfalls and edge cases these come with and doing so would require quite a bit of thought that’s unrelated to the purpose of this example.

PR #3062, opened 2025-04-21
[SwiftParser] Parse @abi attribute arguments even without feature flag

rintaro:
When the parser see `@abi` attribute without the language feature enabled, not parsing the interior causes catastrophic breakage in the tree. To avoid that, parse the argument decl is parsed, but instead of generating `ABIAttributeArgumentsSyntax`, store it as a `UnexpectedNodesSyntax` after the left-paren.

### DIFF
--- a/Tests/SwiftParserTest/AttributeTests.swift
+++ b/Tests/SwiftParserTest/AttributeTests.swift
@@ -1283,6 +1283,49 @@ final class AttributeTests: ParserTestCase {
     )
   }
 
+  func testABIAttributeWithoutFeature() throws {
+    assertParse(
+      """
+      @abi(1️⃣func fn() -> Int2️⃣)
+      func fn1() -> Int { }
+      """,
+      substructure: FunctionDeclSyntax(
+        attributes: [
+          .attribute(
+            AttributeSyntax(
+              attributeName: TypeSyntax("abi"),
+              leftParen: .leftParenToken(),
+              [Syntax(try FunctionDeclSyntax("func fn() -> Int"))],
+              arguments: .argumentList([]),
+              rightParen: .rightParenToken()
+            )
+          )
+        ],
+        name: "fn1",
+        signature: FunctionSignatureSyntax(
+          parameterClause: FunctionParameterClauseSyntax {},
+          returnClause: ReturnClauseSyntax(type: TypeSyntax("Int"))
+        )
+      ) {},
+      diagnostics: [
+        DiagnosticSpec(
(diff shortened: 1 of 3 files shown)


rintaro:
@swift-ci Please test

xedin:
@rintaro I think if the feature is not enabled the whole attribute is not supported, is it possible to move UnexpectedNodesSyntax up to cover not just argument but the attribute node itself?

xedin:
I think the new diagnostic should be about `@abi` attribute itself instead of its argument...

rintaro:
Is there a specific reason that made you suggest that?

xedin:
I was just looking at the diagnostics you added and they don't seem super useful, it feels like we should either accept the presence of the attribute and validate it fully or skip it all together if the feature is not enabled. That's something we do in TypeCheckAttr by diagnosing the attribute and marking it as invalid when the feature is not there...

beccadax:
I'll have to tweak a `#if` in one of the macro tests in #79937, but once I do that, this fix should work fine.

Tests/SwiftParserTest/AttributeTests.swift
```diff
@@ -1283,6 +1283,49 @@ final class AttributeTests: ParserTestCase {
...
+      @abi(1️⃣func fn() -> Int2️⃣)
+      func fn1() -> Int { }
+      """,
+      substructure: FunctionDeclSyntax(
+        attributes: [
+          .attribute(
+            AttributeSyntax(
+              attributeName: TypeSyntax("abi"),
+              leftParen: .leftParenToken(),
+              [Syntax(try! FunctionDeclSyntax("func fn() -> Int"))],
```
ahoppen, line 1298: Could make the test function `throws` and then avoid the `try!` here.
 
```suggestion
              [Syntax(try FunctionDeclSyntax("func fn() -> Int"))],
```

ahoppen (commented):
I want to note that this is technically source breaking in case somebody has a custom attribute defined as `@abi`, eg. the following was correctly parsable using the new parser before this PR and now no longer is. But we broke this kind of code in the C++ parser already, so I don’t think it’s a big deal.

```swift
@globalActor
actor abi {
    static let shared: abi = abi()
}

@abi struct Foo {}
```

ahoppen (approved):

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

PR #3068, opened 2025-04-30
[SwiftLexicalLookup] Unqualified lookup caching

MAJKFL:
This PR introduces optional caching support to `SwiftLexicalLookup`. In order to use it, clients can pass an instance of `LookupCache` as a parameter to the lookup function. 

`LookupCache` keeps track of cache member hits. In order to prevent the cache from taking too much memory, clients can call the `LookupCache.evictEntriesWithoutHit` function to remove members without a hit and reset the hit property for the remaining members. Calling this function every time after lookup effectively maintains one path from a leaf to the root of the scope tree in cache.

Clients can also optionally set the `drop` value:
```swift
/// Creates a new unqualified lookup cache.
/// `drop` parameter specifies how many eviction calls will be
/// ignored before evicting not-hit members of the cache.
///
/// Example cache eviction sequences (s - skip, e - evict):
/// - `drop = 0` - `e -> e -> e -> e -> e -> ...`
/// - `drop = 1` - `s -> e -> s -> s -> e -> ...`
/// - `drop = 3` - `s -> s -> s -> e -> s -> ...`
///
/// - Note: `drop = 0` effectively maintains exactly one path of cached results to
/// the root in the cache (assuming we evict cache members after each lookup in a sequence of lookups).
/// Higher the `drop` value, more such paths can potentially be stored in the cache at any given moment.
/// Because of that, a higher `drop` value also translates to a higher number of cache-hits,
/// but it might not directly translate to better performance. Because of a larger memory footprint,
/// memory accesses could take longer, slowing down the eviction process. That's why the `drop` value
/// could be fine-tuned to maximize the performance given file size,
/// number of lookups, and amount of available memory.
public init(drop: Int = 0) {
  self.dropMod = drop + 1
}
```


MAJKFL:
@swift-ci Please test 
https://github.com/swiftlang/swift/pull/81209

MAJKFL:
https://github.com/swiftlang/swift/pull/81209

@swift-ci Please test

ahoppen (commented):
Without diving too deeply into the details: I am a little concerned about the cache eviction behavior and the fact that you need to manually call `evictEntriesWithoutHit` (which incidentally doesn’t seem to be called in this PR or https://github.com/swiftlang/swift/pull/81209) and I think it’s easy for clients to forget to call it. Does this more complex cache eviction policy provide significant benefits over a simple LRU cache that keeps, say 100, cache entries? We could share the `LRUCache` type that we currently have in `SwiftCompilerPluginMessageHandling` for that. Curious to hear your opinion.

ahoppen:
Would it be possible to use an LRU cache and provide an eviction method that can be called to clean up the cache as we know that some parts of it are no longer relevant (what you described in the sketch above). That way we would get reasonable out-of-the-box behavior and don’t have an ever-growing cache but also have the ability to keep the cache size low in cases where the client (here the compiler) cares about it and knows the access patterns.

MAJKFL:
> That way we would get reasonable out-of-the-box behavior and don’t have an ever-growing cache but also have the ability to keep the cache size low in cases where the client (here the compiler) cares about it and knows the access patterns.

Ah yes, that’s a very good idea to have an upper bound for the size of the cache. I haven’t thought about it. I’ll try to look into how to extend `LRUCache` from `SwiftCompilerPluginMessageHandling` with the cleanup algorithm then. Should we hoist `LRUCache` to some other, shared place, or should it remain in `SwiftCompilerPluginMessageHandling`?

MAJKFL:
https://github.com/swiftlang/swift/pull/81209https://github.com/swiftlang/swift/pull/81209

@swift-ci Please test

MAJKFL:
https://github.com/swiftlang/swift/pull/81209

@swift-ci Please test Windows Platform

Sources/SwiftLexicalLookup/LookupCache.swift
```diff
@@ -0,0 +1,105 @@
...
+  ///
+  /// - Note: `drop = 0` effectively maintains exactly one path of cached results to
+  /// the root in the cache (assuming we evict cache members after each lookup in a sequence of lookups).
+  /// Higher the `drop` value, more such paths can potentially be stored in the cache at any given moment.
+  /// Because of that, a higher `drop` value also translates to a higher number of cache-hits,
+  /// but it might not directly translate to better performance. Because of a larger memory footprint,
+  /// memory accesses could take longer, slowing down the eviction process. That's why the `drop` value
+  /// could be fine-tuned to maximize the performance given file size,
+  /// number of lookups, and amount of available memory.
+  public init(capacity: Int, drop: Int = 0) {
```
ahoppen: Just an idea: Would it make sense to move the `drop` parameter to `evictEntriesWithoutHit`. That way clients don’t have to think about the dropping cache eviction policy unless they start calling `evictEntriesWithoutHit`. It would also open up the option to vary the size of the cache dynamically depending on the shape of the code that we’re in (not sure if that’s useful or not). It would also remove the need for `bypassDropCounter` in that function because you could pass `drop: 0` there, I think.

Sources/SwiftLexicalLookup/LookupCache.swift
```diff
@@ -0,0 +1,105 @@
...
+    self.ancestorResultsCache = LRUCache(capacity: (capacity + 1) / 2)
+    self.sequentialResultsCache = LRUCache(capacity: capacity / 2)
+    self.dropMod = drop + 1
+  }
+
+  /// Get cached ancestor results for the given `id`.
+  /// `nil` if there's no cache entry for the given `id`.
+  /// Adds `id` and ids of all ancestors to the cache `hits`.
+  func getCachedAncestorResults(id: SyntaxIdentifier) -> [LookupResult]? {
+    guard let results = ancestorResultsCache[id] else { return nil }
```
ahoppen, line 49: If the user doesn’t call `evictEntriesWithoutHit`, `hits` will keep on growing indefinitely. Should we clear up `hits` periodically for elements that are no longer in the cache? Eg. as a kind of garbage collection if `hits.count > capacity * 2`. Or should we only keep track of `hits` if the user opts into it inside the initializer?

Sources/SwiftLexicalLookup/LookupCache.swift
```diff
@@ -0,0 +1,105 @@
...
+  public func evictEntriesWithoutHit(bypassDropCounter: Bool = false) {
+    if !bypassDropCounter {
+      evictionCount = (evictionCount + 1) % dropMod
+      guard evictionCount != 0 else { return }
+    }
+
+    for key in ancestorResultsCache.keysInCache.union(sequentialResultsCache.keysInCache).subtracting(hits) {
+      ancestorResultsCache[key] = nil
+      sequentialResultsCache[key] = nil
+    }
```
ahoppen: Not sure how performance sensitive this is or how expensive the `Set` operations are but I would imagine it would be more performant if we just did the following because it requires no set modifications.

```swift
    for key in ancestorResultsCache where !hits.contains(key) {
      ancestorResultsCache[key] = nil
    }
    for key in sequentialResultsCache where !hits.contains(key) {
      sequentialResultsCache[key] = nil
    }
```

Sources/SwiftLexicalLookup/Scopes/ScopeImplementations.swift
```diff
@@ -698,7 +735,8 @@ import SwiftSyntax
   public func lookup(
     _ identifier: Identifier?,
     at lookUpPosition: AbsolutePosition,
-    with config: LookupConfig
+    with config: LookupConfig,
+    cache: LookupCache?
```
ahoppen, line 741: Should we default `cache` to `nil` to avoid API breakage?

Sources/SwiftLexicalLookup/Scopes/ScopeImplementations.swift
```diff
@@ -942,10 +991,12 @@ extension SubscriptDeclSyntax: WithGenericParametersScopeSyntax, CanInterleaveRe
         in: (isMember ? [.implicit(.self(DeclSyntax(self)))] : LookupName.getNames(from: self)),
         identifier,
         at: lookUpPosition,
-        with: config
+        with: config,
+        cache: cache
       )
     } else {
-      return lookupInParent(identifier, at: lookUpPosition, with: config)
+      // We're not using `lookupParent` to not cache the results here.
```
ahoppen: Why don’t we want to cache the results here? Same for the other places that have

```
We're not using `lookupParent` to not cache the results here.
```

I think adding the *why* as a comment would be helpful for future reference.

Sources/SwiftSyntax/LRUCache.swift
```diff
@@ -33,12 +32,14 @@ public class LRUCache<Key: Hashable, Value> {
   private unowned var tail: _Node?
 
   public let capacity: Int
+  public private(set) var keysInCache: Set<Key>
```
ahoppen: Should all the `public` members be marked as `package` if `LRUCache` is `package` now? Also, thanks for introducing the ability to use `package` in swift-syntax 🙏🏽

Sources/SwiftSyntax/LRUCache.swift
ahoppen, line 1: @rintaro Could you check if you have any concerns for the changes in LRUCache?

Tests/SwiftCompilerPluginTest/LRUCacheTests.swift
```diff
@@ -11,6 +11,7 @@
 //===----------------------------------------------------------------------===//
 
 @_spi(Testing) import SwiftCompilerPluginMessageHandling
+import SwiftSyntax
```
ahoppen, line 13: If `LRUCache` is in SwiftSyntax`, I think we should also move the tests to SwiftSyntax.

Sources/SwiftLexicalLookup/Scopes/ScopeSyntax.swift
```diff
@@ -42,9 +47,60 @@ extension SyntaxProtocol {
   /// due to the ordering rules within the function body.
   public func lookup(
     _ identifier: Identifier?,
-    with config: LookupConfig = LookupConfig()
+    with config: LookupConfig = LookupConfig(),
+    cache: LookupCache? = nil
   ) -> [LookupResult] {
-    scope?.lookup(identifier, at: self.position, with: config) ?? []
+    if let cache, let identifier {
```
ahoppen: Just an idea: Could we avoid one nesting level by making this

```swift
guard let cache, let identifier else {
  return scope?.lookup(identifier, at: self.position, with: config, cache: cache) ?? []
}
```

ahoppen: Does this mean that we don’t use the cache if you run `lookup` without an `identifier`. Shouldn’t we be able to return the results from the cache in that case without filtering?

ahoppen (commented):
Thanks for addressing my review comments. I just had a chance to look at the PR again and left a few comments inline.

Sources/SwiftLexicalLookup/LookupCache.swift
rintaro: If this Alex suggestion is fine `LRUCache.keysInCache` is not needed.

Sources/SwiftSyntax/LRUCache.swift
rintaro: As Alex suggests, I'm not sure this addition is needed.

But even if we want this API, I don't think this `Set` storage is necessary. This is essentially `table.keys` except `Dictionary.Keys` is not a `Set`. I think something like this should be fine.
```swift
package var keys: some Collection<Key> {
  table.keys
}
```
The caller can create a `Set` from this.

Sources/SwiftSyntax/LRUCache.swift
rintaro: I feel like we should eventually want to make a "support" module to host things like this, but I'm fine with moving this to `SwiftSyntax` for now.

Sources/SwiftLexicalLookup/LookupCache.swift
MAJKFL: That's a good idea. I like the idea of not having too much additional state in the cache. I've moved `drop` to `evictEntriesWithoutHit `.

Sources/SwiftLexicalLookup/LookupCache.swift
MAJKFL: Thanks for noticing that. I think garbage collection makes more sense in this case in combination with `drop` parameter inside `evictEntriesWithoutHit`. This way clients won't have to think about the custom eviction policy unless they specifically want to use it.

Sources/SwiftLexicalLookup/Scopes/ScopeImplementations.swift
MAJKFL: It's only used internally and is not exposed to clients since it's in `@_spi(Experimental) extension`. I've added `@_spi(Experimental)` to the function now also to avoid confusion. The main entry point of the query in `ScopeSyntax.swift` does however default `cache` to `nil`.

Sources/SwiftLexicalLookup/Scopes/ScopeSyntax.swift
MAJKFL: No, the if statement here only checks for the special case where both `identifier` and `cache` are not `nil` at the same time (in which case we need to perform filtering). When the condition fails, at least one of them is `nil` and we directly return the result of `scope?.lookup(identifier, at: self.position, with: config, cache: cache) ?? []`.

Sources/SwiftSyntax/LRUCache.swift
MAJKFL: Yes, I think this is a much better idea that avoids redundancy. Thanks!

Sources/SwiftLexicalLookup/LookupCache.swift
```diff
@@ -0,0 +1,109 @@
...
+  private let ancestorResultsCache: LRUCache<SyntaxIdentifier, [LookupResult]>
+  /// Cached results of `SequentialScopeSyntax.sequentialLookup` calls.
+  /// Identified by `SyntaxIdentifier`.
+  private let sequentialResultsCache: LRUCache<SyntaxIdentifier, [LookupResult]>
+  /// Looked-up scope identifiers during cache accesses.
+  private var hits: Set<SyntaxIdentifier> = [] {
+    didSet {
+      if hits.count > capacity * 2 {
+        hits.removeAll()
+      }
```
ahoppen, line 29: This doesn’t seem right to me. Once `hits.count > capacity * 2`, we pretend that there were no hits. Shouldn’t we be clearing `ancestorResultsCache` and `sequentialResultsCache` in that case?

Sources/SwiftLexicalLookup/Scopes/ScopeImplementations.swift
ahoppen: Ah, sorry. I missed that this was SPI.

Sources/SwiftLexicalLookup/Scopes/ScopeSyntax.swift
ahoppen: Reading my comment again, I’m not sure what I thought when I wrote it. Thanks for your explanation.

ahoppen (commented):
One more comment about the automatic cache eviction, otherwise LGTM.

Sources/SwiftLexicalLookup/LookupCache.swift
MAJKFL: I think it makes sense to just remove hits in this case. Both `ancestorResultsCache` and `sequentialResultsCache` will not exceed the cap since they have LRU eviction. Because of that, clients would be able to also use `LookupCache` like any other LRU cache without having to think about the custom eviction policy. 

In an edge case where a client runs `evictEntriesWithoutHit` after this cleanup, `ancestorResultsCache` and `sequentialResultsCache` would be emptied then.

ahoppen: Hmm, I see. But wouldn’t this produce very surprising results in the following case?

- Client calls `evictEntriesWithoutHit`
- Client makes requests that result in exactly `2 * capacity + 1` hits (more also apply but example is easier this way)
- Client calls `evictEntriesWithoutHit`
- All caches are cleared despite there being plenty of hits that would indicate that the cache shouldn’t be cleared.

So, maybe we do need a parameter in the initializer after all that indicates whether the client is planning to call `evictEntriesWithoutHit` and that the `LookupCache` should thus keep track of the this.

PR #3069, opened 2025-05-02
[Macros] Infer nonisolated conformances in macro-expanded code.

DougGregor:
The combination of isolated conformance inference with global-actor-isolated types has introduced a source compatibility issue with existing macros that produce conformances that are meant to be nonisolated. Introduce a syntactic transform that identifies the presence of nonisolated members and makes the conformances introduced by the enclosing group "nonisolated". For example, an extension macro that expands to:

    extension C: P {
      nonisolated func f() { }
    }

will have the macro expansion adjusted to:

    extension C: nonisolated P {
      nonisolated func f() { }
    }

This is controlled by the new macro implementation customization point `Macro.inferNonisolatedConformances`, which defaults to `true` but can be turned off for macro implementations that want to handle this inference themselves.

Macro part of rdar://150419628

### DIFF
--- a/Sources/SwiftCompilerPluginMessageHandling/CompilerPluginMessageHandler.swift
+++ b/Sources/SwiftCompilerPluginMessageHandling/CompilerPluginMessageHandler.swift
@@ -30,6 +30,9 @@ import SwiftSyntaxMacros
 @_spi(PluginMessage)
 public enum PluginFeature: String {
   case loadPluginLibrary = "load-plugin-library"
+
+  /// Whether the plugin knows how to infer nonisolated conformances.
+  case inferNonisolatedConformances = "infer-nonisolated-conformances"
 }
 
 /// A type that provides the actual plugin functions.
--- a/Sources/SwiftSyntaxMacroExpansion/SyntaxProtocol+NonisolatedConformances.swift
+++ b/Sources/SwiftSyntaxMacroExpansion/SyntaxProtocol+NonisolatedConformances.swift
@@ -1,0 +1,147 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+//
+// This file implements inference of "nonisolated" on the conformances that
+// occur within macro-expanded code. It's meant to provide source compatibility
+//
+
+import SwiftSyntax
+
+extension SyntaxProtocol {
+  /// Given some Swift syntax that may contain type definitions and extensions,
+  /// add "nonisolated" to protocol conformances when there are nonisolated
+  /// members. For example, given:
+  ///
+  ///     extension X: P {
+  ///       nonisolated func f() { }
+  ///     }
+  ///
+  /// this operation will produce:
+  ///
+  ///     extension X: nonisolated P {
+  ///       nonisolated func f() { }
+  ///     }
+  @_spi(Testing) @_spi(Compiler)
+  public func inferNonisolatedConformances() -> Syntax {
+    let rewriter = NonisolatedConformanceRewriter()
+    return rewriter.rewrite(self)
+  }
+}
+
+fileprivate class NonisolatedConformanceRewriter: SyntaxRewriter {
+  override func visitAny(_ node: Syntax) -> Syntax? {
+    // We only care about decl groups (non-protocol nominal types + extensions)
+    // that have nonisolated members and an inheritance clause.
+    guard let declGroup = node.asProtocol(DeclGroupSyntax.self),
+      !declGroup.is(ProtocolDeclSyntax.self),
+      declGroup.containsNonisolatedMembers,
+      let inheritanceClause = declGroup.inheritanceClause
+    else {
+      return nil
+    }
+
+    var skipFirst =
+      declGroup.is(ClassDeclSyntax.self)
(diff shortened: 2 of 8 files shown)


DougGregor:
@swift-ci please test

Sources/SwiftCompilerPluginMessageHandling/CompilerPluginMessageHandler.swift
```diff
@@ -30,6 +30,9 @@ import SwiftSyntaxMacros
 @_spi(PluginMessage)
 public enum PluginFeature: String {
   case loadPluginLibrary = "load-plugin-library"
+
+  /// Whether the plugin knows how to infer nonisolated conformances.
+  case inferNonisolatedConformances = "infer-nonisolated-conformances"
```
rintaro, line 35: This doesn't seem to be used.

DougGregor: It ends up getting passed back to the compiler, which checks for its presence. See the [compiler PR](https://github.com/swiftlang/swift/pull/81260) I just put up.

DougGregor:
https://github.com/swiftlang/swift/pull/81260

@swift-ci please test

Sources/SwiftSyntaxMacroExpansion/SyntaxProtocol+NonisolatedConformances.swift
```diff
@@ -0,0 +1,147 @@
...
+
+    return Syntax(
+      fromProtocol: declGroup.with(
+        \.inheritanceClause,
+        inheritanceClause.with(
+          \.inheritedTypes,
+          InheritedTypeListSyntax(inheritedTypes)
+        )
+      )
+    )
```
rintaro, line 77: Returning something from `SyntaxRewriter.visitAny(_:)` prevents it from waling into children. If we what to rewrite things like
```swift
extension MyStruct: P {
  nonisolated func f() { }
  struct Inner: Q {
    nonisolated func b() {}
  }
}
```
to
```swift
extension MyStruct: nonisolated P {
  nonisolated func f() { }
  struct Inner: nonisolated Q {
    nonisolated func b() {}
  }
}
```
You'd need 
```swift
.with(
  \.memberBlock
  self.rewrite(declGroup.memberBlock, detach: true)
)
```

Sources/SwiftSyntaxMacroExpansion/SyntaxProtocol+NonisolatedConformances.swift
```diff
@@ -0,0 +1,147 @@
...
+      if skipFirst {
+        skipFirst = false
+        return inheritedType
+      }
+
+      return inheritedType.with(\.type, "nonisolated \(inheritedType.type)")
+    }
+
+    return Syntax(
+      fromProtocol: declGroup.with(
```
rintaro, line 70:
```suggestion
      fromProtocol: declGroup.detached.with(
```
so that it doesn't rewrite the whole tree. Same for other `.with(..)`

Sources/SwiftSyntaxMacroExpansion/SyntaxProtocol+NonisolatedConformances.swift
```diff
@@ -0,0 +1,147 @@
...
+        return inheritedType
+      }
+
+      if skipFirst {
+        skipFirst = false
+        return inheritedType
+      }
+
+      return inheritedType.with(\.type, "nonisolated \(inheritedType.type)")
+    }
```
rintaro, line 67: If this doesn't change anything, can we return `nil` here? To save the rewrite cost.

Sources/SwiftCompilerPluginMessageHandling/CompilerPluginMessageHandler.swift
rintaro: Okay, but I consider `PluginFeature` is more like variation of plugin executable. E.g. single-module-plugin v.s. plugin-server. For something like this, I'd use `PluginMessage.PROTOCOL_VERSION_NUMBER`. Like, in `PluginHost.swift` in the swift repo: 
```swift
extension CompilerPlugin {
  hasIinferNonisolatedConformances: Bool { capability.protcolVersion >= 8 }
}
```

Also, you only added this feature to `LibraryPluginProvider`, which doesn't cover single-module executable  plugins.

Sources/SwiftSyntaxMacroExpansion/SyntaxProtocol+NonisolatedConformances.swift
```diff
@@ -0,0 +1,147 @@
...
+    let inheritedTypes = inheritanceClause.inheritedTypes.map { inheritedType in
+      // If there's already a 'nonisolated' or some kind of custom attribute
+      if inheritedType.type.hasNonisolatedOrCustomAttribute {
+        return inheritedType
+      }
+
+      if skipFirst {
+        skipFirst = false
+        return inheritedType
+      }
```
ahoppen, line 64: Should this check be before the `hasNonisolatedOrCustomAttribute` check? Not that it makes much sense but we skip the second type using the `skipFirst` check if you have something like `enum Foo: nonisolated Int, MyProtocol`.

Sources/SwiftSyntaxMacros/MacroProtocols/Macro.swift
```diff
@@ -15,4 +15,29 @@ public protocol Macro {
...
+  ///
+  /// to
+  ///
+  ///     extension C: nonisolated P {
+  ///       nonisolated func f() { }
+  ///     }
+  ///
+  /// This operation defaults to `true`. Macros can implement it to return
+  /// `false` to prevent this adjustment to the macro-expanded code.
+  static var inferNonisolatedConformances: Bool { get }
```
ahoppen, line 36: Could you create an RFC for the API addition?

DougGregor:
Not going to go in this direction

PR #3070, opened 2025-05-03
[SwiftParser] Parse operator function with value generics

rintaro:
Operator function parsing has a heuristics to determine if `<` is a part of the operator name or the generic parameter clause. Handle `let` there because value generics uses it.

rdar://149556573


bnbarham (approved):

rintaro:
https://github.com/swiftlang/swift/pull/81265
@swift-ci Please test

PR #3073, opened 2025-05-08
[6.1][bazel] Update for new libraries

keith:


keith:
@swift-ci please test

keith:
nbd if this is too late we can just close

bnbarham:
@swift-ci please test Linux platform

bnbarham:
Where is this needed @keith? Just on the release branch or in a 601.* tag?

keith:
just in case there is a new tag probably, since i think folks are successfully using the existing tag. i guess the other case is if these libraries existed in the latest tag and someone wanted to use them directly

PR #3091, opened 2025-06-02
SwiftSyntax support for module selectors

beccadax:
This PR adds module selectors to SwiftSyntax and SwiftParser (draft of matching compiler feature in swiftlang/swift#34556). This feature was [pitched ages ago](https://forums.swift.org/t/pitch-fully-qualified-name-syntax/28482); a proper proposal is ~~on my todo list~~ ~~[now available](https://gist.github.com/beccadax/115349c9aaefb6a35a18e6f1d147151a)~~ under review as SE-0491.

~~**Note**: This PR is still a work in progress—in particular, I need to go back and improve the tests, both by expanding their coverage and by adding assertions about the resulting syntax trees—but I'd appreciate feedback on the design while I'm working on that.~~

**Reviewers**: If you review this commit-by-commit, I've separated out the big dumb mechanical changes from the ones that actually change parsing logic.

### DIFF
--- a/CodeGeneration/Sources/SyntaxSupport/CompatibilityLayer.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/CompatibilityLayer.swift
@@ -312,6 +312,7 @@ public struct DeprecatedMemberInfo {
       }
     }
 
+    @_optimize(none)  // Workaround for SIL optimizer crash (rdar://158353230)
     private func firstIndexOfChild(named targetName: String) -> Int {
       guard let i = children.firstIndex(where: { $0.name == targetName }) else {
         fatalError(


nkcsgexi:
🥳

Sources/SwiftParser/Names.swift
```diff
@@ -17,34 +17,328 @@
...
+  /// Inject `moduleSelector` into `node`, its first child, its first child's first child, etc. If necessary,
+  /// `moduleSelector`'s tokens will be inserted into an unexpected syntax node rather than the selector itself being
+  /// inserted. This should only be called when `node` comes *immediately* after `moduleSelector`.
+  ///
+  /// Use this method to "prepend" a module selector to a tree of syntax nodes. Doing so can avoid threading the module
+  /// selector through a lot of code, which can consume a lot of stack space and be fairly error-prone.
+  ///
+  /// - Precondition: `node` must, at minimum, have a descendant with an unexpected nodes child; it therefore cannot be
+  ///   a token or an empty collection.
+  func attach<Node: RawSyntaxNodeProtocol>(_ moduleSelector: RawModuleSelectorSyntax?, to node: Node) -> Node {
```
beccadax: Retroactively rewriting a node to insert new information is a pretty unconventional way for SwiftParser to do things, so I thought I should justify this design.

Because a module selector is a prefix on a different syntax and requires two tokens, it's difficult to peek past one and make decisions based on the tokens that follow it. Instead, I found it easier to parse them in relatively high-level productions; that gets the tokens out of the way, but it means they're often parsed a significant distance from the node that will become their parent. I tried two other approaches to coping with this problem:

1. Adding new `ModuleSelectorExprSyntax` and `ModuleSelectorTypeSyntax` nodes. The issue was that I want to make sure we could parse *invalid* module selectors—even in declaration syntax, patterns, etc.—into unexpected nodes, and that didn't offer a good way to do so. (I also found that the module selectors on member lookups wouldn't be able to be handled in this way; I didn't really like that.)

2. Passing the `ModuleSelectorSyntax` node down as a parameter and threading it through to wherever it was needed. This required me to add a lot of parameters and manually insert unexpected module selectors into a lot of nodes, but that's not actually what scuttled the idea—it's that it dramatically increased stack usage. At one point I had to reduce the maximum recursion in development builds to 10, which isn't even enough to handle the swift-syntax repo itself.

Retroactively attaching module selectors in this fashion allows more of the parser to ignore the fact that an invalid module selector might have been parsed earlier on, while also avoiding the stack usage problems I mentioned.

rintaro: Could you explain why we want

> I want to make sure we could parse invalid module selectors—even in declaration syntax, patterns, etc.

What makes module selectors different from other invalid things? E.g.:
```
let Foo.Bar = 12
```

beccadax: A couple reasons:

1. The difference between a name and a name reference is a bit subtle and hasn't previously been grammatically important; I can imagine that developers might be confused about where they are allowed to use module selectors.

2. If a developer writes `Foo::Bar` in a place where module selectors are not valid, we can be almost certain that they want to keep `Bar` (because we know that `Foo` was supposed to be a module name), but the naïve recovery behavior will treat `Foo` as the name and `::Bar` as unexpected syntax. Tailoring the recovery gives us a tree that better reflects the user's likely intent. (Whereas with `let Foo.Bar` the developer might have wanted `let Bar`, or `let Foo`, or `let Foo = .Bar`, or any number of other possibilities; it's harder to be certain of their intent.)

beccadax:
@swift-ci please test

beccadax:
@swift-ci please test

beccadax:
@swift-ci please test

beccadax:
@swift-ci please test

nkcsgexi:
@swift-ci please test macOS

nkcsgexi:
@swift-ci please test Linux

nkcsgexi:
This build failure seems to be real:
```
12:29:57  /Users/ec2-user/jenkins/workspace/swift-syntax-PR-macOS/branch-main/swift-syntax/Sources/SwiftSyntax/generated/SyntaxTraits.swift:209:7: error: protocol requirement 'moduleSelector' cannot be declared '@_spi' without a default implementation in a protocol extension
12:29:57  207 | 
12:29:57  208 |   @_spi(ExperimentalLanguageFeatures)
12:29:57  209 |   var moduleSelector: ModuleSelectorSyntax? {
12:29:57      |       `- error: protocol requirement 'moduleSelector' cannot be declared '@_spi' without a default implementation in a protocol extension
12:29:57  210 |     get
12:29:57  211 |     set
```

beccadax:
I can't actually reproduce that failure locally, but I'm pushing a speculative fix.

beccadax:
@swift-ci please test

beccadax:
@swift-ci please test macOS

beccadax:
@swift-ci please test Linux

beccadax:
@swift-ci please test Windows

nkcsgexi:
We are good now with macOS and Linux. Let's try Windows again. @swift-ci please test Windows.

nkcsgexi:
hmm, the Windows CI hit a compiler crasher:
```
[519/528] Compiling SwiftDiagnostics Convenience.swift
error: compile command failed due to exception 3 (use -v to see invocation)
SIL memory lifetime failure in @$s11SwiftParser13TokenConsumerPAAE27consumeModuleSelectorTokens0C0Qz22moduleNameOrUnexpected_AF010colonColonC0SayAFG5extratSgyF: memory is initialized, but shouldn't be
```

CodeGeneration/Sources/SyntaxSupport/CommonNodes.swift
```diff
@@ -233,6 +233,14 @@ public let COMMON_NODES: [Node] = [
...
     ],
     children: [
+      Child(
+        name: "moduleSelector",
+        kind: .node(kind: .moduleSelector),
+        experimentalFeature: .moduleSelector,
+        documentation:
+          "A module selector. Some expressions can be prefixed with module selectors, so if one is parsed before an invalid expression, it will be inserted here.",
+        isOptional: true
+      ),
```
rintaro: I don't feel using `MissingExprSyntax` as dangling module selector is the way to go. `MissingExprSyntax` is a "placeholder" for expression with _unknown_ kind. But IMO, we can reasonably assume the user is to add an identifier after a module selector to form `DeclReferenceExprSyntax`.
I feel it's more natural to model it as `DeclReferenceExprSyntax` with missing `baseName`.

Same for `MissingTypeSyntax`

Sources/SwiftParser/Names.swift
rintaro: I've been thinking about this. but I'm still not convinced. Let's say for `let Foo::Bar = <expr>`,  I don't think we can say the user tried to specify the module name. The user might mistyped `let Foo: Bar = <expr>`.  

Whether developers might get confused between name declarations and name references is subjective, so I won’t argue that. But I don’t think it’s worth adding this extra complexity to the implementation. It looks overly complex. Also, rewriting parsed nodes leaves abandoned nodes in the arena, which is not great for memory usage. IMO, the module selector parsing should be mostly contained in `parseDeclReferenceExpr()`, `(can)parseTypeIdentifier()`, and `(can)parseSimpleType()` (for member types). Of course we should do our best to emit better diagnostics. But I don't feel this is the way. Also we can improve diagnostics later. Can we make it simple for the initial implementation?

Sources/SwiftParser/Names.swift
```diff
@@ -11,40 +11,373 @@
...
+    guard self.peek(isAt: .colonColon) else {
+      // ...however, we will also allow the *current* token to be `::`. `consumeModuleSelectorTokens()` will create a
+      // missing identifier.
+      return self.at(.colonColon)
+    }
+
+    // Technically the current token *should* be an identifier, but we also want to diagnose other tokens that might be
+    // used by accident (particularly keywords and `_`). However, we don't want to consume tokens which would make the
+    // surrounding structure mis-parse.
+    return self.at(anyIn: StructuralTokens.self) == nil
```
rintaro: I don't think it's a good idea to (almost) always consider `<token>` `::` as a module qualifier. E..g.
```
class
::
```
This looks to me an incomplete `class` declaration and just an orphan `::`. Even if it's on the same line with no-space, I don't think we need to parse it as a module selector.

rintaro (commented):
Apologies I didn't review this sooner. I still haven't look into the implementation closely, but here's the first round 🙇

Sources/SwiftParser/Names.swift
bnbarham: > Also, rewriting parsed nodes leaves abandoned nodes in the arena, which is not great for memory usage

Yeah, our policy so far has been that we're okay with having abandoned nodes in the error case but not the success. I agree with Rintaro here - the added complexity doesn't seem worth the difference in the diagnostic (and tree) that we'd end up with.

beccadax:
Okay, this is a dramatically simpler implementation that doesn't try to do anything special to handle module selectors at invalid locations.  I can't say I'm happy with some of the recovery behavior I'm seeing—in particular, parameter lists and tuple patterns seem to just throw up their hands in a pretty ugly way instead of recovering to a `TokenKind.comma`—but maybe that's a more general problem.

Sources/SwiftParser/Names.swift
```diff
@@ -45,6 +45,151 @@ extension Parser {
...
+      }
+      extra.append(self.eat(.colonColon))
+    }
+    return (moduleName, colonColonToken, extra)
+  }
+}
+
+extension Parser {
+  /// Parses one or more module selectors, if present.
+  mutating func parseModuleSelector() -> RawModuleSelectorSyntax? {
```
rintaro: I know there're several parsing functions that doesn't follow this convention, but:
```suggestion
  mutating func parseModuleSelectorIfPresent() -> RawModuleSelectorSyntax? {
```
or check `self.isAtModuleSelector()` before calling this and make this non-optional.

Sources/SwiftParser/Names.swift
```diff
@@ -68,6 +213,9 @@ extension Parser {
   }
 
   mutating func parseDeclReferenceExpr(_ flags: DeclNameOptions = []) -> RawDeclReferenceExprSyntax {
+    // Consume a module selector if present.
+    let moduleSelector = self.parseModuleSelector()
```
rintaro: I think presence of the module selector should affect `flags` (`DeclNameOptions.keywords`).
It would be weird if `ModName::default` is rejected, but `foo.ModName::default` is allowed.
related: https://forums.swift.org/t/pitch-module-selectors/80835/82

Sources/SwiftParser/Names.swift
```diff
@@ -45,6 +45,151 @@ extension Parser {
...
+  mutating func isAtModuleSelector() -> Bool {
+    // If this is a module selector, the next token should be `::`.
+    guard self.peek(isAt: .colonColon) else {
+      // ...however, we will also allow the *current* token to be `::`. `consumeModuleSelectorTokens()` will create a
+      // missing identifier.
+      return self.at(.colonColon)
+    }
+
+    // Technically the current token *should* be an identifier, but we also want to diagnose other tokens that might be
+    // used by accident (particularly keywords and `_`). However, we don't want to consume tokens which would make the
```
rintaro: Would you show me some examples where consuming keywords as a _invalid_ module identifiers improves the diagnostics?

IMO, keyword as a module name rule should follow existing keyword rules should be like:
* Bare keywords as module names are allowed after `.` (following SE-0071). Otherwise it's not a module selector even if there's `::` after that
* Some special keywords such as `Self`, `self`, `_` and `Any` should be parsed as _reserved_ names (even without `.`). And they should be rejected/diagnosed accordingly, either in the parser or the type checker.

Sources/SwiftSyntax/generated/RenamedChildrenCompatibility.swift
```diff
@@ -1473,13 +1473,12 @@ extension ConsumeExprSyntax {
 }
 
 extension DeclReferenceExprSyntax {
-  @available(*, deprecated, renamed: "unexpectedBeforeBaseName")
-  public var unexpectedBeforeIdentifier: UnexpectedNodesSyntax? {
```
rintaro, line 1477: This accessor is removed now? I think it's okay, but any idea how much effort would it take to preserve this?

Sources/SwiftParser/Declarations.swift
```diff
@@ -445,16 +447,25 @@ extension Parser {
...
     var keepGoing: RawTokenSyntax? = nil
     var loopProgress = LoopProgressCondition()
     repeat {
       let name = self.parseAnyIdentifier()
-      keepGoing = self.consume(if: .period)
+      keepGoing = self.consume(ifAnyIn: ImportPathComponentSyntax.TrailingPeriodOptions.self)
+
+      // If we got `::` when it's not allowed, reject it
+      var unexpectedBeforeTrailingPeriod: RawUnexpectedNodesSyntax?
+      if keepGoing?.tokenKind == .colonColon && !hasImportKind {
```
rintaro: This allows:
```
import struct Foo::Bar::Baz
```
which I don't think intentional?

Also nit: If you use optional `var` please initialize it with `nil` explicitly. Or make it `let` and make `else` branch to initialize it.

Sources/SwiftParser/Names.swift
beccadax: A module selector is never mandatory, so I'm going to rename the method.

Sources/SwiftParser/Declarations.swift
beccadax: `import struct Foo.Bar.Baz` is also semantically invalid but permitted by SwiftParser. ASTGen needs to diagnose both, and the same check would handle both of them.

rintaro: Clang submodules are valid and accepted in `import-path`, so ASTGen won't diagnose it.
E.g. `import class Foundation.NSData.NSData` works. I know this is equivalent with `import class Foundation.NSData`, but we cannot just reject it.

Sources/SwiftSyntax/generated/RenamedChildrenCompatibility.swift
beccadax: Ah, there was a bug in the changes I made to hide experimental children in compatibility layers. Will fix.

Sources/SwiftParser/Declarations.swift
beccadax: Huh—that fact dropped out of my memory, but I must have known it when I created `ImportPath` in the compiler back in 2020, because it's clearly designed to support it. Weird how that happens sometimes.

I'll have to decide if I want to support `import class Foundation.NSData::NSData` or not. (I'm thinking no—we allow `::` there so that we allow the same syntax that's valid at a use site, and in the current proposal a submodule wouldn't be valid at a use site.)

rintaro: I agree we should NOT support `import class Foundation.NSData::NSData`.
If `isAtModuleSpecifier()` after `import-kind`, I think it's okay to just parse the module specifier and the following identifier, then return it.

Sources/SwiftParser/Names.swift
beccadax: I'll narrow this to accept `.wildcard`, `.binaryOperator`, and the keywords `self`, `Self`, `super`, and `Any` in addition to identifiers. (These extra tokens are still not valid—they're just assumed to be attempts to write an identifier.) Does that sound like a reasonable list?

rintaro: Hmm I'm not 100% sure about `.binaryOperator`. Actually that brought up another question which is whether `:` should be a "bound" character. (i.e. `*::foo` is "binary" or "prefix" operator). Either way, for example for `foo + ::bar`, for me, this look like a missing module identifier before `::`. Maybe special casing `*` with any operator kind would be more reasonable.
(though I personally can't imagine `*::ignite()` would be a thing because it can't be use after `.` I.e. for `foo.*::ignite()`, `.*` is an operator)

`.wildcard`, `self`, `Self`, `super` and `Any` sounds good to me.

rintaro: Ah `foo + ::bar` is parsed as a regular binary operator expression either way. Hmm maybe fine then...  🤔

Sources/SwiftParser/Declarations.swift
```diff
@@ -163,6 +163,8 @@ extension TokenConsumer {
       }
 
       return true
+    case .rhs(let introducer):
+      return subparser.shouldParsePatternBinding(introducer: introducer)
```
rintaro: Could you explain what this change is for?

Sources/SwiftParser/Expressions.swift
```diff
@@ -1199,6 +1207,10 @@ extension Parser {
     pattern: PatternContext,
     flavor: ExprFlavor
   ) -> RawExprSyntax {
+    if self.isAtModuleSelector() && (!pattern.admitsBinding || !self.lookahead().isInBindingPatternPosition()) {
+      return self.parseIdentifierExpression(flavor: flavor)
+    }
```
rintaro: Could you add comments why this is hoisted?

Sources/SwiftParser/Patterns.swift
```diff
@@ -272,7 +272,10 @@ extension Parser.Lookahead {
     // than likely need to be made contextual as well before finalizing their
     // grammar.
     case ._borrowing, .borrowing:
-      return peek(isAt: TokenSpec(.identifier, allowAtStartOfLine: false))
+      return peek(
+        isAt: TokenSpec(.identifier, allowAtStartOfLine: false),
+        TokenSpec(.wildcard, allowAtStartOfLine: false)
```
rintaro, line 277: This formatting is confusing. Could you put these on a same line?
```suggestion
      return peek(
        isAt: TokenSpec(.identifier, allowAtStartOfLine: false), TokenSpec(.wildcard, allowAtStartOfLine: false)
```
 If not, how about:
```suggestion
      return peek(
        isAt:
          TokenSpec(.identifier, allowAtStartOfLine: false),
          TokenSpec(.wildcard, allowAtStartOfLine: false)
```

Sources/SwiftParser/Names.swift
```diff
@@ -45,6 +45,114 @@ extension Parser {
...
+  ///
+  /// Returns a tuple comprised of:
+  ///
+  ///  - `moduleNameOrUnexpected`: The module name if present; in a valid module selector, this will be a present
+  ///     identifier, but either of those can be untrue in invalid code.
+  ///  - `colonColonToken`: The `::` indicating this module selector. Always `.colonColon`, always present.
+  ///  - `extra`: Tokens for additional trailing module selectors. There is no situation in which two module selectors
+  ///    can be validly chained.
+  ///  - `skipQualifiedName`: True if the next token should be interpreted as a different statement.
+  mutating func consumeModuleSelectorTokens() -> (
```
rintaro: 🙇 
```suggestion
  mutating func consumeModuleSelectorTokensIfPresent() -> (
```

Sources/SwiftParser/Declarations.swift
beccadax: This appears to no longer be necessary. (It used to be needed because `isAtModuleSelector()` would accept any keyword as an invalid module name.)

Sources/SwiftParser/Patterns.swift
beccadax: swift-format wants to format this either the way it's written here, or like this: 
```suggestion
      return peek(
        isAt:
          TokenSpec(.identifier, allowAtStartOfLine: false),
        TokenSpec(.wildcard, allowAtStartOfLine: false)
```
Which I doubt is any better.

(It's technically correct for it to do so; the API Design Guidelines say that descriptions which apply to all arguments equally should be part of the base name, not the first argument label.)

Sources/SwiftParser/Expressions.swift
beccadax: Its primary purpose is to ensure that invalid module selectors are handled correctly (although valid ones went through there too), but now that we allow a much narrower set of invalid tokens, we can just handle them all in a `switch` case, which will be clearer. I'll do that instead.

Sources/SwiftParser/Declarations.swift
beccadax: I've removed this from the relevant commit.

Sources/SwiftParser/Names.swift
beccadax: I ended up removing operators, as I realized that if you try to use one in an expression, the parser will likely consume it to form a `BinaryOperatorExpr` or `PrefixOperatorExpr` before `parsePrimaryExpression` ever gets a chance to see it.

beccadax:
@swift-ci please test

beccadax:
@swift-ci please test

beccadax:
@swift-ci please test

beccadax:
@swift-ci please test

Sources/SwiftParser/Types.swift
```diff
@@ -227,35 +247,18 @@ extension Parser {
...
-            arena: self.arena
-          )
-        )
-      }
-      return node
-    }
-
     var base: RawTypeSyntax
-    switch self.at(anyIn: TypeBaseStart.self)?.spec {
+    switch self.isAtModuleSelector() ? .identifier : self.at(anyIn: TypeBaseStart.self)?.spec {
```
rintaro, line 251: Hmm, could you teach me how this handles `let _: Self::Foo` etc?  
I thought `.identifier` only matches identifiers not including `Self` etc. So this ends up `case nil`. but apparently not?

beccadax: This basically means that, if we are at a module selector (which might have a missing module name or one of the tokens we accept as an invalid module name) we will always go into `case .identifier`, rather than `case nil` or one of the other cases. `Self` and `Any` are part of that case anyway, so the interesting tokens are:

```swift
let w: _::Foo        // would otherwise match `case .wildcard`
let x: self::Foo     // would otherwise match `case nil`
let y: super::Foo    // would otherwise match `case nil`
let z: ::Foo         // would otherwise match `case nil`
```

Rather than calling  `parsePlaceholderType()` or constructing a `RawMissingTypeSyntax`, these will go into `parseTypeIdentifier()`, which will parse that token into a module selector with unexpected and/or missing tokens, and then parse whatever's after the `::` as the actual name of the type (or, if it's not sufficiently identifier-like, parse a missing identifier).

A different way to write this might be:

```swift
    switch (self.at(anyIn: TypeBaseStart.self)?.spec, self.isAtModuleSelector()) {
    case (.Self, false), (.Any, false), (.identifier, false), (_, true):
      base = RawTypeSyntax(self.parseTypeIdentifier())
    case (.leftParen, false):
      base = RawTypeSyntax(self.parseTupleTypeBody())
    case (.leftSquare, false):
      base = self.parseCollectionType()
    case (.wildcard, false):
      base = RawTypeSyntax(self.parsePlaceholderType())
    case (nil, false):
      return RawTypeSyntax(RawMissingTypeSyntax(arena: self.arena))
    }
```

beccadax:
@swift-ci please test

Sources/SwiftParser/Types.swift
rintaro: Oooh, I see. I misunderstood the code. Thanks for explaining.

beccadax:
@swift-ci please test

Sources/SwiftParser/Types.swift
beccadax: @rintaro If there aren't any more comments, could I get an approve on this? Probably shouldn't merge without one.

rintaro (approved):
This looks good to me, after working around the failures! Thank you!

beccadax:

beccadax:
@swift-ci please test

PR #3092, opened 2025-06-04
Handle macro expansion args in placeholder expansion

hamishknight:
Update `ExpandEditorPlaceholdersToLiteralClosures` and `CallToTrailingClosures` such that they can handle macro expansion exprs and decls. This unfortunately requires changing them such that they take a `Syntax` input and output to satisfy the conformance to `SyntaxRefactoringProvider`, but that seems preferable to making the refactoring types themselves generic.

If in the future we decide to expose `CallLikeSyntax` as a public protocol, we could decide to expose additional generic `refactor` overloads here. However it's not clear there's enough clients of these APIs to motivate doing that in this PR.

### DIFF
--- a/Sources/SwiftRefactor/CallLikeSyntax.swift
+++ b/Sources/SwiftRefactor/CallLikeSyntax.swift
@@ -1,0 +1,35 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2025 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+#if compiler(>=6)
+public import SwiftSyntax
+#else
+import SwiftSyntax
+#endif
+
+// TODO: We ought to consider exposing this as a public syntax protocol.
+@_spi(Testing) public protocol CallLikeSyntax: SyntaxProtocol {
+  var arguments: LabeledExprListSyntax { get set }
+  var leftParen: TokenSyntax? { get set }
+  var rightParen: TokenSyntax? { get set }
+  var trailingClosure: ClosureExprSyntax? { get set }
+  var additionalTrailingClosures: MultipleTrailingClosureElementListSyntax { get set }
+}
+@_spi(Testing) extension FunctionCallExprSyntax: CallLikeSyntax {}
+@_spi(Testing) extension MacroExpansionExprSyntax: CallLikeSyntax {}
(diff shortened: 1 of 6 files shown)


hamishknight:
@swift-ci please test

hamishknight:
@swift-ci please test

Sources/SwiftRefactor/CallLikeSyntax.swift
```diff
@@ -0,0 +1,44 @@
...
+  var trailingClosure: ClosureExprSyntax? { get set }
+  var additionalTrailingClosures: MultipleTrailingClosureElementListSyntax { get set }
+}
+@_spi(Testing) extension FunctionCallExprSyntax: CallLikeSyntax {}
+@_spi(Testing) extension MacroExpansionExprSyntax: CallLikeSyntax {}
+@_spi(Testing) extension MacroExpansionDeclSyntax: CallLikeSyntax {}
+
+extension SyntaxProtocol {
+  @_spi(Testing) public var asCallLikeSyntax: (any CallLikeSyntax)? {
+    if let function = self.as(FunctionCallExprSyntax.self) {
```
ahoppen: Could this be `self.as(SyntaxProtocol.self) as? CallLikeSyntax` so that we don’t need to add new entries here if we add new conformances to `CallLikeSyntax`?

Sources/SwiftRefactor/CallLikeSyntax.swift
```diff
@@ -0,0 +1,44 @@
...
+  var rightParen: TokenSyntax? { get set }
+  var trailingClosure: ClosureExprSyntax? { get set }
+  var additionalTrailingClosures: MultipleTrailingClosureElementListSyntax { get set }
+}
+@_spi(Testing) extension FunctionCallExprSyntax: CallLikeSyntax {}
+@_spi(Testing) extension MacroExpansionExprSyntax: CallLikeSyntax {}
+@_spi(Testing) extension MacroExpansionDeclSyntax: CallLikeSyntax {}
+
+extension SyntaxProtocol {
+  @_spi(Testing) public var asCallLikeSyntax: (any CallLikeSyntax)? {
```
ahoppen: If we want to follow the same convention as for the other traits, this should be 

```swift
func asProtocol(_: CallLikeSyntax.Protocol) -> (any CallLikeSyntax)? {
```

ahoppen (commented):
Looks good to me, just two improvement suggestions on `CallLikeSyntax`.

Sources/SwiftRefactor/CallLikeSyntax.swift
hamishknight: Ah nice

hamishknight:
@swift-ci please test

bnbarham (approved):

ahoppen (approved):
Thanks, LGTM

ahoppen:
@swift-ci Please test Windows

PR #3104, opened 2025-06-16
Fix a crash if we had a dollar identifier as the second parameter label in a closure not followed by a colon

ahoppen:
There was a mismatch between the call to `eat` and `at`. Switch the pair to a `consume(if:)` to avoid the issue.

Fixes swiftlang/swift-syntax#3103


ahoppen:
@swift-ci Please test

hamishknight:
@swift-ci please test Windows

PR #3105, opened 2025-06-16
Modify SyntaxVisitor to support throwing errors during traversal.

thunderseethe:
I've found it helpful when writing syntax visitors to support early termination by throwing an exception. Replicating this functionality with the existing error mechanisms is difficult. You have to store the error in a property and set a flag that returns `.skipChildren` from every visit method to end traversal. The addition of typed throws in Swift 6 provides an avenue to improve error handling.

We can introduce a new `ThrowingSyntaxVisitor<E>` that marks every method `throws (E)`, allowing for users to throw an error of a specified type. The existing `SyntaxVisitor` can be made a type alias for `ThrowingSyntaxVisitor<Never>` to avoid requiring any changes to existing code.

In this PR, I've migrated `SyntaxVisitor` and `SyntaxAnyVisitor` to `ThrowingSyntaxVisitor` and `ThowingSyntaxAnyVisitor` respectively. This requires updating CodeGeneration's swift syntax dependency from 510 to 600 which may be a breaking semver change.

### DIFF
--- a/Sources/SwiftSyntax/generated/SyntaxAnyVisitor.swift
+++ b/Sources/SwiftSyntax/generated/SyntaxAnyVisitor.swift
@@ -2300,6 +2300,16 @@ open class ThrowingSyntaxAnyVisitor<E: Error>: ThrowingSyntaxVisitor<E> {
     try visitAnyPost(node._syntaxNode)
   }
 
+  @_spi(ExperimentalLanguageFeatures)
+  override open func visit(_ node: UsingDeclSyntax) throws(E) -> SyntaxVisitorContinueKind {
+    return try visitAny(node._syntaxNode)
+  }
+
+  @_spi(ExperimentalLanguageFeatures)
+  override open func visitPost(_ node: UsingDeclSyntax) throws(E) {
+    try visitAnyPost(node._syntaxNode)
+  }
+
   override open func visit(_ node: ValueBindingPatternSyntax) throws(E) -> SyntaxVisitorContinueKind {
     return try visitAny(node._syntaxNode)
   }
--- a/Sources/SwiftSyntax/generated/SyntaxVisitor.swift
+++ b/Sources/SwiftSyntax/generated/SyntaxVisitor.swift
@@ -5780,7 +5780,6 @@ open class ThrowingSyntaxVisitor<E: Error> {
   }
 
   @inline(never)
-<<<<<<< HEAD
   private func visitUsingDeclSyntaxImpl(_ node: Syntax) throws (E) {
     if try visit(UsingDeclSyntax(unsafeCasting: node)) == .visitChildren {
       try visitChildren(node)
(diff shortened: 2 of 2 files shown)


CodeGeneration/Package.swift
```diff
@@ -12,7 +12,7 @@ let package = Package(
     .executable(name: "generate-swift-syntax", targets: ["generate-swift-syntax"])
   ],
   dependencies: [
-    .package(url: "https://github.com/swiftlang/swift-syntax", from: "510.0.0"),
+    .package(url: "https://github.com/swiftlang/swift-syntax", from: "600.0.0"),
```
ahoppen, line 15: Any reason that you didn’t go to `601.0.0` while you’re at it?

thunderseethe: I assumed this was kept low for compatibility reasons. Happy to update it if that's not the case

thunderseethe:
Collected some numbers. My machine is an Apple M1 Max 64GB running Sequioa 15.5.

**Baseline**

testEmptyAnyVisitorPerformance() number of instructions: 15,009,691
testEmptyRewriterPerformance() number of instructions: 17,531,729
testEmptyVisitorPerformance() number of instructions: 13,374,929

**Throwing**

testEmptyAnyVisitorPerformance() number of instructions: 392,664,189
testEmptyRewriterPerformance() number of instructions: 18,512,487
testEmptyVisitorPerformance() number of instructions: 38,7092,563

These numbers are not good. The throwing visitor is an order of magnitude more instructions than the baseline. I tried out generating just a `throws` (no generic) visitor and the numbers for that were much closer to baseline:

testEmptyVisitorPerformance() number of instructions: 16,219,363

I also ran under debug:

**Baseline**

testEmptyAnyVisitorPerformance() number of instructions: 415,638,162
testEmptyRewriterPerformance() number of instructions: 457,683,193
testEmptyVisitorPerformance() number of instructions: 407,289,419

**Throwing**

testEmptyAnyVisitorPerformance() number of instructions: 434,424,748
testEmptyRewriterPerformance() number of instructions: 458,698,432
testEmptyVisitorPerformance() number of instructions: 421,628,452

Debug shows a consistent regression of ~20 million instructions. Given how much the gap shrinks under a debug build (and the lack of gap for `throws`) I suspect the generic `E: Error` parameter is causing the visitor to invoke some witness table methods that aren't being monomorphized away? If you have any insight, I'd love to hear it.

CodeGeneration/Package.swift
ahoppen: It’s just 510.0.0 because nobody has touched it since. Updating to the latest released version makes sense.

ahoppen:
Ooof, that performance regression is quite significant, thanks for measuring. I suspect that we just can’t optimize the syntax visitor as much because of the generics. One thing that might be worth trying is to specifically specialize the syntax visitor for `Never` and `any Error` using `@specialized` (https://github.com/swiftlang/swift-evolution/blob/main/proposals/0460-specialized.md). That might be able to recover the performance gap. If you don’t have a Swift 6.2 compiler, you can use `@_specialized` with an underscore, which has been available for quite a while.

thunderseethe:
I've tried a couple variations on `@specialize` (I have swift 6.1.2 so I'm using `@_specialize`) without much success. If I just annotate the `open func visit*` suite of methods it works, but doesn't meaningfully affect instruction count. If I annotate methods that actually do the traversal (dispatch, visit*Impl, walk, etc) the compiler crashes in EagerSpecialization. I can't figure out why it's crashing, but it's happening somewhere in https://github.com/swiftlang/swift/blob/main/lib/SILOptimizer/Transforms/EagerSpecializer.cpp#L104

I suspect it's because I'm trying to specialize a typed throws which is an unintended usecase for specialization. I'm not familiar enough with the inner workings of swiftc to determine where specifically the crash is occurring and why.

I also attempted to do a partial specialization of just `walk` with the hope that the compiler would inline and specialize enough of the remaining functionality to make a difference. This causes a crash in the GenericSpecializer when trying to compile `addSourceFile` in SwiftOperators.

Happy to do more debugging work if you know other things I could try. Otherwise, I think we might have to table this till the specialize crashes can be resolved in swiftc.

ahoppen:
Thanks a lot for giving `@specialize` a try. If you get a chance to file issues for the compiler crashes that you saw at https://github.com/swiftlang/swift, that would be greatly appreciated (one per different crash would be great). Best if you could just attach the state of swift-syntax in which you saw the crash as well as the command you used to build it.

More generally, I am very hesitant to decrease the performance of syntax visitor by more than 10% because it does tend to get used in performance-sensitive code paths. If we can’t achieve that using generics and we need to duplicate the types that would be quite unfortunate and we’d need to check how much overhead to code size the new visitor would add. Anyway, I’d like to hear @rintaro’s thoughts as well.

PR #3112, opened 2025-06-23
Parse @unsafe nonisolated conformance

DougGregor:
A protocol conformance can have a conformance attribute (like `@unsafe`) along with `nonisolated`. Extend the syntax tree to allow type specifiers after the type attributes, and parse `nonisolated` in that position.

Fixes https://github.com/swiftlang/swift-syntax/issues/3109.

### DIFF
--- a/CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxNodesFile.swift
+++ b/CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxNodesFile.swift
@@ -176,7 +176,7 @@ func syntaxNode(nodesStartingWith: [Character]) -> SourceFileSyntax {
           // If needed, this could be added in the future, but for now withUnexpected should be sufficient.
           if let childNode = SYNTAX_NODE_MAP[child.syntaxNodeKind]?.collectionNode,
             !child.isUnexpectedNodes,
-            case .collection(_, collectionElementName: let childElt?, _, _) = child.kind
+             case .collection(_, collectionElementName: let childElt?, _, _, generateDeprecatedAddFunction: true) = child.kind
           {
             let childEltType = childNode.collectionElementType.syntaxBaseName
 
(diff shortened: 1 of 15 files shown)


DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test Windows

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test Windows

CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxNodesFile.swift
```diff
@@ -176,7 +176,8 @@ func syntaxNode(nodesStartingWith: [Character]) -> SourceFileSyntax {
           // If needed, this could be added in the future, but for now withUnexpected should be sufficient.
           if let childNode = SYNTAX_NODE_MAP[child.syntaxNodeKind]?.collectionNode,
             !child.isUnexpectedNodes,
-            case .collection(_, collectionElementName: let childElt?, _, _) = child.kind
+            case .collection(_, collectionElementName: let childElt?, _, _, generateDeprecatedAddFunction: true) = child
+              .kind
```
ahoppen, line 180: Interesting formatting choice by swift-format here. swift-format seems to be happy with the following as well, which I’d prefer

```suggestion
            case .collection(_, collectionElementName: let childElt?, _, _, generateDeprecatedAddFunction: true) = 
              child.kind
```

ahoppen (approved):
Thank you. It’s a little sad that we are not consistent with the ordering of attributes and specifiers in front of types but I assume we can’t do anything about that anymore, can we? That’s probably also why I’m not a huge fan of the name `lateSpecifiers` but I don’t have a better suggestion, it’s about as clear as it gets.

Would like to hear @rintaro’s feedback as well.

CodeGeneration/Sources/generate-swift-syntax/templates/swiftsyntax/SyntaxNodesFile.swift
DougGregor: Formatting fix in https://github.com/swiftlang/swift-syntax/pull/3114 because I too eagerly pushed "merge"

DougGregor:
> Thank you. It’s a little sad that we are not consistent with the ordering of attributes and specifiers in front of types but I assume we can’t do anything about that anymore, can we? That’s probably also why I’m not a huge fan of the name `lateSpecifiers` but I don’t have a better suggestion, it’s about as clear as it gets.
> 
> Would like to hear @rintaro’s feedback as well.

An alternative approach here would be to handle the specifiers different for the places where we can have modifiers, i.e. parameters (where things like `inout` go before the type) and conformances (where `nonisolated` goes after the attributes). I suppose we could revisit the language design here to force specifiers like this to be first, i.e., `nonisolated @preconcurrency`. Parameters are inconsistent with declarations, so conformances could choose which one to align with.

PR #3117, opened 2025-06-29
.swift-format: Specify an exhaustive configuration

AnthonyLatsis:
Not specifying a rule disables checking for that rule rather than assumes the default value, which was not the expected effect.

This applies the default configuration, keeping current settings and disabling 2 default-enabled rules: `DoNotUseSemicolons` and `NoAccessLevelOnExtensionDeclaration`.

### DIFF
--- a/.swift-format
+++ b/.swift-format
@@ -1,20 +1,20 @@
 {
-    "version": 1,
-    "lineLength": 120,
+    "indentConditionalCompilationBlocks": false,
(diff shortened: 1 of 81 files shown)


AnthonyLatsis:
@swift-ci please test

AnthonyLatsis:
@swift-ci please test macOS

bnbarham (approved):

PR #3120, opened 2025-07-01
Add Source Location Support to makeUniqueName for Robust Macro-Generated Identifiers

twhitt14:
This PR improves the uniqueness of macro-generated identifiers by extending the `makeUniqueName` API to accept optional source location information. When provided, the source location (file, line, column) is incorporated into the generated name, ensuring that functions or declarations with the same name at different locations produce unique identifiers.

#### Motivation
Previously, macros such as `@Test` could generate duplicate symbols if two functions had the same name, leading to compiler crashes or invalid redeclarations. This change generalizes the solution by making unique name generation robust for all macros, not just `@Test`.

#### Changes
- Added an overloaded `makeUniqueName(_:sourceLocation:)` method to the `MacroExpansionContext` protocol.
- Updated `BasicMacroExpansionContext` and `PluginMacroExpansionContext` to implement the new method.
- Incorporated a hash of the source location (file, line, column) into the generated name when available.
- Maintained backward compatibility with the original API.

#### Impact
- Macro authors can now generate unique names that are robust to duplicate declarations at different source locations.
- No breaking changes for existing macro code.

Resolves swiftlang/swift#82631

### DIFF
--- a/Sources/SwiftCompilerPluginMessageHandling/PluginMacroExpansionContext.swift
+++ b/Sources/SwiftCompilerPluginMessageHandling/PluginMacroExpansionContext.swift
@@ -281,6 +281,11 @@ class PluginMacroExpansionContext {
 extension PluginMacroExpansionContext: MacroExpansionContext {
   /// Generate a unique name for use in the macro.
   public func makeUniqueName(_ providedName: String) -> TokenSyntax {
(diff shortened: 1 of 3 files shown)


grynspan:
This change is not needed. The bug you're looking to address is a compiler crash. We already have the ability to incorporate source location information into a unique symbol name if we want to do that.

PR #3121, opened 2025-07-02
FixItApplier: Misc improvements

AnthonyLatsis:
* Tests for the overload that takes source edits.
* Optimizations.
* Bug fixes.
* Control over whether to skip duplicate insertions. In `swift package migrate`, this will enable us to prevent repeated fix-it application in various scenarios where multiple elements in a protocol composition type are diagnosed with the same insertion fix-it.

### DIFF
--- a/Sources/SwiftIDEUtils/FixItApplier.swift
+++ b/Sources/SwiftIDEUtils/FixItApplier.swift
@@ -32,7 +32,7 @@ public enum FixItApplier {
   public static func applyFixes(
     from diagnostics: [Diagnostic],
     filterByMessages messages: [String]?,
-    to tree: any SyntaxProtocol
+    to tree: some SyntaxProtocol
   ) -> String {
     let messages = messages ?? diagnostics.compactMap { $0.fixIts.first?.message.message }
 
@@ -54,7 +54,7 @@ public enum FixItApplier {
   /// - Returns: A `String` representation of the modified syntax tree after applying the edits.
   public static func apply(
     edits: [SourceEdit],
-    to tree: any SyntaxProtocol
+    to tree: some SyntaxProtocol
   ) -> String {
     var edits = edits
     var source = tree.description
--- a/Sources/SwiftIDEUtils/FixItApplier.swift
+++ b/Sources/SwiftIDEUtils/FixItApplier.swift
@@ -27,12 +27,15 @@ public enum FixItApplier {
   ///   - filterByMessages: An optional array of message strings to filter which Fix-Its to apply.
   ///     If `nil`, the first Fix-It from each diagnostic is applied.
   ///   - tree: The syntax tree to which the Fix-Its will be applied.
+  ///   - allowDuplicateInsertions: Whether to apply duplicate insertions.
+  ///     Defaults to `true`.
   ///
   /// - Returns: A `String` representation of the modified syntax tree after applying the Fix-Its.
   public static func applyFixes(
     from diagnostics: [Diagnostic],
     filterByMessages messages: [String]?,
-    to tree: some SyntaxProtocol
+    to tree: some SyntaxProtocol,
+    allowDuplicateInsertions: Bool = true
   ) -> String {
     let messages = messages ?? diagnostics.compactMap { $0.fixIts.first?.message.message }
 
@@ -43,7 +46,7 @@ public enum FixItApplier {
       .filter { messages.contains($0.message.message) }
       .flatMap(\.edits)
 
-    return self.apply(edits: edits, to: tree)
+    return self.apply(edits: edits, to: tree, allowDuplicateInsertions: allowDuplicateInsertions)
   }
 
   /// Applies the given edits to the given syntax tree.
@@ -51,10 +54,14 @@ public enum FixItApplier {
   /// - Parameters:
   ///   - edits: The edits to apply.
   ///   - tree: The syntax tree to which the edits should be applied.
+  ///   - allowDuplicateInsertions: Whether to apply duplicate insertions.
+  ///     Defaults to `true`.
+  ///
   /// - Returns: A `String` representation of the modified syntax tree.
   public static func apply(
     edits: [SourceEdit],
-    to tree: some SyntaxProtocol
+    to tree: some SyntaxProtocol,
+    allowDuplicateInsertions: Bool = true
   ) -> String {
     var edits = edits
     var source = tree.description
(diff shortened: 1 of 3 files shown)


AnthonyLatsis:
@swift-ci please test

Sources/SwiftIDEUtils/FixItApplier.swift
```diff
@@ -27,12 +27,15 @@ public enum FixItApplier {
   ///   - filterByMessages: An optional array of message strings to filter which Fix-Its to apply.
   ///     If `nil`, the first Fix-It from each diagnostic is applied.
   ///   - tree: The syntax tree to which the Fix-Its will be applied.
+  ///   - allowDuplicateInsertions: Whether to apply duplicate insertions.
+  ///   Default to `true`.
```
ahoppen:
```suggestion
  ///     Defaults to `true`.
```

Same comment in a couple places below as well

Tests/SwiftIDEUtilsTest/FixItApplierTests.swift
```diff
@@ -0,0 +1,378 @@
...
+
+            Actual output:
+              \"\(actualOutput)\"
+            Expected output:
+              either of \(possibleOutputs)
+            Edits:
+              \(editsPermutation)
+            allowDuplicateInsertions:
+              \(allowDuplicateInsertionsValue)
+            """
```
ahoppen: Could you forward `file:` and `line:` here so that the failure message appears on the source location where `assertAppliedEdits` was called?

Tests/SwiftIDEUtilsTest/FixItApplierTests.swift
```diff
@@ -0,0 +1,378 @@
...
+      outputs: [
+        .either(["_ = 1", "vaar x = 1"])
+      ]
+    )
+  }
+}
+
+private enum Output {
+  case output(String, allowDuplicateInsertions: Bool? = nil)
+  case either([String], allowDuplicateInsertions: Bool? = nil)
```
ahoppen: Why do we need this case? Is `FixItApplier` non-deterministic?

Edit: I see that this is probably because you permute the edits. I think it’s worth adding that as a comment for an explanation. Maybe also worth documenting what `allowDuplicateInsertions: nil` means.

Sources/SwiftIDEUtils/FixItApplier.swift
```diff
@@ -43,51 +46,86 @@ public enum FixItApplier {
...
-      let startIndex = source.utf8.index(source.utf8.startIndex, offsetBy: edit.startUtf8Offset)
-      let endIndex = source.utf8.index(source.utf8.startIndex, offsetBy: edit.endUtf8Offset)
+        func shouldDropRemainingEdit() -> Bool {
+          // Insertions never conflict between themselves, unless we were asked
+          // to drop duplicate insertions.
+          if edit.range.isEmpty && remainingEdit.range.isEmpty {
+            guard allowDuplicateInsertions else {
+              return edit == remainingEdit
+            }
+            return false
```
ahoppen: Maybe it’s just me but I think this reads better as an `if` than a `guard`.

```suggestion
            if allowDuplicateInsertions {
              return false
            }
            return edit == remainingEdit
```

Sources/SwiftIDEUtils/FixItApplier.swift
```diff
@@ -43,51 +46,86 @@ public enum FixItApplier {
...
+          }
 
-      source.replaceSubrange(startIndex..<endIndex, with: edit.replacement)
+          // Anything else that is not disjoint bound-wise will conflict.
+          return edit.endUtf8Offset > remainingEdit.startUtf8Offset
+            && edit.startUtf8Offset < remainingEdit.endUtf8Offset
```
ahoppen, line 114: Could we continue to use `overlaps` here? I feel like that reads better than the `<` comparisons. I feel like that reads more natural.

Sources/SwiftIDEUtils/FixItApplier.swift
```diff
@@ -43,51 +46,86 @@ public enum FixItApplier {
...
+      do {
+        let startIndex = source.utf8.index(source.utf8.startIndex, offsetBy: edit.startUtf8Offset)
+        let endIndex = source.utf8.index(source.utf8.startIndex, offsetBy: edit.endUtf8Offset)
+
+        source.replaceSubrange(startIndex..<endIndex, with: edit.replacement)
+      }
+
+      // Drop any subsequent edits that conflict with one we just applied, and
+      // adjust the range of the rest.
+      while true {
```
ahoppen: Am I missing something or could this be a for loop like

```swift
for adjustmentEditIndex in editIndex..<edits.endIndex
```

Tests/SwiftIDEUtilsTest/FixItApplierTests.swift
```diff
@@ -0,0 +1,378 @@
...
+  precondition(!outputs.isEmpty)
+
+  NEXT_OUTPUT: for output in outputs {
+    let allowDuplicateInsertionsValues =
+      switch output.allowDuplicateInsertions {
+      case .none:
+        [true, false]
+      case .some(let value):
+        [value]
+      }
```
ahoppen: Would using `if let` be more swifty here?

```suggestion
      if let value = output.allowDuplicateInsertions {
        [value]
      } else {
        [true, false]
      }
```

Sources/SwiftIDEUtils/FixItApplier.swift
AnthonyLatsis:
```swift
for adjustmentEditIndex in (editIndex + 1)..<edits.endIndex
```
, yeah. Sounds good.

AnthonyLatsis: I just remembered. I did this because it felt less safe to have 2 index variables in scope in the inner loop.

Sources/SwiftIDEUtils/FixItApplier.swift
AnthonyLatsis: This needs a better comment:
* `Range.overlaps` answers "do they have a common element?".
* We need an answer to "do they have a common element, or is one of them empty and
  `other.start < empty.start < other.end`?", which, interestingly, is equivalent to this condition. 
  
  So `0..<2` also conflicts with `1..<1` even though there is no common element, but not with `0..<0` or `2..<2`.

Tests/SwiftIDEUtilsTest/FixItApplierTests.swift
AnthonyLatsis: > I see that this is probably because you permute the edits.

Yes. Might be worth rewriting `FixItApplier.apply` on top of a pre-sorted array of edits (later).

Sources/SwiftIDEUtils/FixItApplier.swift
AnthonyLatsis: I crammed the increment into the `while` condition but I don’t think you’re going to like this 😅

Sources/SwiftIDEUtils/FixItApplier.swift
AnthonyLatsis: Added comment.

ahoppen: Ah, I see. Thanks for adding the comment

ahoppen (approved):
LGTM, could you add a short RFC post for the new API as described in https://github.com/swiftlang/swift-syntax/blob/main/Contributor%20Documentation/RFC%20Process.md?

AnthonyLatsis:
Thank you, loved the feedback!

> Declarations marked with @_spi are not considered part of the public API and don’t require an RFC

For the new parameter you mean? Even though `FixItApplier` is SPI?

ahoppen:
Sorry, didn’t realize that `FixItApplier` is SPI, just saw the public methods inside. In that case we don’t need an RFC. Good to go 🚀

AnthonyLatsis:
Ah, to be fair, I did add a `Strideable` conformance to `AbsolutePosition`. Let me know if that warrants an RFC 🙂

AnthonyLatsis:
@swift-ci please test

AnthonyLatsis:
@swift-ci please test Windows

PR #3133, opened 2025-08-02
[SwiftParser] Handle nested '#if' in attribute list correctly

rintaro:
Parse the following '#if' directive as an attribute list for the func decl.
  ```
  #if condition
  #if condition
  @attr
  #endif
  #endif
  func foo()
  ```


rintaro:
@swift-ci Please test

PR #2944, opened 2025-01-20
[Syntax] Use no-op PlatformMutex implemtation for wasm32-unknown-wasi

rintaro:
Can be detected with `defined(__wasi__) && !defined(_REENTRANT)`

Fixes https://github.com/kkebo/swift-syntax/issues/145

PR #3135, opened 2025-08-05
[CI] Enable Wasm SDK build

kkebo:
This helps us to prevent regressions like #2944.

(By the way, we can even [run tests](https://github.com/kkebo/swift-syntax/blob/wasm32-wasi-test/.github/workflows/swiftwasm.yml) on a Wasm runtime if we add more changes. However, I don't think it's a good time to do that since swiftlang/github-workflows does not support such a workflow yet.)

### DIFF
--- a/Sources/SwiftLibraryPluginProvider/LibraryPluginProvider.swift
+++ b/Sources/SwiftLibraryPluginProvider/LibraryPluginProvider.swift
@@ -10,6 +10,8 @@
 //
 //===----------------------------------------------------------------------===//
 
+#if !os(WASI)
+
 #if compiler(>=6)
 public import SwiftSyntaxMacros
 @_spi(PluginMessage) public import SwiftCompilerPluginMessageHandling
@@ -221,3 +223,5 @@ extension UnsafeMutableBufferPointer {
   }
 }
 #endif
+
+#endif  // !os(WASI)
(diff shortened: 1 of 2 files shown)


MaxDesiatov:
@swift-ci test

Sources/SwiftLibraryPluginProvider/LibraryPluginProvider.swift
```diff
@@ -10,6 +10,8 @@
 //
 //===----------------------------------------------------------------------===//
 
+#if !os(WASI)
```
MaxDesiatov, line 13: Should the whole file be excluded? Does all of it rely on `dlopen` and dynamic linking or are there any parts that we could keep enabled?

bnbarham: Nothing useful IMO - `LibraryPluginProvider` is all about loading libraries and is only used in `tools/swift-plugin-server` in swift.

bnbarham:
@swift-ci please test

Sources/SwiftLibraryPluginProvider/LibraryPluginProvider.swift
kkebo: I agree with @bnbarham. That's why I did this.

MaxDesiatov:
@swift-ci test windows

PR #3136, opened 2025-08-07
[SwiftParser] Improve attribute parsing and recovering

rintaro:
Attributes followed by non-attribute `#if` were not `atStartOfDeclaration()` nor `parseStatementItem()`, so the rest of the code including the attribute were parsed as "unexpected code".

* Improve attibute list checking in `atStartOfDeclaration()` and parse attributes without decl introducer as `MissingDeclSyntax`
* Make sure `#if` contains only attributes before parsing them as an attribute list element
* Remove `#if` parsing from `parseAttribute()`, and introduce `parseAttributeListElement()` for attribute list element parsing including `#if`. This prevents a crash when  `#if` is passed to `AttributeSyntax.parse(from:)`

https://github.com/swiftlang/swift-syntax/issues/3134

### DIFF
--- a/Sources/SwiftParser/Attributes.swift
+++ b/Sources/SwiftParser/Attributes.swift
@@ -17,18 +17,56 @@
 #endif
 
 extension Parser {
+  private mutating func parseAttributeListElement() -> RawAttributeListSyntax.Element {
+    if self.at(.poundIf) {
+      // 'consumeIfConfigOfAttributes()' check in 'parseAttributeList()' already guarantees
+      // that this '#if' only contains attribute list elements. We don't check
+      // `consumeIfConfigOfAttributes()` here again because it's not only redundant, but it
+      // does *not* work for cases like:
+      //
+      //    #if COND1
+      //       @attr
+      //       #if COND2
+      //       #endif
+      //    #endif
+      //    func fn() {}
+      //
+      // In such cases, the second `#if` is not `consumeIfConfigOfAttributes()`.
+      return .ifConfigDecl(
+        self.parsePoundIfDirective { (parser, _) -> RawAttributeListSyntax.Element in
+          return parser.parseAttributeListElement()
+        } syntax: { parser, attributes in
+          return .attributes(RawAttributeListSyntax(elements: attributes, arena: parser.arena))
+        }
+      )
+    } else {
+      return .attribute(self.parseAttribute())
+    }
+  }
+
   mutating func parseAttributeList() -> RawAttributeListSyntax {
-    guard self.at(.atSign, .poundIf) else {
-      return self.emptyCollection(RawAttributeListSyntax.self)
+    var elements = [RawAttributeListSyntax.Element]()
+
+    func shouldContinue() -> Bool {
+      if self.at(.atSign) {
+        return true
+      }
+      if self.at(.poundIf) && self.withLookahead({ $0.consumeIfConfigOfAttributes() }) {
+        return true
+      }
+      return false
     }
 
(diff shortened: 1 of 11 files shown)


Sources/SwiftParser/Declarations.swift
```diff
@@ -492,11 +498,11 @@ extension Parser {
       arena: self.arena
     )
 
-    if self.at(.atSign), case .attribute(let attribute) = self.parseAttribute() {
+    if self.at(.atSign) {
       return RawUsingDeclSyntax(
         unexpectedBeforeUsingKeyword,
         usingKeyword: usingKeyword,
-        specifier: .attribute(attribute),
+        specifier: .attribute(self.parseAttribute()),
```
rintaro, line 505: This `.attribute(let attribute)` pattern matching was scary 😅 Since we were checking `at(.atSign)`, there was no way it returned non-`RawAttributeSyntax`, but it looked like we were discarding the parsed nodes unless it's `RawAttributeSyntax`.

Tests/SwiftParserTest/DeclarationTests.swift
```diff
@@ -567,11 +567,15 @@ final class DeclarationTests: ParserTestCase {
...
     assertParse(
       """
-      1️⃣private(
+      private(1️⃣
       """,
       diagnostics: [
-        DiagnosticSpec(message: "extraneous code 'private(' at top level")
-      ]
+        DiagnosticSpec(message: "expected 'set)' to end modifier", fixIts: ["insert 'set)'"]),
+        DiagnosticSpec(message: "expected declaration after 'private' modifier", fixIts: ["insert declaration"]),
```
rintaro, line 574: Now decl modifier without decl introducer is also considered `MissingDeclSyntax`.

Tests/SwiftParserTest/AttributeTests.swift
```diff
@@ -1400,4 +1400,17 @@ final class AttributeTests: ParserTestCase {
...
+  func testAttributeParsable() {
+    let source = """
+      #if true
+      @discardableResult
+      #endif
+      """
+    var parser = Parser(source)
+    let attr = AttributeSyntax.parse(from: &parser)
+
+    XCTAssertTrue(attr.description == source)
```
bnbarham: You can pass a closure for parsing to `assertParse`. And rather than the missing check, could check substructure?

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

bnbarham (approved):
As long as we're fixing the recovery that caused the decl/statement switch in a later PR, 👍

rintaro:
@swift-ci Please test Windows

Sources/SwiftParser/Attributes.swift
```diff
@@ -17,18 +17,56 @@
...
+      if self.at(.poundIf) && self.withLookahead({ $0.consumeIfConfigOfAttributes() }) {
+        return true
+      }
+      return false
+    }
+
     var loopProgress = LoopProgressCondition()
-    repeat {
-      let attribute = self.parseAttribute()
+    while self.hasProgressed(&loopProgress) && shouldContinue() {
```
ahoppen, line 61: Nitpick, `hasProgressed` is always the last check. Probably not so relevant here but there are other parts of the parser where a loop iteration doesn’t consume anything and the check is expected to return `false`. If we have the `hasProgressed` check first, we’ll hit an assertion failure where it’s not necessary. So, I think it’s mostly nice to stay consistent about the ordering. Maybe also worth adding this as a comment to `hasProgressed`.

PR #3139, opened 2025-08-13
[SwiftRefactor] PackageManifest: Make a couple of properties external…

xedin:
…ly accessible

In it's current state `AddPackageTarget` does multiple things:

1. Updates the manifest to introduce new targets and dependencies.

2. Generates one or more auxiliary files depending on the type of target being added (i.e. a primary file for the target and sometimes a macro specific defintions file).

We'd like to scope the refactoring action down to manifest updates only.

The callers are going to be resposble for #2 from the list. In order to do that `PackageTarget.sanitizedName` and `PackageTarget.Dependency.name` have to be externally accessible (under `@_spi(PackageRefactor)` still) to form the content of the file.

### DIFF
--- a/Sources/SwiftRefactor/PackageManifest/AddPackageTarget.swift
+++ b/Sources/SwiftRefactor/PackageManifest/AddPackageTarget.swift
@@ -295,18 +295,6 @@ public struct AddPackageTarget: ManifestEditRefactoringProvider {
   }
 }
 
-fileprivate extension PackageTarget.Dependency {
-  /// Retrieve the name of the dependency
-  var name: String {
-    switch self {
-    case .target(let name),
-      .byName(let name),
-      .product(let name, package: _):
-      return name
-    }
-  }
-}
-
 /// The array of auxiliary files that can be added by a package editing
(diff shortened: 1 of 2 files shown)


Sources/SwiftRefactor/PackageManifest/AddPackageTarget.swift
```diff
@@ -346,7 +334,8 @@ fileprivate extension PackageDependency {
   }
 }
 
-fileprivate extension PackageTarget {
+@_spi(PackageRefactor)
+public extension PackageTarget {
```
bnbarham: Any reason to keep this one as an extension and not `name`? I'd probably just move it as well

xedin: Sure, there is `localizedFirstWordCapitalized` but we can deal with that later.

xedin: ✅

xedin:
@swift-ci please test

PR #3142, opened 2025-08-23
[SwiftRefactor] EditRefactoringProvider: Make `textRefactor` and `ref…

xedin:
…actor` requirements throwing

The result of both has been made non-optional as well. The idea here is to be able to produce a description for every failure instead of `nil` result, this is something that package model refactoring actions already do.

The only error we currently have is "not applicable" but that's just a start.

### DIFF
--- a/Sources/SwiftRefactor/CallToTrailingClosures.swift
+++ b/Sources/SwiftRefactor/CallToTrailingClosures.swift
@@ -55,39 +55,43 @@ public struct CallToTrailingClosures: SyntaxRefactoringProvider {
 
   /// Apply the refactoring to a given syntax node. If either a
   /// non-function-like syntax node is passed, or the refactoring fails,
-  /// `nil` is returned.
-  // TODO: Rather than returning nil, we should consider throwing errors with
-  // appropriate messages instead.
+  /// an error is thrown.
   public static func refactor(
     syntax: Syntax,
     in context: Context = Context()
-  ) -> Syntax? {
-    guard let call = syntax.asProtocol(CallLikeSyntax.self) else { return nil }
-    return Syntax(fromProtocol: _refactor(syntax: call, in: context))
+  ) throws -> Syntax {
+    guard let call = syntax.asProtocol(CallLikeSyntax.self) else {
+      throw RefactoringNotApplicableError("not a call")
+    }
+    return try Syntax(fromProtocol: _refactor(syntax: call, in: context))
   }
 
   @available(*, deprecated, message: "Pass a Syntax argument instead of FunctionCallExprSyntax")
   public static func refactor(
     syntax call: FunctionCallExprSyntax,
     in context: Context = Context()
-  ) -> FunctionCallExprSyntax? {
-    _refactor(syntax: call, in: context)
+  ) throws -> FunctionCallExprSyntax {
+    try _refactor(syntax: call, in: context)
   }
 
   internal static func _refactor<C: CallLikeSyntax>(
     syntax call: C,
     in context: Context = Context()
-  ) -> C? {
-    let converted = call.convertToTrailingClosures(from: context.startAtArgument)
-    return converted?.formatted().as(C.self)
+  ) throws -> C {
+    let converted = try call.convertToTrailingClosures(from: context.startAtArgument)
+
+    guard let formatted = converted.formatted().as(C.self) else {
+      throw RefactoringNotApplicableError("format error")
+    }
+
+    return formatted
   }
 }
 
 extension CallLikeSyntax {
-  fileprivate func convertToTrailingClosures(from startAtArgument: Int) -> Self? {
+  fileprivate func convertToTrailingClosures(from startAtArgument: Int) throws -> Self {
     guard trailingClosure == nil, additionalTrailingClosures.isEmpty, leftParen != nil, rightParen != nil else {
-      // Already have trailing closures
-      return nil
+      throw RefactoringNotApplicableError("call already uses trailing closures")
     }
 
     var closures = [(original: LabeledExprSyntax, closure: ClosureExprSyntax)]()
@@ -106,7 +110,7 @@ extension CallLikeSyntax {
     }
 
     guard !closures.isEmpty else {
-      return nil
+      throw RefactoringNotApplicableError("no arguments to convert to closures")
     }
 
     // First trailing closure won't have label/colon. Transfer their trivia.
--- a/Sources/SwiftRefactor/ConvertComputedPropertyToStored.swift
+++ b/Sources/SwiftRefactor/ConvertComputedPropertyToStored.swift
@@ -17,11 +17,15 @@ import SwiftSyntax
 #endif
 
 public struct ConvertComputedPropertyToStored: SyntaxRefactoringProvider {
-  public static func refactor(syntax: VariableDeclSyntax, in context: ()) -> VariableDeclSyntax? {
-    guard syntax.bindings.count == 1, let binding = syntax.bindings.first,
-      let accessorBlock = binding.accessorBlock, case let .getter(body) = accessorBlock.accessors, !body.isEmpty
+  public static func refactor(syntax: VariableDeclSyntax, in context: ()) throws -> VariableDeclSyntax {
+    guard syntax.bindings.count == 1, let binding = syntax.bindings.first else {
+      throw RefactoringNotApplicableError("unsupported variable declaration")
+    }
+
+    guard let accessorBlock = binding.accessorBlock,
+      case let .getter(body) = accessorBlock.accessors, !body.isEmpty
     else {
-      return nil
+      throw RefactoringNotApplicableError("getter is missing or empty")
     }
 
     let refactored = { (initializer: InitializerClauseSyntax) -> VariableDeclSyntax in
@@ -55,7 +59,7 @@ public struct ConvertComputedPropertyToStored: SyntaxRefactoringProvider {
     }
 
     guard body.count == 1, let item = body.first?.item else {
-      return nil
+      throw RefactoringNotApplicableError("getter body is not a single expression")
     }
 
     if let item = item.as(ReturnStmtSyntax.self), let expression = item.expression {
@@ -79,6 +83,6 @@ public struct ConvertComputedPropertyToStored: SyntaxRefactoringProvider {
       )
     }
 
-    return nil
+    throw RefactoringNotApplicableError("could not extract initial value of stored property")
   }
 }
--- a/Sources/SwiftRefactor/ConvertStoredPropertyToComputed.swift
+++ b/Sources/SwiftRefactor/ConvertStoredPropertyToComputed.swift
@@ -17,9 +17,9 @@ import SwiftSyntax
 #endif
 
 public struct ConvertStoredPropertyToComputed: SyntaxRefactoringProvider {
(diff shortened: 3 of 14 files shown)


Sources/SwiftRefactor/CallToTrailingClosures.swift
```diff
@@ -55,31 +55,37 @@ public struct CallToTrailingClosures: SyntaxRefactoringProvider {
 
   /// Apply the refactoring to a given syntax node. If either a
   /// non-function-like syntax node is passed, or the refactoring fails,
-  /// `nil` is returned.
-  // TODO: Rather than returning nil, we should consider throwing errors with
-  // appropriate messages instead.
+  /// error is thrown.
```
ahoppen:
```suggestion
  /// an error is thrown.
```

Sources/SwiftRefactor/CallToTrailingClosures.swift
```diff
@@ -55,31 +55,37 @@ public struct CallToTrailingClosures: SyntaxRefactoringProvider {
...
 
   internal static func _refactor<C: CallLikeSyntax>(
     syntax call: C,
     in context: Context = Context()
-  ) -> C? {
-    let converted = call.convertToTrailingClosures(from: context.startAtArgument)
-    return converted?.formatted().as(C.self)
+  ) throws -> C {
+    guard let converted = call.convertToTrailingClosures(from: context.startAtArgument) else {
+      throw RefactoringNotApplicableError("")
```
ahoppen:
```suggestion
      throw RefactoringNotApplicableError("failed to convert call to trailing closures")
```

Sources/SwiftRefactor/CallToTrailingClosures.swift
```diff
@@ -55,31 +55,37 @@ public struct CallToTrailingClosures: SyntaxRefactoringProvider {
...
+    }
+    if let formatted = converted.formatted().as(C.self) {
+      return formatted
+    }
+
+    throw RefactoringNotApplicableError("")
```
ahoppen:
```suggestion
    throw RefactoringNotApplicableError("failed to format converted expression")
```

Sources/SwiftRefactor/ConvertComputedPropertyToStored.swift
```diff
@@ -79,6 +83,6 @@ public struct ConvertComputedPropertyToStored: SyntaxRefactoringProvider {
       )
     }
 
-    return nil
+    throw RefactoringNotApplicableError("")
```
ahoppen:
```suggestion
    throw RefactoringNotApplicableError("failed to extract initial value of stored variable")
```

Sources/SwiftRefactor/ConvertComputedPropertyToStored.swift
```diff
@@ -55,7 +59,7 @@ public struct ConvertComputedPropertyToStored: SyntaxRefactoringProvider {
     }
 
     guard body.count == 1, let item = body.first?.item else {
-      return nil
+      throw RefactoringNotApplicableError("getter body is not a single-expression")
```
ahoppen:
```suggestion
      throw RefactoringNotApplicableError("getter body is not a single expression")
```

Sources/SwiftRefactor/ConvertStoredPropertyToComputed.swift
```diff
@@ -17,17 +17,19 @@ import SwiftSyntax
...
     }
 
     var codeBlockSyntax: CodeBlockItemListSyntax
 
     if let functionExpression = initializer.value.as(FunctionCallExprSyntax.self),
       let closureExpression = functionExpression.calledExpression.as(ClosureExprSyntax.self)
     {
-      guard functionExpression.arguments.isEmpty else { return nil }
+      guard functionExpression.arguments.isEmpty else {
+        throw RefactoringNotApplicableError("not a nullary call")
```
ahoppen:
```suggestion
        throw RefactoringNotApplicableError("cannot convert stored property whose initializer is a closure that takes arguments")
```

Sources/SwiftRefactor/RefactoringProvider.swift
```diff
@@ -16,6 +16,14 @@ public import SwiftSyntax
 import SwiftSyntax
 #endif
 
+public struct RefactoringNotApplicableError: Error, CustomStringConvertible {
```
ahoppen, line 23:
```suggestion
/// Error that refactoring actions can throw when the refactoring fails.
/// 
/// The reason should start with a single character and should be formatted similarly to compiler error messages.
public struct RefactoringNotApplicableError: Error, CustomStringConvertible {
```

Sources/SwiftRefactor/CallToTrailingClosures.swift
xedin: Sorry, I forgot to commit a change to sync this down to `convertToTrailingClosures` 🤦

Sources/SwiftRefactor/RefactoringProvider.swift
xedin: Thanks! I was thinking whether I should add a comment there or it's self-evident what this is since none other errors declared in swift-syntax do...

Sources/SwiftRefactor/CallToTrailingClosures.swift
xedin: ✅

Sources/SwiftRefactor/CallToTrailingClosures.swift
xedin: ✅

Sources/SwiftRefactor/ConvertComputedPropertyToStored.swift
xedin: ✅

Sources/SwiftRefactor/ConvertComputedPropertyToStored.swift
xedin: ✅

Sources/SwiftRefactor/ConvertStoredPropertyToComputed.swift
xedin: ✅

xedin:
https://github.com/swiftlang/sourcekit-lsp/pull/2251
@swift-ci please test

Sources/SwiftRefactor/CallToTrailingClosures.swift
```diff
@@ -55,39 +55,43 @@ public struct CallToTrailingClosures: SyntaxRefactoringProvider {
...
     syntax call: C,
     in context: Context = Context()
-  ) -> C? {
-    let converted = call.convertToTrailingClosures(from: context.startAtArgument)
-    return converted?.formatted().as(C.self)
+  ) throws -> C {
+    let converted = try call.convertToTrailingClosures(from: context.startAtArgument)
+
+    guard let formatted = converted.formatted().as(C.self) else {
+      throw RefactoringNotApplicableError("cannot cast formatted call to \(C.self)")
```
bnbarham: I'd prefer keeping them more user-focused. This one is a little weird since it suggests more of an internal failure, maybe just "format error"?

Sources/SwiftRefactor/ConvertComputedPropertyToStored.swift
```diff
@@ -79,6 +83,6 @@ public struct ConvertComputedPropertyToStored: SyntaxRefactoringProvider {
       )
     }
 
-    return nil
+    throw RefactoringNotApplicableError("failed to extract initial value of stored variable")
```
bnbarham: The rest are all failures too, how about "could not extract initial value of stored variable"?

Sources/SwiftRefactor/ConvertStoredPropertyToComputed.swift
```diff
@@ -17,17 +17,21 @@ import SwiftSyntax
...
 
     var codeBlockSyntax: CodeBlockItemListSyntax
 
     if let functionExpression = initializer.value.as(FunctionCallExprSyntax.self),
       let closureExpression = functionExpression.calledExpression.as(ClosureExprSyntax.self)
     {
-      guard functionExpression.arguments.isEmpty else { return nil }
+      guard functionExpression.arguments.isEmpty else {
+        throw RefactoringNotApplicableError(
+          "cannot convert stored property whose initializer is a closure that takes arguments"
```
bnbarham: Just "initializer is a closure that takes arguments" seems fine? The "cannot convert stored property" is essentially the error itself

Sources/SwiftRefactor/RefactoringProvider.swift
bnbarham: What does "The reason should start with a single character" mean?

Sources/SwiftRefactor/RefactoringProvider.swift
```diff
@@ -29,10 +40,11 @@ public protocol EditRefactoringProvider {
   /// - Parameters:
   ///   - syntax: The syntax to transform.
   ///   - context: Contextual information used by the refactoring action.
+  /// - Throws: Throws an error if the refactoring action although applicable
+  ///           fails for some other reason.
```
bnbarham: Is this true? Aren't we throwing if it isn't applicable, or did I miss a catch?

Sources/SwiftRefactor/RemoveSeparatorsFromIntegerLiteral.swift
```diff
@@ -31,10 +31,8 @@ import SwiftSyntax
 /// 0xFFFFFFFFF
 /// ```
 public struct RemoveSeparatorsFromIntegerLiteral: SyntaxRefactoringProvider {
-  public static func refactor(syntax lit: IntegerLiteralExprSyntax, in context: Void) -> IntegerLiteralExprSyntax? {
-    guard lit.literal.text.contains("_") else {
-      return lit
-    }
+  public static func refactor(syntax lit: IntegerLiteralExprSyntax, in context: Void) -> IntegerLiteralExprSyntax {
+    guard lit.literal.text.contains("_") else { return lit }
```
bnbarham, line 35: Shouldn't this throw with "literal does not contain '_'"?

xedin: I originally changed it but then reverted, I'd rather fix it in a separate PR.

Sources/SwiftRefactor/CallToTrailingClosures.swift
xedin: Sure

Sources/SwiftRefactor/ConvertComputedPropertyToStored.swift
xedin: Sounds good, I think it should say "property" instead of a "variable" to be consistent with compiler diagnostics.

Sources/SwiftRefactor/RefactoringProvider.swift
xedin: Thanks for catching this! There were a few places left from the time when `Output` was still optional in `refactor`.

xedin:
https://github.com/swiftlang/sourcekit-lsp/pull/2251
@swift-ci please test

bnbarham (approved):

xedin:
https://github.com/swiftlang/sourcekit-lsp/pull/2251

PR #3145, opened 2025-09-01
Test new @_expose(!Cxx) attribute syntax

Xazax-hun:
Seems to work out of the box


Xazax-hun:
@swift-ci please test

ahoppen:
Yeah, this will be parsed correctly as a custom attribute. I personally don’t think we need a test for it. But don’t mind it either.

PR #3146, opened 2025-09-05
[SwiftParser] Update whitespace rule between attribute name and '('

rintaro:
In Swift 6 and later, whitespace is no longer permitted between an attribute name and the opening parenthesis. Update the parser so that in Swift 6+, a '(' without preceding whitespace is always treated as the start of the argument list, while a '(' with preceding whitespace is considered the start of the argument list only when it is unambiguous.

This change makes the following closure be parsed correctly:

  ```swift
  { @MainActor (arg: Int) in ... }
  ```

rdar://147785544

### DIFF
--- a/Sources/SwiftParser/Attributes.swift
+++ b/Sources/SwiftParser/Attributes.swift
@@ -196,28 +196,35 @@ extension Parser {
       atSign = atSign.tokenView.withTokenDiagnostic(tokenDiagnostic: diagnostic, arena: self.arena)
     }
     let attributeName = self.parseAttributeName()
+    let attributeNameHasTrailingSpace = attributeName.raw.trailingTriviaByteLength > 0
+
     let shouldParseArgument: Bool
     switch argumentMode {
     case .required:
       shouldParseArgument = true
     case .customAttribute:
-      shouldParseArgument = self.withLookahead { $0.atAttributeOrSpecifierArgument() }
+      shouldParseArgument = self.withLookahead {
+        $0.atAttributeOrSpecifierArgument(lastTokenHadSpace: attributeNameHasTrailingSpace, forCustomAttribute: true)
+      }
     case .optional:
-      shouldParseArgument = self.at(TokenSpec(.leftParen, allowAtStartOfLine: false))
+      shouldParseArgument = self.withLookahead {
+        $0.atAttributeOrSpecifierArgument(lastTokenHadSpace: attributeNameHasTrailingSpace, forCustomAttribute: false)
+      }
     case .noArgument:
       shouldParseArgument = false
     }
     if shouldParseArgument {
       var (unexpectedBeforeLeftParen, leftParen) = self.expect(TokenSpec(.leftParen, allowAtStartOfLine: false))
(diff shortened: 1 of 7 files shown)


rintaro:
https://github.com/swiftlang/swift/pull/84128
@swift-ci Please test

Sources/SwiftParser/Attributes.swift
```diff
@@ -1074,44 +1081,70 @@ extension Parser {
...
+    guard
+      let numTypeTokens = self.withLookahead({ $0.canParseSimpleType() ? $0.tokensConsumed : nil }),
+      numTypeTokens >= 1
+    else {
       return false
     }
+    // Check if the last token had trailing white spaces.
+    for _ in 0..<numTypeTokens - 1 {
+      self.consumeAnyToken()
+    }
```
hamishknight, line 1143: IMO it would be cleaner to do something like:

```swift
do {
  guard lookahead.canParseSimpleType() else { return false }
  self = lookahead
}
```

Sources/SwiftParser/Lookahead.swift
```diff
@@ -163,40 +163,31 @@ extension Parser.Lookahead {
...
-    // These are keywords that we accept as attribute names.
-    guard self.at(.identifier) || self.at(.keyword(.in), .keyword(.inout)) else {
-      return
-    }
+  /// Skip *any* single attribute. I.e. a type attribute, a decl attribute, or
+  /// a custom attribute.
+  mutating func consumeAnyAttribute() {
+    self.eat(.atSign)
+
+    let nameHadSpace = self.currentToken.trailingTriviaByteLength > 0;
```
hamishknight:
```suggestion
    let nameHadSpace = self.currentToken.trailingTriviaByteLength > 0
```

Sources/SwiftParser/Attributes.swift
rintaro: Hm, I'm not sure what you suggest here.
What it's doing is (1) skip the type or return, and (2) check if the last token of the skipped type had any trailing trivia. I'm not a fan of this, but (2) is needed for checking if `(` is immediately following the attribute name. 

Maybe like this?
```swift
let hasSpace: Bool
do {
  guard lookahead.canParseSimpleType() else {
    return false
  }
  for _ in 0 ..< (lookahead.tokensConsumed - 1) {
    self.consumeAnyToken()
  }
  hasSpace = self.currentToken.trailingTriviaByteLength > 0
  self.consumeAnyToken()
}
```

hamishknight: Ah sorry I misunderstood what this code is doing, what you have currently is fine

rintaro:
https://github.com/swiftlang/swift/pull/84128
@swift-ci Please test

rintaro:
https://github.com/swiftlang/swift/pull/84128
@swift-ci Please test Windows

rintaro:
https://github.com/swiftlang/swift/pull/84128
@swift-ci Please test Windows

PR #3149, opened 2025-09-09
Use an aspect hint instead of the `swift_module` tag

brentleyjones:
This is needed when using rules_swift 2.x+, otherwise inputs aren’t properly tracked and RBE doesn’t work.

Also includes some of the version bumps from https://github.com/bazelbuild/bazel-central-registry/pull/4879.


keith (approved):

keith:
@swift-ci please test

keith:
@swift-ci please test

PR #3150, opened 2025-09-09
Add support for borrow and mutate accessors

meg-gupta:


meg-gupta:
@swift-ci test

Tests/SwiftParserTest/DeclarationTests.swift
```diff
@@ -3473,6 +3473,28 @@ final class DeclarationTests: ParserTestCase {
...
+  func testBorrowAndMutateAccessors() {
+    assertParse(
+      """
+      public class Klass {}
+
+      public struct Wrapper {
+        var _otherK: Klass
+
+        var k1: Klass {
+          borrow {
```
ahoppen, line 3485: I just want to mention that this will be source breaking for the following code, which is currently valid. I’ll leave it up to you to decide whether that source breakage is acceptable.

```swift
func borrow<T>(_ x: () -> T) -> T {
    return x()
}

var x: Int {
    borrow {
        1
    }
}
```

meg-gupta: Thanks for pointing that out. This is currently experimental and is not an immediate problem. We may gate the feature on language version in the future if source breakage becomes a concern.

meg-gupta:
@swift-ci test

meg-gupta:
@swift-ci smoke test

PR #3154, opened 2025-09-26
[SwiftIfConfig] Add StaticBuildConfiguration

DougGregor:
Introduce StaticBuildConfiguration, a data structure that describes all of the aspects of a build configuration that are needed for `#if` evaluation. It is Codable so that it can be stored and replayed later.


Sources/SwiftIfConfig/StaticBuildConfiguration.swift
```diff
@@ -0,0 +1,431 @@
...
+    targetArchitectures: Set<String> = [],
+    targetEnvironments: Set<String> = [],
+    targetRuntimes: Set<String> = [],
+    targetPointerAuthenticationSchemes: Set<String> = [],
+    targetObjectFileFormats: Set<String> = [],
+    targetPointerBitWidth: Int = 64,
+    targetAtomicBitWidths: [Int] = [],
+    endianness: Endianness = .little,
+    languageVersion: VersionTuple = VersionTuple(6),
+    compilerVersion: VersionTuple = VersionTuple(6, 2)
```
bnbarham: I'm not super keen on having more places that we need to update version, does this need to be defaulted here?

DougGregor:
> Is `TestingBuildConfiguration` still needed, or can it be fully subsumed by `StaticBuildConfiguration`?

It's still used to test places where the evaluation can throw errors. I subsumed it where I could.

Sources/SwiftIfConfig/StaticBuildConfiguration.swift
DougGregor: I suppose we don't have to default it here, no

DougGregor: No, we can do without

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

Sources/SwiftCompilerPluginMessageHandling/Macros.swift
```diff
@@ -53,6 +55,7 @@ extension PluginProviderMessageHandler {
     macro: PluginMessage.MacroReference,
     macroRole pluginMacroRole: PluginMessage.MacroRole?,
     discriminator: String,
+    staticBuildConfiguration: StaticBuildConfiguration?,
     expandingSyntax: PluginMessage.Syntax,
     lexicalContext: [PluginMessage.Syntax]?
   ) -> PluginToHostMessage {
```
rintaro, line 72: Could you plumb `staticBuildConfiguration.languageVersion` and `.features` to `SourceManager.add()` -> `ParsedSyntaxRegistry.get()` -> `ParsedSyntaxRegistry.parse()` -> `Parser.init()` so the parsing can be performed with the correct swift version and the experimental feature set.

DougGregor: Great idea! Implemented in https://github.com/swiftlang/swift-syntax/pull/3154/commits/6342a41981112ea5e2b4ce154041f4fa8a6e4cfc

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

Sources/SwiftCompilerPluginMessageHandling/Macros.swift
rintaro: Nice! Thank you!

DougGregor:
@swift-ci please test

Sources/SwiftParser/Parser.swift
```diff
@@ -137,7 +137,7 @@ public struct Parser {
   #endif
 
   /// The Swift version as which source files should be parsed if no Swift version is explicitly specified in the parser.
-  static let defaultSwiftVersion: SwiftVersion = .v6
+  public static let defaultSwiftVersion: SwiftVersion = .v6
```
ahoppen, line 140: Should this be `package` instead of `public`? I don’t see a reason why we need to expose this to clients since all APIs take `nil` to indicate that the default version should be used.

Sources/SwiftSyntaxMacros/MacroExpansionContext.swift
```diff
@@ -70,6 +72,13 @@ public protocol MacroExpansionContext: AnyObject {
   /// This array can be empty if there is no context, for example when a
   /// freestanding macro is used at file scope.
   var lexicalContext: [Syntax] { get }
+
+  /// Returns the build configuration that will be used with the generated
+  /// source code. Macro implementations can use this information to determine
+  /// the context into which they are generating code.
+  ///
+  /// When the build configuration is not known, this returns nil.
```
ahoppen: Would it be worth documenting in which situations the build configuration will be unknown and thus `nil`?

SwiftParserCLI/Sources/swift-parser-cli/Commands/Print.swift
```diff
@@ -0,0 +1,34 @@
...
+  static var configuration = CommandConfiguration(
+    commandName: "print",
+    abstract: "Print the parsed source file after applying any other arguments"
+  )
+
+  @OptionGroup
+  var arguments: ParseArguments
+
+  @Flag(name: .long, help: "Include trivia in the output")
+  var includeTrivia: Bool = false
```
ahoppen: This argument is unused, right?

SwiftParserCLI/Sources/swift-parser-cli/ParseCommand.swift
```diff
@@ -50,4 +59,52 @@ extension ParseCommand {
...
   /// Whether sequence folding using standard operators should be performed
   var foldSequences: Bool { arguments.foldSequences }
+
+  /// Parse the source file, applying any additional configuration options
+  /// such as sequence folding, and provide it to the given closure.
+  func withParsedSourceFile<R>(
+    wantDiagnostics: Bool = true,
+    body: (SourceFileSyntax, [Diagnostic]) throws -> R
+  ) throws -> R {
+    return try sourceFileContents.withUnsafeBufferPointer { sourceBuffer in
```
ahoppen, line 68: Since we intern the source buffer into the syntax arena, I think you could just return `(SourceFileSyntax, [Diagnostic])` here and avoid the trip through the `body` closure. That should make the code on the call side look a little more straight-line, I think.

Sources/SwiftIfConfig/StaticBuildConfiguration.swift
```diff
@@ -0,0 +1,402 @@
...
+  /// Custom build conditions can be set by the `-D` command line option to
+  /// the Swift compiler. For example, `-DDEBUG` sets the custom condition
+  /// named `DEBUG`, which could be checked with, e.g.,
+  ///
+  /// ```swift
+  /// #if DEBUG
+  /// // ...
+  /// #endif
+  /// ```
+  public var customConditions: Set<String> = []
```
ahoppen, line 71: The compiler calls these *conditional compilation flags*. Should the variable be called `conditionalCompilationFlags` or similar to match the terms?

> `-D <value>              Marks a conditional compilation flag as true`

Sources/SwiftIfConfig/StaticBuildConfiguration.swift
```diff
@@ -0,0 +1,402 @@
...
+  ///
+  /// ```swift
+  /// #if hasAttribute(available)
+  /// // ...
+  /// #endif
+  /// ```
+  public var attributes: Set<String> = []
+
+  /// The active target OS names, e.g., "Windows", "iOS".
+  public var targetOSNames: Set<String> = []
```
ahoppen: Out of curiosity: How can this contain multiple target OSs? Wouldn’t you always be building for a single OS? Same for `targetArchitectures` and `targetEnvironments`.

ahoppen: I would call this `targetOSs` since none of the other properties contain the `Names` suffix, eg. it’s not called `targetArchitectureNames`.

ahoppen: I really liked the doc comment that links these properties to the clauses in `#if` that can be used to check them in the previous properties, like on `attributes` because it easily explains what the values mean. Do you think you could add equivalent documentation for all the other properties as well?

Sources/SwiftIfConfig/StaticBuildConfiguration.swift
```diff
@@ -0,0 +1,402 @@
...
+  ///
+  /// This lists all of the bit widths for which the target provides support
+  /// for atomic operations. It can be queried with
+  /// `_hasAtomicBitWidth(_<bitwidth>)`, e.g.
+  ///
+  /// ```swift
+  /// #if _hasAtomicBitWidth(_64)
+  /// // 64-bit atomics are available
+  /// #endif
+  public var targetAtomicBitWidths: [Int] = []
```
ahoppen, line 194: Why is this an `Array` when eg. `targetEnvironments` is a `Set`?

Sources/SwiftIfConfig/StaticBuildConfiguration.swift
```diff
@@ -0,0 +1,402 @@
...
+  public var endianness: Endianness = .little
+
+  /// The effective language version, which can be set by the user (e.g., 5.0).
+  ///
+  /// The language version can be queried with the `swift` directive that checks
+  /// how the supported language version compares, as described by
+  /// [SE-0212](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0212-compiler-version-directive.md). For example:
+  ///
+  /// ```swift
+  /// #if swift(>=5.5)
```
ahoppen: Isn’t this example a little misleading because there is no Swift 5.5 language mode?

Sources/SwiftIfConfig/StaticBuildConfiguration.swift
```diff
@@ -0,0 +1,402 @@
...
+  /// // Hooray, we can use tasks!
+  /// ```
+  public var languageVersion: VersionTuple
```
ahoppen: Shouldn’t this be called `languageMode` to align with https://github.com/swiftlang/swift-evolution/blob/main/proposals/0441-formalize-language-mode-terminology.md?

ahoppen (commented):
Nice improvement. As usual, I have a few nitpicky / API design comments 😉 

Also, should we add a way that allows users of `assertMacroExpansion` to pass in a build configuration?

Sources/SwiftParser/Parser.swift
DougGregor: I needed this exposed because we have a cache key for parsed sources that needs to take it into account. If we don't expose the value, then there will be pointlessly separate cache entries for "no language mode specified" and "the language mode specified match the default," which is unfortunate. Why wouldn't we provide access to our defaults?

Sources/SwiftSyntaxMacros/MacroExpansionContext.swift
DougGregor: ... sure

SwiftParserCLI/Sources/swift-parser-cli/Commands/Print.swift
DougGregor: Yes, copy-pasta.

SwiftParserCLI/Sources/swift-parser-cli/ParseCommand.swift
DougGregor: I had missed that the source buffer always gets interned. Alright, we can simplify this.

Sources/SwiftIfConfig/StaticBuildConfiguration.swift
DougGregor: We already had the "custom condition" terminology in BuildConfiguration. Aligning with that seems more valuable for the clients of this API.

Sources/SwiftIfConfig/StaticBuildConfiguration.swift
DougGregor: There can be aliases. For example, `OSX` and `macOS`.

DougGregor: `targetOSs` reads so weirdly. I guess we can do that.

Sources/SwiftIfConfig/StaticBuildConfiguration.swift
DougGregor: This way, it matches the corresponding requirement in `BuildConfiguration`.

Sources/SwiftIfConfig/StaticBuildConfiguration.swift
DougGregor: Sure, we can use 6.0

Sources/SwiftIfConfig/StaticBuildConfiguration.swift
DougGregor: Sure. We'll still need to have a `languageVersion` for the conformance to `BuildConfiguration`, though.

Sources/SwiftIfConfig/StaticBuildConfiguration.swift
DougGregor: ... alright ...

DougGregor:
> Also, should we add a way that allows users of `assertMacroExpansion` to pass in a build configuration?

Easy enough, sure.

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test Windows

Sources/SwiftParser/Parser.swift
ahoppen: I don’t have any concrete objections, just trying to not unnecessarily increase the API surface area. 

Which cache is this? Did I miss it in this PR or is it in the compiler?

Also, I just realized that this defaults to Swift 6 while the default language mode in the compiler is still Swift 5. Could that cause any issues? If that is the case, wouldn’t it be better if the compiler always explicitly specified a language version.

Sources/SwiftIfConfig/StaticBuildConfiguration.swift
ahoppen: That’s fair. Maybe an unfortunate choice of terminology on `BuildConfiguration` then, but that ship has sailed. 🤷🏽

Sources/SwiftIfConfig/StaticBuildConfiguration.swift
ahoppen: If you have a chance to add that to a doc comment while fixing CI failures, I think that would be useful information to include but it’s not critical.

ahoppen: I thought the same the `Names` suffix also seems inconsistent. If you have a semi-strong opinion on the naming here, I can also be convinced to stick with `targetOSNames`.

Sources/SwiftIfConfig/StaticBuildConfiguration.swift
ahoppen: Ah, forgot about that again. Maybe we need to revise the terminology in `BuildConfiguration` at some point…

DougGregor:
https://github.com/swiftlang/sourcekit-lsp/pull/2312

@swift-ci please test

Sources/SwiftIfConfig/StaticBuildConfiguration.swift
j-f1: Maybe `targetOSes`? `targetOperatingSystems` could also work, although it might feel a bit strange to expand OS in a way that is not done in most settings anymore.

DougGregor: We're already abbreviating `OperatingSystem` to `OS` in the corresponding `BuildConfiguration`, so I don't want to write it out expanded here. `targetOSes` and `targetOSs` are... equally weird to me.

DougGregor:
https://github.com/swiftlang/sourcekit-lsp/pull/2312

@swift-ci please test macOS

PR #3160, opened 2025-10-03
[NFC] Reduce SwiftParser stack usage in debug mode

beccadax:
I’ve had trouble running SwiftParser’s tests—particularly the torture tests in the swiftlang/swift repo—locally in debug mode because some common parser functions would use many kilobytes of stack for each call, consuming the stack too quickly for even a low `maximumNestingLevel` to control. Break up a number of problematic functions to help with this, and also insert a missing stack overflow check in `parseClosureExpression()`.

Each of the functions I've broken up previously used 4-11K of stack and appeared in a backtrace of a test case that crashed because of a stack overflow, either repeatedly or very close to the top of the stack.

I haven't determined if these changes regress release-mode performance; if they do, we probably shouldn't take them.

### DIFF
--- a/Sources/SwiftParser/Expressions.swift
+++ b/Sources/SwiftParser/Expressions.swift
@@ -449,7 +449,7 @@ extension Parser {
     flavor: ExprFlavor,
     pattern: PatternContext
   ) -> RawExprSyntax? {
-    EXPR_PREFIX: switch self.at(anyIn: ExpressionModifierKeyword.self) {
+    switch self.at(anyIn: ExpressionModifierKeyword.self) {
     case (.await, let handle)?:
       let awaitTok = self.eat(handle)
       let sub = self.parseSequenceExpressionElement(
@@ -482,7 +482,7 @@ extension Parser {
       )
     case (.unsafe, let handle)?:
       if !atContextualKeywordPrefixedSyntax(exprFlavor: flavor, acceptClosure: true, preferPostfixExpr: false) {
-        break EXPR_PREFIX
+        break
       }
 
       let unsafeTok = self.eat(handle)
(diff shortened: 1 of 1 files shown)


bnbarham:
Ooo, thanks for looking into this @beccadax!

`swift-parser-cli` (`Sources/swift-parser-cli`) has a `performance-test` (outputs time + instruction count) that you can use to compare before/after your changes. @hamishknight what repo(s) do you use for the weekly performance tests?

beccadax:
I'll wait for Hamish's input, but a quick run on the swift-syntax/Sources directory was actually about 1% faster than main, so that's promising.

beccadax:
@swift-ci please test

beccadax:

hamishknight:
Running the stress tester job on https://github.com/swiftlang/swift/pull/58827 to get performance data

Sources/SwiftParser/Expressions.swift
```diff
@@ -415,21 +418,37 @@ extension Parser {
...
       }
     }
+    return nil
+  }
 
+  private mutating func parseSequenceExpressionElementPrefixIfPresent(
+    flavor: ExprFlavor,
+    pattern: PatternContext
+  ) -> RawExprSyntax? {
     EXPR_PREFIX: switch self.at(anyIn: ExpressionModifierKeyword.self) {
```
hamishknight: While here, could you get rid of this statement label? Seems like was already unnecessary, all the uses of it are directly within the switch (and now with your patch could actually be turned into `return nil`)

Sources/SwiftParser/Expressions.swift
```diff
@@ -1659,71 +1788,64 @@ extension Parser {
...
-            arena: self.arena
-          )
-          if element.isEmpty {
-            break COLLECTION_LOOP
-          } else {
-            elements.append(RawSyntax(element))
-          }
+        let element = elementKind!.makeElement(trailingComma: keepGoing, arena: self.arena)
+        if element.isEmpty {
+          break COLLECTION_LOOP
```
hamishknight: Can we also get rid of this statement label now? (I really dislike statement labels 😄)

beccadax:
@swift-ci smoke test

beccadax:
@swift-ci please test

beccadax:

beccadax:

PR #3165, opened 2025-10-10
chore: restrict GitHub workflow permissions - future-proof

incertum:
See https://github.com/swiftlang/github-workflows/issues/167 for additional context

This approach aligns with security best practices, as detailed in the following documentation:

- https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions
- https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#defining-access-for-the-github_token-scopes
- https://openssf.org/blog/2024/08/12/mitigating-attack-vectors-in-github-workflows/


incertum:
> Thanks a lot for the PR, @incertum 🙏🏽 Two quick questions:
> 
> * Out of curiosity, do you know what the default permissions are if they are not specified? 

The default GITHUB_TOKEN permissions are defined at the repository level. This PR modifies the workflow-level overrides to conform to OpenSSF best practices -> defense in depth.

> * The `automerge` workflow requires `contents: write` permission in the `create_merge_pr` job. Do you know / did you check if it’s possible that a job has more permissions than the surrounding workflow?

I replied to a similar question here: https://github.com/swiftlang/swift-build/pull/858#discussion_r2427266414 Hope it answers the question?

ahoppen:
@swift-ci Please test

PR #3170, opened 2025-10-16
Update syntax for #objectFormat per SE-0492

kubamracek:
See <https://github.com/swiftlang/swift-evolution/blob/main/proposals/0492-section-control.md>, which is now accepted.

### DIFF
--- a/Sources/SwiftIfConfig/BuildConfiguration.swift
+++ b/Sources/SwiftIfConfig/BuildConfiguration.swift
@@ -309,8 +309,7 @@ public protocol BuildConfiguration {
 /// Default implementation of BuildConfiguration, to avoid a revlock with the
 /// swift repo, and breaking clients with the new addition to the protocol.
 extension BuildConfiguration {
-  /// FIXME: This should be @_spi(ExperimentalLanguageFeatures) but cannot due
-  /// to rdar://147943518, https://github.com/swiftlang/swift/issues/80313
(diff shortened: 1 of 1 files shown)


kubamracek:
@swift-ci please test

kubamracek:
@swift-ci please test

Sources/SwiftIfConfig/BuildConfiguration.swift
```diff
@@ -309,7 +311,7 @@ public protocol BuildConfiguration {
 extension BuildConfiguration {
   /// FIXME: This should be @_spi(ExperimentalLanguageFeatures) but cannot due
   /// to rdar://147943518, https://github.com/swiftlang/swift/issues/80313
```
ahoppen: I think this FIXME is no longer relevant now, right?

Also, could you mark this default implementation as deprecated so that clients conforming to `BuildConfiguration` without implementing `isActiveTargetObjectFormat` get a warning, similar to what we do in https://github.com/swiftlang/swift-syntax/blob/ccfe3fcc9db129dce2dca5ab81851e241de5c90f/Sources/SwiftSyntaxMacros/MacroExpansionContext.swift#L103

kubamracek:
@swift-ci please test

kubamracek:

DougGregor (approved):

PR #3171, opened 2025-10-16
[SwiftParse] Update for accessor block disambiguation marker

rintaro:
Update for https://github.com/swiftlang/swift/pull/84972
If the block after a pattern binding initializer starts with `{ @_accessorBlock` it's an accessor block, not a trailing closure.

rdar://140943107

### DIFF
--- a/Sources/SwiftParser/Lookahead.swift
+++ b/Sources/SwiftParser/Lookahead.swift
@@ -274,6 +274,13 @@ extension Parser.Lookahead {
     var lookahead = self.lookahead()
     lookahead.eat(.leftBrace)
 
+    // '@_accessorBlock' is a builtin disambiguation marker.
+    if lookahead.peek(isAt: .identifier),
+      lookahead.peek().tokenText == "_accessorBlock"
+    {
+      return true
+    }
+
     // Eat attributes, if present.
     while lookahead.consume(if: .atSign) != nil {
       guard lookahead.consume(if: .identifier) != nil else {
--- a/Tests/SwiftParserTest/DeclarationTests.swift
+++ b/Tests/SwiftParserTest/DeclarationTests.swift
@@ -3768,4 +3768,41 @@ final class UsingDeclarationTests: ParserTestCase {
       """
     )
   }
+
+  func testAccessorBlockDisambiguationMarker() {
+    assertParse(
+      """
+      var value = initialValue { @_accessorBlock
+        get
+      }
+      """,
+      substructure: VariableDeclSyntax(
+        bindingSpecifier: .keyword(.var),
+        bindings: [
+          PatternBindingSyntax(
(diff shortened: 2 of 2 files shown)


rintaro:
https://github.com/swiftlang/swift/pull/84972
@swift-ci Please test

rintaro:
https://github.com/swiftlang/swift/pull/84972
@swift-ci Please test

Sources/SwiftParser/Lookahead.swift
```diff
@@ -274,6 +274,13 @@ extension Parser.Lookahead {
     var lookahead = self.lookahead()
     lookahead.eat(.leftBrace)
 
+    // '@_accessorBlock' is a builtin disambiguation marker.
+    if lookahead.peek(isAt: .identifier),
+      lookahead.peek().tokenText == "_accessorBlock"
+    {
+      return true;
```
ahoppen:
```suggestion
      return true
```

Sources/SwiftParser/Lookahead.swift
```diff
@@ -274,6 +274,13 @@ extension Parser.Lookahead {
     var lookahead = self.lookahead()
     lookahead.eat(.leftBrace)
 
+    // '@_accessorBlock' is a builtin disambiguation marker.
+    if lookahead.peek(isAt: .identifier),
```
ahoppen, line 278: Don’t you also need to consume the `@` before looking at the identifier? Also, I think a test case would be good.

Also, I think this would probably belong in the ` while lookahead.consume(if: .atSign) != nil {` loop below to make it resilient so `@_accessorBlock` doesn’t necessarily need to be the first attribute.

rintaro: > Don’t you also need to consume the @ before looking at the identifier? 

We know the current token is `@` at line 269 above.

> I think this would probably belong in the  while lookahead.consume(if: .atSign) != nil { loop below to make it resilient so @_accessorBlock doesn’t necessarily need to be the first attribute.

`ASTPrinter` only print `@_accessorBlock` right after `{`. I don't see a reason to put it in the loop

> Also, I think a test case would be good.

Sure, I will add a test case.

ahoppen: > We know the current token is `@` at line 269 above.

Ah, I got a little confused about `nextToken` and then using a fresh `lookahead` here. Opened https://github.com/swiftlang/swift-syntax/pull/3172 to make this slightly easier to read.

> `ASTPrinter` only print `@_accessorBlock` right after `{`. I don't see a reason to put it in the loop

OK, your choice. But I find it odd that order of attributes matters here when it doesn’t in any other cases.

rintaro:
https://github.com/swiftlang/swift/pull/84972
@swift-ci Please test

rintaro:
https://github.com/swiftlang/swift/pull/84972
@swift-ci Please test Windows

PR #3172, opened 2025-10-17
Minor cleanup from `if` to `guard` statement

ahoppen:
I think the guard is a lot easier to read here, noticed while reviewing https://github.com/swiftlang/swift-syntax/pull/3171.

### DIFF
--- a/Sources/SwiftParser/Lookahead.swift
+++ b/Sources/SwiftParser/Lookahead.swift
@@ -266,7 +266,7 @@ extension Parser.Lookahead {
     }
 
     // If we don't have attributes, then it cannot be an accessor block.
-    if nextToken.rawTokenKind != .atSign {
+    guard self.peek(isAt: .atSign) else {
       return false
     }
 
--- a/Tests/SwiftParserTest/DeclarationTests.swift
+++ b/Tests/SwiftParserTest/DeclarationTests.swift
@@ -3831,4 +3831,16 @@ final class UsingDeclarationTests: ParserTestCase {
       )
     )
   }
+
+  func testAccessorBlockAfterPatternBindingDeclWithAttribute() {
+    assertParse(
+      """
+      var x: Int = foo()
+      {
+        @available(*, deprecated)
+        didSet {}
+      }
+      """
+    )
+  }
 }


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test macOS

Sources/SwiftParser/Lookahead.swift
```diff
@@ -266,7 +266,7 @@ extension Parser.Lookahead {
     }
 
     // If we don't have attributes, then it cannot be an accessor block.
-    if nextToken.rawTokenKind != .atSign {
+    guard self.at(.atSign) else {
```
rintaro: Ah sorry, `nextToken` is `peek()`. This is a behavior change.

ahoppen: Great catch, I guess this shows why `at` and `peek(isAt:)` are more explicit. Added a test case for this.

Tests/SwiftParserTest/DeclarationTests.swift
```diff
@@ -3769,40 +3769,15 @@ final class UsingDeclarationTests: ParserTestCase {
...
 
-  func testAccessorBlockDisambiguationMarker() {
+  func testAccessorBlockAfterPatternBindingDeclWithAttribute() {
     assertParse(
       """
-      var value = initialValue { @_accessorBlock
-        get
+      var x: Int = 2
+      {
+        @something
```
ahoppen: @rintaro Do you have an idea what kind of attribute would make sense here, I can’t think of anything.

rintaro: I don't think there's valid attribute for `will/didSet` at this point, at least I couldn't find a test case in swift repo. I believe handling attributes here is mostly for recovery/diagnostics. So `@something` is fine for me, maybe an attribute with argument clause (e.g. `@available(...)`) in addition to that, for extra coverage.

Tests/SwiftParserTest/DeclarationTests.swift
```diff
@@ -3768,4 +3768,16 @@ final class UsingDeclarationTests: ParserTestCase {
       """
     )
   }
+
+  func testAccessorBlockAfterPatternBindingDeclWithAttribute() {
+    assertParse(
+      """
+      var x: Int = 2
```
rintaro: Avoid literal here. We don't try parsing trailing closure for literals
https://github.com/swiftlang/swift-syntax/blob/06f123a9a18757759a4d81a9d6507ab31b9a9c5b/Sources/SwiftParser/Expressions.swift#L936-L937

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test macOS

ahoppen:
@swift-ci Please test macOS

PR #3173, opened 2025-10-18
[Parser] Consolidate recovery codepaths

hamishknight:
We had two different codepaths for recovery, which could result in cases where we attempt to recover, but `expect` is then not able to do so. Consolidate them into a single function.


hamishknight:
@swift-ci please test

hamishknight:
@swift-ci please test

PR #3174, opened 2025-10-20
Add 'SwiftWarningControl' library for source-level lexical warning group behavior settings

artemcm:
This library provides support for a mechanism to control the behavior of specific diagnostic groups for a given declaration's lexical scope with the `@warn` attribute.

The syntax tree and its parser do not reason about warning group controls. The syntax tree produced by the parser reprsents the `@warn` attribute in a generic fashion, as it would any other basic attribute on a declaration. The per-declaration nature of the attribute means that for any given declaration's lexical scope, the behavior of a given diagnosic group can be queried by checking for the presence of this attribute in its parent declaration scope.

The `SwiftWarningControl` library provides a utility to determine, for a given source location and diagnostic group identifier, whether or not its behavior is affected by an `@warn` attribute of any of its parent declaration scope.

This utility relies on constructing a warning control tree data structure to represent all warning control lexical scopes in a given source file. This data structure all of the `@warn` diagnostic group behavior controls within the given syntax node, indicating each group's active behavior at a given position.

For example, given code like the following:
```swift
1:  @warn(Deprecate, as: error)
2:  func foo() {
3:     let a = dep
4:     @warn(Deprecate, as: warning)
5:     func bar() {
6:         let b = dep
7:         @warn(Deprecate, as: ignored)
8:         func baz() {
9:           let c = dep
10:        }
11:        @warn(Deprecate, as: error)
12:        @warn(OtherGroup, as: ignored)
13:        func qux() {
14:            let d = dep
15:            @warn(SomeOtherGroup, as: warning)
16:            func corge() {
17:              let e = dep
18:            }
19:        }
20:     }
21: }
```
the result will be:
```
  - [`Deprecate`:`error`] region within `foo` lexical scope (lines 1-22)
    - [`Deprecate`:`warning`] region within `bar` lexical scope (lines 4-20)
      - [`Deprecate`:`ignored`] region within `baz` lexical scope (lines 7-10)
      - [`Deprecate`:`error`, `OtherGroup`:`ignored`] region within `qux` lexical scope (lines 12-19)
        - [`SomeOtherGroup`:`warning`] region within `corge` lexical scope (lines 15-18)
```

The data structure represents these regions and their nesting relationships as an interval tree where interval nodes can only be nested or disjoint, and where a given interval corresponds to a diagnostic control region for one or more diagnostic group, with a behavior specifier for each.

Intervals cannot partially overlap, and each node's child intervals are always kept sorted. The tree has a single top-level node (root) representing file-level scope.

Once the tree is computed, lookup of a diagnostic group behavior at a given position is performed by recursively descending into the child node containing the given position (located with a binary search of child nodes of a given parent node). Once the position's depth in the interval tree is reached, we walk back the traversal until we find the first containing region which specifies warning behavior control for the given diagnostic group id.


Sources/SwiftWarningControl/WarningControlDeclSyntax.swift
```diff
@@ -0,0 +1,100 @@
...
+}
+
+// Compute a dictionary of all `@warn` diagnostic group behaviors
+// specified on this warning control declaration scope.
+extension WarningControlDeclSyntax {
+  func allWarningGroupControls() -> [String: WarningGroupBehavior] {
+    return attributes.reduce(into: [String: WarningGroupBehavior]()) { result, attr in
+      if case .attribute(let attributeSyntax) = attr,
+         attributeSyntax.attributeName.as(IdentifierTypeSyntax.self)?.name.text == "warn" {
+        if let diagnosticGroupID = attributeSyntax.arguments?.as(LabeledExprListSyntax.self)?.first?.expression.as(DeclReferenceExprSyntax.self)?.baseName.text {
```
artemcm: @rintaro I am sure there is a better way to do all this. Any tips?

artemcm:
@swift-ci test

artemcm:

Sources/SwiftWarningControl/SyntaxProtocol+WarningControl.swift
```diff
@@ -0,0 +1,26 @@
...
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftDiagnostics
+import SwiftSyntax
+
+extension SyntaxProtocol {
+  /// Get the warning emission behavior for the specified diagnostic group
+  /// by determining its containing WarningControlScope, if one is present.
```
ahoppen:
```suggestion
  /// by determining its containing `WarningControlScope`, if one is present.
```

Sources/SwiftWarningControl/WarningControlDeclSyntax.swift
```diff
@@ -0,0 +1,104 @@
...
+protocol WarningControlDeclSyntax {
+  var attributes: AttributeListSyntax { get }
+  var position: AbsolutePosition { get }
+  var endPosition: AbsolutePosition { get }
+}
+
+// Compute a dictionary of all `@warn` diagnostic group behaviors
+// specified on this warning control declaration scope.
+extension WarningControlDeclSyntax {
+  func allWarningGroupControls() -> [String: WarningGroupBehavior] {
```
ahoppen: Should this be a computed property?

Sources/SwiftWarningControl/SyntaxProtocol+WarningControl.swift
```diff
@@ -0,0 +1,26 @@
...
+  @_spi(ExperimentalLanguageFeatures)
+  public func warningGroupBehavior(
+    for diagnosticGroupIdentifier: String
```
ahoppen: Just an idea: What do you think about representing diagnostic group identifiers using a struct wrapping a `String`. That way you can’t accidentally mix up a diagnostic group identifier with some other string and it’s always clear what this string represents. Ie. something like


```swift
struct DiagnosticGroupIdentifier: Hashable, Sendable, ExpressibleByStringLiteral {
  let value: String
}

File: Sources/SwiftWarningControl/WarningControlDeclSyntax.swift [hunk 5]
```diff
@@ -0,0 +1,104 @@
...
+    return attributes.reduce(into: [String: WarningGroupBehavior]()) { result, attr in
+      if case .attribute(let attributeSyntax) = attr,
+        attributeSyntax.attributeName.as(IdentifierTypeSyntax.self)?.name.text == "warn"
+      {
+        if let diagnosticGroupID = attributeSyntax.arguments?.as(LabeledExprListSyntax.self)?.first?.expression.as(
+          DeclReferenceExprSyntax.self
+        )?.baseName.text {
```
ahoppen: Slightly nicer formatting suggestion (also accepted by swift-format)

```suggestion
        if let diagnosticGroupID = attributeSyntax.arguments?.as(LabeledExprListSyntax.self)?.first?.expression
          .as(DeclReferenceExprSyntax.self)?.baseName.text {
```

Sources/SwiftWarningControl/WarningControlDeclSyntax.swift
```diff
@@ -0,0 +1,104 @@
...
+  var position: AbsolutePosition { get }
+  var endPosition: AbsolutePosition { get }
+}
+
+// Compute a dictionary of all `@warn` diagnostic group behaviors
+// specified on this warning control declaration scope.
+extension WarningControlDeclSyntax {
+  func allWarningGroupControls() -> [String: WarningGroupBehavior] {
+    return attributes.reduce(into: [String: WarningGroupBehavior]()) { result, attr in
+      if case .attribute(let attributeSyntax) = attr,
```
ahoppen: You could change all of the `if`s in this function to `guard` statements to avoid some nesting level.

Sources/SwiftWarningControl/WarningControlDeclSyntax.swift
```diff
@@ -0,0 +1,104 @@
...
+        attributeSyntax.attributeName.as(IdentifierTypeSyntax.self)?.name.text == "warn"
+      {
+        if let diagnosticGroupID = attributeSyntax.arguments?.as(LabeledExprListSyntax.self)?.first?.expression.as(
+          DeclReferenceExprSyntax.self
+        )?.baseName.text {
+          if let behaviorText = attr.as(AttributeSyntax.self)?
```
ahoppen: We already know that `attr` is an `AttributeSyntax`, right?

```suggestion
          if let behaviorText = attributeSyntax
```

Sources/SwiftWarningControl/WarningControlDeclSyntax.swift
```diff
@@ -0,0 +1,104 @@
...
+            .arguments?.as(LabeledExprListSyntax.self)?
+            .dropFirst().first?
+            .expression.as(DeclReferenceExprSyntax.self)?
```
ahoppen: Should you check the argument label for `as` here?

Sources/SwiftWarningControl/WarningControlDeclSyntax.swift
```diff
@@ -0,0 +1,104 @@
...
+      }
+    }
+  }
+}
+
+// All syntax nodes which carry attributes and may represent a warning control scope
+
+// @warn(Deprecate, as: ignored)
+// func foo() {...}
+extension FunctionDeclSyntax: WarningControlDeclSyntax {}
```
ahoppen: Could you just use `WithAttributesSyntax` for this instead of introducing a new protocol?

Sources/SwiftWarningControl/WarningControlRegionBuilder.swift
```diff
@@ -0,0 +1,119 @@
...
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
+/// Compute the full set of warning control regions in this syntax node
+extension SyntaxProtocol {
+  @_spi(ExperimentalLanguageFeatures)
+  public func warningGroupControlRegionTree() -> WarningControlRegionTree {
```
ahoppen: Same here, should this be a computed property?

Sources/SwiftWarningControl/WarningControlRegionBuilder.swift
```diff
@@ -0,0 +1,119 @@
...
+    )
+    visitor.walk(self)
+    return visitor.tree
+  }
+}
+
+/// Add this warning control decl syntax node warning group controls (as specified with `@warn`)
+/// to the specified WarningControlRegionTree.
+extension WarningControlDeclSyntax {
+  func addWarningControlRegionsToTree(_ tree: inout WarningControlRegionTree) {
```
ahoppen: Wouldn’t it make more sense if this was a method on `WarningControlRegionTree` that takes `some WarningControlDeclSyntax` as argument?

Sources/SwiftWarningControl/WarningControlRegionBuilder.swift
```diff
@@ -0,0 +1,119 @@
...
+private class WarningControlRegionVisitor: SyntaxVisitor {
+  /// The tree of warning control regions we have found so far
+  var tree: WarningControlRegionTree
+
+  init(start: AbsolutePosition, end: AbsolutePosition) {
+    self.tree = WarningControlRegionTree(start: start, end: end)
+    super.init(viewMode: .fixedUp)
+  }
+
+  override func visit(_ node: FunctionDeclSyntax) -> SyntaxVisitorContinueKind {
```
ahoppen: Instead of adding all these `visit` functions and risking that you forget to add a type that conforms to `WarningControlDeclSyntax`, you can use `SyntaxAnyVisitor` and in `visitAny` check if the node conforms to `WarningControlDeclSyntax` using `node.as(SyntaxProtocol.self) as? WarningControlDeclSyntax` (or `node.asProtocol(WithAttributesSyntax.self)` if you chose to re-use `WithAttributesSyntax` here.

Sources/SwiftWarningControl/WarningControlRegions.swift
```diff
@@ -0,0 +1,313 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
+/// A single warning control region, consisting of a start and end positions,
+/// a diagnostic group identifier, and an emission behavior specifier.
+@_spi(ExperimentalLanguageFeatures)
+public struct WarningControlRegion {
+  let bounds: (start: AbsolutePosition, end: AbsolutePosition)
```
ahoppen: Should this just be a `Range<AbsolutePosition>`. Same for the argument to the initializer and other places where you’re passing a `start` and `end` position.

Sources/SwiftWarningControl/WarningControlRegions.swift
```diff
@@ -0,0 +1,313 @@
...
+/// Once the tree is computed, lookup of a diagnostic group behavior at a given position
+/// is performed by recursively descending into the child node containing
+/// the given position (located with a binary search of child nodes of a given parent node).
+/// Once the position's depth in the interval tree is reached, we walk back the
+/// traversal until we find the first containing region which specifies warning
+/// behavior control for the given diagnostic group id.
+///
+/// TODO: Capture global configuration from command-line arguments
+///       to represent global rules, such as `-Werror`, `-Wwarning`,
+///       and `-suppress-warnings` as *the* root region node.
```
ahoppen, line 103: If you think that we should add this, could you file an issue for this or fix it in this PR? My experience is that `TODO` comments tend to just lie forgotten.

Sources/SwiftWarningControl/WarningControlRegions.swift
```diff
@@ -0,0 +1,313 @@
...
+    parent.children.removeAll { child in absorbedChildren.contains { $0 === child } }
+    for absorbed in absorbedChildren {
+      node.addChild(absorbed)
+    }
+
+    // Add as direct child of parent
+    parent.addChild(node)
+  }
+
+  public func debugDescription() -> String {
```
ahoppen: Shouldn’t you conform `WarningControlRegionTree` to `CustomDebugStringConvertible` and make this a computed property, so this gets picked dup when you run `po` in the debugger for this object?

Tests/SwiftWarningControlTest/WarningControlTests.swift
```diff
@@ -0,0 +1,272 @@
...
+  line: UInt = #line
+) throws {
+  // Pull out the markers that we'll use to dig out nodes to query.
+  let (markerLocations, source) = extractMarkers(markedSource)
+
+  var parser = Parser(source)
+  let tree = SourceFileSyntax.parse(from: &parser)
+
+  let warningControlRegions = tree.warningGroupControlRegionTree()
+  print(warningControlRegions.debugDescription())
```
ahoppen: Leftover print statement?

Tests/SwiftWarningControlTest/WarningControlTests.swift
```diff
@@ -0,0 +1,272 @@
...
+        "1️⃣": .error,
+        "2️⃣": .warning,
+      ]
+    )
+  }
+}
+
+/// Assert that the various marked positions in the source code have the
+/// expected warning behavior controls.
+private func validateWarningGroupControl(
```
ahoppen: We generally call these functions `assert*`


```suggestion
private func assertWarningGroupControl(
```

Tests/SwiftWarningControlTest/WarningControlTests.swift
```diff
@@ -0,0 +1,272 @@
...
+    )
+  }
+
+  func testNestedFunctionWarningGroupControl() throws {
+    try validateWarningGroupControl(
+      """
+      @warn(SomeOtherGroup, as: warning)
+      @warn(GroupID, as: error)
+      @warn(YetAnotherGroup, as: warning)
+      func foo() {
```
ahoppen, line 42: A little nitpicky but should the code in the test cases be indented by 2 spaces to esthetically match the surrounding Swift style?

Sources/SwiftWarningControl/WarningControlRegions.swift
```diff
@@ -0,0 +1,313 @@
...
+  ) -> WarningGroupBehavior? {
+    return rootRegion.innermostContainingRegion(at: location, for: diagnosticGroupIdentifier)?.behavior
+  }
+}
+
+/// A node in the warning control region tree, representing a region and
+/// references to its nested child regions.
+private class WarningControlRegionNode {
+  let bounds: (start: AbsolutePosition, end: AbsolutePosition)
+  var regions: [String: WarningControlRegion] = [:]  // Keyed by diagnostic group identifier
```
ahoppen: This is the kind of comment that wouldn’t be necessary if diagnostic groups had their own type 😉

Sources/SwiftWarningControl/WarningControlRegions.swift
```diff
@@ -0,0 +1,313 @@
...
+  var children: [WarningControlRegionNode] = []
+  weak var parent: WarningControlRegionNode?
```
ahoppen: Do we need to keep track of the `parent` here? I don’t see where it’s used and if it is, aren’t we already traversing the tree top-down, so that we would know the parent by the traversal order?

Sources/SwiftWarningControl/WarningControlRegions.swift
```diff
@@ -0,0 +1,313 @@
...
+  }
+
+  /// Check if this node's bounds match another node's
+  func hasSameBoundaries(with node: WarningControlRegionNode) -> Bool {
+    bounds.start == node.bounds.start && bounds.end == node.bounds.end
+  }
+
+  /// Add a region with the same bounds as this node
+  func addRegion(_ region: WarningControlRegion) {
+    precondition(bounds.start == region.bounds.start && bounds.end == region.bounds.end)
```
ahoppen: Storing the range of the warning control region both in `WarningControlRegionNode` and in `WarningControlRegion` seems a little superfluous to me. Wouldn’t it be easier if you only stored the range in `WarningControlRegionNode` as well as `regions: [String: WarningGroupBehavior]` and then form `WarningControlRegion` from that information when needed? That would also remove the need for these preconditions.

Sources/SwiftWarningControl/WarningControlRegions.swift
```diff
@@ -0,0 +1,313 @@
...
+
+  /// Find the exact child node containing the specified interval node, or `nil` if one is not found.
+  func findChildContaining(_ node: WarningControlRegionNode) -> WarningControlRegionNode? {
+    precondition(contains(node))
+    return findChildContaining(node.bounds.start)
+  }
+
+  /// Binary search to find the exact child containing the specified location,
+  /// or `nil` if one is not found, meaning it is only the `self` node itself that contains the location.
+  private func findChildContaining(_ location: AbsolutePosition) -> WarningControlRegionNode? {
```
ahoppen: General comment: Let’s call these variables `position` instead of `location`. `SourceLocation` is a type that also contains the file name and is line-column based. Let’s try and not mix the two concepts of location and position here.

Sources/SwiftWarningControl/WarningControlRegions.swift
```diff
@@ -0,0 +1,313 @@
...
+    precondition(contains(location))
```
ahoppen: Why not just return `nil` if the location` is outside of the current node. No need to crash IMO.

Sources/SwiftWarningControl/WarningControlRegions.swift
```diff
@@ -0,0 +1,313 @@
...
+    {
+      return nestedResult
+    }
+    // If no child region has this warning group, check self
+    return getRegion(for: diagnosticGroupIdentifier)
+  }
+
+  /// Find the exact child node containing the specified interval node, or `nil` if one is not found.
+  func findChildContaining(_ node: WarningControlRegionNode) -> WarningControlRegionNode? {
+    precondition(contains(node))
```
ahoppen: I don’t think this function matches the description. The comment says that it matches by interval but it only looks up by start location. FWIW, I think we don’t really need this anyway and could just call into `findChildContaining(_:AbsolutePosition)`

Sources/SwiftWarningControl/WarningControlDeclSyntax.swift
artemcm: Changed to computed property, thanks.

Sources/SwiftWarningControl/SyntaxProtocol+WarningControl.swift
artemcm: I like it, I also found it a bit uneasy to just use plain strings around. Changed to use a wrapper structure.

Sources/SwiftWarningControl/WarningControlDeclSyntax.swift
artemcm: Added a label check, thanks.

Sources/SwiftWarningControl/WarningControlDeclSyntax.swift
artemcm: Switched to just using `WithAttributesSyntax` instead of a custom protocol.

Sources/SwiftWarningControl/WarningControlRegionBuilder.swift
artemcm: Changed, thanks.

Sources/SwiftWarningControl/WarningControlRegionBuilder.swift
artemcm: Switched over to a `SyntaxAnyVisitor`, thanks for pointing it out.

Sources/SwiftWarningControl/WarningControlRegions.swift
artemcm: Replaced all `start` and `end` with a single range, thanks for pointing it out.

Sources/SwiftWarningControl/WarningControlRegions.swift
artemcm: Will file an issue. I am definitely intending to do this, just wanted to get the library basics landed and then co-design the next part together with the compiler-side implementation.

Sources/SwiftWarningControl/WarningControlRegions.swift
artemcm: I'd like to keep bounds information in `WarningControlRegion` because it's public API and I think it may be useful. I did remove `regions: WarningControlRegion` from each node, instead keeping around `[DiagnosticGroupIdentifier: WarningGroupBehavior]` and only constructing a `WarningControlRegion` on-demand as a return value for a query. This way there's no duplicate bounds info in the tree nodes, still.

artemcm:
Thanks for the detailed feedback so far, @ahoppen !

artemcm:
@swift-ci test

artemcm:
@swift-ci test

Sources/SwiftWarningControl/WarningControlRegions.swift
ahoppen: Sounds good to me 👍🏽

Sources/SwiftWarningControl/SwiftWarningControl.md
```diff
@@ -0,0 +1,30 @@
...
+    ...
+    @warn(Deprecate, as: warning)
+    func bar() {
+        ...
+        @warn("Deprecate", as: ignored, reason: "Foo")
+        func baz() {
+          ...
+        }
+    }
+}
```
ahoppen, line 23: Another nitpick: Would you indent this by two spaces to match the rest of the repo?

Sources/SwiftWarningControl/SyntaxProtocol+WarningControl.swift
```diff
@@ -0,0 +1,26 @@
...
+import SwiftSyntax
+
+extension SyntaxProtocol {
+  /// Get the warning emission behavior for the specified diagnostic group
+  /// by determining its containing `WarningControlScope`, if one is present.
+  @_spi(ExperimentalLanguageFeatures)
+  public func warningGroupBehavior(
+    for diagnosticGroupIdentifier: DiagnosticGroupIdentifier
+  ) -> WarningGroupBehavior? {
+    let warningControlRegions = root.warningGroupControlRegionTree
```
ahoppen: One thought: Would it make sense to allow a warning group control region tree to be constructed only for a certain source range? Ie. `warningGroupControlRegionTree` becomes a function again that takes an optional `Range<AbsolutePosition>`. It then passes that range to `WarningControlRegionVisitor`, which skips all nodes where `!node.range.overlaps(thatRange)` (or maybe `overlapsOrTouches`, not sure)? That should significantly speed up one-off warning group behavior queries. And then in here you just pass in `self.position..<self.position`

Sources/SwiftWarningControl/WarningControlDeclSyntax.swift
```diff
@@ -0,0 +1,56 @@
...
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
+// Compute a dictionary of all `@warn` diagnostic group behaviors
+// specified on this warning control declaration scope.
+extension WithAttributesSyntax {
+  var allWarningGroupControls: [DiagnosticGroupIdentifier: WarningGroupBehavior] {
```
ahoppen:
```suggestion
extension WithAttributesSyntax {
  /// Compute a dictionary of all `@warn` diagnostic group behaviors
  /// specified on this warning control declaration scope.
  var allWarningGroupControls: [DiagnosticGroupIdentifier: WarningGroupBehavior] {
```

Sources/SwiftWarningControl/WarningControlRegionBuilder.swift
```diff
@@ -0,0 +1,57 @@
...
+
+import SwiftSyntax
+
+/// Compute the full set of warning control regions in this syntax node
+extension SyntaxProtocol {
+  @_spi(ExperimentalLanguageFeatures)
+  public var warningGroupControlRegionTree: WarningControlRegionTree {
+    let visitor = WarningControlRegionVisitor(
+      range: self.range
+    )
```
ahoppen:
```suggestion
    let visitor = WarningControlRegionVisitor(range: self.range)
```

Sources/SwiftWarningControl/WarningControlRegions.swift
```diff
@@ -0,0 +1,329 @@
...
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
+/// A single warning control region, consisting of a start and end positions,
+/// a diagnostic group identifier, and an emission behavior specifier.
+@_spi(ExperimentalLanguageFeatures)
+public struct WarningControlRegion {
```
ahoppen, line 18: This type will be a little useless as public API if none of its members are public 😉

Sources/SwiftWarningControl/WarningControlRegions.swift
```diff
@@ -0,0 +1,329 @@
...
+    behavior: WarningGroupBehavior
+  ) {
+    self.range = range
+    self.diagnosticGroupIdentifier = diagnosticGroupIdentifier
+    self.behavior = behavior
+  }
+}
+
+/// A struct representing a diagnostic group identifier string
+/// @_spi(ExperimentalLanguageFeatures)
```
ahoppen: Did you mean?

```suggestion
@_spi(ExperimentalLanguageFeatures)
```

Sources/SwiftWarningControl/WarningControlRegions.swift
```diff
@@ -0,0 +1,329 @@
...
+public struct DiagnosticGroupIdentifier: Hashable, Sendable, ExpressibleByStringLiteral {
+  public init(stringLiteral value: String) {
+    self.value = value
+  }
+  public init(_ value: String) {
+    self.value = value
+  }
+  let value: String
```
ahoppen: Should we make this public so that people can extract the raw name? And then maybe rename the property to `identifier`?

Sources/SwiftWarningControl/WarningControlRegions.swift
```diff
@@ -0,0 +1,329 @@
...
+/// 15:            @warn(SomeOtherGroup, as: warning)
+/// 16:            func corge() {
+/// 17:              let e = dep
+/// 18:            }
+/// 19:        }
+/// 20:     }
+/// 21: }
+/// 22: func grault() {
+/// 23:   let f = dep
+/// 24: }
```
ahoppen:
```suggestion
/// ```
/// 1:  @warn(Deprecate, as: error)
/// 2:  func foo() {
/// 3:     let a = dep
/// 4:     @warn(Deprecate, as: warning)
/// 5:     func bar() {
/// 6:         let b = dep
/// 7:         @warn(Deprecate, as: ignored)
/// 8:         func baz() {
/// 9:           let c = dep
/// 10:        }
/// 11:        @warn(Deprecate, as: error)
/// 12:        @warn(OtherGroup, as: error)
/// 13:        func qux() {
/// 14:            let d = dep
/// 15:            @warn(SomeOtherGroup, as: warning)
/// 16:            func corge() {
/// 17:              let e = dep
/// 18:            }
/// 19:        }
/// 20:     }
/// 21: }
/// 22: func grault() {
/// 23:   let f = dep
/// 24: }
/// ```
```

So this renders nicely when exporting docc. Also completely nitpicky, but would `|` instead of `:` would look nice and match how we format source code in diagnostics?

Sources/SwiftWarningControl/WarningControlRegions.swift
```diff
@@ -0,0 +1,329 @@
...
+      return
+    }
+
+    // Check if this node should absorb any of the parent's children
+    var absorbedChildren: [WarningControlRegionNode] = []
+    for child in parent.children {
+      if node.contains(child) && !node.hasSameBoundaries(with: child) {
+        absorbedChildren.append(child)
+      }
+    }
```
ahoppen: Can this ever happen? We are traversing the syntax tree top-down so nodes that are visited later should never contain any of the previously visited nodes.

Sources/SwiftWarningControl/WarningControlRegions.swift
```diff
@@ -0,0 +1,329 @@
...
+    _ node: WarningControlRegionNode,
+    parent: WarningControlRegionNode
+  ) {
+    // Check if the new region has the same boundaries as the parent
+    if parent.hasSameBoundaries(with: node) {
+      for (diagnosticGroupIdentifier, control) in node.warningGroupControls {
+        parent.addWarningGroupControl(for: diagnosticGroupIdentifier, behavior: control)
+      }
+      return
+    }
```
ahoppen, line 140: When can this happen? If we have a parent node that had warning controls, then any child node must have a different range because it won’t contain the parent node’s attributes.

Edit: I think the only reason this may happen is because we start off with a `rootRegionNode` that encompasses the entire source file but is empty. If we started of with `rootRegionNode` being `nil`, this shouldn’t happen anymore. It would also remove the need to pass a range to the `WarningControlRegionTree` initializer, which somehow kept confusing me.

Sources/SwiftWarningControl/WarningControlRegions.swift
```diff
@@ -0,0 +1,329 @@
...
+  /// Check if this node's bounds contain a position
+  /// The regions are `[start, end)`, meaning the region's
+  /// starting point is included in the interval, but the end point
+  /// is not. This is done in order to support fully-adjacent regions,
+  /// e.g. two function declarations which are alongside each other with
+  /// a `@warn` attribute on each will result in two warning control regions
+  /// where the first region ends at the exact same `AbsolutePosition` as
+  /// the start of the next region.
+  func contains(_ position: AbsolutePosition) -> Bool {
+    range.lowerBound <= position && position < range.upperBound
```
ahoppen:
```suggestion
    range.contains(position)
```

Sources/SwiftWarningControl/WarningControlRegions.swift
```diff
@@ -0,0 +1,329 @@
...
+  }
+
+  /// Check if this node's bounds contain another node
+  func contains(_ node: WarningControlRegionNode) -> Bool {
+    range.lowerBound <= node.range.lowerBound && node.range.upperBound <= range.upperBound
```
ahoppen:
```suggestion
    range.contains(node.range)
```

Sources/SwiftWarningControl/WarningControlRegions.swift
```diff
@@ -0,0 +1,329 @@
...
+  }
+
+  /// Check if this node's bounds match another node's
+  func hasSameBoundaries(with node: WarningControlRegionNode) -> Bool {
+    range.lowerBound == node.range.lowerBound && range.upperBound == node.range.upperBound
```
ahoppen:
```suggestion
    range == node.range
```

Which kind of makes you wonder if any of these `contains`-style methods are needed or if callers should just access `range` and call whatever they want on it.

Sources/SwiftWarningControl/WarningControlRegions.swift
```diff
@@ -0,0 +1,329 @@
...
+    }
+    return nil
+  }
+}
+
+extension WarningControlRegionNode: Comparable {
+  static func == (lhs: WarningControlRegionNode, rhs: WarningControlRegionNode) -> Bool {
+    return lhs.range.lowerBound == rhs.range.lowerBound && lhs.range.upperBound == rhs.range.upperBound
+  }
+  static func < (lhs: WarningControlRegionNode, rhs: WarningControlRegionNode) -> Bool {
```
ahoppen, line 288:
```suggestion
  }

  static func < (lhs: WarningControlRegionNode, rhs: WarningControlRegionNode) -> Bool {
```

I thought that swift-format would complain about this, but maybe I’m wrong.

Sources/SwiftWarningControl/WarningControlRegions.swift
artemcm: I did, thanks for catching.

Sources/SwiftWarningControl/WarningControlRegions.swift
artemcm: Reformatted with `|`.

Sources/SwiftWarningControl/WarningControlRegions.swift
artemcm: It technically cannot, I was writing the code with more generality in mind, but you're right that it is not necessary. I removed this logic and added a comment to clarify.

Sources/SwiftWarningControl/WarningControlRegions.swift
artemcm: Previously this could happen separately from the root region node because we called `addWarningGroupControl` on the same syntax node multiple times when it has multiple `@warn` attributes. I refactored this to be a single call to only create one node containing all the controls.

Still, I would like to keep the root node spanning the source file as-is because one follow-up change I am planning is to add support for:
```swift
using @warn(DiagGroupID, as: ignored)
```
to define file-scoped controls. This will populate the root node's regions. I think it's also helpful to have the tree hold its full source range in the root node.

Sources/SwiftWarningControl/SyntaxProtocol+WarningControl.swift
artemcm: I added an `internal` `warningGroupControlRegionTree` function which takes an optional position which it uses to skip all nodes which do not contain said position. 

Then the `public` method `warningGroupBehavior()` uses that to get individual query speedup you describe. I'd like to keep the `public` API for `warningGroupControlRegionTree` as-is though because I think it's valuable to keep it simple and have it present a complete tree that clients can then execute individual queries on.

Sources/SwiftWarningControl/WarningControlRegions.swift
artemcm: Good point. I've removed these methods, it is simpler to just have client code use the `range` property directly, indeed.

Sources/SwiftWarningControl/SyntaxProtocol+WarningControl.swift
ahoppen: Should we make the `warningGroupControlRegionTree` computed property a method then? That way, if we decide that we want to expose the `position` parameter publicly, we can do so in almost non-API-breaking way.

Sources/SwiftWarningControl/WarningControlRegionBuilder.swift
```diff
@@ -0,0 +1,63 @@
...
+
+/// Helper class that walks a syntax tree looking for warning behavior control regions.
+private class WarningControlRegionVisitor: SyntaxAnyVisitor {
+  /// The tree of warning control regions we have found so far
+  var tree: WarningControlRegionTree
+  let containingPosition: AbsolutePosition?
+
+  init(_ topLevelRange: Range<AbsolutePosition>, containing position: AbsolutePosition? = nil) {
+    self.tree = WarningControlRegionTree(range: topLevelRange)
+    containingPosition = position;
```
ahoppen:
```suggestion
    containingPosition = position
```

Sources/SwiftWarningControl/WarningControlRegionBuilder.swift
```diff
@@ -0,0 +1,63 @@
...
+    super.init(viewMode: .fixedUp)
+  }
+
+  override func visitAny(_ node: Syntax) -> SyntaxVisitorContinueKind {
+    if let containingPosition,
+      !node.range.contains(containingPosition)
+    {
+      return .visitChildren
```
ahoppen: Did you mean `.skipChildren`? We can skip the entire subtree if the node doesn’t contain the given position. This is where the major speed-up would come from.

Sources/SwiftWarningControl/WarningControlRegions.swift
ahoppen: Ah, OK. With that in mind, this might make sense.

Sources/SwiftWarningControl/WarningControlRegionBuilder.swift
ahoppen: Haha, already fixed it, I saw 😉

Sources/SwiftWarningControl/SyntaxProtocol+WarningControl.swift
artemcm: Yeah, that's reasonable. Done!

artemcm:
@swift-ci test

artemcm:

ahoppen (approved):
Looks good from my side

PR #3175, opened 2025-10-21
Add swiftLanguageModes to the set of labels that occur after package targets

rconnell9:
Add swiftLanguageModes to the set of labels that occur after package targets.

Fixes case where user adds a package dependency for the first time after previously including swiftLanguageMode, eg:

* Create an empty Package and include the swiftLanguagesModes label:
```
// swift-tools-version: 6.3
// The swift-tools-version declares the minimum version of Swift required to build this package.

import PackageDescription

let package = Package(
    name: "tmp",
    swiftLanguageModes: [.v6]
)
```
* swift package describe works as expected:
```
Name: tmp
Manifest display name: tmp
Path: /Users/rconnell/tmp
Tools version: 6.3
Dependencies:
Platforms:
Products:
Targets:
Swift languages versions:
    6

```
* Adding a dependency results in a malformed manifest:
```
 %swift package add dependency https://github.com/apple/swift-syntax.git

% swift package describe
error: 'tmp': Invalid manifest (compiled with: ["...
...
```

### DIFF
--- a/Sources/SwiftRefactor/PackageManifest/AddPackageTarget.swift
+++ b/Sources/SwiftRefactor/PackageManifest/AddPackageTarget.swift
@@ -34,6 +34,7 @@ public struct AddPackageTarget: EditRefactoringProvider {
   /// argument in the Package initializers.
   private static let argumentLabelsAfterTargets: Set<String> = [
     "swiftLanguageVersions",
+    "swiftLanguageModes",
     "cLanguageStandard",
     "cxxLanguageStandard",
   ]


rconnell9:
@swift-ci test

rconnell9:
@swift-ci test macos

dschaefer2:
This is exactly why moving these refactorings to swift-syntax is problematic. They're assuming knowledge of the concepts provided by SwiftPM and can get that wrong, or at the very least not up-to-date as SwiftPM introduces changes. And there are a lot of changes to the manifest coming over the next couple of years.

bnbarham:
> This is exactly why moving these refactorings to swift-syntax is problematic

I'm not sure I really understand the argument here -`swiftLanguageModes` was added well before these were moved. They can be out of date in SwiftPM just as much as they can be out of date in swift-syntax (and that was the case here). The repo they're in doesn't really change anything.

dschaefer2:
This is one example. Over the next couple of years we're planning a lot of additions to the package manifest. It would be a lot easier for those working on SwiftPM to update if these refactorings were in SwiftPM.

PR #3177, opened 2025-10-26
Work around a retroactive conformance bug in Swift 6.2

tshortli:
The Swift 6.2 compiler emits spurious warnings about retroactive conformances to protocols that are inherited through a protocol that is written module qualified in the inheritance clause (the Swift compiler bug is tracked by https://github.com/swiftlang/swift/issues/85153). Suppress the warnings by making the inherited conformances explicit and module qualified.


tshortli:
@swift-ci please test

tshortli:
@swift-ci please test Windows

tshortli:
> Do we have an issue for that bug? If so, could we add a comment that references it and a comment that says that we can remove this workaround once we no longer support Swift 6.2?

Sure, there wasn't an issue but I filed one for tracking purposes and mentioned it by the workarounds.

tshortli:
@swift-ci please test

ahoppen:
@swift-ci Please test

PR #3180, opened 2025-10-30
[SwiftWarningControl] Add ability to query warning group behavior including client-specified enclosing control rules

artemcm:
Adds an optional parameter `with enclosingControls: [DiagnosticGroupIdentifier: WarningGroupBehavior] = [:]` to both:
- `SyntaxProtocol.warningGroupControlRegionTree`
- `SyntaxProtocol.warningGroupBehavior` to specify enclosing global warning group behavior controls which span the entire syntax node.

This mechanism will be used by the compiler to specify global controls over certain diagnostic groups from the compilation configuration. For example, a command-line-specified `-Werror Group` will become an input to this mechanism to be used as *the* behavior control at source positions where no syntactic control (`@warn`) is specified.

With this in place, the `SwiftWarningControl` API can be used as a single query which captures both syntactic and command-line configuration of warning diagnostic group behavior.

### DIFF
--- a/Sources/SwiftWarningControl/SyntaxProtocol+WarningControl.swift
+++ b/Sources/SwiftWarningControl/SyntaxProtocol+WarningControl.swift
@@ -14,13 +14,25 @@ import SwiftDiagnostics
 import SwiftSyntax
 
 extension SyntaxProtocol {
-  /// Get the warning emission behavior for the specified diagnostic group
+  /// Get the warning emission behavior control for the specified diagnostic group
   /// by determining its containing `WarningControlRegion`, if one is present.
+  /// Returns the syntactic control for the given diagnostic group, or `nil` if
+  /// there is not one.
+  /// - Parameters:
+  ///   - for diagnosticGroupIdentifier: The identifier of the diagnostic group.
+  ///   - globalControls: The global controls to consider, specified by the client (compiler)
+  ///     representing module-wide diagnostic group emission configuration, for example
+  ///     with `-Wwarning` and `-Werror` flags. These controls can be overriden at
+  ///     finer-grained scopes with the `@warn` attribute.
   @_spi(ExperimentalLanguageFeatures)
-  public func warningGroupBehavior(
-    for diagnosticGroupIdentifier: DiagnosticGroupIdentifier
-  ) -> WarningGroupBehavior? {
-    let warningControlRegions = root.warningGroupControlRegionTreeImpl(containing: self.position)
-    return warningControlRegions.warningGroupBehavior(at: self.position, for: diagnosticGroupIdentifier)
+  public func warningGroupControl(
+    for diagnosticGroupIdentifier: DiagnosticGroupIdentifier,
+    globalControls: [DiagnosticGroupIdentifier: WarningGroupControl] = [:]
+  ) -> WarningGroupControl? {
+    let warningControlRegions = root.warningGroupControlRegionTreeImpl(
+      globalControls: globalControls,
+      containing: self.position
+    )
+    return warningControlRegions.warningGroupControl(at: self.position, for: diagnosticGroupIdentifier)
   }
 }
--- a/Sources/SwiftWarningControl/WarningControlRegions.swift
+++ b/Sources/SwiftWarningControl/WarningControlRegions.swift
@@ -13,21 +13,21 @@
 import SwiftSyntax
 
 /// A single warning control region, consisting of a start and end positions,
-/// a diagnostic group identifier, and an emission behavior specifier.
+/// a diagnostic group identifier, and an emission behavior control specifier.
 @_spi(ExperimentalLanguageFeatures)
 public struct WarningControlRegion {
   public let range: Range<AbsolutePosition>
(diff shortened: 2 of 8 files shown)


artemcm:
@swift-ci test

artemcm:
@swift-ci test

artemcm:
@swift-ci test

Sources/SwiftWarningControl/SyntaxProtocol+WarningControl.swift
```diff
@@ -18,9 +18,13 @@ extension SyntaxProtocol {
   /// by determining its containing `WarningControlRegion`, if one is present.
   @_spi(ExperimentalLanguageFeatures)
   public func warningGroupBehavior(
-    for diagnosticGroupIdentifier: DiagnosticGroupIdentifier
+    for diagnosticGroupIdentifier: DiagnosticGroupIdentifier,
+    enclosingControls: [DiagnosticGroupIdentifier: WarningGroupBehavior] = [:]
```
ahoppen: Could you explain what `enclosingControls` does in the doc comment?

Also, I feel like there’s a better term than `enclosingControls`. My thoughts are:
- Would `global` be better than `enclosing`. There isn’t really anything that’s enclosing a file’s syntax tree.
- Maybe a little more generally, we seem to use the terms `Behavior` and `Control` somewhat interchangeably in this library. Would it make sense to unify them to either always be `Behavior` or `Control`? Or is there a meaningful distinction between these two terms?

Sources/SwiftWarningControl/WarningControlRegions.swift
```diff
@@ -157,7 +158,8 @@ extension WarningControlRegionTree: CustomDebugStringConvertible {
       let spacing = String(repeating: "  ", count: indent)
       result += "\(spacing)[\(node.range.lowerBound), \(node.range.upperBound)]"
       if !node.warningGroupControls.isEmpty {
-        result += " id(s): \(node.warningGroupControls.keys.map { $0.identifier }.joined(separator: ", "))\n"
+        result +=
+          " control(s): \(node.warningGroupControls.map { $0.key.identifier + ":" + $0.value.rawValue }.joined(separator: ", "))\n"
```
ahoppen: Nitpick, think this will look nicer with a space 😉 
```suggestion
          " control(s): \(node.warningGroupControls.map { $0.key.identifier + ": " + $0.value.rawValue }.joined(separator: ", "))\n"
```

Sources/SwiftWarningControl/WarningControlRegions.swift
```diff
@@ -104,7 +104,7 @@ public struct DiagnosticGroupIdentifier: Hashable, Sendable, ExpressibleByString
 @_spi(ExperimentalLanguageFeatures)
 public struct WarningControlRegionTree {
   /// Root region representing top-level (file) scope
-  private var rootRegionNode: WarningControlRegionNode
+  var rootRegionNode: WarningControlRegionNode
```
ahoppen: Any reason why this needs to be `internal` now?

Sources/SwiftWarningControl/WarningControlRegions.swift
```diff
@@ -185,7 +187,7 @@ extension WarningControlRegionTree {
 
 /// A node in the warning control region tree, representing a collection of warning
 /// group controls and references to its nested child regions.
-private class WarningControlRegionNode {
+class WarningControlRegionNode {
```
ahoppen: Any reason why this can’t be `private` anymore?

Tests/SwiftWarningControlTest/WarningControlTests.swift
```diff
@@ -236,6 +287,7 @@ private func assertWarningGroupControl(
   _ markedSource: String,
   diagnosticGroupID: DiagnosticGroupIdentifier,
   states: [String: WarningGroupBehavior?],
+  enclosingControls: [DiagnosticGroupIdentifier: WarningGroupBehavior] = [:],
```
ahoppen: Would it make sense to move `enclosingControls` after `markedSource`. That way you start by specifying the test input followed by the expected states instead of interleaving them.

Sources/SwiftWarningControl/SyntaxProtocol+WarningControl.swift
artemcm: Added a comment and renamed to `globalControls`. I also went ahead and tried to unify the naming to use the word `Control`, since that's also in the library name. Some of the documentation still talks about diagnostic emission **behavior**, but the library primarily deals with user-specified **controls** of said behavior so I think this terminology makes sense.

Sources/SwiftWarningControl/WarningControlRegions.swift
artemcm: That was from some testing code, restored `private`, thanks for the catch.

artemcm:
@swift-ci test

ahoppen (approved):
LGTM, thank you!

ahoppen:
@swift-ci Please test Windows

PR #3181, opened 2025-10-31
Fix Bazel build

SimplyDanny:
Building with Bazel doesn't work at the moment. That's mainly due to the missing dependencies for `SwiftRefactorTest` and `SwiftSyntaxMacros`. However, there is also an issue with `rules_python` which gets fixed by the update of the Bazel version.

To alleviate the risk of causing an outdated Bazel configuration again, I also added an action that at least builds everything in PRs.


keith:
@swift-ci please test

SimplyDanny:
Probably need to clone the repo in the build job as well ... 🤦‍♂️

SimplyDanny:
Could you please approve the workflow once more, @keith?

keith:
Started

utils/bazel/swift_syntax_library.bzl
```diff
@@ -66,6 +66,6 @@ def swift_syntax_test(name, deps):
         deps = [name + ".library"],
         # Keep in sync with Package.swift
         minimum_os_version = "13.0",
-        tags = ["exclusive"],
+        tags = ["manual", "exclusive"],
```
SimplyDanny: That's to disable `.ios` tests being run on Linux. Is `manual` okay or should another dedicated tag be used instead?

keith: Hmm wr should be able to use some sort of target_compatible_with which would be nicer so it doesn't affect local testing

SimplyDanny: That makes more sense. Better now?

keith: yea, im not sure if that's the smallest possible one or not, but fine either way

keith:
@swift-ci please test

keith:
cc @ahoppen is it ok to add a non-blocking job for this?

ahoppen:
I’m personally fine with running the Bazel tests in GitHub Actions (thanks for adding them) but I’d like to hear @bnbarham’s opinion as well before merging.

bnbarham:
@swift-ci please test

ahoppen:
@swift-ci Please test

PR #3182, opened 2025-11-02
Exclude documentation from build

SimplyDanny:
This otherwise causes the warning

```
    /Users/danny/Projects/swift-syntax/Sources/SwiftWarningControl/SwiftWarningControl.md
```


ahoppen:
@swift-ci Please test

PR #3183, opened 2025-11-03
[SwiftWarningControl] Add ability to specify diagnostic sub-group links

artemcm:
These links establish group inheritance relationships amongst diagnostic groups. This takes form of a new parameter: `subGroupLinks: [DiagnosticGroupIdentifier: [DiagnosticGroupIdentifier]] = [:]` on both:
- `SyntaxProtocol.warningGroupBehavior`
- `SyntaxProtocol.warningGroupControlRegionTree` Where the dictionary key is a super-group and dictionary values are its sub-groups.

Upon encountering a warning group control (`@warn`), its corresponding region is populated with its identifier and behavior, as well as the same corresponding behavior for each of its sub-groups, direct and transitive.

### DIFF
--- a/Sources/SwiftWarningControl/DiagnosticGroupInheritanceTree.swift
+++ b/Sources/SwiftWarningControl/DiagnosticGroupInheritanceTree.swift
@@ -1,0 +1,76 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2025 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// A struct wrapper for a diagnostic group inheritance tree
+/// represented with a dictionary of a group identifier to an array
+/// of its sub-group identifiers.
+@_spi(ExperimentalLanguageFeatures)
+public struct DiagnosticGroupInheritanceTree {
+  private let subGroups: [DiagnosticGroupIdentifier: [DiagnosticGroupIdentifier]]
+  public init(subGroups: [DiagnosticGroupIdentifier: [DiagnosticGroupIdentifier]]) throws {
+    self.subGroups = subGroups
+    if hasCycle() {
+      throw WarningControlError.groupInheritanceCycle
+    }
+  }
+  init() {
+    self.subGroups = [:]
+  }
+
+  /// Diagnostic groups that inherit from `group`.
+  func subgroups(of group: DiagnosticGroupIdentifier) -> [DiagnosticGroupIdentifier] { subGroups[group] ?? [] }
+}
+
+extension DiagnosticGroupInheritanceTree {
+  // Check the subgroup tree for possible cycles
+  func hasCycle() -> Bool {
+    var visited: Set<DiagnosticGroupIdentifier> = []
+    var recursionStack: Set<DiagnosticGroupIdentifier> = []
+    func hasCycleFromGroup(_ group: DiagnosticGroupIdentifier) -> Bool {
+      if visited.insert(group).inserted {
+        recursionStack.insert(group)
+        let subgroups = self.subgroups(of: group)
+        for subGroup in subgroups {
+          if recursionStack.contains(subGroup) {
+            return true
+          } else if !visited.contains(subGroup), hasCycleFromGroup(subGroup) {
+            return true
+          }
+        }
+      }
+      recursionStack.remove(group)
+      return false
+    }
+
+    for group in subGroups.keys {
+      if !visited.contains(group), hasCycleFromGroup(group) {
+        return true
+      }
+    }
+    return false
+  }
+}
+
+/// Describes the kinds of diagnostics that can occur when processing warning
+/// group control queries. This is an Error-conforming type so we can throw errors when
+/// needed.
+@_spi(ExperimentalLanguageFeatures)
+public enum WarningControlError: Error, CustomStringConvertible {
+  case groupInheritanceCycle
+
+  public var description: String {
+    switch self {
+    case .groupInheritanceCycle:
+      return "cycle detected in the warning group inheritance hierarchy"
+    }
+  }
+}
--- a/Sources/SwiftWarningControl/SyntaxProtocol+WarningControl.swift
+++ b/Sources/SwiftWarningControl/SyntaxProtocol+WarningControl.swift
@@ -27,10 +27,12 @@ extension SyntaxProtocol {
   @_spi(ExperimentalLanguageFeatures)
   public func warningGroupControl(
     for diagnosticGroupIdentifier: DiagnosticGroupIdentifier,
-    globalControls: [DiagnosticGroupIdentifier: WarningGroupControl] = [:]
+    globalControls: [DiagnosticGroupIdentifier: WarningGroupControl] = [:],
+    groupInheritanceTree: DiagnosticGroupInheritanceTree? = nil
   ) -> WarningGroupControl? {
     let warningControlRegions = root.warningGroupControlRegionTreeImpl(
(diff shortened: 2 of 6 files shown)


artemcm:
@swift-ci test

Sources/SwiftWarningControl/WarningControlRegions.swift
```diff
@@ -103,8 +103,16 @@ public struct WarningControlRegionTree {
   /// Root region representing top-level (file) scope
   private var rootRegionNode: WarningControlRegionNode
 
-  init(range: Range<AbsolutePosition>) {
-    rootRegionNode = WarningControlRegionNode(range: range)
+  /// Inheritance relationships between diagnostic group identifiers
+  /// Key: super-group, Values: sub-groups
+  let subGroupLinks: [DiagnosticGroupIdentifier: [DiagnosticGroupIdentifier]]
```
ahoppen: You already call this inheritance in the doc comment and I think inheritance describes it better than `subGroupLinks`. It might be worth wrapping the dictionary in a struct so that we can have nice accessors that tell you what the sub and super-groups are. Something like the following

```swift
struct DiagnosticGroupInheritanceTree {
  private let subGroups: [DiagnosticGroupIdentifier: [DiagnosticGroupIdentifier]]

  /// Diagnostic groups that inherit from `group`.
  func subgroups(of group: DiagnosticGroupIdentifier) -> [DiagnosticGroupIdentifier] { subGroups[group] }
  ...
}
```

That way we might also add a static member `default` at some point if we want to, similar to how we have `standardOperators` in `SwiftOperators`.

artemcm: Re-named to `groupInheritanceTree` and added the wrapper struct, which I think is a good idea. Thanks.

artemcm:
> How large do you envision `subGroupLinks` to be? Just thinking if it would be more efficient to take the sub group relationships into account when querying for warning controls instead of when populating the tree. Kind of depends on whether you expect to have more queries or more `@warn` controls.

I considered both options and went with this one to take the relationships into account at tree construction: my main reason for this is that I expect the primary client use-case (such as the compiler in https://github.com/swiftlang/swift/pull/85036) to pay the cost of constructing a tree and then run queries against said tree. I do expect there to be a lot of queries against a constructed tree, and it's tricky to predict how many group relationships we will end up with since that concept is still quite new. 

Baking these relationships into the constructed tree does also result in simpler query logic on the client side.

artemcm:
@swift-ci test

Sources/SwiftWarningControl/SyntaxProtocol+WarningControl.swift
```diff
@@ -27,10 +27,12 @@ extension SyntaxProtocol {
   @_spi(ExperimentalLanguageFeatures)
   public func warningGroupControl(
     for diagnosticGroupIdentifier: DiagnosticGroupIdentifier,
-    globalControls: [DiagnosticGroupIdentifier: WarningGroupControl] = [:]
+    globalControls: [DiagnosticGroupIdentifier: WarningGroupControl] = [:],
+    groupInheritanceTree: [DiagnosticGroupIdentifier: [DiagnosticGroupIdentifier]] = [:]
```
ahoppen: Shouldn’t this also take a DiagnosticGroupInheritanceTree? I’d make DiagnosticGroupInheritanceTree part of the SPI.

Same for all the other functions that take a dictionary.

Sources/SwiftWarningControl/WarningControlRegions.swift
```diff
@@ -103,8 +103,16 @@ public struct WarningControlRegionTree {
   /// Root region representing top-level (file) scope
   private var rootRegionNode: WarningControlRegionNode
 
-  init(range: Range<AbsolutePosition>) {
-    rootRegionNode = WarningControlRegionNode(range: range)
+  /// Inheritance relationships between diagnostic group identifiers
+  /// Key: super-group, Values: sub-groups
```
ahoppen: Comment is now outdated.

Sources/SwiftWarningControl/WarningControlRegions.swift
```diff
@@ -115,7 +123,14 @@ public struct WarningControlRegionTree {
     guard !controls.isEmpty else { return }
     let newNode = WarningControlRegionNode(range: range)
     for (diagnosticGroupIdentifier, control) in controls {
-      newNode.addWarningGroupControl(for: diagnosticGroupIdentifier, control: control)
+      var groups: [DiagnosticGroupIdentifier] = [diagnosticGroupIdentifier]
```
ahoppen, line 127: What if we have the following sub-groups?

```
A: B, C
B: C
```

Then we’d report `C` twice if you have a warning control for `A`. When using a `Set`, we need to think about whether we want to guarantee a deterministic output ordering of the diagnostic groups (which I think we should), so we’d need to sort somewhere. So maybe it would be easier to avoid duplicates in an `Array`.

Also, since the inheritance tree is user-provided, I think we need to think about cycles somewhere. Either make it the responsibility of `DiagnosticGroupInheritanceTree` to ensure there are no cycles (and throw if there are), or try and find cycles here. I’d prefer the former option.

artemcm: That's a great catch, thank you. 
I added code to keep track of already-added groups for a given control to ensure we don't add duplicates to this array.
I also added a cycle check in the initializer of a `DiagnosticGroupInheritanceTree` and factored it out into a separate file `DiagnosticGroupInheritanceTree.swift`. It will throw an error if initialized with a subgroup dictionary which contains cycles.

artemcm:
@swift-ci test

Sources/SwiftWarningControl/DiagnosticGroupInheritanceTree.swift
```diff
@@ -0,0 +1,85 @@
...
+}
+
+extension DiagnosticGroupInheritanceTree {
+  // Check the subgroup tree for possible cycles
+  func hasCycle() -> Bool {
+    var visited: Set<DiagnosticGroupIdentifier> = []
+    var recursionStack: Set<DiagnosticGroupIdentifier> = []
+    func hasCycleFromGroup(_ group: DiagnosticGroupIdentifier) -> Bool {
+      if !visited.contains(group) {
+        visited.insert(group);
```
ahoppen: Slightly more efficient

```suggestion
      if visited.insert(group).inserted {
```

Sources/SwiftWarningControl/DiagnosticGroupInheritanceTree.swift
```diff
@@ -0,0 +1,85 @@
...
+            cycleFound = true
+          } else if !visited.contains(subGroup),
+            hasCycleFromGroup(subGroup)
+          {
+            cycleFound = true
+          }
+        }
+        if cycleFound {
+          return true
+        }
```
ahoppen: I don’t think we need the `cycleFound` variable.

```suggestion
        let subgroups = self.subgroups(of: group)
        for subGroup in subgroups {
          if recursionStack.contains(subGroup) {
            return true
          } else if !visited.contains(subGroup),
            hasCycleFromGroup(subGroup)
          {
            return true
          }
        }
```

Sources/SwiftWarningControl/DiagnosticGroupInheritanceTree.swift
```diff
@@ -0,0 +1,85 @@
...
+      }
+      recursionStack.remove(group)
+      return false
+    }
+
+    for group in subGroups.keys {
+      if !visited.contains(group),
+        hasCycleFromGroup(group)
+      {
```
ahoppen: Nitpick, fits on one line

```suggestion
      if !visited.contains(group), hasCycleFromGroup(group) {
```

Sources/SwiftWarningControl/DiagnosticGroupInheritanceTree.swift
artemcm: This is neat. ✅

artemcm:
@swift-ci test

artemcm:

artemcm:

PR #3186, opened 2025-11-10
Wrap the characters mentioned in the descriptions of `TriviaPieces` in backticks

### LINKED ISSUE #3185: Documentation describes TriviaPiece.blockComment incorrectly (Markdown escaping issue)
### Description

The rendered documentation for [`TriviaPiece.blockComment`](https://swiftpackageindex.com/swiftlang/swift-syntax/main/documentation/swiftsyntax/triviapiece/blockcomment(_:)) looks like this:

> A developer block comment, starting with ‘/<em>’ and ending with ’</em>/’.

Screenshot:

Image

Note that the documentation shows the wrong tokens "/" and "/" for a block comment and renders the text in between in italics. The reason is apparently that the DocC Markdown renderer (correctly, from its perspective) interprets the asterisks in `/*` and `*/` as emphasis markers.

The fix would be to change the doc comment string [in the `let TRIVIAS` array](https://github.com/swiftlang/swift-syntax/blob/02f8220bccfa03df68cc50631e11891c7ebcca0c/CodeGeneration/Sources/SyntaxSupport/Trivia.swift#L100):

```swift
public let TRIVIAS: [Trivia] = [
  …
  Trivia(
    name: "BlockComment",
    comment: #"A developer block comment, starting with '/*' and ending with '*/'."#,
    traits: [.comment]
  ),
  …
```

The comment should either use code voice (backticks) for the tokens or escape the asterisks with backslashes (which is the official Markdown way to escape special characters). I'd do this myself, but I thought I'd post here first because I don't know if such a change would interfere with other places where the comment field is used as plain text where Markdown formatting or escaping would be a problem.

### Steps to Reproduce

_No response_

ahoppen:
Fixes #3185

### DIFF
--- a/CodeGeneration/Sources/SyntaxSupport/Trivia.swift
+++ b/CodeGeneration/Sources/SyntaxSupport/Trivia.swift
@@ -97,73 +97,73 @@ public let TRIVIAS: [Trivia] = [
 
   Trivia(
     name: "BlockComment",
-    comment: #"A developer block comment, starting with '/*' and ending with '*/'."#,
+    comment: #"A developer block comment, starting with `/*` and ending with `*/`."#,
     traits: [.comment]
   ),
 
   Trivia(
     name: "CarriageReturn",
-    comment: #"A newline '\r' character."#,
+    comment: #"A newline `\r` character."#,
     characters: "\r",
     traits: [.whitespace, .newline]
   ),
 
   Trivia(
     name: "CarriageReturnLineFeed",
-    comment: #"A newline consists of contiguous '\r' and '\n' characters."#,
+    comment: #"A newline consists of contiguous `\r` and `\n` characters."#,
     characters: "\r\n",
     traits: [.whitespace, .newline]
   ),
 
   Trivia(
     name: "DocBlockComment",
-    comment: #"A documentation block comment, starting with '/**' and ending with '*/'."#,
+    comment: #"A documentation block comment, starting with `/**` and ending with `*/`."#,
     traits: [.comment]
   ),
 
   Trivia(
     name: "DocLineComment",
-    comment: #"A documentation line comment, starting with '///' and excluding the trailing newline."#,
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

CodeGeneration/Sources/SyntaxSupport/Trivia.swift
```diff
@@ -97,73 +97,73 @@ public let TRIVIAS: [Trivia] = [
...
 
   Trivia(
     name: "DocLineComment",
-    comment: #"A documentation line comment, starting with '///' and excluding the trailing newline."#,
+    comment: #"A documentation line comment, starting with `///` and excluding the trailing newline."#,
     traits: [.comment]
   ),
 
-  // Swift don't support form feed '\f' so we use the raw unicode
+  // Swift don`t support form feed `\f` so we use the raw unicode
```
ole:
```suggestion
  // Swift doesn't support form feed `\f` so we use the raw unicode
```

CodeGeneration/Sources/SyntaxSupport/Trivia.swift
```diff
@@ -173,10 +173,10 @@ public let TRIVIAS: [Trivia] = [
     comment: #"Any skipped unexpected text."#
   ),
 
-  // Swift don't support vertical tab '\v' so we use the raw unicode
+  // Swift don`t support vertical tab `\v` so we use the raw unicode
```
ole:
```suggestion
  // Swift doesn't support vertical tab `\v` so we use the raw unicode
```

ole (approved):

CodeGeneration/Sources/SyntaxSupport/Trivia.swift
ahoppen: Oops, good catch, thank you @ole 🙏🏽

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PR #3187, opened 2025-11-13
[NFC] Avoid use of unavailable range 'contains' method

artemcm:
This method appears to be missing in the host standard library on some CI platforms, as it is sufficiently new:
https://github.com/swiftlang/swift/pull/76891

e.g. https://ci.swift.org/job/swift-PR-macos/25670/, https://ci.swift.org/job/swift-PR-Linux/24824/

### DIFF
--- a/Sources/SwiftWarningControl/WarningControlRegions.swift
+++ b/Sources/SwiftWarningControl/WarningControlRegions.swift
@@ -254,7 +254,7 @@ private class WarningControlRegionNode {
 
   /// Add a child region that is directly nested within this region
   func addChild(_ node: WarningControlRegionNode) {
-    precondition(range.contains(node.range))
+    precondition(range.lowerBound <= node.range.lowerBound && node.range.upperBound <= range.upperBound)
     children.append(node)
     children.sort()
   }
--- a/Tests/SwiftWarningControlTest/WarningControlTests.swift
+++ b/Tests/SwiftWarningControlTest/WarningControlTests.swift
@@ -386,8 +386,6 @@ private func assertWarningGroupControl(
       groupInheritanceTree: groupInheritanceTree
     )
 
-    print(warningControlRegions.debugDescription)
-
     let groupControl = token.warningGroupControl(
       for: diagnosticGroupID,
       globalControls: globalControls,


artemcm:
@swift-ci test

Sources/SwiftWarningControl/WarningControlRegions.swift
```diff
@@ -254,7 +254,7 @@ private class WarningControlRegionNode {
 
   /// Add a child region that is directly nested within this region
   func addChild(_ node: WarningControlRegionNode) {
-    precondition(range.contains(node.range))
+    precondition(range.lowerBound <= node.range.lowerBound && range.upperBound >= node.range.upperBound)
```
ahoppen: `Range` is a half open range, `ClosedRange` is a closed range. So this should be `>` on the upper bound. Also, I personally like to have `range.<something>`  on the outside of the comparisons, just feel like it reads more nicely if the think you’re checking for is in the middle.


```suggestion
    precondition(range.lowerBound <= node.range.lowerBound && node.range.upperBound < range.upperBound)
```

Tests/SwiftWarningControlTest/WarningControlTests.swift
```diff
@@ -386,8 +386,6 @@ private func assertWarningGroupControl(
       groupInheritanceTree: groupInheritanceTree
     )
 
-    print(warningControlRegions.debugDescription)
```
ahoppen, line 389: Ooops, good catch. Thanks.

Sources/SwiftWarningControl/WarningControlRegions.swift
artemcm: I took this straight from the implementation here: 
https://github.com/swiftlang/swift/blob/0e6c6bcfc185c4e6601ae9c8e513bd34f794835f/stdlib/public/core/Range.swift#L1079
I believe the semantic here is that the half-open range contains all elements **up to** the upper bound, so if the argument to `contains` has the same upper bound, then technically both ranges include the same `upperBound-1` element.

artemcm:
@swift-ci test

artemcm:

artemcm:

Sources/SwiftWarningControl/WarningControlRegions.swift
ahoppen: I thought this was contains of another element not contains of another range. Should have read more clearly. 🤦🏽

ahoppen (approved):
Thanks

PR #3189, opened 2025-11-15
[SE-0474] Support `yielding borrow` and `yielding mutate` accessors

tbkka:
`yielding borrow` and `yielding mutate` are the final approved names for the accessors previously known as `read` and `modify` (which are themselves improvements over the earlier non-standard `_read` and `_modify` accessors).

Since some people are already using these under the `read` and `modify` names, we want to support the old and new names as synonyms for a transition period until everyone has had a chance to migrate.

RFC discussion: https://forums.swift.org/t/rfc-multiple-modifiers-in-accessordeclsyntax-for-se-0474/83316

### DIFF
--- a/Sources/SwiftSyntax/SwiftSyntaxCompatibility.swift
+++ b/Sources/SwiftSyntax/SwiftSyntaxCompatibility.swift
@@ -17,7 +17,11 @@ extension AccessorDeclSyntax {
   // Note: Does _not_ need @_disfavoredOverload because `modifier:`
   // does not have a default value.  This will _only_ get used when
   // someone explicitly provides a legacy `modifier:` argument.
-  @available(*, deprecated, message: "use `modifiers:` instead of `modifier:`")
+  @available(
+    *,
+    deprecated,
+    message: "use `modifiers:` instead of `modifier:`"
+  )
   public init(
     leadingTrivia: Trivia? = nil,
     _ unexpectedBeforeAttributes: UnexpectedNodesSyntax? = nil,
@@ -54,7 +58,11 @@ extension AccessorDeclSyntax {
     )
   }
 
-  @available(*, deprecated, message: "use `modifiers` to see all modifiers on an accessor declaration; `modifier` only returns the first one")
+  @available(
+    *,
+    deprecated,
+    message: "use `modifiers` to see all modifiers on an accessor declaration; `modifier` only returns the first one"
+  )
   public var modifier: DeclModifierSyntax? {
     get {
       // The legacy `modifier` API predates `yielding`


tbkka:
This isn't working out so well.   There seem to be two basic problems:
* Treating `yielding borrow` as a synonym for `read`
* Having `yielding borrow` be a "keyword" that's really two words

tbkka:
It's not clear from SE-0474 whether `yielding mutating borrow` should be accepted, or only `mutating yielding borrow`.   @DougGregor @rjmccall -- Any insights here?

If the former is accepted, then it might make sense to model `yielding` as a modifier here; otherwise, I'll need to do something more complicated.

tbkka:
Can anyone suggest how this should work?   @bnbarham @rintaro @hamishknight ??   I'm completely stumped.  What I have here seems entirely wrong, but I have no idea what approach should be used here.  In particular:
* Mapping `yielding borrow` to `read` breaks round-tripping, so it seems like I somehow need to model `yielding borrow` as a distinct construct from `read`
* It seems like `yielding borrow` must be modeled as two Nodes?  Otherwise, trivia between those tokens can't be modeled?
* `yielding` can't be a "modifier" without other changes, since `mutating yielding borrow` is legal, and that would have two modifiers, which isn't supported in the current accessor modeling.  So either accessors need to support a set of modifiers?  Or something different?

Sources/SwiftParser/Declarations.swift
```diff
@@ -1692,11 +1697,28 @@ extension Parser {
       modifier = nil
     }
 
+    if yielding {
+      _ = self.expect(TokenSpec(.yielding))
+    }
+
     let (unexpectedBeforeIntroducer, introducer) = self.expect(kind.spec)
+
+    // Map `yielding borrow` => `read`, etc.
```
tbkka: This breaks round-tripping, so I can't do this mapping here.

tbkka:
Maybe I could expand `AccessorIntroducer` with a new `yielding` field like this?
```
  struct AccessorIntroducer {
    var attributes: RawAttributeListSyntax
    var modifier: RawDeclModifierSyntax?
    var yielding: RawAccessorYieldingSyntax?
    var kind: AccessorDeclSyntax.AccessorSpecifierOptions
    var unexpectedBeforeToken: RawUnexpectedNodesSyntax?
    var token: RawTokenSyntax
  }
```
and then I'd have to reverse-engineer the code generator to construct `RawAccessorYieldingSyntax` and update everywhere an `AccessorIntroducer` was used to properly handle the new field?   Does that make any sense?

tbkka:
It would be so helpful if the generated code included comments indicating where the data came from for each particular generated type and/or file.

bnbarham:
> It's not clear from SE-0474 whether yielding mutating borrow should be accepted, or only mutating yielding borrow. @DougGregor @rjmccall -- Any insights here?

I haven't been following the proposal closely, but from a quick read, it has eg.
> Like a set accessor, a yielding mutate accessor in a struct or enum property is mutating by default (unless explicitly declared a nonmutating yielding mutate)

Confusing set of terms IMO (nonmutating mutate?), but `set` accessors allow specifying `mutating` (despite being redundant), so seems reasonable to allow here.

> yielding can't be a "modifier" without other changes, since mutating yielding borrow is legal, and that would have two modifiers, which isn't supported in the current accessor modeling. So either accessors need to support a set of modifiers? Or something different?

Given the proposal mentions the possibility of a non-yielding borrow/mutate, seems that `yielding` is most like a modifier. So yes, I'd go with:
> So either accessors need to support a set of modifiers?

The next question is whether there should be an order to these, ie. is `yielding nonmutating mutate` allowed?

Sources/SwiftParser/Declarations.swift
```diff
@@ -1671,9 +1672,14 @@ extension Parser {
     var look = self.lookahead()
     let _ = look.consumeAttributeList()
     let hasModifier = look.consume(ifAnyIn: AccessorModifier.self) != nil
+    let hasYielding = look.consume(if: TokenSpec(.yielding)) != nil
```
tbkka: These changes to `Declarations.swift` seem to make sense to me.  This should verify the basic syntax.

Sources/SwiftParser/Declarations.swift
```diff
@@ -1659,6 +1659,7 @@ extension Parser {
   struct AccessorIntroducer {
     var attributes: RawAttributeListSyntax
     var modifier: RawDeclModifierSyntax?
+    var yielding: RawAccessorYieldingModifierSyntax?
```
tbkka: I need to either add a new field here or change `modifier` to an array.  No idea which is more appropriate.

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
```diff
@@ -180,6 +180,14 @@ public let DECL_NODES: [Node] = [
     ]
   ),
 
+  Node(
```
tbkka: Is there any documentation for how to do this?  @bnbarham ?  @hamishknight ? @rintaro ?

tbkka:
What does this mean?
```
Build of product 'generate-swift-syntax' complete! (0.08s)
SyntaxSupport/Child.swift:370: Fatal error: unexpected node has no siblings?
```

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
tbkka: @hamishknight suggested not doing this.

tbkka:
@swift-ci Please test

tbkka:
I ended up just making `yielding` be another modifier, handled the same as `mutating` or `nonmutating`.  The current PR allows any set of modifiers in any order -- more work will be required downstream in ASTGen or Sema to diagnose conflicts or ordering issues.

CodeGeneration/Sources/SyntaxSupport/KeywordSpec.swift
```diff
@@ -692,6 +693,8 @@ public enum Keyword: CaseIterable {
       return KeywordSpec("wrt")
     case .yield:
       return KeywordSpec("yield")
+    case .yielding:
+      return KeywordSpec("yielding", experimentalFeature: .coroutineAccessors)
```
hamishknight: Is this meant to still be an experimental feature given the proposal was accepted? Should we be removing `borrowAndMutateAccessors` too?

Tests/SwiftParserTest/DeclarationTests.swift
```diff
@@ -3467,15 +3467,21 @@ final class DeclarationTests: ParserTestCase {
...
+        yielding borrow {
+          yield _i
+        }
         _modify {
           yield &_i
         }
         modify {
           yield &_i
         }
+        yielding mutate {
```
hamishknight: Worth testing some different combinations of modifiers?

tbkka:
In the RFC discussion, @allevato suggested:
> We'll still need to keep modifier around as deprecated for source compatibility so that clients have an opportunity to migrate.

@hamishknight What's the right way to implement this?   I suppose I could put back the `modifier` field in `DeclNodes.swift`, but that won't get the deprecation right.  I could also hand-code the `modifier` field in an extension, I suppose.  Thoughts?

hamishknight:
We have a compatibility layer in `SwiftSyntaxCompatibility.swift` you could add to, I'm a little weary of adding one for `modifier` though, I think I would rather clients immediately migrate after bumping their swift-syntax version, unless it's particularly prevalent. AFAIK we don't guarantee source compatibility between swift-syntax versions, the compatibility overloads just serve as a way to ease the burden of migrating. It could potentially be worth adding a compatibility overload for the AccessorDecl constructor though since that can just forward a single modifier. What do you think @bnbarham?

hamishknight:
Though even if we don't add a compatibility layer for `modifier`, we could still add an `unavailable` stub for it that points users towards the new name

allevato:
Unless it's technically infeasible (e.g., in cases where the entire structure of a node must change), I don't think we should omit compatibility shims in general. Even if swift-syntax doesn't strictly offer source compatibility, it's still often the right thing to do when it's possible, because macro authors get the opportunity to migrate on their terms, _and_ you don't have to pair this change with parallel changes to other swiftlang/swift-* projects that might to be updated, like swift-format and possibly sourcekit-lsp.

In this case, I don't think there are any technical challenges here. The singular property can be marked deprecated but return either the first modifier or it could look through the list for `mutating` and return that if present or nil if not.

CodeGeneration/Sources/SyntaxSupport/KeywordSpec.swift
tbkka: Hmmmm....  The regular `borrow` and `mutate` accessors have not been approved.  But syntactically, since SE-0474 is approved, we do have to allow those keywords without a feature flag.   (CC: @meg-gupta)

Done.

hamishknight:
> you don't have to pair this change with parallel changes to other swiftlang/swift-* projects that might to be updated

Well that's the thing, I think we definitely want to make sure we update any uses in swiftlang projects since otherwise there's a good chance we're not handling this new language feature correctly. I think the same logic also extends to macros when they choose to adopt a new version of the language, as long as this isn't prevalent enough for that to be a burden. I'll let @bnbarham decide though.

CodeGeneration/Sources/SyntaxSupport/KeywordSpec.swift
tbkka: Done.

Tests/SwiftParserTest/DeclarationTests.swift
tbkka: I added a few more.

tbkka:
@allevato I implemented backwards-compatibility shims.  Let me know if this looks right to you.

Sources/SwiftParser/Declarations.swift
tbkka: Changed `modifier` to `modifiers` that takes a list.

Sources/SwiftParser/Declarations.swift
tbkka: Implemented `yielding` as a modifier for syntactic purposes.   ASTGen will be more interesting.

bnbarham:
> What do you think @bnbarham?

The main issue I see is with the setter + initializer, since a macro that's intending to copy a whole node could now remove a modifier. Maybe best to discuss in the RFC thread?

Sources/SwiftParser/CollectionNodes+Parsable.swift
```diff
@@ -67,7 +67,7 @@ extension AccessorDeclListSyntax: SyntaxParseable {
     } makeMissing: { remainingTokens, arena in
       return RawAccessorDeclSyntax(
         attributes: RawAttributeListSyntax(elements: [], arena: arena),
-        modifier: nil,
+        modifiers: RawDeclModifierListSyntax(elements: [], arena: arena),
```
ahoppen: We can save a few bytes in the syntax tree by using a shared empty collection. Also for `RawAttributeListSyntax` now that I look at it.


```suggestion
        attributes: self.emptyCollection(RawAttributeListSyntax.self),
        modifiers: self.emptyCollection(RawDeclModifierListSyntax.sel),
```

ahoppen (approved):
One minor suggestion, otherwise LGTM.

Sources/SwiftParser/CollectionNodes+Parsable.swift
tbkka: Done.  (I think?  I had to shuffle your suggestion a fair bit to get it to compile.)

tbkka:
@swift-ci Please test

tbkka:
@swift-ci Please test

tbkka:
@swift-ci Please test Linux Platform

tbkka:
Missed a backwards-compat shim that's needed for the old Observation macro.

tbkka:
@swift-ci Please test Linux Platform

tbkka:
@swift-ci Please test Linux Platform

tbkka:
@swift-ci Please test Linux Platform

tbkka:
@swift-ci Please test macOS Platform

tbkka:
@swift-ci Please test

tbkka:
@swift-ci Please test Linux Platform

tbkka:
@swift-ci Please test macOS Platform

tbkka:
Windows is failing because of this:
```
warning: In-process dependency scan query failed due to incompatible libSwiftScan (C:\Users\ContainerAdministrator\AppData\Local\Programs\Swift\Toolchains\0.0.0+Asserts\usr\bin\_InternalSwiftScan.dll). Fallback to `swift-frontend` dependency scanner invocation. Specify '-nonlib-dependency-scanner' to silence this warning.
```

tbkka:

PR #3191, opened 2025-11-20
Implement BasicMacroExpansionContext.buildConfiguration

jmschonfeld:
While testing out the new `buildConfiguration` property on the macro expansion context, I noticed that when my macro was expanding in unit tests the macro was not receiving the build configuration. The root cause was that `BasicMacroExpansionContext` did not implement `var buildConfiguration` directly and was instead picking up the default implementation which always returns `nil`. This change implements the property by returning the value from the shared state.

I couldn't find existing tests that exercise the build configuration as part of an expansion context, but if there's a good place to add a unit test let me know and I can try to add that.


jmschonfeld:
@swift-ci please test

DougGregor (approved):
Ah, thank you!

ahoppen:
@swift-ci Please test Windows

jmschonfeld:
@swift-ci please test

DougGregor:
Once this lands, we should also cherry-pick to the 6.3 branch.

PR #3192, opened 2025-11-21
Add support for Swift Testing in SwiftSyntaxMacrosTestsSupport

0xTim:
- **Explanation**:
  As described in https://github.com/swiftlang/swift-syntax/issues/2720 `SwiftSyntaxMacrosTestsSupport` does not work with Swift Testing, which is an issue given Swift Testing is the way to test projects and XCTest is deprecated. If produces false positives (where tests pass even if they shouldn't) which is a major issue, especially as there are no warnings.

This PR adds support for Swift Testing so that tests fail correctly. This **does not** introduce an issue with a circular dependency on Swift Testing. There is a circular dependency at the package level, but this is allowed due to https://github.com/swiftlang/swift-package-manager/pull/7530. There is no circular dependency between targets.

- **Scope**: No changes to existing code unless tests are passing when they shouldn't be. Only affects `SwiftSyntaxMacrosTestsSupport`
- **Issues**: https://github.com/swiftlang/swift-syntax/issues/2720
- **Risk**: Low risk
- **Testing**: Tests written to ensure it fails as expected
- **Reviewers**:

### DIFF
--- a/Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
+++ b/Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
@@ -37,6 +37,10 @@ public typealias DiagnosticSpec = SwiftSyntaxMacrosGenericTestSupport.Diagnostic
 /// Assert that expanding the given macros in the original source produces
 /// the given expanded source code.
 ///
+/// - Warning: If you call this function inside a Swift Testing test case,
+/// the test will report no assertion failures even if the test fails.
+/// Use ``expectMacroExpansion(_:expandedSource:diagnostics:macroSpecs:applyFixIts:fixedSource:testModuleName:testFileName:indentationWidth:buildConfiguration:fileID:file:line:column:)`` instead.
+///
 /// - Parameters:
 ///   - originalSource: The original source code, which is expected to contain
 ///     macros in various places (e.g., `#stringify(x + y)`).
@@ -51,7 +55,7 @@ public typealias DiagnosticSpec = SwiftSyntaxMacrosGenericTestSupport.Diagnostic
 ///   - indentationWidth: The indentation width used in the expansion.
 ///   - buildConfiguration: a build configuration that will be made available
 ///     to the macro implementation
-/// - SeeAlso: ``assertMacroExpansion(_:expandedSource:diagnostics:macroSpecs:applyFixIts:fixedSource:testModuleName:testFileName:indentationWidth:buildConfiguration:fileID:file:line:column:)``
+/// - SeeAlso: ``assertMacroExpansion(_:expandedSource:diagnostics:macroSpecs:applyFixIts:fixedSource:testModuleName:testFileName:indentationWidth:buildConfiguration:file:line:)``
 ///   to also specify the list of conformances passed to the macro expansion.
 public func assertMacroExpansion(
   _ originalSource: String,
@@ -64,10 +68,8 @@ public func assertMacroExpansion(
   testFileName: String = "test.swift",
   indentationWidth: Trivia = .spaces(4),
   buildConfiguration: (any BuildConfiguration)? = nil,
-  fileID: StaticString = #fileID,
   file: StaticString = #filePath,
-  line: UInt = #line,
-  column: UInt = #column
+  line: UInt = #line
 ) {
   let specs = macros.mapValues { MacroSpec(type: $0) }
   assertMacroExpansion(
@@ -81,15 +83,17 @@ public func assertMacroExpansion(
     testFileName: testFileName,
     indentationWidth: indentationWidth,
     buildConfiguration: buildConfiguration,
-    fileID: fileID,
     file: file,
-    line: line,
-    column: column
+    line: line
   )
 }
 
 /// Assert that expanding the given macros in the original source produces
 /// the given expanded source code.
+///
+/// - Warning: If you call this function inside a Swift Testing test case,
+/// the test will report no assertion failures even if the test fails.
+/// Use ``expectMacroExpansion(_:expandedSource:diagnostics:macroSpecs:applyFixIts:fixedSource:testModuleName:testFileName:indentationWidth:buildConfiguration:fileID:file:line:column:)`` instead.
 ///
 /// - Parameters:
 ///   - originalSource: The original source code, which is expected to contain
@@ -117,6 +121,60 @@ public func assertMacroExpansion(
   testFileName: String = "test.swift",
   indentationWidth: Trivia = .spaces(4),
   buildConfiguration: (any BuildConfiguration)? = nil,
+  file: StaticString = #filePath,
+  line: UInt = #line
+) {
+  SwiftSyntaxMacrosGenericTestSupport.assertMacroExpansion(
+    originalSource,
+    expandedSource: expectedExpandedSource,
+    diagnostics: diagnostics,
+    macroSpecs: macroSpecs,
+    applyFixIts: applyFixIts,
+    fixedSource: expectedFixedSource,
+    testModuleName: testModuleName,
+    testFileName: testFileName,
+    indentationWidth: indentationWidth,
+    buildConfiguration: buildConfiguration,
+    failureHandler: {
+      XCTFail($0.message, file: $0.location.staticFilePath, line: $0.location.unsignedLine)
+    },
+    fileID: "",  // Not used in the failure handler
+    filePath: file,
+    line: line,
+    column: 0  // Not used in the failure handler
+  )
+}
+
+/// Expect that expanding the given macros in the original source produces
+/// the given expanded source code.
+/// This function is the Swift Testing equivalent of ``assertMacroExpansion(_:expandedSource:diagnostics:macroSpecs:applyFixIts:fixedSource:testModuleName:testFileName:indentationWidth:buildConfiguration:file:line:)``
+///
+/// - Parameters:
+///   - originalSource: The original source code, which is expected to contain
+///     macros in various places (e.g., `#stringify(x + y)`).
+///   - expectedExpandedSource: The source code that we expect to see after
+///     performing macro expansion on the original source.
+///   - diagnostics: The diagnostics when expanding any macro
+///   - macroSpecs: The macros that should be expanded, provided as a dictionary
+///     mapping macro names (e.g., `"CodableMacro"`) to specification with macro type
+///     (e.g., `CodableMacro.self`) and a list of conformances macro provides
+///     (e.g., `["Decodable", "Encodable"]`).
+///   - applyFixIts: If specified, filters the Fix-Its that are applied to generate `fixedSource` to only those whose message occurs in this array. If `nil`, all Fix-Its from the diagnostics are applied.
+///   - fixedSource: If specified, asserts that the source code after applying Fix-Its matches this string.
+///   - testModuleName: The name of the test module to use.
+///   - testFileName: The name of the test file name to use.
+///   - indentationWidth: The indentation width used in the expansion.
+public func expectMacroExpansion(
+  _ originalSource: String,
+  expandedSource expectedExpandedSource: String,
+  diagnostics: [DiagnosticSpec] = [],
+  macroSpecs: [String: MacroSpec],
+  applyFixIts: [String]? = nil,
+  fixedSource expectedFixedSource: String? = nil,
+  testModuleName: String = "TestModule",
+  testFileName: String = "test.swift",
+  indentationWidth: Trivia = .spaces(4),
+  buildConfiguration: (any BuildConfiguration)? = nil,
   fileID: StaticString = #fileID,
   file: StaticString = #filePath,
   line: UInt = #line,
@@ -134,27 +192,72 @@ public func assertMacroExpansion(
     indentationWidth: indentationWidth,
     buildConfiguration: buildConfiguration,
     failureHandler: {
-      #if canImport(Testing)
-      if Test.current != nil {
-        Issue.record(
-          Comment(rawValue: $0.message),
-          sourceLocation: .init(
-            fileID: fileID.description,
-            filePath: file.description,
-            line: Int(line),
-            column: Int(column)
-          )
+      Issue.record(
+        Comment(rawValue: $0.message),
+        sourceLocation: .init(
+          fileID: fileID.description,
+          filePath: file.description,
+          line: Int(line),
+          column: Int(column)
         )
-      } else {
-        XCTFail($0.message, file: $0.location.staticFilePath, line: $0.location.unsignedLine)
-      }
-      #else
-      XCTFail($0.message, file: $0.location.staticFilePath, line: $0.location.unsignedLine)
-      #endif
+      )
     },
     fileID: "",  // Not used in the failure handler
     filePath: file,
     line: line,
     column: 0  // Not used in the failure handler
+  )
+}
+
+/// Expect that expanding the given macros in the original source produces
+/// the given expanded source code.
+/// This function is the Swift Testing equivalent of ``assertMacroExpansion(_:expandedSource:diagnostics:macroSpecs:applyFixIts:fixedSource:testModuleName:testFileName:indentationWidth:buildConfiguration:file:line:)``
+///
+/// - Parameters:
+///   - originalSource: The original source code, which is expected to contain
+///     macros in various places (e.g., `#stringify(x + y)`).
+///   - expectedExpandedSource: The source code that we expect to see after
+///     performing macro expansion on the original source.
+///   - diagnostics: The diagnostics when expanding any macro
+///   - macros: The macros that should be expanded, provided as a dictionary
+///     mapping macro names (e.g., `"stringify"`) to implementation types
+///     (e.g., `StringifyMacro.self`).
+///   - applyFixIts: If specified, filters the Fix-Its that are applied to generate `fixedSource` to only those whose message occurs in this array. If `nil`, all Fix-Its from the diagnostics are applied.
+///   - fixedSource: If specified, asserts that the source code after applying Fix-Its matches this string.
+///   - testModuleName: The name of the test module to use.
+///   - testFileName: The name of the test file name to use.
+///   - indentationWidth: The indentation width used in the expansion.
+public func expectMacroExpansion(
+  _ originalSource: String,
+  expandedSource expectedExpandedSource: String,
+  diagnostics: [DiagnosticSpec] = [],
+  macros: [String: Macro.Type],
+  applyFixIts: [String]? = nil,
+  fixedSource expectedFixedSource: String? = nil,
+  testModuleName: String = "TestModule",
+  testFileName: String = "test.swift",
+  indentationWidth: Trivia = .spaces(4),
+  buildConfiguration: (any BuildConfiguration)? = nil,
+  fileID: StaticString = #fileID,
+  file: StaticString = #filePath,
+  line: UInt = #line,
+  column: UInt = #column
+) {
+  let specs = macros.mapValues { MacroSpec(type: $0) }
+  expectMacroExpansion(
+    originalSource,
+    expandedSource: expectedExpandedSource,
+    diagnostics: diagnostics,
+    macroSpecs: specs,
+    applyFixIts: applyFixIts,
+    fixedSource: expectedFixedSource,
+    testModuleName: testModuleName,
+    testFileName: testFileName,
+    indentationWidth: indentationWidth,
+    buildConfiguration: buildConfiguration,
+    fileID: fileID,
+    file: file,
+    line: line,
+    column: column
   )
 }
--- a/Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
+++ b/Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
@@ -51,7 +51,7 @@ public typealias DiagnosticSpec = SwiftSyntaxMacrosGenericTestSupport.Diagnostic
 ///   - indentationWidth: The indentation width used in the expansion.
 ///   - buildConfiguration: a build configuration that will be made available
 ///     to the macro implementation
-/// - SeeAlso: ``assertMacroExpansion(_:expandedSource:diagnostics:macroSpecs:applyFixIts:fixedSource:testModuleName:testFileName:indentationWidth:buildConfiguration:file:line:)``
+/// - SeeAlso: ``assertMacroExpansion(_:expandedSource:diagnostics:macroSpecs:applyFixIts:fixedSource:testModuleName:testFileName:indentationWidth:buildConfiguration:fileID:file:line:column:)``
 ///   to also specify the list of conformances passed to the macro expansion.
 public func assertMacroExpansion(
   _ originalSource: String,
--- a/Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
+++ b/Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
@@ -68,8 +68,10 @@ public func assertMacroExpansion(
   testFileName: String = "test.swift",
   indentationWidth: Trivia = .spaces(4),
   buildConfiguration: (any BuildConfiguration)? = nil,
+  fileID: StaticString = #fileID,
   file: StaticString = #filePath,
-  line: UInt = #line
+  line: UInt = #line,
+  column: UInt = #column
 ) {
   let specs = macros.mapValues { MacroSpec(type: $0) }
   assertMacroExpansion(
@@ -83,8 +85,10 @@ public func assertMacroExpansion(
     testFileName: testFileName,
     indentationWidth: indentationWidth,
     buildConfiguration: buildConfiguration,
+    fileID: fileID,
     file: file,
-    line: line
+    line: line,
+    column: column
   )
 }
 
@@ -121,8 +125,10 @@ public func assertMacroExpansion(
   testFileName: String = "test.swift",
   indentationWidth: Trivia = .spaces(4),
   buildConfiguration: (any BuildConfiguration)? = nil,
+  fileID: StaticString = #fileID,
   file: StaticString = #filePath,
-  line: UInt = #line
+  line: UInt = #line,
+  column: UInt = #column
 ) {
   SwiftSyntaxMacrosGenericTestSupport.assertMacroExpansion(
     originalSource,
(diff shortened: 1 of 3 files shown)


Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
```diff
@@ -125,7 +134,15 @@ public func assertMacroExpansion(
     indentationWidth: indentationWidth,
     buildConfiguration: buildConfiguration,
     failureHandler: {
+      #if canImport(Testing)
+      if Test.current != nil {
```
grynspan: This is not the correct test to determine which library is in use because code can run in a detached task. See https://github.com/swiftlang/swift-testing/issues/475

grynspan (changes requested):

Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
```diff
@@ -17,6 +17,9 @@ public import SwiftSyntaxMacroExpansion
 public import SwiftSyntaxMacros
 @_spi(XCTestFailureLocation) public import SwiftSyntaxMacrosGenericTestSupport
 private import XCTest
+#if canImport(Testing)
+import Testing
```
grynspan: This _must_ be a `private import`.

grynspan (dismissed):

Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
0xTim: What is the correct way? Should we split it out into a `expect` function instead?

grynspan: There is no correct way at this time, which is why that issue is still open. Jerry's work should allow us to just call `#expect()` here and have it work under all configurations.

0xTim: The pitch looks like it will solve the issue, but still require work in the library to migrate over to Swift Testing APIs. What I propose is that we land this now, as it solves a problem that exists for users today (and potentially provide a release in the next monthly Linux release/Swift patch release) and then fix forward when the proposal lands. Given it's still in the pitch phase it likely won't be landed until 6.4 and waiting 10 months for a solution seems like a bad idea.

Regarding the `Test.current` issue - from my understanding this works in all instances apart from those running in a detached task. For this specific case, I can't see a scenario when a user would be using the `assertMacroExpansion` from a detached task so we can fix this for the majority of the users and those attempting to use it from a detached task will see no change in behaviour.

grynspan: @stmontgomery Your take? You okay with this presumably being nonfunctional with the package build?

> For this specific case, I can't see a scenario when a user would be using the assertMacroExpansion from a detached task so we can fix this for the majority of the users and those attempting to use it from a detached task will see no change in behaviour.

Let's at least document it as unsupported in the symbol's Markup?

stmontgomery: First of all, thank you @0xTim for taking a stab at this. You have essentially beaten us to the punch; I was anticipating that one of us who contribute to Swift Testing would eventually propose a similar change to what you have here once @jerryjrchen's interoperability work had all landed. But yes, I do see some advantage to proceeding with this interim step forward now, even while I acknowledge @grynspan's point about it not working for detached tasks.

My understanding is that right now, since interoperability is neither available yet nor an approved feature, if we land this PR and a user has a dependency on both the swift-syntax _and_ swift-testing packages, then what will happen is that the `SwiftSyntaxMacrosTestsSupport` module from their swift-syntax build will import and link the copy of the `Testing` module from the toolchain they are building with. However, since they have a copy of swift-testing in the package graph as well, the user's own tests will import and link the copy of the `Testing` module from that package dependency, ignoring the toolchain copy. As a result, the new call to `#expect` in `SwiftSyntaxMacrosTestsSupport` will interact with the "wrong" copy of the `Testing` module (the toolchain copy) and will effectively be a no-op because the toolchain copy won't have any active tests running, only the package copy will. Therefore, in this scenario, the fact that a user has a package dependency on swift-testing causes this fix to not have the desired effect, but it also is no worse than the status quo.

Assuming that understanding is correct (and please correct me if not), I think it's not an ideal situation but also not worrisome enough in practice to avoid proceeding with this change. And the good news is that if/when interoperability does land, it should solve for this because it will centralize things to ensure that there is always only at most one "fallback handler" in each process.

I think this is definitely a good suggestion:

> Let's at least document it as unsupported in the symbol's Markup?

And I also think we should find a place to document the special considerations and pitfalls when you have both swift-syntax and swift-testing package dependencies. In general, we greatly encourage using Swift Testing from the toolchain and this is one reason why.

grynspan: > As a result, the new call to `#expect` in `SwiftSyntaxMacrosTestsSupport` will interact with the "wrong" copy of the `Testing` module (the toolchain copy) and will effectively be a no-op because the toolchain copy won't have any active tests running, only the package copy will.

Akshully, this could go even worse on non-Apple platforms (no two-level linker namespace) and result in duplicate symbol errors at link time.

stmontgomery: > this could go even worse on non-Apple platforms (no two-level linker namespace) and result in duplicate symbol errors at link time.

That's a good callout. And I suppose _any_ testing library[^1] vended by any package would have that problem if it uses Swift Testing, assuming the package doesn't declare its own dependency on swift-testing and relies on the toolchain copy instead. I don't think problem is unique to swift-syntax, in other words, and IMO we shouldn't recommend that packages avoid using Swift Testing in test-helper libraries due to this problem.

[^1]: Note that the concept of a 'testing library' is not yet formally defined in Swift package manager, but I mean any library product in a package which imports `Testing`. See also: [swiftlang/swift-package-manager#9343](https://github.com/swiftlang/swift-package-manager/pull/9343)

So _if_ we proceed with this change, it would probably be best to communicate to clients of swift-syntax that if they use `SwiftSyntaxMacrosTestsSupport` with Swift Testing tests, they should be sure to use the toolchain copy and not the swift-testing package. (Perhaps that could go in the release notes?) And if they insist on using the swift-testing package copy, they can instead use the "generic" test library (`SwiftSyntaxMacrosGenericTestSupport`). That would allow them to provide their own handler function, which can use their package dependency-provided copy of `#expect`.

FranzBusch: @stmontgomery @grynspan Thanks for clarifying the situation around the dependencies here. If we can agree here that `SwiftSyntax` can provide `Testing` support by depending on the toolchain's `Testing` module can we unblock the PR on this matter?

I think there is still the larger question on how this is currently implemented. The `if Test.current != nil ` check in this method is certainly less than ideal. Instead I would propose that we add an entirely new method called `expectMacroExpansion` that is based on `Testing` unconditionally. So existing users of `assertMacroExpansion` can continue to use it with `XCTest` and `Testing` users can start adopting the new one. How does that sound to everyone?

grynspan: > If we can agree here that `SwiftSyntax` can provide `Testing` support by depending on the toolchain's `Testing` module can we unblock the PR on this matter?

I'd be morally okay with saying "the features in the swift-syntax repository are dependent on the built-in copy of Swift Testing even if you include a package dependency" however this will break builds on non-Apple platforms with flat linker namespaces due to duplicate symbols at link time.

> I think there is still the larger question on how this is currently implemented. The `if Test.current != nil ` check in this method is certainly less than ideal. Instead I would propose that we add an entirely new method called `expectMacroExpansion` that is based on `Testing` unconditionally. So existing users of `assertMacroExpansion` can continue to use it with `XCTest` and `Testing` users can start adopting the new one. How does that sound to everyone?

Once @jerryjrchen's work on the _interop_ feature lands, it will be possible to implement this in a way that depends on neither XCTest _nor_ Swift Testing. It may be a better idea to just wait until that work is done and revisit the problem at that point.

FranzBusch: > I'd be morally okay with saying "the features in the swift-syntax repository are dependent on the built-in copy of Swift Testing even if you include a package dependency" however this will break builds on non-Apple platforms with flat linker namespaces due to duplicate symbols at link time.

Only if there is both `Testing` from the toolchain and from the package right? Which I thought we agreed is only really valid in development environments.

grynspan: > Only if there is both Testing from the toolchain and from the package right? Which I thought we agreed is only really valid in development environments.

All testing environments are development environments.

If swift-syntax explicitly links the copy of Swift Testing in the toolchain, that will break developers who want to test their macros and have a package dependency on Swift Testing.

In addition, if they are using the package copy of Swift Testing and you are using the toolchain's copy, your calls to e.g. `#expect()` won't be routed to the infrastructure the developer's test target links against, so a failure will be invisible.

Jerry's work should give us an escape hatch for this problem, so we should wait until it lands and then make the necessary changes here.

0xTim: My concern with waiting is that it could be a year until that's shipped, meanwhile users have no indications their tests are passing incorrectly, whereas we could fix it today for the majority of use cases

grynspan: I appreciate the concern, but there are technical blockers here. We must not cause build failures for teams using Swift Testing as a package.

0xTim:
Ok all, I've introduced a separate function to be used from Swift testing and added some documentation to the `assert` functions. Let me know what you think!

cc @FranzBusch

grynspan:
I think this is still ultimately a moot point with @jerryjrchen's upcoming work?

FranzBusch:
> I think this is still ultimately a moot point with @jerryjrchen's upcoming work?

I thought we identified in the other thread that this is not the case. While the interoperability will allow the existing XCTest based method to work. It won't work with the `strict` mode which seems limiting. I think the new documentation in the `swift-testing` makes it pretty clear that the primary use-case of `swift-testing` as a package is for local development.

0xTim:
To add some more context as well, I was talking to a couple of maintainers at FOSDEM who have both hit this issue with false positives



> I think this is still ultimately a moot point with @jerryjrchen's upcoming work?

This work is still ongoing correct? Because it won't land until 6.4 at the earliest I'd assume

grynspan:
Right but that doesn't mean we should ship new API that immediately becomes obsolete with the next Swift release.

grynspan:
> While the interoperability will allow the existing XCTest based method to work. It won't work with the `strict` mode which seems limiting.

The end result of Jerry's work is, as I understand it, that the existing API in this library can adopt `#expect()` in place of `XCTAssertTrue()` (etc. etc.) and it'll work in both libraries without problems.

> I think the new documentation in the `swift-testing` makes it pretty clear that the primary use-case of `swift-testing` as a package is for local development.

I'm not sure I understand how that relates to my comment, but maybe you're referring to something earlier in the conversation I missed?

jerryjrchen:
> While the interoperability will allow the existing XCTest based method to work. It won't work with the `strict` mode which seems limiting.

We can resolve the `strict` mode restriction against using XCTest API in a Swift Testing test by adopting `#expect` in place of XCTest API (I can see `XCTAssert*` and `XCTFail` from a quick search). Interop also ensures the conversion to `#expect` still works as expected if the helpers are being called from an XCTest test case.

stmontgomery:
Hi folks, I want to suggest a path forward here which I hope will unblock progress on this PR and allow it to proceed in the short term (if accepted by the repo's code owners, of course!) while laying out a plan for how it will align with our anticipated and fast-approaching interoperability work. 

1. Keep just one "assert" function, with the existing name `assertMacroExpansion()`
    - …instead of introducing a separate function `expectMacroExpansion()` and having one for XCTest, one for Swift Testing.
    - This is to avoid bifurcation between the XCTest and Swift Testing worlds.
1. Modify `assertMacroExpansion()` so that it calls both `Issue.record()` (Swift Testing) and `XCTFail()` (XCTest), one after the other.
    - For right now, this is sufficient to ensure failures are surfaced regardless of which testing library is active.
    - Note: to achieve full reporting fidelity, the function will need to gain `fileID:` and `column:` parameters with default values and plumb those through to the "generic" funnel point so Swift Testing can reference them.
1. (Important) Communicate the potential effects of this change to SwiftSyntax users
    - This could happen via a detailed release note in the tag this change lands in. See the [602.0.0](https://github.com/swiftlang/swift-syntax/releases/tag/602.0.0) notes for example of the format of the last release.
    - This note should do several things:
        - First, celebrate the useful new functionality—Swift Testing support for macro testing!
        - Then, explain the known caveat(s). The main one being that if you have a package dependency on Swift Testing, you upgrade to this (major) version of SwiftSyntax, and you attempt to validate macros from a Swift Testing test, then it's possible failures may not be surfaced correctly.
        - Link to [my new document](https://github.com/swiftlang/swift-testing/blob/main/Documentation/Distributions.md) in the Swift Testing repo about this topic, and emphasize that it's due to using a non-built in copy.
        - Offer two solutions for anyone affected by this:
            - Simplest: Remove the package dependency on Swift Testing and switch to using a built-in copy, as the [document](https://github.com/swiftlang/swift-testing/blob/main/Documentation/Distributions.md) recommends.
            - Otherwise, switch from importing `SwiftSyntaxMacrosTestSupport` to `SwiftSyntaxMacrosGenericTestSupport` and specify your own `failureHandler:` closure in which you call `Issue.record()` yourself. That should work even when using a package copy of Swift Testing.
1. Later, in a subsequent PR, adjust this logic for Swift Testing interoperability
    - If/when the pitched [Interoperability](https://github.com/jerryjrchen/swift-evolution/blob/swt-xct-interop/proposals/testing/NNNN-targeted-interoperability-swift-testing-and-xctest.md) feature arrives, `assertMacroExpansion()` will need to be adjusted to avoid redundant test issues being recorded due to it calling _both_ `Issue.record()` and `XCTFail()`.
    - To avoid that, at that point in time `assertMacroExpansion()` can be modified to stop calling `XCTFail()` starting in the Swift version where interoperability is first enabled.
    - (The call to `XCTFail()` shouldn't be removed altogether though, since SwiftSyntax could presumably still be built and used from older Swift versions.)

I've assembled a branch (atop this PR branch) with my suggested code changes for this PR in a commit you can see here: https://github.com/stmontgomery/swift-syntax/commit/c772dc6c5949870845063bac7ce7254a357a2829. Thoughts?

stmontgomery:
There are two other points above I wanted to respond to:

> While the interoperability will allow the existing XCTest based method to work. It won't work with the `strict` mode which seems limiting.

Building on what @jerryjrchen [said](https://github.com/swiftlang/swift-syntax/pull/3192#issuecomment-3837940201), the semantics of strict mode will be that in any situation where Swift Testing offers an API, that should be preferred over the analogous XCTest variant. It does not mandate that you must only call the APIs corresponding to the active test framework, in case that's something folks are concerned about here. I do think/hope strict mode can eventually be used in conjunction with this work, if end users choose to enable it.

> We must not cause build failures for teams using Swift Testing as a package.

I recognize that this plan does have the potential to break users who have a package dependency on Swift Testing. I suspect the practical impact of that is likely small due to few users relying on it, and I think the tradeoff here is very worthwhile since we'll be delivering a frequently requested feature. The [document](https://github.com/swiftlang/swift-testing/blob/main/Documentation/Distributions.md) I wrote about distribution locations of Swift Testing was intended to help clarify our recommendations regarding use of the package copy, and frankly, to discourage its use more strongly than we have in the past.

This change will roll out in a new _major_ tag of SwiftSyntax, meaning (per SemVer) it will be an explicit action users must take so the timing of the upgrade will be in their control. And I believe we can mitigate and help users make the transition gracefully with a clearly-worded release note (discussed above).

0xTim:
@stmontgomery looks good to me - happy to pull those changes into my branch as I have done (just some extra changes to add if you agree around the can import stuff and expanding the test info) or have you submit the PR

rintaro:
Sorry I haven't been following this PR, so I may be missing some context.

FWIW, the original intent behind the current structure was to take advantage of [cross-import overlays](https://forums.swift.org/t/cross-import-overlays/36710) once it's accepted and SwiftPM supports them.
The idea was:
 * Rename the current XCTest-based `SwiftSyntaxMacrosTestSupport` to `SwiftSyntaxMacrosTestSupport_XCTest`
 * Rename the current `SwiftSyntaxMacrosGenericTestSupport` to `SwiftSyntaxMacrosTestSupport`
 * Introduce `SwiftSyntaxMacrosTestSupport_SwiftTesting`, which provides the swift-testing implementation of `assertMacroExpansion()`
With that in place, users could simply write:
```swift
import SwiftSyntaxMacrosTestSupport
import SwiftTesting
```
and have `SwiftSyntaxMacrosTestSupport_SwiftTesting` picked up automatically via cross-import overlays.

That's said I'm a bit skeptical that cross-import overlays will be available in the near future. Also, as I understand it, we likely can't utilize [Package traits](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0450-swiftpm-package-traits.md) because we treat XCTest and Testing as system libraries. Given that, plus the direction of [the interoperability work](https://github.com/jerryjrchen/swift-evolution/blob/swt-xct-interop/proposals/testing/NNNN-targeted-interoperability-swift-testing-and-xctest.md), this PR (and the plan from @stmontgomery) seems to be the right direction.


A couple of questions:

> Modify assertMacroExpansion() so that it calls both Issue.record() (Swift Testing) and XCTFail() (XCTest), one after the other.

Just to confirm, this is safe because [XCTAssert](https://developer.apple.com/documentation/xctest/equality-and-inequality-assertions) failure in a Swift Testing test or an [#expect](https://developer.apple.com/documentation/testing/expectations) failure in an XCTest test is silently ignored (from [the interoperability proposal](https://github.com/jerryjrchen/swift-evolution/blob/swt-xct-interop/proposals/testing/NNNN-targeted-interoperability-swift-testing-and-xctest.md#introduction))?

What would happen _after_ the interoperability is implemented? Would a failure be reported twice? Or the interoperability feature would consolidate the duplicated failure into one?

>  if you have a package dependency on Swift Testing, you upgrade to this (major) version of SwiftSyntax, and you attempt to validate macros from a Swift Testing test, then it's possible failures may not be surfaced correctly.

Could you elaborate on this? Any example that would break with this change? I don't believe a simple setup like this would work even with the current situation.
```
import SwiftSyntaxMacroTestSupport
import Testing

@Test func testMyMacro() {
  assertMacroExpansion(...)
}
```

Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
```diff
@@ -34,6 +37,10 @@ public typealias DiagnosticSpec = SwiftSyntaxMacrosGenericTestSupport.Diagnostic
 /// Assert that expanding the given macros in the original source produces
 /// the given expanded source code.
 ///
+/// - Warning: If you call this function inside a Swift Testing test case,
+/// the test will report no assertion failures even if the test fails.
+/// Use ``expectMacroExpansion(_:expandedSource:diagnostics:macroSpecs:applyFixIts:fixedSource:testModuleName:testFileName:indentationWidth:buildConfiguration:fileID:file:line:column:)`` instead.
+///
```
rintaro: This (and the one on the another function) isn't true now.

0xTim: Since which Swift version? Because it's definitely the case in 6.2.3

rintaro:
> Could you elaborate on this? Any example that would break with this change? I don't believe a simple setup like this would work even with the current situation.

Ah, I see it now. `SwiftSytaxMacrosTestingSupport` might be built against the toolchain Testing, but the clients will be built against the just-built Testing because they have the dependencies. And that will cause the issue.

Sources/SwiftSyntaxMacrosTestSupport/Assertions.swift
rintaro: I meant after your latest change. You're not even adding `expectMacroExpansion` anymore. :)

0xTim: Ah yes, very good point!

0xTim:
Anything I can do to push this along?

grynspan:
Help review the proposal for interop? https://forums.swift.org/t/st-0021-targeted-interoperability-between-swift-testing-and-xctest/84965/

0xTim:
Right but that's targeting 6.4, assuming it's gets accepted (which for the record I'm in favour of). 

As far as I can tell there's nothing stopping this being merged and it could, if there was the desire land in 6.3 and not force people to hit this issue for another 6 months. This PR has already been open for 3 months. It would be good to fix it sooner rather than later.

grynspan:
The review period is only a couple of weeks. Let's at least wait that out.

stmontgomery:
Hey @rintaro ,
> A couple of questions:
> 
> > Modify assertMacroExpansion() so that it calls both Issue.record() (Swift Testing) and XCTFail() (XCTest), one after the other.
> 
> Just to confirm, this is safe because [XCTAssert](https://developer.apple.com/documentation/xctest/equality-and-inequality-assertions) failure in a Swift Testing test or an [#expect](https://developer.apple.com/documentation/testing/expectations) failure in an XCTest test is silently ignored (from [the interoperability proposal](https://github.com/jerryjrchen/swift-evolution/blob/swt-xct-interop/proposals/testing/NNNN-targeted-interoperability-swift-testing-and-xctest.md#introduction))?

Correct, that's the behavior prior to the interoperability feature.

> What would happen _after_ the interoperability is implemented? Would a failure be reported twice? Or the interoperability feature would consolidate the duplicated failure into one?

I think what we should do is proactively land this change with the call to `XCTFail()` guarded by `#if compiler(<6.4)` in anticipation of the interoperability feature being included in 6.4.

As of this writing no outcome has been announced for [ST-0021](https://github.com/swiftlang/swift-evolution/blob/main/proposals/testing/0021-targeted-interoperability-swift-testing-and-xctest.md), so we shouldn't actually land this PR with that guard in place until that happens to be safe. But assuming it does get accepted _soon_, in my opinion this PR could be landed and it could be included in 6.3 too. And that would mean that even when using a 6.3-series tag of SwiftSyntax with a future 6.4 toolchain which has Interoperability, things would work correctly and test failures would not be reported twice (redundantly).

> > if you have a package dependency on Swift Testing, you upgrade to this (major) version of SwiftSyntax, and you attempt to validate macros from a Swift Testing test, then it's possible failures may not be surfaced correctly.
> 
> Could you elaborate on this? Any example that would break with this change? I don't believe a simple setup like this would work even with the current situation.
> 
> ```
> import SwiftSyntaxMacroTestSupport
> import Testing
> 
> @Test func testMyMacro() {
>   assertMacroExpansion(...)
> }
> ```

Here's an example to illustrate what I mean. Let's assume for the sake of discussion we move forward with this SwiftSyntax PR. Imagine a user has a package which has dependencies on _both_ the `swift-syntax` and `swift-testing` packages — that is, they choose not to use the copy of Swift Testing included in the toolchain, but use its package copy instead. In that scenario, code in `swift-syntax` will now have `import Testing` and call its APIs. However, the `swift-syntax` package intentionally does not declare a package dependency on the `swift-testing` package itself; it just assumes that it will use the copy of that module from the toolchain. So it's an ambiguous dependency.

Because of that, the order the targets in the graph build in could lead to nondeterministic behavior. It's possible that the `swift-syntax` modules which `import Testing` may build before `swift-testing` package and thus, those `swift-syntax` modules will import and link the toolchain copy of `libTesting`. Or it's possible the order will be inverted and everything will link to the package copy of `swift-testing`. If the former happens, SwiftSyntax will record issues but do so against a copy of the testing library which is different than the one the end users' tests are using, so the issues will be silently ignored. But again, there are several potential outcomes and the whole situation is nondeterministic.

I lay all this out _not_ to worry people; I fully support the plan above. I'm just trying to illustrate the known pitfalls of using a package copy of Swift Testing, and that's why I wrote a [documentation article](https://github.com/swiftlang/swift-testing/blob/main/Documentation/Distributions.md) trying to discourage it. I do not think this concern should prevent us from proceeding with landing this PR.

rintaro:
> I think what we should do is proactively land this change with the call to XCTFail() guarded by #if compiler(<6.4) in anticipation of the interoperability feature being included in 6.4.

This sounds good to me :+1:

0xTim:
Ping again on this @stmontgomery given the timescales

0xTim:
Closing as there seems to be no desire to merge

grynspan:
Reopening. Yes, there is, I promise!

stmontgomery:
I spent some time on this last night and wanted to share an update. My proposal for how to proceed with this _had_ been the following:

1. Fix a couple small things I saw in this PR while reviewing it closely (`fileID:` and `column:` args not being propagated, and an update to a DocC symbol reference).
1. Guard the call to `XCTFail()` with `#if compiler(<6.4)`, so that in toolchains where the [ST-0021: Interoperability](https://github.com/swiftlang/swift-evolution/blob/main/proposals/testing/0021-targeted-interoperability-swift-testing-and-xctest.md) feature is enabled, only `Issue.record()` will be called.
1. Test that using both 6.3 and 6.4 snapshot toolchains, to confirm both the pre- and post-interoperability scenarios work as expected.
1. Once qualified, merge this PR into `main`, and then request to cherry-pick it into the `release/6.3` branch for an upcoming 6.3.x release.

I pushed the changes for 1 directly to this PR branch since they're trivial (https://github.com/swiftlang/swift-syntax/pull/3192/commits/fe826cf2c4410e386394855195aedef91c245627).

However, after making the change in 2 and qualifying them (3) in a 6.4 snapshot toolchain I realized there's a problem with this rollout plan: the default **interop mode**. The mode is `limited` in the 6.4 toolchains, except for packages whose tools version is set to 6.4 or later (in which case it will be `complete`). In `limited` mode, issues propagated across the interop boundary are recorded with **warning** severity, meaning they don't cause a test failure. In practice, what this means is that when we initially roll this change out, for existing XCTests which use `assertMacroExpansion()`, if the assertion fails when using a 6.4 toolchain, `Issue.record()` will be called but it will only record an XCTest _warning_, and the test status will be Success instead of Failure.

I'm pretty sure that isn't the experience we want and that would be considered a regression; we need the final XCTest status to be a true Failure in that situation. So I also experimented with some workarounds and I do have one potentially viable alternative which keys off of both the interop environment variable and `Test.current`: see https://github.com/stmontgomery/swift-syntax/commit/6e07d36c7c627910c2ef1a936cfd4925ff00d1e6. However, I'm beginning to consider some Swift Testing enhancements to streamline this workflow and reduce the need for this kind of workaround. Still figuring that out, and I'd like to talk with @jerryjrchen and @grynspan about it, but wanted to give that update. I personally still want to find a way to get some version of this into a 6.3-series release if possible.

FranzBusch:
> However, I'm beginning to consider some Swift Testing enhancements to streamline this workflow and reduce the need for this kind of workaround. Still figuring that out, and I'd like to talk with @jerryjrchen and @grynspan about it, but wanted to give that update. I personally still want to find a way to get some version of this into a 6.3-series release if possible.

Thanks for sharing all of this. Just to take a step back what if we consider a previous alternative again - Introduce an entirely new method that is only using `Testing` expectation APIs so we don't try to make a single method work in all contexts.

stmontgomery:
> Thanks for sharing all of this. Just to take a step back what if we consider a previous alternative again - Introduce an entirely new method that is only using `Testing` expectation APIs so we don't try to make a single method work in all contexts.

Yes, I believe that would work. (We should test it though of course, including in a few interoperability scenarios, to confirm.) Part of why I've been getting involved and experimenting in this space is because I suspect the transition challenges we're observing here aren't unique to SwiftSyntax; other packages which vend test-helper libraries will encounter them too. So I've been looking for ways to avoid the need for clients to introduce "parallel" sets of helper functions (one for XCTest, one for Swift Testing) since it can often be difficult to know which is the correct one to call—especially if the helper is within a wrapper and isn't called directly from a "leaf node" test function. So while that approach may work for SwiftSyntax specifically, for other use cases that approach might be problematic.

Still, it's worth considering (again) in this PR, given that most macro tests do call this utility directly from the test function. Supposing we did take that approach, and introduced a new "**expect**MacroExpansion()" function, that would support interoperability too of course. So once it existed, it would be usable from XCTests too (with the severity caveat above), and over time as ST-0021 phases in and peoples' interop modes transition from `limited` to `complete`, that newer wrapper could become the "primary" one, and eventually `assertMacroExpansion()` could be deprecated outright.

FranzBusch:
> Still, it's worth considering (again) in this PR, given that most macro tests do call this utility directly from the test function. Supposing we did take that approach, and introduced a new "**expect**MacroExpansion()" function, that would support interoperability too of course. So once it existed, it would be usable from XCTests too (with the severity caveat above), and over time as ST-0021 phases in and peoples' interop modes transition from `limited` to `complete`, that newer wrapper could become the "primary" one, and eventually `assertMacroExpansion()` could be deprecated outright.

Right. I would really like to take the next step as something that we can actually land. An `expectMacroExpansion` method seems reasonable to me and we can clearly document when to use what. It also provides a nice migration path for developers. Not as nice as automatic interop but it exists.

grynspan:
The reason we didn't pursue that solution already is that library authors can't tell which function to call on behalf of their own calers, and there's no reliable `XCTest.isRunning`- or `Testing.isRunning`-style API (for which we have had a bunch of different PRs and issues opened over the last year or so).

0xTim:
> The reason we didn't pursue that solution already is that library authors can't tell which function to call on behalf of their own calers, and there's no reliable `XCTest.isRunning`- or `Testing.isRunning`-style API (for which we have had a bunch of different PRs and issues opened over the last year or so).

If we expose an `expectMacroExpansion` and document it's only expected to be used from Swift Testing, that would be a suitable compromise until the XCTest/Swift Testing interop stuff lands in 6.4 right? Then at that point, we can add support for that in the function and adjust the comment

stmontgomery:
I had a chat with @jerryjrchen and recently, he independently discovered the limitation I [described above](https://github.com/swiftlang/swift-syntax/pull/3192#issuecomment-4237396772). He's preparing an amendment to ST-0021 with a change meant to help solve this. And if that proceeds, it means the original plan I outlined above would once again be viable, and we wouldn't need to introduce a new function here. The amendment would likely be considered for 6.4 as well, which means we could still land this change relatively quickly in a 6.3.x tag as previously planned.

stmontgomery:
> I had a chat with @jerryjrchen […]. He's preparing an amendment to ST-0021 with a change meant to help solve this. And if that proceeds, it means the original plan I outlined above would once again be viable, and we wouldn't need to introduce a new function here. 

That proposal amendment is now up for review: https://forums.swift.org/t/amendment-st-0021-targeted-interoperability-between-swift-testing-and-xctest/86142

I have gone ahead and pushed another commit to this branch which implements the [earlier plan](https://github.com/swiftlang/swift-syntax/pull/3192#issuecomment-4237396772), which should work under this amendment, so it's ready if/when that is approved.

rintaro:
@swift-ci Please test

stmontgomery:
@swift-ci Please test

stmontgomery:
The amendment to ST-0021 was [accepted](https://forums.swift.org/t/amended-st-0021-targeted-interoperability-between-swift-testing-and-xctest/86479) and implemented in Swift 6.4. I just validated that this current PR branch works as expected now that that's in place and everything's looking good for me now locally.

rintaro:
@swift-ci Please test Windows

stmontgomery (approved):

rintaro (approved):
LGTM.

PR #3198, opened 2025-12-02
[SwiftWarningControl] Specify control inputs to tree node in order

artemcm:
Specifically for global controls, the ordering is important because later controls override earlier controls. Similarly, this will result in the same semantic for syntactic controls, with the first attribute being able to be overriden by subsequent attributes.

The control configuration logic as-is handles that correctly *except* for iterating over input controls which are supplied in an unordered dictionary. This refactors the API surface to accept an array of pairs '(groupIdentifier, groupControl)', instead of a dictionary.


artemcm:
@swift-ci test

artemcm:
@swift-ci test macOS platform

PR #3199, opened 2025-12-07
Check for delimiter before lookahead in `advanceIfMultilineStringDelimiter`

hamishknight:
Avoid doing an unnecessary lookahead if we're not even at a multiline delimiter. This also fixes an issue where we'd scan ahead on the next line for `#"\n`, although this was benign in practice since the delimiter check would have previously then failed afterwards. This matches the logic in swiftlang/swift#85889.

### DIFF
--- a/Sources/SwiftParser/Lexer/Cursor.swift
+++ b/Sources/SwiftParser/Lexer/Cursor.swift
@@ -814,11 +814,19 @@ extension Lexer.Cursor {
     openingRawStringDelimiters: Int?
   ) -> Bool {
     precondition(self.previous == #"""#)
+
+    var delimEnd = self
+    guard delimEnd.advance(matching: #"""#),
+      case let delimLast = delimEnd,
+      delimEnd.advance(matching: #"""#)
+    else {
+      return false
+    }
+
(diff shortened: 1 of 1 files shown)


hamishknight:
@swift-ci please test

Sources/SwiftParser/Lexer/Cursor.swift
```diff
@@ -814,11 +814,19 @@ extension Lexer.Cursor {
     openingRawStringDelimiters: Int?
   ) -> Bool {
     precondition(self.previous == #"""#)
+
+    var delimEnd = self
+    guard delimEnd.advance(matching: #"""#),
+      case let delimLast = delimEnd,
```
ahoppen, line 820: Is `case` needed here? If so, why?

```suggestion
      let delimLast = delimEnd,
```

hamishknight: It's not optional, we just want a binding here to avoid needing to split this into 2 separate guard statements

ahoppen: Oh, I never even considered that option. Interesting. Good with me 👍🏽

PR #3211, opened 2025-12-15
[SwiftWarningControl] Add support for file-scoped 'using @warn'

artemcm:
`using @warn(DiagGroupID, as: warning)` 
specifies a file-scoped diagnostic group control. This change adds detection of these controls and upon encountering one, adds a corresponding control to the file-scoped root of the warning control region tree. 

This change also removes the optimization on `warningGroupControlRegionTreeImpl` construction which only constructs a tree that considers nodes which contain a queried location because we now have a scenario where diagnostic group behavior at a specific location can be affected by an `using @warn` node elsewhere in the file. I expect clients to run queries against a once-built tree for the source file, so I do not think this will have a meaningful impact, in practice.

### DIFF
--- a/Sources/SwiftWarningControl/WarningControlRegionBuilder.swift
+++ b/Sources/SwiftWarningControl/WarningControlRegionBuilder.swift
@@ -10,7 +10,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-import SwiftSyntax
+@_spi(ExperimentalLanguageFeatures) import SwiftSyntax
 
 /// Compute the full set of warning control regions in this syntax node
 extension SyntaxProtocol {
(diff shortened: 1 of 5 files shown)


ahoppen (commented):
One high-level comment before I dive into a detailed review: Is `using @warn` only allowed at the top level or also inside declarations? 
- If it is only allowed at the top level: We could still have an API that takes a position and just doesn’t walk into declarations
- If it is allowed inside type declarations: Does the `using` declaration only apply within that type, similar to applying it as an attribute to the type or does it still apply to the whole file?

artemcm:
> * If it is only allowed at the top level: We could still have an API that takes a position and just doesn’t walk into declarations

This is correct, `using @warn` is only allowed at the top level. I've restored the position API and changed the logic to specifically skip Declarations which do not contain the specified position and it seems to work well. Thanks.

artemcm:
@swift-ci test

Tests/SwiftWarningControlTest/WarningControlTests.swift
```diff
@@ -395,15 +423,16 @@ private func assertWarningGroupControl(
...
   diagnosticGroupID: DiagnosticGroupIdentifier,
   states: [String: WarningGroupControl?],
   file: StaticString = #filePath,
   line: UInt = #line
 ) throws {
   // Pull out the markers that we'll use to dig out nodes to query.
   let (markerLocations, source) = extractMarkers(markedSource)
-
-  var parser = Parser(source)
+  let experimentalFeatures = experimentalFeatures ?? Parser.ExperimentalFeatures(rawValue: 0)
```
ahoppen:
```suggestion
  let experimentalFeatures = experimentalFeatures ?? []
```

artemcm:
@ahoppen I am going to have non-top-level `using` ignored for now, since I think diagnosing declaration-nested `using` statements should apply to all of them, rather than just `using @warn`, which is something I can look into later. 
Added a corresponding test.

Sources/SwiftWarningControl/WarningControlRegionBuilder.swift
```diff
@@ -73,14 +73,40 @@ private class WarningControlRegionVisitor: SyntaxAnyVisitor {
...
     }
-    if let withAttributesSyntax = node.asProtocol(WithAttributesSyntax.self) {
-      tree.addWarningControlRegions(for: withAttributesSyntax)
-    }
     return .visitChildren
   }
 }
+
+extension SyntaxProtocol {
+  /// Determines if this syntax node is top-level code (not contained in any declaration).
```
artemcm: @ahoppen is there an existing mechanism to do this by chance?

Sources/SwiftWarningControl/WarningControlRegionBuilder.swift
```diff
@@ -73,14 +73,40 @@ private class WarningControlRegionVisitor: SyntaxAnyVisitor {
...
+  func isTopLevelCode() -> Bool {
+    var current = Syntax(self)
+    while let parent = current.parent {
+      if parent.isProtocol(DeclSyntaxProtocol.self) {
+        return false
+      }
+      current = parent
+    }
```
ahoppen: With this logic, the following `using` statement would be considered top-level, which it probably shouldn’t (`if` is top-level code of a script file here).

```swift
if false {
  using @warn …
}
```

I would instead check if the parent is a `CodeBlockItemSyntax`, its parent is `CodeBlockItemListSyntax` and then, importantly, the parent is a `SourceFileSyntax`

artemcm: That's a good point, thanks. 
I rewrote the check to validate this specific parent chain instead.

artemcm:
@swift-ci test

artemcm:
@swift-ci test

artemcm:
@swift-ci test

artemcm:
@swift-ci test macOS platform

PR #3215, opened 2025-12-22
[LexicalLookup] Make `GenericParameterScopeSyntax` public

rintaro:
Using a method in a extension of non-public protocol to witness a public protocol requirement is not allowed in future version of Swift and is diagnosed as a warning in the current main.

Make `GenericParameterScopeSyntax` a `@_spi(Experimetnal) public` the same accessibility as the concrete type.

Fixes build failures e.g. https://github.com/swiftlang/swift-syntax/actions/runs/20438687858/job/58726284454#step:14:31
since we're building it with `-warnings-as-errors`


rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

PR #3216, opened 2025-12-22
🚥 Introduce a new Lexical query `lookupControlStructure` to the `SwiftLexicalLookup` Library

lokesh-tr:
⚠️ Under Progress ⚠️

- **Explanation**:
Introduces a new Lexical query `lookupControlStructure` to the `SwiftLexicalLookup` Library which answers "which function / closure / init / deinit / accessor gets terminated by this return / throw statement?", also answers "which loop does this break / continue statement correspond to?"
- **Scope**:
Purely Additive. Shouldn't break anything. New API exposed under `@_spi(Experimental)`.
- **Issues**:
This API will be used by SourceKit-LSP to fix (https://github.com/swiftlang/sourcekit-lsp/issues/2400).
- **Original PRs**:
Link to the SourceKit-LSP PR: https://github.com/swiftlang/sourcekit-lsp/pull/2403
- **Risk**:
No Real Risks to the best of my knowledge. This is my first contribution to SwiftSyntax, it's possible that I overlooked something. If that's the case, enlighten me :)
- **Testing**:
Tests and Docs are yet to be written. After some time, will polish this PR description as well.
- **Reviewers**:
Discussed with @ahoppen on some fun things that i can implement in sourcekit-lsp, he suggested this 😁. Open for anyone to review, the current version makes the feature work, might require some polishing.

### DIFF
--- a/Sources/SwiftLexicalLookup/SimpleLookupQueries.swift
+++ b/Sources/SwiftLexicalLookup/SimpleLookupQueries.swift
@@ -13,6 +13,43 @@
 import SwiftSyntax
 
 extension SyntaxProtocol {
+
+  @_spi(Experimental) public func lookupControlStructure() -> Syntax? {
+    let syntax = Syntax(self)
+    
+    switch syntax.as(SyntaxEnum.self) {
+    case .throwStmt, .returnStmt:
+      return lookupEnclosingControlStructure(for: [
+        FunctionDeclSyntax.self,
+        InitializerDeclSyntax.self,
+        DeinitializerDeclSyntax.self,
+        AccessorDeclSyntax.self,
+        ClosureExprSyntax.self,
+      ])
+    case .breakStmt, .continueStmt:
+      return lookupEnclosingControlStructure(for: [
+        WhileStmtSyntax.self,
+        RepeatStmtSyntax.self,
+        ForStmtSyntax.self,
+      ])
+    default:
+      return nil
+    }
+  }
+
+  private func lookupEnclosingControlStructure(for types: [SyntaxProtocol.Type]) -> Syntax? {
+    var nextSyntax: Syntax? = Syntax(self).parent
+    
+    while let currentSyntax = nextSyntax {
+      if types.contains(where: { currentSyntax.is($0) }) {
+        return currentSyntax
+      }
+      nextSyntax = currentSyntax.parent
+    }
+    
+    return nil
+  }
+
   /// Returns all labeled statements available at a particular syntax node.
   ///
   /// - Returns: Available labeled statements at a particular syntax node


ahoppen:
CC @MAJKFL if you’re interested in helping review this and contributing your expertise in name lookup edge cases. 😉

Sources/SwiftLexicalLookup/SimpleLookupQueries.swift
```diff
@@ -13,6 +13,43 @@
 import SwiftSyntax
 
 extension SyntaxProtocol {
+
+  @_spi(Experimental) public func lookupControlStructure() -> Syntax? {
```
rintaro, line 17: I don't feel comfortable with having this method on `SyntaxProtocol`.
Instead, how about making
```swift
@_spi(Experimental) public protocol ControlTransferStmtSyntax {
  var controlTransferTarget: Syntax? { get }
}
```
And make [`break`, `continue`, `return`, `throw`, and `fallthrough`](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/statements#Control-Transfer-Statements) conform to it?

Sources/SwiftLexicalLookup/SimpleLookupQueries.swift
```diff
@@ -13,6 +13,43 @@
...
+    case .throwStmt, .returnStmt:
+      return lookupEnclosingControlStructure(for: [
+        FunctionDeclSyntax.self,
+        InitializerDeclSyntax.self,
+        DeinitializerDeclSyntax.self,
+        AccessorDeclSyntax.self,
+        ClosureExprSyntax.self,
+      ])
+    case .breakStmt, .continueStmt:
+      return lookupEnclosingControlStructure(for: [
```
rintaro, line 30: Unfortunately, `break` and `continue` target is not that simple. You'd need to handle labels e.g. `break LOOP`.
Also, they can target other non-loop labeled statements , e.g.
```swift
OUTER: do {
  for val in values {
    if val.condition {
      break OUTER
    }
  }
}
```
In the actual compiler, the rule is implemented in [`swift::findBreakOrContinueStmtTarget` ](https://github.com/swiftlang/swift/blob/4f19313f1493845d5698d1e9d598b49d3b828f5f/lib/Sema/TypeCheckStmt.cpp#L676-L732)

Sources/SwiftLexicalLookup/SimpleLookupQueries.swift
```diff
@@ -13,6 +13,43 @@
 import SwiftSyntax
 
 extension SyntaxProtocol {
+
+  @_spi(Experimental) public func lookupControlStructure() -> Syntax? {
+    let syntax = Syntax(self)
+    
+    switch syntax.as(SyntaxEnum.self) {
+    case .throwStmt, .returnStmt:
```
rintaro, line 21: The target of `throw` can be do-catch statements.

rintaro (commented):
Thank you!
Overall, I think this is a great addition to SwiftLexicalLookup.

I know this PR is still in draft, but I left some comments anyway.

Sources/SwiftLexicalLookup/SimpleLookupQueries.swift
rintaro: Actually, `SyntaxProtoco.lookupCatchNode()` should be usable for `throw` case.

Sources/SwiftLexicalLookup/SimpleLookupQueries.swift
```diff
@@ -13,6 +13,43 @@
...
+        DeinitializerDeclSyntax.self,
+        AccessorDeclSyntax.self,
+        ClosureExprSyntax.self,
+      ])
+    case .breakStmt, .continueStmt:
+      return lookupEnclosingControlStructure(for: [
+        WhileStmtSyntax.self,
+        RepeatStmtSyntax.self,
+        ForStmtSyntax.self,
+      ])
```
MAJKFL, line 34: I think there is a problem when you consider a nested function like this:
```swift
for i in 0..<42 {
  func foo() {
    break
  }
}
```
The code is obviously not correct because of the `break` inside of the `foo` body, but running the query on the `break` will now yield the enclosing `for` as the targeted construct. Handling this case would improve correctness for clients operating on invalid or intermediate syntax.

Other queries stop at the function boundary via [`SyntaxProtocol.collectNodesOfTypeUpToFunctionBoundary`](https://github.com/swiftlang/swift-syntax/blob/948625fc9fe2c035ff4401a17b1e78baa4c51ff2/Sources/SwiftLexicalLookup/SimpleLookupQueries.swift#L150-L167). Should `lookupEnclosingControlStructure` adopt a similar early-return strategy?

Sources/SwiftLexicalLookup/SimpleLookupQueries.swift
```diff
@@ -13,6 +13,43 @@
...
+    default:
```
MAJKFL, line 35: I think the query is missing a case for `fallthrough`. We could handle it with [`SyntaxProtocol.lookupFallthroughSourceAndDestintation`](https://github.com/swiftlang/swift-syntax/blob/948625fc9fe2c035ff4401a17b1e78baa4c51ff2/Sources/SwiftLexicalLookup/SimpleLookupQueries.swift#L217).

MAJKFL (commented):
Thank you for tagging me in! 

The PR looks like a very nice addition. I’ve left a couple of suggestions.

Out of curiosity, have you thought whether the query could be tested against the compiler? As opposed to the unqualified lookup, simple queries are only covered by the test suite in [`SimpleQueryTests.swift`](https://github.com/swiftlang/swift-syntax/blob/948625fc9fe2c035ff4401a17b1e78baa4c51ff2/Tests/SwiftLexicalLookupTest/SimpleQueryTests.swift) which might not be exhaustive. I think it could be interesting to look into, but it might be an overkill.

PR #3226, opened 2026-01-06
fix: “Convert to computed property” removes leading trivia

PhantomInTheWire:
fixes: https://github.com/swiftlang/sourcekit-lsp/issues/2431
explicitly preserve the leading trivia in refactor()

### DIFF
--- a/Sources/SwiftRefactor/ConvertZeroParameterFunctionToComputedProperty.swift
+++ b/Sources/SwiftRefactor/ConvertZeroParameterFunctionToComputedProperty.swift
@@ -58,12 +58,7 @@ public struct ConvertZeroParameterFunctionToComputedProperty: SyntaxRefactoringP
       rightBrace: body.rightBrace
     )
 
-    let bindingSpecifier = TokenSyntax(
-      .keyword(.var),
-      leadingTrivia: syntax.funcKeyword.leadingTrivia,
-      trailingTrivia: syntax.funcKeyword.trailingTrivia,
-      presence: .present
-    )
+    let bindingSpecifier = syntax.funcKeyword.detached.with(\.tokenKind, .keyword(.var))
 
     let patternBinding = PatternBindingSyntax(
       pattern: variableName,


Sources/SwiftRefactor/ConvertZeroParameterFunctionToComputedProperty.swift
```diff
@@ -60,6 +60,7 @@ public struct ConvertZeroParameterFunctionToComputedProperty: SyntaxRefactoringP
     )
 
     return VariableDeclSyntax(
+      leadingTrivia: syntax.leadingTrivia,
```
ahoppen: Doesn’t this duplicate the leading trivia if the syntax node contains a modifier? I think we just need to transfer the trivia from `func` to `var`.

PhantomInTheWire: good catch. idk why i missed that. I've updated the fix to only transfer trivia from the func keyword to the var keyword (bindingSpecifier) when there are no attributes or modifiers:
```swift
if syntax.attributes.isEmpty && syntax.modifiers.isEmpty {
  result.bindingSpecifier.leadingTrivia = syntax.funcKeyword.leadingTrivia
}
```

PhantomInTheWire:
i've added the tests as requested, everything seems to pass

Sources/SwiftRefactor/ConvertZeroParameterFunctionToComputedProperty.swift
```diff
@@ -59,12 +59,19 @@ public struct ConvertZeroParameterFunctionToComputedProperty: SyntaxRefactoringP
...
       modifiers: syntax.modifiers,
       .var,
       name: variableName,
       type: variableType,
       accessorBlock: accessorBlock
     )
+
+    // Transfer leading trivia from the func keyword to the var keyword.
+    if syntax.attributes.isEmpty && syntax.modifiers.isEmpty {
+      result.bindingSpecifier.leadingTrivia = syntax.funcKeyword.leadingTrivia
```
ahoppen: Is there any reason why we can’t unconditionally transfer the trivia from the `funcKeyword` to the binding specifier and also transfer the trailing trivia so that we don’t loose a comment in eg. `func /* comment */ foo() {}`?

I’m thinking that we might use `syntax.funcKeyword.with(\.tokenKind, .keyword(.var))`. That way we can’t use the current convenience initializer anymore but that also made me notice that we weren’t transferring attributes from the function to the variable, which we should fix as well (a test case for that would also be appreciated).

PhantomInTheWire: sure, added test cases and fixed to use syntax.funcKeyword.with(\.tokenKind, .keyword(.var)).

Sources/SwiftRefactor/ConvertZeroParameterFunctionToComputedProperty.swift
```diff
@@ -59,12 +58,21 @@ public struct ConvertZeroParameterFunctionToComputedProperty: SyntaxRefactoringP
...
-      type: variableType,
+    let bindingSpecifier = syntax.funcKeyword.with(\.tokenKind, .keyword(.var))
+
+    let patternBinding = PatternBindingSyntax(
+      pattern: variableName,
+      typeAnnotation: variableType,
       accessorBlock: accessorBlock
     )
+
+    let result = VariableDeclSyntax(
```
ahoppen: Can be simplified slightly again 😉 

```suggestion
    return VariableDeclSyntax(
```

ahoppen (commented):
Small nitpick, otherwise LGTM.

Sources/SwiftRefactor/ConvertZeroParameterFunctionToComputedProperty.swift
PhantomInTheWire: done.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

PhantomInTheWire:
```
Test Suite 'ConvertZeroParameterFunctionToComputedPropertyTests' started at 2026-01-07 08:53:56.973
Test Case 'ConvertZeroParameterFunctionToComputedPropertyTests.testRefactoringFunctionToComputedProperty' started at 2026-01-07 08:53:56.973
SwiftSyntax/RawSyntaxValidation.swift:1374: Fatal error: Error validating child at index 5 of functionDecl:
Expected token with one of [keyword('func')] but received keyword with text 'var'

See "RawSyntax Validation" in CONTRIBUTING.md to reproduce the failure locally.

*** Signal 4: Backtracing from 0x7f3e599c8628... done ***
```

raw syntax issue should be fixed now

Sources/SwiftRefactor/ConvertZeroParameterFunctionToComputedProperty.swift
```diff
@@ -58,7 +58,12 @@ public struct ConvertZeroParameterFunctionToComputedProperty: SyntaxRefactoringP
       rightBrace: body.rightBrace
     )
 
-    let bindingSpecifier = syntax.funcKeyword.with(\.tokenKind, .keyword(.var))
+    let bindingSpecifier = TokenSyntax(
+      .keyword(.var),
+      leadingTrivia: syntax.funcKeyword.leadingTrivia,
+      trailingTrivia: syntax.funcKeyword.trailingTrivia,
+      presence: .present
+    )
```
ahoppen: Ah, completely forgot about the issue that the keyword still carries around its parent. This should be fixable by just detaching the token from its parent before changing the token kind. This is slightly more concise than creating a new `TokenSyntax`

```swift
let bindingSpecifier = syntax.funcKeyword.detached.with(\.tokenKind, .keyword(.var))
```

PhantomInTheWire: done.

ahoppen:
@swift-ci Please test

ahoppen:
Windows failure is unrelated. Let’s try again

@swift-ci Please test Windows

ahoppen:
Let’s try Windows one more time, failed for a different reason now.

@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

ahoppen:
Windows failed because of https://github.com/swiftlang/swift/pull/86403, trying again.

@swift-ci Please test Windows

PhantomInTheWire:
image
seems like another unrelated ci failure

ahoppen:
We’ll get there one day. Let’s try again.

@swift-ci Please test Windows

PhantomInTheWire:
@ahoppen looks like the day has come

ahoppen:
🎉👯🙌🏽

PR #3228, opened 2026-01-06
[6.3] Implement BasicMacroExpansionContext.buildConfiguration

jmschonfeld:
- **Explanation**: Adds an implementation of `BasicMacroExpansionContext.buildConfiguration` so that unit test-provided build configurations get passed through to generic callers in swift-syntax that use the expansion context
  
- **Scope**: Only impacts uses of `buildConfiguration` on a `BasicMacroExpansionContext` which are typically created for tests
  
- **Issues**:
  
- **Original PRs**: https://github.com/swiftlang/swift-syntax/pull/3191
  
- **Risk**: Low - the scope is small and the change size is minimal
  
- **Testing**: I was not able to add a unit test for this, but the change is trivial and straightforward so I have high confidence it is correct and does not introduce a regression
  
- **Reviewers**: @DougGregor @ahoppen


jmschonfeld:
@swift-ci please test

DougGregor (approved):

PR #3229, opened 2026-01-07
Add new features to ConvertStoredPropertyToComputed

myaumura:
As part of https://github.com/swiftlang/sourcekit-lsp/issues/2424.
Add new functionality to the existing code action analogous to  [ConvertToComputedProperty.cpp](https://github.com/swiftlang/swift/blob/main/lib/Refactoring/ConvertToComputedProperty.cpp)  and [tryComputedPropertyFixIts](https://github.com/swiftlang/swift/blob/main/lib/Sema/CSDiagnostics.cpp#L3659).

- Support removing `lazy` attribute if present.
- Support removing the `=` sign if variable like `var foo: Int = { return 0 }()`.

I'm happy to make changes if needed

### DIFF
--- a/Sources/SwiftRefactor/ConvertStoredPropertyToComputed.swift
+++ b/Sources/SwiftRefactor/ConvertStoredPropertyToComputed.swift
@@ -22,6 +22,14 @@ public struct ConvertStoredPropertyToComputed: SyntaxRefactoringProvider {
       throw RefactoringNotApplicableError("unsupported variable declaration")
     }
 
+    var syntax = syntax
+
+    if let lazyKeyword = syntax.modifiers.first(where: { $0.name.tokenKind == .keyword(.lazy) }) {
+      syntax = DeclModifierRemover { $0.id == lazyKeyword.id }
+        .rewrite(syntax)
+        .cast(VariableDeclSyntax.self)
+    }
+
     var codeBlockSyntax: CodeBlockItemListSyntax
 
     if let functionExpression = initializer.value.as(FunctionCallExprSyntax.self),
--- a/Sources/SwiftRefactor/DeclModifierRemover.swift
+++ b/Sources/SwiftRefactor/DeclModifierRemover.swift
@@ -1,0 +1,70 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+@_spi(RawSyntax) import SwiftSyntax
+
+final class DeclModifierRemover: SyntaxRewriter {
+  private let predicate: (DeclModifierSyntax) -> Bool
+
+  private var triviaToAttachToNextToken: Trivia = Trivia()
+
+  /// Initializes a modifier remover with a given predicate to determine which modifiers to remove.
+  ///
+  /// - Parameter predicate: A closure that determines whether a given `AttributeSyntax` should be removed.
+  ///   If this closure returns `true` for an attribute, that attribute will be removed.
+  public init(removingWhere predicate: @escaping (DeclModifierSyntax) -> Bool) {
+    self.predicate = predicate
+    super.init()
+  }
+
+  public override func visit(_ node: DeclModifierListSyntax) -> DeclModifierListSyntax {
+    var filteredModifiers: [DeclModifierListSyntax.Element] = []
+
+    for modifier in node {
+      guard self.predicate(modifier) else {
+        filteredModifiers.append(prependAndClearAccumulatedTrivia(to: modifier))
+        continue
+      }
+
+      // Removing modifier before comment leaves space before comment intact — doesn’t merge with following trivia.
+      let trailingTrivia = modifier.trailingTrivia.trimmingPrefix(while: \.isSpaceOrTab)
+      triviaToAttachToNextToken += modifier.leadingTrivia.merging(trailingTrivia)
+    }
+
+    if !triviaToAttachToNextToken.isEmpty, !filteredModifiers.isEmpty {
+      filteredModifiers[filteredModifiers.count - 1].trailingTrivia = filteredModifiers[filteredModifiers.count - 1]
+        .trailingTrivia
+        .merging(triviaToAttachToNextToken)
+      triviaToAttachToNextToken = Trivia()
+    }
+
+    return DeclModifierListSyntax(filteredModifiers)
+  }
+
+  public override func visit(_ token: TokenSyntax) -> TokenSyntax {
+    return prependAndClearAccumulatedTrivia(to: token)
+  }
+
+  /// Prepends the accumulated trivia to the given node's leading trivia.
+  ///
+  /// To preserve correct formatting after attribute removal, this function reassigns
+  /// significant trivia accumulated from removed attributes to the provided subsequent node.
+  /// Once attached, the accumulated trivia is cleared.
+  ///
+  /// - Parameter node: The syntax node receiving the accumulated trivia.
+  /// - Returns: The modified syntax node with the prepended trivia.
+  private func prependAndClearAccumulatedTrivia<T: SyntaxProtocol>(to syntaxNode: T) -> T {
+    guard !triviaToAttachToNextToken.isEmpty else { return syntaxNode }
+    defer { triviaToAttachToNextToken = Trivia() }
+    return syntaxNode.with(\.leadingTrivia, triviaToAttachToNextToken.merging(syntaxNode.leadingTrivia))
+  }
+}
--- a/Tests/SwiftRefactorTest/ConvertStoredPropertyToComputedTest.swift
+++ b/Tests/SwiftRefactorTest/ConvertStoredPropertyToComputedTest.swift
@@ -170,6 +170,130 @@ final class ConvertStoredPropertyToComputedTest: XCTestCase {
 
     try assertRefactorConvert(baseline, expected: nil)
   }
+
+  func testRefactoringStoredPropertyWithLazyKeyword() throws {
+    let baseline: DeclSyntax = """
+      lazy var defaultColor: Color = .red
+      """
+
+    let expected: DeclSyntax = """
+      var defaultColor: Color { .red }
+      """
+
+    try assertRefactorConvert(baseline, expected: expected)
+  }
+
+  func testRefactoringStoredPropertyWithModifiers() throws {
+    let baseline: DeclSyntax = """
+      private lazy var defaultColor: Color = .red
+      """
+
+    let expected: DeclSyntax = """
+      private var defaultColor: Color { .red }
+      """
+
+    try assertRefactorConvert(baseline, expected: expected)
+  }
+
+  func testRefactoringStoredPropertyWithModifiers2() throws {
+    let baseline: DeclSyntax = """
+      lazy private var defaultColor: Color = .red
+      """
+
+    let expected: DeclSyntax = """
+      private var defaultColor: Color { .red }
+      """
+
+    try assertRefactorConvert(baseline, expected: expected)
+  }
+
+  func testRefactoringStoredPropertyWithModifiersAndComment() throws {
+    let baseline: DeclSyntax = """
+      lazy /* some comment */ private var defaultColor: Color = .red
+      """
+
+    let expected: DeclSyntax = """
+      /* some comment */ private var defaultColor: Color { .red }
+      """
+
+    try assertRefactorConvert(baseline, expected: expected)
+  }
+
+  func testRefactoringStoredPropertyWithModifiersAndComment2() throws {
+    let baseline: DeclSyntax = """
+      private /* comment */ lazy var defaultColor: Color = .red
+      """
+
+    let expected: DeclSyntax = """
+      private /* comment */ var defaultColor: Color { .red }
+      """
+
+    try assertRefactorConvert(baseline, expected: expected)
+  }
+
+  func testRefactoringStoredPropertyWithModifierAndComment() throws {
+    let baseline: DeclSyntax = """
+      lazy /* comment */ var defaultColor: Color = .red
+      """
+
+    let expected: DeclSyntax = """
+      /* comment */ var defaultColor: Color { .red }
+      """
+
+    try assertRefactorConvert(baseline, expected: expected)
+  }
+
+  func testRefactoringStructStoredPropertiyWithModifiers() throws {
+    let baseline: DeclSyntax = """
+      struct Foo {
+        lazy private var defaultColor: Color = .red
+      }
+      """
+
+    let expected: DeclSyntax = """
+      struct Foo {
+        private var defaultColor: Color { .red }
+      }
+      """
+
+    try assertRefactorStructConvert(baseline, expected: expected)
+  }
+
+  func testRefactoringStructStoredPropertiyWithModifiers2() throws {
+    let baseline: DeclSyntax = """
+      struct Foo {
+        private
+        /* comment */ lazy var defaultColor: Color = .red
+      }
+      """
+
+    let expected: DeclSyntax = """
+      struct Foo {
+        private
+        /* comment */ var defaultColor: Color { .red }
+      }
+      """
+
+    try assertRefactorStructConvert(baseline, expected: expected)
+  }
+
+  func testRefactoringStructStoredPropertiyWithModifiers3() throws {
+    let baseline: DeclSyntax = """
+      struct Foo {
+        private /* comment */
+        /* another comment */ lazy var defaultColor: Color = .red
+      }
+      """
+
+    let expected: DeclSyntax = """
+      struct Foo {
+        private /* comment */
+        /* another comment */ var defaultColor: Color { .red }
+      }
+      """
+
+    try assertRefactorStructConvert(baseline, expected: expected)
+  }
 }
 
 private func assertRefactorConvert(
@@ -186,4 +310,24 @@ private func assertRefactorConvert(
     file: file,
     line: line
   )
+}
+
+private func assertRefactorStructConvert(
+  _ callDecl: DeclSyntax,
+  expected: DeclSyntax,
+  file: StaticString = #filePath,
+  line: UInt = #line
+) throws {
+
+  let structCallDecl = try XCTUnwrap(callDecl.as(StructDeclSyntax.self))
+  let variable = try XCTUnwrap(structCallDecl.memberBlock.members.first?.decl.as(VariableDeclSyntax.self))
+  let refactored = try ConvertStoredPropertyToComputed.refactor(syntax: variable, in: ())
+
+  let members = MemberBlockItemListSyntax {
+    MemberBlockItemSyntax(decl: DeclSyntax(refactored))
+  }
+
+  let refactoredMemberBlock = structCallDecl.memberBlock.with(\.members, members)
+  let refactoredStruct = structCallDecl.with(\.memberBlock, refactoredMemberBlock)
+  assertStringsEqualWithDiff(refactoredStruct.description, expected.description)
 }
--- a/Sources/SwiftRefactor/CMakeLists.txt
+++ b/Sources/SwiftRefactor/CMakeLists.txt
@@ -14,6 +14,7 @@ add_swift_syntax_library(SwiftRefactor
   ConvertComputedPropertyToZeroParameterFunction.swift
   ConvertStoredPropertyToComputed.swift
   ConvertZeroParameterFunctionToComputedProperty.swift
+  DeclModifierRemover.swift
   ExpandEditorPlaceholder.swift
   FormatRawStringLiteral.swift
   IntegerLiteralUtilities.swift
--- a/Sources/SwiftSyntax/Trivia.swift
+++ b/Sources/SwiftSyntax/Trivia.swift
@@ -215,3 +215,27 @@ extension RawTriviaPiece: CustomDebugStringConvertible {
     TriviaPiece(raw: self).debugDescription
   }
 }
+
+@_spi(RawSyntax)
+public extension Trivia {
+  func trimmingPrefix(
+    while predicate: (TriviaPiece) -> Bool
+  ) -> Trivia {
+    Trivia(pieces: self.drop(while: predicate))
+  }
+
+  func trimmingSuffix(
+    while predicate: (TriviaPiece) -> Bool
+  ) -> Trivia {
+    Trivia(
+      pieces: self[...]
+        .reversed()
+        .drop(while: predicate)
+        .reversed()
+    )
+  }
+
+  var startsWithNewline: Bool {
+    self.first?.isNewline ?? false
+  }
+}
--- a/Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
+++ b/Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift
@@ -14,14 +14,14 @@
 internal import SwiftDiagnostics
 internal import SwiftOperators
 @_spi(MacroExpansion) internal import SwiftParser
-public import SwiftSyntax
+@_spi(RawSyntax) public import SwiftSyntax
 internal import SwiftSyntaxBuilder
 @_spi(MacroExpansion) @_spi(ExperimentalLanguageFeature) public import SwiftSyntaxMacros
 #else
 import SwiftDiagnostics
 import SwiftOperators
 @_spi(MacroExpansion) import SwiftParser
-import SwiftSyntax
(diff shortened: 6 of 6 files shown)


Sources/SwiftRefactor/ConvertStoredPropertyToComputed.swift
```diff
@@ -41,16 +41,33 @@ public struct ConvertStoredPropertyToComputed: SyntaxRefactoringProvider {
         closureExpression.trailingTrivia + closureExpression.rightBrace.leadingTrivia
         + closureExpression.rightBrace.trailingTrivia + functionExpression.trailingTrivia
     } else {
-      var body = CodeBlockItemListSyntax([
-        CodeBlockItemSyntax(
-          item: .expr(initializer.value)
-        )
-      ])
+
+      var body = CodeBlockItemListSyntax([])
```
ahoppen: I don’t think you need to initialize `body` here, initializing it in both branches should be sufficient.

```suggestion
      var body: CodeBlockItemListSyntax
```

Sources/SwiftRefactor/ConvertStoredPropertyToComputed.swift
```diff
@@ -41,16 +41,33 @@ public struct ConvertStoredPropertyToComputed: SyntaxRefactoringProvider {
...
       body.leadingTrivia = initializer.equal.trailingTrivia + body.leadingTrivia
-      body.trailingTrivia += .space
       codeBlockSyntax = body
     }
 
+    var modifiers = syntax.modifiers
+
+    if let lazyKeyword = modifiers.first(where: { decl in
+      decl.name.tokenKind == .keyword(.lazy)
+    }), let index = modifiers.index(of: lazyKeyword) {
```
ahoppen: Nitpick, I think this reads a little nicer as

```suggestion
    if let lazyKeyword = modifiers.first(where: { $0.name.tokenKind == .keyword(.lazy) })
      let index = modifiers.index(of: lazyKeyword) 
    {
```

Tests/SwiftRefactorTest/ConvertStoredPropertyToComputedTest.swift
```diff
@@ -170,6 +170,48 @@ final class ConvertStoredPropertyToComputedTest: XCTestCase {
...
+    try assertRefactorConvert(baseline, expected: expected)
+  }
+
+  func testRefactoringStoredPropertyWithClosureStatements() throws {
+    let baseline: DeclSyntax = """
+      static let defaultColor: Color = { 
+        let someColor = .red
+        return someColor
+      }
+      """
```
ahoppen, line 201: Shouldn’t the test case input be

```suggestion
    let baseline: DeclSyntax = """
      static let defaultColor: Color = { 
        let someColor = .red
        return someColor
      }()
      """
```

Currently, this isn’t valid Swift code because you’re trying to initialize a variable of type `Color` with a closure type.

ahoppen (commented):
Thanks for your contribution, @myaumura.

myaumura:
@ahoppen, Thanks for your review!

Sources/SwiftRefactor/ConvertStoredPropertyToComputed.swift
```diff
@@ -41,16 +41,32 @@ public struct ConvertStoredPropertyToComputed: SyntaxRefactoringProvider {
...
     } else {
-      var body = CodeBlockItemListSyntax([
-        CodeBlockItemSyntax(
-          item: .expr(initializer.value)
-        )
-      ])
+
+      var body: CodeBlockItemListSyntax
+
+      if let closure = initializer.value.as(ClosureExprSyntax.self) {
```
ahoppen: I just looked at this again. I don’t think we ever hit this case. We already checked for in line 28. So, I think all this PR really does is remove the `lazy` keyword.

Tests/SwiftRefactorTest/ConvertStoredPropertyToComputedTest.swift
```diff
@@ -170,6 +170,48 @@ final class ConvertStoredPropertyToComputedTest: XCTestCase {
...
+
+    let expected: DeclSyntax = """
+      static var defaultColor: Color { .red }
+      """
+
+    try assertRefactorConvert(baseline, expected: expected)
+  }
+
+  func testRefactoringStoredPropertyWithClosureStatements() throws {
```
ahoppen: Just looked at this again. Isn’t this exactly what we’re testing in `testRefactoringStoredPropertyWithMultipleStatements`?

Tests/SwiftRefactorTest/ConvertStoredPropertyToComputedTest.swift
```diff
@@ -170,6 +170,48 @@ final class ConvertStoredPropertyToComputedTest: XCTestCase {
...
+      """
+
+    let expected: DeclSyntax = """
+      var defaultColor: Color { .red }
+      """
+
+    try assertRefactorConvert(baseline, expected: expected)
+  }
+
+  func testRefactoringStoredPropertyWithClosure() throws {
```
ahoppen: Isn’t this already covered by `testRefactoringStoredPropertyWithMultipleStatements`?

Tests/SwiftRefactorTest/ConvertStoredPropertyToComputedTest.swift
```diff
@@ -170,6 +170,48 @@ final class ConvertStoredPropertyToComputedTest: XCTestCase {
 
     try assertRefactorConvert(baseline, expected: nil)
   }
+
+  func testRefactoringStoredPropertyWithLazyKeyword() throws {
+    let baseline: DeclSyntax = """
+      lazy var defaultColor: Color = .red
+      """
```
ahoppen, line 177: Could you test that we don’t remove necessary leading trivia of the `lazy` keyword, eg. if the test case is as follows?

```suggestion
    let baseline: DeclSyntax = """
      struct Foo {
        lazy var defaultColor: Color = .red
      }
      """
```

ahoppen (commented):
Looking at this again, I noticed that the `= { … }()` should already be handled.

Sources/SwiftRefactor/ConvertStoredPropertyToComputed.swift
myaumura: Sorry, my fault. I tested with an incorrect example that got parsed as a closure

Tests/SwiftRefactorTest/ConvertStoredPropertyToComputedTest.swift
myaumura: added support for leading trivia

Sources/SwiftRefactor/ConvertStoredPropertyToComputed.swift
```diff
@@ -46,11 +46,21 @@ public struct ConvertStoredPropertyToComputed: SyntaxRefactoringProvider {
...
     }
 
+    var modifiers = syntax.modifiers
+    var modifiersLeadingTrivia = Trivia()
+
+    if let lazyKeyword = modifiers.first(where: { $0.name.tokenKind == .keyword(.lazy) }),
+      let index = modifiers.index(of: lazyKeyword)
+    {
+      modifiers.remove(at: index)
+      modifiersLeadingTrivia = lazyKeyword.leadingTrivia
```
ahoppen: Instead of adding the trivia before the new binding specifier, I’d add it to the leading trivia of the next token. That way we don’t unexpectedly move trivia eg. in

```swift 
struct Foo {
  private 
  /* comment */ lazy var defaultColor: Color = .red
}
```

I think we have two options here: 
1. Only transfer the trivia of `lazy` if it is the first token in the variable declaration and don’t worry about trivia inside the declaration. This is the easiest solution, slightly incorrect but probably still good enough.
2. Implement a `SyntaxRewriter` that is able to remove nodes and properly transfer their trivia to the next token. This will be a little more work but could be useful for other refactoring actions in the future as well (`Trivia.merging` will be useful to avoid duplicate spaces because of this).

Tests/SwiftRefactorTest/ConvertStoredPropertyToComputedTest.swift
ahoppen: Could you add a test case for this?

Sources/SwiftRefactor/ConvertStoredPropertyToComputed.swift
```diff
@@ -46,11 +46,21 @@ public struct ConvertStoredPropertyToComputed: SyntaxRefactoringProvider {
           item: .expr(initializer.value)
         )
       ])
-      body.leadingTrivia = initializer.equal.trailingTrivia + body.leadingTrivia
       body.trailingTrivia += .space
+      body.leadingTrivia = initializer.equal.trailingTrivia + body.leadingTrivia
```
ahoppen, line 57: Any reason why you changed this order?

Sources/SwiftRefactor/ConvertStoredPropertyToComputed.swift
myaumura: Good evening, @ahoppen! I found the [`AttributeRemover`](https://github.com/swiftlang/swift-syntax/blob/d1bab1f815d193e6343eb56cf2f477165414911f/Sources/SwiftSyntaxMacroExpansion/MacroSystem.swift#L525) implementation, which seems similar to what’s needed. Would it be better to extend it to support new types and move it out of the macros module, or to implement a new one specifically for this module?

ahoppen: Nice find, I completely forgot that we had that implementation. Generalizing that would be the best way to go. I remember getting `AttributeRemover` to function properly was more pain than you’d think and sharing the implementations means we can’t make the same mistakes twice.

Sources/SwiftSyntax/NodeRemover.swift
myaumura, line 1: I’m not sure it should be in this module

Sources/SwiftRefactor/ConvertStoredPropertyToComputed.swift
myaumura: Updated the implementation. Happy to fix anything if needed

Tests/SwiftRefactorTest/ConvertStoredPropertyToComputedTest.swift
myaumura: Done

Sources/SwiftSyntax/NodeRemover.swift
```diff
@@ -0,0 +1,161 @@
...
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+
+/// Removes attributes from a syntax tree while maintaining their surrounding trivia.
+@_spi(Testing)
+public class NodeRemover: SyntaxRewriter {
```
ahoppen: Let’s change this to the `package` access level. I think we didn’t use to have it when this was written.

Sources/SwiftSyntax/NodeRemover.swift
```diff
@@ -0,0 +1,161 @@
...
+//
+// Copyright (c) 2014 - 2025 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
```
ahoppen: swift-syntax must not depend on Foundation because it’s lower in the stack than `Foundation`. Could you remove the Foundation dependency?

Sources/SwiftSyntax/NodeRemover.swift
```diff
@@ -0,0 +1,161 @@
...
+      } else if let lastAttribute = filteredAttributes.last {
+        filteredAttributes[filteredAttributes.count - 1].trailingTrivia = lastAttribute
+          .trailingTrivia
+          .trimmingSuffix(while: \.isSpaceOrTab)
+      }
+    }
+    return AttributeListSyntax(filteredAttributes)
+  }
+
+  public override func visit(_ node: DeclModifierListSyntax) -> DeclModifierListSyntax {
```
ahoppen: Hmm, I was hoping that we could extend `NodeRemover` to be able to remove any arbitrary node. This doesn’t seem to be the case right now and I don’t want to have a semi-general type that promises to remove arbitrary nodes but isn’t actually able to do so. Maybe the better solution in that case is to extract all the common functionality (like the trivia convenience accessors) to the SwiftSyntax module and write another dedicated rewriter to remove decl modifiers.

Sources/SwiftSyntax/NodeRemover.swift
```diff
@@ -0,0 +1,161 @@
...
+    var filteredModifiers: [DeclModifierListSyntax.Element] = []
+
+    for modifier in node {
+      if self.predicate(Syntax(modifier)) {
```
ahoppen: Would this read easier as a `guard self.predicate(Syntax(modifier)) else {`? At least it saves an indentation level.

Sources/SwiftSyntax/NodeRemover.swift
```diff
@@ -0,0 +1,161 @@
...
+        let trailingTrivia = modifier.trailingTrivia.trimmingPrefix(while: \.isSpaceOrTab)
```
ahoppen: Do we need to trim the spaces and tabs here. Shouldn’t that be handled already when merging `triviaToAttachToNextToken` into existing trivia? Or am I missing something?

Sources/SwiftSyntax/NodeRemover.swift
```diff
@@ -0,0 +1,161 @@
...
+        triviaToAttachToNextToken += modifier.leadingTrivia.merging(trailingTrivia)
+
+        if triviaToAttachToNextToken.isEmpty,
+          let previousToken = modifier.previousToken(viewMode: .sourceAccurate),
+          previousToken.trailingTrivia.isEmpty
+        {
+          triviaToAttachToNextToken = modifier.trailingTrivia
+        }
```
ahoppen: Why is this check necessary? Could you maybe add a doc comment that explains it?

Sources/SwiftSyntax/NodeRemover.swift
```diff
@@ -0,0 +1,161 @@
...
+  ///
+  /// To preserve correct formatting after attribute removal, this function reassigns
+  /// significant trivia accumulated from removed attributes to the provided subsequent node.
+  /// Once attached, the accumulated trivia is cleared.
+  ///
+  /// - Parameter node: The syntax node receiving the accumulated trivia.
+  /// - Returns: The modified syntax node with the prepended trivia.
+  private func prependAndClearAccumulatedTrivia<T: SyntaxProtocol>(to syntaxNode: T) -> T {
+    defer { triviaToAttachToNextToken = Trivia() }
+    return syntaxNode.with(\.leadingTrivia, triviaToAttachToNextToken + syntaxNode.leadingTrivia)
```
ahoppen: I think we should get a little better performance if we don’t modify the node if `triviaToAttachToNextToken` is empty (that way we don’t need to allocate a new node), ie. just do an early exist if `triviaToAttachToNextToken` is empty.

Sources/SwiftRefactor/ConvertStoredPropertyToComputed.swift
```diff
@@ -51,6 +51,15 @@ public struct ConvertStoredPropertyToComputed: SyntaxRefactoringProvider {
       codeBlockSyntax = body
     }
 
+    var modifiers = syntax.modifiers
+
+    if let lazyKeyword = modifiers.first(where: { $0.name.tokenKind == .keyword(.lazy) }) {
+      modifiers = NodeRemover {
+        guard let keyword = $0.as(DeclModifierSyntax.self) else { return false }
+        return keyword.name.tokenKind == lazyKeyword.name.tokenKind
```
ahoppen: Shouldn’t we be able to use the identity on `Syntax` to find the token to remove, ie. do something like the following

```suggestion
        $0.id == lazyKeyword.id
```

Sources/SwiftSyntax/NodeRemover.swift
myaumura: yeah, my miss

Sources/SwiftSyntax/NodeRemover.swift
myaumura: I’ll try to work on this idea

Sources/SwiftSyntax/NodeRemover.swift
myaumura: When `lazy` is first before the comment, removing `lazy` leaves the space before the comment – it doesn't merge with trivia after comment.

case handled in test `testRefactoringStoredPropertyWithModifiersAndComment`

ahoppen: Oh, I see. That makes sense. Thanks for the explanation. Can you add that as a comment for future reference?

Sources/SwiftSyntax/NodeRemover.swift
myaumura: I’d like to clarify — is the current purpose to extract the base logic into a `NodeRemover` class and have `AttributeRemover` and `DeclModifierRemover` inherit from it, with only strictly typed  `visit`  methods remaining in them?

ahoppen: No, I’m not proposing a type hierarchy. Essentially just create two separate classes and whatever you can factor out, factor out. I was mostly talking about at least sharing `Trivia.trimmingPrefix` etc.

Sources/SwiftSyntax/NodeRemover.swift
myaumura: removed this code

Sources/SwiftSyntax/NodeRemover.swift
myaumura: Added comment

Sources/SwiftSyntax/NodeRemover.swift
myaumura: done

Sources/SwiftSyntax/NodeRemover.swift
myaumura: Hope I got it right this time, could you please check and sorry for the previous misunderstanding

Sources/SwiftRefactor/DeclModifierRemover.swift
```diff
@@ -0,0 +1,65 @@
...
+  public override func visit(_ node: DeclModifierListSyntax) -> DeclModifierListSyntax {
+    var filteredModifiers: [DeclModifierListSyntax.Element] = []
+
+    for modifier in node {
+      guard self.predicate(modifier) else {
+        filteredModifiers.append(prependAndClearAccumulatedTrivia(to: modifier))
+        continue
+      }
+
+      //Removing modifier before comment leaves space before comment intact — doesn’t merge with following trivia.
```
ahoppen: Nitpick

```suggestion
      // Removing modifier before comment leaves space before comment intact — doesn’t merge with following trivia.
```

Sources/SwiftRefactor/ConvertStoredPropertyToComputed.swift
```diff
@@ -51,6 +51,14 @@ public struct ConvertStoredPropertyToComputed: SyntaxRefactoringProvider {
       codeBlockSyntax = body
     }
 
+    var modifiers = syntax.modifiers
+
+    if let lazyKeyword = modifiers.first(where: { $0.name.tokenKind == .keyword(.lazy) }) {
+      modifiers = DeclModifierRemover {
+        $0.id == lazyKeyword.id
+      }.rewrite(syntax.modifiers).as(DeclModifierListSyntax.self)!
```
ahoppen: Nitpick, we generally prefer `cast` instead of force-unwrapping `as`. Just a style decision.

```suggestion
      }.rewrite(syntax.modifiers).cast(DeclModifierListSyntax.self)
```

Sources/SwiftRefactor/DeclModifierRemover.swift
```diff
@@ -0,0 +1,65 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
+package class DeclModifierRemover: SyntaxRewriter {
```
ahoppen: I think this can now be `internal` instead of `package`. At which point you can also remove the access modifiers from all members.

Sources/SwiftRefactor/DeclModifierRemover.swift
```diff
@@ -0,0 +1,65 @@
...
+  let predicate: (DeclModifierSyntax) -> Bool
+
+  var triviaToAttachToNextToken: Trivia = Trivia()
+
```
ahoppen: Let’s make these two `private`. It helps with local reasoning if you know they can’t be modified/accessed from the outside.

Sources/SwiftRefactor/DeclModifierRemover.swift
```diff
@@ -0,0 +1,65 @@
...
+    if !triviaToAttachToNextToken.isEmpty, !filteredModifiers.isEmpty {
+      filteredModifiers[filteredModifiers.count - 1].trailingTrivia = filteredModifiers[filteredModifiers.count - 1]
+        .trailingTrivia
+        .merging(triviaToAttachToNextToken)
+    }
+
+    return DeclModifierListSyntax(filteredModifiers)
+  }
+
```
ahoppen, line 52: Do we also need a `visit` function for `Token` that calls `prependAndClearAccumulatedTrivia` to transfer trivia to the token after the last modifier, eg. in `lazy /* comment */ var test: Int`?

Sources/SwiftRefactor/DeclModifierRemover.swift
```diff
@@ -0,0 +1,65 @@
...
+  /// To preserve correct formatting after attribute removal, this function reassigns
+  /// significant trivia accumulated from removed attributes to the provided subsequent node.
+  /// Once attached, the accumulated trivia is cleared.
+  ///
+  /// - Parameter node: The syntax node receiving the accumulated trivia.
+  /// - Returns: The modified syntax node with the prepended trivia.
+  private func prependAndClearAccumulatedTrivia<T: SyntaxProtocol>(to syntaxNode: T) -> T {
+    guard !triviaToAttachToNextToken.isEmpty else { return syntaxNode }
+    defer { triviaToAttachToNextToken = Trivia() }
+    return syntaxNode.with(\.leadingTrivia, triviaToAttachToNextToken + syntaxNode.leadingTrivia)
```
ahoppen: Should we call `merging` here? I don’t have a concrete example because you’re already removing spaces and tabs above but it just generally yields better results when unifying two trivias.

```suggestion
    return syntaxNode.with(\.leadingTrivia, triviaToAttachToNextToken.merging(syntaxNode.leadingTrivia))
```

Sources/SwiftRefactor/DeclModifierRemover.swift
ahoppen, line 1: Could you add this to CMakeLists.txt for the Windows build?

Sources/SwiftRefactor/DeclModifierRemover.swift
ahoppen: Could you add a test case for this so we don’t regress it in the future?

ahoppen:
Just https://github.com/swiftlang/swift-syntax/pull/3229#discussion_r2700804687 left, after that this is good to go 🚀

Sources/SwiftRefactor/DeclModifierRemover.swift
myaumura: Done

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
@myaumura We can ignore the Bazel build failure, it is because of https://github.com/swiftlang/swift-syntax/pull/3232#discussion_r2696297914.

myaumura:
@ahoppen, Thank you for the thorough review and detailed explanations—they really helped clarify the issues! 

Should I squash commits?

ahoppen:
Yes, squashing would be appreciated, it makes for a cleaner git history.

And it appears that the CMake build doesn’t support the `package` access level yet either (not just the Bazel build). Could you use `@_spi(RawSyntax) public` in the meantime instead?

myaumura:
Sure, doing it now!

ahoppen:
@swift-ci Please test

ahoppen:
Also FWIW, adding support for the `package` access modifier in https://github.com/swiftlang/swift-syntax/pull/3239

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

ahoppen:
Congrats to your first contribution to Swift, @myaumura 🚀

PR #3230, opened 2026-01-11
Add docCommentValue property to Trivia

PhantomInTheWire:
fixes: https://github.com/swiftlang/swift-syntax/issues/1890

- Adds Trivia.docCommentValue to conveniently extract the text content of the last documentation comment in a trivia collection.
- Automatically strips comment markers (///, /**, */) and intelligently removes common indentation from multi-line block comments.
- Implements stateful parsing to correctly handle section breaks, ensuring only the most immediate doc comment is processed.
- Includes extensive test coverage for various scenarios including mixed comment styles, complex indentation, and interleaved non-doc trivia.

### DIFF
--- a/Sources/SwiftSyntax/Trivia+commentValue.swift
+++ b/Sources/SwiftSyntax/Trivia+commentValue.swift
@@ -68,8 +68,11 @@ extension Trivia {
       // because it won't have any indentation since it starts with /**
       let indentation = lines.dropFirst()
         .map { $0.prefix(while: { $0 == " " || $0 == "\t" }) }
-        .reduce(nil as Substring?) { (acc: Substring?, element: Substring.SubSequence) in
-          acc.map { commonPrefix($0, element) } ?? element
+        .reduce(nil as Substring?) { (acc: Substring?, element: Substring) in
+          guard let acc else {
+            return element
+          }
+          return commonPrefix(acc, element)
         }
 
       guard let firstLine = lines.first else {


ahoppen:
If I understand it correctly, this PR is based on top of https://github.com/swiftlang/swift-syntax/pull/2966, right? Thanks for picking that stale PR up! To make sure we give @zyadtaha credit where credit is due for all the work he put into the original PR, would you mind basing you changes on top of his branch so that we retain his original commit and keep him listed as that commit’s *author*? The author field should be retained if you cherry-pick a commit or rebase it (feel free to read up on the distinction between *committer* and *author* in git if you are interested 😉). 

A real review will follow, I think it might be easier to get the git logistics sorted out first though.

PhantomInTheWire:
@ahoppen I think this should be good now

Sources/SwiftSyntax/Trivia+commentValue.swift
```diff
@@ -14,26 +14,17 @@ extension Trivia {
   /// The contents of the last doc comment piece with any comment markers removed and indentation whitespace stripped.
   public var docCommentValue: String? {
     var comments: [Substring] = []
-
-    /// Keep track of whether we have seen a line or block comment trivia piece.
-    var hasBlockComment = false
-
     var currentLineComments: [Substring] = []
```
ahoppen: Does this comment still apply or did you not address it for a reason? https://github.com/swiftlang/swift-syntax/pull/2966#discussion_r2185056408

Sources/SwiftSyntax/Trivia+commentValue.swift
```diff
@@ -43,9 +34,15 @@ extension Trivia {
...
+
+      case .newlines(let n), .carriageReturns(let n), .carriageReturnLineFeeds(let n):
+        if n == 1 {
+          consecutiveNewlines += 1
+        } else {
+          consecutiveNewlines = 0
+        }
+
+        if consecutiveNewlines != 1 {
           processSectionBreak()
```
ahoppen, line 42: Should we reset `consecutiveNewlines` to 0 here? At which point the `if` above could just become `consecutiveNewlines += n` again.

Sources/SwiftSyntax/Trivia+commentValue.swift
```diff
@@ -59,71 +56,51 @@ extension Trivia {
       var lines = text.dropPrefix("/**").dropSuffix("*/")
         .split(omittingEmptySubsequences: false, whereSeparator: \.isNewline)
 
-      // If the comment content starts on the same line as the `/**` marker or ends on the same line as the `*/` marker,
-      // it is common to separate the marker and the actual comment using spaces. Strip those spaces if they exists.
-      // If there are non no-space characters on the first / last line, then the comment doesn't start / end on the line
-      // with the marker, so don't do the stripping.
```
ahoppen: Any particular reason why you removed this comment? Same for the other comments in the file you removed.

Sources/SwiftSyntax/Trivia+commentValue.swift
```diff
@@ -59,71 +56,51 @@ extension Trivia {
...
-        // but be safe and return `nil` here anyway.
         return nil
       }
 
       var unindentedLines = [firstLine] + lines.dropFirst().map { $0.dropPrefix(indentation ?? "") }
 
-      // If the first line only contained the comment marker, don't include it. We don't want to start the comment value
-      // with a newline if `/**` is on its own line. Same for the end marker.
-      if unindentedLines.first?.allSatisfy({ $0 == " " }) ?? false {
+      while unindentedLines.first?.allSatisfy({ $0 == " " }) == true {
```
ahoppen: If I read the comment above correctly, we explicitly only wanted to skip the first line, not any following and thus this was an `if` instead of a `while` on purpose. Any reason why you changed it?

Same below.

Release Notes/602.md
```diff
@@ -25,6 +25,10 @@
   - Pull request: https://github.com/swiftlang/swift-syntax/pull/3030
   - Migration stems: None required.
 
+- `Trivia` has a new `docCommentValue` property.
```
ahoppen: Could you start a release notes document for 604 where this would go?

Tests/SwiftSyntaxTest/TriviaCommentValueTests.swift
```diff
@@ -0,0 +1,280 @@
...
+
+  func testDocLineCommentValues() {
+    assertCommentValue("///", docCommentValue: "")
+    assertCommentValue("/// Some doc line comment", docCommentValue: "Some doc line comment")
+    assertCommentValue(
+      """
+      /// Some doc line comment
+      /// Another
+      """,
+      docCommentValue: "Some doc line comment\nAnother"
```
ahoppen: This would also make a nice multi-line string literal instead of using `\n`. Same for other string literals with `\n` in them.

Sources/SwiftSyntax/Trivia+commentValue.swift
PhantomInTheWire: I think I confused this with this https://github.com/swiftlang/swift-syntax/pull/2966#discussion_r1970432891 and made both [Substring], but if i understand this correctly now we want comments to be [Substring] and currentLineComments to be [String] right? that said I think having both as [String] or [Substring] looks cleaner

Sources/SwiftSyntax/Trivia+commentValue.swift
PhantomInTheWire: i think i messed up here. I've reverted this back to if statements and added a regression test that verifies blank lines within the comment content are preserved.

Sources/SwiftSyntax/Trivia+commentValue.swift
PhantomInTheWire: I initially did not start with the old commit as a base, but I can add them

Sources/SwiftSyntax/Trivia+commentValue.swift
PhantomInTheWire: makes sense. done.

Tests/SwiftSyntaxTest/TriviaCommentValueTests.swift
PhantomInTheWire: done.

Release Notes/602.md
PhantomInTheWire: sure done.

Sources/SwiftSyntax/Trivia+commentValue.swift
ahoppen: Ah, I thought we could get rid of the `[…]` altogether but I was mistaking. Anyway, I think it’s slightly nicer the way we have it now.

Sources/SwiftSyntax/Trivia+commentValue.swift
ahoppen: Would you mind squashing the commits you authored so we have a cleaner git history that preserves @zyadtaha as the author of these comments?

Sources/SwiftSyntax/Trivia+commentValue.swift
```diff
@@ -70,17 +64,13 @@ extension Trivia {
...
       // because it won't have any indentation since it starts with /**
-      for line in lines.dropFirst() {
-        let lineIndentation = line.prefix(while: { $0 == " " || $0 == "\t" })
-        guard let previousIndentation = indentation else {
-          indentation = lineIndentation
-          continue
+      let indentation = lines.dropFirst()
+        .map { $0.prefix(while: { $0 == " " || $0 == "\t" }) }
+        .reduce(nil as Substring?) { (acc: Substring?, element: Substring.SubSequence) in
+          acc.map { commonPrefix($0, element) } ?? element
```
ahoppen: I have a personal dislike for `map` on optionals because `map` really implies collections to me and usually find that alternative spellings are cleaner. Ie. in this case 

```swift
guard let acc else {
  return element
}
return commonPrefix(acc, element)
```

Sources/SwiftSyntax/Trivia+commentValue.swift
```diff
@@ -70,17 +64,13 @@ extension Trivia {
...
       // Find the lowest indentation that is common among all lines in the block comment. Do not consider the first line
       // because it won't have any indentation since it starts with /**
-      for line in lines.dropFirst() {
-        let lineIndentation = line.prefix(while: { $0 == " " || $0 == "\t" })
-        guard let previousIndentation = indentation else {
-          indentation = lineIndentation
-          continue
+      let indentation = lines.dropFirst()
+        .map { $0.prefix(while: { $0 == " " || $0 == "\t" }) }
+        .reduce(nil as Substring?) { (acc: Substring?, element: Substring.SubSequence) in
```
ahoppen: The subsequence of `Substring` is `Substring`, so this can be simplified to

```swift
 .reduce(nil as Substring?) { (acc: Substring?, element: Substring) in
```

ahoppen (commented):
Just two style comments, otherwise LGTM.

Sources/SwiftSyntax/Trivia+commentValue.swift
PhantomInTheWire: done.

Sources/SwiftSyntax/Trivia+commentValue.swift
PhantomInTheWire: makes sense. done.

PhantomInTheWire:
@swift-ci please test

Sources/SwiftSyntax/Trivia+commentValue.swift
PhantomInTheWire: cool

PhantomInTheWire:
@swift-ci test macos

PhantomInTheWire:
@swift-ci test linux

PhantomInTheWire:
@swift-ci please test macos

PhantomInTheWire:
RFC: https://forums.swift.org/t/rfc-add-trivia-doccommentvalue-property/84192

PR #1527, opened 2023-04-13
Add documentation for `InitializerDeclSyntax` and all its children

ahoppen:
Mostly a prototype of how the documentation for a syntax node could look like.

PR #3231, opened 2026-01-12
Documentation for Switch, Return, Throw and Type syntax nodes

Padmashree06:
This PR adds documentation comments for the following Syntax nodes:

- `SwitchCaseLabelSyntax`
-  `SwitchDefaultLabelSyntax`
-  `SwitchCaseSyntax`
-  `ThrowStmtSyntax`
-  `ReturnClauseSyntax`
-  `ReturnStmtSyntax`
-  `TypeInitializerClauseSyntax`
-  `TypeAnnotationSyntax`

The documentation follows the format used in #1527 and includes
brief explanations and examples where applicable.

This contributes to issue #1528.

### DIFF
--- a/Release Notes/604.md
+++ b/Release Notes/604.md
@@ -1,0 +1,25 @@
+# Swift Syntax 604 Release Notes
+
+## New APIs
+
+- `Trivia` has a new `docCommentValue` property.
+  - Description: Extracts sanitized comment text from doc comment trivia pieces, omitting leading comment markers (`///`, `/**`).
+  - Pull Request: https://github.com/swiftlang/swift-syntax/pull/2966
+
+
+## API Behavior Changes
+
+## Deprecations
+
+## API-Incompatible Changes
+
+## Template
+
+- *Affected API or two word description*
+  - Description: *A 1-2 sentence description of the new/modified API*
+  - Issue: *If an issue exists for this change, a link to the issue*
+  - Pull Request: *Link to the pull request(s) that introduces this change*
+  - Migration steps: Steps that adopters of swift-syntax should take to move to the new API (required for deprecations and API-incompatible changes).
+  - Notes: *In case of deprecations or API-incompatible changes, the reason why this change was made and the suggested alternative*
+
+*Insert entries in chronological order, with newer entries at the bottom*
--- a/Sources/SwiftLexicalLookup/Scopes/GenericParameterScopeSyntax.swift
+++ b/Sources/SwiftLexicalLookup/Scopes/GenericParameterScopeSyntax.swift
@@ -16,7 +16,7 @@ import SwiftSyntax
 /// futher lookup to its `WithGenericParametersScopeSyntax`
 /// parent scope's parent scope (i.e. on return, bypasses names
 /// introduced by its parent).
-protocol GenericParameterScopeSyntax: ScopeSyntax {}
+@_spi(Experimental) public protocol GenericParameterScopeSyntax: ScopeSyntax {}
 
 @_spi(Experimental) extension GenericParameterScopeSyntax {
   /// Returns names matching lookup and bypasses
--- a/Sources/SwiftRefactor/CMakeLists.txt
+++ b/Sources/SwiftRefactor/CMakeLists.txt
@@ -14,6 +14,7 @@ add_swift_syntax_library(SwiftRefactor
   ConvertComputedPropertyToZeroParameterFunction.swift
   ConvertStoredPropertyToComputed.swift
   ConvertZeroParameterFunctionToComputedProperty.swift
+  DeclModifierRemover.swift
   ExpandEditorPlaceholder.swift
   FormatRawStringLiteral.swift
   IntegerLiteralUtilities.swift
--- a/Sources/SwiftRefactor/ConvertStoredPropertyToComputed.swift
+++ b/Sources/SwiftRefactor/ConvertStoredPropertyToComputed.swift
@@ -22,6 +22,14 @@ public struct ConvertStoredPropertyToComputed: SyntaxRefactoringProvider {
       throw RefactoringNotApplicableError("unsupported variable declaration")
     }
 
+    var syntax = syntax
+
+    if let lazyKeyword = syntax.modifiers.first(where: { $0.name.tokenKind == .keyword(.lazy) }) {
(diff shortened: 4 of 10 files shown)


CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
```diff
@@ -2437,6 +2455,20 @@ public let DECL_NODES: [Node] = [
     kind: .typeInitializerClause,
     base: .syntax,
     nameForDiagnostics: nil,
+    documentation: """
+      A clause that specifies the underlying type of a type declaration.
```
ahoppen:
```suggestion
      A clause that specifies the underlying type of a `typealias` or `associatedtype` declaration.
```

CodeGeneration/Sources/SyntaxSupport/PatternNodes.swift
```diff
@@ -181,6 +181,28 @@ public let PATTERN_NODES: [Node] = [
...
+      ```
+
+      ### Examples
+
+      ```swift
+      let count: Int
+      ```
+
+      ```swift
+      func greet(name: String) {}
```
ahoppen: Function parameters actually don’t use `TypeAnnotationSyntax`, see https://swiftpackageindex.com/swiftlang/swift-syntax/main/documentation/swiftsyntax/functionparametersyntax

CodeGeneration/Sources/SyntaxSupport/StmtNodes.swift
```diff
@@ -493,6 +493,18 @@ public let STMT_NODES: [Node] = [
...
+
+         Written as:
+          ```swift
+            return
+          ```
+
+          ```swift
+            return <expr>
+          ```
+      """,
```
ahoppen: The indentation seems a little off to me here, both within the code blocks and overall.

Same below

CodeGeneration/Sources/SyntaxSupport/StmtNodes.swift
```diff
@@ -510,14 +522,24 @@ public let STMT_NODES: [Node] = [
...
     children: [
       Child(
         name: "throwKeyword",
         kind: .token(choices: [.keyword(.throw)])
       ),
       Child(
         name: "expression",
-        kind: .node(kind: .expr)
+        kind: .node(kind: .expr),
+        nameForDiagnostics: "error"
```
ahoppen: Did you intentionally change the name for diagnostics here?

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
```diff
@@ -2171,6 +2171,24 @@ public let DECL_NODES: [Node] = [
     kind: .returnClause,
     base: .syntax,
     nameForDiagnostics: nil,
+    documentation: """
+      A clause that specifies the return type of a function or closure.
```
rintaro: Actually function, subscript, type, or closure.
```swift
// function
func f() -> Int { }
// subscript
subscript(I: Int) -> Int {}
// type
let a: () -> Int = { 1 }
// closure
_ = { () -> Int in 1 }
```
But I'm not sure we should list everything here. This information is automatically rendered as "Contained in" section.
So maybe `A clause that specifies the return type` is enough.

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
```diff
@@ -2171,6 +2171,24 @@ public let DECL_NODES: [Node] = [
...
+
+      Written as:
+      ```swift
+      -> <type>
+      ```
```
rintaro: This technically a duplicated information with `### Children` section in the rendered doc-comments. Although this simplified code might be nice to read, my personal preference is not to have this.

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
```diff
@@ -2171,6 +2171,24 @@ public let DECL_NODES: [Node] = [
...
+
+      ### Examples
```
rintaro: I'm not really sure we should write a section (`###`) here. It's the same header level as other generated sections (i.e. `### Children` and `### Contained in`), If we want another section for "examples", we might need another property on `Node`.

But I _think_ something like this would be enough.
````
A clause that specifies the return type.

For example, the `-> Int` part in
```swift
func foo() throws -> Int { ... }
```
````

Padmashree06: Thank you for reviewing @rintaro . I will make the suggested change.

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
ahoppen: My proposal would be `A clause that specifies the return type, typically of a function or subscript`. That leaves room for the slightly more edge cases in the `Contained In` section, while still giving you an idea of what this is about.

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
ahoppen: We have a lot of precedence for `### Examples` sections in the documentation, just search for that string in `CodeGeneration`. So, I vote to keep it. 

I’m not opposed to moving the examples into a different property of `Node` but I think that should be a different PR.

CodeGeneration/Sources/SyntaxSupport/DeclNodes.swift
ahoppen: Fair point, I don’t have a strong opinion here either way

Padmashree06:
I have made the necessary changes please let me know if anything else needs to be modified.

Padmashree06:
The requested changes have been addressed, and the test suite passes without issues.

Padmashree06:
@ahoppen  @rintaro 
I have made the necessary changes. 
Could you please review it.
Thank you for your time.

ahoppen:
@swift-ci Please test

kimdv:
@swift-ci please test MacOS

Padmashree06:
Thank you for merging !

ahoppen:
Thanks for your contribution, @Padmashree06!

Padmashree06:
Thank you

